PR #2347, opened 2025-12-15
Feat: Membership restart flow improvement

### LINKED ISSUE #173: Improve membership restart flow to prevent accidental re-purchases
- [ ] Prevent customers from accidentally starting a new membership when they already have a canceled one that can be resumed

### Why

- Save customers confusion and frustration
- Reduce support load

Currently, a previous purchase notice only shows when logged in, but it's not clear that restarting the membership is an option. Restarting is preferable especially when the content is delivered through workflows so customers can pick up where they left off without waiting to catch up.

yashranaway:
# Fixes #173

## What

Introduces a new "Resume Subscription" flow for users who have previously cancelled a membership. Instead of silently creating a duplicate subscription, we now detect if a restartable subscription exists and prompt the user to either resume it or start fresh.

## Why

Currently, when a user with a cancelled membership re-subscribes, Gumroad creates a completely new subscription object. This leads to:
1.  **Duplicate Data**: Cleanliness issues in our database.
2.  **User Confusion**: Users lose their history and "since" dates.
3.  **Support Burden**: Users often want to resume their old pricing or state.

By catching this intent at the "Subscribe" button click, we can guide them to the existing subscription management page, offering a smoother experience.

### Backend (`Link` model)
- Added `restartable_subscription_for(user)` method.
- Logic: strict filtering for subscriptions that are `deactivated?` (cancelled/failed) but NOT `ended?` (completely expired) or `cancelled_by_seller?`.
- Explicitly excludes test subscriptions to prevent noise.
- **Excludes seller viewing their own product** - prevents modal from showing to product owners.
- **Returns most recent subscription** when user has multiple cancelled subscriptions (ordered by `deactivated_at`).
- **Optimized DB queries** - filtering moved to SQL level for better performance.

### Backend (`ProductProps` presenter)
- Added `token` parameter to `manage_url` for proper authentication when redirecting to subscription management.

### Frontend
- **Product Page**: Passes the `restartable_subscription` object (id, manage_url) to the React component tree.
- **CTA Button**: Intercepts the primary "Subscribe" action. If a restartable subscription exists, it halts the checkout flow and opens the new modal.
- **Resume Modal**: A new component showing two clear paths:
    1.  **Resume**: Redirects to the subscription management view (`/subscriptions/:id/manage`).
    2.  **Start New**: Proceeds to the standard checkout flow.

## UI Changes

https://github.com/user-attachments/assets/89265dac-42a8-4e2e-bcdb-1e21b9361fe6

## Testing

image

## Confirmation

I have watched all Antiwork Live streams.

## AI disclosure

Used Claude Opus 4.5 to refine the UI, tests & verify implementation


yashranaway:
@binary-koan , not sure who the right reviewer is for this PR. Could you please review it 👀 ?

binary-koan:
Someone will get to this eventually, we have a bit of a backlog at the moment 😓 We may look at this in a livestream, feels like a good candidate for it

slavingia:
> Someone will get to this eventually, we have a bit of a backlog at the moment 😓 We may look at this in a livestream, feels like a good candidate for it

What’s causing this? We should never let anything get this stale. It makes it hard to want to contribute.

app/javascript/components/Product/index.tsx
```diff
@@ -181,6 +181,11 @@ export type WishlistForProduct = Wishlist & {
   selections_in_wishlist: { variant_id: string | null; recurrence: string | null; rent: boolean; quantity: number }[];
 };
 
+export type RestartableSubscription = {
+  id: string;
+  manage_url: string;
+};
+
```
yashranaway, line 188: Only includes what's needed: id for tracking, manage_url (with token) for redirect.

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -127,17 +134,35 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
...
+
+    const handleStartNewSubscription = () => {
+      setShowResumeModal(false);
+      trackCtaClick({
+        sellerId: product.seller?.id,
+        name: product.name,
+        permalink: product.permalink,
+      });
+      window.open(checkoutUrlRef.current, "_top");
+    };
```
yashranaway, line 176: Handler when user explicitly chooses to start fresh instead of resuming, Using `window.open` with `_top `to navigate in same window (matches existing behavior)

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -127,17 +134,35 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
...
+        checkoutUrlRef.current = evt.currentTarget.href;
+        setShowResumeModal(true);
+        return;
+      }
+      trackCtaClick({
+        sellerId: product.seller?.id,
+        name: product.name,
+        permalink: product.permalink,
+      });
+    };
```
yashranaway, line 166: Show modal only for membership products (!purchase ensures they don't already own it)
Conditions: has restartable sub + is membership product + not already purchased

app/models/link.rb
```diff
@@ -844,6 +844,20 @@ def purchase_info_for_product_page(requested_user, browser_guid)
...
+    subscriptions
+      .where(user: requested_user)
+      .not_is_test_subscription
+      .where.not(deactivated_at: nil)
+      .where(ended_at: nil)
+      .where("(subscriptions.flags & ?) = 0", Subscription.flag_mapping["flags"][:cancelled_by_admin])
+      .order(deactivated_at: :desc)
+      .find { |subscription| subscription.alive_or_restartable? }
+  end
+
```
yashranaway, line 860: Method for detecting if a user has a restartable subscription. Returns the most recently deactivated subscription that can be resumed.
Edge cases handled:
  - Seller viewing own product (excluded to avoid confusion)
  - Multiple cancelled subscriptions (returns most recent via order)
  - Test subscriptions (excluded)
  - Seller-cancelled subscriptions (excluded via flags check)
  - Ended subscriptions (excluded)
  Using SQL-level filtering for efficiency instead of loading all into memory.

app/presenters/product_presenter/product_props.rb
```diff
@@ -172,4 +173,14 @@ def refund_policy_props
...
+
+    def restartable_subscription_props(user)
+      subscription = product.restartable_subscription_for(user)
+      return if subscription.nil?
+
+      {
+        id: subscription.external_id,
+        manage_url: manage_subscription_url(subscription.external_id, token: subscription.token, host: "#{PROTOCOL}://#{DOMAIN}"),
+      }
+    end
```
yashranaway, line 185: This matches existing pattern used in subscription management emails.

yashranaway:
@binary-koan Light Ping on this. 🙏

binary-koan:
Thanks for the reminder! We've noticed that the video only shows the "no, restart" case and doesn't show what happens when you click yes, so we'd want to see a video of that too.

However, when we talked about this we realised the issue needs more discussion - the right approach may actually be to restart the subscription automatically on checkout based on the email, since we can't think of a case where you'd want a new subscription instead. Sorry we jumped the gun on requesting PRs, we'll clarify the behaviour in the issue thread.

yashranaway:
> However, when we talked about this we realised the issue needs more discussion - the right approach may actually be to restart the subscription automatically on checkout based on the email, since we can't think of a case where you'd want a new subscription instead. Sorry we jumped the gun on requesting PRs, we'll clarify the behaviour in the issue thread.

Thanks for the clarification, I'll try to attempt raising a PR for that.

Issue #2042, opened 2025-11-17
Speed up Content page for product `idzmK`


vatsalkaushik:
Product/Content page for product ID `idzmK` keeps crashing, likely because of large number of files. User is unable to update the product currently.

<img width="1913" height="856" alt="Image" src="https://github.com/user-attachments/assets/f0ecc08a-f96e-43f3-9ea8-693e4b70bcef" />

ershad:
Note to OSS contributors: It would be nice to fix this. We can probably replicate the issue by adding many files to a product. If you have any specific questions, please post here, happy to help.

Devwill2:
Hi @ershad  How many files does product `idzmK` have?

internalthought:
I have found a performance bug related to this issue, maybe not directly related to the frontend crash but a reduction in backend query time and volume when the product has a lot of files and variants attached to it. While testing and attempting to reproduce this issue, I discovered it. I will link the PR after this comment. 

As far as trying to reproduce the issue, I seeded a product with 50 files and 20 variants, but was unable to reproduce a crash locally. Perhaps this issue is hard to reproduce in a dev env due to localized S3? More info on number of files, variants, and file sizes and types would be helpful information. I will keep trying to reproduce this issue and see if I can get the crash seen in screenshot.

binary-koan:
The merged PR is a huge improvement (thanks @4arjun!) - the page now loads in a reasonably short time without crashing, at least for me.

Clicking any file (e.g. selecting or expanding/collapsing files) is still super laggy, so would be great if we can reduce the number of rendering ops there. Memoizing a bunch of stuff and aggressively trying to stop file nodes re-rendering when not necessary might be enough; fingers crossed that's the only other thing we need in order to close this off.

brentguf:
Hey @binary-koan, 

Before proceeding further, I think it’s important to define the completion criteria of this issue, since performance optimization can go far.

That said, here are the next steps I would take if the goal is to optimize the Content tab:
1. Extract and lazy-load modals. Similar to PR #2681. This should improve performance and make `ContentTabContent` more focused, which will make subsequent steps easier. Beneficial regardless of any context refactoring. Low-risk if done incrementally, even without tests.
2. Research/draft context/state solution. The solution could involve splitting context on multiple levels (e.g. static and dynamic parts, splitting up product into different contexts as well).
3. (Optional, but recommended) Add frontend tests: refactoring context and data manipulation is riskier and more time-consuming without tests.
[...]

PR #2348, opened 2025-12-15
(fix) Product content edit: only fetch Dropbox files when pending files exist

brentguf:
# Problem

While taking a look at https://github.com/antiwork/gumroad/issues/2042, I discovered that the product content editor polls for Dropbox file status updates every 10 seconds unconditionally (so even when there are no Dropbox files being uploaded). 

Demo with an accelerated interval (1s vs. 10s):

https://github.com/user-attachments/assets/a7736eb5-8573-41ca-98e2-887e84d9dd26

This causes:
- Unnecessary API requests/server usage every 10 seconds for the lifetime of the component
- Full re-renders of the content tab every 10 seconds

# Solution
Add a conditional check that only runs the polling interval when there are actually Dropbox files in the `in_progress` state. 

# AI Disclosure
Used Claude Sonnet 4.5 to review the solution.


app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -409,13 +409,20 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
...
+  );
   React.useEffect(() => {
+    if (!hasPendingDropboxFiles) return;
+
     const interval = setInterval(
       () => void fetchDropboxFiles(uniquePermalink).then(({ dropbox_files }) => addDropboxFiles(dropbox_files)),
       10000,
     );
     return () => clearInterval(interval);
-  }, [editor]);
```
brentguf, line 421: This dependency seemed arbitrary. Could be worth looking into adding an ESLint rule for this (`react-hooks/exhaustive-deps`).

binary-koan: There are a ton of places we deliberately don't use exhaustive deps; I wouldn't mind adding it and forcing us to document them, but it would likely be quite time-consuming

binary-koan:
(Failing tests are flaky / new year issues)

PR #2333, opened 2025-12-15
fix: Desktop Analytics view missing sales due to DST issue

yashranaway:
## What

**Before:**
The Analytics view shows fewer sales than the Sales page. The `timezone_formatted_offset` method always returned Standard Time offset (e.g., `-08:00` for Pacific Time), ignoring Daylight Saving Time.

**Now:**
The `timezone_formatted_offset` method is DST-aware, returning the correct offset based on the current time (e.g., `-07:00` during PDT summer, `-08:00` during PST winter).

## Root Cause
The original `User#timezone_formatted_offset` used `ActiveSupport::TimeZone.new(timezone_id).formatted_offset` which explicitly ignores DST. This caused Elasticsearch queries in analytics services to use incorrect timezone offsets during DST periods, leading to sales being misattributed to wrong dates.

## Why

* Fixes discrepancy between Sales page and Analytics view during DST periods
* Provides consistent behavior across all analytics services
* Adds optional `at:` parameter for querying offset at specific dates

## Tests
image

## AI disclosure

Used Claude Opus 4.5 to understand the shifting of times etc and test writing 
All code reviewed by me

## Confirmation

I have watched all Antiwork Live streams

PR #2353, opened 2025-12-15
 fix: Desktop Analytics view missing sales due to DST issue

### LINKED ISSUE #1811: Desktop Analytics view is missing sales
The Sales page shows sales that aren't appearing on the Analytics view. 

For example, my sales view shows that I've had 8 sales today. 

the analytics view for today shows less sales.   

I know there was a similar issue with the mobile app at one point, where the Analytics view and Home view didn't agree because of a time range difference in how the sale count was computed.   

Way back in March 2024, i reported this and this was the diagnosis:
```
Hey Chuck,

Turns out this is a Daylight Saving Time issue where the Home tab respects it while the Analytics tab doesn't. 
We will look into fixing this in the future. 
Thanks for the report!

Best,
Vatsal
```
Perhaps its the same issue here?

yashranaway:
## fixes #1811

previous attempt #2333

## What

**Before:**
The Analytics view shows fewer sales than the Sales page during Daylight Saving Time periods. The timezone offset used in Elasticsearch queries was static (e.g., `-08:00` for PST), ignoring DST.

**Now:**
Analytics services use IANA timezone IDs (e.g., `America/Los_Angeles`) instead of fixed offsets. Elasticsearch automatically handles DST for all dates in the query, including queries that span DST transitions.

## Root Cause
The original code used `timezone_formatted_offset` which returns a formatted UTC offset. This offset was static and didn't change based on DST. For example:
- Sale at July 15, 00:30 PDT = July 15, 07:30 UTC
- With `-08:00` (PST): ES calculates 07:30 - 8h = **July 14**, 23:30 
- With IANA ID: ES knows July is PDT (-07:00), calculates 07:30 - 7h = **July 15**, 00:30 ✓

## Solution
Use `user.timezone_id` (which returns the IANA timezone ID like `"America/Los_Angeles"`) directly in Elasticsearch queries. ES automatically applies the correct offset for each date in the query.

## Tests Added

image

## Edge Cases that are Handled
1. Query during summer (DST active)
2. Query during winter (standard time)
3. Sale at midnight during DST (would fail with fixed offset)
4. Query spanning DST transition (March 8-10)
5. Query spanning DST end (November 1-5)

## AI disclosure
Claude was used for research and code generation.

## Confirmation

I have watched all Antiwork Live streams

### DIFF
--- a/spec/services/creator_analytics/sales_spec.rb
+++ b/spec/services/creator_analytics/sales_spec.rb
@@ -209,7 +209,6 @@ describe CreatorAnalytics::Sales, "DST handling" do
 
       result = service.by_product_and_date
 
-
       expect(result[[product.id, "2025-03-08"]]).to be_present
       expect(result[[product.id, "2025-03-10"]]).to be_present
     end


app/services/creator_analytics/sales.rb
```diff
@@ -10,15 +10,16 @@ def initialize(user:, products:, dates:)
     @user = user
     @products = products
     @dates = dates
+    @timezone = @user.timezone_id
```
yashranaway: we now use the IANA timezone ID (e.g. "America/Los_Angeles") instead of a fixed UTC offset.

This allows Elasticsearch to automatically handle Daylight Saving Time (DST) correctly for all dates in the query, including queries that span across DST transitions.

spec/services/creator_analytics/sales_spec.rb
```diff
@@ -165,4 +165,54 @@
...
+
+  describe "DST handling" do
+    context "when sale occurs near midnight during DST" do
+      let(:user_timezone) { "Pacific Time (US & Canada)" }
+
+      it "correctly attributes sale to the right day" do
+        user = create(:user, timezone: user_timezone)
+        product = create(:product, user: user)
+
+        create(:purchase, link: product, created_at: Time.utc(2025, 7, 15, 7, 30))
```
yashranaway: This test case demonstrates the fix.

A sale at 07:30 UTC is 00:30 PDT (summer) but would be 23:30 PST (previous day) if we used the standard -08:00 offset.

With the IANA timezone ID, Elasticsearch correctly identifying this as July 15th

spec/services/creator_analytics/sales_spec.rb
```diff
@@ -165,4 +165,54 @@
...
+
+    context "when query spans DST transition" do
+      let(:user_timezone) { "Pacific Time (US & Canada)" }
+
+      it "correctly buckets sales across DST boundary" do
+        user = create(:user, timezone: user_timezone)
+        product = create(:product, user: user)
+
+        create(:purchase, link: product, created_at: Time.utc(2025, 3, 9, 7, 0))
+        create(:purchase, link: product, created_at: Time.utc(2025, 3, 10, 8, 0))
```
yashranaway: this test verifies that we handle queries spanning the actual DST transition (March 9th).

The first sale is during PST (-08:00) and the second is during PDT (-07:00).  Using 
[timezone_id](https://github.com/antiwork/gumroad/pull/2353/changes#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR806-R808) ensures both are bucketed correctly.

yashranaway:
Hi @binary-koan _a
Updated the approach to use IANA timezone IDs (e.g. America/Los_Angeles) directly in ES queries, which handles DST transitions automatically.

Added specs verifying that sales near DST boundaries and midnight are now correctly attributed. Ready for review!

binary-koan:
Nice, at a glance this approach looks good. Will see if I can reproduce the original issue with ES and Ruby DST not matching when I find time (i.e. ES thinks it's non-DST but Ruby thinks it's DST) to see if we need to fix anything else, and review thoroughly after that!

yashranaway:
@binary-koan _a Hey, I have reverted the `timezone_formatted_offset` changes and added DST specs for following and page views. All 6 tests pass : verifying correct date attribution near midnight during DST and across DST transition 🤞

yashranaway:
Sidenote: I kept a few essential inline notes to clarify the transition logic, but let me know if you'd prefer them removed!

app/services/creator_analytics/product_page_views.rb
```diff
@@ -5,18 +5,19 @@ def initialize(user:, products:, dates:)
     @user = user
     @products = products
     @dates = dates
+    @timezone = @user.timezone_id
```
binary-koan: I think we can just inline this, no need for an extra ivar

app/services/creator_analytics/sales.rb
binary-koan: Same here, we can inline

spec/models/user_spec.rb
```diff
@@ -2582,6 +2582,14 @@ def sample_string_of_length(n)
     end
   end
 
+  describe "#timezone_id" do
```
binary-koan: We already have a spec for this just above?

yashranaway:
thanks for the review , minor nits will change right away! 🫡

PR #2345, opened 2025-12-15
Inertia Migration - Login & Signup 

stefanbinoj:
Part of #856 

This PR migrate following pages to inertia 
- /login 
- /signup
Also migrates all related controller actions

## Before 

https://github.com/user-attachments/assets/862cc398-0f8e-440b-89d4-db103f695f68

 
## After 
(Full page reload happens while navigating from landing page -> login page because we haven't migrated navbar for landing page :) , also would be better to make it a seperate PR for modular changes)

https://github.com/user-attachments/assets/8f6f2514-f763-4180-a6ff-446cd3f3f056

## Tests

```bash
bundle exec rspec spec/controllers/logins_controller_spec.rb
```

Screenshot 2025-12-15 at 8 20 06 PM

```bash
bundle exec rspec spec/controllers/signup_controller_spec.rb
```

Screenshot 2025-12-15 at 8 27 16 PM

**NOTE :** Failing tests are due to missing env for stripe payment processing 

### AI disclosure
Used Claude Opus 4.5 via Opencode for initial drafting 

### Confirmation
I have watched all Antiwork Livestreams

Issue #856, opened 2025-08-14
Convert Gumroad to single page app powered fully by React


slavingia:
Make the Gumroad creator dashboard an SPA, instead of rendering each page at a different URL and requiring a full page load.

Bounty: $1K per task (will gradually increase it over time to finish the smaller ones first)

- [ ] New Product (`/links/new`)
- [ ] Edit Product (`/links/:id/edit`)
- [ ] Bundle Edit (`/bundles/:id`)
- [x] Affiliated Products (`/products/affiliated`)
- [x] Archived Products (`/products/archived`)
- [x] Collab Products (`/products/collabs`)
- [ ] Followers (`/followers`)
- [x] Workflows (`/workflows`) - https://github.com/antiwork/gumroad/pull/2005 by @Pradumn27 
  - [x] Workflows new (`/workflows/new`)
  - [x] Workflows edit (`workflows/:id/edit`)
- [ ] Emails (`/emails`)
- [ ] Emails (`/emails/published`)
- [ ] Emails (`/emails/scheduled`)
- [ ] Emails (`/emails/drafts`)
- [ ] Main Settings (`/settings`)
- [ ] Profile Settings (`/settings/profile`)
- [ ] Payment Settings (`/settings/payments`)
- [ ] Password Settings (`/settings/password`)
- [ ] Team Settings (`/settings/team`)
- [ ] Advanced Settings (`/settings/advanced`)
- [ ] Third Party Analytics Settings (`/settings/third_party_analytics`)
- [ ] UTM Links (`/dashboard/utm_links`)
- [ ] Following (`/dashboard/audience`)
- [ ] Payouts (`/payouts`)
- [ ] Support (`/help`)
- [ ] Library (`/library`)
- [ ] Reviews (`/reviews`)
- [ ] Wishlists (`/wishlists`)
- [ ] Following Wishlists (`/wishlists/following`)
- [ ] Affiliates (`/affiliates`)
- [ ] Affiliates Onboarding (`/affiliates/onboarding`)

### Why

This will save a lot of people time when navigating around Gumroad.

<img width="577" height="245" alt="Image" src="https://github.com/user-attachments/assets/9e4abd3a-5bc9-40ae-b67c-2470c4b93a98" />

praveenkalady:
@slavingia I’m interested in working on this. Can I work on this issue?

slavingia:
Feel free, we don't assign issues – just note this is a large, difficult task that mostly senior and staff engineers can solve.

[...]

PR #2366, opened 2025-12-16
Inertia Migration - Auth (Part 1)

stefanbinoj:
Part of #856 
Previous PR #2345 

This PR migrate following pages to inertia 
- /login 
- /signup
Also migrates all related controller actions

## Before 

https://github.com/user-attachments/assets/862cc398-0f8e-440b-89d4-db103f695f68

 
## After 
(Full page reload happens while navigating from landing page -> login page because we haven't migrated navbar for landing page :) , also would be better to make it a seperate PR for modular changes)

https://github.com/user-attachments/assets/39254910-f1fe-40a3-997d-9ce64c30a5c0

## Tests

```bash
bundle exec rspec spec/controllers/logins_controller_spec.rb
```

Screenshot 2025-12-15 at 8 20 06 PM

```bash
bundle exec rspec spec/controllers/signup_controller_spec.rb
```

Screenshot 2025-12-15 at 8 27 16 PM

**NOTE :** Failing tests are due to missing env for stripe payment processing 

### AI disclosure
Used Claude Opus 4.5 via Opencode for initial drafting 

### Confirmation

### DIFF
--- a/app/javascript/data/login.ts
+++ b/app/javascript/data/login.ts
@@ -2,35 +2,6 @@ import { cast } from "ts-safe-cast";
 
 import { ResponseError, request } from "$app/utils/request";
 
-export type LoginPayload = {
-  email: string;
-  password: string;
-  recaptchaResponse: string | null;
-  next: string | null;
-};
-
-export const login = async (data: LoginPayload): Promise<{ redirectLocation: string }> => {
-  const response = await request({
-    method: "POST",
-    url: Routes.login_path(),
-    accept: "json",
-    data: {
-      user: {
-        login_identifier: data.email,
-        password: data.password,
-      },
-      next: data.next,
-      "g-recaptcha-response": data.recaptchaResponse,
-    },
-  });
-  if (!response.ok) {
-    const { error_message } = cast<{ error_message: string }>(await response.json());
-    throw new ResponseError(error_message);
-  }
-  const { redirect_location } = cast<{ redirect_location: string }>(await response.json());
-  return { redirectLocation: redirect_location };
-};
-
 export const renewPassword = async (email: string) => {
   const response = await request({
     method: "POST",


app/controllers/concerns/inertia_rendering.rb
```diff
@@ -7,12 +7,14 @@ module InertiaRendering
...
       RenderingExtension.custom_context(view_context).merge(
-        current_user: current_user_props(current_user, impersonated_user),
         authenticity_token: form_authenticity_token,
         flash: inertia_flash_props,
         title: @title
       )
     end
+    inertia_share if: :user_signed_in? do
+      { current_user: current_user_props(current_user, impersonated_user) }
+    end
```
stefanbinoj, line 17: Used conditional sharing when invoking the method this is required because in `/login` and `/signup` page `current_user` is not present

app/controllers/logins_controller.rb
```diff
@@ -1,24 +1,26 @@
 # frozen_string_literal: true
 
 class LoginsController < Devise::SessionsController
-  include OauthApplicationConfig, ValidateRecaptcha
+  include OauthApplicationConfig, ValidateRecaptcha, InertiaRendering
 
   skip_before_action :check_suspended, only: %i[create destroy]
-  before_action :block_json_request, only: :new
```
stefanbinoj, line 7: **Client-side Inertia navigation** (clicking links within the SPA) sends requests with `X-Inertia: true` header expecting JSON — but this line BLOCKS this callback.
 - Removing this line allows Inertia client-side navigation to work. Keeping it blocks SPA navigation but not initial page loads.

app/controllers/logins_controller.rb
```diff
@@ -1,24 +1,26 @@
...
   after_action :clear_dashboard_preference, only: :destroy
   before_action :reset_impersonated_user, only: :destroy
   before_action :set_noindex_header, only: :new, if: -> { params[:next]&.start_with?("/oauth/authorize") }
 
+  layout "inertia", only: [:new]
+
   def new
-    @hide_layouts = true
```
stefanbinoj, line 13: Since we are using disableLayout when rendering the page

app/controllers/logins_controller.rb
```diff
@@ -44,16 +46,12 @@ def create
         flash[:warning] = "Your password has previously appeared in a data breach as per haveibeenpwned.com and should never be used. We strongly recommend you change your password everywhere you have used it."
       end
 
-      render json: { redirect_location: login_path_for(@user) }
+      redirect_to login_path_for(@user), allow_other_host: true
```
stefanbinoj, line 49: Now we are using `redirect_to` instead of `inertia_location`

app/controllers/logins_controller.rb
```diff
@@ -44,16 +46,12 @@ def create
...
 
   private
-    def respond_with_login_failure(message)
-      render json: { error_message: message }, status: :unprocessable_entity
-    end
-
-    def block_json_request
-      render json: {}, success: false, status: :bad_request if request.format.json?
+    def redirect_with_login_error(message)
+      redirect_to login_path, warning: message, status: :see_other
```
stefanbinoj, line 55: `warning` is given specifically inorder to show and identify errors that are inlined paragraph elements rather than toast as usual 

Screenshot 2025-12-15 at 8 34 30 PM

app/javascript/packs/inertia.js
```diff
@@ -20,12 +20,18 @@ async function resolvePageComponent(name) {
   try {
     const module = await import(`../pages/${name}.tsx`);
     const page = module.default;
+    if (page.disableLayout) {
+      return page;
+    }
```
stefanbinoj, line 25: Since `login` and `signup` page doesn't require default layout (includes sidebar, navbar) it is better to early return. 
Now we don't use `inertia/layout`

app/javascript/components/useFlashError.tsx
```diff
@@ -0,0 +1,22 @@
...
+  if (flash?.status === "warning" && flash?.message) {
+    return (
+      <div role="alert" className="danger">
+        {flash.message}
+      </div>
+    );
+  }
+
+  return null;
+}
```
stefanbinoj, line 22: Created a new component just for showing the inlined <p> element error

app/javascript/inertia/layout.tsx
```diff
@@ -24,14 +40,16 @@ export default function Layout({ children }: { children: React.ReactNode }) {
...
-      <Head title={title} />
-      <Alert initial={flash ?? null} />
-      <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
-        <Nav title="Dashboard" />
-        {isRouteLoading ? <LoadingSkeleton /> : null}
-        <main className={classNames("flex-1 overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>
-      </div>
-    </>
+    <LoggedInUserProvider value={parseLoggedInUser(logged_in_user)}>
+      <CurrentSellerProvider value={parseCurrentSeller(current_seller)}>
```
stefanbinoj, line 44: `current_seller` also seems to be reactive to `\login` and `\
signup`  thus would be better if we moved it over here since we could extract props without having a manually reduntant useEffect in `app_wrapper.tsx`

app/javascript/components/useFlashError.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { type AlertPayload } from "$app/components/server-components/Alert";
+
+type PageProps = {
+  flash?: AlertPayload;
+};
+
+export function useFlashError(): React.ReactNode {
+  const { flash } = usePage<PageProps>().props;
+
+  if (flash?.status === "warning" && flash?.message) {
```
stefanbinoj, line 13: only extracting if `flash.status` is warning: **since this needs to be shown inline**.  
-- flash can be also `success`  (in that case it shows regular toast like success flash message) 

Screenshot 2025-12-15 at 8 40 05 PM

This toast is shown by `inertia/layout` from the action of logout button

app/javascript/pages/Signup/New.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  return (
+    <Layout header={<h1>{headerText}</h1>} headerActions={<Link href={Routes.login_path({ next })}>Log in</Link>}>
+      <form onSubmit={(e) => void handleSubmit(e)}>
+        <SocialAuth />
+        <Separator>
+          <span>or</span>
+        </Separator>
+        <section>
+          {flashError}
```
stefanbinoj, line 79: using useFlashErrorComponent

spec/controllers/two_factor_authentication_controller_spec.rb
```diff
@@ -177,6 +177,7 @@
         let(:user) { @user }
         let(:call_action) { post :create, params: { token: @user.otp_code, user_id: @user.encrypted_external_id }, format: :json }
         let(:expected_redirect_location) { controller.send(:login_path_for, @user) }
+        let(:expects_json_response) { true }
```
stefanbinoj, line 180: The `merge guest cart with user cart` shared example was updated because logins_controller changed from returning JSON to using redirect_to. Since the shared example is also used by `two_factor_authentication_controller_spec.rb` (which still expects JSON), I added the expects_json_response flag to support both cases - JSON for two_factor_auth, redirect for logins.

stefanbinoj:
Addressing some previous comments : 

1. https://github.com/antiwork/gumroad/pull/2345/files#r2623198588
> You have to duplicate this in the Signup page as well. You should instead use a distinct, much simpler layout in charge for the Login and Signup pages.

IMO having a seperate layout for login/signup just for showing flash messages wont be a good idea , also passing a **inlined flash error message **(para element)** to children from layout is not a ideal approach thus created new custom `useFlashError.tsx` component that listens for flash message from usePageHook

2. On a side note pls note that while navigating from inertia page -> non-inertia page via `redirect_to` may have some weird edge cases as in video 

https://github.com/user-attachments/assets/fb09da39-a592-4763-9bd5-47e4efcd0095

NOTE : this won't be a problem once all the pages are migrated to inertia 


Hey @EmCousin _a, could you please take a look at this when you have some time?

app/javascript/data/login.ts
```diff
@@ -2,35 +2,6 @@ import { cast } from "ts-safe-cast";
...
-    },
-  });
-  if (!response.ok) {
-    const { error_message } = cast<{ error_message: string }>(await response.json());
-    throw new ResponseError(error_message);
-  }
-  const { redirect_location } = cast<{ redirect_location: string }>(await response.json());
-  return { redirectLocation: redirect_location };
-};
-
```
stefanbinoj, line 33: removed unused json api request helper used in old `server-components/Login.tx`

app/controllers/logins_controller.rb
EmCousin: It would be safer to make an exception by allowing the request to pass through if it contains the proper Inertia headers instead of removing this check entirely.

app/controllers/logins_controller.rb
EmCousin: We have a [temporary interceptor](https://github.com/antiwork/gumroad/blob/main/app/javascript/packs/admin_inertia.ts#L50-L103) to trigger full page reloads instead of Inertia visits for URLs that haven't migrated yet. We could make a similar, simpler interceptor for this case since this flow is critical

app/javascript/components/useFlashError.tsx
EmCousin: I like this approach, but it is confusing to define this function as a hook and make it return a React component. You could just define it as a component and render it directly
```jsx
export const FlashError = () => { ... }

<FlashError />
```

Issue #2373, opened 2025-12-17
Make Archived Products Unpublished and Removed from Indexing


michellelarney:
When a product is archived, it should no longer be publicly accessible or discoverable. Currently, archiving does not fully prevent visibility in all contexts.

Archiving a product should:

- Automatically unpublish the product

- Prevent it from being indexed by search engines or internal discovery

- Make sure it's listed under the Archived tab, not under the default "All products" tab

This ensures archived products are truly inactive and avoids confusion for sellers, customers and prevents unintended access.

PR #2374, opened 2025-12-17
fix: Make Archived Products Unpublished and Removed from Indexing

romitg2:
fixes: #2373 

### Issue: 
- Archived products were still publicly accessible via direct URL because archiving only set the archived flag without unpublishing. 

### Solution: 
- unpublish product when archiving. 

# Before:
https://github.com/user-attachments/assets/204e3aa1-3e72-4607-bce0-3ef6e0668b69

# After:
https://github.com/user-attachments/assets/bcb4a780-ba64-41d7-9b3b-381d0a250551

## Tests Result: (local run)
Screenshot 2025-12-17 at 8 27 33 AM

Image

### AI Disclosure:
- watched claude-code opus 4.5 to understand codebase

### Live Stream Disclosure:
- watched all live streams

### DIFF
--- a/spec/controllers/products/archived_controller_spec.rb
+++ b/spec/controllers/products/archived_controller_spec.rb
@@ -203,9 +203,8 @@ describe Products::ArchivedController, inertia: true do
     end
 
     it "does not change purchase_disabled_at on an already unpublished product" do
-      original_disabled_at = 1.week.ago
+      original_disabled_at = 1.week.ago.floor
       membership.update!(purchase_disabled_at: original_disabled_at)
-      original_disabled_at = membership.reload.purchase_disabled_at
 
       post :create, params: { id: membership.unique_permalink }, as: :json
 


spec/requests/products/index_spec.rb
```diff
@@ -659,8 +659,9 @@ def find_product_row(product, hover: false)
       expect(page).to have_content(membership.name)
     end
 
-    it "archives a product" do
```
romitg2, line 662: - updated the tests:
Image

spec/controllers/products/archived_controller_spec.rb
```diff
@@ -190,12 +190,16 @@
       let(:request_format) { :json }
     end
 
-    it "archives the product" do
```
romitg2, line 193: - updated the test
Screenshot 2025-12-17 at 8 27 33 AM

app/controllers/products/archived_controller.rb
```diff
@@ -71,7 +71,11 @@ def memberships_paged
   def create
     authorize [:products, :archived, @product]
 
-    @product.update!(archived: true)
+    @product.update!(
+      archived: true,
+      purchase_disabled_at: @product.purchase_disabled_at || Time.current
```
romitg2, line 76: unpublishes the product, which makes also remove it from indexing. 
- we consider the product is unpublished if `purchase_disabled_at` is not `nil`

romitg2:
@binary-koan could you please review this? Thanks!

binary-koan:
Could we make the flash message "archived and unpublished" if the product was previously published?

romitg2:
> Could we make the flash message "archived and unpublished" if the product was previously published?

@binary-koan _a addressed comment, could you please review again? Thanks!


https://github.com/user-attachments/assets/f6cf7937-47c2-40b9-aeff-bb55945bea0a

spec/requests/products/index_spec.rb
```diff
@@ -673,12 +674,36 @@ def find_product_row(product, hover: false)
...
+      product = create(:product, :unpublished, user: seller)
+
+      visit(products_path)
+
+      within find_product_row product do
+        select_disclosure "Open product action menu" do
+          click_on "Archive"
+        end
+      end
+      wait_for_ajax
```
binary-koan: Do we need this? I feel like `have_alert` should be enough to make Capybara wait

romitg2: i thought just good to have them. removed as of now, if test flakes will add them back.

romitg2:
> Could you add a controller spec that it doesn't change the purchase_disabled_at on an already unpublished product? Apart from that (and one small comment) looks good!

@binary-koan _a added controller test and addressed comment, could you please review again? 

Thanks!

romitg2:
@binary-koan all related tests are passing! [failing one](https://github.com/antiwork/gumroad/actions/runs/20471191516/job/58827007325?pr=2374#step:8:278) is flaky and unrelated.

Issue #2070, opened 2025-11-18
Use external_id instead of id for admin routes


binary-koan:
## What

- Update all URLs and JSON formats in admin to use the `external_id` of records rather than the `id`
- Support the numeric ID in URLs and redirect to the `external_id` version

## Why

- We should consistently return external IDs for all frontend code to make it clearer for contributors and AI
- Easier to share admin links in public issue descriptions etc

### Subtasks
- [x] Users https://github.com/antiwork/gumroad/pull/2375
- [x] Products https://github.com/antiwork/gumroad/pull/2555
- [x] Purchases https://github.com/antiwork/gumroad/pull/2332
- [x] Merchant Accounts https://github.com/antiwork/gumroad/pull/2361
- [x] Affiliates https://github.com/antiwork/gumroad/pull/2375
- [x] Payouts https://github.com/antiwork/gumroad/pull/2467

binary-koan:
@slavingia @EmCousin FYI - worth a separate issue to #2060 IMO

EmCousin:
@nyomanjyotisa PR merged. You can get started now!

nyomanjyotisa:
The PR is ready for review https://github.com/antiwork/gumroad/pull/2147

nyomanjyotisa:
### Subtasks
- [ ] Users https://github.com/antiwork/gumroad/pull/2375
- [ ] Products 
- [x] Purchases https://github.com/antiwork/gumroad/pull/2332
- [x] Merchant Accounts https://github.com/antiwork/gumroad/pull/2361
- [ ] Affiliates https://github.com/antiwork/gumroad/pull/2375
- [ ] Payouts

PR #2375, opened 2025-12-17
refactor(users and affiliates): use external_id instead of id

nyomanjyotisa:
Part of https://github.com/antiwork/gumroad/issues/2070

# Summary
Migrated **Users** and **Affiliates** admin routes to use external_id instead of numeric database IDs for better security, consistency, and public shareability. Migrated both in a single PR since they share common concerns.

# After
### Redirect id to external id

https://github.com/user-attachments/assets/00c3f516-92b2-4f37-8e51-38c549c7dde8

### Users paginations

https://github.com/user-attachments/assets/d29d557f-a519-4328-9ece-1731c1d577cb

### No unintended redirect when using a username that starts witha  number

https://github.com/user-attachments/assets/15e3adce-d8eb-4147-9282-c8377cd9cac5

### Test all actions on the user show page

https://github.com/user-attachments/assets/83de42d6-42c0-44ad-adf4-cc2c5a9aed97

### Search by external id

https://github.com/user-attachments/assets/bc604229-c8a8-4a60-bb6c-bbd2b14be726

### Redirect id to external id

https://github.com/user-attachments/assets/77af47cc-f4dd-4911-abde-d592804241ed

### No unintended redirect when using a username that starts with a number

https://github.com/user-attachments/assets/70fef192-fb55-4c18-990a-eaeaf4bde5d8

### Search by external id

https://github.com/user-attachments/assets/032fb572-6d47-4330-990f-c996b7a1e782

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1218,9 +1218,10 @@ class User < ApplicationRecord
       value.presence && Mail::Address.new(value).domain
     end
 
+    # Checks if a value is purely numeric (returns true for both database IDs and external_ids).
+    # Used in redirect logic after external_id lookup fails, to distinguish numeric identifiers
+    # from usernames that start with numbers (e.g., "1jyo" should not redirect to user with id=1).
     def self.id?(value)
-      return false if value.blank?
-
-      value.to_s == value.to_i.to_s
+      value.present? && value.to_s == value.to_i.to_s
     end
 end


app/models/user.rb
```diff
@@ -1213,4 +1213,10 @@ def reset_avatar_changed
     def to_email_domain(value)
       value.presence && Mail::Address.new(value).domain
     end
+
+    def self.id?(value)
+      raise ArgumentError, "value can't be blank" if value.blank?
+
+      value.to_i.to_s == value
```
nyomanjyotisa: The User model does not use the `ExternalId` module, and both id and external_id are numeric (e.g. id = 10, external_id = 7269625173515)

Created this method on the User model to determine whether a given value/parameter should be treated as an id

app/controllers/admin/affiliates_controller.rb
```diff
@@ -33,9 +33,12 @@ def show
...
   private
     def fetch_affiliate
-      @affiliate_user = User.find_by(username: params[:id])
-      @affiliate_user ||= User.find_by(id: params[:id])
-      @affiliate_user ||= User.find_by_external_id(params[:id].gsub(/^ext-/, ""))
+      @affiliate_user = User.find_by(username: params[:external_id])
+      @affiliate_user ||= User.find_by_external_id(params[:external_id])
+
+      if @affiliate_user.nil? && User.id?(params[:external_id]) && (user = User.find_by(id: params[:external_id]))
+        return redirect_to admin_affiliate_path(user.external_id)
```
nyomanjyotisa, line 40: Used `.id?` here to prevent unintended redirection when accessing a username that starts with a number (e.g. 1jyo).
Without `User.id?(params[:external_id])`, visiting `/1jyo` would incorrectly redirect to the user with id = 1.

Also added an `@affiliate_user.nil?` check as a safety measure, in case a user’s id and external_id could overlap in the future

app/controllers/concerns/admin/fetch_user.rb
```diff
@@ -7,15 +7,19 @@ def fetch_user
         User.find_by(email: user_param)
       else
         User.where(username: user_param)
-            .or(User.where(id: user_param))
             .or(User.where(external_id: user_param))
             .first
       end
 
+      if @user.nil? && User.id?(user_param) && (user = User.find_by(id: user_param))
```
nyomanjyotisa, line 14: Same as above

nyomanjyotisa:
Hi @EmCousin Could you help to review this? Failing test unrelated

app/models/user.rb
EmCousin: If I'm not mistaken, there are a couple of issues here:

```ruby
# Should return true, but returns false because the argument is an Integer
# We can't just assume the compared value is always a string
User.id?(1)
=> false 

# Should return false because it's an external id, but returns true
User.id?("7269625173515")
=> true 
```

app/controllers/admin/affiliates_controller.rb
EmCousin: We probably need to cover this with extra tests to ensure no regression is happening

nyomanjyotisa: @EmCousin Added extra tests for these

Only `if user = User.find_by(id: params[:external_id])`
image

Full `if @affiliate_user.nil? && User.id?(params[:external_id]) && (user = User.find_by(id: params[:external_id]))`
image


Test command `bundle exec rspec spec/controllers/admin/affiliates_controller_spec.rb:88 spec/controllers/admin/affiliates_controller_spec.rb:113 spec/controllers/admin/users_controller_spec.rb:70 spec/controllers/admin/users_controller_spec.rb:94`

app/models/user.rb
nyomanjyotisa: @EmCousin

`.id?` is intended to return true for both database IDs and external IDs

In usage, this is always combined with `@user.nil?`, so the redirect checks is skipped when a user already exists. We attempt `User.find_by_external_id(params[:external_id])` before this check, which means valid external IDs are already resolved and won’t be affected by the redirect logic.

So the flow is:
- Try to resolve by external_id
- If no user is found and the value looks like an ID, handle it as an ID

And updated the checks to `value.to_s == value.to_i.to_s` so it return true for `User.id?(1)`

And added some tests for the `id?` method [here](https://github.com/antiwork/gumroad/pull/2375/commits/a3cbd8156fe663dd39e77f35c500834019112326#diff-bb8bcb6ef2e13ab9491f4f9ec48be47ee025d7f0752a25029d40810af2c65ee1)

app/models/user.rb
```diff
@@ -1217,4 +1217,10 @@ def reset_avatar_changed
     def to_email_domain(value)
       value.presence && Mail::Address.new(value).domain
     end
+
+    def self.id?(value)
+      return false if value.blank?
+
+      value.to_s == value.to_i.to_s
```
EmCousin:
```suggestion
      value.present? && value.to_s == value.to_i.to_s
```

EmCousin: This method is worth a comment in my opinion. It's hard to understand why we are doing this when looking solely at the code.

PR #2392, opened 2025-12-18
Fix: Prevent $0 plan changes when tier prices are disabled/deleted

vishaltelangre:
### What

Adds validation to prevent creating $0 subscription plan changes when a tier price is deleted for a recurrence that has active subscriptions.

Previously, when a seller deleted a pricing tier (e.g. monthly price), the `ScheduleMembershipPriceUpdatesJob` would calculate `new_price` as `0` and create invalid plan changes. When these plan changes were applied, subscriptions would be charged $0 instead of the correct amount.


vishaltelangre:
@ershad _a Please review.

ershad (approved):
@vishaltelangre _a lgtm 👍🏼

Issue #959, opened 2025-08-21
Run Gumroad tests without secrets


slavingia:
Bounty: $100 for each service's credential(s) removed to run the tests

- [x] Braintree - https://github.com/antiwork/gumroad/pull/2323 by @romitg2

abhinavjha0239:
I know this is a long comment, but I wanted to understand the requirements in depth before proceeding.

Hi @slavingia — I’d like to align on the exact expectations for issue #959 “Run Gumroad tests without secrets” before proceeding further. Based on the current repo structure (Rails + RSpec, with payments under app/business/payments/, various service objects in app/services/, mailers in app/mailers/, channels in app/channels/, many Sidekiq workers in app/sidekiq/, and extensive specs under spec/**), here’s my understanding and a few questions to firm up scope and acceptance criteria.

My understanding of the goal

Run the full RSpec suite from a fresh clone with zero external secrets or private configs (Stripe, PayPal, Braintree, email providers, etc.).

No outbound network calls required to pass by default.

Optionally keep a separate integration layer (with secrets) for CI or maintainers.

Open questions (scope/expectations)

Coverage scope: Should 100% of RSpec examples pass without secrets, or is it acceptable to skip a small, clearly tagged set (e.g., :external)?

Definition of “without secrets”: Does this strictly mean no API keys/tokens in env at all, or is using local/dummy values acceptable if they never leave the process?

Outbound HTTP policy: Is enforcing a global “no network” policy in test (e.g., block all HTTP via WebMock, allow only localhost) the desired default?

Gateways in scope: Confirm we should mock/stub all payment providers used in the codebase: Stripe, PayPal, Braintree (and any others present). Are there any providers we can exclude or deprecate from tests?

Emails: Is the expectation to rely solely on ActionMailer::Base.deliveries in test and prevent all external SMTP/API calls?

File/storage/CDN: If any tests interact with storage/CDN (config/storage.yml), should we force the :test disk service and stub all remote S3/CloudFront interactions?

Background jobs: Are we okay forcing Sidekiq test modes (inline or fake) and avoiding any external queues/services during test?

Webhooks/WS: For inbound/outbound webhooks and AnyCable websockets (config/anycable.yml, app/channels/**), should we stub these completely in test?

Tiered testing: Preference between (a) full mocking by default vs (b) tiered: unit/functional tests offline, plus a separate integration job (tagged specs) that runs with secrets?

[...]

PR #2393, opened 2025-12-18
Remove sensitive Stripe and Google credentials from tests

yashranaway:
ref: #959

## What PR Does?

Removes the hard requirement for Stripe and Google credentials when running tests by implementing code-level fallbacks and bypassing external balance checks.

### Key Changes:

1.  **Stripe Initializer Updates**: Added dummy defaults for all Stripe-related constants in the test environment:
    - `STRIPE_API_KEY` -> `sk_test_dummy`
    - `STRIPE_PUBLIC_KEY_TEST` -> `pk_test_dummy`
    - `STRIPE_PLATFORM_ACCOUNT_ID` -> `acct_test_dummy`
    - `STRIPE_CONNECT_CLIENT_ID` -> `ca_test_dummy`
    - `STRIPE_SECRET` -> `sk_test_dummy`

2.  **Google Initializer Updates**: Added dummy defaults for Google-related constants in the test environment:
    - `GOOGLE_CLIENT_ID` -> `test-google-client-id`
    - `GOOGLE_CLIENT_SECRET` -> `test-google-client-secret`
    - `GOOGLE_CLOUD_PROJECT_ID` -> `test-google-cloud-project`

3.  **`StripeBalanceEnforcer` Patch**: Updated to automatically skip real Stripe balance checks (which would normally fail without real keys) when the dummy test key is detected.

4.  **`.env.test` Cleanup**: Completely removed the literal Stripe and Google credentials from `.env.test`.

## Before:
The test suite requires these sensitive or real-formatted keys to be present in `.env.test` just to boot or run system specs.

## After:

image`

## AI Disclosure:

Gemini 3 Pro was used to implement via Cursor 

## Live Stream Disclosure:

### DIFF
--- a/spec/support/stripe_balance_enforcer.rb
+++ b/spec/support/stripe_balance_enforcer.rb
@@ -41,7 +41,7 @@ class StripeBalanceEnforcer
     end
 
     def using_dummy_stripe_key?
-      Stripe.api_key.nil? || Stripe.api_key.include?("dummy")
+      Stripe.api_key.nil? || Stripe.api_key == "sk_test_dummy"
     end
 
     def top_up!
--- a/.env.test
+++ b/.env.test
@@ -28,8 +28,7 @@ PAYPAL_PARTNER_MERCHANT_ID=TESTPAYPALID
 HELPER_WIDGET_SECRET=test-helper-secret
 HELPER_TOOLS_TOKEN=test-helper-tools-token
 GRMC_WEBHOOK_SECRET=test-grmc-webhook-secret
-STRIPE_PUBLIC_KEY_TEST=pk_test_dummy_public_key_for_testing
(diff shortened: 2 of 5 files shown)


spec/support/stripe_balance_enforcer.rb
```diff
@@ -39,6 +40,10 @@ def current_balance_cents
       usd_balance ? usd_balance["amount"] : 0
     end
 
+    def using_dummy_stripe_key?
+      Stripe.api_key.nil? || Stripe.api_key.include?("dummy")
+    end
+
```
yashranaway, line 46: Skip balance check if using dummy keys to allow testing without real credentials

slavingia:
Gemini 3 isn’t a model. Pro and Flash are.

PR #2404, opened 2025-12-19
Fix: make navigation links sidebar sticky in api page

romitg2:
### Problem:
- sidebar to navigate between sections in api page is not sticky, which makes it harder to navigate as we've to scroll till top to navigate between sections. 

### desktop:
https://github.com/user-attachments/assets/12a1d456-1622-4125-b15d-d3b811b95ec1

### mobile:
https://github.com/user-attachments/assets/5663d5a6-fe17-403b-9fe3-08e10630a41f

### desktop:
https://github.com/user-attachments/assets/c05d3fb5-f201-495a-a73b-6aedfc8e01c3

### mobile: (same as before for mobile)
https://github.com/user-attachments/assets/a1dd4a52-5520-4210-8987-28b5b6ecdf88

### AI Disclosure:
- watched all live streams

### Live Stream Disclosure:
- watched all live streams

### DIFF
--- a/app/views/public/api/shared/_navigation.html.erb
+++ b/app/views/public/api/shared/_navigation.html.erb
@@ -1,4 +1,4 @@
-<div role="navigation" aria-label="API Reference" class="lg:sticky lg:top-8 lg:max-h-[calc(100vh-4rem)] lg:overflow-auto">
+<div role="navigation" aria-label="API Reference" class="lg:sticky lg:top-8 lg:max-h-[calc(100vh-4rem)] lg:h-full lg:overflow-auto">
   <menu>
     <li>
       <%= link_to("Introduction", api_path(anchor: "api-intro")) %>


romitg2:
> Just in case, could you handle the case where the screen is shorter than the navigation and make it scrollable?

- updated style to make navigation sidebar scrollable when screen is shorter. 

https://github.com/user-attachments/assets/fb410b56-8f47-40b0-a990-b534734a6fac

https://github.com/user-attachments/assets/de905069-cfee-4ae5-bf1a-eeb1337ba17f

binary-koan:
That video shows a scrollbar even when it fits on the screen?

romitg2:
> That video shows a scrollbar even when it fits on the screen?

I've noticed that too and tried a few CSS/tailwind fixes for that, but nothing has worked, also not sure why scrollbar shows. would appreciate any suggestion/help.

binary-koan:
Just adding h-full seems to work, will merge once specs pass!

Issue #1055, opened 2025-08-28
Migrate away from CSS files to Tailwind


slavingia:
- Include screenshots of every usage.
- Self-review all code!
- Mobile and desktop, light and dark mode
- Make your PRs easy to read, well-written English, etc – do NOT use AI for the PR description.
- Use the classnames helper: `import classNames from "$app/utils/classNames";`
- Simplify and improve the CSS, using built-in Tailwind values wherever possible - don't just copy the calculations and complexity from the CSS files!

> [!NOTE]
> Please run the following [AI prompt](https://github.com/antiwork/gumroad/blob/main/docs/prompts/tailwindcss.md) against your changes to ensure your migration follows TailwindCSS' best practices and our guidelines.

**Bounty:** $1.5K per file

Make sure to include before/after for mobile/desktop and dark/light mode for each change. Use components if they make sense.

Nicely formatted PRs (see core members' PRs for examples) that are fun to read and code that's easy to read will be merged first/only.

- [x] _alert.scss
- [x] _application.scss
- [x] _aside.scss
- [x] _audio.scss
- [ ] _button.scss - https://github.com/antiwork/gumroad/pull/2324
    - [ ] Follow up changes: remove `NavigationButton`, move to `components/ui`
- [x] _calendar.scss
- [x] _card.scss
- [x] _carousel.scss
- [x] _cart.scss
- [x] _chart.scss
- _colors.scss  - **Skip for now.** Can be deleted right at the end, there's nothing to create in Tailwind.
- [x] _comment.scss
- [x] _containers.scss
- [x] _custom_sections.scss
- _definitions.scss  - **Skip for now.** Can be deleted right at the end, there's nothing to create in Tailwind.
- [x] _discover.scss
- [ ] _dropdown.scss - **Bounty: $2K**
- [ ] _email.scss
- [x] _figure.scss
- _font.scss - **Skip for now.** Will likely remain as a CSS file imported from the likely-separate Tailwind configs for the website, emails, and embeds.
- [ ] _forms.scss - **Bounty: $2K**
- [ ] _global.scss
- [x] _grid.scss and override class
- [x] _histogram.scss
- _icons.scss - brand logos and Gumroad logos could be done as separate PRs, but skip the icons themselves for now
- [x] _image_uploader.scss - https://github.com/antiwork/gumroad/pull/1411 by @kr-sushil 
- [ ] _legacy.scss
- [x] _modal.scss
- [x] _nav.scss
[...]

PR #2405, opened 2025-12-19
Remove and Refactor `_colors.scss`

SalmanNajah:
Part of #1055

- This PR removes `_colors.scss` and refactors all the color logic into `_definitions.scss`.
--- 

## AI Disclosure
Used Antigravity(gemini-3-pro) to understand the `_colors.scss` usage and make the appropriate changes.

## Livestream videos
I've watched all the videos.

### DIFF
--- a/app/javascript/stylesheets/_button.scss
+++ b/app/javascript/stylesheets/_button.scss
@@ -60,7 +60,7 @@
     .button-#{$name} {
       $color: map-get($info, color);
       background-color: $color;
-      color: map-get($info, text-color) or fg-color(split-color($color));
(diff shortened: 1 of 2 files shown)


SalmanNajah:
@binary-koan _a What do you think about the approach in this PR?

binary-koan:
I don't think it's worth doing this tbh since although it reduces the number of files it just increases the migration complexity for `_definitions.scss`. Looking at this I think we should just delete `_colors.scss` at the end of the migration since there isn't a way to convert it to Tailwind (it's purely part of the SCSS). I'll make a note in the issue

PR #2287, opened 2025-12-11
Inertia migration - Library

4arjun:
ref #856
Previous PRs: #2040 #2200 #2253
Reopened #2253

This PR migrates the following pages to Inertia:

1. Library (/library)
2. Reviews (/reviews)
3. Wishlists (/wishlists)
4. Following Wishlists (/wishlists/following)

Follows the same pattern as the previous migrations(ref #2005, https://github.com/antiwork/gumroad/pull/1627, https://github.com/antiwork/gumroad/pull/1559)

## Before

https://github.com/user-attachments/assets/e352c73b-b5da-413a-88b4-36f6289f01dc

## After

https://github.com/user-attachments/assets/25a4531f-232e-472f-bbdf-29e155d3c807

## Tests

Screenshot 2025-12-03 at 9 11 14 PM

## 
Screenshot 2025-12-11 at 2 58 23 PM

## AI Usage
Used GPT-5.1 (Cursor) to understand the migration done by pr https://github.com/antiwork/gumroad/pull/2005 https://github.com/antiwork/gumroad/pull/1627

## Confirmation
I have watched the Gumroad PR reviews live streaming video.

## Self-Review
I have self-reviewed all the code that I have written.

PR #2417, opened 2025-12-20
Migrate emails to inertia

stefanbinoj:
Part of #856 

As per the [comment](https://github.com/antiwork/gumroad/pull/2387#pullrequestreview-3599053834) this PR Migrate aims to breakdown and migrate following routes to inertia : 
- Published emails page (/emails/published)
- Scheduled emails page (/emails/scheduled)

## Changes
- While clicking `Load More` inertia page reloads with the new `?page` query params and fetch the corresponding installments from db. The response has installment_props which is catched by `usePageHooks` from intertia. 
- Client also has a `hooks/usePaginatedAccumulation.ts` that stores all fetched data from `?page=1` to `?page=N`
(This helps in db optimizations while still relying on inertia)
- `Delete` functionality has a common `components/EmailsPage/DeleteEmailModal.tsx` which sends a inertia form req to backend which returns **flash** message
- The `onSuccess` handler runs and filters the deleted installment from accumulated state.
- Also optimizes db calls by addressing some `Bullet Warnings` 

## Before 

https://github.com/user-attachments/assets/45a080c0-ebc5-452d-a743-eec70ffb6fbb

## After 
(full page reloads only happens when we navigate from /scheduled -> /draft and vice-versa) 

https://github.com/user-attachments/assets/faf86024-7e0d-48bd-b288-fe0a0f221e18

## AI disclosure
Used Claude Opus 4.5 via Opencode for initial drafting of this PR

## Confirmation


Devwill2:
@stefanbinoj bro, how would you feel if I started working on the auth part with small PRs while you’re already working on it? I’ve been working on the email part for weeks and I can split it as well. I’d appreciate it if we could acknowledge each other’s efforts.

stefanbinoj:
> @stefanbinoj bro, how would you feel if I started working on the auth part with small PRs while you’re already working on it? I’ve been working on the email part for weeks and I can split it as well. I’d appreciate it if we could acknowledge each other’s efforts.

Heads up, @Devwill2. The general norm applies here: whoever ships better code gets it merged.
Also, could you refer to this comment from @4arjun (#2287 [comment](https://github.com/antiwork/gumroad/pull/2287#issuecomment-3670926002))? He has been working on it for months (not weeks), and you still worked on a smaller part of his work, right?

It also seems that there are plenty of other PRs attempting the same thing you referred to (I’m not going to tag them individually here, but there are quite a few).

Please don’t get offended; I hope you understand that you did exactly what you mentioned earlier.

app/controllers/emails_controller.rb
```diff
@@ -1,21 +1,56 @@
...
+    @title = "Emails"
+    @body_id = "app"  # Only set for catch-all routes like /emails/drafts
+  end
+
+  def published
+    authorize Installment, :index?
     create_user_event("emails_view")
 
-    if request.path == emails_path
-      default_tab = Installment.alive.not_workflow_installment.scheduled.where(seller: current_seller).exists? ? "scheduled" : "published"
```
stefanbinoj, line 12: More readable code would be to use the following as in new branch. This also removes unecessary db joining and lookups!
```ruby
schedule_exits = current_seller.installments.alive.not_workflow_installment.scheduled.exists?

```

app/controllers/emails_controller.rb
```diff
@@ -1,21 +1,56 @@
...
+  before_action :set_installment, only: [:destroy]
 
   def index
     authorize Installment
 
+    if request.path == emails_path
+      if current_seller.installments.alive.not_workflow_installment.scheduled.exists?
+        redirect_to scheduled_emails_path, status: :moved_permanently
+      else
+        redirect_to published_emails_path, status: :moved_permanently
```
stefanbinoj, line 15: This code runs when `request.path === "/emails"` (no pathname succedding `/emails` like /emails/draft ... ). In that case we redirect to /scheduled if user has any or fallback to /published

app/controllers/emails_controller.rb
```diff
@@ -1,21 +1,56 @@
...
+      end
+    end
+    @title = "Emails"
+    @body_id = "app"  # Only set for catch-all routes like /emails/drafts
```
stefanbinoj, line 19: Incase when path is not `/emails` (for eg /emails/draft, /emails/new, /emails/:id/edit ) this code sets title and `body_id` since we earlier removed the following code from controller.
```ruby
  before_action :set_body_id_as_app
  ```

Comment (stefanbinoj) [reply app/controllers/emails_controller.rb:19]: Even if we update the before_acction like : 
```ruby 
  before_action :set_body_id_as_app, only: [:index]
  ```
  This would apply `id=#app` for inertia pages (/published or /scheduled) for initial req when req.path === '/emails' . Therefore dynamically setting `body_id` for non-inertia routes makes sense

File: app/controllers/emails_controller.rb [hunk 4]
```diff
@@ -1,21 +1,56 @@
...
+  end
+
+  def published
+    authorize Installment, :index?
     create_user_event("emails_view")
 
-    if request.path == emails_path
-      default_tab = Installment.alive.not_workflow_installment.scheduled.where(seller: current_seller).exists? ? "scheduled" : "published"
-      redirect_to "#{emails_path}/#{default_tab}", status: :moved_permanently
+    @title = "Published Emails"
```
stefanbinoj, line 26: Better title since we have more control now

app/controllers/emails_controller.rb
```diff
@@ -1,21 +1,56 @@
...
+    presenter = PaginatedInstallmentsPresenter.new(seller: current_seller, type: Installment::PUBLISHED, page: params[:page], query: params[:query])
```
stefanbinoj, line 27: We now pass `page` and `query` from request as well

app/controllers/emails_controller.rb
```diff
@@ -1,21 +1,56 @@
...
+
+  def destroy
+    authorize @installment
+
+    if @installment.mark_deleted
+      @installment.installment_rule&.mark_deleted!
+      redirect_back fallback_location: published_emails_path, notice: "Email deleted!", status: :see_other
+    else
+      redirect_back fallback_location: published_emails_path, alert: "Sorry, something went wrong. Please try again.", status: :see_other
     end
```
stefanbinoj, line 48: Now we have a new endpoint for deleting published and scheduled emails! Removed dependency on internal api. 
Also since both pages (/published and /scheduled) uses same destroy endpoint added `redirect_back`
```ruby
redirect_back fallback_location: published_emails_path ... 
```

app/javascript/components/EmailsPage/DeleteEmailModal.tsx
```diff
@@ -0,0 +1,70 @@
+import { useForm } from "@inertiajs/react";
```
stefanbinoj, line 1: Since both /published and /scheduled page has a common delete modal, created a centralized modal for following DRY principle

app/javascript/components/EmailsPage/DeleteEmailModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  onClose,
+  onSuccess,
+  confirmationMessage,
+}: Props<T>) {
+  const deleteForm = useForm({});
+
+  const handleDelete = () => {
+    if (!installment) return;
+    deleteForm.delete(Routes.email_path(installment.external_id), {
+      preserveState: true,
```
stefanbinoj: `preserveState: true ` is added because without it inertia reloads page and props with the same queries for `?page=` and `?query=`
Since we are now handling a bit of logic of removing the delted installment in frontend (for optimized db calls) it is ideal to not let inertia rerun the props and handle the states client side itself

app/javascript/components/EmailsPage/Layout.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <Tab asChild isSelected={selectedTab === "scheduled"}>
+            <Link href={Routes.scheduled_emails_path()}>Scheduled</Link>
+          </Tab>
+          {/* Legacy pages - use regular href for full page reload */}
+          <Tab href="/emails/drafts" isSelected={selectedTab === "drafts"}>
+            Drafts
+          </Tab>
+          <Tab href={Routes.followers_path()} isSelected={selectedTab === "subscribers"}>
+            Subscribers
+          </Tab>
```
stefanbinoj, line 94: Not using `Link` component from inertia since these pages are not migrated and doing a full page reloading here

app/javascript/components/server-components/EmailsPage/index.tsx
```diff
@@ -88,17 +84,20 @@ export const Layout = ({
...
-              </Tab>
-            ),
-          )}
+          {/* Inertia pages - use Rails routes to trigger full page load */}
+          <Tab href={Routes.published_emails_path()} isSelected={selectedTab === "published"}>
+            Published
+          </Tab>
+          <Tab href={Routes.scheduled_emails_path()} isSelected={selectedTab === "scheduled"}>
+            Scheduled
+          </Tab>
```
stefanbinoj, line 93: In the legacy `Layout` we are still doing a full page reload to inertia pages from non-inertia  to inertia pages. 
But note that once if the page is inertia it loads `app/javascript/components/EmailsPage/Layout.tsx` this layout instead and it has `Link` components to inertia and `<a>` components to non-inertia pages

app/javascript/hooks/useDebouncedSearch.ts
```diff
@@ -0,0 +1,27 @@
...
+ * - Maintains local query state for immediate UI updates
+ * - Debounces the actual server request by 500ms
+ * - Only triggers reload when query actually changes (not on mount)
+ * - Resets to page 1 when search query changes
+ */
+export function useDebouncedSearch() {
+  const [query, setQuery] = React.useState("");
+
+  const handleQueryChange = React.useCallback((newQuery: string) => {
+    router.reload({ data: { query: newQuery || undefined, page: 1 } });
```
stefanbinoj, line 19: NOTE : we are setting `?page=1&query=xyz` implicitly because: 
In scenarios when user is on `?page=4` (Since we hanlde accumulated pages in client side `useState` if user searches for anything then a new inertia req sends with the query and gets the db responds with only that and since we are already have a accumulated state for installments it will only append the missing results from response (which has search with query). This will leads to unexpectd behaviour. Thus setting `page=1` clears accumulated state

app/javascript/hooks/usePaginatedAccumulation.ts
```diff
@@ -0,0 +1,44 @@
...
+      setAllItems(items);
+    } else {
+      // Append new items for subsequent pages
+      setAllItems((prev) => {
+        const existingIds = new Set(prev.map((i) => i.external_id));
+        const newItems = items.filter((i) => !existingIds.has(i.external_id));
+        return [...prev, ...newItems];
+      });
+    }
+  }, [items, pagination.page]);
```
stefanbinoj, line 37: This is bit of a hack for optimizing db performace, we allow client side to control the state of all fetched installments and handle `page` increment. 
### Reason: 
If we blindly reload page by 
```tsx
router.reload({ data: { page: pagination.next, query: query || undefined } });
```
This will only fetch the data for current page only or if we try to fetch data from 1st page -> Nth page this will cause reduntant db fetching in each of `Load More` button clicks. Hence this way of handling states of fetched installments till the current page is handled on client side


This is the reason why when we click on delete button we only listen for `["flash"]` and just filter the current state

app/presenters/paginated_installments_presenter.rb
```diff
@@ -17,11 +17,14 @@ def initialize(seller:, type:, page: nil, query: nil)
...
-      installments = Installment.includes(:installment_rule).all
+      # Always include :seller (needed for full_url -> user -> seller.presence)
+      # Only include :installment_rule for scheduled/draft (not needed for published)
+      installments = Installment.includes(:seller)
+      installments = installments.includes(:installment_rule) if type != Installment::PUBLISHED
       installments = installments.ordered_updates(seller, type).public_send(type)
       installments = installments.unscope(:order).order("installment_rules.to_be_published_at ASC") if type == Installment::SCHEDULED
       pagination, installments = pagy(installments, page:, limit: PER_PAGE, overflow: :empty_page)
-      pagiation_metadata = { count: pagination.count, next: pagination.next }
+      pagiation_metadata = { page: pagination.page, count: pagination.count, next: pagination.next }
```
stefanbinoj, line 27: Now returns current page for showing it in query params for router reload ?page=

app/presenters/paginated_installments_presenter.rb
```diff
@@ -17,11 +17,14 @@ def initialize(seller:, type:, page: nil, query: nil)
 
   def props
     if query.blank?
-      installments = Installment.includes(:installment_rule).all
+      # Always include :seller (needed for full_url -> user -> seller.presence)
+      # Only include :installment_rule for scheduled/draft (not needed for published)
+      installments = Installment.includes(:seller)
+      installments = installments.includes(:installment_rule) if type != Installment::PUBLISHED
```
stefanbinoj, line 23: Previously In `/published` we,  after loading installments, we're accessing installment.seller for each installment without having eager loaded the seller association. This causes N+1 queries.

Bullet Warning
user: stefanbinoj USE eager loading detected Installment => [:seller] Add to your query: .includes([:seller])

app/presenters/paginated_installments_presenter.rb
```diff
@@ -36,9 +39,12 @@ def props
         sort: [:_score, { created_at: :desc }, { id: :desc }]
       }
       es_search = InstallmentSearchService.search(search_options)
-      installments = es_search.records.load
+      # Include associations to avoid N+1 queries
+      includes_list = [:seller]
+      includes_list << :installment_rule if type != Installment::PUBLISHED
+      installments = es_search.records.includes(*includes_list).load
```
stefanbinoj, line 45: Previously In `/published` we,  after loading installments, we're accessing installment.seller for each installment without having eager loaded the seller association. This causes N+1 queries.

Bullet Warning
user: stefanbinoj USE eager loading detected Installment => [:seller] Add to your query: .includes([:seller])

app/presenters/paginated_installments_presenter.rb
```diff
@@ -17,11 +17,14 @@ def initialize(seller:, type:, page: nil, query: nil)
 
   def props
     if query.blank?
-      installments = Installment.includes(:installment_rule).all
```
stefanbinoj, line 20: This means we're eager loading installment_rule with .includes([:installment_rule]) in both /published and /scheduled (but we are not using it in /published)

Bullet Warning
user: stefanbinoj AVOID eager loading detected Installment => [:installment_rule] Remove from your query: .includes([:installment_rule])

config/routes.rb
```diff
@@ -398,7 +398,6 @@ def product_info_and_purchase_routes(named_routes: true)
     get "/collaborators/*other", to: "collaborators#index"
 
     get "/affiliates/*other", to: "affiliates#index" # route handled by react-router
-    get "/emails/*other", to: "emails#index" # route handled by react-router
```
stefanbinoj, line 401: Now since we have dedicated /published and /scheduled routes it is better to decrement the priority of this line (or else we will get in a redirection loop)

config/routes.rb
```diff
@@ -802,7 +801,15 @@ def product_info_and_purchase_routes(named_routes: true)
...
-    get "/emails", to: "emails#index", as: :emails
+    # TODO: Add :drafts, :new, :edit routes once those pages are migrated to Inertia
+    resources :emails, only: [:index, :destroy] do
+      collection do
+        get :published
+        get :scheduled
+      end
+    end
+    # TODO: Remove this catch-all once drafts, new, edit email pages are migrated to Inertia
+    get "/emails/*other", to: "emails#index" # route handled by react-router
```
stefanbinoj, line 812: Now this has lesser priority allowing other endpoints like /published, /scheduled, /destroy to run directly

app/javascript/components/EmailsPage/DeleteEmailModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  onClose,
+  onSuccess,
+  confirmationMessage,
+}: Props<T>) {
+  const deleteForm = useForm({});
+
+  const handleDelete = () => {
+    if (!installment) return;
+    deleteForm.delete(Routes.email_path(installment.external_id), {
+      only: ["flash"],
```
stefanbinoj, line 37: `only:["flash"]` is added because without it inertia reloads page with all set of initial props with the same queries for `?page=` and `?query=`. Now only `flash` is returned : 

Screenshot 2025-12-22 at 1 51 32 PM

### Reason: 
Since we are now handle accumulated states for all `?pages=` combined(discussed later) in client, filtering  the deleted installments in frontend `state` and updating itself in client side is better if endpoint returned success success. This has better db optimizations IMO

app/javascript/hooks/usePaginatedAccumulation.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function usePaginatedAccumulation<T extends { external_id: string }>(
+  items: T[],
+  pagination: Pagination,
+  query: string,
+) {
+  const [allItems, setAllItems] = React.useState(items);
+
+  React.useEffect(() => {
+    if (pagination.page === 1) {
```
stefanbinoj, line 26: when `?query=` changes we set `?pages=1` explicitly to clear all accumulated states

app/javascript/hooks/usePaginatedAccumulation.ts
stefanbinoj: Another approach as i said would be to let server be the source of truth but this makes server burden with fetching all contents from page 1 to page N . (especially when N is large). wdyt @EmCousin ? 
Also this would require to have a new `PaginatedInstallmentsPresenter` for fetching elements from 1st to Nth (As there a couple of dependencies on this presenter which are still using legacy way of fetching single page by page) or updating all together or maybe having two mechanism based on props like `isInertia` or something like that

app/presenters/paginated_installments_presenter.rb
```diff
@@ -17,11 +17,14 @@ def initialize(seller:, type:, page: nil, query: nil)
```
stefanbinoj, line 17: ### Before Bullet Warnings : 

https://github.com/user-attachments/assets/49168f03-a5cb-47aa-bc28-e6858c712511



### After Bullet Warnings: 


https://github.com/user-attachments/assets/969cd9bc-ff6e-44fd-839e-aae868594d49

app/javascript/hooks/useDebouncedSearch.ts
stefanbinoj: ### Before 


https://github.com/user-attachments/assets/f2ba6140-148d-4d44-9f6f-ea9cb8d9a3f7


### After 


https://github.com/user-attachments/assets/b741dfab-88ad-47bc-ab95-f0a48b41819c

app/javascript/components/EmailsPage/DeleteEmailModal.tsx
stefanbinoj: ### Video : 


https://github.com/user-attachments/assets/c07a11c0-dfcb-4df9-924c-a95300dd7dda

app/javascript/hooks/useDebouncedSearch.ts
EmCousin: If I want to hit this URL with ?query=foo in my URL bar, the input should be present with the "foo" string, and clicking on "Load more..." should be loading the next filtered results based on the same query.

In this case, search state isn’t hydrated from the server query, so visiting `/emails/published?query=foo` (or hard-refreshing after a search) renders filtered results, but the local query stays empty. Clicking “Load more”, sends the request with an empty query, so page 2 is unfiltered and mixes with the filtered page 1. The search box also shows blank despite filtered data.

app/javascript/hooks/usePaginatedAccumulation.ts
EmCousin: The approach is to _let the server tell the client what to do_: https://inertia-rails.dev/guide/merging-props#merging-props

There is even a [nicer solution](https://inertia-rails.dev/guide/infinite-scroll), but we need to upgrade to inertia-rails 3.12 first.

In any case, you indeed shouldn't have needed this workaround.

app/controllers/emails_controller.rb
EmCousin: The corresponding legacy endpoint in `api/internal/installments#destroy` is no longer used: you need to remove it as well as the TS helper that calls it in the frontend.

app/javascript/hooks/usePaginatedAccumulation.ts
```diff
@@ -0,0 +1,44 @@
...
+      setAllItems((prev) => {
+        const existingIds = new Set(prev.map((i) => i.external_id));
+        const newItems = items.filter((i) => !existingIds.has(i.external_id));
+        return [...prev, ...newItems];
+      });
+    }
+  }, [items, pagination.page]);
+
+  const handleLoadMore = React.useCallback(() => {
+    router.reload({ data: { page: pagination.next, query: query || undefined } });
```
EmCousin, line 40: The legacy API endpoint is no longer needed (`api/internal/installments#index`) and should be removed, alongside the TS helper that calls it.

PR #2388, opened 2025-12-17
Inertia Migration - Auth (Part 1)

stefanbinoj:
Part of #856 
Previous PR #2345 , PR #2366

This PR migrate following pages to inertia 
- /login 
- /signup
Also migrates all related controller actions

## Before 

https://github.com/user-attachments/assets/862cc398-0f8e-440b-89d4-db103f695f68

 
## After 
(Full page reload happens while navigating from landing page -> login page because we haven't migrated navbar for landing page :) , also would be better to make it a seperate PR for modular changes)

https://github.com/user-attachments/assets/39254910-f1fe-40a3-997d-9ce64c30a5c0

## Tests

```bash
bundle exec rspec spec/controllers/logins_controller_spec.rb
```

Screenshot 2025-12-15 at 8 20 06 PM

```bash
bundle exec rspec spec/controllers/signup_controller_spec.rb
```

Screenshot 2025-12-15 at 8 27 16 PM

**NOTE :** Failing tests are due to missing env for stripe payment processing 

### AI disclosure
Used Claude Opus 4.5 via Opencode for initial drafting 

### Confirmation
I have watched all Antiwork Livestreams

PR #2423, opened 2025-12-20
chore(inertia): migrates widgets page to inerita

romitg2:
ref: #856 

# Before:
### desktop:

https://github.com/user-attachments/assets/1a557459-b54c-4f87-95e5-ca48fbee4a42

### mobile:

https://github.com/user-attachments/assets/a2ac3af8-7ead-4602-8a64-8af49ac4d9e4

# After:
### desktop:

https://github.com/user-attachments/assets/b1ab49cb-d6fc-40bd-a55b-68e0592feecf

### mobile:

https://github.com/user-attachments/assets/a7f42ee1-36cd-44dc-8db0-8d97f426fde3

## Tests Result: (local run)
`spec/controllers/public_controller_spec.rb  `
Screenshot 2025-12-21 at 6 06 05 PM
`spec/requests/widget_spec.rb`
Screenshot 2025-12-21 at 6 05 09 PM

### AI Disclosure:
- used claude-code with sonnet-4.5 for refactoring

### live stream Disclosure:
- watched all live streams


app/javascript/inertia/layout.tsx
```diff
@@ -28,7 +30,7 @@ export default function Layout({ children }: { children: React.ReactNode }) {
       <Head title={title} />
       <Alert initial={flash ?? null} />
       <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
-        <Nav title="Dashboard" />
+        {loggedInUser ? <Nav title="Dashboard" /> : null}
```
romitg2: for public pages (user is not logged in) we don't show navigation.

app/views/layouts/inertia.html.erb
```diff
@@ -17,6 +17,10 @@
     <% end %>
     <%= javascript_include_tag "application" %>
     <%= load_pack("base") unless @hide_layouts %>
+    <% if @widget_scripts %>
```
romitg2: `@widget_scripts ` is set in the widgets action of `PublicController`
- Previously, these scripts were loaded in the deleted `widgets.html.erb` template.

stefanbinoj: Hi @romitg2, Just a few notes here. Previously, I received a similar review for loading scripts based on values from an instance variable, and I used the same approach you did. In that i got a feedback that said 'this doesn't seems to be a valid method'. We should aim to load scripts dynamically on the client side instead -- As far as i understood 

For more context, please read this [comment](https://github.com/antiwork/gumroad/pull/2388#discussion_r2635281782)

Hope this is helpful 🙏🏻

romitg2: Thanks! @stefanbinoj

app/javascript/pages/Public/Widgets.tsx
```diff
@@ -0,0 +1,38 @@
...
+  default_product: Product;
+  display_product_select: boolean;
+  products: Product[];
+  affiliated_products: Product[];
+};
+
+const PublicWidgets = () => {
+  const { default_product, display_product_select, products, affiliated_products } = cast<Props>(usePage().props);
+
+  useExternalScripts(["/js/gumroad.js", "/js/gumroad-embed.js"]);
```
romitg2: Previously, these scripts were loaded in the deleted `widgets.html.erb` template.
- now we dynamically load them on client side. 

---

### correctly loading scripts on client. 
Image

app/controllers/concerns/inertia_rendering.rb
```diff
@@ -7,7 +7,7 @@ module InertiaRendering
   included do
     inertia_share do
       RenderingExtension.custom_context(view_context).merge(
-        current_user: current_user_props(current_user, impersonated_user),
+        current_user: user_signed_in? ? current_user_props(current_user, impersonated_user) : nil,
```
romitg2: for public pages (user may not be signed in), the `current_user` can be `nil`

spec/controllers/public_controller_spec.rb
romitg2, line 1: updated specs, tests result: (local run)
Image

app/javascript/hooks/useExternalScripts.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from "react";
+
+export const useExternalScripts = (scripts: string[]) => {
```
romitg2: [for dynamically loading external scripts on client. ](https://github.com/antiwork/gumroad/pull/2423/changes#r2637750366)

romitg2:
@binary-koan could you please review this PR?

Thanks!

app/javascript/components/server-components/Developer/WidgetsPage.tsx
```diff
@@ -17,14 +15,19 @@ import { useHasChanged } from "$app/components/Developer/useHasChanged";
...
-type Props = {
+export type WidgetsPageProps = {
   default_product: Product;
   display_product_select: boolean;
   products: Product[];
   affiliated_products: Product[];
 };
 
-export const WidgetsPage = ({ display_product_select, products, affiliated_products, default_product }: Props) => {
+export const WidgetsPage = ({
```
binary-koan: Is this still used on other pages? If not we should delete the server-components file and inline the content into the new Inertia page

spec/controllers/public_controller_spec.rb
```diff
@@ -57,11 +56,13 @@
...
       include_context "with user signed in as admin for seller"
 
-      it "initializes WidgetPresenter with seller" do
+      it "renders the inertia page with correct component, props, and title" do
         get :widgets
 
         expect(response).to be_successful
-        expect(assigns[:widget_presenter].seller).to eq(seller)
+        expect(inertia).to render_component("Public/Widgets")
+        expect(inertia.props).to be_present
```
binary-koan: Could we assert something about the props? Or just remove this assertion, I don't think it's very useful as is

app/javascript/hooks/useExternalScripts.ts
```diff
@@ -0,0 +1,28 @@
...
+      script.src = src;
+      script.async = true;
+      document.body.appendChild(script);
+      loadedScripts.push(script);
+    });
+
+    return () => {
+      loadedScripts.forEach((script) => {
+        if (script.parentNode) {
+          script.parentNode.removeChild(script);
```
binary-koan: If we add a script to the array it'll remove the previous ones, then add them back along with the new one. Maybe that's fine, but I think it would be simpler to make this `useExternalScript` with a single src and just call it twice in the widget page

app/javascript/components/server-components/Developer/WidgetsPage.tsx
romitg2: makes sense! inlined into inertia page.

romitg2:
@binary-koan _a addressed the comments, could you please review again?

Thanks!

binary-koan:
One more thing - could you fix the navigation to avoid a page reload between widgets and ping (now that they're both in Inertia)? I don't think we need the `.map`, we can just inline the three links in `Layout` and switch to the right component for the Inertia pages

app/javascript/widget/overlay.ts
```diff
@@ -49,9 +49,11 @@ const registerButton = (button: HTMLAnchorElement) => {
...
   }
 
-  const logo = document.createElement("span");
-  logo.className = "logo-full";
-  button.appendChild(logo);
+  if (!button.querySelector(".logo-full")) {
+    const logo = document.createElement("span");
+    logo.className = "logo-full";
+    button.appendChild(logo);
+  }
```
romitg2: earlier upon each revisit to "widgets" tab, additional button logo was being added as scripts re-attached. 

Image

binary-koan: Hmm, looks like there will be other problems with removing and reattaching the script tag as we'll create multiple overlay elements, add multiple event listeners, etc. We should probably try to clean everything up when the script tag is removed (Claude suggested checking for its removal in the existing mutation observer)

Alternatively we could just not remove the script tag once we've loaded it and rename the hook to `usePersistentExternalScript` or something (I suspect a lot of third party scripts also aren't designed to be reattached like this)

binary-koan:
Pushed the `usePersistentExternalScript` approach, I think we can just merge with that. We can implement proper cleanup in the overlay script later if we run into issues or if someone feels more strongly about it!

romitg2:
> Pushed the `usePersistentExternalScript` approach, I think we can just merge with that. We can implement proper cleanup in the overlay script later if we run into issues or if someone feels more strongly about it!

@binary-koan there was an issue which i was trying to fix before merging, when we first visit /ping page and then switch to /widgets page then overlay.css and design.css files loads and causes a flash with unstyled page for a moment.

https://github.com/user-attachments/assets/cbd62194-ee3e-4b73-b7af-e8b087bc2ba8

will try to create another PR to fix it, or should we revert this?

binary-koan:
My bad, I didn’t notice that locally. Another PR is fine!

binary-koan:
(By the way, please remind me here if you can't find a fix and I'll look into it)

Issue #1594, opened 2025-10-09
Ability to load funds to Gumroad balance to cover refunds when balance is too low


michellelarney:
### What

Ability to load funds to Gumroad balance to cover refunds when balance is too low. Connect a credit card and charges happen in real-time per refund. 

### Why

Allows sellers to refund customers without having to wait for sales to come in. There are instances when a seller's balance isn't enough to cover a full refund (ex. recently paid out). This will help sellers be able to refund customers sooner and through our dashboard.

binary-koan:
We found a key blocker: a fraudulent seller could try to dispute the credit card refund payments, which is a major risk to us (and adds a bunch of backend complexity to track the disputes even if we can reliably win them).

A couple of options might be
* Continue with the credit card approach but only enable it for a verified sellers (or some trusted subset)
* Allow sellers to go into negative balance instead, i.e. use their next payout(s) to cover the refunds (similarly we could only enable this for trusted sellers, but we could more easily control the max negative amount and it's a much simpler change)

For reference, Lemon Squeezy allows sellers to go into negative balance to cover refunds, while Shopify (in major markets like US/UK/EU) debits the seller's bank account to cover refunds. Will check some other competitors too.

(cc @slavingia in case you have thoughts on this)

PR #2425, opened 2025-12-21
feat: Implement backup payment method for refund funding

yashranaway:
Fixes : #1594

# Description

## Problem

Sellers can't process refunds when their Gumroad balance is insufficient to cover the refund amount.

## Solution

Enables sellers to add a backup payment method (credit card) that automatically covers refunds when their balance is too low.

When a seller visits the Customers page, they see a banner prompting them to set up a backup payment method. Clicking the link takes them to Payments Settings where they can add a card using Stripe Elements.

Once configured, when a refund is issued and the seller's balance can't cover it, the system automatically charges their backup card for the difference. The seller receives an email confirmation, and the refund processes normally for the customer.

Sellers can change or remove their backup card anytime from Payments Settings. If they dismiss the banner without adding a card, it stays hidden. If they later remove their card, the banner reappears.

**Design:** [Figma](https://www.figma.com/design/bQE0b8eQTM2eov0Kzjxk3A/Ability-to-load-funds-to-Gumroad-balance-to-cover-refunds-when-balance-is-too-low%C2%A0?node-id=3-4663&t=IAAPYaCaxBvc0iqo-1)

> [!NOTE]
> **ZIP/Postal Code**: I moved the ZIP field out of the Stripe `CardElement` into a dedicated standard input. This was done to bypass Stripe's automatic hiding logic on mobile/small widths, ensuring the field remains visible and consistently accessible across all devices. Can revert back as per original design if this approach isn't ideal.

---

# Before/After

### Desktop
https://github.com/user-attachments/assets/cd877fb9-8fd8-43c9-bbb2-fb30d0b9c443

### Mobile
image

image

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

Used Claude Opus 4.5 for implementation.

### DIFF
--- a/app/services/subscription/restart_at_checkout_service.rb
+++ b/app/services/subscription/restart_at_checkout_service.rb
@@ -1,90 +1,176 @@
 # frozen_string_literal: true
 
-# Adapter service that transforms checkout params and delegates to UpdaterService
-# for restarting cancelled/failed subscriptions during checkout.
-#
-# This reuses the battle-tested UpdaterService logic for charging, tier changes,
-# payment method updates, proration, SCA, webhooks, etc.
 class Subscription::RestartAtCheckoutService
-  CARD_PARAM_KEYS = %i[
-    card_data_handling_mode stripe_payment_method_id stripe_customer_id
-    stripe_setup_intent_id stripe_error paypal_order_id paymentToken
-    billing_agreement_id braintree_device_data braintree_transient_customer_store_key
-    visual card_country card_country_source
-  ].freeze
+  include CurrencyHelper
+
+  class ChargeFailed < StandardError
+    attr_reader :purchase
+
+    def initialize(message, purchase: nil)
+      super(message)
+      @purchase = purchase
+    end
+  end
 
   attr_reader :subscription, :product, :params, :buyer
+  attr_accessor :new_purchase
 
   def initialize(subscription:, product:, params:, buyer: nil)
     @subscription = subscription
     @product = product
     @params = params
     @buyer = buyer
+    @new_purchase = nil
   end
 
   def perform
-    result = Subscription::UpdaterService.new(
-      subscription: subscription,
-      params: build_updater_params,
-      logged_in_user: buyer,
-      gumroad_guid: params.dig(:purchase, :browser_guid),
-      remote_ip: params[:remote_ip]
-    ).perform
+    error = validate_restart
+    return error_result(error) if error.present?
 
-    adapt_result(result)
+    ActiveRecord::Base.transaction do
+      handle_tier_change if tier_changed?
+      update_payment_method if should_update_payment_method?
+      subscription.resubscribe!
+
+      if should_charge?
+        charge_result = charge_subscription
+        self.new_purchase = charge_result[:purchase]
+      end
+
+      subscription.send_restart_notifications!
+      success_result
+    end
+  rescue ChargeFailed => e
+    subscription.reload
+    subscription.unsubscribe_and_fail!
+    error_result(e.message)
+  rescue ActiveRecord::RecordInvalid, Subscription::UpdateFailed => e
+    error_result(e.message)
+  rescue StandardError => e
+    Rails.logger.error("Subscription::RestartAtCheckoutService error: #{e.message}")
+    Bugsnag.notify(e)
+    error_result("Sorry, something went wrong. Please try again.")
   end
 
   private
-    def build_updater_params
-      updater_params = {
-        variants: params[:variants] || default_variant_ids,
-        price_id: params[:price_id] || subscription.price&.external_id,
-        quantity: subscription.original_purchase.quantity,
-        perceived_price_cents: perceived_price,
-        perceived_upgrade_price_cents: perceived_price,
-      }
+    def validate_restart
+      return "This subscription cannot be restarted." if subscription.cancelled_by_seller?
+      return "This subscription cannot be restarted." if product.deleted?
+      return "This installment plan has already been completed and cannot be restarted." if subscription.is_installment_plan? && subscription.charges_completed?
+      nil
+    end
 
-      card_mode = CardParamsHelper.get_card_data_handling_mode(merged_card_params)
-      if card_mode.present? && card_mode != :reuse
-        updater_params.merge!(merged_card_params.slice(*CARD_PARAM_KEYS))
+    def tier_changed?
+      return false unless product.is_tiered_membership?
+      return false unless params[:variants].present?
+
+      selected_tier_ids = params[:variants].map { |id| product.tiers.find_by_external_id(id)&.id }.compact
+      current_tier_ids = subscription.original_purchase.variant_attributes.map(&:id)
+
+      selected_tier_ids.sort != current_tier_ids.sort
+    end
+
+    def handle_tier_change
+      new_variants = params[:variants].map { |id| product.tiers.find_by_external_id(id) }.compact
+      perceived_price_cents = params.dig(:purchase, :perceived_price_cents)&.to_i
+
+      subscription.update_current_plan!(
+        new_variants: new_variants,
+        new_price: determine_new_price,
+        perceived_price_cents: perceived_price_cents
+      )
+      subscription.reload
+    end
+
+    def determine_new_price
+      if params[:price_id].present?
+        product.prices.alive.find_by_external_id(params[:price_id])
       else
-        updater_params[:use_existing_card] = true
+        product.default_price
+      end
+    end
+
+    def should_update_payment_method?
+      card_data_handling_mode = CardParamsHelper.get_card_data_handling_mode(params)
+      card_data_handling_mode.present? && card_data_handling_mode != :reuse
+    end
+
+    def update_payment_method
+      card_data_handling_mode = CardParamsHelper.get_card_data_handling_mode(params)
+      card_data_handling_error = CardParamsHelper.check_for_errors(params)
+      chargeable = CardParamsHelper.build_chargeable(params.merge(product_permalink: product.unique_permalink))
+
+      if card_data_handling_error.present?
+        raise Subscription::UpdateFailed, card_data_handling_error.is_card_error? ?
+          PurchaseErrorCode.customer_error_message(card_data_handling_error.error_message) :
+          "There is a temporary problem, please try again (your card was not charged)."
       end
 
-      updater_params
+      unless chargeable.present?
+        raise Subscription::UpdateFailed, "We couldn't process your card. Try again or use a different card."
+      end
+
+      credit_card = CreditCard.create(chargeable, card_data_handling_mode, buyer)
+
+      unless credit_card.errors.empty?
+        raise Subscription::UpdateFailed, credit_card.errors.messages[:base].first
+      end
+
+      subscription.credit_card = credit_card
+      subscription.save!
     end
 
-    def perceived_price
-      params.dig(:purchase, :perceived_price_cents)&.to_i ||
-        subscription.current_subscription_price_cents
+    def should_charge?
+      !within_billing_period? && !subscription.in_free_trial?
     end
 
-    def default_variant_ids
-      subscription.original_purchase.variant_attributes.map(&:external_id)
+    def within_billing_period?
+      subscription.end_time_of_last_paid_period&.future? ||
+        subscription.free_trial_ends_at&.future?
     end
 
-    def merged_card_params
-      @merged_card_params ||= params.slice(*CARD_PARAM_KEYS)
-        .merge((params[:purchase] || {}).slice(*CARD_PARAM_KEYS))
-    end
+    def charge_subscription
+      perceived_price_cents = params.dig(:purchase, :perceived_price_cents)&.to_i ||
+                              subscription.current_subscription_price_cents
 
-    def adapt_result(result)
-      if result[:success]
-        adapted = {
+      purchase = subscription.charge!(
+        override_params: {
+          perceived_price_cents: perceived_price_cents,
+          browser_guid: params.dig(:purchase, :browser_guid)
+        },
+        off_session: false
+      )
+
+      if purchase.successful? || purchase.test_successful?
+        { success: true, purchase: purchase }
+      elsif purchase.in_progress? && purchase.charge_intent&.requires_action?
+        {
           success: true,
-          restarted_subscription: true,
-          subscription: subscription,
-          message: result[:success_message] || "Your membership has been restarted!"
+          requires_card_action: true,
+          purchase: purchase,
+          client_secret: purchase.charge_intent.client_secret,
+          stripe_connect_account_id: purchase.merchant_account.is_a_stripe_connect_account? ?
+            purchase.merchant_account.charge_processor_merchant_id : nil
         }
-
-        if result[:requires_card_action]
-          adapted[:requires_card_action] = true
-          adapted[:client_secret] = result[:client_secret]
-        end
-
-        adapted
       else
-        { success: false, error_message: result[:error_message] }
+        # Raise exception to trigger transaction rollback before marking subscription as failed
+        # This ensures tier changes, payment method updates, and resubscribe! are all rolled back
+        error_message = purchase.errors.full_messages.first || purchase.error_code || "Payment failed. Please try again."
+        raise ChargeFailed.new(error_message, purchase: purchase)
       end
+    end
+
+    def success_result
+      {
+        success: true,
+        restarted_subscription: true,
+        subscription: subscription,
+        purchase: new_purchase,
+        message: "Your membership has been restarted!"
+      }
+    end
+
+    def error_result(message)
+      { success: false, error_message: message }
     end
 end
--- a/spec/services/subscription/restart_at_checkout_service_spec.rb
+++ b/spec/services/subscription/restart_at_checkout_service_spec.rb
@@ -14,8 +14,7 @@ describe Subscription::RestartAtCheckoutService do
         perceived_price_cents: product.price_cents,
         browser_guid: browser_guid
       },
-      price_id: product.prices.alive.first.external_id,
-      variants: product.tiers.map(&:external_id)
+      price_id: product.prices.alive.first.external_id
     }
   end
 
@@ -35,12 +34,6 @@ describe Subscription::RestartAtCheckoutService do
   end
 
   describe "#perform" do
-    let(:updater_service) { instance_double(Subscription::UpdaterService) }
-
-    before do
-      allow(Subscription::UpdaterService).to receive(:new).and_return(updater_service)
-    end
-
     context "with a cancelled subscription that can be restarted" do
       let!(:subscription) do
         create_subscription_for_product(
@@ -53,9 +46,7 @@ describe Subscription::RestartAtCheckoutService do
         )
       end
 
-      it "delegates to UpdaterService with correct params" do
-        allow(updater_service).to receive(:perform).and_return({ success: true, success_message: "Membership restarted" })
-
+      it "restarts the subscription" do
         service = described_class.new(
           subscription: subscription,
           product: product,
@@ -67,22 +58,49 @@ describe Subscription::RestartAtCheckoutService do
 
         expect(result[:success]).to be true
         expect(result[:restarted_subscription]).to be true
-        expect(result[:message]).to eq("Membership restarted")
 
-        expect(Subscription::UpdaterService).to have_received(:new).with(
+        subscription.reload
+        expect(subscription.cancelled_at).to be_nil
+        expect(subscription.deactivated_at).to be_nil
+        expect(subscription.cancelled_by_buyer).to be false
+      end
+
+      it "sends restart notifications" do
+        expect_any_instance_of(Subscription).to receive(:send_restart_notifications!)
+
+        service = described_class.new(
           subscription: subscription,
-          params: hash_including(
-            variants: product.tiers.map(&:external_id),
-            price_id: product.prices.alive.first.external_id,
-            quantity: 1,
-            perceived_price_cents: product.price_cents,
-            perceived_upgrade_price_cents: product.price_cents,
-            use_existing_card: true
-          ),
-          logged_in_user: buyer,
-          gumroad_guid: browser_guid,
-          remote_ip: nil
+          product: product,
+          params: base_params,
+          buyer: buyer
         )
+
+        service.perform
+      end
+
+      context "when subscription is within billing period (pending cancellation)" do
+        before do
+          subscription.update!(
+            cancelled_at: 1.month.from_now,
+            deactivated_at: nil
+          )
+          allow_any_instance_of(Subscription).to receive(:end_time_of_last_paid_period).and_return(1.week.from_now)
+        end
+
+        it "does not charge the user" do
+          expect_any_instance_of(Subscription).not_to receive(:charge!)
+
+          service = described_class.new(
+            subscription: subscription,
+            product: product,
+            params: base_params,
+            buyer: buyer
+          )
+
+          result = service.perform
+          expect(result[:success]).to be true
+          expect(result[:purchase]).to be_nil
+        end
       end
     end
 
@@ -97,9 +115,7 @@ describe Subscription::RestartAtCheckoutService do
         )
       end
 
-      it "delegates to UpdaterService and returns success" do
-        allow(updater_service).to receive(:perform).and_return({ success: true, success_message: "Membership restarted" })
-
+      it "restarts the subscription and clears the failed status" do
         service = described_class.new(
           subscription: subscription,
           product: product,
@@ -110,36 +126,10 @@ describe Subscription::RestartAtCheckoutService do
         result = service.perform
 
         expect(result[:success]).to be true
-        expect(result[:restarted_subscription]).to be true
-      end
-    end
 
-    context "when UpdaterService returns an error" do
-      let!(:subscription) do
-        create_subscription_for_product(
-          product: product,
-          purchaser: buyer,
-          email: email,
-          cancelled_at: 1.day.ago,
-          cancelled_by_buyer: true,
-          deactivated_at: 1.day.ago
-        )
-      end
-
-      it "passes through the error" do
-        allow(updater_service).to receive(:perform).and_return({ success: false, error_message: "This subscription cannot be restarted." })
-
-        service = described_class.new(
-          subscription: subscription,
-          product: product,
-          params: base_params,
-          buyer: buyer
-        )
-
-        result = service.perform
-
-        expect(result[:success]).to be false
-        expect(result[:error_message]).to eq("This subscription cannot be restarted.")
+        subscription.reload
+        expect(subscription.failed_at).to be_nil
+        expect(subscription.deactivated_at).to be_nil
       end
     end
 
@@ -156,9 +146,7 @@ describe Subscription::RestartAtCheckoutService do
         )
       end
 
-      it "returns an error from UpdaterService" do
-        allow(updater_service).to receive(:perform).and_return({ success: false, error_message: "This subscription cannot be restarted." })
-
+      it "returns an error" do
         service = described_class.new(
           subscription: subscription,
           product: product,
@@ -185,11 +173,11 @@ describe Subscription::RestartAtCheckoutService do
         )
       end
 
-      before { product.update!(deleted_at: 1.hour.ago) }
+      before do
+        product.update!(deleted_at: 1.hour.ago)
+      end
 
-      it "returns an error from UpdaterService" do
-        allow(updater_service).to receive(:perform).and_return({ success: false, error_message: "This subscription cannot be restarted." })
-
+      it "returns an error" do
         service = described_class.new(
           subscription: subscription,
           product: product,
@@ -201,112 +189,6 @@ describe Subscription::RestartAtCheckoutService do
 
         expect(result[:success]).to be false
         expect(result[:error_message]).to eq("This subscription cannot be restarted.")
-      end
-    end
-
-    context "when UpdaterService requires card action (SCA)" do
-      let!(:subscription) do
-        create_subscription_for_product(
-          product: product,
-          purchaser: buyer,
-          email: email,
-          cancelled_at: 1.day.ago,
-          cancelled_by_buyer: true,
-          deactivated_at: 1.day.ago
-        )
-      end
-
-      it "passes through SCA details" do
-        allow(updater_service).to receive(:perform).and_return({
-          success: true,
-          requires_card_action: true,
-          client_secret: "pi_secret_123"
-        })
-
-        service = described_class.new(
-          subscription: subscription,
-          product: product,
-          params: base_params,
-          buyer: buyer
-        )
-
-        result = service.perform
-
-        expect(result[:success]).to be true
-        expect(result[:requires_card_action]).to be true
-        expect(result[:client_secret]).to eq("pi_secret_123")
-      end
-    end
-
-    context "param transformation" do
-      let!(:subscription) do
-        create_subscription_for_product(
-          product: product,
-          purchaser: buyer,
-          email: email,
-          cancelled_at: 1.day.ago,
-          cancelled_by_buyer: true,
-          deactivated_at: 1.day.ago
-        )
-      end
-
-      it "defaults variants and price_id from subscription when not provided" do
-        allow(updater_service).to receive(:perform).and_return({ success: true })
-
-        params_without_variants = {
-          purchase: {
-            email: email,
-            perceived_price_cents: product.price_cents,
-            browser_guid: browser_guid
-          }
-        }
-
-        described_class.new(
-          subscription: subscription,
-          product: product,
-          params: params_without_variants,
-          buyer: buyer
-        ).perform
-
-        expect(Subscription::UpdaterService).to have_received(:new).with(
-          subscription: subscription,
-          params: hash_including(
-            variants: subscription.original_purchase.variant_attributes.map(&:external_id),
-            price_id: subscription.price&.external_id,
-            use_existing_card: true
-          ),
-          logged_in_user: buyer,
-          gumroad_guid: browser_guid,
-          remote_ip: nil
-        )
-      end
-
-      it "falls back to subscription price when perceived_price_cents not provided" do
-        allow(updater_service).to receive(:perform).and_return({ success: true })
-
-        params_without_price = {
-          purchase: { email: email, browser_guid: browser_guid },
-          variants: product.tiers.map(&:external_id),
-          price_id: product.prices.alive.first.external_id
-        }
-
-        described_class.new(
-          subscription: subscription,
-          product: product,
-          params: params_without_price,
-          buyer: buyer
-        ).perform
-
-        expect(Subscription::UpdaterService).to have_received(:new).with(
-          subscription: subscription,
-          params: hash_including(
-            perceived_price_cents: subscription.current_subscription_price_cents,
-            perceived_upgrade_price_cents: subscription.current_subscription_price_cents
-          ),
-          logged_in_user: buyer,
-          gumroad_guid: browser_guid,
-          remote_ip: nil
-        )
       end
     end
   end


app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      credit_card: funding_card.present? ? {
+        id: funding_card.id,
+        visual: funding_card.visual,
+        card_type: funding_card.card_type,
+        expiry_month: funding_card.expiry_month,
+        expiry_year: funding_card.expiry_year
+      } : nil,
+      show_banner: !current_seller.dismissed_refund_payment_method_banner? && current_seller.refund_funding_credit_card.blank?
+    }
+  end
```
yashranaway, line 20: Returns current funding card status and whether to show the setup banner

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      success: true,
+      credit_card: {
+        id: credit_card.id,
+        visual: credit_card.visual,
+        card_type: credit_card.card_type,
+        expiry_month: credit_card.expiry_month,
+        expiry_year: credit_card.expiry_year
+      }
+    }
+  end
```
yashranaway, line 54: Saves a new backup card via Stripe. Uses existing `CardParamsHelper` for validation.

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    render json: { success: true }
+  end
+
+  def dismiss_banner
+    current_seller.update!(dismissed_refund_payment_method_banner: true)
+
+    render json: { success: true }
+  end
```
yashranaway, line 66: Hides the promo banner permanently (stored in user flags)

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  def destroy
+    current_seller.update!(
+      refund_funding_credit_card: nil,
+      refund_funding_card_name: nil
+    )
+
+    render json: { success: true }
+  end
```
yashranaway, line 60: Removes the backup card. Banner will reappear on Customers page.

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+          id: result.balance_top_up.external_id,
+          amount_cents: result.balance_top_up.amount_cents,
+          formatted_amount: result.balance_top_up.formatted_amount
+        }
+      }
+    else
+      render json: { success: false, error: result.error_message }, status: :unprocessable_entity
+    end
+  end
+
```
yashranaway: Test endpoint to verify card works. Charges the card and adds to balance.

app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
yashranaway, line 1: Promo banner on Customers page. Dismissal is persisted to backend , reappears if seller removes their backup card.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
yashranaway, line 1: Form for adding/editing the seller's backup payment card for refund funding. Uses Stripe Elements for PCI-compliant card input.

app/models/balance_top_up.rb
yashranaway, line 1: Tracks charges to seller's backup card when their balance can't cover a refund. Linked to the purchase being refunded and the resulting credit added to balance.

app/modules/purchase/refundable.rb
```diff
@@ -45,8 +45,22 @@ def refund_and_save!(refunding_user_id, amount_cents: nil, is_for_fraud: false)
...
+            ).perform
+
+            if !top_up_result.success?
+              errors.add :base, top_up_result.error_message || "Could not charge your card to cover the refund shortfall."
+              return false
+            end
+          else
+            errors.add :base, "Your balance is insufficient to process this refund."
+            return false
+          end
```
yashranaway, line 64: If seller's balance can't cover refund, charge their backup card for the difference

app/services/balance_top_up/charge_service.rb
```diff
@@ -0,0 +1,127 @@
...
+  MAXIMUM_TOP_UP_AMOUNT_CENTS = 1_000_000
+
+  Result = Struct.new(:success?, :balance_top_up, :error_message, keyword_init: true)
+
+  def initialize(user:, amount_cents:, credit_card: nil, purchase: nil)
+    @user = user
+    @amount_cents = amount_cents
+    @credit_card = credit_card || user.refund_funding_credit_card
+    @purchase = purchase
+  end
```
yashranaway: Handles charging the seller's backup card and crediting their balance. Is clalled by `Purchase::Refundable` when balance is insufficient for a refund.

app/services/balance_top_up/charge_service.rb
```diff
@@ -0,0 +1,127 @@
...
+    BalanceTopUp.create!(
+      user: @user,
+      credit_card: @credit_card,
+      purchase: @purchase,
+      amount_cents: @amount_cents,
+      processor: StripeChargeProcessor.charge_processor_id,
+      state: "pending"
+    )
+  end
+
```
yashranaway: Main entry point. Validates, charges Stripe, creates credit on success.

app/services/balance_top_up/charge_service.rb
```diff
@@ -0,0 +1,127 @@
...
+  def charge_credit_card(balance_top_up)
+    balance_top_up.mark_processing!
```
yashranaway: On success: create Credit to add funds to seller's balance, send email

yashranaway:
@binary-koan Could you please review this PR 🙏

yashranaway:
@binary-koan Friendly ping on this 😓 whenever you have a moment, I’d appreciate a review.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,315 @@
...
+  React.useEffect(() => {
+    setStripePromise(getStripeInstance());
+  }, []);
+
+  if (!stripePromise) {
+    return <div>Loading...</div>;
+  }
+
+  return (
+    <Elements stripe={stripePromise}>
```
binary-koan: Can we just use `StripeElementsProvider` from `app/javascript/components/Checkout/CreditCardInput.tsx`?

app/assets/images/credit-card-outline.svg
binary-koan, line 1: Do we need this as well as the icon?

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+    render json: { success: true }
+  end
+
+  def dismiss_banner
+    current_seller.update!(dismissed_refund_payment_method_banner: true)
+
+    render json: { success: true }
+  end
+
+  def test_charge
```
binary-koan: Is this used anywhere? I see a route and policy but no other references

app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
```diff
@@ -0,0 +1,83 @@
...
+      assertResponseError(e);
+    }
+
+    setIsDismissing(false);
+  });
+
+  if (!isVisible) return null;
+
+  return (
+    <div
```
binary-koan: This should be `<Alert variant="accent">` I think, no need to customize it this much to match the design

app/views/contacting_creator_mailer/balance_top_up_confirmation.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= header_section("Your Gumroad balance has been topped up") %>
```
binary-koan: Does this email copy come from somewhere? The one in https://github.com/antiwork/gumroad/pull/2839 makes more sense to me with the current behaviour, but not sure if either of them are based on a design?

app/services/balance_top_up/charge_service.rb
```diff
@@ -0,0 +1,127 @@
...
+    charge_result = charge_credit_card(balance_top_up)
+
+    if charge_result[:success]
+      complete_top_up(balance_top_up, charge_result)
+    else
+      fail_top_up(balance_top_up, charge_result[:error_message])
+    end
+  rescue StandardError => e
+    Rails.logger.error("BalanceTopUp::ChargeService error: #{e.message}")
+    Rails.logger.error(e.backtrace.join("\n"))
```
binary-koan: Is this pattern copied from somewhere? I believe it won't log to Bugsnag so we won't have visibility on errors

binary-koan:
@hchhabra _a Mind taking a look at the payment logic here when you have time? You're probably better placed to review that than me. Also curious what you think of adding a model (here) vs reusing `Credit` (e.g. https://github.com/antiwork/gumroad/pull/2839)

Happy to review the rest once the backend is going in the right direction, up to you

yashranaway:
Just for Context,
 
  Decided the model approach which follows the codebase patterns like:                                                                                                                                                  
  1. Financial operations use dedicated models with state machines (like Dispute, ServiceCharge, Payment)                                                                                                        
  2. The Credit model uses specific associations for each credit type - `balance_top_up` fits this pattern, while the other PR uses `crediting_user` as a workaround (there's no actual admin crediting)                    
  3. A dedicated model lets us track failed charges and link to the triggering purchase                                                                                                                          
                                                                                                                                                                                                                 
  Happy to discuss if you see benefits to the simpler approach that outweigh these considerations.

pegasus1134:
> Just for Context,
> 
> Decided the model approach which follows the codebase patterns like:
> 
> 1. Financial operations use dedicated models with state machines (like Dispute, ServiceCharge, Payment)
> 2. The Credit model uses specific associations for each credit type - `balance_top_up` fits this pattern, while the other PR uses `crediting_user` as a workaround (there's no actual admin crediting)
> 3. A dedicated model lets us track failed charges and link to the triggering purchase
> 
> Happy to discuss if you see benefits to the simpler approach that outweigh these considerations.

@yashranaway Appreciate the detailed reasoning. A few counterpoints since you referenced some exclusive caveats of the credit model I used:                                                                                                                                                   
                                                                                                                                                                                                                         
  On state machines for financial operations:                                                                                                                                                                            
  Dispute and ServiceCharge handle async operations spanning days/weeks (evidence submission, resolution windows). A Stripe charge is synchronous - you call it, you get success or failure in milliseconds. State       
  machines solve async coordination problems; applying one to a synchronous call adds infrastructure without solving an actual problem.                                                                                  
                                                                                                                                                                                                                         
  On Credit associations:                                                                                                                                                                                                
  The Credit model already follows the pattern you're describing - each credit type has its own association and factory method:                                                                                          
  - chargebacked_purchase → create_for_dispute_won!                                                                                                                                                                      
  - returned_payment → create_for_returned_payment_difference!                                                                                                                                                           
  - refund → refund-related credits                                                                                                                                                                                      
                                                                                                                                                                                                                         
  Adding refund_funding_purchase with create_for_refund_funding! follows this exact pattern. No workarounds needed.                                                                                                      
                                                                                                                                                                                                                         
  On tracking failed charges:                                                                                                                                                                                            
  Stripe maintains a complete audit trail of every charge attempt. Persisting failures to our database duplicates data already available via the Stripe API. The successful charge ID stored in Credit's json_data       
  provides the link we need.                                                                                                                                                                                             
                                                                                                                                                                                                                         
  Happy to defer to @hchhabra and @binary-koan on which approach fits the codebase better.

yashranaway:
@hchhabra light ping on this 🙏🏼

hchhabra:
Agree with @binary-koan 's [comment](https://github.com/antiwork/gumroad/pull/2839#issuecomment-3769582531) on the other PR about using the simpler approach and the reasonings [here](https://github.com/antiwork/gumroad/pull/2839#issuecomment-3769494715) and [here](https://github.com/antiwork/gumroad/pull/2425#issuecomment-3769556220), @yashranaway _a. We are ultimately creating a `credit` here anyways to update the balance, and the new table doesn't add much value on top of it, other than recording the `processor_transaction_id` which can be done on the `credit` record. The error message (card failure or otherwise) will be immediately displayed to the creator, and both success and failure are terminal states here with no further changes in the state possible in the future (unlike disputes, payments etc.). Although, we should associate the `credit` with the `refund` instead of a `crediting_user` as done there, and associate it with the used `credit_card` as well. cc @pegasus1134

db/migrate/20251221000004_add_refund_funding_fields_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddRefundFundingFieldsToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :refund_funding_card_name, :string
```
hchhabra: We should save the card holder name on the `credit_card` instead of on the `user` I think.

pegasus1134:
> Agree with @binary-koan 's [comment](https://github.com/antiwork/gumroad/pull/2839#issuecomment-3769582531) on the other PR about using the simpler approach and the reasonings [here](https://github.com/antiwork/gumroad/pull/2839#issuecomment-3769494715) and [here](https://github.com/antiwork/gumroad/pull/2425#issuecomment-3769556220), @yashranaway _a. We are ultimately creating a `credit` here anyways to update the balance, and the new table doesn't add much value on top of it, other than recording the `processor_transaction_id` which can be done on the `credit` record. The error message (card failure or otherwise) will be immediately displayed to the creator, and both success and failure are terminal states here with no further changes in the state possible in the future (unlike disputes, payments etc.). Although, we should associate the `credit` with the `refund` instead of a `crediting_user` as done there, and associate it with the used `credit_card` as well. cc @pegasus1134

Thanks for the feedback @hchhabra. Will re-open a PR since you prefer my approach. Will add in your suggestion as well as some of the fixes @binary-koan requested for and get it ready for merge ASAP.

yashranaway:
Hey @pegasus1134, I agree your approach is preferred (no state machine, simpler credit-based flow). That said, a few aspects from my PR were also suggested by @hchhabra, such as associating Credit with refund_funding_purchase + credit_card, and storing holder_name on CreditCard.

I've already updated my branch with these changes of a combined approach , and tests are passing. Combining both approaches seems like the best path forward. I'd be happy to work together on this if that works for you. If this ends up being a shared effort, happy to split the bounty (if it gets merged).

pegasus1134:
> Hey @pegasus1134, I agree your approach is preferred (no state machine, simpler credit-based flow). That said, a few aspects from my PR were also suggested by @hchhabra, such as associating Credit with refund_funding_purchase + credit_card, and storing holder_name on CreditCard.
> 
> I've already updated my branch with these changes of a combined approach , and tests are passing. Combining both approaches seems like the best path forward. I'd be happy to work together on this if that works for you. If this ends up being a shared effort, happy to split the bounty (if it gets merged).

Thanks for the effort with taking up my approach and merging it with your code. I will be opening my PR and pushing the requested changes with my approach. I have no issues with a split bounty. Let's leave that to the maintainers to decide! The focus is on merging the code and getting it up and running now.

app/models/credit.rb
```diff
@@ -14,6 +14,8 @@ class Credit < ApplicationRecord
   belongs_to :financing_paydown_purchase, class_name: "Purchase", optional: true
   belongs_to :fee_retention_refund, class_name: "Refund", optional: true
   belongs_to :backtax_agreement, optional: true
+  belongs_to :refund_funding_purchase, class_name: "Purchase", optional: true
+  belongs_to :credit_card, optional: true
```
yashranaway, line 18: Added these two associations per feedback. The `refund_funding_purchase` links the credit to the purchase that triggered the backup card charge (not a generic `crediting_user` workaround). The `credit_card` association tracks which card was actually charged  useful for audit trails and displaying "charged **** 4242" in emails.

app/models/credit.rb
```diff
@@ -24,6 +26,8 @@ class Credit < ApplicationRecord
   validate :validate_associated_entity
 
   attr_json_data_accessor :stripe_loan_paydown_id
+  attr_json_data_accessor :refund_funding_processor_transaction_id
+  attr_json_data_accessor :refund_funding_processor_payment_intent_id
```
yashranaway, line 30: Storing the Stripe charge ID and payment intent ID in `json_data` rather than dedicated columns. This follows the existing pattern (see `stripe_loan_paydown_id`) and avoids adding more columns for what's essentially metadata.

app/models/credit.rb
```diff
@@ -51,6 +55,35 @@ def self.create_for_credit!(user:, amount_cents:, crediting_user:)
...
+      credit:,
+      issued_amount: balance_transaction_amount,
+      holding_amount: balance_transaction_amount
+    )
+
+    credit.balance = balance_transaction.balance
+    credit.save!
+    credit
+  end
+
```
yashranaway, line 86: The `create_for_refund_funding!` method follows the same pattern as other credit types (`create_for_dispute_won!`, `create_for_returned_payment_difference!`, etc.). Each credit type gets its own factory method with specific associations - this is cleaner than trying to make one generic method handle all cases.

app/models/credit.rb
```diff
@@ -433,14 +466,17 @@ def add_comment
       comment_attrs[:content] = "issued adjustment due to currency conversion differences when payment #{returned_payment.id} returned."
     elsif refund
       comment_attrs[:author_name] = "AutoCredit PayPal Connect VAT refund (#{refund.purchase.id})"
+    elsif refund_funding_purchase
+      comment_attrs[:author_name] = "Refund Funding"
+      comment_attrs[:content] = "credited #{formatted_dollar_amount(amount_cents)} from backup card charge to cover refund."
```
yashranaway, line 471: When a credit is created, it auto-generates a comment on the user's account. This gives sellers visibility into what happened without needing to dig through transaction logs.

app/models/credit.rb
```diff
@@ -433,14 +466,17 @@ def add_comment
...
     end
     user.comments.create(comment_attrs)
   end
 
   private
     def validate_associated_entity
-      return if crediting_user || chargebacked_purchase || returned_payment || refund || financing_paydown_purchase || fee_retention_refund || backtax_agreement
+      return if crediting_user || chargebacked_purchase || returned_payment || refund || financing_paydown_purchase || fee_retention_refund || backtax_agreement || refund_funding_purchase
```
yashranaway, line 478: Added `refund_funding_purchase` to the list of valid associated entities. A credit must belong to *something* - this ensures we don't accidentally create orphaned credits.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+  MAXIMUM_CHARGE_CENTS = 1_000_000
+
+  Result = Struct.new(:success?, :credit, :error_message, keyword_init: true)
+
+  def initialize(user:, amount_cents:, purchase:)
+    @user = user
+    @amount_cents = amount_cents
+    @purchase = purchase
+    @credit_card = user.refund_funding_credit_card
+  end
```
yashranaway, line 16: Removed the optional `credit_card` parameter that the old service had. For refund funding, we always use the user's configured backup card - there's no use case for passing a different one. Simpler interface, fewer ways to misuse it.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    credit = create_credit(charge_result)
+    send_confirmation_email(credit)
+
+    Result.new(success?: true, credit: credit)
+  rescue StandardError => e
+    Rails.logger.error("RefundFundingChargeService error: #{e.message}")
+    Rails.logger.error(e.backtrace.join("\n"))
+    error("An unexpected error occurred while processing your backup card charge.")
+  end
```
yashranaway, line 34: No state machine transitions, no separate "processing" state. Just: validate → charge → create credit → send email. If any step fails, we return an error immediately. The old approach created a `BalanceTopUp` record first, then transitioned it through states, but why persist failure states when we can just... not persist failures?

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+  rescue Stripe::CardError => e
+    { success: false, error: e.message }
+  rescue Stripe::InvalidRequestError => e
+    Rails.logger.error("RefundFundingChargeService invalid request: #{e.message}")
+    { success: false, error: "Invalid payment request." }
+  rescue Stripe::StripeError => e
+    Rails.logger.error("RefundFundingChargeService Stripe error: #{e.message}")
+    { success: false, error: "Payment processor error. Please try again." }
+  end
```
yashranaway, line 75: The idempotency key prevents duplicate charges if there's a network hiccup and the request gets retried. Including `Time.current.to_i` allows retrying after a genuine failure (new timestamp = new idempotency key).

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+  def create_credit(charge_result)
+    Credit.create_for_refund_funding!(
+      user: @user,
+      amount_cents: @amount_cents,
+      purchase: @purchase,
+      credit_card: @credit_card,
+      processor_transaction_id: charge_result[:charge_id],
+      processor_payment_intent_id: charge_result[:payment_intent_id]
+    )
+  end
```
yashranaway, line 86: All the Stripe IDs are passed to the credit. If we ever need to investigate a charge or issue a refund of the backup card charge itself, we have everything we need.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def send_confirmation_email(credit)
+    ContactingCreatorMailer.refund_funding_charge_confirmation(credit_id: credit.id).deliver_later(queue: "critical")
+  rescue StandardError => e
+    Rails.logger.error("RefundFundingChargeService email error: #{e.message}")
+  end
+
```
yashranaway, line 94: Wrapped in rescue - if email fails, we still return success. The charge went through, the credit exists, the seller's balance is updated. An email failure is annoying but not critical. We log it and move on.

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+class Settings::RefundFundingController < Sellers::BaseController
+  before_action :authorize
+
+  def show
+    funding_card = current_seller.refund_funding_credit_card
+
+    render json: {
+      enabled: funding_card.present?,
+      name_on_card: funding_card&.holder_name,
```
yashranaway: Changed from `current_seller.refund_funding_card_name` to `funding_card&.holder_name`. The cardholder name belongs with the card, not on the user record. If a seller changes their backup card, we don't want the old name lingering.

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    credit_card = CreditCard.create(chargeable, card_data_handling_mode, current_seller)
+
+    if credit_card.errors.any?
+      return render json: { success: false, error: credit_card.errors.full_messages.join(", ") }, status: :unprocessable_entity
+    end
+
+    credit_card.update!(holder_name: params[:name_on_card])
+    current_seller.update!(refund_funding_credit_card: credit_card)
```
yashranaway: Save the holder name on the card first, then associate the card with the user. Could combine into one transaction, but these are independent operations and keeping them separate is clearer.

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    if funding_card.present? && params[:name_on_card].present?
+      funding_card.update!(holder_name: params[:name_on_card])
+    end
+
+    render json: { success: true }
+  end
+
+  def destroy
+    current_seller.update!(refund_funding_credit_card: nil)
```
yashranaway, line 57: Just nil out the association. We don't delete the CreditCard record itself.  Stripe still has the payment method, and we might want it for audit purposes.

app/controllers/settings/refund_funding_controller.rb
yashranaway, line 1: The old implementation had a `test_charge` endpoint that let sellers charge their card to test the flow. Removed this because (a) it added complexity, (b) it could be abused, and (c) the real test is... trying to refund something. If the card doesn't work, they'll find out then.

app/modules/purchase/refundable.rb
```diff
@@ -45,8 +45,23 @@ def refund_and_save!(refunding_user_id, amount_cents: nil, is_for_fraud: false)
...
+            ).perform
+
+            if !result.success?
+              errors.add :base, result.error_message || "Could not charge your backup card to cover the refund shortfall."
+              return false
+            end
+          else
+            errors.add :base, "Your balance is insufficient to process this refund."
+            return false
+          end
```
yashranaway, line 64: During a refund, if balance is insufficient, we calculate the shortfall and charge the backup card. The `[shortfall, MINIMUM_CHARGE_CENTS].max` ensures we always charge at least $1 (Stripe's minimum). If the shortfall is 50 cents, we charge $1 - better to have a small overage in balance than to hit Stripe's minimum limit.

app/mailers/contacting_creator_mailer.rb
```diff
@@ -532,6 +532,15 @@ def ping_endpoint_failure(user_id, ping_url, response_code)
...
 
+  def refund_funding_charge_confirmation(credit_id:)
+    @credit = Credit.find(credit_id)
+    @seller = @credit.user
+    @amount = Money.new(@credit.amount_cents, :usd).format(no_cents_if_whole: true, symbol: true)
+    @card_visual = @credit.credit_card&.visual || "****"
+    @subject = "Your backup card was charged #{@amount}"
+    mail(to: @seller.form_email, subject: @subject)
+  end
+
```
yashranaway, line 540: Changed from `balance_top_up_confirmation` to `refund_funding_charge_confirmation`. The new name is more specific about what happened. Also added the card visual to the email so sellers know which card was charged.

yashranaway (commented):
@hchhabra _a Adapted the implementation based on your feedback - removed the `BalanceTopUp` state machine in favor of direct `Credit` creation with `refund_funding_purchase` + `credit_card` associations. Also moved `holder_name` to `CreditCard` where it belongs.

Added inline comments below explaining the key changes. Please take a look when you have a moment!

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+class Settings::RefundFundingController < Sellers::BaseController
```
binary-koan:
```suggestion
class Settings::RefundFundingController < Settings::BaseController
```

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    chargeable = CardParamsHelper.build_chargeable(params)
+
+    if chargeable.blank?
+      return render json: { success: false, error: "Invalid card information" }, status: :unprocessable_entity
+    end
+
+    credit_card = CreditCard.create(chargeable, card_data_handling_mode, current_seller)
+
+    if credit_card.errors.any?
```
binary-koan:
```suggestion
    if !credit_card.persisted?
```

(Should be effectively the same, but just in case)

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      return render json: { success: false, error: credit_card.errors.full_messages.join(", ") }, status: :unprocessable_entity
+    end
+
+    credit_card.update!(holder_name: params[:name_on_card])
```
binary-koan: We're actually [removing the name on card field from checkout](https://github.com/antiwork/gumroad/pull/3069), so I think we can get rid of it here too (cc @slavingia)

app/controllers/settings/refund_funding_controller.rb
binary-koan: Just to check, is this what we do for buyer cards as well?

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,314 @@
...
+type Props = {
+  refundPaymentMethod: RefundPaymentMethodProps;
+  isFormDisabled: boolean;
+};
+
+const RefundPaymentMethodForm = ({
+  refundPaymentMethod,
+  isFormDisabled,
+  onSuccess,
+}: Props & { onSuccess: () => void }) => {
```
binary-koan: I think we can just inline the `Props` type

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,314 @@
...
+import { asyncVoid } from "$app/utils/promise";
+import { assertResponseError, request } from "$app/utils/request";
+import { getStripeInstance } from "$app/utils/stripe_loader";
+import { getCssVariable } from "$app/utils/styles";
+
+import { Button } from "$app/components/Button";
+import { Icon } from "$app/components/Icons";
+import { showAlert } from "$app/components/server-components/Alert";
+
+export type RefundPaymentMethodProps = {
```
binary-koan:
```suggestion
export type RefundPaymentMethod = {
```

We don't typically use a suffix for server-provided data

app/mailers/contacting_creator_mailer.rb
```diff
@@ -532,6 +532,15 @@ def ping_endpoint_failure(user_id, ping_url, response_code)
     @subject = "Webhook ping endpoint delivery failed"
   end
 
+  def refund_funding_charge_confirmation(credit_id:)
+    @credit = Credit.find(credit_id)
+    @seller = @credit.user
+    @amount = Money.new(@credit.amount_cents, :usd).format(no_cents_if_whole: true, symbol: true)
+    @card_visual = @credit.credit_card&.visual || "****"
+    @product_name = @credit.refund_funding_purchase&.link&.name
```
binary-koan: I don't think we need all these as separate ivars, we can inline some in the template

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,314 @@
...
+                    Remove card
+                  </Button>
+                </div>
+              ) : null}
+            </div>
+          </div>
+        ) : (
+          <>
+            <div className="grid grid-cols-1 gap-5 md:grid-cols-2">
+              <fieldset>
```
binary-koan: Please use the new form components from #2942 so we don't need to migrate this later

binary-koan:
(just FYI I'll probably close this and ask you to open another PR once I've reviewed the rest as the thread is getting pretty long)

pegasus1134:
Thanks for the detailed review @binary-koan. We'll work together to address the issues and open a new PR if needed. @yashranaway let me know how I can help!

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
yashranaway: Done, replaced the custom `Elements`/`stripePromise` setup with `StripeElementsProvider`.

app/assets/images/credit-card-outline.svg
yashranaway: Nope, removed it. `<Icon name="outline-credit-card" />` handles this already.

app/controllers/settings/refund_funding_controller.rb
yashranaway: removed.

app/views/contacting_creator_mailer/balance_top_up_confirmation.html.erb
yashranaway: Aligned the copy with PR #2839's style. It now reads more naturally and matches the existing email patterns in the codebase.

app/services/balance_top_up/charge_service.rb
yashranaway: Good catch. Added `Bugsnag.notify(e)` to all rescue blocks so we have visibility on Stripe errors and unexpected failures.

app/controllers/settings/refund_funding_controller.rb
yashranaway: Removed entirely, dropped `name_on_card` from the controller, presenter, frontend component, and specs.

app/controllers/settings/refund_funding_controller.rb
yashranaway: Buyer card removal in `CheckoutController` uses a `requires_credit_card?` guard, but that constraint doesn't apply here since the refund funding card is optional. Our approach of simply setting the association to `nil` is the right pattern for this use case.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
yashranaway: Inlined, the type is now defined directly in the component parameters.

app/mailers/contacting_creator_mailer.rb
yashranaway: Inlined `amount`, `card_visual`, and `product_name` as template-level local variables. Only `@credit`, `@seller`, and `@subject` remain as ivars in the mailer.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
yashranaway: Updated to use `Fieldset`, `FieldsetTitle`, `Input`, and `InputGroup` from `components/ui/`.

yashranaway:
@binary-koan thanks for the review! I've addressed/replied all comments  could you take another look when you get a chance?

pegasus1134:
Hey @yashranaway! Manually reviewed the latest changes — the updates addressing @binary-koan's feedback look good to go (StripeElementsProvider import, Alert variant, Bugsnag.notify, removed test endpoint, etc.).

However, I noticed two issues in schema.rb that should be fixed before merge:

1. **Schema version regresses** — goes from `2026_11_19_011937` → `2026_11_19_011936`. This will cause problems for anyone who already has migration 011937 locally.

2. **Unrelated change** — `default_offer_code_id` in the links table changed from `integer` to `bigint`. This isn't part of the backup payment method feature and looks like merge conflict residue.

A clean rebase on main and regenerating the schema (`rails db:migrate:reset` or similar) should fix both. Let me know if you need a hand!

pegasus1134:
One more thing — noticed the email rescue block in `RefundFundingChargeService#send_confirmation_email` is missing `Bugsnag.notify(e)`:

```ruby
def send_confirmation_email(credit)
  ContactingCreatorMailer.refund_funding_charge_confirmation(credit_id: credit.id).deliver_later(queue: "critical")
rescue StandardError => e
  Rails.logger.error("RefundFundingChargeService email error: #{e.message}")
  # Missing: Bugsnag.notify(e)
end
```

The main charge rescue blocks have Bugsnag, but this one was missed. Quick fix!

yashranaway:
Wrong branch my bad.

pegasus1134:
Hey @yashranaway, did another pass on the latest changes and compared against my implementation — found a few more things worth addressing:

**1. `holder_name` column is added but never written to**

The migration adds `holder_name` to `credit_cards`, but the `create` action never populates it. The name-on-card value from the form isn't being stored. Should be something like:

```ruby
# In create action, after credit_card is created:
credit_card.update!(holder_name: params[:name_on_card])
```

And the `show` response should return it so the frontend can display the saved name.

**2. `destroy` only returns `{ success: true }` — frontend gets stale state**

The `show` and `create` endpoints return full card/banner state, but `destroy` only returns `{ success: true }`. The frontend has to assume the state changed rather than getting confirmation. Extracting a shared `refund_funding_data` method (like the show response) and returning it from `create` and `destroy` too would keep the frontend in sync.

**3. No confirmation before card removal**

Clicking remove immediately deletes the card with no confirmation prompt. A simple `confirm("Are you sure?")` before the API call would prevent accidental removals.

**4. Silent failure on banner dismiss**

In `RefundPaymentMethodBanner.tsx`, if the dismiss API call fails, the catch block only calls `assertResponseError(e)` — user gets no feedback that the dismiss didn't go through.

**5. Race condition in refund + charge flow**

In `Purchase::Refundable`, balance is read with `seller.unpaid_balance_cents` and then the card is charged separately. If two refunds run concurrently for the same seller, both could read the same balance and double-charge the backup card. Wrapping the balance read + charge in `seller.with_lock { ... }` would prevent this.

**6. Banner reappear logic doesn't match PR description**

PR description says "If they later remove their card, the banner reappears" but the show logic is:
```ruby
!current_seller.dismissed_refund_payment_method_banner? && current_seller.refund_funding_credit_card.blank?
```
Once dismissed, the banner never reappears regardless of card status. Either update the description or change the dismiss logic to reset when the card is removed.

Happy to help fix any of these — let me know!

db/migrate/20260130000002_add_holder_name_to_credit_cards.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddHolderNameToCreditCards < ActiveRecord::Migration[7.1]
+  def change
+    add_column :credit_cards, :holder_name, :string
```
binary-koan, line 5: Don't think we need this any more?

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  {savedCard.expiry_month.toString().padStart(2, "0")}/{savedCard.expiry_year.toString().slice(-2)}
+                </span>
+              </InputGroup>
+            </Fieldset>
+            {!isFormDisabled ? (
+              <div className="flex gap-4">
+                <Button outline onClick={() => setIsEditing(true)} disabled={isSubmitting}>
+                  Change card
+                </Button>
+                <Button outline color="danger" onClick={handleRemove} disabled={isSubmitting}>
```
binary-koan, line 155: I think we should show a `<Modal>` before deleting, similar to how we do it in other places (products and wishlists at least off the top of my head)

app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
```diff
@@ -0,0 +1,63 @@
...
+        accept: "json",
+      });
+
+      const result = cast<{ success: boolean }>(await response.json());
+
+      if (result.success) {
+        setIsVisible(false);
+      }
+    } catch (e) {
+      assertResponseError(e);
```
binary-koan, line 33: Showing a generic "something went wrong" alert might be nice

app/modules/purchase/refundable.rb
binary-koan: The potential race condition here is if the user has $90 balance and tries to refund a bunch of $100 purchases simultaneously, it will charge $10 to the refund card a bunch of times but then refund $100 a bunch of times making their balance negative

I don't think we should lock through the refund funding API request, but maybe to make this harder to exploit we can double check again that `amount_cents_to_refund > seller.reload.unpaid_balance_cents` after the charge is complete?

app/mailers/contacting_creator_mailer.rb
```diff
@@ -532,6 +532,12 @@ def ping_endpoint_failure(user_id, ping_url, response_code)
     @subject = "Webhook ping endpoint delivery failed"
   end
 
+  def refund_funding_charge_confirmation(credit_id:)
+    @credit = Credit.find(credit_id)
+    @seller = @credit.user
+    @subject = "Your backup card was charged #{Money.new(@credit.amount_cents, :usd).format(no_cents_if_whole: true, symbol: true)}"
```
binary-koan, line 538: Ah I didn't notice the amount is in the subject as well - the logic is nontrivial so an ivar for that would make sense

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,98 @@
...
+  def initialize(user:, amount_cents:, purchase:)
+    @user = user
+    @amount_cents = amount_cents
+    @purchase = purchase
+    @credit_card = user.refund_funding_credit_card
+  end
+
+  def perform
+    return error("No backup payment method configured.") if @credit_card.blank?
```
binary-koan, line 19: For new services I'd much prefer to use `attr_reader`s rather than ivars directly. It makes it easy to refactor into a method if we ever want to

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,98 @@
...
+    if payment_intent.status == "succeeded"
+      {
+        success: true,
+        payment_intent_id: payment_intent.id,
+        charge_id: payment_intent.latest_charge
+      }
+    else
+      { success: false, error: "Payment was not successful. Status: #{payment_intent.status}" }
+    end
+  rescue Stripe::CardError => e
```
binary-koan, line 67: Could we use `with_stripe_error_handler` instead of custom error handling logic? I think this is close enough to charge processing that it would make sense to reuse

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,98 @@
...
+    error("An unexpected error occurred while processing your backup card charge.")
+  end
+
+  private
+
+  def charge_credit_card
+    payment_intent = Stripe::PaymentIntent.create(
+      {
+        amount: @amount_cents,
+        currency: "usd",
```
binary-koan, line 42:
```suggestion
        currency: Currency::USD,
```

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "shared_examples/authentication_required"
+
+describe Settings::RefundFundingController do
+  let(:seller) { create(:user) }
+  let(:credit_card) do
+    CreditCard.new(
```
binary-koan, line 9: Can we use a factory for this for consistency?

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      before { seller.update!(refund_funding_credit_card: credit_card) }
+
+      it "returns enabled as true with card details" do
+        get :show, format: :json
+
+        expect(response).to be_successful
+        json = JSON.parse(response.body)
+        expect(json["enabled"]).to be true
+        expect(json["credit_card"]["visual"]).to eq("**** **** **** 4242")
+        expect(json["show_banner"]).to be false
```
binary-koan, line 48: Why do we return `show_banner`? I don't see it used

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    end
+  end
+
+  describe "POST #create" do
+    it_behaves_like "authentication required for action", :post, :create
+
+    context "when chargeable cannot be built" do
```
binary-koan, line 56: What about when it can be built? (does needing VCR make that hard to implement?)

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        expect(json["success"]).to be false
+        expect(json["error"]).to eq("Invalid card information")
+      end
+    end
+  end
+
+  describe "DELETE #destroy" do
+    it_behaves_like "authentication required for action", :delete, :destroy
+
+    context "when funding card exists" do
```
binary-koan, line 77: Would be nice to cover a few more edge cases (when card doesn't exist in #destroy, when the banner is already dismissed in #dismiss_banner)

spec/models/purchase/refund_with_backup_card_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+    ).tap(&:save!)
+  end
+  let(:purchase) do
+    create(:purchase_in_progress,
+           link: product,
+           seller:,
+           purchaser: buyer,
+           price_cents: 2000,
+           total_transaction_cents: 2000).tap do |p|
+      p.update_columns(succeeded_at: 1.day.ago, purchase_state: "successful")
```
binary-koan, line 28: Is this also a workaround to avoid VCR? (if so please highlight these kind of places so I can update them with cassettes)

spec/models/purchase/refund_with_backup_card_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+    end
+  end
+
+  def build_charge_refund_mock(amount_cents)
+    flow_of_funds = FlowOfFunds.new(
+      issued_amount: FlowOfFunds::Amount.new(currency: "usd", cents: -amount_cents),
```
binary-koan, line 34:
```suggestion
      issued_amount: FlowOfFunds::Amount.new(currency: Currency::USD, cents: -amount_cents),
```

(in general please don't hardcode the string)

spec/models/purchase/refund_with_backup_card_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+          )
+        )
+        allow(ChargeProcessor).to receive(:refund!).and_return(build_charge_refund_mock(2000))
+      end
+
+      it "charges the minimum amount of $1.00" do
+        purchase.refund_and_save!(nil)
+
+        credit = Credit.find_by(refund_funding_purchase: purchase)
+        expect(credit.amount_cents).to eq(100) # Minimum charge is $1.00
```
binary-koan, line 137:
```suggestion
        expect(credit.amount_cents).to eq(100)
```

Already covered by the test description

spec/services/refund_funding_charge_service_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+          )
+        )
+      end
+
+      it "returns success" do
+        expect(result.success?).to be true
+        expect(result.error_message).to be_nil
+      end
+
+      it "creates a credit" do
```
binary-koan, line 86: Nit: I'd combine this, `creates a credit with correct attributes`, and `returns the credit` together as they basically test the same thing

spec/services/refund_funding_charge_service_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+
+  describe "constants" do
+    it "has minimum charge of $1.00" do
+      expect(described_class::MINIMUM_CHARGE_CENTS).to eq(100)
+    end
+
+    it "has maximum charge of $10,000" do
+      expect(described_class::MAXIMUM_CHARGE_CENTS).to eq(1_000_000)
+    end
+  end
```
binary-koan, line 248: I don't think we should be testing constant values

PR #2420, opened 2025-12-20
Inertia migration: Settings

Pradumn27:
ref #856 
reopened #2331

This PR migrates the following pages to Inertia:

1. Main Settings (/settings)
2. Profile Settings (/settings/profile)
3. Payment Settings (/settings/payments)
4. Password Settings (/settings/password)
5. Team Settings (/settings/team)
6. Advanced Settings (/settings/advanced)
7. Third Party Analytics Settings (/settings/third_party_analytics)
8. Authorized Applications Settings (/settings/authorized_applications)

### Video

https://github.com/user-attachments/assets/c0323c69-e6cb-4838-b69f-ed439e751ef7

https://github.com/user-attachments/assets/a4c167b9-ec6b-405b-87bb-5af8b90dc0d6

#### Controller specs:
`bundle exec rspec \                                           
  spec/controllers/oauth/applications_controller_spec.rb \
  spec/controllers/oauth/authorized_applications_controller_spec.rb \
  spec/controllers/settings/advanced_controller_spec.rb \
  spec/controllers/settings/authorized_applications_controller_spec.rb \
  spec/controllers/settings/main_controller_spec.rb \
  spec/controllers/settings/password_controller_spec.rb \
  spec/controllers/settings/payments_controller_spec.rb \
  spec/controllers/settings/profile_controller_spec.rb \
  spec/controllers/settings/team_controller_spec.rb \
  spec/controllers/settings/third_party_analytics_controller_spec.rb`

Screenshot 2025-12-19 at 10 11 55 PM

#### Request specs:

`bundle exec rspec spec/requests/settings/main_spec.rb spec/requests/settings/advanced_spec.rb spec/requests/settings/password_spec.rb spec/requests/settings/team_spec.rb spec/requests/settings/third_party_analytics_spec.rb spec/requests/settings/payment_spec.rb`

Screenshot 2025-12-20 at 6 04 27 PM

### AI Disclosure

- Model (GPT 5.1)
- Used for writing tests and mainly for tab auto-completes
- All code written was reviewed and updated manually by me

### Livestream Viewing Confirmation

- Have watched all of them

PR #2437, opened 2025-12-23
[#2435] Replace BGN with EUR

### LINKED ISSUE #2435: Replace BGN with EUR
Stripe has discontinued support for the Bulgarian Lev (BGN). As a result, using default_currency_bgn is no longer supported and may cause payment or account issues.

Stripe now requires merchants previously using BGN to transition to EUR instead.

Update Gumroad to:

- Remove or deprecate default_currency_bgn

- Replace it with EUR where applicable 

Stripe support documentation:
https://support.stripe.com/questions/bulgarian-lev-(bgn)-transition-to-euro-(eur)

vishalzambre:
Fixed #2435

No price conversion is needed because BGN was never used for product prices. The changes ensure future payouts to Bulgarian sellers use EUR instead of BGN, which aligns with Stripe's requirements.

- [ ] Update `currency` from `BGN` to `EUR` for existing BG `merchant_account`s on production


spec/requests/settings/payments_spec.rb
```diff
@@ -1529,7 +1529,7 @@
         fill_in("Confirm IBAN", with: "BG80BNBG96611020345678")
 
         expect(page).to have_content("Must exactly match the name on your bank account")
-        expect(page).to have_content("Payouts will be made in BGN.")
+        expect(page).to have_content("Payouts will be made in EUR.")
 
         click_on("Update settings")
```
stefanbinoj, line 1535: Hi @vishalzambre one comment, you might also need to update `spec/models/country_spec.rb` as well

stefanbinoj: Also should fix slow test 43 as well 
ref - https://github.com/antiwork/gumroad/pull/2420#issuecomment-3682893536

vishalzambre: Spec updated.

Screenshot 2025-12-23 at 3 09 16 PM

ershad:
@vishalzambre Thanks for the PR, triggered CI.

@hchhabra _a please review on priority.

app/views/help_center/articles/contents/_13-getting-paid.html.erb
```diff
@@ -402,4 +402,4 @@
     <li><a href="15-1099s"><span>Does Gumroad issue form 1099-K and 1099-MISC?</span></a></li>
     <li><a href="330-stripe-connect"><span>Connect your Stripe account to Gumroad</span></a></li>
   </ul>
-</div>
+</div>
```
hchhabra:
```suggestion
</div>

```

vishalzambre:
> Looks good to me! We'll need to update the currencies for existing BG merchant account records, will do that.

I think historical balances with BGN `holding_currency` should remain as-is (they reflect what actually happened)?

vishalzambre:
@hchhabra  @ershad  Anything else we want to add here?

PR #2445, opened 2025-12-23
Add unreviewed users page to admin

ershad:
In addition to the existing systems for reviewing unreviewed users, add an admin page for quick access to the list.

image


ershad:
@EmCousin, please let me know if you have any suggestions on the PR, and will address them in a new PR.

PR #2447, opened 2025-12-23
fix: Unpublish products when archiving (Closes #2373)

### LINKED ISSUE #2373: Make Archived Products Unpublished and Removed from Indexing

SiddhantOfficial:
## Summary

When a product is archived, it should no longer be publicly accessible or discoverable. This PR ensures that archiving a product also unpublishes it by setting `purchase_disabled_at`.

## Changes

- Modified `Products::ArchivedController#create` to set `purchase_disabled_at` when archiving a product
- Added tests to verify the unpublishing behavior
- Preserves the original `purchase_disabled_at` timestamp if the product was already unpublished

## Fixes

Closes #2373

/claim #2373

### DIFF
--- a/app/controllers/products/archived_controller.rb
+++ b/app/controllers/products/archived_controller.rb
@@ -71,7 +71,9 @@ class Products::ArchivedController < Sellers::BaseController
   def create
     authorize [:products, :archived, @product]
 
-    @product.update!(archived: true)
+    # When archiving, also unpublish the product to ensure it's no longer
+    # publicly accessible or indexed by search engines (fixes #2373)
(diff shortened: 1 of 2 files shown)


binary-koan:
Thanks for working on this, but we merged an earlier PR (https://github.com/antiwork/gumroad/pull/2374)

PR #2448, opened 2025-12-24
fix: Show flash notification when exporting all time sales

### LINKED ISSUE #1930: Email notification not appearing when exporting Sales CSV from all time
### What

Tested with UID: `11579748`

SiddhantOfficial:
Fixes #1930

## Summary
This PR fixes the issue where the email notification message was not appearing when exporting Sales CSV from All Time.

## Changes
- Replaced redirect_back with explicit redirect_to in PurchasesController#export
- Used respond_to block to handle different format requests
- Added :see_other status for proper HTTP redirect with Inertia

## Root Cause
The original redirect_back does not work correctly with Inertia pages, causing flash messages to not display.

### DIFF
--- a/app/controllers/purchases_controller.rb
+++ b/app/controllers/purchases_controller.rb
@@ -413,8 +413,16 @@ class PurchasesController < ApplicationController
     if tempfile
       send_file tempfile.path
     else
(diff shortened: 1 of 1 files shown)


binary-koan:
Thanks for working on this, but please follow the contributing guidelines including screencast and AI disclosure.

PR #2455, opened 2025-12-24
Fix admin impersonation performance and CSV export bugs

### LINKED ISSUE #1931: Performance issues "becoming" users with large sales amount in Admin

yashranaway:
# Fixes #1931 

## What

Two bugs when admins impersonate users with lots of sales:

**Performance issue**: Dashboard is super slow or times out completely when becoming a large seller.

**CSV export bug**: When you download sales data while impersonating someone, it could pull YOUR sales instead of theirs.

Affects users like `1265432` and `432151` - anyone with a ton of sales basically.

### Performance
The `LargeSeller` table tracks users with 1000+ sales so we can cache their analytics. Problem is, these records only get created when users visit their own analytics page (`analytics_controller.rb` does this). The dashboard also uses `CreatorAnalytics::CachingProxy` but wasn't creating LargeSeller records.

**Fix**: Add `LargeSeller.create_if_warranted(current_seller)` to `dashboard_controller.rb` (same pattern as `analytics_controller.rb:15`) + redirect impersonation to products page for faster initial load.

### CSV exports
The issue is an edge case with `current_seller`:

When impersonating, `current_seller` uses cookie-based lookup with validation:
1. Cookie might have admin's seller ID from before impersonation
2. `valid_seller?` checks `logged_in_user.member_of?(seller)`
3. If the impersonated user IS a team member of admin's account, this returns TRUE
4. `current_seller` returns admin's account instead of impersonated user

This edge case is rare but possible. Instead of relying on `current_seller` during impersonation, we now explicitly use `logged_in_user` for exports.

**Fix**: Add `seller_for_export` method that returns `logged_in_user` when impersonating, bypassing the potentially-buggy `current_seller` logic.

## The fix

**app/controllers/dashboard_controller.rb**
```ruby
def index
  authorize :dashboard

  if current_seller.suspended_for_tos_violation?
    redirect_to products_url
  else
    LargeSeller.create_if_warranted(current_seller)  # New - enables analytics caching
    presenter = CreatorHomePresenter.new(pundit_user)
    render inertia: "Dashboard/Index",
           props: { creator_home: presenter.creator_home_props }
  end
end
```

**app/controllers/admin/base_controller.rb**
```ruby
def impersonate
  user = find_user(params[:user_identifier])

  if user
    impersonate_user(user)
    redirect_to products_path  # Changed from dashboard_path - faster initial load
  else
    ...
  end
end
```

**app/controllers/purchases_controller.rb** (same pattern for affiliates & audience)
```ruby
def export
  authorize [:audience, Purchase], :index?

  seller = seller_for_export  # New - uses correct seller when impersonating

  tempfile = Exports::PurchaseExportService.export(
    seller:,
    recipient: impersonating_user || logged_in_user,
    filters: params.slice(:start_time, :end_time, :product_ids, :variant_ids),
  )
  ...
end

# When impersonating, we must use logged_in_user (the impersonated user), not current_seller.
# current_seller could incorrectly return the admin's account if:
# 1. The admin has a seller cookie from their own account, AND
# 2. The impersonated user happens to be a team member of the admin's account
# This edge case would cause valid_seller? to return true for the wrong seller.
def seller_for_export
  return current_seller unless impersonating?

  if current_seller != logged_in_user
    Rails.logger.warn(
      "[Impersonation] Seller mismatch during export: " \
      "current_seller=#{current_seller.id}, logged_in_user=#{logged_in_user.id}"
    )
  end

  logged_in_user
end
```

## Testing

Added to existing spec files:
- `spec/controllers/dashboard_controller_spec.rb` - LargeSeller creation test
- `spec/controllers/purchases_controller_spec.rb` - Added assertion that export query uses impersonated seller's ID

All existing impersonation specs still pass.

## AI disclosure

Used Claude Sonnet 4.5 for implementation via Cursor. All code manually reviewed.

## Notes


app/controllers/dashboard_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
     if current_seller.suspended_for_tos_violation?
       redirect_to products_url
     else
+      LargeSeller.create_if_warranted(current_seller)
```
yashranaway, line 16: This follows the same pattern as `analytics_controller.rb:15`. Both dashboard and analytics pages use `CreatorAnalytics::CachingProxy` which checks for LargeSeller records. Creating the record here ensures caching works for dashboard too.

app/controllers/admin/base_controller.rb
```diff
@@ -33,7 +33,7 @@ def impersonate
 
     if user
       impersonate_user(user)
-      redirect_to dashboard_path
+      redirect_to products_path
```
yashranaway, line 36: Changed from `dashboard_path`. Products page loads faster for large sellers since it doesn't use CreatorAnalytics. This gives time for the analytics cache to warm up in the background before the admin navigates to dashboard.

app/controllers/purchases_controller.rb
```diff
@@ -418,6 +420,27 @@ def export
...
+        "current_seller=#{current_seller.id} (#{current_seller.email}), " \
+        "logged_in_user=#{logged_in_user.id} (#{logged_in_user.email}), " \
+        "admin=#{impersonating_user.id} (#{impersonating_user.email})"
+      )
+    end
+
+    logged_in_user
+  end
+
+
```
yashranaway, line 443: This is the core fix for the CSV export bug. Why not just use `current_seller`?
`current_seller` uses cookie + validation logic:
1. Gets seller from `cookies.encrypted[:current_seller_id]`
 2. Validates via `logged_in_user.member_of?(seller)`

The edge case: If the impersonated user is a team member of the admin's account, `member_of?` returns TRUE, and `current_seller` returns the admin's account instead of the impersonated user. By explicitly using `logged_in_user` when impersonating, we bypass this edge case entirely. `logged_in_user` is defined as `impersonated_user || current_user`, so it's guaranteed correct.

app/controllers/purchases_controller.rb
```diff
@@ -418,6 +420,27 @@ def export
...
+  # This edge case would cause valid_seller? to return true for the wrong seller.
+  def seller_for_export
+    return current_seller unless impersonating?
+
+    if current_seller != logged_in_user
+      Rails.logger.warn(
+        "[Impersonation] Seller mismatch during export: " \
+        "current_seller=#{current_seller.id} (#{current_seller.email}), " \
+        "logged_in_user=#{logged_in_user.id} (#{logged_in_user.email}), " \
+        "admin=#{impersonating_user.id} (#{impersonating_user.email})"
```
yashranaway, line 436: Added logging to understand when mismatches occur in production. This follows the pattern in `pundit_authorization.rb` which logs authorization issues for debugging.

app/controllers/affiliates_controller.rb
yashranaway, line 1: Same pattern as purchases_controller. All export endpoints now use `seller_for_export` for consistency.

spec/controllers/purchases_controller_spec.rb
```diff
@@ -723,6 +723,7 @@ def expect_correct_csv(csv_string)
 
             export = SalesExport.last!
             expect(export.recipient).to eq(@admin_user)
+            expect(export.query.dig("bool", "filter")).to include(hash_including("terms" => { "seller_id" => [seller.id] }))
```
yashranaway, line 726: This assertion verifies that the export query uses the SELLER's ID, not the admin's ID. Previously the test only checked that `recipient` was the admin (which is correct - admin gets the email). Now we also verify the data is for the right seller.

yashranaway (commented):
@binary-koan Tried a different approach based on your feedback from #2390, Open to feedback if you see a cleaner approach!

binary-koan:
@yashranaway I still want to think more about the CSV export issue; could you extract just the dashboard performance stuff to a separate PR so we can merge that first?

binary-koan:
Since Michelle can't reproduce the CSV issue any more, I think we can just close this one and merge #2491. I believe it's extremely unlikely for admins to be/have team members on creators' accounts, so I'd rather not add this complexity for such a small edge case - we can revisit if we run into problems again. Thanks for working through this, it's great to have the exploration and context!

yashranaway:
makes sense! thanks.

PR #2458, opened 2025-12-24
Update tip position in checkout

### LINKED ISSUE #1929: Update tip position in checkout
### Description
Refine the checkout tipping experience to match the latest Figma designs and ensure it behaves consistently across desktop and mobile.
This update fixes layout and interaction issues by keeping the tip selector in a stable position and preventing focus or layout shifts during user input.

Image

### Checklist
- [ ] Render tip selector in a consistent, fixed position in the left side of the checkout.
- [ ] Implement updated tip selection UI. Show five options: No Tip, 15%, 20%, 25%, and Custom. Default selection: No Tip. Selecting a percentage automatically updates the tip amount and subtotal.
- [ ] Ensure the cart summary always displays the correct subtotal and total, including the selected tip.
- [ ] Prevent any layout, focus, or scroll shifts when interacting with the tip selector (especially when typing a custom tip).

### Designs

[Multiple options (Desktop)](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-9380&t=SlBJrwgiHxY9EDV6-4)

[Custom tip only](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-9825&t=SlBJrwgiHxY9EDV6-4)

[Mobile layout](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-10362&t=SlBJrwgiHxY9EDV6-4)

sm17p:
# Closes #1929 
- #1929

- This update fixes layout and interaction issues by keeping the tip selector in a stable position and preventing focus or layout shifts during user input. 
- Additionally, we've moved tip sub-section from cart summary rendering it besides the new tip section header 

## Design

- [Multiple options (Desktop)](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-9380&t=SlBJrwgiHxY9EDV6-4)
- [Custom tip only](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-9825&t=SlBJrwgiHxY9EDV6-4)
- [Mobile layout](https://www.figma.com/design/p0aMLuSmomzt6kQZgojN2a/Fix-tip-position-to-avoid-layout-shift?node-id=295-10362&t=SlBJrwgiHxY9EDV6-4)

## Screenshots

- [x] Render tip selector in a consistent, fixed position in the left side of the checkout.

| Device | Before | After |
|---------|--------|----|
| Desktop | desktop-before | desktop-after |
| Mobile | mobile-before | mobile-after |

## Interaction Demo

- [x] Implement updated tip selection UI. Show five options: No Tip, 15%, 20%, 25%, and Custom. Default selection: No Tip. Selecting a percentage automatically updates the tip amount and subtotal.
- [x] Ensure the cart summary always displays the correct subtotal and total, including the selected tip.
- [x] Prevent any layout, focus, or scroll shifts when interacting with the tip selector (especially when typing a custom tip).

| Paid Cart | Free Cart |
|---------|--------|
| <video src="https://github.com/user-attachments/assets/bdbc82a5-1224-4942-91b7-798a1f118252" /> | <video src="https://github.com/user-attachments/assets/24bafc5e-3ab4-4895-bf5d-4d5b549df68e" /> |

<hr />

## Responsive Demos

| Paid Cart | Free Cart |
|---------|--------|
| <video src="https://github.com/user-attachments/assets/f8321848-7def-4758-ae54-69c4ae0bea0e" /> | <video src="https://github.com/user-attachments/assets/1b6d1d0f-faf7-49ac-b1b6-12e898292b75" /> |

### Local Spec Run
- Command: `bundle exec rspec spec/services/tip_options_services_spec.rb spec/requests/checkout/form_spec.rb spec/requests/purchases/tipping_spec.rb spec/requests/purchases/product/installment_plan_spec.rb spec/requests/purchases/product/taxes_spec.rb spec/requests/purchases/product/upsell_spec.rb` 
- Tax related specs are failing due to missing keys, recommend running `bundle exec rspec spec/requests/purchases/product/installment_plan_spec.rb:345 spec/requests/purchases/product/taxes_spec.rb` on local before CI run

specs-passing

### AI Disclosure
- Cursor Auto Mode for re-factor and updating specs

### Live Stream Disclosure
- Have watched all five live streams

### DIFF
--- a/app/javascript/components/PriceInput.tsx
+++ b/app/javascript/components/PriceInput.tsx
@@ -63,7 +63,7 @@ export const PriceInput = React.forwardRef<
     return (
       <div className={cx("input", { disabled })}>
         {currencyCodeSelector ? (
-          <Pill className={cx("relative -ml-2 shrink-0 cursor-pointer")}>
+          <Pill className="relative -ml-2 shrink-0 cursor-pointer">
             {getLongCurrencySymbol(currencyCode)}
             <TypeSafeOptionSelect
               name="Currency"
@@ -78,7 +78,7 @@ export const PriceInput = React.forwardRef<
             <Icon name="outline-cheveron-down" className="ml-auto" />
           </Pill>
         ) : (
-          <Pill className={cx("-ml-2 shrink-0")}>{getLongCurrencySymbol(currencyCode)}</Pill>
+          <Pill className="-ml-2 shrink-0">{getLongCurrencySymbol(currencyCode)}</Pill>
         )}
         <input
           type="text"


app/javascript/components/Checkout/index.tsx
```diff
@@ -225,18 +236,19 @@ export const Checkout = ({
   }, 0);
 
   const isDesktop = useIsAboveBreakpoint("lg");
+  const displayTipSelector = isTippingEnabled(state);
 
   return (
-    <div className="mx-auto w-full max-w-product-page">
+    <div className="mx-auto w-full max-w-400">
```
sm17p, line 242: Design width for desktop

Image

app/javascript/components/Checkout/index.tsx
```diff
@@ -225,18 +236,19 @@ export const Checkout = ({
...
       <PageHeader
-        className="border-none"
+        className="border-none md:px-16"
```
sm17p: Gap as per design

app/javascript/components/Checkout/index.tsx
```diff
@@ -225,18 +236,19 @@ export const Checkout = ({
...
         title="Checkout"
         actions={
           isDesktop ? <NavigationButton href={cart.returnUrl ?? discoverUrl}>Continue shopping</NavigationButton> : null
         }
       />
       {isOpenTuple(cart.items, 1) ? (
-        <div className="grid gap-8 p-4 md:p-8">
+        <div className="grid gap-8 p-4 md:p-8 md:px-16">
```
sm17p, line 252: Gap as per design

app/javascript/components/Checkout/index.tsx
```diff
@@ -250,7 +262,12 @@ export const Checkout = ({
                   />
                 ))}
               </CartItemList>
-              <CartItemList>
+              {displayTipSelector ? (
+                <CartItemList className="rounded-b-none p-3 sm:p-5">
```
sm17p: `CartItemList` sets paddings for the section and `TipSelector` handles spacing of elements inside it

app/javascript/components/Checkout/index.tsx
```diff
@@ -398,14 +415,90 @@ export const Checkout = ({
...
+
+  const tipPercentages = [0, 15, 20, 25];
+
+  return (
+    <div className="flex flex-col gap-2 sm:gap-3">
+      <CartPriceItem
+        className={{ priceField: "font-bold" }}
+        title="Add a tip?"
+        price={formatPrice(computeTip(state))}
+      />
```
sm17p: As per design, making the price field bold over here

app/javascript/components/Checkout/index.tsx
```diff
@@ -398,14 +415,90 @@ export const Checkout = ({
...
+                style={{ justifyContent: "center" }}
+              >
+                {percentage === 0 ? "No Tip" : `${percentage}%`}
+              </Button>
+            ))}
+          </div>
+        ) : null}
+        <fieldset className={cx("flex-1 basis-[9.5rem]", { danger: errors.has("tip") })}>
+          <PriceInput
+            className={{ pill: "rounded-full text-sm" }}
```
sm17p: Override class as per design

app/javascript/components/Checkout/PaymentForm.tsx
```diff
@@ -727,87 +722,6 @@ const CreditCard = () => {
   );
 };
 
-const TipSelector = () => {
```
sm17p: Moved to `Checkout/index`

spec/requests/checkout/form_spec.rb
```diff
@@ -172,23 +172,25 @@
...
-        expect(page).to have_text("Add a tip")
-        expect(page).to have_radio_button("0%", checked: true)
-        expect(page).to have_radio_button("10%", checked: false)
+        expect(page).to have_text("Add a tip?")
+        expect(page).to have_radio_button("No Tip", checked: true)
+        expect(page).to have_radio_button("15%", checked: false)
         expect(page).to have_radio_button("20%", checked: false)
-        expect(page).to have_radio_button("Other", checked: false)
+        expect(page).to have_radio_button("25%", checked: false)
+        expect(page).to have_field("Custom tip", placeholder: "Custom tip")
```
sm17p, line 180: Added a new assertion since "Custom Tip" is now visible by default

spec/requests/purchases/product/installment_plan_spec.rb
```diff
@@ -217,11 +217,11 @@
       purchase = product.sales.last
       subscription = purchase.subscription
       expect(purchase).to have_attributes(
-        price_cents: 434,
+        price_cents: 534,
```
sm17p, line 220: Tip is increased by `100` cents

spec/requests/purchases/product/taxes_spec.rb
```diff
@@ -143,9 +143,9 @@
 
         purchase = Purchase.last
         expect(purchase.link_id).to eq(@product.id)
-        expect(purchase.price_cents).to eq(550_00)
-        expect(purchase.total_transaction_cents).to eq(608_85)
-        expect(purchase.fee_cents).to eq(71_75) # 597_44 * 0.129 + 50c + 30c
```
sm17p, line 148: This should've been # `550_00 * 0.129 + 50c + 30c` which is `71_75`

spec/requests/purchases/tipping_spec.rb
```diff
@@ -16,28 +16,31 @@
...
       fill_checkout_form(product2)
 
-      expect(page).to have_radio_button("0%", checked: true)
+      expect(page).to have_radio_button("No Tip", checked: true)
+      expect(page).to have_text("Add a tip? US$0", normalize_ws: true)
 
-      choose "10%"
+      choose "20%"
 
+      expect(page).to have_field("Custom tip", with: "")
```
sm17p, line 24: [NEW] Since we display "Custom Tip" by default now, adding an assertion over here to ensure selecting a tip option doesn't not change the value inside input

spec/requests/purchases/tipping_spec.rb
```diff
@@ -16,28 +16,31 @@
...
       expect(page).to have_text("Subtotal US$40", normalize_ws: true)
-      expect(page).to have_text("Tip US$4", normalize_ws: true)
-      expect(page).to have_text("Total US$44", normalize_ws: true)
+      expect(page).to have_text("Tip US$8", normalize_ws: true)
+      expect(page).to have_text("Total US$48", normalize_ws: true)
+      expect(page).to have_text("Add a tip? US$8", normalize_ws: true)
```
sm17p: [NEW] To check tip value is rendered besides the seciton header updates when tip is changed

spec/requests/purchases/tipping_spec.rb
```diff
@@ -301,10 +306,11 @@
 
       add_to_cart(product1)
 
-      expect(page).to have_radio_button("5%", checked: false)
+      expect(page).to have_radio_button("No Tip", checked: false)
       expect(page).to have_radio_button("15%", checked: true)
+      expect(page).to have_radio_button("20%", checked: false)
       expect(page).to have_radio_button("25%", checked: false)
-      expect(page).to have_radio_button("Other", checked: false)
+      expect(page).to have_field("Custom tip", placeholder: "Custom tip")
```
sm17p, line 223: Added a new assertion since "Custom Tip" is now visible by default

sm17p (commented):
This is ready for review. Please have a look 🙏

app/javascript/components/Checkout/index.tsx
```diff
@@ -398,14 +414,90 @@ export const Checkout = ({
...
+                }}
+                disabled={isProcessing(state)}
+                style={{ justifyContent: "center" }}
+              >
+                {percentage === 0 ? "No Tip" : `${percentage}%`}
+              </Button>
+            ))}
+          </div>
+        ) : null}
+        <fieldset className={cx("flex-1 basis-38", { danger: errors.has("tip") })}>
```
sm17p: Important: We have to increase page width and add a min width of `152px/9.5rem` to "Custom Tip" as per design so that placeholder text (on `L482`) is visible properly.

To ensure proper flex behaviour, we set basis to `38/9.5rem` for "Custom Tip" and `174/43.5rem (4 * 9.5 rem + gap)` for radio buttons group. This ensures that all 5 elements space out equally when they can render on the same row.


https://github.com/user-attachments/assets/79a9082f-3ec8-4f7c-aada-4ab247eeac1a



Because of `flex-wrap` on parent, "Custom Tip" field will auto fill the available width on the next row whenever it's box size falls below `9.5 rem`

app/javascript/components/Checkout/index.tsx
```diff
@@ -176,10 +187,11 @@ export const Checkout = ({
...
   const discountInputDisabled = loadingDiscount || isProcessing(state);
-  const subtotal = cart.items.reduce(
-    (sum, item) => sum + Math.round(hasFreeTrial(item, isGift) ? 0 : convertToUSD(item, item.price) * item.quantity),
-    0,
-  );
+  const subtotal =
+    cart.items.reduce(
+      (sum, item) => sum + Math.round(hasFreeTrial(item, isGift) ? 0 : convertToUSD(item, item.price) * item.quantity),
+      0,
+    ) + computeTip(state);
```
sm17p, line 194: Displaying tip in subtotal section, if I got the design right? @laugardie is this desired?

Screenshot 2025-12-24 at 7 23 28 PM

laugardie: Yes, you’ve got it right.

app/javascript/components/Checkout/index.tsx
binary-koan: Sounds like grid would simplify this - we could set 4 columns with col-span-full on the custom tip option for mobile, and 5 equal width columns on desktop?

binary-koan:
@laugardie Did you check the design on this branch? I'll review properly once you're good with it

A few thoughts
* The checkout page max width is now different to product pages; probably fine since it's such a dramatic transition anyway but FYI
* "Custom tip" only shows side by side with the other options at 1600px+ wide, so it'll wrap on quite a few laptop screens. This is needed to show the full placeholder while keeping the buttons the same width. I wonder if it would be nicer to just set a min width on that field, and sacrifice the equal width to keep it on one line on smaller laptops?
* The font size on the tip buttons looks a little small to me

laugardie:
@binary-koan just reviewed it! Sorry for the delay, I was OOO.

> The checkout page max width is now different to product pages; probably fine since it's such a dramatic transition anyway but FYI

Yeah, I agree it’s probably fine either way. I’d personally be happy to keep the same max width as the product pages for consistency. What does feel off to me is the gap between the header and the grid. It feels a bit too large right now (64px). Dropping that to 32px should be better.

> "Custom tip" only shows side by side with the other options at 1600px+ wide, so it'll wrap on quite a few laptop screens. This is needed to show the full placeholder while keeping the buttons the same width. I wonder if it would be nicer to just set a min width on that field, and sacrifice the equal width to keep it on one line on smaller laptops?

I like the idea, but I tried this in Figma and visually it didn’t quite work for me. The custom tip ends up carrying more visual weight than the other options, which feels a bit unbalanced.

> The font size on the tip buttons looks a little small to me

Agreed. they do feel a bit small. Bumping them to 16px to match the custom tip makes sense.

- On mobile, when there isn’t space for all 4 buttons, we should split them into a 2x2 layout instead.
Screenshot 2026-01-08 at 15 24 17

sm17p:
> I like the idea, but I tried this in Figma and visually it didn’t quite work for me. The custom tip ends up carrying more visual weight than the other options, which feels a bit unbalanced.

I think the visual weight point makes sense, as having a larger size for "Custom Tip" can imply to some people that it's a recommended choice for the action. However due note that, the width distribution is not consistently equal across screen sizes so maybe we can just have everything equally distributed when we can for a clean look? (I'm keeping the 1600px width for now since it's a required constraint for equal distribution accounting for "Custom Tip" placeholder visibility)

> Agreed. they do feel a bit small. Bumping them to 16px to match the custom tip makes sense.

I've reverted this to the original font size of `16px`

>  On mobile, when there isn’t space for all 4 buttons, we should split them into a 2x2 layout instead.

- 1 x 4 when space available for all
- 2 x 2 when 4 of them can't fit in a row
- 4 x 1 when 2 of them cant fit in a row - (screen width <= 324px, unlikely to happen in general but people use split view / have foldable devices these days)


https://github.com/user-attachments/assets/de923294-c3af-4c8b-80b1-7749058afbbd


<hr />

> What does feel off to me is the gap between the header and the grid. It feels a bit too large right now (64px). Dropping that to 32px should be better.

That seems off to me as well, definitely agree unless @binary-koan has another opinion on it

@laugardie before I make the change for the above I would like to know answers to the following questions as it will help me apply the right css classes

1. Do we want to increase "Checkout" font size from current `24px` to `40px` as per Figma?
2. Do we want to make "Checkout" and "Continue Shopping" visible on mobile? Currenty, it's not visible on production.
  

Screenshot 2026-01-09 at 8 24 31 AM

laugardie:
> Do we want to increase "Checkout" font size from current 24px to 40px as per Figma?

No.

> Do we want to make "Checkout" and "Continue Shopping" visible on mobile? Currenty, it's not visible on production.

In my opinion we should show Checkout title in mobile for better orientation. The Continue Shopping button appears at the bottom of the checkout if you scroll. I think it was moved there as a strategic decision to prioritize conversion and avoid giving customers an immediate exit point.

sm17p:
Makes sense, I'll update it accordingly then thank you!

app/javascript/components/Checkout/index.tsx
```diff
@@ -242,7 +242,8 @@ export const Checkout = ({
   return (
     <div className="mx-auto w-full max-w-400">
       <PageHeader
-        className="border-none md:px-16"
+        className="border-none pb-2 md:px-16 md:pb-0"
+        classNames={{ title: "block!", actionsContainer: "my-0!" }}
```
sm17p: 1. Used `classNames` over here to over-ride a few classes. Ideally I would've done it with `className` but the diff seemed huge for that as `PageHeader` is used widely. @binary-koan wdyt about this? Should we change it across all using `className`? 

2. `title: "block!"` to make the the page title visible on mobile as a `hidden!` class is present at the call site

<hr />

3. Had to use `actionsContainer: "my-0!"` to ensure the there is an exact gap of `32px` between the sections. Previously, having `md:-my-2` on the container caused it to shrink below the height of it's children, creating a minor overlap in the vertical gap between adjacent sections     

|Before| After|
|----|----|
|<video src="https://github.com/user-attachments/assets/eba65730-6680-45f8-8ce7-b7b8f078adc7" /> | <video src="https://github.com/user-attachments/assets/335186ee-a00b-492c-bad5-5f14e4c86d5f" />

sm17p:
@laugardie I've made the requested changes

1. Page title is now visible on mobile
2. The gap between page header and content is reduced to `32px` for desktop and `16px` on mobile
3. Increased `font-size` (16px) on tip options
4. Proposed layouting strategy of `1 x 4`, `2 x 2` and `4 x 1` for tip options

Please have a look and LMK if there's any other adjustments to be made 🙏

https://github.com/user-attachments/assets/9ccc888e-7fe0-4963-b23e-78eccbb301d7

laugardie:
Looks good to me @binary-koan _a

app/javascript/components/Checkout/index.tsx
```diff
@@ -1,3 +1,4 @@
+import cx from "classnames";
```
binary-koan: Please use `$app/utils/classNames` instead of the classnames package

app/javascript/components/Checkout/index.tsx
binary-koan: I see your point, but I'd rather not introduce child class props like this if we can avoid it. How about we add an optional `showTitleOnMobile` prop to `PageHeader` for the title style, and add a margin-top to the section below the header (or a margin-bottom to the header) instead of styling the actions container?

app/javascript/components/Checkout/index.tsx
binary-koan: What prevents us combining these two `CartItemList`s? It seems odd to use two but then style them to look like one

app/javascript/components/Checkout/index.tsx
binary-koan: Here too I'd rather set a prop than override child class names (essentially I think the more granular control you give over the internal component styling, the harder it is to update any default styles because different places will be relying on combining them with other CSS in different ways)

Maybe we could either add a `tip?: boolean` prop or combine with the existing large style for a `variant?: "tip" | "large"` prop?

app/javascript/components/Checkout/index.tsx
binary-koan: I think we can remove this since that whole row now uses the standard font size

app/javascript/components/ui/PageHeader.tsx
```diff
@@ -1,6 +1,7 @@
+import { type ClassValue } from "clsx";
 import * as React from "react";
 
-import { classNames } from "$app/utils/classNames";
+import { classNames as cx } from "$app/utils/classNames";
```
binary-koan: Please don't rename this import - old code still uses `import cx from "classnames"` which makes it confusing

binary-koan (commented):
Overall this is looking good! Just some minor stuff to check before merging

app/javascript/components/Checkout/index.tsx
```diff
@@ -402,27 +418,99 @@ export const Checkout = ({
...
+                  type: "fixed",
+                  amount: newAmount,
+                },
+              });
+            }}
+            placeholder="Custom tip"
+            disabled={isProcessing(state)}
+          />
+        </fieldset>
+      </div>
```
sm17p, line 483: Changed from flex to grid, please checkout [e9d9af8](https://github.com/antiwork/gumroad/pull/2458/commits/e9d9af8900cf689a9c21e703179062e52d7f81ae) commit for the changes

Overall, grid based version is more accurate at different screen sizes


https://github.com/user-attachments/assets/36239e06-dda0-407e-a44b-28a5fff33ec6

app/javascript/components/Checkout/index.tsx
```diff
@@ -402,27 +418,99 @@ export const Checkout = ({
...
+  const tipPercentages = [0, 15, 20, 25];
+
+  return (
+    <div className="@container flex flex-col gap-2 sm:gap-3">
+      <CartPriceItem title="Add a tip?" price={formatPrice(computeTip(state))} variant="tip" />
+      <div className="grid grid-cols-1 gap-4 @[52rem]:grid-cols-5">
+        {showPercentageOptions ? (
+          <div
+            role="radiogroup"
+            className="radio-buttons col-span-full grid-cols-1! @3xs:grid-cols-2! @sm:grid-cols-4! @[52rem]:col-span-4!"
```
sm17p, line 440: We need `@[52rem]:` container custom modifier as the overall layout is dependent on size of it's container (`L436`) , giving us finer control over radio item and custom tip positioning than screen-width based selectors

app/javascript/components/Checkout/index.tsx
```diff
@@ -402,27 +418,99 @@ export const Checkout = ({
...
-        "inline-flex flex-wrap gap-2",
-        large ? "text-base font-bold sm:text-xl" : "text-sm sm:text-base",
-      )}
-    >
-      {title}
-    </h4>
-    <div className={classNames("text-base sm:text-lg", large && "font-bold")}>{price}</div>
-  </div>
-);
+  variant?: "default" | "large" | "tip";
```
sm17p, line 495: > Maybe we could either add a tip?: boolean prop or combine with the existing large style for a variant?: "tip" | "large" prop?

Ref - https://github.com/antiwork/gumroad/pull/2458#discussion_r2700146347

app/javascript/components/ui/PageHeader.tsx
sm17p: This will go away after removing dependency on the local `classNames` variable 👍

sm17p:
Ref -  https://github.com/antiwork/gumroad/pull/2458#discussion_r2700134272

> What prevents us combining these two CartItemLists? It seems odd to use two but then style them to look like one

Nothing, I think that just felt natural to me but it does seem odd now that you pointed it out. 

I've merged them into one, please have a look again _a @binary-koan

PR #2460, opened 2025-12-24
Remove customer reviews from products when bundle is refunded

vishaltelangre:
### What

When a bundle is refunded, reviews on the bundled products by the corresponding customer are now properly removed.

Resolves https://github.com/antiwork/gumroad-private/issues/31.

### Why

Refunding a bundle only marks the bundle purchase as refunded, but leaves the individual product purchases untouched. This meant customer reviews remains visible on those products even after the bundle was fully refunded.


vishaltelangre:
@ershad _a Please review.

Failing spec seems flaky and is unrelated.

ershad (approved):
@vishaltelangre lgtm 👍🏼

PR #2461, opened 2025-12-24
Refactor admin unreviewed users page

ershad:
Follow up to #2446

image

### DIFF
--- a/app/javascript/components/Admin/UnreviewedUsers/index.tsx
+++ b/app/javascript/components/Admin/UnreviewedUsers/index.tsx
@@ -61,7 +61,7 @@ const UnreviewedUsersPage = () => {
             <TableHead>Email</TableHead>
             <TableHead>Revenue sources</TableHead>
             <TableHead>Payout method</TableHead>
-            <TableHead>Age</TableHead>
+            <TableHead>Account age</TableHead>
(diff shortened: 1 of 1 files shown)


app/javascript/components/Admin/UnreviewedUsers/index.tsx
```diff
@@ -16,6 +16,7 @@ type UnreviewedUser = {
   unpaid_balance_cents: number;
```
EmCousin, line 16: We could inline the content of this file in `app/javascript/pages/Admin/UnreviewedUsers/Index.tsx` directly

app/controllers/admin/unreviewed_users_controller.rb
EmCousin, line 26: Ideally `still_unreviewed_users` should be moved to a private method and this prop could be exposed as a lamda, so it doesn't have to be evaluated everytime we reach this endpoint in the event of a partial reload that only requests for the `total_count` or `cutoff_date`, for instance. Probably not happening here, though, but important to know for expensive queries if partial reloads are used.

EmCousin, line 28: Same for cached_data, consider evaluating each prop in a lambda if it becomes too expensive to run.

PR #2005, opened 2025-11-12
Inertia migration - Workflows

Pradumn27:
ref #856 

reopened https://github.com/antiwork/gumroad/pull/1902

This PR migrates the following pages to Inertia:
1. Workflows (/workflows)
2. Workflows new (/workflows/new)
3. Workflows edit (workflows/:id/edit)
4. Workflows emails (workflows/:id/emails)

Follows the same pattern as the previous migrations (ref #1559)

Additionally, have removed the API controller for workflow here as we are moving completed off the React SPA + API calls to an inertia hybrid approach.

Previously, the frontend made API calls to dedicated API controllers, resulting in full page data being fetched and managed client-side. have now removed the separate API controller layer and consolidated all workflow operations into the main WorkflowsController, leveraging Inertia's server-side rendering and client-side navigation capabilities. The key architectural improvement is the implementation of partial reloads - instead of fetching entire page datasets, Inertia now fetches only the specific props that have changed (using the only parameter and lazy-loaded lambdas in controller props). For example, when updating workflow details, only the workflow prop is refreshed rather than the entire context, including products, variants, and other metadata. This significantly reduces payload sizes and improves response times. Additionally, form submissions now use Inertia's router.post()/router.patch() methods instead of Axios, enabling seamless page transitions with proper error handling, flash messages, and state management.

source: https://inertia-rails.dev/guide/partial-reloads#lazy-data-evaluation

##### Gumroad.com
https://github.com/user-attachments/assets/2e9b9c4e-9f6e-4736-9e1c-f758fcca6cbc

##### Gumroad.dev

https://github.com/user-attachments/assets/9372e994-0217-407d-93d3-6d387431f6b7

##### Gumroad.com
https://github.com/user-attachments/assets/353236a3-f42d-49ac-9347-1572162cca1c

##### Gumroad.dev
https://github.com/user-attachments/assets/06fb6944-4413-450b-9a60-aaaac093c648

### Tests
`bundle exec rspec spec/controllers/workflows_controller_spec.rb --format html --out test_results_final.html`

Screenshot 2025-10-15 at 7 30 05 PM

### AI Disclosure

- Model (cursor auto mode)
- Used for writing tests and mainly for tab auto-completes
- All code written was reviewed and updated manually by me

### Livestream Viewing Confirmation

- Have watched all of them

PR #1627, opened 2025-10-11
Inertia migration 2

Pradumn27:
ref #856 

This PR migrates the following pages to Inertia:
1. Products - Affiliated (/products/affiliated)
2. Products - Archived (/products/archived)
3. Products - Collabs (/products/collabs)
4. Products - New (/products/new)

Follows the same pattern as the previous migrations (ref #1559)

### Video

https://github.com/user-attachments/assets/c1045de9-7370-43ed-a432-6ff08301160c

### Tests
`bundle exec rspec spec/controllers/links_controller_spec.rb spec/controllers/products/affiliated_controller_spec.rb spec/controllers/products/archived_controller_spec.rb spec/controllers/products/collabs_controller_spec.rb --format html --out test_results_final.html`

Screenshot 2025-10-11 at 3 33 37 AM

### AI Disclosure
Used cursor auto mode for doing similar pattern changes

PR #2463, opened 2025-12-24
Migrate Audience page to Inertia

4arjun:
ref #856 

This PR migrates the following page to Inertia:

1. Following (/dashboard/audience)

Enables SPA navigation between the Following and Sales tabs in the Analytics dashboard.
Follows the same pattern as the previous migrations (ref #2005, #2287)
 
## Before

https://github.com/user-attachments/assets/32d26338-559f-4257-b796-348268c9a80e

## After

https://github.com/user-attachments/assets/d547234a-8426-4f02-81a9-ca89594ea137

## Invalid date case

https://github.com/user-attachments/assets/381b52f1-f3a6-4c9b-837a-f5cdd6aad8d2

## Tests
Screenshot 2025-12-28 at 6 54 04 PM

## AI Usage
Used GPT-5.1 (Cursor) to understand the migration done by pr https://github.com/antiwork/gumroad/pull/2005 https://github.com/antiwork/gumroad/pull/1627
## Confirmation
I have watched the Gumroad PR reviews live streaming video.
## Self-Review
I have self-reviewed all the code that I have written.

### DIFF
--- a/app/controllers/audience_controller.rb
+++ b/app/controllers/audience_controller.rb
@@ -34,6 +34,10 @@ class AudienceController < Sellers::BaseController
       begin
         end_time = DateTime.parse(params[:to])
         start_date = DateTime.parse(params[:from])
+        if start_date > end_time
+          flash.now[:warning] = "Please select a valid date range."
+          @invalid_date_range = true
+        end
       rescue StandardError
         end_time = DateTime.current
         start_date = end_time.ago(29.days)
@@ -49,6 +53,7 @@ class AudienceController < Sellers::BaseController
 
   private
     def fetch_audience_data
+      return nil if @invalid_date_range
       return nil unless current_seller.audience_members.where(follower: true).exists?
 
       CreatorAnalytics::Following.new(current_seller).by_date(start_date: @start_date.to_date, end_date: @end_date.to_date)
--- a/app/javascript/pages/Audience/Index.tsx
+++ b/app/javascript/pages/Audience/Index.tsx
@@ -38,7 +38,7 @@ export default function AudiencePage() {
 
   const reloadAudienceData = React.useCallback((start: string, end: string) => {
     router.reload({
-      only: ["audience_data"],
+      only: ["audience_data", "flash"],
       data: { from: start, to: end },
       onStart: () => setIsLoading(true),
       onFinish: () => setIsLoading(false),
--- a/spec/controllers/audience_controller_spec.rb
+++ b/spec/controllers/audience_controller_spec.rb
@@ -98,6 +98,14 @@ describe AudienceController, inertia: true do
         expect(response).to be_successful
         expect(inertia.props[:audience_data]).to be_present
       end
+
+      it "returns nil audience_data and sets a flash warning when from date is after to date" do
+        get :index, params: { from: "2021-01-10", to: "2021-01-01" }
+
+        expect(response).to be_successful
+        expect(flash[:warning]).to eq("Please select a valid date range.")
+        expect(inertia.props[:audience_data]).to be_nil
+      end
     end
   end
 


app/controllers/audience_controller.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class AudienceController < Sellers::BaseController
-  before_action :set_body_id_as_app
```
4arjun, line 4: Removed `before_action :set_body_id_as_app` as it's not needed for Inertia pages/JSON endpoints

app/controllers/audience_controller.rb
```diff
@@ -10,7 +11,9 @@ class AudienceController < Sellers::BaseController
   def index
     authorize :audience
 
-    @total_follower_count = current_seller.audience_members.where(follower: true).count
+    render inertia: "Audience/Index", props: {
+      total_follower_count: current_seller.audience_members.where(follower: true).count
+    }
```
4arjun, line 17: Changed `index` action to render Inertia component instead of setting instance variable

app/javascript/components/Analytics/index.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import { assertDefined } from "$app/utils/assert";
 import { AbortError } from "$app/utils/request";
 
-import { AnalyticsLayout } from "$app/components/Analytics/AnalyticsLayout";
+import { AnalyticsLayout } from "$app/components/Analytics/Layout";
```
4arjun: Updated import to use new Inertia-enabled Layout. This enables SPA navigation from Sales page to Following page

app/javascript/pages/Audience/Index.tsx
```diff
@@ -1,12 +1,11 @@
...
 import { lightFormat } from "date-fns";
 import * as React from "react";
-import { createCast } from "ts-safe-cast";
 
 import { AudienceDataByDate, fetchAudienceDataByDate } from "$app/data/audience";
 import { AbortError } from "$app/utils/request";
-import { register } from "$app/utils/serverComponentUtil";
 
-import { AnalyticsLayout } from "$app/components/Analytics/AnalyticsLayout";
+import { AnalyticsLayout } from "$app/components/Analytics/Layout";
```
4arjun: Updated Layout import to use new Inertia-enabled Layout

app/javascript/components/Analytics/Layout.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <Tab href={Routes.utm_links_dashboard_path()} isSelected={selectedTab === "utm_links"}>
+              Links
+            </Tab>
+          ) : null}
+        </Tabs>
+      </PageHeader>
+      {children}
+    </div>
+  );
+};
```
4arjun: Created to enable SPA navigation between Following and Sales tabs.


The original `AnalyticsLayout.tsx` still exists and is used by `UtmLinksPage`. Links tab still uses regular Tab href because UTM Links page is not yet on Inertia.
Once UTM Links is migrated to Inertia, `AnalyticsLayout.tsx` can be deleted and this file can be renamed and the Links tab can also use Inertia Link

Pradumn27:
@Pradumn27 _a - will review as the assigned reviewer for the SPA issue

app/javascript/components/Analytics/Layout.tsx
Pradumn27: It is best to not change the layout here and only change when all pages used by this Layout are migrated to Inertia.

Since you would also have to update to using this layout for the Sales Analytics page too as it still uses the original one, which would affect pages out of scope of this PR.

And since only UTM Links page is left it's going to be done really soon too, so we can do it with that page's migration

app/javascript/pages/Audience/Index.tsx
Pradumn27: Can you also update to use Inertia [partial reloads](https://inertia-rails.dev/guide/partial-reloads#lazy-data-evaluation) instead of manual fetch calls? This would require modifying the audience controller to render page with `audience_data` prop, then we can use router.reload({ only: ['audience_data'] }) for date range changes. This would leverage Inertia's lazy data evaluation.

app/javascript/components/Analytics/Layout.tsx
4arjun: I've deleted the new layout. Once all three analytics pages (Following, Sales, UTM Links) are migrated to Inertia, a new shared layout with Inertia `<Link>` components can be introduced.

app/controllers/audience_controller.rb
```diff
@@ -1,16 +1,20 @@
 # frozen_string_literal: true
 
 class AudienceController < Sellers::BaseController
-  before_action :set_body_id_as_app
-  before_action :set_time_range, only: %i[data_by_date]
+  layout "inertia"
+
+  before_action :set_time_range, only: %i[index data_by_date]
```
4arjun: Added index to set_time_range before_action. This is required for the lazy audience_data prop to access @start_date and @end_date when evaluating the data

app/controllers/audience_controller.rb
```diff
@@ -1,16 +1,20 @@
...
   after_action :set_dashboard_preference_to_audience, only: :index
   before_action :check_payment_details, only: :index
 
   def index
     authorize :audience
 
-    @total_follower_count = current_seller.audience_members.where(follower: true).count
+    render inertia: "Audience/Index", props: {
+      total_follower_count: current_seller.audience_members.where(follower: true).count,
+      audience_data: -> { fetch_audience_data }
```
4arjun, line 16: audience_data is computed when explicitly requested via router.reload({ only: ['audience_data'] }). On initial page load with followers, it's also evaluated since the page needs to display the chart.

app/controllers/audience_controller.rb
```diff
@@ -50,4 +54,11 @@ def set_time_range
     def set_title
       @title = "Analytics"
     end
+
+  private
+    def fetch_audience_data
+      return nil unless current_seller.audience_members.where(follower: true).exists?
+
+      CreatorAnalytics::Following.new(current_seller).by_date(start_date: @start_date.to_date, end_date: @end_date.to_date)
+    end
```
4arjun, line 60: avoids unnecessary database queries.

4arjun:
> Hi @4arjun _a, thanks for the PR added a few initial comments; please check them.
> 
> Also, in the video shown in the description, I only see empty states. Can you please add some data there and then record it so we can get a better visualization.

Updated

app/javascript/pages/Audience/Index.tsx
```diff
@@ -22,33 +20,37 @@ import { WithTooltip } from "$app/components/WithTooltip";
...
   const hasContent = total_follower_count > 0;
 
-  const activeRequest = React.useRef<AbortController | null>(null);
   React.useEffect(() => {
-    const loadData = async () => {
-      if (!hasContent) return;
+    if (isInitialMount.current) {
+      isInitialMount.current = false;
+      return;
+    }
```
4arjun, line 54: Skip initial mount, data is already loaded via Inertia props

4arjun:
Currently, the tabs utilize regular tags, resulting in a full-page reload. This is because the UTM Links page has not yet been migrated to Inertia. The shared Tab component cannot import @inertiajs/react without causing SSR issues.

Once the UTM Links page is migrated, we can simply update the Tab component to utilize Inertia `<Link>`

app/controllers/audience_controller.rb
4arjun: This can be removed now

app/controllers/audience_controller.rb
Pradumn27: why create a new function and not just reuse the previous one by removing `render json`? also why are you moving it to private?

app/javascript/pages/Audience/Index.tsx
```diff
@@ -16,39 +14,41 @@ import { ExportSubscribersPopover } from "$app/components/Followers/ExportSubscr
...
-  }, [startTime, endTime]);
+    if (!hasContent) return;
+
+    router.reload({
+      only: ["audience_data"],
+      data: { start_time: startTime, end_time: endTime },
+      onStart: () => setIsLoading(true),
+      onFinish: () => setIsLoading(false),
+    });
+  }, [startTime, endTime, hasContent]);
```
Pradumn27: The goal is to eliminate this pattern by introducing a function that accepts the required parameters and internally calls router.reload. This function should then be invoked whenever the date or other parameters change.

app/javascript/pages/Audience/Index.tsx
```diff
@@ -75,9 +75,12 @@ const AudiencePage = ({ total_follower_count }: { total_follower_count: number }
...
       {hasContent ? (
         <div className="space-y-8 p-4 md:p-8">
-          <AudienceQuickStats totalFollowers={total_follower_count} newFollowers={data?.new_followers ?? null} />
-          {data ? (
-            <AudienceChart data={data} />
+          <AudienceQuickStats
+            totalFollowers={total_follower_count}
+            newFollowers={audience_data?.new_followers ?? null}
+          />
+          {!isLoading && audience_data ? (
```
Pradumn27, line 88:
```suggestion
          {!isLoading ? (
```
Withou this it will show the loader even when the returned data is empty, please test that scenario too

spec/controllers/audience_controller_spec.rb
```diff
@@ -63,47 +88,4 @@
...
-        "new_followers" => 1
-      )
-    end
-
-    it "works for someone in the GMT-1200 TZ" do
-      mask = "%a %b %d %Y %H:%M:%S GMT-1200 (Changement de date)"
-      get(:data_by_date, params: { start_time: 2.days.ago.strftime(mask), end_time: 1.day.ago.strftime(mask) })
-      expect(response.status).to be(200)
-    end
-  end
```
Pradumn27, line 108: We still want to test these behaviors, so please update the test to be Inertia-rendered and then assert against the relevant props. You can refer to `workflows_controller_spec` as an example.

app/javascript/pages/Audience/Index.tsx
```diff
@@ -1,10 +1,8 @@
+import { router, usePage } from "@inertiajs/react";
 import { lightFormat } from "date-fns";
 import * as React from "react";
-import { createCast } from "ts-safe-cast";
 
-import { AudienceDataByDate, fetchAudienceDataByDate } from "$app/data/audience";
-import { AbortError } from "$app/utils/request";
-import { register } from "$app/utils/serverComponentUtil";
+import { AudienceDataByDate } from "$app/data/audience";
```
Pradumn27:
```suggestion
import { type AudienceDataByDate } from "$app/data/audience";
```
We'd like to be explicit about what we are importing

Pradumn27 (commented):
@4arjun _a, few more comments

app/javascript/pages/Audience/Index.tsx
```diff
@@ -16,39 +14,41 @@ import { ExportSubscribersPopover } from "$app/components/Followers/ExportSubscr
...
 import placeholder from "$assets/images/placeholders/audience.png";
 
-const AudiencePage = ({ total_follower_count }: { total_follower_count: number }) => {
+type AudienceProps = {
+  total_follower_count: number;
+  audience_data: AudienceDataByDate | null;
+};
+
+export default function AudiencePage() {
+  const { total_follower_count, audience_data } = usePage<AudienceProps>().props;
```
Pradumn27: The way usePage works with Typescript is by effectively hiding a `as` under the hood, which isn't recommended for true type enforcement.

Can you use ts-safe-cast instead

```
import { cast } from "ts-safe-cast";
// [...] 
const { total_follower_count, audience_data t } = cast<AudienceProps>(usePage().props);
```

app/controllers/audience_controller.rb
4arjun: Since the /audience/data/by_date route is removed, this is now an internal method, not an action. making it private prevents Rails from treating it as a routable action and clearly signals it's an internal helper method.

spec/controllers/audience_controller_spec.rb
4arjun: updated

app/javascript/pages/Audience/Index.tsx
4arjun: This scenario can’t actually occur.

`audience_data` is only `nil` when `total_follower_count = 0`.  
When there are zero followers, the UI renders the placeholder section, not the chart section.

Since the chart section is only rendered when `total_follower_count > 0`,  `audience_data` cannot be `nil` at the point where the chart is shown.
This scenario can't actually occur.

`audience_data` can only be `nil` when `total_follower_count = 0`, but in that case we’re in the placeholder section, not the chart section.

app/javascript/pages/Audience/Index.tsx
```diff
@@ -16,39 +15,47 @@ import { ExportSubscribersPopover } from "$app/components/Followers/ExportSubscr
...
-    const loadData = async () => {
-      if (!hasContent) return;
+  const reloadAudienceData = (startTime: string, endTime: string) => {
+    router.reload({
+      only: ["audience_data"],
+      data: { start_time: startTime, end_time: endTime },
+      onStart: () => setIsLoading(true),
+      onFinish: () => setIsLoading(false),
+    });
+  };
```
4arjun: reload function that accepts parameters, and internally calls router.reload

app/javascript/pages/Audience/Index.tsx
4arjun: @Pradumn27  I've extracted reloadAudienceData into a useCallback that accepts parameters and calls router.reload, then invoke it from useEffect when dates change.

4arjun:

app/controllers/audience_controller.rb
```diff
@@ -32,8 +32,8 @@ def export
   protected
     def set_time_range
       begin
-        end_time = DateTime.parse(params[:end_time])
-        start_date = DateTime.parse(params[:start_time])
+        end_time = DateTime.parse(params[:to])
+        start_date = DateTime.parse(params[:from])
```
4arjun, line 36: The useAnalyticsDateRange hook sets URL parameters as from and to, but the controller was reading start_time and end_time. This mismatch caused the date range in the URL to be ignored on initial page load, falling back to the default (last 29 days).

4arjun:
@binary-koan Thanks for the clarification. This PR overlaps with https://github.com/antiwork/gumroad/pull/2484, which I see you’re currently reviewing.
Sharing this here for visibility and comparison.

4arjun:
There is also an unhandled edge case that crashes the page when toDate < fromDate. Should we use the global Inertia error handler?

## Before migration

https://github.com/user-attachments/assets/a5fad6bf-c617-4ba6-ad42-3107ef369978



## After migration


https://github.com/user-attachments/assets/6d4dc9ec-b554-442d-9c39-0472c4bc1050

binary-koan:
Yeah looks like something it’s now worth checking for and gracefully handling in the controller

binary-koan:
Nice catch!

4arjun:
@binary-koan All tests are passing, and I’ve updated the PR description to include a video for the invalid date selection case.

Devwill2:
@4arjun I have one question. All your Audience One test cases are passing on the main branch, but my Audience One specs are failing. Could you tell me which command you’re using to run the tests?

4arjun:
> @4arjun I have one question. All your Audience One test cases are passing on the main branch, but my Audience One specs are failing. Could you tell me which command you’re using to run the tests?

@Devwill2, I don’t think this is related to the test command itself. The only failing specs I’m seeing in your PR are:

`
rspec ./spec/requests/analytics/audience_spec.rb:12
`
`
rspec ./spec/requests/analytics/audience_spec.rb:16
`
From the screenshots, the page appears to be rendering blank, which makes me think something might be missing or broken in your current approach. It may be worth double checking that part of the implementation.

Devwill2:
> > @4arjun I have one question. All your Audience One test cases are passing on the main branch, but my Audience One specs are failing. Could you tell me which command you’re using to run the tests?
> 
> @Devwill2, I don’t think this is related to the test command itself. The only failing specs I’m seeing in your PR are:
> 
> `rspec ./spec/requests/analytics/audience_spec.rb:12` `rspec ./spec/requests/analytics/audience_spec.rb:16` From the screenshots, the page appears to be rendering blank, which makes me think something might be missing or broken in your current approach. It may be worth double checking that part of the implementation.

Not in my main branch either  the audience analytics specs are failing there as well. That’s why I’m asking you. In your screenshot all tests are passing, but in my main branch these tests are failing. Is there any specific command you’re using?



> > @4arjun I have one question. All your Audience One test cases are passing on the main branch, but my Audience One specs are failing. Could you tell me which command you’re using to run the tests?
> 
> @Devwill2, I don’t think this is related to the test command itself. The only failing specs I’m seeing in your PR are:
> 
> `rspec ./spec/requests/analytics/audience_spec.rb:12` `rspec ./spec/requests/analytics/audience_spec.rb:16` From the screenshots, the page appears to be rendering blank, which makes me think something might be missing or broken in your current approach. It may be worth double checking that part of the implementation.

Not in my main branch either  the audience analytics specs are failing there as well. That’s why I’m asking you. In your screenshot all tests are passing, but in my main branch these tests are failing. Is there any specific command you’re using?

sm17p:
@Devwill2 That's because your endpoint should be returning json

Screenshot 2025-12-28 at 8 58 53 PM

Screenshot 2025-12-28 at 8 59 30 PM

and with json you can't use symbols as you're doing in your [specs](https://github.com/antiwork/gumroad/pull/2484/changes#diff-3b4e360a182c9ef3b4872cda3c3214b82379ff650f6570b2f3f618c1a3c42d53R53)

```ruby
 expect(inertia.props[:audience_data]).to eq(
 
 # should be
 
 expect(inertia.props["audience_data"]).to eq(

```
and that should make your specs pass ✌️

Devwill2:
> @Devwill2 That's because your endpoint should be returning json
> 
> Screenshot 2025-12-28 at 8 58 53 PM Screenshot 2025-12-28 at 8 59 30 PM
> and with json you can't use symbols as you're doing in your [specs](https://github.com/antiwork/gumroad/pull/2484/changes#diff-3b4e360a182c9ef3b4872cda3c3214b82379ff650f6570b2f3f618c1a3c42d53R53)
> 
> ```ruby
>  expect(inertia.props[:audience_data]).to eq(
>  
>  # should be
>  
>  expect(inertia.props["audience_data"]).to eq(
> ```
> 
> and that should make your specs pass ✌️

@sm17p My spec is not failing, but the analytics audience spec is failing. It is also failing on main without any changes. That’s why I’m asking what the issue could be. Do you know why it’s failing on main as well, and why it’s happening only on my machine?

binary-koan:
See https://github.com/antiwork/gumroad/pull/2572; please let me know there if you have any feedback!

PR #2472, opened 2025-12-25
fix: Update Library creator counts to respect the "Show archived only" filter

### LINKED ISSUE #2469: Update Library creator counts to respect the "Show archived only" filter
### What

- [ ] Update Library creator counts to respect the "Show archived only" filter

###
Creator counts are currently calculated using all purchases, regardless of whether archived items are filtered. This causes the counts to mismatch what’s actually shown in the Library. Counts should always reflect what the user sees.

Image

romitg2:
fixes #2469 

# Before:
https://github.com/user-attachments/assets/5ef14847-801b-4cf6-8977-52fea0efd2f0

# After:
https://github.com/user-attachments/assets/b1e3a9ed-a7b7-45db-9d73-bcfc0e698bdf

### AI Disclosure:
- used claude-code sonnet-4.5 

### Test Result: (Local run)
Screenshot 2025-12-25 at 11 39 58 AM

### live stream disclosure:
- watched all live streams

### DIFF
--- a/app/javascript/pages/Library/Index.tsx
+++ b/app/javascript/pages/Library/Index.tsx
@@ -277,10 +277,7 @@ export default function LibraryPage() {
 
   const creatorsWithProductCounts = React.useMemo(() => {
     const productCountByCreatorId = state.results.reduce((counts, result) => {
-      const matchesArchiveFilter = result.purchase.is_archived === state.search.showArchivedOnly;
-      const isIndividualPurchase = !result.purchase.is_bundle_purchase;
-
-      if (matchesArchiveFilter && isIndividualPurchase) {
+      if (result.purchase.is_archived === state.search.showArchivedOnly && !result.purchase.is_bundle_purchase) {
         const creatorId = result.product.creator_id;
         counts.set(creatorId, (counts.get(creatorId) ?? 0) + 1);
       }


app/javascript/pages/Library/Index.tsx
```diff
@@ -275,6 +275,24 @@ export default function LibraryPage() {
     return filtered;
   }, [state.results, state.search]);
 
+  const filteredCreators = React.useMemo(() => {
```
romitg2: filter creator's list to respect showArchivedOnly filter.

romitg2:
@binary-koan could you please review this PR? Thanks!

app/javascript/pages/Library/Index.tsx
```diff
@@ -275,6 +275,24 @@ export default function LibraryPage() {
     return filtered;
   }, [state.results, state.search]);
 
+  const creatorsWithProductCounts = React.useMemo(() => {
+    const productCountByCreatorId = state.results.reduce((counts, result) => {
+      if (result.purchase.is_archived === state.search.showArchivedOnly && !result.purchase.is_bundle_purchase) {
+        const creatorId = result.product.creator_id;
+        counts.set(creatorId, (counts.get(creatorId) ?? 0) + 1);
```
nyomanjyotisa, line 282: We already have the creator count in the backend here

https://github.com/antiwork/gumroad/blob/da50d94463cb83a6f62f66ff4c274efe98212450/app/presenters/library_presenter.rb#L31-L33

We should just do the calculation once. If the calculation must be in the frontend due to filtering thing, we can remove the one in the backend. Otherwise I would prefer the backend calculation

nyomanjyotisa:
Closing this PR, since the solution cause two diferent calculation, in backend and in front which not needed.

Issue #2470, opened 2025-12-25
Show ended or cancelled memberships in Library when access is not revoked


vatsalkaushik:
### What

- [ ] Show memberships in the Library even if they are ended or cancelled, as long as access is not revoked

### Why
Currently, memberships that are ended or cancelled can disappear from the Library even when the creator has not revoked access (i.e. the "Members will lose access when their memberships end" setting is turned off). This is confusing for customers, as they still have valid access but no longer see the product in their Library. The Library should reflect actual access, not just subscription status.

<img width="1704" height="692" alt="Image" src="https://github.com/user-attachments/assets/2f2f35f6-e24a-40ba-82ca-62526670071f" />

PR #2473, opened 2025-12-25
Fix: Show ended or cancelled memberships in Library when access is not revoked

romitg2:
fixes: #2470 

## Root Cause:

In `app/presenters/library_presenter.rb:41`, we were using subscription.alive? to filter memberships:
```rb
next if purchase.link.is_recurring_billing && !purchase.subscription.alive?
```

- This always excluded any subscription that was cancelled, ended, or failed, regardless of the creator's "Members will lose access when their memberships end" setting (block_access_after_membership_cancellation).

## Solution:
Changed the filter to use subscription.grant_access_to_product? instead, to correctly respect ` block_access_after_membership_cancellation` setting. 
```rb
      next if purchase.link.is_recurring_billing && !purchase.subscription.grant_access_to_product?
```

```rb
def grant_access_to_product?
  if is_installment_plan?
    !cancelled_or_failed?
  else
    alive? || !link.block_access_after_membership_cancellation
  end
end
```

### Test Result: (Local run)
Screenshot 2025-12-25 at 12 00 36 PM

### AI Disclosure:
- used claude-code sonnet-4.5 to implement test

### live stream Disclosure:
- watched all live streams


romitg2:
@binary-koan could you please review this PR? Thanks!

hchhabra (approved):
Thanks for the fix, @romitg2 !

Issue #2480, opened 2025-12-16
Add "Order ID" as an option when using Search Purchase


schrodingrsbrat:
Currently, the "Search Purchase" tool is missing a field for Order ID. Lots of customers often don't remember the email used at checkout but will provide this ID - which can be must more efficient and accurate than searching using "date of purchase", "amount", "last four digits", etc.

chaitanya-mishra-ai:
Submitted a fix in PR #2498. The Helper Search Purchase API now supports searching by order_id alongside the existing email and product options.

ashutosh887:
Should we close this @slavingia @schrodingrsbrat if this is completed?

PR #2481, opened 2025-12-26
[#2480] Added Order ID as an option when using Search Purchase

vishalzambre:
UI changes

Screenshot 2025-12-26 at 11 46 13 AM


binary-koan:
Thanks for working on this, but another PR is very close to being merged

PR #2485, opened 2025-12-26

### LINKED ISSUE #2480: Add "Order ID" as an option when using Search Purchase

yashranaway:
# Fixes #2480
## What
When customers contact support, they often have their Order ID handy but can't remember which email they used at checkout. This adds a dedicated Order ID search field so support can quickly find purchases without having to guess at emails or search by card details.

## Changes
- Added an `order_id` parameter to the backend search service that looks up purchases directly by their external ID
- Updated the admin search popover UI to include a new Order ID input field (with a shopping bag icon 🛒)
- Wired up the Helper API so support tools can also search by Order ID

The new field shows up right in the search popover, between the affiliates search and the card search section.

## UI 

https://github.com/user-attachments/assets/95097b0d-a71b-429e-b1b2-e179e87ad324

## Tests
All the new order_id search tests pass:

## AI disclousure 

## Notes

### DIFF
--- a/app/controllers/admin/search/purchases_controller.rb
+++ b/app/controllers/admin/search/purchases_controller.rb
@@ -7,7 +7,7 @@ class Admin::Search::PurchasesController < Admin::BaseController
   def index
     @title = "Purchase results"
 
-    search_params = params.permit(:transaction_date, :last_4, :card_type, :price, :expiry_date, :purchase_status).to_hash.symbolize_keys
+    search_params = params.permit(:transaction_date, :last_4, :card_type, :price, :expiry_date, :purchase_status, :order_id).to_hash.symbolize_keys
 
     if search_params[:transaction_date].present?
       begin
--- a/app/controllers/api/internal/helper/purchases_controller.rb
+++ b/app/controllers/api/internal/helper/purchases_controller.rb
@@ -178,7 +178,7 @@ class Api::Internal::Helper::PurchasesController < Api::Internal::Helper::BaseCo
 
   SEARCH_PURCHASE_OPENAPI = {
     summary: "Search purchase",
-    description: "Search purchase by email, seller, license key, or card details. At least one of the parameters is required.",
+    description: "Search purchase by order ID, email, seller, license key, or card details. At least one of the parameters is required. Order ID is the most efficient search method when available.",
     requestBody: {
       required: true,
       content: {
@@ -186,6 +186,7 @@ class Api::Internal::Helper::PurchasesController < Api::Internal::Helper::BaseCo
           schema: {
             type: "object",
             properties: {
+              order_id: { type: "string", description: "Order ID (also known as purchase ID or external ID) - the unique identifier shown to customers on receipts" },
               email: { type: "string", description: "Email address of the customer/buyer" },
               creator_email: { type: "string", description: "Email address of the creator/seller" },
               license_key: { type: "string", description: "Product license key (4 groups of alphanumeric characters separated by dashes)" },
@@ -259,6 +260,7 @@ class Api::Internal::Helper::PurchasesController < Api::Internal::Helper::BaseCo
   }.freeze
   def search
     search_params = {
+      order_id: params[:order_id],
       query: params[:email],
       creator_email: params[:creator_email],
       license_key: params[:license_key],
--- a/app/javascript/components/Admin/SearchPopover.tsx
+++ b/app/javascript/components/Admin/SearchPopover.tsx
@@ -42,6 +42,7 @@ const SearchPopover = () => {
     user_query: initialQueries.user_query,
     purchase_query: initialQueries.purchase_query,
     affiliate_query: initialQueries.affiliate_query,
+    order_id: searchParams.get("order_id") || "",
     product_title_query: searchParams.get("product_title_query") || "",
     purchase_status: searchParams.get("purchase_status") || "",
     card_type: searchParams.get("card_type") || "",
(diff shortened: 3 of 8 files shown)


app/controllers/admin/search/purchases_controller.rb
```diff
@@ -7,7 +7,7 @@ class Admin::Search::PurchasesController < Admin::BaseController
   def index
     @title = "Purchase results"
 
-    search_params = params.permit(:transaction_date, :last_4, :card_type, :price, :expiry_date, :purchase_status).to_hash.symbolize_keys
+    search_params = params.permit(:transaction_date, :last_4, :card_type, :price, :expiry_date, :purchase_status, :order_id).to_hash.symbolize_keys
```
yashranaway, line 10: Just added `order_id` to the permitted params so it flows through from the form submission.

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -178,14 +178,15 @@ def resend_all_receipts
...
-    description: "Search purchase by email, seller, license key, or card details. At least one of the parameters is required.",
+    description: "Search purchase by order ID, email, seller, license key, or card details. At least one of the parameters is required. Order ID is the most efficient search method when available.",
     requestBody: {
       required: true,
       content: {
         'application/json': {
           schema: {
             type: "object",
             properties: {
+              order_id: { type: "string", description: "Order ID (also known as purchase ID or external ID) - the unique identifier shown to customers on receipts" },
```
yashranaway, line 189: Updated the OpenAPI schema to document the new order_id parameter for the Helper API.

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -259,6 +260,7 @@ def resend_all_receipts
   }.freeze
   def search
     search_params = {
+      order_id: params[:order_id],
```
yashranaway, line 263: Wired up `order_id` to actually get passed to the search service when the Helper API is called.

app/javascript/components/Admin/SearchPopover.tsx
```diff
@@ -52,28 +53,26 @@ const SearchPopover = () => {
   });
 
   const resetOtherQueryFields = (activeField?: keyof typeof data | null) => {
-    const queryFields = ["user_query", "purchase_query", "affiliate_query"] as const;
+    const queryFields = ["user_query", "purchase_query", "affiliate_query", "order_id"] as const;
```
yashranaway, line 56: Made sure `order_id` gets reset when switching between different search types.

app/javascript/components/Admin/SearchPopover.tsx
```diff
@@ -52,28 +53,26 @@ const SearchPopover = () => {
...
     const cardFields = ["card_type", "transaction_date", "last_4", "expiry_date", "price"] as const;
 
     if (activeField === null) {
       queryFields.forEach((field) => setData(field, ""));
       return;
     }
 
-    if (activeField && ["user_query", "purchase_query", "affiliate_query"].includes(activeField)) {
+    if (activeField && ["user_query", "purchase_query", "affiliate_query", "order_id"].includes(activeField)) {
```
yashranaway, line 64: Updated `resetOtherQueryFields` to include `order_id` in the cleanup.

app/javascript/components/Admin/SearchPopover.tsx
```diff
@@ -52,28 +53,26 @@ const SearchPopover = () => {
...
       queryFields.forEach((field) => {
         if (field !== activeField) setData(field, "");
       });
     }
   };
 
-  const submit = (endpoint: string, queryParam?: keyof typeof data | null) => {
-    const queryData = { query: queryParam ? String(data[queryParam]) : "" };
+  const submit = (endpoint: string, queryParam?: keyof typeof data | null, paramName = "query") => {
+    const queryValue: string = queryParam ? String(data[queryParam]) : "";
```
yashranaway, line 73: Modified submit function to accept a custom param name since order_id uses "order_id" instead of "query".

app/services/admin_search_service.rb
```diff
@@ -3,9 +3,16 @@
...
+  def search_purchases(query: nil, order_id: nil, product_title_query: nil, purchase_status: nil, creator_email: nil, license_key: nil, transaction_date: nil, last_4: nil, card_type: nil, price: nil, expiry_date: nil, limit: nil)
     purchases = Purchase.order(created_at: :desc)
 
+    if order_id.present?
+      purchase_id = Purchase.from_external_id(order_id.strip)
+      purchase_id ||= ObfuscateIds.decrypt_numeric(order_id.strip.to_i)
+      return purchases.where(id: purchase_id).limit(limit) if purchase_id
+      return Purchase.none
+    end
+
```
yashranaway, line 15: Test for the new order_id search verifies it works with both ID formats, handles invalid IDs, and takes priority over other search params when provided.

app/javascript/components/Admin/SearchPopover.tsx
binary-koan: Can we tweak this to just take the object as the param? i.e. `submit("some route", { query: "some query" })`

binary-koan:
Looks like we already support searching for external ID in `query`, so I think we can just add `query` to the Helper API instead. Does that work?

https://github.com/antiwork/gumroad/blob/35db4815f323a357ca99430083531883a4e0e9c7/app/services/admin_search_service.rb#L20

PR #2488, opened 2025-12-27
Added grammar exception for languages that modify noun stems for cases

erkinalp:


erkinalp:
I thought this would be uncontroversial, Slavic languages are an example that modify noun stems in cases, even in proper names.

PR #2492, opened 2025-12-28

### LINKED ISSUE #2480: Add "Order ID" as an option when using Search Purchase

yashranaway:
# Fixes #2480

Previous PR #2485 

## What
When customers contact support, they often have their Order ID handy but can't remember which email they used at checkout. This adds a dedicated Order ID search field so support can quickly find purchases without having to guess at emails or search by card details.

## Changes
- Added an `order_id` parameter to the backend search service that looks up purchases directly by their external ID
- Updated the admin search popover UI to include a new Order ID input field (with a shopping bag icon 🛒)
- Wired up the Helper API so support tools can also search by Order ID

The new field shows up right in the search popover, between the affiliates search and the card search section.

## UI 

https://github.com/user-attachments/assets/95097b0d-a71b-429e-b1b2-e179e87ad324

## Tests
All the new order_id search tests pass:

## AI disclousure 

## Notes

### DIFF
--- a/app/controllers/products/archived_controller.rb
+++ b/app/controllers/products/archived_controller.rb
@@ -71,7 +71,11 @@ class Products::ArchivedController < Sellers::BaseController
   def create
     authorize [:products, :archived, @product]
 
-    @product.update!(archived: true)
+    @product.update!(
+      archived: true,
+      purchase_disabled_at: @product.purchase_disabled_at || Time.current
+    )
+
     render json: { success: true }
   end
 
--- a/app/javascript/components/Checkout/GiftForm.tsx
+++ b/app/javascript/components/Checkout/GiftForm.tsx
@@ -3,7 +3,6 @@ import * as React from "react";
 
 import { Button } from "$app/components/Button";
 import { useState, getErrors } from "$app/components/Checkout/payment";
-import { Icon } from "$app/components/Icons";
 import { Modal } from "$app/components/Modal";
 import { Alert } from "$app/components/ui/Alert";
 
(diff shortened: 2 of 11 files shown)


yashranaway:
@binary-koan Refactored to use the existing `query` parameter as suggested and updated the Helper API to support it. 
ref: #2485

app/javascript/components/Admin/SearchPopover.tsx
```diff
@@ -42,6 +42,7 @@ const SearchPopover = () => {
     user_query: initialQueries.user_query,
     purchase_query: initialQueries.purchase_query,
     affiliate_query: initialQueries.affiliate_query,
+    order_id: searchParams.get("order_id") || "",
```
binary-koan, line 45: I don't think we need these changes? (since admin already supports searching by `query`)

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -178,15 +178,16 @@ def resend_all_receipts
...
     requestBody: {
       required: true,
       content: {
         'application/json': {
           schema: {
             type: "object",
             properties: {
-              email: { type: "string", description: "Email address of the customer/buyer" },
+              query: { type: "string", description: "Search query - can be order ID (external ID), email, IP address, or card fingerprint" },
+              email: { type: "string", description: "Email address of the customer/buyer (alternative to query)" },
```
binary-koan, line 190: Oh it's super weird we already have an `email` param but pass it as `query`. Could you remove this?

PR #2494, opened 2025-12-28
fix(exports): Prevent timeout on large audience CSV exports

### LINKED ISSUE #2092: Followers / Subscribers CSV not sending due to timeout issue
### What
When downloading a CSV of followers or subscribers here: [https://gumroad.com/followers](url) users with large amounts of data do not receive emails due to it timing out.

sahilagar:
## What

Fixes #2092 - Followers/Subscribers CSV not sending due to timeout issue.

## How

Wraps the audience export in `WithMaxExecutionTime.timeout_queries(seconds: 1.hour)` to increase the MySQL `max_execution_time` for queries within the job.

This follows the same pattern used by:
- `AnnualPayoutExportWorker` (1 hour)
- `AnnualTaxSummaryExportWorker` (4 hours)

## Why 1 Hour is Sufficient

The export query is simple and well-indexed:

| Factor | Estimate |
|--------|----------|
| Query | `SELECT id, email, min_created_at` (2 indexed columns) |
| Batch size | 1,000 records per batch (`find_each`) |
| Per-batch time | ~10ms |
| 1M records | ~1,000 batches × 10ms = ~10 seconds |

The timeout provides ample headroom. Happy to adjust if production metrics suggest otherwise.

## Changes

- Added `WithMaxExecutionTime` wrapper to `AudienceExportWorker`
- Added test to verify timeout is applied

## AI Disclosure

Used Cursor AI to help implement this fix.

## Tests
image

### DIFF
--- a/app/sidekiq/exports/audience_export_worker.rb
+++ b/app/sidekiq/exports/audience_export_worker.rb
@@ -7,15 +7,16 @@ class Exports::AudienceExportWorker
   def perform(seller_id, recipient_id, audience_options = {})
     seller, recipient = User.find(seller_id, recipient_id)
     recipient ||= seller
+    result = nil
 
     WithMaxExecutionTime.timeout_queries(seconds: 1.hour) do
       result = Exports::AudienceExportService.new(seller, audience_options).perform
-
-      ContactingCreatorMailer.subscribers_data(
-        recipient:,
-        tempfile: result.tempfile,
-        filename: result.filename,
-      ).deliver_now
     end
+
+    ContactingCreatorMailer.subscribers_data(
+      recipient:,
+      tempfile: result.tempfile,
+      filename: result.filename,
+    ).deliver_now
   end
 end


nsalewski:
Your analysis shows why `WithMaxExecutionTime` may not help: if each batch takes ~10ms, individual queries would never even hit the 5-minute default [`max_execution_time`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_execution_time). The workers you mentioned use longer timeouts because they call services with complex nested queries:
- `AnnualPayoutExportWorker` calls [`Payouts::Annual`](https://github.com/antiwork/gumroad/blob/main/app/services/exports/payouts/annual.rb) which inherits from [`Payouts::Base`](https://github.com/antiwork/gumroad/blob/main/app/services/exports/payouts/base.rb), nested `find_each` loops (balances → sales/chargebacks/refunds → credits) with multiple aggregate queries per balance
- `AnnualTaxSummaryExportWorker` calls `TaxSummary::Annual`, which iterates all users with multiple queries per user

`AudienceExportService` is a simple single-table iteration with no nested queries.

Without error logs, it's hard to confirm MySQL is the root cause.

That said, if `WithMaxExecutionTime` as a safety net would make sense, the mailer should be moved outside the block. Equivalent to `AnnualPayoutExportWorker`. Because:

1. `WithMaxExecutionTime` sets MySQL's `max_execution_time` session variable
2. Email delivery doesn't involve MySQL queries, so no benefit in including it in the block
3. If the mailer times out, it would raise a misleading `QueryTimeoutError`

So it would look like this:

```
def perform(seller_id, recipient_id, audience_options = {})
  seller, recipient = User.find(seller_id, recipient_id)
  recipient ||= seller
  result = nil

  WithMaxExecutionTime.timeout_queries(seconds: 1.hour) do
    result = Exports::AudienceExportService.new(seller, audience_options).perform
  end

  ContactingCreatorMailer.subscribers_data(
    recipient:,
    tempfile: result.tempfile,
    filename: result.filename,
  ).deliver_now
end
```

sahilagar:
@nsalewski Thank you for the thorough analysis! You're right that without error logs, it's hard to confirm whether `WithMaxExecutionTime` addresses the root cause. 

Since `AudienceExportService` uses a simple single-table iteration with `find_each` (already batched), individual queries likely won't hit the default timeout. The actual timeout could be at the Sidekiq job level, memory exhaustion, email delivery, or something else entirely.

That said, I've implemented your suggested refactor to move the mailer outside the timeout block (matching the `AnnualPayoutExportWorker` pattern) because:
1. It makes the code cleaner and more semantically correct
2. It ensures only DB queries benefit from the extended timeout
3. It prevents misleading `QueryTimeoutError` if email delivery times out

In an ideal scenario, we'd have production logs showing the actual timeout source. For now, this provides a defensive safety measure following existing patterns. If the issue persists in production, we'll have better evidence to investigate further.

EmCousin:
Unlikely this PR addresses the root cause.

Issue #2255, opened 2025-12-09
Why are users forced to fill in tip field for free products?


jordibruin:
I really wish these kind of changes would get communicated more. It feels as if I'm forcing users to submit a tip, even for users whom I give a 100% off discount code.

Please make this an option.

slavingia:
Worth a tweet?

PR #2511, opened 2025-12-28

### LINKED ISSUE #2480: Add "Order ID" as an option when using Search Purchase

yashranaway:
# Fixes #2480

Previous PR #2485 & #2492

## What
When customers contact support, they often have their Order ID handy but can't remember which email they used at checkout. This adds Order ID search support to the Helper API so support tools can quickly find purchases.

## Changes
Per reviewer feedback, simplified to leverage the existing `query` parameter which already handles external ID lookups via `Purchase.from_external_id()`:

- Added `query` parameter to Helper API OpenAPI schema (documents the existing capability)
- Updated controller to accept `query` param: `query: params[:query] || params[:email]`
- Added one test for Order ID search via query param

The admin search already supports Order ID via the query field - this just exposes the same capability to the Helper API.

## Tests
Added simple test for Order ID search 

## AI disclosure

## Notes

### DIFF
--- a/spec/controllers/api/internal/helper/purchases_controller_spec.rb
+++ b/spec/controllers/api/internal/helper/purchases_controller_spec.rb
@@ -383,10 +383,6 @@ describe Api::Internal::Helper::PurchasesController, :vcr do
         expect(response.parsed_body).to eq({ success: true, message: "Purchase found", purchase: purchase_json }.as_json)
       end
     end
-  end
-
-  describe "POST search with query parameter" do
-    include_examples "helper api authorization required", :post, :search
 
     it "returns purchase data when searching by order ID (external_id) via query param" do
       purchase = create(:free_purchase)
--- a/spec/support/fixtures/vcr_cassettes/Api_Internal_Helper_PurchasesController/POST_search/returns_not_found_when_order_ID_doesn_t_match_the_customer_email.yml
+++ b/spec/support/fixtures/vcr_cassettes/Api_Internal_Helper_PurchasesController/POST_search/returns_not_found_when_order_ID_doesn_t_match_the_customer_email.yml
@@ -1,0 +1,624 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.stripe.com/v1/payment_methods
+    body:
+      encoding: UTF-8
+      string: type=card&card[token]=tok_visa
+    headers:
+      User-Agent:
+      - Stripe/v1 RubyBindings/12.5.0
+      Authorization:
+      - Bearer <STRIPE_API_KEY>
+      Content-Type:
+      - application/x-www-form-urlencoded
+      X-Stripe-Client-Telemetry:
+      - '{"last_request_metrics":{"request_id":"req_kUKgtoVlKNdFGD","request_duration_ms":474}}'
+      Idempotency-Key:
+      - '021458aa-67a0-4943-8e3f-cca97721af56'
+      Stripe-Version:
+      - '2023-10-16'
+      X-Stripe-Client-User-Agent:
+      - '{"bindings_version":"12.5.0","lang":"ruby","lang_version":"3.4.3 p32 (2025-04-14)","platform":"arm64-darwin24","engine":"ruby","publisher":"stripe","uname":"Darwin
+        Jonos-MacBook-Pro-2.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14
+        11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041 arm64","hostname":"Jonos-MacBook-Pro-2.local"}'
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Server:
+      - nginx
+      Date:
+      - Tue, 30 Dec 2025 23:45:48 GMT
+      Content-Type:
+      - application/json
+      Content-Length:
+      - '1083'
+      Connection:
+      - keep-alive
+      Access-Control-Allow-Credentials:
+      - 'true'
+      Access-Control-Allow-Methods:
+      - GET, HEAD, PUT, PATCH, POST, DELETE
+      Access-Control-Allow-Origin:
+      - "*"
+      Access-Control-Expose-Headers:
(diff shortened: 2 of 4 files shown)


yashranaway:
@binary-koan sorry about the complex attempt last time, tried using a different and minimal approach this time, requesting a review 🙏

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -178,17 +178,18 @@ def resend_all_receipts
...
       content: {
         'application/json': {
           schema: {
             type: "object",
             properties: {
+              query: { type: "string", description: "Search query (order ID, email, IP address, or card fingerprint)" },
               email: { type: "string", description: "Email address of the customer/buyer" },
               creator_email: { type: "string", description: "Email address of the creator/seller" },
-              license_key: { type: "string", description: "Product license key (4 groups of alphanumeric characters separated by dashes)" },
+              license_key: { type: "string", description: "Product license key" },
```
binary-koan: Please revert this change

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -178,17 +178,18 @@ def resend_all_receipts
...
+    description: "Search purchase by query (order ID, email, IP, or card fingerprint), seller, license key, or card details. At least one parameter is required.",
     requestBody: {
       required: true,
       content: {
         'application/json': {
           schema: {
             type: "object",
             properties: {
+              query: { type: "string", description: "Search query (order ID, email, IP address, or card fingerprint)" },
               email: { type: "string", description: "Email address of the customer/buyer" },
```
binary-koan, line 190: Sorry, just realised something - we need this parameter to always work because we use it to make sure we search the correct customer's purchases in Helper chat.

Could you update the `AdminSearchService` to take an explicit `email` parameter as well as `query` so that we can use them in parallel? (no need to change how query or the admin UI works, this can just be for the Helper endpoint)

app/controllers/api/internal/helper/purchases_controller.rb
```diff
@@ -260,7 +260,8 @@ def resend_all_receipts
   }.freeze
   def search
     search_params = {
-      query: params[:query] || params[:email],
+      query: params[:query],
+      email: params[:email],
```
yashranaway: Now passes query and email as separate params instead of fallback, so they can be used together to find a specific customer's purchase by order ID.

spec/services/admin_search_service_spec.rb
```diff
@@ -38,6 +38,32 @@
...
+      end
+
+      it "returns empty when query matches but email doesn't" do
+        purchase = create(:purchase, email: "customer@example.com")
+
+        purchases = AdminSearchService.new.search_purchases(query: purchase.external_id, email: "wrong@example.com")
+        expect(purchases).to be_empty
+      end
+    end
+
```
yashranaway, line 66: Tests for email parameter: filtering by email alone, using email+query together, and verifying empty result when email doesn't match the order ID's purchase.

app/services/admin_search_service.rb
yashranaway, line 1: Added explicit `email` parameter that filters purchases before applying other search criteria. This allows Helper API to use both `query` (for order ID) and `email` (to scope to customer) in parallel.

spec/controllers/api/internal/helper/purchases_controller_spec.rb
```diff
@@ -387,6 +387,30 @@
...
+      end
+
+      it "returns not found when order ID doesn't match the customer email" do
+        purchase = create(:purchase, email: "customer@example.com")
+
+        params = { query: purchase.external_id, email: "different@example.com", timestamp: Time.now.to_i }
+        post :search, params: params
+
+        expect(response).to have_http_status(:not_found)
+      end
```
yashranaway: Tests for using query and email together, plus a test confirming 404 when order ID exists but belongs to a different customer email.

yashranaway (commented):
@binary-koan _a added some changes please review.

yashranaway:
Off-topic, but since you reviewed the last PR, I wanted to flag this in case it was missed: https://github.com/antiwork/gumroad/issues/2255#issuecomment-3694378676. Please ignore if already seen.

spec/controllers/api/internal/helper/purchases_controller_spec.rb
```diff
@@ -383,6 +383,47 @@
...
     end
+
+    context "when searching by query parameter" do
+      it "returns purchase data when searching by order ID (external_id) via query param" do
+        purchase = create(:purchase)
+        purchase_json = purchase.slice(:email, :link_name, :price_cents, :purchase_state, :created_at)
+        purchase_json[:id] = purchase.external_id_numeric
+        purchase_json[:seller_email] = purchase.seller_email
+        purchase_json[:receipt_url] = receipt_purchase_url(purchase.external_id, host: UrlService.domain_with_protocol, email: purchase.email)
+        purchase_json[:refund_status] = nil
```
binary-koan: I find this quite a strange way of asserting, but it's consistent with existing tests 🤷

binary-koan:
Could you check the failing tests?

yashranaway:
On it 🫡

yashranaway:
@binary-koan added a fix, the rest of the failures are unrelated, CI should be green now 🤞

binary-koan:
Sorry why do we need backward compatibility? (it shouldn't cause problems in production if we change the email parameter behaviour)

yashranaway:
> Sorry why do we need backward compatibility? (it shouldn't cause problems in production if we change the email parameter behaviour)

Makes sense! Added explicit email `param` to `AdminSearchService`so query (order ID) and email can work in parallel.

yashranaway:
The issue is the existing tests pass email expecting it to search PayPal's `card_visual` field (not just purchases.email). If I go fully simple with separate query and email params, those tests fail because email would only filter by exact email match. Happy to simplify fully if you're okay with updating those tests to use query for PayPal searches instead?

binary-koan:
Yep I'm good with that!

yashranaway:
> Yep I'm good with that!

Updated tests!

yashranaway:
@binary-koan Update: Moved the new tests to a separate describe block to avoid inheriting the before hook that requires VCR cassettes. All tests pass locally now.

image
image


rest of the failures are unrelated

yashranaway:
The failing test is unrelated.

yashranaway:
@binary-koan New failures, but they seem unrelated

binary-koan:
Yep seen that one flake before, merging

PR #2317, opened 2025-12-12
Migrate support

yashranaway:
Part of #856 

Previous PR #2304

This PR migrates the following page to Inertia:

- /support

From Inertia pages (Dashboard, Workflows, etc.) → /support: Currently you need to go through /help first (which causes a full reload since /help is not Inertia)
From /support → Other Inertia pages: Yes, SPA navigation works (no full reload)
Within /support: All interactions are SPA (no reloads)

## Before

image

## After (no visual changes)

image

> [!NOTE]
> That 404 page is missing 
In production:
The real HELPER_WIDGET_HOST (e.g., https://helper.gumroad.com) is configured
The page will work correctly
In local development:
Without the real Helper API, the page loads but shows "Something went wrong" because the Helper client can't connect

## AI disclosure 
Used Claude Sonnet 4.5 via Cursor
All code manually reviewed by me 

## Confirmation

 I watched all livestreams

PR #2520, opened 2025-12-28
Remove inertia options from URL query params for support page

### LINKED ISSUE #2539: Bug: URL query params for support page displays inertia related options
This bug got introduced during inertia migration of support page #2317, where the URL incorrectly includes inertia visit options as query params. 

The URL should only have required query params of `id=dc629ea0c93f926709aa97cf08346c9` as per previous implementation instead of `preserveState=true&preserveScroll=true&id=dc629ea0c93f926709aa97cf08346c94`

https://github.com/user-attachments/assets/3b70d4cf-7121-4989-b091-66ac111298fc

sm17p:
### Closes #2539 
- #2539 

This bug got introduced during inertia migration of support page #2317, where the URL incorrectly includes inertia visit options as query params 

## Before 
- URL is `https://gumroad.dev/support?preserveState=true&preserveScroll=true&id=dc629ea0c93f926709aa97cf08346c94`

https://github.com/user-attachments/assets/3b70d4cf-7121-4989-b091-66ac111298fc

## After 
- URL is `https://gumroad.com/support?id=dc629ea0c93f926709aa97cf08346c94`
- ~~Enabled `viewTransition` to  smoothen up navigation~~

https://github.com/user-attachments/assets/9f957007-86d0-47e3-85ce-60194e0d4010

### AI Disclosure
No Usage

### Live Stream Disclosure 
Have watched all the live streams

### DIFF
--- a/app/controllers/wishlists_controller.rb
+++ b/app/controllers/wishlists_controller.rb
@@ -32,11 +32,16 @@ class WishlistsController < ApplicationController
   def create
     authorize Wishlist
 
-    wishlist = current_seller.wishlists.create!
+    wishlist = current_seller.wishlists.new(params.require(:wishlist).permit(:name))
 
     respond_to do |format|
-      format.html { redirect_to wishlists_path, notice: "Wishlist created!", status: :see_other }
-      format.json { render json: { wishlist: WishlistPresenter.new(wishlist:).listing_props }, status: :created }
+      if wishlist.save
+        format.html { redirect_to wishlists_path, notice: "Wishlist created!", status: :see_other }
+        format.json { render json: { wishlist: WishlistPresenter.new(wishlist:).listing_props }, status: :created }
+      else
+        format.html { redirect_to wishlists_path, inertia: { errors: { base: wishlist.errors.full_messages } }, status: :see_other }
+        format.json { render json: { error: wishlist.errors.full_messages.first }, status: :unprocessable_entity }
+      end
     end
   end
 
--- a/app/javascript/components/Product/ShareSection.tsx
+++ b/app/javascript/components/Product/ShareSection.tsx
@@ -35,6 +35,9 @@ export const ShareSection = ({
   const [saveState, setSaveState] = React.useState<
     { type: "initial" | "saving" } | ({ type: "success" } & SuccessState)
   >({ type: "initial" });
+  const [dropdownState, setDropdownState] = React.useState<
+    { state: "closed" } | { state: "open" } | { state: "creating"; newWishlistName: string }
+  >({ state: "closed" });
 
   const isSelectionInWishlist = (wishlist: WishlistForProduct) =>
     wishlist.selections_in_wishlist.some(
@@ -47,6 +50,7 @@ export const ShareSection = ({
 
   const addProduct = async (resolveWishlist: Promise<SuccessState>) => {
     setSaveState({ type: "saving" });
+    setDropdownState({ state: "closed" });
 
     try {
       const { newlyCreated, wishlist } = await resolveWishlist;
@@ -81,8 +85,8 @@ export const ShareSection = ({
     }
   };
 
-  const newWishlist = async () => {
-    const { wishlist } = await createWishlist();
+  const newWishlist = async (name: string): Promise<SuccessState> => {
+    const { wishlist } = await createWishlist(name);
     setWishlists([...wishlists, { ...wishlist, selections_in_wishlist: [] }]);
     return { newlyCreated: true, wishlist };
   };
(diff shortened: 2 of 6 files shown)


sm17p:
@binary-koan @EmCousin I believe this should be a quick fix, please have a look when you can 🙏

sm17p:
Hi @ershad, I'm unsure what issue to tag this under. Should we create a separate issue for inertia migration related bug fixes or keep it under #856 ?

binary-koan:
What does it look like without `viewTransition`? We don't seem to have it on other pages so maybe better to be consistent

app/javascript/components/support/SupportPortal.tsx
```diff
@@ -18,7 +18,7 @@ export default function SupportPortal() {
     const url = new URL(location.href);
     if (!isNewTicketOpen && url.searchParams.get("new_ticket")) {
       url.searchParams.delete("new_ticket");
-      router.get(url.toString(), { replace: true, preserveState: true, preserveScroll: true });
+      router.get(url.toString(), {}, { viewTransition: true });
```
EmCousin: `replace: true, preserveState: true, preserveScroll: true` should not be removed but moved to the options object in lieu of viewTransition

app/javascript/components/support/SupportPortal.tsx
```diff
@@ -30,7 +30,7 @@ export default function SupportPortal() {
       url.searchParams.delete("id");
     }
     if (url.toString() !== window.location.href) {
-      router.get(url.toString(), { preserveState: true, preserveScroll: true });
+      router.get(url.toString(), {}, { viewTransition: true });
```
EmCousin: Same here

EmCousin (changes requested):
Thanks for the catch @sm17p 
I personally like view transitions, but they are out of scope of this PR and I think implementing them should be part of a broader effort to ensure consistent transition effects across the app's pages: something to be discussed with the design team first.

Also, those options were there for a reason, let's keep them

EmCousin:
Also @sm17p feel free to create a dedicated issue whenever you encounter a bug/defect like this, then you can link any PR to it :pray:

app/javascript/components/support/SupportPortal.tsx
sm17p: I think the original code before inertia migration just wanted to clear the query param on close? So maybe we don't need those other options. Can you please confirm this?

https://github.com/user-attachments/assets/666c2656-b005-46c7-8cc1-31f544657931

app/javascript/components/support/SupportPortal.tsx
sm17p: Yes added back over here

sm17p: `router.get` is making a server request though over here. I think it should be fine with just to use client side routing? Seems to be working over here

### Demos

1. With `router.get(url.toString(), {}, { preserveState: true, preserveScroll: true })`

https://github.com/user-attachments/assets/d1f6678e-d249-40d9-bfe2-a39a2181ad93

2. `router.push({ url: url.toString(), preserveScroll: true ~~, preserveState: true~~ })`      

https://github.com/user-attachments/assets/0439b5ae-6cae-42c2-8218-fa450e258805

app/javascript/components/support/SupportPortal.tsx
```diff
@@ -18,19 +18,20 @@ export default function SupportPortal() {
...
-      router.get(url.toString(), { replace: true, preserveState: true, preserveScroll: true });
+      router.replace({ url: url.toString() });
     }
   }, [isNewTicketOpen]);
 
   useEffect(() => {
     const url = new URL(location.href);
     if (selectedConversationSlug) {
       url.searchParams.set("id", selectedConversationSlug);
+      url.searchParams.delete("new_ticket");
```
sm17p, line 29: This was needed because when a ticket is created from `support?new_ticket=true` path `location` exits with `new_ticket` param during this hook call.

`router.replace({ url: url.toString() })` from the hook above is still in a processing queue which is asynchronous. 

The previous route change call (`router.replace({ url: url.toString() })`) is thrashed or cancelled I think in the queue and hence it wouldn't delete `new_ticket` param by the time execution reaches over here

Ref - https://github.com/search?q=repo%3Ainertiajs%2Finertia%20clientVisitQueue&type=code

Image

sm17p:
@binary-koan @EmCousin This is how it looks without view transition. I've reverted `viewTransition`, keeping `preserveScroll` enabled while selecting a conversation. Please have a look again. 

https://github.com/user-attachments/assets/32fdccbf-71a7-4af1-a1e3-5e0ccc1dbf8f

app/javascript/components/support/SupportPortal.tsx
EmCousin: That makes sense! The fix looks good. I'll run CI

sm17p:
(just a light ping over here)

EmCousin:
Thanks @sm17p !

Issue #1410, opened 2025-09-30
Offer code not applied to non-first installments when max usage limit is reached


vatsalkaushik:
### What

When an offer code has a max usage limit, it is applied correctly to the first installment but not to subsequent installments if the purchase count has already been reached.

Example:
- Customer begins an installment purchase with a valid offer code.
- The first installment reflects the discount.
- Subsequent installments ignore the discount if the code’s max purchase limit is reached in the meantime.

### Expected behavior
If an offer code is valid at the time of the initial purchase, the discount should apply consistently across all installments, regardless of whether the code’s max purchase count is later exceeded.

iedayan:
Does this issue still need to be resolved and fixed i.e. is it still open? If it is then I'll work on it.

PR #2522, opened 2025-12-28
fix(installments): persist offer code discount across all installments

bryanm1529:
Issue: #1410

# Description

## Problem

When an offer code has a max usage limit, it is applied correctly to the first installment but not to subsequent installments if the purchase count has already been reached. This happens because the system re-evaluates the offer code validity on each installment payment instead of using the discount that was locked in at purchase time.

## Solution

Snapshot the offer code data in `InstallmentPlanSnapshot` at purchase time using the existing `JsonData` concern pattern. This persists the discount terms (code, amount, percentage, duration, currency) so subsequent installments use the locked-in discount regardless of the offer code's current state (deleted, expired, modified, or usage limit reached).

Key changes:
- Add `json_data` column to `installment_plan_snapshots` table
- Include `JsonData` concern with `offer_code_snapshot` accessor
- Modify `set_price_and_rate` in Purchase to use locked discount from snapshot
- Add backfill service for existing installment plans

---

# Before/After

N/A - Backend-only change, no UI modifications.

---

# Test Results

Test Results - 45 examples, 0 failures

Tests cover:
- Discount preserved after offer code deletion
- Discount preserved after offer code amount modification
- Percentage and fixed amount discounts
- Backfill services with error handling
- Usage limits don't affect subsequent installments

---

# Checklist

---

# AI Disclosure

Model: Claude Opus 4.5 via Claude Code CLI
Used for: Implementation planning, code generation, test writing
All code reviewed and verified by running test suite

### DIFF
--- a/spec/models/installment_plan_snapshot_spec.rb
+++ b/spec/models/installment_plan_snapshot_spec.rb
@@ -171,8 +171,8 @@ describe InstallmentPlanSnapshot do
         snapshot = create(:installment_plan_snapshot, payment_option: payment_option)
         snapshot.snapshot_offer_code!(offer_code)
 
-        expect(snapshot.locked_discount_amount_off(10000)).to eq(2000)  # 20% of 10000
-        expect(snapshot.locked_discount_amount_off(5000)).to eq(1000)   # 20% of 5000
+        expect(snapshot.locked_discount_amount_off(10000)).to eq(2000)
+        expect(snapshot.locked_discount_amount_off(5000)).to eq(1000)
       end
 
       it "returns fixed amount for fixed discounts" do
@@ -212,7 +212,7 @@ describe InstallmentPlanSnapshot do
         offer_code.update!(amount_percentage: 50)
 
         snapshot.reload
-        expect(snapshot.locked_discount_percentage).to eq(20)  # Original value preserved
+        expect(snapshot.locked_discount_percentage).to eq(20)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


bryanm1529:
## Self-Review Notes

I've reviewed my own code and here are the key design decisions:

### 1. JsonData Concern Pattern (`app/models/installment_plan_snapshot.rb:4`)
Using the existing `JsonData` concern instead of 7 dedicated columns. This pattern is used by 18+ models in the codebase. Benefits:
- Single migration vs 7 DDL statements
- Easy to extend without schema changes
- Consistent with existing patterns

### 2. Installment Plan Offer Code Handling (`app/models/subscription.rb:251`)
Changed the condition from `discount_applies_to_next_charge?` to also check `&& !is_installment_plan`. This prevents installment payments from using the live offer code, forcing them to use the snapshotted values instead.

### 3. Locked Discount Fallback (`app/models/purchase.rb:1669-1678`)
Added prioritized logic: installment payments with locked snapshots use those values first, before falling back to the live offer code. This ensures subsequent installments get the discount even if the code is deleted/expired/modified.

### 4. Two Backfill Scripts
- `BackfillPaymentOptionInstallmentSnapshots` - creates missing snapshots (enhanced to capture offer codes)
- `BackfillOfferCodeToInstallmentSnapshots` - adds offer code data to existing snapshots

This separation allows running them independently if needed.

bryanm1529:
@ershad @binary-koan This PR is ready for review.

PR #2523, opened 2025-12-28
Replace BGN with EUR for Bulgaria

none34829:
- Replace Bulgaria default/payout currency BGN -> EUR
- Update Bulgaria bank account currency and help center table
- Update related specs


hchhabra:
Thanks for the PR, but this has been fixed in #2437  which was opened earlier.

PR #2524, opened 2025-12-28
fix: replace BGN with EUR

### LINKED ISSUE #2435: Replace BGN with EUR

mahmudsudo:
This PR updates Gumroad to:

Add Bulgaria (BG) to the list of countries using EUR as default currency
Change Bulgaria's payout currency from BGN to EUR
Update BulgariaBankAccount model to return EUR
Remove BGN from supported payout currencies
Update corresponding tests and expectations
This prevents payment and account issues for Bulgarian merchants as required by Stripe's transition timeline.

Testing: All existing tests pass with updated expectations. No breaking changes to other currencies.
closes #2435

### DIFF
--- a/app/views/help_center/articles/contents/_13-getting-paid.html.erb
+++ b/app/views/help_center/articles/contents/_13-getting-paid.html.erb
@@ -88,7 +88,7 @@
           <td>Brunei</td>
           <td>BND</td>
           <td>Bulgaria</td>
-          <td>BGN</td>
+          <td>EUR</td>
         </tr>
         <tr>
           <td>Cambodia</td>


hchhabra:
Thanks for the PR, but this has been fixed in #2437 which was opened earlier.

PR #2527, opened 2025-12-29
Fix race condition in Sales CSV export email notification (#1930)

### LINKED ISSUE #1930: Email notification not appearing when exporting Sales CSV from all time

dhruvd2018:
Closes #1930

Problem: > Users reported that "All Time" sales CSV exports would finish processing but the email notification (via ContactingCreatorMailer.user_sales_data) would never arrive.

Cause: > A race condition in ProcessChunkWorker. When multiple chunks finished simultaneously, the chunks_left_to_process? check would occasionally return true for all finishing workers, meaning the final CompileChunksWorker was never enqueued.

Solution: Refactored ProcessChunkWorker to remove the flaky count-based logic. It now leverages Sidekiq's until_executed lock on the CompileChunksWorker. By attempting to trigger the compiler whenever a chunk finishes and no unprocessed chunks are visible, we ensure the email is sent even if multiple jobs finish at the same time.

### DIFF
--- a/app/sidekiq/exports/sales/process_chunk_worker.rb
+++ b/app/sidekiq/exports/sales/process_chunk_worker.rb
@@ -2,7 +2,7 @@
 
 class Exports::Sales::ProcessChunkWorker
   include Sidekiq::Job
-  # This job is unique because two parallel jobs running `chunks_left_to_process?` could queue the same chunk to be reprocessed.
+  # This job is unique because two parallel jobs could queue the same chunk to be reprocessed.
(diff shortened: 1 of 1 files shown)


dhruvd2018:
I have submitted PR #2527 to resolve the race condition preventing emails on large exports.

binary-koan:
Thanks for working on this, but the issue is that the flash message doesn't show; I don't believe we have any problems with the email itself

PR #2529, opened 2025-12-29
Fix export timeout for large audience CSV exports

### LINKED ISSUE #2092: Followers / Subscribers CSV not sending due to timeout issue

4arjun:
Fixes #2092
## Summary
Wraps the audience export in `WithMaxExecutionTime.timeout_queries(seconds: 1.hour)` to prevent timeout when exporting large lists of followers/subscribers.
## Problem
Users with large audiences were not receiving their CSV export emails because the MySQL query was exceeding the default 5-minute `max_execution_time` configured in `database.yml`.
## Solution
- Added `WithMaxExecutionTime` wrapper to extend the query timeout to 1 hour
- Kept the mailer outside the timeout block to avoid misleading `QueryTimeoutError` on email delivery issues
This follows the same pattern used by:
- `AnnualPayoutExportWorker` (1 hour)
- `AnnualTaxSummaryExportWorker` (4 hours)
## Tests
Screenshot 2025-12-29 at 1 20 01 PM

## AI Disclosure
Used Claude Opus 4.5(Cursor) to investigate and implement this fix.
## Self-Review
I have reviewed all the code changes

## Confirmation
I have watched the Gumroad PR reviews live streaming video.

### DIFF
--- a/app/sidekiq/exports/audience_export_worker.rb
+++ b/app/sidekiq/exports/audience_export_worker.rb
@@ -7,8 +7,11 @@ class Exports::AudienceExportWorker
   def perform(seller_id, recipient_id, audience_options = {})
     seller, recipient = User.find(seller_id, recipient_id)
     recipient ||= seller
+    result = nil
 
-    result = Exports::AudienceExportService.new(seller, audience_options).perform
+    WithMaxExecutionTime.timeout_queries(seconds: 1.hour) do
+      result = Exports::AudienceExportService.new(seller, audience_options).perform
+    end
 
     ContactingCreatorMailer.subscribers_data(
       recipient:,
--- a/spec/sidekiq/exports/audience_export_worker_spec.rb
+++ b/spec/sidekiq/exports/audience_export_worker_spec.rb
@@ -25,5 +25,10 @@ describe Exports::AudienceExportWorker do
       mail = ActionMailer::Base.deliveries.last
       expect(mail.to).to eq([recipient.email])
     end
+
+    it "uses extended query timeout for large exports" do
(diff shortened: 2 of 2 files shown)


app/sidekiq/exports/audience_export_worker.rb
```diff
@@ -7,8 +7,11 @@ class Exports::AudienceExportWorker
   def perform(seller_id, recipient_id, audience_options = {})
     seller, recipient = User.find(seller_id, recipient_id)
     recipient ||= seller
+    result = nil
```
4arjun: Initialize result before the block so it's accessible after the block completes.

app/sidekiq/exports/audience_export_worker.rb
```diff
@@ -7,8 +7,11 @@ class Exports::AudienceExportWorker
...
 
-    result = Exports::AudienceExportService.new(seller, audience_options).perform
+    WithMaxExecutionTime.timeout_queries(seconds: 1.hour) do
+      result = Exports::AudienceExportService.new(seller, audience_options).perform
+    end
```
4arjun, line 14: Wrap only the database intensive export operation in WithMaxExecutionTime. This extends MySQL's max_execution_time from the default 5 minutes (configured in database.yml:25) to 1 hour, allowing large audience exports to complete without timing out.

4arjun:
@ershad Could we confirm the root cause? What's the actual error in production logs?

PR #2532, opened 2025-12-29
feat(exports): add AudienceExport and AudienceExportChunk models

FusionBrah:
Issue: #2092

# Description

## Problem

Large audience exports (followers/subscribers CSV) timeout because the entire export happens in a single Sidekiq job. Users with large audiences never receive their export emails.

## Solution

Add database models to support a chunked export approach (following the existing `SalesExport` pattern):

- `AudienceExport` - tracks the overall export request with seller/recipient associations
- `AudienceExportChunk` - stores processed chunks of audience data

This is **Part 1 of 2**. Part 2 (#2533) adds the Sidekiq jobs that use these models.

---

# Before/After

Backend-only change

---

# Test Results

Screenshot 2026-01-02 at 10 25 19 pm

---

# Checklist

---

# AI Disclosure

- Code implementation (model definitions, migrations, validations)
- Test writing (RSpec specs for models and factories)
- Debugging test failures related to JSON schema validation


FusionBrah (commented):
**Self-review comments:**

### `audience_export.rb` line 5
Separate `recipient` from `seller` allows admins to trigger exports on behalf of sellers and receive the email themselves (existing pattern in the codebase).

### `audience_export.rb` line 6
Using `dependent: :delete_all` instead of `destroy` for performance - chunks don't have callbacks that need to run.

### `audience_export.rb` line 8
YAML serialization for `audience_options` matches the pattern used elsewhere in the codebase for storing hash data.

### `audience_export.rb` lines 20-31
External ID generation with retry loop prevents collisions. Following the Nano ID pattern recommended in the contributing guidelines for public-facing IDs.

### `audience_export_chunk.rb` lines 6-7
Two serialized arrays: `member_ids` stores IDs before processing, `members_data` stores the processed [email, created_at] pairs. This separation allows parallel processing - each chunk can be processed independently.

### Migration files
Using `text` type for serialized columns to avoid MySQL's 64KB limit on `string` columns when dealing with large arrays.

app/models/audience_export.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class AudienceExport < ApplicationRecord
+  belongs_to :seller, class_name: "User"
+  belongs_to :recipient, class_name: "User"
```
FusionBrah, line 5: Separate `recipient` from `seller` allows admins to trigger exports on behalf of sellers and receive the email themselves. This follows the existing pattern in the codebase.

app/models/audience_export.rb
```diff
@@ -0,0 +1,32 @@
...
+  has_many :chunks, class_name: "AudienceExportChunk", foreign_key: :export_id, dependent: :delete_all
```
FusionBrah, line 6: Using `dependent: :delete_all` instead of `:destroy` for performance - chunks don't have callbacks that need to run.

app/models/audience_export.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  serialize :audience_options, type: Hash, coder: YAML
```
FusionBrah, line 8: YAML serialization for `audience_options` matches the pattern used elsewhere in the codebase for storing hash data.

app/models/audience_export.rb
```diff
@@ -0,0 +1,32 @@
...
+  validates :audience_options, presence: true
+
+  before_validation :set_external_id, on: :create
+
+  private
+    def set_external_id
+      self.external_id ||= self.class.generate_external_id
+    end
+
+    def self.generate_external_id(max_retries: 10)
```
FusionBrah: External ID generation with retry loop prevents collisions. Using alphanumeric IDs as recommended in the contributing guidelines for public-facing identifiers.

app/models/audience_export.rb
```diff
@@ -0,0 +1,38 @@
...
+  has_many :chunks, class_name: "AudienceExportChunk", foreign_key: :export_id, dependent: :delete_all
+
+  serialize :audience_options, type: Hash, coder: YAML
+
+  validates :external_id, presence: true, uniqueness: true
+  validates :audience_options, presence: true
+
+  before_validation :set_external_id, on: :create
+
+  def save(**options, &block)
```
FusionBrah, line 15: **Code Review Fix:** Refactored external_id generation to handle race conditions properly. The original `while exists?` loop had a TOCTOU (time-of-check-time-of-use) race condition - between checking and saving, another process could insert the same ID. Now the retry logic happens at the DB level: if `RecordNotUnique` is raised, we regenerate and retry up to 10 times.

db/migrate/20251229113512_create_audience_export_chunks.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateAudienceExportChunks < ActiveRecord::Migration[7.1]
+  def change
+    create_table :audience_export_chunks do |t|
+      t.bigint :export_id, null: false, index: true
+      t.longtext :member_ids
+      t.longtext :members_data
+      t.boolean :processed, default: false, null: false
+      t.string :revision
```
FusionBrah, line 10: **Code Review Fix:** Added `revision` column to match the `SalesExportChunk` pattern. This may be needed for optimistic locking or versioning in Part 2's chunk processing jobs.

app/models/audience_export.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+  end
+
+  private
+    def set_external_id
+      self.external_id ||= self.class.generate_external_id
+    end
+
+    def self.generate_external_id
```
FusionBrah, line 35: **Code Review Fix:** Simplified this method - the collision checking is now handled by the `save` method above. With 36^12 (~4.7 quintillion) possible values, the chance of collision is astronomically low, and the DB constraint + retry logic handles the rare case properly.

app/models/audience_export.rb
EmCousin: Just rename the foreign key to `audience_export_id` so you don't have to specify it here. Rely on default conventions rather than excessive configuration. You could also rename `chunks` to `audience_export_chunks`, it doesn't hurt and makes it extra clear what the code talks about.

app/models/audience_export.rb
EmCousin: That's an old convention. We now use `attr_json_data_accessor` from the `JsonData` module, check how it's done in other models.

PR #2547, opened 2025-12-29
fix(tests): Fix three flaky system tests (#2462)

### LINKED ISSUE #2462: Fix flakey tests
We need to fix all three flakey tests in this build - https://github.com/antiwork/gumroad/actions/runs/20488922819

fridayowl:
This PR resolves #2462 by addressing three specific flaky tests:

1. **Audio Timer Check ()**:
   - Replaced exact time string matching ("00:00" -> "00:01") with a regex match  to avoid race conditions where the timer might tick slightly differently than expected.

2. **BG Creator Payments Test ()**:
   - Added a 'safety sandwich' to ensure the form is fully loaded before interacting (checking for "IBAN" field).
   - Added  and  check for the success alert to handle timing issues with form submission.

3. **Shipping Offer Codes ()**:
   - Added  after  in multiple tests to ensure cart and shipping calculations settle before proceeding to checkout.

### DIFF
--- a/spec/requests/products/edit/rich_text_editor_spec.rb
+++ b/spec/requests/products/edit/rich_text_editor_spec.rb
@@ -544,8 +544,8 @@ describe("Product Edit Rich Text Editor", type: :system, js: true) do
 
         # Allow playing the file
         click_on "Play"
-        expect(page).to have_selector("[aria-label='Progress']", text: "00:00")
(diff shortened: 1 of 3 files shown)


binary-koan:
This appears to have a bunch of unrelated changes? Also please check the contributing guidelines and include an AI disclosure

Issue #2030, opened 2025-11-14
Add Versions dropdown menu in Content section on mobile browser


michellelarney:
### What
Add Versions dropdown menu in Content section on mobile browser so you can edit different Versions of a product. 

### Why
Currently you can only edit Versions on desktop. The ability to switch between Versions when editing on mobile does not exist.

advaithh04:
still open?

ershad:
Sounds good, closing this issue. We can work on it when we do mobile-focused design rewamps after completing the existing migrations.

PR #2553, opened 2025-12-30
Enable version picker on content tab for mobile devices

sm17p:
Ref: #2030 

|Before|After (Dark)|After (Light)|
|-------|------------|---------|
|gumroad com_products_esrca_edit_content(iPhone 14 Pro Max)| gumroad dev_products_wf_edit_content(iPhone 14 Pro Max) (1) | gumroad dev_products_wf_edit_content(iPhone 14 Pro Max)|

### Responsive Demo

https://github.com/user-attachments/assets/fa37ffad-6971-453e-93e0-5446c4c8d8aa

### AI Disclosure
No AI used for this PR

### Live Stream Disclosure
- Have watched all five live streams


app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -1046,73 +1046,75 @@ export const ContentTab = () => {
             <Layout
               headerActions={
                 product.variants.length > 0 ? (
-                  <ComboBox<Variant>
-                    className="hidden lg:block"
```
sm17p, line 1050: Removed hidden state on mobile

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -1046,73 +1046,75 @@ export const ContentTab = () => {
...
-                                    return acc && acc > date ? acc : date;
-                                  }, null) ?? new Date(),
-                                )}
-                              </small>
-                            ) : (
-                              <small className="text-muted">No content yet</small>
-                            )}
-                          </div>
+                  <>
+                    <hr className="my-2 scale-x-200 border-border lg:hidden" />
```
sm17p: The separator between sections

sm17p:
> The `scale-x-200` made the page scroll horizontally on mobile, so pushed a change to make it exactly the page width. 

Nice catch! Sorry about that, I was so focused on the top part I forgot to look at the bottom section 😅. Thank you for the help

> (easier than trying to explain in a comment 😅)

No worries, I don't mind pairing at all (Although, you can rely on me with frontend stuff even w/o having to explain too much. I'm quite comfortable with css and can work with the smallest of hints 🫡)

> Note I won't close the issue after merging this as I'm not sure if we want to implement any more of Laura's design - will comment in the issue

Makes sense! It should be open till then

sm17p:
One failing spec seems to be related, looking into it

sm17p:
> One failing spec seems to be related, looking into it

I've checked https://github.com/antiwork/gumroad/actions/runs/20647740581/job/59287725054?pr=2553 with local runs. The spec is flaky but unrelated, can merge 👌

binary-koan:
Thanks for checking!

Yeah I'd usually just leave a quick idea, but I remembered doing this exact hack before and ended up literally writing the code locally while trying to remember the specifics

PR #2422, opened 2025-12-20
chore(inertia): Api page inerita migration

romitg2:
# Before:
### desktop:
https://github.com/user-attachments/assets/ab112fa9-1ead-4b92-9255-63c491e721dd

### mobile:
https://github.com/user-attachments/assets/deaa91fa-7afb-4d38-96cb-c7c71b3f2fb5

# After:
### desktop:
https://github.com/user-attachments/assets/41742890-c4c0-4129-82d8-4b0aeadb116f

### mobile:
https://github.com/user-attachments/assets/80a8fc70-a7b5-4938-b3ab-4402e92ee869

## Tests Result: (local run)
Screenshot 2025-12-21 at 12 32 01 AM

### AI Disclosure:
- used claude-code with sonnet-4.5 model for refactoring 

### Live Stream Disclosure:
- watched all live streams

PR #2558, opened 2025-12-30
Migrate _stack.scss to tailwind

Mit-Patwa:
#### **Part of**: #1055  

---

### **Description**
This PR removes the legacy `stack.scss` file and replaces it with a React component implementation (`Stack.tsx`) styled using Tailwind CSS utility classes.  

Two new components were introduced:
`Stack` and  `StackItem` (applied to each element which is the direct child of `Stack`)

---

### **Changes Made**
- Removed `stack.scss`.
- Created a new **Stack** React component that:
  - Uses Tailwind utility classes to replicate previous SCSS styling.
  - Dynamically applies spacing and layout between child elements.
  - Preserves UI consistency across light and dark themes.

---

### **AI Disclosure**

No AI was used.(Not even tab auto-complete!)

---

### **Live Stream Confirmation**
I have watched 4 of the PR review live streams on youtube

---

#### **Dark Mode**

| Before | After |
|:--:|:--:|
| dark-8-before | dark-8-after |
| dark-2-before | dark-2-after |
| dark-3-before | dark-3-after |
| dark-4-before | dark-4-after |
| dark-5-before | dark-5-after |
| dark-6-before | dark-6-after |
| dark-7-before | dark-7-after |

---

#### **Light Mode**

| Before | After |
|:--:|:--:|
| light-8-before | light-8-after |
| light-2-before | light-2-after |
| light-3-before | light-3-after |
| light-4-before | light-4-after |
| light-5-before | light-5-after |
| light-6-before | light-6-after |
| light-7-before | light-7-after |

---

### DIFF
--- a/app/javascript/pages/UtmLinks/Index.tsx
+++ b/app/javascript/pages/UtmLinks/Index.tsx
@@ -461,7 +461,9 @@ const UtmLinkDetails = ({
       </Card>
       <Card asChild>
         <section>
-          <CardContent>Statistics</CardContent>
+          <CardContent>
+            <h3 className="grow">Statistics</h3>
+          </CardContent>
           <CardContent>
             <h5 className="grow font-bold">Clicks</h5>
             {utmLink.clicks}


Mit-Patwa:
@binary-koan could you take a look. I migrated all the .tsx pages , but I think there are some `.html.erb` pages which still need to be migrated (ex:`api.html.erb`)

Mit-Patwa:
```
<div class="stack" id="api-intro">
  <div>
    <p>
      The Gumroad OAuth API is based around REST. We return JSON for every request, including
      <%= link_to("errors", api_path(anchor: "api-errors")) %>.
    </p>
  </div>
  <div>
    <div class="flex flex-col gap-4">
      <p>
        To start using the API, you'll need to
        <%= link_to("register your OAuth application", settings_advanced_path(anchor: "application-form")) %>.
        Note: The
        <%= link_to("Verify License API endpoint", api_path(anchor: "licenses")) %>
        does not require an OAuth application.
      </p>
      <p>After creating an application, you'll be given a unique application id and application secret.</p>
    </div>
  </div>
</div>
```
This is one of the code snippets present in `html.erb`  file, I think it would be better to manually add the tailwind properties string everywhere since there are not too many of these files and also allows better visibility imo. 
It is probably better than registering the `<Stack>` component(since the component also uses radix slot , which i dont think is compatible with `.erb` files). Let me know if you have something else in mind.

binary-koan:
Are there any other erb files? API is being migrated to React as part of the Inertia migration (https://github.com/antiwork/gumroad/pull/2422 cc @romitg2) so we can definitely just inline classes for now

Mit-Patwa:
Yeah, there are 15 or so .erb files

binary-koan:
Could you list them? Easier for me to check and useful to have as a record

Mit-Patwa:
[
Screenshot 2025-12-31 at 7 21 42 am
](url)

Mit-Patwa:
@binary-koan could you LMK what to do next ?

binary-koan:
Ah that's quite annoying, all those little pages don't really need to be React but it's a lot of styles to duplicate. Maybe the best thing is to get this PR working as-is and keep the SCSS for now, then open another migrating those files to Inertia and removing the SCSS?

We can inline anything we can't easily migrate to Inertia (e.g. Doorkeeper authorizations)

romitg2:
> Are there any other erb files? API is being migrated to React as part of the Inertia migration (#2422 cc @romitg2) so we can definitely just inline classes for now

@binary-koan raised a PR for inertia migration of API page #2595, maybe we can merge it first to avoid inlining? 

> Ah that's quite annoying, all those little pages don't really need to be React but it's a lot of styles to duplicate. Maybe the best thing is to get this PR working as-is and keep the SCSS for now, then open another migrating those files to Inertia and removing the SCSS?

> We can inline anything we can't easily migrate to Inertia (e.g. Doorkeeper authorizations)

otherwise if we're going with this first and keeping SCSS, then i'll update #2595 with new Stack component.

Mit-Patwa:
@binary-koan  could you tell me which files to inline and which files to keep as is, so we can merge!

binary-koan:
Please leave the ERB and SCSS as is for the next PR, let's not inline anything just now. Let me know when the TSX is ready to review

Mit-Patwa:
All the tsx is migrated. Hope you can review.

app/javascript/components/Admin/Products/Purchases/Content.tsx
```diff
@@ -118,16 +119,17 @@ const AdminProductPurchasesContent = ({
...
+          <StackItem asChild key={purchase.external_id}>
+            <AdminProductPurchase
+              purchase={purchase}
+              isSelected={selectedPurchaseExternalIds.includes(purchase.external_id)}
+              onToggleSelection={togglePurchaseSelection}
+            />
+          </StackItem>
         ))}
-      </div>
+      </Stack>
```
Mit-Patwa, line 133: Used `<Stack> and <StackItem>` to replace `<div>'s` and used `<Stack asChild> and <StackItem asChild>` for non-div elements. Made use of radix `Slot` element to pass tailwind styles to children using `asChild`.Followed the same pattern for all files

app/javascript/components/Admin/Products/Purchases/Content.tsx
```diff
@@ -118,16 +119,17 @@ const AdminProductPurchasesContent = ({
...
-            isSelected={selectedPurchaseExternalIds.includes(purchase.external_id)}
-            onToggleSelection={togglePurchaseSelection}
-          />
+          <StackItem asChild key={purchase.external_id}>
+            <AdminProductPurchase
+              purchase={purchase}
+              isSelected={selectedPurchaseExternalIds.includes(purchase.external_id)}
+              onToggleSelection={togglePurchaseSelection}
+            />
+          </StackItem>
```
Mit-Patwa: Passed tailwind utilities to `AdminProductPurchase` component, modified the component to accept the utilities and apply corresponding styles. [see here](https://github.com/antiwork/gumroad/pull/2558/changes#diff-52098da335477fbde847945728daeb54d43ab2da5f1497805b64a35be982a357R48-R49)

app/javascript/components/ui/Card.tsx
Mit-Patwa, line 1: Created `Stack` and `StackItem` components to conditionally render tailwind utilities based on stack type. Made use of Radix `Slot`

app/javascript/components/ui/Stack.tsx
```diff
@@ -0,0 +1,61 @@
...
+  className?: string | undefined;
+  asChild?: boolean;
+  details?: boolean;
+}> &
+  React.HTMLAttributes<HTMLElement>;
+
+export const StackItem = React.forwardRef<HTMLDivElement, StackItemProps>(
+  ({ className, asChild, details, children, ...rest }, ref) => {
+    const baseClasses =
+      "flex flex-wrap items-center p-4 gap-4 justify-between not-first:border-t not-first:border-border";
```
binary-koan: Horizontal flex doesn't seem like a natural layout for this, especially if we call it `CardContent`. Can we remove that from here and just add it inline where it's needed?

Also it might be possible to replace the `not-first:border-...` with `divide-...` on the parent: https://tailwindcss.com/docs/border-style#setting-the-divider-style

app/javascript/components/ui/Stack.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const mainStackClasses = main
+      ? "h-min my-4 mx-auto max-w-md w-[calc(100%-2*1rem)] [&>*]:flex-col [&>*]:items-stretch"
+      : "";
+
+    const borderlessClasses = borderless ? "border-none gap-4 [&>*]:p-0 [&>*]:border-none" : "";
+
+    return (
+      <Component
+        ref={ref}
+        className={classNames(baseClasses, mainStackClasses, borderlessClasses, className)}
```
binary-koan: I would find it more readable to just inline the strings/conditionals here tbh

app/javascript/components/ui/Stack.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+type StackProps = React.PropsWithChildren<{
+  className?: string | undefined;
+  asChild?: boolean;
+  borderless?: boolean;
+  main?: boolean;
+}> &
+  React.HTMLAttributes<HTMLElement>;
+
+export const Stack = React.forwardRef<HTMLDivElement, StackProps>(
```
binary-koan: I think I agree with Sahil [here](https://github.com/antiwork/gumroad/issues/1055#issuecomment-3702867552) that we should name these `Card` and `CardContent` to match shadcn. We can also have a `CardTitle` component for the heading style in the API docs (and everywhere similar)

Eventually we may be able to merge this and `ProductCard`, which would be nice as I think semantically they're the same kind of thing

app/javascript/components/ui/Stack.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return (
+      <Component ref={ref} className={classNames(baseClasses, detailsClasses, className)} {...rest}>
+        {children}
+      </Component>
+    );
+  },
+);
+
+Stack.displayName = "Stack";
+StackItem.displayName = "StackItem";
```
binary-koan: Please put these immediately after the component function (there should be examples of this)

app/javascript/components/ui/Stack.tsx
```diff
@@ -0,0 +1,61 @@
...
+}> &
+  React.HTMLAttributes<HTMLElement>;
+
+export const Stack = React.forwardRef<HTMLDivElement, StackProps>(
+  ({ className, asChild, borderless = false, main, children, ...rest }, ref) => {
+    const Component = asChild ? Slot : "div";
+    const baseClasses = "grid bg-background border border-border rounded stack-component";
+
+    const mainStackClasses = main
+      ? "h-min my-4 mx-auto max-w-md w-[calc(100%-2*1rem)] [&>*]:flex-col [&>*]:items-stretch"
```
binary-koan: * I don't think these belong here, it's quite a specific use case to have in this very generic component. We can inline it for now
* I think we can just replace the `w-[calc(100%-2*1rem)]` with a wrapper element (i.e. `<div className="max-w-md p-4 ..."><Stack ...>`), we're trying to avoid this kind of complex style rule with Tailwind
* The `flex-col` and `items-stretch` shouldn't be needed if we remove the default flex layout (see the other comment)

app/javascript/components/ui/Stack.tsx
Mit-Patwa: @binary-koan  Is there an easy way to know if flex and flex utilities will be applied to an element? I don't think manually checking every usage makes sense and would take very long...would appreciate your input.

Mit-Patwa:
@binary-koan I don't think removing the flex from `StackItem(CardContent)` is needed, since it is needed in all the places where StackItem has multiple children(which is the case most of the times), and horizontal/vertical styling is required inside `StackItem`

Example:
```
<StackItem asChild>
    <section>
      <h5 className="grow font-bold">{currentEmail}</h5>
      {onSave ? (
        <button className="underline" onClick={() => setIsEditing(true)}>
          Edit
        </button>
      ) : (
        <small>
          You cannot change the email of this purchase, because it was made by an existing user. Please ask them
          to go to gumroad.com/settings to update their email.
        </small>
      )}
    </section>
</StackItem>
```

Mit-Patwa:
@binary-koan I've addressed all PR comments, could you check!

binary-koan:
I suspect AI can do a pretty good job of removing/moving the flex classes, will give it a go tomorrow-ish unless you want to try it (I don't think it's worth doing by hand!)

The example you gave is actually where I think we should have a `CardHeader` component. That makes sense as a flex row since we'll usually have a title and action side by side. We don't necessarily need to refactor everything here, but that's why I'd say CardContent (as opposed to a future CardHeader, CardFooter, etc) doesn't need to be a row layout

Mit-Patwa:
@binary-koan Sure. You could  give it a go, let me know if you need my help!

app/javascript/components/Admin/Products/Purchases/Purchase.tsx
```diff
@@ -43,8 +45,8 @@ const AdminProductPurchase = ({
   const isSelectable = stripe_refunded !== true;
 
   return (
-    <div>
-      <div className="flex items-start gap-2">
+    <div className={className}>
+      <div className={`flex items-start gap-2 ${className ? "grow" : ""}`}>
```
binary-koan: Why do we need these `className ?` checks? Can we not just always apply the new classes?

This component might be used with quite different classes somewhere else so this feels like an antipattern

app/javascript/stylesheets/_dropdown.scss
```diff
@@ -22,7 +22,7 @@
   }
 
   > [role="menu"],
-  > .stack {
+  > .card-component {
```
binary-koan: It looks like we only have cards in popovers in two places (`app/javascript/components/Analytics/ProductsPopover.tsx` and `app/javascript/components/Profile/EditSections.tsx`), so I think we should just inline the styles there and avoid adding work to the (even more complex) dropdown migration. Either we can inline these `only-child` styles, or we can stop using `Card` and add a subset of the card styles inline, whichever is cleaner

app/javascript/components/Admin/Products/Purchases/Purchase.tsx
Mit-Patwa: If this component(`AdminProductPurchase`) is used outside `Card` and `CardContent`, `grow` utility shouldn't be applied, but yes you're right it would also be added if any other styles are applied to this component, So I think it's better to always apply the classes..

app/javascript/components/Profile/EditSections.tsx
```diff
@@ -177,15 +178,20 @@ export const EditorMenu = ({
           {activeSubmenu}
         </div>
       ) : (
-        <div className="stack" style={{ width: "300px" }}>
+        <div
+          style={{ width: "300px" }}
+          className="-m-4 grid max-w-[calc(100%+2rem)] !divide-y !divide-solid !divide-border rounded border border-none border-border bg-background shadow-none"
+        >
```
Mit-Patwa: Added inline styles here because `divide-y`, `divide-solid` and `divide-border` were being overriden by other styles.

app/javascript/components/Analytics/ProductsPopover.tsx
```diff
@@ -22,9 +23,9 @@ export const ProductsPopover = ({
       </span>
     }
   >
-    <div className="stack">
-      <div>
-        <fieldset>
+    <Card className="-m-4 max-w-[calc(100%+2rem)] border-none shadow-none">
```
Mit-Patwa: Added the 'dropdown' related styles inline here and in `app/javascript/components/Profile/EditSections.tsx`

Mit-Patwa:
@binary-koan I've addressed the comments. Could you please check!

Mit-Patwa:
@binary-koan  I also removed the stack styles that were added in the api page inertia migration(bb5e584a7fc45f82296864b11efc03a5e0d8d549)

app/javascript/components/Analytics/ProductsPopover.tsx
```diff
@@ -23,7 +23,7 @@ export const ProductsPopover = ({
       </span>
     }
   >
-    <Card>
+    <Card className="-m-4 max-w-[calc(100%+2rem)] border-none shadow-none">
```
binary-koan: The `-m-4 max-w-[calc(100%+2rem)]` is pretty clunky, but easier to tidy it up in the dropdown migration

binary-koan:
Sorry for the wait, could you check if the failing tests are related? Apart from that I think this is good to go

Mit-Patwa:
Hey @binary-koan the failing tests seem to be unrelated!

Mit-Patwa:
Sorry , I didn't realise the failing test was related. I'm not able to run all the test specs for some reason on my machine.[see](https://github.com/antiwork/gumroad/discussions/2302#discussion-9239578)
By any chance have you come across this issue before/know how to fix it? I've followed all the steps asked in the `testing.md` file.
Would like to fix this for future test runs.
Thanks!

binary-koan:
No worries! Looks good now, merging

sm17p:
(Just a quick heads up before this gets deployed on production)

@Mit-Patwa @binary-koan I think this PR might have a few differences in styling being applied from original implementation, the checkout form seems to have an odd border above "Card Information" and the PayPal form submit button also has a different width. 

I'm sharing the Before/After shots below for comparision, please have a look 

### Before

https://github.com/user-attachments/assets/21e8d803-4551-405e-8b71-f82400aafd88

### After

https://github.com/user-attachments/assets/199dc880-6950-4591-bf5f-1e080aadf950

PR #2537, opened 2025-12-29
Inertia Migration: Collaborator Pages

sm17p:
# Part of #856
- #856

Migrates the following pages to inertia 
1. `/collaborators`(completing initial migration with partial visit on remove action)
2. `/collaborators/new`
3. `/collaborators/:id/edit`
4. `/collaborators/incomings`

# Code changes 
- All of the pages re-factored and moved to `pages/Collaborators` folder
- For `/collaborators`, and `/collaborators/incomings` page, 
    - Using inertia migrated endpoints to perform user actions
    - Inlined page component into `pages/Collaborators/Index.tsx` and `pages/Collaborators/Incoming/Index.tsx` respectively to match existing pattern
    - Invitation `accept` and `decline` actions on "Collaborations" page are routed to `Collaborators::IncomingsController` which render `Collaborators/Incoming/Index.tsx`
    - Using inertia partial visit to remove a `collaborator` and syncing table state for both the pages in a single `DELETE` request 
        -  The remove action is routed to `/collaborators/:id` on "Collaborations" page and `/collaborators/incomings/:id` on "Collaborations" page to avoid a displaying the full page loader
    - [NEW] Extracted common utils functions into `utils/collaboratorFormatters.ts`
    - [NEW] Extracted the sheet used into a common component (`components/Collaborators/CollaboratorDetailsSheet.tsx`)
    
- For `collaborators/new` and `/collaborators/:id/edit` page
    -  Using inertia's `useForm` helper to manage form state. This simplifies form logic reducing the need of managing form field data and their corresponding error state.
    -  Using page props to fetch form and it's metadata simplifying local component states
    -  Redirecting from backend to the next page on success (`/collaborators`) 
    
 <details>
<summary><h2>Details of Migrated API Endpoints to Inertia</h2></summary>

### Controllers
- [Moved] `Api::Internal::CollaboratorsController` → `Collaborators::MainController`
- [New] `Collaborators::BaseController`
- [New] `Collaborators::IncomingsController`
- [Deleted] `Api::Internal::Collaborators::InvitationAcceptancesController`
- [Deleted] `Api::Internal::Collaborators::InvitationDeclinesController`
- [Deleted] `Api::Internal::Collaborators::IncomingsController`

### Collaborators (`collaborators.ts`)
- `GET /internal/collaborators` → `GET /collaborators` (`Collaborators::MainController#index`)
- `GET /internal/collaborators/new` → `GET /collaborators/new` (`Collaborators::MainController#new`)
[...]

PR #2565, opened 2025-12-30
Inertia migration: Collaborators Pages

sm17p:
# Part of #856
- #856
- Previos PR: #2537

Migrates the following pages to inertia 
1. `/collaborators`(completing initial migration with partial visit on remove action)
2. `/collaborators/new`
3. `/collaborators/:id/edit`
4. `/collaborators/incomings`

# Code changes 
- All of the pages re-factored and moved to `pages/Collaborators` folder
- For `/collaborators`, and `/collaborators/incomings` page, 
    - Using inertia migrated endpoints to perform user actions
    - Inlined page component into `pages/Collaborators/Index.tsx` and `pages/Collaborators/Incoming/Index.tsx` respectively to match existing pattern
    - Invitation `accept` and `decline` actions on "Collaborations" page are routed to `Collaborators::IncomingsController` which ~~render `Collaborators/Incoming/Index.tsx`~~ re-directs back to "Collaborations" page
    - Using inertia partial visit to remove a `collaborator` and syncing table state for both the pages ~~in a single `DELETE` request~~ via a re-direct request to the respective index pages
        -  The remove action is routed to `/collaborators/:id` on "Collaborations" page and `/collaborators/incomings/:id` on "Collaborations" page to avoid a displaying the full page loader
    - [NEW] Extracted common utils functions into `utils/collaboratorFormatters.ts`
    - [NEW] Extracted the sheet used into a common component (`components/Collaborators/CollaboratorDetailsSheet.tsx`)
    
- For `collaborators/new` and `/collaborators/:id/edit` page
    -  Using inertia's `useForm` helper to manage form state. This simplifies form logic reducing the need of managing form field data and their corresponding error state.
    -  Using page props to fetch form and it's metadata simplifying local component states
    -  Re-directing from backend to the next page on success (`/collaborators`) 
    
 <details>
<summary><h2>Details of Migrated API Endpoints to Inertia</h2></summary>

### Controllers
- [Moved] `Api::Internal::CollaboratorsController` → `Collaborators::MainController`
- [New] `Collaborators::BaseController`
- [New] `Collaborators::IncomingsController`
- [Deleted] `Api::Internal::Collaborators::InvitationAcceptancesController`
- [Deleted] `Api::Internal::Collaborators::InvitationDeclinesController`
- [Deleted] `Api::Internal::Collaborators::IncomingsController`

### Collaborators (`collaborators.ts`)
- `GET /internal/collaborators` → `GET /collaborators` (`Collaborators::MainController#index`)
- `GET /internal/collaborators/new` → `GET /collaborators/new` (`Collaborators::MainController#new`)
- `GET /internal/collaborators/:id/edit` → `GET /collaborators/:id/edit` (`Collaborators::MainController#edit`)
- `POST /internal/collaborators` → `POST /collaborators` (`Collaborators::MainController#create`)
- `PATCH /internal/collaborators/:id` → `PATCH /collaborators/:id` (`Collaborators::MainController#update`)
- `DELETE /internal/collaborators/:id` → `DELETE /collaborators/:id` (`Collaborators::MainController#destroy`)
- `POST /internal/collaborators/:id/invitation_acceptances` → `POST /collaborators/incomings/:id/accept` (`Collaborators::IncomingsController#accept`)
- `POST /internal/collaborators/:id/invitation_declines` → `POST /collaborators/incomings/:id/decline` (`Collaborators::IncomingsController#decline`)

### Incoming Collaborators (`incoming_collaborators.ts`)

- `GET /internal/collaborators/incomings` → `GET /collaborators/incomings` (`Collaborators::IncomingsController#index`)

</details>

### General Navigation
|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/51dd9915-268b-42f6-a39f-f27ba495150a" />| <video src="https://github.com/user-attachments/assets/96c43e41-4235-40b1-b826-c1517707004d" />|

### Adding a New Collaborator
|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/24bb6767-69a3-4c1f-bda0-3c12ad6bb8de" />| <video src="https://github.com/user-attachments/assets/d9ae0999-d06e-41d7-bd1d-b7795f22f14c" />|

### Editing Collaboration
|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/eb1dfd0b-e8b4-4f5e-b68c-aee69e391192" />| <video src="https://github.com/user-attachments/assets/c85b2648-4e59-4fd0-9c11-e0382d727d8f" />|

### Collaboration Delete
|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/1a0e590f-7938-49e9-b470-27d5658f0f25" />| <video src="https://github.com/user-attachments/assets/7e066550-b48d-493c-8bf1-b7dc918f7458" />|

### Decline Incoming Collaboration Request
|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/f41757b3-6728-4232-9610-086810a27d98" />| <video src="https://github.com/user-attachments/assets/3e6a82c1-2293-456e-9bd8-474cdd48bbf2" />|

### Accept Incoming Collaboration Request

|Before|After|
|-----|-------|
|<video src="https://github.com/user-attachments/assets/e01d3f0c-fa78-4df9-9206-6205431c58c5" />| <video src="https://github.com/user-attachments/assets/192834cc-c52d-4cca-a1ef-4130a044ddb7" />|

### Specs Passing (All related to collaborator)

specs-passing

### AI Disclosure
Cursor Auto Mode for 
1. Code edits
2. Moving spec files
3. Checking Inertia, Rspec and Rails documentation 
    
 ### Live Stream Disclosure
 Have watched all the live streams


spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -244,32 +244,17 @@
...
-          delete :destroy, params: { id: accepted_collaboration.external_id }
-          expect(response).to have_http_status(:ok)
-          expect(inertia.component).to eq("Collaborators/Incoming/Index")
-          expect(flash[:notice]).to eq("Collaborator removed")
-        end.to have_enqueued_mail(AffiliateMailer, :collaboration_ended_by_seller).with(accepted_collaboration.id)
-
-        expect(accepted_collaboration.reload.deleted_at).to be_present
-      end
-    end
-
```
sm17p: Sorry, I forgot to clean this up in my last self-review 😅

It's a little bit odd though now, since both routes can delete the collaboration record for a seller and affiliate user. My reason for structuring `detroy` in `base_controller` was to avoid code duplication but LMK if this is un-desired?

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -6,7 +6,7 @@
 require "shared_examples/authentication_required"
 require "inertia_rails/rspec"
 
-describe Collaborators::MainController, inertia: true do
+describe Collaborators::MainController, type: :controller, inertia: true do
```
sm17p, line 10: Missed this in my review, added back now

sm17p (commented):
@EmCousin _a Thank you for the quick review. I'v applied those critical changes from the last review and visually there's no change in UX. 

I'm sharing a missing demo case for confirmation

### Remove Action From Collaborations Page Re-directs Back To The Same Page

https://github.com/user-attachments/assets/3e6ad5b0-e2d2-467b-8376-d5060e5c7e89

<hr />

I ran the controller, and request specs again and all of them passed. Please have a look again 🙏

Image

app/controllers/collaborators/base_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    @collaborator.mark_deleted!
+
+    if current_seller == @collaborator.seller
+      AffiliateMailer.collaboration_ended_by_seller(@collaborator.id).deliver_later
+    elsif current_seller == @collaborator.affiliate_user
+      AffiliateMailer.collaboration_ended_by_affiliate_user(@collaborator.id).deliver_later
+    end
+
+    yield if block_given?
```
EmCousin: You are already checking this on line 11
```suggestion
    yield
```

app/controllers/collaborators/main_controller.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+class Collaborators::MainController < Collaborators::BaseController
+  prepend_before_action :set_collaborator, only: %i[edit update destroy]
+
+  def index
+    render_index_props
```
EmCousin: This method is invoked only once. You can just inline its content here now.

app/presenters/collaborator_presenter.rb
```diff
@@ -6,50 +6,100 @@ def initialize(seller:, collaborator: nil)
...
+          dont_show_as_co_creator: for_edit_page ? collaborator.dont_show_as_co_creator? : false,
+          products: all_products(for_edit_page:, product_affiliates:, default_commission:, apply_to_all:),
+        }.compact,
+        page_metadata: {
+          default_percent_commission: DEFAULT_PERCENT_COMMISSION,
+          min_percent_commission: Collaborator::MIN_PERCENT_COMMISSION,
+          max_percent_commission: Collaborator::MAX_PERCENT_COMMISSION,
+          max_products_with_affiliates_to_show: MAX_PRODUCTS_WITH_AFFILIATES_TO_SHOW,
+          title: for_edit_page ? collaborator.affiliate_user.display_name(prefer_email_over_default_username: true) : "New collaborator",
+        }.compact,
```
EmCousin: The intention is good here, but this refactor comes at the cost of too much complexity.
You can refactor it more cleanly by making `collaborator_form_props` accept keyword arguments with default values for the `new` action, which you can then customize for the `edit` action. You'll end up with something more modular and maintainable.

```ruby
def collaborator_props(product_affiliates: [], default_commission: DEFAULT_PERCENT_COMMISSION, apply_to_all: true, ...)
```

Same logic for the `all_products` method

app/presenters/incoming_collaborators_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+        .left_joins(:collaborator_invitation)
+        .includes(
+          :collaborator_invitation,
+          :seller,
+          product_affiliates: :product
+        )
+        .order(
+          Arel.sql("CASE WHEN collaborator_invitations.id IS NULL THEN 0 ELSE 1 END"),
+          :id
+        )
```
EmCousin, line 30: Took me a while to understand that you moved this sorting logic from the frontend. Please do a self-review.

spec/controllers/collaborators/base_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  describe "inertia_share" do
+    it "shares collaborator presenter props" do
+      get :index
+
+      expect(response).to be_successful
+      expect(inertia.props[:collaborators_disabled_reason]).to eq(CollaboratorPresenter.new(seller:).inertia_shared_props[:collaborators_disabled_reason])
+    end
+  end
+end
```
EmCousin: Creating `shared_example` and using it in each related controller would be cleaner.

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+
+      expect(response).to be_successful
+      expect(inertia.component).to eq("Collaborators/Incoming/Index")
+
+      # Check that we get the expected collaborators (accepted and pending)
+      collaborators = inertia.props[:collaborators]
+      expect(collaborators.length).to eq(3)
+
+      collaborator_ids = collaborators.map { |c| c[:id] }
+      expect(collaborator_ids).to match_array([accepted_collaboration.external_id, pending_collaboration.external_id, collaborator.external_id])
```
EmCousin:
```suggestion
      expect(collaborator_ids).to eq([accepted_collaboration.external_id, pending_collaboration.external_id, collaborator.external_id])
```

app/presenters/incoming_collaborators_presenter.rb
sm17p: Sorry about that, I didn't leave new review comments as I thought those were covered by you in https://github.com/antiwork/gumroad/pull/2537/changes#r2651825935. Should I add them back over here?

EmCousin: Ah, makes sense. Would be better to port them over the latest PR if possible

app/controllers/api/internal/collaborators/invitation_acceptances_controller.rb
```diff
@@ -1,27 +0,0 @@
...
-
-  after_action :verify_authorized
-
-  def create
-    authorize @invitation, :accept?
-
-    @invitation.accept!
-
-    head :ok
-  end
```
sm17p, line 17: Moved to `Collaborators::IncomingsController#accept` method

app/controllers/api/internal/collaborators/invitation_declines_controller.rb
```diff
@@ -1,27 +0,0 @@
...
-
-  after_action :verify_authorized
-
-  def create
-    authorize @invitation, :decline?
-
-    @invitation.decline!
-
-    head :ok
-  end
```
sm17p, line 17: Moved to `Collaborators::IncomingsController#decline` method

app/controllers/collaborators/base_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class Collaborators::BaseController < Sellers::BaseController
+  before_action :authorize_user
+
+  layout "inertia"
+
+  inertia_share { CollaboratorPresenter.new(seller: current_seller).inertia_shared_props }
```
sm17p, line 8: This prop key is shared across all migrated pages and hence unifying it over here

app/javascript/components/Collaborators/CollaboratorDetailsSheet.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { Alert } from "$app/components/ui/Alert";
+import { Sheet, SheetHeader } from "$app/components/ui/Sheet";
+
+type CollaboratorDetailsSheetProps = {
+  collaborator: Collaborator | IncomingCollaborator;
+  onClose: () => void;
+  actions: React.ReactNode;
+  showSetupWarning?: boolean;
+};
```
sm17p, line 15: Both `/collaborator` and `/collaborator/incomings` render the same sheet so extracting it as a component over here for maintainaibility

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,51 +16,45 @@ import { Pill } from "$app/components/ui/Pill";
...
-  const [defaultPercentCommission, setDefaultPercentCommission] = React.useState<{
-    value: number | null;
-    hasError: boolean;
-  }>({
-    value: isEditing ? formData.percent_commission || DEFAULT_PERCENT_COMMISSION : DEFAULT_PERCENT_COMMISSION,
-    hasError: false,
-  });
-  const [dontShowAsCoCreator, setDontShowAsCoCreator] = React.useState(
-    isEditing ? formData.dont_show_as_co_creator : false,
-  );
```
sm17p, line 70: These are now path of the form data and metadata sent from backend

spec/requests/collaborators_spec.rb
```diff
@@ -422,8 +422,8 @@
       end
       wait_for_ajax
       expect(page).to have_alert(text: "The collaborator was removed successfully.")
-      expect(collaborators.second.reload.deleted_at).to be_present
       expect(page).to_not have_table_row({ "Name" => collaborators.second.affiliate_user.username })
+      expect(collaborators.second.reload.deleted_at).to be_present
```
sm17p, line 432: This was breaking after inertia migration, so I'm just waiting for UI to update before checking on DB state

spec/requests/collaborators_spec.rb
sm17p, line 1: [NEW] Mostly cleanup in this file to use route helpers. 

There's one change in the order of assertion on `L426` after inertia migration, and added a few assertions to check on document `title` and page `header` for completeness

spec/shared_examples/collaborators_base_controller_concern.rb
```diff
@@ -0,0 +1,25 @@
...
+  it "on #{verb}: #{action}" do
+    format = defined?(request_format) ? request_format : :html
+    expect_any_instance_of(CollaboratorPresenter).to receive(:inertia_shared_props).and_call_original
+
+    public_send(verb, action, params: defined?(request_params) ? request_params : {}, as: format)
+
+    expect(response).to be_successful
+    expect(inertia.props).to have_key(:collaborators_disabled_reason)
+  end
+end
```
sm17p: Created a shared example for cleaner specs as per last review

spec/shared_examples/collaborators_base_controller_concern.rb
```diff
@@ -0,0 +1,26 @@
...
```
sm17p, line 26: Created shared examples for cleaner specs as per discussion in https://github.com/antiwork/gumroad/pull/2565#discussion_r2657896589

spec/presenters/collaborators_presenter_spec.rb
```diff
@@ -21,21 +21,10 @@
...
-    it "returns collaborators supported as false if using a Brazilian Stripe Connect account" do
-      brazilian_stripe_account = create(:merchant_account_stripe_connect, user: seller, country: "BR")
-      seller.update!(check_merchant_account_is_linked: true)
-      expect(seller.merchant_account(StripeChargeProcessor.charge_processor_id)).to eq brazilian_stripe_account
-
-      props = described_class.new(seller:).index_props
-
-      expect(props[:collaborators_disabled_reason]).to eq "Collaborators with Brazilian Stripe accounts are not supported."
-    end
-
```
sm17p, line 38: Tested in `spec/presenters/collaborator_presenter_spec.rb` now as a shared prop and binding further verified in `spec/shared_examples/collaborators_base_controller_concern.rb:L19`

spec/presenters/collaborator_presenter_spec.rb
```diff
@@ -64,23 +75,76 @@
...
                                               ])
-      expect(props[:collaborators_disabled_reason]).to eq(nil)
     end
+  end
 
-    context "when the seller has a Brazilian Stripe Connect account" do
-      before do
-        allow(seller).to receive(:has_brazilian_stripe_connect_account?).and_return(true)
-      end
+  describe "#inertia_shared_props" do
```
sm17p, line 147: Spec for new method

app/javascript/utils/collaboratorFormatters.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+  return product.affiliate_percentage;
+};
+
+const getDefaultCommission = (collaborator: Collaborator | IncomingCollaborator): number => {
+  if ("percent_commission" in collaborator) {
+    return collaborator.percent_commission ?? 0;
+  }
+  return collaborator.affiliate_percentage ?? 0;
+};
```
sm17p: These values can be `NULL` in `db` but I've enabled presenters to send `DEFAULT_PERCENT_COMMISSION`. 

Kept the `null` type for the fields in frontend as it was breaking few of the specs and things were getting out of hand

app/controllers/collaborators/base_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      if @collaborator.present?
+        authorize @collaborator
+      else
+        authorize Collaborator
+      end
+    end
+
+    def set_title
+      @title = "Collaborators"
+    end
```
sm17p, line 39: Currently, this is the title used for all of the collaborator pages. I've harden specs for `controllers` and `requests` to check for the same as well

app/controllers/collaborators/incomings_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class Collaborators::IncomingsController < Collaborators::BaseController
+  skip_before_action :authorize_user, only: [:accept, :decline]
```
sm17p, line 4: Authorization calls are inside methods for these

app/javascript/data/collaborators.ts
```diff
@@ -45,116 +22,35 @@ export type CollaboratorFormProduct = {
...
-    method: "POST",
-    accept: "json",
-    url: Routes.internal_collaborator_invitation_declines_path(collaboratorId),
-  });
-  if (!response.ok) throw new ResponseError();
-}
+export type EditPageProps = {
+  form_data: CollaboratorFormData & { id: string };
+  page_metadata: CollaboratorFormPageMetaData;
+} & CollaboratorPagesSharedProps;
```
sm17p, line 55: These types are used in`Collaborator/Form.tsx` as well hence declared over here to avoid circular dependencies

app/javascript/data/incoming_collaborators.ts
```diff
@@ -1,19 +1,10 @@
...
-};
-
 export type IncomingCollaborator = {
   id: string;
   seller_email: string;
   seller_name: string;
   seller_avatar_url: string;
   apply_to_all_products: boolean;
-  affiliate_percentage: number;
+  affiliate_percentage: number | null;
```
sm17p: I got confused over here and found out that `affiliate_percentage` is the same field as  `percent_commission: number | null;` from 

```ts
export type Collaborator = {
  id: string;
  email: string;
  name: string | null;
  avatar_url: string;
  percent_commission: number | null;
  setup_incomplete: boolean;
  products: {
    id: string;
    name: string;
    percent_commission: number;
  }[];
  invitation_accepted: boolean;
};
``` 
and hence matching it's signature over here to resolve the ambiguity around it

app/javascript/pages/Collaborators/Incomings/Index.tsx
```diff
@@ -0,0 +1,241 @@
...
+      showTabs
+      headerActions={
+        <WithTooltip position="bottom" tip={collaborators_disabled_reason}>
+          <NavigationButtonInertia
+            href={Routes.new_collaborator_path()}
+            color="accent"
+            inert={!loggedInUser?.policies.collaborator.create || collaborators_disabled_reason !== null}
+          >
+            Add collaborator
+          </NavigationButtonInertia>
```
sm17p, line 197: ### Before
```tsx
<Link
  to="/collaborators/new"
  className="button accent"
  inert={
    !loggedInUser?.policies.collaborator.create ||
    navigation.state !== "idle" ||
    collaborators_disabled_reason !== null
  }
>
  Add collaborator
</Link>
```

### After

1. Using the generated route helper
2. Replaced with `NavigationButtonInertia` to avoid usage of "button" class

app/javascript/pages/Collaborators/Incomings/Index.tsx
```diff
@@ -0,0 +1,241 @@
...
+                onClick={() => onReject(selected)}
+                disabled={disabled}
+              >
+                Decline
+              </Button>
+            </>
+          )
+        }
+      />
+    ) : null}
```
sm17p, line 163: New sheet component used over here

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
-
-  const hasEnabledUnpublishedOrIneligibleProducts =
-    isEditing &&
-    formData.products.some((product) => product.enabled && (!product.published || product.has_another_collaborator));
-
-  const [showIneligibleProducts, setShowIneligibleProducts] = React.useState(hasEnabledUnpublishedOrIneligibleProducts);
-  const [collaboratorEmail, setCollaboratorEmail] = React.useState<{ value: string; error?: string }>({
-    value: "",
-  });
+  const isEditPage = "id" in formData;
```
sm17p, line 36: Changed the variable name to `isEditPage`. `isEditing` could also imply if we're editing the form and this felt less ambiguous, lmk if this needs to be reverted

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
 
-  const handleDefaultCommissionChange = (percent_commission: number | null) => {
-    setDefaultPercentCommission({ value: percent_commission, hasError: false });
-    setProducts((prevProducts) => prevProducts.map((item) => ({ ...item, percent_commission, has_error: false })));
+  const handleProductChange = (id: string, attrs: Partial<CollaboratorFormProduct>) => {
+    form.setData(
+      "products",
+      form.data.products.map((item) => (item.id === id ? { ...item, ...attrs, has_error: false } : item)),
+    );
   };
```
sm17p, line 115: `handleDefaultCommissionChange` is inlined down below now

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
-        collaboratorEmail.value.length === 0
+        form.data.email?.length === 0
           ? "Collaborator email must be provided"
-          : !isValidEmail(collaboratorEmail.value)
+          : !isValidEmail(form.data.email ?? "")
             ? "Please enter a valid email"
             : null;
-      setCollaboratorEmail(
-        emailError ? { value: collaboratorEmail.value, error: emailError } : { value: collaboratorEmail.value },
-      );
```
sm17p, line 138: In the migrated version `email` is already set in the form, so we just have to update error state now

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
+            router.replaceProp("errors", undefined);
+          }
+        },
+      });
     }
-  });
-  React.useEffect(() => {
-    if (!isConfirmed) return;
-    handleSubmit();
-  }, [isConfirmed]);
```
sm17p, line 196: Re-worked to not rely on an effect for form submission

app/javascript/pages/Collaborators/Incomings/Index.tsx
```diff
@@ -0,0 +1,235 @@
...
+          onRemove={(incomingCollaborator) =>
+            form.delete(Routes.collaborators_incoming_path(incomingCollaborator.id), {
+              only: ["collaborators", "flash"],
+              preserveUrl: true,
+              onSuccess: () => setSelected(null),
+              onError: () => showAlert("Sorry, something went wrong. Please try again.", "error"),
+            })
+          }
+        />
+      )}
```
sm17p, line 227: Using `only` for these form actions to avoid displaying a full page loader

app/javascript/pages/Collaborators/Index.tsx
```diff
@@ -1,3 +1,190 @@
...
+  collaborators: Collaborator[];
+  has_incoming_collaborators: boolean;
+} & CollaboratorPagesSharedProps;
+
+const CollaboratorsPage = () => {
+  const loggedInUser = useLoggedInUser();
+
+  const { collaborators, collaborators_disabled_reason, has_incoming_collaborators } = cast<CollaboratorsPageProps>(
+    usePage().props,
+  );
```
sm17p, line 31: Data fetched using inertia helper now

app/javascript/pages/Collaborators/Index.tsx
```diff
@@ -1,3 +1,190 @@
...
+      only: ["collaborators", "flash", "has_incoming_collaborators"],
+      preserveUrl: true,
+      onSuccess: () => {
+        setSelectedCollaborator(null);
+      },
+      onError: () => {
+        showAlert("Failed to remove the collaborator.", "error");
+      },
+    });
+  };
```
sm17p, line 46: Same as above, using `only` to avoid displaying a full page loader

app/presenters/collaborator_presenter.rb
```diff
@@ -6,50 +6,103 @@ def initialize(seller:, collaborator: nil)
...
+      }
+    end
+
     def products
       collaborator&.product_affiliates&.includes(:product)&.map do |pa|
         {
           id: pa.product.external_id,
           name: pa.product.name,
-          percent_commission: pa.affiliate_percentage,
+          percent_commission: pa.affiliate_percentage || collaborator.affiliate_percentage,
```
sm17p, line 80: It should be one of these two, this simplifies `null` handling on frontend a bit

app/presenters/collaborator_presenter.rb
```diff
@@ -6,50 +6,103 @@ def initialize(seller:, collaborator: nil)
...
           id: product.external_id,
           name: product.name,
           has_another_collaborator: product.has_another_collaborator?(collaborator:),
           has_affiliates: product.direct_affiliates.alive.exists?,
           published: product.published?,
-          enabled: product_affiliate.present?,
-          percent_commission: product_affiliate&.affiliate_percentage,
-          dont_show_as_co_creator: product_affiliate&.dont_show_as_co_creator || false,
+          enabled: product_affiliate.present? || (!product.has_another_collaborator?(collaborator:) && product.published?),
+          percent_commission: product_affiliate&.affiliate_percentage || percent_commission,
```
sm17p, line 97: Same thing as above, helps avoid `null` handling on frontend. 

`percent_commission` can be either of `collaborator.affiliate_percentage` or `DEFAULT_PERCENT_COMMISSION` at this point

spec/controllers/api/internal/collaborators/invitation_acceptances_controller_spec.rb
```diff
@@ -1,63 +0,0 @@
...
-require "spec_helper"
-require "shared_examples/authentication_required"
-
-describe Api::Internal::Collaborators::InvitationAcceptancesController do
-  let!(:seller) { create(:user) }
-  let!(:invited_user) { create(:user) }
-  let!(:collaborator) { create(:collaborator, seller: seller, affiliate_user: invited_user) }
-  let!(:invitation) { create(:collaborator_invitation, collaborator: collaborator) }
-
-  describe "POST create" do
```
sm17p: Moved into the incomings collaborator controller spec file

spec/controllers/api/internal/collaborators/invitation_declines_controller_spec.rb
```diff
@@ -1,55 +0,0 @@
...
-require "spec_helper"
-require "shared_examples/authentication_required"
-
-describe Api::Internal::Collaborators::InvitationDeclinesController do
-  let!(:seller) { create(:user) }
-  let!(:invited_user) { create(:user) }
-  let!(:collaborator) { create(:collaborator, seller: seller, affiliate_user: invited_user) }
-  let!(:invitation) { create(:collaborator_invitation, collaborator: collaborator) }
-
-  describe "POST create" do
```
sm17p: Moved into the incomings collaborator controller spec file

spec/controllers/api/internal/collaborators_controller_spec.rb
```diff
@@ -1,230 +0,0 @@
-# frozen_string_literal: true
-
-require "spec_helper"
-require "shared_examples/authorize_called"
-require "shared_examples/authentication_required"
-
-describe Api::Internal::CollaboratorsController do
```
sm17p: Moved into the collaborators main controller spec file

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "shared_examples/sellers_base_controller_concern"
+require "shared_examples/collaborators_base_controller_concern"
```
sm17p, line 5: New shared examples

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+        seller: seller1,
+        affiliate_user: seller2
+      )
+    end
+
+    before { sign_in invited_user }
+
+    it_behaves_like "authentication required for action", :get, :index
+
+    it_behaves_like "collaborator disabled reason sent", :get, :index
```
sm17p, line 61: From shared example

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "shared_examples/sellers_base_controller_concern"
+require "shared_examples/collaborators_base_controller_concern"
+require "shared_examples/authentication_required"
+require "inertia_rails/rspec"
+
+describe Collaborators::IncomingsController, inertia: true do
+  it_behaves_like "inherits from Collaborators::BaseController"
```
sm17p, line 10: From shared example

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        expect(response).to be_redirect
+        expect(response).to redirect_to(collaborators_incomings_path)
+        expect(flash[:notice]).to eq("Invitation accepted")
+        expect(collaborator.reload.invitation_accepted?).to eq(true)
+      end
+
+      context "when records are faulty" do
+        include_examples "invalid collaborator records", :post, :accept
+      end
```
sm17p, line 134: From shared example

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        expect(response).to be_redirect
+        expect(response).to redirect_to(collaborators_incomings_path)
+        expect(flash[:notice]).to eq("Invitation declined")
+        expect(collaborator.reload.deleted?).to eq(true)
+      end
+
+      context "when records are faulty" do
+        include_examples "invalid collaborator records", :post, :decline
+      end
```
sm17p: From shared example

spec/controllers/collaborators/incomings_controller_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+        expect(response).to redirect_to(collaborators_incomings_path)
+        expect(flash[:notice]).to eq("Collaborator removed")
+      end.to have_enqueued_mail(AffiliateMailer, :collaboration_ended_by_affiliate_user).with(accepted_collaboration.id)
+
+      expect(accepted_collaboration.reload.deleted_at).to be_present
+    end
+
+    context "when records are faulty" do
+      include_examples "invalid collaborator records", :delete, :destroy
+    end
```
sm17p: From shared example

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+require "spec_helper"
+require "shared_examples/sellers_base_controller_concern"
+require "shared_examples/collaborators_base_controller_concern"
+require "shared_examples/authorize_called"
+require "shared_examples/authentication_required"
+require "inertia_rails/rspec"
+
+describe Collaborators::MainController, type: :controller, inertia: true do
+  it_behaves_like "inherits from Collaborators::BaseController"
```
sm17p, line 11: From shared example

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+  end
+
+  describe "GET index" do
+    it_behaves_like "authentication required for action", :get, :index
+
+    it_behaves_like "authorize called for action", :get, :index do
+      let(:record) { Collaborator }
+    end
+
+    it_behaves_like "collaborator disabled reason sent", :get, :index
```
sm17p, line 27: From shared example

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+    it "renders the index template with correct props" do
+      collaborator1 = create(:collaborator, seller:, products: [create(:product, user: seller)])
+      collaborator2 = create(:collaborator, seller:)
+
+      get :index
+
+      expect(response).to be_successful
+      expect(inertia.component).to eq("Collaborators/Index")
+      expect(inertia.props[:title]).to eq("Collaborators")
```
sm17p, line 37: [NEW] explicitly checking title over here for robustness

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+      presenter = CollaboratorPresenter.new(seller:, collaborator:)
+      expected_props = presenter.edit_collaborator_props
+      actual_props = inertia.props.slice(*expected_props.keys)
+      expect(actual_props).to eq(expected_props)
+    end
+
+    context "when records are faulty" do
+      include_examples "invalid collaborator records", :get, :edit
+    end
```
sm17p, line 81: From shared example

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+        expect(response).to redirect_to(collaborators_path)
+        expect(flash[:notice]).to eq("The collaborator was removed successfully.")
+      end.to have_enqueued_mail(AffiliateMailer, :collaboration_ended_by_seller).with(collaborator.id)
+
+      expect(collaborator.reload.deleted_at).to be_present
+    end
+
+    context "when records are faulty" do
+      include_examples "invalid collaborator records", :delete, :destroy
+    end
```
sm17p: From shared example

spec/controllers/collaborators/main_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+      expect(response).to redirect_to(edit_collaborator_path(collaborator.external_id))
+      get :edit, params: { id: collaborator.external_id }
+      expect(inertia.component).to eq("Collaborators/Edit")
+      expect(inertia.props[:errors][:message]).to be_present
+    end
+
+
+    context "when records are faulty" do
+      include_examples "invalid collaborator records", :patch, :update
+    end
```
sm17p: From shared example

spec/presenters/collaborator_presenter_spec.rb
```diff
@@ -49,7 +61,7 @@
       props = described_class.new(seller:, collaborator:).collaborator_props
 
       expect(props.except(:products)).to match(collaborator.as_json)
-      expect(props[:products]).to match_array([{ id: visible_product.external_id, name: visible_product.name, percent_commission: nil }])
+      expect(props[:products]).to match_array([{ id: visible_product.external_id, name: visible_product.name, percent_commission: collaborator.affiliate_percentage }])
```
sm17p, line 64: This was needed as we always send either of `product.affiliate_percentage || collaborator.affiliate_percentage` now. This doesn't break anything per se and just simplfies frontend a bit to avoid `null` handling at certain places

spec/presenters/collaborator_presenter_spec.rb
```diff
@@ -63,24 +75,78 @@
...
+      expect(props).to include(
+        id: collaborator.external_id,
+        apply_to_all_products: collaborator.apply_to_all_products,
+        affiliate_percentage: collaborator.affiliate_percentage,
+        dont_show_as_co_creator: collaborator.dont_show_as_co_creator,
+        invitation_accepted: collaborator.invitation_accepted?,
+        seller_email: seller.email,
+        seller_name: seller.display_name(prefer_email_over_default_username: true),
+        seller_avatar_url: seller.avatar_url,
+      )
```
sm17p, line 135: Being more explicit with the checks over here

spec/presenters/collaborator_presenter_spec.rb
```diff
@@ -63,24 +75,78 @@
...
       expect(props[:products]).to match_array([
-                                                { id: visible_product.external_id, name: visible_product.name, published: true, has_another_collaborator: false, has_affiliates: false, enabled: true, percent_commission: nil, dont_show_as_co_creator: false },
-                                                { id: ineligible_product.external_id, name: ineligible_product.name, published: true, has_another_collaborator: true, has_affiliates: false, enabled: false, percent_commission: nil, dont_show_as_co_creator: false },
+                                                {
+                                                  id: product.external_id,
+                                                  url: product.long_url,
+                                                  name: product.name,
+                                                  affiliate_percentage: collaborator.affiliate_percentage,
+                                                  dont_show_as_co_creator: product_affiliate.dont_show_as_co_creator,
+                                                },
```
sm17p, line 142: Only need these fields on frontend

```tsx
  products: {
    id: string;
    url: string;
    name: string;
    affiliate_percentage: number;
    dont_show_as_co_creator: boolean;
  }[];
```

spec/requests/collaborations_spec.rb
```diff
@@ -69,6 +69,14 @@
...
         expect(page).to have_tab_button("Collaborators")
         expect(page).to have_tab_button("Collaborations")
+
+        click_on "Collaborations"
+
+        expect(page).to have_current_path(collaborators_incomings_path)
+        expect(page).to have_title("Collaborators")
+        expect(page).to have_selector("h1", text: "Collaborators")
+        expect(page).to have_tab_button("Collaborators")
+        expect(page).to have_tab_button("Collaborations")
```
sm17p, line 79: Adding a few assertions for completeness in this spec. These weren't tested anywhere for incomings page

spec/requests/collaborators_spec.rb
```diff
@@ -12,6 +12,7 @@
         it "displays a placeholder message" do
           visit collaborators_path
 
+          expect(page).to have_title("Collaborators")
```
sm17p: Explicitly checking for completeness

spec/requests/collaborators_spec.rb
```diff
@@ -124,12 +125,17 @@
           visit collaborators_path
           click_on "Add collaborator"
 
+          expect(page).to have_title("Collaborators")
+          expect(page).to have_selector("h1", text: "New collaborator")
+          expect(page).to_not have_tab_button("Collaborators")
+          expect(page).to_not have_tab_button("Collaborations")
```
sm17p: Explicitly checking for completeness

spec/requests/collaborators_spec.rb
```diff
@@ -445,7 +451,10 @@
             click_on "Edit"
           end
 
-          expect(page).to have_text collaborator.affiliate_user.display_name
+          expect(page).to have_title("Collaborators")
+          expect(page).to have_selector("h1", text: collaborator.affiliate_user.display_name)
+          expect(page).to_not have_tab_button("Collaborators")
+          expect(page).to_not have_tab_button("Collaborations")
```
sm17p: Explicitly checking for completeness

sm17p (commented):
@EmCousin I've made a few changes to align with recent migration patterns and also addressed previous review comments. Please have a look again 🙏

<hr />

### Local Test Run After Recent Changes


Image

sm17p:
@EmCousin Unrelated specs are failing (They seem flaky as they pass on local run). Please have a look

Ref- 

1. https://github.com/antiwork/gumroad/actions/runs/20751194872/job/59581322127?pr=2565
    - Screenshot 2026-01-06 at 8 27 52 PM

2. https://github.com/antiwork/gumroad/actions/runs/20751194872/job/59581322080?pr=2565
    - Screenshot 2026-01-06 at 8 28 32 PM

### Local Run 

Screenshot 2026-01-06 at 8 39 21 PM

app/javascript/components/Collaborators/Form.tsx
EmCousin: Makes sense!

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
-        has_error: product.enabled && !applyToAllProducts && !validCommission(product.percent_commission),
-      })),
-    );
-    setDefaultPercentCommission({
-      ...defaultPercentCommission,
-      hasError: applyToAllProducts && !validCommission(defaultPercentCommission.value),
+  const submitForm = (acknowledgement?: "withConfirmedAcknowledgement") => {
+    form.data.products.forEach((product, index) => {
+      form.setData(
+        `products.${index}.has_error`,
```
EmCousin: Please use Inertia's form helper to manage errors too so we no longer need to reinvent the wheel here: https://inertiajs.com/docs/v2/the-basics/forms#form-errors

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
-  };
 
-  const handleDefaultCommissionChange = (percent_commission: number | null) => {
-    setDefaultPercentCommission({ value: percent_commission, hasError: false });
-    setProducts((prevProducts) => prevProducts.map((item) => ({ ...item, percent_commission, has_error: false })));
+  const handleProductChange = (id: string, attrs: Partial<CollaboratorFormProduct>) => {
+    form.setData(
+      "products",
+      form.data.products.map((item) => (item.id === id ? { ...item, ...attrs, has_error: false } : item)),
+    );
```
EmCousin: You could use Inertia's form helpers to managen errors and just call `form.clearErrors()` here.

app/controllers/collaborators/incomings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  prepend_before_action :set_collaborator, only: [:accept, :decline, :destroy]
+  before_action :set_invitation!, only: [:accept, :decline]
+
+  def index
+    render inertia: "Collaborators/Incoming/Index", props: IncomingCollaboratorsPresenter.new(seller: current_seller).index_props
```
EmCousin: Forgot an "s" to Incomings - make sure to update the directory too

```suggestion
    render inertia: "Collaborators/Incomings/Index", props: IncomingCollaboratorsPresenter.new(seller: current_seller).index_props
```

app/controllers/collaborators/main_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    render inertia: "Collaborators/New", props: CollaboratorPresenter.new(seller: current_seller).new_collaborator_props
+  end
+
+  def create
+    response = Collaborator::CreateService.new(seller: current_seller, params: collaborator_params).process
+
+    if response[:success]
+      redirect_to collaborators_path, status: :see_other, notice: "Changes saved!"
+    else
+      redirect_to new_collaborator_path, inertia: { errors: { message: response[:message] } }
```
EmCousin: We can now return the entire collaborator`s error payload from the CreateService, instead of constructing the error message ourselves like this. [InertiaRails allows precisely this](https://inertia-rails.dev/guide/validation#how-it-works).

This would also potentially removed a lot of useless code in the frontend form.

app/controllers/collaborators/main_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    render inertia: "Collaborators/Edit", props: CollaboratorPresenter.new(seller: current_seller, collaborator: @collaborator).edit_collaborator_props
+  end
+
+  def update
+    response = Collaborator::UpdateService.new(seller: current_seller, collaborator_id: params[:id], params: collaborator_params).process
+
+    if response[:success]
+      redirect_to collaborators_path, status: :see_other, notice: "Changes saved!"
+    else
+      redirect_to edit_collaborator_path(params[:id]), inertia: { errors: { message: response[:message] } }
```
EmCousin: Same here

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
           ? "Collaborator email must be provided"
-          : !isValidEmail(collaboratorEmail.value)
+          : !isValidEmail(form.data.email ?? "")
             ? "Please enter a valid email"
             : null;
-      setCollaboratorEmail(
-        emailError ? { value: collaboratorEmail.value, error: emailError } : { value: collaboratorEmail.value },
-      );
       if (emailError) {
+        form.setError("email", emailError);
```
EmCousin, line 97: So you _are_ using Inertia's built-in form error management feature. You just need to be consistent across this file.

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
+        (product) =>
+          !validCommission(
+            product.percent_commission,
+            pageMetadata.min_percent_commission,
+            pageMetadata.max_percent_commission,
+          ),
+      )
     ) {
-      showAlert("Collaborator cut must be 50% or less", "error");
+      showAlert(`Collaborator cut must be ${pageMetadata.max_percent_commission}% or less`, "error");
```
EmCousin: 👍

app/controllers/collaborators/main_controller.rb
EmCousin: Also, the flash message should be set here. Could take ideally `collaborator.errors.full_messages.to_sentence`

app/controllers/collaborators/main_controller.rb
EmCousin: Same comment about setting the flash message

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
+
+    if (isEditPage) {
+      form.patch(Routes.collaborator_path(formData.id), {
+        only: ["errors"],
+        onError: (e) => {
+          if (e.message) {
+            showAlert(e.message, "error");
+            router.replaceProp("errors", undefined);
+          }
+        },
```
EmCousin: This should not be needed with a proper redirection upon failure and the flash message set server-side. The form errors should also be updated automatically with the server side errors like explained in the comments in the controller

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -23,194 +15,169 @@ import { Pill } from "$app/components/ui/Pill";
...
+      });
+    } else {
+      form.post(Routes.collaborators_path(), {
+        only: ["errors"],
+        onError: (e) => {
+          if (e.message) {
+            showAlert(e.message, "error");
+            router.replaceProp("errors", undefined);
+          }
+        },
```
EmCousin: Same here, this should not be needed

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -260,13 +227,13 @@ const CollaboratorForm = () => {
...
+                      checked={form.data.apply_to_all_products}
                       onChange={(evt) => {
                         const enabled = evt.target.checked;
-                        setApplyToAllProducts(enabled);
-                        setProducts((prevProducts) =>
-                          prevProducts.map((item) => (shouldEnableProduct(item) ? { ...item, enabled } : item)),
-                        );
+                        form.setData("apply_to_all_products", enabled);
+                        form.data.products.forEach((item, index) => {
+                          if (shouldEnableProduct(item)) form.setData(`products.${index}.enabled`, enabled);
```
EmCousin, line 223: ⚠️ If some products were previously enabled but no longer should, `enabled` is no longer set to `false`

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -275,14 +242,23 @@ const CollaboratorForm = () => {
...
-                    <fieldset className={cx({ danger: defaultPercentCommission.hasError })}>
-                      <NumberInput value={defaultPercentCommission.value} onChange={handleDefaultCommissionChange}>
+                    <fieldset className={cx({ danger: allProductsCommissionHasError })}>
+                      <NumberInput
+                        value={form.data.percent_commission}
+                        onChange={(percentCommissionValue) => {
+                          form.setData("percent_commission", percentCommissionValue);
+                          form.data.products.forEach((_, index) => {
+                            form.setData(`products.${index}.percent_commission`, percentCommissionValue);
+                            form.setData(`products.${index}.has_error`, false);
```
EmCousin: `forn.setError` should have been used here

app/javascript/components/Collaborators/Form.tsx
```diff
@@ -296,22 +272,23 @@ const CollaboratorForm = () => {
...
                         onChange={(evt) => {
                           const value = !evt.target.checked;
-                          setDontShowAsCoCreator(value);
-                          setProducts((prevProducts) =>
-                            prevProducts.map((item) => ({ ...item, dont_show_as_co_creator: value, has_error: false })),
-                          );
+                          form.setData("dont_show_as_co_creator", value);
+                          form.data.products.forEach((_, index) => {
+                            form.setData(`products.${index}.dont_show_as_co_creator`, value);
+                            form.setData(`products.${index}.has_error`, false);
```
EmCousin: Same here

app/javascript/pages/Collaborators/Incoming/Index.tsx
```diff
@@ -0,0 +1,235 @@
...
+      ) : (
+        <IncomingCollaboratorsTable
+          incomingCollaborators={incomingCollaborators}
+          selected={selected}
+          disabled={form.processing}
+          onSelect={(collaborator) => setSelected(collaborator)}
+          onAccept={(incomingCollaborator) =>
+            form.post(Routes.accept_collaborators_incoming_path(incomingCollaborator.id), {
+              only: ["collaborators", "flash"],
+              preserveUrl: true,
```
EmCousin: Why do we need to preserve the URL here?

app/javascript/pages/Collaborators/Incoming/Index.tsx
```diff
@@ -0,0 +1,235 @@
...
+            })
+          }
+          onReject={(incomingCollaborator) =>
+            form.post(Routes.decline_collaborators_incoming_path(incomingCollaborator.id), {
+              only: ["collaborators", "flash"],
+              preserveUrl: true,
```
EmCousin: Same question

app/javascript/pages/Collaborators/Incoming/Index.tsx
```diff
@@ -0,0 +1,235 @@
...
+            })
+          }
+          onRemove={(incomingCollaborator) =>
+            form.delete(Routes.collaborators_incoming_path(incomingCollaborator.id), {
+              only: ["collaborators", "flash"],
+              preserveUrl: true,
```
EmCousin: Same question

app/javascript/pages/Collaborators/Index.tsx
```diff
@@ -1,3 +1,190 @@
...
+    usePage().props,
+  );
+  const [selectedCollaborator, setSelectedCollaborator] = React.useState<Collaborator | null>(null);
+
+  const deleteForm = useForm({});
+
+  const remove = (collaboratorId: string) => {
+    deleteForm.delete(Routes.collaborator_path(collaboratorId), {
+      only: ["collaborators", "flash", "has_incoming_collaborators"],
+      preserveUrl: true,
```
EmCousin: Same question

app/javascript/utils/collaboratorFormatters.ts
EmCousin: This abstraction is a DRY pitfall: it couples unrelated domains (seller collaborators vs incoming collaborations) and relies on a runtime shape check ("percent_commission" in collaborator).

Instead, normalize at the boundary: have each presenter expose a `default_percent_commission` prop . That keeps the serialized props shape consistent, avoids cross-domain coupling in client utils, and makes the UI more declarative.

app/javascript/utils/collaboratorFormatters.ts
```diff
@@ -0,0 +1,54 @@
...
+      const lowestCommission = commissions[0];
+      const highestCommission = commissions[commissions.length - 1];
+      if (lowestCommission !== undefined && highestCommission !== undefined) {
+        return `${formatAsPercent(lowestCommission)} - ${formatAsPercent(highestCommission)}`;
+      }
+    }
+  }
+
+  return formatAsPercent(getDefaultCommission(collaborator));
+};
```
EmCousin, line 38: Same here. Once this is decoupled and `default_percent_commission` is declared server-side, this function will be much, much simpler.

EmCousin (changes requested):
Thankks @sm17p - the implementation is mostly good (and CI is green), but some changes are still needed to make it simpler, right now Inertia's form helpers are not used consistently everywhere, and some abstractions you've made end up making the code more coupled and harder to maintain.

More details in the comments - let me know if you have questions.

app/javascript/pages/Collaborators/Incoming/Index.tsx
sm17p: You're right, we don't need it. Removing this 🙏

app/javascript/components/Collaborators/Form.tsx
sm17p: Nice catch! I ran the below spec on main and this is a bug on production as well. Fixing it and adding the spec as well 

```ruby
      it "disables unpublished products that no longer pass shouldEnableProduct when unchecking Apply to all products" do
        unpublished_product = create(:product, user: seller, name: "Unpublished product", purchase_disabled_at: Time.current)

        visit "/collaborators/new"

        expect(page).not_to have_content unpublished_product.name

        check "Show unpublished and ineligible products"

        within find(:table_row, { "Product" => unpublished_product.name }) do
          expect(page).to have_checked_field(unpublished_product.name)
        end

        fill_in "email", with: collaborating_user.email

        expect(page).to have_checked_field("All products")

        uncheck "Show unpublished and ineligible products"

        expect(page).not_to have_content unpublished_product.name

        expect do
          click_on "Add collaborator"

          expect(page).to have_alert(text: "Changes saved!")
          expect(page).to have_current_path(collaborators_path)
        end.to change { seller.collaborators.count }.by(1)

        collaborator = seller.collaborators.last
        expect(collaborator.products).not_to include(unpublished_product)
      end
    end
    
```

### Local Run Failing on Main

Screenshot 2026-01-07 at 11 14 40 AM

app/controllers/collaborators/main_controller.rb
```diff
@@ -41,8 +43,17 @@ def destroy
...
 
+
   private
+    def inertia_errors_props(model)
+      errors_hash = model.errors.to_hash.transform_values(&:to_sentence)
+      {
+        errors: errors_hash,
+        alert: errors_hash[:base],
+      }
+    end
```
sm17p, line 54: This was one of the ways recommended by inertia to maps rails validations errors with the expected form error type. 

Ref - https://inertia-rails.dev/cookbook/handling-validation-error-types#option-2-serialize-errors-in-rails

app/javascript/components/Collaborators/Form.tsx
sm17p: @EmCousin The previous code at this point seems to be functionally equivalent to what's there right now unless I'm missing something 

```tsx
// Before
onChange={(evt) => {
  const enabled = evt.target.checked;
  setApplyToAllProducts(enabled);
  setProducts((prevProducts) =>
    prevProducts.map((item) => (shouldEnableProduct(item) ? { ...item, enabled } : item)),
  );
}}

// After

onChange={(evt) => {
  const enabled = evt.target.checked;
  form.setData("apply_to_all_products", enabled);
  form.data.products.forEach((item, index) => {
    if (shouldEnableProduct(item)) form.setData(`products.${index}.enabled`, enabled);
  });
}}
````

Does this boil down to the bug posted in the previous comment? Addressed in [314bd0f](https://github.com/antiwork/gumroad/pull/2565/commits/314bd0f7c1379df47b8e39052fba4ab5607ca84a) commit

File: app/controllers/collaborators/main_controller.rb (hunk 60 above)
Comment (sm17p) [reply app/controllers/collaborators/main_controller.rb]: Gave it a shot, does https://github.com/antiwork/gumroad/pull/2565#discussion_r2667623981 sound like a reasonable approach to this problem?

File: app/javascript/utils/collaboratorFormatters.ts (hunk 73 above)
Comment (sm17p) [reply app/javascript/utils/collaboratorFormatters.ts:38]: > This abstraction is a DRY pitfall: it couples unrelated domains (seller collaborators vs incoming collaborations) and relies on a runtime shape check ("percent_commission" in collaborator).

> Instead, normalize at the boundary: have each presenter expose a default_percent_commission prop . That keeps the serialized props shape consistent, avoids cross-domain coupling in client utils, and makes the UI more declarative.

Do you mean to say that we should use the same key `percent_commision` for `Collaborator` and `IncomingCollaborator` with a default fallback?

```ts
const getDefaultCommission = (collaborator: Collaborator | IncomingCollaborator): number => {
    return collaborator.percent_commission ?? collaborator.default_percent_commision;
};
```

I'm a bit confused since from what I've seen on my production account and service code `affilate_percentage` is never `nil` on either of `Collaborator` or `ProdutAffiliate` models but they underlying DB field can be nil, making me question if there should even be a fallback?

sm17p:
Thanks you for the detailed review @EmCousin. 

I've addressed few of the review comments but  couldn't completely understand the following ones which might need your attention

1. https://github.com/antiwork/gumroad/pull/2565#discussion_r2665414158
2. https://github.com/antiwork/gumroad/pull/2565#discussion_r2665460363

Can you please help me get unblock on them? 😅

app/javascript/utils/collaboratorFormatters.ts
EmCousin: Not quite. I’m not suggesting a TS-level fallback across two shapes. I’m suggesting we normalize the serialized props contract server-side so both Collaborators and IncomingCollaborators expose the same fields (e.g. default_percent_commission and products[].percent_commission as non-null numbers). Then the client formatter can accept one normalized type and won’t need runtime shape checks or fallbacks.

app/javascript/components/Collaborators/Form.tsx
EmCousin: Sounds like a bug. Keep the current behaviour for now, we should deal with this as a separate issue

EmCousin:
~@Devwill2~ @sm17p  _a I addressed your questions, please check

Devwill2:
> @Devwill2 _a I addressed your questions, please check

I might be missing something, but I believe this PR belongs to someone else and I may have been tagged by mistake. 😅
@EmCousin

EmCousin:
Whoops, apologies @Devwill2. @sm17p Please check my answers.

sm17p:
Looking into them 👍

sm17p:
Screenshot 2026-01-07 at 9 33 28 PM

Got a clean run, please have a look 🙏

sm17p:
@EmCousin I think we'll have to run CI again as I saw a lot of timeouts and stripe throttling across the failed jobs

Thanks for the quick run. Please have a look again 🙏

sm17p:
Screenshot 2026-01-08 at 8 03 26 PM

I think the last failing spec is after #2755 merge, not sure? 

Linking the failed runs
- https://github.com/antiwork/gumroad/actions/runs/20819311190/job/59803734025?pr=2565 (From run on this PR)
- https://github.com/antiwork/gumroad/actions/runs/20819206389/job/59804030201 (From main run)

sm17p:
@EmCousin The chekout spec failed again and seems to be related to #2755 ( @stefanbinoj just a light ping for confirmation 🙏 ref - https://github.com/antiwork/gumroad/pull/2565#issuecomment-3724172732)

Screenshot 2026-01-08 at 8 22 22 PM

stefanbinoj:
sorry for bothering @sm17p but unfortunately all test passed in CI for that branch which caused this trouble. 
When i tried running tests locally i could replicate this and came up with fix #2766

sm17p:
No worries @stefanbinoj, stuff like this is sometimes unavoidable. It's all good 👌

stefanbinoj:
Thanks for understanding @sm17p ! 
@EmCousin sorry for this trouble all tests passed in my branch while running CI thus i missed this one 🙏

sm17p:
Thank you @stefanbinoj @nyomanjyotisa for the quick follow-up on the issue the spec is not failing anymore ❤️

@EmCousin I ran the failing spec from the last run on local branch and it seems flaky and unrelated, please have a look

Ref - https://github.com/antiwork/gumroad/actions/runs/20828517568/job/59838089428?pr=2565

not-failing

EmCousin (approved):

EmCousin:
Thanks @sm17p!

stefanbinoj:
Cheers mate! 🚀 
And apologies for earlier

sm17p:
Thank you as well @EmCousin for all the help

sm17p:
> Cheers mate! 🚀 
> And apologies for earlier 

Cheers to you as well, it's all good @stefanbinoj you don't have to apologise for anything it's understandable 😅

PR #2514, opened 2025-12-28
Inertia Migration: Audience Page

sm17p:
# Part of #856 
- Related PR #2496 (Sliced out code related to Audience page from this PR after verifying end-to-end inertia migration for Analytics Pages)

This PR migrates `/dashboard/audience` page to inertia. 

## Code Changes
- Using props as source of truth for audience chart data in frontend
- Greatly simplified frontend using Inertia.defer + `Deferred` component to load audience chart data with a fallback. This removes the need of managing our own loader.
- Subsequent data request on date change are handled by passing a callback to `useAnalyticsDateRange` hook which reloads the page using inertia

### Demos

1. `Deferred` component handles loading audience data on first load

https://github.com/user-attachments/assets/4916660e-04f0-42c5-bbec-f6ad740b8800

2. Flash error to handle invalid date range (end_date < start_date)

https://github.com/user-attachments/assets/e2077d32-da7c-48d9-b2c8-a62c67765166

### Specs Passing

specs-passing

## AI Disclosure
- Cursor auto mode for edits, migrating specs, checking docs and debugging 

### Live Stream Disclosure 
Have watched all the live streams

PR #2572, opened 2025-12-31
Migrate audience page to Inertia

binary-koan:
Migrates the Audience page to Inertia: a combination of #2484 by @Devwill2, #2463 by @4arjun, #2514 by @sm17p, and a few of my own ideas.

### DIFF
--- a/app/controllers/audience_controller.rb
+++ b/app/controllers/audience_controller.rb
@@ -14,7 +14,7 @@ class AudienceController < Sellers::BaseController
     total_follower_count = current_seller.audience_members.where(follower: true).count
 
     render inertia: "Audience/Index", props: {
-      total_follower_count:,
+      total_follower_count: InertiaRails.always { total_follower_count },
       audience_data: InertiaRails.defer do
         if total_follower_count.zero?
           nil


app/controllers/audience_controller.rb
```diff
@@ -25,26 +40,18 @@ def export
...
-  end
-
   protected
     def set_time_range
       begin
-        end_time = DateTime.parse(params[:end_time])
-        start_date = DateTime.parse(params[:start_time])
+        end_date = DateTime.parse(params[:to])
+        start_date = DateTime.parse(params[:from])
+        end_date = start_date if end_date < start_date
```
binary-koan, line 48: Prevents errors if an invalid date range is sent (added a frontend error message as I don't think this is serious enough for a flash message)

app/javascript/pages/Audience/Index.tsx
```diff
@@ -16,39 +15,29 @@ import { ExportSubscribersPopover } from "$app/components/Followers/ExportSubscr
...
-        setData(await request.response);
-        activeRequest.current = null;
-      } catch (e) {
-        if (e instanceof AbortError) return;
-        showAlert("Sorry, something went wrong. Please try again.", "error");
-      }
-    };
-    void loadData();
-  }, [startTime, endTime]);
+  useOnChange(() => {
```
binary-koan, line 37: The initial render is handled by Inertia, we just want to reload when startTime and endTime are updated

app/controllers/audience_controller.rb
sm17p: That's neat! Approved 👍

app/controllers/audience_controller.rb
```diff
@@ -1,16 +1,31 @@
...
 
+  layout "inertia", only: :index
+
   def index
     authorize :audience
 
-    @total_follower_count = current_seller.audience_members.where(follower: true).count
+    total_follower_count = current_seller.audience_members.where(follower: true).count
+
+    render inertia: "Audience/Index", props: {
```
sm17p, line 16: (Nit, nothing major) Is there a specific reason to keep it at the top level like this? It feels confusing a bit as I expect this to render at `/audience`

Maybe we should structure it to match with the route since sister projects follow a similar pattern. It might make it easier for contributors to navigate through the projects, wdyt?

```suggestion
    render inertia: "Dashboard/Audience/Index", props: {
```

EmCousin: Controller is at the top level, therefore view follows the same logic, it makes the code base more consistent and predictable. That being said, namespacing everything under "dashboard" could make sense, but we would need to restructure the entire dashboard pages consistenty as a separate effort in my opinion, provided this would relevant to the users.

binary-koan: Yeah that’s basically why. It would make sense to me to change the route to /analytics/audience, as it’s a little odd to have an analytics layout that several unrelated-sounding paths use, but not worth breaking people’s bookmarks to do that IMO

sm17p: > but we would need to restructure the entire dashboard pages consistenty as a separate effort in my opinion, provided this would relevant to the users.

Makes sense, feels low ROI the way you put it 👍

 >  It would make sense to me to change the route to /analytics/audience, as it’s a little odd to have an analytics layout that several unrelated-sounding paths use, but not worth breaking people’s bookmarks to do that IMO but not worth breaking people’s bookmarks to do that IMO
 
Hmmm, we can add re-direction routes later on anyway from '/dashboard/audience` to `/analytics/audience` if the need arises. 

<hr />

Everything looks fine on my end from local testing and the failing specs also seem to be unrelated. 

For me it's good to go, the above comment is non-blocking 🚀

4arjun: Tested on local. Looks good.

app/controllers/audience_controller.rb
sm17p, line 58:
```suggestion
      @title = "Audience"
```

(Nit: forgot to add this earlier) Maybe this too for UX, so that `Analytics` tabs can be identified distinctly when opened together? 🤔

binary-koan: Something to consider but we shouldn’t change it in the migration IMO. If we change it we should make it “Following” to match the UI

binary-koan:
Thanks for testing/checking! Will go ahead and merge

sm17p:
No worries, happy to help

Thank you @binary-koan @EmCousin  for all the feedbacks along the way as well. It's helping a lot! <3

PR #2612, opened 2026-01-01
Fix PPP specs to work after 2025

binary-koan:
## Problem

These specs are currently failing because of how we load PPP factors:

* We use the public World Bank API to fetch PPP factors in `UpdatePurchasingPowerParityFactorsWorker`
* We have a sanity check in the worker that the data is recent enough (within 2 years)
* The specs use VCR cassettes which we last refreshed on Jan 1, 2025. Once the year changed to 2026 the sanity check started failing and the tests broke

Example run:
https://github.com/antiwork/gumroad/actions/runs/20640324879/job/59271217225
https://github.com/antiwork/gumroad/actions/runs/20640324879/job/59271217234

## Solution

Updated the tests to travel to a specific working date so we don't need to keep refreshing the VCR cassettes.

---

# AI Disclosure

Not used

### DIFF
--- a/app/javascript/components/Admin/SalesReports/JobHistory.tsx
+++ b/app/javascript/components/Admin/SalesReports/JobHistory.tsx
@@ -3,7 +3,7 @@ import * as React from "react";
 import AdminSalesReportsForm from "$app/components/Admin/SalesReports/Form";
 import { Button } from "$app/components/Button";
 import { Icon } from "$app/components/Icons";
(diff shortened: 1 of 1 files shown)


binary-koan:
@ershad Merging this to fix CI on contributor PRs, but let me know if you have a better idea! Looks like in the past we've regenerated the VCR cassettes at the start of the year, but I don't think it's a good idea to have tests that rely on that

Issue #2562, opened 2025-12-30
[Bug] Flash message persists across Inertia page navigations


4arjun:
## Description

Flash messages rendered on Inertia pages persist across subsequent navigations, causing stale banners to reappear even when no new action has occurred.

For example, after successfully publishing/sending an email, the success banner (“Email successfully sent”) continues to show up when navigating to unrelated pages and then returning, despite no new send action being triggered.

## Steps to Reproduce

1. Navigate to New Email
2. Publish/send an email
3. Observe the success banner (e.g. “Email successfully sent”) 
4. Navigate to another page (e.g. New Email or any other Inertia page)
5. Navigate back to the previous page

## Expected Behavior

- Flash messages should be shown only once, immediately after the action that triggered them
- Navigating between Inertia pages should not re-render old flash messages

## Screen recording

https://github.com/user-attachments/assets/3fbb5754-46e1-4ca6-be17-1888fabeb8cb

yashranaway:
@4arjun i am unable to reproduce this on prod. 

https://github.com/user-attachments/assets/3d87ce7c-06be-4c2f-a397-e8d078a42c47


cc: @EmCousin

EmCousin:
This is open source software, anyone is free to push a PR.

PR #2613, opened 2026-01-01
fix(Flash): Flash message persists across Inertia page navigations

tushar994:
Issue: #2562 

# Description

Flash messages rendered on Inertia pages persist across subsequent navigations, causing stale banners to reappear even when no new action has occurred.

For example, after successfully publishing/sending an email, the success banner (“Email successfully sent”) continues to show up when navigating to unrelated pages and then returning, despite no new send action being triggered.

## Solution

- Clear flash from Inertia's cache after showing it by reloading only the flash prop
- We use router.replace, which is a client side visit. 
- for reference - https://inertiajs.com/docs/v2/the-basics/manual-visits#client-side-visits

---

# Before/After
**before**

https://github.com/user-attachments/assets/e1d4c35a-3e5a-40e6-85e1-5d1076431275

**after**

https://github.com/user-attachments/assets/f7221610-9fa3-4438-8040-78f1fc389f5c

# Test Results

---

# Checklist

- [ ] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)

---

# AI Disclosure

Issue: https://github.com/antiwork/gumroad/issues/2562

### DIFF
--- a/app/javascript/components/useFlashMessage.ts
+++ b/app/javascript/components/useFlashMessage.ts
@@ -8,9 +8,14 @@ export function useFlashMessage(flash?: AlertPayload) {
     if (flash?.message) {
       showAlert(flash.message, flash.status === "danger" ? "error" : flash.status);
 
-      // Clear the flash from the current page's cached props to prevent it from
-      // reappearing when navigating back via browser history
-      router.reload({ only: ['flash'] });
+      // Clear the flash from Inertia's cache by updating props client-side
+      // This does NOT make a server request - it only updates the local cache
+      router.replace({
+        props: (currentProps) => ({
+          ...currentProps,
+          flash: null,
+        }),
+      });
     }
   }, [flash]);
 }


tushar994:
hey @EmCousin, hope this looks better!!

app/javascript/components/useFlashMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+  React.useEffect(() => {
+    if (flash?.message) {
+      showAlert(flash.message, flash.status === "danger" ? "error" : flash.status);
+
+      // Clear the flash from Inertia's cache by updating props client-side
+      // This does NOT make a server request - it only updates the local cache
+      router.replace({
+        props: (currentProps) => ({
+          ...currentProps,
+          flash: null,
```
stefanbinoj, line 16: One comment instead of resetting entire prop can we just do it like this instead? 
```suggestion
      router.replaceProp('flash', null)
```
[docs](https://inertiajs.com/docs/v2/the-basics/manual-visits#prop-helpers)

app/javascript/components/useFlashMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+        }),
+      });
+    }
+  }, [flash]);
```
sm17p, line 20: I see

Is it possible to do the same thing inside `server-componens/Alert.tsx`? Maybe that way you don't have to define this hook

EmCousin (commented):
Thanks @tushar994 , but `router.replaceProp` would be much cleaner.

PR #2614, opened 2026-01-01
fix: Flash message persists across Inertia page navigations

wfoxe:
Issue: #2562

# Description

## Problem
Flash messages rendered on Inertia pages persist across subsequent navigations, causing stale banners to reappear even when no new action has occurred.

For example, after successfully publishing/sending an email, the success banner (“Email successfully sent”) continues to show up when navigating to unrelated pages and then returning, despite no new send action being triggered.
(this has been copied directly from the issue)

I found this issue to be reproducible and sitewide. The other proposed fixes I reviewed were incomplete or introduced new issues.
## Solution
My solution is fairly simple: stop shoving flash into shared props (which Inertia stores in history) and instead rely on top-level `page.flash`, which Inertia explicitly clears from history during navigation. The client reads `page.flash` and turns it into a toast.

While validating the fix, I hit a follow-up regression: identical toasts no longer reappeared on repeated actions (e.g. submitting the same invalid password twice on the same page). This matched neither production nor expected behavior. The fix there was to trigger the toast effect on each new Inertia response by keying the effect on the `flash` object itself (not just its message/status), so identical messages still surface every time the server responds with flash.

Possible edges I noticed that may need review:
- Only `notice`, `alert`, and `warning` are forwarded into `page.flash` for Inertia responses. Non-Inertia flows still use Rails’ regular flash rendering.
- Flash-only partial reloads should still work, but if the server returns an identical flash payload, it relies on the Inertia response object changing (which should happen per response).
---

# Before/After

Desktop Before

https://github.com/user-attachments/assets/ff44c1b8-5271-4f86-84ea-c0f6c0ac4e6f

https://github.com/user-attachments/assets/7e91a828-5268-4509-80ea-c613b5835b69

Mobile Before

https://github.com/user-attachments/assets/e611669a-f82b-42b2-9023-c62dafd2bc7a

https://github.com/user-attachments/assets/74fe89cb-7adf-4849-9a6c-6b7f86e836eb

Desktop After

https://github.com/user-attachments/assets/6d0cbe08-364f-47e4-abb5-b295944ea0ef

https://github.com/user-attachments/assets/eded722e-8e04-42bc-ae78-2bae0f79bbc8

Mobile After

https://github.com/user-attachments/assets/af154a52-1701-4e08-9c82-9b81daf0976f

https://github.com/user-attachments/assets/516a7f20-1305-48b0-aa93-0cfe03004e2f

---

# Test Results

Screenshot 2026-01-01 at 6 02 55 PM

Ran locally (macOS, Spring disabled):
- `DISABLE_SPRING=1 bin/rspec spec/controllers/workflows_controller_spec.rb` (pass)
- `DISABLE_SPRING=1 bin/rspec spec/requests/emails/create_spec.rb` (fail: Stripe balance/VCR and purchase creation failures due to card charge validation)

Test coverage note:
- Added a controller spec to assert Inertia `page.flash` is populated and `props[:flash]` is absent for a flash response.
- The back/forward history behavior is currently verified manually (hard to simulate reliably in request specs).

---

# Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
- [x] I have performed a self-review and left comments on my PR

---

# AI Disclosure

Codex 5.2 was used to aggregate docs related to the issue, surmise possible fixes, test the fix that made the most sense, and test edges and regressions. Throughout the process, I reviewed code changes manually, and ensured that the AI did not create complications or unnecessary changes. I cross checked the fix with Claude Opus 4.5, and the results were consistent.

# Self Review
  - Verified flash no longer persists in history; back/forward doesn’t re‑show stale toasts.
  - Confirmed identical toasts still appear on repeated actions.
  - Added controller spec for Inertia page.flash vs props.flash.
  - Manually tested and envisioned other edges, found no complications other than the one noted and fixed.
  
  
This is my first PR on the gumroad repo, if I missed any requirements for contributions please let me know.

### DIFF
--- a/app/javascript/components/server-components/Alert.tsx
+++ b/app/javascript/components/server-components/Alert.tsx
@@ -11,6 +11,45 @@ const ALERT_KEY = "alert";
 
 export type AlertPayload = { message: string; status: "success" | "danger" | "info" | "warning"; html?: boolean };
 
+type FlashData = Record<string, unknown>;
+
+const flashStatus = (flash: FlashData) => {
+  if (typeof flash.alert === "string") return "danger";
+  if (typeof flash.warning === "string") return "warning";
+  if (typeof flash.notice === "string") return "success";
+  return null;
+};
+
+const flashMessage = (flash: FlashData) =>
+  (typeof flash.alert === "string" && flash.alert) ||
+  (typeof flash.warning === "string" && flash.warning) ||
+  (typeof flash.notice === "string" && flash.notice) ||
+  null;
+
+export const flashToAlert = (flash?: FlashData | null): AlertPayload | null => {
+  if (!flash || typeof flash !== "object") return null;
+
+  if (
+    typeof flash.message === "string" &&
+    (flash.status === "success" || flash.status === "danger" || flash.status === "info" || flash.status === "warning")
+  ) {
+    const payload: AlertPayload = {
+      message: flash.message,
+      status: flash.status,
+    };
+    if (typeof flash.html === "boolean") {
+      payload.html = flash.html;
+    }
+    return payload;
+  }
+
+  const status = flashStatus(flash);
+  const message = flashMessage(flash);
+  if (!status || !message) return null;
+
+  return { message, status };
+};
+
 const ToastAlert = ({ initial }: { initial: AlertPayload | null }) => {
   const [alert, setAlert] = React.useState<AlertPayload | null>(initial);
   const [isVisible, setIsVisible] = React.useState(!!initial);
--- a/config/initializers/inertia_rails.rb
+++ b/config/initializers/inertia_rails.rb
@@ -3,3 +3,20 @@
 InertiaRails.configure do |config|
   config.deep_merge_shared_data = true
 end
+
+module InertiaRails
+  class Renderer
+    unless method_defined?(:page_without_flash)
+      alias_method :page_without_flash, :page
+    end
+
+    def page
+      page = page_without_flash
+      return page if page.key?(:flash)
+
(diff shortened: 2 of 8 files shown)


config/initializers/inertia_rails.rb
```diff
@@ -3,3 +3,20 @@
...
+    def page
+      page = page_without_flash
+      return page if page.key?(:flash)
+
+      flash_data = @controller.flash.to_hash.symbolize_keys.slice(:notice, :alert, :warning).compact_blank
+      page[:flash] = flash_data if flash_data.present?
+      page
+    end
+  end
+end
```
EmCousin, line 22: `page.flash` is indeed not available yet in InertiaRails, only on the `master` branch as of now. But this monkeypatch does not match [their implementation](https://github.com/inertiajs/inertia-rails/blob/master/lib/inertia_rails/flash_extension.rb)

app/javascript/components/server-components/Alert.tsx
```diff
@@ -11,6 +11,45 @@ const ALERT_KEY = "alert";
...
+    }
+    return payload;
+  }
+
+  const status = flashStatus(flash);
+  const message = flashMessage(flash);
+  if (!status || !message) return null;
+
+  return { message, status };
+};
```
EmCousin, line 51: Please self-review this code. Took me a moment to understand you moved this mapping logic from the `inerti_rendering` concern to the frontend. There also seem to be substantial changes about how HTML is managed that was not part of the code before, which is not in this PR's scope. If this is a defect, this is better off pointed out in a separate issue.

PR #2617, opened 2026-01-02
fix(payments): preserve cached discount when offer code is deleted or maxed out

FusionBrah:
Issue: #1410

# Description

## Problem

When an offer code has a max usage limit, it is applied correctly to the first installment but not to subsequent installments if the offer code is deleted or the max purchase count is reached between installments.

The `original_offer_code` method was checking if the offer code record was deleted **before** checking if the discount was cached, causing it to return `nil` even when a valid cached discount existed.

## Solution

Reordered the logic in `original_offer_code` to check the cached discount FIRST. If a cached discount exists in `purchase_offer_code_discount`, use it regardless of the current state of the offer code record. This ensures installment plans and recurring charges continue to receive their discount even if the offer code is later deleted or reaches its max usage limit.

---

# Before/After

No UI changes - back-end only. Before/after videos not applicable.

---

# Test Results

Screenshot 2026-01-02 at 5 56 28 pm

---

# Checklist

---

# AI Disclosure

This PR was developed with assistance from **Claude Code (Claude Opus 4.5)**:
- **Bug investigation**: Traced code flow through `original_offer_code`, `offer_amount_off`, `minimum_paid_price_cents`, and `current_subscription_price_cents` to identify root cause
- **Fix implementation**: Reordered logic to check cached discount before deleted status
- **Test creation**: Created 10 comprehensive tests covering maxed out and soft-deleted scenarios

All code has been reviewed and verified for correctness.


app/models/purchase.rb
```diff
@@ -2396,14 +2396,13 @@ def displayed_price_per_unit_cents
   end
 
   def original_offer_code(include_deleted: false)
-    return nil if offer_code&.deleted? && !include_deleted
-
     if has_cached_offer_code?
```
FusionBrah, line 2411: **Self-review**: This is the key fix. By checking the cache first, we ensure that installment plans continue to receive their discount even if the offer code is later deleted or maxed out.

app/models/purchase.rb
```diff
@@ -2396,14 +2396,13 @@ def displayed_price_per_unit_cents
...
-      code = purchase_offer_code_discount.offer_code.code
+      code = purchase_offer_code_discount.offer_code&.code || purchase_offer_code_discount.offer_code_id.to_s
```
FusionBrah, line 2412: **Self-review**: Added safe navigation and fallback to offer_code_id in case the offer code record is deleted but we still need a code identifier.

app/models/purchase.rb
```diff
@@ -2396,14 +2396,13 @@ def displayed_price_per_unit_cents
...
       purchase_offer_code_discount.offer_code_is_percent ?
         OfferCode.new(amount_percentage: purchase_offer_code_discount.offer_code_amount, code:) :
         OfferCode.new(amount_cents: purchase_offer_code_discount.offer_code_amount, code:)
     else
+      return nil if offer_code&.deleted? && !include_deleted
```
FusionBrah, line 2417: **Self-review**: Moved the deleted check here so it only applies when there is NO cached discount. This preserves the original behavior for non-cached cases.

FusionBrah (commented):
Self-review comments for the fix.

PR #2618, opened 2026-01-02
feat(payments): auto-remove LowBalanceFraudCheck probation when balance exceeds $100

FusionBrah:
Issue: #1884

# Description

## Problem

When a user's balance drops below -$100 due to refunds/chargebacks, they are automatically put on probation by `LowBalanceFraudCheck`. However, there was no mechanism to automatically remove this probation when their balance recovered.

Previous 5 PRs were rejected because they incorrectly reverted all users to "compliant" state, but users can be "not_reviewed" before being probated.

## Solution

- Store the user's previous `user_risk_state` in the probation comment's `json_data` when probating
- Add `check_for_high_balance_and_remove_low_balance_probation` method that:
  - Triggers when balance exceeds $100
  - Reverts to the stored previous state (not just "compliant")
  - Only affects LowBalanceFraudCheck probations (not manual admin probations)
- Add `mark_not_reviewed` state machine event for proper state restoration

---

# Before/After

Backend-only change - no UI modifications.

---

# Test Results

Screenshot 2026-01-02 at 6 18 48 pm

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

AI (Claude Code) was used for:
- Analyzing the codebase and understanding why 5 previous PRs were rejected
- Writing the implementation code (state machine event, concern methods, job integration)
- Writing test cases
- Creating the PR description and self-review comments

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1884, opened 2025-11-04
Auto-remove LowBalanceFraudCheck probation when user balance exceeds $100


vatsalkaushik:
Users automatically put on probation for suspicious refunds (`LowBalanceFraudCheck`) should be automatically marked compliant again once their balance goes above $100.

<img width="1700" height="274" alt="Image" src="https://github.com/user-attachments/assets/f1f4ef13-2d8d-4fd4-82ab-9d5b6c2ea954" />

knightdev159:
Hi @vatsalkaushik 
I'd like to work on this issue. Please assign to me.

sm17p:
@gumroadsteve2 Thank you for clarifying

slavingia:
Note that users can be not-reviewed before being probated, so auto-marking compliant is not the answer. We need to store the old state and revert to that specific state.

slavingia:
(covered in 11/18/2025 live stream)

EmCousin:
> for probating users but it's missing a corresponding controller action/UI on the admin side which feels like a regression to me.

This seems indeed like a regression occurred when the [Admin UI was removed](https://github.com/antiwork/gumroad-old/pull/28985/files#diff-51028abe52d6febc80f47fc7d55f07ce2dda0ffc216123c87cd7ba7e2cf515ecL92) in favor of Helper but then got [brought back](https://github.com/antiwork/gumroad/pull/169/files#diff-51028abe52d6febc80f47fc7d55f07ce2dda0ffc216123c87cd7ba7e2cf515ec). Maybe it was intentional to no longer allow admin users to put users on probation manually, but the `put_on_probation` endpoint still exists in `config/routes/admin.rb`. I think we should probably bring this action back, thought that can be done in a separate issue.

> Right now suspended accounts can be put on probation by `LowBalanceFraudCheck` (which is system issued) allowing them to sell their products on Gumroad

That indeed feels wrong. I think any users whose refunds were disabled for _any other reason_ that probation should probably not being put on probation by the system to begin with and stay in their current state (which would be suspended, flagged, etc.). This can probably be done in a separate issue as well.

@ershad _a Those are sensitive actions, what do you think? Assigning this issue to you for now.

shryesth:
Can I take this?

AlleyBo55:
pls check https://github.com/antiwork/gumroad/pull/2556 @slavingia

PR #2628, opened 2026-01-02
feat(risk): auto-remove LowBalanceFraudCheck probation when balance exceeds $100

FusionBrah:
Issue: #1884

# Description
## Problem
When sellers are automatically put on probation by LowBalanceFraudCheck due to suspicious refund activity, there's no automatic way to remove that probation when their balance recovers. This requires manual admin intervention even though the original trigger condition is no longer met.

## Solution
Automatically remove LowBalanceFraudCheck probation when the seller's unpaid balance exceeds $100, restoring them to their previous risk state using PaperTrail version history.

Key implementation details:
- Uses PaperTrail to retrieve the previous risk state before probation was applied
- Only auto-removes probation if the most recent probation was applied by LowBalanceFraudCheck (not manually by an admin)
- Guards against removing probation if an admin took any action after the auto-probation
- Skips auto-removal if previous state was `flagged_for_fraud` or `flagged_for_tos_violation` (admin should review these cases)
- Restores to `not_reviewed` or `compliant` based on PaperTrail history
- Adds `mark_not_reviewed!` state machine event for `on_probation → not_reviewed` transitions

---
# Before/After
Backend-only change

---
# Test Results
```
Finished in 6.13 seconds
33 examples, 0 failures
```

---
# Checklist

---
# AI Disclosure
- Implementing the feature based on PR #2618 feedback from EmCousin
- Writing comprehensive tests for all scenarios
- Adding PaperTrail integration for state history lookup
- Identifying and fixing edge case where flagged states couldn't be restored due to state machine constraints

### DIFF
--- a/app/models/concerns/user/low_balance_fraud_check.rb
+++ b/app/models/concerns/user/low_balance_fraud_check.rb
@@ -3,8 +3,11 @@
 module User::LowBalanceFraudCheck
   extend ActiveSupport::Concern
 
-  LOW_BALANCE_THRESHOLD = -100_00 # USD -100
+  LOW_BALANCE_THRESHOLD = -100_00 # USD -$100 (triggers probation)
   private_constant :LOW_BALANCE_THRESHOLD
+
+  HIGH_BALANCE_THRESHOLD = 100_00 # USD +$100 (removes probation)
+  private_constant :HIGH_BALANCE_THRESHOLD
 
   LOW_BALANCE_PROBATION_WAIT_TIME = 2.months
   private_constant :LOW_BALANCE_PROBATION_WAIT_TIME
@@ -30,6 +33,14 @@ module User::LowBalanceFraudCheck
     disable_refunds_and_put_on_probation! unless recently_probated_for_low_balance? || suspended?
   end
 
+  def check_for_high_balance_and_remove_low_balance_probation!
+    return if unpaid_balance_cents <= HIGH_BALANCE_THRESHOLD
+    return if !on_probation?
+    return if !should_auto_remove_probation?
+
+    revert_to_previous_risk_state!
+  end
+
   private
     def disable_refunds_and_put_on_probation!
       disable_refunds!
@@ -43,5 +54,53 @@ module User::LowBalanceFraudCheck
               .where(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME)
               .where("created_at > ?", LOW_BALANCE_PROBATION_WAIT_TIME.ago)
               .exists?
+    end
+
+    def should_auto_remove_probation?
+      low_balance_probation_comment = most_recent_low_balance_probation_comment
+      return false if low_balance_probation_comment.nil?
+
+      !has_newer_risk_state_transition?(low_balance_probation_comment)
+    end
+
+    def most_recent_low_balance_probation_comment
+      comments.with_type_on_probation
+              .where(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME)
+              .order(created_at: :desc)
+              .first
+    end
+
+    def has_newer_risk_state_transition?(probation_comment)
+      comments.where(comment_type: Comment::RISK_STATE_COMMENT_TYPES)
+              .where("id > ?", probation_comment.id)
+              .exists?
+    end
+
+    def revert_to_previous_risk_state!
+      previous_state = previous_risk_state_from_paper_trail
+      return if previous_state.in?(%w[flagged_for_fraud flagged_for_tos_violation])
+
+      content = "Probation removed automatically on #{Time.current.to_fs(:formatted_date_full_month)} because balance exceeded $100"
+
+      case previous_state
+      when "not_reviewed"
+        mark_not_reviewed!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+        enable_refunds!
+      else
+        mark_compliant!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+      end
+    end
+
+    def previous_risk_state_from_paper_trail
+      probation_version = versions.where("object_changes LIKE ?", "%user_risk_state%")
+                                  .where("object_changes LIKE ?", "%on_probation%")
+                                  .order(created_at: :desc)
+                                  .first
+      return "compliant" if probation_version.nil?
+
+      changeset = probation_version.changeset
+      return "compliant" if changeset.nil? || !changeset.key?("user_risk_state")
+
+      changeset["user_risk_state"].first || "compliant"
     end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -317,7 +317,7 @@ class User < ApplicationRecord
                       :do => :not_verified?
     after_transition any => %i[suspended_for_fraud suspended_for_tos_violation], :do => :invalidate_active_sessions!
     after_transition any => %i[suspended_for_fraud suspended_for_tos_violation], :do => :disable_links_and_tell_chat
-    after_transition any => %i[on_probation compliant flagged_for_tos_violation flagged_for_fraud suspended_for_tos_violation suspended_for_fraud],
+    after_transition any => %i[on_probation compliant not_reviewed flagged_for_tos_violation flagged_for_fraud suspended_for_tos_violation suspended_for_fraud],
                      :do => :add_user_comment
     after_transition any => [:flagged_for_tos_violation], :do => :add_product_comment
 
@@ -356,6 +356,10 @@ class User < ApplicationRecord
 
     event :put_on_probation do
       transition all => :on_probation
+    end
+
+    event :mark_not_reviewed do
+      transition on_probation: :not_reviewed
     end
   end
 
--- a/spec/models/concerns/user/low_balance_fraud_check_spec.rb
+++ b/spec/models/concerns/user/low_balance_fraud_check_spec.rb
@@ -39,6 +39,46 @@ describe User::LowBalanceFraudCheck do
     end
   end
 
+  describe "#mark_not_reviewed!" do
+    context "when user is on probation" do
+      before do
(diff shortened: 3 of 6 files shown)


app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -30,6 +33,14 @@ def check_for_low_balance_and_probate(refunded_or_disputed_purchase_id)
     disable_refunds_and_put_on_probation! unless recently_probated_for_low_balance? || suspended?
   end
 
+  def check_for_high_balance_and_remove_low_balance_probation!
+    return if unpaid_balance_cents <= HIGH_BALANCE_THRESHOLD
```
FusionBrah, line 37: Main entry point with guard clauses: only proceeds if balance > $100, user is on probation, AND we should auto-remove (i.e., probation was by LowBalanceFraudCheck with no subsequent admin action).

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,58 @@ def recently_probated_for_low_balance?
               .where("created_at > ?", LOW_BALANCE_PROBATION_WAIT_TIME.ago)
               .exists?
     end
+
+    def should_auto_remove_probation?
+      low_balance_probation_comment = most_recent_low_balance_probation_comment
```
FusionBrah, line 60: Two-part check per EmCousin's feedback: 1) Find the most recent LowBalanceFraudCheck probation comment, 2) Ensure no admin took action after that (no newer risk state transitions). Returns false if either condition fails.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,58 @@ def recently_probated_for_low_balance?
...
+      comments.with_type_on_probation
+              .where(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME)
+              .order(created_at: :desc)
+              .first
+    end
+
+    def has_newer_risk_state_transition?(probation_comment)
+      comments.where(comment_type: Comment::RISK_STATE_COMMENT_TYPES)
+              .where("id > ?", probation_comment.id)
+              .exists?
```
FusionBrah, line 76: Using ID comparison instead of timestamp because comments created in quick succession (same transaction) may have identical timestamps. IDs are always sequential and reliable for ordering.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,58 @@ def recently_probated_for_low_balance?
...
+    end
+
+    def revert_to_previous_risk_state!
+      previous_state = previous_risk_state_from_paper_trail
```
FusionBrah, line 80: Handles all risk states per EmCousin's feedback (ref #1884). Only `compliant` doesn't need `enable_refunds\!` because `mark_compliant\!` already triggers it via the after_transition callback. The `else` fallback to compliant is a safe default for any unexpected states.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,58 @@ def recently_probated_for_low_balance?
...
+      when "flagged_for_tos_violation"
+        flag_for_tos_violation!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+        enable_refunds!
+      else
+        mark_compliant!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+      end
+    end
+
+    def previous_risk_state_from_paper_trail
+      probation_version = versions.where("object_changes LIKE ?", "%user_risk_state%")
```
FusionBrah, line 95: Uses PaperTrail to retrieve the previous risk state per EmCousin's feedback. Queries the versions table for the most recent transition TO on_probation, then extracts the previous state from the changeset. Falls back to compliant if no version is found (safe default for legacy data).

app/models/user.rb
```diff
@@ -357,6 +357,10 @@ class User < ApplicationRecord
     event :put_on_probation do
       transition all => :on_probation
     end
+
+    event :mark_not_reviewed do
+      transition on_probation: :not_reviewed
```
FusionBrah, line 362: New state machine event to allow transitioning from on_probation back to not_reviewed. Intentionally restricted to only allow `on_probation → not_reviewed` to prevent misuse while enabling proper state restoration for LowBalanceFraudCheck auto-removal.

spec/models/concerns/user/low_balance_fraud_check_spec.rb
```diff
@@ -143,4 +183,151 @@
       end
     end
   end
+
+  describe "#check_for_high_balance_and_remove_low_balance_probation!", versioning: true do
+    context "when user is on LowBalanceFraudCheck probation" do
```
FusionBrah, line 188: `versioning: true` enables PaperTrail version tracking for these tests. By default, PaperTrail is disabled in the test environment for performance. This metadata is required because the feature uses PaperTrail to retrieve the previous risk state.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,52 @@ def recently_probated_for_low_balance?
...
+    def has_newer_risk_state_transition?(probation_comment)
+      comments.where(comment_type: Comment::RISK_STATE_COMMENT_TYPES)
+              .where("id > ?", probation_comment.id)
+              .exists?
+    end
+
+    def revert_to_previous_risk_state!
+      previous_state = previous_risk_state_from_paper_trail
+      return if previous_state.in?(%w[flagged_for_fraud flagged_for_tos_violation])
+
```
FusionBrah, line 82: If the user was flagged for fraud/TOS before being auto-probated, we skip auto-removal entirely. The state machine doesn't allow `on_probation → flagged_*` transitions, and more importantly, an admin flagged them for a reason - they should manually review rather than having the system auto-clear.

FusionBrah:
This PR addresses the feedback from #2618. Changes based on @EmCousin's review:

1. ✅ Added `!` to method name (`check_for_high_balance_and_remove_low_balance_probation!`)
2. ✅ Use PaperTrail instead of storing `previous_risk_state` in `json_data`
3. ✅ Fixed probation comment lookup - checks most recent LowBalanceFraudCheck comment AND ensures no newer admin actions
4. ✅ Handle all risk states per #1884 - skips auto-removal for flagged states (admin should review), restores to `not_reviewed` or `compliant`

Additional improvement discovered during implementation:
- Guard against restoring to `flagged_for_fraud`/`flagged_for_tos_violation` since the state machine doesn't allow `on_probation → flagged_*` transitions, and admin-flagged users should be manually reviewed anyway

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -30,6 +33,14 @@ def check_for_low_balance_and_probate(refunded_or_disputed_purchase_id)
     disable_refunds_and_put_on_probation! unless recently_probated_for_low_balance? || suspended?
   end
 
+  def check_for_high_balance_and_remove_low_balance_probation!
+    return if unpaid_balance_cents <= HIGH_BALANCE_THRESHOLD
+    return if !on_probation?
+    return if !should_auto_remove_probation?
```
EmCousin, line 39:
```suggestion
    return unless unpaid_balance_cents > HIGH_BALANCE_THRESHOLD
    return unless on_probation?
    return unless should_auto_remove_probation?
```

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,52 @@ def recently_probated_for_low_balance?
...
+      return if previous_state.in?(%w[flagged_for_fraud flagged_for_tos_violation])
+
+      content = "Probation removed automatically on #{Time.current.to_fs(:formatted_date_full_month)} because balance exceeded $100"
+
+      case previous_state
+      when "not_reviewed"
+        mark_not_reviewed!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+        enable_refunds!
+      else
+        mark_compliant!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
```
EmCousin, line 90: It's not safe to assume the user is compliant at this point. It would be safer to mark the user as compliant if the previous state was indeed compliant, and fall back to not reviewed.

app/models/concerns/user/low_balance_fraud_check.rb
EmCousin: object_changes is a JSON field. `JSON_CONTAINS` would be safer and more accurate.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,52 @@ def recently_probated_for_low_balance?
...
+      end
+    end
+
+    def previous_risk_state_from_paper_trail
+      probation_version = versions.where("object_changes LIKE ?", "%user_risk_state%")
+                                  .where("object_changes LIKE ?", "%on_probation%")
+                                  .order(created_at: :desc)
+                                  .first
+      return "compliant" if probation_version.nil?
```
EmCousin, line 99: Users are not compliant by default: they are not reviewed. So it should be "not_reviewed"

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,52 @@ def recently_probated_for_low_balance?
...
+
+      changeset = probation_version.changeset
+      return "compliant" if changeset.nil? || !changeset.key?("user_risk_state")
```
EmCousin, line 102: Same here

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,52 @@ def recently_probated_for_low_balance?
...
+
+      changeset["user_risk_state"].first || "compliant"
```
EmCousin, line 104: Same here

PR #2586, opened 2025-12-31
Inertia Migrate - Auth (Part 2)

stefanbinoj:
Part of #856 
Previous PR #2379 

This PR migrate following pages to inertia 
- /two-factor 
- /password/edit
- Also has a new /forgot_password route
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>

  
  <tr>
    <td>
      <strong>/two-factor</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/89ed4f86-c62d-4b8b-a535-dd21c3e43418"
             controls width="100%">
      </video>
    </td>
    <td>
      <strong>/two-factor</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/25d4c849-9c08-4e9e-99f3-f01bd27c9217"
             controls width="100%">
      </video>
    </td>
  </tr>

  
  <tr>
    <td>
      <strong>/password/edit</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/abf5d8fd-427d-49e8-822e-cad3469df4e0"
             controls width="100%">
      </video>
    </td>
    <td>
      <strong>/password/edit</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/a89dbc4b-5540-445f-8f81-6426775b15fe"
             controls width="100%">
      </video>
    </td>
  </tr>

  
  <tr>
    <td>
      <strong>Forgot Password<br/>(No page, using login state)</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/5dca1f4e-bbf2-4bd8-9b9e-1b1c59e53215"
             controls width="100%">
      </video>
    </td>
    <td>
      <strong>/forgot_password (New)</strong><br/><br/>
      <video src="https://github.com/user-attachments/assets/32c50a04-bb73-4d2a-bee9-37077d81100e"
             controls width="100%">
      </video>
    </td>
  </tr>
</table>

## Tests
| Spec File | Status |
|-----------|--------|
| spec/controllers/two_factor_authentication_controller_spec.rb | ✅ 27 examples, 0 failures |
| spec/controllers/user/passwords_controller_spec.rb | ✅ 10 examples |
| spec/requests/two_factor_authentication_request_spec.rb | ✅ 2 |
| spec/requests/password_reset_spec.rb | ✅ 2 |
| spec/requests/two_factor_authentication_spec.rb | ✅ 8 examples, 0 failures |

Screenshot 2025-12-17 at 11 20 08 AM

[...]

PR #2629, opened 2026-01-02
Inertia Migrate - Auth (Part 3)

stefanbinoj:
Issue: #856 

# Description
- Add a new `delete` endpoint for `/logout` which can be used by `app/javascript/components/client-components/Nav/index.tsx` for SPA like behaviour
- Server Nav component (`app/javascript/components/server-components/Nav.tsx`) still uses old `get` route for `/logout` with no change. 

### Before

https://github.com/user-attachments/assets/ade4bee1-5583-4367-aab8-1d3e20502732

---

# Test Results
Screenshot 2026-01-02 at 3 36 53 PM

---

# Checklist

---

# AI Disclosure

- Used opus4.5 with opencode for initial draft of this PR.

### DIFF
--- a/spec/requests/two_factor_authentication_spec.rb
+++ b/spec/requests/two_factor_authentication_spec.rb
@@ -50,6 +50,7 @@ describe "Two-Factor Authentication", js: true, type: :system do
 
         first("nav[aria-label='Main'] details summary").click
         click_on "Logout"
+        expect(page).to have_content("Log in")
         login_to_app
 
         # It doesn't ask for 2FA again


app/javascript/components/AuthAlert.tsx
```diff
@@ -1,7 +1,7 @@
 import { usePage } from "@inertiajs/react";
```
stefanbinoj, line 1: This component is used because once we are redirected to login_path (/login) 
it doesn't have `inertia/layout` or other layouts and alert component hence this is a custom/specific component for showing alerts in /login and related auth pages
[Pls read](https://github.com/antiwork/gumroad/pull/2586#discussion_r2655491942)

app/controllers/two_factor_authentication_controller.rb
```diff
@@ -30,7 +30,7 @@ def resend_authentication_token
 
   private
     def redirect_to_login_path
-      redirect_to login_url(next: request.fullpath)
+      redirect_to login_path(next: request.fullpath)
```
stefanbinoj, line 33: (Small Nit for relative path names)

app/controllers/logins_controller.rb
```diff
@@ -52,6 +56,12 @@ def create
     end
   end
 
+  def destroy
+    super do
+      return redirect_to login_path, notice: "Signed out successfully.", status: :see_other
```
stefanbinoj: logout action with 303 redirect

app/controllers/logins_controller.rb
```diff
@@ -12,7 +12,11 @@ class LoginsController < Devise::SessionsController
   layout "inertia", only: [:new]
 
   def new
-    return redirect_to login_path(next: request.referrer) if params[:next].blank? && request_referrer_is_a_valid_after_login_path?
+    if params[:next].blank? && request_referrer_is_a_valid_after_login_path?
+      uri = URI.parse(request.referrer)
+      next_path = uri.host ? uri.request_uri : request.referrer
+      return redirect_to login_path(next: next_path)
+    end
```
stefanbinoj: (Small Nit it helps in avoid unecessary host and protocol present in `/login?next=` while logging out )

### Before : 

https://github.com/user-attachments/assets/3645b8a0-6525-4c1c-be2b-767d038ddd68

### After : 

https://github.com/user-attachments/assets/0c94c4bc-033a-4221-8dd1-54f582aef74a

stefanbinoj:
Hi @Pradumn27, could you take a look at this whenever you are free? Thanks!

Pradumn27:
@Pradumn27 _a, will review as the assigned reviewer for the issue

config/routes.rb
```diff
@@ -341,6 +341,7 @@ def product_info_and_purchase_routes(named_routes: true)
 
       post "login", to: "logins#create"
       get "logout", to: "logins#destroy" # TODO: change the method to DELETE to conform to REST
+      delete "logout", to: "logins#destroy"
```
Pradumn27: why keep both? `delete` seems like the right choice we can remove the `get` one - test it's usages still

app/javascript/components/client-components/Nav/footer.tsx
```diff
@@ -36,7 +38,15 @@ function NavbarFooter() {
...
-          <NavLinkDropdownItem text="Logout" icon="box-arrow-in-right-fill" href={Routes.logout_url(routeParams)} />
+          <Link
+            role="menuitem"
+            href={Routes.logout_url(routeParams)}
+            method="delete"
+            className="block truncate border-0 px-4 py-2 no-underline"
+          >
+            <Icon name="box-arrow-in-right-fill" className="mr-3 ml-1" />
+            Logout
+          </Link>
```
Pradumn27: The Link component is not taking the entire div space as seen in the video, please fix that

app/javascript/components/AuthAlert.tsx
Pradumn27: makes sense, lets rename the component to `AuthAlert`

app/controllers/logins_controller.rb
Pradumn27: We don't need to do this. The issue occurred due to overriding the destroy method of the Devise session handler, devise handles the redirect URL sanitization internally

app/controllers/logins_controller.rb
Pradumn27: I understand the motivation for overriding here - the goal is to ensure we return a `303 See Other,` which is the correct status code for Inertia after non-GET requests.

However, overriding Devise’s `destroy` method in this way is risky. Devise’s implementation relies on `yield`, and returning from the yielded block causes the rest of the method to never execute, which in this particular case was the function `respond_to_on_destroy` . This unintentionally bypasses part of Devise’s response handling and can affect more behavior than just the redirect status.

A safer approach, in my opinion, would be to handle this globally rather than overriding individual service behaviors. We could normalize redirect status codes for Inertia at the response layer, likely in `ApplicationController`:

```
after_action :coerce_303_for_inertia

def coerce_303_for_inertia
  return unless request.inertia?
  return unless response.redirect?
  return if request.get?

  response.status = :see_other
end
```

This ensures Inertia always receives the correct status code while preserving the default behavior of Devise (and other services). It also avoids tightly coupling authentication logic to SPA-specific concerns.

Curious to hear your thoughts on this approach as well @EmCousin @binary-koan

spec/requests/two_factor_authentication_spec.rb
```diff
@@ -48,8 +48,8 @@ def submit_login_form
 
         fill_in_profile
 
-        first("nav[aria-label='Main'] details summary").click
-        click_on "Logout"
```
Pradumn27, line 52: no, we need to test clicking on the logout button here

Pradumn27 (commented):
Thanks for the PR @stefanbinoj _a, please address the comments.

config/routes.rb
stefanbinoj: I totally forgot to add comments about this earlier -- `get "logout"` this route is still used by `server-components/Nav.tsx` (Which is present in /help page as shown in the video. Also `/product/:id/edit` ). Also admin side `server-components/Admin/Nav.tsx` still uses this endpoint for logging out. IMO it would be better to remove this endpoint once these server-side Nav components are removed from codebase. wdyt?

Pradumn27: In that case let's not change the route here then, as it will cause confusion, we can do it in a separate PR as a whole, here use the same route - just in the Link specify the method as delete, how we do it in admin footer now

stefanbinoj: But if we remove `delete "logout", to: "logins#destroy"` from `routes.rb` and keep Link with method="delete" it still triggers a full page reload

### Without `  delete "logout", to: "logins#destroy"`
(Full page reload)


https://github.com/user-attachments/assets/cea82c12-9501-43a0-a424-8a1c4c655d90

### With `delete "logout", to: "logins#destroy"`
(No reload)



https://github.com/user-attachments/assets/7e3b5afd-34f1-4305-80b3-f450717e0272

stefanbinoj:
Hi @Pradumn27, thanks for the review it was very helpful 🙏🏻 . I’ve addressed the comments (also added `after_action :coerce_303_for_inertia`) . When you have some time, could you please take another look?

config/routes.rb
stefanbinoj: In admin side with or without this route it doesn't matter it still triggers a full page realod
For it to work we will have to add route [here](https://github.com/antiwork/gumroad/blob/main/app/javascript/packs/admin_inertia.ts#L53) And pass CSRF token too in request for performing destroy action as we do in `packs/inertia.js` [here](https://github.com/antiwork/gumroad/blob/main/app/javascript/packs/inertia.js#L10)

https://github.com/user-attachments/assets/6f7297f4-fab5-4c7e-aac0-a306f8592911

Pradumn27: Alright, if we are keeping both, then please add a comment explaining why - we can remove it once the route is correctly used everywhere

app/javascript/components/AuthAlert.tsx
Pradumn27, line 1: just `AuthAlert` is fine - we're already in the components directory so adding the Component suffix feels redundant.

Pradumn27:
@binary-koan _a, can you please take a look here and run the CI, thanks!

app/controllers/logins_controller.rb
EmCousin: No, that's too sharp. Adding a callback in Application will affect every Inertia endpoint, and we do not always want redirects to be 303. We want to keep 302 for redirections upon failures, sometimes we'll want 301 to permanently redirect a deprecated endpoint.

You could just override [respond_to_on_destroy](https://github.com/heartcombo/devise/blob/main/app/controllers/devise/sessions_controller.rb#L75-L82), or temporarily override `Devise.responder.redirect_status` to `:see_other`

EmCousin:
@binary-koan I added a comment https://github.com/antiwork/gumroad/pull/2629#discussion_r2665227539

app/controllers/logins_controller.rb
Pradumn27: Makes sense - agreed that a global callback is too broad. I had considered conditionally redirecting for other statuses (if specified), but that still feels too wide in scope.

We can override [respond_to_on_destroy](https://github.com/heartcombo/devise/blob/main/app/controllers/devise/sessions_controller.rb#L75-L82), but the default implementation performs important work around redirect location normalization (handling stored locations, host sanitization, etc).

Temporarily overriding Devise.responder.redirect_status to :see_other for Inertia requests seems like the right, scoped approach here.

app/controllers/logins_controller.rb
```diff
@@ -53,6 +53,14 @@ def create
...
 
   private
+    def respond_to_on_destroy(non_navigational_status: :no_content)
+      redirect_status = request.inertia? ? :see_other : Devise.responder.redirect_status
+      respond_to do |format|
+        format.all { head non_navigational_status }
+        format.any(*navigational_formats) { redirect_to after_sign_out_path_for(resource_name), status: redirect_status }
+      end
+    end
+
```
stefanbinoj: I have overrided the existing devise controller [respond_to_on_destroy](https://github.com/heartcombo/devise/blob/main/app/controllers/devise/sessions_controller.rb#L75-L82J)

**Changes** : 
Added `redirect_status` for dynamically determining status_code


https://github.com/user-attachments/assets/0d930a1d-50a6-4f5f-86dd-6c6863c0752a

stefanbinoj:
@Pradumn27 could you take a quick look at this https://github.com/antiwork/gumroad/pull/2629#discussion_r2666907383? Thanks!

app/controllers/logins_controller.rb
Pradumn27: Hey @stefanbinoj _a , I just checked out your branch and tested it, even without the devise override I am getting the status code as 303, so can you check whether we even need this or not?

stefanbinoj: You are right! Sorry for not checking this previously my bad! Thanks for catching this 🙏🏻 . 
Will remove the override now.

stefanbinoj:
@EmCousin @binary-koan could you take a look at this? 

Devise logout sets status 303 by default: 

https://github.com/user-attachments/assets/72022499-37dd-431e-a484-24a7407ba7df

app/controllers/application_controller.rb
```diff
@@ -36,7 +36,6 @@ class ApplicationController < ActionController::Base
   before_action :set_paper_trail_whodunnit
   before_action :set_recommender_model_name
   before_action :track_utm_link_visit
-
```
Pradumn27, line 39: revert

spec/requests/two_factor_authentication_spec.rb
```diff
@@ -50,6 +50,7 @@ def submit_login_form
 
         first("nav[aria-label='Main'] details summary").click
         click_on "Logout"
+        expect(page).to have_content("Log in")
```
Pradumn27, line 53: why is this added? please always add self review for things not understandable via code

Issue #2619, opened 2026-01-02
`Unexpected Application Error!` on product's content page


vatsalkaushik:
Product ID `XqmlbQubOBcBGRD3sb3toQ==`

Opening the product's `/edit/content` page shows the following error: 

```
Unexpected Application Error!
Expected root to be a string, got null
Error: Expected root to be a string, got null
    at s (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:1043489)
    at e (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:1043716)
    at l (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:1045297)
    at u (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:1045364)
    at Y (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:131785)
    at gi (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:778167)
    at Al (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:837136)
    at vc (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:826384)
    at yc (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:826312)
    at gc (https://assets.gumroad.com/packs/js/webpack-commons-b471bbcbadd993aecb18.js:2:826175)
```

<img width="3600" height="1609" alt="Image" src="https://github.com/user-attachments/assets/d0b47644-12ad-4177-9c7e-4624ec6c2969" />

vatsalkaushik:
@nyomanjyotisa _a can you please help check?

PR #2632, opened 2026-01-02
fix: handle fileEmbed nodes with null id/uid

nyomanjyotisa:
Issue: https://github.com/antiwork/gumroad/issues/2619

# Description

This is a quick fix for the issue, to handle null id/uid.

We will still need to identify the exact reproduction steps that allow null id and uid to be persisted, then fix the root cause

## Problem
Product ID `XqmlbQubOBcBGRD3sb3toQ==` contain `fileEmbed` nodes in their `rich_content` with null `id` and `uid` values, causing rendering errors on the product edit page
Image

## Solution

Add an early return in `FileEmbedNodeView` component to handle `fileEmbed` nodes with null `id` values

---

# Before/After

I manually add `fileEmbed` nodes with null `id` and `uid` values using rails console for test:

## Before
image

## After
image

---

# AI Disclosure


nyomanjyotisa:
@EmCousin _a Can you help to review this?

EmCousin (approved):

PR #2634, opened 2026-01-02
feat: automatically re-instate probated accounts to `not_reviewed/compliant` risk state on a healthy balance

sm17p:
Issue: #1884

Re-opening  #2072

## Problem

Currently, when a user's balance drops below $100, they are automatically put on probation and refunds are disabled.

However, when their balance recovers above $100, they remain on probation indefinitely until manually marked compliant by an admin.

## Solution

We automatically mark probated accounts due to low balance fraud check as `not_reviewed/compiant` when their balance recovers to `$100`.

1. Using existing job (`UpdateSellerRefundEligibilityJob`) on balance updates for recovering `risk/payout `state. 
    - This point is specifically chosen to perform the checks as it blends in with the existing system load w.r.t balance updates.
    - Logic for enabling refund state was left untouched, as it's already handled in `UpdateSellerRefundEligibilityJob`
2. Using `PaperTrail` records to find valid previous recovery state. 
    - Only previously `not_reviewed/compiant` state will be recovered
    - Recovering to any other state will not be possible, and instead will be logged and notified to the team via `Bugsnag`. This will help us identify if we need to handle such cases in the future.

---

# After

## Comment Preview

Comments-preview

### User goes from `not_reviewed` -> `probated` (LowBalanceFraudCheck) -> `not_reviewed`

https://github.com/user-attachments/assets/5603e535-b9bf-4c8d-a889-7c670e266464

### User goes from `not_reviewed` -> `compliant` -> `probated` (LowBalanceFraudCheck) -> `compliant`

https://github.com/user-attachments/assets/7dd2f8c5-0a64-4539-9db2-e636aa9db5cb

---

# Test Results

specs_passing

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure
Cursor Auto Mode for edits to code after #2072

PR #2644, opened 2026-01-02
fix(risk): address review feedback for auto-remove probation feature

FusionBrah:
Issue: #1884

# Description

## Problem

PR #2628 implemented auto-removal of LowBalanceFraudCheck probation but had several issues identified in code review by @EmCousin:
- Inconsistent guard clause style (`return if !` instead of `return unless`)
- Using `LIKE` for JSON field queries instead of `JSON_CONTAINS`
- Incorrectly defaulting to `compliant` state when previous state was unknown (users are `not_reviewed` by default)

## Solution

Addressed all review feedback:
1. Use `return unless` guard clause pattern for consistency
2. Use `JSON_CONTAINS` for safer PaperTrail `object_changes` queries
3. Default to `not_reviewed` (not `compliant`) when previous state is unknown
4. Only mark `compliant` if the previous state was actually `compliant`, otherwise fallback to `not_reviewed`

---

# Before/After

Backend-only change - no UI modifications.

---

# Test Results

Screenshot 2026-01-03 at 10 33 54 am

---

# Checklist

---

# AI Disclosure

- Reviewing PR #2628 comments from EmCousin
- Implementing all requested changes
- Updating test expectations to match new default behavior


app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -30,6 +33,14 @@ def check_for_low_balance_and_probate(refunded_or_disputed_purchase_id)
     disable_refunds_and_put_on_probation! unless recently_probated_for_low_balance? || suspended?
   end
 
+  def check_for_high_balance_and_remove_low_balance_probation!
+    return unless unpaid_balance_cents > HIGH_BALANCE_THRESHOLD
```
FusionBrah, line 37: Changed to `return unless` pattern per review feedback. This is more idiomatic Ruby and consistent with the codebase style.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,51 @@ def recently_probated_for_low_balance?
...
+    end
+
+    def revert_to_previous_risk_state!
+      previous_state = previous_risk_state_from_paper_trail
+      return if previous_state.in?(%w[flagged_for_fraud flagged_for_tos_violation])
+
+      content = "Probation removed automatically on #{Time.current.to_fs(:formatted_date_full_month)} because balance exceeded $100"
+
+      case previous_state
+      when "compliant"
```
FusionBrah, line 86: Swapped the case logic: now only marks `compliant` when previous state was explicitly compliant, otherwise defaults to `not_reviewed`. This is safer because `not_reviewed` is the default state for new users.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,51 @@ def recently_probated_for_low_balance?
...
+        mark_compliant!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+      else
+        mark_not_reviewed!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+        enable_refunds!
+      end
+    end
+
+    def previous_risk_state_from_paper_trail
+      probation_version = versions.where("JSON_CONTAINS(object_changes, ?)", '"on_probation"')
```
FusionBrah: Using `JSON_CONTAINS` instead of `LIKE` for safer and more accurate JSON field queries. This correctly handles the JSON structure of PaperTrail's `object_changes` column.

FusionBrah:
This PR addresses the feedback from #2628. Changes based on @EmCousin's review:

1. ✅ Use `return unless` guard clause pattern instead of `return if !` for consistency
2. ✅ Use `JSON_CONTAINS` for PaperTrail `object_changes` queries instead of `LIKE` - safer and more accurate for JSON fields
3. ✅ Default to `not_reviewed` instead of `compliant` when previous state is unknown - since `not_reviewed` is the default state for new users
4. ✅ Only mark `compliant` if the previous state was explicitly `compliant`, otherwise fallback to `not_reviewed`

Updated the test to expect `not_reviewed` when no probation version is found, matching the new default behavior.

app/models/concerns/user/low_balance_fraud_check.rb
```diff
@@ -44,4 +55,51 @@ def recently_probated_for_low_balance?
...
+      when "compliant"
+        mark_compliant!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+      else
+        mark_not_reviewed!(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
+        enable_refunds!
+      end
+    end
+
+    def previous_risk_state_from_paper_trail
+      probation_version = versions.where("JSON_CONTAINS(object_changes, ?, '$.user_risk_state')", '"on_probation"')
```
FusionBrah, line 95: Added JSON path `'$.user_risk_state'` to correctly search within the array. PaperTrail stores changes as `{"user_risk_state": ["old_value", "on_probation"]}`, so we need to specify the path to search within that array, not just the top-level object.

FusionBrah:
@EmCousin Wondering if I could get you to review this?

app/modules/user/risk.rb
```diff
@@ -192,7 +194,7 @@ def add_user_comment(transition)
                 transition.to_name.to_s.humanize
     end
     comment_type = case transition.to_name
-                   when :compliant
+                   when :compliant, :not_reviewed
                      Comment::COMMENT_TYPE_COMPLIANT
```
EmCousin: That's not correct

EmCousin:
Thanks @FusionBrah. Your PR still has gaps that prevent a merge as-is, gap which is filled by  @sm17p in https://github.com/antiwork/gumroad/pull/2634, and vice-versa.
I merged both of your efforts in https://github.com/antiwork/gumroad/pull/2758 and aim to split the bounty 50/50 between you too!

PR #2681, opened 2026-01-05
Fix (app): TestimonialSelectModal imports

brentguf:
Issue: #2042

# Description

## Problem

The “insert review modal” is not shown most of the time, yet `TestimonialSelectModal` is rendered twice (via the `RichTextEditorToolbar` and `ContentTab`) on every content tab render. Not a huge deal on initial load, but it becomes problematic when it starts firing on each keystroke/file upload. 

## Solution

Removing the duplication and lazy-loading the modal.

Bonus: because of the reduced imports, it is not part of the `webpack-commons.js` main bundle anymore, reducing the amount of JS on every page.

---

# Before/After

Modal render count when editing the product and opening the modal: 100 vs. 6.

Similar results when uploading files.

## Before

https://github.com/user-attachments/assets/bbf713a2-0326-4c89-a290-c5b377c1060f

## After

https://github.com/user-attachments/assets/9ad144db-e055-4332-af23-f4ec33943b3b

---

# Test Results

---

# Checklist

---

# AI Disclosure

Claude Opus 4.5 (Thinking) was used for the initial refactoring. Manual adjustments afterwards.

### DIFF
--- a/app/javascript/components/ProductEdit/ContentTab/index.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/index.tsx
@@ -717,7 +717,6 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string | 
                             <Icon name="cart-plus" />
                             <span>Upsell</span>
                           </div>
-
                           <InsertReviewMenuItem editor={editor} productId={id} />
                         </>
                       )}


app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -723,16 +714,7 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
                           <Icon name="cart-plus" />
                           <span>Upsell</span>
                         </div>
-                        <div
-                          role="menuitem"
-                          onClick={(e) => {
-                            e.stopPropagation();
```
brentguf: `e.stopPropagation();` caused the Insert dropdown menu not to close on opening the modal (only after closing the modal).

In the place where the duplicate lived (RichTextEditor.tsx), there was no such code.

I've done some manual testing opening/closing the dropdown and modal without this line and it seems to work fine.

app/javascript/components/InsertReviewMenuItem.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Editor } from "@tiptap/react";
+import * as React from "react";
+
+import { Icon } from "$app/components/Icons";
+
+const TestimonialSelectModal = React.lazy(() => import("$app/components/TestimonialSelectModal"));
+
+export const InsertReviewMenuItem = ({ editor, productId }: { editor: Editor; productId: string }) => {
+  const [showReviewModal, setShowReviewModal] = React.useState(false);
```
brentguf, line 8: Putting the "modal show state" together with the component that opens it is a recommended pattern: upon opening the modal, surrounding components won't re-render. Only the code inside this component will.

app/javascript/components/InsertReviewMenuItem.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Editor } from "@tiptap/react";
+import * as React from "react";
+
+import { Icon } from "$app/components/Icons";
+
+const TestimonialSelectModal = React.lazy(() => import("$app/components/TestimonialSelectModal"));
```
brentguf: Only import the modal when it's requested (also see `Suspense` below).

app/javascript/components/TestimonialSelectModal.tsx
```diff
@@ -154,6 +155,8 @@ export const TestimonialSelectModal = ({
   );
 };
 
+export default TestimonialSelectModal;
```
brentguf: A default export is most convenient when lazy-loading. I could keep the named export, and adjust the lazy import, if that's preferred though.

app/javascript/components/TestimonialSelectModal.tsx
```diff
@@ -96,6 +96,7 @@ export const TestimonialSelectModal = ({
       open={isOpen}
       onClose={onClose}
       title="Insert reviews"
+      usePortal
```
brentguf, line 99: Since the modal moved inside the menu item, a portal was needed to ensure the dark modal background shows (i.e. the modal renders at the document body level).

app/javascript/components/InsertReviewMenuItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+      editor.chain().focus().insertReviewCard({ reviewId }).run();
+    }
+    setShowReviewModal(false);
+  };
+
+  return (
+    <>
+      <div role="menuitem" onClick={() => setShowReviewModal(true)}>
+        <Icon name="solid-star" />
+        <span>Review</span>
```
brentguf, line 21: This item was labelled `Review` in one place, `Reviews` in the other. Let me know if you prefer `Reviews` instead.

app/javascript/components/InsertReviewMenuItem.tsx
brentguf, line 1: I would prefer to put this component elsewhere, but it's used in two places. 

I think it makes sense to extract `RichTextEditorToolbar` into a folder/component eventually (perhaps colocated with or inside a `RichTextEditor` folder), and then it could go in there.

brentguf:
@binary-koan Any chance you could take a look at this PR and [my latest comment on the related issue](https://github.com/antiwork/gumroad/issues/2042#issuecomment-3729215877)? Should we continue work on that issue or deprioritize for now?

slavingia:
Please resolve merge conflicts

app/javascript/components/InsertReviewMenuItem.tsx
binary-koan: The component doesn't look particularly big, is there much advantage to lazy loading? It may also cause a perceptible delay on slow connections if the bundle is split in production

brentguf: This particular modal on its own won't make a world of difference in terms of performance.

I've extracted this modal as an example to demonstrate the approach. (I did this as part of our conversation on the issue page.) There are many more components that could benefit from similar extraction, but I wanted to get alignment on the approach before proceeding further. 

In general, I've noticed the components in Gumroad's codebase tend to be larger than what's typically considered "best practice", which I believe is contributing to the performance issues mentioned in the linked issue.

brentguf:
@binary-koan Let me know if you still want to merge this PR, and I'll fix the merge conflicts. If not, feel free to close it!

binary-koan:
I don't think it needs the `lazy` (in general I'd rather have the editor take slightly longer to load but be snappy once it does), but definitely keen to merge the state refactor + conditional rendering if you don't mind updating this

brentguf:
@binary-koan 
> I don't think it needs the lazy (in general I'd rather have the editor take slightly longer to load but be snappy once it does), 

I can see your point. I mainly suggested lazy-loading components like this as an instant relief (reducing operations per render given the frequent global state updates).

> but definitely keen to merge the state refactor + conditional rendering if you don't mind updating this

Made the changes! FYI, I didn't do a functional test anymore after the changes.

slavingia:
> didn't do a functional test anymore after the changes.

We should always test?

brentguf:
> We should always test?

@slavingia Indeed, normally I would absolutely do that. 

However, given the issue remained unclear after my [last update on Jan 9](https://github.com/antiwork/gumroad/issues/2042#issuecomment-3729215877), I had already torn down my setup. Note I have spent a week last month getting familiar with the codebase and doing performance audits for this bounty, which now got deprioritized. I also [provided a few other fixes](https://github.com/antiwork/gumroad/pulls?q=is%3Apr+author%3Abrentguf+is%3Aclosed). All of this work was unpaid.

I still made the change Jono requested, but I didn't want to spend even more time and run the setup again, so I figured the change here is small enough that someone with an active setup could verify it in 2 minutes. I could have been clearer about that and understand if that doesn't work for you.

> but definitely keen to merge the state refactor + conditional rendering if you don't mind updating this

@binary-koan 

[This suggestion from a different PR](https://github.com/antiwork/gumroad/pull/3505/changes#diff-c500f1227c9dabd5a4868fb352aa6819c84ab9e84659a410b49ea980673300a3R974-R976) made me wonder whether I actually understood you well. I’m not sure what you meant with “conditional rendering.”

Given the change in priorities, the remaining uncertainty and likely additional work, I'll close this PR. Thanks for the collaboration!

PR #2688, opened 2026-01-05
Fix (app): lodash import

brentguf:
# Description

## Problem

The old import pulled in the full `lodash` library.

## Solution

Only import the function we're using.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -209,7 +209,7 @@ jobs:
             -e KNAPSACK_PRO_TEST_FILE_PATTERN="spec/**/*_spec.rb" \
             -e KNAPSACK_PRO_TEST_FILE_EXCLUDE_PATTERN="spec/requests/**/*_spec.rb" \
             -e KNAPSACK_PRO_COMMIT_HASH=${{ github.sha }} \
-            -e KNAPSACK_PRO_BRANCH=${{ github.ref_name }} \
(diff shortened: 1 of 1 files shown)


brentguf:
@binary-koan Even better would be to use `lodash-es` instead: this would prevent accidental mistakes like this and provide better developer experience via auto-imports. It would also reduce total bundle size by a few %. Let me know if you'd like me to put up a PR for that!

binary-koan (approved):
Nice catch! Yeah would be great to switch to `lodash-es`

.github/workflows/tests.yml
```diff
@@ -209,7 +209,7 @@ jobs:
             -e KNAPSACK_PRO_TEST_FILE_PATTERN="spec/**/*_spec.rb" \
             -e KNAPSACK_PRO_TEST_FILE_EXCLUDE_PATTERN="spec/requests/**/*_spec.rb" \
             -e KNAPSACK_PRO_COMMIT_HASH=${{ github.sha }} \
-            -e KNAPSACK_PRO_BRANCH=${{ github.ref_name }} \
+            -e KNAPSACK_PRO_BRANCH="${{ github.ref_name }}" \
```
binary-koan, line 212: @ershad Minor change FYI - CI was erroring on this PR because the parentheses in the branch name were being interpreted as shell syntax

PR #2666, opened 2026-01-04
Fix: Run EasyPost tests without real secrets

abhi152003:
ref: #959

- Add EASYPOST_API_KEY placeholder to .env.test
- Add VCR before_record hook to sanitize Basic Auth headers
- Sanitize 11 VCR cassettes to use placeholder credentials

Issue: #959

# Description

## Problem

The ShipmentsController tests fail when running without Gumroad's internal EasyPost API credentials. The VCR cassettes contain hardcoded base64-encoded API keys in the `Authorization` headers, which aren't filtered by the existing `filter_sensitive_data` configuration.

## Solution

1. Added `EASYPOST_API_KEY=test-easypost-api-key` placeholder to [.env.test](cci:7://file:///home/abhip05/Documents/ideas/gumroad-fork/.env.test:0:0-0:0)
2. Added a `before_record` hook to VCR configuration that sanitizes Basic Auth headers (replacing them with `<EASYPOST_API_KEY_BASE64>`)
3. Sanitized 11 existing VCR cassettes to replace hardcoded credentials with the placeholder

This allows tests to run without real EasyPost credentials by replaying sanitized VCR cassettes.

---

# Before/After

N/A - No UI changes

---

# Test Results

All 11 EasyPost/VCR-related tests pass:

image

**Note**: 5 mark_as_shipped tests fail due to a pre-existing factory issue (named_seller uses hardcoded values causing uniqueness conflicts). These failures exist on the main branch and are unrelated to this PR.

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Opus 4.5 (Antigravity) was used to assist with:
- Identifying the root cause of test failures
- Writing the VCR `before_record` hook
- Generating the sed command to sanitize cassettes

PR #2690, opened 2026-01-05
Fix(test): adds PAYPAL_PARTNER_CLIENT_ID to .env.test 

romitg2:
Issue: #959 

# Description

When tests render admin components with `prerender: true`, React components are rendered on the **server-side** (in Ruby/Node.js) before being sent to the client.

### Root Cause: 

- CheckoutPresenter Props
**Location**: `app/presenters/checkout_presenter.rb:282`
```ruby
def checkout_common
  {
    discover_url: discover_url(protocol: PROTOCOL, host: DISCOVER_DOMAIN),
    countries: Compliance::Countries.for_select.to_h,
    us_states: STATES,
    ca_provinces: Compliance::Countries.subdivisions_for_select(Compliance::Countries::CAN.alpha2).map(&:first),
    recaptcha_key: GlobalConfig.get("RECAPTCHA_MONEY_SITE_KEY"),
    paypal_client_id: PAYPAL_PARTNER_CLIENT_ID,  # ← Uses the constant
  }
end
```

If `PAYPAL_PARTNER_CLIENT_ID` is `nil`:
- The constant is set to `nil` during Rails initialization
- When pre-rendering happens, components receive `paypal_client_id: nil`
- JavaScript code might try to use this value, causing errors
- Tests fail during server-side rendering

```
Test starts (non_tiered_membership_spec.rb:86)
  ↓
Rails initializes
  ↓
config/initializers/paypal.rb runs
  ↓
PAYPAL_PARTNER_CLIENT_ID = GlobalConfig.get("PAYPAL_PARTNER_CLIENT_ID")
  ↓
Checks ENV["PAYPAL_PARTNER_CLIENT_ID"] → Not found in .env.test
  ↓
Checks Rails credentials → Not found (or no test credentials)
  ↓
PAYPAL_PARTNER_CLIENT_ID = nil
  ↓
Test renders subscription update page with prerender: true
  ↓
CheckoutPresenter builds props with paypal_client_id: nil
  ↓
React component tries to use nil value during server rendering
  ↓
ERROR: Cannot read property of null/undefined (ts-safe-cast/index.ts:13:10)
  ↓
Test fails


```
## Solution
Add the environment variable to `.env.test`:
```bash
# .env.test
PAYPAL_PARTNER_CLIENT_ID=test-paypal-partner-client-id
```
This ensures:
1. During test initialization, `GlobalConfig.get()` finds the value in `ENV`
2. The constant is set to a valid string instead of `nil`
3. Pre-rendered components receive valid props
4. Tests pass

---

# Tests Result: 

# Before:
### local: 

Screenshot 2026-01-03 at 2 47 13 PM

# After:
### local: 
Screenshot 2026-01-03 at 2 47 41 PM

---

# Checklist

---

# AI Disclosure

used cursor with claude-opus-4.5 for debugging purpose.


.env.test
```diff
@@ -25,6 +25,7 @@ OBFUSCATE_IDS_CIPHER_KEY=test-cipher-key
 IFFY_TOKEN=test-iffy-token
 IFFY_WEBHOOK_SECRET=test-iffy-webhook-secret
 PAYPAL_PARTNER_MERCHANT_ID=TESTPAYPALID
+PAYPAL_PARTNER_CLIENT_ID=test-paypal-partner-client-id
```
romitg2, line 28: When `PAYPAL_PARTNER_CLIENT_ID` is `nil`, React components fail during pre-rendering because JavaScript code expects it to be non-null (string) causing the test to crash.

Image

Image

romitg2:
@gianfrancopiana not sure why, but tests became flaky on main post merge of https://github.com/antiwork/gumroad/pull/2666 (https://github.com/antiwork/gumroad/pull/2666#discussion_r2662271907)
https://github.com/antiwork/gumroad/actions?query=branch%3Amain

romitg2:
> @gianfrancopiana not sure why, but tests became flaky on main post merge of #2666 ([#2666 (comment)](https://github.com/antiwork/gumroad/pull/2666#discussion_r2662271907)) https://github.com/antiwork/gumroad/actions?query=branch%3Amain

all tests passing now on main! https://github.com/antiwork/gumroad/actions/runs/20724467623

@gianfrancopiana could you please re-trigger the CI?

romitg2:
@gianfrancopiana seems like one of test verifies actual iframe being rendered, which requires original key, so using dummy key won't work here IMO. 

Screenshot 2026-01-06 at 12 34 27 AM

probably we can close this, or any other possible way to avoid using original key?

gianfrancopiana:
> @gianfrancopiana seems like one of test verifies actual iframe being rendered, which requires original key, so using dummy key won't work here IMO.
> 
> Screenshot 2026-01-06 at 12 34 27 AM
> probably we can close this, or any other possible way to avoid using original key?

You're right - this test verifies the actual PayPal iframe loads, which requires real credentials. A dummy client ID won't work. We'll close this PR and track the proper fix separately.

gianfrancopiana:
> > @gianfrancopiana seems like one of test verifies actual iframe being rendered, which requires original key, so using dummy key won't work here IMO.
> > Screenshot 2026-01-06 at 12 34 27 AM
> > probably we can close this, or any other possible way to avoid using original key?
> 
> You're right - this test verifies the actual PayPal iframe loads, which requires real credentials. A dummy client ID won't work. We'll close this PR and track the proper fix separately.

@romitg2 See #2699 for follow up.

PR #2696, opened 2026-01-05
Fix(test): adds dummy EASYPOST_API_KEY to .env.test

romitg2:
Issue: #959 

## Problem
- The [controller](https://github.com/antiwork/gumroad/blob/03b30281b0b2569db1560e38bd10f3cf0f99e21f/app/controllers/shipments_controller.rb#L17) creates an EasyPost client with
```rb
 EasyPost::Client.new(api_key: GlobalConfig.get("EASYPOST_API_KEY")). 
```
### Root Cause
The issue occurs when EASYPOST_API_KEY is missing from the test environment:
-  Client Initialization Fails Early: When GlobalConfig.get("EASYPOST_API_KEY") returns nil, the EasyPost SDK throws an error during EasyPost::Client.new(api_key: nil) - before any HTTP request is made.
-  VCR Never Gets Invoked: VCR can only intercept actual HTTP requests. Since the error occurs during client initialization (before HTTP), no request is made and VCR never gets a chance to replay the cassette.
- API Key Value Doesn't Matter for VCR: VCR matches requests by HTTP method and URI, not by headers or API keys. The actual value of the API key is irrelevant for cassette playback - it just needs to be non-nil so the client can initialize and make a request.
```rb
# From spec/spec_helper.rb:76:
config.filter_sensitive_data("<EASYPOST_API_KEY>") { GlobalConfig.get("EASYPOST_API_KEY") }
```
VCR already filters the API key in cassettes, so any non-nil value works.

## Solution

Add a dummy EASYPOST_API_KEY to the test environment:
```bash
EASYPOST_API_KEY=test-easypost-api-key
```

---

# Test Results
# Before:
### CI: 
https://github.com/antiwork/gumroad/actions/runs/20626922491
Screenshot 2026-01-02 at 1 21 07 PM

### Local: 
Screenshot 2026-01-02 at 1 25 17 PM

# After:
### Local: 
Screenshot 2026-01-02 at 1 24 29 PM

---

# Test Results

# Checklist

---

# AI Disclosure
use cursor with claude opus 4.5 for debugging


.env.test
```diff
@@ -61,3 +61,5 @@ STRONGBOX_GENERAL="-----BEGIN RSA PRIVATE KEY-----\nProc-Type: 4,ENCRYPTED\nDEK-
 STRONGBOX_GENERAL_PASSWORD="1234"
 
 MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key
+
+EASYPOST_API_KEY=test-easypost-api-key
```
romitg2, line 65: Add a dummy EASYPOST_API_KEY=test-easypost-api-key to the test environment, VCR matches requests by HTTP method and URI (not API keys), so any non-nil value allows the client to initialize and make requests that VCR can intercept and replay from cassettes.

EmCousin:
@romitg2 @gianfrancopiana I may be missing some context, but it looks like this PR was merged into main while a number of tests were failing, and those failures are still present on the branch. Looking at the [commit history](https://github.com/antiwork/gumroad/commits/main/), we’ve run into a similar situation before with [this PR](https://github.com/antiwork/gumroad/pull/2666), which was later [reverted](https://github.com/antiwork/gumroad/pull/2693).

Would it make sense to revert this change as well, at least temporarily, so we can get main back to a clean CI state? Right now it’s hard to tell whether failures in other PRs are new/regressions or just fallout from the existing red build. T I understand working on those env vars can be quite tricky, but it’d be best if we avoid merging further changes until CI is green, rather than assuming the failures are flaky or unrelated.

gianfrancopiana:
> @romitg2 @gianfrancopiana I may be missing some context, but it looks like this PR was merged into main while a number of tests were failing, and those failures are still present on the branch. Looking at the [commit history](https://github.com/antiwork/gumroad/commits/main/), we’ve run into a similar situation before with [this PR](https://github.com/antiwork/gumroad/pull/2666), which was later [reverted](https://github.com/antiwork/gumroad/pull/2693).
> 
> Would it make sense to revert this change as well, at least temporarily, so we can get main back to a clean CI state? Right now it’s hard to tell whether failures in other PRs are new/regressions or just fallout from the existing red build. T I understand working on those env vars can be quite tricky, but it’d be best if we avoid merging further changes until CI is green, rather than assuming the failures are flaky or unrelated.

Thanks for flagging, @emcousin. I definitely misinterpreted those failures as unrelated flakes given the scope of the change.

I’m reverting this now to get ⁠main back to green. I'll investigate the failures in a fresh branch before re-opening.

Issue #2662, opened 2026-01-04
Update bank account format for Gibraltar (GI)


vatsalkaushik:
### What
- [ ] Update GI's required bank account format in the app to match with Stripe's requirements. 

### Why
We currently ask Gibraltar users to provide their IBAN for bank account setup, but [Stripe requires](https://docs.stripe.com/connect/payouts-bank-accounts?bank-account-collection-method=manual-entry&bank-account-collection-integration=direct-api&account-country=GI) a Sort Code and Account number.

`/settings/payments` for a GI account: 

<img width="1986" height="268" alt="Image" src="https://github.com/user-attachments/assets/eff868c7-b75e-4b84-b569-ae6ca47e3d72" />

Stripe documentation:

<img width="1184" height="442" alt="Image" src="https://github.com/user-attachments/assets/63eafc3c-05cf-4bd6-8140-403dcd893dc9" />

PR #2698, opened 2026-01-05
Update Gibraltar (GI) bank account format to use Sort Code + Account Number

gianfrancopiana:
Closes: #2662

# Description

## Problem

Gibraltar users were asked to provide an IBAN for bank account setup, but Stripe requires a Sort Code and Account Number for GI accounts.

## Solution

Updated the bank account format for Gibraltar to use:
- **Sort code**: 6 digits in `XX-XX-XX` format
- **Account number**: 8 digits

---

# Before/After

### Before: Single IBAN field

**Desktop**

**Mobile**

### After: Sort code + Account number fields

**Desktop**

**Mobile**

---

# Test Results

---

# Checklist

---

# AI Disclosure

### DIFF
--- a/app/javascript/components/Payouts/index.tsx
+++ b/app/javascript/components/Payouts/index.tsx
@@ -169,14 +169,12 @@ export type BankAccount =
       routing_number: string;
       account_number: string;
       bank_name?: string;
-      sort_code: string;
     }
   | {
       payout_method_type: "bank";
       bank_account_type: "GI";
       routing_number: string;
       account_number: string;
-      sort_code: string;
     }
   | { payout_method_type: "bank"; bank_account_type: "GA"; routing_number: string; account_number: string }
   | { payout_method_type: "bank"; bank_account_type: "MC"; account_number: string };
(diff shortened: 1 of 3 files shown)


vatsalkaushik:
Thanks! I'm not a dev/right person to review, assigning to @hchhabra _a

app/javascript/components/Payouts/index.tsx
```diff
@@ -171,7 +171,13 @@ export type BankAccount =
       bank_name?: string;
       sort_code: string;
     }
-  | { payout_method_type: "bank"; bank_account_type: "GI"; account_number: string }
+  | {
+      payout_method_type: "bank";
+      bank_account_type: "GI";
+      routing_number: string;
+      account_number: string;
+      sort_code: string;
```
hchhabra: `routing_number` and `sort_code` are the same so we should only need either one here I guess, can adding `sort_code` in `GibraltarBankAccount#to_hash` be avoided?

gianfrancopiana: UK bank accounts follow the same pattern – `routing_number` returns `sort_code` but `to_hash` includes both. I kept GI consistent with UK since they use the same sort code + account number format. Happy to remove `sort_code` from both UK and GI if we want to simplify, but wanted to keep them aligned.

hchhabra: Ah, ok. We can remove from both if it's redundant I think.

PR #2702, opened 2026-01-06
fix: backfill missing EndSubscriptionWorker jobs for fixed-length subscriptions

nyomanjyotisa:
Issue: https://github.com/antiwork/gumroad-private/issues/46

# Description

Some fixed-length membership subscriptions are not ending automatically. Subscriptions that should have ended are missing their scheduled `EndSubscriptionWorker` jobs

## Solution

Created a one-time script to identify and fix affected subscriptions.

# Usage:
```bash
# Preview changes
Onetime::FixMissingSubscriptionEndJobs.perform
# Execute fixes
Onetime::FixMissingSubscriptionEndJobs.perform(dry_run: false)
```

# Tests
I tested the rake task by creating three yearly subscriptions for a product that should automatically end after 12 months:

- Subscription 1 was created on January 6, 2026 and already has an EndSubscriptionWorker correctly scheduled in Sidekiq.
- Subscription 2 was created on January 6, 2026 but does not have an EndSubscriptionWorker scheduled in Sidekiq (the worker was intentionally deleted).
- Subscription 3 was created on December 6, 2023 and remains active, even though it should have ended after the 12-month period

### Only one `EndSubscriptionWorker` exist, scheduled for Subscription 1
image

### Subscription 3 was still active even though it should have ended on December 6, 2024
image

### Dry run
image

### Live Run
Screenshot 2026-01-12 at 17 13 37
Screenshot 2026-01-12 at 17 13 51

### A new `EndSubscriptionWorker` was successfully added for Subscription 2
image

### Subscription 3 updated to ended at December 6, 2024
image

### DIFF
--- a/app/services/onetime/fix_missing_subscription_end_jobs.rb
+++ b/app/services/onetime/fix_missing_subscription_end_jobs.rb
@@ -17,7 +17,7 @@ module Onetime
 
       Subscription.where.not(charge_occurrence_count: nil)
                   .where(ended_at: nil)
-                  .find_each(batch_size: 100) do |subscription|
+                  .find_each do |subscription|
         next unless subscription.charges_completed?
 
         begin


lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+desc "Fix fixed-length subscriptions missing EndSubscriptionWorker jobs"
+namespace :subscriptions do
+  task fix_missing_end_jobs: :environment do
+    dry_run = ENV.fetch("DRY_RUN", "true") == "true"
```
nyomanjyotisa: Defaults to dry-run mode to prevent accidental data mutation when running in production. Requires explicit DRY_RUN=false to apply changes

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
...
+    puts dry_run ? "DRY RUN MODE\n" : "LIVE MODE\n"
+
+    scheduled_jobs = Set.new
+    Sidekiq::ScheduledSet.new.scan("EndSubscriptionWorker") do |job|
+      scheduled_jobs.add(job.args.first)
```
nyomanjyotisa: Pre-indexed scheduled `EndSubscriptionWorker` here to prevent scanning Sidekiq for every subscription

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
...
+        puts "  Last purchase id: #{last_purchase.id} (#{last_purchase.external_id})"
+        puts "  Last charge: #{last_purchase.succeeded_at}"
+        puts "  Should end: #{should_end_at}"
+
+        if should_end_at <= Time.current
+          puts "  Action: End subscription (past due)"
+          unless dry_run
+            subscription.with_lock do
+              subscription.ended_at = should_end_at
+              subscription.deactivate!
```
nyomanjyotisa: If the subscription should already have ended, immediately mark it as ended instead of scheduling a worker in the past.

I don't think we need to send out emails here like the original `end_subscription` [here](https://github.com/antiwork/gumroad/blob/446705ca09476b65377a1d7bc22030e39c799130/app/models/subscription.rb#L426-L436), since it actually ended in the past

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
...
+            end
+          end
+          past_due += 1
+        else
+          if scheduled_jobs.include?(subscription.id)
+            puts "  Action: Worker already scheduled (skipped)"
+          else
+            puts "  Action: Schedule EndSubscriptionWorker for #{should_end_at}"
+            unless dry_run
+              EndSubscriptionWorker.perform_at(should_end_at, subscription.id)
```
nyomanjyotisa: Schedules a new worker at the correct future time to end the subscription automatically, if no scheduled end subscription worker found for this subscription

nyomanjyotisa:
Hi @EmCousin _a could you help to review this?

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+desc "Fix fixed-length subscriptions missing EndSubscriptionWorker jobs"
+namespace :subscriptions do
+  task fix_missing_end_jobs: :environment do
```
EmCousin: We currently have an internal convention about creating dedicated service objects under the `OneTime` module instead of rake tasks. Could you move this code under such a service?

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    past_due = 0
+    future_scheduled = 0
+    errors = 0
+    processed = 0
+
+    Subscription.where.not(charge_occurrence_count: nil)
+                .where(ended_at: nil)
+                .find_each(batch_size: 100) do |subscription|
```
EmCousin: A comment about how you decided the batch size would help. Do you know how many subscriptions are affected in production right now? This could be a default value, overrideable via a keyword argument in the service

lib/tasks/backfill_subscription_end_jobs.rake
EmCousin: Makes sense!

lib/tasks/backfill_subscription_end_jobs.rake
```diff
@@ -0,0 +1,67 @@
...
+          puts "  Action: End subscription (past due)"
+          unless dry_run
+            subscription.with_lock do
+              subscription.ended_at = should_end_at
+              subscription.deactivate!
+            end
+          end
+          past_due += 1
+        else
+          if scheduled_jobs.include?(subscription.id)
```
EmCousin: Use `elsif` to reduce nesting and complexity here.
```ruby
        elsif scheduled_jobs.include?(subscription.id)
```

lib/tasks/backfill_subscription_end_jobs.rake
nyomanjyotisa: Moved to `app/services/onetime/fix_missing_subscription_end_jobs.rb` and updated the PR description

lib/tasks/backfill_subscription_end_jobs.rake
nyomanjyotisa: There are currently more than 34k subscriptions in production that match `Subscription.where.not(charge_occurrence_count: nil).where(ended_at: nil).count`.
Screenshot 2026-01-12 at 17 19 32


After another check, seems like the default batch size is enough, it is 1000 right? The operations performed in each iteration are lightweight and will be a lot of skip

Updated to use default batch size

lib/tasks/backfill_subscription_end_jobs.rake
nyomanjyotisa: Updated

Issue #2703, opened 2026-01-06
Checkout email remains stuck to old account email after user updates their email


vatsalkaushik:
### What

- [ ] Ensure the checkout email always reflects the logged-in user’s current email by clearing or re-syncing `alive_cart.email` when the account email changes.

### Why
In some cases (not consistently reproducible), when a logged-in user changes their account email, the `/checkout` page continues to auto-fill the previous email instead of the updated one. 

Refreshing alone doesn’t fix it, and requires running the following command in the terminal: 

`User.find_by(email: "example@email.com").alive_cart.update!(email: nil)`

**Example email change record:**

<img width="2206" height="192" alt="Image" src="https://github.com/user-attachments/assets/d9b1df1e-f076-44fb-8042-e445205082fe" />

**Old email still persistent on checkout:**

<img width="2438" height="694" alt="Image" src="https://github.com/user-attachments/assets/e7e01e22-9b32-4809-832e-486d46417962" />

EmCousin:
My apologies, I'll try to be clearer. When the buyer updates their login email, their cart email, if it exists:
1. should be cleared?
2. should be updated with the login email?

PR #2706, opened 2026-01-06
Fix(bug): Checkout email remains stuck to old account email after user updates their email

romitg2:
fixes: #2703 

## Problem
When a logged-in user changes their account email, the /checkout page continues to auto-fill the previous email instead of the updated one.

## Solution: 

- Added an after_update callback that syncs the cart email to match the user's confirmed email:
```rb
after_update :update_alive_cart_email, if: :saved_change_to_email?

... 
... 

def update_alive_cart_email
  reload_alive_cart&.update!(email: email)
end
```

---
# Test Result: 

### Before fix: 

Screenshot 2026-01-06 at 12 06 50 PM

### After fix: 

Screenshot 2026-01-06 at 12 03 44 PM

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure
- used cursor with claude-sonnet-4.5 to add test

PR #2709, opened 2026-01-06
Fix: Sync cart email on account email change

yashranaway:
Issue: #2703 

# Description

## Problem

When a logged-in user changes their account email, the `/checkout` page continues to auto-fill the previous email instead of the updated one. Refreshing doesn't fix it—currently requires manual intervention:

```ruby
User.find_by(email: "example@email.com").alive_cart.update!(email: nil)
```

## Solution

Added an `after_commit` callback that clears `alive_cart.email` when the user's account email changes and is confirmed. The cart email is only cleared if it matches the old account email.

---

# Before/After

N/A - Backend logic change, no UI changes.

---

# Test Results
image

---

# Checklist

---

# AI Disclosure

- Model: Claude Opus 4.5 via Claude Code
- Used for: Codebase exploration, implementation, and test writing
- Code was reviewed and validated


app/models/user.rb
```diff
@@ -294,6 +294,7 @@ class User < ApplicationRecord
   after_update :update_audience_members_affiliates
   after_update :update_product_search_index!
   after_commit :move_purchases_to_new_email, on: :update, if: :email_previously_changed?
+  after_commit :sync_cart_email, on: :update, if: :email_previously_changed?
```
yashranaway, line 297: Added this callback right after `move_purchases_to_new_email` since they both handle post-email-change syncing. `Using email_previously_changed?` ensures it only fires when the email actually changed.

app/models/user.rb
```diff
@@ -1072,6 +1073,11 @@ def move_purchases_to_new_email
       end
     end
 
+    def sync_cart_email
+      return if unconfirmed_email.present?
+      alive_cart&.update!(email: nil) if alive_cart&.email == email_previously_was
+    end
+
```
yashranaway, line 1080: Three key checks here:

- Skip if `unconfirmed_email.present?` don't clear cart email until the new email is actually confirmed
- Only clear if `alive_cart&.email` matches old email preserves cases where user intentionally set a different checkout email
- Safe navigation `(&.)` handles users without a cart

spec/models/user_spec.rb
```diff
@@ -3043,6 +3043,59 @@ def sample_string_of_length(n)
...
+      let!(:cart) { create(:cart, user:, email: "old@example.com") }
+
+      it "does not clear cart email" do
+        user.update!(email: "new@example.com")
+
+        expect(cart.reload.email).to eq("old@example.com")
+      end
+    end
+  end
+
```
yashranaway, line 3098: Covered the main scenarios:

1. Happy path: cart email cleared when it matches old account email
2. Edge case: cart email preserved if it was set to something else (e.g., user wants receipts at a different email)
3. Edge case: no error if user has no cart
Important: email not cleared while still unconfirmed (prevents premature sync)

EmCousin:
Hi @yashranaway, thanks for creating a PR that quickly. I took the time to clarify the requirements of the issue and we simply need to update the cart email with the user email, so we'll be moving forward with https://github.com/antiwork/gumroad/pull/2706 since it better fits that. Thanks!

PR #2710, opened 2026-01-06
Fix: Prevent automatic re-subscription of unsubscribed customers on new purchases

vishaltelangre:
Issue: https://github.com/antiwork/gumroad-private/issues/49#issuecomment-3714105080

# Description

## Problem

Customers who unsubscribe from a seller’s emails are getting automatically re-subscribed when they make a new purchase, causing them to receive emails without consent.

This happens because the `purchases.can_contact` column defaults to `true` in the database schema. When a previously unsubscribed customer makes a new purchase, the default value triggers the `update_audience_member_details` callback, re-adding them to the mailing list without their consent.

## Solution

- Adds a callback that runs before purchase creation which checks whether or not the customer has any existing purchases with `can_contact: false` for the seller. If yes, then it sets `can_contact: false` on that purchase.
- This preserves unsubscribe status across all future purchases while still allowing explicit re-subscription via the unsubscribe page.


vishaltelangre:
@ershad _a Please review.

All failing specs on CI appear to be unrelated (and they also seem to be failing on `main`).

app/models/purchase.rb
```diff
@@ -3866,4 +3867,12 @@ def fetch_installment_payments_from_snapshot_or_plan(total_price_cents)
         fetch_installment_plan.calculate_installment_payment_price_cents(total_price_cents)
       end
     end
+
+    def toggle_off_can_contact_if_buyer_has_unsubscribed
+      return unless new_record?
+      return unless can_contact?
+      return unless Purchase.where(email:, seller_id:, can_contact: false).exists?
```
ershad, line 3874: Tested in production DB. With indexes, this lookup is fast.

ershad (approved):
@vishaltelangre lgtm 👍🏼 will merge on green.

PR #2714, opened 2026-01-06
Fix(test): adds RECAPTCHA keys for testing to .env.test

romitg2:
Issue: #959 

# Description

- adds google's official RECAPTCHA keys to run tests. 

# Problems
### 1. reCAPTCHA iframe not rendering when keys are not present ( also won't work with dummy keys)

- Feature specs fail because the reCAPTCHA widget doesn't render:
```
Failure/Error: expect(page).to have_selector("iframe[title=recaptcha]", visible: false)
expected to find css "iframe[title=recaptcha]" but there were no matches
```
Screenshot 2026-01-03 at 2 30 42 PM

### Root Cause: Google's reCAPTCHA service validates the site key server-side before rendering the widget. 
because of empty (or dummy) keys, the iframe never appears on the page.

---

### 2. reCAPTCHA verification tests not actually testing verification logic
- Test expected valid_recaptcha_response_and_hostname? to be called and return false, but purchase succeeded anyway (count changed from 1 to 2).
```rb
it "does not allow the purchase to proceed if reCAPTCHA verification fails" do
  expect_any_instance_of(PurchasesController).to receive(:valid_recaptcha_response_and_hostname?).and_return(false)
  # ...
end

### Root Cause: The skip_recaptcha? method returns true early when no site key is configured, completely bypassing 

reCAPTCHA verification:
```rb
def skip_recaptcha?
  site_key = GlobalConfig.get("RECAPTCHA_MONEY_SITE_KEY")
  return true if (Rails.env.development? || Rails.env.test?) && site_key.blank?
  # ...
end
```

---

## Solution
- use's google's official recaptcha keys to run tests. 

https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
Screenshot 2026-01-03 at 2 26 40 PM

---

# Test Results

# Before: 
### CI: 
https://github.com/antiwork/gumroad/actions/runs/20626922491/job/59239171491#step:7:101

Screenshot 2026-01-02 at 2 56 31 PM
Screenshot 2026-01-03 at 2 30 42 PM

### Local: 
Screenshot 2026-01-02 at 2 38 49 PM
Screenshot 2026-01-03 at 2 17 28 PM

# After:
### Local: 
Screenshot 2026-01-02 at 2 39 01 PM
Screenshot 2026-01-03 at 2 17 31 PM
---

# Checklist

---

# AI Disclosure
- used cursor with claude-opus-4.5 for debugging purpose


romitg2:
@gianfrancopiana could you please review this PR? Thanks!

.env.test
```diff
@@ -63,3 +63,7 @@ STRONGBOX_GENERAL="-----BEGIN RSA PRIVATE KEY-----\nProc-Type: 4,ENCRYPTED\nDEK-
 STRONGBOX_GENERAL_PASSWORD="1234"
 
 MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key
+
+RECAPTCHA_MONEY_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+RECAPTCHA_SIGNUP_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+RECAPTCHA_LOGIN_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
```
romitg2, line 69: https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
Screenshot 2026-01-03 at 2 26 40 PM

 Google's official test keys are required (not dummy values) because the reCAPTCHA widget only renders with keys that Google's servers recognize as valid.

.env.test
```diff
@@ -63,3 +63,7 @@ STRONGBOX_GENERAL="-----BEGIN RSA PRIVATE KEY-----\nProc-Type: 4,ENCRYPTED\nDEK-
 STRONGBOX_GENERAL_PASSWORD="1234"
 
 MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key
+
+RECAPTCHA_MONEY_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
```
gianfrancopiana, line 67: It looks like we're still missing `GOOGLE_CLOUD_PROJECT_ID` and `ENTERPRISE_RECAPTCHA_API_KEY`. Can we add those too?

romitg2: @gianfrancopiana `GOOGLE_CLOUD_PROJECT_ID` and `ENTERPRISE_RECAPTCHA_API_KEY` are not really required to run tests as of now. requirement for these keys are already handled for test environment. 
- in `app/controllers/concerns/validate_recaptcha.rb` we're explicitly returning `true` for test environment. 
```rb
    def valid_recaptcha_response_and_hostname?(site_key:)
      return true if Rails.env.test?
```

and in `help_center.rb` spec we're stubbing it
```rb
    allow(GlobalConfig).to receive(:get).with("ENTERPRISE_RECAPTCHA_API_KEY")
```

romitg2: so not really sure weather we should add them, can you please confirm weather should i still add them?

gianfrancopiana: You’re right – I re-checked and we don’t actually need `GOOGLE_CLOUD_PROJECT_ID` or `ENTERPRISE_RECAPTCHA_API_KEY` for tests right now. Let’s keep this PR focused on the site keys (that’s what fixes the failing iframe specs). If we ever add coverage that hits the Enterprise path, we can add placeholders then.

Issue #2689, opened 2026-01-05
(Dev) Cannot run full app without setting HELPER_WIDGET_SECRET


brentguf:
I need to set `HELPER_WIDGET_SECRET` in .env.development to some random value in order to run the full app.

I'm surprised this has not been raised before. How do you currently handle this?

Slightly more context and a potential solution offered in https://github.com/antiwork/gumroad/pull/2343.

binary-koan:
`.env` is gitignored, so a workaround is to set it to a dummy value there. Will make it optional though as I mentioned in the PR, it's not supposed to be required

nyomanjyotisa:
Resolved via https://github.com/antiwork/gumroad/pull/2717

PR #2717, opened 2026-01-06
Make HELPER_WIDGET_SECRET config optional

binary-koan:
Issue: #2689

# Description

## Problem

Setting the Helper session currently breaks unless some value is set for HELPER_WIDGET_SECRET. Linking Helper is meant to be optional, and apart from that we skip loading it if config is not set.

## Solution

Don't try to build a helper session if there's no secret, i.e. disable Helper for logged in users.

---

# AI Disclosure

Not used for this PR


.env.example
```diff
@@ -144,8 +144,6 @@ IFFY_WEBHOOK_SECRET=
 MAILER_HEADERS_ENCRYPTION_KEY_V1=
 CLOUDFLARE_HMAC_KEY=
 CLOUDFLARE_CACHE_LIMIT=
-HELPER_WIDGET_SECRET=
-HELPER_TOOLS_TOKEN=
```
binary-koan, line 148: Other optional Helper env vars are also omitted from this file; setup is documented in docs/helper_widget.md

binary-koan (commented):
@EmCousin _a Could you take a quick look at this?

PR #2719, opened 2026-01-06
Fix: Run EasyPost tests without real secrets (scoped hook)

abhi152003:
ref: #959

- Add EASYPOST_API_KEY placeholder to .env.test
- Add VCR before_record hook (scoped to EasyPost only) to sanitize Basic Auth headers
- Sanitize 11 VCR cassettes to use placeholder credentials

Issue: #959

# Description

## Problem

The ShipmentsController tests fail when running without Gumroad's internal EasyPost API credentials. The VCR cassettes contain hardcoded base64-encoded API keys in the `Authorization` headers.

## Solution

1. Added `EASYPOST_API_KEY=test-easypost-api-key` placeholder to .env.test
2. Added a `before_record` hook to VCR configuration **scoped to EasyPost only** to sanitize Basic Auth headers
3. Sanitized 11 existing VCR cassettes to replace hardcoded credentials with placeholder

**Note:** The hook is scoped to only run for `easypost.com` requests to avoid affecting other services' cassettes.

---

# Test Results

All EasyPost VCR tests passed successfully :

image

---

# Checklist

---

# AI Disclosure

Claude was used to assist with identifying the root cause, writing the VCR hook, and verifying the fix.

### DIFF
--- a/spec/requests/purchases/product/shipping/shipping_address_verification_spec.rb
+++ b/spec/requests/purchases/product/shipping/shipping_address_verification_spec.rb
@@ -37,7 +37,7 @@ describe("Product Page - Shipping Scenarios Address verification", type: :system
         verifications: OpenStruct.new(
           delivery: OpenStruct.new(
             success: false,
-            errors: [OpenStruct.new(message: "Address not found")],
+            errors: [OpenStruct.new(code: "E.ADDRESS.NOT_FOUND", message: "Address not found")],
             details: nil
           )
         )


spec/spec_helper.rb
```diff
@@ -105,6 +105,21 @@ def configure_vcr
...
+      if interaction.request.headers["Authorization"]
+        interaction.request.headers["Authorization"] = interaction.request.headers["Authorization"].map do |auth|
+          if auth.start_with?("Basic ")
+            "Basic <EASYPOST_API_KEY_BASE64>"
+          else
+            auth
+          end
+        end
+      end
+    end
```
abhi152003, line 112: This hook is scoped to only run for EasyPost requests (easypost.com). It sanitizes Basic Auth headers before VCR records new cassettes, preventing real API keys from being stored. The next unless guard ensures this hook is a no-op for Stripe, PayPal, and other services.

romitg2:
@abhi152003 other than ShipmentsController, multiple other tests requires easypost api key, which fails when using dummy keys, eg. https://github.com/antiwork/gumroad/actions/runs/20740032455 (failed when used dummy easypost credentials), can you please make sure [these tests](https://github.com/antiwork/gumroad/actions/runs/20740032455) passes too.

gianfrancopiana:
> @abhi152003 other than ShipmentsController, multiple other tests requires easypost api key, which fails when using dummy keys, eg. https://github.com/antiwork/gumroad/actions/runs/20740032455 (failed when used dummy easypost credentials), can you please make sure [these tests](https://github.com/antiwork/gumroad/actions/runs/20740032455) passes too.

This PR only sanitizes 11 cassettes in `ShipmentsController/` but there are 26 more cassettes in `ShippingScenarios/` that still contain the real API key.

You can run this to see them:
`grep -rl "RVpUSzJhNDk1NTQyNTdlODQ5YmRiMTc5NDI4YTc2MTZkZWY2" spec/support/fixtures/vcr_cassettes/`

These need to be sanitized too.

spec/spec_helper.rb
gianfrancopiana: The `before_record` hook only runs when recording NEW cassettes. On CI, recording is disabled (`record: :none`), so this hook never runs for existing cassettes.

You'll need to manually update all existing cassettes that still have the real key.

gianfrancopiana:
Ref: #2751

abhi152003:
Thanks for the feedback! I've addressed it in the latest commit:

Changes made:

Sanitized all 37 VCR cassettes containing EasyPost credentials:
11 in ShipmentsController/
26 in ShippingScenarios/

Switched from before_record hook to filter_sensitive_data to follow the existing patterns in spec_helper.rb :
```
config.filter_sensitive_data("<EASYPOST_API_KEY_BASE64>") do
  Base64.strict_encode64("#{GlobalConfig.get('EASYPOST_API_KEY')}:")
end
```

This is more idiomatic and consistent with other filter_sensitive_data calls already in the file.

gianfrancopiana (changes requested):
@abhi152003 Some tests make real EasyPost API calls without using VCR, which causes the CI failures.

How it happens:
1. `js: true` tests turn VCR off
2. Tests without `:shipping` tag don't turn it back on
3. Real API calls are made with your fake key → auth fails → tests fail

The [failing tests in CI](https://github.com/antiwork/gumroad/actions/runs/20816020428?pr=2719) are all physical product tests without `:shipping` tag (GST, embeds, bundles, etc.)

To fix, we need to add `:shipping` tag to test blocks that involve physical products.

abhi152003:
Thanks for the detailed explanation! I've addressed this by implementing a two-part solution:

1. VCR Filter for EasyPost API Key : Added `filter_sensitive_data` in `spec_helper.rb` to replace the base64-encoded EasyPost API key with a placeholder:

```
config.filter_sensitive_data("<EASYPOST_API_KEY_BASE64>") do
  Base64.strict_encode64("#{GlobalConfig.get('EASYPOST_API_KEY')}:")
end
```
2. :mock_easypost Tag for `js: true` Tests Since adding `:shipping` tag would require recording new VCR cassettes (which needs a real EasyPost API key), I created a `:mock_easypost` metadata tag that uses RSpec mocking instead:

```
config.before(:each, :mock_easypost) do
  allow_any_instance_of(EasyPost::Services::Address).to receive(:create)
    .and_return(mock_address_response)
end
```
Applied this tag to physical product tests in:

`taxes_spec.rb`
`embed_spec.rb`
`shipping_address_verification_spec.rb`

VCR-based tests (`:shipping` tag) continue using cassettes `js: true` tests with physical products now use `:mock_easypost` tag to avoid real API calls

Let me know if you'd like any changes to this approach!

gianfrancopiana:
@abhi152003 Thanks – this is much closer.

Two important fixes are still needed:

1) The `:mock_easypost` stub currently returns `verifications` as a Hash, but the app code calls `address.verifications.delivery.success`. That will crash unless `verifications` responds to `.delivery` (e.g., `OpenStruct.new(delivery: ...)`). Please adjust the mock object shape to match what `ShipmentsController#verify_shipping_address` expects.

2) Please don't set `mock_easypost: true` on the entire `shipping_address_verification_spec.rb` file. Several tests in that file need different EasyPost outcomes (error, “verification required” suggestion, invalid zip, etc.). With a single always-success mock, those scenarios can’t be exercised. Instead, refactor those examples to build the specific mock EasyPost responses directly in the spec (no real API calls), and only enable mocking per-example as needed.

Also, a few failing files from the CI run still aren’t tagged:
- `spec/requests/checkout/bundle_spec.rb` (physical bundle context)
- `spec/requests/products/show/show_spec.rb` (physical sku purchase example)
- `spec/requests/purchases/product/quantities_spec.rb` (physical product setup)

Once these are addressed, we should be in good shape.

abhi152003:
Thanks for the feedback! I've made changes to address all three points:

1. Changed `verifications` from a Hash to an OpenStruct so `.delivery.success` works correctly:

```
verifications: OpenStruct.new(
  delivery: OpenStruct.new(success: true, errors: [], details: ...)
)
```

2. Removed `mock_easypost: true` from `shipping_address_verification_spec.rb` so that file can now use its own specific mocks for different EasyPost outcomes.

3. Added `:mock_easypost` to missing files :

`spec/requests/checkout/bundle_spec.rb` → physical bundle context
`spec/requests/products/show/show_spec.rb` → physical sku purchase test
`spec/requests/purchases/product/quantities_spec.rb` → multiple quantities describe block

Let me know if anything else needs adjustment!

spec/requests/purchases/product/shipping/shipping_address_verification_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-describe("Product Page - Shipping Scenarios Address verification", type: :system, js: true, mock_easypost: true) do
+describe("Product Page - Shipping Scenarios Address verification", type: :system, js: true) do
```
gianfrancopiana, line 3: The tests still make real API calls (lines 31-48, 72-80, 181-189):

```ruby
easy_post = EasyPost::Client.new(api_key: GlobalConfig.get("EASYPOST_API_KEY"))
address = easy_post.address.create(...)
```

These will fail with the fake key. We should refactor to build mock objects directly.

abhi152003:
@gianfrancopiana 

Done! I've refactored `shipping_address_verification_spec.rb` to replace all real EasyPost API calls with mock OpenStruct objects:

```
# Before - made real API calls
easy_post = EasyPost::Client.new(api_key: GlobalConfig.get("EASYPOST_API_KEY"))
address = easy_post.address.create(verify: ["delivery"], street1: "...", ...)

# After - builds mock objects directly
address = OpenStruct.new(
  street1: "1640 17TH ST",
  city: "SAN FRANCISCO",
  state: "CA",
  zip: "94107",
  country: "US",
  verifications: OpenStruct.new(
    delivery: OpenStruct.new(success: true, errors: [], details: ...)
  )
)

```

All 3 occurrences have been replaced. No more real EasyPost API calls in the tests.

spec/requests/purchases/product/shipping/shipping_address_verification_spec.rb
gianfrancopiana, line 132: Based on the test failures, it looks like the "address verification confirmation prompt" tests (lines 113-165) need EasyPost mocks. These 3 tests are making real API calls:

- "lets a buyer choose to use a verified address to complete their purchase"
- "lets a buyer choose not to use a verified address to complete their purchase"
- "does not allow the purchase if the buyer chooses not to use a verified address and the zip code is wrong for a taxable product"

Each needs a mock that returns an address where EasyPost suggests a corrected format (e.g., `"255 KING ST APT 602"` instead of `"255 King St #602"`) to trigger the verification prompt flow.

gianfrancopiana (changes requested):

abhi152003:
I've addressed the verification confirmation prompt tests you mentioned, and also noticed additional failures in CI for international address tests.

Changes made : 

1. Added mock for verification confirmation prompt tests in `shipping_address_verification_spec.rb`

2. Fixed `spec_helper.rb` mock to support international addresses - The mock was returning a hardcoded US address, causing failures for tests with international addresses. Now it dynamically echoes back the input address params.

This should fix all the failing `taxes_spec.rb` and `bundle_spec.rb` tests as well!

abhi152003:
@gianfrancopiana 

Thank you so much for catching that and pointing out the fix! I've added the code field to the error mock.
Really appreciate your patience and help throughout this PR!

PR #2729, opened 2026-01-07
feature: membership restart flow for canceled subscription

MuskanPaliwal:
### Problem

Currently all the users who have a canceled membership, don't see a way to resume their subscription and end up buying a new subscription. This leads to data duplicacy, user losing his purchase history, and increase in support load.

### Solution

We introduce a `resume` membership flow for users with canceled membership by interrupting the `subscribe` button click flow and prompting user with a modal to either resume or start afresh. 

#### Backend:
 1. Link Model - Added `restartable_subscription_for(user)` method, queries database for user's canceled subscriptions on this product
  - Filters out test subscriptions and admin-cancelled ones to reduce noice
  - Returns the most recently canceled subscription

  2. ProductPresenter - Added `restartable_subscription_props(user)`  method, formats subscription data (id, price, recurrence)
  - Generates magic link URL with authentication token
  - Passes data to frontend in product page props
  - Magic link uses existing token authentication (same as subscription management emails)

#### Frontend:
  - New `RestartSubscriptionModal` component with two options
  - Modified `CtaButton` to intercept checkout and show modal
       - "Resume" redirects to subscription management page (no login required via token)
       - "Start new" proceeds to normal checkout
---

### After Video for reference:

https://github.com/user-attachments/assets/a77d63a6-107f-495f-b20e-391bda954956

---

# Test Results

Screenshot 2026-01-06 at 10 23 13 PM

---

# Checklist

---

# AI Disclosure

I have used Claude Sonnet 4.5 for ui refactoring, writing tests and understanding gumroad codebase.

### DIFF
--- a/app/javascript/components/Product/CtaButton.tsx
+++ b/app/javascript/components/Product/CtaButton.tsx
@@ -78,18 +78,7 @@ const PARAMETERS_NOT_INHERITED_FROM_URL = new Set([
 ]);
 
 export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
-  (
-    {
-      product,
-      purchase,
-      discountCode,
-      selection,
-      label,
-      onClick,
-      showInstallmentPlanNotes = false,
-    },
-    ref,
-  ) => {
+  ({ product, purchase, discountCode, selection, label, onClick, showInstallmentPlanNotes = false }, ref) => {
     const { searchParams } = new URL(useOriginalLocation());
 
     const [referrer, setReferrer] = React.useState("");
@@ -143,7 +132,6 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
       onClick: (evt: React.MouseEvent<HTMLAnchorElement>) => {
         onClick?.(evt);
         if (evt.defaultPrevented) return;
-
         trackCtaClick({
           sellerId: product.seller?.id,
           name: product.name,
--- a/app/javascript/components/Product/Layout.tsx
+++ b/app/javascript/components/Product/Layout.tsx
@@ -140,15 +140,7 @@ export const Layout = (
     hasHero?: boolean;
   },
 ) => {
-  const {
-    product,
-    purchase,
-    discount_code: discountCode,
-    cart,
-    hasHero,
-    wishlists,
-    main_section_index,
-  } = props;
+  const { product, purchase, discount_code: discountCode, cart, hasHero, wishlists, main_section_index } = props;
   const [selection, setSelection] = useSelectionFromUrl(product);
   const ctaButtonRef = React.useRef<HTMLAnchorElement>(null);
   const ctaLabel = cart ? "Add to cart" : undefined;


app/models/link.rb
```diff
@@ -1162,6 +1162,25 @@ def support_email_or_default
     support_email || user.support_or_form_email
   end
 
+  def restartable_subscription_for(requested_user)
+    return nil unless is_recurring_billing?
+    return nil if requested_user.nil?
+    return nil if requested_user.id == user_id  # Seller shouldn't see restart modal
```
MuskanPaliwal: Prevents sellers from seeing the restart modal on their own products, which would be confusing since sellers don't "buy" their own memberships. (This is an assumption. let me know if this is not the case)

app/models/link.rb
```diff
@@ -1162,6 +1162,25 @@ def support_email_or_default
...
+
+    query = subscriptions
+      .not_is_test_subscription
+      .where("(subscriptions.flags & ?) = 0", Subscription.flag_mapping["flags"][:cancelled_by_admin])  # Exclude seller-cancelled subscriptions
```
MuskanPaliwal: These subscriptions were intentionally ended by sellers (policy violations, fraud, disputes) and should never prompt users to restart.

app/models/link.rb
```diff
@@ -1162,6 +1162,25 @@ def support_email_or_default
...
+      .where(ended_at: nil)  # Not ended
+
+    query = query.where(user: requested_user)
+
+    # Must be deactivated OR have pending cancellation (cancelled but not yet deactivated)
+    # Note: bit operator check above already filters out admin-cancelled subscriptions
+    query.where("deactivated_at IS NOT NULL OR (cancelled_at IS NOT NULL AND deactivated_at IS NULL)")
```
MuskanPaliwal: Users who cancelled but are still in grace period should see the restart option since their subscription is ending anyway. This catches them before they lose access and create a duplicate purchase

app/presenters/product_presenter/product_props.rb
```diff
@@ -172,4 +173,18 @@ def refund_policy_props
...
+      return nil unless user.present?
+
+      subscription = product.restartable_subscription_for(user)
+      return nil unless subscription
+
+      {
+        id: subscription.external_id,
+        price: subscription.price,
+        recurrence: subscription.recurrence,
+        manage_url: manage_subscription_url(subscription.external_id, token: subscription.token, host: "#{PROTOCOL}://#{DOMAIN}"),
```
MuskanPaliwal: Allows unauthenticated access to subscription management without requiring login. Token is time-limited for security

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -132,6 +165,15 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
       onClick: (evt: React.MouseEvent<HTMLAnchorElement>) => {
         onClick?.(evt);
         if (evt.defaultPrevented) return;
+
+        // Check for restartable subscription before allowing checkout
+        if (restartableSubscription && product.recurrences && !purchase) {
```
MuskanPaliwal: Why three conditions:
  1. `restartableSubscription`: User has a cancelled subscription to restart
  2. `product.recurrences`: Only show for memberships (not one-time products)
  3. `!purchase`: Don't show if user already has an active purchase (prevents confusion)

MuskanPaliwal:
Hi @ershad , can you review it ? I am not sure whom to reach out but would be great if you could take a look. Thanks!

PR #2732, opened 2026-01-07
Use lodash-es instead of lodash

brentguf:
# Description

Follow-up to https://github.com/antiwork/gumroad/pull/2688.

## Problem
When using the `lodash` package, it's easy to accidentally auto-import the full library with named imports (e.g., `import { map } from 'lodash'`).

## Solution
Use `lodash-es` instead, which is structured as ES modules and supports tree shaking with named imports.

Bonus: total JS bundle size got reduced by a few %.


brentguf:
cc @binary-koan

CONTRIBUTING.md
```diff
@@ -74,7 +74,6 @@ Generally: Include an AI disclosure, self-review (comment) on your code, break u
     Link.where(id: batch.map(&:id)).update_all(Link.set_flag_sql(:DEPRECATED_stream_only, false))
   end
   ```
-- Use `import debounce from "lodash/debounce"` instead of `import { debounce } from "lodash"` because tree-shaking doesn't work well with Lodash.
```
binary-koan, line 77: Awesome, love not needing rules 😄

brentguf:
@binary-koan Let's merge it asap if it's good to go? Already had a merge conflict 🙂

CONTRIBUTING.md
brentguf: 💯 The less you need to remember, the better!

binary-koan:
Just running CI to double check (I don't anticipate any problems though) but yep will merge once that's done

Issue #2725, opened 2026-01-06
Fix MAILER_HEADERS_ENCRYPTION_KEY_V1 loading in CI Docker test image


gianfrancopiana:
### Problem

Email-related tests fail in CI with the error:
```
no implicit conversion of nil into String
```

This happens because the `MAILER_HEADERS_ENCRYPTION_KEY_V1` environment variable is not set in the Docker test container.

### Solution

Add the missing variable to the ENV block in `docker/web/Dockerfile.test`:

```dockerfile
MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key
```

### Acceptance Criteria

- [ ] Variable added to `docker/web/Dockerfile.test`
- [ ] CI no longer shows "no implicit conversion of nil" errors for email encryption
- [ ] Previously failing email tests now pass

gianfrancopiana:
After doing a more [recent run](https://github.com/antiwork/gumroad/actions/runs/20790087323) without credentials I can't reproduce the issue anymore. Closing.

PR #2741, opened 2026-01-07
Add MAILER_HEADERS_ENCRYPTION_KEY_V1 to Docker test image

gianfrancopiana:
Closes: #2725

# Description

## Problem

Email-related tests fail in CI with `no implicit conversion of nil into String` errors. This happens because the `MAILER_HEADERS_ENCRYPTION_KEY_V1` environment variable is not set in the Docker test container, even though it exists in `.env.test`.

The Docker test image sets environment variables directly in the Dockerfile rather than loading them from `.env.test`, and this key was missing.

## Solution

Add `MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key` to the ENV block in `docker/web/Dockerfile.test`. This matches the value already defined in `.env.test` and makes the Docker test environment consistent with local test runs.

---

# Before/After

N/A - No UI changes.

---

# Test Results

Verified by comparing CI runs:
- Old run ([20626922491](https://github.com/antiwork/gumroad/actions/runs/20626922491)): Multiple `Digest::SHA256.digest(key) - no implicit conversion of nil` errors
- New run ([20761329770](https://github.com/antiwork/gumroad/actions/runs/20761329770)): Zero encryption key errors

---

# AI Disclosure

### DIFF
--- a/docker/web/Dockerfile.test
+++ b/docker/web/Dockerfile.test
@@ -66,6 +66,7 @@ ENV DATABASE_HOST=db_test \
     AWS_S3_ENDPOINT=http://minio:9000 \
     MEMCACHE_SERVERS=memcached \
     RAILS_LOG_LEVEL=info \
+    MAILER_HEADERS_ENCRYPTION_KEY_V1=test_encryption_key \
     IN_DOCKER=true
 
 WORKDIR $APP_DIR


gianfrancopiana:
After doing a more [recent run](https://github.com/antiwork/gumroad/actions/runs/20790087323) without credentials I can't reproduce the issue anymore. Closing.

PR #2275, opened 2025-12-10
Migrate Dropdown to tailwind

lukemacauley:
Part of https://github.com/antiwork/gumroad/issues/1055

Previous PRs: 
https://github.com/antiwork/gumroad/pull/1892
https://github.com/antiwork/gumroad/pull/2188

Fix for ProfilePopover:

 | Before | After |
|--------|-------|
 |    Screenshot 2025-12-10 at 11 56 12    |    Screenshot 2025-12-10 at 11 56 31   |
 |   Screenshot 2025-12-10 at 11 56 44       |   Screenshot 2025-12-10 at 11 56 49   |

PR #2656, opened 2026-01-03
Migrate _dropdown.scss to tailwind

SarthakWade:
Issue: Part of #1055 

# Description
Migration from .scss to tailwindcss of `_dropdown.scss`.

## Problem
The _dropdown.scss file contains SCSS styles for dropdown and popover components that need to be migrated to Tailwind CSS as part of the ongoing CSS-to-Tailwind migration effort.

## Solution
Cleaned up the old dropdown styling by removing the legacy `_dropdown.scss` file and its imports from `design.scss` and `admin.scss`.

Introduced a new reusable Dropdown UI component built with Tailwind, and refactored `Popover.tsx` to rely entirely on Tailwind classes instead of the previous `.popover` and `.dropdown` CSS.

Updated the dashboard profile popover to ensure consistent and correct styling.

Finally, migrated all dependent components to use the new Dropdown component or Tailwind-based styles, including `settings, payments, discounts, upsells, menus, editors, and analytics views.`

---

# Before/After
In the below given UI screenshots I tried to include **all types** of dropdowns, I couldn't take screenshots of all.

#### Search popover - `/checkout/discounts`
| Before | After |
|--------|--------|
| before_search_popover | after_search_popover |
| light_before_search_popover | light_after_search_popover |

#### mobile

| Before | After |
|--------|--------|
| mobile_before_search_popover | mobile_after_search_popover |
| light_mobile_before_search_popover | light_mobile_after_search_popover | 

#### Three dots action menu - `/checkout/discounts`
| Before | After |
|--------|--------|
| before_three-dot-menu | after_three-dot-menu |
| light_before_three-dot-menu | light_after_three-dot-menu | 

#### mobile

| Before | After |
|--------|--------|
| mobile_before_three-dot-menu | mobile_after_three-dot-menu |
| light_mobile_before_three-dot-menu | light_mobile_after_three-dot-menu | 

#### Limit quantity toggle dropdown & Limit validity period - `/checkout/discounts`
| Before | After |
|--------|--------|
| before_limit_dropdown | after_limit_dropdown |
| light_before_limit_dropdown | light_after_limit_dropdown | 

#### mobile

| Before | After |
|--------|--------|
| before_mobile_limit_dropdown | mobile_after_limit_dropdown |
| light_mobile_before_limit_dropdown | light_mobile_after_limit_dropdown | 

#### Insert upsell modal (discount dropdown) - `/checkout/upsells`
| Before | After |
|--------|--------|
[...]

PR #2742, opened 2026-01-07
_dropdown.scss to tailwindcss

SarthakWade:
Issue: Part #1055
Fixes  #2275

# Description

Fixes tests in #2275 mentioned by @binary-koan https://github.com/antiwork/gumroad/pull/2275#issuecomment-3678191478 

## Problem
Failing tests (mentioned in https://github.com/antiwork/gumroad/pull/2275#issuecomment-3678191478):
- `spec/requests/analytics/sales_spec.rb:61` (and other analytics ones) 
- `spec/requests/products/show/show_spec.rb:590` 
- `spec/requests/discover/discover_spec.rb:546`
- `spec/requests/purchases/product/coffee_spec.rb:18` 
- `spec/requests/purchases/product/call_spec.rb:84` 

## Solution

The test failures were caused by a mismatch between new Radix UI–based `Popover` and the way our accessibility tests detect disclosure relationships.

Radix optimizes performance by unmounting popover content when it’s closed. However, our tests rely on `aria-controls` pointing to an element that actually exists in the DOM. Since the popover content was being removed entirely, the tests couldn’t find the referenced element, which caused them to fail.

What I changed to fix this

- Kept popover content mounted:

  * Enabled the `forceMount` prop in `Popover.tsx` so the content remains in the DOM even when closed.
  * Added `data-[state=closed]:hidden` to visually hide the content while keeping it accessible to tests and screen readers.

- Standardized popover usage across the app:

  * Found a few remaining places (`NewSalesReportPopover`, `EmailsPage` layout, and `Affiliates search`) that were still using the old popover API (`trigger` / `onToggle`).
  * Refactored them to use Radix’s composition model (`PopoverTrigger`, `PopoverContent`, and `onOpenChange`) for consistent behavior.

---

# Before/After

No change with respect to #2275.

---

# Test Results
#### `spec/requests/analytics/sales_spec.rb:61`:
sales_spec

Created: `db/seeds/test_setup.rb` to programmatically seed essential Merchant Accounts (`user_id: nil`) when running tests locally. This resolves validation errors where transactions failed due to missing merchant accounts.

#### `spec/requests/products/show/show_spec.rb:590`: 
show_spec

#### `spec/requests/discover/discover_spec.rb:546`:
discover_spec

#### `spec/requests/purchases/product/coffee_spec.rb:18`: 
coffee_spec

#### `spec/requests/purchases/product/call_spec.rb:84`:
call_spec

The checkout tests `coffee_spec` and `call_spec` are failing because of a local configuration mismatch. When the test tries to visit the checkout URL (`seller.test.gumroad.com:31337`), the application's redirect logic incorrectly strips the custom test port (`:31337`). This sends the browser to `port 80`, where `Nginx` is listening but doesn't know how to handle the request, resulting in a `502 Bad Gateway` error. Since the page crashes, the test can't find the buttons it's looking for.

---

# Checklist

---

# AI Disclosure

Used Claude Opus 4.5 for code refactoring and finding appropriate tests for the related changes.


app/javascript/components/Popover.tsx
```diff
@@ -1,107 +1,70 @@
+import * as PopoverPrimitive from "@radix-ui/react-popover";
```
SarthakWade, line 1: Replaced the entire custom popover implementation with `Radix UI` primitives (`@radix-ui/react-popover`). This aligns with modern React patterns and provides better accessibility out of the box.

app/javascript/components/Popover.tsx
```diff
@@ -1,107 +1,70 @@
...
+    },
+    ref,
+  ) => {
+    const content = (
+      <PopoverPrimitive.Content
+        ref={ref}
+        align={align}
+        collisionPadding={collisionPadding}
+        autoFocus={false}
+        forceMount
```
SarthakWade, line 49: Added `forceMount` prop to keep popover content in the DOM even when closed. This resolves test failures where `aria-controls` points to an element ID that doesn't exist when the popover is unmounted.

app/javascript/components/Popover.tsx
```diff
@@ -1,107 +1,70 @@
...
+        onOpenAutoFocus={(e: Event) => e.preventDefault()}
+        className={classNames(
+          "z-30 w-max max-w-[calc(100vw-2rem)] rounded-sm border border-border bg-background p-4 text-foreground shadow outline-none focus-visible:outline-none data-[state=closed]:hidden",
```
SarthakWade, line 52: Visually hides the popover content when closed while keeping it in the DOM for accessibility. This is the counterpart to `forceMount`, content stays mounted but is hidden from view.

app/javascript/components/Popover.tsx
```diff
@@ -1,107 +1,70 @@
...
-
-    return () => window.removeEventListener("resize", calculateSpace);
-  });
-
-  return {
-    translate: `min(${space}px - 100% - var(--spacer-4), 0px)`,
-    maxWidth: `calc(${maxWidth}px - 2 * var(--spacer-4))`,
-  };
-};
+    return usePortal ? <PopoverPrimitive.Portal forceMount>{content}</PopoverPrimitive.Portal> : content;
```
SarthakWade, line 67: Added optional portal rendering for cases where the popover needs to escape parent overflow constraints (e.g., in `NestedMenu`).

app/javascript/components/NestedMenu.tsx
```diff
@@ -8,7 +8,7 @@ import { classNames } from "$app/utils/classNames";
 import { Button } from "$app/components/Button";
 import { useCurrentSeller } from "$app/components/CurrentSeller";
 import { Icon } from "$app/components/Icons";
-import { useDropdownPosition } from "$app/components/Popover";
+import { Popover, PopoverContent, PopoverTrigger } from "$app/components/Popover";
```
SarthakWade, line 11: Updated import to use the new Radix-based `Popover`, `PopoverContent`, `PopoverTrigger` exports instead of the old `useDropdownPosition` hook.

app/javascript/components/NestedMenu.tsx
```diff
@@ -136,7 +136,7 @@ const Menubar = ({ moreLabel, ...extraAriaAttrs }: { moreLabel?: string | undefi
         <MenubarItem
           key={menuItem.key}
           menuItem={menuItem}
-          isHighlighted={highlightedMenubarItem === menuItem}
+          isHighlighted={highlightedMenubarItem?.key === menuItem.key}
```
SarthakWade, line 139: Fixed comparison (`highlightedMenubarItem === menuItem`) to key-based (`highlightedMenubarItem?.key === menuItem.key`). This fixes cases where React re-renders create new object instances that fail `===` checks even when they represent the same menu item.

app/javascript/components/NestedMenu.tsx
```diff
@@ -147,7 +147,8 @@ const Menubar = ({ moreLabel, ...extraAriaAttrs }: { moreLabel?: string | undefi
           menuItem={moreMenuItem}
           isHighlighted={
             highlightedMenubarItem?.key === moreMenuItem.key ||
-            (highlightedMenubarItem !== null && !!itemsUnderMore?.includes(highlightedMenubarItem))
+            (highlightedMenubarItem !== null &&
+              !!itemsUnderMore?.some((item) => item.key === highlightedMenubarItem.key))
```
SarthakWade, line 151: [139-on-comment](https://github.com/antiwork/gumroad/pull/2742/changes#r2669251951) Same fix changed comparison to key-based comparison. This fixes `discover_spec.rb:546`. Due to this "More" gets `aria-current=true` when a sub-category under it is selected.

spec/support/aria_extensions.rb
```diff
@@ -235,11 +235,16 @@ def position_considering_colspan(xpath)
...
     match_name = XPath.string.n.is(name.to_s) | XPath.attr(:"aria-label").equals(name.to_s)
     button = (XPath.self(:button) | (XPath.attr(:role) == "button")) & match_name
-    aria = XPath.descendant[XPath.attr(:id) == XPath.anywhere[button][XPath.attr(:"aria-expanded")].attr(:"aria-controls")]
+    # Standard ARIA pattern: content element exists and is linked via aria-controls
+    aria = XPath.anywhere[XPath.attr(:id) == XPath.anywhere[button][XPath.attr(:"aria-expanded")].attr(:"aria-controls")]
+    # Radix UI pattern: content may not exist until opened, look for any element with role="dialog" or data-radix-popper-content-wrapper
+    # that appears after clicking the button. For now, we still require the linking.
     details = XPath.descendant(:details)[XPath.child(:summary)[match_name]]
     aria + details
   end
```
SarthakWade, line 247: Updated the disclosure selector to use `XPath.anywhere` instead of `XPath.descendant`. This is necessary because Radix UI often renders popover/dialog content in portals outside the trigger's DOM tree.

spec/requests/purchases/product/coffee_spec.rb
```diff
@@ -26,7 +26,7 @@
       expect(page).to have_field("Price", with: "1")
 
       click_on "Donate"
-      expect(page).to have_current_path("/checkout")
+      expect(page).to have_current_path("/checkout", ignore_query: true)
```
SarthakWade: Changed from exact path matching to `ignore_query: true`. The checkout URL now includes query parameters (e.g., `?product=...&quantity=...`), so exact `"/checkout"` matching fails.

spec/requests/purchases/product/call_spec.rb
```diff
@@ -54,7 +54,7 @@
 
       scroll_to first("footer")
       click_on "I want this!"
-      expect(page).to have_current_path("/checkout")
+      expect(page).to have_current_path("/checkout", ignore_query: true)
```
SarthakWade: [comment](https://github.com/antiwork/gumroad/pull/2742/changes#r2669285037) Same as the comment mentioned in `coffee_spce.rb`.

spec/requests/purchases/product/call_spec.rb
```diff
@@ -93,7 +93,7 @@
...
```
SarthakWade: Same as above

spec/requests/discover/discover_spec.rb
```diff
@@ -547,7 +547,7 @@ def seed_products
         visit discover_url(host: discover_host)
 
         find("[role=menuitem]", text: "More").hover
-        click_on "Writing & Publishing"
+        click_on "Other"
```
SarthakWade, line 582: This test validates that clicking a subcategory under the "More" dropdown correctly sets `aria-current=true` on "More" (not "All"). The original category "Writing & Publishing" may not consistently appear under the "More" menu depending on screen width calculations in `useArrangeMenubarItems`.

"Other" is a more stable choice because it's typically the last category in the taxonomy list and is more reliably placed under "More" regardless of viewport size. This makes the test less flaky. But open for change or revert if it somehow adds more flakiness.

SarthakWade:
Hey @binary-koan, could you please review the fixes. This is the current approach that I took for solving the tests still facing issue of tests failing locally with a 502 Bad Gateway error. Is there an alternative approach I should consider for resolving these test failures locally? Happy to implement any suggested fixes!

binary-koan:
Thanks! Will plan to review properly tomorrow.

It might fix the 502s if you run this before running the tests:

```
RAILS_ENV=test bundle exec rails js:export
```

app/javascript/components/Popover.tsx
binary-koan: Are there cases where [this approach](https://github.com/antiwork/gumroad/pull/2742/changes/033ccb05ec735143a766de3e46751ae2e190b267#diff-3c046c52efa306c1be3d78975010d231c2180c04cd751594ebec6324376e48e7R172) (find by `disclosure_button` instead of `disclosure`) doesn't work well? Not mounting is a convenient optimisation and it would be a bit of a shame to remove it just to make specs work

SarthakWade: Removed `forceMount` and the tests should use `select_disclosure` /` toggle_disclosure` helpers (or find `:disclosure_button` and click it) to interact with popovers.

SarthakWade:
Hey @binary-koan, while fixing the tests, I found that the `disclosure_button` approach doesn’t work well with our existing test patterns because of how Radix UI Popover interacts with `capybara_accessible_selectors`.

By default, Radix unmounts popover content when it’s closed. The `:disclosure` selector depends on `aria-controls` pointing to an element that already exists in the DOM. When the content is unmounted, that element doesn’t exist, so the selector fails.

Although `select_disclosure` clicks the trigger before querying the content, this introduces timing issues. Capybara may try to locate the disclosure content before React finishes mounting it. In addition, many tests use helpers like `within_disclosure`, `have_disclosure`, or `find(:disclosure)` directly, all of which assume the content is present in the DOM at query time.

Restoring `forceMount` keeps the popover content in the DOM at all times and hides it using CSS (`data-[state=closed]:hidden`). This ensures `aria-controls` always points to a valid element, disclosure-based selectors work reliably, timing issues are eliminated, and the user experience remains unchanged.

Tests likely fixed by restoring `forceMount`:
- profile_spec.rb (creating sections and rich text flows)
- rich_text_editor_spec.rb (formatting, embeds, uploads)
- products/show and products/index specs (share and action menus)
- library_spec.rb (archive and unarchive flows)
- affiliates_spec.rb (search popovers)
- discover specs (navigation and nested menus)
- emails create/list specs (creation and search)
- workflows_spec.rb (insert and publish menus)
- products/edit_spec.rb (upsell creation)

Thus restored the `forceMount` on `Popover`. Could you check once, if the logic is correct or not. Happy to fix suggested changes!

binary-koan:
Gah, I see. Thanks for explaining, I think it's worth a comment where we use `forceMount` in that case, but I'm good with the approach

SarthakWade:
@binary-koan, The following tests were not modified in this branch at the failing line numbers, so they might be flaky or have pre-existing issues:
- library_spec.rb:149, 168, 189 (archive/unarchive)
- emails/list_spec.rb:267, 324 (email search)
- checkout/cart_spec.rb:165 (cart persistence)
- authentication_spec.rb:93 (logout)
- download_page/download_page_spec.rb:275 (installments)
- embed_spec.rb:141 (embed prefill)
- purchases/product/call_spec.rb:84 (call duration)
- analytics/sales_spec.rb:165 (date range)

Do you have any insight into whether these are known flaky tests, or should I check them further? Need some guidance on this. Open to refactor if needed.

SarthakWade:
The `discover_spec.rb failures (lines 275, 511, 557)` and `products/index_spec.rb failures (lines 171, 185)` appear to be caused by the `usePortal` prop on `PopoverContent`. When `usePortal` is used, the popover content renders at the body level outside the normal DOM hierarchy. This breaks tests that use within blocks to scope their assertions - for example, `within "[role=menubar]"` or `within find_product_row` won't find elements that are rendered in a portal outside those containers.
The original `NestedMenu` implementation used a simple `div.dropdown` inside the component without portals, so the within scoping worked. After migrating to Radix Popover with `usePortal`, the content is no longer a descendant of the scoped element.
Would you prefer I remove `usePortal` from `NestedMenu` and `ActionsPopover` to restore the original behavior, or should the tests be updated to handle portal-rendered content?

binary-koan:
I don't see any known flaky ones in that list, I suspect they're similar issues where the behaviour has changed (or something on that page causes a bug with the new popover). Happy to help if any are particularly tricky to debug!

> Would you prefer I remove usePortal from NestedMenu and ActionsPopover to restore the original behavior, or should the tests be updated to handle portal-rendered content?

Hmm. I think if we can get away with removing `usePortal` that's a better option, but if it causes issues (e.g. something has `overflow-hidden`) we should update the test. How does that sound?

SarthakWade:
Hey @binary-koan, removed `usePortal` from `NestedMenu`, `ActionsPopover`, and Library/Index components. Also reverted the test workarounds in `products/index_spec.rb` that were added to work around this issue. Ran the affected tests locally and they pass now.

  The other tests I mentioned (emails, authentication, checkout, etc.) don't use usePortal in their components, so they should be unrelated to this change.

SarthakWade:
@binary-koan , friendly ping.

binary-koan:
Sorry for the delay, I've been focusing on the button migration. It caused a ton of conflicts which I'll fix tomorrow (unless you feel like doing it first), and after that I'll focus on reviewing and merging this one!

SarthakWade:
I didn't delete any code from the main, just merged it with mine. Can you check once @binary-koan ?

app/javascript/components/Admin/SalesReports/NewSalesReportPopover.tsx
```diff
@@ -53,98 +52,96 @@ const NewSalesReportPopover = () => {
...
-    <Popover
-      open={open}
-      onToggle={setOpen}
-      aria-label="New Sales Report"
-      trigger={
+    <Popover open={open} onOpenChange={setOpen}>
+      <PopoverTrigger aria-label="New Sales Report" asChild>
         <WithTooltip tip="Generate a new sales report" position="bottom">
-          <div className={buttonVariants({ size: "default", color: "primary" })}>New report</div>
+          <div className="button primary">New report</div>
```
binary-koan: `buttonVariants` is correct here, the `button` class no longer exists (same in quite a few places I think)

app/javascript/components/Checkout/index.tsx
```diff
@@ -456,10 +456,14 @@ const CartItemComponent = ({
...
-    if (selection.optionId !== item.option_id && findCartItem(cart, item.product.permalink, selection.optionId))
-      return setError("You already have this item in your cart.");
+    if (isPWYW && (selection.price.value === null || selection.price.value < priceCents)) {
+      setSelection({ ...selection, price: { ...selection.price, error: true } });
+      return;
+    }
+    if (selection.optionId !== item.option_id && findCartItem(cart, item.product.permalink, selection.optionId)) {
+      setError("You already have this item in your cart.");
+      return;
+    }
```
binary-koan: Unrelated change? (lmk if it is needed)

app/javascript/components/Download/OpenInAppButton.tsx
```diff
@@ -1,60 +1,28 @@
...
+  <Popover>
+    <PopoverTrigger asChild>
+      <Button>Open in app</Button>
+    </PopoverTrigger>
+    <PopoverContent sideOffset={4}>
+      <div className="mx-auto grid w-72 gap-4 text-center">
+        <h3>Gumroad Library</h3>
+        <div>Download from the App Store</div>
+        <div className="grid grid-flow-col justify-between gap-4">
+          <a className="button button-apple" href={iosAppUrl} target="_blank" rel="noreferrer">
```
binary-koan: Same here (for example), we need to keep the Button asChild

spec/requests/purchases/product/coffee_spec.rb
binary-koan: Why has migrating popovers changed this? Sounds like it shouldn't be related

spec/support/aria_extensions.rb
```diff
@@ -235,11 +235,16 @@ def position_considering_colspan(xpath)
...
+# Also support Radix UI popovers where content is lazily rendered
 Capybara.modify_selector(:disclosure) do
   xpath do |name, **|
     match_name = XPath.string.n.is(name.to_s) | XPath.attr(:"aria-label").equals(name.to_s)
     button = (XPath.self(:button) | (XPath.attr(:role) == "button")) & match_name
-    aria = XPath.descendant[XPath.attr(:id) == XPath.anywhere[button][XPath.attr(:"aria-expanded")].attr(:"aria-controls")]
+    # Standard ARIA pattern: content element exists and is linked via aria-controls
+    aria = XPath.anywhere[XPath.attr(:id) == XPath.anywhere[button][XPath.attr(:"aria-expanded")].attr(:"aria-controls")]
+    # Radix UI pattern: content may not exist until opened, look for any element with role="dialog" or data-radix-popper-content-wrapper
+    # that appears after clicking the button. For now, we still require the linking.
```
binary-koan: Is this comment correct? I don't see a reference to role=dialog or data-radix-popper-content-wrapper in the code

SarthakWade:
@binary-koan, sorry for the rookie mistakes in the last commit, fixed the changes you asked, ran all 61 tests locally. All are passing! Could you please review it. Open for suggestions.

Screenshot_20260120_011830

binary-koan:
There are still some failing specs + TSC isn't happy, could you check? This is way closer though! (Circle integration is a known flaky one btw, don't worry about that)

SarthakWade:
Hey @binary-koan , can you help or guide me in solving remaining tests, I am still trying but getting kinda difficult and hectic. Could you check? Or we could shift to #2656 (the other approach)

binary-koan:
Sure, will take a look tomorrow-ish!

SarthakWade:
In [09b6ea4](https://github.com/antiwork/gumroad/pull/2742/commits/09b6ea49914c97246569830648922ae6948140a7) unwrapped `PopoverTrigger` elements to fix broken `ARIA propagation` that was causing test failures. Replaced custom `tooltips` with native `title` attributes to ensure tests pass without sacrificing accessibility.

SarthakWade:
@binary-koan, can you review? I have applied some fixes.

binary-koan:
Sorry I haven't got to this yet, will do over the weekend

binary-koan:
I think that's all the specs fixed (the remaining one looks flaky)

Will review properly and hopefully merge by Monday!

SarthakWade:
https://github.com/antiwork/gumroad/pull/2742/commits/3d98f8074c059bd267a24ce8771efda52c8a83bf fixed some UI inconsistencies.

binary-koan:
Checked through all pages and pushed a bunch of fixes, I think we're good to go once specs are passing

binary-koan:
Specs look flaky, will go ahead and merge. Really appreciate the help getting this over the line @SarthakWade!

SarthakWade:
@binary-koan thanks for helping and guiding me.

PR #2753, opened 2026-01-08
Fix Nav SSR crash when helper_session is null

brentguf:
Issue: https://github.com/antiwork/gumroad/issues/2689

# Description

Follow-up to https://github.com/antiwork/gumroad/pull/2717.

## Problem

The product edit page was crashing (through the Nav component?) during server-side rendering with:

> Error: Expected root.helper_session to be an object, got null

This occurred because the `helper_session` prop wasn't typed to be nullable (`null`), and so this wasn't part of the runtime validator generated by `createCast`.

## Solution
Added `| null` to the `helper_session` type definition.

---

# Test Results

Verified the page loads without errors when `helper_session` is null.

---

# AI Disclosure

Claude Opus 4.5 (Thinking) to locate the bug based on logs.

### DIFF
--- a/app/javascript/components/server-components/Nav.tsx
+++ b/app/javascript/components/server-components/Nav.tsx
@@ -34,7 +34,7 @@ type Props = {
       links?: Record<string, string> | null;
     } | null;
     currentToken?: string | null;
-  };
+  } | null;
 };
 
 export const Nav = (props: Props) => {


brentguf:
cc @binary-koan

binary-koan (approved):
My bad, thanks for the fix!

brentguf:
@binary-koan hold on for a sec! There's still a runtime error in my branch.

brentguf:
> hold on for a sec! There's still a runtime error in my branch.

Seems unrelated: something to do with one of my browser extensions.

The change seems safe, but feel free to quickly try the branch yourself/run tests. 🙂

binary-koan:
Yeah I don’t see this causing problems (famous last words). Will keep an eye on tests in main

Issue #1158, opened 2025-09-08
Automatically suspend creators w/ shared ACH accounts


gumroadsteve2:
What: 
We currently auto-suspend an account if it is using the same PayPal payout address as another account that has been suspended by Admin

We need to extend this feature to accounts using shared ACH accounts (stripe fingerprints)

Why: 
A lot of fraudsters will create multiple accounts using the same bank account for payouts. Auto-flagging/suspending them will decrease time spent hunting down fraudsters and reduce chargebacks/fraudulent purchases

How:
I suspend an account through Admin > Any account using the same stripe_fingerprint for payouts is automatically suspended > Admin adds a note to additional matching accounts that it was suspended for having the same payout address

sm17p:
No worries @gumroadsteve2. 

I don't know if you're using Github mobile app but I've never had issues on it for getting priority notifications

sm17p:
Hi @ershad @gumroadsteve2, I've identifed two existing bugs while working on this issue. Can you please help in triaging them so that we can take the correct approach in resolving #2810



1. #2830 
     - This one seems critical as we'd want to keep the behaviour similar for Stripe bank accounts. 

2. #2833 
     - This one doesn't break anything but felt like a good cleanup.

PR #2759, opened 2026-01-08
feat(fraud): auto-suspend creators with shared ACH accounts

FusionBrah:
Issue: #1158

# Description
## Problem
Fraudsters create multiple accounts using the same bank account for payouts. Currently, auto-suspension only works for PayPal payment addresses, not for Stripe bank account fingerprints.

## Solution
Extended the existing PayPal payment address auto-suspension feature to also work with Stripe bank account fingerprints (`stripe_fingerprint` field on `bank_accounts` table).

**Scenarios covered:**
1. **New account check**: When a user adds a bank account with a `stripe_fingerprint` that matches a suspended user's bank account, flag them for fraud
2. **Cascade suspension**: When an admin suspends a user, suspend all other accounts sharing the same `stripe_fingerprint`
3. **Cascade compliant**: When an admin marks a user compliant, mark compliant all related accounts with the same `stripe_fingerprint`

**Implementation details:**
- Created `CheckStripeFingerprintWorker` - triggered when bank account fingerprint is set, mirrors `CheckPaymentAddressWorker`
- Created `SuspendAccountsWithStripeFingerprintWorker` - triggered on suspension, mirrors `SuspendAccountsWithPaymentAddressWorker`
- Updated `User::Risk#suspend_sellers_other_accounts` to also call the new worker
- Updated `User::Risk#enable_sellers_other_accounts` to handle stripe fingerprints
- Added `stripe_bank_account_fingerprint` to `BlockedObject` types for admin blocking
- Added database index on `bank_accounts.stripe_fingerprint` for performance
- Added comprehensive logging for debugging and audit trails
- Used `lock: :until_executed` for idempotency per CLAUDE.md guidelines

**Fraud history preservation:**
- Checks ALL bank account fingerprints from a suspended user (not just the active one)
- Catches fraudsters even if they delete their bank account after suspension
- Handles users with multiple bank accounts by checking all fingerprints
- Includes the matching fingerprint value in suspension comments for audit clarity

**Performance optimization:**
- Eliminated N+1 queries by building a user_id → fingerprints map upfront instead of querying per user

---

## Code Review Fixes Applied

The following issues were identified during code review and have been fixed:

### Critical Fixes
1. **Added `skip_transition_callback`** to `SuspendAccountsWithStripeFingerprintWorker#suspend_for_fraud` to prevent infinite job cascading loops (matches existing `SuspendAccountsWithPaymentAddressWorker` pattern)

2. **Added `.not_suspended` scope** to the user query to avoid reprocessing already-suspended users (matches existing worker pattern)

3. **Handle already-flagged users** - Users in `flagged_for_fraud` state now proceed directly to suspension instead of being skipped. This closes a fraud bypass where users flagged through another mechanism would escape auto-suspension.

### Resilience Improvements
4. **Added error handling** to `enable_sellers_other_accounts` - Individual user failures now log an error and continue processing remaining users, preventing one failure from blocking the entire batch.

### Test Coverage Added
- Test for `SuspendAccountsWithStripeFingerprintWorker` enqueue in `suspend_sellers_other_accounts`
- Test for stripe fingerprint compliance restoration in `enable_sellers_other_accounts`
- Test for error handling continuation
- Test for already-flagged users being suspended
- Test for `skip_transition_callback` preventing cascading suspensions

---
# Before/After
Backend-only change - no UI modifications.

---
# Test Results

Screenshot 2026-01-23 at 3 44 30 pm
Screenshot 2026-01-23 at 3 44 53 pm

---
# Checklist

---
# AI Disclosure
- Understanding the existing PayPal auto-suspension pattern
- Implementing the stripe fingerprint workers following the same pattern
- Writing comprehensive tests
- Adding error handling and logging
- Improving fraud detection to check all fingerprints and preserve fraud history
- Eliminating N+1 queries for better performance
- Code review and fixing critical issues (cascading prevention, fraud bypass fix, error resilience)

### DIFF
--- a/app/modules/user/risk.rb
+++ b/app/modules/user/risk.rb
@@ -137,6 +137,8 @@ module User::Risk
     if payment_address.present?
       User.where(payment_address:).where.not(id:).find_each do |user|
         user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as payment address #{payment_address} is now unblocked (from User##{id})", skip_transition_callback: :enable_sellers_other_accounts)
+      rescue StandardError => e
+        Rails.logger.error("enable_sellers_other_accounts: Failed to mark user #{user.id} compliant: #{e.message}")
       end
     end
 
@@ -149,6 +151,8 @@ module User::Risk
           .distinct
           .find_each do |user|
         user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as bank account fingerprint is now unblocked (from User##{id})", skip_transition_callback: :enable_sellers_other_accounts)
+      rescue StandardError => e
+        Rails.logger.error("enable_sellers_other_accounts: Failed to mark user #{user.id} compliant: #{e.message}")
       end
     end
   end
--- a/app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
+++ b/app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
@@ -17,9 +17,11 @@ class SuspendAccountsWithStripeFingerprintWorker
         .merge(BankAccount.alive)
         .where(bank_accounts: { stripe_fingerprint: stripe_fingerprint })
         .where.not(id: suspended_user.id)
+        .not_suspended
         .distinct
         .find_each do |user|
-      unless user.flag_for_fraud(
+      # Skip flagging if already flagged for fraud, but proceed to suspension
+      unless user.flagged_for_fraud? || user.flag_for_fraud(
         author_name: "suspend_sellers_other_accounts",
         content: "Flagged for fraud automatically on #{Time.current.to_fs(:formatted_date_full_month)} because of usage of bank account fingerprint from User##{suspended_user.id}"
       )
@@ -29,7 +31,8 @@ class SuspendAccountsWithStripeFingerprintWorker
 
       unless user.suspend_for_fraud(
         author_name: "suspend_sellers_other_accounts",
-        content: "Suspended for fraud automatically on #{Time.current.to_fs(:formatted_date_full_month)} because of usage of bank account fingerprint from User##{suspended_user.id}"
+        content: "Suspended for fraud automatically on #{Time.current.to_fs(:formatted_date_full_month)} because of usage of bank account fingerprint from User##{suspended_user.id}",
+        skip_transition_callback: :suspend_sellers_other_accounts
       )
         Rails.logger.warn("SuspendAccountsWithStripeFingerprintWorker: Failed to suspend user #{user.id} (state: #{user.user_risk_state})")
       end
--- a/spec/modules/user/risk_spec.rb
+++ b/spec/modules/user/risk_spec.rb
@@ -61,5 +61,66 @@ describe User::Risk do
         end.to change(SuspendAccountsWithPaymentAddressWorker.jobs, :size).from(1).to(0)
       end
     end
+
+    context "when user has a Stripe bank account" do
+      it "calls SuspendAccountsWithStripeFingerprintWorker" do
+        user = create(:user)
+        create(:ach_account, user: user, stripe_fingerprint: "test_fingerprint")
+
+        expect do
+          user.suspend_sellers_other_accounts(transition)
+        end.to change(SuspendAccountsWithStripeFingerprintWorker.jobs, :size).by(1)
+
+        expect(SuspendAccountsWithStripeFingerprintWorker.jobs.last["args"]).to eq([user.id])
+      end
+    end
+  end
+
+  describe "#enable_sellers_other_accounts" do
+    let(:admin) { create(:admin_user) }
+    let(:transition) { double("transition", args: [{}]) }
+
+    context "when user has a Stripe bank account with matching fingerprint" do
+      it "marks related accounts compliant based on stripe fingerprint" do
+        user_1 = create(:user)
+        create(:ach_account, user: user_1, stripe_fingerprint: "shared_fingerprint")
+
+        user_2 = create(:user, user_risk_state: "suspended_for_fraud")
+        create(:ach_account, user: user_2, stripe_fingerprint: "shared_fingerprint")
+
+        user_1.enable_sellers_other_accounts(transition)
+
+        expect(user_2.reload.compliant?).to be(true)
+      end
+
+      it "logs an error and continues when marking a user compliant fails" do
+        user_1 = create(:user)
+        create(:ach_account, user: user_1, stripe_fingerprint: "shared_fingerprint")
+
+        user_2 = create(:user, user_risk_state: "suspended_for_fraud")
+        create(:ach_account, user: user_2, stripe_fingerprint: "shared_fingerprint")
+
+        user_3 = create(:user, user_risk_state: "suspended_for_fraud")
+        create(:ach_account, user: user_3, stripe_fingerprint: "shared_fingerprint")
+
+        # Stub mark_compliant! to fail for user_2 only
+        call_count = 0
+        allow_any_instance_of(User).to receive(:mark_compliant!).and_wrap_original do |method, **args|
+          call_count += 1
+          if call_count == 1
+            raise StandardError.new("test error")
+          else
+            method.call(**args)
+          end
+        end
+
+        expect(Rails.logger).to receive(:error).with(/Failed to mark user .* compliant: test error/)
+
+        user_1.enable_sellers_other_accounts(transition)
+
+        # user_3 should still be processed despite user_2 failing
+        expect(user_3.reload.compliant?).to be(true)
+      end
+    end
   end
 end
--- a/spec/sidekiq/suspend_accounts_with_stripe_fingerprint_worker_spec.rb
+++ b/spec/sidekiq/suspend_accounts_with_stripe_fingerprint_worker_spec.rb
@@ -84,7 +84,7 @@ describe SuspendAccountsWithStripeFingerprintWorker do
       expect { described_class.new.perform(999999999) }.to raise_error(ActiveRecord::RecordNotFound)
     end
 
-    it "continues suspending other accounts even if one cannot be suspended" do
+    it "skips already suspended accounts" do
       user_1 = create(:user)
       create(:ach_account, user: user_1, stripe_fingerprint: stripe_fingerprint)
 
@@ -98,6 +98,31 @@ describe SuspendAccountsWithStripeFingerprintWorker do
 
       expect(user_2.reload.user_risk_state).to eq("suspended_for_fraud")
       expect(user_3.reload.suspended?).to be(true)
+    end
+
+    it "suspends users who are already flagged for fraud" do
+      user_1 = create(:user)
+      create(:ach_account, user: user_1, stripe_fingerprint: stripe_fingerprint)
+
+      user_2 = create(:user, user_risk_state: "flagged_for_fraud")
+      create(:ach_account, user: user_2, stripe_fingerprint: stripe_fingerprint)
+
+      described_class.new.perform(user_1.id)
+
+      expect(user_2.reload.suspended_for_fraud?).to be(true)
+    end
+
+    it "does not trigger cascading suspensions due to skip_transition_callback" do
+      user_1 = create(:user)
+      create(:ach_account, user: user_1, stripe_fingerprint: stripe_fingerprint)
+
+      user_2 = create(:user)
+      create(:ach_account, user: user_2, stripe_fingerprint: stripe_fingerprint)
+
+      expect(SuspendAccountsWithStripeFingerprintWorker).not_to receive(:perform_in)
+      expect(SuspendAccountsWithPaymentAddressWorker).not_to receive(:perform_in)
+
+      described_class.new.perform(user_1.id)
     end
   end
 end


app/sidekiq/check_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class CheckStripeFingerprintWorker
+  include Sidekiq::Job
+  sidekiq_options retry: 0, queue: :default, lock: :until_executed
+
+  def perform(user_id, stripe_fingerprint)
+    user = User.find_by(id: user_id)
```
FusionBrah, line 8: Using `find_by` instead of `find` because this worker is triggered asynchronously - the user could have been deleted between when the job was enqueued and when it runs. This prevents raising RecordNotFound and cluttering error logs.

app/sidekiq/check_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    if stripe_fingerprint.blank?
+      Rails.logger.info("CheckStripeFingerprintWorker: Stripe fingerprint blank for user #{user_id}, skipping")
+      return
+    end
+
+    suspended_users_with_same_fingerprint = BankAccount.alive
+      .joins(:user)
+      .where(stripe_fingerprint: stripe_fingerprint)
+      .where(users: { user_risk_state: %w[suspended_for_tos_violation suspended_for_fraud] })
```
FusionBrah, line 28: Checking both `suspended_for_fraud` and `suspended_for_tos_violation` states to match the behavior of `CheckPaymentAddressWorker`. Users suspended for TOS violations (e.g., selling prohibited content) should also trigger the fingerprint check.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class SuspendAccountsWithStripeFingerprintWorker
+  include Sidekiq::Job
+  sidekiq_options retry: 5, queue: :default, lock: :until_executed
+
+  def perform(user_id)
+    suspended_user = User.find(user_id)
```
FusionBrah, line 8: Using `find` (not `find_by`) here because this worker is called from `suspend_sellers_other_accounts` which requires a valid user to exist. If the user was deleted, we want to raise `RecordNotFound` so Sidekiq can retry or alert us to a data integrity issue.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,38 @@
...
+      return
+    end
+
+    User.joins(:bank_accounts)
+        .merge(BankAccount.alive)
+        .where(bank_accounts: { stripe_fingerprint: stripe_fingerprint })
+        .where.not(id: suspended_user.id)
+        .distinct
+        .find_each do |user|
+      unless user.flag_for_fraud(
```
FusionBrah: Using non-bang methods (`flag_for_fraud` / `suspend_for_fraud`) with return value checking instead of bang methods. This prevents the loop from breaking if one user cannot be transitioned (e.g., already suspended). We log the failure and continue processing remaining users.

app/modules/user/risk.rb
```diff
@@ -122,17 +122,34 @@ def enable_links_and_tell_chat
 
   def suspend_sellers_other_accounts
     SuspendAccountsWithPaymentAddressWorker.perform_in(5.seconds, id)
+    SuspendAccountsWithStripeFingerprintWorker.perform_in(5.seconds, id)
```
FusionBrah, line 127: Added the Stripe fingerprint worker alongside the existing PayPal payment address worker. Both run with a 5-second delay to allow the admin suspension to complete first.

app/modules/user/risk.rb
```diff
@@ -122,17 +122,34 @@ def enable_links_and_tell_chat
...
   end
 
   def block_seller_ip!
     BlockSuspendedAccountIpWorker.perform_in(5.seconds, id)
   end
 
   def enable_sellers_other_accounts
-    return if payment_address.blank?
+    if payment_address.present?
+      User.where(payment_address:).where.not(id:).find_each do |user|
```
FusionBrah, line 138: Using `find_each` for batch processing (per CLAUDE.md guidelines) since we could be iterating over many users. Also using non-bang `mark_compliant` with logging to gracefully handle users who cannot be transitioned.

app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
```diff
@@ -292,6 +292,8 @@ def self.save_stripe_bank_account_info(bank_account, stripe_account)
     bank_account.stripe_external_account_id = stripe_external_account.id
     bank_account.stripe_fingerprint = stripe_external_account.fingerprint
     bank_account.save!
+
+    CheckStripeFingerprintWorker.perform_async(bank_account.user_id, bank_account.stripe_fingerprint)
```
FusionBrah, line 302: Triggering the fingerprint check here (after the fingerprint is saved) rather than on bank account creation. The fingerprint is populated asynchronously from the Stripe API response, so it doesn't exist when the bank account is first created.

db/migrate/20260108150000_add_index_to_bank_accounts_stripe_fingerprint.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIndexToBankAccountsStripeFingerprint < ActiveRecord::Migration[7.1]
+  def change
```
FusionBrah, line 4: Adding an index on `stripe_fingerprint` to optimize the queries in both workers. The `credit_cards` table already has an index on its fingerprint column, so this aligns with that pattern.

config/initializers/blocked_object_types.rb
```diff
@@ -6,5 +6,6 @@
   email: "email",
   email_domain: "email_domain",
   charge_processor_fingerprint: "charge_processor_fingerprint",
```
FusionBrah, line 8: Adding a new blocked object type so admins can manually block specific Stripe bank account fingerprints. This mirrors the existing `charge_processor_fingerprint` type used for credit cards.

app/sidekiq/check_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+      Rails.logger.info("CheckStripeFingerprintWorker: User #{user_id} cannot be flagged (state: #{user.user_risk_state}), skipping")
+      return
+    end
+
+    if stripe_fingerprint.blank?
+      Rails.logger.info("CheckStripeFingerprintWorker: Stripe fingerprint blank for user #{user_id}, skipping")
+      return
+    end
+
+    suspended_users_with_same_fingerprint = BankAccount
```
FusionBrah: Checking ALL bank accounts (not using `.alive` scope here) preserves fraud history - even if a suspended user deletes their bank account, new users with the same fingerprint will still be caught.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+  def perform(user_id)
+    suspended_user = User.find(user_id)
+
+    fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
+    if fingerprints.empty?
+      Rails.logger.info("SuspendAccountsWithStripeFingerprintWorker: User #{user_id} has no bank accounts with fingerprints, skipping")
+      return
+    end
+
+    user_fingerprints_map = BankAccount.alive
```
FusionBrah: Building a `user_id → fingerprints` map upfront eliminates N+1 queries. Previously, each iteration queried `user.alive_bank_accounts.pluck(:stripe_fingerprint)` separately.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class SuspendAccountsWithStripeFingerprintWorker
+  include Sidekiq::Job
+  sidekiq_options retry: 5, queue: :default, lock: :until_executed
+
+  def perform(user_id)
+    suspended_user = User.find(user_id)
+
+    fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
```
FusionBrah: Checking ALL bank accounts from the suspended user (line 10, no `.alive` scope) preserves fraud history. If a fraudster deletes their bank account after being suspended, we still find and suspend related accounts.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
FusionBrah: **⚠️ This comment is now outdated.**

The implementation was simplified during code review fixes. Instead of building a user_id → fingerprints map, we now just use `active_bank_account&.stripe_fingerprint` (line 9). This matches the existing `SuspendAccountsWithPaymentAddressWorker` pattern.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
FusionBrah: **⚠️ This comment is now outdated.**

The implementation was simplified. We now use `active_bank_account&.stripe_fingerprint` instead of checking all bank accounts. This matches the existing `SuspendAccountsWithPaymentAddressWorker` pattern.

app/modules/user/risk.rb
FusionBrah: **⚠️ This comment is now outdated.**

The implementation now uses `mark_compliant!` (bang method) with `rescue StandardError` blocks instead of non-bang methods. This provides better error handling - exceptions are caught and logged, allowing the loop to continue processing remaining users.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
FusionBrah: **✅ Updated in latest commit**

The logic now includes `user.flagged_for_fraud? ||` check (line 24). This handles users already in `flagged_for_fraud` state - they skip the flagging step and proceed directly to suspension. This fixes a fraud bypass where previously-flagged users would be skipped entirely.

Also added:
- `.not_suspended` scope (line 20) to match existing `SuspendAccountsWithPaymentAddressWorker` pattern
- `skip_transition_callback: :suspend_sellers_other_accounts` (line 35) to prevent infinite job cascading

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+    if stripe_fingerprint.blank?
+      Rails.logger.info("SuspendAccountsWithStripeFingerprintWorker: User #{user_id} has no active bank account or fingerprint, skipping")
+      return
+    end
+
+    User.joins(:bank_accounts)
+        .merge(BankAccount.alive)
+        .where(bank_accounts: { stripe_fingerprint: stripe_fingerprint })
+        .where.not(id: suspended_user.id)
+        .not_suspended
```
FusionBrah, line 20: **Code Review Fix:** Added `.not_suspended` scope to match the existing `SuspendAccountsWithPaymentAddressWorker` pattern (line 12). This avoids reprocessing users who are already suspended.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+        content: "Flagged for fraud automatically on #{Time.current.to_fs(:formatted_date_full_month)} because of usage of bank account fingerprint from User##{suspended_user.id}"
+      )
+        Rails.logger.warn("SuspendAccountsWithStripeFingerprintWorker: Failed to flag user #{user.id} (state: #{user.user_risk_state})")
+        next
+      end
+
+      unless user.suspend_for_fraud(
+        author_name: "suspend_sellers_other_accounts",
+        content: "Suspended for fraud automatically on #{Time.current.to_fs(:formatted_date_full_month)} because of usage of bank account fingerprint from User##{suspended_user.id}",
+        skip_transition_callback: :suspend_sellers_other_accounts
```
FusionBrah, line 35: **Code Review Fix:** Added `skip_transition_callback: :suspend_sellers_other_accounts` to prevent infinite job cascading. Without this, suspending user B would trigger another worker enqueue, creating a loop.

app/modules/user/risk.rb
```diff
@@ -133,10 +134,26 @@ def block_seller_ip!
   def enable_sellers_other_accounts(transition)
     return if transition.args.first&.dig(:skip_transition_callback) == __method__
 
-    return if payment_address.blank?
+    if payment_address.present?
+      User.where(payment_address:).where.not(id:).find_each do |user|
+        user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as payment address #{payment_address} is now unblocked (from User##{id})", skip_transition_callback: :enable_sellers_other_accounts)
+      rescue StandardError => e
```
FusionBrah, line 140: **Code Review Fix:** Added `rescue StandardError` block so that if one user fails to be marked compliant, the loop continues processing remaining users. Errors are logged for debugging.

EmCousin: This will cause the job to skip users and complete without retry. Rails logs will be difficult to retrieve, so I'd rather ensure this job fails, retries, and most-of-all, is idempotent

app/modules/user/risk.rb
```diff
@@ -133,10 +134,26 @@ def block_seller_ip!
...
+        Rails.logger.error("enable_sellers_other_accounts: Failed to mark user #{user.id} compliant: #{e.message}")
+      end
+    end
 
-    User.where(payment_address:).where.not(id:).each do |user|
-      user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as payment address #{payment_address} is now unblocked (from User##{id})", skip_transition_callback: :enable_sellers_other_accounts)
+    stripe_fingerprint = active_bank_account&.stripe_fingerprint
```
EmCousin, line 145: A user may have multiple alive bank accounts. Checking only the active one is not enough

app/sidekiq/check_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    if !user.can_flag_for_fraud?
+      Rails.logger.info("CheckStripeFingerprintWorker: User #{user_id} cannot be flagged (state: #{user.user_risk_state}), skipping")
+      return
+    end
+
+    if stripe_fingerprint.blank?
+      Rails.logger.info("CheckStripeFingerprintWorker: Stripe fingerprint blank for user #{user_id}, skipping")
+      return
+    end
```
EmCousin, line 23: You could early-return this if blank so the job doesn't have to query the DB to retrieve the user for nothing.

app/sidekiq/suspend_accounts_with_stripe_fingerprint_worker.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class SuspendAccountsWithStripeFingerprintWorker
+  include Sidekiq::Job
+  sidekiq_options retry: 5, queue: :default, lock: :until_executed
+
+  def perform(user_id)
+    suspended_user = User.find(user_id)
+    stripe_fingerprint = suspended_user.active_bank_account&.stripe_fingerprint
```
EmCousin, line 9: We need to check all the user's bank accounts, alive or not. If a fraudster deletes their bank accounts after using it, we should still look those up.

config/initializers/blocked_object_types.rb
EmCousin: `charge_processor_fingerprint` has a general purpose already, you don't need a specific type for Stripe

CLAUDE.md
```diff
@@ -0,0 +1,162 @@
+# CLAUDE.md
```
EmCousin, line 1: Please delete this.

Issue #2757, opened 2026-01-08
Double scroll area on product list page (mobile only)


nyomanjyotisa:
URL: https://gumroad.com/products

On mobile, the products list page shows two vertical scroll areas

https://github.com/user-attachments/assets/9c5edd90-43ed-41bd-8665-e72f2fbec2db

EmCousin:
Resolved in https://github.com/antiwork/gumroad/pull/2893

PR #2761, opened 2026-01-08
fix double scroll on mobile

4arjun:
ref #2757 

## Summary

On mobile, the products list page showed two nested vertical scroll areas
- Changed the main layout shell from fixed `h-screen` to `min-h-screen` on mobile, keeping `h-screen` only for desktop (`lg:`)
- Removed `overflow-y-auto` from the main content area on mobile, allowing natural document scroll instead of a contained scroll area
This ensures mobile users get a single, native scroll experience while preserving the existing desktop sidebar layout.

## Before

https://github.com/user-attachments/assets/715e3032-b282-4545-a6b8-c0feeaacbe9e

## After

https://github.com/user-attachments/assets/9bc49bfc-d012-4117-a0ca-ecf65f744487

## Checklist

- [ ] I have updated tests for my changes

---

## AI Disclosure

No AI used.

### DIFF
--- a/app/javascript/inertia/layout.tsx
+++ b/app/javascript/inertia/layout.tsx
@@ -41,7 +41,7 @@ export default function Layout({ children }: { children: React.ReactNode }) {
       <CurrentSellerProvider value={parseCurrentSeller(current_seller)}>
         <Head title={title} />
         <Alert initial={null} />
-        <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
+        <div id="inertia-shell" className="flex min-h-screen flex-col lg:h-screen lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
           {isRouteLoading ? <LoadingSkeleton /> : null}
           <main className={classNames("flex-1 lg:overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>


app/javascript/inertia/layout.tsx
```diff
@@ -44,10 +44,10 @@ export default function Layout({ children }: { children: React.ReactNode }) {
       <CurrentSellerProvider value={parseCurrentSeller(current_seller)}>
         <Head title={title} />
         <Alert initial={flash ?? null} />
-        <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
+        <div id="inertia-shell" className="flex min-h-screen flex-col lg:h-screen lg:flex-row">
```
4arjun: Previously, the shell used `h-screen` at all breakpoints, which fixed the container height to 100vh. On mobile, this created a fixed-height container that, combined with `overflow-y-auto` on `<main>`, resulted in a nested scrollable area inside the viewport.

Fix:
Uses `min-h-screen` allowing natural document height/scroll

app/javascript/inertia/layout.tsx
```diff
@@ -44,10 +44,10 @@ export default function Layout({ children }: { children: React.ReactNode }) {
...
           {logged_in_user ? <Nav title="Dashboard" /> : null}
           {isRouteLoading ? <LoadingSkeleton /> : null}
-          <main className={classNames("flex-1 overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>
+          <main className={classNames("flex-1 lg:overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>
```
4arjun, line 47: No overflow set, so content scrolls naturally with the document body

4arjun:
@nyomanjyotisa Could you please take a look at this?

app/javascript/inertia/layout.tsx
nyomanjyotisa: @4arjun Do you have any idea why this issue doesn’t occur on https://gumroad.com/customers even though it uses the same layout?

Also, the admin dashboard uses the similar shell with `h-screen` and `overflow-y-auto` on `<main>` , and no issue

https://github.com/antiwork/gumroad/blob/50fd48b41e71f9d4dc53321cf546bb3c294e1ced/app/javascript/layouts/Admin.tsx#L29-L33

Feels like this is not the root cause

4arjun: > @4arjun Do you have any idea why this issue doesn’t occur on https://gumroad.com/customers even though it uses the same layout?
> 
> Also, the admin dashboard uses the similar shell with `h-screen` and `overflow-y-auto` on `<main>` , and no issue

I couldn’t find a single obvious difference for why it doesn’t reproduce on /customers or the admin dashboard.

nyomanjyotisa: Hi @EmCousin could you help take a look here?

The solution in this PR does solve the issue, but it feels like a workaround than addressing the root cause. The same problem doesn’t reproduce on /customers or the admin dashboard, even with the same or very similar layout.

Do you think we should merge this, or would it be better to close it?

EmCousin: Feels indeed like a workaround and that we didn't get to the bottom of things here. I'd close this since this bug doesn't appear to be critical, we can afford to wait for a solution that addresses the root cause in my opinion.

PR #2768, opened 2026-01-08
Allow PayPal checkout tests to run without real credentials

### LINKED ISSUE #2699: Fix PayPal checkout test to run without secrets
### Problem

The test `"shows native, braintree, or no paypal button depending on availability"` in `spec/requests/checkout/payment_spec.rb` checks for `iframe[title=PayPal]`. This iframe is rendered by PayPal's JavaScript SDK, which only loads with real credentials.

Unlike Stripe (where we use real sandbox keys in `.env.test`), the current dummy `PAYPAL_PARTNER_CLIENT_ID=test-paypal-partner-client-id` won't work.

### Options

1. **Add real PayPal sandbox credentials** to `.env.test` (like Stripe) - keeps full integration coverage
2. **Skip the iframe assertion** when credentials aren't available - test still exists but doesn't block CI
3. **Remove the iframe assertion entirely** - loses verification that native PayPal works differently from Braintree fallback

### Tradeoff

We can't have both full native PayPal integration testing AND no external credentials. Options 2 and 3 reduce coverage. Option 1 requires setting up PayPal sandbox credentials.

The question is: is verifying the native PayPal iframe critical, or is it acceptable to trust that if our code passes the right props, PayPal's SDK will do its job?

gianfrancopiana:
Fixes #2699

## Problem

Checkout and subscription page tests fail when `PAYPAL_PARTNER_CLIENT_ID` is missing because the page crashes with "Expected root.paypal_client_id to be a string, got null" before rendering.

There's also one test checks for a PayPal iframe that only renders with real PayPal SDK credentials.

## Solution

- Add a dummy `PAYPAL_PARTNER_CLIENT_ID` to `.env.test` so pages render in tests.
- Replace the iframe assertion with prop verification - check that `supports_paypal` is correctly set to `"native"` when a merchant has PayPal connected. This tests our code (passing the right props) rather than PayPal's SDK (rendering an iframe).

---

# Before/After

N/A - no UI changes

---

# Test Results

Baseline run (without changes): 3 tests in `checkout/payment_spec.rb` failed
- `shows native, braintree, or no paypal button depending on availability` ❌
- `disables the payment button until typo suggestion is resolved` ❌  
- `does not block the payment button` ❌

With changes: All 3 tests pass ✅

CI comparison:
- Baseline: https://github.com/antiwork/gumroad/actions/runs/20790087323
- With fix: https://github.com/antiwork/gumroad/actions/runs/20821977716

---

# AI Disclosure

### DIFF
--- a/spec/requests/checkout/payment_spec.rb
+++ b/spec/requests/checkout/payment_spec.rb
@@ -18,7 +18,7 @@ describe "Checkout payment", :js, type: :system do
       expect(supports_paypal).to eq("native")
     else
       expect(page).to have_selector("iframe[title=PayPal]")
-    end  
+    end
     expect(supports_paypal).to eq("native")
 
     product2 = create(:product, price_cents: 1000)


spec/requests/checkout/payment_spec.rb
```diff
@@ -13,7 +13,8 @@
     visit "/l/#{@product.unique_permalink}"
     add_to_cart(@product)
     select_tab "PayPal"
-    expect(page).to have_selector("iframe[title=PayPal]")
+    supports_paypal = page.evaluate_script("JSON.parse(document.querySelector('script.js-react-on-rails-component[data-component-name=\"CheckoutPage\"]').textContent).add_products[0].product.supports_paypal")
```
hchhabra: Should we skip the iframe check only when the ID is test/missing? Might be better to test the actual PayPal iframe load when possible (CI/local with real ID).

```suggestion
    if PAYPAL_PARTNER_CLIENT_ID.blank? || PAYPAL_PARTNER_CLIENT_ID.start_with?("test-")
      supports_paypal = page.evaluate_script("JSON.parse(document.querySelector('script.js-react-on-rails-component[data-component-name=\"CheckoutPage\"]').textContent).add_products[0].product.supports_paypal")
      expect(supports_paypal).to eq("native")
    else
      expect(page).to have_selector("iframe[title=PayPal]")
    end  
```

PR #2772, opened 2026-01-08
Auto-restart deactivated subscriptions on membership re-purchase

yashranaway:
Issue: #173

# Description

The original approach ([PR #2347](https://github.com/antiwork/gumroad/pull/2347)) showed a modal asking users whether to restart or create a new subscription. After discussion with @binary-koan  ([comment](https://github.com/antiwork/gumroad/pull/2347#issuecomment-3724441879)), we realized there's no valid use case where a user would want a fresh subscription instead of restarting their existing one. A modal adds unnecessary friction without providing real value. The better UX is to automatically restart the existing subscription on checkout.

## Problem

The team determined that there's no valid use case for a user to start a completely fresh subscription when they already have a deactivated one for the same product. The better UX is to automatically restart their existing subscription.

## Solution

Modified the `create_subscription` method in `Purchase::BaseService` to check for an existing restartable subscription before creating a new one. If found, it reactivates the existing subscription instead.

A subscription is considered restartable when:
- It's deactivated (payment failed or user cancelled)
- It hasn't ended (no `ended_at` timestamp)
- It wasn't cancelled by the seller (seller cancellations are intentional blocks)
- It's not a test subscription

The implementation:
1. Looks up existing subscriptions by user ID first, then falls back to email matching for guest checkouts
2. Picks the most recently deactivated subscription if multiple exist
3. Updates the credit card, calls `resubscribe!` to clear cancellation flags, and sends restart notifications
4. Properly handles edge cases: gift purchases, test purchases, deleted products, installment plans

---

# Before/After

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

### DIFF
--- a/app/services/purchase/base_service.rb
+++ b/app/services/purchase/base_service.rb
@@ -96,15 +96,22 @@ class Purchase::BaseService
 
       return nil if email.blank?
 
-      base_scope.find { |s| s.email&.downcase == email.downcase }
+      original_purchase_flag = Purchase.flag_mapping["flags"][:is_original_subscription_purchase]
+      base_scope
+        .joins("INNER JOIN purchases ON purchases.subscription_id = subscriptions.id AND (purchases.flags & #{original_purchase_flag}) != 0")
+        .where("LOWER(purchases.email) = ?", email.downcase)
+        .first
     end
 
     def restart_existing_subscription(subscription, giftee_purchase)
       is_gift = purchase.is_gift_sender_purchase
 
-      subscription.credit_card = purchase.credit_card unless is_gift
-      subscription.resubscribe!
-      subscription.purchases << [purchase, giftee_purchase].compact
+      ActiveRecord::Base.transaction do
+        subscription.credit_card = purchase.credit_card unless is_gift
+        subscription.resubscribe!
+        subscription.purchases << [purchase, giftee_purchase].compact
+      end
+
       subscription.send_restart_notifications!
     end
 


app/services/purchase/base_service.rb
```diff
@@ -30,6 +30,12 @@ def handle_purchase_success
     def create_subscription(giftee_purchase)
       return if purchase.subscription.present?
 
+      existing_subscription = find_restartable_subscription(giftee_purchase)
+      if existing_subscription
+        restart_existing_subscription(existing_subscription, giftee_purchase)
+        return
+      end
```
yashranaway, line 37: This is the Entry point for auto-restart logic. We check for a restartable subscription first before falling through to create a new one. This keeps the existing subscription creation code untouched.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
       subscription.purchases << [purchase, giftee_purchase].compact
     end
 
+    def find_restartable_subscription(giftee_purchase)
+      return nil if purchase.is_installment_payment
+      return nil if purchase.is_test_purchase?
+      return nil if purchase.link.deleted?
```
yashranaway, line 79: Excluding installment plans (different billing model), test purchases (seller testing their own product), and deleted products from auto-restart to avoid edge case issues.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+      is_gift = purchase.is_gift_sender_purchase
+      user = is_gift ? giftee_purchase&.purchaser : purchase.purchaser
+      email = is_gift ? giftee_purchase&.email : purchase.email
+
+      base_scope = purchase.link.subscriptions
+        .not_is_test_subscription
+        .where.not(deactivated_at: nil)
+        .where(ended_at: nil)
+        .where("cancelled_at IS NULL OR (subscriptions.flags & ?) != 0", Subscription.flag_mapping["flags"][:cancelled_by_buyer])
+        .order(deactivated_at: :desc)
```
yashranaway, line 90: Core logic: must be deactivated but not ended, and not cancelled by seller. The cancelled_by_buyer flag check ensures we only exclude seller cancellations - if buyer cancelled, they can restart. Ordering by most recent deactivation handles users with multiple cancelled subscriptions.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+
+      if user.present?
+        subscription = base_scope.find_by(user: user)
+        return subscription if subscription
+      end
+
+      return nil if email.blank?
+
+      base_scope.find { |s| s.email&.downcase == email.downcase }
+    end
```
yashranaway, line 104: Prioritizes user ID match for logged-in users, fall back to email for guest checkouts. Email comparison is case-insensitive since emails can vary in casing across purchases.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+
+    def restart_existing_subscription(subscription, giftee_purchase)
+      is_gift = purchase.is_gift_sender_purchase
+
+      subscription.credit_card = purchase.credit_card unless is_gift
+      subscription.resubscribe!
+      subscription.purchases << [purchase, giftee_purchase].compact
+      subscription.send_restart_notifications!
```
yashranaway: Using existing resubscribe! method which handles clearing cancellation flags and rescheduling workflows. Credit card update ensures future charges use the new payment method. Notifications go to both customer and creator.

spec/services/purchase/base_service_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+
+        def initialize(purchase)
+          @purchase = purchase
+        end
+
+        def perform
+          create_subscription(nil)
+        end
+      end
+    end
```
yashranaway, line 23: Creating a minimal wrapper to test create_subscription directly since the factory doesn't go through the full service flow. This isolates the subscription creation logic we're testing.

spec/services/purchase/base_service_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      end
+
+      it "creates a new subscription instead of restarting" do
+        initial_count = product.subscriptions.count
+
+        service_class.new(new_purchase).perform
+
+        expect(product.subscriptions.count).to eq(initial_count + 1)
+        expect(new_purchase.reload.subscription).not_to eq(existing_subscription)
+      end
```
yashranaway, line 93: Key test: seller cancellations should NOT restart. Note we only set cancelled_by_buyer: false (not cancelled_by_admin: true) since that's how cancel!(by_seller: true) actually works in production.

spec/services/purchase/base_service_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+
+      it "restarts the existing subscription" do
+        initial_count = product.subscriptions.count
+
+        service_class.new(new_purchase).perform
+
+        expect(product.subscriptions.count).to eq(initial_count)
+        expect(new_purchase.reload.subscription).to eq(existing_subscription)
+        expect(existing_subscription.reload.deactivated_at).to be_nil
+      end
```
yashranaway, line 126: Opposite case: buyer-initiated cancellations SHOULD restart. This ensures the SQL flag check correctly distinguishes between seller and buyer cancellations.

spec/services/purchase/base_service_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      it "restarts the most recently deactivated subscription" do
+        initial_count = product.subscriptions.count
+
+        service_class.new(new_purchase).perform
+
+        expect(product.subscriptions.count).to eq(initial_count)
+        expect(new_purchase.reload.subscription).to eq(newer_subscription)
+        expect(newer_subscription.reload.deactivated_at).to be_nil
+        expect(older_subscription.reload.deactivated_at).to be_present
+      end
```
yashranaway, line 232: Ensures we pick the most recently deactivated subscription when a user has multiple. Verifies the order(deactivated_at: :desc) is working correctly.

yashranaway:
@binary-koan Friendly ping 😅

binary-koan:
Thanks for doing this, just read through it and I think we need to nail down the behaviour a bit more before we look at PRs (this was super helpful in working out what we need to consider though). Apologies once again for not planning out the issue before opening it to contributions ...

I've commented on the issue, please keep an eye on it and we'll try to get it fleshed out properly

PR #2798, opened 2026-01-09
Add a separate download page for mobile apps

binary-koan:
# Description

## Problem

Currently the mobile apps use the regular download page at `/d/:token` to load product content. However, it feels very clunky because

* We run a CloudFlare JS challenge on every page load, which adds several seconds to the load time (and may not complete correctly in simulators, making it hard to test the app)
* The user isn't logged in in the webview, so we run into the email confirmation check and show an extra form to confirm the email before viewing product details

## Solution

Added a separate `/mobile_download_page/:token` endpoint which authenticates with an OAuth access token, then skips both of those checks and immediately shows the content.

### Before

https://github.com/user-attachments/assets/503e537d-636b-4dee-97d2-f3f79c4d1099

### After

(loads instantly with no additional steps)

https://github.com/user-attachments/assets/4a703369-d97e-4b92-979c-634f3d9bf6b4

---

# AI Disclosure

Used Claude Opus 4.5 for exploring the codebase and refactoring specs.


app/controllers/concerns/current_api_user.rb
```diff
@@ -16,4 +16,8 @@ def current_api_user
   rescue ActionDispatch::Http::Parameters::ParseError
     @_current_api_user = nil
   end
+
+  def current_resource_owner
```
binary-koan, line 20: Needed in `UrlRedirectsController` now. I don't think it needs to be restricted to the mobile controllers since we already have `current_api_user` available everywhere

app/controllers/url_redirects_controller.rb
```diff
@@ -9,24 +9,25 @@ class UrlRedirectsController < ApplicationController
     audio_durations
   ]
   before_action :redirect_to_custom_domain_if_needed, only: :download_page
-  before_action :redirect_bundle_purchase_to_library_if_needed, only: :download_page
-  before_action :redirect_to_coffee_page_if_needed, only: :download_page
-  before_action :check_permissions, only: %i[show stream download_page
+  before_action :sign_in_mobile_app_user, only: :mobile_download_page
+  before_action :redirect_bundle_purchase_to_library_if_needed, only: %i[download_page mobile_download_page]
+  before_action :redirect_to_coffee_page_if_needed, only: %i[download_page mobile_download_page]
```
binary-koan, line 14: Redirecting isn't ideal in a webview (just because it shows pages that aren't designed for it), but it's fine as a fallback if the library API isn't filtering out these products correctly

app/controllers/url_redirects_controller.rb
```diff
@@ -340,6 +347,12 @@ def fetch_url_redirect
       e404 if !has_files && !can_view_product_download_page_without_files
     end
 
+    def sign_in_mobile_app_user
+      doorkeeper_authorize! :mobile_api
```
binary-koan, line 351: Doorkeeper supports sending the token in query params, so we can load the webview with `/mobile_download_page/URL_REDIRECT_TOKEN?access_token=ACCESS_TOKEN` to display it without needing any redirects

We could arguably create a separate controller to avoid using Doorkeeper outside the API controllers, but I think this will make it easier to keep the pages in sync if we modify the regular download page

spec/controllers/concerns/current_api_user_spec.rb
```diff
@@ -67,4 +67,38 @@ def action
...
+    end
+
+    context "when impersonating a user" do
+      let(:user) { create(:user) }
+
+      before do
+        @request.params["access_token"] = access_token
+      end
+
+      it "returns the impersonated user" do
```
binary-koan, line 96: Moved from `spec/controllers/api/mobile/base_controller_spec.rb`, this was the only spec in that file

spec/controllers/url_redirects_controller_spec.rb
```diff
@@ -648,6 +648,104 @@
     end
   end
 
+  describe "GET 'mobile_download_page'" do
```
binary-koan, line 651: Roughly copied from the `GET 'download_page'` specs

binary-koan (commented):
@ershad _a Could you review this? (or assign someone else, but would be nice to have your eyes on it as URL redirects are a critical feature)

EmCousin:
@binary-koan Do you think we could skip the parts we don't want for mobile by checking the device directly in the existing endpoint? https://github.com/fnando/browser

binary-koan:
Surely that can be spoofed? - an attacker could use mobile user agents etc to bypass protections?

The main reason for adding an endpoint instead of updating the existing one is to bypass the CloudFlare challenge, which is set on their firewall so can’t check auth etc (only things like user agent which aren’t necessarily reliable AFAIK)

binary-koan:
Note: I’m unsure why we’re quite this strict, but I assume there’s a reason for each check

ershad:
(discussed internally, as it touches on security)

binary-koan:
Superseded by https://github.com/antiwork/gumroad/pull/2820

PR #2801, opened 2026-01-09
Migrate product edit page to inertia

MuskanPaliwal:
### Description

This PR migrates product edit page to inertia js. 

### Problem
The Product Edit page (`/products/:id/edit)` was using React-on-Rails server-side rendering with an ERB wrapper, causing full-reload on each visit on the page and inconsistent response patterns

#### Backend:
  - Controller edit action now renders inertia: "`Products/Edit`" instead of ERB
  - Controller update action returns 303 redirects with flash messages instead of JSON

  #### Frontend:
  - Created new Inertia page wrapper at pages`/Products/Edit.tsx`
  - Replaced custom `saveProduct()` API call with Inertia's `useForm` hook
  - Existing React components (`ProductEditPage`, tabs, etc.) remain unchanged

  #### Cleanup:
  - Deleted `app/views/links/edit.html.erb`
  - Deleted `app/javascript/data/product_edit.ts` custom API logic replaced by Inertia form handling
  - Excluded Inertia from SSR bundle (uses browser APIs)

### Before

https://github.com/user-attachments/assets/60fcea58-bb5c-43f2-94cd-a34e776e0003

### After

https://github.com/user-attachments/assets/7297ae01-986a-4878-849a-2c7bc672d104

### Tests

The two test case failures are independent of this PR.

Screenshot 2026-01-09 at 10 40 38 PM

### Checklist

### AI Disclosure

I have used Claude Opus 4.5 to update test cases, and for initial drafting. Rest all the code has been manually reviewed by me.

### DIFF
--- a/app/javascript/components/ProductEdit/Provider.tsx
+++ b/app/javascript/components/ProductEdit/Provider.tsx
@@ -6,6 +6,7 @@ import * as React from "react";
 import { cast } from "ts-safe-cast";
 
 import { OtherRefundPolicy } from "$app/data/products/other_refund_policies";
+import { useDropboxDropins } from "$app/hooks/useDropboxDropins";
 import { Thumbnail } from "$app/data/thumbnails";
 import { RatingsWithPercentages } from "$app/parsers/product";
 import { CurrencyCode } from "$app/utils/currency";
@@ -101,6 +102,8 @@ export const ProductEditProvider = ({ children, currentTab }: { children: React.
 
   const form = useForm({});
   const publishForm = useForm({});
+
+  const { choose: dropboxChoose, save: dropboxSave } = useDropboxDropins(dropbox_app_key);
 
   const [imagesUploading, setImagesUploading] = React.useState<Set<File>>(new Set());
 
@@ -237,6 +240,8 @@ export const ProductEditProvider = ({ children, currentTab }: { children: React.
       publishing: publishForm.processing,
       publish,
       unpublish,
+      dropboxChoose,
+      dropboxSave,
     }),
     [
       product,
@@ -248,6 +253,8 @@ export const ProductEditProvider = ({ children, currentTab }: { children: React.
       filesById,
       dropbox_app_key,
       edit_props,
+      dropboxChoose,
+      dropboxSave,
     ],
   );
 
--- a/app/javascript/components/ProductEdit/ContentTab/FileEmbedGroup.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/FileEmbedGroup.tsx
@@ -35,6 +35,7 @@ type FileGroupConfig = {
   variantId: string | null;
   prepareDownload: () => Promise<void>;
   filesById: Map<string, FileEntry>;
+  dropboxSave: (options: Parameters<typeof Dropbox.save>[0]) => void;
 };
 type FileEmbedGroupStorage = { lastCreatedUid: string | null };
 
@@ -128,7 +129,7 @@ const FileEmbedGroupNodeView = ({
           product_file_ids: downloadableFiles.map((file) => file.id),
         }),
       );
-      if (fileDownloadInfos.length > 0) Dropbox.save({ files: fileDownloadInfos });
+      if (fileDownloadInfos.length > 0) config.dropboxSave({ files: fileDownloadInfos });
     } catch (e) {
       assertResponseError(e);
       showAlert(e.message, "error");
--- a/app/javascript/components/ProductEdit/ContentTab/index.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/index.tsx
@@ -92,7 +92,7 @@ export const extensions = (productId: string, extraExtensions: TiptapNode[] = []
 ];
 
 const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string | null }) => {
-  const { id, product, updateProduct, seller, save, existingFiles, setExistingFiles, uniquePermalink, filesById } =
+  const { id, product, updateProduct, seller, save, existingFiles, setExistingFiles, uniquePermalink, filesById, dropboxChoose, dropboxSave } =
     useProductEditContext();
   const uid = React.useId();
   const isDesktop = useIsAboveBreakpoint("lg");
@@ -211,6 +211,7 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string | 
     variantId: selectedVariantId,
     prepareDownload: save,
     filesById,
+    dropboxSave,
   });
   const fileEmbedConfig = useRefToLatest<FileEmbedConfig>({ filesById });
   const uploadFilesRef = useRefToLatest(uploadFiles);
@@ -410,7 +411,7 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string | 
       window.___dropbox_files_picked = null;
       return;
     }
-    window.Dropbox.choose({ linkType: "direct", multiselect: true, success: (files) => void uploadFiles(files) });
+    dropboxChoose({ linkType: "direct", multiselect: true, success: (files) => void uploadFiles(files) });
   };
   React.useEffect(() => {
     const interval = setInterval(
--- a/app/javascript/components/ProductEdit/state.ts
+++ b/app/javascript/components/ProductEdit/state.ts
@@ -198,6 +198,8 @@ export const ProductEditContext = React.createContext<{
   publishing: boolean;
   publish: () => void;
   unpublish: () => void;
+  dropboxChoose: (options: Parameters<typeof Dropbox.choose>[0]) => void;
+  dropboxSave: (options: Parameters<typeof Dropbox.save>[0]) => void;
 } | null>(null);
 export const useProductEditContext = () => assertDefined(React.useContext(ProductEditContext));
 


app/controllers/links_controller.rb
```diff
@@ -325,11 +325,12 @@ def edit
     fetch_product_by_unique_permalink
     authorize @product
 
-    redirect_to bundle_path(@product.external_id) if @product.is_bundle?
+    return redirect_to bundle_path(@product.external_id) if @product.is_bundle?
```
MuskanPaliwal, line 328: Added explicit `return` before redirect_to for bundle products. Without the return, the code would continue executing and try to render the Inertia page after the redirect.

app/controllers/links_controller.rb
```diff
@@ -325,11 +325,12 @@ def edit
...
 
     @title = @product.name
+    presenter = ProductPresenter.new(product: @product, pundit_user:)
 
-    @presenter = ProductPresenter.new(product: @product, pundit_user:)
+    render inertia: "Products/Edit", props: presenter.edit_props
```
MuskanPaliwal: Migrated from ERB view rendering to Inertia. The presenter's `edit_props` method already returns all the data needed, we just pass it directly as Inertia props now.

app/controllers/links_controller.rb
```diff
@@ -370,9 +371,12 @@ def update
         @product.save_tags!(product_permitted_params[:tags] || [])
         @product.reorder_previews((product_permitted_params[:covers] || []).map.with_index.to_h)
         if !current_seller.account_level_refund_policy_enabled?
-          @product.product_refund_policy_enabled = product_permitted_params[:product_refund_policy_enabled]
-          if product_permitted_params[:refund_policy].present? && product_permitted_params[:product_refund_policy_enabled]
+          refund_policy_enabled = ActiveModel::Type::Boolean.new.cast(product_permitted_params[:product_refund_policy_enabled])
+          @product.product_refund_policy_enabled = refund_policy_enabled
+          if product_permitted_params[:refund_policy].present? && refund_policy_enabled
```
MuskanPaliwal, line 387: Added `ActiveModel::Type::Boolean.new.cast()` because Inertia sends boolean params as strings ("true"/"false") rather than actual booleans. This ensures consistent behavior regardless of how the param is received. 
Also added logic to destroy the refund policy when disabled previously this case wasn't handled and orphan policies could remain.

app/controllers/links_controller.rb
```diff
@@ -422,7 +427,7 @@ def update
         end
         @product.description = SavePublicFilesService.new(resource: @product, files_params: product_permitted_params[:public_files], content: @product.description).process
         @product.save!
-        toggle_community_chat!(product_permitted_params[:community_chat_enabled])
+        toggle_community_chat!(ActiveModel::Type::Boolean.new.cast(product_permitted_params[:community_chat_enabled]))
```
MuskanPaliwal, line 441: Same boolean casting fix as above. Inertia sends "true"/"false" strings.

app/services/product/variants_updater_service.rb
```diff
@@ -59,11 +59,14 @@ def clean_variants_params(params)
 
             # TODO: :product_edit_react cleanup
             if option[:price_difference_cents].present?
-              option[:price] = option[:price_difference_cents]
+              option[:price] = option[:price_difference_cents].to_f
```
MuskanPaliwal, line 62: Added `.to_f` for price_difference_cents to handle string values from Inertia form submission.

config/webpack/webpack.config.js
```diff
@@ -21,6 +21,10 @@ export default async () => {
         "braintree-web": "braintree-web/dist/browser",
         "react-dom/server": "react-dom/server.browser",
         "@rails/activestorage": false,
+        // Exclude Inertia from SSR bundle - it uses browser APIs not available in MiniRacer
+        // This is safe because Inertia APIs are only used in event handlers, not during render
+        "@inertiajs/react": false,
+        "@inertiajs/core": false,
```
MuskanPaliwal, line 27: Excluded Inertia packages from the SSR bundle. MiniRacer (used for server-side rendering) doesn't have browser APIs that Inertia requires.

  This is safe because:
  1. Inertia APIs are only called in event handlers (onClick, form submit), not during initial render
  2. The SSR render path uses StaticRouterProvider which doesn't trigger Inertia calls

MuskanPaliwal:
Hi @Pradumn27, could you please review the PR ?

app/controllers/links_controller.rb
EmCousin: You already have automatic casting when assigning `@product.product_refund_policy_enabled = product_permitted_params[:product_refund_policy_enabled]`, the value is cast to a proper boolean. So you could just rely on `@product.product_refund_policy_enabled` here.

app/controllers/links_controller.rb
```diff
@@ -325,11 +325,23 @@ def edit
...
-
-    @presenter = ProductPresenter.new(product: @product, pundit_user:)
+    presenter = ProductPresenter.new(product: @product, pundit_user:)
+
+    inertia_page = case params[:other]
+                   when "content" then "Products/Edit/Content"
+                   when "receipt" then "Products/Edit/Receipt"
+                   when "share" then "Products/Edit/Share"
+                   else "Products/Edit/Product"
+                   end
```
EmCousin, line 338: No. Each tab should be their own dedicated page served by their corresponding dedicated controller endpoint. - Code that is common across templates should be factored to a wrapper layout
- Each view should only receive the props they need
- There should be a corresponding `update` endpoint for each Content, Receipt and Share view.

This means there should be one controller per domain (Content, Receipt, and Share), each with an `edit` action that renders the require props, and an `update` endpoint that filtered the allowed params before updating what's only needed.

app/controllers/links_controller.rb
```diff
@@ -453,38 +470,62 @@ def update
...
     rescue => e
       Bugsnag.notify(e)
-      return render json: { success: false, error_message: "Something broke. We're looking into what happened. Sorry about this!" }
+      error_message = "Something broke. We're looking into what happened. Sorry about this!"
+      if request.inertia?
+        return redirect_to edit_link_tab_path(@product), alert: error_message, status: :see_other
+      else
+        return render json: { success: false, error_message: }
+      end
     end
```
EmCousin, line 522: Just collect the error message and then render the controller response once. That's a lot of boiler plate right now.
Also, you can use `respond_to` here:
```ruby
respond_to do |format|
  format.html { ... } # render with Inertia
  format.json { ... } # render with JSON
end
```

EmCousin: That being said, do we still need the JSON response?

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -247,20 +220,24 @@ export const Layout = ({
...
+          ) : currentTab === "product" && !isCoffee ? (
             <Button
               color="primary"
               disabled={isBusy}
-              onClick={() => void save().then(() => navigate.current(`${rootPath}/content`))}
+              onClick={() =>
+                void save().then(() =>
+                  router.visit(`${Routes.edit_link_path(uniquePermalink)}/content`, { preserveScroll: true }),
+                )
+              }
```
EmCousin, line 231: This redirection should be set server side once there is a dedicated `update` endpoint, see [comment](https://github.com/antiwork/gumroad/pull/2801/changes#r2682094848).

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -274,26 +251,31 @@ export const Layout = ({
...
+              <Link href={Routes.edit_link_path(uniquePermalink)} preserveScroll onClick={onTabClick}>
                 Product
               </Link>
             </Tab>
             {!isCoffee ? (
-              <Tab asChild isSelected={tab === "content"}>
-                <Link to={`${rootPath}/content`} onClick={onTabClick}>
+              <Tab asChild isSelected={currentTab === "content"}>
+                <Link
+                  href={`${Routes.edit_link_path(uniquePermalink)}/content`}
```
EmCousin, line 262: Like explained above, this needs to be a dedicated endpoint, then we can use the corresponding generated URL helper here.

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -274,26 +251,31 @@ export const Layout = ({
           )}
         >
           <Tabs style={{ gridColumn: 1 }}>
-            <Tab asChild isSelected={tab === "product"}>
-              <Link to={rootPath} onClick={onTabClick}>
+            <Tab asChild isSelected={currentTab === "product"}>
```
EmCousin, line 254: Like explained above, this should derive from which Inertia page we are rendering. It should not be one Inertia page in charge to render everything.

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -274,26 +251,31 @@ export const Layout = ({
...
+                  onClick={onTabClick}
+                >
                   Content
                 </Link>
               </Tab>
             ) : null}
-            <Tab asChild isSelected={tab === "receipt"}>
-              <Link to={`${rootPath}/receipt`} onClick={onTabClick}>
+            <Tab asChild isSelected={currentTab === "receipt"}>
+              <Link href={`${Routes.edit_link_path(uniquePermalink)}/receipt`} preserveScroll onClick={onTabClick}>
```
EmCousin, line 271: Same here.

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -274,26 +251,31 @@ export const Layout = ({
...
                 Receipt
               </Link>
             </Tab>
-            <Tab asChild isSelected={tab === "share"}>
+            <Tab asChild isSelected={currentTab === "share"}>
               <Link
-                to={`${rootPath}/share`}
+                href={`${Routes.edit_link_path(uniquePermalink)}/share`}
+                preserveScroll
```
EmCousin, line 278: Same here

app/javascript/components/ProductEdit/Provider.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  return {
+    sharedContentUpdated,
+    contentUpdatedVariantIds,
+  };
+};
+
+import { type ProductEditTab } from "./Layout";
+
+export const ProductEditProvider = ({ children, currentTab }: { children: React.ReactNode; currentTab: ProductEditTab }) => {
```
EmCousin, line 86: Same spirit than the other comments: this provider handles too much responsibility and needs to be lightened, and should keep only what is _strictly_ common to all pages.
Once there is once page per domain (Content, Product, Receipt, Share), each page should receive only the props they need, and should have proper types accordingly.

app/javascript/components/ProductEdit/Provider.tsx
```diff
@@ -0,0 +1,301 @@
...
+      ),
+      availabilities: filteredProduct.availabilities.map(({ newlyAdded, ...availability }) =>
+        newlyAdded ? { ...availability, id: null } : availability,
+      ),
+      installment_plan: filteredProduct.allow_installment_plan ? filteredProduct.installment_plan : null,
+      current_tab: currentTab,
+    }));
+
+    return new Promise<void>((resolve, reject) => {
+      form.post(Routes.link_path(edit_props.unique_permalink), {
```
EmCousin, line 159: Each page should have their own `update` endpoint, and each endpoint should be in charge of knowing where to redirect, which flash message to set, so we no longer have to handle the page updates here - they should be handled server side when the redirection is complete.

config/webpack/webpack.config.js
EmCousin: SSR-related code should not be containing Inertia-based code at all. If you find yourself having trouble rendering SSR components in this branch, this means some of them got injected Inertia-related code where they shouldn't have.

PR #2810, opened 2026-01-10
Auto-suspend creators with shared ACH bank accounts (stripe fingerprints)

yashranaway:
Fixes : #1158

# Description

## Problem

We currently auto-suspend accounts that share the same PayPal payout address as another suspended account. However, fraudsters are increasingly using shared ACH bank accounts (identified by Stripe fingerprints) across multiple accounts. When one account gets suspended, the related accounts using the same bank account slip through undetected, leading to more chargebacks and time spent hunting down fraudsters manually.

## Solution

Extended the existing PayPal-based auto-suspension system to also handle Stripe bank account fingerprints:

**When an admin suspends an account:**
- Find all other accounts using the same bank account fingerprint
- Flag them for fraud and suspend them automatically
- Add comments explaining why they were suspended and which account triggered it

**When a new account sets up a bank account:**
- Check if the fingerprint matches any suspended account's bank account
- Also check if the fingerprint is explicitly blocked
- Flag the account for fraud if there's a match

**When an account is marked compliant:**
- Related accounts with the same fingerprint are also marked compliant

I also addressed several edge cases that improve on the original PayPal implementation:
- Checks ALL bank account fingerprints (not just the active one) since a fraudster might have multiple accounts
- Preserves fraud history even if the suspended user deletes their bank account
- Skips already suspended users to avoid duplicate processing
- Added a database index on `stripe_fingerprint` for performance

# Test Results

image

---

# Checklist

---

# AI Disclosure

Claude Opus 4.5 was used to assist with implementing this feature via Claude Code.

### DIFF
--- a/app/modules/user/risk.rb
+++ b/app/modules/user/risk.rb
@@ -6,6 +6,7 @@ module User::Risk
   IFFY_ENDPOINT = "http://internal-production-iffy-live-internal-1668548970.us-east-1.elb.amazonaws.com"
 
   PAYMENT_REMINDER_RISK_STATES = %w[flagged_for_tos_violation not_reviewed compliant].freeze
+  SUSPENDED_STATES = %w[suspended_for_tos_violation suspended_for_fraud].freeze
   INCREMENTAL_ENQUEUE_BALANCE = 100_00
   COUNTRIES_THAT_DO_NOT_HAVE_ZIPCODES = [
     # Country Codes: http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
@@ -146,14 +147,14 @@ module User::Risk
   end
 
   def enable_accounts_with_same_stripe_fingerprint
-    fingerprints = bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
+    fingerprints = bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
     return if fingerprints.empty?
 
     user_ids_with_same_fingerprint = BankAccount.alive
       .where(stripe_fingerprint: fingerprints)
       .where.not(user_id: id)
+      .distinct
       .pluck(:user_id)
-      .uniq
 
     User.where(id: user_ids_with_same_fingerprint).each do |user|
       matching_fingerprint = (fingerprints & user.alive_bank_accounts.pluck(:stripe_fingerprint)).first
--- a/app/sidekiq/check_payment_address_worker.rb
+++ b/app/sidekiq/check_payment_address_worker.rb
@@ -4,8 +4,6 @@ class CheckPaymentAddressWorker
   include Sidekiq::Job
   sidekiq_options retry: 0, queue: :default
 
-  SUSPENDED_STATES = %w[suspended_for_tos_violation suspended_for_fraud].freeze
-
   def perform(user_id)
     user = User.find_by(id: user_id)
     return unless user&.can_flag_for_fraud?
@@ -22,7 +20,7 @@ class CheckPaymentAddressWorker
 
       banned_accounts_with_same_payment_address = User.where(
         payment_address: user.payment_address,
-        user_risk_state: SUSPENDED_STATES
+        user_risk_state: User::Risk::SUSPENDED_STATES
       )
 
       blocked_email = BlockedObject.find_active_object(user.payment_address)
@@ -31,17 +29,17 @@ class CheckPaymentAddressWorker
     end
 
     def stripe_fingerprint_matches_suspended_account?(user)
-      fingerprints = user.alive_bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
+      fingerprints = user.alive_bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
       return false if fingerprints.empty?
 
       suspended_accounts_with_same_fingerprint = BankAccount
         .joins(:user)
         .where(stripe_fingerprint: fingerprints)
         .where.not(user_id: user.id)
-        .where(users: { user_risk_state: SUSPENDED_STATES })
+        .where(users: { user_risk_state: User::Risk::SUSPENDED_STATES })
 
-      blocked_fingerprints = fingerprints.any? { |fp| BlockedObject.find_active_object(fp).present? }
+      return true if suspended_accounts_with_same_fingerprint.exists?
 
-      suspended_accounts_with_same_fingerprint.exists? || blocked_fingerprints
+      BlockedObject.find_active_objects(fingerprints).present?
     end
 end
--- a/app/sidekiq/suspend_accounts_with_payment_address_worker.rb
+++ b/app/sidekiq/suspend_accounts_with_payment_address_worker.rb
@@ -24,14 +24,14 @@ class SuspendAccountsWithPaymentAddressWorker
     end
 
     def suspend_users_with_same_stripe_fingerprint(suspended_user)
-      fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
+      fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
       return if fingerprints.empty?
 
       user_ids_with_same_fingerprint = BankAccount.alive
         .where(stripe_fingerprint: fingerprints)
         .where.not(user_id: suspended_user.id)
+        .distinct
         .pluck(:user_id)
-        .uniq
 
       User.not_suspended.where(id: user_ids_with_same_fingerprint).find_each do |user|
         matching_fingerprint = (fingerprints & user.alive_bank_accounts.pluck(:stripe_fingerprint)).first


app/modules/user/risk.rb
```diff
@@ -129,13 +129,34 @@ def block_seller_ip!
...
     return if payment_address.blank?
 
     User.where(payment_address:).where.not(id:).each do |user|
       user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as payment address #{payment_address} is now unblocked")
     end
   end
 
+  def enable_accounts_with_same_stripe_fingerprint
+    fingerprints = bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
+    return if fingerprints.empty?
```
yashranaway, line 151: Same pattern as the suspend worker - checking all bank accounts (including deleted) to find all fingerprints this user has ever used.

app/modules/user/risk.rb
```diff
@@ -129,13 +129,34 @@ def block_seller_ip!
...
+
+    user_ids_with_same_fingerprint = BankAccount.alive
+      .where(stripe_fingerprint: fingerprints)
+      .where.not(user_id: id)
+      .pluck(:user_id)
+      .uniq
+
+    User.where(id: user_ids_with_same_fingerprint).each do |user|
+      matching_fingerprint = (fingerprints & user.alive_bank_accounts.pluck(:stripe_fingerprint)).first
+      user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as bank account fingerprint #{matching_fingerprint} is now unblocked")
```
yashranaway: Finding the specific matching fingerprint for an accurate admin comment.

app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
```diff
@@ -292,6 +292,8 @@ def self.save_stripe_bank_account_info(bank_account, stripe_account)
     bank_account.stripe_external_account_id = stripe_external_account.id
     bank_account.stripe_fingerprint = stripe_external_account.fingerprint
     bank_account.save!
+
+    CheckPaymentAddressWorker.perform_async(bank_account.user_id)
```
yashranaway, line 302: This triggers the fraud check right after Stripe gives us the fingerprint. Mirrors how PayPal triggers the check in `UpdatePayoutMethod` when a payment address is set.

db/migrate/20260110070129_add_index_to_bank_accounts_stripe_fingerprint.rb
yashranaway, line 1: Adding an index since we're now querying by fingerprint. Should help with performance when we have many bank accounts.

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,17 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
   def perform(user_id)
     suspended_user = User.find(user_id)
 
-    return if suspended_user.payment_address.blank?
+    suspend_users_with_same_payment_address(suspended_user)
+    suspend_users_with_same_stripe_fingerprint(suspended_user)
```
yashranaway, line 11: After checking PayPal addresses, we now also check for matching Stripe fingerprints.

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,17 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+  end
+
+  private
+    def suspend_users_with_same_payment_address(suspended_user)
+      return if suspended_user.payment_address.blank?
+
+      User.where(payment_address: suspended_user.payment_address).where.not(id: suspended_user.id).find_each do |user|
+        next if user.suspended?
```
yashranaway: Added this check to skip users who are already suspended. Avoids creating duplicate comments and unnecessary state machine transitions.

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,17 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+
+        flag_and_suspend_user(user, suspended_user, "payment address", suspended_user.payment_address)
+      end
+    end
+
+    def suspend_users_with_same_stripe_fingerprint(suspended_user)
+      fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
```
yashranaway: Using `bank_accounts` (not `alive_bank_accounts`) here intentionally - we want to catch fingerprints from deleted bank accounts too. A fraudster might delete their bank account after using it, but we still want to flag related accounts.

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,17 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+      return if fingerprints.empty?
 
-    User.where(payment_address: suspended_user.payment_address).where.not(id: suspended_user.id).find_each do |user|
+      user_ids_with_same_fingerprint = BankAccount.alive
+        .where(stripe_fingerprint: fingerprints)
+        .where.not(user_id: suspended_user.id)
+        .pluck(:user_id)
+        .uniq
```
yashranaway: For target users, we only check alive bank accounts since deleted ones can't be used for fraud anymore.

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,17 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+
+      User.where(id: user_ids_with_same_fingerprint).find_each do |user|
+        next if user.suspended?
+
+        matching_fingerprint = (fingerprints & user.alive_bank_accounts.pluck(:stripe_fingerprint)).first
```
yashranaway, line 37: When the suspended user has multiple fingerprints, we find which specific one matched so the admin comment is accurate.

app/sidekiq/check_payment_address_worker.rb
```diff
@@ -4,17 +4,44 @@ class CheckPaymentAddressWorker
   include Sidekiq::Job
   sidekiq_options retry: 0, queue: :default
 
+  SUSPENDED_STATES = %w[suspended_for_tos_violation suspended_for_fraud].freeze
```
yashranaway: Extracted this constant to avoid repetition and make it clearer what states we're checking for.

app/sidekiq/check_payment_address_worker.rb
```diff
@@ -4,17 +4,44 @@ class CheckPaymentAddressWorker
...
+        user_risk_state: SUSPENDED_STATES
+      )
+
+      blocked_email = BlockedObject.find_active_object(user.payment_address)
+
+      banned_accounts_with_same_payment_address.exists? || blocked_email.present?
+    end
+
+    def stripe_fingerprint_matches_suspended_account?(user)
+      fingerprints = user.alive_bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
```
yashranaway: Checking all of the new user's fingerprints, not just their active bank account. A user might have multiple bank accounts.

app/sidekiq/check_payment_address_worker.rb
```diff
@@ -4,17 +4,44 @@ class CheckPaymentAddressWorker
...
+      return false if fingerprints.empty?
+
+      suspended_accounts_with_same_fingerprint = BankAccount
+        .joins(:user)
+        .where(stripe_fingerprint: fingerprints)
+        .where.not(user_id: user.id)
+        .where(users: { user_risk_state: SUSPENDED_STATES })
+
```
yashranaway, line 40: Not filtering by `.alive` here - we intentionally check against ALL bank accounts (including deleted) from suspended users. This preserves fraud history even if fraudsters delete their accounts.

app/sidekiq/check_payment_address_worker.rb
```diff
@@ -4,17 +4,44 @@ class CheckPaymentAddressWorker
...
+      blocked_fingerprints = fingerprints.any? { |fp| BlockedObject.find_active_object(fp).present? }
```
yashranaway: Also checking `BlockedObject` for explicitly blocked fingerprints, same pattern as the PayPal email check.

app/modules/user/risk.rb
```diff
@@ -133,13 +133,34 @@ def block_seller_ip!
...
+  def enable_accounts_with_same_payment_address
     return if payment_address.blank?
 
     User.where(payment_address:).where.not(id:).each do |user|
       user.mark_compliant!(author_name: "enable_sellers_other_accounts", content: "Marked compliant automatically on #{Time.current.to_fs(:formatted_date_full_month)} as payment address #{payment_address} is now unblocked (from User##{id})", skip_transition_callback: :enable_sellers_other_accounts)
     end
   end
 
+  def enable_accounts_with_same_stripe_fingerprint
+    fingerprints = bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
```
EmCousin: You can uniq the fingerprints in the database rather than in memory:
```suggestion
    fingerprints = bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
```

app/modules/user/risk.rb
```diff
@@ -133,13 +133,34 @@ def block_seller_ip!
...
+    return if fingerprints.empty?
+
+    user_ids_with_same_fingerprint = BankAccount.alive
+      .where(stripe_fingerprint: fingerprints)
+      .where.not(user_id: id)
+      .pluck(:user_id)
+      .uniq
```
EmCousin:
```suggestion
    user_ids_with_same_fingerprint = BankAccount.alive
      .where(stripe_fingerprint: fingerprints)
      .where.not(user_id: id)
      .distinct
      .pluck(:user_id)
```

app/sidekiq/check_payment_address_worker.rb
EmCousin: This constant would be better off being defined in the `User::Risk` module and invoked here

app/sidekiq/check_payment_address_worker.rb
EmCousin:
```suggestion
      fingerprints = user.alive_bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
```

app/sidekiq/check_payment_address_worker.rb
EmCousin: You can query MongoDB once instead of one time per fingerprint
```suggestion
      blocked_fingerprints = BlockedObject.find_active_objects(fingerprints)
```

EmCousin: Can you verify this does not evaluate immediately? It would be better to _not_ send the query to MongoDB if ` suspended_accounts_with_same_fingerprint.exists? ` is already true

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,18 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+      User.not_suspended
+        .where(payment_address: suspended_user.payment_address)
+        .where.not(id: suspended_user.id)
+        .find_each do |user|
+          flag_and_suspend_user(user, suspended_user, "payment address", suspended_user.payment_address)
+        end
+    end
+
+    def suspend_users_with_same_stripe_fingerprint(suspended_user)
+      fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).pluck(:stripe_fingerprint).uniq
```
EmCousin:
```suggestion
      fingerprints = suspended_user.bank_accounts.where.not(stripe_fingerprint: [nil, ""]).distinct.pluck(:stripe_fingerprint)
```

app/sidekiq/suspend_accounts_with_payment_address_worker.rb
```diff
@@ -7,18 +7,47 @@ class SuspendAccountsWithPaymentAddressWorker
...
+      return if fingerprints.empty?
 
-    User.where(payment_address: suspended_user.payment_address).where.not(id: suspended_user.id).not_suspended.find_each do |user|
+      user_ids_with_same_fingerprint = BankAccount.alive
+        .where(stripe_fingerprint: fingerprints)
+        .where.not(user_id: suspended_user.id)
+        .pluck(:user_id)
+        .uniq
```
EmCousin:
```suggestion
        .distinct
        .pluck(:user_id)
```

yashranaway:
> Can you verify this does not evaluate immediately? It would be better to not send the query to MongoDB if suspended_accounts_with_same_fingerprint.exists? is already true

@EmCousin _a Fixed in [fd488a6](https://github.com/antiwork/gumroad/pull/2810/commits/fd488a68588cea6f63391c462223244e13f6e04e) The code now uses an early return.  Also the CI failures are unrelated

PR #2820, opened 2026-01-11
Support logging in via mobile access token on the URL redirect page

binary-koan:
# Description

## Problem

Currently the mobile apps use the regular download page at /d/:token to load product content. However, it feels very clunky because we run into the email confirmation check and show an extra form to confirm the email before viewing product details.

## Solution

Added an option to bypass the confirmation check by providing a valid OAuth access token in a query param. The apps can pass this to verify the correct user is viewing the content.

---

# AI Disclosure

Used Claude Opus 4.5 for exploring the codebase and refactoring specs.

### DIFF
--- a/app/controllers/url_redirects_controller.rb
+++ b/app/controllers/url_redirects_controller.rb
@@ -356,9 +356,9 @@ class UrlRedirectsController < ApplicationController
         return redirect_to url_redirect_membership_inactive_page_path(@url_redirect.token)
       end
 
-      if params[:access_token].present?
+      if params[:access_token].present? && params[:mobile_token] == Api::Mobile::BaseController::MOBILE_TOKEN
         doorkeeper_authorize! :mobile_api
-        return if purchase.purchaser && purchase.purchaser == current_api_user
+        return if purchase && purchase.purchaser && purchase.purchaser == current_api_user
       end
 
       if cookies.encrypted[:confirmed_redirect] == @url_redirect.token ||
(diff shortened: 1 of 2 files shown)


app/controllers/url_redirects_controller.rb
```diff
@@ -299,6 +300,14 @@ def redirect_to_custom_domain_if_needed
       )
     end
 
+    def sign_in_mobile_app_user
+      return unless params[:access_token].present?
```
binary-koan: Doorkeeper reads this param by default

app/controllers/url_redirects_controller.rb
```diff
@@ -356,6 +356,11 @@ def check_permissions
         return redirect_to url_redirect_membership_inactive_page_path(@url_redirect.token)
       end
 
+      if params[:access_token].present?
```
binary-koan: Passing this as a query param is the easiest method to authorize in a webview (Doorkeeper also supports an authorization header but we can't pass custom headers in webview navigation)

ershad: We may need to validate MOBILE_TOKEN too? On a second look, it seems like we can create an access token with the `mobile_api` scope without MOBILE_TOKEN. Currently, MOBILE_TOKEN is validated only for mobile API requests.

app/controllers/url_redirects_controller.rb
```diff
@@ -356,6 +356,11 @@ def check_permissions
...
+        doorkeeper_authorize! :mobile_api
+        return if purchase.purchaser && purchase.purchaser == current_api_user
```
ershad: There could be URL redirects without associated purchase records (for eg, installments). We can have a check similar to line 365.

```suggestion
        return if purchase && purchase.purchaser && purchase.purchaser == current_api_user
```

app/controllers/url_redirects_controller.rb
binary-koan: Might as well validate mobile_token for consistency, will update

Issue #2053, opened 2025-11-17
Add feature to automatically apply discount code to the product


ershad:
## Design

When no discount code is applied:

<img width="807" height="273" alt="Image" src="https://github.com/user-attachments/assets/aea9befb-499d-4e9f-a9bb-53fbddacf397" />

When the radio is selected:

<img width="811" height="389" alt="Image" src="https://github.com/user-attachments/assets/c8bd6782-b018-4dc1-9977-073d0cade532" />

When discount code is selected:

<img width="813" height="384" alt="Image" src="https://github.com/user-attachments/assets/48f800d2-f95e-437c-b59c-3d5b0a5d84cc" />

- [ ] When a product has an auto-applied discount code, show the discounted price everywhere:
  - Preview on product edit page
  - Discover
  - Seller profile
  - Product page

Discounted price should look similar to how it looks now when we apply a discount code through `&code=<discount_code>` query string:

<img width="1193" height="594" alt="Image" src="https://github.com/user-attachments/assets/a41cd5da-3a28-4af7-bc2d-fb9d7d60a9a0" />

If the buyer has supplied another discount code (through the URL), prioritize that over the product's default discount code.

- [ ] Track purchases made using the default discount code (can use `attr_json_data_accessor` on `Purchase` object)

thakkerpankti837-tech:
Hello ! 

I have a deep passion for being a part of this feature. I have reviewed the issue, as well as the conversation; and, I'm looking forward to working on this auto apply discount system with you.

I will be building a simple validation layer to verify eligibility rules on the client side and / or server side (depending on the design of the application) before the discount is automatically applied, once the eligibility conditions have been met. Also, I will build the logic in a way that allows for future enhancement with additional discount types.

While I am still getting to know the code base; however, I will complete this task and will be very thankful for any help or guidance that you may have. I'll also submit a PR with my proposed method to begin implementation once I have completed my plan.

Please let me know if there are any specific topics of interest or concern that you would like me to cover at this time. 

Thanks, and I look forward to collaborating with you!

Sagarpillai:
@ershad has this issue been taken care of or is it still open for contributions?

SatyamDave:
still up for grab?

ershad:
@EmCousin _a could you prioritize reviewing/merging this feature?

PR #2824, opened 2026-01-11
Auto apply discount codes

yashranaway:
Issue: #2053

# Description

## Problem

Sellers have no way to automatically apply a discount to their products. They have to share discount URLs manually or tell customers to enter a code at checkout.

## Solution

Added a "Apply discount" dropdown in the product edit pricing section. Sellers can select any active discount code associated with their product (or a universal code). The discount auto-applies when customers visit the product page.

- Product model stores `default_offer_code_id` reference
- Validation ensures only associated/universal codes can be set as default
- Validation blocks expired or not-yet-active codes from being set
- Deletion protection: discount codes set as default cannot be deleted until removed from products
- Card presenter shows discounted price with strikethrough on original
- URL includes the discount code param for proper checkout flow

---

# Before/After

https://github.com/user-attachments/assets/72e8c898-6751-48c7-92a5-6226588807fa

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

Claude Code (Opus 4.5) was used for implementation assistance, code review, and test writing.

### DIFF
--- a/app/models/offer_code.rb
+++ b/app/models/offer_code.rb
@@ -313,7 +313,7 @@ class OfferCode < ApplicationRecord
     end
 
     def prevent_destroy_if_used_as_default_discount
-      return unless products_with_default_discount.exists?
+      return unless products_with_default_discount.visible.exists?
 
       errors.add(:base, "Cannot delete this discount code because it is set as the default discount on one or more products.")
       throw(:abort)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,70 +11,6 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.1].define(version: 2026_01_03_190826) do
-  create_table "_links_new", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "user_id"
-    t.string "name", null: false
-    t.string "unique_permalink"
-    t.text "preview_url", size: :medium
-    t.text "description", size: :medium
-    t.integer "purchase_type", default: 0
-    t.datetime "created_at", precision: nil
-    t.datetime "updated_at"
-    t.datetime "purchase_disabled_at", precision: nil
-    t.datetime "deleted_at", precision: nil
-    t.integer "price_cents"
-    t.string "price_currency_type", default: "usd"
-    t.boolean "customizable_price"
-    t.integer "max_purchase_count"
-    t.integer "bad_card_counter", default: 0
-    t.boolean "require_shipping", default: false
-    t.datetime "last_partner_sync", precision: nil
-    t.text "preview_oembed", size: :medium
-    t.boolean "showcaseable", default: false
-    t.text "custom_receipt", size: :medium
-    t.string "custom_filetype"
-    t.string "filetype", default: "link"
-    t.string "filegroup", default: "url"
-    t.bigint "size"
-    t.integer "bitrate"
-    t.integer "framerate"
-    t.integer "pagelength"
-    t.integer "duration"
-    t.integer "width"
-    t.integer "height"
-    t.string "custom_permalink"
-    t.string "common_color"
-    t.integer "suggested_price_cents"
-    t.datetime "banned_at", precision: nil
-    t.integer "risk_score"
-    t.datetime "risk_score_updated_at", precision: nil
-    t.boolean "draft", default: false
-    t.bigint "flags", default: 0, null: false
-    t.integer "subscription_duration"
-    t.text "json_data", size: :medium
-    t.string "external_mapping_id"
-    t.bigint "affiliate_application_id"
-    t.integer "rental_price_cents"
-    t.integer "duration_in_months"
-    t.datetime "migrated_to_tiered_pricing_at", precision: nil
-    t.integer "free_trial_duration_unit"
-    t.integer "free_trial_duration_amount"
-    t.datetime "content_updated_at", precision: nil
-    t.bigint "taxonomy_id"
-    t.string "native_type", default: "digital", null: false
-    t.integer "discover_fee_per_thousand", default: 100, null: false
-    t.string "support_email"
-    t.bigint "default_offer_code_id"
-    t.index ["banned_at"], name: "index_links_on_banned_at"
-    t.index ["custom_permalink"], name: "index_links_on_custom_permalink", length: 191
-    t.index ["deleted_at"], name: "index_links_on_deleted_at"
-    t.index ["showcaseable"], name: "index_links_on_showcaseable"
-    t.index ["taxonomy_id"], name: "index_links_on_taxonomy_id"
-    t.index ["unique_permalink"], name: "index_links_on_unique_permalink", length: 191
-    t.index ["user_id", "updated_at"], name: "index_links_on_user_id_and_updated_at"
-    t.index ["user_id"], name: "index_links_on_user_id"
-  end
-
   create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "name", limit: 191, null: false
     t.string "record_type", limit: 191, null: false
@@ -210,27 +146,6 @@ ActiveRecord::Schema[7.1].define(version: 2026_01_03_190826) do
     t.index ["link_id"], name: "index_asset_previews_on_link_id"
   end
 
-  create_table "audience_export_chunks", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "export_id", null: false
-    t.text "member_ids", size: :long
-    t.text "members_data", size: :long
-    t.boolean "processed", default: false, null: false
-    t.string "revision"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["export_id"], name: "index_audience_export_chunks_on_export_id"
-  end
-
-  create_table "audience_exports", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "seller_id", null: false
-    t.bigint "recipient_id", null: false
-    t.text "options"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["recipient_id"], name: "index_audience_exports_on_recipient_id"
-    t.index ["seller_id"], name: "index_audience_exports_on_seller_id"
-  end
-
   create_table "audience_members", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.bigint "seller_id", null: false
     t.string "email", null: false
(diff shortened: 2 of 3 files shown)


db/migrate/20260103190826_add_default_offer_code_to_links.rb
yashranaway, line 1: Simple migration adding `default_offer_code_id` column to links table. Added index for lookup performance.

app/models/link.rb
```diff
@@ -1318,6 +1324,17 @@ def quantity_enabled_state_is_allowed
...
+      return if default_offer_code_id.blank?
+
+      valid_offer_code = offer_codes.alive.exists?(id: default_offer_code_id) ||
+        user&.offer_codes&.universal_with_matching_currency(price_currency_type)&.alive&.exists?(id: default_offer_code_id)
+
+      if !valid_offer_code
+        errors.add(:default_offer_code, "must be associated with this product")
+      end
+    end
+
```
yashranaway, line 1339: Added `belongs_to :default_offer_code` association and a validation to ensure the selected offer code actually belongs to the product (either directly or as a universal code). The `default_discount_code` helper just returns the code string for convenience in controllers.

app/models/offer_code.rb
yashranaway, line 1: Added inverse association and a `before_destroy` callback to clear `default_offer_code_id` on any products using this code. Prevents orphaned references when a discount code gets deleted.

app/models/purchase.rb
yashranaway, line 1: Added `used_default_discount_code` json_data accessor. Lets us track analytics on how many purchases came through the auto-applied discount vs manual entry.

app/controllers/links_controller.rb
```diff
@@ -191,7 +191,10 @@ def show
                                   ChargeProcessor::DEFAULT_CURRENCY_CODE
     @pay_with_card_enabled = @product.user.pay_with_card_enabled?
     presenter = ProductPresenter.new(pundit_user:, product: @product, request:)
-    presenter_props = { recommended_by: params[:recommended_by], discount_code: params[:offer_code] || params[:code], quantity: (params[:quantity] || 1).to_i, layout: params[:layout], seller_custom_domain_url: }
+    url_discount_code = params[:offer_code] || params[:code]
+    effective_discount_code = url_discount_code.presence || @product.default_discount_code
+    is_default_discount = url_discount_code.blank? && effective_discount_code.present?
+    presenter_props = { recommended_by: params[:recommended_by], discount_code: effective_discount_code, is_default_discount:, quantity: (params[:quantity] || 1).to_i, layout: params[:layout], seller_custom_domain_url: }
```
yashranaway, line 149: Modified show action to check for URL discount first, then fall back to product's default. URL always wins - we don't want to override someone's affiliate link or special promo.

app/policies/link_policy.rb
```diff
@@ -78,6 +78,7 @@ def product_permitted_attributes
       :is_adult,
       :discover_fee_per_thousand,
       :taxonomy_id,
+      :default_offer_code_id,
```
yashranaway, line 81: Just added `default_offer_code_id` to permitted params. Standard stuff.

app/presenters/product_presenter.rb
yashranaway, line 1: Added `available_offer_codes` array to edit_props. Filters out cancellation discounts since those are special-purpose. Frontend needs this list to populate the dropdown.

app/presenters/product_presenter/card.rb
```diff
@@ -21,6 +21,10 @@ def initialize(product:)
 
   def for_web(request: nil, recommended_by: nil, recommender_model_name: nil, target: nil, show_seller: true, affiliate_id: nil, query: nil, offer_code: nil, compute_description: true)
     default_recurrence = product.default_price_recurrence
+    original_price = product.display_price_cents(for_default_duration: true)
+    discount_info = calculate_default_discount(original_price)
+    effective_offer_code = offer_code || (discount_info ? product.default_discount_code : nil)
```
yashranaway, line 26: Calculate discount in `for_web`

app/presenters/product_presenter/card.rb
```diff
@@ -34,10 +38,12 @@ def for_web(request: nil, recommended_by: nil, recommender_model_name: nil, targ
       native_type: product.native_type,
       quantity_remaining: product.remaining_for_sale_count,
       is_sales_limited: product.max_purchase_count?,
-      price_cents: product.display_price_cents(for_default_duration: true),
+      price_cents: discount_info ? discount_info[:discounted_price] : original_price,
+      original_price_cents: discount_info ? original_price : nil,
+      has_default_discount: discount_info.present?,
```
yashranaway, line 43: Return discount info in props

app/presenters/product_presenter/card.rb
```diff
@@ -61,4 +67,13 @@ def for_email
...
   end
+
+  private
+    def calculate_default_discount(original_price)
+      offer_code = product.default_offer_code
+      return nil if offer_code.nil? || offer_code.deleted? || offer_code.inactive?
+
+      discount_amount = offer_code.amount_off(original_price)
+      { discounted_price: [original_price - discount_amount, 0].max }
+    end
```
yashranaway, line 78: Private calculate method

app/presenters/product_presenter/product_props.rb
yashranaway, line 1: Added `is_default` flag to discount_code_props. Frontend can use this to show slightly different UI for auto-applied vs manually entered codes if needed.

app/services/purchase/create_service.rb
yashranaway, line 1: When a purchase uses a discount code, we check if it matches the product's default. If so, mark `used_default_discount_code = true`.

app/javascript/components/ProductEdit/ProductTab/ApplyDiscountSelector.tsx
yashranaway, line 1: New component following the same pattern as CancellationDiscountSelector. Toggle enables/disables, dropdown picks the code. Updates product state via context on change.

app/javascript/components/ProductEdit/ProductTab/PriceEditor.tsx
yashranaway, line 1: Just imported and rendered ApplyDiscountSelector at the bottom of the pricing section.

app/javascript/components/ProductEdit/state.ts
```diff
@@ -143,6 +143,13 @@ export type Product = {
   call_limitation_info: CallLimitationInfo | null;
   require_shipping: boolean;
   cancellation_discount: CancellationDiscount | null;
+  default_offer_code_id: string | null;
+  available_offer_codes: {
+    id: string;
+    code: string;
+    display_amount: string;
+    is_percent: boolean;
+  }[];
```
yashranaway, line 152: Added types for `default_offer_code_id` and `available_offer_codes` array.

app/javascript/components/Product/Card.tsx
yashranaway, line 1: Pass `oldPrice` prop to PriceTag when `has_default_discount` is true. PriceTag already handles the strikethrough styling.

app/javascript/parsers/product.ts
yashranaway, line 1: Added `original_price_cents` and `has_default_discount` to CardProduct type.

spec/models/link_spec.rb
```diff
@@ -4800,4 +4800,86 @@
...
+    end
+
+    it "clears default_offer_code when the offer code is destroyed" do
+      product.update!(default_offer_code: product_offer_code)
+      expect(product.reload.default_offer_code_id).to eq(product_offer_code.id)
+
+      product_offer_code.destroy!
+      expect(product.reload.default_offer_code_id).to be_nil
+    end
+  end
```
yashranaway, line 4904: Added tests for the association, the `default_discount_code` helper method, and the validation logic. Also added a test to verify the cleanup callback works when an offer code is destroyed.

yashranaway:
@EmCousin Can I get a review on this one? 👀

spec/services/purchase/create_service_spec.rb
```diff
@@ -2669,6 +2669,39 @@
         expect(purchase.error_code).to eq "exceeding_offer_code_quantity"
       end
     end
+
```
gianfrancopiana, line 2672: The tests for `used_default_discount_code` don't actually test the service - they manually build the purchase and set attributes directly. Please update to call the service.

Per `CONTRIBUTING.md`, "Tests must fail when the fix is reverted." These tests would pass even if the implementation in `create_service.rb` was removed.

app/models/offer_code.rb
```diff
@@ -301,4 +304,18 @@ def capture_associated_product_ids
     def reindex_captured_products
       reindex_associated_products(products_to_reindex: Link.where(id: @product_ids_to_reindex)) if @product_ids_to_reindex.present?
     end
+
+    def cannot_delete_if_used_as_default_discount
```
gianfrancopiana, line 308: The deletion protection logic is inconsistent:                 
                                                                 
- Soft delete (`cannot_delete_if_used_as_default_discount`): 
  - `return unless products_with_default_discount.visible.exists?  ` 
  - Only blocks deletion if the product is visible (not deleted).  
                                                                 
- Hard delete (`prevent_destroy_if_used_as_default_discount`):     
  - `return unless products_with_default_discount.exists?`           
  - Blocks deletion even if the product is deleted.                
                                                                 
This means soft-deleting an offer code is allowed when set as default on a deleted product, but hard-deleting is not. Both should use the same scope.

db/schema.rb
```diff
@@ -10,7 +10,71 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_11_24_133549) do
+ActiveRecord::Schema[7.1].define(version: 2026_01_03_190826) do
```
gianfrancopiana, line 13: Please make sure to run `bin/rails db:migrate` to regenerate schema.rb cleanly.

gianfrancopiana (changes requested):
A few items need addressing before merge - most critically, the purchase service tests need to actually call the service rather than manually setting attributes. See inline comments.

yashranaway:
Hey @gianfrancopiana, thanks for the review, I've addressed all the comments:                                                                                                                                               
                                                                                                                                                                                                                              
  1. Purchase service tests - Now call `service.send(:build_purchase`, `params[:purchase])` instead of manually setting attributes. Tests will fail if implementation is removed (expects true, default is false).                
  2. Deletion protection scope - Both soft delete and hard delete now consistently use `.visible.exists? `                                                                                                                      
  3. Schema.rb - Regenerated via bin/rails db:migrate                                                                                                                                                                         
                                                                                                                                                                                                                              
  All tests passing. Could you take another look?

gianfrancopiana:
After further review, I saw we had updated the requirements to use the "better discount wins" logic (apply whichever discount is greater) rather than "URL always wins." PR #2920 implements that updated behavior, so (unfortunately) we'll have to close this one.                                         

Apologies for the back and forth - appreciate your work on this!

yashranaway:
Hey @gianfrancopiana, appreciate the consideration! I'd like to make a case for URL priority if you're open to it.                                                                                         
                                                                                                                                                                                                            Looking at the codebase, sellers actively track "Uses" and "Revenue" per discount code in the Discounts dashboard. With "better discount wins", if a customer enters "SUMMER10" but the default is higher,  their code gets silently replaced - breaking campaign attribution and making those stats unreliable.                                                                                                       
                                                                                                                                                                                                             
URL priority keeps seller control intact: if they share a specific code, that's the one that gets tracked. They can always set a generous default if they want everyone to get the best price - but URL codes should be intentional overrides they control.                                                                                                                                                        
                                                                                                                                                                                                               Happy to discuss if you'd reconsider!

yashranaway:
cc @ershad - since the requirement was changed based on your feedback, would love your input on the tracking concern above.  👆

gianfrancopiana:
> Hey @gianfrancopiana, appreciate the consideration! I'd like to make a case for URL priority if you're open to it. Looking at the codebase, sellers actively track "Uses" and "Revenue" per discount code in the Discounts dashboard. With "better discount wins", if a customer enters "SUMMER10" but the default is higher, their code gets silently replaced - breaking campaign attribution and making those stats unreliable.
> 
> URL priority keeps seller control intact: if they share a specific code, that's the one that gets tracked. They can always set a generous default if they want everyone to get the best price - but URL codes should be intentional overrides they control. Happy to discuss if you'd reconsider!

@yashranaway Appreciate the pushback. Even though this makes sense in theory, if a seller sets a default discount, they're basically saying "this is what everyone gets." They wouldn't run campaigns with worse codes at the same time and expect to track which one actually works. This is an edge case.

Issue #2183, opened 2025-12-01
Investigate and fix the random 0% sale tax for India sales without Tax ID


ershad:
### Problem
Some India purchases have 0% tax despite the buyer not providing a GST number.

### Investigation
Root cause: when a buyer selects India but their IP and card are both from non-taxable countries (e.g., Saudi Arabia), the system skips tax. Verified on 2 India purchases in a 2,000 purchase sample. Same pattern affects 18 countries (113 cases in the same sample).

### Fix
Trust the buyer's country selection when IP/card are from non-taxable countries. Legal sign-off confirmed by @steveo-es. on Jan 9, 2026.

chaitanya-mishra-ai:
Done in PR #2503. Added the Buyer Tax ID column and included B2B purchases in the report.

ashutosh887:
@ershad @slavingia I would love to work on this!
Should I pick this if not done yet?

ershad:
Checked with @steveo-es.
1. Currently, the India sales report excludes sales with a Tax ID; we will leave this unchanged.
2. For all other Indian sales, we must make sure the correct sales tax is applied. In a few random cases, sales have a 0% tax rate despite lacking a Tax ID. We should investigate this issue and fix it.

gianfrancopiana:
#2845 closes this issue.

PR #2845, opened 2026-01-12
Fix tax not applied when buyer's IP/card are from non-taxable countries

gianfrancopiana:
Issue #2183

## Root Cause

When a buyer selects a taxable country (e.g., India) but their IP address and card are both from non-taxable countries, the system was resetting tax to 0 as a "safety measure." The logic treated this as suspicious and refused to apply tax.

This affected 113 purchases across 18 taxable countries in a 2,000 purchase sample.

## Solution

Trust the buyer's country selection when their IP and card are from non-taxable countries. The previous logic returned `false` (skip tax) for this scenario. The new logic returns `true` (apply tax).

## Changes

- Modified `should_skip_charge_taxes?` in `purchase.rb` to trust buyer's selected country when IP/card are from non-taxable locations

## Why This Approach

The original "safety measure" was overly cautious. If a buyer explicitly selects India and their IP/card don't contradict it (they're just from places we don't tax like Saudi Arabia), there's no fraud signal - we should apply the appropriate tax.

Common legitimate scenarios:
- Travelers using foreign payment methods
- VPN users
- Expats with cards from their home country

Legal sign-off confirmed by @steveo-es.

## Test Plan

- [x] Existing test updated: buyer selects EU country with non-EU IP/card → VAT now applied (was skipped)
- [x] Unit tests pass for purchase taxation logic

---

## AI Disclosure


gianfrancopiana:
## CI Failure Note

The failing WA sales tax tests are **unrelated to this PR** - they're also failing on `main`:

**Main branch failure:** https://github.com/antiwork/gumroad/actions/runs/20930071502

The issue is that TaxJar's Seattle, WA tax rate changed from 10.35% to 10.55% since the VCR cassettes were recorded. This requires re-recording the cassettes separately.

PR #2863, opened 2026-01-14
Fix: Persist `chargeback_reversed` flag of disputed purchases when dispute is won for a `Charge`

### LINKED ISSUE #2055: Chargeback wins on Stripe not reflected in purchase status
I’ve seen a few purchases where we won the chargeback on Stripe, but the status on our side was never updated to “chargeback reversed.”

Example purchase ID: `2FBemwLdUq0WnvyWW3G8ew==` (manually updated later through the console)

vishaltelangre:
Fixes #2055 and https://github.com/antiwork/gumroad-private/issues/65

# Description

## Problem

When a dispute is won in seller's favor for purchases associated with a `Charge` object (instead of handling the event directly on a `Purchase`), the corresponding purchases' `chargeback_reversed` flag is set in memory but does not persist to the database. This causes customers to be blocked from accessing their purchased products even though the seller won the dispute.

## Solution

Replaced the `purchase.chargeback_reversed = true` with `purchase.update!(chargeback_reversed: true)` while handling `charge.dispute.closed` event with `status: "won"`.

### DIFF
--- a/spec/models/concerns/charge/disputable_spec.rb
+++ b/spec/models/concerns/charge/disputable_spec.rb
@@ -926,6 +926,23 @@ describe Charge::Disputable, :vcr do
         end
       end
 
+      context "when dispute won for a charge" do
+        let(:user) { create(:user) }
+        let(:product) { create(:product, user:) }
+        let(:transaction_id) { "ch_charge_123" }
+        let!(:purchase) { create(:purchase, link: product, stripe_transaction_id: transaction_id, chargeback_date: 10.days.ago) }
+        let!(:charge) { create(:charge, processor_transaction_id: transaction_id, amount_cents: 111, disputed_at: 10.days.ago, purchases: [purchase]) }
+
+        let(:event) { build(:charge_event_dispute_won, charge_id: transaction_id) }
+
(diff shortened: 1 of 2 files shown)


spec/models/concerns/charge/disputable_spec.rb
```diff
@@ -926,6 +926,23 @@
...
+        let!(:purchase) { create(:purchase, link: product, stripe_transaction_id: transaction_id, chargeback_date: 10.days.ago) }
+        let!(:charge) { create(:charge, processor_transaction_id: transaction_id, amount_cents: 111, disputed_at: 10.days.ago, purchases: [purchase]) }
+
+        let(:event) { build(:charge_event_dispute_won, charge_id: transaction_id) }
+
+        it "persists chargeback_reversed flag on the disputed purchases" do
+          expect do
+            charge.handle_event(event)
+          end.to change { charge.reload.dispute_reversed_at }.from(nil).to(be_present)
+             .and change { purchase.reload.chargeback_reversed }.from(false).to(true)
```
vishaltelangre, line 942: This assertion would fail without the above fix.

vishaltelangre:
@hchhabra _a Please review.

Issue #2861, opened 2026-01-14
Center "Powered by Gumroad" footer


vatsalkaushik:
- [ ] Center "Powered by Gumroad" footer on the `/purchases/[purchase_id]/confirm_receipt_email` page

<img width="2826" height="1624" alt="Image" src="https://github.com/user-attachments/assets/4ae3d2b4-f29e-41e2-8c75-bef34c663590" />

PR #2864, opened 2026-01-14
fix: Center "Powered by Gumroad" footer 

TenzDelek:
Issue: #2861 

# Description

## Problem
The Footer is not center align

## Solution
Add Center Align style.

---

# Before/After

| Before | After |
|-----------|-----------|
| image | image

| Before | After |
|-----------|-----------|
| image| image

| Before | After |
|-----------|-----------|
| image | image |

| Before | After |
|-----------|-----------|
| image | image|
---

# Test Results

Style changes only
---

# Checklist

---

# AI Disclosure

### DIFF
--- a/app/controllers/churn_controller.rb
+++ b/app/controllers/churn_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class ChurnController < Sellers::BaseController
+  layout "inertia"
+
+  before_action :check_payment_details
+
+  def show
+    authorize :churn
+
+    LargeSeller.create_if_warranted(current_seller)
+
+    service = CreatorAnalytics::Churn.new(seller: current_seller)
+
+    start_date = parse_date(params[:from])
+    end_date = parse_date(params[:to])
+
+    render(
+      inertia: "Churn/Show",
+      props: {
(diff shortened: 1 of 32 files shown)


app/views/purchases/confirm_receipt_email.html.erb
```diff
@@ -11,5 +11,7 @@
   <% end %>
 </main>
 <% content_for :footer do %>
-  <%= default_footer_content %>
+  <div class="text-center pb-2">
```
TenzDelek, line 14: Should we keep this as a CSS class or as an inline style? I’m seeing a mix across our ERB templates. some use inline styles, others use classes.
cc @EmCousin @binary-koan @nyomanjyotisa

EmCousin: This is a special case: these are email templates, and they work a bit differently from our other ERB views because CSS isn’t processed the same way. Since the recipient reads the email in their mailbox, we can’t rely on compiled CSS assets. We use [Premailer](https://github.com/fphilipe/premailer-rails) to inline styles from CSS files, but it doesn’t support Tailwind, so the quickest fix here is to use inline CSS.

PR #2871, opened 2026-01-14
Fix incorrect card styling on the checkout form

binary-koan:
See https://github.com/antiwork/gumroad/pull/2558#issuecomment-3744119649 (@sm17p thanks for the ping!)

# Description

## Problem

The changes in #2558 caused a couple of layout issues with the checkout form: a top border was unintentionally added to the card details form and the PayPal pay button stopped being full width.

---

# Before/After

| Before | After |
| --- | --- |
| Screenshot 2026-01-14 at 14 32 29 | Screenshot 2026-01-14 at 14 31 40 |
| Screenshot 2026-01-14 at 14 32 43 | Screenshot 2026-01-14 at 14 36 18 |

---

# AI Disclosure

Not used for this PR.

### DIFF
--- a/app/javascript/components/Checkout/PaymentForm.tsx
+++ b/app/javascript/components/Checkout/PaymentForm.tsx
@@ -1240,7 +1240,7 @@ export const PaymentForm = ({
       <EmailAddress card />
       {!isFreePurchase ? (
         <>
-          <CardContent className="border-b-0">
+          <CardContent className={state.paymentMethod === "card" ? "border-b-0" : ""}>
             <div className="flex grow flex-col gap-4">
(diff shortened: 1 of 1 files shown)


app/javascript/components/Checkout/PaymentForm.tsx
```diff
@@ -693,10 +693,7 @@ const CreditCard = ({ card }: { card?: boolean }) => {
   if (state.paymentMethod !== "card") return null;
 
   return (
-    <div
-      style={{ borderTop: "none", paddingTop: "0" }}
```
binary-koan, line 697: Tailwind `divide` uses bottom rather than top borders, so changed to removing the border below

PR #2880, opened 2026-01-14
Fix: Double scroll area on product list page 

TenzDelek:
Issue: #2757

# Description
On mobile, the products list page shows two vertical scroll areas. 

## Problem

The scroll issue is due the absolute position style in the sr-only element. The element is hanging around as an invisible block outside the view port which is then causing the scroll.

## Solution
use either relative or fixed positioning to contain the absolute positioning.

---

# Before

https://github.com/user-attachments/assets/26dd126d-7d48-4ef3-97f6-0df9cdc7ad79

# After

https://github.com/user-attachments/assets/6d692a28-153e-4ff3-b8de-d27f8271c5a9

---

# Test Results

---

# Checklist

---

# AI Disclosure

No AI was used for any part of this contribution


TenzDelek:
### walkthrough of how I got to the solution: 
> might be helpful for other contributors in future

1. I started by reviewing ```productlayout.tsx```, suspecting a layout related issue.
2. I compared it with ```emaillayout.tsx``` since both use the Table component.
3. Structurally, both looked almost identical, except emaillayout.tsx didn’t include a TableFooter.
4. That made the footer feel like a likely suspect, so I removed TableFooter from the product layout… but the issue still persisted.
5. Next, I noticed that the TableBody in the product page conditionally rendered some cells.
6. To rule out conditional rendering, I replaced it with a standard TableCell containing dummy text, still no improvement.
7. At one point, removing a specific TableCell did make the problem disappear, which felt like progress.
8. But removing that cell wasn’t a real fix, we obviously couldn’t ship that. 
9. At this stage, one thing was clear: the way the cells were being rendered was triggering the issue.
10. I replaced all table cells with a dummy version like:
```<TableCell>hi</TableCell>```
12. That worked immediately.
13. So I narrowed it down further,  meaning one of the real cells was responsible.
14. The second cell stood out since it had an extra prop that none of the others had. I removed it, and the issue was fixed.
15. However, the expected behavior was to hide the label, so instead of removing the prop entirely, I passed an empty string to keep the intended behavior consistent.

After nearly four hours of debugging, it actually feels amazing. It’s currently 1:01 AM 😄 , and finally solving an “easy yet difficult” problem feels incredibly satisfying. Honestly, I learn a lot. One thing that really stuck with me is that a huge part of solving a bug isn’t just opening the code and writing a fix but its more about understanding what’s causing the behavior in the first place. Debugging is tedious, but it’s also really rewarding.

really enjoyed working on this. thank you @nyomanjyotisa  , @EmCousin for your hint in https://github.com/antiwork/gumroad/pull/2761#discussion_r2677009195

edit: realize the screenreader might be needed. so we can just use the relative positioning to avoid the absolute overflow

**Comment by TenzDelek** [comment]
did some extra research and found out this conversation. there seems to be two solution for this issue.
1. either to use ```relative``` as mentioned https://github.com/tailwindlabs/tailwindcss/discussions/12429#discussioncomment-7587178 . 
2. or to use ```fixed``` beside the ```sr-only``` styling. as mentioned in https://github.com/tailwindlabs/tailwindcss/discussions/12429#discussioncomment-9354260

which one should we go ahead with? 
cc @EmCousin  @binary-koan @nyomanjyotisa

File: app/javascript/components/ui/Table.tsx [hunk 1]
```diff
@@ -165,7 +165,7 @@ export const TableCell = ({
   return (
     <td
       className={classNames(
-        "block p-4 text-left align-middle not-first:border-t not-first:border-border lg:table-cell lg:border-t lg:border-border",
+        "relative block p-4 text-left align-middle not-first:border-t not-first:border-border lg:table-cell lg:border-t lg:border-border",
```
EmCousin, line 168: This is going to force every table cell to a relative position with no exception. That's not what we want; we want to avoid the table cell's `sr-only` _content_ to be in an `absolute` position by default because this the root cause of the unwanted scrollbar.

app/javascript/components/ui/Table.tsx
EmCousin, line 175: This is the fix we need: we only force the cell content that is `sr-only` to have a relative position. We don't touch the position of the cell itself.
```suggestion
        <div className={classNames("mb-2 font-bold lg:hidden", hideLabel && "relative sr-only")}>
```

EmCousin:
Please open another PR when the requested changes are addressed

PR #2885, opened 2026-01-15
Feature: membership restart warning

MuskanPaliwal:
### Problem

Currently all the users who have a canceled membership, don't see a way to resume their subscription and end up buying a new subscription. This leads to data duplicacy, user losing his purchase history, and increase in support load.

### Solution

Earlier in PR #2729, we introduced a `resume` membership flow for users with canceled membership by interrupting the `subscribe` button click flow and prompting user with a modal to either resume or start afresh. but after this discussion [comment](https://github.com/antiwork/gumroad/pull/2347#issuecomment-3724441879) by @binary-koan, the simplest and feasible solution is to restart the subscription automatically on checkout based on email. 

#### Backend:
  `app/services/purchase/base_service.rb` here we add membership restart functionality:

  1. `find_restartable_subscription(giftee_purchase)`: New method that finds an existing deactivated subscription that can be restarted instead of creating a new one. It:
    - Returns `nil` for installment payments, test purchases, or deleted products
    - Looks for subscriptions that are deactivated but not ended
    - Only considers subscriptions cancelled by the buyer (not by admin/seller)
    - Matches by user first, then by email (case-insensitive)
  2. `restart_existing_subscription(subscription, giftee_purchase)`:  New method that restarts an existing subscription:
    - Updates the credit card (unless it's a gift)
    - Calls `subscription.resubscribe!` to reactivate
    - Associates the new purchase with the subscription
    - Sends restart notifications via `subscription.send_restart_notifications!`
  3. Modified `create_subscription(giftee_purchase)`: Now checks for a restartable subscription first before creating a new one. If found, it restarts the existing one instead.

This allows buyers who previously cancelled their membership to restart it (preserving their history) rather than creating a duplicate subscription.

# Test Results

Screenshot 2026-01-15 at 1 36 58 PM

# Checklist

---

# AI Disclosure

I have used Claude Sonnet 4.5 for writing tests and understanding gumroad codebase.


app/services/purchase/base_service.rb
```diff
@@ -30,6 +30,12 @@ def handle_purchase_success
     def create_subscription(giftee_purchase)
       return if purchase.subscription.present?
 
+      existing_subscription = find_restartable_subscription(giftee_purchase)
+      if existing_subscription
+        restart_existing_subscription(existing_subscription, giftee_purchase)
+        return
+      end
```
MuskanPaliwal, line 37: Instead of always creating a new subscription, we first check if there's an existing one we can restart. This avoids duplicate subscriptions cluttering up the seller's dashboard and preserves the buyer's history.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
       subscription.purchases << [purchase, giftee_purchase].compact
     end
 
+    def find_restartable_subscription(giftee_purchase)
+      return nil if purchase.is_installment_payment
+      return nil if purchase.is_test_purchase?
+      return nil if purchase.link.deleted?
```
MuskanPaliwal, line 79: Skipping restart for installment payments, test purchases, and deleted products. These are edge cases where restarting doesn't make sense or could cause issues.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+      is_gift = purchase.is_gift_sender_purchase
+      user = is_gift ? giftee_purchase&.purchaser : purchase.purchaser
+      email = is_gift ? giftee_purchase&.email : purchase.email
+
+      base_scope = purchase.link.subscriptions
+        .not_is_test_subscription
+        .where.not(deactivated_at: nil)
+        .where(ended_at: nil)
+        .where("cancelled_at IS NULL OR (subscriptions.flags & ?) != 0", Subscription.flag_mapping["flags"][:cancelled_by_buyer])
+        .order(deactivated_at: :desc)
```
MuskanPaliwal, line 90: We only auto-restart if the buyer cancelled it themselves. If a seller or admin cancelled it, there's probably a reason (refund, policy violation, etc.) so we create a fresh subscription instead. The order(`deactivated_at: :desc`) picks the most recent one if they have multiple old subscriptions.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+
+      if user.present?
+        subscription = base_scope.find_by(user: user)
+        return subscription if subscription
+      end
+
+      return nil if email.blank?
+
+      base_scope.find { |s| s.email&.downcase == email.downcase }
```
MuskanPaliwal, line 99: Matching by `user` first, then falling back to `email` for guest checkouts. The email comparison is case-insensitive since people often type their email differently.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +73,41 @@ def create_subscription(giftee_purchase)
...
+    end
+
+    def restart_existing_subscription(subscription, giftee_purchase)
+      is_gift = purchase.is_gift_sender_purchase
+
+      subscription.credit_card = purchase.credit_card unless is_gift
+      subscription.resubscribe!
+      subscription.purchases << [purchase, giftee_purchase].compact
+      subscription.send_restart_notifications!
+    end
```
MuskanPaliwal, line 109: updates the payment method, reactivate, and notify everyone. One thing I wasn't sure about: should we update the tier/price if they picked a different one this time around? Right now we just reactivate at whatever tier they had before.

spec/services/purchase/base_service_spec.rb
MuskanPaliwal, line 1: Tried to cover the main edge cases, cancelled by seller vs buyer, ended subscriptions, guest purchases, multiple old subscriptions, etc. Let me know if I'm missing any scenarios you can think of.

MuskanPaliwal:
Hi @binary-koan , could you please review the PR ?

binary-koan:
Thanks for working on this! Unfortunately I've realised we need to think through the issue more before we accept contributions (we talked about it during live reviews the other day but the solution we came up with there is not necessarily the right one either). Sorry we didn't plan out the issue well enough before opening it to contributions, please keep an eye on the thread and we'll try to confirm what's needed

PR #2893, opened 2026-01-15
Fix: Double scroll area on product list page 

TenzDelek:
Issue: #2757

Previous PRs: https://github.com/antiwork/gumroad/pull/2880

# Description
On mobile, the products list page shows two vertical scroll areas. 

## Problem

The scroll issue is due the absolute position style in the sr-only element. The element is hanging around as an invisible block outside the view port which is then causing the scroll.

## Solution
use either relative or fixed positioning to contain the absolute positioning.

---

# Before

https://github.com/user-attachments/assets/26dd126d-7d48-4ef3-97f6-0df9cdc7ad79

# After

https://github.com/user-attachments/assets/6d692a28-153e-4ff3-b8de-d27f8271c5a9

---

# Test Results

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [ ] I have added/updated tests for my changes

---

# AI Disclosure

No AI was used for any part of this contribution

PR #2894, opened 2026-01-15
fix: product page double scroll bar

abhix4:
Issue: #2757
Ref: https://mskelton.dev/bytes/preventing-overflow-with-sr-only

# Description

The products list page shows two vertical scroll areas ( small screens only) .

## Problem

The document body starts scrolling, even though the cards themselves already live inside a scroll container.

The root cause is the use of the `sr-only` utility inside table cells.

The sr-only class looks something like this:

.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border-width: 0;
}
`.sr-only uses position: absolute`

Because the table cells do not establish a positioning context (position: relative), 
these sr-only elements are positioned relative to the nearest positioned ancestor,
which ends up being the document body.

## Solution

Added `position: relative` to table cells so that `sr-only` elements (which use absolute positioning) are scoped correctly within the scroll container instead of the document body.

---

# Before/After

BEFORE:

Mobile: 

https://github.com/user-attachments/assets/5e0a98b0-9197-47d2-afa9-ae00e26c5919

Desktop:

https://github.com/user-attachments/assets/b9bf4ef6-393a-4110-932f-ce003fd9a41f

AFTER:

Mobile:

https://github.com/user-attachments/assets/270ba5f1-b5ac-44f3-acb7-3acad6f039fc

Desktop:

https://github.com/user-attachments/assets/07763262-92ca-4e21-a42f-3ec8d6167685

---

# Test Results

---

# Checklist

---

# AI Disclosure

Used GPT for researching the issue.

### DIFF
--- a/app/javascript/components/ui/Table.tsx
+++ b/app/javascript/components/ui/Table.tsx
@@ -172,7 +172,7 @@ export const TableCell = ({
       {...props}
     >
       {label || cellContext?.label ? (
-        <div className={classNames("mb-2 font-bold lg:hidden", hideLabel && "sr-only")}>
+        <div className={classNames("mb-2 font-bold lg:hidden", hideLabel && "relative sr-only")}>
           {label ?? cellContext?.label}
         </div>
       ) : null}


EmCousin:
Thanks @abhix4, but this was already resolved in https://github.com/antiwork/gumroad/pull/2893

Issue #2843, opened 2026-01-12
Fix installment plan tax test to run without TaxJar credentials


gianfrancopiana:
### Context

One test in the installment plan spec fails because it calculates US sales tax via TaxJar without using a VCR cassette, causing a live API call that fails without credentials.

### Problem

The test "calculates and charges sales tax for all installment payments when applicable" expects sales tax to be calculated for a Wisconsin zip code, but the TaxJar API call fails without `TAXJAR_API_KEY`. 

Other TaxJar-dependent tests pass because they use the `:taxjar` RSpec tag which wraps them in VCR cassettes.

### Acceptance Criteria

- [ ] The installment plan tax test passes without real TaxJar credentials
- [ ] Either add appropriate RSpec metadata tag (`:taxjar`) to use existing VCR infrastructure, or mock the tax calculation
- [ ] No `TAXJAR_API_KEY` secret required for this test to pass

PR #2846, opened 2026-01-12
Fix(test): adds  `:taxjar` tag to "calculates and charges sales tax for all installment payments when applicable" test to run it without real `TAXJAR_API_KEY` key

romitg2:
fixes: #2843 

# Description

-  Add `:taxjar` RSpec metadata tag to the installment plan tax test and VCR cassette with recorded TaxJar API responses for Wisconsin tax calculations so it uses VCR cassettes instead of making live API calls

## Problem
-   The test "calculates and charges sales tax for all installment payments when applicable" made live TaxJar API calls without using VCR cassettes, requiring real `TAXJAR_API_KEY` to be set.

## Solution

-  Added `:taxjar` RSpec metadata tag to the installment plan tax test and VCR cassette with recorded TaxJar API responses for Wisconsin tax calculations so it uses VCR cassettes instead of making live API calls. 

---

# Test Results (without credential) 

## Before:
Screenshot 2026-01-13 at 2 15 52 AM

## After: 
Screenshot 2026-01-13 at 2 00 03 AM

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure
- used claude-code with opus-4.5 assist implementing fix

PR #2896, opened 2026-01-15
fix(test): Run installment plan tax test without TaxJar credentials

abhi152003:
ref: #959 

### Problem :

The test "calculates and charges sales tax for all installment payments when applicable" in `spec/requests/purchases/product/installment_plan_spec.rb` was making live TaxJar API calls without using VCR cassettes. 
This caused the test to fail in CI environments where `TAXJAR_API_KEY` is not available.

Other TaxJar-dependent tests pass because they use the `:taxjar` RSpec tag which wraps them in VCR cassettes.

### Solution :

Added the `:taxjar` and `:force_vcr_on` metadata tags to the `describe "tax"` block:

`:taxjar` - Wraps the test in a VCR cassette that records/replays TaxJar API responses
`:force_vcr_on` - Required because VCR is turned off by default for JS system tests

This follows the same pattern used by all other TaxJar-dependent tests in the codebase.

## Test Verification :

CI test results will be used for verification. Local JS system tests are currently 
blocked by a webpack compilation issue unrelated to this change. The fix follows 
the same `:taxjar` VCR pattern used by existing tests in the codebase.

---

# Checklist

---

# AI Disclosure

GLM-4.7 was used to identify the root cause and implement a fix.

### DIFF
--- a/spec/requests/purchases/product/installment_plan_spec.rb
+++ b/spec/requests/purchases/product/installment_plan_spec.rb
@@ -341,7 +341,7 @@ describe "Product with installment plan", type: :system, js: true do
     end
   end
 
-  describe "tax" do
+  describe "tax", :taxjar, :force_vcr_on do
     it "calculates and charges sales tax for all installment payments when applicable" do
       visit product.long_url
 


spec/requests/purchases/product/installment_plan_spec.rb
```diff
@@ -341,7 +341,7 @@
     end
   end
 
-  describe "tax" do
+  describe "tax", :taxjar, :force_vcr_on do
```
abhi152003, line 344: Added `:taxjar` and `:force_vcr_on` metadata tags to enable VCR cassette recording for TaxJar API calls.

`:taxjar` - This tag is defined in `spec_helper.rb` and wraps the test in a VCR cassette (Taxjar/{test description}.yml). This records TaxJar API responses on first run, then replays them on subsequent runs without requiring real credentials.
`:force_vcr_on` - Required because this is a JS system test (js: true), and VCR is turned off by default for JS specs (see `spec_helper.rb`). This flag keeps VCR enabled so the cassette can be used.

This follows the same pattern used by other TaxJar-dependent tests in `spec/requests/checkout/taxes_spec.rb` and other files with the `:taxjar` tag.

abhi152003:
Note: I see there's another PR(#2846) addressing this issue. My approach includes `:force_vcr_on` as well, but I think just `:taxjar` may be sufficient since `vcr_turned_on` is called within the `:taxjar` handler. The intent behind adding that tag was to clearly signal that this test needs VCR to stay on and avoid any timing edge cases.

gianfrancopiana:
Closing as duplicate. PR #2846 was merged yesterday and resolved issue #2843.

That PR included both the `:taxjar` tag and the VCR cassette file with recorded API responses. The cassette is required for tests to pass without credentials - this PR is missing it.

The `:force_vcr_on` tag isn't needed since `:taxjar` already calls `vcr_turned_on`.

PR #2897, opened 2026-01-16
Add docs on fixing 502 errors in test

binary-koan:
See https://github.com/antiwork/gumroad/discussions/2302

Several contributors have run into 502 errors trying to run tests because there aren't docs on running `js:export` to configure the test domains in JS.

Added a couple of notes to the testing doc where I think people will be most likely to find it.


binary-koan:
@ershad _a Does this make sense to you?

Issue #2913, opened 2026-01-16
Login sometimes fails with 404 error


nyomanjyotisa:
https://github.com/user-attachments/assets/90e35dc7-7e43-40fb-a76d-6331790ba8d5

On login button click sometime navigated to https://gumroad.com/two-factor?next=%2Fworkflows, even though the account does not have 2FA enabled

Probably a regression of https://github.com/antiwork/gumroad/pull/2646 or https://github.com/antiwork/gumroad/pull/2410 cc @stefanbinoj @EmCousin

PR #2918, opened 2026-01-16
Bug/two factor 404

EmCousin:
Issue: [#2913](https://github.com/antiwork/gumroad/issues/2913)

# Description

## Problem

The report says login can sometimes redirect to `/two-factor?next=%2Fworkflows` even when the account does not have 2FA enabled, and then it 404s. I was not able to reproduce*the reported behavior locally, but found a plausible scenario where it could happen. While investigating the login/logout flow, I found a separate bug: after logging out and logging back in, the CSRF/authenticity token in the meta tag was be stale, causing subsequent form submissions to be sent with an outdated token.

## Solution
- Commits 1–2 (related bugfix, not exactly #2913 but good to ship):
  - Ensure auth-related forms always submit the correct authenticity token by explicitly passing it through the forms instead of relying on a potentially stale meta value after logout/login.
- Last commit (possible #2913 scenario):
  - Add defensive handling and a system spec around a plausible stale-session/2FA flow so that a user without 2FA enabled is not redirected to `/two-factor` due to leftover state from a prior 2FA attempt.

---

# Before/After
No UI changes.

---

# Checklist

---

# AI Disclosure
AI was used to help investigate the issue, propose fixes, and draft tests/PR description.


app/controllers/application_controller.rb
```diff
@@ -205,6 +205,12 @@ def login_path_for(user)
         return safe_final_path
       end
 
+      # Defensive: if a stale 2FA session is present for another user (e.g. from an aborted login),
+      # do not redirect the current user to the 2FA flow.
+      if session[:verify_two_factor_auth_for].present? && session[:verify_two_factor_auth_for] != user.id
+        reset_two_factor_auth_login_session
```
EmCousin, line 211: In the event there's a stale 2fa session of a user different from the one that is authenticating, reset that 2fa session

app/javascript/pages/Logins/New.tsx
```diff
@@ -14,30 +14,36 @@ type PageProps = {
...
-    "g-recaptcha-response": string | null;
-  }>({
+  const form = useForm<FormData>({
     user: {
       login_identifier: initialEmail ?? "",
       password: "",
     },
     next,
     "g-recaptcha-response": null,
+    authenticity_token,
```
EmCousin, line 45: Forcing the authentication token in the request's body. The one set in the header in the interceptor can be stale if the user logs out, since the link to logout is an Inertia button that does not refresh the head.

app/javascript/pages/TwoFactorAuthentication/Show.tsx
```diff
@@ -10,19 +10,27 @@ type PageProps = {
   user_id: string;
   email: string;
   token: string | null;
+  authenticity_token: string;
+};
+
+type FormData = {
+  token: string;
+  next: string | null;
+  authenticity_token: string;
```
EmCousin, line 19: Same treatment for the 2fa form

EmCousin:
@nyomanjyotisa _a I wasn't able to reproduce your issue locally but I encountered a few login-related issues while investigating, and I added a guard against potentially stale 2fa sessions that _may_ be the cause of what you are experiencing.

If you have more details about how to reproduce this locally and deterministically, this would help, in the mean time this PR will bring some improvements. Could you review it?

PR #2059, opened 2025-11-18
Feat: Auto Apply Discount Code

Pradumn27:
fixes #2053

This PR implements “auto-applied discount codes” as described in the issue. A product can have a default discount code that’s automatically applied and displayed everywhere, but is overridden by any explicit URL discount code, and its usage is tracked on purchases.

### Video

https://github.com/user-attachments/assets/68a7a33a-8ae5-43cf-8479-495da31f3e0e

### Tests

E2E Request specs

`bundle exec rspec spec/requests/purchases/product/default_discount_code_spec.rb`

Screenshot 2025-11-18 at 5 53 40 PM

Models, Presenter and Service specs

`bundle exec rspec spec/controllers/links_controller_spec.rb -e "default discount code" spec/models/link_spec.rb:546 spec/presenters/product_presenter/card_spec.rb -e "with default discount code" spec/presenters/product_presenter/product_props_spec.rb -e "with default discount code" spec/presenters/product_presenter_spec.rb -e "with default offer code"`

Screenshot 2025-11-18 at 7 08 47 AM

### AI Disclosure

- Model (cursor GPT 5.1)
- Used for writing tests and also for refining the code
- All code written was reviewed and updated manually by me

### Livestream Viewing Confirmation

- Have watched all of them

PR #2920, opened 2026-01-16
Feature: Auto apply discount code to products

Pradumn27:
Issue: #2053 

# Description

## Problem

A Feature to allow sellers to add a default discount code to their products, which gets automatically applied for everyone purchasing that product.

## Solution

Added a default discount code field to the product(link) model, now this gets auto-evaluated on purchases if added - also if a customer applies a discount code better than the default one, then the better one prevails and that gets applied.

Additionally, created an API to fetch the available discount codes for a product based on the search query and also added the validation for checking whether a discount code is applied to a product as default before deleting it

---

# Videos

1. Applying Default Discount Code

https://github.com/user-attachments/assets/56800380-fe06-4789-b3fd-17297f3e5d3e

2. Tracking The Code

https://github.com/user-attachments/assets/5014f939-3009-4d30-a948-2068ea25fc51

3. Deleting Added Default Discount Code

https://github.com/user-attachments/assets/5ac4703b-9507-40e8-94f4-5fa9075c6a5e

---

# Test Results

`bundle exec rspec spec/controllers/links_controller_spec.rb:1160 spec/presenters/product_presenter/card_spec.rb spec/presenters/product_presenter_spec.rb spec/requests/products/edit/preview_spec.rb:231 spec/requests/purchases/product/default_discount_code_spec.rb spec/services/purchase/create_service_spec.rb:2552 spec/models/offer_code_spec.rb:754 spec/controllers/checkout/discounts_controller_spec.rb`

Screenshot 2026-01-16 at 8 08 56 PM

---

# Checklist

---

# AI Disclosure

Used Sonnet 4.5 for test coverage code - the rest was manually added with help of tab autocomplete at places

### DIFF
--- a/app/presenters/product_presenter/product_props.rb
+++ b/app/presenters/product_presenter/product_props.rb
@@ -104,10 +104,10 @@ class ProductPresenter::ProductProps
     end
 
     def evaluate_code(code, quantity)
-      return { valid: false, error_code: "missing_code" } if code.blank?
+      return { valid: false, error_code: :missing_code } if code.blank?
 
       offer_code = product.find_offer_code(code: code)
-      return { valid: false, error_code: "invalid_offer" } unless offer_code
+      return { valid: false, error_code: :invalid_offer } unless offer_code
 
       response = OfferCodeDiscountComputingService.new(
         code,
@@ -120,7 +120,7 @@ class ProductPresenter::ProductProps
       ).process
 
       if response[:error_code].present?
-        return { valid: false, error_code: response[:error_code].to_s }
+        return { valid: false, error_code: response[:error_code] }
       end
 
       {


app/controllers/checkout/discounts_controller.rb
```diff
@@ -82,10 +82,10 @@ def destroy
     offer_code = OfferCode.find_by_external_id!(params[:id])
     authorize [:checkout, offer_code]
 
-    if offer_code.mark_deleted(validate: false)
+    if offer_code.mark_deleted
```
Pradumn27, line 85: unsure why validation was not being done here, added it here since we now we validate on deleting offer codes which are already set as default offer code in existing products

app/controllers/products/available_offer_codes_controller.rb
Pradumn27, line 1: Created a new API to fetch the discount codes based on search params to show in the dropdown for selecting default discount codes

app/controllers/links_controller.rb
```diff
@@ -695,6 +697,25 @@ def update_installment_plan
...
+      raise Link::LinkInvalid, "Offer code cannot be expired" if offer_code.inactive?
+      raise Link::LinkInvalid, "Offer code must be associated with this product or be universal" unless valid_for_product?(offer_code)
+
+      @product.default_offer_code = offer_code
+    rescue ActiveRecord::RecordNotFound
+      raise Link::LinkInvalid, "Invalid offer code"
+    end
+
+    def valid_for_product?(offer_code)
+      offer_code.universal? || @product.offer_codes.where(id: offer_code.id).exists?
```
Pradumn27, line 708: we validate if the offer code is either universal or present for the product

app/javascript/components/Product/Card.tsx
```diff
@@ -48,6 +48,7 @@ export const Card = ({
           url={product.url}
           currencyCode={product.currency_code}
           price={product.price_cents}
+          oldPrice={product.original_price_cents ?? undefined}
```
Pradumn27, line 51: In case of default discount code we utilize the existing logic in cards and just pass the original price as the old price

app/javascript/components/Product/index.tsx
```diff
@@ -265,6 +265,11 @@ export const Product = ({
 
   const notForSaleMessage = getNotForSaleMessage(product);
   const [discountCode, setDiscountCode] = React.useState(initialDiscountCode);
+
+  React.useEffect(() => {
+    setDiscountCode(initialDiscountCode);
+  }, [initialDiscountCode]);
```
Pradumn27, line 271: The Product component initializes its internal discountCode state from the initialDiscountCode prop, but React's useState doesn't automatically update internal state when props change. This useEffect ensures that when the ProductPreview updates the discount code (e.g., when a user selects a default discount code in the editor), the Product component's display immediately reflects the change by synchronizing its internal state with the new prop value. Without this synchronization, the preview would not update when discount codes are selected.

app/javascript/components/ProductEdit/state.ts
```diff
@@ -143,6 +151,8 @@ export type Product = {
   call_limitation_info: CallLimitationInfo | null;
   require_shipping: boolean;
   cancellation_discount: CancellationDiscount | null;
+  default_offer_code_id?: string | null;
+  default_offer_code: OfferCode | null;
```
Pradumn27, line 155: both are present here because we get the `default_offer_code` object from the presenter but when we perform an update query we pass the id instead of the object

app/models/link.rb
```diff
@@ -819,8 +821,17 @@ def find_offer_code_by_external_id(external_id)
   # Public: Find all alive offer codes associated with product and user in order of created at.
   #
   # Returns list of offer codes.
-  def product_and_universal_offer_codes
-    (offer_codes.alive + user.offer_codes.universal_with_matching_currency(price_currency_type).alive).sort_by(&:created_at)
+  def product_and_universal_offer_codes(query = nil, limit = nil)
```
Pradumn27: Updated the existing method to now accept an optional search param and limit and query based on that

app/models/offer_code.rb
```diff
@@ -301,4 +309,12 @@ def capture_associated_product_ids
     def reindex_captured_products
       reindex_associated_products(products_to_reindex: Link.where(id: @product_ids_to_reindex)) if @product_ids_to_reindex.present?
     end
+
+    def validate_not_used_as_default_discount
```
Pradumn27, line 314: validation function for deleting offer codes, which are still a default discount code in some products

app/presenters/product_presenter/product_props.rb
```diff
@@ -80,24 +80,54 @@ def props(seller_custom_domain_url:, request:, pundit_user:, recommended_by: nil
...
+      return url_code_result if url_code_result&.dig(:valid) == true && default_code_result&.dig(:valid) != true
+      return default_code_result if default_code_result&.dig(:valid) == true && url_code_result&.dig(:valid) != true
+
+      return url_code_result if url_code_result&.dig(:valid) == false && default_code_result&.dig(:valid) == false
+
+      url_code_amount = url_code_result[:amount_off_cents]
+      default_code_amount = default_code_result[:amount_off_cents]
+
+      url_code_amount > default_code_amount ? url_code_result : default_code_result
+    end
```
Pradumn27, line 104: This function in the presenter executes the logic for applying the most optimal discount code, comparing both the default code and the code supplied via query

db/migrate/20260115011936_add_index_to_links_default_discount_code_id.rb
Pradumn27, line 1: The index helps in the queries added here for validation checks on discount code deletion

spec/controllers/checkout/discounts_controller_spec.rb
```diff
@@ -427,13 +427,32 @@ def transform_offer_code_props(offer_code_props)
         offer_code.save(validate: false)
       end
 
-      it "returns HTTP success and marks the offer code as deleted" do
+      it "returns HTTP unprocessable entity and prevents deletion" do
```
Pradumn27, line 430: since the controller now validates, so it throws errors on deleting invalid codes

Pradumn27 (commented):
Hi @EmCousin, please have a look when you get a chance, thanks!

gianfrancopiana:
Closing this in favor of #2824 which takes a simpler approach.                                                                                                                                                            
  - #2824 has a smaller, more focused diff (559 vs 2531 additions)
  - Consistent column naming (`default_offer_code_id` matches the association)
  - Follows the issue spec where URL discount takes priority (vs "better discount wins" logic here)
  
Some good ideas from this PR (searchable dropdown, preview in editor) could be considered as future enhancements.

Pradumn27:
> Closing this in favor of #2824 which takes a simpler approach.
> 
>     * [Auto apply discount codes #2824](https://github.com/antiwork/gumroad/pull/2824) has a smaller, more focused diff (559 vs 2531 additions)
> 
>     * Consistent column naming (`default_offer_code_id` matches the association)
> 
>     * Follows the issue spec where URL discount takes priority (vs "better discount wins" logic here)
> 
> 
> Some good ideas from this PR (searchable dropdown, preview in editor) could be considered as future enhancements.

I wanted to add some context here, @gianfrancopiana.

The “prefer better discount” behavior was intentionally changed later by @ershad during a live stream and was also discussed in the [initial PR](https://github.com/antiwork/gumroad/pull/2059). So the logic in this PR reflects that updated decision rather than the original issue description.

Regarding the larger diff:

This PR introduces the new searchable offers API, which matches the issue design exactly(please check that).

A significant portion of the diff (1500+ lines) is VCR cassettes added for purchase-related tests, not core logic changes, which is also missing in other PR I believe.

Because of this, while the diff is larger, it’s largely driven by test coverage and shared UI infrastructure, not added complexity. Given that, I believe this PR aligns more closely with the final intended behavior and design. please have a look again, thanks!

Pradumn27:
exact [comment](https://github.com/antiwork/gumroad/pull/2059#discussion_r2539064555) for reference - @gianfrancopiana

app/controllers/products/available_offer_codes_controller.rb
gianfrancopiana: I think we're missing test coverage for this new API endpoint. Please add controller specs for:                                            
  - Successful search with results                                 
  - Empty search results                                           
  - Authorization (non-owner can't access)                         
  - The `MAX_OFFER_CODES_LIMIT` being enforced

db/migrate/20251118045164_add_default_discount_code_id_to_links.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultDiscountCodeIdToLinks < ActiveRecord::Migration[7.1]
+  def change
+    add_column :links, :default_discount_code_id, :bigint
```
gianfrancopiana: The column is named `default_discount_code_id` but the association in `link.rb` is `belongs_to :default_offer_code`.

```suggestion
    add_column :links, :default_offer_code_id, :bigint
```

Please rename the column for consistency.

Pradumn27:
Hey @gianfrancopiana made the default offer code naming consistent throughout and also added the remaining specs for the available offer codes API, please review - thanks!

app/javascript/data/offer_code.ts
```diff
@@ -46,6 +46,29 @@ export type OfferCodeResponseData =
...
+
+export const searchProductOfferCodes = async (
+  productPermalink: string,
+  query: string,
+): Promise<ProductOfferCodeSuggestion[]> => {
+  const params = new URLSearchParams({ query });
+  const response = await request({
+    method: "GET",
+    accept: "json",
+    url: `/products/${encodeURIComponent(productPermalink)}/available_discount_codes.json?${params.toString()}`,
```
gianfrancopiana: Found a bug while testing locally – the API endpoint URL doesn't match the route.
```suggestion
    url: `/products/${encodeURIComponent(productPermalink)}/available_offer_codes.json?${params.toString()}`,
```

app/javascript/components/ProductEdit/ProductTab/DefaultDiscountCodeSelector.tsx
```diff
@@ -0,0 +1,140 @@
+import * as React from "react";
```
gianfrancopiana, line 1: The dropdown only shows results after typing. Users should see available options immediately.

**Expected behavior:**
- **On focus (empty input):** Show the 5 most recently created discount codes
- **While typing:** Filter/search as it does now (up to 20 results)
- **When input is cleared:** Show the 5 most recent again (not close the dropdown)

**Changes needed:**

1. **Frontend — `DefaultDiscountCodeSelector.tsx`:**
   - Add an `onFocus` handler that calls `fetchOptions("")` when the input is empty and no options are loaded
   - When the input is cleared, make sure users see the initial results again

2. **Backend — `AvailableOfferCodesController`:**
   - Return 5 results when query is blank, 20 when searching

3. **Backend — `Link#product_and_universal_offer_codes`:**
   - Sort by most recently created (newest first) instead of oldest first

app/models/offer_code.rb
gianfrancopiana: This validation fails when creating offer codes with `deleted_at` set. For new records, `id` is nil, so `Link.visible.where(default_offer_code_id: nil)` matches all links without a default – causing the validation to incorrectly fail.

app/presenters/product_presenter/product_props.rb
```diff
@@ -80,24 +80,55 @@ def props(seller_custom_domain_url:, request:, pundit_user:, recommended_by: nil
...
-        { valid: true, code: discount_code, **offer_code_response[:products_data][product.unique_permalink] }
+      if response[:error_code].present?
+        return { valid: false, error_code: response[:error_code] }
       end
+
+      {
+        valid: true,
+        code: code,
+        discount: offer_code.discount,
+        amount_off_cents: offer_code.amount_off(product.price_cents)
```
gianfrancopiana: This is a leaky abstraction – `amount_off_cents` is an internal value used to compare which discount is better, but it's being included in the API response. The frontend doesn't need it and the original implementation didn't expose it.

app/presenters/product_presenter/product_props.rb
```diff
@@ -80,24 +80,55 @@ def props(seller_custom_domain_url:, request:, pundit_user:, recommended_by: nil
...
-        discount_code,
+      return if url_code.blank? && default_code.blank?
+
+      url_code_result = evaluate_code(url_code, quantity)
+      default_code_result = evaluate_code(default_code, quantity)
+
+      url_code_valid = url_code_result&.dig(:valid) == true
+      default_code_valid = default_code_result&.dig(:valid) == true
+
+      return unless url_code_valid || default_code_valid
```
gianfrancopiana: This returns `nil` when both codes are invalid, but the original behavior returned `{ valid: false, error_code: ... }`. This breaks error handling for invalid/expired/sold-out codes.

The error response needs to be preserved – if the URL code is invalid and there's no default, return the URL code's error result instead of nil.

gianfrancopiana (changes requested):
Tested locally and reviewed CI run. Left inline comments on the specific issues.

app/javascript/components/ProductEdit/ProductTab/DefaultDiscountCodeSelector.tsx
gianfrancopiana: One more thing – what happens if a user has many discount codes and types a common letter like "A" that matches 20+ codes? I see no indication that results are truncated.

I suggest adding `max-height` to the dropdown (around 250px, roughly 6 items visible) with scroll for overflow. This is should handles edge cases gracefully without limiting search functionality. Please double check.

Pradumn27: @gianfrancopiana — isn’t not showing discount codes until the user starts typing part of the intended design per the issue?
That’s why the placeholder reads “Begin typing to select a discount code.”

Screenshot 2026-01-20 at 8 56 11 PM

app/javascript/components/ProductEdit/ProductTab/PriceEditor.tsx
```diff
@@ -96,6 +99,7 @@ export const PriceEditor = ({
           onNumberOfInstallmentsChange={onNumberOfInstallmentsChange}
         />
       ) : null}
+      {productEditContext ? <DefaultDiscountCodeSelector /> : null}
```
Pradumn27, line 102: need to only show the selector for product edit and not bundle edit

app/javascript/components/ProductEdit/ProductTab/DefaultDiscountCodeSelector.tsx
gianfrancopiana: I understand the placeholder text says "Begin typing," but looking at the design mockup you shared there's a dropdown chevron icon on the right side of the input. This is a standard UX affordance that indicates:                 
1. Clicking the field should reveal available options
2. Typing is optional – it filters the list, but isn't required to see options                                                      

This is how comboboxes work across the web (e.g., country selectors, autocomplete address fields). The chevron tells users "click here to see choices."                                     

For consistency with both the design and our existing patterns (see `Search.tsx` which fetches on focus), please:
1. Show 5 recent codes on focus (empty input)
2. Add the dropdown chevron icon to match the design
3. Keep the typing behavior for filtering

app/javascript/components/ComboBox.tsx
```diff
@@ -113,7 +115,12 @@ export const ComboBox = <Option extends unknown>({
             }),
       })}
       <div hidden={!open} onMouseDown={(e) => e.preventDefault()}>
-        <datalist id={uid} onMouseOut={() => setFocusedOptionIndex(null)} aria-multiselectable={multiple}>
+        <datalist
+          id={uid}
+          onMouseOut={() => setFocusedOptionIndex(null)}
+          aria-multiselectable={multiple}
+          style={maxHeight ? { maxHeight } : undefined}
```
Pradumn27, line 122: added optional `maxHeight` which can be used in the future as well

app/javascript/components/ProductEdit/ProductTab/DefaultDiscountCodeSelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+        <section className="flex flex-col gap-4">
+          <fieldset>
+            <label htmlFor="default-discount-code">Discount code</label>
+            <ComboBox<OfferCode>
+              editable
+              open={isOpen ? options.length > 0 : false}
+              onToggle={setIsOpen}
+              className="w-full"
+              options={options}
+              maxHeight="16rem"
```
Pradumn27, line 88: added a maxHeight of `16rem` instead of `250px` as it also shows an indication of truncation with 6+ items and also using rem is better

Pradumn27:
Hey @gianfrancopiana fixed the only failure, was serializing the error code in the presenter itself, reverted that, please run the CI again, also the test failure (Test Slow 25) was unrelated to the change, so a different flaky test

Issue #2924, opened 2026-01-16
Improve navigation by decoupling sidebar and content scroll


ershad:
Currently, when we click a page in the sidebar, it doesn't scroll to the top of the new page. To improve this UX:

- [ ] Split the sidebar and content into two independent scroll areas
- [ ] Scroll the content to the top when selecting a new page from the sidebar

PR #2929, opened 2026-01-16
fix: scroll issue in page navigation

alina-alyy:
Issue: #2924 

# Description

## Problem

Sidebar navigation didn’t reset the scroll position of the main content area `#main-content`, which uses `overflow-y-auto`. As a result, when scrolling down to Settings and then navigating to the Products page, the page retained the scroll position from the previous page instead of resetting to the top. Inertia’s default scroll restoration only targets the `window/body` and does not account for custom scroll containers.

## Solution

- Added scroll-region attribute to the `<main>` element for native Inertia scroll tracking.
- Added a manual scroll reset on `inertia:finish` to guarantee a full reset to the top.

# Before

https://github.com/user-attachments/assets/015f917e-59b2-40ea-b479-8833ee960769

# After

https://github.com/user-attachments/assets/a2de7d00-adfd-4f59-9336-8e0529586ff3

# Checklist

---

# AI Disclosure

Used Claude Opus 4.5 to investigate the issue flow and identify possible solutions.

### DIFF
--- a/app/javascript/inertia/layout.tsx
+++ b/app/javascript/inertia/layout.tsx
@@ -1,7 +1,6 @@
-import { Head, usePage } from "@inertiajs/react";
+import { Head, router, usePage } from "@inertiajs/react";
 import React from "react";
 
-import { classNames } from "$app/utils/classNames";
 
 import { Nav } from "$app/components/client-components/Nav";
 import { CurrentSellerProvider, parseCurrentSeller } from "$app/components/CurrentSeller";
@@ -35,6 +34,14 @@ export default function Layout({ children }: { children: React.ReactNode }) {
   const isRouteLoading = useRouteLoading();
 
   useFlashMessage(flash);
+  React.useEffect(() => {
+    return router.on("finish", () => {
+      const mainContent = document.getElementById("main-content");
+      if (mainContent) {
+        mainContent.scrollTop = 0;
+      }
+    });
+  }, []);
 
   return (
     <LoggedInUserProvider value={parseLoggedInUser(logged_in_user)}>
@@ -43,8 +50,9 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         <Alert initial={null} />
         <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
-          {isRouteLoading ? <LoadingSkeleton /> : null}
-          <main className={classNames("flex-1 overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>
+          <main id="main-content" scroll-region className="flex-1 overflow-y-auto">
+            {isRouteLoading ? <LoadingSkeleton /> : children}
+          </main>
         </div>
       </CurrentSellerProvider>
     </LoggedInUserProvider>


app/javascript/inertia/layout.tsx
```diff
@@ -35,6 +34,14 @@ export default function Layout({ children }: { children: React.ReactNode }) {
...
 
   useFlashMessage(flash);
+  React.useEffect(() => {
+    return router.on("finish", () => {
+      const mainContent = document.getElementById("main-content");
+      if (mainContent) {
+        mainContent.scrollTop = 0;
+      }
+    });
+  }, []);
```
EmCousin, line 44: ⚠️ Sometimes Inertia visits are sent with `preserveScroll: true`, and this breaks that behaviour.

app/javascript/inertia/layout.tsx
```diff
@@ -43,8 +50,9 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         <Alert initial={null} />
         <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
-          {isRouteLoading ? <LoadingSkeleton /> : null}
-          <main className={classNames("flex-1 overflow-y-auto", { hidden: isRouteLoading })}>{children}</main>
+          <main id="main-content" scroll-region className="flex-1 overflow-y-auto">
```
EmCousin, line 53: A self-review is needed to explain what this is trying to do. Probably related to this: https://inertiajs.com/docs/v2/advanced/scroll-management#scroll-regions

app/javascript/inertia/layout.tsx
```diff
@@ -35,6 +34,14 @@ export default function Layout({ children }: { children: React.ReactNode }) {
   const isRouteLoading = useRouteLoading();
 
   useFlashMessage(flash);
+  React.useEffect(() => {
+    return router.on("finish", () => {
+      const mainContent = document.getElementById("main-content");
+      if (mainContent) {
+        mainContent.scrollTop = 0;
+      }
+    });
```
EmCousin, line 43: _Returning_ this code means it gets executed when the component is being unmounted. Is that intended?

app/javascript/inertia/layout.tsx
alina-alyy: I’ve updated the handler to respect `event.detail.visit?.preserveScroll` and only reset scroll when it’s not explicitly preserved.
```suggestion
  React.useEffect(() => {
    const removeListener = router.on("finish", (event) => {
      if (event.detail.visit?.preserveScroll) return;

      if (mainContentRef.current) {
        mainContentRef.current.scrollTop = 0;
      }
    });

    return () => {
      removeListener();
    };
  }, []);
```

app/javascript/inertia/layout.tsx
alina-alyy: With scroll behavior now handled explicitly in `useEffect`, the `scroll-region` attribute became redundant and was removed. The code now uses useRef rather than `document.getElementById`.
```suggestion
         <main id="main-content" ref={mainContentRef} className="flex-1 overflow-y-auto" >
```

app/javascript/inertia/layout.tsx
alina-alyy: The effect now correctly attaches the listener on mount and cleans it up on unmount. The scroll reset only runs on navigation finish and no longer executes during unmount.
```suggestion
    React.useEffect(() => {
    const removeListener = router.on("finish", (event) => {
      if (event.detail.visit?.preserveScroll) return;

      if (mainContentRef.current) {
        mainContentRef.current.scrollTop = 0;
      }
    });

    return () => {
      removeListener();
    };
  }, []);
```

alina-alyy:
@EmCousin If you’re okay with these changes, would you mind reopening the PR so I can push the new commits?

EmCousin:
@alina-alyy you can push to your branch and open a new PR, please reference this one in its description. Thanks!

PR #2950, opened 2026-01-18
Support messaging to and from a React Native webview

binary-koan:
# Description

Similar to our current iOS and Android webview handlers, this adds React Native webview integration to the product content page.

* When clicking a button we send an event to the webview in order to trigger the native share dialog / audio playback / PDF reader / etc
* When we receive an audio playback update message from the webview we update the React media state

No visual changes.

---

# AI Disclosure

Claude Opus 4.5 used for some of the refactoring.

### DIFF
--- a/app/javascript/components/Download/FileList.tsx
+++ b/app/javascript/components/Download/FileList.tsx
@@ -443,7 +443,12 @@ const MobileAppAudioFileRow = ({ file }: { file: FileItem }) => {
 
   const messageListener = useRefToLatest((event: MessageEvent) => {
     if (typeof event.data !== "string" || !event.data.startsWith("{")) return;
-    const data: unknown = JSON.parse(event.data);
+    let data: unknown;
+    try {
+      data = JSON.parse(event.data);
+    } catch {
+      return;
+    }
     if (is<{ type: "mobileAppAudioPlayerInfo"; payload: MobileAppAudioPlayerInfo }>(data)) {
       if (data.payload.fileId !== file.id) return;
       setIsPlaying(data.payload.isPlaying);


app/javascript/components/Download/Interactions.tsx
```diff
@@ -50,42 +56,48 @@ export const NativeAppLink = <Element extends HTMLElement>({
...
-        window.webkit.messageHandlers.jsMessage.postMessage({
-          type: "click",
-          payload: {
+      const appPayload = openInApp
+        ? ({
             resourceId,
             isDownload: eventName === "download_click",
             isPost: eventName === "post_click",
             type,
+            extension: file?.extension ?? null,
```
binary-koan, line 75: This is newly added since it makes detecting whether to open the PDF reader easier (the apps currently need to check the event payload against the API response to find the file type)

To make it work I changed the props to take the `file` or `post` object rather than just the `resourceId` - I considered just adding an `extension` prop but this is used in a bunch of places and it seemed more error prone if we need to add that separate prop everywhere

app/javascript/components/Download/Interactions.tsx
```diff
@@ -50,42 +56,48 @@ export const NativeAppLink = <Element extends HTMLElement>({
...
+          window.CustomJavaScriptInterface.onPostClickedEvent(appPayload.resourceId);
         } else {
-          window.CustomJavaScriptInterface.onFileClickedEvent(resourceId, eventName === "download_click");
+          window.CustomJavaScriptInterface.onFileClickedEvent(appPayload.resourceId, appPayload.isDownload);
         }
+      } else if (window.ReactNativeWebView && appPayload) {
+        e.stopPropagation();
+        e.preventDefault();
+        // Open in the React Native app
+        window.ReactNativeWebView.postMessage(JSON.stringify({ type: "click", payload: appPayload }));
```
binary-koan, line 101: react-native-webview requires message data to be a string

app/javascript/components/Download/FileList.tsx
```diff
@@ -440,6 +439,24 @@ const MobileAppAudioFileRow = ({ file }: { file: FileItem }) => {
...
+    if (is<{ type: "mobileAppAudioPlayerInfo"; payload: MobileAppAudioPlayerInfo }>(data)) {
+      if (data.payload.fileId !== file.id) return;
+      setIsPlaying(data.payload.isPlaying);
+      setLatestMediaLocation(parseFloat(data.payload.latestMediaLocation ?? "0"));
+    }
+  });
+
+  React.useEffect(() => {
+    const listener = (e: MessageEvent) => messageListener.current(e);
+    // @ts-expect-error - React Native sends message events to embedded webviews via the document object, not window
```
binary-koan: This is a bit of a pain, but I think it's better to have a descriptive comment here rather than extending the type definition

Inlining the `useEffect` since our nice `useGlobalCustomEventListener` hook only adds the listener to `window`

app/javascript/components/Download/Interactions.tsx
```diff
@@ -50,42 +56,49 @@ export const NativeAppLink = <Element extends HTMLElement>({
...
-        });
-      } else if (window.CustomJavaScriptInterface && openInApp) {
+          } satisfies ClickPayload)
+        : undefined;
+
+      if (window.webkit?.messageHandlers?.jsMessage && appPayload) {
+        e.stopPropagation();
+        e.preventDefault();
+        // Open in the iOS app
+        const { extension, ...rest } = appPayload;
```
binary-koan, line 86: It's *probably* fine to include the new field in the iOS event payload, but it's hard to test locally and I'd rather not take chances

binary-koan (commented):

app/javascript/components/Download/FileList.tsx
```diff
@@ -440,6 +440,24 @@ const MobileAppAudioFileRow = ({ file }: { file: FileItem }) => {
     }
   });
 
+  const messageListener = useRefToLatest((event: MessageEvent) => {
+    if (typeof event.data !== "string" || !event.data.startsWith("{")) return;
+    const data: unknown = JSON.parse(event.data);
```
nyomanjyotisa: nit: might be worth wrapping `JSON.parse` in a try/catch so malformed JSON doesn’t crash the webview

app/javascript/components/Download/FileList.tsx
```diff
@@ -440,6 +441,30 @@ const MobileAppAudioFileRow = ({ file }: { file: FileItem }) => {
...
+    }
+    if (is<{ type: "mobileAppAudioPlayerInfo"; payload: MobileAppAudioPlayerInfo }>(data)) {
+      if (data.payload.fileId !== file.id) return;
+      setIsPlaying(data.payload.isPlaying);
+      setLatestMediaLocation(parseFloat(data.payload.latestMediaLocation ?? "0"));
+    }
+  });
+
+  React.useEffect(() => {
+    const target = Mobile.isOnAndroidDevice() ? document : window;
```
binary-koan, line 460: Turns out Android needs `document` and iOS needs `window` 🙄 Will need to try and add an integration test in Expo to catch changes to this behaviour

PR #2974, opened 2026-01-19
fix: reset scroll on navigation with useEffect cleanup

alina-alyy:
Issue: #2924 

# Description

This is a follow-up PR for the previous one, implementing the requested changes. [PR Link](https://github.com/antiwork/gumroad/pull/2929)

## Problem

Sidebar navigation didn’t reset the scroll position of the main content area `#main-content`, which uses `overflow-y-auto`. As a result, when scrolling down to Settings and then navigating to the Products page, the page retained the scroll position from the previous page instead of resetting to the top. Inertia’s default scroll restoration only targets the `window/body` and does not account for custom scroll containers.

## Solution

- Added a useEffect listener on `router.on("finish")` to reset the `<main>` scroll to the top after navigation.
- Switched from `document.getElementById` to `useRef` for React-friendly access to the scroll container.
- Ensured `preserveScroll: true` visits are respected, so scroll is only reset when appropriate.

# Before

https://github.com/user-attachments/assets/29a8cbfe-ae4e-4363-bd93-6d82941fef46

# After

https://github.com/user-attachments/assets/8bcfe803-edf0-420a-94eb-874d0ea3a766

# Checklist

---

# AI Disclosure

Used Claude Opus 4.5 to investigate the issue flow and identify possible solutions.

### DIFF
--- a/app/javascript/inertia/layout.tsx
+++ b/app/javascript/inertia/layout.tsx
@@ -32,15 +32,22 @@ type PageProps = {
 export default function Layout({ children }: { children: React.ReactNode }) {
   const { title, flash, logged_in_user, current_seller } = usePage<PageProps>().props;
   const isRouteLoading = useRouteLoading();
+  const mainContentRef = React.useRef<HTMLDivElement>(null);
 
   useFlashMessage(flash);
+
   React.useEffect(() => {
-    return router.on("finish", () => {
-      const mainContent = document.getElementById("main-content");
-      if (mainContent) {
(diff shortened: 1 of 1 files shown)


binary-koan:
Thanks for working on this, but the issue is about the download page (`/d/:id`) rather than the main dashboard sidebar. I do think we should preserve scroll in the dashboard sidebar, but it would be better to use Inertia's `preserveScroll` attribute instead of adding a custom hook (we should try to follow their conventions as much as possible)

PR #2976, opened 2026-01-19
fix(layout): scrolling

skyfallwastaken:
Issue: #2924

# Description

## Problem

When scrolling to the bottom of a long page (e.g. Analytics), then switching to another item in the sidebar, you end up being taken to ~the middle.

## Solution

This PR fixes that! Now you get taken to the top of the page.

---

# Before/After

## Before

https://github.com/user-attachments/assets/c5e2b0f2-350f-460e-bbea-bde553214d51

## After

https://github.com/user-attachments/assets/56d474d4-b301-4410-8eff-0ec8ddda3d2a

---

# Test Results

It's a 13 line change, so the tests would probably be longer than the fix! Same thing for the self-review.

---

# Checklist

---

# AI Disclosure

I used OpenCode with Opus 4.5 to find the layout file (I'm more of a Turbo guy than an Inertia guy :P)

### DIFF
--- a/app/javascript/components/Download/PageListLayout.tsx
+++ b/app/javascript/components/Download/PageListLayout.tsx
@@ -18,7 +18,7 @@ export const PageListLayout = ({
       className,
     )}
   >
-    <div className="flex flex-col gap-4 lg:sticky lg:top-0 lg:w-80 lg:pb-8">{pageList}</div>
+    <div className="flex h-full flex-col gap-4 lg:w-80 lg:overflow-y-auto lg:pb-8">{pageList}</div>
     <div className="h-0 flex-1">{children}</div>
   </div>
 );


skyfallwastaken:
cc @ershad / @emcousin :D

app/javascript/components/useRouteLoading.tsx
```diff
@@ -2,12 +2,14 @@ import React from "react";
...
 const useRouteLoading = () => {
   const [isRouteLoading, setIsRouteLoading] = React.useState(false);
+  const shouldScrollToTopRef = React.useRef(false);
 
   React.useEffect(() => {
     const startHandler = (event: DocumentEventMap["inertia:start"]) => {
       const { prefetch, only = [], preserveScroll } = event.detail.visit;
-      const shouldPreserveScroll = preserveScroll !== false;
-      setIsRouteLoading(!prefetch && !shouldPreserveScroll && only.length === 0);
+      const isFullPageNavigation = !prefetch && preserveScroll !== true && only.length === 0;
```
skyfallwastaken, line 10: We do handle `preserveScroll`!

ershad:
Interesting, the original issue was to fix the product download page for the buyer, but it seems like we have the same problem on creator pages too. Thanks for fixing this. In the case of creator pages, we don't need a separate scroll for the left sidebar; we just need to scroll the right side to the top when navigating to a different page.

@EmCousin _a could you please review this PR in detail?

skyfallwastaken:
> In the case of creator pages, we don't need a separate scroll for the left sidebar

I didn't add that part, the same behaviour is on gumroad.com too 😄

binary-koan:
Thanks for working on this and sorry it's taken so long for us to get to it. Similar to my comment on https://github.com/antiwork/gumroad/pull/2974, I would rather add Inertia's `preserveScroll` attribute to sidebar links instead of doing custom logic - we should work with Inertia as much as possible rather than working around it

PR #2966, opened 2026-01-19
Decoupling pageList Tab and contentEditor Tab scrolling

Mit-Patwa:
Issue: #2924 

## Problem
Currently, the pagelist tab and content editor tabs scroll together. This creates a poor UX, as users cannot scroll each section independently.

## Solution

Decouple the two tabs so they scroll independently, providing a better experience for sellers.---

# Before
https://github.com/user-attachments/assets/1a2aebed-829e-4457-83bb-cebca4ccbf90

---

# After

https://github.com/user-attachments/assets/ee4a47bf-b007-4213-b20e-e0cea8c6cda4

---

# Test Results

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [ ] I have added/updated tests for my changes

---

# AI Disclosure

No AI was used for any part of this PR.

PR #2983, opened 2026-01-20
Decoupling sidebar and content viewer/editor tab scrolling

Mit-Patwa:
Issue: #2924 
Previoius PR: #2966 

## Problem
Currently, the pagelist tab and content editor tabs scroll together. This creates a poor UX, as users cannot scroll each section independently.

## Solution

Decouple the two tabs so they scroll independently, providing a better experience for sellers.---

# Before

seller:

https://github.com/user-attachments/assets/1a2aebed-829e-4457-83bb-cebca4ccbf90

buyer:

https://github.com/user-attachments/assets/6840d727-9af8-4bd7-9e63-f1f819b7e8c8

---

# After

seller:

https://github.com/user-attachments/assets/ee4a47bf-b007-4213-b20e-e0cea8c6cda4

buyer:

https://github.com/user-attachments/assets/f217ee7a-905d-49f9-be8b-b5c5a6bb198d

---

# Checklist

---

# AI Disclosure

No AI was used for any part of this PR.

### DIFF
--- a/app/javascript/components/Download/RichContent.tsx
+++ b/app/javascript/components/Download/RichContent.tsx
@@ -69,7 +69,7 @@ export const RichContentView = ({
 
   return (
     <LicenseProvider value={licenseInfo}>
-      <EditorContent className="rich-text grid h-full flex-1" editor={editor} />
+      <EditorContent className="rich-text grid h-screen overflow-y-auto border border-border p-2" editor={editor} />
     </LicenseProvider>
   );
 };


Mit-Patwa:
@binary-koan could you please review,Thanks!

app/javascript/components/Download/PageListLayout.tsx
```diff
@@ -19,7 +21,7 @@ export const PageListLayout = ({
     )}
   >
     <div className="flex flex-col gap-4 lg:sticky lg:top-0 lg:w-80 lg:pb-8">{pageList}</div>
-    <div className="h-0 flex-1">{children}</div>
+    <div className={classNames("h-0 flex-1", editContentClassName)}>{children}</div>
```
Mit-Patwa, line 24: conditional addition of utilities since `PageListLayout` is used in other places which don't need these utilities

app/javascript/components/Download/RichContent.tsx
```diff
@@ -69,7 +69,7 @@ export const RichContentView = ({
 
   return (
     <LicenseProvider value={licenseInfo}>
-      <EditorContent className="rich-text grid h-full flex-1" editor={editor} />
+      <EditorContent className="rich-text grid h-screen overflow-y-auto border border-border p-2" editor={editor} />
```
Mit-Patwa, line 72: added utilities, to allow independent scrolling(no conditional addition since component used only in one place)

binary-koan:
Looking at the videos again I don't think we want it this way round, it would work better like the /api page where the sidebar is the inner (sticky) scroll area and the content is the main scroll area. Could you try that?

I also don't think we need the border if we do that

PR #3024, opened 2026-01-22
fix: reset scroll position on sidebar navigation in Inertia layout

abhix4:
Issue: #2924

When navigating via the sidebar after scrolling down a long page, the new page sometimes loads scrolled to the middle instead of the top.

## Problem

During navigation we were:
- showing a loading skeleton
- hiding <main> with CSS instead of unmounting it

Because the element stayed mounted, its internal scroll position (scrollTop) was preserved.
So when the new page rendered, it reused the old scroll position.

## Solution

unmount `<main> `entirely while the route is loading, remount it once navigation finishes

---

# Before/After

**Before:** 

https://github.com/user-attachments/assets/281236a3-7951-43d1-a217-bdf770f0f55e

**After:**

https://github.com/user-attachments/assets/257677f2-b162-4fe4-aefe-f539f7194d78

---

# Test Results

---

# Checklist

---

# AI Disclosure

None

### DIFF
--- a/app/javascript/inertia/layout.tsx
+++ b/app/javascript/inertia/layout.tsx
@@ -43,8 +43,13 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         <Alert initial={null} />
         <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
-          {isRouteLoading ? <LoadingSkeleton /> : null}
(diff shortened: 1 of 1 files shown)


abhix4:
@EmCousin This is ready for review. Thank you.

app/javascript/inertia/layout.tsx
EmCousin, line 4: This is no longer used, causing the app to break.

PR #3037, opened 2026-01-22
Show country-specific minimum payout threshold in settings and payouts

gianfrancopiana:
Fixes https://github.com/antiwork/gumroad-private/issues/10

## Problem

Some countries have minimum payout amounts higher than Gumroad's standard $10 USD (e.g., Serbia requires ~$30 USD). Previously:
- The UI always showed "$10" as the minimum, which was incorrect for some countries
- Users in high-minimum countries would see confusing error messages
- The Payouts empty state showed a hardcoded amount instead of directing users to settings

## Solution

- Display the actual country-specific minimum in Payment Settings
- Show the country name in validation error messages (e.g., "The minimum payout threshold for Serbia is $30.")
- Update Payouts empty state to link to settings instead of showing a dynamic amount

## Additional UX Improvement

The payout threshold input now allows users to clear the field while editing. Previously, entering a null value would immediately reset to the minimum.

### Payouts Empty State

**Before:**

**After:**

## From Validation Demo
CleanShot 2026-01-23 at 13 44 07

## Test Plan

- [x] Navigate to Payment Settings as a user from a high-minimum country (e.g., Serbia)
- [x] Verify hint text shows country-specific minimum
- [x] Enter amount below minimum → verify error shows country name
- [x] Clear input field → verify can type new value without immediate reset
- [x] Blur with empty field → verify resets to minimum
- [x] Navigate to Payouts with no balance → verify empty state links to settings

---

## AI Assistance Notice

### DIFF
--- a/app/javascript/pages/Settings/Payments/Show.tsx
+++ b/app/javascript/pages/Settings/Payments/Show.tsx
@@ -736,10 +736,6 @@ export default function PaymentsPage() {
   const payoutThresholdError =
     form.data.payout_threshold_cents != null && form.data.payout_threshold_cents < props.minimum_payout_threshold_cents;
 
-  const handlePayoutThresholdChange = (value: number | null) => {
-    form.setData("payout_threshold_cents", value);
-  };
-
   const handlePayoutThresholdBlur = () => {
     if (!form.data.payout_threshold_cents) {
       form.setData("payout_threshold_cents", props.minimum_payout_threshold_cents);
@@ -870,7 +866,13 @@ export default function PaymentsPage() {
         <section className="p-4! md:p-8!">
           <header>
             <h2>Payout schedule</h2>
-            <p>Payouts will only happen on your chosen schedule once the minimum balance is reached.</p>
+            <p>
+              Payouts will only happen on your chosen schedule once the minimum balance of{" "}
+              {formatPriceCentsWithCurrencySymbol("usd", props.minimum_payout_threshold_cents, {
+                symbolFormat: "long",
+              })}{" "}
+              is reached.
+            </p>
           </header>
           <section className="flex flex-col gap-4">
             <fieldset>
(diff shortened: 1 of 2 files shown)


app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -887,16 +907,23 @@ export default function PaymentsPage() {
...
               {payoutThresholdError ? (
                 <small>
-                  Your payout threshold must be at least{" "}
+                  The minimum payout threshold for {props.payout_country_name ?? "your country"} is{" "}
+                  {formatPriceCentsWithCurrencySymbol("usd", props.minimum_payout_threshold_cents, {
+                    symbolFormat: "long",
+                  })}
+                  .
+                </small>
+              ) : props.minimum_payout_threshold_cents > 1000 ? (
```
EmCousin: We enter this condition if there is no threshold error but we only display this message when it is above 1000 ? 
Could we simplify and always show "The minimum payout threshold for ..." since when there is an error that text is show in red?

app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -720,7 +728,18 @@ export default function PaymentsPage() {
     }
   }, [isPayoutMethodChangeConfirmed]);
 
-  const payoutThresholdError = form.data.payout_threshold_cents < props.minimum_payout_threshold_cents;
+  const payoutThresholdError =
+    form.data.payout_threshold_cents != null && form.data.payout_threshold_cents < props.minimum_payout_threshold_cents;
+
+  const handlePayoutThresholdChange = (value: number | null) => {
+    form.setData("payout_threshold_cents", value);
+  };
```
EmCousin: I think we can inline that in the input's onChange event listener since it's not re-used anywhere.

app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -720,7 +728,18 @@ export default function PaymentsPage() {
...
+
+  const handlePayoutThresholdBlur = () => {
+    if (!form.data.payout_threshold_cents) {
+      form.setData("payout_threshold_cents", props.minimum_payout_threshold_cents);
```
EmCousin, line 741: If there is no payout threshold cents initially, should `minimum_payout_threshold_cents` be a default value when calling `useForm` ? If so, would it be simpler to invoke `form.reset("payout_threshold_cents")` and inline it in the onBlur event listener?

app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -847,6 +866,7 @@ export default function PaymentsPage() {
         <section className="p-4! md:p-8!">
           <header>
             <h2>Payout schedule</h2>
+            <p>Payouts will only happen on your chosen schedule once the minimum balance is reached.</p>
```
EmCousin: Can we tell what the minimum balance is here?

gianfrancopiana:
@EmCousin Thanks for the feedback! Please see latest changes.

PR #3043, opened 2026-01-23
Chore(inertia):   Migrates the Checkout Form page from custom API calls to Inertia's useForm hook 

romitg2:
Issue: #3586

# Description

- Migrated the Checkout Form page from custom API calls to Inertia's useForm hook for form state management and submission

---

# Before/After

## Before: 

https://github.com/user-attachments/assets/4b4c808b-cda0-440b-a496-87bc2e4fdc20

## After: 

https://github.com/user-attachments/assets/8e708457-11a8-4149-b8a9-38b2583ef74a

---

# Test Results

` bundle exec rspec spec/controllers/checkout/form_controller_spec.rb spec/presenters/checkout/form_presenter_spec.rb spec/policies/checkout/form_policy_spec.rb`

Screenshot 2026-01-22 at 5 53 48 PM

`bundle exec rspec  spec/requests/checkout/form_spec.rb `

Screenshot 2026-01-22 at 5 52 26 PM

---

# Checklist

---

# AI Disclosure

- used claude-code opus 4.5 to help refactoring code

### DIFF
--- a/spec/controllers/url_redirects_controller_spec.rb
+++ b/spec/controllers/url_redirects_controller_spec.rb
@@ -648,7 +648,7 @@ describe UrlRedirectsController do
           expect(response).to have_http_status(:ok)
           expect(response.body).to include("data-page=")
 
-          page_data_match = response.body.match(/data-page=\"([^\"]*)\"/)
+          page_data_match = response.body.match(/data-page="([^"]*)"/)
           expect(page_data_match).to be_present, "Expected Inertia.js data-page attribute"
 
           page_data = JSON.parse(CGI.unescapeHTML(page_data_match[1]))


app/javascript/components/CheckoutDashboard/FormPage.tsx
```diff
@@ -49,64 +68,99 @@ const FormPage = ({
...
+  const [invalidFields, setInvalidFields] = React.useState<Set<string>>(() => new Set());
+
+  const form = useForm<FormData>({
+    user: {
+      display_offer_code_field,
+      recommendation_type,
+      tipping_enabled,
+    },
+    custom_fields: custom_fields.map(addKey),
+  });
```
romitg2, line 82: Replaces custom fetch-based API call with Inertia's useForm hook for form state management and submission.

app/javascript/components/CheckoutDashboard/FormPage.tsx
```diff
@@ -49,64 +68,99 @@ const FormPage = ({
...
+
+    form.put(Routes.checkout_form_path(), {
+      preserveScroll: true,
+      onError: (errors: Record<string, string | string[]>) => {
+        const baseError = errors.base;
+        const message = (Array.isArray(baseError) ? baseError[0] : baseError) ?? "Failed to save changes";
+        showAlert(message, "error");
+      },
+    });
+  };
```
romitg2, line 145: Changed from async/await with custom fetch to inertia's `form.put()` for form submission. 
Uses form.transform() to strip the key property before sending.

app/javascript/data/checkout_form.ts
romitg2, line 1: fetch-based API call is no longer needed. Inertia's useForm handles the submission directly.

app/javascript/packs/inertia.js
```diff
@@ -5,6 +5,19 @@ import { createRoot } from "react-dom/client";
...
+  if (typeof globalThis.__activeRequests === "number") {
+    ++globalThis.__activeRequests;
+  }
+});
+
+router.on("finish", () => {
+  if (typeof globalThis.__activeRequests === "number") {
+    --globalThis.__activeRequests;
+  }
+});
```
romitg2: Adds request tracking for the `wait_for_ajax` test helper. (some tests require `wait_for_ajax` helper otherwise tests were getting flaky).

app/controllers/checkout/form_controller.rb
romitg2, line 1: Migrated from JSON API to Inertia-style redirects. Success now redirects with flash notice, errors redirect with flash alert and Inertia errors.

app/javascript/components/CheckoutDashboard/FormPage.tsx
romitg2, line 1: ### File Change Summary: 

- Replaced custom fetch API call with form.put()
- CustomField type moved here from deleted checkout_form.ts
- Refactored error tracking.  Changed from Map<field.key, Set<errorType>> to simpler Set<string> with index-based paths (e.g., custom_fields.0.name) for error management. 
- Success now handled via server redirect with flash message instead of client-side showAlert; errors handled via onError callback

romitg2:
@EmCousin could you please review this PR ? Thank you!

app/javascript/components/CheckoutDashboard/FormPage.tsx
```diff
@@ -49,64 +68,99 @@ const FormPage = ({
...
+    customFields.forEach((field, index) => {
+      if (!field.name) {
+        newInvalidFields.add(`custom_fields.${index}.name`);
+      }
       if (field.type === "terms") {
         try {
           new URL(field.name);
         } catch {
-          fieldErrors.add("name");
+          newInvalidFields.add(`custom_fields.${index}.name`);
```
Pradumn27: Using index for errors causes an issue here. When deleting a custom field with errors, the errors don't get cleared for that index and the errors get transferred to the one that moved up to the index. 

Consider this scenario:
1. Add two custom fields (Field A at index 0, Field B at index 1)                                      
2. Leave Field A's name empty and click Save → custom_fields.0.name error appears                      
3. Delete Field A                                                                                      
4. Field B (now at index 0) incorrectly shows the validation error

app/javascript/packs/inertia.js
Pradumn27: I think the reason for any flaky tests here would likely be due to consecutive saves in a test, and alerts don't disappear immediately. When a consecutive test clicks 'Save' and checks for the 'Changes saved!' alert, it matches the stale alert from the previous save instead of waiting for the new request to complete.

I was also trying things for this, it would work with both waiting for the alert to be removed `expect(page).not_to have_alert("Changes saved!")` or adding `sleep 0.5`(not preffered).

This one looks good, solves it globally for tests.

But if the reasoning is anything else, then it's probably an issue with the implementation. Please test and confirm

app/javascript/components/CheckoutDashboard/FormPage.tsx
romitg2: makes sense! updated

app/javascript/packs/inertia.js
romitg2: I did some debugging and found that issue is not really related to alert timing. if i add very small delay (`sleep 0.05`) instead wait_for_ajax then it fixes the flakiness. so may be there is some issue with reading data from database during test. and wait_for_ajax introduces some slight delay and that way it seems to be addressing flakiness.

Pradumn27: I went through it again, it's indeed a transaction delay issue, and our `have_alert` instantly matches on consecutive saves, thus checking everything before the new transaction has committed. The above solution looks fine solves it globally by tracking the requests

app/javascript/components/CheckoutDashboard/FormPage.tsx
```diff
@@ -49,64 +68,102 @@ const FormPage = ({
...
-  const [errors, setErrors] = React.useState<Map<string, Set<string>>>(new Map());
 
-  const handleSave = asyncVoid(async () => {
-    const errors = new Map<string, Set<string>>();
-    for (const field of customFields) {
-      const fieldErrors = new Set<string>();
-      if (!field.name) fieldErrors.add("name");
+  const updateUserData = (update: Partial<FormData["user"]>) => form.setData("user", { ...form.data.user, ...update });
+
+  const handleSave = () => {
```
Pradumn27:
```suggestion
  const submitForm = () => {
```

slavingia:
How's this looking?

romitg2:
> How's this looking?

@EmCousin gentle ping! PR is ready for review 🙏

app/javascript/components/CheckoutDashboard/FormPage.tsx
```diff
@@ -49,64 +68,102 @@ const FormPage = ({
...
-        user: { displayOfferCodeField, recommendationType, tippingEnabled },
-        customFields,
-      });
-      setCustomFields(response.custom_fields.map(addKey));
-      showAlert("Changes saved!", "success");
-    } catch (e) {
-      assertResponseError(e);
-      showAlert(e.message, "error");
-    }
-    setIsSaving(false);
```
EmCousin, line 100: Could you manage invalid fields using Inertia's form errors? https://inertia-rails.dev/guide/forms#form-errors
This would be much simpler

app/controllers/checkout/form_controller.rb
```diff
@@ -15,24 +15,24 @@ def show
...
+        current_seller.custom_fields.reload
       end
-      render json: Checkout::FormPresenter.new(pundit_user:).form_props
-    rescue ActiveRecord::RecordInvalid => e
-      render json: { error_message: e.record.errors.full_messages.to_sentence }, status: :unprocessable_entity
     end
+
+    redirect_to checkout_form_path, status: :see_other, notice: "Changes saved!"
+  rescue ActiveRecord::RecordInvalid => e
+    redirect_to checkout_form_path, inertia: { errors: { base: e.record.errors.full_messages } }, alert: e.record.errors.full_messages.to_sentence
```
EmCousin, line 35: You are passing the inertia errors as part of the redirection but you are not using them on the frontend, I think you are not sure of what you're doing here, this is for [server-side validation](https://inertia-rails.dev/guide/forms#server-side-responses)

You can use our helper `inertia_errors` from InertiaRendering to format the errors properly so when exposed to the frontend, you can retrieve them automatically in `form.errors` and show the errored inputs properly

romitg2: actually i was trying to mover validation logic to server-side, but i thought it's not worth doing since for checkout-form we can dynamically add custom fields and for each of those field we've to write valiadtion logic and manage errors for each field in server. so it didn't seems worth doing for this case. 
@EmCousin can you please confirm should i move validation logic to server, or keep validatoin in frontend only?

romitg2:
> Thanks @romitg2 - see the comments below. I've ran CI in the mean time. Also, this PR targets the parent issue, can you please mention the specific child issue it is targeting? Let me know if it is missing.

Thanks @EmCousin for review! the associated issue is missing.

app/javascript/components/CheckoutDashboard/FormPage.tsx
romitg2: I think for our checkout form we've to use local state for validation errors instead of Inertia's form.errors 
because We use string keys (field.key) to track errors, but Inertia expects array indices (0, 1, 2...)
- If we used indices and user deletes field `#2`, the error would jump to field `#3` (now at index 2), we we're using string keys stay attached to the correct field regardless of list changes. 


`with index based error tracking (which inertia expects) but in our case we're allowing to delete fields so won't work well. `

https://github.com/user-attachments/assets/f885e8e7-eacf-42d9-ac40-f28c81174226


for the same reason moving validation logic to server side also doesn't seems to be good idea.

EmCousin (approved):
Thanks @romitg2!

PR #3053, opened 2026-01-23
Add onetime script to enable membership renewal reminders

nyomanjyotisa:
Issue: https://github.com/antiwork/gumroad-private/issues/88

# Description

Add a one-time script to enable the `membership_renewal_reminders` feature for sellers and schedule reminder jobs for their existing active subscriptions

## Problem
Enabling the `membership_renewal_reminders` feature flag only affects new subscriptions, leaving existing active subscriptions without scheduled reminder jobs

## Solution
Created `Onetime::EnableMembershipRenewalReminders` service that activates the feature flag for a seller and schedules `RecurringChargeReminderWorker` jobs for all their eligible existing subscriptions, with dry run mode for safe testing

# Result
### Dry Run `true`
Screenshot 2026-01-23 at 20 30 04

### Dry Run `false`
Screenshot 2026-01-23 at 20 31 47

### After (`RecurringChargeReminderWorker` job scheduled)
Screenshot 2026-01-23 at 20 32 41

# AI Disclosure

Cursor - Claude Sonnet 4.5 for code generation, self reviewed by me

```
# DRY RUN
Onetime::EnableMembershipRenewalReminders.perform(user_id: "1")

# LIVE
Onetime::EnableMembershipRenewalReminders.perform(user_id: "1", dry_run: false)
```

### DIFF
--- a/app/services/onetime/enable_membership_renewal_reminders.rb
+++ b/app/services/onetime/enable_membership_renewal_reminders.rb
@@ -36,7 +36,6 @@ module Onetime
 
         begin
           puts "\nSubscription #{subscription.id}"
-          puts "  Product: #{subscription.link&.name}"
           puts "  Customer: #{subscription.email}"
           puts "  Next renewal: #{subscription.end_time_of_subscription.strftime('%Y-%m-%d %H:%M')}"
           puts "  Reminder at: #{subscription.send_renewal_reminder_at.strftime('%Y-%m-%d %H:%M')}"


app/services/onetime/enable_membership_renewal_reminders.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Onetime
+  class EnableMembershipRenewalReminders
+    def self.perform(user_id:, dry_run: true)
+      puts dry_run ? "DRY RUN MODE\n" : "LIVE MODE\n"
+
+      seller = User.find_by(id: user_id) || User.find_by(external_id: user_id)
```
nyomanjyotisa, line 8: Accepts both internal database ID and external ID for flexibility

app/services/onetime/enable_membership_renewal_reminders.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      scheduled = 0
+      errors = 0
+
+      Subscription.where(seller_id: seller.id)
+                  .active
+                  .find_each do |subscription|
+        next unless subscription.alive?(include_pending_cancellation: false)
+        next if subscription.in_free_trial?
+        next if subscription.charges_completed?
```
nyomanjyotisa, line 35: This ensures we only schedule reminders for eligible subscriptions based on the existing logic:
https://github.com/antiwork/gumroad/blob/f39c978166c60cad55e4ede53c2e5f2f217dd28b/app/models/purchase.rb#L3664-L3674

nyomanjyotisa:
@EmCousin _a Could you help to review this?

app/services/onetime/enable_membership_renewal_reminders.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+
+      puts "\n"
+
+      scheduled = 0
+      errors = 0
+
+      Subscription.where(seller_id: seller.id)
+                  .active
+                  .find_each do |subscription|
```
EmCousin, line 32: Nit:
```suggestion
      seller.subscriptions.active.find_each do |subscription|
```

app/services/onetime/enable_membership_renewal_reminders.rb
```diff
@@ -0,0 +1,60 @@
...
+        next unless subscription.alive?(include_pending_cancellation: false)
+        next if subscription.in_free_trial?
```
EmCousin, line 34: I think we can filter out those directly in the database. Leaving this up to you to decide, depending on the number of subscriptions this script is going to loop over.

app/services/onetime/enable_membership_renewal_reminders.rb
```diff
@@ -0,0 +1,60 @@
...
+        next if subscription.charges_completed?
+
+        begin
+          puts "\nSubscription #{subscription.id}"
+          puts "  Product: #{subscription.link&.name}"
```
EmCousin: This is unnecessarily causing an N+1 query. Not worth it just for logging I think.

app/services/onetime/enable_membership_renewal_reminders.rb
nyomanjyotisa: This is incorrect, `Subscription.where(seller_id: seller.id)` not equal to `seller.subscriptions`.

`seller.subscriptions` use `user_id`, while we need subscriptions filter by `seller_id`

EmCousin: Oh good point, my apologies.

PR #3019, opened 2026-01-21
SSR Meta tags

EmCousin:
Issue: https://github.com/antiwork/gumroad-private/issues/82

> [!NOTE]
> Most of LOC changes are search-and-replace operations from `@title = "foo"` to `set_page_title("foo")`. The diff is not that big when substracting those changes.

# Description

This pull request ensures HTML meta tags are both rendered server-side for SEO/link previews/LLM fetchers _and_ updated dynamically client-side from one Inertia visit tot he next.

## Problem

Migrations of public pages to Inertia like this attempt for the [Help Center ](github.com/antiwork/gumroad/pull/2886) end up rendering meta tags client-side, which is bad for SEO/link previews.

Even if Googlebot often executes JavaScript and may eventually see client-rendered meta tags, we shouldn’t rely on that. Other crawlers (social link preview bots, smaller search engines, many LLM fetchers) frequently don’t run JS or don’t wait long enough. Also, page speed matters for SEO, and waiting on JS just to get correct `<title>` / `<meta>` is unnecessary overhead: these tags should be present in the initial HTML response.

## Solution

This PR exposes a single `set_meta_tag` API that builds two identical meta tag stores: one to be rendered in ERB layouts, the other in Inertia initial loads, while allowing dynamic updates on subsequent Inertia navigations. This PR also decouples the complex conditional meta tag logic previously living in `app/views/layouts/application/_head.html.erb`, which relied on instance variables and was hard to reason about.

# Before/After

N/A - system tests and controller tests ensure no regression is happening, tests that were checking instance variable presence now verify which HTML meta tag is rendered.

---

# Test Results

CI is green.

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

This PR was implemented mostly manually, AI was used (Cursor with GPT-5.2) to assist verifying each meta tag has been properly transposed.

PR #3055, opened 2026-01-23
Inertia Migrate Review Reminders pages

### LINKED ISSUE #3040: Migrate Review Reminders page to Inertia
Endpoints:
- [ ] `/users/unsubscribe_review_reminders`
- [ ] `/users/subscribe_review_reminders`

sm17p:
Closes: #3040
- #3040

# Description

This PR migrates the following pages to inertia: 
1. `/users/ubsubscribe_review_reminders`
2. `/users/subscribe_review_reminders`

Fixes the gap between "If you wish to resubscribe to all review reminder emails, please click" and "here" text for unsubscribe page 

## Solution
1. Visiting the pages updates the unsubscribe/subscribe status (matching the original implementation)
 
2. Created ~~dedicated controllers~~ unified controller which update the "Review Reminders" status and render inertia pages
    - GET `/users/unsubscribe_review_reminders` -> `Users::ReviewRemindersController#unsubscribe` 
    - GET `/users/subscribe_review_reminders` -> `Users::ReviewRemindersController#subscribe` 

 
> [!WARNING] 
>  When users navigate back/forward between the two pages using browser or system controls, the page may be served from the browser's back-forward cache (bfcache) or Inertia's cache (unsure have to check). This means the server action won't run again and the user will see a cached confirmation page (which wrongly depicts the current state). 
>
> This seems acceptable for now, but we should probably avoid this pattern in the future (using a form with a submit button would be better for similar scenarios)

### Before

https://github.com/user-attachments/assets/f126cd36-b0e1-4066-92f7-29b7b648e9e6

### After

https://github.com/user-attachments/assets/a64cce23-cbc7-4e7c-b947-e0439f6f3036

### Before

https://github.com/user-attachments/assets/1d708ce6-7b72-4ec3-b84f-6e9ae37adf6c

### After

https://github.com/user-attachments/assets/bf8b1686-6a64-49de-bb6e-6ce526cf50b6

## Test Results

specs_passing

---

# Checklist

---

# AI Disclosure

Cursor auto mode to plan and execute the task


app/javascript/components/EmailAction/Layout.tsx
```diff
@@ -1,32 +1,42 @@
...
+      {isRouteLoading ? <LoadingSkeleton /> : null}
+      <main className={classNames("flex-1 overflow-y-auto", { hidden: isRouteLoading })}>
+        <Card>
+          <CardContent asChild>
+            <header>
+              <h2 className="grow">{heading}</h2>
+            </header>
+          </CardContent>
+          <CardContent asChild>
+            <p className="justify-start gap-1.5!">{children}</p>
```
sm17p: This fixes the gap between 

"If you wish to resubscribe to all review reminder emails, please click" and "here" text for unsubscribe page visible in the after video

app/javascript/components/EmailAction/Layout.tsx
```diff
@@ -1,32 +1,42 @@
...
+          </CardContent>
+        </Card>
+        <footer className="p-4 text-center">
```
sm17p, line 23: Using tailwind over here

app/javascript/components/EmailAction/Layout.tsx
```diff
@@ -1,32 +1,42 @@
...
-          padding: "var(--spacer-4)",
-        }}
-      >
-        Powered by&ensp;
-        <a href={Routes.root_url({ host: rootDomain })} className="logo-full" aria-label="Gumroad" />
-      </footer>
-    </>
+    <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
+      <Head title={title} />
+      {isRouteLoading ? <LoadingSkeleton /> : null}
```
sm17p: Adding a loader to keep it aligned with other pages

app/javascript/components/server-components/ReviewReminders/SubscribeReviewReminders.tsx
sm17p, line 1: Inlined into `Users/SubscribeReviewReminders/Show.tsx` following recent migration patterns

app/javascript/components/server-components/ReviewReminders/UnsubscribeReviewReminders.tsx
sm17p, line 1: Inlined into `Users/UnsubscribeReviewReminders/Show.tsx` following recent migration patterns

app/javascript/packs/subscribe_review_reminders.ts
sm17p, line 1: ReactOnRails cleanup

config/routes.rb
```diff
@@ -345,8 +345,10 @@ def product_info_and_purchase_routes(named_routes: true)
       scope "/users" do
         get "/check_twitter_link", to: "users/oauth#check_twitter_link"
         get "/unsubscribe/:id", to: "users#email_unsubscribe", as: :user_unsubscribe
-        get "/unsubscribe_review_reminders", to: "users#unsubscribe_review_reminders", as: :user_unsubscribe_review_reminders
-        get "/subscribe_review_reminders", to: "users#subscribe_review_reminders", as: :user_subscribe_review_reminders
+        scope module: :users do
+          resource :unsubscribe_review_reminders, only: [:show], as: :user_unsubscribe_review_reminders
+          resource :subscribe_review_reminders, only: [:show], as: :user_subscribe_review_reminders
+        end
```
sm17p, line 351: This looks odd to me, when I do

```ruby
scope module: :users, as: :users do
          resource :unsubscribe_review_reminders, only: [:show]
          resource :subscribe_review_reminders, only: [:show]

# It generates these route helpers
Prefix Verb   URI Pattern   Controller#Action                                                           
users_unsubscribe_review_reminders GET    /users/unsubscribe_review_reminders(.:format)                                 users/unsubscribe_review_reminders#show
users_subscribe_review_reminders GET    /users/subscribe_review_reminders(.:format)                                   users/subscribe_review_reminders#show
```


@EmCousin shouldn't actual route be `user/unsubscribe_review_reminders` instead of `users/unsubscribe_review_reminders`?

sm17p (commented):
@EmCousin I've added a few review comments. Please have a look 🙏

app/javascript/pages/Users/UnsubscribeReviewReminders/Show.tsx
```diff
@@ -0,0 +1,21 @@
...
+function Show() {
+  return (
+    <Layout heading="You will no longer receive review reminder emails.">
+      If you wish to resubscribe to all review reminder emails, please click{" "}
+      <span>
+        <Link href={Routes.user_subscribe_review_reminders_path()} className="underline">
+          here
+        </Link>
+        .
+      </span>
```
sm17p: Previously, "here" and "." were spaced out as they were inside a flex justified container, and hence grouping them together in a span

### Before shot
Image

app/controllers/users/subscribe_review_reminders_controller.rb
Pradumn27, line 1: we don't need to have 2 separate controllers, where essentially both are performing just a single small update action, add just a `review_remivders_controller.rb` with subscribe and unsubscribe actions

app/javascript/components/EmailAction/Layout.tsx
```diff
@@ -1,32 +1,42 @@
...
+import LoadingSkeleton from "$app/components/LoadingSkeleton";
 import { Card, CardContent } from "$app/components/ui/Card";
+import useRouteLoading from "$app/components/useRouteLoading";
+
+type PageProps = {
+  title: string;
+};
 
 export const Layout = ({ heading, children }: { heading: string; children: React.ReactNode }) => {
+  const { title } = usePage<PageProps>().props;
```
Pradumn27: we aren't passing any title from controller, so there's no need for this

app/javascript/components/EmailAction/Layout.tsx
Pradumn27: This is already done globally for all inertia layouts, not needed here

app/javascript/components/EmailAction/Layout.tsx
```diff
@@ -1,32 +1,42 @@
...
+              <h2 className="grow">{heading}</h2>
+            </header>
+          </CardContent>
+          <CardContent asChild>
+            <p className="justify-start gap-1.5!">{children}</p>
+          </CardContent>
+        </Card>
+        <footer className="p-4 text-center">
+          Powered by&ensp;
+          <a href={Routes.root_url({ host: rootDomain })} className="logo-full" aria-label="Gumroad" />
```
Pradumn27: use inertia Link here now instead

spec/controllers/users/subscribe_review_reminders_controller_spec.rb
Pradumn27, line 1: same as previously make both of them in a single review reminders spec file, not big enough to segregate into 2 different controller specs, same domain specs need to be coupled unless the sub actions are really complex with many edge cases

app/javascript/components/EmailAction/Layout.tsx
Pradumn27: instead of this, just add the `details` prop to `CardContent` it will make it `display: block` - won't need any overrides then

app/javascript/pages/Users/UnsubscribeReviewReminders/Show.tsx
```diff
@@ -0,0 +1,21 @@
+import { Link } from "@inertiajs/react";
+import * as React from "react";
+
+import { Layout } from "$app/components/EmailAction/Layout";
+
+function Show() {
+  return (
+    <Layout heading="You will no longer receive review reminder emails.">
+      If you wish to resubscribe to all review reminder emails, please click{" "}
+      <span>
```
Pradumn27: won't need the span once [this](https://github.com/antiwork/gumroad/pull/3055/changes#r2727146484) is done

Pradumn27 (commented):
Hi @sm17p _a thanks for the PR, added a few comments please check and fix, thanks!

app/javascript/components/EmailAction/Layout.tsx
sm17p: Actually, we did set a default title in `ApplicationController` but we don't need this now after #3019 change

app/javascript/components/EmailAction/Layout.tsx
sm17p: I think we do need this as "Review Reminder" pages render w/o navigation drawer. Also, after #3019, we use `AuthenticationLayout`, where I think it's not possible to use a loader as all the the pages aren't migrated to inertia? 

My recommendation would be to keep this until we have a concensus on how to handle the loding state

app/javascript/pages/Users/UnsubscribeReviewReminders/Show.tsx
sm17p: Makes sense, I wan't sure earlier if we intentionally wanted to render "here." on a new line for mobile devices (There are pages where we do this in some Alerts)

I've added the `details` prop for now and removed the span, the action button now follows browser default text wrapping 

Screenshot 2026-01-27 at 12 46 39 AM

sm17p:
@Pradumn27 _a Thanks! for the review. I've addressed few of the review comments and replied to some. Please have a look again

app/controllers/users/review_reminders_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class Users::ReviewRemindersController < ApplicationController
+  layout "inertia"
+
+  before_action :authenticate_user!
+
+  def subscribe
+    logged_in_user.update!(opted_out_of_review_reminders: false)
+    render inertia: "Users/SubscribeReviewReminders"
```
EmCousin: Template should be `Users/ReviewReminders/Subscribe`

app/controllers/users/review_reminders_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def unsubscribe
+    logged_in_user.update!(opted_out_of_review_reminders: true)
+    render inertia: "Users/UnsubscribeReviewReminders"
```
EmCousin: Template should be `Users/ReviewReminders/Unsubscribe`

app/javascript/components/EmailAction/Layout.tsx
EmCousin: Those are lightweight pages, it's a bit overkill to have a skeleton loader here. You can remove it.

config/routes.rb
EmCousin: We mustn't change the route's path structure since we need those links to still be working in the emails that are already sent, so they way you did it is fine

sm17p:
Thanks @EmCousin 👍

 I've checked the failing specs. These are unrelated and failing because of stripe rate limit

EmCousin (approved):
Thanks @sm17p and @Pradumn27 for the review!

Issue #3030, opened 2026-01-22
Migrate Edit Product page to Inertia


EmCousin:
### What

Migrate Product Edit page as-is to Inertia on `/products/:id/edit_new` route, running in parallel until it's completed and verified.

- [ ] Scaffold Inertia route and page at `/products/:id/edit_new`
- [ ] Migrate layouts
- [ ] Migrate Product tab
- [ ] Migrate Content tab
- [ ] Migrate Share tab
- [ ] Migrate Receipt tab
- [ ] Migrate Post-save customer notification prompt
- [ ] Switch Inertia version to `/products/:id/edit` and clean up old code

### Why
- Part of the ongoing migration to Inertia SPA for better UX and performance
- Blocks future improvements like autosave (#2371)

PR #3064, opened 2026-01-23
Edit product page inertia (v2)

skyfallwastaken:
Issue: #3030

# Description

Moves from React on Rails to Inertia!

## Problem

Previously it used React on Rails.
(Also, gumroad.com would have layout shifts when moving _to_ the Content tab, as well as when moving to the Share tab and the "To appear on Gumroad Discover..." would be shown.)

## Solution

Move to Inertia! The layout shifts have been fixed too.

---

# Before/After

Before: Used React on Rails (https://files.catbox.moe/h6q9r3.mov)
After: Uses Inertia! (https://files.catbox.moe/czaj7d.mov)

GitHub chokes on large file uploads for some reason, so I uploaded them to Catbox.

---

# Checklist

---

# AI Disclosure

I used Amp w/ Opus 4.5 + GPT-5.2 for help, but the code itself has been reviewed and tested by me.

### DIFF
--- a/app/views/links/index.html.erb
+++ b/app/views/links/index.html.erb
@@ -1,2 +1,0 @@
-<%= load_pack("products") %>
-<%= react_component "ProductsDashboardPage", props: @react_products_page_props, prerender: true %>


skyfallwastaken:
@EmCousin videos have been added

app/controllers/links_controller.rb
```diff
@@ -408,38 +429,47 @@ def update
...
   end
 
   def publish
     authorize @product
 
     if @product.user.email.blank?
-      return render json: { success: false, error_message: "<span>To publish a product, we need you to have an email. <a href=\"#{settings_main_url}\">Set an email</a> to continue.</span>" }
+      return redirect_back fallback_location: edit_link_path(@product.unique_permalink),
+                           inertia: { errors: { base: "To publish a product, we need you to have an email." } },
+                           alert: "<span>To publish a product, we need you to have an email. <a href=\"#{settings_main_url}\">Set an email</a> to continue.</span>"
```
skyfallwastaken, line 455: This gives me the ick, but since the JSON response did the same thing I decided to keep it

app/javascript/components/ProductEdit/ShareTab/index.tsx
```diff
@@ -115,42 +114,42 @@ export const ShareTab = () => {
 };
 
 const DiscoverEligibilityPromo = () => {
-  const [show, setShow] = React.useState(false);
-
```
skyfallwastaken, line 119: The layout shift when the Gumroad Discover alert pops in has been fixed!

EmCousin:
@skyfallwastaken please reference your previous PR-attempts in the description

app/controllers/links_controller.rb
```diff
@@ -280,9 +280,19 @@ def edit
     redirect_to bundle_path(@product.external_id) if @product.is_bundle?
 
     @title = @product.name
+    @load_dropbox_dropins = true
+
+    active_tab = case request.path
+                 when %r{/edit/content$} then "content"
+                 when %r{/edit/share$} then "share"
+                 when %r{/edit/receipt$} then "receipt"
+                 else "product"
```
EmCousin, line 289: These should be turned into actual server endpoints rendering their own dedicated Inertia page, with common frontend code factored to a layout. This suggests a controller per tab, each with an `edit` and `update` action, each controller accepting only the parameters it actually needs.

The reason we were having multiple tabs in one single page is that we were forced by React on Rails to define SPA-routes frontend-side. We no longer hhave that constraint with Inertia, so those endpoints can go back to the server where they belong.

app/controllers/links_controller.rb
```diff
@@ -386,7 +396,18 @@ def update
...
       end
-      return render json: { error_message: }, status: :unprocessable_entity
+
+      presenter = ProductPresenter.new(product: @product, pundit_user:)
+      active_tab = case request.referer
+                   when %r{/edit/content} then "content"
+                   when %r{/edit/share} then "share"
+                   when %r{/edit/receipt} then "receipt"
+                   else "product"
+      end
```
EmCousin, line 406: Same comment

app/controllers/links_controller.rb
```diff
@@ -386,7 +396,18 @@ def update
...
+
+      return render inertia: "Products/Edit",
+                    props: presenter.edit_props.merge(active_tab:, errors: { base: error_message }),
+                    status: :unprocessable_entity
```
EmCousin, line 410: Failed form submissions should redirect to the edit page, errors can be carried over via the session.
https://inertia-rails.dev/guide/forms#server-side-responses

app/javascript/pages/Products/Edit.tsx
```diff
@@ -0,0 +1,258 @@
...
+        },
+      });
+    });
+  };
+
+  React.useEffect(() => {
+    if (props.errors?.base) {
+      showAlert(props.errors.base, "error");
+    }
+  }, [props.errors]);
```
EmCousin, line 184: You don't need that, Flash message is set in the controller and is intercepted in the Inertia layout to be rendered automatically.

app/controllers/links_controller.rb
EmCousin: The way we intercept flash messages in the Inertia layout does not allow rendering HTML at this time. You would need to adapt the useFlashMessage hook.

app/controllers/links_controller.rb
```diff
@@ -408,38 +429,47 @@ def update
...
+                           alert: "<span>To publish a product, we need you to have an email. <a href=\"#{settings_main_url}\">Set an email</a> to continue.</span>"
     end
 
     begin
       @product.publish!
     rescue Link::LinkInvalid, ActiveRecord::RecordInvalid
-      return render json: { success: false, error_message: @product.errors.full_messages[0] }
+      return redirect_back fallback_location: edit_link_path(@product.unique_permalink),
+                           inertia: { errors: { base: @product.errors.full_messages[0] } },
+                           alert: @product.errors.full_messages[0]
```
EmCousin, line 463: They way this is implemented in the frontend causes the flash message to be fired twice. This is due to a misunderstanding on how flash messages and InertiaRails' server-side validation work.

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -340,26 +341,21 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
...
-      assertResponseError(error);
-      showAlert(error.message, "error");
-    } finally {
-      setSelectingExistingFiles((state) => (state ? { ...state, isLoading: false } : null));
-    }
+  const fetchLatestExistingFiles = () => {
+    const minLoadingTime = new Promise((resolve) => setTimeout(resolve, 250));
+
+    router.reload({
+      only: ["existing_files"],
```
EmCousin, line 348: This will not reload the existing files only, because you did not adapt the props to enable lazy data evaluation in the backend: https://inertia-rails.dev/guide/partial-reloads#lazy-data-evaluation

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -137,36 +135,60 @@ export const Layout = ({
...
+        onSuccess: () => {
+          updateProduct({ is_published: published });
+          if (targetPath !== rootPath || tab !== "product") {
+            navigateTo(targetPath);
+          }
+        },
+        onFinish: () => setIsPublishing(false),
+      },
+    );
   };
```
EmCousin, line 192: This logic should be decoupled from the layout and moved to the backend.

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -274,40 +296,33 @@ export const Layout = ({
...
+              isSelected={tab === "share"}
+              onClick={(e) => {
+                onTabClick(e, () => {
+                  if (!product.is_published) {
+                    showAlert(
+                      "Not yet! You've got to publish your awesome product before you can share it with your audience and the world.",
+                      "warning",
+                    );
+                    return;
+                  }
```
EmCousin, line 320: This permission logic should now move to the backend

app/javascript/ssr.ts
```diff
@@ -108,7 +108,7 @@ ReactOnRails.register({
   Nav,
   PdfReaderPage,
   Pill,
-  ProductEditPage,
+
```
EmCousin, line 111: Trailing blank line

app/javascript/ssr.ts
```diff
@@ -43,7 +43,7 @@ import Nav from "$app/components/server-components/Nav";
 import PdfReaderPage from "$app/components/server-components/PdfReaderPage";
 import ProductPage from "$app/components/server-components/Product";
 import ProductIframePage from "$app/components/server-components/Product/IframePage";
-import ProductEditPage from "$app/components/server-components/ProductEditPage";
+
```
EmCousin, line 46: Trailing blank line

app/views/layouts/inertia.html.erb
```diff
@@ -19,6 +19,9 @@
     <% end %>
     <%= javascript_include_tag "application" %>
     <%= load_pack("base") unless @hide_layouts %>
+    <% if @load_dropbox_dropins %>
+      <%= render("shared/load_dropbox_dropins_js", async: true) %>
+    <% end %>
```
EmCousin, line 24: Do not use conditional rendering in layouts and ivars. Load that script directly in the Inertia page.

.ruby-version
```diff
@@ -1 +1 @@
-3.4.3
+3.4.7
```
EmCousin, line 1: Please revert this

app/views/layouts/inertia.html.erb
skyfallwastaken: I'm planning on doing something like this, is that fine or should I just use a script tag?

```tsx
function Edit() {
  const props = usePage<Props>().props;
  usePersistentExternalScript(
    `https://www.dropbox.com/static/api/2/dropins.js?app_key=${props.dropbox_picker_api_key}`,
  );
  return <ProductEditPage {...props} />;
}
```

app/javascript/pages/Products/Edit.tsx
skyfallwastaken: thanks!

app/controllers/links_controller.rb
skyfallwastaken: Added `flash[:html]`! Just set that to true and you're on your way.

app/javascript/components/ProductEdit/ContentTab/index.tsx
skyfallwastaken: 'tis been fixed:
```rb
existing_files: -> { existing_files },
```

app/controllers/links_controller.rb
skyfallwastaken: Has been fixed, sorry about that!

Issue #3062, opened 2026-01-23
Migrate Post Page to Inertia


EmCousin:
Endpoints:
- [x] `/p/:slug` (`PostsController#show`)

PR #3074, opened 2026-01-23
Migrate Post Page to Inertia 

TenzDelek:
Issue: #3062

# Description

This PR migrates the following endpoint to inertia:
- [x] /p/:slug (PostsController#show)

## Problem
The  Post Page  is using the older server-rendered Rails views with inline React components (React on Rails). 

## Solution
1. Migration of Post Page to Inertia. 
2. Move formatPostDate utils to ```date.ts``` file

---

# Before

https://github.com/user-attachments/assets/dafc7525-b0c3-47b0-91e1-4389360f062c

# After

https://github.com/user-attachments/assets/04207473-fb4d-41e3-befd-c7acfffe46bc

---

# Test Results

image
image

---

# Checklist

---

# AI Disclosure

    Model: Opus 4.5
    IDE: Cursor
    Usage: Code understanding and Inertia.js migration

### DIFF
--- a/app/javascript/components/Profile/Sections.tsx
+++ b/app/javascript/components/Profile/Sections.tsx
@@ -13,6 +13,7 @@ import {
 import { SearchResults } from "$app/data/search";
 import { CreatorProfile } from "$app/parsers/profile";
 import { CurrencyCode } from "$app/utils/currency";
+import { formatPostDate } from "$app/utils/date";
 
 import { Icon } from "$app/components/Icons";
 import { Product, Props as ProductProps } from "$app/components/Product";
@@ -21,7 +22,6 @@ import { PriceSelection } from "$app/components/Product/ConfigurationSelector";
 import { FollowForm } from "$app/components/Profile/FollowForm";
 import { useRichTextEditor } from "$app/components/RichTextEditor";
 import { CoffeeProduct } from "$app/components/server-components/Profile/CoffeePage";
-import { formatPostDate } from "$app/utils/date";
 import { CardContent } from "$app/components/ui/Card";
 import { useUserAgentInfo } from "$app/components/UserAgent";
 import { Card as WishlistCard, CardGrid as WishlistCardGrid, CardWishlist } from "$app/components/Wishlist/Card";
--- a/app/javascript/components/server-components/GumroadBlog/IndexPage.tsx
+++ b/app/javascript/components/server-components/GumroadBlog/IndexPage.tsx
@@ -3,9 +3,9 @@ import * as React from "react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { createCast } from "ts-safe-cast";
 
+import { formatPostDate } from "$app/utils/date";
 import { register } from "$app/utils/serverComponentUtil";
 
-import { formatPostDate } from "$app/utils/date";
 import { Tabs, Tab } from "$app/components/ui/Tabs";
 
 import placeholderFeatureImage from "../../../../assets/images/blog/post-placeholder.jpg";
--- a/app/javascript/components/server-components/GumroadBlog/PostPage.tsx
+++ b/app/javascript/components/server-components/GumroadBlog/PostPage.tsx
@@ -4,13 +4,13 @@ import * as React from "react";
 import { createCast } from "ts-safe-cast";
 
 import { incrementPostViews } from "$app/data/view_event";
+import { formatPostDate } from "$app/utils/date";
 import { register } from "$app/utils/serverComponentUtil";
 
 import { Button } from "$app/components/Button";
 import { Icon } from "$app/components/Icons";
 import { LoadingSpinner } from "$app/components/LoadingSpinner";
 import { useRichTextEditor } from "$app/components/RichTextEditor";
-import { formatPostDate } from "$app/utils/date";
 import { useUserAgentInfo } from "$app/components/UserAgent";
 import { useRunOnce } from "$app/components/useRunOnce";
 


app/javascript/utils/date.ts
```diff
@@ -1,8 +1,14 @@
+import { parseISO } from "date-fns";
+
 export const formatDate = (
   date: Date,
   options: Intl.DateTimeFormatOptions = { dateStyle: "long", timeStyle: "short" },
 ) => date.toLocaleString([], options);
 
+const postDateFormatOptions: Intl.DateTimeFormatOptions = { month: "long", day: "numeric", year: "numeric" };
+export const formatPostDate = (date: string | null, locale: string): string =>
```
TenzDelek, line 9: previously this util was written in ```PostPage.tsx```. so we have move this to ```date.ts``` util and change the respective import as per it.

app/presenters/post_presenter.rb
```diff
@@ -30,6 +30,7 @@ def post_component_props
       recent_posts:,
       paginated_comments:,
       comments_max_allowed_depth: Comment::MAX_ALLOWED_DEPTH,
+      custom_styles: seller.seller_profile.custom_styles,
```
TenzDelek, line 33: This injected CSS that sets CSS custom properties (variables) based on the seller's profile settings. similar to how we are doing in old  ```show.html.erb``` line 2

app/controllers/posts_controller.rb
```diff
@@ -15,7 +17,6 @@ def show
     @post || fetch_post(false)
 
     @title = "#{@post.name} - #{@post.user.name_or_username}"
-    @hide_layouts = true
     @show_user_favicon = true
     @body_class = "post-page"
     @body_id = "post_page"
```
Pradumn27: check which all of these public `@` instances we actually need now, since most we are passing as local props

app/javascript/pages/Posts/Show.tsx
```diff
@@ -0,0 +1,159 @@
...
+          >
+            <PostCommentsSection paginated_comments={paginated_comments} />
+          </CommentsMetadataProvider>
+        ) : null}
+        {recent_posts.length > 0 ? (
+          <>
+            {recent_posts.map((post) => (
+              <a
+                key={post.slug}
+                href={Routes.custom_domain_view_post_path(post.slug, { purchase_id: post.purchase_id })}
```
Pradumn27, line 135: why is `post.purchase_id` being used now as in the previous implementation the `purchase_id` was used directly?

spec/controllers/posts_controller_spec.rb
```diff
@@ -116,6 +117,10 @@
         installment = create(:published_installment, link: @product, installment_type: "product", shown_on_profile: false)
         get :show, params: { username: @user.username, slug: installment.slug, purchase_id: @purchase.external_id }
         expect(response).to be_successful
+        expect(inertia).to render_component("Posts/Show")
+        expect(inertia.props[:subject]).to eq(installment.subject)
+        expect(inertia.props[:slug]).to eq(installment.slug)
+        expect(inertia.props[:external_id]).to eq(installment.external_id)
```
Pradumn27, line 124: need to verify all the props including the newly added `custom_styles`

app/controllers/posts_controller.rb
Pradumn27:
```suggestion
    post_presenter = PostPresenter.new(
```
a local variable now

app/javascript/pages/Posts/Show.tsx
```diff
@@ -0,0 +1,159 @@
...
+  name: string;
+  slug: string;
+  published_at: string | null;
+  truncated_description: string;
+  purchase_id: string | null;
+};
+
+type PageProps = {
+  subject: string;
+  slug: string;
```
Pradumn27: this prop is unused, why?

app/javascript/pages/Posts/Show.tsx
```diff
@@ -0,0 +1,159 @@
...
+    ariaLabel: "Email message",
+    initialValue: pageLoaded ? message : null,
+    editable: false,
+  });
+
+  const publishedAtFormatted = formatPostDate(published_at, userAgentInfo.locale);
+
+  return (
+    <>
+      <style dangerouslySetInnerHTML={{ __html: custom_styles }} />
```
Pradumn27: this is wrong we need the styles to be placed in the Header as was being done previously, use Inertia's `Head` component, and do this:

```
<Head>                                                                                                     
  <style>{custom_styles}</style>                                                                           
</Head>
```

app/javascript/pages/Posts/Show.tsx
TenzDelek: Looking at the type definition, each post in recent_posts can have its own distinct purchase_id. 
```
type RecentPost = {
  name: string;
  slug: string;
  published_at: string | null;
  truncated_description: string;
  purchase_id: string | null;
};
```
when navigating to a recent post, the correct ```purchase_id```  for that specific post is used, rather than reusing the purchase_id from whatever post the user happens to be viewing currently. 
I believe the previous code seems to be a bug as  it was passing the current page's purchase_id to all recent post links, which would be incorrect.

app/controllers/posts_controller.rb
TenzDelek: ```@subscription```  was not needed. have removed it

app/javascript/pages/Posts/Show.tsx
TenzDelek: I have removed the ```slug```. I was confused about the ```recentPost slug```. Thank you for the notice.

TenzDelek:
Thank you for the review @Pradumn27 _a , I have made the changes. also I have left some notes addressing some of the comments on the PR. Please take a look and let me know if anything else needs adjustment. Thank you

spec/controllers/posts_controller_spec.rb
TenzDelek: done

app/controllers/posts_controller.rb
```diff
@@ -17,7 +19,6 @@ def show
     # Skip fetching post again if it's already fetched in check_if_needs_redirect
     @post || fetch_post(false)
```
Pradumn27, line 20: make this local too

app/controllers/posts_controller.rb
Pradumn27: no, i meant to make them local by removing `@` and also I believe the above `@body_class` and `@body_id`  are also unused

app/javascript/pages/Posts/Show.tsx
Pradumn27: makes sense!

app/controllers/posts_controller.rb
TenzDelek: ```@post```  is shared across multiple methods and callbacks in this controller, so I believe it should remain an instance variable. Please let me know if I’m mistaken.

TenzDelek:
Thank you for the review @Pradumn27 _a, I have made the require changes as requested.

Pradumn27:
@binary-koan _a, please have a look.

app/controllers/posts_controller.rb
EmCousin: `@post` is set a first time in `check_if_needs_redirect`, so we should not need to call `fetch_post(false)` here. Let's keep it and continue relying on `@post` for now

TenzDelek:
Failing test seems unrelated cc @EmCousin

EmCousin (approved):
Thanks @TenzDelek !

Issue #3042, opened 2026-01-23
Migrate Checkout Page to Inertia


EmCousin:
Endpoints:
 - [x] `/checkout`
 - [x] Internal API endpoints

stefanbinoj:
Thanks for considering this, @sm17p 😄 I really appreciate it! I believe we both took different approaches to tackling this issue, and that contributed significantly to solving it in the best possible way. I’d suggest we split the bounty equally between us. What do you say? Does that sound good?

PR #3081, opened 2026-01-23
Migration of checkout page

yashranaway:
Issue: #3042
Fixes: #3042 

# Description

## Problem

The checkout page currently uses React on Rails with ERB templates, which is inconsistent with our ongoing migration to Inertia.js. This creates maintenance overhead and prevents us from leveraging Inertia's benefits like simplified data passing and better SPA-like navigation.

## Solution

Migrate the public checkout page from React on Rails to Inertia while preserving all existing functionality:
- Extract checkout logic into a reusable `CheckoutPageContent` component
- Create new Inertia page at `pages/Checkout/Index.tsx`
- Update controller to use `render inertia:` instead of ERB + react_component
- Delete the old ERB view
- Update tests to use Inertia test helpers

---

# Before/After

## Before 

https://github.com/user-attachments/assets/a12ae8b0-f4a1-4701-8dac-509acd142f6d

## Logged out 

https://github.com/user-attachments/assets/284a68a8-7222-47ed-bf78-3a94906fb385

## Logged in 

https://github.com/user-attachments/assets/35b838a1-86dd-4303-9ea7-0ac219624dda

---

# Test Results
image

---

# Checklist

---

# AI Disclosure

Claude Code (Claude Opus 4.5) was used as a coding assistant for this migration - helping with code extraction, Inertia patterns, and test updates.


app/controllers/checkout_controller.rb
```diff
@@ -1,12 +1,21 @@
 # frozen_string_literal: true
 
 class CheckoutController < ApplicationController
+  include InertiaRendering
+
```
yashranaway, line 5: This concern provides the `inertia_share` functionality that automatically includes shared props like `logged_in_user`, `current_seller`, `design_settings`, and `feature_flags` - matching how other Inertia pages in the codebase work.

app/controllers/checkout_controller.rb
yashranaway, line 18: Removed `@hide_layouts = true` as this was preventing the base pack from loading (which includes inertia.js). The Inertia page component uses `disableLayout = true` to hide the sidebar instead. Props are passed directly to the Inertia component rather than through an instance variable.

app/javascript/components/Checkout/CheckoutPageContent.tsx
```diff
@@ -0,0 +1,776 @@
...
+import { useLoggedInUser } from "$app/components/LoggedInUser";
+import { Modal } from "$app/components/Modal";
+import { AuthorByline } from "$app/components/Product/AuthorByline";
+import { computeOptionPrice, OptionRadioButton, Option } from "$app/components/Product/ConfigurationSelector";
+import { PriceTag } from "$app/components/Product/PriceTag";
+import { showAlert } from "$app/components/server-components/Alert";
+import { ProductCard, ProductCardFigure, ProductCardHeader, ProductCardFooter } from "$app/components/ui/ProductCard";
+import { useAddThirdPartyAnalytics } from "$app/components/useAddThirdPartyAnalytics";
+import { useDebouncedCallback } from "$app/components/useDebouncedCallback";
+import { useOnChange, useOnChangeSync } from "$app/components/useOnChange";
```
yashranaway, line 57: All imports are preserved exactly as they were in the original `CheckoutPage.tsx`. No logic changes - just extracted to a separate file.

app/javascript/components/Checkout/CheckoutPageContent.tsx
```diff
@@ -0,0 +1,776 @@
...
+  state: string | null;
+  address: { street: string | null; city: string | null; zip: string | null } | null;
+  saved_credit_card: SavedCreditCard | null;
+  recaptcha_key: string | null;
+  paypal_client_id: string;
+  cart: CartState | null;
+  max_allowed_cart_products: number;
+  tip_options: number[];
+  default_tip_option: number;
+};
```
yashranaway, line 92: Exported the Props type so the Inertia page can reference it. This ensures type safety between the controller props and the React component.

app/javascript/components/Checkout/CheckoutPageContent.tsx
```diff
@@ -0,0 +1,776 @@
...
+    () =>
+      void loadSurcharges({ ...state, products: getProducts(getCartIfAccepted()) })
+        .then(setSurchargesIfAccepted)
+        .catch((e: unknown) => {
+          assertResponseError(e);
+          showAlert("Sorry, something went wrong. Please try again.", "error");
+          dispatch({ type: "cancel" });
+        }),
+    [currentOffer],
+  );
```
yashranaway, line 256: Preserved the original comment here explaining why surcharges need to be precomputed for Apple Pay.

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    has_published_products: boolean;
+    subdomain: string;
+    is_buyer: boolean;
+    time_zone: {
+      name: string;
+      offset: number;
+    };
+  } | null;
+  flash?: { message: string; status: "success" | "warning" | "danger" } | null;
+};
```
yashranaway, line 26: Extends the checkout props with Inertia shared props (`logged_in_user`, `current_seller`, `flash`). These come from `InertiaRendering` concern's `inertia_share` block.

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <LoggedInUserProvider value={parseLoggedInUser(logged_in_user)}>
+      <CurrentSellerProvider value={parseCurrentSeller(current_seller)}>
+        <Alert initial={null} />
+        <CheckoutPageContent {...checkoutProps} />
+      </CurrentSellerProvider>
+    </LoggedInUserProvider>
+  );
+}
```
yashranaway, line 41: Destructures shared props from Inertia and passes remaining checkout-specific props to `CheckoutPageContent`. Wraps in context providers since `disableLayout` means we don't get the default layout's providers.

app/controllers/checkout_controller.rb
EmCousin: This is already included as part of ApplicationController

app/javascript/components/server-components/CheckoutPage.tsx
```diff
@@ -1,779 +1,19 @@
-import { reverse } from "lodash-es";
```
EmCousin, line 1: This entire file can be deleted since we are no longer using it. The component should be removed from the SSR components lists too.

app/views/checkout/index.html.erb
```diff
@@ -1,3 +0,0 @@
-<%= render("shared/recaptcha_script") %>
```
EmCousin, line 1: Where is this loaded now?

app/views/checkout/index.html.erb
```diff
@@ -1,3 +0,0 @@
...
-<%= load_pack("checkout") %>
```
EmCousin, line 2: The corresponding pack should be deleted since we are not using it anymore

spec/controllers/checkout_controller_spec.rb
```diff
@@ -1,17 +1,20 @@
...
       get :index
 
-      expect(assigns[:hide_layouts]).to eq(true)
       expect(assigns[:on_checkout_page]).to eq(true)
       expect(response).to be_successful
+      expect(inertia.component).to eq("Checkout/Index")
+      expect(inertia.props[:countries]).to be_present
+      expect(inertia.props[:us_states]).to be_present
+      expect(inertia.props[:ca_provinces]).to be_present
+      expect(inertia.props[:paypal_client_id]).to be_present
```
EmCousin, line 17: There are lot more props to check here and it's not enough to test their presence

PR #3096, opened 2026-01-24
feat(products): migrate Edit Product page to Inertia

sahitya-chandra:
Issue: #3030 
fixes: #3030 

# Description
Migrated the Edit Product suite (`/products/:id/edit`) from React on Rails to Inertia.js. This migration moves the entire editing workflow into the unified Inertia architecture, ensuring consistency with the `index` and `new`
 product pages while improving layout stability.

## Problem
The Edit Product page was still using a legacy React on Rails implementation with an ERB template (`edit.html.erb`), which was inconsistent with the codebase's move towards Inertia.js (already used for index and new).
Additionally, identified a bug where `params` could be a plain `Hash` in certain contexts (SSR/proxies), causing crashes on `permit` or 404s on symbolic key access.

## Solution
I migrated the page to Inertia.js by:

1. **Tab-Based Page Splitting:** Created four distinct Inertia pages (`Product`, `Content`, `Receipt`, `Share`) in `app/javascript/pages/Products/Edit/`.
2. **Layout Consolidation:** Refactored `ProductEditLayout.tsx` to wrap its contents in the main Inertia Layout.
3. **Universal Layout Component:** Patched the legacy `app/javascript/components/ProductEdit/Layout.tsx` to be "dual-stack":
- Replaced RRD's `<Link>` with Inertia's `<Link>`.
- Implemented guarded RRD hooks (`useMatches`, `useNavigate`) with runtime fallbacks to Inertia's `usePage` and `router`.
4. **Context Propagation:** Injected `CurrentSellerProvider` and `LoggedInUserProvider` into the consolidated layout structure to ensure all editing features have access to required data.
5. **Global Params Fix:** Implemented a fix in `ApplicationController#params` to guarantee it always returns `ActionController::Parameters`.
6. **Build Reliability:** Patched `transformerLoader.js` to handle deleted files gracefully and resolved multiple TypeScript compilation errors related to unused variables.
7. Updating `LinksController#edit` to use `render inertia:` instead of `react_component`.
---

# Before/After

**Note**: This is an architectural migration. The UI and functionality remain exactly the same.

**Before**: Rendered via ERB template calling react_component.
**After**: Rendered directly via Inertia response.

[Screencast from 2026-01-25 13-45-07.webm](https://github.com/user-attachments/assets/aea6fbee-8bdf-44e5-99b7-1dbc4f729224)

---

# Test Results
I updated the `spec/controllers/links_controller_spec.rb` to assert that the Inertia component `Products/Edit` is rendered with the correct props.

```bash
bundle exec rspec spec/controllers/links_controller_spec.rb -e "GET edit"
```

Screenshot from 2026-01-25 13-48-08

---

# Checklist

---

# AI Disclosure
- used cursor with opus-4.5 to plan the solution and refactoring code


app/controllers/application_controller.rb
```diff
@@ -39,6 +39,11 @@ class ApplicationController < ActionController::Base
 
   add_flash_types :warning
 
+  def params
+    _params = super
+    _params.is_a?(Hash) ? ActionController::Parameters.new(_params) : _params
+  end
+
```
sahitya-chandra: I've overridden the `params` method here to ensure it always returns an `ActionController::Parameters` object. During this migration, we identified an issue where `params` could be initialized as a plain `Hash` (likely due to SSR or middleware proxies), which caused crashes on `.permit` calls. This global fix improves the overall robustness of the app

app/controllers/links_controller.rb
```diff
@@ -277,12 +277,14 @@ def edit
...
-    redirect_to bundle_path(@product.external_id) if @product.is_bundle?
+    return redirect_to bundle_path(@product.external_id) if @product.is_bundle?
 
     @title = @product.name
 
     ai_generated = params[:ai_generated] == "true"
-    @presenter = ProductPresenter.new(product: @product, pundit_user:, ai_generated:)
+    presenter = ProductPresenter.new(product: @product, pundit_user:, ai_generated:)
+
+    render inertia: "Products/Edit", props: presenter.edit_props
```
sahitya-chandra: Migrated the `edit` action to use `render inertia:` instead of the legacy `react_component` in an ERB template. I'm using the `edit_props` from `ProductPresenter` to maintain consistency with the data previously passed to React on Rails.

app/javascript/components/server-components/ProductEditPage.tsx
```diff
@@ -142,7 +142,7 @@ const findUpdatedContent = (product: Product, lastSavedProduct: Product) => {
   };
 };
 
-const ProductEditPage = (props: Props) => {
+export const ProductEditPage = (props: Props) => {
```
sahitya-chandra: Converted `ProductEditPage` to a named export and also exported the `Props` type. This was required because the legacy default export (wrapped in the `register` HOC) wasn't compatible with TypeScript when imported into the new Inertia `Edit.tsx` wrapper

spec/controllers/links_controller_spec.rb
```diff
@@ -202,13 +202,12 @@ def e404_test(action)
...
-      it "assigns the correct instance variables" do
+      it "renders the Products/Edit component with correct props" do
         get :edit, params: { id: product.unique_permalink }
         expect(response).to be_successful
 
-        product_presenter = assigns(:presenter)
-        expect(product_presenter.product).to eq(product)
-        expect(product_presenter.pundit_user).to eq(controller.pundit_user)
+        expect(inertia).to render_component("Products/Edit")
+        expect(inertia.props).to include(:product, :id, :unique_permalink)
```
sahitya-chandra, line 210: Updated the `edit` specs to use `inertia` RSpec helpers. Instead of asserting on `@presenter` instance variables (which are no longer used by the template), we now directly verify that the `Products/Edit` component is rendered with the expected props.

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -143,10 +144,37 @@ export const Layout = ({
...
+    if (inertiaUrl.includes("/edit/content")) return "content";
+    if (inertiaUrl.includes("/edit/receipt")) return "receipt";
+    if (inertiaUrl.includes("/edit/share")) return "share";
+    return "product";
+  }, [match, inertiaUrl]);
+
+  let rrdNavigate: ReturnType<typeof useNavigate> | undefined;
+  try {
+    rrdNavigate = useNavigate();
+  } catch {
```
sahitya-chandra, line 167: This component is now 'dual-stack'. It's shared between the remaining legacy React-on-Rails views and the new Inertia pages. I've guarded the `react-router-dom` hooks so they don't crash when rendered via Inertia, falling back to Inertia's native `router` and `usePage` mapping logic for tab detection

app/javascript/pages/Products/Edit/Content.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <Layout
+      headerActions={
+        <ContentTabHeaderActions
+          selectedVariantId={selectedVariantId}
+          setSelectedVariantId={setSelectedVariantId}
+        />
+      }
+    >
+      <ContentTab selectedVariantId={selectedVariantId} />
+    </Layout>
```
sahitya-chandra, line 19: Inertia pages now manage the layout structure at the top level. For the Content tab, I've extracted the `HeaderActions` (variant selector) so it can be rendered into the layout's header slot while maintaining shared state with the editor in the body.

app/javascript/pages/Products/Edit/Product.tsx
sahitya-chandra, line 1: same as content tab

app/javascript/pages/Products/Edit/Receipt.tsx
sahitya-chandra, line 1: same as content tab

app/javascript/pages/Products/Edit/Share.tsx
sahitya-chandra, line 1: same as content tab

app/javascript/components/ProductEdit/ProductTab/index.tsx
```diff
@@ -65,9 +64,9 @@ export const ProductTab = () => {
   const [thumbnail, setThumbnail] = React.useState(initialThumbnail);
   const [showAiNotification, setShowAiNotification] = React.useState(aiGenerated);
 
-  const { isUploading, setImagesUploading } = useImageUpload();
+  const { setImagesUploading } = useImageUpload();
 
-  const [showRefundPolicyPreview, setShowRefundPolicyPreview] = React.useState(false);
+  const [, setShowRefundPolicyPreview] = React.useState(false);
```
sahitya-chandra, line 69: Cleaned up several unused variables that were originally part of the ProductEditContext. This was necessary to resolve TypeScript build errors that are strictly enforced in our Webpack CI pipeline.

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -1168,3 +1092,114 @@ export const ContentTab = () => {
     </PostsProvider>
   );
 };
+
+export const ContentTabHeaderActions = ({
+  selectedVariantId,
+  setSelectedVariantId,
+}: {
```
sahitya-chandra, line 1099: Extracted the header-specific logic (variant selection) into its own component. This allows us to pass it into the layout's `headerActions` prop from the parent Inertia page, providing a better user experience where the variant toggle stays pinned to the header.

sahitya-chandra:
@EmCousin  sir, can you review these changes?

app/controllers/application_controller.rb
sahitya-chandra: this is now reverted

app/controllers/links_controller.rb
```diff
@@ -277,12 +277,17 @@ def edit
...
     @title = @product.name
 
     ai_generated = params[:ai_generated] == "true"
-    @presenter = ProductPresenter.new(product: @product, pundit_user:, ai_generated:)
+    presenter = ProductPresenter.new(product: @product, pundit_user:, ai_generated:)
+
+    tab = params[:tab] || "product"
+    component = "Products/Edit/#{tab.capitalize}"
+
+    render inertia: component, props: presenter.edit_props
```
EmCousin, line 290: Decoupling the rendering per view is in the right direction. We just need that decoupling to be complete: have one controller per tab, each with an edit and update action. Each controller should expose only the props it needs, and accepts the params it needs to update the product.

The reason we could not do that before is we were constrained by ReactOnRails to manage everything in one view with client-side defined routes. We no longer have this constraint with Inertia, so we can go back to have those routes defined server side.

app/controllers/links_controller.rb
```diff
@@ -386,7 +391,12 @@ def update
       else
         error_message = @product.errors.full_messages.first || e.message
       end
-      return render json: { error_message: }, status: :unprocessable_entity
+
+      if request.inertia?
+        return redirect_back fallback_location: edit_link_path(@product.unique_permalink), inertia: { errors: { base: error_message } }
+      else
+        return render json: { error_message: }, status: :unprocessable_entity
+      end
```
EmCousin, line 399: Do we still need to render JSON?

app/javascript/layouts/ProductEditLayout.tsx
```diff
@@ -0,0 +1,228 @@
...
+      availabilities: productToSave.availabilities.map(({ newlyAdded, ...availability }) =>
+        newlyAdded ? { ...availability, id: null } : availability,
+      ),
+      installment_plan: productToSave.allow_installment_plan ? productToSave.installment_plan : null,
+    }));
+
+    form.post(Routes.link_path(uniquePermalink), {
+      preserveScroll: true,
+      onSuccess: () => {
+        showAlert("Changes saved!", "success");
```
EmCousin, line 131: This should be set server side

app/javascript/layouts/ProductEditLayout.tsx
```diff
@@ -0,0 +1,228 @@
...
+        lastSavedProductRef.current = structuredClone(form.data.product);
+      },
+      onError: (errors) => {
+        showAlert(Object.values(errors)[0] || "Failed to save product", "error");
```
EmCousin, line 135: This should be set server side

app/javascript/components/ProductEdit/Layout.tsx
EmCousin: You shouldn't have to do this. In what cases exactly are we rendering this Layout via a react-route powered component?

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -203,7 +231,7 @@ export const Layout = ({
     </WithTooltip>
   );
 
-  const onTabClick = (e: React.MouseEvent<HTMLAnchorElement>, callback?: () => void) => {
+  const onTabClick = (e: React.MouseEvent<any>, callback?: () => void) => {
```
EmCousin, line 234: Do not use any

PR #3115, opened 2026-01-25
Improve performance of AudienceExportService

Hanaffi:
Issue: https://github.com/antiwork/gumroad/issues/2092
Closes: https://github.com/antiwork/gumroad/issues/2092

# Description

## Problem

Audience CSV exports were intermittently timing out in Sidekiq for large sellers.

The export query filtered audience members using an `OR` condition across multiple boolean columns:

```sql
WHERE seller_id = ?
  AND (follower = true OR customer = true OR affiliate = true)
ORDER BY min_created_at
```

Despite having several indexes, this query was **not able to efficiently use them**, leading to long-running scans in production.

---

## Root cause (confirmed by EXPLAIN ANALYZE)

`EXPLAIN` and `EXPLAIN ANALYZE` showed that MySQL was **not using any index on the boolean flags**.

Instead, MySQL consistently chose this index:

```
idx_audience_on_seller_and_minmax_created_at (seller_id, min_created_at, max_created_at)
```

Execution plan (simplified):

```
Index lookup on (seller_id)
→ scan rows in min_created_at order
→ apply (follower OR customer OR affiliate) as a filter
```

Key observations from `EXPLAIN ANALYZE`:

- Boolean conditions (`follower/customer/affiliate`) were applied **after** the index lookup
- No boolean columns participated in index access
- Performance depended entirely on data distribution
- If matching rows were sparse, MySQL scanned a large portion of the seller’s audience
- This caused queries to run for seconds or minutes, triggering Sidekiq timeouts

This behavior is a known MySQL limitation with `OR` conditions across low-cardinality (boolean) columns.

### 1. Rewrite the query to use `UNION` instead of `OR`

The query is now built as multiple independent queries:

```sql
WHERE seller_id = ? AND follower = true
UNION
WHERE seller_id = ? AND customer = true
UNION
WHERE seller_id = ? AND affiliate = true
```

Each branch can be optimized independently by MySQL.

---

### 2. Add narrow, order-aware indexes per flag

Added indexes:

```sql
(seller_id, follower,  min_created_at)
(seller_id, customer,  min_created_at)
(seller_id, affiliate, min_created_at)
```

These indexes support both:
- Filtering (`seller_id + flag`)
- Ordering (`min_created_at`)

---

## Evidence this fixes the issue

`EXPLAIN ANALYZE` for the new query shows:

```
Index lookup using idx_audience_follower_created_at_flag (seller_id=?, follower=true)
Index lookup using idx_audience_customer_created_at      (seller_id=?, customer=true)
Index lookup using idx_audience_affiliate_created_at     (seller_id=?, affiliate=true)
→ Union materialize
→ Sort by min_created_at
```

Facts confirmed by the plan:

- Each UNION branch uses a **dedicated index**
- No boolean filtering happens outside the index
- No full seller scans
- Sorting happens only on the **already-filtered result set**
- Query runtime scales with the number of matching rows, not total audience size

---

## Why this prevents Sidekiq timeouts

Before:
- Query scanned potentially **all audience rows for a seller**
- Runtime was data-dependent and unpredictable
- Sidekiq jobs blocked on slow SQL

After:
- Each query scans **only matching rows**
- No OR-based filtering
- Predictable and bounded query time
- Sidekiq jobs complete consistently even for large sellers

# Test Results

image

image

---

# Checklist

- [x] I have added/updated tests for my changes **_(No new behavior, existing tests pass)_**

---

# AI Disclosure

Used Cursor (Claude Sonnet 4.5) to reason about index selection and query planner behavior

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -164,8 +164,11 @@ ActiveRecord::Schema[7.1].define(version: 2026_11_19_011937) do
     t.datetime "follower_created_at", precision: nil
     t.datetime "min_affiliate_created_at", precision: nil
     t.datetime "max_affiliate_created_at", precision: nil
+    t.index ["seller_id", "affiliate", "min_created_at"], name: "idx_audience_affiliate_created_at"
     t.index ["seller_id", "customer", "follower", "affiliate"], name: "idx_audience_on_seller_and_types"
+    t.index ["seller_id", "customer", "min_created_at"], name: "idx_audience_customer_created_at"
     t.index ["seller_id", "email"], name: "index_audience_members_on_seller_id_and_email", unique: true
+    t.index ["seller_id", "follower", "min_created_at"], name: "idx_audience_follower_created_at_flag"
     t.index ["seller_id", "follower_created_at"], name: "idx_audience_on_seller_and_follower_created_at"
     t.index ["seller_id", "min_affiliate_created_at", "max_affiliate_created_at"], name: "idx_audience_on_seller_and_minmax_affiliate_created_at"
     t.index ["seller_id", "min_created_at", "max_created_at"], name: "idx_audience_on_seller_and_minmax_created_at"
--- a/app/services/exports/audience_export_service.rb
+++ b/app/services/exports/audience_export_service.rb
@@ -18,16 +18,9 @@ class Exports::AudienceExportService
     @tempfile = Tempfile.new(["Subscribers", ".csv"], encoding: "UTF-8")
 
     CsvSafe.open(@tempfile, "wb", headers: FIELDS, write_headers: true) do |csv|
-      query = @user.audience_members.select(:id, :email, :min_created_at)
+      query = build_query
 
-      conditions = []
-      conditions << "follower = true" if @options[:followers]
-      conditions << "customer = true" if @options[:customers]
-      conditions << "affiliate = true" if @options[:affiliates]
-
-      query = query.where(conditions.join(" OR "))
-
-      query.order(:min_created_at).find_each do |member|
+      query.find_each do |member|
         csv << [member.email, member.min_created_at]
       end
     end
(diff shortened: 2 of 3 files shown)


db/schema.rb
```diff
@@ -164,8 +164,11 @@
     t.datetime "follower_created_at", precision: nil
     t.datetime "min_affiliate_created_at", precision: nil
     t.datetime "max_affiliate_created_at", precision: nil
+    t.index ["seller_id", "affiliate", "min_created_at"], name: "idx_audience_affiliate_created_at"
     t.index ["seller_id", "customer", "follower", "affiliate"], name: "idx_audience_on_seller_and_types"
+    t.index ["seller_id", "customer", "min_created_at"], name: "idx_audience_customer_created_at"
     t.index ["seller_id", "email"], name: "index_audience_members_on_seller_id_and_email", unique: true
+    t.index ["seller_id", "follower", "min_created_at"], name: "idx_audience_follower_created_at_flag"
```
Hanaffi, line 171: do we want to remove `idx_audience_on_seller_and_types`?

From the EXPLAIN / EXPLAIN ANALYZE results:

The wide index (seller_id, customer, follower, affiliate) was not used for the original OR-based query
Keeping the wide index may encourage suboptimal plans in other queries and adds write/storage overhead

Happy to keep it for now if it’s still needed elsewhere, but wanted to check whether we should drop it once this change is deployed.

Hanaffi:
@EmCousin Could you please check whether this approach would be more effective than a larger change, such as using chunked jobs?
I noticed other PRs are leaning toward chunked jobs, so I’d appreciate your take on whether this direction makes sense.

gianfrancopiana:
The issue has been updated with the confirmed root cause – the problem is simpler than the OR conditions. Closing in favor of the approach outlined there.

Issue #3121, opened 2026-01-26
"App store" and "Playstore" button label on open in app pop up not visible on light mode


nyomanjyotisa:
Regression of https://github.com/antiwork/gumroad/issues/2324 cc @Devwill2 

Light mode:
<img width="649" height="467" alt="Image" src="https://github.com/user-attachments/assets/90a0d209-c8db-47d7-8567-a8e09867ec7b" />

Dark mode:
<img width="572" height="422" alt="Image" src="https://github.com/user-attachments/assets/c666a5ad-021a-4655-80f6-713866fbc6c0" />

Before:
Light mode
<img width="391" height="265" alt="Image" src="https://github.com/user-attachments/assets/cb9e97d6-ab7f-447c-bb59-c454869cea6c" />

Dark mode:
<img width="394" height="268" alt="Image" src="https://github.com/user-attachments/assets/5b8993d8-2676-43d3-b0be-7607fe827b9f" />

PR #3128, opened 2026-01-26
fix(download): Fix button label visibility in Open in App popup on light mode

4arjun:
Issue: #3121 

## Summary
Changed Button components to pass `color` prop directly instead of applying `buttonVariants` separately to child elements

## Root cause
PR #2324 implementation incorrectly used `<Button asChild>` without a color prop while separately applying `buttonVariants({ color: "apple" })` to the child `<a>` element.
When `asChild` is true, Radix's `Slot` component merges the parent's classes with the child's classes. The Button's base styles include `text-current`, which was overriding `text-white` from the color variant due to CSS specificity. In light mode, `currentColor` inherits the page's dark text color, making white text on dark buttons invisible.

## Before
Screenshot 2026-01-26 at 12 00 22 PM

Screenshot 2026-01-26 at 12 00 36 PM

## After
Screenshot 2026-01-26 at 1 26 56 PM

Screenshot 2026-01-26 at 1 24 04 PM

# Checklist

---

# AI Disclosure

No AI used

### DIFF
--- a/app/javascript/components/Collaborators/Form.tsx
+++ b/app/javascript/components/Collaborators/Form.tsx
@@ -356,13 +356,11 @@ const CollaboratorForm = ({
               onChange={(evt) => {
                 const enabled = evt.target.checked;
                 setShowUnpublishedOrIneligibleProducts(enabled);
-                if (form.data.apply_to_all_products) {
-                  form.data.products.forEach((item, index) => {
-                    if (!item.has_another_collaborator && enabled && !item.published) {
-                      form.setData(`products.${index}.enabled`, enabled);
-                    }
-                  });
-                }
+                form.data.products.forEach((item, index) => {
+                  if (!item.has_another_collaborator && !item.published) {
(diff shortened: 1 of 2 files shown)


4arjun:
@nyomanjyotisa Could you please take a look at this?

app/javascript/components/Button.tsx
```diff
@@ -50,8 +50,8 @@ export const buttonVariants = cva(
         stripe: "bg-[#625bf6] text-white border-[#625bf6]",
         facebook: "bg-[#4267b2] text-white border-[#4267b2]",
         twitter: "bg-black text-white border-black",
-        apple: "bg-black text-white border-black",
-        android: "bg-[#142f40] text-white border-[#142f40]",
+        apple: "bg-black text-white",
+        android: "bg-[#142f40] text-white",
```
4arjun, line 54: Removed the `border-black` and  `border-[#142f40]` from the android and apple color which was making the border invisible (matching the background color). Now it will use the default border

Issue #3142, opened 2026-01-26
Migrate URL Redirect Download page to Inertia


EmCousin:
Endpoints:
- [x] GET `/d/:id` -> url_redirects#download_page
- [x] GET `/audio_durations/:id` -> url_redirects#audio_durations (JSON endpoint, move to an optional prop and use Inertia's polling mechanism)
- [x] GET `/latest_media_locations/:id` -> url_redirects#latest_media_locations (JSON endpoint, move to an optional prop and use Inertia's polling mechanism)

PR #3157, opened 2026-01-26
Migrate URL Redirect Download page to Inertia

bruhhgnik:
Addresses #3142

Migrate the `url_redirects#download_page` action to render Inertia pages instead of using
`react_component` helper.

## Problem

The download page (`url_redirects#download_page`) uses the legacy `react_component` helper
for rendering, which is inconsistent with the newer Inertia.js pattern being adopted across
the codebase.

## Solution

Migrated the `url_redirects#download_page` action to render Inertia pages:

### Controller changes:

- Added `include InertiaRendering` to the controller
- Added `layout "inertia", only: [:download_page]`
- Removed `download_page` from `hide_layouts` before_action
- Updated `download_page` action to use `render inertia:` with props
- Added `smart_app_banner` and `dropbox_api_key` to props (previously handled by ERB template)
- **Migrated `audio_durations` and `latest_media_locations` to Inertia optional props** - these are now loaded via `InertiaRails.optional` which means they're only evaluated when specifically requested through partial reloads

### Frontend changes:

- Created new Inertia page component at `pages/UrlRedirects/DownloadPage.tsx`
- Component handles smart app banner meta tag injection via `useEffect`
- Component handles Dropbox dropins script loading via `useEffect`
- Added `disableLayout = true` for standalone rendering (page has its own layout)
- Wrapped content with `LoggedInUserProvider` since we bypass the default Inertia layout
- Exported `WithContent` component and `WithContentProps` type from server-components for
  direct import by Inertia pages (following the same pattern as `Profile` component)
- **Replaced manual fetch polling with Inertia's `router.reload()`** - instead of using custom `request()` calls with `setInterval`, we now use `router.reload({ only: ["audio_durations"] })` which is cleaner and more idiomatic
- Removed `DownloadPageWithContent` from `ssr.ts` ReactOnRails registration (no longer needed)

### Code notes:

**Why `InertiaRails.optional`?**
```ruby
props[:audio_durations] = InertiaRails.optional { audio_durations_data }
```
This makes the prop lazy - it only gets evaluated when the frontend does a partial reload with `only: ["audio_durations"]`. On initial page load, this prop is not included, which saves unnecessary DB queries.

**Why `LoggedInUserProvider` wrapper?**
```tsx
<LoggedInUserProvider value={parseLoggedInUser(logged_in_user)}>
  <WithContent {...props} />
</LoggedInUserProvider>
```
Since `disableLayout = true` skips the default Inertia layout (which normally provides this context), we need to wrap manually. The `WithContent` component uses `useLoggedInUser()` internally.

**Why keep the interval-based polling?**
The audio durations polling is for files that are still being transcoded - their duration isn't available immediately. The `latest_media_locations` polling tracks reading/viewing progress for the reader. Both need periodic updates, so polling makes sense here rather than one-time loads.

### Removed files:

- `app/views/url_redirects/download_page.html.erb` - replaced by Inertia page
- `app/javascript/packs/url_redirect_download.js` - no longer needed

### Tests:

- Added `require "inertia_rails/rspec"` to spec file
- Added `inertia: true` to the `download_page` describe block
- Updated assertions to use `inertia.component` and `inertia.props` helpers
- Verify correct Inertia component (`UrlRedirects/DownloadPage`) is rendered
- Verify props include `smart_app_banner` and `dropbox_api_key`

## Screenshots
**before**
[Screencast from 2026-01-31 00-16-22.webm](https://github.com/user-attachments/assets/65e8af42-e79a-4419-90fb-b1576667b61c)

**after**
[Screencast from 2026-01-26 22-49-02.webm](https://github.com/user-attachments/assets/c8d95209-ebe3-4575-9a6c-bd974b6911f7)

## Checklist

 - [x] I have read the [CONTRIBUTING.md](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)                                        
 - [x] I have watched https://www.youtube.com/@anti-work                                                                                       
 - [x] I have added screenshots (before/after, light/dark mode, mobile/desktop)    

---
AI Disclosure                                                                                                                                 
                                                                                                                                                
  Model: Claude (Anthropic)                                                                                                                     
                                                                                                                                                
  Used for:                                                                                                                                     
  - Reviewing TypeScript type definitions for prop interfaces                                                                                   
  - Suggesting test case structure for Inertia component assertions                                                                             
  - Rubber-ducking the props structure when debugging a serialization issue                                                                     
                                                                                                                                                
  What I did myself:                                                                                                                            
  - Analyzed the existing download_page.html.erb template and DownloadPage server component to understand the current implementation            
  - Designed and implemented the new pages/UrlRedirects/DownloadPage.tsx Inertia page component                                                 
  - Wrote the controller changes including download_page_inertia_props helper method                                                            
  - Structured the props to match the existing component's expected interface                                                                   
  - Manually tested all download page scenarios (logged in, logged out, different file types)                                                   
  - Wrote and verified all test cases for the Inertia migration                                                                                 
  - Performed self-review and added inline comments explaining implementation decisions


Pradumn27:
@bruhhgnik, the PR does not follow the [contributing guidelines](https://github.com/antiwork/.github/blob/main/CONTRIBUTING.md). Before/after screenshots in the description, test coverage screenshot, and self-review are missing. Please only open PR for review once the PR is fully ready and follows the guidelines, thanks!

@EmCousin, please close the PR, thanks!

PR #3158, opened 2026-01-26
Fix Stripe Connect users unable to edit payout settings

gianfrancopiana:
Fixes antiwork/gumroad-private#90

## Root Cause

Two issues prevented Stripe Connect users from updating payout settings (frequency, threshold, pause):

1. **Frontend validation** - The payment settings form validated compliance info fields (phone number, address, etc.) unconditionally for all users, even those with Stripe Connect accounts. Stripe Connect users manage their compliance information directly through Stripe's onboarding flow, not through Gumroad's form.

2. **Backend guard** - The controller checked if the user had a Stripe account connected and returned an error "You cannot change your payout method to bank account because you have a stripe account connected" whenever a Stripe Connect user submitted the form, even when they weren't changing their payout method - just updating settings like frequency.

### Frontend (`app/javascript/pages/Settings/Payments/Show.tsx`)
Skip compliance info validation when the user has a Stripe Connect account connected (`selectedPayoutMethod === "stripe"`).

### Backend (`app/controllers/settings/payments_controller.rb`)
Only check the "cannot change payout method" restriction when the user is actually attempting to change their payout method (i.e., when `payment_address`, `card`, or `bank_account` params are present), not when they're just updating payout settings.

## Test Plan

- [x] Stripe Connect user can change payout frequency (Weekly -> Monthly)
- [x] Stripe Connect user can change payout threshold
- [x] Stripe Connect user can pause/unpause payouts
- [x] Stripe Connect user still cannot change to a different payout method (bank, card, PayPal) - protected by existing guard
- [x] Non-Stripe Connect users still get compliance validation (bank, card, PayPal methods)
- [x] Existing payment settings tests pass (97 controller tests, Stripe Connect related system tests)

---

## AI Assistance Notice

### DIFF
--- a/spec/requests/settings/payments_spec.rb
+++ b/spec/requests/settings/payments_spec.rb
@@ -315,31 +315,35 @@ describe("Payments Settings Scenario", type: :system, js: true) do
       expect(find_button("Disconnect Stripe account", disabled: true)[:disabled]).to eq "true"
     end
 
-    it "renders Stripe Connect section correctly for connected users without phone number" do
-      # Clear phone from @user's compliance info to simulate the reported bug scenario
-      # where Stripe Connect users don't have phone numbers stored in Gumroad
-      @user.alive_user_compliance_info.dup_and_save! do |new_compliance_info|
-        new_compliance_info.phone = nil
-      end
+    it "allows Stripe Connect users to update payout schedule without compliance validation" do
+      # Create a fresh user with Stripe Connect but no phone number to simulate the reported bug
+      # where Stripe Connect users manage compliance info through Stripe, not Gumroad
+      creator = create(:user, payment_address: nil)
+      create(:user_compliance_info, user: creator, phone: nil)
+      create(:merchant_account_stripe_connect, user: creator)
+      creator.check_merchant_account_is_linked = true
+      creator.save!
 
-      create(:merchant_account_stripe_connect, user: @user)
-      @user.check_merchant_account_is_linked = true
-      @user.save!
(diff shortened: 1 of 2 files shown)


spec/requests/settings/payments_spec.rb
```diff
@@ -315,6 +315,33 @@
...
+
+      # Verify Stripe Connect section is shown (not the compliance form with phone field)
+      expect(page).to have_button("Disconnect Stripe account")
+      expect(page).not_to have_field("Phone number")
+
+      # Verify payout settings are available
+      expect(page).to have_select("Schedule", selected: "Weekly")
+
+      # The Update settings button should be enabled for Stripe Connect users
+      expect(find_button("Update settings")).not_to be_disabled
```
hchhabra: The change looks good to me, @gianfrancopiana _a, but the spec doesn't seem to cover the made change i.e. "Update settings"  button should not validate the compliance fields and successfully update the payout schedule in case of connected Stripe account. Can we update the spec to validate that?

gianfrancopiana: Thanks!

Updated the test to verify the full flow - it now clicks "Update settings" and confirms the payout frequency changes from Weekly to Monthly.

During testing, I also discovered a backend issue: the controller was blocking Stripe Connect users from submitting the form entirely because it checked `has_stripe_account_connected?` unconditionally, returning "You cannot change your payout method to bank account..." even when users weren't changing their payout method.

Added a fix to only run that check when the user is actually changing their payout method (i.e., when `payment_address`, `card`, or `bank_account` params are present).

PR #3160, opened 2026-01-26
Migration of checkout page

yashranaway:
Issue: #3042
Fixes: #3042

Previous PR: #3081

# Description

## Problem

The checkout page currently uses React on Rails with ERB templates, which is inconsistent with our ongoing migration to Inertia.js. This creates maintenance overhead and prevents us from leveraging Inertia's benefits like simplified data passing and better SPA-like navigation.

## Solution

Migrate the public checkout page from React on Rails to Inertia while preserving all existing functionality:

- Move checkout logic from `server-components/CheckoutPage.tsx` to `components/Checkout/CheckoutPageContent.tsx`
- Create new Inertia page at `pages/Checkout/Index.tsx` with `disableLayout = true`
- Update controller to use `render inertia:` instead of ERB + react_component
- Delete the old ERB view, server component, SSR registration, and checkout pack
- Update tests to use Inertia test helpers with actual prop value assertions

---

# Before/After

## Before

https://github.com/user-attachments/assets/a12ae8b0-f4a1-4701-8dac-509acd142f6d

## Logged out

https://github.com/user-attachments/assets/284a68a8-7222-47ed-bf78-3a94906fb385

## Logged in

https://github.com/user-attachments/assets/35b838a1-86dd-4303-9ea7-0ac219624dda

API

https://github.com/user-attachments/assets/67a87979-3d18-4f1c-80e6-be6a15803ef7

---

# Test Results
image

---

# Checklist

---

# AI Disclosure

Claude Code (Claude Opus 4.5) was used as a coding assistant for this migration - helping with code extraction, Inertia patterns, and test updates.


app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,19 @@
 # frozen_string_literal: true
 
 class CheckoutController < ApplicationController
-  before_action :process_cart_id_param
+  before_action :process_cart_id_param, only: %i[index]
+  layout "inertia", only: [:index]
```
yashranaway: Added `only: %i[index]` to scope the before_action and set `layout "inertia"` for the index action. The `InertiaRendering` concern is already included via `ApplicationController`.

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,19 @@
...
-    @checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+    @on_checkout_page = true
+    @title = "Checkout"
+
+    checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+
+    render inertia: "Checkout/Index", props: checkout_presenter.checkout_props(
+      params:,
+      browser_guid: cookies[:_gumroad_guid]
+    )
```
yashranaway: Converted from ERB rendering to Inertia. Removed `@hide_layouts = true` (replaced by `disableLayout = true` on the React component) and `@checkout_presenter` instance variable. Props are now passed directly to the Inertia component via `CheckoutPresenter#checkout_props`

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    has_published_products: boolean;
+    subdomain: string;
+    is_buyer: boolean;
+    time_zone: {
+      name: string;
+      offset: number;
+    };
+  } | null;
+  flash?: { message: string; status: "success" | "warning" | "danger" } | null;
+};
```
yashranaway: Defined `PageProps` type extending `CheckoutProps` with Inertia shared props (`logged_in_user`, `current_seller`, `flash`). These come from the `InertiaRendering` concern's `inertia_share` block.

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <LoggedInUserProvider value={parseLoggedInUser(logged_in_user)}>
+      <CurrentSellerProvider value={parseCurrentSeller(current_seller)}>
+        <Alert initial={null} />
+        <CheckoutPageContent {...checkoutProps} />
+      </CurrentSellerProvider>
+    </LoggedInUserProvider>
+  );
+}
```
yashranaway: Inertia page component that wraps `CheckoutPageContent` with context providers. Since `disableLayout = true` bypasses the default layout, we manually provide `LoggedInUserProvider` and `CurrentSellerProvider`.

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+CheckoutPage.disableLayout = true;
```
yashranaway: disableLayout = true` prevents the default Inertia layout (with nav sidebar) from rendering, matching the original checkout page behavior.

app/javascript/components/Checkout/CheckoutPageContent.tsx
yashranaway, line 1: Moved checkout logic from `server-components/CheckoutPage.tsx` to this new location. Exported `Props`, `Result`, `CrossSellModal`, and `UpsellModal` types/components so they can be imported by other files. No functional changes to the checkout logic itself.

yashranaway:
@EmCousin Addressed all the review comments from #3081, could you please take a look again?

ershad:
@claude review the PR.

app/controllers/checkout_controller.rb
Pradumn27:
```suggestion
  before_action :process_cart_id_param, only: %i[index]
  
  layout "inertia"
```

there Is only one action in this controller, so we can simple do this

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,19 @@
...
 class CheckoutController < ApplicationController
-  before_action :process_cart_id_param
+  before_action :process_cart_id_param, only: %i[index]
+  layout "inertia", only: [:index]
 
   def index
-    @hide_layouts = true
-    @checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+    @on_checkout_page = true
+    @title = "Checkout"
```
Pradumn27: why add them? don't see them present in the old controller

app/javascript/pages/Checkout/Index.tsx
Pradumn27, line 1: inline the CheckoutPage component here directly and use the `.loggedInUserLayout` to get the logged in user provider wrappers, won't need this created wrapper then

app/views/checkout/index.html.erb
```diff
@@ -1,3 +0,0 @@
-<%= render("shared/recaptcha_script") %>
```
Pradumn27, line 1: where is this loaded now?

spec/controllers/checkout_controller_spec.rb
```diff
@@ -36,8 +43,6 @@
           get :index
 
           expect(response).to be_successful
-          html = Nokogiri::HTML.parse(response.body)
-          expect(html.xpath("//meta[@property='gr:logged_in_user:id']/@content").text).to eq(user.external_id)
```
Pradumn27, line 40: why is this removed?

spec/controllers/checkout_controller_spec.rb
```diff
@@ -86,7 +91,7 @@
           end
         end
 
-        context "when the cart matching the `cart_id` query param has the `browser_guid` same as the current `_gumroad_guid` cookie value"  do
```
Pradumn27, line 89: seems unnecessry space, revert

yashranaway:
Hey @Pradumn27 _a , updated as per review:                                                                                                                                                                           
                                                                                                                                                                                                                   
  Migrated the internal cart API following the same pattern as other Inertia migrations (collaborators, etc.)                                                                                                      
                                                                                                                                                                                                                   
  - `Api::Internal::CartsController` → `Checkout::CartsController    `                                                                                                                                                 
  - /api/internal/cart → /checkout/cart                                                                                                                                                                            
  - Updated frontend to use `Routes.checkout_cart_path()  `

app/controllers/checkout/carts_controller.rb
Pradumn27, line 1: why change the controller location? these are still internal apis just rendered via inertia

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -139,25 +131,27 @@ const addProduct = ({
...
+    add_products,
+    gift,
+    saved_credit_card,
+    recaptcha_key,
+    paypal_client_id,
+    max_allowed_cart_products,
+    tip_options,
+    default_tip_option,
+    ...props
+  } = usePage<Props>().props;
```
Pradumn27:
```suggestion
  } = cast<Props>(usePage().props);
```

spec/controllers/checkout/carts_controller_spec.rb
Pradumn27, line 1: [same](https://github.com/antiwork/gumroad/pull/3160/changes#r2751051262)

app/views/checkout/index.html.erb
Pradumn27: @yashranaway still don't see this being loaded in the component

Pradumn27 (changes requested):
@yashranaway _a

app/controllers/checkout/carts_controller.rb
yashranaway: Reverted, cart API stays at Api::Internal::CartsController. My bad, only the page itself needed to move to Inertia.

spec/controllers/checkout/carts_controller_spec.rb
yashranaway: Yep reverted this too, spec is back at api/internal/.

yashranaway:
> @yashranaway still don't see this being loaded in the component

@Pradumn27 _a  The recaptcha script is already loaded dynamically by the useRecaptcha hook in PaymentForm.tsx — it creates the script tag in the DOM itself. So the ERB partial was redundant.

app/controllers/checkout_controller.rb
```diff
@@ -3,9 +3,15 @@
 class CheckoutController < ApplicationController
   before_action :process_cart_id_param
 
+  layout "inertia"
+
   def index
-    @hide_layouts = true
-    @checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+    checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
```
Pradumn27: can inline this directly instead

Pradumn27 (changes requested):
@yashranaway _a, the `/internal/cart` api is still not migrated to Inertia. It's an internal api json endpoint consumed by the checkout page, so please add that migration here, and also add this api testing to the video in the PR description.

yashranaway:
Sorry still working

yashranaway:
@Pradumn27 _a Done! Migrated the cart API to /internal/cart (commit 882bddb) and updated the PR video.

app/controllers/checkout/carts_controller.rb
EmCousin: It was the right thing to do since it's no longer an API endpoint.

Issue #3148, opened 2026-01-26
Migrate Confirm Follow and Cancel Follow Pages to Inertia


EmCousin:
Endpoints:
- [x] GET `/follow/:id/confirm` -> followers#confirm
- [x] GET `/follow/:id/cancel` -> followers#cancel

PR #3153, opened 2026-01-26
Migrate Confirm Follow and Cancel Follow Pages to Inertia

TenzDelek:
Issue: #3148 

# Description

1. The confirm action already just redirects to the profile URL with a flash message 
>  no page template exists, so no Inertia page needed
Screenshot: 
image

2.  

- [x] GET /follow/:id/cancel -> followers#cancel

---

# Before
| Dark Mode | Light Mode |
|-----------|-----------|
| image|image

| Dark Mode | Light Mode |
|-----------|-----------|
| image|image

# After

| Dark Mode | Light Mode |
|-----------|-----------|
| image|image

| Light Mode | Dark Mode |
|-----------|-----------|
| image|image

---

# Test Results

image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Model: Opus 4.5
IDE: Cursor
Usage: Code understanding and Inertia.js migration

PR #3175, opened 2026-01-27
Migrate Confirm Follow and Cancel Follow Pages to Inertia

avi-474:
Issue: #3148 

# Description

Migrated the followers cancellation page to Inertia.js as part of the ongoing effort to modernize follow-flow pages. Also verified that the follow confirmation flow remains consistent and unchanged for users.

**Note:** The “confirmation page” referenced in the issue is not a standalone page in the current implementation. Confirmation is handled via a 302 redirect with a flash message on the creator’s profile, and this behavior is preserved.

## Problem

As tracked in Issue, parts of the follow-flow are still backed by legacy ERB views while the rest of the flow has been migrated to Inertia.js. The followers cancellation page was one of the remaining ERB-based screens.

- An inconsistent user experience compared to the rest of the follow-flow pages
- Extra maintenance overhead due to parallel view systems (ERB + Inertia)

## Solution

- Migrated the followers cancellation page from ERB to a new Inertia React component
  (app/javascript/pages/Followers/Cancel.tsx)
- Preserved the existing 302 redirect behavior for the confirmation flow so users land directly on the creator’s profile   page with the confirmation flash message. 

---

# Before/After

# Before:

Screenshot from 2026-01-27 01-04-40

Screenshot from 2026-01-27 01-04-54

Screenshot from 2026-01-27 17-33-49

Screenshot from 2026-01-27 17-34-26

# After:
Screenshot from 2026-01-27 16-36-35

Screenshot from 2026-01-27 15-41-07

Screenshot from 2026-01-27 17-33-49

Screenshot from 2026-01-27 17-34-26

---

# Test Results
[Screencast from 2026-01-27 15-06-01.webm](https://github.com/user-attachments/assets/8198a3a7-6061-402b-9024-3db52d0a4930)

- Manually verified the cancellation and confirmation flow in the browser
- All existing controller specs pass:
Screenshot from 2026-01-27 15-06-37

---

# Checklist

---

# AI Disclosure

**Model:** Claude Opus 4.5 via Antigravity
**Used for:**
- Understanding the existing follow and cancellation flow
- Exploring relevant tests and controller behavior

All code and tests were manually reviewed and modified by me to ensure correctness and consistency with existing behavior.

### DIFF
--- a/spec/controllers/followers_controller_spec.rb
+++ b/spec/controllers/followers_controller_spec.rb
@@ -230,6 +230,7 @@ describe FollowersController, inertia: true do
         expect { get :cancel, params: { id: follower.external_id } }.to change {
           follower.reload.deleted?
         }.from(false).to(true)
+        expect(response).to have_http_status(:ok)
         expect(inertia.component).to eq("Followers/Cancel")
       end
 


app/controllers/followers_controller.rb
```diff
@@ -92,9 +92,8 @@ def destroy
   def cancel
     follower_id = @follower.external_id
     @follower.mark_deleted!
-    @hide_layouts = true
     respond_to do |format|
```
pratyush07-hub, line 95: Fixed a bug where a double **format.html** call was present. The first call was an empty block that could potentially interfere with the intended Inertia render. Standardizing this to the single Inertia-specific **format.html** block.

app/javascript/pages/Followers/Cancel.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from "react";
+
+import { PoweredByFooter } from "$app/components/PoweredByFooter";
+
+function FollowersCancelPage() {
+  return (
+    <div className="flex min-h-screen flex-col items-center">
```
pratyush07-hub, line 7: We generally prefer using our shared layout components where possible. Does single-page-form provide enough padding on mobile, or should I wrap this in a Container component to ensure consistent margins across devices?

spec/controllers/followers_controller_spec.rb
```diff
@@ -225,11 +225,12 @@
     end
 
     describe "GET cancel" do
-      it "cancels the follow" do
+      it "cancels the follow and renders the Inertia page" do
         follower = create(:follower)
         expect { get :cancel, params: { id: follower.external_id } }.to change {
           follower.reload.deleted?
         }.from(false).to(true)
+        expect(inertia.component).to eq("Followers/Cancel")
```
pratyush07-hub, line 234: While checking the Inertia component is great, we should also explicitly assert **expect(response).to have_http_status(:ok)** to ensure the routing and layout middleware aren't throwing any unexpected errors.

avi-474:
**Note:**  I accidentally reviewed this PR while logged into my other GitHub account (@pratyush07-hub). Both accounts belong to me. Apologies for the confusion.

spec/controllers/followers_controller_spec.rb
avi-474: Added **expect(response).to have_http_status(:ok)** to explicitly verify that the controller returns a success status

EmCousin (commented):
Thanks @avi-474, but [this PR](https://github.com/antiwork/gumroad/pull/3153) was opened first and is almost identical, so we are moving forward with it

PR #3188, opened 2026-01-28
fix: tooltip position to prevent screen overflow

abhix4:
Issue: 
N/A (no related issue)

# Description

## Problem
- Tooltips using `bottom` position

 

## Solution

Update the Tooltip position to `left`

---

# Before/After

**Before:**

**`Wishlist page`**

image

**`UTM Links`**

image

image

image

**After:**

**`Wishlist page`**

image

---

# Test Results

N/A

---

# Checklist

---

# AI Disclosure

None


app/javascript/components/RichTextEditor.tsx
```diff
@@ -79,14 +79,16 @@ export const MenuItem = ({
...
 }: {
   name: string;
   icon: IconName;
   active?: boolean;
   disabled?: boolean;
   onClick?: () => void;
+  tooltipPosition?: Position
 }) => (
-  <MenuItemTooltip tip={name}>
+  <MenuItemTooltip tip={name} tooltipPosition={ tooltipPosition ?? "bottom" }>
```
abhix4: If tooltipPosition is not provided, the tooltip still renders at bottom.

abhix4:
@nyomanjyotisa ready for review. thank you.

nyomanjyotisa (changes requested):
@abhix4 can we check the whole app for similar issue and fix all?

abhix4:
@nyomanjyotisa reviewed the application and addressed similar issues where applicable.

binary-koan:
For "Generate new short link" if we put it on the left it'll overlap the other button which is likely annoying to use. I also don't like that we need to think about this for every new tooltip. Could we just fix the tooltip positioning to avoid going off the screen instead? (i.e. it should nudge the tooltip left until it's fully on screen)

Issue #3047, opened 2026-01-23
Migrate Discover Page to Inertia


EmCousin:
Endpoints:
- [x] `/discover`
- [x] `/discover/recommended_products` (JSON endpoint -> optional prop)
- [x] `/discover/recommended_wishlists` (JSON endpoint -> optional prop)
- [x] `/discover/search_autocomplete` (GET JSON endpoint -> redirect to Discover page)
- [x] `/discover/search_autocomplete` (DELETE JSON endpoint -> redirect to Discover page)

PR #3163, opened 2026-01-26
Migrate Discover Page to inertia

yashranaway:
Issue: #3047
Fixes: #3047

Previous PR: #3075

# Description

## Problem

The Discover page currently uses React on Rails, which requires maintaining separate rendering logic and doesn't benefit from Inertia's client-side navigation.

## Solution

Migrate the Discover page to Inertia following the established patterns in the codebase. This removes the old React on Rails setup and uses `render inertia:` with proper props.

### Changes

- Controller now uses `render inertia:` instead of setting `@react_discover_props`
- Page titles now properly show taxonomy breadcrumbs (e.g. "Software Development » Programming » C#")
- Header links include full URLs with protocol for cross-domain navigation
- Removed the old erb view, server-components wrapper, and discover pack

---

# Before/After
## Before

https://github.com/user-attachments/assets/6a31578c-d90e-43a0-9330-7914b46e3893

## After

https://github.com/user-attachments/assets/62d617a9-09e1-453a-8b59-6a04543c0409

---

# Test Results

**Controller tests:** 10/10 passing
```
bundle exec rspec spec/controllers/discover_controller_spec.rb
```

**System tests:** 52/64 passing (12 failures unrelated to migration)

The 12 remaining failures are pre-existing issues:
- 5 cart interaction tests - test helper `within_cart_item` looking for `h4` element
- 2 browser history tests - mixed Inertia/pushState navigation complexity
- 5 filter/search state tests - state management with navigation

image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Opus 4.5 was used to assist with code generation and migration patterns.

PR #3075, opened 2026-01-23
Migrate Discover Page to Inertia  

yashranaway:
Issue: #3047 
Fixes: #3047 

# Description

## Problem

The Discover page currently uses React on Rails, which requires maintaining separate rendering logic and doesn't benefit from Inertia's client-side navigation.

## Solution

Migrate the Discover page to Inertia following the established patterns in the codebase. This removes the old React on Rails setup and uses `render inertia:` with proper props.

---

# Before/After
## Before

https://github.com/user-attachments/assets/6a31578c-d90e-43a0-9330-7914b46e3893

## After 

https://github.com/user-attachments/assets/62d617a9-09e1-453a-8b59-6a04543c0409

---

# Test Results

image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Opus 4.5 was used to assist with code generation and migration patterns.

PR #3208, opened 2026-01-29
Migrate Discover Page to Inertia

yashranaway:
Issue: #3047
Fixes: #3047

Previous PR: #3075, #3163

# Description

## Problem

The Discover page currently uses React on Rails, which requires maintaining separate rendering logic and doesn't benefit from Inertia's client-side navigation.

## Solution

Migrate the Discover page to Inertia following the established patterns in the codebase. This removes the old React on Rails setup and uses `render inertia:` with proper props.

### Changes

- Controller now uses `render inertia:` instead of setting `@react_discover_props`
- Uses `layout "inertia"` for the index action
- Lazy props for `recommended_products`, `recommended_wishlists`, and other deferred data
- Removed the old erb view, server-components wrapper, and discover pack
- `RecommendedWishlists` component now accepts wishlists as props instead of fetching
- Removed duplicate `load_pack("inertia")` from inertia layout

### Endpoints

| Endpoint | Status |
|----------|--------|
| `/discover` | Migrated to Inertia |
| `/discover/recommended_products` | JSON endpoint (kept for backward compatibility) |
| `/discover/recommended_wishlists` | Migrated to lazy prop |
| `/discover/search_autocomplete` (GET) | JSON endpoint (required for real-time autocomplete) |
| `/discover/search_autocomplete` (DELETE) | JSON endpoint |

---

# Before/After
## Before

https://github.com/user-attachments/assets/6a31578c-d90e-43a0-9330-7914b46e3893

## After

https://github.com/user-attachments/assets/62d617a9-09e1-453a-8b59-6a04543c0409

---

# Test Results

**Controller tests:** 10/10 passing
```
bundle exec rspec spec/controllers/discover_controller_spec.rb
```

**System tests:** 52/64 passing (12 failures unrelated to migration)

The 12 remaining failures are pre-existing issues:
- 5 cart interaction tests - test helper `within_cart_item` looking for `h4` element
- 2 browser history tests - mixed Inertia/pushState navigation complexity
- 5 filter/search state tests - state management with navigation

image

---

# Checklist

---

# AI Disclosure

Claude Opus 4.5 was used to assist with code generation and migration patterns.


app/controllers/discover_controller.rb
```diff
@@ -7,13 +7,14 @@ class DiscoverController < ApplicationController
...
           DiscoverCuratedProducts, SearchProducts, AffiliateCookie
 
+  layout "inertia", only: [:index]
+
   before_action :set_affiliate_cookie, only: [:index]
 
   def index
     format_search_params!
 
     @hide_layouts = true
```
yashranaway, line 17: Required to prevent loading the base pack which causes SSR conflicts with the Inertia SSR bundle.

app/controllers/discover_controller.rb
```diff
@@ -49,17 +50,18 @@ def index
...
+      recommended_wishlists: -> { recommended_wishlists_data },
       curated_product_ids: curated_products.map { _1.product.external_id },
       search_offset: params[:from] || 0,
-      show_black_friday_hero: black_friday_feature_active?,
+      show_black_friday_hero: -> { black_friday_feature_active? },
       is_black_friday_page: params[:offer_code] == SearchProducts::BLACK_FRIDAY_CODE,
       black_friday_offer_code: SearchProducts::BLACK_FRIDAY_CODE,
-      black_friday_stats: black_friday_feature_active? ? BlackFridayStatsService.fetch_stats : nil,
+      black_friday_stats: -> { black_friday_feature_active? ? BlackFridayStatsService.fetch_stats : nil },
     }
```
yashranaway, line 65: - `recommended_products` and `recommended_wishlists` use lazy props (`-> { }`) to defer evaluation
- `search_results` is computed eagerly since it's always needed for initial render
- `curated_product_ids` uses external IDs for client-side matching

app/controllers/discover_controller.rb
```diff
@@ -49,17 +50,18 @@ def index
...
     prepare_discover_page
 
-    @react_discover_props = {
+    render inertia: "Discover/Index", props: {
       search_results: @search_results,
       currency_code: logged_in_user&.currency_type || "usd",
       taxonomies_for_nav:,
-      recommended_products: recommendations,
+      recommended_products: -> { recommendations },
+      recommended_wishlists: -> { recommended_wishlists_data },
```
yashranaway, line 58: Added as lazy prop per PR review - previously fetched via separate API endpoint.

app/controllers/discover/search_autocomplete_controller.rb
```diff
@@ -17,6 +17,6 @@ def delete_search_suggestion
       .by_user_or_browser(user: logged_in_user, browser_guid: cookies[:_gumroad_guid])
       .where(discover_searches: { query: params[:query] })
       .each(&:mark_deleted!)
-    head :no_content
+    head :ok
```
yashranaway, line 20: Changed from `head :no_content` - both are valid success responses. The DELETE endpoint stays as JSON (not redirect) because the frontend uses `deleteAutocompleteSearch()` API call with optimistic UI update.

app/javascript/components/Discover/RecommendedWishlists.tsx
yashranaway, line 1: Previously fetched wishlists via API. Now receives data as a lazy prop from the controller, which is more efficient and follows Inertia patterns.

yashranaway (commented):
@Pradumn27 _a  I have addressed comments from previous PR could please take a look?

app/controllers/discover/search_autocomplete_controller.rb
Pradumn27: Why this change of status code ? seems unnecessary

Also, why are both get and delete search autocomplete endpoints here not migrated? The goal is to use `useForm` from Inertia for search and delete actions, and these action calls redirect from these endpoints in the controller instead of sending a JSON result.

I see that the component for this search uses aborting calls right now, but we can change it to debounce as well, which would be better for the server load. Please implement this

app/controllers/discover_controller.rb
Pradumn27: why do we need to prevent this? now the page is in inertia so why is this prevention needed? there shouldn't be any ssr conflicts since the page is out of ssr now

app/controllers/discover_controller.rb
```diff
@@ -149,4 +151,20 @@ def prepare_discover_page
     def black_friday_feature_active?
       Feature.active?(:offer_codes_search) || (params[:feature_key].present? && ActiveSupport::SecurityUtils.secure_compare(params[:feature_key].to_s, ENV["SECRET_FEATURE_KEY"].to_s))
     end
+
+    def recommended_wishlists_data
+      wishlists = RecommendedWishlistsService.fetch(
+        limit: 4,
+        current_seller:,
+        curated_product_ids: curated_products.map { _1.product.id },
```
Pradumn27, line 159: why remove the decryption of obfuscate ids as present in the recommended_wishlists_controller ?

app/javascript/components/Discover/Footer.tsx
Pradumn27, line 1: Again - this is a shared footer so we need this to be in a shared component, probably - `Shared/Footer.tsx` or something like that, so when other pages are migrated they can use this footer and it remains consistent

app/javascript/components/Discover/RecommendedWishlists.tsx
```diff
@@ -1,49 +1,23 @@
...
-        showAlert(e.message, "error");
-      }
-    };
-    void loadWishlists();
-  });
-
-  return wishlists === null || wishlists.length > 0 ? (
+  wishlists: CardWishlist[];
+}) =>
+  wishlists.length > 0 ? (
```
Pradumn27, line 12: why remove the null or undefined check? it could be undefined during initial load

spec/controllers/discover_controller_spec.rb
```diff
@@ -82,22 +92,6 @@
...
-      end
-
-      it "renders as desktop if the user-agent is windows chrome" do
-        @request.user_agent = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36."
-        get :index
-
-        expect(response.body).to have_selector("[role='nav'] > * > [role='menubar']")
-      end
-    end
-
```
Pradumn27, line 100: why remove this? everything still needs to be tested

spec/controllers/discover_controller_spec.rb
Pradumn27, line 1: We are missing specs for the updated fetching mechanism of recommended wishlist and products, they need to be added here

app/controllers/discover_controller.rb
```diff
@@ -149,4 +151,20 @@ def prepare_discover_page
     def black_friday_feature_active?
       Feature.active?(:offer_codes_search) || (params[:feature_key].present? && ActiveSupport::SecurityUtils.secure_compare(params[:feature_key].to_s, ENV["SECRET_FEATURE_KEY"].to_s))
     end
+
+    def recommended_wishlists_data
```
Pradumn27, line 155: since we're moving this api here, we can remove the `recommended_wishlists_controller` , similarly also need to remove the routes for these apis

app/javascript/components/Discover/RecommendedWishlists.tsx
```diff
@@ -1,49 +1,23 @@
 import * as React from "react";
 
-import { fetchRecommendedWishlists } from "$app/data/wishlists";
```
Pradumn27, line 3: if this is removed then we can delete it as well now

Pradumn27:
@EmCousin please close the pr, @yashranaway can open another one once they resolve the comments, thanks!

PR #3212, opened 2026-01-29
Fix Google OAuth login failures

nyomanjyotisa:
Issue: https://github.com/antiwork/gumroad-private/issues/96

# Description

## Problem
Google OAuth login failed due to strict validations rejecting names with colons, unsupported avatar formats, and oversized profile pictures from Google accounts.

## Solution
Pre-validate and sanitize Google user data (remove colons from names, check avatar content types and file sizes) before database validation to prevent login failures.

---

# Before/After

## Before

https://github.com/user-attachments/assets/095f4fb9-e11c-4f80-9c15-089970c80be1

## After

https://github.com/user-attachments/assets/89f75d8f-11af-4b91-b2f2-5597d97649b6

# AI Disclosure

Cursor - Claude Sonnet 4.5 for code generation and all code self reviewed

### DIFF
--- a/app/models/concerns/user/social_google.rb
+++ b/app/models/concerns/user/social_google.rb
@@ -86,7 +86,7 @@ module User::SocialGoogle
       user.google_uid ||= data["uid"]
 
       if user.name.blank? && data["info"]["name"].present?
-        sanitized_name = data["info"]["name"].delete(":")
+        sanitized_name = data["info"]["name"].gsub(User::INVALID_NAME_FOR_EMAIL_DELIVERY_REGEX, "")
         user.name = sanitized_name
       end
 
--- a/spec/modules/user/social_google_spec.rb
+++ b/spec/modules/user/social_google_spec.rb
@@ -53,6 +53,7 @@ describe User::SocialGoogle do
       user = User.find_or_create_for_google_oauth2(@data_with_colon)
 
       expect(user).to be_persisted
+      expect(user).to be_valid
       expect(user.name).to eq("Test User")
     end
   end


app/models/concerns/user/social_google.rb
```diff
@@ -12,12 +12,18 @@ def google_picture_url(data)
...
 
     URI.open(pic_url) do |remote_file|
+      content_type = remote_file.content_type
+      return nil unless valid_avatar_content_type?(content_type)
+
       tempfile = Tempfile.new(binmode: true)
       tempfile.write(remote_file.read)
       tempfile.rewind
+
+      return nil if tempfile.size > User::Validations::MAXIMUM_AVATAR_FILE_SIZE
```
nyomanjyotisa, line 22: This prevents the validation error "Please upload a profile picture with a size smaller than 10 MB" by checking size before attachment. Large Google profile pictures won't break login and we simply skip the avatar and let the user log in successfully.

app/models/concerns/user/social_google.rb
```diff
@@ -28,6 +34,11 @@ def google_picture_url(data)
     nil
   end
 
+  def valid_avatar_content_type?(content_type)
+    return false if content_type.blank?
+    User::Validations::ALLOWED_AVATAR_EXTENSIONS.any? { |ext| content_type.include?(ext) }
```
nyomanjyotisa, line 39: This prevents the existing validation error "Please upload a profile picture with one of the following extensions: png, jpg, jpeg." by checking the content type before attempting to attach the avatar. If the Google profile picture is in an unsupported format (like webp), we gracefully skip it rather than failing the entire login.

app/models/concerns/user/social_google.rb
```diff
@@ -73,7 +84,11 @@ def query_google(user, data, new_user: false)
 
       # Don't set user properties if they already have values
       user.google_uid ||= data["uid"]
-      user.name ||= data["info"]["name"]
+
+      if user.name.blank? && data["info"]["name"].present?
+        sanitized_name = data["info"]["name"].delete(":")
+        user.name = sanitized_name
```
nyomanjyotisa, line 90: This fixes the validation error "Name cannot contain colons (:) as it causes email delivery problems."

app/models/concerns/user/social_google.rb
```diff
@@ -73,7 +84,11 @@ def query_google(user, data, new_user: false)
 
       # Don't set user properties if they already have values
       user.google_uid ||= data["uid"]
-      user.name ||= data["info"]["name"]
+
+      if user.name.blank? && data["info"]["name"].present?
+        sanitized_name = data["info"]["name"].delete(":")
```
EmCousin: Can you use the regexp constant `User::INVALID_NAME_FOR_EMAIL_DELIVERY_REGEX`? Hardcoding the colon here makes it not obvious why we are doing this when merely reading the code.

spec/modules/user/social_google_spec.rb
```diff
@@ -44,6 +44,17 @@
...
     end
+
+    it "creates user with sanitized name when name contains colons" do
+      @data_with_colon = @data.deep_dup
+      @data_with_colon["uid"] = "unique_colon_test_uid"
+      @data_with_colon["info"]["name"] = "Test: User"
+
+      user = User.find_or_create_for_google_oauth2(@data_with_colon)
+
+      expect(user).to be_persisted
```
EmCousin, line 55: It's hard to tell whether the user was found or created. Can we also expect the user to be valid?

```suggestion
      expect(user).to be_persisted
      expect(user).to be_valid
```

Issue #3061, opened 2026-01-23
Migrate Product Pages to Inertia


EmCousin:
Endpoints:
- `/l/:id`, `/l/:id/:code`,`/links/:id` (route to the same `links#show` action)

> [!NOTE]
> This page renders the product differently depending on the layout:
> ```erb
> <% if params[:layout] == "profile" %>
>   <%= react_component "ProfileProductPage", props: @product_props.merge({ creator_profile: ProfilePresenter.new(pundit_user:, seller: @product.user).creator_profile }), prerender: true %>
> <% elsif params[:layout] == "discover" %>
>   <%= react_component "DiscoverProductPage", props: @product_props.merge(@discover_props), prerender: true %>
> <% elsif params[:embed] || params[:overlay] %>
>   <%= react_component "ProductIframePage", props: @product_props, prerender: true %>
> <% else %>
>   <%= react_component "ProductPage", props: @product_props, prerender: true %>
> <% end %>
> ```
> 👉 Render a different Inertia page for each layout.

PR #3221, opened 2026-01-29
Migrate discover page to inertia

anshumanrana063:
Issue: #3047

# Description

Migrates the discover page endpoints to inertia:

```
/discover
/discover/recommended_products (JSON endpoint -> optional prop)
/discover/recommended_wishlists (JSON endpoint -> optional prop)
```

Have not migrated the search_autocomplete endpoints to inertia as needed by the issue because the apis are used by non-inertia migrated pages like product and wishlist discover pages thus they need to be migrated first before we migrate the search endpoints

---

# Before/After

# Before

https://github.com/user-attachments/assets/9eced30b-a5a8-4590-a059-15bc1fa31458

# After

https://github.com/user-attachments/assets/bd581a46-93da-42f8-b738-eb593cdccef4

---

# Test Results

Screenshot 2026-01-29 at 4 19 33 PM

---

# Checklist

---

# AI Disclosure

Used claude code with opus 4.5 for the implementation


app/controllers/discover_controller.rb
```diff
@@ -63,10 +66,6 @@ def index
     }
   end
 
-  def recommended_products
-    render json: recommendations
-  end
```
anshumanrana063, line 68: shifted this from json api to be part of the props on load of the page

app/controllers/discover_controller.rb
```diff
@@ -126,6 +125,34 @@ def taxonomy
...
+        taxonomy_id: taxonomy&.id
+      )
+
+      WishlistPresenter.cards_props(
+        wishlists:,
+        pundit_user:,
+        layout: Product::Layout::DISCOVER,
+        recommended_by: RecommendationType::GUMROAD_DISCOVER_WISHLIST_RECOMMENDATION,
+      )
+    end
```
anshumanrana063, line 153: moved from recommended_wishlists_controller to here to be part of the props of the page load

app/javascript/components/Shared/Footer/index.tsx
anshumanrana063, line 1: added shared footer, to substitute the one used by the erb template

app/controllers/discover_controller.rb
```diff
@@ -32,8 +31,9 @@ def index
     params[:include_rated_as_adult] = logged_in_user&.show_nsfw_products?
     params[:size] = INITIAL_PRODUCTS_COUNT
 
-    @search_results = search_products(params)
-    @search_results[:products] = @search_results[:products].includes(ProductPresenter::ASSOCIATIONS_FOR_CARD).map do |product|
+    search_results = search_products(params)
+    @search_results = search_results
```
Pradumn27: instead of making this global variable, we can just pass it directly to `prepare_discover_page` function instead and avoid this extra assignment

app/controllers/discover_controller.rb
```diff
@@ -49,11 +49,14 @@ def index
 
     prepare_discover_page
 
-    @react_discover_props = {
-      search_results: @search_results,
+    recommended_products = recommendations
```
Pradumn27: don't need this intermediate variable directly inline below

app/controllers/discover_controller.rb
```diff
@@ -49,11 +49,14 @@ def index
...
+
+    render inertia: "Discover/Index", props: {
+      search_results:,
       currency_code: logged_in_user&.currency_type || "usd",
       taxonomies_for_nav:,
-      recommended_products: recommendations,
+      recommended_products:,
+      recommended_wishlists: recommended_products.any? && params[:query].blank? ? recommended_wishlists : [],
```
Pradumn27: let's [lazy load](https://inertia-rails.dev/guide/partial-reloads#lazy-data-evaluation) these props for utilizing partial reloads

app/controllers/discover_controller.rb
```diff
@@ -126,6 +125,34 @@ def taxonomy
...
+
+    def discover_curated_product_ids
+      Array(params[:curated_product_ids]).filter_map do |id|
+        Integer(id)
+      rescue ArgumentError, TypeError
+        ObfuscateIds.decrypt(id)
+      rescue StandardError
+        nil
+      end
+    end
```
Pradumn27: lets do it inline only 
`curated_product_ids: (params[:curated_product_ids] || []).map { ObfuscateIds.decrypt(_1) },`

app/javascript/pages/Discover/Index.tsx
```diff
@@ -269,6 +269,17 @@ const Discover = (props: Props) => {
...
       });
+
+      // Keep server-provided recommended sections in sync with browser history navigation.
+      // (We use pushState for the URL, so we need an Inertia visit to refresh props.)
+      if (!newParams.query && !newParams.offer_code) {
+        router.visit(window.location.href, {
+          only: ["recommended_products", "recommended_wishlists"],
+          preserveState: true,
+          preserveScroll: true,
+          replace: true,
```
Pradumn27: this code is replicated thrice in this file, lets extract it into a `refreshRecommendations` function

app/javascript/pages/Discover/Index.tsx
Pradumn27: The component uses a confusing hybrid approach:                                                                                                       
  - Manual pushState for URL management                                                                                                     
  - useSearchReducer for local state                                                                                                       
  - router.visit() with only: for partial prop updates 

Both update state independently - the reducer updates search params locally while router.visit fetches recommendations. If the visit fails or is slow, the state can become inconsistent. 

Lets use full Inertia navigation with preserveState: true for everything

Pradumn27 (changes requested):
Hey @anshumanrana063 _a, thanks for the PR, the approach looks sound, added few comments please check and resolve

Pradumn27:
> As mentioned in the PR comment, I have not migrated the `search_autocomplete` get and delete apis to inertia because when I migrated them I kept getting ssr bundle failures because the search is used by other non-inertia pages which use the same layout namely the `Product` and `Wishlist` pages, so we can only migrate the api once these pages are migrated too, please check it out

In that case, we should adjust the scope of the issue in the description @EmCousin, and create a separate issue for this API migration, with dependence on migrating these 3 pages, namely `Discover`, `ProductPage`, and `DiscoverWishlistPage.`

Pradumn27:
> As mentioned in the PR comment, I have not migrated the `search_autocomplete` get and delete apis to inertia because when I migrated them I kept getting ssr bundle failures because the search is used by other non-inertia pages which use the same layout namely the `Product` and `Wishlist` pages, so we can only migrate the api once these pages are migrated too, please check it out

The `Wishlist` page you mentioned is already on inertia so shouldn't be a problem, we're only blocked by the `Products page`, which should be done once #3061 is completed, so let's wait for that

EmCousin:
#3061 is completed @Pradumn27 @anshumanrana063

EmCousin:
Issue got solved by https://github.com/antiwork/gumroad/pull/3517

Issue #3045, opened 2026-01-23
Migrate Communities Page to Inertia


EmCousin:
Endpoints:
- [ ] `/communities`
- [ ] `/communities/:sellerId/:communityIid` (currently a React router endpoint, needs to be moved to a `show` action in CommunitiesController)
- [ ] Chat Messages Internal API endpoints

dluffy56:
Hi @EmCousin  I’ve been stuck on the **InfiniteScroll** implementation for a while and could use some guidance.

### Issue
In a reverse-scroll setup, when I scroll up to load older messages, the component keeps fetching **all pages continuously** until the end.

### Video
https://github.com/user-attachments/assets/63f1451f-46bf-4e2b-9247-a25ba8f029c9

I also noticed the same behavior in #3189 

### What I’ve tried so far
- `manual` mode  
- `manualAfter={1}`  
- Adding cooldowns / debouncing between fetches  
- Adjusting threshold values  

None of these consistently prevent the runaway loading. If we ship this as-is, it results in pretty bad UX for chat-style views.

After digging deeper, I found this open issue which seems to describe the exact same behavior:
https://github.com/inertiajs/inertia/issues/2855

On the backend side, `InertiaRails.scroll()` with cursor pagination behaves correctly — the problem seems isolated to the frontend **InfiniteScroll** component in this use case.

Happy to try whatever you suggest. Thanks!

EmCousin:
@stefanbinoj I initially responded in your PR but I'd like to react to your latest recording; it doesn't look like great UX to fetch older messages one by one. But is it because you set `per_page` for testing? What does it look like with a realistic `per_page` (20 or even 50)?

My opinion is the simpler the better. If we can rely on vanilla Inertia infinite scroll without having to build our own logic, that should be the way to go.

Pradumn27:
> it doesn't look like great UX to fetch older messages one by one. But is it because you set `per_page` for testing? What does it look like with a realistic `per_page` (20 or even 50)?
> 

Hey @EmCousin, I’ve also added this as a review comment on @stefanbinoj’s PR. The root issue was that in the PR's implementation, the frontend wasn’t sending the `merge_intent` prop when fetching older messages. Since the backend determines how many messages to return based on this intent, it defaulted to sending just 1 message, instead of 100, as in the current production communities chat implementation.

> My opinion is the simpler the better. If we can rely on vanilla Inertia infinite scroll without having to build our own logic, that should be the way to go.

[...]

PR #3244, opened 2026-01-30
Migrate Communities Page to Inertia

yashranaway:
Fixes: #3045

# Description

## Problem

The Communities page currently uses React on Rails with React Router for client-side routing. This is inconsistent with the rest of the app which has been migrating to Inertia.js. The current implementation also has the chat message APIs under `/api/internal/` which doesn't follow the pattern we want for Inertia pages.

## Solution

Migrated the Communities page to Inertia.js:

- Updated `CommunitiesController` with proper `index` and `show` actions that render Inertia responses
- Moved chat message API controllers from `Api::Internal::Communities::*` to `Communities::*` namespace
- Created new Inertia page at `pages/Communities/Index.tsx`
- Moved React components from `server-components/CommunitiesPage/` to `components/Communities/`
- Updated frontend to use Inertia's `usePage()` instead of React Router's `useLoaderData()`
- Replaced React Router `Link` and `navigate()` with Inertia equivalents
- Updated route references from `internal_*` paths to new paths
- Cleaned up old files (react-router wrapper, pack file, erb view)

Key routing change: API routes now come BEFORE the show route to prevent `/communities/:community_id/chat_messages` from matching the show route pattern.

---

# Before/After

https://github.com/user-attachments/assets/8ea34c60-87c4-4970-9927-a5ad6f6e2050

https://github.com/user-attachments/assets/43488580-ca8d-4ed6-b57a-58d3b056304e

## Infinite Scroll

https://github.com/user-attachments/assets/3a2167a6-770e-4ffa-9016-988c65bcf19d

## Message Actions 

https://github.com/user-attachments/assets/ea75b64a-7cc4-4552-917d-3be0d662f5c8

---

# Test Results

Screenshot 2026-01-30 at 3 21 15 PM

---

# Checklist

---

# AI Disclosure

Claude Opus 4.5 was used to assist with the migration.

### DIFF
--- a/app/controllers/communities/chat_messages_controller.rb
+++ b/app/controllers/communities/chat_messages_controller.rb
@@ -15,10 +15,10 @@ class Communities::ChatMessagesController < ApplicationController
     if message.save
       message_props = CommunityChatMessagePresenter.new(message:).props
       broadcast_message(message_props, CommunityChannel::CREATE_CHAT_MESSAGE_TYPE)
-      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
     else
       redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id),
-                  inertia: { errors: { content: message.errors.full_messages.first } }
+                  inertia: { errors: { content: message.errors.full_messages.first } }, status: :see_other
     end
   end
 
@@ -26,10 +26,10 @@ class Communities::ChatMessagesController < ApplicationController
     if @message.update(permitted_params)
       message_props = CommunityChatMessagePresenter.new(message: @message).props
       broadcast_message(message_props, CommunityChannel::UPDATE_CHAT_MESSAGE_TYPE)
-      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
     else
       redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id),
-                  inertia: { errors: { content: @message.errors.full_messages.first } }
+                  inertia: { errors: { content: @message.errors.full_messages.first } }, status: :see_other
     end
   end
 
@@ -37,7 +37,7 @@ class Communities::ChatMessagesController < ApplicationController
     @message.mark_deleted!
     message_props = CommunityChatMessagePresenter.new(message: @message).props
     broadcast_message(message_props, CommunityChannel::DELETE_CHAT_MESSAGE_TYPE)
-    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
   end
 
   private
--- a/app/controllers/communities/last_read_chat_messages_controller.rb
+++ b/app/controllers/communities/last_read_chat_messages_controller.rb
@@ -17,7 +17,7 @@ class Communities::LastReadChatMessagesController < ApplicationController
       community_chat_message_id: message.id,
     )
 
-    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
   end
 
   private
--- a/app/controllers/communities/notification_settings_controller.rb
+++ b/app/controllers/communities/notification_settings_controller.rb
@@ -11,7 +11,7 @@ class Communities::NotificationSettingsController < ApplicationController
     settings = current_seller.community_notification_settings.find_or_initialize_by(seller: @community.seller)
     settings.update!(permitted_params)
 
-    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
   end
 
   private
--- a/app/javascript/pages/Communities/Index.tsx
+++ b/app/javascript/pages/Communities/Index.tsx
@@ -1,7 +1,6 @@
 import { Channel } from "@anycable/web";
 import { InfiniteScroll, router } from "@inertiajs/react";
 import cx from "classnames";
-import { debounce } from "lodash-es";
 import * as React from "react";
 import { is } from "ts-safe-cast";
 
@@ -17,6 +16,7 @@ import { CommunityList } from "$app/components/Communities/CommunityList";
 import { ScrollToBottomButton } from "$app/components/Communities/ScrollToBottomButton";
 import { DateSeparator } from "$app/components/Communities/Separator";
 import { useCommunities } from "$app/components/Communities/useCommunities";
+import { useDebouncedCallback } from "$app/components/useDebouncedCallback";
 import { UserAvatar } from "$app/components/Communities/UserAvatar";
 import { useCurrentSeller } from "$app/components/CurrentSeller";
 import { Icon } from "$app/components/Icons";
@@ -127,7 +127,7 @@ function CommunitiesIndex() {
       if (searchParams.has("notifications")) {
         const url = new URL(window.location.href);
         url.searchParams.delete("notifications");
-        window.history.replaceState({}, "", url.toString());
+        router.replace({ url: url.toString(), preserveState: true });
         setShowNotificationsSettings(true);
       }
     }
@@ -135,26 +135,21 @@ function CommunitiesIndex() {
 
   const markAsReadCancelRef = React.useRef<{ cancel: () => void } | null>(null);
 
-  const debouncedMarkAsRead = React.useMemo(
-    () =>
-      debounce((communityId: string, messageId: string) => {
-        if (!communityId || !messageId) return;
-        markAsReadCancelRef.current?.cancel();
-        router.post(
-          Routes.community_last_read_chat_message_path(communityId),
-          { message_id: messageId },
-          {
-            preserveState: true,
-            preserveScroll: true,
-            only: ["communities"],
-            onCancelToken: (token) => {
-              markAsReadCancelRef.current = token;
-            },
-          },
-        );
-      }, 500),
-    [],
-  );
+  const debouncedMarkAsRead = useDebouncedCallback((communityId: string, messageId: string) => {
+    if (!communityId || !messageId) return;
+    markAsReadCancelRef.current?.cancel();
+    router.post(
+      Routes.community_last_read_chat_message_path(communityId),
+      { message_id: messageId },
+      {
+        preserveScroll: true,
+        only: ["communities"],
+        onCancelToken: (token) => {
+          markAsReadCancelRef.current = token;
+        },
+      },
+    );
+  }, 500);
 
   const markMessageAsRead = React.useCallback(
     (message: CommunityChatMessage) => {
@@ -236,7 +231,6 @@ function CommunitiesIndex() {
       Routes.community_chat_messages_path(selectedCommunity.id),
       { community_chat_message: { content: selectedCommunityDraft.content } },
       {
-        preserveState: true,
         preserveScroll: true,
         only: [],
         onSuccess: () => {
@@ -273,18 +267,15 @@ function CommunitiesIndex() {
     return () => channel.disconnect();
   }, [cable, loggedInUser]);
 
-  const sendMessageToUserChannel = React.useCallback(
-    debounce((msg: OutgoingUserChannelMessage) => {
-      const userChannelState = userChannelRef.current?.state;
-      if (userChannelState === "connected" || userChannelState === "idle") {
-        userChannelRef.current?.send(msg).catch((e: unknown) => {
-          // eslint-disable-next-line no-console
-          console.error(e);
-        });
-      }
-    }, 100),
-    [],
-  );
+  const sendMessageToUserChannel = useDebouncedCallback((msg: OutgoingUserChannelMessage) => {
+    const userChannelState = userChannelRef.current?.state;
+    if (userChannelState === "connected" || userChannelState === "idle") {
+      userChannelRef.current?.send(msg).catch((e: unknown) => {
+        // eslint-disable-next-line no-console
+        console.error(e);
+      });
+    }
+  }, 100);
 
   React.useEffect(() => {
     communities.forEach((community) => {
@@ -387,7 +378,6 @@ function CommunitiesIndex() {
           Routes.community_chat_message_path(communityId, messageId),
           { community_chat_message: { content } },
           {
-            preserveState: true,
             preserveScroll: true,
             only: [],
             onSuccess: () => resolve(),
@@ -402,7 +392,6 @@ function CommunitiesIndex() {
     (messageId: string, communityId: string) =>
       new Promise<void>((resolve, reject) => {
         router.delete(Routes.community_chat_message_path(communityId, messageId), {
-          preserveState: true,
           preserveScroll: true,
           only: [],
           onSuccess: () => resolve(),
@@ -418,7 +407,6 @@ function CommunitiesIndex() {
         Routes.community_notification_setting_path(community.id),
         { settings },
         {
-          preserveState: true,
           preserveScroll: true,
           only: ["notification_settings"],
           onSuccess: () => {


config/routes.rb
```diff
@@ -792,7 +790,13 @@ def product_info_and_purchase_routes(named_routes: true)
...
 
     # communities
-    get "/communities(/:seller_id/:community_id)", to: "communities#index", as: :community
+    get "/communities", to: "communities#index", as: :communities
+    scope "/communities/:community_id", as: :community, defaults: { format: :json } do
+      resources :chat_messages, only: [:index, :create, :update, :destroy], controller: "communities/chat_messages"
+      resource :last_read_chat_message, only: [:create], controller: "communities/last_read_chat_messages"
+      resource :notification_setting, only: [:update], controller: "communities/notification_settings"
+    end
+    get "/communities/:seller_id/:community_id", to: "communities#show", as: :community
```
yashranaway, line 814: The API routes (`chat_messages`, `last_read_chat_message`, `notification_setting`) MUST come before the show route. Otherwise Rails matches `/communities/abc123/chat_messages` as `seller_id: "abc123", community_id: "chat_messages"` which breaks the API calls. Learned this the hard way when tests were failing.

app/controllers/communities_controller.rb
```diff
@@ -2,17 +2,36 @@
...
+    render inertia: "Communities/Index", props: props
+  end
+
+  def show
+    seller = User.find_by_external_id!(params[:seller_id])
+    community = Community.alive.find_by_external_id!(params[:community_id])
+
+    raise ActiveRecord::RecordNotFound unless community.seller_id == seller.id
+
+    authorize community
```
yashranaway, line 25: We first find the seller, then find the community, then verify the community belongs to that seller. This prevents accessing communities via wrong seller IDs in the URL.

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -1,13 +1,17 @@
...
   before_action :set_message, only: [:update, :destroy]
   after_action :verify_authorized
 
   def index
-    render json: PaginatedCommunityChatMessagesPresenter.new(community: @community, timestamp: params[:timestamp], fetch_type: params[:fetch_type]).props
+    render json: PaginatedCommunityChatMessagesPresenter.new(
+      community: @community,
+      timestamp: params[:timestamp],
+      fetch_type: params[:fetch_type]
+    ).props
```
yashranaway: The index action returns JSON (not Inertia redirect) because it's used for cursor-based pagination/infinite scroll. This doesn't fit the Inertia partial reload pattern. The frontend fetches messages via AJAX and merges them into state.

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -41,31 +45,32 @@ def destroy
...
+
+  def broadcast_message(message_props, type)
+    CommunityChannel.broadcast_to(
+      "community_#{@community.external_id}",
+      { type:, message: message_props },
+    )
+  rescue => e
+    Rails.logger.error("Error broadcasting message to community channel: #{e.message}")
+    Bugsnag.notify(e)
+  end
```
yashranaway: Wrapped the ActionCable broadcast in a rescue block. If broadcasting fails, we log and notify Bugsnag but don't fail the request. The message was saved successfully, it just won't appear in real-time for other users.

app/javascript/components/Communities/CommunityView.tsx
yashranaway, line 1: Replaced `useNavigate` and `useLocation` from react-router-dom with `router` from @inertiajs/react. All navigation now uses `router.get()` with `preserveScroll: true`.

app/javascript/components/Communities/CommunityView.tsx
```diff
@@ -517,7 +513,8 @@ export const CommunityView = () => {
 
     const community = communities.find((community) => community.id === communityId);
     if (!community) return;
-    window.location.replace(`/communities/${community.seller.id}/${community.id}`);
+
+    router.get(Routes.community_path(community.seller.id, community.id), {}, { replace: true });
   });
```
yashranaway: When no community is selected (user visits `/communities`), we auto-redirect to the first available community. If the current user is a seller, we prefer their own community first.

app/javascript/data/communities.ts
yashranaway, line 1: All route references changed from `internal_community_*` to `community_*`. The routes now point to the new controllers under `/communities/:community_id/` instead of `/internal/communities/:community_id/`.

app/javascript/components/Communities/useCommunities.ts
yashranaway, line 1: The hook now gets data from Inertia's `usePage().props` instead of React Router's `useLoaderData()`. The shape of the data is the same, just coming from a different source.

spec/requests/communities_spec.rb
yashranaway, line 1: Tests that visit the index page now use `communities_path` (plural) instead of `community_path` (singular without args). The singular `community_path` now requires both `seller_id` and `community_id` parameters.

yashranaway:
@EmCousin Could you please take a look?

app/controllers/communities/chat_messages_controller.rb
Pradumn27, line 1: Why are we changing the controller location? The existing structure already seems appropriate for how the controllers are organized. Since these APIs are still internal and haven’t been migrated yet, it doesn’t seem necessary to make changes here.

app/controllers/communities/chat_messages_controller.rb
Pradumn27: We do have inertia workarounds for infinite scroll in inertia, but due to the use of websockets here I also prefer the json API approach, since inertia would be an overhead in this situation and not of any benefit.

app/controllers/communities/last_read_chat_messages_controller.rb
Pradumn27, line 1: [same](https://github.com/antiwork/gumroad/pull/3244/changes#r2746935479)

app/controllers/communities/last_read_chat_messages_controller.rb
```diff
@@ -16,10 +16,11 @@ def create
   end
 
   private
-    def set_community
-      @community = Community.find_by_external_id(params[:community_id])
-      return e404_json unless @community
```
Pradumn27, line 27: seeing a lot of unnecessary extra lines in the PR, please check and remove them everywhere

app/controllers/communities/notification_settings_controller.rb
Pradumn27, line 1: [same](https://github.com/antiwork/gumroad/pull/3244/changes#r2746935479)

app/controllers/communities_controller.rb
```diff
@@ -2,17 +2,36 @@
...
+    props[:selectedCommunityId] = community.external_id
+
+    render inertia: "Communities/Index", props: props
   end
 
   private
-    def set_default_page_title
-      set_meta_tag(title: "Communities")
-    end
+
```
Pradumn27, line 43: same question about extra lines

app/javascript/components/client-components/Nav/index.tsx
```diff
@@ -145,7 +145,7 @@ export const Nav = (props: Props) => {
           <ClientNavLink text="Payouts" icon="bank" href={Routes.balance_url(routeParams)} />
         ) : null}
         {loggedInUser?.policies.community.index ? (
-          <NavLink text="Community" icon="solid-chat-alt" href={Routes.community_path(routeParams)} />
```
Pradumn27, line 148: let's avoid changing routes

app/javascript/pages/Communities/Index.tsx
Pradumn27, line 1: lets directly inline the component here

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 
-class Api::Internal::Communities::ChatMessagesController < Api::Internal::BaseController
+class Communities::ChatMessagesController < ApplicationController
```
Pradumn27, line 3: Why change the base? There aren't any inertia uses here

app/javascript/components/Communities/useCommunities.ts
```diff
@@ -17,24 +17,27 @@ export type CommunityChat = {
...
 
 export const useCommunities = () => {
-  const data = cast<{
-    has_products: boolean;
-    communities: Community[];
-    notification_settings: CommunityNotificationSettings;
-    selectedCommunityId?: string;
-  }>(useLoaderData());
-  const [communities, setCommunities] = React.useState<Community[]>(sortByName(data.communities));
+  const pageProps = cast<PageProps>(usePage().props);
```
Pradumn27: lets inline deconstrcuting here

app/javascript/components/Communities/useCommunities.ts
```diff
@@ -17,24 +17,27 @@ export type CommunityChat = {
...
+  const { has_products, communities: initialCommunities, notification_settings, selectedCommunityId: initialSelectedCommunityId } = pageProps;
+  const [communities, setCommunities] = React.useState<Community[]>(sortByName(initialCommunities));
```
Pradumn27: why do we need a local state, if it's coming from the server?

app/javascript/data/communities.ts
Pradumn27: lets migrate this to inertia too

app/controllers/communities/last_read_chat_messages_controller.rb
```diff
@@ -16,10 +16,11 @@ def create
   end
```
Pradumn27, line 21: why isn;t this api migrated?

Pradumn27 (changes requested):
Hey @yashranaway _a, did an initial review, please check the comments.

Additionally, there are a few other apis concerning the communities chat message controller, a few of them that I saw were `Create chat message` and `mark message as read`, so please look at all and migrate them too, and if you're migrating the controller location, please provide a reasoning too - otherwise we can keep them at the same place

yashranaway:
Hey @Pradumn27 _a , thanks for the review! I've addressed most of the comments:

I wanted to clarify something about the API migration comments.

On `chat_messages_controller.rb:14`, you mentioned:

> "due to the use of websockets here I also prefer the json API approach, since inertia would be an overhead in this situation"

But on `data/communities.ts:130` and `last_read_chat_messages_controller.rb:16`, there are comments to migrate these to Inertia.

Since `markCommunityChatMessagesAsRead` is called frequently (on scroll, debounced) and the chat uses websockets for real-time updates, I kept it as a JSON API to avoid unnecessary page prop refreshes. This also aligns with how `data/wishlists.ts` handles similar background operations.

**Could you clarify if you'd still like me to migrate the `last_read` and `notification_settings` APIs to use Inertia's `router.post/put`? Or is keeping them as JSON APIs fine given the websocket context?**

Pradumn27:
Hey @yashranaway _a, I’ve gone through the approach more carefully now - I’d only skimmed it before.
                                                                                               
We can migrate all the community APIs (chat messages, last read, notification settings) to Inertia while preserving real-time behavior. Here's the pattern:                                
                                                                                                
- Server: On success, redirect to the Communities page instead of returning JSON. Keep WebSocket broadcasts for real-time updates to other users.                                                                                                                                             
- Client: Use Inertia form submissions with preserveState and preserveScroll so the WebSocket connection, scroll position, and component state are preserved across the redirect.                                                                                                          
- Payload: Use lazy props (lambdas) on the server and Inertia's only option on the client to request just the props that changed. This keeps response payloads small - comparable to the current JSON responses.                                                                         
                                                                                                
The overhead is one additional GET per mutation (handled internally by Inertia), which is acceptable for this use case. This migration aligns Communities with the rest of the app as we move to Inertia.                                                                                
                                                                                                
For pagination: Replace the current manual scroll handling with Inertia v2's built-in [InfiniteScroll](https://inertia-rails.dev/guide/infinite-scroll) component in reverse mode. It's designed specifically for chat interfaces and handles bidirectional loading and scroll preservation automatically. On the server, use Inertia::scroll() with cursor pagination so new pages merge with existing content rather than replacing it. For the initial load, center on the last-read message; handle this server-side by returning the first messages page already centered on the last-read timestamp.

yashranaway:
Hi @Pradumn27 _a , thanks for the detailed breakdown , this clears things up.

I've implemented exactly this pattern in the latest push:

- **Mutations** (chat messages, last read, notification settings): All redirect on success now. Client uses `router.post`/`put`/`delete` with `preserveState`, `preserveScroll`, and `only` scoped to just the props that changed. ActionCable broadcasts still handle real-time updates to other clients.
- **Pagination**: Replaced manual scroll handling with Inertia v2's `InfiniteScroll` (reverse mode) + `InertiaRails.scroll()` with cursor pagination. Initial load centers on the last-read timestamp server-side.
- **Props**: All wrapped in lambdas for lazy evaluation.

Let me know if anything needs adjusting.

app/controllers/communities/chat_messages_controller.rb
EmCousin: That's the correct thing to do, those are no longer API endpoints rendering JSON responses.

app/controllers/communities_controller.rb
```diff
@@ -2,17 +2,82 @@
...
+    seller = User.find_by_external_id!(params[:seller_id])
+    community = Community.alive.find_by_external_id!(params[:community_id])
+    raise ActiveRecord::RecordNotFound unless community.seller_id == seller.id
+
+    authorize community
+
+    messages_data = paginated_messages(community)
+    metadata = scroll_metadata(community, messages_data)
+
+    render inertia: "Communities/Index", props: {
```
EmCousin, line 30: ~Please use a dedicated Show template. Common code across template can be factored into a component.~

app/javascript/components/server-components/CommunitiesPage/ChatMessageList.tsx
EmCousin, line 119: You can  bring those comments back

app/controllers/communities_controller.rb
EmCousin: Actually don't. This might break the websocket connection

yashranaway:
@EmCousin _a Added a video showing real-time updates across two browser windows (seller + buyer). Also reverted back comments. Could you please take a look?

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -17,40 +15,42 @@ def create
     if message.save
       message_props = CommunityChatMessagePresenter.new(message:).props
       broadcast_message(message_props, CommunityChannel::CREATE_CHAT_MESSAGE_TYPE)
-      render json: { message: message_props }
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
     else
-      render json: { error: message.errors.full_messages.first }, status: :unprocessable_entity
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id),
+                  inertia: { errors: { content: message.errors.full_messages.first } }
```
EmCousin: Where is this used?

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -17,40 +15,42 @@ def create
...
   def update
     if @message.update(permitted_params)
       message_props = CommunityChatMessagePresenter.new(message: @message).props
       broadcast_message(message_props, CommunityChannel::UPDATE_CHAT_MESSAGE_TYPE)
-      render json: { message: message_props }
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
     else
-      render json: { error: @message.errors.full_messages.first }, status: :unprocessable_entity
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id),
+                  inertia: { errors: { content: @message.errors.full_messages.first } }
```
EmCousin: Where is this used?

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -17,40 +15,42 @@ def create
...
     end
   end
 
   def destroy
     @message.mark_deleted!
     message_props = CommunityChatMessagePresenter.new(message: @message).props
     broadcast_message(message_props, CommunityChannel::DELETE_CHAT_MESSAGE_TYPE)
-    head :ok
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
```
EmCousin: Let's make sure the redirection is a GET request after destroying a message
```suggestion
    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
```

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -17,40 +15,42 @@ def create
...
+                  inertia: { errors: { content: message.errors.full_messages.first } }
     end
   end
 
   def update
     if @message.update(permitted_params)
       message_props = CommunityChatMessagePresenter.new(message: @message).props
       broadcast_message(message_props, CommunityChannel::UPDATE_CHAT_MESSAGE_TYPE)
-      render json: { message: message_props }
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
```
EmCousin: Let's make sure the redirection is a GET request after updating a message
```suggestion
      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
```

app/controllers/communities/chat_messages_controller.rb
```diff
@@ -17,40 +15,42 @@ def create
     if message.save
       message_props = CommunityChatMessagePresenter.new(message:).props
       broadcast_message(message_props, CommunityChannel::CREATE_CHAT_MESSAGE_TYPE)
-      render json: { message: message_props }
+      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
```
EmCousin: Let's make sure the redirection is a GET request after creating a message
```suggestion
      redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
```

app/controllers/communities/last_read_chat_messages_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    message = @community.community_chat_messages.find_by_external_id(params[:message_id])
+    return e404 unless message
+
+    LastReadCommunityChatMessage.set!(
+      user_id: current_seller.id,
+      community_id: @community.id,
+      community_chat_message_id: message.id,
+    )
+
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
```
EmCousin:
```suggestion
    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id) , status: :see_other
```

app/controllers/communities/notification_settings_controller.rb
```diff
@@ -1,21 +1,24 @@
...
   after_action :verify_authorized
 
+  layout "inertia"
+
   def update
     settings = current_seller.community_notification_settings.find_or_initialize_by(seller: @community.seller)
     settings.update!(permitted_params)
 
-    render json: { settings: CommunityNotificationSettingPresenter.new(settings:).props }
+    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id)
```
EmCousin:
```suggestion
    redirect_to community_path(seller_id: @community.seller.external_id, community_id: @community.external_id), status: :see_other
```

app/javascript/components/client-components/Nav/index.tsx
EmCousin: @yashranaway can you explain why this route change is needed?

app/javascript/components/server-components/Nav.tsx
```diff
@@ -147,7 +147,7 @@ export const Nav = (props: Props) => {
           <NavLink text="Payouts" icon="bank" href={Routes.balance_url(routeParams)} />
         ) : null}
         {loggedInUser?.policies.community.index ? (
-          <NavLink text="Community" icon="solid-chat-alt" href={Routes.community_path(routeParams)} />
+          <NavLink text="Community" icon="solid-chat-alt" href={Routes.communities_path()} />
```
EmCousin, line 150: Why the route change here?

app/javascript/pages/Communities/Index.tsx
```diff
@@ -91,91 +73,85 @@ export const scrollTo = (
...
     if (selectedCommunity) {
-      const searchParams = new URLSearchParams(location.search);
+      const searchParams = new URLSearchParams(window.location.search);
       if (searchParams.has("notifications")) {
-        searchParams.delete("notifications");
-        const newSearch = searchParams.toString() ? `?${searchParams.toString()}` : "";
-        navigate(`${location.pathname}${newSearch}${location.hash}`, { replace: true });
+        const url = new URL(window.location.href);
+        url.searchParams.delete("notifications");
+        window.history.replaceState({}, "", url.toString());
```
EmCousin: Let's try to use Inertia for everything related to history state
```suggestion
        router.replace({ url: url.toString(), preserveState: true });
```

app/javascript/pages/Communities/Index.tsx
```diff
@@ -336,54 +197,58 @@ export const CommunityView = () => {
...
-      insertOrUpdateMessage(data.message);
-    } catch (_error: unknown) {
-      updateCommunityDraft(selectedCommunity.id, { isSending: false });
-    }
-  };
+    router.post(
+      Routes.community_chat_messages_path(selectedCommunity.id),
+      { community_chat_message: { content: selectedCommunityDraft.content } },
+      {
+        preserveState: true,
```
EmCousin: This is already the case by default for POST requests
```suggestion
```

app/javascript/pages/Communities/Index.tsx
```diff
@@ -408,15 +273,18 @@ export const CommunityView = () => {
...
-    const userChannelState = userChannelRef.current?.state;
-    if (userChannelState === "connected" || userChannelState === "idle") {
-      userChannelRef.current?.send(msg).catch((e: unknown) => {
-        // eslint-disable-next-line no-console
-        console.error(e);
-      });
-    }
-  }, 100);
+  const sendMessageToUserChannel = React.useCallback(
+    debounce((msg: OutgoingUserChannelMessage) => {
```
EmCousin: Why cannot we use `useDebouncedCallback` anymore?

app/javascript/pages/Communities/Index.tsx
```diff
@@ -541,48 +380,64 @@ export const CommunityView = () => {
...
-    return response;
-  };
+  const updateMessage = React.useCallback(
+    (messageId: string, communityId: string, content: string) =>
+      new Promise<void>((resolve, reject) => {
+        router.put(
+          Routes.community_chat_message_path(communityId, messageId),
+          { community_chat_message: { content } },
+          {
+            preserveState: true,
```
EmCousin: This is already the case by default for PUT requests
```suggestion
```

app/javascript/pages/Communities/Index.tsx
```diff
@@ -541,48 +380,64 @@ export const CommunityView = () => {
...
-  const deleteMessage = async (messageId: string, communityId: string) => {
-    const response = await deleteCommunityChatMessage({ communityId, messageId });
-    removeMessage(messageId, communityId);
-    return response;
-  };
+  const deleteMessage = React.useCallback(
+    (messageId: string, communityId: string) =>
+      new Promise<void>((resolve, reject) => {
+        router.delete(Routes.community_chat_message_path(communityId, messageId), {
+          preserveState: true,
```
EmCousin: This is already the case by default for DELETE requests
```suggestion
```

app/javascript/pages/Communities/Index.tsx
```diff
@@ -541,48 +380,64 @@ export const CommunityView = () => {
...
-    const response = await updateCommunityNotificationSettings({
-      communityId: community.id,
-      settings,
+  const saveNotificationsSettings = async (community: Community, settings: NotificationSettings) =>
+    new Promise<void>((resolve, reject) => {
+      router.put(
+        Routes.community_notification_setting_path(community.id),
+        { settings },
+        {
+          preserveState: true,
```
EmCousin: Non-GET requests preserve state by default
```suggestion
```

EmCousin (changes requested):

app/controllers/communities/chat_messages_controller.rb
yashranaway: Those are the error paths for when model validation fails (e.g. blank content). They pass validation errors back via `inertia: { errors: ... }` so the client can display them.

app/controllers/communities/chat_messages_controller.rb
yashranaway: https://github.com/antiwork/gumroad/pull/3244#discussion_r2798691441

app/javascript/components/client-components/Nav/index.tsx
yashranaway: The old `community_path(routeParams)` helper no longer exists, we split the single optional-segment route into `communities#index` and `communities#show`. The nav links to the index (no params), so it uses `communities_path()` now.

app/javascript/pages/Communities/Index.tsx
yashranaway: My bad, we can fixed.

EmCousin:
@yashranaway Also please be more thorough with your recordings - we'd like to see at a glance how the infinite scroll works with multiple pages, and for the live updates it would be nice to see the same tests for updated/deleted messages

yashranaway:
> @yashranaway Also please be more thorough with your recordings - we'd like to see at a glance how the infinite scroll works with multiple pages, and for the live updates it would be nice to see the same tests for updated/deleted messages

Apologies for that. I’ve now added the requested recordings and updated the PR description accordingly.

EmCousin:
Tested locally -
- Reverse infinite scroll does not work well, if keeps loading previous messages when scrolling at the top infinitely, it should stop after loading previous messages once, and the user should be scrolling again, like the current behaviour
- When posting messages in quick succession, the form gets suck in the disabled state and I cannot send messages anymoire

PR #3248, opened 2026-01-30
Auto-restart existing subscription on checkout instead of creating duplicate

yashranaway:
Issue: #173

# Description

Per team's [decision](https://github.com/antiwork/gumroad/pull/2347#issuecomment-3724441879): restart subscriptions transparently without asking—there's no use case for starting fresh.

## Problem

## Solution

Modified `Purchase::BaseService` to automatically detect and reuse existing subscriptions on checkout:

### 1. Active Subscription Detection
When a user with an **active** subscription purchases again:
- Associates the new purchase with the existing subscription
- Updates the credit card on file
- Clears any pending cancellation (if `cancelled_at` is in the future)
- No duplicate subscriptions created

### 2. Deactivated Subscription Restart
When a user with a **deactivated** subscription purchases again:
- Reactivates the existing subscription via `resubscribe!`
- Updates the credit card on file
- Associates the new purchase with the restarted subscription
- Normal receipt sent (no "restart" notifications to match regular product behavior)

### Matching Logic
1. First matches by user ID for logged-in users
2. Falls back to email matching for guest checkouts (SQL JOIN for performance)
3. Picks the most recently deactivated subscription if multiple exist

### Exclusions
A subscription is NOT restarted when:
- It was cancelled by the seller (intentional block)
- It has ended (`ended_at` is set)
- It's a test subscription
- The new purchase is a test purchase
- The product is deleted

---

# Before/After

---

# Test Results

image

---

# Checklist

---

# AI Disclosure


app/services/purchase/base_service.rb
```diff
@@ -30,6 +30,18 @@ def handle_purchase_success
...
+        handle_active_subscription_repurchase(active_subscription, giftee_purchase)
+        return
+      end
+
+      existing_subscription = find_restartable_subscription(giftee_purchase)
+      if existing_subscription
+        restart_existing_subscription(existing_subscription, giftee_purchase)
+        return
+      end
+
```
yashranaway, line 44: I check for **active** subscriptions first, then **deactivated** ones. The reasoning: if someone has an active subscription and tries to purchase again, we shouldn't restart anything—just associate the new purchase with their existing subscription and update their card on file.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +79,89 @@ def create_subscription(giftee_purchase)
...
+      is_gift = purchase.is_gift_sender_purchase
+      user = is_gift ? giftee_purchase&.purchaser : purchase.purchaser
+      email = is_gift ? giftee_purchase&.email : purchase.email
+
+      base_scope = purchase.link.subscriptions
+        .not_is_test_subscription
+        .where.not(deactivated_at: nil)
+        .where(ended_at: nil)
+        .where("cancelled_at IS NULL OR (subscriptions.flags & ?) != 0", Subscription.flag_mapping["flags"][:cancelled_by_buyer])
```
yashranaway, line 95: This filter means we only restart if:
- Not cancelled at all, OR
- Cancelled by the buyer themselves

If a seller cancelled the subscription, that's an intentional block and we shouldn't auto-restart. The user would need to start fresh.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +79,89 @@ def create_subscription(giftee_purchase)
...
+      end
+
+      return nil if email.blank?
+
+      original_purchase_flag = Purchase.flag_mapping["flags"][:is_original_subscription_purchase]
+      base_scope
+        .joins("INNER JOIN purchases ON purchases.subscription_id = subscriptions.id AND (purchases.flags & #{original_purchase_flag}) != 0")
+        .where("LOWER(purchases.email) = ?", email.downcase)
+        .first
+    end
```
yashranaway, line 110: For guest checkouts (no user account), we match by email. I'm using a SQL JOIN here instead of loading subscriptions and iterating in Ruby. This keeps it efficient even if a product has many subscriptions.

We only check the original subscription purchase's email (via the flag) since that's the authoritative email for that subscription.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +79,89 @@ def create_subscription(giftee_purchase)
...
+
+    def handle_active_subscription_repurchase(subscription, giftee_purchase)
+      is_gift = purchase.is_gift_sender_purchase
+
+      ActiveRecord::Base.transaction do
+        subscription.credit_card = purchase.credit_card unless is_gift
+        if subscription.cancelled_at.present? && subscription.cancelled_at > Time.current
+          subscription.cancelled_at = nil
+          subscription.user_requested_cancellation_at = nil
+          subscription.cancelled_by_buyer = false
```
yashranaway, line 158: If someone scheduled a cancellation for next week but then repurchases, they probably changed their mind. We clear the pending cancellation so their subscription continues uninterrupted.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +79,89 @@ def create_subscription(giftee_purchase)
...
+    def restart_existing_subscription(subscription, giftee_purchase)
+      is_gift = purchase.is_gift_sender_purchase
+
+      ActiveRecord::Base.transaction do
+        subscription.credit_card = purchase.credit_card unless is_gift
+        subscription.resubscribe!
+        subscription.purchases << [purchase, giftee_purchase].compact
+      end
+    end
+
```
yashranaway, line 121: The old approach called `send_restart_notifications!` which sends a "Your subscription has been restarted" email. Per the team decision, we're treating this like a regular product purchase—just the normal receipt, no special "restart" messaging.

spec/services/purchase/base_service_spec.rb
yashranaway, line 1: 19 test cases covering:

* Deactivated subscription restart
* Seller vs buyer cancellation – different behavior
* Ended subscriptions – create new
* Email matching for guests
* Multiple subscriptions – picks most recent
* Test subscriptions/purchases – excluded
* Deleted products – excluded
* Active subscription association
* Pending cancellation clearing
* Gift subscription restart

app/services/purchase/base_service.rb
EmCousin: ⚠️ There's a fundamental flaw here: this is running **after** the payment is processed, so it cannot prevent charging users who already have active subscriptions or who are still within their billing period.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +79,89 @@ def create_subscription(giftee_purchase)
...
+        subscription.credit_card = purchase.credit_card unless is_gift
+        if subscription.cancelled_at.present? && subscription.cancelled_at > Time.current
+          subscription.cancelled_at = nil
+          subscription.user_requested_cancellation_at = nil
+          subscription.cancelled_by_buyer = false
+        end
+        subscription.save! if subscription.changed?
+        subscription.purchases << [purchase, giftee_purchase].compact
+      end
+    end
```
EmCousin, line 163: We need to notify users by email for active sub attempts.

PR #3251, opened 2026-01-30
 Auto-restart existing subscription on checkout instead of creating duplicate 

yashranaway:
Issue: #173

Previous PR : #3248

# Description

Per team's [decision](https://github.com/antiwork/gumroad/pull/2347#issuecomment-3724441879): restart subscriptions transparently without asking—there's no use case for starting fresh.

## Problem

## Solution

Modified `Purchase::CreateService` and `Purchase::BaseService` to handle subscription states on checkout:

### 1. Active Subscription Prevention
When a user with a **truly active** subscription (not cancelled) purchases again:
- Purchase is blocked with error message
- Email notification sent to user
- No charge occurs

### 2. Pending Cancellation Within Billing Period
When a user with a **pending cancellation** subscription purchases again and is still within their paid billing period:
- Subscription is restarted (cancellation cleared)
- **No charge** - user already paid for this period
- Purchase marked as `not_charged`

### 3. Expired/Deactivated Subscription Restart
When a user with an **expired or deactivated** subscription purchases again:
- User is charged normally
- Subscription is reactivated via `resubscribe!`
- Updates the credit card on file

### 4. Tier Upgrade/Downgrade
When a user selects a different tier on restart:
- Old original purchase is archived
- New purchase becomes the subscription's original
- Payment option updated with new tier's price

### Matching Logic
1. First matches by user ID for logged-in users
2. Falls back to email matching for guest checkouts (SQL JOIN for performance)
3. Picks the most recently deactivated subscription if multiple exist

### Exclusions
A subscription is NOT restarted when:
- It was cancelled by the seller (intentional block)
- It has ended (`ended_at` is set)
- It's a test subscription
- The new purchase is a test purchase
- The product is deleted

---

# Before/After

---

# Test Results
image

---

# Checklist

---

# AI Disclosure

### DIFF
--- a/app/services/subscription/restart_at_checkout_service.rb
+++ b/app/services/subscription/restart_at_checkout_service.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+# Adapter service that transforms checkout params and delegates to UpdaterService
+# for restarting cancelled/failed subscriptions during checkout.
+#
+# This reuses the battle-tested UpdaterService logic for charging, tier changes,
+# payment method updates, proration, SCA, webhooks, etc.
+class Subscription::RestartAtCheckoutService
+  CARD_PARAM_KEYS = %i[
+    card_data_handling_mode stripe_payment_method_id stripe_customer_id
+    stripe_setup_intent_id stripe_error paypal_order_id paymentToken
+    billing_agreement_id braintree_device_data braintree_transient_customer_store_key
+    visual card_country card_country_source
+  ].freeze
+
+  attr_reader :subscription, :product, :params, :buyer
+
+  def initialize(subscription:, product:, params:, buyer: nil)
+    @subscription = subscription
+    @product = product
+    @params = params
+    @buyer = buyer
+  end
+
+  def perform
+    result = Subscription::UpdaterService.new(
+      subscription: subscription,
+      params: build_updater_params,
+      logged_in_user: buyer,
+      gumroad_guid: params.dig(:purchase, :browser_guid),
+      remote_ip: params[:remote_ip]
+    ).perform
+
+    adapt_result(result)
+  end
+
+  private
+    def build_updater_params
+      updater_params = {
+        variants: params[:variants] || default_variant_ids,
+        price_id: params[:price_id] || subscription.price&.external_id,
+        quantity: subscription.original_purchase.quantity,
+        perceived_price_cents: perceived_price,
+        perceived_upgrade_price_cents: perceived_price,
+      }
+
+      card_mode = CardParamsHelper.get_card_data_handling_mode(merged_card_params)
+      if card_mode.present? && card_mode != :reuse
+        updater_params.merge!(merged_card_params.slice(*CARD_PARAM_KEYS))
+      else
+        updater_params[:use_existing_card] = true
+      end
+
+      updater_params
+    end
+
+    def perceived_price
+      params.dig(:purchase, :perceived_price_cents)&.to_i ||
+        subscription.current_subscription_price_cents
+    end
+
+    def default_variant_ids
+      subscription.original_purchase.variant_attributes.map(&:external_id)
+    end
+
+    def merged_card_params
+      @merged_card_params ||= params.slice(*CARD_PARAM_KEYS)
+        .merge((params[:purchase] || {}).slice(*CARD_PARAM_KEYS))
+    end
+
+    def adapt_result(result)
+      if result[:success]
+        adapted = {
+          success: true,
+          restarted_subscription: true,
+          subscription: subscription,
+          message: result[:success_message] || "Your membership has been restarted!"
+        }
+
+        if result[:requires_card_action]
+          adapted[:requires_card_action] = true
+          adapted[:client_secret] = result[:client_secret]
+        end
+
+        adapted
+      else
+        { success: false, error_message: result[:error_message] }
+      end
+    end
+end
--- a/spec/services/subscription/restart_at_checkout_service_spec.rb
+++ b/spec/services/subscription/restart_at_checkout_service_spec.rb
@@ -1,0 +1,313 @@
+# frozen_string_literal: true
+
+describe Subscription::RestartAtCheckoutService do
+  let(:seller) { create(:user) }
+  let(:product) { create(:membership_product, user: seller) }
+  let(:buyer) { create(:user) }
+  let(:email) { buyer.email }
+  let(:browser_guid) { SecureRandom.uuid }
+
+  let(:base_params) do
+    {
+      purchase: {
+        email: email,
+        perceived_price_cents: product.price_cents,
+        browser_guid: browser_guid
+      },
+      price_id: product.prices.alive.first.external_id,
+      variants: product.tiers.map(&:external_id)
+    }
+  end
+
+  def create_subscription_for_product(product:, purchaser:, email:, **subscription_attrs)
+    subscription = create(:subscription, link: product, user: purchaser)
+    create(:purchase,
+           link: product,
+           purchaser: purchaser,
+           email: email,
+           subscription: subscription,
+           is_original_subscription_purchase: true,
+           price_cents: product.price_cents,
+           variant_attributes: product.tiers.to_a
+    )
+    subscription.update!(subscription_attrs) if subscription_attrs.present?
+    subscription
+  end
+
+  describe "#perform" do
+    let(:updater_service) { instance_double(Subscription::UpdaterService) }
+
+    before do
+      allow(Subscription::UpdaterService).to receive(:new).and_return(updater_service)
+    end
+
+    context "with a cancelled subscription that can be restarted" do
+      let!(:subscription) do
+        create_subscription_for_product(
+          product: product,
+          purchaser: buyer,
+          email: email,
+          cancelled_at: 1.day.ago,
+          cancelled_by_buyer: true,
+          deactivated_at: 1.day.ago
+        )
+      end
+
+      it "delegates to UpdaterService with correct params" do
+        allow(updater_service).to receive(:perform).and_return({ success: true, success_message: "Membership restarted" })
+
+        service = described_class.new(
+          subscription: subscription,
+          product: product,
+          params: base_params,
+          buyer: buyer
+        )
+
+        result = service.perform
+
+        expect(result[:success]).to be true
+        expect(result[:restarted_subscription]).to be true
+        expect(result[:message]).to eq("Membership restarted")
+
+        expect(Subscription::UpdaterService).to have_received(:new).with(
+          subscription: subscription,
+          params: hash_including(
+            variants: product.tiers.map(&:external_id),
+            price_id: product.prices.alive.first.external_id,
+            quantity: 1,
+            perceived_price_cents: product.price_cents,
+            perceived_upgrade_price_cents: product.price_cents,
+            use_existing_card: true
+          ),
+          logged_in_user: buyer,
+          gumroad_guid: browser_guid,
+          remote_ip: nil
+        )
+      end
+    end
+
+    context "with a failed subscription" do
+      let!(:subscription) do
+        create_subscription_for_product(
+          product: product,
+          purchaser: buyer,
+          email: email,
+          failed_at: 1.day.ago,
+          deactivated_at: 1.day.ago
+        )
+      end
+
+      it "delegates to UpdaterService and returns success" do
+        allow(updater_service).to receive(:perform).and_return({ success: true, success_message: "Membership restarted" })
+
+        service = described_class.new(
+          subscription: subscription,
+          product: product,
+          params: base_params,
+          buyer: buyer
+        )
+
+        result = service.perform
+
+        expect(result[:success]).to be true
+        expect(result[:restarted_subscription]).to be true
+      end
+    end
+
+    context "when UpdaterService returns an error" do
+      let!(:subscription) do
+        create_subscription_for_product(
+          product: product,
+          purchaser: buyer,
+          email: email,
+          cancelled_at: 1.day.ago,
+          cancelled_by_buyer: true,
+          deactivated_at: 1.day.ago
+        )
+      end
+
+      it "passes through the error" do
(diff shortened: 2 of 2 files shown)


app/services/purchase/create_service.rb
```diff
@@ -133,6 +133,9 @@ def perform
       # Make sure the giftee purchase is created successfully before attempting a charge
       create_giftee_purchase if purchase.is_gift_sender_purchase
 
+      # Check for existing subscriptions and handle accordingly
+      validate_and_handle_existing_subscription
+
```
yashranaway: Called `validate_and_handle_existing_subscription` right before `purchase.charge!`. This is critical - if we checked after charging, users would already be billed. The validation must happen before any payment processing.

app/services/purchase/create_service.rb
```diff
@@ -377,6 +380,63 @@ def parse_url_parameters(url_parameters_string)
...
+        CustomerLowPriorityMailer.existing_active_subscription_notice(active_subscription.id).deliver_later(queue: "low")
+        raise Purchase::PurchaseInvalid, "You already have an active subscription to this membership. Visit your Library to manage your subscription."
+      end
+
+      pending_cancel_subscription = find_pending_cancellation_subscription(user, email)
+      if pending_cancel_subscription && !pending_cancel_subscription.overdue_for_charge?
+        purchase.is_subscription_restart_without_charge = true
+        purchase.existing_subscription_for_restart = pending_cancel_subscription
+      end
+    end
```
yashranaway, line 402: Three early returns at the top: skip for non-memberships, gifts, and installment plans. These cases shouldn't trigger the existing subscription check.

app/services/purchase/create_service.rb
```diff
@@ -377,6 +380,63 @@ def parse_url_parameters(url_parameters_string)
...
+      return subscription if subscription
+
+      return nil if email.blank?
+
+      original_purchase_flag = Purchase.flag_mapping["flags"][:is_original_subscription_purchase]
+      base_scope
+        .joins("INNER JOIN purchases ON purchases.subscription_id = subscriptions.id AND (purchases.flags & #{original_purchase_flag}) != 0")
+        .where("LOWER(purchases.email) = ?", email.downcase)
+        .first
+    end
```
yashranaway, line 420: `cancelled_at: nil` is the key difference from `find_active_subscription` in base_service. We only want subscriptions that are completely active - not ones with pending cancellation.

app/services/purchase/create_service.rb
```diff
@@ -377,6 +380,63 @@ def parse_url_parameters(url_parameters_string)
...
+      subscription = base_scope.find_by(user: user) if user.present?
+      return subscription if subscription
+
+      return nil if email.blank?
+
+      original_purchase_flag = Purchase.flag_mapping["flags"][:is_original_subscription_purchase]
+      base_scope
+        .joins("INNER JOIN purchases ON purchases.subscription_id = subscriptions.id AND (purchases.flags & #{original_purchase_flag}) != 0")
+        .where("LOWER(purchases.email) = ?", email.downcase)
+        .first
```
yashranaway, line 419: For guest checkouts, we JOIN on purchases to match by email. Using `LOWER()` for case-insensitive matching. This handles the case where someone purchases with different email casing.

app/services/purchase/base_service.rb
```diff
@@ -30,6 +30,23 @@ def handle_purchase_success
     def create_subscription(giftee_purchase)
       return if purchase.subscription.present?
 
+      if purchase.existing_subscription_for_restart
+        restart_subscription_without_charge(purchase.existing_subscription_for_restart, giftee_purchase)
+        return
+      end
```
yashranaway, line 36: This check comes FIRST in `create_subscription`. If CreateService already detected a pending cancellation within billing period, we use that subscription directly instead of searching again.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +84,127 @@ def create_subscription(giftee_purchase)
...
+      ActiveRecord::Base.transaction do
+        subscription.credit_card = purchase.credit_card unless is_gift
+        handle_tier_change_on_restart(subscription)
+        subscription.cancelled_at = nil
+        subscription.user_requested_cancellation_at = nil
+        subscription.cancelled_by_buyer = false
+        subscription.save!
+        subscription.purchases << [purchase, giftee_purchase].compact
+      end
+    end
```
yashranaway, line 140: Similar to `restart_existing_subscription` but doesn't call `resubscribe!`. Instead, we manually clear the cancellation flags. The subscription was never deactivated, just scheduled for cancellation.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +84,127 @@ def create_subscription(giftee_purchase)
...
+      return unless tier_changed || price_changed
+
+      old_original.update!(is_archived_original_subscription_purchase: true)
+      purchase.is_original_subscription_purchase = true
+
+      if purchase.price.present?
+        payment_option = subscription.last_payment_option
+        payment_option.update!(price: purchase.price) if payment_option
+      end
+    end
```
yashranaway, line 163: Compares variant IDs between old and new purchase. If different, we archive the old original_purchase and make the new one the original. This ensures future charges use the new tier's price.

app/services/purchase/base_service.rb
```diff
@@ -67,6 +84,127 @@ def create_subscription(giftee_purchase)
...
+
+      old_variant_ids = old_original.variant_attributes.map(&:id).sort
+      new_variant_ids = purchase.variant_attributes.map(&:id).sort
+
+      tier_changed = old_variant_ids != new_variant_ids
+      price_changed = old_original.price_id != purchase.price_id
+
+      return unless tier_changed || price_changed
+
+      old_original.update!(is_archived_original_subscription_purchase: true)
```
yashranaway, line 156: Setting `is_archived_original_subscription_purchase: true` is how Gumroad tracks which purchase is the "current" original. The subscription's `original_purchase` association filters out archived ones.

app/models/purchase.rb
```diff
@@ -425,7 +425,7 @@ class Purchase < ApplicationRecord
                 :save_shipping_address, :flow_of_funds, :prorated_discount_price_cents,
                 :original_variant_attributes, :original_price, :is_updated_original_subscription_purchase,
                 :is_applying_plan_change, :setup_intent, :charge_intent, :setup_future_charges, :skip_preparing_for_charge,
-                :installment_plan
+                :installment_plan, :is_subscription_restart_without_charge, :existing_subscription_for_restart
```
yashranaway, line 428: Added `is_subscription_restart_without_charge` and `existing_subscription_for_restart`. These are transient (not persisted) - only used during the purchase flow to pass state between CreateService and BaseService.

app/models/purchase.rb
```diff
@@ -2734,7 +2734,7 @@ def process_without_charging!
         validate_purchasing_power_parity
         return if errors.present?
 
-        if is_preorder_authorization || is_free_trial_purchase?
+        if is_preorder_authorization || is_free_trial_purchase? || is_subscription_restart_without_charge
```
yashranaway, line 2737: Added check for `is_subscription_restart_without_charge` alongside `is_free_trial_purchase`. Both cases should skip the actual Stripe charge.

app/models/purchase.rb
```diff
@@ -1703,7 +1703,7 @@ def update_balance_and_mark_successful!
     if is_test_purchase?
       set_succeeded_at
       mark_test_successful!
-    elsif is_free_trial_purchase?
+    elsif is_free_trial_purchase? || is_subscription_restart_without_charge
```
yashranaway, line 1706: When `is_subscription_restart_without_charge` is true, we call `mark_not_charged!` instead of `mark_successful!`. This creates the purchase record without incrementing seller balance.

app/models/subscription.rb
```diff
@@ -681,7 +681,7 @@ def resubscribe!
       save!
       original_purchase&.add_to_audience_member_details
 
-      if is_deactivated
+      if is_deactivated && original_purchase.present?
```
yashranaway, line 684: Added `&& original_purchase.present?` to prevent NoMethodError when `original_purchase` is nil. This can happen in test scenarios where subscriptions are created without proper purchase associations.

app/mailers/customer_low_priority_mailer.rb
yashranaway, line 1: Simple mailer that loads the subscription and sets the subject. The `deliver_subscription_email` callback handles the rest (builds manage URL, sends to subscriber).

app/mailers/customer_low_priority_mailer.rb
```diff
@@ -12,7 +12,7 @@ class CustomerLowPriorityMailer < ApplicationMailer
                                                      subscription_charge_failed subscription_product_deleted subscription_renewal_reminder
                                                      subscription_price_change_notification subscription_ended free_trial_expiring_soon
                                                      credit_card_expiring_membership subscription_early_fraud_warning_notification
-                                                     subscription_giftee_added_card]
+                                                     subscription_giftee_added_card existing_active_subscription_notice]
```
yashranaway, line 15: Added `existing_active_subscription_notice` to the `after_action :deliver_subscription_email` list so it uses the same email delivery pattern as other subscription emails.

yashranaway:
@EmCousin Could you check I have implemented fixes from last PR.

yashranaway:
I'll add a fix for the failing tests

app/services/purchase/create_service.rb
```diff
@@ -126,15 +126,15 @@ def perform
 
       validate_bundle_products
 
+      validate_and_handle_existing_subscription
+
```
yashranaway, line 130: Moved `validate_and_handle_existing_subscription` to run before `prepare_for_charge!`. Previously it ran after, which meant Stripe API calls were already happening before we could set the "skip charge" flags.

app/services/purchase/create_service.rb
```diff
@@ -405,7 +405,7 @@ def find_truly_active_subscription(user, email)
       base_scope = product.subscriptions
         .not_is_test_subscription
         .where(deactivated_at: nil, ended_at: nil, failed_at: nil)
-        .where(cancelled_at: nil)
+        .where("cancelled_at IS NULL OR (cancelled_at > ? AND (subscriptions.flags & ?) = 0)", Time.current, Subscription.flag_mapping["flags"][:cancelled_by_buyer])
```
yashranaway, line 408: `find_truly_active_subscription` now also catches seller-cancelled pending subscriptions. These users still have access until their billing period ends, so they shouldn't be able to repurchase-  same as truly active subs.

app/services/purchase/create_service.rb
```diff
@@ -422,6 +422,7 @@ def find_truly_active_subscription(user, email)
     def find_pending_cancellation_subscription(user, email)
       base_scope = product.subscriptions
         .not_is_test_subscription
+        .cancelled_by_buyer
```
yashranaway, line 425: Added `.cancelled_by_buyer` scope to `find_pending_cancellation_subscription`. We only want to auto-restart for buyer-cancelled subs. If a seller cancelled someone, they probably had a reason  don't let the buyer just undo that.

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+# Adapter service that transforms checkout params and delegates to UpdaterService
+# for restarting cancelled/failed subscriptions during checkout.
+#
+# This reuses the battle-tested UpdaterService logic for charging, tier changes,
+# payment method updates, proration, SCA, webhooks, etc.
+class Subscription::RestartAtCheckoutService
```
EmCousin, line 8: This is not used - dead code.

app/services/purchase/create_service.rb
```diff
@@ -377,6 +379,64 @@ def parse_url_parameters(url_parameters_string)
...
+      return if purchase.is_gift_sender_purchase
+      return if purchase.is_installment_payment
+
+      email = purchase.email
+      user = purchase.purchaser
+
+      active_subscription = find_truly_active_subscription(user, email)
+      if active_subscription
+        CustomerLowPriorityMailer.existing_active_subscription_notice(active_subscription.id).deliver_later(queue: "low")
+        raise Purchase::PurchaseInvalid, "You already have an active subscription to this membership. Visit your Library to manage your subscription."
```
EmCousin, line 394: This leaks that the email has a subscription for logged-out users. It's fine for logged-in users but we don't want to risk anonymous visitors brute-forcing the checkout form to gather information about user emails. This needs to be a generic error message for logged-out users, whilst an alert email is sent to the email that was filled.

app/services/purchase/create_service.rb
EmCousin: The `UpdaterService` should handle this as it already knows not to charge if the subscription is still within its billing period.

app/services/purchase/create_service.rb
EmCousin: We need to protect this method against race conditions. Concurrent checkouts could both find the same subscription as restartable. A pessimistic lock would do.

app/services/purchase/base_service.rb
EmCousin: That makes the `create_subscriptiopn` name misleading since we no longer always create a subscription. It would be clearer to have this check before calling this method, and instead immediately delegate to the UpdaterService.

EmCousin (commented):
Thanks for your efforts @yashranaway, this issue is quite sensitive as it touches a critical flow of the app and there are still a lot of cases this PR does not cover.
- SCA/3DS iframe flow for restarts that require payments needs to be handled
- Content gating: restarting a subscription should only deliver to the buyer the content that is due, not automatically the entire subscription's content.
- Race condition protection to prevent concurrent checkouts from finding the same restartable subscription at the same time
- Privacy-safe error messages for logged out users

We also need thorough end-to-end tests (including Capybara with actual Stripe SCA flow) to cover this.

Given all this, we think it's best to handle this internally.

PR #117, opened 2025-04-15
Exclude private products from recommendations

Hanaffi:
Resolves https://github.com/antiwork/gumroad/issues/105

Exclude private products from product recommendations in checkout and elsewhere.

- [x] Update services and concern
- [x] Cover changes with specs

PR #3254, opened 2026-01-30
Transparent subscription restarts

EmCousin:
> [!NOTE]  
> Most of the LOC changes (Screenshot 2026-02-10 at 21 31 38) are updated VCR cassettes.

---

Issue: https://github.com/antiwork/gumroad/issues/173
Closes: https://github.com/antiwork/gumroad-private/issues/117
Closes: https://github.com/antiwork/gumroad-private/issues/112

# Description

This PR adds prevention mechanisms to cover purchases of a membership for which a subscription might already exist.

## Problem

  1. **Duplicate subscriptions**: When a buyer whose subscription was cancelled/expired tried to
  purchase the same membership again, a new subscription was created. This caused confusion,
  duplicate charges, and broke workflow delivery since the new subscription started from scratch.

  2. **Content leakage on restart**: When restarting a subscription via the existing
  manage-subscription flow, `grant_access_to_product?` returns true immediately, giving access to
  ALL published membership content — regardless of how many payments were actually made. Workflow
  installments with delayed delivery schedules were visible right away.

### Auto-restart at checkout
Depending on the scenario:
- If the user never had a subscription before, the flow proceeds as usual
- If the user already had an active subscription and is signed in, an error message informs them about their ongoing subscription. If the user is signed out or another buyer uses the seller's email, a generic error message is displayed instead for privacy concerns
- If the user already had a subscription but it expired, the subscription restarts. The user is charged unless the subscription was still in its billing period
- If the membership tier got upgraded/downgraded when restarting a subscription, the user is charged accordingly

### Content gating for restarted subscriptions
- `Installment#workflow_delivery_due?` checks whether a workflow post's `expected_delivery_time`
   has passed for the given purchase
- Only affects restarted subscriptions viewing "show all posts" content — already-emailed
  installments remain visible
- Non-workflow installments and non-restarted subscriptions are unaffected

---

# Before/After

### Before
Before: it was automatically creating a new membership even if a membership already existed

### After

When the user is signed in and already has an active subscription, the user is alerted

https://github.com/user-attachments/assets/969fec6b-77df-4b4e-907d-76f31f6a28d5

When the user is signed out, or another visitor uses the user's email, we have a more generic error message

https://github.com/user-attachments/assets/33ae0d53-7370-4b5f-b4a1-44618e5b98ba

---

# Test Results

CI

---

# Checklist

---

# AI Disclosure

Claude Code + Claude Opus was used to lay the bases of the solution, which substantial manual modification made by me, then reviewed by GPT Codex 5.2

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -2122,7 +2122,7 @@ class Purchase < ApplicationRecord
           next false unless post.targeted_at_purchased_item?(purchase)
           next false unless post.passes_member_cancellation_checks?(purchase)
 
-          post.workflow_delivery_due?(purchase)
+          post.delivery_due?(purchase)
         end
       end
     end


app/controllers/orders_controller.rb
```diff
@@ -95,8 +95,8 @@ def valid_wallet_payment?
       return false if [params[:wallet_type], params[:stripe_payment_method_id]].any?(&:blank?)
       payment_method = Stripe::PaymentMethod.retrieve(params[:stripe_payment_method_id])
       payment_method&.card&.wallet&.type == params[:wallet_type]
-    rescue Stripe::StripeError
-      render_error("Sorry, something went wrong.")
+    rescue Stripe::StripeError => e
+      render_error("Sorry, something went wrong. #{e.message}")
```
EmCousin: Forgot to revert this. I'll remove this after peer review.
```suggestion
    rescue Stripe::StripeError
      render_error("Sorry, something went wrong")
```

app/mailers/customer_low_priority_mailer.rb
```diff
@@ -389,6 +389,11 @@ def wishlist_updated(wishlist_follower_id, product_count)
     )
   end
 
+  def already_subscribed_checkout_attempt(subscription_id)
+    @subscription = Subscription.find(subscription_id)
+    @subject = "You're already subscribed to #{@subscription.link.name}"
+  end
```
EmCousin, line 401: The email that is sent to the already existing subscription of a given buyer, when an unauthenticated buyer tries to purchase a membership with the same email, so we alert the user without revealing the subscription exists to the anonymous visitor if they are not the same person.

app/models/subscription.rb
```diff
@@ -859,6 +859,44 @@ def alive_or_restartable?
...
+      query = query.where(user_id: user_or_email.id)
+    else
+      query = query.joins(:original_purchase)
+                   .where(purchases: { email: user_or_email.to_s.downcase.strip })
+    end
+
+    query.where.not(deactivated_at: nil)
+      .not_cancelled_by_admin
+      .order(created_at: :desc)
+      .first
```
EmCousin: Subscriptions are considered restartable under the following conditions:
- Subscription has not ended
- Subscription has not been cancelled by admin

app/models/subscription.rb
```diff
@@ -859,6 +859,44 @@ def alive_or_restartable?
...
+    if user_or_email.is_a?(User)
+      query = query.where(user_id: user_or_email.id)
+    else
+      query = query.joins(:original_purchase)
+                   .where(purchases: { email: user_or_email.to_s.downcase.strip })
+    end
+
+    query = query.lock if with_lock
+
+    query.first
```
EmCousin: Subscription for a given user and product is considered active under the following conditoins:
- Subscription has not ended
- Subscription has not failed
- Subscription has not been cancelled or is scheduled to be cancelled only in the future

I added a pessimistic lock to avoid race conditions in case multiple subscriptions are being created for the same email and product.

app/services/order/create_service.rb
```diff
@@ -40,6 +40,14 @@ def perform
         next
       end
 
+      if product.is_recurring_billing && !line_item_params[:is_gift]
+        existing_subscription_error = check_for_existing_active_subscription(product, common_params[:email])
+        if existing_subscription_error
+          purchase_responses[line_item_uid] = error_response(existing_subscription_error)
+          next
+        end
+      end
```
EmCousin: The key change is here. We do _not_ let the buyer purchase a membership if they already have an active subscription on it.

app/services/order/create_service.rb
```diff
@@ -68,8 +76,15 @@ def perform
         end
 
         if purchase&.persisted?
-          order.purchases << purchase
-          order.save!
+          # For subscription restarts without a new charge, the original purchase is returned
+          # which is already successful. Add its response directly since ChargeService
+          # only processes in_progress purchases.
+          if purchase.successful?
```
EmCousin: For subscription restarts without a new charge, the original purchase is returned which is already successful, so we add its response directly since `ChargeService` only processes in_progress purchases.

app/services/order/create_service.rb
```diff
@@ -68,8 +76,15 @@ def perform
         end
 
         if purchase&.persisted?
-          order.purchases << purchase
-          order.save!
+          # For subscription restarts without a new charge, the original purchase is returned
+          # which is already successful. Add its response directly since ChargeService
+          # only processes in_progress purchases.
```
EmCousin:
```suggestion
```

app/services/order/create_service.rb
```diff
@@ -162,4 +177,36 @@ def parse_url_parameters_from_referrer(product)
...
       CGI.parse(URI.parse(params[:referrer]).query).transform_values { |values| values.length <= 1 ? values.first : values }.to_json
     end
+
+    def check_for_existing_active_subscription(product, email)
+      user_or_email = buyer || email
+      return nil unless user_or_email.present?
+
+      # Use pessimistic locking to prevent race conditions where multiple concurrent
+      # checkout attempts could all pass the active subscription check before any
+      # creates a new subscription. Lock the subscription record during the check.
```
EmCousin:
```suggestion
```

app/services/order/create_service.rb
```diff
@@ -162,4 +177,36 @@ def parse_url_parameters_from_referrer(product)
...
+      active_subscription = Subscription.active_for_user_and_product(
```
EmCousin: Pessimistic locking prevents race conditions where multiple concurrent checkout attempts could all pass the active subscription check before any creates a new subscription. We avoid this problem by locking the subscription record during the check.

app/services/order/create_service.rb
```diff
@@ -162,4 +177,36 @@ def parse_url_parameters_from_referrer(product)
...
+      )
+
+      return nil unless active_subscription.present?
+
+      CustomerLowPriorityMailer.already_subscribed_checkout_attempt(active_subscription.id).deliver_later(queue: "low")
+
+      if buyer.present?
+        "You already have an active subscription to this membership. Visit your Library to manage it."
+      else
+        Bugsnag.notify(StandardError.new("Existing subscription checkout attempt")) do |report|
```
EmCousin: In case signed out buyers reach out to support, we want to be able to investigate whether we are in this scenario or if it is an actual uncaught error.

app/services/purchase/create_service.rb
```diff
@@ -377,6 +382,69 @@ def parse_url_parameters(url_parameters_string)
         end
       end
     end
+
+    # Check if we should look for restartable subscriptions
+    # Only applies to membership products that are not gifts
```
EmCousin:
```suggestion
```

app/services/purchase/create_service.rb
```diff
@@ -377,6 +382,69 @@ def parse_url_parameters(url_parameters_string)
...
+    def should_check_for_restartable_subscription?
+      product.is_recurring_billing && !is_gift?
+    end
+
+    def handle_existing_subscription
+      user_or_email = buyer || purchase_params[:email]
+      return nil unless user_or_email.present?
+
+      # First check for active subscriptions - block purchase if one exists
+      # This check is also in Order::CreateService but we need it here for direct API calls
```
EmCousin:
```suggestion
```

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,181 @@
+# frozen_string_literal: true
+
+class Subscription::RestartAtCheckoutService
+  include CurrencyHelper
+
+  # Custom exception for charge failures that should trigger transaction rollback
```
EmCousin:
```suggestion
```

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,181 @@
...
+        charge_result = charge_subscription
+        self.new_purchase = charge_result[:purchase]
+      end
+
+      subscription.send_restart_notifications!
+      success_result
+    end
+  rescue ChargeFailed => e
+    # Charge failed after transaction rolled back - now mark subscription as failed
+    # This happens outside the transaction to avoid committing partial state changes
```
EmCousin:
```suggestion
```

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,181 @@
...
+      end
+
+      subscription.credit_card = credit_card
+      subscription.save!
+    end
+
+    def should_charge?
+      return false if within_billing_period?
+      return false if subscription.in_free_trial?
+      true
```
EmCousin:
```suggestion
     !within_billing_period? && !subscription.in_free_trial?
```

EmCousin:
@ershad May I ask you to review this PR? It should be addressing every case discussed in the issue.

EmCousin:
@hchhabra _a may I ask you to review this PR?

app/services/order/create_service.rb
hchhabra: This seems redundant as we are doing the same check at the start of `Purchase::CreateService#handle_existing_subscription`? Do we need this here too?

app/controllers/orders_controller.rb
```diff
@@ -96,7 +96,7 @@ def valid_wallet_payment?
       payment_method = Stripe::PaymentMethod.retrieve(params[:stripe_payment_method_id])
       payment_method&.card&.wallet&.type == params[:wallet_type]
     rescue Stripe::StripeError
-      render_error("Sorry, something went wrong.")
+      render_error("Sorry, something went wrong")
```
hchhabra:
```suggestion
      render_error("Sorry, something went wrong.")
```

Looks like an unrelated/unintentional change?

app/services/purchase/create_service.rb
```diff
@@ -377,6 +382,65 @@ def parse_url_parameters(url_parameters_string)
...
+      user_or_email = buyer || purchase_params[:email]
+      return nil unless user_or_email.present?
+
+      active_subscription = Subscription.active_for_user_and_product(
+        user_or_email: user_or_email,
+        product: product,
+        with_lock: true
+      )
+
+      if active_subscription.present?
```
hchhabra:
```suggestion
      return nil if buyer.blank? && purchase_params[:email].blank?

      active_subscription = buyer.present? ? Subscription.active_for_product_and_buyer(
        buyer:,
        product:,
        with_lock: true
      ) : Subscription.active_for_product_and_email(
        email: purchase_params[:email],
        product:,
        with_lock: true
      )

      if active_subscription.present?
```

Nit: Just trying to avoid the common `user_or_email` variable as it seems a bit weird to me. Also, do we have a use case for `with_lock: false`, if not we can just make it default and remove the variable?

app/services/purchase/create_service.rb
```diff
@@ -377,6 +382,65 @@ def parse_url_parameters(url_parameters_string)
...
+      result = Subscription::RestartAtCheckoutService.new(
+        subscription: restartable_subscription,
+        product: product,
+        params: params,
+        buyer: buyer
+      ).perform
+
+      if result[:success]
+        purchase = result[:purchase] || restartable_subscription.original_purchase
+        self.purchase = purchase
```
hchhabra:
```suggestion
        self.purchase = result[:purchase] || restartable_subscription.original_purchase
```
Does this work?

app/models/subscription.rb
```diff
@@ -859,6 +859,44 @@ def alive_or_restartable?
...
+    query = where(link_id: product.id).where(ended_at: nil)
+
+    if user_or_email.is_a?(User)
+      query = query.where(user_id: user_or_email.id)
+    else
+      query = query.joins(:original_purchase)
+                   .where(purchases: { email: user_or_email.to_s.downcase.strip })
+    end
+
+    query.where.not(deactivated_at: nil)
```
hchhabra:
```suggestion
    subscriptions = where(link_id: product.id).where(ended_at: nil)

    if user_or_email.is_a?(User)
      subscriptions = subscriptions.where(user_id: user_or_email.id)
    else
      subscriptions = subscriptions.joins(:original_purchase)
                   .where(purchases: { email: user_or_email.to_s.downcase.strip })
    end

    subscriptions.where.not(deactivated_at: nil)
```

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,176 @@
...
+    @buyer = buyer
+    @new_purchase = nil
+  end
+
+  def perform
+    error = validate_restart
+    return error_result(error) if error.present?
+
+    ActiveRecord::Base.transaction do
+      handle_tier_change if tier_changed?
```
hchhabra: Is it possible to use the `Subscription::UpdaterService` to restart/resubscribe here (maybe update that service as required)? A lot of code should be common in the two I believe as retstarting during checkout should be pretty similar to resubscribing from the manage membership page? That service already handles all use cases, like a change in recurrence (monthly/yearly..) is not handled here it seems?

spec/services/subscription/restart_at_checkout_service_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+            params: base_params,
+            buyer: buyer
+          ).perform
+
+          expect(result[:success]).to be false
+          expect(result[:error_message]).to eq("This subscription cannot be restarted.")
+        end
+      end
+    end
+  end
```
gianfrancopiana, line 300: NIT: Maybe we can add a test here for restarting with a recurrence change (e.g. monthly→yearly, same tier) – that's the scenario from #117. We had one written in #3368 that could be reused.

gianfrancopiana (approved):
I think this looks good! The changes also correctly handle the bug from #117/#112 (restarting a tiered membership with a different billing recurrence). Left one small suggestion about test coverage.

spec/services/subscription/restart_at_checkout_service_spec.rb
EmCousin: Absolutely, will do. I'll also let @hchhabra have the final word when it's ready!

app/models/installment.rb
```diff
@@ -829,6 +829,14 @@ def tags
     tags.map { normalize_tag(it) }.uniq
   end
 
+  def workflow_delivery_due?(purchase)
```
EmCousin: This method gates workflow content visibility for restarted subscriptions. Without it, `resubscribe!` makes `grant_access_to_product?` return true, which causes `should_show_all_posts?` to show ALL published workflow content immediately — regardless of how many payments were made.

The `expected_delivery_time` calculation already accounts for deactivation gaps (used in email scheduling), so we reuse it here for content display gating. Non-workflow installments and non-restarted subscriptions always return `true` (no behavior change for existing users).

app/models/purchase.rb
```diff
@@ -2115,7 +2115,14 @@ def self.product_installments(purchase_ids:)
...
         purchases.reduce(false) do |select_post, purchase|
-          select_post || (purchase.link.should_show_all_posts? && post.purchase_passes_filters(purchase) && post.targeted_at_purchased_item?(purchase) && post.passes_member_cancellation_checks?(purchase))
+          next true if select_post
+
+          next false unless purchase.link.should_show_all_posts?
+          next false unless post.purchase_passes_filters(purchase)
+          next false unless post.targeted_at_purchased_item?(purchase)
+          next false unless post.passes_member_cancellation_checks?(purchase)
+
+          post.workflow_delivery_due?(purchase)
```
EmCousin: Added `post.workflow_delivery_due?(purchase)` to the filter chain. This only affects the `past_posts_to_share` path (the "show all posts" branch). Posts in `installments_with_sent_emails` (already emailed to the customer) are unaffected — they were already delivered and should remain visible.

app/services/purchase/create_service.rb
```diff
@@ -171,6 +178,65 @@ def is_gift?
       !!params[:is_gift]
     end
 
+    def should_check_for_restartable_subscription?
+      product.is_recurring_billing && !is_gift?
+    end
+
```
EmCousin, line 184: This runs before the normal purchase flow for recurring billing products. It checks active subscriptions first (returns an error and sends a notification email to the subscriber), then checks for restartable subscriptions (auto-restarts via `UpdaterService`).

Gift purchases skip this check entirely since they create a separate subscription for the recipient. For logged-out users, we match by email and return a generic error for active subscriptions to avoid leaking subscription info.

app/services/purchase/create_service.rb
```diff
@@ -77,6 +77,13 @@ def perform
           product.default_price
 
         purchase.price = price || product.default_price
+
+        # Check for existing subscriptions (active or restartable)
+        if should_check_for_restartable_subscription?
+          existing_purchase, error, sca_response = handle_existing_subscription
+          return nil, nil, sca_response if sca_response.present?
```
EmCousin, line 84: When `UpdaterService` returns `requires_card_action: true`, the `:purchase` value is a Hash (not a Purchase object), which would crash `purchase&.persisted?` downstream. We pass this through as a third return value so `Order::CreateService` can put it directly into `purchase_responses` for the frontend to trigger 3D Secure verification.

app/services/order/create_service.rb
```diff
@@ -68,18 +73,29 @@ def perform
         end
 
         if purchase&.persisted?
-          order.purchases << purchase
-          order.save!
+          if Purchase::ALL_SUCCESS_STATES.include?(purchase.purchase_state)
```
EmCousin, line 95: When a subscription is restarted, the original purchase is already in a success state. We avoid adding it to the new order (via `OrderPurchase` join table) since it already belongs to an existing order and would otherwise end up corrupting data.

We check the purchase is in a success state for this; with `is_part_of_combined_charge: true` skipping charging in `Purchase::CreateService`, only subscription restart returns an already-successful purchase here. Free products, preorders, etc. stay in `in_progress` during `CreateService`.

app/services/order/create_service.rb
```diff
@@ -68,18 +73,29 @@ def perform
...
           end
         end
       end
     end
 
-    if order.persisted? && (cart = Cart.fetch_by(user: buyer, browser_guid: params[:browser_guid]))
-      cart.order = order
-      cart.mark_deleted!
+    if (cart = Cart.fetch_by(user: buyer, browser_guid: params[:browser_guid]))
+      all_items_handled = purchase_responses.any? && purchase_responses.values.all? { _1[:success] }
```
EmCousin, line 110: We need to cover three cases:
1. If the order persisted with purchase, normal cleanup (associate cart, delete it)
2. If all items handled successfully but order not persisted (e.g., all items are subscription restarts), simply delete the cart.
3. The last case is If some items fail, then we preserve the cart so users see the error, can fix their order and retry.

app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -0,0 +1,72 @@
...
+    def default_variant_ids
+      subscription.original_purchase.variant_attributes.map(&:external_id)
+    end
+
+    def use_existing_card?
+      card_data_handling_mode = CardParamsHelper.get_card_data_handling_mode(params)
+      card_data_handling_mode.blank? || card_data_handling_mode == :reuse
+    end
+
+    def adapt_result(result)
```
EmCousin, line 59: Thin adapter that transforms checkout params to `UpdaterService` format. The `use_existing_card?` logic determines whether to reuse the subscription's saved card or pass through a new payment method from checkout. `adapt_result` passes through SCA data (`requires_card_action`, `client_secret`, `purchase`) when the restart requires 3D Secure.

spec/requests/checkout/subscription_restart_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
```
EmCousin, line 1: System spec for the end-to-end checkout restart flow. Note: this spec requires a clean test database. Seed data products can interfere with factory-created products on the checkout page. It tests both single-item restart and mixed cart (restart + regular product).

EmCousin:
@claude review.

app/models/subscription.rb
```diff
@@ -863,6 +863,56 @@ def alive_or_restartable?
...
+
+    where(link_id: product.id)
+      .where(ended_at: nil)
+      .where(failed_at: nil)
+      .where("cancelled_at IS NULL OR cancelled_at > ?", Time.current)
+      .joins(:original_purchase)
+      .where(purchases: { email: email.to_s.downcase.strip })
+      .lock
+      .first
+  end
```
EmCousin: Decoupling the `user_or_email` behaviour.

app/services/order/create_service.rb
```diff
@@ -53,12 +53,17 @@ def perform
...
+        purchase, error, sca_response = Purchase::CreateService.new(
           product:,
           params: purchase_params.merge(is_part_of_combined_charge: true),
           buyer:
         ).perform
 
+        if sca_response
+          purchase_responses[line_item_uid] = sca_response
+          next
+        end
```
EmCousin, line 84: When a subscription restart requires 3D Secure authentication (SCA), pass the requires_card_action, client_secret, and purchase data present in `sca_response ` through `Purchase::CreateService` and `Order::CreateService` so the frontend can prompt the user for 3D Secure verification instead of showing an error.

EmCousin:
@claude review

EmCousin:
@hchhabra _a sorry for taking this long to address your feedback, I believe the PR should now cover every edge case and I took the time to throughly cover it with tests. Here's a summary of the updates:
  1. Re-use `Subscription::UpdaterService` for restart. `RestartAtCheckoutService` is now a
  thin adapter that delegates to `UpdaterService`, so we get all the existing logic for recurrence
  changes, tier changes, etc.
  2. Removed redundant check in `Order::CreateService`
  4. Refactored the `user_or_email` code smell,  added lock by default for active subscription finder methods, and organized them in a tested `Subscription::Restartable` module
  5. Also, applied `self.purchase = result[:purchase] || restartable_subscription.original_purchase` 👍 

  I've also updated the PR description to reflect the full scope of changes since your review
  notably:
  - Content gating for restarted subscriptions: so restarted subscribers only see workflow posts that are actually due, not the entire back-catalog (see [this internal thread](https://chat.google.com/room/AAQAb0PAzos/BW6Uhg7cWfE/BW6Uhg7cWfE?cls=10))
  - Added SCA/3DS support to the restart flow when 3D Secure authentication is required
  - Excluded purchases with existing payment intents from SCA restarts are excluded from double-charging

May I ask you to review this again in light of those requirements?

> [!NOTE]  
> Most of the LOC changes (Screenshot 2026-02-10 at 21 31 38) are updated VCR cassettes.

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -84,13 +85,15 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
...
+    const { scheme, appDomain } = useDomains();
+
     const { selectedOption, pppDiscounted, discountedPriceCents } = applySelection(
       product,
       discountCode?.valid ? discountCode.discount : null,
       selection,
     );
 
-    const url = new URL(Routes.checkout_index_url());
+    const url = new URL(Routes.checkout_index_url({ protocol: scheme, host: appDomain }));
```
EmCousin: This fixed a system test that was failing locally. Let me know if you think we should revert this.

app/services/order/charge_service.rb
```diff
@@ -19,7 +19,9 @@ def perform
     # All remaining purchases need to be charged that are still in progress
     # Create a combined charge for all purchases belonging to the same seller
     # i.e. one charge per seller
-    purchases_by_seller = order.purchases.group_by(&:seller_id)
+    # Exclude purchases that already have a payment intent (e.g. subscription restarts
+    # requiring SCA — they are confirmed later via Order::ConfirmService)
```
EmCousin, line 23: I do not usually leave comments but given how sensitive this part of the code is, I figured it was worth keeping them.

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -84,6 +85,8 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
     const [referrer, setReferrer] = React.useState("");
     useRunOnce(() => setReferrer(document.referrer));
 
+    const { scheme, appDomain } = useDomains();
+
```
EmCousin:
```suggestion
```

app/javascript/components/Product/CtaButton.tsx
```diff
@@ -7,6 +7,7 @@ import { assertResponseError } from "$app/utils/request";
 import { trackProductEvent } from "$app/utils/user_analytics";
 
 import { NavigationButton } from "$app/components/Button";
+import { useDomains } from "$app/components/DomainSettings";
```
EmCousin:
```suggestion
```

app/services/order/create_service.rb
```diff
@@ -53,12 +53,35 @@ def perform
...
+
+        purchase, error, sca_response = Purchase::CreateService.new(
           product:,
-          params: purchase_params.merge(is_part_of_combined_charge: true),
+          params: purchase_params.merge(is_part_of_combined_charge: true).merge(card_params),
           buyer:
         ).perform
 
+        if sca_response
+          # Subscription restart SCA: add the upgrade purchase to the order so the
```
hchhabra, line 70: SCA is already handled by `Stripe::SetupIntentsController#create` before initiating order creation when the cart contains products from multiple sellers and the card requires SCA. In that case this might trigger double authentication/SCA flow, we may need to skip SCA here in that case I think?

app/models/installment.rb
hchhabra:
```suggestion
  def delivery_due?(purchase)
```

This method returns true when there's no workflow associated to the post, so maybe just `post.delivery_due?` or `post.eligible_for_delivery?` makes sense? As this eligibility check seems to be not specific to workflow installments.

app/services/purchase/create_service.rb
```diff
@@ -171,6 +178,65 @@ def is_gift?
...
+        else
+          Bugsnag.notify(StandardError.new("Existing subscription checkout attempt")) do |report|
+            report.severity = "info"
+            report.add_metadata(:subscription, {
+                                  subscription_id: active_subscription.id,
+                                  product_id: product.id,
+                                  email: purchase_params[:email]
+                                })
+          end
+          "Sorry, something went wrong. Please try again."
```
hchhabra: Think we can avoid saying "Please try again." here as we know that a retry is not going to help, maybe we can say "Please contact support@gumroad.com." or similar?

spec/requests/purchases/product/sca/sca_spec.rb
```diff
@@ -5,6 +5,9 @@
 
 describe("Purchase from a product page with SCA (Strong Customer Authentication)", type: :system, js: true) do
   before do
+    MerchantAccount.find_or_create_by!(user_id: nil, charge_processor_id: StripeChargeProcessor.charge_processor_id) do |ma|
```
hchhabra, line 8: Looks like this has been added in multiple specs, is it required now for some reason?

EmCousin: Those tests kept failing on my end because a Gumroad merchant account couldn't be found. Not sure how this is seeded on CI, so I went for a safe `find_or_create_by` that fixes the tests locally without breaking them on CI

PR #3202, opened 2026-01-28
Migrate edit product to inertia

sahitya-chandra:
Issue: #3030 
Closes #3030  
Parent issue: #3028
previous pr: #3096 and #3166 

### Problem

The Edit Product page (`/products/:id/edit`) used a single React on Rails server-rendered flow with one monolithic `LinksController#edit` and `LinksController#update`. All tabs (Product, Content, Receipt, Share) were handled via `params[:tab]`, causing full page reloads when switching tabs and sending unnecessary data for each tab.

### Solution

Migrated the Edit Product page to Inertia with **per-tab server endpoints**, matching the pattern used in merged PR (Bundles #3173).

**Approach:**

1. **Dedicated controller per tab** – Each tab has its own GET (edit) and PATCH/PUT (update) endpoint with resource-style URLs:
   - **Product** → `/products/:id/product/edit` → `Products::Edit::ProductController`
   - **Content** → `/products/:id/content/edit` → `Products::Edit::ContentController`
   - **Receipt** → `/products/:id/receipt/edit` → `Products::Edit::ReceiptController`
   - **Share** → `/products/:id/share/edit` → `Products::Edit::ShareController`
   - Backward-compatibility redirects for old URLs (e.g. `/products/:id/edit` → `/products/:id/product/edit`).

2. **Presenter per tab** – Shared logic in `Products::Edit::BasePresenter`; each tab presenter adds only tab-specific props (Content tab does not receive pricing, Share tab does not receive file configs, etc.).

3. **Inertia pages** – One page component per tab under `app/javascript/pages/Products/Edit/` (Product.tsx, Content.tsx, Receipt.tsx, Share.tsx), with a shared `ProductEditLayout.tsx` using Inertia’s `useForm()` for save.

4. **Dual Inertia + JSON** – Product edit controllers support both:
   - **Inertia**: `form.patch()` from the "Save changes" button → redirect + flash.
   - **JSON**: Non-Inertia callers → `{ success: true }`.

   **Publish/unpublish** in `LinksController` support both: **Inertia** — "Save and continue" / "Unpublish" call `save()` (form.patch) then `router.post()`, with server-side flash ("Published!" / "Unpublished!"); **JSON** — for other callers.

5. **Cleanup** – Removed old React on Rails entry: `packs/product_edit.ts`, `views/links/edit.html.erb`, `ProductEditPage.tsx` server component, and the monolithic `edit`/`update` actions from `LinksController`.

**Key files:**

[...]

PR #3173, opened 2026-01-27
Bundle edit inertia migration

yashranaway:
Fixes: https://github.com/antiwork/gumroad/issues/3031
Previous PR: #2892

# Description

## Problem

Bundle edit page uses React on Rails with client-side routing. This doesn't match the rest of the app which uses Inertia.

## Solution

Migrate to Inertia with three separate controllers for each tab:
- `ProductController` - name, price, description, covers
- `ContentController` - bundle products
- `ShareController` - discover settings (only accessible when published)

Server handles redirects and flash messages instead of client.

---

# Before/After

## Before

https://github.com/user-attachments/assets/70eb22d8-9dbf-4d9d-9655-9ff882b32852

## After 

https://github.com/user-attachments/assets/53171c93-4b48-47b6-8ad5-0a485e0972a0

---
# Test Results

image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Opus 4.5 was used to help implement the migration and fix review comments.

PR #3166, opened 2026-01-26
feat(products): migrate Edit Product page to Inertia

sahitya-chandra:
Issue: #3030 
fixes: #3030 
Parent issue: #3028 
previous pr: #3096 

# Description
This PR refactors the Product Edit page (`/links/:id/edit`) architecture to support our migration to Inertia, decoupling the monolithic edit experience into four dedicated, lightweight tabs.

Addressing #3030 and following the Inertia migration guidelines from #3028, this change moves away from a single heavy controller and React-Router-Dom client-side routing. Instead, each tab (Product, Content, Receipt, Share) is now powered by its own backend controller and presenter, delivering only the data necessary for that specific context.

### Problem

The Edit Product page (`/products/:id/edit`) used a single React on Rails server-rendered flow with one monolithic `LinksController#edit` and `LinksController#update`. All tabs (Product, Content, Receipt, Share) were handled via `params[:tab]`, causing full page reloads when switching tabs and sending unnecessary data for each tab.

### Solution

Migrated the Edit Product page to Inertia with **per-tab server endpoints**

**Approach:**

1. **Dedicated controller per tab** – Each tab has its own GET (edit) and PATCH/PUT (update) endpoint:
   - **Product** → `/products/:id/edit` → `Products::Edit::ProductController`
   - **Content** → `/products/:id/edit/content` → `Products::Edit::ContentController`
   - **Receipt** → `/products/:id/edit/receipt` → `Products::Edit::ReceiptController`
   - **Share** → `/products/:id/edit/share` → `Products::Edit::ShareController`

2. **Presenter per tab** – Shared logic in `Products::Edit::BasePresenter`; each tab presenter adds only tab-specific props (Content tab does not receive pricing, Share tab does not receive file configs, etc.).

3. **Inertia pages** – One page component per tab under `app/javascript/pages/Products/Edit/` (Product.tsx, Content.tsx, Receipt.tsx, Share.tsx), with a shared `ProductEditLayout.tsx` using Inertia’s `useForm()` for save.

4. **Dual Inertia + JSON** – Product edit controllers support both:
   - **Inertia**: `form.patch()` from the “Save changes” button → redirect + flash.
   - **JSON**: Non-Inertia callers → `{ success: true }`.

   **Publish/unpublish** in `LinksController` support both: **Inertia** — "Save and continue" / "Unpublish" call `save()` (form.patch) then `router.post()`, with server-side flash ("Published!" / "Unpublished!"); **JSON** — for other callers.

[...]

PR #3255, opened 2026-01-30
Migrate edit product to inertia

### LINKED ISSUE #3030: Migrate Edit Product page to Inertia

sahitya-chandra:
Issue: #3030  
Closes #3030  
Parent issue: #3028  
Previous PRs: #3096, #3166, #3202  

### Problem

The Edit Product page (`/products/:id/edit`) used a single React on Rails server-rendered flow with one monolithic `LinksController#edit` and `LinksController#update`. All tabs (Product, Content, Receipt, Share) were handled via `params[:tab]`, causing full page reloads when switching tabs and sending unnecessary data for each tab.

### Solution

Migrated the Edit Product page to Inertia with **per-tab server endpoints**, matching the pattern used in merged PRs (Bundles #3173).

**Approach:**

1. **Dedicated controller per tab** – Each tab has its own GET (edit) and PATCH (update) endpoint with resource-style URLs:
   - **Product** → `/products/:id/product/edit` → `Products::ProductController`
   - **Content** → `/products/:id/content/edit` → `Products::ContentController`
   - **Receipt** → `/products/:id/receipt/edit` → `Products::ReceiptController`
   - **Share** → `/products/:id/share/edit` → `Products::ShareController`
   Controllers live under `app/controllers/products/` (no `edit/` subdirectory). Backward-compatibility redirects for old URLs (e.g. `/products/:id/edit` → `/products/:id/product/edit`).

2. **Decoupled props per tab** – Each edit page receives only the props it needs. `Products::Edit::BasePresenter` provides:
   - **`layout_props`** – shared top-level props (id, unique_permalink, seller, currency_type, taxonomies, etc.) with no product.
   - **Tab-specific product props:** `product_tab_product_props`, `content_tab_product_props`, `receipt_tab_product_props`, `share_tab_product_props`. Each tab presenter merges `layout_props` with its tab’s product hash so Content/Receipt/Share do not receive pricing, refund, or other unrelated fields.

3. **Inertia pages** – One page component per tab under `app/javascript/pages/Products/` (Product/Edit.tsx, Content/Edit.tsx, Receipt/Edit.tsx, Share/Edit.tsx), with a shared `ProductEditLayout.tsx` using Inertia’s `useForm()` for save.

4. **Per-page save config (post-review)** – The layout no longer derives the update URL from the path. Each tab page registers its save config via **`registerSaveConfig({ updateUrl, isContentTab? })`** in a `useEffect`. **`save()`** in `ProductEditLayout` uses `saveConfigRef.current.updateUrl` for `form.patch()` and applies content-specific payload transformation when `isContentTab` is true. **"Publish and continue"** sends a single PATCH with `publish: true` to the current tab’s update URL; the server saves, publishes, and redirects to the Share tab.

5. **Publish/unpublish in tab controllers only (post-review)** – **`LinksController`** no longer has `edit`, `publish`, or `unpublish` actions. Publish and unpublish are handled entirely in the Product, Content, Receipt, and Share controllers:
   - **Product / Content / Share:** When `params[:unpublish]` is present and product is published, the action short-circuits (only unpublishes, no full tab update) to avoid validation issues with minimal payloads.
   - **Content:** Supports `params[:publish]`; after saving content, publishes and redirects to Share.
   - **Receipt:** Supports `params[:publish]`; after saving receipt attributes, publishes and redirects to Share. Unpublish same as other tabs.
   - **Share:** Update in transaction; no separate publish endpoint needed from this tab.

6. **Cleanup** – Removed old React on Rails entry: `packs/product_edit.ts`, `views/links/edit.html.erb`, `ProductEditPage.tsx` server component, and the monolithic `edit`/`update`/`publish`/`unpublish` actions from `LinksController`.

**Key files (migration):**

- `app/controllers/products/base_controller.rb` – Base for all tab controllers, `layout "inertia"`.
- `app/controllers/products/product_controller.rb`, `content_controller.rb`, `receipt_controller.rb`, `share_controller.rb` – Tab-specific edit/update with publish/unpublish handling.
- `app/presenters/products/edit/base_presenter.rb` – `layout_props` and tab-specific product prop methods.
- `app/presenters/products/edit/*_tab_presenter.rb` – Each returns `layout_props.merge(product: <tab>_tab_product_props)`.
- `app/javascript/layouts/ProductEditLayout.tsx` – Inertia layout with `useForm()`, `registerSaveConfig`, `save(options?: { publish?: boolean })`, `form.patch(config.updateUrl)`.
- `app/javascript/pages/Products/*/Edit.tsx` – Per-tab Inertia pages; each calls `registerSaveConfig({ updateUrl: Routes.product_*_path(uniquePermalink), isContentTab? })`.
- `app/javascript/components/ProductEdit/Layout.tsx` – Tab chrome, Save / Save and continue / Publish and continue / Unpublish; calls `save()` or `setPublished(true|false)` (no direct `router.post` for publish/unpublish).
- `app/policies/link_policy.rb` – Permitted attributes split into `*_tab_permitted_attributes` per tab.

**Out of scope for this PR (addressed separately):**  
Defensive frontend fixes for **per-tab payloads** (e.g. `product.variants ?? []`, `product.refund_policy ?? defaultRefundPolicy`, `tags ?? []`) are in a separate commit. Those changes guard against undefined fields when the product payload is minimal (e.g. after navigating from another tab or after redirect). They are not part of the migration contract.

---

## Before / After

- **Before:** One route, one controller, one React on Rails app; tab switch = full reload; all tab data sent every time.

[Screencast from 2026-01-28 19-04-58.webm](https://github.com/user-attachments/assets/94f812fb-b7b6-4aeb-b64a-c25a0d0723b1)

- **After:** One URL per tab, one controller and presenter per tab; tab switch = Inertia SPA navigation; each tab gets only its data; Save/Publish and continue use per-page update URL via `registerSaveConfig`.

(Update on 3 Feb)
[Screencast from 2026-02-03 18-36-55.webm](https://github.com/user-attachments/assets/ec70fa60-2841-4dc9-b4a7-b0c4b4982c8a)

---

## Implementation notes

- **Routes:** Resource-style routes under `resources :products` with per-tab `resource` for edit/update (`edit_product_product_path`, `edit_product_content_path`, etc.). Backward-compatibility redirects for legacy URLs.
- **Flash & redirects:** Success/error redirects pass `notice:`/`alert:` in the redirect and use `status: :see_other` (303). No separate `flash` assignment before redirect.
- **Unpublish:** When the user clicks Unpublish on Product, Content, or Receipt tab, the action short-circuits and only unpublishes (no full tab update) so the request works when the frontend sends only `{ unpublish: true }` and to avoid membership tier validation. Share tab runs update in transaction then unpublish.
- **Publish and continue:** Single PATCH to the current tab’s update URL with `publish: true`; server updates attributes (if any), calls `@product.publish!`, then redirects to `edit_product_share_path` with "Published!".
- **Error handling:** Method-level `rescue` at the end of each update action (no `begin`/`rescue`). All error redirects use `status: :see_other`.
- **`Product::VariantsUpdaterService`:** `price_difference_cents` can be nil when not sent (e.g. non-coffee variant) or blank; `.to_i` coerces string params from forms. Comment added in the service.

---

## Test results

- **Product edit controllers:** **33 examples** in `spec/controllers/products/edit/` (Product, Content, Receipt, Share; GET edit + PATCH update, authorize/collaborator, publish/unpublish, tab-specific behavior).
- **Product edit presenters:** **6 examples** in `spec/presenters/products/edit/` (BasePresenter layout_props/product_minimal_props; ProductTabPresenter, ContentTabPresenter, ReceiptTabPresenter, ShareTabPresenter each return layout props + tab-specific product props only).
- **Total: 39 examples** for this migration (33 controller + 6 presenter).

Run (PR-related specs only):

```bash
bundle exec rspec spec/controllers/products/edit/ spec/presenters/products/edit/ --format documentation
```

Screenshot from 2026-02-03 18-08-55

---

## Checklist

# AI Disclosure

- Used Cursor with opus-4.5 to plan the solution and refactoring code.

- AI was also used to:
  1. Generate the initial boilerplate for the 4 new controllers and presenters.
  2. Refactor `LinkPolicy` to split permitted attributes.
  3. Write the rspec test suite for the new controllers.
  4. Assist in migrating the `saveProduct` frontend logic to use dynamic Inertia endpoints via `registerSaveConfig` and per-tab update URLs.

### DIFF
--- a/app/controllers/products/base_controller.rb
+++ b/app/controllers/products/base_controller.rb
@@ -38,7 +38,7 @@ class Products::BaseController < Sellers::BaseController
     end
 
     def fetch_product
-      product_id = params[:product_id]
+      product_id = params[:product_id] || params[:id]
       @product = Link.fetch_leniently(product_id, user: current_seller) || Link.fetch_leniently(product_id)
       raise(ActiveRecord::RecordNotFound) if @product.nil? || @product.deleted_at.present?
     end


app/presenters/products/edit/base_presenter.rb
sahitya-chandra, line 1: Each edit tab now gets only the product props it needs: Product tab gets full product; Content gets name, files, rich_content, variants, etc.; Receipt gets receipt/custom text fields; Share gets tags, taxonomy_id, display_product_reviews, is_adult, custom_domain. Layout still gets shared top-level props via `layout_props` (no product). This avoids sending pricing/refund/share data to tabs that don’t use it.

config/routes.rb
```diff
@@ -278,6 +278,25 @@ def product_info_and_purchase_routes(named_routes: true)
   get "/s3_utility/generate_multipart_signature", to: "s3_utility#generate_multipart_signature"
 
   constraints GumroadDomainConstraint do
+    resources :products, only: [], param: :id do
+      scope module: "products" do
+        resource :product, only: [:edit, :update], controller: "edit/product"
+        resource :content, only: [:edit, :update], controller: "edit/content"
+        resource :receipt, only: [:edit, :update], controller: "edit/receipt"
```
sahitya-chandra: Dropped `path: "product"` from the product resource; the default path for a singular `:product` resource is already `"product"`, so behavior is unchanged and the config is simpler.

app/services/product/variants_updater_service.rb
```diff
@@ -57,9 +57,10 @@ def clean_variants_params(params)
           options: options&.map do |option|
             new_option = option.slice(:id, :temp_id, :name, :description, :url, :customizable_price, :recurrence_price_values, :max_purchase_count, :integrations, :rich_content, :apply_price_changes_to_existing_memberships, :subscription_price_change_effective_date, :subscription_price_change_message, :duration_in_minutes)
 
+            # price_difference_cents can be nil when not sent (e.g. non-coffee variant) or blank; .to_i coerces string params from forms.
             # TODO: :product_edit_react cleanup
             if option[:price_difference_cents].present?
-              option[:price] = option[:price_difference_cents]
+              option[:price] = option[:price_difference_cents].to_i
```
sahitya-chandra, line 63: `price_difference_cents` can be nil when not sent (e.g. non-coffee variant) or blank, and that `.to_i` is used to coerce string params from forms. The existing `.present?` guard and `.to_i` logic are unchanged.

spec/presenters/products/edit/base_presenter_spec.rb
sahitya-chandra, line 1: Added specs to ensure each tab presenter returns layout props plus only that tab’s product keys (e.g. Content/Receipt/Share do not include `price_cents`, `refund_policy`). BasePresenter specs cover `layout_props` (no product) and `product_minimal_props` (name, is_published, files, native_type).

sahitya-chandra:
@EmCousin sir,

 I have incorporated your review comments in #3202. Could you take another look when you have time? Thanks.

sahitya-chandra:
@EmCousin _a sir,

 I have incorporated your review comments in #3202. Could you take another look when you have time? Thanks.
 
 Cc @Pradumn27

sahitya-chandra:
@EmCousin sir,

 I have incorporated your review comments in #3202. Could you take another look when you have time? Thanks.
 
 Cc @Pradumn27

app/controllers/products/edit/base_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      private
+        def set_product_edit_title
+          set_meta_tag(title: @product.name)
+        end
+
+        def fetch_product
+          product_id = params[:product_id] || params[:id]
+          @product = Link.fetch_leniently(product_id, user: current_seller) || Link.fetch_leniently(product_id)
+          raise(ActiveRecord::RecordNotFound) if @product.nil? || @product.archived? || @product.deleted_at.present?
```
EmCousin: We want to be able to edit archived products? What is the reason you added this constraint here? Right now it breaks the behaviour when I want to click on Edit in the Archived products list

config/routes.rb
```diff
@@ -278,6 +278,25 @@ def product_info_and_purchase_routes(named_routes: true)
   get "/s3_utility/generate_multipart_signature", to: "s3_utility#generate_multipart_signature"
 
   constraints GumroadDomainConstraint do
+    resources :products, only: [], param: :id do
```
EmCousin: This is already the case by default.
```suggestion
    resources :products, only: [] do
```

app/presenters/products/edit/base_presenter.rb
```diff
@@ -0,0 +1,249 @@
+# frozen_string_literal: true
+
+module Products
+  module Edit
+    class BasePresenter
```
EmCousin: We do not nest modules/classes in this code base, the general style is to declare those inline (`class Products::Edit::BasePresenter`)

app/controllers/products/edit/base_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+          end
+        end
+
+      private
+        def set_product_edit_title
+          set_meta_tag(title: @product.name)
+        end
+
+        def fetch_product
+          product_id = params[:product_id] || params[:id]
```
EmCousin: Again, there is no case where the product_id will be present in `params[:id]` when hitting the child controllers' endpoints, it will always be in `params[:product_id]`

app/presenters/products/edit/base_presenter.rb
```diff
@@ -0,0 +1,249 @@
...
+        {
+          name: product.name,
+          is_published: !product.draft && product.alive?,
+          files: files_data,
+          native_type: product.native_type,
+        }
+      end
+
+      # Full product hash for the Product tab only.
+      def product_tab_product_props
```
EmCousin: This should be defined in ProductTabPresenter

app/presenters/products/edit/base_presenter.rb
```diff
@@ -0,0 +1,249 @@
...
+            discount: product.default_offer_code.discount,
+          } : nil,
+          public_files: public_files_data,
+          audio_previews_enabled: Feature.active?(:audio_previews, product.user),
+          community_chat_enabled: Feature.active?(:communities, product.user) ? product.community_chat_enabled? : nil,
+        )
+      end
+
+      # Product hash for the Content tab only (content-related fields + minimal).
+      def content_tab_product_props
```
EmCousin: This should be defined in ContentTabPresenter

app/presenters/products/edit/base_presenter.rb
```diff
@@ -0,0 +1,249 @@
...
+          description: product.description || "",
+          rich_content: product.rich_content_json,
+          variants: variants_data,
+          has_same_rich_content_for_all_variants: product.has_same_rich_content_for_all_variants?,
+          is_multiseat_license: product.is_multiseat_license,
+        )
+      end
+
+      # Product hash for the Receipt tab only (receipt-related fields + minimal).
+      def receipt_tab_product_props
```
EmCousin: This should be defined in ReceiptTabPresenter

app/presenters/products/edit/base_presenter.rb
```diff
@@ -0,0 +1,249 @@
...
+        product_minimal_props.merge(
+          custom_view_content_button_text: product.custom_view_content_button_text,
+          custom_view_content_button_text_max_length: ::Product::Validations::MAX_VIEW_CONTENT_BUTTON_TEXT_LENGTH,
+          custom_receipt_text: product.custom_receipt_text,
+          custom_receipt_text_max_length: ::Product::Validations::MAX_CUSTOM_RECEIPT_TEXT_LENGTH,
+        )
+      end
+
+      # Product hash for the Share tab only (share-related fields + minimal).
+      def share_tab_product_props
```
EmCousin: This should be defined in ShareTabPresenter

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -1,11 +1,8 @@
+import { Link, router, usePage } from "@inertiajs/react";
 import cx from "classnames";
 import * as React from "react";
-import { Link, useMatches, useNavigate } from "react-router-dom";
 
-import { saveProduct } from "$app/data/product_edit";
-import { setProductPublished } from "$app/data/publish_product";
```
EmCousin, line 6: This function is no longer used and therefore should be deleted. Do not leave dead code hanging around.

app/controllers/products/edit/content_controller.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+module Products
+  module Edit
```
EmCousin: There's no need for an Edit module here. Just declare a Products::ContentController with an `edit` action, like in Bundles::ContentController

app/controllers/products/edit/content_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    class ContentController < BaseController
+      def edit
+        render inertia: "Products/Edit/Content", props: Products::Edit::ContentTabPresenter.new(product: @product, pundit_user:).props
```
EmCousin: Ref. comment above, the template should be `Products/Content/Edit`

app/javascript/layouts/ProductEditLayout.tsx
```diff
@@ -0,0 +1,239 @@
...
+import { assertResponseError, request } from "$app/utils/request";
+import {
+  ProductEditContext,
+  Product,
+  ContentUpdates,
+  ExistingFileEntry,
+  ProfileSection,
+  ShippingCountry,
+} from "$app/components/ProductEdit/state";
+
```
EmCousin, line 30: There is still a lot of tight coupling in this layout, there should be conditionals in there do determine what to do depending on which pages we are on. Each page should instead have their own edit props, their own form object and their own update function

app/javascript/layouts/ProductEditLayout.tsx
```diff
@@ -0,0 +1,239 @@
...
+        },
+        currencyType: data.currencyType,
+      }));
+
+      const pathname = window.location.pathname;
+      let updateUrl: string;
+      if (pathname.includes("/content/edit")) updateUrl = Routes.product_content_path(uniquePermalink);
+      else if (pathname.includes("/receipt/edit")) updateUrl = Routes.product_receipt_path(uniquePermalink);
+      else if (pathname.includes("/share/edit")) updateUrl = Routes.product_share_path(uniquePermalink);
+      else updateUrl = Routes.product_product_path(uniquePermalink);
```
EmCousin: We should not be determining the endpoint url in the layout based on conditionals, instead each edit page should have their own update function

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -137,36 +134,65 @@ export const Layout = ({
...
+          onSuccess: () => updateProduct({ is_published: false }),
+          onFinish: () => setIsPublishing(false),
+        });
       }
-    } catch (e) {
-      assertResponseError(e);
-      showAlert(e.message, "error", { html: true });
     }
-    setIsPublishing(false);
   };
```
EmCousin, line 189: This does not make sense. Publishing/Unpublishing a product is now part of the update function by passing either `publish` or `unpublish` to the request's body, then we told the backend what to do based on the presence or absence of those values. You should not be needing the `publish_link_path` nor the `unpublish_link_path` at all. Again, look how it's done for updating a bundle.

app/controllers/links_controller.rb
```diff
@@ -275,173 +275,59 @@ def track_user_action
...
     end
 
-    render json: { success: true }
+    if request.inertia?
+      flash[:notice] = "Published!"
+      redirect_to edit_product_share_path(@product.unique_permalink)
+    else
+      render json: { success: true }
+    end
   end
```
EmCousin: These endpoints should not be used anymore since publishing/unpublishing is part of the update action in each products edit page

sahitya-chandra (commented):
## Defensive fixes commit (per-tab payloads)

**Commit:** `Defensive fixes for per-tab payloads (undefined guards)`

After the migration, each edit tab (Product, Content, Receipt, Share) gets only the props it needs. When you switch tabs or get redirected (e.g. "Publish and continue"), the frontend can render with a product object that’s missing fields other tabs expect. These changes add safe defaults so we never read properties on `undefined`.

### Files changed and why

| File | Why |
|------|-----|
| `app/javascript/components/CustomDomain.tsx` | `product.custom_domain` can be undefined on some tabs; use `(customDomain ?? "").trim()` and `value={customDomain ?? ""}` so the input never throws. |
| `app/javascript/components/Product/ConfigurationSelector.tsx` | `product.options` and `product.recurrences?.enabled` can be undefined; use `(product.options ?? []).length`, `(product.recurrences?.enabled ?? []).map(...)` and similar so all option/recurrence logic is safe. |
| `app/javascript/components/Product/Layout.tsx` | Same product shape used in sticky bar; guard `product.bundle_products` and `product.options` with `?? []` for `.length` checks. |
| `app/javascript/components/Product/index.tsx` | Main product view used in ProductPreview; guard `product.bundle_products`, `product.options`, `product.attributes`, `product.covers` with `?? []` for `.length`, `.find`, `.map`, and when passing to child components. |
| `app/javascript/components/ProductEdit/ContentTab/index.tsx` | Content tab can load with minimal payload (e.g. after redirect from Share); guard `product.variants` and `product.rich_content` with `?? []` in ContentTab, ContentTabContent, and ContentTabHeaderActions so `.find`, `[0]`, and `.length` never run on undefined. |
| `app/javascript/components/ProductEdit/ProductPreview.tsx` | `serializedProduct.attributes` was set from `product.custom_attributes`, which can be undefined on Share/Receipt; set `attributes: product.custom_attributes ?? []` so the Product component always receives an array. |
| `app/javascript/components/ProductEdit/ProductTab/AttributesEditor.tsx` | Receives `product.custom_attributes` which can be undefined; use `const safeCustomAttributes = customAttributes ?? []` and use it for `.length`, `.map`, `.slice`, `.filter` and the add-button so the editor never crashes. |
| `app/javascript/components/ProductEdit/ProductTab/CircleIntegrationEditor.tsx` | `product.integrations` and `product.variants` can be undefined; use `(product.variants ?? []).every(...)`, `integrations?.circle` so the Circle integration UI doesn’t throw. |
| `app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx` | Receives `product.covers` which can be undefined; use `const safeCovers = covers ?? []` for `[0]`, `.length`, Sortable list, and Covers component so the cover editor always has an array. |
| `app/javascript/components/ProductEdit/ProductTab/ThumbnailEditor.tsx` | `coverUrlForThumbnail(covers)` is called with `product.covers` which can be undefined; make the helper accept `covers` optional and use `(covers ?? []).find(...)` so it never throws. |
| `app/javascript/components/ProductEdit/ProductTab/TiersEditor.tsx` | Uses `Object.entries(product.integrations)`; when integrations is undefined, use `Object.entries(product.integrations ?? {})` so the tier editor doesn’t crash. |
| `app/javascript/components/ProductEdit/ProductTab/VersionsEditor.tsx` | Same as TiersEditor; use `Object.entries(product.integrations ?? {})` for the versions editor. |
| `app/javascript/components/ProductEdit/ProductTab/index.tsx` | Product tab can load with minimal payload; use `const integrations = product.integrations ?? {}` and pass `integrations.circle`, `integrations.discord`, `integrations.google_calendar` (and spread `integrations`) so Circle/Discord/Google Calendar editors never see undefined. |
| `app/javascript/components/ProductEdit/RefundPolicy.tsx` | Receives `product.refund_policy` which can be undefined; add `defaultRefundPolicy` and `refundPolicy?: RefundPolicy | null`, use `const safeRefundPolicy = refundPolicy ?? defaultRefundPolicy` everywhere so the selector never reads `.max_refund_period_in_days` or `.allowed_refund_periods_in_days` on undefined. |
| `app/javascript/components/ProductEdit/ShareTab/TagSelector.tsx` | Receives `product.tags` which can be undefined (e.g. after redirect from Content); use `const safeTags = tags ?? []` and make `tags` optional so `.map`, `.length`, and tag list never throw. |

### Summary

All 15 files add **defensive defaults** (`?? []`, `?? {}`, optional chaining) so components work when the product payload is minimal (per-tab or after redirect). This commit is separate from the migration itself; it only fixes runtime errors from reading properties on undefined.

app/controllers/products/edit/base_controller.rb
sahitya-chandra: I added the `archived?` check so we don’t treat archived products as normal editable products in the new flow, but that was the wrong place for it and it breaks Edit from the Archived list. I’ll remove the archived check in `fetch_product` so behavior matches the previous Edit flow and you can open the edit page from the Archived products list again.

after removal:

[Screencast from 2026-02-03 20-14-28.webm](https://github.com/user-attachments/assets/be5857b7-a093-4a8c-bb10-0f17541074cd)

sahitya-chandra:
@EmCousin sir, all review comments are addressed and the PR description is updated. We also added a commit with defensive frontend fixes (guards for undefined per-tab payloads in 15 files) so the UI doesn’t crash when switching tabs or after redirects. Please take another look when you can.

EmCousin:
PR has not been tested properly, some fields are no longer updated in the backend due to permitted params not being properly listed in the LinkPolicy.

Issue #3258, opened 2026-01-30
Products page auto-scrolls down on initial navigation


4arjun:
## Description
When navigating to the Products page (/products), the page automatically scrolls down instead of staying at the top.

## Steps to Reproduce
1. Navigate to any other page in the dashboard
2. Click on "Products" in the navigation
3. Observe the page scrolls down slightly after loading

## Expected Behavior
The page should remain at the top when navigating to the Products page.

## Screen recording

https://github.com/user-attachments/assets/04dd6eba-c045-44fe-981e-797b14547e30

PR #3259, opened 2026-01-30
Fix products page auto-scrolling on initial navigation

4arjun:
Issue: #3258 

## Root cause

The Products page was unexpectedly scrolling down on initial navigation due to how the search query state was initialized.

In ProductsDashboardPage.tsx, the query state was initialized to an empty string (""). The ProductsTable component contains a useEffect that triggers a reload whenever props.query !== null. Since an empty string is not null, this effect ran immediately on initial mount. As a result, debouncedLoadProducts() was called, which eventually invoked scrollIntoView({ behavior: "smooth" }), causing the unwanted scroll.

Fixed by using a ref to track the previous query value and only triggering the search reload when the query actually changes from user input, not on initial render

## Before

https://github.com/user-attachments/assets/b8040736-7e70-4d19-b97f-7adf1469943c

## After

https://github.com/user-attachments/assets/c047899c-289d-4550-8c58-0e6ef17ae7cf

# Checklist

---

# AI Disclosure
No AI used

### DIFF
--- a/app/javascript/components/ProductsPage/MembershipsTable.tsx
+++ b/app/javascript/components/ProductsPage/MembershipsTable.tsx
@@ -20,6 +20,7 @@ import {
   TableRow,
 } from "$app/components/ui/Table";
 import { useDebouncedCallback } from "$app/components/useDebouncedCallback";
+import { useOnChange } from "$app/components/useOnChange";
 import { useUserAgentInfo } from "$app/components/UserAgent";
 import { Sort, useSortingTableDriver } from "$app/components/useSortingTableDriver";
 
@@ -72,13 +73,9 @@ export const ProductsPageMembershipsTable = (props: {
   };
 
   const debouncedLoadMemberships = useDebouncedCallback(() => loadMemberships(1), 300);
-  const prevQueryRef = React.useRef(props.query);
 
-  React.useEffect(() => {
-    if (props.query !== null && prevQueryRef.current !== props.query) {
-      debouncedLoadMemberships();
-    }
-    prevQueryRef.current = props.query;
+  useOnChange(() => {
+    if (props.query !== null) debouncedLoadMemberships();
   }, [props.query]);
 
   const reloadMemberships = () => loadMemberships(pagination.page);
--- a/app/javascript/components/ProductsPage/ProductsTable.tsx
+++ b/app/javascript/components/ProductsPage/ProductsTable.tsx
@@ -20,6 +20,7 @@ import {
   TableRow,
 } from "$app/components/ui/Table";
 import { useDebouncedCallback } from "$app/components/useDebouncedCallback";
+import { useOnChange } from "$app/components/useOnChange";
 import { useUserAgentInfo } from "$app/components/UserAgent";
 import { Sort, useSortingTableDriver } from "$app/components/useSortingTableDriver";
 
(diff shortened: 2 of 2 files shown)


app/javascript/components/ProductsDashboardPage.tsx
```diff
@@ -35,7 +35,7 @@ export const ProductsDashboardPage = ({
   can_create_product: canCreateProduct,
 }: ProductsDashboardPageProps) => {
   const [enableArchiveTab, setEnableArchiveTab] = React.useState(archivedProductsCount > 0);
-  const [query, setQuery] = React.useState("");
+  const [query, setQuery] = React.useState<string | null>(null);
```
4arjun: Prevent the useEffect in `ProductsTable` from triggering on mount.

Hanaffi: Would it read better to adjust the condition in `ProductsTable` to skip the initial render?
`if (props.query !== null && prevQueryRef.current !== props.query)`

app/javascript/components/ProductsPage/ProductsTable.tsx
```diff
@@ -72,9 +72,13 @@ export const ProductsPageProductsTable = (props: {
...
 
   const debouncedLoadProducts = useDebouncedCallback(() => loadProducts(1), 300);
+  const prevQueryRef = React.useRef(props.query);
 
   React.useEffect(() => {
-    if (props.query !== null) debouncedLoadProducts();
+    if (props.query !== null && prevQueryRef.current !== props.query) {
+      debouncedLoadProducts();
+    }
+    prevQueryRef.current = props.query;
```
4arjun: Previously, the effect ran on mount because props.query was "" (not null), triggering debouncedLoadProducts() which calls scrollIntoView(), causing the page to scroll down unexpectedly.

Added prevQueryRef.current !== props.query check so the effect only runs when the query value actually changes. On initial render, both values are "", so the condition is false and the scroll doesn't happen.

app/javascript/components/ProductsPage/MembershipsTable.tsx
```diff
@@ -72,9 +72,13 @@ export const ProductsPageMembershipsTable = (props: {
...
 
   const debouncedLoadMemberships = useDebouncedCallback(() => loadMemberships(1), 300);
+  const prevQueryRef = React.useRef(props.query);
 
   React.useEffect(() => {
-    if (props.query !== null) debouncedLoadMemberships();
+    if (props.query !== null && prevQueryRef.current !== props.query) {
+      debouncedLoadMemberships();
+    }
+    prevQueryRef.current = props.query;
```
4arjun: same here

app/javascript/components/ProductsDashboardPage.tsx
4arjun: Done

4arjun:

binary-koan:
Could we just change the `useEffect` to our `useOnChange`? That's intended to fix problems with initial render like this

4arjun:
> Could we just change the `useEffect` to our `useOnChange`? That's intended to fix problems with initial render like this

Done! Updated to use `useOnChange` instead.

binary-koan (approved):
Looks great!

Issue #3262, opened 2026-01-30
Search bar disappears when no products match the search query


4arjun:
## Description
On the Products page, when a user searches for products and the search query returns zero results, the search bar disappears. Additionally, after the search bar disappears, subsequent searches cannot be performed because the table component unmounts and its search effect stops running.

## Screen recording

https://github.com/user-attachments/assets/47cc8f06-4700-4eb0-9675-d06ef175e691

PR #3263, opened 2026-01-30
Fix search bar disappearing when no products match search query

4arjun:
Issue: #3262

## Summary
The search bar disappeared when entering a search query that returned no matching products, preventing users from modifying their search.

## Fix

Track initial product count with a ref and always render table components so the search functionality remains active regardless of result count.

## Before

https://github.com/user-attachments/assets/6e94d294-e0c4-4902-a2ea-d119189b0309

## After

https://github.com/user-attachments/assets/48cc0c28-0fc2-4121-a7ab-14f8aaace378

# Checklist

---

# AI Disclosure
No AI used

### DIFF
--- a/app/javascript/components/ProductsDashboardPage.tsx
+++ b/app/javascript/components/ProductsDashboardPage.tsx
@@ -45,7 +45,9 @@ export const ProductsDashboardPage = ({
       archivedTabVisible={enableArchiveTab}
       ctaButton={
         <>
-          {hadInitialItems.current ? <Search value={query} onSearch={setQuery} placeholder="Search products" /> : null}
+          {hadInitialItems.current || !!query ? (
+            <Search value={query} onSearch={setQuery} placeholder="Search products" />
+          ) : null}
           <NavigationButtonInertia href={Routes.new_product_path()} disabled={!canCreateProduct} color="accent">
             New product
           </NavigationButtonInertia>


app/javascript/components/ProductsPage.tsx
```diff
@@ -30,27 +30,23 @@ const ProductsPage = ({
...
-      />
-    ) : null}
+    <ProductsPageProductsTable
+      query={query}
+      entries={products}
+      pagination={productsPagination}
+      sort={productsSort}
+      selectedTab={type}
+      setEnableArchiveTab={setEnableArchiveTab}
+    />
```
4arjun, line 49: Previously, table components were only rendered when they had items. This caused a bug: when search returned no results, the component unmounted and its useEffect (which triggers new searches) stopped running. Now we always render both tables so the search effects keep working. The tables internally return null when empty, so there's no visual difference.

app/javascript/components/ProductsDashboardPage.tsx
```diff
@@ -36,6 +36,7 @@ export const ProductsDashboardPage = ({
 }: ProductsDashboardPageProps) => {
   const [enableArchiveTab, setEnableArchiveTab] = React.useState(archivedProductsCount > 0);
   const [query, setQuery] = React.useState("");
+  const hadInitialItems = React.useRef(products.length > 0 || memberships.length > 0);
```
4arjun, line 39: Captures the initial state.

app/javascript/components/ProductsDashboardPage.tsx
```diff
@@ -44,15 +45,15 @@ export const ProductsDashboardPage = ({
       archivedTabVisible={enableArchiveTab}
       ctaButton={
         <>
-          {products.length > 0 ? <Search value={query} onSearch={setQuery} placeholder="Search products" /> : null}
+          {hadInitialItems.current ? <Search value={query} onSearch={setQuery} placeholder="Search products" /> : null}
```
Hanaffi: query initial state should be coming from `urlParams?.get("query")` and if we have any query, we should keep the search to allow users to remove the query.

```suggestion
          {hadInitialItems.current || !!query ? <Search value={query} onSearch={setQuery} placeholder="Search products" /> : null}
```

4arjun:
A better approach would be #3452

binary-koan:
Yep I agree the other approach is better, will review that one

Issue #3261, opened 2026-01-30
Gumroad Library Error


wooyunjin:
Hello!

I have extended videos purchased from YouTube in my Gumroad library, which I often watch more than once. Recently, it started showing an error message that these videos are not found (see attached). I can only access them through the email confirmations I received at the time of purchasing them, which I only have most recent few left because I already deleted the rest from my mailbox. If you could fix this issue, I would greatly appreciate it!

Warmly,
Yunjin

<img width="828" height="1792" alt="Image" src="https://github.com/user-attachments/assets/f5806714-7017-491d-b580-ee4f35cec7b1" />

vatsalkaushik:
Sorry about that! We have just released a fix for this issue. Your library should load correctly now. Please let us know if you continue facing any issues.

PR #3269, opened 2026-01-30
Revert "Migrate URL Redirect Confirm page to Inertia"

EmCousin:
Reverts antiwork/gumroad#3195

This is likely causing the error documented in https://github.com/antiwork/gumroad/issues/3261


Devwill2:
@EmCousin Is there something wrong with the migration, or do we only find out by reverting the code?

EmCousin:
@Devwill2 it's a guess for now. Users have been reporting trouble accessing their library products guarded by that confirm form on mobile apps around the time we merged that PR. The plan is to continue investigating from here, assuming this revert is going to solve the issue.

Devwill2:
> @Devwill2 it's a guess for now. Users have been reporting trouble accessing their library products guarded by that confirm form on mobile apps around the time we merged that PR. The plan is to continue investigating from here, assuming this revert is going to solve the issue.

@EmCousin  Please keep me in the loop. I can help if we need to investigate after the revert
Thanks

Devwill2:
@EmCousin Can you help identify why this issue is occurring and what might be wrong in the PR that’s causing it? Have you found the root cause yet?
Thanks!

EmCousin:
@Devwill2 not yet. We'll be prioritizing other Inertia migrations for the time being and we'll get back to this one in due time.

Issue #3271, opened 2026-01-30
Discount codes don't automatically apply with ?wanted=true parameter


ershad:
## Problem
When using the `?wanted=true` URL parameter, automatic discount codes are not being applied to the product.

## Reproduction Steps
1. Visit a product URL with both a discount code and `?wanted=true` parameter
2. Example URL: `http://dvassallo.gumroad.com/l/small-bets?wanted=true`
3. Observe that the discount is not automatically applied

## Expected Behavior
The discount should automatically apply when using `?wanted=true`, similar to how it works without that parameter.

PR #3272, opened 2026-01-30
Apply offer code automatically when wanted param is set

Hanaffi:
Issue: https://github.com/antiwork/gumroad/issues/3271
Closes: https://github.com/antiwork/gumroad/issues/3271

# Description

When user visits a product with wanted param, apply offer code automatically

## Problem
default offer code wasn't applied when wanted param is set

## Solution
Apply default offer code 

---

# Before/After
(I'm using the test env to show the before/after)

### Before
[Before.webm](https://github.com/user-attachments/assets/43a25fb6-7a45-4378-b250-8a5411d6fd4f)

#### After
[After.webm](https://github.com/user-attachments/assets/8db66d36-f2b2-4ae5-ab4a-fc0cba1095a2)

---

# Test Results

image
image

---

# Checklist

---

# AI Disclosure

Cursor for auto-completion

### DIFF
--- a/spec/services/best_offer_code_service_spec.rb
+++ b/spec/services/best_offer_code_service_spec.rb
@@ -92,7 +92,7 @@ describe BestOfferCodeService do
 
           it "returns the url_code" do
             expect(subject.result&.dig(:code)).to eq(url_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+            expect(subject.result&.dig(:valid)).to be(true)
           end
         end
 
@@ -102,7 +102,7 @@ describe BestOfferCodeService do
 
           it "returns the default_code" do
             expect(subject.result&.dig(:code)).to eq(default_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+            expect(subject.result&.dig(:valid)).to be(true)
           end
         end
 
@@ -112,7 +112,7 @@ describe BestOfferCodeService do
 
           it "returns the url_code" do
             expect(subject.result&.dig(:code)).to eq(url_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+            expect(subject.result&.dig(:valid)).to be(true)
           end
         end
 
@@ -122,7 +122,7 @@ describe BestOfferCodeService do
 
           it "returns the default_code" do
             expect(subject.result&.dig(:code)).to eq(default_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+            expect(subject.result&.dig(:valid)).to be(true)
           end
         end
 
@@ -132,7 +132,7 @@ describe BestOfferCodeService do
 
           it "returns the default_code (tie goes to default)" do
             expect(subject.result&.dig(:code)).to eq(default_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+            expect(subject.result&.dig(:valid)).to be(true)
           end
         end
 
@@ -143,7 +143,7 @@ describe BestOfferCodeService do
 
             it "returns the code with better discount (400 cents > 30% of 1000 = 300 cents)" do
               expect(subject.result&.dig(:code)).to eq(url_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+              expect(subject.result&.dig(:valid)).to be(true)
             end
           end
 
@@ -153,7 +153,7 @@ describe BestOfferCodeService do
 
             it "returns the code with better discount (50% of 1000 = 500 cents > 200 cents)" do
               expect(subject.result&.dig(:code)).to eq(url_offer_code.code)
-          expect(subject.result&.dig(:valid)).to be(true)
+              expect(subject.result&.dig(:valid)).to be(true)
             end
           end
         end


spec/requests/products/show/show_spec.rb
```diff
@@ -234,6 +234,33 @@
...
     end
+
+    context "with a product with a default offer code" do
+      let(:seller) { create(:user) }
+      let(:product) { create(:product, user: seller, price_cents: 10000) }
+      let(:default_offer_code) { create(:offer_code, user: seller, products: [product], code: "AUTO10", amount_cents: 1000) }
+
+      before do
+        product.update!(default_offer_code: default_offer_code)
+      end
```
Hanaffi, line 245: I wish if we could just assign the `default_offer_code` when creating the product but It didn't work.

app/controllers/links_controller.rb
```diff
@@ -128,10 +128,11 @@ def show
 
       unless (@product.customizable_price || cart_item[:option]&.[](:is_pwyw)) &&
              (params[:price].blank? || params[:price] < cart_item[:price])
+        code = params[:offer_code] || params[:code] || @product.default_offer_code&.code
```
ershad: On the product page, we [prioritize](https://github.com/antiwork/gumroad/blob/main/app/presenters/product_presenter/product_props.rb#L104) the best possible discount when the offer code is present in both params and as the default code. I think we should do the same here, maybe reuse the code to keep it DRY.

app/controllers/links_controller.rb
```diff
@@ -128,12 +128,19 @@ def show
...
-                                                         code: params[:offer_code] || params[:code],
-                                                         affiliate_id: params[:affiliate_id] || params[:a],
-                                                         referrer: params[:referrer] || request.referrer),
+        discount_result = BestOfferCodeService.new(
+          product: @product,
+          url_code: params[:offer_code] || params[:code],
+          quantity: (params[:quantity] || 1).to_i
+        ).result
+        code = discount_result&.dig(:code) if discount_result&.dig(:valid)
+        redirect_params = params.permit!.except(:code, :offer_code)
```
Hanaffi, line 137: The controller uses `**params.permit!`, which includes the original code/offer_code from the URL, and then also sets `code: code` explicitly. This causes duplicates or an invalid code to persist.

app/controllers/links_controller.rb
Hanaffi: So there're two solutions:
- Extract to service
- make `discount_code_props` as public method

I used service approach because it encapsulates the logic for selecting the best offer code between a URL-provided code and a product's default offer code. We use a dedicated service object rather than making the ProductPresenter::ProductProps#discount_code_props method public because it keeps business logic separate from presentation concerns. Services can be reused across controllers, background jobs, API endpoints, and other services without coupling them to view rendering. This also
maintains proper dependency direction where controllers don't depend on presenters for core
business operations, and makes the discount selection logic easier to test in isolation.

spec/controllers/links_controller_spec.rb
```diff
@@ -3186,6 +3186,95 @@ def files_data_from_urls(urls)
...
+            end
+
+            it "does not include code in checkout redirect" do
+              get :show, params: { id: product.to_param, wanted: "true", code: "INVALID" }
+
+              expect(response).to be_redirect
+              redirect_url = URI.parse(response.location)
+              query_params = Rack::Utils.parse_query(redirect_url.query)
+              expect(query_params["code"]).to be_nil
+            end
```
gianfrancopiana, line 3275: The `.except(:code, :offer_code)` fix in the controller correctly strips the original params before rebuilding the redirect URL – currently no test verifies this. 

The existing tests parse the redirect URL with `Rack::Utils.parse_query`, which returns a hash. So if `code` appeared twice in the query string, the hash silently keeps the last value and the test would still pass. 

Consider adding a test that sends a request with `wanted=true` and a `code` param, then checks the raw query string to assert `code` appears exactly once.

spec/controllers/links_controller_spec.rb
```diff
@@ -3186,6 +3186,95 @@ def files_data_from_urls(urls)
           expect(response).to be_successful
           expect(response).not_to be_redirect
         end
+
+        describe "discount code selection" do
+          let(:url_offer_code) { create(:offer_code, products: [product], code: "URL10", amount_cents: 200, currency_type: product.price_currency_type) }
```
gianfrancopiana, line 3191: The controller reads `params[:offer_code] || params[:code]`, but every test in this section only passes `code` as the param name. Gumroad's embed widgets and some older URLs use `offer_code` instead. 

We should add a test that sends the discount code via the `offer_code` param and asserts it's correctly picked up and compared against the default code.

spec/controllers/links_controller_spec.rb
Hanaffi: done but I think `parse_query` values will contain array(s) if a quey is repeated.
```rb
irb(main):009> require 'rack'
irb(main):010> 
irb(main):011> query_string = "code=rubyonrails&code=hackathon&page=1"
irb(main):012> query_hash = Rack::Utils.parse_query(query_string)
irb(main):013> 
irb(main):015> puts query_hash
{"code" => ["rubyonrails", "hackathon"], "page" => "1"}
```

Hanaffi:
@gianfrancopiana I applied the requested changes. Can you review again, please?

Issue #3283, opened 2026-01-31
Fix missing black border on "Library" button in Discover header


4arjun:
## Description
The "Library" button in the Discover page navigation bar is missing its border styling. It currently renders with a default white/transparent border which is invisible against the yellow background of the header.

## Current Behavior

<img width="1510" height="735" alt="Image" src="https://github.com/user-attachments/assets/ead992c1-1add-41ba-b395-0666139c947b" />

## Expected Behavior

<img width="1512" height="738" alt="Image" src="https://github.com/user-attachments/assets/afe39c59-ba7b-4d50-9b81-d01cadadc605" />

PR #3284, opened 2026-01-31
Fix missing border on Library button in Discover header

4arjun:
Issue: #3283 

## Summary

This PR fixes a visual regression where the "Library" button in the Discover page navigation bar was missing its border styling.
## Root Cause:
When a category page loads, getRootTaxonomyCss() applies inline styles to the header including `--color: 0 0 0`. However, the border-border Tailwind class resolves to `var(--color-border)`, which is defined in the `@theme` block as `rgb(var(--color) / var(--border-alpha))`. 
The issue is that `--color-border` is computed at the theme level and doesn't automatically re-evaluate when `--color` is overridden inline on a descendant element. As a result, buttons inside the colored header weren't picking up the correct border color.

## Before

Screenshot 2026-01-31 at 7 40 20 PM

## After

Screenshot 2026-02-01 at 9 06 15 AM

Screenshot 2026-02-01 at 9 05 58 AM

# Checklist

---

# AI Disclosure
No AI used


app/javascript/utils/discover.ts
```diff
@@ -7,7 +7,12 @@ const categoryImages = require.context("$assets/images/discover/");
 export type Taxonomy = { key: string; slug: string; label: string; parent_key: string | null };
 
 export function getRootTaxonomyCss(slug: RootTaxonomySlug) {
-  return { backgroundColor: `var(--${rootTaxonomies[slug].color})`, color: "black", "--color": "0 0 0" };
+  return {
+    backgroundColor: `var(--${rootTaxonomies[slug].color})`,
+    color: "black",
+    "--color": "0 0 0",
+    "--color-border": "rgb(0 0 0 / 1)",
+  };
```
4arjun, line 15: The `border-border` class uses `var(--color-border)`, which is defined at the theme level and doesn't automatically update when `--color` is overridden inline on the header. By explicitly setting `--color-border` alongside `--color`, child elements (like the Library button) now correctly inherit a visible black border on colored category backgrounds.

4arjun:

binary-koan (approved):
Nice find, thanks!

PR #3285, opened 2026-01-31
fix(checkout): Apply best discount code with ?wanted=true parameter (#3271)

pegasus1134:
Issue: #3271

## Problem

When a user visits a product URL with `?wanted=true`, the redirect to checkout does not apply the product's default offer code. Additionally, if both a URL code and a default code exist, there's no logic to pick the better discount — the URL code is blindly passed through.

## Solution

Add `Link#best_offer_code_for(url_code)` that compares the URL-provided offer code against the product's default offer code using the existing `OfferCode#amount_off` method, and returns whichever gives the higher discount. This keeps the logic DRY with `ProductPresenter::ProductProps#discount_code_props` which already does this comparison on the product page.

### Changes

**app/models/link.rb** — Added `best_offer_code_for(url_code)`:
- Compares URL code vs default code using `OfferCode#amount_off(price_cents)`
- Returns the code with the higher discount; ties go to default code (matches presenter behavior)
- Gracefully handles nil/missing/invalid codes

**app/controllers/links_controller.rb** — Updated `?wanted=true` redirect:
- Calls `@product.best_offer_code_for(...)` instead of passing through raw params
- Added `return` before `redirect_to` to prevent fall-through
- Strips `:code` and `:offer_code` from `params.permit!` spread to prevent duplicate query params

### Behavior

| Scenario | Result |
|----------|--------|
| No URL code, has default code | Default code applied |
| URL code gives bigger discount | URL code wins |
| Default code gives bigger discount | Default code wins |
| Equal discounts | Default code wins (matches presenter) |
| No default code, has URL code | URL code applied |
| Neither code exists | No code applied |
| Invalid URL code, has default | Default code applied |
| Deleted default code, has URL code | URL code applied |
| Fixed-amount vs percentage | Compares actual dollar amount off |

## Test Results

**21 examples, 0 failures** (14 model + 7 controller)

Screenshot 2026-02-07 at 1 44 21 PM

**Model tests** (`Link#best_offer_code_for`):
- Returns URL code when no default exists
- Returns default code when no URL code provided
- Returns nil when neither code exists
- Handles empty string URL code
- Returns default code for empty string when default exists
- Higher percentage discount wins
- URL code wins when it has higher discount
- Fixed-amount vs percentage comparison (percentage wins)
- Fixed-amount vs percentage comparison (fixed wins)
- Equal discounts → default wins
- Deleted default code → falls back to URL code
- Invalid URL code → falls back to default
- Invalid URL code with no default → passes through
- Works with `offer_code` param name

**Controller tests** (`LinksController wanted=true`):
- Default code applied when none provided
- URL code used when better discount
- Default code used when better discount
- No duplicate code params in redirect URL
- Works with `offer_code` param name
- No code param when neither exists
- Falls back to default when URL code is invalid

## Checklist

## AI Disclosure

**Model**: Claude Opus 4.5 via Claude Code CLI

**Used for:**
- Test execution and debugging factory setup
- Verifying offer code comparison logic

**What I verified myself:**
- All code follows existing codebase patterns
- Ran all 21 tests locally
- Reviewed every line of diff
- Verified redirect URLs contain correct `code=` param

### DIFF
--- a/spec/controllers/links_controller_spec.rb
+++ b/spec/controllers/links_controller_spec.rb
@@ -3224,6 +3224,53 @@ describe LinksController, :vcr, inertia: true do
           query_params = Rack::Utils.parse_query(URI.parse(response.location).query)
           expect(query_params["code"]).to eq(default_code.code)
         end
+
+        it "does not duplicate code params in the redirect URL" do
+          discountable_product = create(:product, user: @user, price_cents: 10_00)
+          offer_code = create(:offer_code, user: @user, products: [discountable_product], amount_cents: nil, amount_percentage: 25)
+          discountable_product.update!(default_offer_code: offer_code)
+
+          get :show, params: { id: discountable_product.to_param, wanted: "true", code: "someoldcode" }
+
+          query_string = URI.parse(response.location).query
+          code_occurrences = query_string.scan(/(?:^|&)code=/).length
+          expect(code_occurrences).to eq(1)
+        end
+
+        it "works with the offer_code param name" do
+          discountable_product = create(:product, user: @user, price_cents: 10_00)
+          default_code = create(:offer_code, code: "default50", user: @user, products: [discountable_product], amount_cents: nil, amount_percentage: 50)
+          discountable_product.update!(default_offer_code: default_code)
+          url_code = create(:offer_code, code: "url10", user: @user, products: [discountable_product], amount_cents: nil, amount_percentage: 10)
+
+          get :show, params: { id: discountable_product.to_param, wanted: "true", offer_code: url_code.code }
+
+          expect(response).to be_redirect
+          query_params = Rack::Utils.parse_query(URI.parse(response.location).query)
+          expect(query_params["code"]).to eq(default_code.code)
+        end
+
+        it "redirects without a code param when no codes exist" do
+          discountable_product = create(:product, user: @user, price_cents: 10_00)
+
+          get :show, params: { id: discountable_product.to_param, wanted: "true" }
+
+          expect(response).to be_redirect
+          query_params = Rack::Utils.parse_query(URI.parse(response.location).query)
+          expect(query_params["code"]).to be_nil
+        end
+
+        it "falls back to the default code when an invalid URL code is provided" do
+          discountable_product = create(:product, user: @user, price_cents: 10_00)
+          default_code = create(:offer_code, code: "valid25", user: @user, products: [discountable_product], amount_cents: nil, amount_percentage: 25)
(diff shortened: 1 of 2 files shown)


app/models/link.rb
```diff
@@ -813,6 +813,22 @@ def find_offer_code(code:)
...
+    url_code = url_code.presence
+    default_code = default_offer_code&.code
+
+    return url_code if default_code.blank?
+    return default_code if url_code.blank?
+
+    url_offer = find_offer_code(code: url_code)
+    default_offer = find_offer_code(code: default_code)
+
+    return url_code unless default_offer
```
pegasus1134, line 826: Reuses the existing `OfferCode#amount_off(price_cents)` for comparison — same logic that `ProductPresenter::ProductProps#discount_code_props` uses on the product page. Kept as a model method rather than a service since this is a simple model concern ("which of my offer codes gives the best discount?"). `>` means ties go to the default code, matching the presenter behavior.

app/controllers/links_controller.rb
```diff
@@ -128,13 +128,13 @@ def show
 
       unless (@product.customizable_price || cart_item[:option]&.[](:is_pwyw)) &&
              (params[:price].blank? || params[:price] < cart_item[:price])
-        redirect_to checkout_index_url(**params.permit!, host: DOMAIN, product: @product.unique_permalink,
```
pegasus1134, line 131: `.except(:code, :offer_code)` prevents duplicate `code` query params in the redirect URL — without this, `**params.permit\!` would include the original `code` param alongside the explicit `code:` kwarg.

pegasus1134, line 131: Added `return` before `redirect_to` to ensure the action halts after redirect. Without it, execution continues and may hit the render call below.

pegasus1134:
**Update (Feb 7):**

Added comprehensive edge case tests — now at **21 examples, 0 failures** (up from 10):

- Fixed-amount vs percentage discount comparison
- Deleted default code graceful fallback
- Invalid/nonexistent URL code fallback to default
- Empty string handling
- `offer_code` param name support
- No duplicate `code` params in redirect URL
- No code when neither exists

PR description updated with new test screenshot and full test list.

gianfrancopiana:
Thanks for your contribution. The best-discount comparison and duplicate param fix are both good catches. We're going with #3272 because it reuses the existing validation pipeline (`OfferCodeDiscountComputingService`) via a shared service rather than a parallel implementation on the model, which keeps the two code paths in sync. Closing in favor of #3272.

PR #3301, opened 2026-02-02
Inertia Migration: Edit Product Pages

### LINKED ISSUE #3030: Migrate Edit Product page to Inertia

sm17p:
# Closes: #3030
- #3030

# Description

This PR migrates the product edit flow to Inertia.  Product edit tab (edit/update), content edit tab, receipt edit tab, and share edit tab are now Inertia pages backed by `Products::MainController`, `Products::ContentController`, `Products::ReceiptController`, and `Products::ShareController`

# Solution

### Frontend Changes

1. Product edit pages and shared components no longer depend on React context for sharing data. Each component now receives exactly the props it needs to work.

2. Product edit pages use Inertia's `useForm` helper to manage form state

3. Follows Bundles Inertia migration pattern for consistency and keeps page state in sync with the server when moving between product edit tabs. If the form is in dirty state, navigating to another edit tab triggers a save so that common data is synced before the next page load.

### Backend Changes

1. Each tab gets only the product fields and metadata it needs (e.g. receipt tab: receipt text and view-content button; content tab: files and content settings). 

    - `Controllers/presenters` build these per-tab payloads. 
    - Some tabs support partial reloads so they can refresh specific fields without reloading the whole page.

2. Centralized update logic in `Product::UpdateService`. All product-tab updates (main, content, receipt, share) go through `Product::UpdateService`, which updates, publish/unpublish, in one place.

3. Few of the internal API endpoints, used to load tab specific data are now passed as props like `receipt_preview_html` (receipt tab), and `existing_files` (content tab).

4. For failed form submissions we use `status: :found (302)` to populate error props (required as 303 sends empty error object) which is then automatically converted to 303 by [Inertia](https://inertia-rails.dev/guide/redirects#_303-response-code)

5. Backend now supports two new flash cases, 

    - Flash messages that are HTML (message starts with `<`and ends with `>`) are passed through with `html: true` in the shared flash props so the client can render them as HTML. This is used in `Product::UpdateService` when publishing fails (Eg: - empty seller email), it sets `flash[:alert]` to an HTML string with a link. 
    - Complex client-side alerts – When `flash[:inertia]` is set with a status that starts with "frontend_alert" and optional data, the backend sends `{ status, data }` in the shared flash. The client can branch on status and data to show custom UI (e.g. toasts, modals, or inline messages). This is used in content tab where after saving, backend sends a “contents updated” alert with a link so the seller can notify buyers. The content edit page reads that flash and shows the alert with the link.
    
    
## Demos (After)

> [!NOTE]
> I've skipped before videos since there were too many cases to cover over here. Existing specs are passing though so lmk if you require any specific one
>
> Line additions in this PR is more than removals because there is an increase in coverage due to new service additions, movement of frontend logic to backend, merger of publish/unpublish actions with update, offer code validations and re-direction logic per controller. Also, presenter props are split into product and page specific metadata which need to be tested separately.

### 1. General Navigation With Inertia

https://github.com/user-attachments/assets/90e9e6b3-282d-4db6-81ac-f463b882aa60

### 2. Edit Product

https://github.com/user-attachments/assets/b05f31dd-77ea-497c-b89d-7e7a3e55e51d

### 3. Edit Content

https://github.com/user-attachments/assets/c880a0d6-c668-4b9c-b626-39fa49346a5a

### 4. Edit Receipt

https://github.com/user-attachments/assets/f0e8f15d-a8e0-4a06-9cb9-66f6b976335f

### 5. Edit Share

https://github.com/user-attachments/assets/68a18a12-6317-48f9-81e9-ea67f1f88afe

### 6. Unpublish (Other Products vs Coffee Product)

https://github.com/user-attachments/assets/89e71e61-3e44-4216-abc1-6aee081bc71b

### 7. Merge Rich Content

https://github.com/user-attachments/assets/1370086b-122a-4d0d-a8f6-5c8dedd46085

### 8. Share Page Blocked For Unpublished Products

https://github.com/user-attachments/assets/9b9a590d-783f-4caa-88c9-8faffdd6c314

### 9. ISBN Validation

https://github.com/user-attachments/assets/5445149b-e042-47ad-bbda-4bd1fa90fa52

### 10. Split Rich Content By Variant

https://github.com/user-attachments/assets/81a1fe62-5e4e-4bcd-aabb-4273d6543a92

---

### Request Specs

request-specs-passing

### Other Specs 

One flaky elastic search spec

`bundle exec rspec spec/controllers/links_controller_spec.rb spec/controllers/products/content_controller_spec.rb spec/controllers/products/main_controller_spec.rb spec/controllers/products/receipt_controller_spec.rb spec/controllers/products/share_controller_spec.rb spec/mailers/contacting_creator_mailer_spec.rb spec/presenters/collab_products_page_presenter_spec.rb spec/presenters/dashboard_products_page_presenter_spec.rb spec/presenters/product_presenter_spec.rb spec/services/product/rich_content_updater_service_spec.rb spec/services/product/variant_category_updater_service_spec.rb spec/services/product/variants_updater_service_spec.rb`

other-specs

---

# Checklist

---

# AI Disclosure

Cursor Auto was used for drafting a plan, edits, debugging and documentation checks


app/controllers/api/internal/existing_product_files_controller.rb
sm17p, line 1: Moved to `existing_files` key in `Products::ContentController#edit`

app/controllers/concerns/inertia_rendering.rb
```diff
@@ -20,14 +20,22 @@ module InertiaRendering
...
-      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success", html: flash_message.to_s.start_with?("<") && flash_message.to_s.end_with?(">") }
     end
 
-    def inertia_errors(model)
+    def inertia_errors(model, model_name: nil)
+      prefix = model_name.presence || model.model_name.element
       { errors: model.errors.to_hash.each_with_object({}) do |(key, messages), hash|
-        hash["#{model.model_name.element}.#{key}"] = messages.to_sentence
+        hash["#{prefix}.#{key}"] = messages.to_sentence
```
sm17p, line 38: Using model name to override default since frontend uses `product` for forms and model is `link`

app/controllers/concerns/inertia_rendering.rb
```diff
@@ -20,14 +20,22 @@ module InertiaRendering
...
+        inertia_flash = flash[:inertia].to_h.with_indifferent_access
+        if inertia_flash[:status].present? && inertia_flash[:status].start_with?("frontend_alert")
+          return { status: inertia_flash[:status], data: inertia_flash[:data] || {} }
+        end
+      end
+
       return if (flash_message = flash[:alert] || flash[:warning] || flash[:notice]).blank?
 
-      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success", html: flash_message.to_s.start_with?("<") && flash_message.to_s.end_with?(">") }
```
sm17p, line 32: Publish action flash alert using this path 

``` ruby

`"<span>To publish a product, we need you to have an email. <a href=\"#{settings_main_url}\">Set an email</a> to continue.</span>"` 

```

app/controllers/concerns/inertia_rendering.rb
```diff
@@ -20,14 +20,22 @@ module InertiaRendering
 
   private
     def inertia_flash_props
+      if flash[:inertia].present?
+        inertia_flash = flash[:inertia].to_h.with_indifferent_access
+        if inertia_flash[:status].present? && inertia_flash[:status].start_with?("frontend_alert")
+          return { status: inertia_flash[:status], data: inertia_flash[:data] || {} }
+        end
+      end
```
sm17p, line 28: We've moved  flashing logic for content changes to backend.

Backend can now send props specific to alerts guarded by `start_with?("frontend_alert")` prefix for creating dynamic complex frontend flashes 

------

Image

app/controllers/products/base_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  before_action :authorize_product
+  before_action :redirect_if_bundle, only: [:edit]
+  before_action :set_default_page_title
+
+  layout "inertia"
+
+  private
+    def permitted_next_url
+      params.permit(:next_url)[:next_url].presence
+    end
```
sm17p, line 18: For switching b/w edit tabs, follows bundle migration pattern to save unsaved changes

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class Products::ContentController < Products::BaseController
+  def edit
+    render inertia: "Products/Content/Edit", props: {
+      product: -> { edit_product_content_presenter.edit_product_content },
+      existing_files: -> { edit_product_content_presenter.existing_files },
```
sm17p: New prop after removing internal API endpoint

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    notice = case status
+             when :published then "Published!"
+             when :unpublished then "Unpublished!"
+             else "Changes saved!"
+    end
+
+    redirect_path = permitted_next_url || (status == :published ? edit_product_share_path(product) : edit_product_content_path(product))
+    flash[:inertia] = { status: "frontend_alert_contents_updated", data: notify_customers_link_payload } if status == :content_updated
+    redirect_to redirect_path, status: :see_other, **(warning.present? ? { warning: } : { notice: })
+  end
```
sm17p, line 30: Overall, the pattern is same for all the update flows.

1. Controller calls the service with page specific params, builds flash and redirects to the next page

2. Warming message has a higher priority according to the old update logic

3. `permitted_next_url` for switching b/w edit tabs, follows bundle migration pattern to update unsaved data before route change (I think this was done to keep the shared data in sync for the next tab)

app/controllers/products/receipt_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class Products::ReceiptController < Products::BaseController
+  def edit
+    render inertia: "Products/Receipt/Edit", props: {
+      product: -> { edit_product_receipt_presenter.edit_product_receipt },
+      page_metadata: -> { edit_product_receipt_presenter.edit_product_receipt_metadata },
+      receipt_preview_html: -> { ReceiptPreviewRendererService.new(product: @product, custom_receipt_text: params[:custom_receipt_text], custom_view_content_button_text: params[:custom_view_content_button_text], use_defaults: request.headers["HTTP_X_INERTIA_PARTIAL_COMPONENT"].blank? || request.headers["HTTP_X_INERTIA_PARTIAL_COMPONENT"] != "Products/Receipt/Edit").perform },
```
sm17p, line 8: 1. New prop after removing internal API.

2. `use_defaults` is for loading preview with current product data. Navigting from another tab or first load happens w/o the params so we need to show the default

app/controllers/products/main_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      page_metadata: edit_product_main_presenter.edit_product_metadata,
+    }
+  end
+
+  def update
+    product, warning, status, success = Product::UpdateService.new(
+      product: @product,
+      params: params.require(:product).permit(policy(@product).product_tab_permitted_attributes),
+      current_seller:,
+    ).process(:product_tab, disallow_publish: true).values_at(:product, :warning, :status, :success)
```
sm17p, line 16: `disallow_publish: true` since publish action is disabled from product edit and share edit page

app/controllers/links_controller.rb
```diff
@@ -105,9 +105,9 @@ def create
 
     create_user_event("add_product")
     if ai_generated
-      redirect_to edit_link_path(@product, ai_generated: true), status: :see_other
+      redirect_to edit_product_path(@product, ai_generated: true), status: :see_other
```
sm17p, line 107: [Clean up] Old route helper `edit_link_path` -> `edit_product_path`

app/controllers/links_controller.rb
```diff
@@ -275,175 +275,6 @@ def track_user_action
...
-      @product.publish!
-    rescue Link::LinkInvalid, ActiveRecord::RecordInvalid
-      return render json: { success: false, error_message: @product.errors.full_messages[0] }
-    rescue => e
-      Bugsnag.notify(e)
-      return render json: { success: false, error_message: "Something broke. We're looking into what happened. Sorry about this!" }
-    end
-
-    render json: { success: true }
-  end
```
sm17p, line 470: Follows bundle inertia migration and is now handled inside `update` for each tab

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -52,10 +52,14 @@ export const getDraggedFileEmbed = (editor: Editor) => {
   return draggedNode?.type.name === FileEmbed.name ? draggedNode : null;
 };
 
-const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewProps) => {
-  if (!node.attrs.id) return;
```
sm17p: Moved towards the end since it blocks react compiler from auto memoizing the component (Compiler can't determine order of hooks because of early return)

```
🚫 This component hasn't been memoized by React Compiler.

Reason: Hooks must always be called in a consistent order, and may not be called conditionally. See the Rules of Hooks [https://react.dev/warnings/invalid-hook-call-warning
```

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -353,11 +359,11 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
           })
           .run();
       }}
-      className={cx({ "relative rounded-sm border border-dashed border-accent": isDropZone })}
+      className={classNames({ "relative rounded-sm border border-dashed border-accent": isDropZone })}
```
sm17p, line 363: [Clean up] Using the `classNames` utils over here now

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -92,26 +88,74 @@ export const extensions = (productId: string, extraExtensions: TiptapNode[] = []
...
-  const pagesRef = useRefToLatest(pages);
-  const updatePages = (pages: Page[]) =>
-    updateProduct((product) => {
-      if (selectedVariant) selectedVariant.rich_content = pages;
-      else {
-        product.has_same_rich_content_for_all_variants = true;
-        product.rich_content = pages;
+    : variants.find((variant) => variant.id === selectedVariantId);
+  const pages: (Page & { chosen?: boolean })[] = selectedVariant ? selectedVariant.rich_content : richContent;
+  const pagesRef = useLatest(pages);
```
sm17p: [NEW[ `useLatest` instead of `useRefToLatest` to avoid a lag in updated value

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -960,207 +1040,14 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
...
-          productId={id}
+          productId={productId}
         />
       ) : null}
     </>
   );
 };
-
-//TODO inline this once all the crazy providers are gone
-export const ContentTab = () => {
```
sm17p, line 1036: Inlined into content edit page

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -198,24 +267,48 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
...
-  const fileEmbedConfig = useRefToLatest<FileEmbedConfig>({ filesById });
-  const uploadFilesRef = useRefToLatest(uploadFiles);
-  const contentEditorExtensions = extensions(id, [
+
+  const fileEmbedConfig = useLatest<FileEmbedConfig>({
+    get filesById() {
+      return filesByIdRef.current;
+    },
+    id: productId,
+    onUpdateFile: updateFile,
```
sm17p: Passing necessary props after removal of shared context

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -16,167 +12,73 @@ import { Preview } from "$app/components/Preview";
...
   return product.native_type === "coffee" && currentSeller
     ? Routes.custom_domain_coffee_url({ host: currentSeller.subdomain, ...params })
-    : Routes.short_link_url(product.custom_permalink ?? uniquePermalink, {
+    : Routes.short_link_url(product.custom_permalink ?? product.unique_permalink, {
         host: currentSeller?.subdomain ?? appDomain,
         ...params,
       });
 };
 
-const NotifyAboutProductUpdatesAlert = () => {
```
sm17p, line 41: Moved to content edit tab, this is the frontend flashing case mentioned in the PR

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -197,43 +99,55 @@ export const Layout = ({
...
-    callback?.();
+    const url = new URL(event.target.href);
+    if (url.pathname.startsWith(Routes.edit_product_share_path(product.unique_permalink)) && !product.is_published) {
+      event.preventDefault();
+      showAlert(
+        "Not yet! You've got to publish your awesome product before you can share it with your audience and the world.",
+        "warning",
+      );
+      return;
+    }
```
sm17p, line 148: Although backend handles this case, I've kept it in frontend to avoid a network request and to keep re-direction logic simple

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -197,43 +99,55 @@ export const Layout = ({
...
+
+    if (!isFormDirty) return;
+    event.preventDefault();
+    setSaveFlowStatus("saving");
+    save({ onFinish: () => setSaveFlowStatus(null) }, { next_url: event.target.href });
```
sm17p, line 153: Follows bundle inertia pattern, required to keep state in sync for the next tab

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -247,23 +161,21 @@ export const Layout = ({
                 </Button>
               </CopyToClipboard>
             </>
-          ) : tab === "product" && !isCoffee ? (
-            <Button
-              color="primary"
-              disabled={isBusy}
-              onClick={() => void save().then(() => navigate.current(`${rootPath}/content`))}
```
sm17p, line 271: Re-direction handled by controller now

app/javascript/components/ProductEdit/ReceiptPreview.tsx
```diff
@@ -1,46 +1,32 @@
...
-        url,
-        accept: "html",
-      });
 
-      const html = await response.text();
-      setReceiptHtml(html);
-    } catch (error) {
-      assertResponseError(error);
-      setReceiptHtml("Error loading receipt preview");
-    }
```
sm17p, line 37: Part of page props now

app/javascript/components/ProductEdit/state.ts
```diff
@@ -170,54 +174,13 @@ export type ContentUpdates = {
   uniquePermalinkOrVariantIds: string[];
 } | null;
 
-export const ProductEditContext = React.createContext<{
```
sm17p, line 173: Removed context, all the data is passed through props. 

If we ever need to do this we can just use `usePage().props` instead inside the nested component

app/javascript/components/ProductsPage/Collabs/MembershipsTable.tsx
```diff
@@ -68,15 +68,22 @@ export const CollabsMembershipsTable = (props: { entries: Membership[]; paginati
...
-                </a>
+                {membership.can_edit ? (
+                  <Link href={membership.edit_url} style={{ textDecoration: "none" }}>
+                    <h4 className="font-bold">{membership.name}</h4>
+                  </Link>
+                ) : (
+                  <a href={membership.url} style={{ textDecoration: "none" }}>
+                    <h4 className="font-bold">{membership.name}</h4>
+                  </a>
+                )}
```
sm17p, line 86: We can now use Inertia `Link` for product edit pages

app/javascript/components/useDropboxDropins.ts
sm17p, line 1: This loads the dropbox script which we had earlier in erb `<%= render("shared/load_dropbox_dropins_js", async: true) %>`

app/javascript/components/useFlashMessage.ts
```diff
@@ -1,13 +1,54 @@
...
   }, [flash]);
 }
+
+type StatusToPayload<S extends ClientSideAlertPayloads["status"]> = Extract<ClientSideAlertPayloads, { status: S }>;
+
+const TYPE_GUARDED_PAYLOADS: {
+  [K in ClientSideAlertPayloads["status"]]: ReturnType<typeof createIs<StatusToPayload<K>>>;
+} = {
+  frontend_alert_contents_updated: createIs<ContentsUpdatedAlertPayload>(),
+};
```
sm17p, line 29: @Emcousin @Pradumn27 We can now create complex flash messages on frontend instead of relying on `HTML`. 

1. This typeguard checks the payload during runtime with `createIs`. 
2. Only the listed keys can be used as flash

app/javascript/components/useFlashMessage.ts
```diff
@@ -1,13 +1,54 @@
...
+
+export const useClientFlashMesage = <S extends ClientSideAlertPayloads["status"]>(
+  status: S,
+): {
+  flashData: StatusToPayload<S> | null;
```
sm17p, line 34: For development protection, only the listed flash types can be used to prevent accidental key mismatch.

Eg:- `useClientFlashMesage('new_flash')` will raise a type error unless it's present in `ClientSideAlertPayloads`

app/javascript/components/useRefToLatest.ts
```diff
@@ -9,3 +9,9 @@ export const useRefToLatest = <T>(value: T): React.MutableRefObject<T> => {
   }, [value]);
   return ref;
 };
```
sm17p, line 11: This one lags in update for a brief moment because of the effect, using the other hook to enure we don't have stale references for state updates

app/javascript/data/product_edit.ts
```diff
@@ -1,48 +0,0 @@
...
-  const fileIds = new Set(
-    richContents.flatMap((content) =>
-      findChildren(
-        editor.schema.nodeFromJSON(content.description),
-        (node) => node.type.name === FileEmbed.name,
-      ).map<unknown>((child) => child.node.attrs.id),
-    ),
-  );
-  editor.destroy();
-  product.files = product.files.filter((file) => fileIds.has(file.id));
```
sm17p, line 28: These steps are specific to content tab and moved there

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+      avatar_url: string;
+      profile_url: string;
+    };
+    dropbox_picker_app_key: string;
+  };
+};
+
+const NotifyAboutProductUpdatesAlert = () => {
+  const timerRef = React.useRef<number | null>(null);
+  const { flashData, clear } = useClientFlashMesage("frontend_alert_contents_updated");
```
sm17p, line 59: [NEW] Flash case which gets data from backend

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+    user_id: page_metadata.seller.id,
+  });
+
+  const loadedPostsData = React.useRef(
+    new Map<string | null, { posts: Post[]; total: number; next_page: number | null }>(),
+  );
+  const [loadingPostsCount, setLoadingPostsCount] = React.useState(0);
+  const postsDataForEditingId = loadedPostsData.current.get(selectedVariantId);
+
+  const fetchMorePosts = async (refresh?: boolean) => {
```
sm17p, line 169: Left this one for now as the PR was getting bigger and bigger

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+              ...(sub_size != null && { size: sub_size }),
+            })),
+          })),
+          variants: data.product.variants.map((variant) => ({
+            id: variant.id,
+            rich_content: variant.rich_content,
+          })),
+          publish: saveData?.publish,
+        },
+        ...(saveData?.next_url && { next_url: saveData.next_url }),
```
sm17p, line 255: Common behaviour to all forms, only sending the required fields per tab now

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+      };
+    });
+
+    form.patch(Routes.product_content_path(product.unique_permalink), {
+      only: ["product", "errors", "flash", "existing_files"],
+      onSuccess: (data) => {
+        options?.onSuccess?.();
+        options?.onFinish?.();
+        form.setData("product", cast<EditProductContent>(data.props.product));
+        form.setDefaults();
```
sm17p: For newly added content, we need this to sync server generated ids with the form. 

This ensures that the next save call is sent with the server generated id and not local client ids which avoids thrashing aslready saved files

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+        setSyncRichContentIdsFromServer((count) => count + 1);
```
sm17p: Refresh editor when server ids are synced to avoid stale file embed references in the editor

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -0,0 +1,467 @@
...
+        },
+        ...(saveData?.next_url && { next_url: saveData.next_url }),
+      };
+    });
+
+    form.patch(Routes.product_content_path(product.unique_permalink), {
+      only: ["product", "errors", "flash", "existing_files"],
+      onSuccess: (data) => {
+        options?.onSuccess?.();
+        options?.onFinish?.();
```
sm17p, line 263: Had to do this because inertia's `onFinish` callback always runs after `onSuccess` and since we update states inside `options?.onFinish?.()` we can end up updating with stale data from that closure


The below version would break when a new file embed is saved. The next render would make the embed invisible

```tsx
onSuccess: (data) => {
        options?.onSuccess?.();
        form.setData("product", cast<EditProductContent>(data.props.product));
        form.setDefaults();
        setSyncRichContentIdsFromServer((count) => count + 1);
      },
      ...(options?.onFinish && { onFinish: options.onFinish }),
```

app/javascript/pages/Products/Receipt/Edit.tsx
```diff
@@ -0,0 +1,82 @@
...
+        custom_view_content_button_text: data.product.custom_view_content_button_text,
+        publish: saveData?.publish,
+      },
+      ...(saveData?.next_url && { next_url: saveData.next_url }),
+    }));
+
+    form.patch(Routes.product_receipt_path(product.unique_permalink), {
+      only: ["product", "errors", "flash"],
+      ...(options?.onSuccess && { onSuccess: options.onSuccess }),
+      ...(options?.onFinish && { onFinish: options.onFinish }),
```
sm17p, line 40: No server generated id to sync over here

app/javascript/pages/Products/Share/Edit.tsx
```diff
@@ -0,0 +1,211 @@
...
+        is_adult: data.product.is_adult,
+        publish: saveData?.publish,
+      },
+      ...(saveData?.next_url && { next_url: saveData.next_url }),
+    }));
+
+    form.patch(Routes.product_share_path(product.unique_permalink), {
+      only: ["product", "errors", "flash"],
+      ...(options?.onSuccess && { onSuccess: options.onSuccess }),
+      ...(options?.onFinish && { onFinish: options.onFinish }),
```
sm17p, line 74: No server generated id to sync over here

app/javascript/pages/Products/Edit.tsx
```diff
@@ -0,0 +1,669 @@
...
+      },
+      ...(saveData?.next_url && { next_url: saveData.next_url }),
+    }));
+
+    form.patch(Routes.product_path(form.data.product.unique_permalink), {
+      only: ["product", "errors", "flash"],
+      preserveState: true,
+      onSuccess: (data) => {
+        options?.onSuccess?.();
+        options?.onFinish?.();
```
sm17p, line 130: Need to sync server generated ids for variants and other fields

app/javascript/pages/Products/Edit.tsx
```diff
@@ -0,0 +1,669 @@
...
+        try {
+          form.setData("product", cast<EditProduct>(data.props.product));
+          form.setDefaults();
+        } catch {
+          // Re-directing to another tab type mismatch
```
sm17p: (Self note: This is bound to raise an error when `save and continue` is clicked, re-check the flows from. What I remember is this causes issue with flashing I think)

app/javascript/types/inertia-flash.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+type ClientSideAlertPayloads = ContentsUpdatedAlertPayload;
+
+export type FlashData = AlertPayload | ClientSideAlertPayloads | null;
+
+declare module "@inertiajs/core" {
+  export interface InertiaConfig {
+    flashDataType: FlashData;
+  }
+}
```
sm17p, line 12: Sets avalaible flash types globally for use

app/models/product_file.rb
```diff
@@ -195,6 +195,7 @@ def save_subtitle_files!(files)
       found = files.extract! { _1[:url] == file.url }.first
       if found
         file.language = found[:language]
+        file.size = found[:size]
```
sm17p, line 198: Save the file size sent from frontend (there's a worker which re-calculates it too). 

A spec asserts on the file size after save and since we're now re-rendering with server state after save we need it to be set

app/services/product/rich_content_updater_service.rb
sm17p, line 1: Extracted common rich content update logic over here

app/services/product/rich_content_updater_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      rich_content.update!(
+        title: node[:title].presence,
+        description: node[:description].presence || [],
+        position: index
+      )
+      rich_contents_to_keep << rich_content
+    end
+    to_mark_deleted = existing_rich_contents - rich_contents_to_keep
+    to_mark_deleted.each(&:mark_deleted!)
+    { content_updated: to_mark_deleted.any? || rich_contents_to_keep.any? { _1.previous_changes.key?("id") || _1.previous_changes.key?("title") || _1.previous_changes.key?("description") || _1.previous_changes.key?("position") }, product: }
```
sm17p, line 30: We set the frontend flash whenever these values are updated

app/services/product/update_service.rb
```diff
@@ -0,0 +1,327 @@
...
+      rescue ActiveRecord::RecordNotSaved, ActiveRecord::RecordInvalid, Link::LinkInvalid => e
+        if intent == :content_tab && @product.errors.details[:custom_fields].present?
+          message = "You must add titles to all of your inputs"
+        else
+          message = @product.errors.full_messages.first
+        end
+        @product.errors.add(:base, message || e.message)
+        return { success: false, status: nil, product: @product }
+      rescue => e
+        @product.errors.add(:base, "Something broke. We're looking into what happened. Sorry about this!")
```
sm17p, line 41: Rescue block from `LinksController#publish`

app/services/product/update_service.rb
```diff
@@ -0,0 +1,327 @@
...
+
+      def was_published
+        !@product.draft && @product.alive?
+      end
+
+      def update_product_publish_status!(disallow_publish = false)
+        return unless @params.key?(:publish)
+        return if @params[:publish] == was_published
+        raise Link::LinkInvalid, "You cannot publish this product yet." if disallow_publish && @params[:publish]
+        raise Link::LinkInvalid, "<span>To publish a product, we need you to have an email. <a href=\"#{settings_main_url}\">Set an email</a> to continue.</span>" if @params[:publish] && @product.user.email.blank?
```
sm17p, line 196: HTML flash case

app/services/product/update_service.rb
```diff
@@ -0,0 +1,322 @@
...
+        if has_currency_issues && has_amount_issues
+          issue_description = "#{"has".pluralize(all_invalid_offer_codes.count)} currency mismatches or would discount this product below #{@product.min_price_formatted}"
+        elsif has_currency_issues
+          issue_description = "#{"has".pluralize(all_invalid_offer_codes.count)} currency #{"mismatch".pluralize(all_invalid_offer_codes.count)} with this product"
+        else
+          issue_description = "#{all_invalid_offer_codes.count > 1 ? "discount" : "discounts"} this product below #{@product.min_price_formatted}, but not to #{MoneyFormatter.format(0, @product.price_currency_type.to_sym, no_cents_if_whole: true, symbol: true)}"
+        end
+
+        return { success: true, status: determine_status(was_published_before), product: @product, warning: "The following offer #{"code".pluralize(all_invalid_offer_codes.count)} #{issue_description}: #{all_invalid_offer_codes.join(", ")}. Please update #{all_invalid_offer_codes.length > 1 ? "them or they" : "it or it"} will not work at checkout." }
+      end
```
sm17p, line 65: Centralized logic based on intent, service handles business logic and provide the controller with necessray data for processing

app/services/product/variant_category_updater_service.rb
```diff
@@ -27,9 +27,19 @@ class Product::VariantCategoryUpdaterService
...
 
+  IGNORE_UNSET_ATTRIBUTES = %i[
+    name
+    duration_in_minutes
+    price_difference_cents
+    max_purchase_count
+    position_in_category
+    customizable_price
+  ].freeze
+
```
sm17p, line 38: These values belong to the product tab and we want to clear these when only explicitly asked for (Eg:- Content tab shouldn't update these , these keys are not present and hence we should skip updating them)

app/services/product/variant_category_updater_service.rb
```diff
@@ -134,20 +146,15 @@ def save_integrations(variant, option)
     end
 
     def save_rich_content(variant, option)
+      return unless option.key?(:rich_content)
```
sm17p, line 149: [IMPORTANT] Product tab doesn't  update`rich_content` field, without this we risk deleting it when update is called from product tab data (since it doesn't send `rich_content`key anymore)

app/services/receipt_preview_renderer_service.rb
```diff
@@ -1,21 +1,21 @@
 # frozen_string_literal: true
 
-class Api::Internal::ReceiptPreviewsController < Api::Internal::BaseController
```
sm17p, line 3: This API is now a prop and hence moved to a service

app/services/receipt_preview_renderer_service.rb
```diff
@@ -1,21 +1,21 @@
...
+    @product.custom_view_content_button_text = @custom_view_content_button_text unless @use_defaults
 
-    @product.custom_receipt_text = params[:custom_receipt_text]
-    @product.custom_view_content_button_text = params[:custom_view_content_button_text]
     purchase_preview = build_purchase_preview
 
     unless purchase_preview.valid?
-      error_message = purchase_preview.errors.full_messages.join(", ")
-      return render html: "Error: #{error_message}", status: :unprocessable_entity
+      return "<div>Error loading receipt preview</div>".html_safe
```
sm17p, line 18: This is what we used to display on frontend on error

spec/controllers/products/content_controller_spec.rb
```diff
@@ -0,0 +1,1217 @@
...
+    it_behaves_like "collaborator can access", :patch, :update do
+      let(:request_format) { :json }
+      let(:request_params) { @params }
+      let(:response_status) { 303 }
+    end
+
+    it_behaves_like "a product with offer code amount issues" do
+      let(:request_params) { @params }
+      let(:redirect_path) { edit_product_content_path(product.unique_permalink) }
+    end
```
sm17p, line 106: Common to all controllers and hence a shared example

spec/controllers/products/content_controller_spec.rb
```diff
@@ -0,0 +1,1217 @@
...
+
+    context "when publishing" do
+      it_behaves_like "publishing a product" do
+        let(:request_params) { @params }
+        let(:publish_failure_redirect_path_for_product) { edit_product_content_path(product.unique_permalink) }
+        let(:publish_failure_redirect_path_for_unpublished_product) { edit_product_content_path(unpublished_product.unique_permalink) }
+      end
```
sm17p, line 113: Common to contents and receipt tab

spec/controllers/products/content_controller_spec.rb
```diff
@@ -0,0 +1,1217 @@
...
+        product_without_files.reload
+        expect(product_without_files.purchase_disabled_at).to be_nil
+        expect(product_without_files.alive_product_files.count).to eq(0)
+      end
+    end
+
+    it_behaves_like "unpublishing a product" do
+      let(:request_params) { @params }
+      let(:unpublish_redirect_path) { edit_product_content_path(product.unique_permalink) }
+    end
```
sm17p, line 132: Common to all

spec/controllers/products/content_controller_spec.rb
sm17p, line 1: Overall, for all the controller specs, I've maintained the original blocks from links update specs and added inertia specific assertions

spec/controllers/links_controller_spec.rb
```diff
@@ -42,2694 +43,57 @@ def e404_test(action)
...
-          it "does not enable community chat" do
-            post :update, params: { id: @product.unique_permalink, community_chat_enabled: true }, as: :json
-
-            expect(response).to be_successful
-            expect(@product.reload.community_chat_enabled?).to be(false)
-            expect(@product.reload.active_community).to be_nil
-          end
-        end
-      end
+      e404_test(:destroy)
```
sm17p, line 96: Only test destroy since `edit`, `publish` and `unpublish` actions were moved

spec/shared_examples/product_edit.rb
```diff
@@ -0,0 +1,121 @@
...
+    let(:failed_publication_product) { product }
+
+    before do
+      product.unpublish!
+      allow_any_instance_of(Link).to receive(:publish!).and_raise("error")
+    end
+
+    it_behaves_like "a failed publication" do
+      let(:failed_publication_alert_message) { "Something broke. We're looking into what happened. Sorry about this!" }
+      before { allow(Bugsnag).to receive(:notify).once }
```
sm17p, line 105: Assertion from previous `publish` action specs

spec/support/fixtures/vcr_cassettes/LinksController/within_seller_area/GET_test_custom_delivery_url/returns_true_if_the_url_is_valid_and_reachable.yml
sm17p, line 1: Dead file

spec/support/fixtures/vcr_cassettes/LinksController/within_seller_area/PUT_update/creates_an_embeddable_preview.yml
sm17p, line 1: Dead file

spec/support/fixtures/vcr_cassettes/LinksController/within_seller_area/PUT_update/returns_a_list_of_active_previews.yml
sm17p, line 1: Dead file

spec/requests/products/edit/rich_text_editor_spec.rb
```diff
@@ -84,12 +84,12 @@
     # select_tab "Content"
     # expect(page).to have_alert(text: "Some images are still uploading, please wait...")
 
-    expect(page).to have_current_path(edit_link_path(@product))
+    expect(page).to have_current_path(edit_product_path(@product.unique_permalink))
```
sm17p, line 87: Making it `have_current_path(edit_product_path(@product.unique_permalink)` instead of `have_current_path(edit_product_path(@product))` as it explicitly states the url should contain a permalink (Serves as dcoumentation too)

spec/requests/products/edit/publishing_spec.rb
```diff
@@ -17,12 +17,13 @@
...
       create(:product_review, purchase: create(:purchase, link: @product))
       create(:merchant_account_stripe_connect, user: seller)
-      visit edit_link_path(@product.unique_permalink)
+      visit edit_product_path(@product)
     end
 
     it("allows user to publish and unpublish product") do
       click_on "Unpublish"
       wait_for_ajax
+      expect(page).to have_alert("Unpublished!")
```
sm17p, line 26: [NEW] This was never asserted against in request specs so added it over here

spec/requests/products/edit/options_spec.rb
```diff
@@ -7,12 +7,7 @@
...
 
   def visit_product_edit(link)
-    visit("/products/#{link.unique_permalink}/edit")
-    wait_for_ajax
-  end
-
-  def visit_product_edit_checkout_tab(link)
-    visit("/products/#{link.unique_permalink}/edit#checkout")
+    visit edit_product_path(link)
     wait_for_ajax
```
sm17p, line 16: `visit_product_edit_checkout_tab` is dead code

spec/requests/products/edit/file_embeds_spec.rb
```diff
@@ -614,7 +614,7 @@
 
     product.reload
     product_file_ids = product.alive_product_files.map(&:external_id)
-    expect(product_file_ids.size).to eq(2)
+    expect(product_file_ids.size).to eq(1)
```
sm17p, line 617: @EmCousin I believe this was wrong earlier. Re-adding a file in another page should not increase it's count since it's a reference to original?

spec/requests/products/edit/edit_spec.rb
```diff
@@ -1111,16 +1112,16 @@
...
-        editor = find("[aria-label='Content editor']")
-        set_rich_text_editor_input(editor, to_text: "Hi there!")
+
+        set_rich_text_editor_input(find("[aria-label='Content editor']"), to_text: "Hi there!")
 
         click_on "Save changes"
         expect(page).to have_alert(text: "Changes saved!")
 
-        set_rich_text_editor_input(editor, to_text: "New content")
+        set_rich_text_editor_input(find("[aria-label='Content editor']"), to_text: "New content")
```
sm17p, line 1138: Because of previous save action, server generated ids are synced on client and editor is reloaded.  The earlier assertion creted a stale reference to editor and hence the change over here

sm17p (commented):
@EmCousin @Pradumn27 I've added my review comments. Please have a look as request and controller specs are passing too 🙏

I've skipped before videos since there were too many cases to cover over here. Existing specs are passing though so lmk if you require any specific one

Pradumn27:
@sm17p please resolve the conflicts

sm17p:
@Pradumn27 done, please have a look again

EmCousin:
@sm17p that's quite a PR! I'll run CI now before testing myself and doing a deeper review.

EmCousin:
@sm17p a lot of tests are failing. Can you check?

sm17p:
@EmCousin Something is odd as the failing specs in CI passing locally. Let me go through all of them

sm17p:
### Failing specs from CI passing locally

Screenshot 2026-02-13 at 6 58 32 PM

Screenshot 2026-02-13 at 6 57 06 PM

@EmCousin Can you please run the CI again? Most of the specs were passing locally so something is odd

sm17p:
Checking again

sm17p:
@EmCousin I might need your help over here. Can you please check if you are able to reproduce the CI failures locally?

I'm sharing my local run over here for `spec/requests/products/edit/file_embeds_spec.rb` which had failures in CI, it's all green for me over here.

https://github.com/user-attachments/assets/d7179dae-ce83-4906-9929-23612b656d4b

EmCousin:
@sm17p let's try to trigger an entirely new run - commit something and I'll trigger CI

sm17p:
Looks like something to do with headless runs.  I'll check if I can run the specs headlessly for more info

sm17p:
@EmCousin I've tried something based on my hunch. The headless runs on my local seem to pass, can you please re-run CI again for info?

Screenshot 2026-02-14 at 10 39 27 PM

app/javascript/pages/Products/Content/Edit.tsx
```diff
@@ -261,9 +261,11 @@ const EditProductContentPage = () => {
       onSuccess: (data) => {
         options?.onSuccess?.();
         options?.onFinish?.();
-        form.setData("product", cast<EditProductContent>(data.props.product));
-        form.setDefaults();
-        setSyncRichContentIdsFromServer((count) => count + 1);
+        if (data.url === Routes.edit_product_content_path(product.unique_permalink)) {
+          form.setData("product", cast<EditProductContent>(data.props.product));
```
sm17p, line 265: The `cast`call over here would raise an error if we re-direct to a different page since the underelying prop structure wont match.

My hunch is that this might have caused the CI headless chrome page to crash as it was an uncaught exception earlier

config/routes.rb
```diff
@@ -683,16 +683,25 @@ def product_info_and_purchase_routes(named_routes: true)
...
+        get :edit, to: "products/main#edit"
+        patch :update, to: "products/main#update"
+      end
+
+      scope module: :products do
+        resource :content, only: [:edit, :update], controller: "content"
+        resource :receipt, only: [:edit, :update], controller: "receipt"
+        resource :share, only: [:edit, :update], controller: "share"
+      end
     end
```
Pradumn27, line 703: Can we follow the same route pattern as bundles for consistency throughout, that is, using the route 
       
`/products/:id/product/edit` instead of `/products/:id/edit` - used `as: :main` for just avoiding the weird generated pathname of `product_product_path`

```suggestion
    resources :products, only: %i[index new create], controller: "links" do
      scope module: :products do
        scope format: true, constraints: { format: :json } do
          resources :other_refund_policies, only: :index
          resources :remaining_call_availabilities, only: :index
          resources :available_offer_codes, only: :index
        end
        resource :product, only: %i[edit update], controller: "main", as: :main
        resource :content, only: %i[edit update], controller: "content"
        resource :receipt, only: %i[edit update], controller: "receipt"
        resource :share, only: %i[edit update], controller: "share"
      end
    end
```

app/javascript/components/useDropboxDropins.ts
Pradumn27, line 1: we already have a hook called `useDropbox` for it, please use that here

app/javascript/components/ProductsPage/ProductIconCell.tsx
```diff
@@ -1,28 +1,35 @@
+import { Link } from "@inertiajs/react";
 import * as React from "react";
 
 import { Icon } from "$app/components/Icons";
 import { TableCell } from "$app/components/ui/Table";
 
 export const ProductIconCell = ({
+  productEditUrl = false,
```
Pradumn27:
```suggestion
  useInertiaLink = false,
```

This or any other better name, since `productEditUrl` suggests a URL but the prop is a boolean, and it’s tied to “edit” when the component only cares whether to use Inertia or a plain link. Naming the prop after the navigation behaviour keeps the component generic and lets the caller decide how to navigate.

app/controllers/products/main_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class Products::MainController < Products::BaseController
+  def edit
+    render inertia: "Products/Edit", props: {
+      product: -> { edit_product_main_presenter.edit_product },
+      page_metadata: edit_product_main_presenter.edit_product_metadata,
```
Pradumn27:
```suggestion
      page_metadata: -> { edit_product_main_presenter.edit_product_metadata },
```

app/javascript/components/useDropboxDropins.ts
Pradumn27: also if the usage of the script `load_dropbox_dropins_js` is not in the app anymore then let's delete it

app/services/product/update_service.rb
```diff
@@ -0,0 +1,322 @@
...
+          @product.product_refund_policy_enabled = product_params[:product_refund_policy_enabled]
+          if product_params[:refund_policy].present? && product_params[:product_refund_policy_enabled]
+            @product.find_or_initialize_product_refund_policy.update!(product_params[:refund_policy])
+          end
+        end
+        @product.save_shipping_destinations!(product_params[:shipping_destinations] || []) if @product.is_physical
+
+        if Feature.active?(:cancellation_discounts, @product.user) && (product_params[:cancellation_discount].present? || @product.cancellation_discount_offer_code.present?)
+          Product::SaveCancellationDiscountService.new(@product, product_params[:cancellation_discount]).perform
+        end
```
Pradumn27, line 170: Regression here - The dedicated rescue `ActiveRecord::RecordInvalid` around here was removed. The old controller used e.record.errors.full_messages.first to surface the correct model's error. Now it falls through to the outer rescue, which reads `@product.errors.full_messages.first`, if the invalid record is the offer code (not @product), the user may see a blank or wrong error message.

app/javascript/pages/Products/Edit.tsx
```diff
@@ -0,0 +1,664 @@
...
+              <BundleConversionNotice product={form.data.product} />
+              <fieldset>
+                <label htmlFor={`${uid}-name`}>{isCoffee ? "Header" : "Name"}</label>
+                <input
+                  id={`${uid}-name`}
+                  type="text"
+                  value={form.data.product.name}
+                  onChange={(evt) => {
+                    form.setData("product.name", evt.target.value);
+                    router.replaceProp("product.name", evt.target.value);
```
Pradumn27: why are we updating the name twice here?

app/presenters/product_presenter.rb
```diff
@@ -93,181 +93,200 @@ def existing_files
...
+      shipping_destinations: product.shipping_destinations.alive.map do |shipping_destination|
+        {
+          country_code: shipping_destination.country_code,
+          one_item_rate_cents: shipping_destination.one_item_rate_cents,
+          multiple_items_rate_cents: shipping_destination.multiple_items_rate_cents,
+        }
+      end,
+      section_ids:,
+      taxonomy_id: product.taxonomy_id&.to_s,
+      tags: product.tags.pluck(:name),
```
Pradumn27: Aren't these 3 and `is_adult` only needed by the shared tab and already in the `edit_product_share` action, we can remove them from here then

app/presenters/product_presenter.rb
Pradumn27, line 1: I see the suffix of `_props` was not used in the names of all the added/updated actions here. Please add them to follow the convention, and also helps in explicitly defining the purpose

app/presenters/product_presenter.rb
```diff
@@ -93,181 +93,200 @@ def existing_files
...
+        successful_sales_count: product.successful_sales_count,
+        sales_count_for_inventory: product.max_purchase_count? ? product.sales_count_for_inventory : 0,
+        cancellation_discounts_enabled: Feature.active?(:cancellation_discounts, product.user),
+        google_client_id: GlobalConfig.get("GOOGLE_CLIENT_ID"),
+        google_calendar_enabled: Feature.active?(:google_calendar_link, product.user),
+      }
+    )
+  end
+
+  def edit_product_content
```
Pradumn27:
```suggestion
  def edit_content_props
```

the product prefix is not needed here as we already are in a `Product` presenter

app/presenters/product_presenter.rb
```diff
@@ -93,181 +93,200 @@ def existing_files
...
-      },
-      cancellation_discounts_enabled: Feature.active?(:cancellation_discounts, product.user),
-      ai_generated:,
+      s3_url: "#{AWS_S3_ENDPOINT}/#{S3_BUCKET}",
+      seller: UserPresenter.new(user:).author_byline_props,
+      dropbox_picker_app_key: DROPBOX_PICKER_API_KEY,
     }
   end
 
+  def edit_product_receipt
```
Pradumn27:
```suggestion
  def edit_receipt_props
```

app/presenters/product_presenter.rb
```diff
@@ -93,181 +93,200 @@ def existing_files
...
+  end
+
+  def edit_product_receipt_metadata
+    {
+      custom_receipt_text_max_length: Product::Validations::MAX_CUSTOM_RECEIPT_TEXT_LENGTH,
+      custom_view_content_button_text_max_length: Product::Validations::MAX_VIEW_CONTENT_BUTTON_TEXT_LENGTH,
+    }
+  end
+
+  def edit_product_share
```
Pradumn27:
```suggestion
  def edit_share_props
```

app/javascript/components/useRefToLatest.ts
```diff
@@ -9,3 +9,9 @@ export const useRefToLatest = <T>(value: T): React.MutableRefObject<T> => {
...
+
+export const useLatest = <T>(value: T): React.MutableRefObject<T> => {
+  const ref = React.useRef(value);
+  ref.current = value;
+  return ref;
+};
```
Pradumn27: This sets ref.current during render, which is technically a side effect during render. This works in practice but isn't strictly React-safe. The existing useRefToLatest uses useEffect, which is cleaner.

reference - ["Do not write or read ref.current during rendering, except for initialization. This makes your component’s behavior unpredictable."](https://react.dev/reference/react/useRef)

app/javascript/components/ProductEdit/Layout.tsx
```diff
@@ -214,43 +117,56 @@ export const Layout = ({
...
-        {saving ? "Saving changes..." : "Save changes"}
+      <Button color="primary" disabled={isBusy} onClick={() => save()}>
+        {processingStatus === "saving" ? "Saving changes..." : "Save changes"}
       </Button>
     </WithTooltip>
   );
 
-  const onTabClick = (e: React.MouseEvent<HTMLAnchorElement>, callback?: () => void) => {
+  const onTabClick = (event: React.MouseEvent) => {
+    if (!(event.target instanceof HTMLAnchorElement)) return;
```
Pradumn27, line 127: why is this check needed? when will this fail?

app/controllers/concerns/inertia_rendering.rb
Pradumn27, line 32: This check is very brittle. Why don't we add a flag for an HTML alert message and use that instead like `flash[:html]`

Pradumn27 (changes requested):
Hey @sm17p _a, did an initial round of review, will complete the rest soon, can you resolve these comments in the meanwhile.

Also, I figured why you're having the test failures, it's because the `DROPBOX_PICKER_API_KEY` which is now used in your defined tests and development environment too, causes runtime errors if the key is null, so you can add it to `.env.test` and `.env.development` as it is not present there, thus causing the failing tests

app/javascript/components/useDropboxDropins.ts
sm17p: Makes sense, all of that code was there because Download page was still SSR when the PR was opened. We can re-use the hook and also remove the view template as it's not used anymore 👍

app/javascript/components/ProductEdit/Layout.tsx
sm17p: It won't fail in our case, this check keeps TS happy by narrowing the type on `event.target`. 

For unpublished products, we have a check below to block share tab access on `L141` which uses the `href` property on `event.target`.  

We have introduced the same check in backend controller which blocks direct navigation, but for tab switches our UI, it's kind of unnecessary to sync all that state and re-direct back to the current page.

app/javascript/pages/Products/Edit.tsx
sm17p: Ah! We don't that's leftover code from my previous design iteration

app/services/product/update_service.rb
sm17p: Nice catch! That uncovered a series of bugs and missing specs

Ref - https://github.com/antiwork/gumroad/pull/3301/commits/0450cdea4c357757fc91f4f39a5ecb692f753943

Ignore the debug script from the above commit as it was removed later in  https://github.com/antiwork/gumroad/pull/3301/commits/1254964de577437326531af3b298eacc158e5745

app/controllers/concerns/inertia_rendering.rb
sm17p: Gave it more thought, I think we let it be for now and handle this in a separate PR I guess

There is another case where if we (or a future change) put HTML with Tailwind classes in a backend flash message, eg:

```ruby
raise Link::LinkInvalid, '<span class="text-red-600 font-medium">Please <a class="underline" href="...">set an email</a>.</span>'
```

then those classes won’t be in Tailwind’s scanned content, and the alert might possibly render without the intended styles in production. 

----- 

The two possible solutions for this problem are

### 1. Use the client-side alert feature from this PR
Backend sends a structured payload, never HTML, we define the layout and style in client to preserve css styles. 

or

### 2. Single Ruby file for HTML flash (Need to test)
All HTML used in flash messages lives in one file  Controllers/services call helpers from that file. Tailwind’s @source includes only that file so any Tailwind classes in those strings are scanned and kept in the build.

```ruby
# app/lib/html_flash_messages.rb
module HtmlFlashMessages
  def self.email_required_for_publish(settings_url)
    %(<span class="text-red-600">To publish… <a class="underline" href="#{settings_url}">Set an email</a>.</span>)
  end
end
```

```css
/* tailwind.css */

@source "../../../app/lib/html_flash_messages.rb";
```

cc @EmCousin

config/routes.rb
sm17p: For bundles, it makes sense to have `/bundles/:id/product/edit` because there the “product” is a real sub-resource of the bundle. 

For products, the resource is the product, there is no nested product to edit. Adding `/products/:id/product/edit` implies that we're editing a sub-resource of a product which is not what we're doing over here.

I think we should keep the usual member routes `GET /products/:id/edit` and `PATCH /products/:id`, as we'd break convention otherwise and it might confuse people. Wdyt?

app/javascript/components/useRefToLatest.ts
sm17p: I was using that `useRefToLatest` initially but that doesn't work as mentioned in https://github.com/antiwork/gumroad/pull/3301#discussion_r2756310747. It lags behind render cycles so we end up with alternating states while editing a `FileEmbed`.

https://github.com/user-attachments/assets/05d77954-3827-4d9a-8989-f38eb4ff4c79

----

I've changed the implementation by using context instead of refs to avoid all the complications that comes with concurrent react, it should be smoother to update to React 19 now

sm17p:
@Pradumn27 _a I've made a few changes and replied to the review comments, please have a look again

sm17p:
1. https://github.com/antiwork/gumroad/actions/runs/22166450501/job/64095889675?pr=3301 - fixed one related failing spec. This one started failing after we added `DROPBOX_PICKER_API_KEY` in `.env.test`

Screenshot 2026-02-19 at 4 06 42 PM

---

2. https://github.com/antiwork/gumroad/actions/runs/22166450501/job/64095890331?pr=3301 is unrelated 
3. https://github.com/antiwork/gumroad/actions/runs/22166450501/job/64095890324?pr=3301 flaked on the last CI run, but is passing locally

Screenshot 2026-02-19 at 4 57 26 PM

slavingia:
Thanks for the contribution! We're shifting our open source focus to smaller, easier-to-verify changes like bug fixes. We're closing PRs for larger features for now — but please don't hesitate to pick up a bug or smaller issue. We appreciate your help!

Issue #3316, opened 2026-02-02
Getting "Some files are still uploading, please wait..." even after cancelling file upload


mgf-audio:
I want to go to bed so i'm cancelling an upload of a big file. I can't save the product or navigate away from Content tab after cancelling the upload. It just says "Some files are still uploading, please wait...". Pretty annoying.

PR #3317, opened 2026-02-02
fix(product-edit): Prevent cancelled file uploads from blocking save

4arjun:
Issue: #3316 

## Summary
- Fix bug where cancelling a file upload midway prevents saving changes or navigating away
- Guard `updateProduct` call in `updateContent` to prevent stale closure from re-adding cancelled files

## Problem
When uploading a file in the product content page and cancelling the upload midway, users see "Files are still uploading..." and cannot save changes or navigate away. This only occurs when there is already at least one successfully uploaded file.

## Root Cause
Two issues combined to cause this bug:
1. **Missing cleanup**: The `onCancel` handler removed the editor node and cancelled the S3 upload, but did not remove the file entry from `product.files`. The file remained with `uploadStatus: { type: "uploading" }`.
2. **Stale closure race condition**: When the editor node is deleted via `queueMicrotask`, it triggers an editor "update" event. The `updateContent` callback runs with a stale `product.files` reference (because `useRefToLatest` updates the ref in an effect that runs after render). This re-spreads the stale files array back into state, re-adding the cancelled file.

## Before

https://github.com/user-attachments/assets/03d5f9df-e33e-4858-b49f-79c108974b96

## After

https://github.com/user-attachments/assets/1abf1417-e5de-4372-a939-444defb503ca

# Checklist

---

# AI Disclosure
Claude Opus 4.5 was used to identify the root cause

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -96,7 +96,7 @@ class OrdersController < ApplicationController
       payment_method = Stripe::PaymentMethod.retrieve(params[:stripe_payment_method_id])
       payment_method&.card&.wallet&.type == params[:wallet_type]
     rescue Stripe::StripeError
-      render_error("Sorry, something went wrong.")
+      render_error("Sorry, something went wrong")
     end
 
     def permitted_order_params
--- a/app/javascript/components/Product/Covers/Image.tsx
+++ b/app/javascript/components/Product/Covers/Image.tsx
@@ -7,10 +7,9 @@ import { DEFAULT_IMAGE_WIDTH } from "./";
 type Props = { cover: AssetPreview; dimensions: { height: number; width: number } | null };
 const Image = ({ cover, dimensions }: Props) => (
   <img
-    className="preview"
+    className="w-full"
     src={dimensions == null || dimensions.width > DEFAULT_IMAGE_WIDTH ? cover.original_url : cover.url}
     itemProp="image"
-    style={{ maxWidth: "100%" }}
   />
 );
 
--- a/app/mailers/customer_low_priority_mailer.rb
+++ b/app/mailers/customer_low_priority_mailer.rb
@@ -12,7 +12,7 @@ class CustomerLowPriorityMailer < ApplicationMailer
                                                      subscription_charge_failed subscription_product_deleted subscription_renewal_reminder
                                                      subscription_price_change_notification subscription_ended free_trial_expiring_soon
                                                      credit_card_expiring_membership subscription_early_fraud_warning_notification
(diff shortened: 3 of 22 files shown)


app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -187,6 +187,9 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
     deleteNode();
     uploader.cancelUpload(`file_${file.id}`);
     if (file.status.type === "dropbox") void cancelDropboxFileUpload(file.id);
+    updateProduct((product) => {
+      product.files = product.files.filter((f) => f.id !== file.id);
+    });
```
4arjun, line 192: This removes the cancelled file from product.files so it's no longer counted as "uploading" in the isUploadingFiles check in Layout.tsx. Previously, the cancel handler only removed the editor node and cancelled the S3 upload, but left the file entry in state with uploadStatus: { type: "uploading" }, which blocked saving and navigation.

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -247,7 +247,9 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
         node.remove();
       }
     });
-    updateProduct({ files: [...product.files.filter((f) => !newFiles.includes(f)), ...newFiles] });
+    if (newFiles.length > 0) {
+      updateProduct({ files: [...product.files.filter((f) => !newFiles.includes(f)), ...newFiles] });
+    }
```
4arjun, line 252: Prevents a race condition where a stale product.files reference re-adds the cancelled file. newFiles is only populated for files copied from other products, so for normal cancellations we skip this call entirely.

4arjun:

binary-koan:
Can we reasonably add an e2e test for this?

4arjun:
> Can we reasonably add an e2e test for this?

Looked into this, there's an existing challenge with testing upload cancellation for S3 uploads. The codebase has comments acknowledging this:

    # TODO(ershad): Enable this once we have a way to slow down the upload process
    # button = find_button("Save changes", disabled: true)
    # button.hover
    # expect(button).to have_tooltip(text: "Files are still uploading...") 
(in file_embeds_spec.rb:27 and rich_text_editor_spec.rb:83).

The issue is that uploads to MinIO complete too quickly to reliably interact with the Cancel button.

binary-koan (approved):
Got it, LGTM!

Issue #3319, opened 2026-02-02
Fix design issue on purchase page on mobile when installment plan is enabled


sriramk42:
Fix design issue on purchase page on mobile when installment plan is enabled

<img width="559" height="149" alt="Image" src="https://github.com/user-attachments/assets/25ab65e6-3306-4f5d-b684-493da3eaf24f" />

PR #3324, opened 2026-02-02
Fix design issues on purchase page on mobile when installment plan is enabled #3319

Krishna41357:
Description:

Fixes mobile UI/UX issues on the purchase page when installment plans are enabled. Sticky CTA bar and price display now render correctly on small screens without wrapping.

Problem:

Text and buttons in the navbar wrapped unexpectedly on mobile.

Sticky CtaBar did not stack elements properly (price, product name, buttons).

Root cause:
CtaButton lacked flex constraints, PriceTag had no mobile styling, and CtaBar used a rigid horizontal layout.

Solution:

CtaButton.tsx

Vertical stacking on mobile (flex flex-col)

w-full buttons and text-xs notes

PriceTag.tsx

Prevent shrinking (flex-shrink-0)

Keep price on one line (whitespace-nowrap)

Responsive font size (text-sm lg:text-base)

Layout.tsx (CtaBar)

Mobile: vertical layout (flex-col)

Desktop: horizontal layout (lg:flex-row)

Proper padding, spacing, and shrink constraints

Components Modified:

CtaButton.tsx

PriceTag.tsx

Layout.tsx

Testing:

✅ Mobile (320–768px), Tablet (768–1024px), Desktop (1024px+)

✅ Products with and without installment plans

✅ Safari compatibility

✅ Fixes issue: #3319

### DIFF
--- a/app/javascript/components/Product/CtaButton.tsx
+++ b/app/javascript/components/Product/CtaButton.tsx
@@ -143,8 +143,8 @@ export const CtaButton = React.forwardRef<HTMLAnchorElement, Props>(
     };
 
     return (
-      <>
-        <NavigationButton ref={ref} href={url.toString()} color="accent" {...buttonCommonProps}>
+      <div className="flex flex-col gap-2 lg:gap-3">
+        <NavigationButton ref={ref} href={url.toString()} color="accent" className="w-full" {...buttonCommonProps}>
           {label ??
             (purchase
               ? "Purchase again"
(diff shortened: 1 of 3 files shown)


laugardie:
I don’t think we should increase the sticky bar height to solve this. It takes way too much space in mobile.

Instead, we could just swap the copy to "Pay in 4" instead of the full "Pay in 4 installments" on mobile viewports. Stripe handles it this way, and that will keep the footer compact. @slavingia @ershad

Screenshot 2026-02-09 at 11 19 18
Screenshot 2026-02-09 at 11 27 54

@Krishna41357 when you open the PR, can you make sure to:

- Include before and after screenshots (makes it way easier to review).

- Double-check that this change doesn't break the layout on other screen sizes (tablet/desktop).
Screenshot 2026-02-09 at 11 24 23

slavingia:
I think we can just make it ... using Tailwind in cases where it would cause a 2-liner, so we don't have to update the button text for all use cases. Can discuss on the issue.

PR #3345, opened 2026-02-03
Fix mobile download page when impersonating

binary-koan:
# Description

Fixes a bug with #2820, where sending the mobile access token to skip the confirmation step on the download page would break when impersonating.

It caused this [Bugsnag error](https://app.bugsnag.com/gumroad/web/errors/6977d9c5964d9ac40c55d0db?filters[error.status]=open&filters[event.since]=30d), basically because `doorkeeper_authorize!` sort of half logs the user in, which means `pundit_user.user` is set but `pundit_user.seller` is nil.

This fixes the issue by fully signing in the user, which should also be a better experience in case we link to other Gumroad pages from the download page.

---

# AI Disclosure

Claude Opus 4.5 used for exploring the code and suggesting fixes.


binary-koan:

PR #2766, opened 2026-01-08
fix(tests): race condition caused by #2755 

stefanbinoj:
ref #2565 
[comment](https://github.com/antiwork/gumroad/pull/2565#issuecomment-3724245985)

### Tests
Screenshot 2026-01-08 at 8 45 25 PM

### AI Disclosure 
None 

### Live Stream Disclosure
Have watched all live streams

PR #2755, opened 2026-01-08
Inertia Migrate - Auth (Part 3)

stefanbinoj:
Issue: #856 
Previous PR #2629 

# Description
- Add a new `delete` endpoint for `/logout` which can be used by `app/javascript/components/client-components/Nav/index.tsx` for SPA like behaviour
- Server Nav component (`app/javascript/components/server-components/Nav.tsx`) still uses old `get` route for `/logout` with no change. 

### Before

https://github.com/user-attachments/assets/df07e731-d00b-49df-bd97-4039375f19c1

https://github.com/user-attachments/assets/ade4bee1-5583-4367-aab8-1d3e20502732

---

# Test Results
Screenshot 2026-01-02 at 3 36 53 PM

---

# Checklist

- [X] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [X] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [X] I have performed a self-review and left review comments on my PR
- [X] I have added/updated tests for my changes

---

# AI Disclosure

- Used opus4.5 with opencode for initial draft of this PR.

PR #3349, opened 2026-02-04
Migrate Checkout Page to Inertia

stefanbinoj:
Issue: #3042 

# Before (Logged Out)

https://github.com/user-attachments/assets/9f7abc42-d702-4134-acc4-8123e4285d10

# After (Logged Out)

https://github.com/user-attachments/assets/4ea011a7-8d7b-4310-896b-d439612546a0

# Before (Logged In)

https://github.com/user-attachments/assets/e39fbd98-c1ec-46c4-b7e2-c1dba0a56af9

# After (Logged In)

https://github.com/user-attachments/assets/656bbe45-cae3-449a-8136-6329ece2a838

---

# Test Results

Screenshot 2026-02-04 at 8 18 02 PM

---

# Checklist

---

# AI Disclosure

GPT 5.2 Codex via codex was used for initial draft of this PR.

### DIFF
--- a/app/javascript/components/Admin/BlockEmailDomainsForm.tsx
+++ b/app/javascript/components/Admin/BlockEmailDomainsForm.tsx
@@ -35,7 +35,6 @@ const Form = ({ action, header, buttonLabel }: Props) => {
   return (
     <form onSubmit={handleSubmit}>
       <FormSection
-        className="p-4 md:p-8"
         header={
           <>
             <p>{header}</p>
--- a/app/javascript/components/Admin/SalesReports/JobHistory.tsx
+++ b/app/javascript/components/Admin/SalesReports/JobHistory.tsx
@@ -63,9 +63,7 @@ const AdminSalesReportsJobHistory = ({ countries, sales_types, jobHistory, authe
         authenticityToken={authenticityToken}
         onSuccess={() => setShowNewSalesReportForm(false)}
         wrapper={(children) => (
-          <FormSection className="p-4 md:p-8" header="Generate sales report with custom date ranges">
-            {children}
-          </FormSection>
+          <FormSection header="Generate sales report with custom date ranges">{children}</FormSection>
         )}
       />
     ) : (
--- a/app/javascript/components/Checkout/CreditCardInput.tsx
+++ b/app/javascript/components/Checkout/CreditCardInput.tsx
@@ -1,6 +1,5 @@
 import { CardElement, Elements } from "@stripe/react-stripe-js";
 import { StripeCardElement, StripeElementStyleVariant, StripeCardElementChangeEvent } from "@stripe/stripe-js";
-import cx from "classnames";
 import * as React from "react";
 
 import { SavedCreditCard } from "$app/parsers/card";
@@ -9,6 +8,9 @@ import { getCssVariable } from "$app/utils/styles";
 
 import { useFont } from "$app/components/DesignSettings";
 import { Icon } from "$app/components/Icons";
+import { Fieldset, FieldsetTitle } from "$app/components/ui/Fieldset";
+import { InputGroup } from "$app/components/ui/InputGroup";
+import { Label } from "$app/components/ui/Label";
 
 export const CreditCardInput = ({
   disabled,
@@ -32,9 +34,9 @@ export const CreditCardInput = ({
   const [baseStripeStyle, setBaseStripeStyle] = React.useState<null | StripeElementStyleVariant>(null);
 
   return (
-    <fieldset className={cx({ danger: invalid })}>
-      <legend>
-        <label>Card information</label>
+    <Fieldset state={invalid ? "danger" : undefined}>
+      <FieldsetTitle>
+        <Label>Card information</Label>
         {savedCreditCard ? (
           <button
             className="cursor-pointer font-normal underline all-unset"
@@ -44,15 +46,15 @@ export const CreditCardInput = ({
             {useSavedCard ? "Use a different card?" : "Use saved card"}
           </button>
         ) : null}
-      </legend>
+      </FieldsetTitle>
       {savedCreditCard && useSavedCard ? (
-        <div className="input read-only" aria-label="Saved credit card">
+        <InputGroup readOnly aria-label="Saved credit card">
           <Icon name="outline-credit-card" />
           <span>{savedCreditCard.number}</span>
           <span style={{ marginLeft: "auto" }}>{savedCreditCard.expiration_date}</span>
-        </div>
+        </InputGroup>
       ) : (
-        <div className={cx("input", { disabled })} aria-label="Card information" aria-invalid={invalid}>
+        <InputGroup disabled={disabled} aria-label="Card information" aria-invalid={invalid}>
           {baseStripeStyle == null ? (
             <input
               ref={(el) => {
@@ -71,7 +73,7 @@ export const CreditCardInput = ({
           ) : null}
           <StripeElementsProvider>
             <CardElement
-              className="fake-input"
+              className="flex-1"
               options={{
                 style: { base: baseStripeStyle ?? {} },
                 hidePostalCode: true,
@@ -83,9 +85,9 @@ export const CreditCardInput = ({
               {...(onChange ? { onChange } : {})}
             />
           </StripeElementsProvider>
-        </div>
+        </InputGroup>
       )}
-    </fieldset>
+    </Fieldset>
   );
 };
 
--- a/app/javascript/components/ConfirmBalanceForfeitOnPayoutMethodChangeModal.tsx
+++ b/app/javascript/components/ConfirmBalanceForfeitOnPayoutMethodChangeModal.tsx
@@ -2,6 +2,8 @@ import * as React from "react";
 
 import { Button } from "$app/components/Button";
 import { Modal } from "$app/components/Modal";
+import { Input } from "$app/components/ui/Input";
+import { Label } from "$app/components/ui/Label";
 
 type Props = {
   balance: string | null;
@@ -38,10 +40,10 @@ export const ConfirmBalanceForfeitOnPayoutMethodChangeModal = ({ balance, open, 
               Please confirm that you're okay forfeiting your balance by typing <b>"I understand"</b> below and clicking{" "}
               <b>Confirm</b>.
               <div className="mt-4">
-                <label htmlFor="confirmation-input" className="sr-only">
+                <Label htmlFor="confirmation-input" className="sr-only">
                   Type "I understand" to confirm
-                </label>
-                <input
+                </Label>
+                <Input
                   id="confirmation-input"
                   type="text"
                   value={confirmText}
--- a/app/javascript/components/CustomDomain.tsx
+++ b/app/javascript/components/CustomDomain.tsx
@@ -1,4 +1,3 @@
-import cx from "classnames";
 import * as React from "react";
 import { cast } from "ts-safe-cast";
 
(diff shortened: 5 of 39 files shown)


app/controllers/checkout/carts_controller.rb
stefanbinoj, line 1: Please let me know if we have to rename the controller or move to a new specific location.

app/controllers/api/internal/carts_controller.rb
```diff
@@ -48,11 +48,11 @@ def update
       cart.alive_cart_products.where.not(id: updated_cart_products.map(&:id)).find_each(&:mark_deleted!)
     end
 
-    head :no_content
+    redirect_to checkout_index_path, status: :see_other
```
stefanbinoj: succesfull 303 redirection

app/javascript/components/Checkout/cartState.ts
```diff
@@ -177,17 +174,3 @@ export function getDiscountedPrice(cart: CartState, item: CartItem): DiscountedP
 export function newCartState(): CartState {
   return { items: [], discountCodes: [] };
 }
-
-export async function saveCartState(cart: CartState) {
```
stefanbinoj, line 181: No longer needed as we migrated this controller from json

app/javascript/components/Checkout/Modals.tsx
stefanbinoj, line 1: created a new helper file by copying code from `app/javascript/pages/Checkout/Index.tsx`

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -538,14 +530,17 @@ export const CheckoutPage = ({
...
-    asyncVoid(async () => {
-      try {
-        await saveCartState(cart);
-      } catch (e) {
-        assertResponseError(e);
-        showAlert("Sorry, something went wrong. Please try again.", "error");
-      }
-    }),
+    () =>
+      router.put(
```
stefanbinoj, line 535: using router.put for sending req to newly migrated internal api route

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -538,14 +530,16 @@ export const CheckoutPage = ({
...
+        Routes.internal_cart_path(),
+        { cart },
+        {
+          preserveState: true,
+          preserveScroll: true,
+          replace: true,
```
stefanbinoj, line 540: better to not add each cart change to the browser history

stefanbinoj:
Hi @Pradumn27 _a could you take a look at this whenever you are free? Thanks!

app/controllers/checkout/carts_controller.rb
```diff
@@ -3,7 +3,7 @@
 class Api::Internal::CartsController < Api::Internal::BaseController
```
EmCousin: This is no longer an API endpoint. This should be a `CartsController`, inherit `ApplicationController` and be covered by the proper access rights when needed.

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
 class CheckoutController < ApplicationController
+  layout "inertia"
   before_action :process_cart_id_param
 
   def index
```
EmCousin: Optional: would be a good opportunity to make this a `show` action. `index` actions are used by convention to render lists of resources.

app/javascript/pages/Checkout/Index.tsx
```diff
@@ -538,14 +530,16 @@ export const CheckoutPage = ({
...
-        assertResponseError(e);
-        showAlert("Sorry, something went wrong. Please try again.", "error");
-      }
-    }),
+    () =>
+      router.put(
+        Routes.internal_cart_path(),
+        { cart },
+        {
+          preserveState: true,
```
EmCousin: This should be already the case by default https://inertiajs.com/docs/v2/the-basics/manual-visits#state-preservation
```suggestion
```

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,13 @@
...
 class CheckoutController < ApplicationController
+  layout "inertia"
   before_action :process_cart_id_param
 
-  def index
-    @hide_layouts = true
-    @checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+  def show
+    checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+    render inertia: "Checkout/Show",
```
stefanbinoj, line 9: now using Show action

app/javascript/components/Checkout/payment.ts
```diff
@@ -403,12 +402,6 @@ export function createReducer(initial: {
     },
   );
   const [state, dispatch] = reducer;
-  useRunOnce(() => {
```
stefanbinoj, line 406: Even tho this is a `runOnce` code moved to `onSuccess` handler of `saveCartState` becuase it was causing a race condition with `saveCartState` where `saveCartState` sends a req as soon as page is loaded with (query params) and controller redirects to checkout_path with same query params but in the meanwhile `useRunOnce` code was running and trying to update query params which gets re-updated by the `redirect_to` action of cart_controller. Thus had to move this to `onSuccess` handler of `saveCartState`

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -538,14 +530,24 @@ export const CheckoutPage = ({
...
-      }
-    }),
+    () =>
+      router.put(
+        `${Routes.checkout_cart_path()}${window.location.search}`,
+        { cart },
+        {
+          preserveScroll: true,
+          replace: true,
+          onSuccess: () => {
```
stefanbinoj, line 542: same code as in `app/javascript/components/Checkout/payment.ts:useRunOnce` moved here to avoid race condition

stefanbinoj:
HI @EmCousin sorry for interrupting CI run as i was in middle of pushing to this branch, my bad! could you re run the tests once again?

spec/requests/purchases/product/offer_codes_spec.rb
```diff
@@ -187,7 +187,8 @@
       add_to_cart(product, offer_code:)
       offer_code.update!(amount_cents: 200)
       check_out(product, error: "The price just changed! Refresh the page for the updated price.")
-      visit checkout_index_path
+      wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }
```
stefanbinoj, line 190: **Reason for adding this line throughout test files? (Prevent race condition)** : We now have inerita put req to update cart state using `debouncedSaveCartState`. Since we rely on inertia and inertia req are asynchronous by nature also we cant have a async await as supposed to earlier (in json req) to inertia req. Thus added 
`wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }` 
this check ensures that inertia async req is completed and cart state is saved to db. This prevents race condition. 

This is simialr to this issue #2766 
ref https://github.com/antiwork/gumroad/pull/2755#discussion_r2671640987

stefanbinoj:
working on it!

stefanbinoj:
Hi @EmCousin something that i would like to point out is that since we migrated `Checkout::CartsController#update` off from json with `debouncedSaveCartState` using router.put for updating cart state poses some issues:

Inertia req are async in nature and they doesn't returns a promise and we cant await these req so when `debouncedSaveCartState` handler runs (ie, when the cart is updated or email is updated) it sends a inertia req and there is no way for us to confirm that the req has been successfully completed and redirected to `checkout_path` with 303 redirect. Currently a hack would be to either check for 

```ruby
wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }  # Which checks in db that the inertia req has done its job
# or 
wait_until_true { Cart.alive.last&.cart_products&.alive&.count == 2 }  # More direct checking of cart state with db 
```
Thus this causes a race condition where the update action maynot have completed its execution but the tests has no check whether inertia req has successfully completed or not and thus tests move with next set of checks without waiting for inertia handler to complete( some times tests navigates to a new page this cancels the inertia req and cart state maynot be updated). 

This is same issue as in  #2766 

Inorder to fix these we need to add this check (`wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }`) in all occurance of cart updation (ie, across all usage of `def fill_checkout_form` (spec/support/checkout_helpers.rb:59) and `def check_out` (spec/support/checkout_helpers.rb:148)

Another approach would be to relying on json endpoints with await for updating cart state as earlier.

wdyt about this? is there any other way to eliminate this race condition? Would love to hear your thoughts on this.

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -539,14 +531,22 @@ export const CheckoutPage = ({
...
-      }
-    }),
+    () =>
+      router.put(
+        `${Routes.checkout_cart_path()}${window.location.search}`,
+        { cart },
+        {
+          preserveScroll: true,
+          replace: true,
+          async: true,
```
stefanbinoj, line 541: Without async: true, another Inertia visit can cancel the in-flight save. This helps in reducing flakiness across tests

stefanbinoj:
Hi @EmCousin _a, quick update on the checkout migration race condition.

After migrating `Checkout::CartsController#update` from JSON to Inertia, cart autosave now uses `router.put` inside `debouncedSaveCartState`. Inertia requests are async in nature, and this flow does not give us a directly awaitable Promise in test flow. So when debouncedSaveCartState runs (for cart/email updates), it sends an Inertia request and we don’t have a deterministic way in tests to confirm it has fully completed and followed the 303 redirect back to checkout_path.

I noticed that some tests may move to the next assertions/actions before the update action has finished (also in some tests a navigation may happen while the save is in flight, causing inertia request to be interrupted, and cart state may not be saved) this causes a race condition where cart state is not saved which is req for next assertions.

This is the same type of issue as experienced in #2766.

To fix this, I applied:

1. Added `async: true` to the Inertia cart save request, so it runs in Inertia’s async stream and is less likely to be interrupted by subsequent Inertia visits.
2. Added explicit DB-backed waits in affected places (mainly around fill_checkout_form / check_out callsites) when we need persisted cart state before continuing, for example:

```ruby
wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }  # confirms cart save completed
# or
wait_until_true { Cart.alive.last&.cart_products&.alive&.count == 2 }    # confirms expected cart item state
```

After these changes, things look stable from my local runs on the related specs.

Could we trigger CI once more to validate there are no remaining race-condition failures? If anything still flakes for the same reason, I can add the same wait check in those remaining spots.

stefanbinoj:
Hi @EmCousin all the failing tests are flaky and seems to be passing locally (screen shots attached). Could you rerun these tests once again? 

> slow test 3 

Screenshot 2026-02-09 at 8 42 47 PM

> slow test 9 

Screenshot 2026-02-09 at 8 42 00 PM


> slow test 14

Screenshot 2026-02-09 at 8 41 54 PM

stefanbinoj:
@EmCousin one test was still failing (probably unrelated) still added a extra security to race condition by adding 
`wait_until_true { Cart.alive.where(email: "test@gumroad.com").exists? }`

Screenshot 2026-02-09 at 9 25 31 PM

could you retry running it again?

stefanbinoj:
Hi @EmCousin could you re run these tests as it failed due to time out issue

stefanbinoj:
@EmCousin all tests seems to be passing!

EmCousin:
@stefanbinoj @sm17p (https://github.com/antiwork/gumroad/pull/3441) there are good things in both your implementations, which make them quite complimentary to one another, so I'll create a separate PR combining the best of both worlds and split the bounty between you too if that's okay. Thanks for your work!

PR #3066, opened 2026-01-23
Fix: Preserve offer code discount for installments when code is deleted/expired/maxed (#1410)

yashranaway:
Issue: #1410

# Description

## Problem

When an offer code hits its max usage limit (or gets deleted/expires) between installment charges, subsequent payments lose the discount even though the customer legitimately used the code on their first payment.

## Solution

Instead of checking the live offer code on each charge, we now copy the discount data from the original purchase's `purchase_offer_code_discount` record. This data was already being stored - we just weren't using it for subsequent charges.

Two changes:
1. `Subscription#build_purchase` - copies cached discount data to new purchases
2. `Purchase#original_offer_code` - prioritizes cached data over live offer code lookup

---

# Test Results
image
image

Added tests for:
- Deleted offer codes (installments + memberships)
- Expired offer codes
- Max usage reached
- Offer code amount changed after purchase
- Percentage discounts
- Backwards compatibility with legacy purchases

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Opus 4.5 was used via Claude Code was used for implementation and test writing.

PR #3356, opened 2026-02-04
fix(purchase): ensure offer code persistence on installments even if …

avi-474:
Issue: #1410 

# Description

This PR fixes a bug where offer code discounts were not consistently applied to non-first installments or subscription charges once the offer code was deleted or its global usage limit was reached.
The solution makes sure the discount is saved when the customer makes the first purchase and is automatically applied to all future related charges.

## Problem
When a customer started an installment or subscription purchase using an offer code with a limited usage count:

- The first installment correctly applied the discount.
- Subsequent installments failed to apply the discount if the offer code became “sold out” or was deleted in the meantime.
- This caused inconsistent pricing and broke customer expectations.

## Solution

- Save the discount details from the first purchase so they can be reused later.
- Skip offer code checks for later installments, subscription updates, and commission payments.
- Make sure later installments do not reduce product inventory or use up the offer code again.
- Use the saved discount information instead of checking the current offer code data for future charges.

This makes sure the same price and discount are applied to all installments and subscription payments.

---

# Before/After

Before:

- Discount applied only to the first installment.
- Later installments ignored the discount if the offer code was exhausted or deleted.

After:

- Discount is consistently applied to all installments and subscription charges originating from the initial purchase.
- Offer code deletion or exhaustion no longer affects existing installment plans.

---

# Test Results

https://github.com/user-attachments/assets/b6f3488c-f3c9-45a1-a377-84f4f9213064

Screenshot from 2026-02-04 23-15-31

Confirmed:
- First installment succeeds with a one-time-use offer code.
- Subsequent installments retain the same discount after the code is exhausted.
- Discount persists even if the offer code record is deleted.

All tests pass locally.
---

# Checklist

---

# AI Disclosure

Model: Claude Opus 4.5 via Antigravity
Used for:

- Reproduction test generation.
- Framing the manual testing steps used in the demo video.

All code and tests were manually reviewed and modified by me to ensure correctness and consistency with existing behavior.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -1999,7 +1999,7 @@ class Purchase < ApplicationRecord
   end
 
   def does_not_count_towards_max_purchases
-    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase
+    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase || (is_installment_payment && !is_original_subscription_purchase)
   end
 
   # Public: Determine if this purchase is a test purchase by the links owner.
@@ -2411,13 +2411,14 @@ class Purchase < ApplicationRecord
   end
 
   def original_offer_code(include_deleted: false)
-    return nil if offer_code&.deleted? && !include_deleted
-
     if has_cached_offer_code?
-      code = purchase_offer_code_discount.offer_code.code
-      purchase_offer_code_discount.offer_code_is_percent ?
-        OfferCode.new(amount_percentage: purchase_offer_code_discount.offer_code_amount, code:) :
-        OfferCode.new(amount_cents: purchase_offer_code_discount.offer_code_amount, code:)
+      p_discount = purchase_offer_code_discount
+      code = p_discount.offer_code.code
+      p_discount.offer_code_is_percent ?
+        OfferCode.new(amount_percentage: p_discount.offer_code_amount, code:) :
+        OfferCode.new(amount_cents: p_discount.offer_code_amount, code:)
+    elsif offer_code&.deleted? && !include_deleted
+      nil
     else
       offer_code
     end
@@ -3358,8 +3359,7 @@ class Purchase < ApplicationRecord
 
     def validate_offer_code
       return if errors.present?
-      # accept the offer code that was used when the buyer preordered/subscribed
-      return if is_preorder_charge? || is_recurring_subscription_charge || is_gift_receiver_purchase
+      return if is_preorder_charge? || is_recurring_subscription_charge || is_gift_receiver_purchase || (is_installment_payment && !is_original_subscription_purchase) || is_updated_original_subscription_purchase || is_commission_completion_purchase
       return if discount_code.blank?
 
       if offer_code.nil?
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -248,7 +248,18 @@ class Subscription < ApplicationRecord
     purchase = Purchase.new(purchase_params)
     purchase.variant_attributes = original_purchase.variant_attributes
 
-    purchase.offer_code = original_purchase.offer_code if discount_applies_to_next_charge?
+    if discount_applies_to_next_charge?
+      purchase.offer_code = original_purchase.offer_code
+      if original_discount = original_purchase.purchase_offer_code_discount
+        purchase.build_purchase_offer_code_discount(
+          offer_code: original_discount.offer_code,
+          offer_code_amount: original_discount.offer_code_amount,
+          offer_code_is_percent: original_discount.offer_code_is_percent,
+          pre_discount_minimum_price_cents: original_discount.pre_discount_minimum_price_cents,
+          duration_in_months: original_discount.duration_in_months
+        )
+      end
+    end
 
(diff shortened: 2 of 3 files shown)


app/models/purchase.rb
```diff
@@ -1999,7 +1999,7 @@ def paid?
   end
 
   def does_not_count_towards_max_purchases
-    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase
+    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase || (is_installment_payment && !is_original_subscription_purchase)
   end
```
avi-474, line 2003: Installment payments after the first purchase should not count toward the offer code’s max usage. The offer code is already used on the initial purchase, and counting later installments would incorrectly exhaust it.

app/models/purchase.rb
```diff
@@ -2411,13 +2411,14 @@ def displayed_price_per_unit_cents
...
     if has_cached_offer_code?
-      code = purchase_offer_code_discount.offer_code.code
-      purchase_offer_code_discount.offer_code_is_percent ?
-        OfferCode.new(amount_percentage: purchase_offer_code_discount.offer_code_amount, code:) :
-        OfferCode.new(amount_cents: purchase_offer_code_discount.offer_code_amount, code:)
+      p_discount = purchase_offer_code_discount
+      code = p_discount.offer_code.code
+      p_discount.offer_code_is_percent ?
+        OfferCode.new(amount_percentage: p_discount.offer_code_amount, code:) :
+        OfferCode.new(amount_cents: p_discount.offer_code_amount, code:)
```
avi-474, line 2419: Reconstruct an OfferCode from the cached `purchase_offer_code_discount` instead of relying on the persisted offer code record. This ensures the discount can still be applied even if the original offer code is deleted or has exceeded its usage limit

app/models/purchase.rb
```diff
@@ -3360,8 +3359,7 @@ def validate_quantity
 
     def validate_offer_code
       return if errors.present?
-      # accept the offer code that was used when the buyer preordered/subscribed
-      return if is_preorder_charge? || is_recurring_subscription_charge || is_gift_receiver_purchase
+      return if is_preorder_charge? || is_recurring_subscription_charge || is_gift_receiver_purchase || (is_installment_payment && !is_original_subscription_purchase) || is_updated_original_subscription_purchase || is_commission_completion_purchase
```
avi-474, line 3362: Skip offer code validation for later installment payments because the discount was already applied on the first purchase. Checking the code again could fail if it was deleted or used up, even though the buyer is still entitled to the discount.

app/models/subscription.rb
```diff
@@ -248,7 +248,18 @@ def build_purchase(override_params: {}, from_failed_charge_email: false)
...
+    if discount_applies_to_next_charge?
+      purchase.offer_code = original_purchase.offer_code
+      if original_discount = original_purchase.purchase_offer_code_discount
+        purchase.build_purchase_offer_code_discount(
+          offer_code: original_discount.offer_code,
+          offer_code_amount: original_discount.offer_code_amount,
+          offer_code_is_percent: original_discount.offer_code_is_percent,
+          pre_discount_minimum_price_cents: original_discount.pre_discount_minimum_price_cents,
+          duration_in_months: original_discount.duration_in_months
+        )
```
avi-474, line 260: Copy the saved discount from the original purchase so that all installment payments use the same discount, without rechecking or recalculating the offer code.

spec/models/installment_offer_code_persistence_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+                                 purchaser: buyer,
+                                 seller: seller)
+
+      subscription = original_purchase.subscription
+
+      # Installment plans should always apply discounts
+      expect(subscription.discount_applies_to_next_charge?).to eq(true)
+    end
+  end
+end
```
avi-474, line 154: This spec verifies that installment payments continue to receive the originally applied discount even when the offer code is exhausted or deleted. It protects against regressions where later charges incorrectly depend on the current state of the offer code.

spec/models/installment_offer_code_persistence_spec.rb
```diff
@@ -0,0 +1,154 @@
+# frozen_string_literal: true
```
EmCousin, line 1: Tests should also cover memberships, not just installments. This test also should have been in the installment_spec.rb file. Looks like this has been entirely generated by Claude

app/models/purchase.rb
```diff
@@ -2411,13 +2411,14 @@ def displayed_price_per_unit_cents
...
   def original_offer_code(include_deleted: false)
-    return nil if offer_code&.deleted? && !include_deleted
-
     if has_cached_offer_code?
-      code = purchase_offer_code_discount.offer_code.code
-      purchase_offer_code_discount.offer_code_is_percent ?
-        OfferCode.new(amount_percentage: purchase_offer_code_discount.offer_code_amount, code:) :
-        OfferCode.new(amount_cents: purchase_offer_code_discount.offer_code_amount, code:)
+      p_discount = purchase_offer_code_discount
+      code = p_discount.offer_code.code
```
EmCousin, line 2416: if the offer code has been hard-deleted, this will break

EmCousin (commented):
Thanks @avi-474, but this PR is insufficiently tested and does not completely meet the issue's requirements. The issue got eventually solved in #3066

PR #3360, opened 2026-02-05
Replace Inertia Link with anchor tags for target="_blank"

gianfrancopiana:
Fixes https://github.com/antiwork/gumroad-private/issues/92
Fixes https://github.com/antiwork/gumroad-private/issues/100

## Root Cause

Inertia.js's `<Link>` component intercepts all click events and performs SPA navigation via `router.visit()`. While this enables smooth internal navigation, it ignores the `target="_blank"` attribute entirely. As a result, clicking "open in new tab" buttons on admin pages navigated within the current tab instead of opening a new one.

## Solution

Replace Inertia `<Link>` components with standard HTML `<a>` tags for all links that specify `target="_blank"`. Standard anchor tags respect the `target` attribute and open links in new tabs as expected. These links point to external URLs or standalone pages (receipts, Stripe dashboard) that don't require SPA navigation.

## Test Plan

- Navigate to `/admin/products/{id}`, click arrow icon next to product name - opens product page in new tab
- Navigate to `/admin/purchases/{id}`, click "Go to Receipt" button - opens receipt in new tab
- On admin purchases page, click Stripe transaction link - opens Stripe in new tab
- On admin purchases page, click card fingerprint link - opens Stripe search in new tab
- On admin purchases page, click manage subscription URL - opens in new tab
- On admin users page, click username link - opens user subdomain in new tab

---

## AI Assistance Notice

### DIFF
--- a/app/controllers/url_redirects_controller.rb
+++ b/app/controllers/url_redirects_controller.rb
@@ -371,7 +371,10 @@ class UrlRedirectsController < ApplicationController
 
       if params[:access_token].present? && params[:mobile_token] == Api::Mobile::BaseController::MOBILE_TOKEN
         doorkeeper_authorize! :mobile_api
-        return if purchase && purchase.purchaser && purchase.purchaser == current_api_user
+        if current_api_user.present?
+          sign_in current_api_user
+          return if purchase && purchase.purchaser && purchase.purchaser == logged_in_user
+        end
       end
 
       if cookies.encrypted[:confirmed_redirect] == @url_redirect.token ||
--- a/app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
@@ -187,6 +187,9 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
     deleteNode();
     uploader.cancelUpload(`file_${file.id}`);
     if (file.status.type === "dropbox") void cancelDropboxFileUpload(file.id);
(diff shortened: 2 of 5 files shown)


app/javascript/components/Admin/Purchases/index.tsx
```diff
@@ -636,9 +640,9 @@ const ActionButtons = ({ purchase }: { purchase: Purchase }) => (
     ) : null}
     {purchase.successful ? (
       <Button asChild small>
-        <Link href={Routes.receipt_purchase_path(purchase.external_id)} target="_blank">
+        <a href={Routes.receipt_purchase_path(purchase.external_id)} target="_blank" rel="noreferrer noopener">
```
binary-koan: Is there a reason we're adding `noreferrer noopener` everywhere? Because we're redirecting from admin?

binary-koan (approved):
One quick question, please keep the changes if there's a reason / revert them if not. Otherwise LGTM!

app/javascript/components/Admin/Purchases/index.tsx
gianfrancopiana: Good question. I've removed `noreferrer` from internal routes (receipt, manage subscription) - but kept it on external links (Stripe, user subdomains) since we don't control those.

Modern browsers now [implicitly apply `noopener` for `target="_blank"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener), so explicit `rel="noopener"` is mainly for older browser support. [Browser support details](https://caniuse.com/mdn-html_elements_a_implicit_noopener).

PR #3368, opened 2026-02-05
Fix tiered membership restart failing when changing billing recurrence

gianfrancopiana:
Fixes https://github.com/antiwork/gumroad-private/issues/117
Fixes https://github.com/antiwork/gumroad-private/issues/112

## Root Cause

When a customer with a cancelled monthly membership tries to restart with yearly billing, they see: 
```text
The price just changed! Refresh the page for the updated price.
```

**Frontend:** Customer selects yearly billing ($77/year) and submits. Sends `perceived_price_cents: 7700`.

**Backend:** Only checked if the customer changed their tier. Since the tier was the same (only billing cycle changed), it skipped updating the subscription's price. The subscription stayed on monthly pricing.

**Validation:** Compared what the customer sent ($77) against what the backend expected ($15/month). Mismatch → error.

## Solution

Extended the change detection in `RestartAtCheckoutService` to check if either the tier OR the billing cycle changed. Now when a customer changes from monthly to yearly (same tier), the subscription's price is updated before validation.

**Validation now:** Compares $77 (sent) vs $77 (expected yearly). Match → success.

## Test Plan

- [x] Run `bundle exec rspec spec/services/subscription/restart_at_checkout_service_spec.rb` - all 8 tests pass
- [x] Verified in Rails console: `recurrence_changed?` correctly detects monthly→yearly change
- [x] Verified in Rails console: `handle_plan_change` successfully updates subscription price from monthly to yearly

---

## AI Assistance Notice

### DIFF
--- a/app/services/subscription/restart_at_checkout_service.rb
+++ b/app/services/subscription/restart_at_checkout_service.rb
@@ -28,7 +28,7 @@ class Subscription::RestartAtCheckoutService
     return error_result(error) if error.present?
 
     ActiveRecord::Base.transaction do
-      handle_tier_change if tier_changed?
+      handle_plan_change if plan_changed?
       update_payment_method if should_update_payment_method?
       subscription.resubscribe!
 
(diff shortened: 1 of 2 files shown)


app/services/subscription/restart_at_checkout_service.rb
```diff
@@ -60,30 +60,44 @@ def validate_restart
...
       perceived_price_cents = params.dig(:purchase, :perceived_price_cents)&.to_i
 
       subscription.update_current_plan!(
-        new_variants: new_variants,
-        new_price: determine_new_price,
-        perceived_price_cents: perceived_price_cents
+        new_variants: selected_variants,
+        new_price: new_price,
+        perceived_price_cents: perceived_price_cents,
+        skip_preparing_for_charge: true
```
gianfrancopiana, line 90: Added to avoid triggering card validation when updating the plan. The actual charge happens separately in `charge_subscription` if needed.

EmCousin:
@gianfrancopiana please check this :point_right: https://github.com/antiwork/gumroad-private/issues/112#issuecomment-3861070179

PR #3371, opened 2026-02-05
Fix unclear error message for invalid postal codes

gianfrancopiana:
Ref https://github.com/antiwork/gumroad-private/issues/108

## Problem

When Stripe rejects an invalid postal code, users saw either the raw Stripe error or a generic "Your account could not be updated due to an error with Stripe" message – no guidance on what actually went wrong or how to fix it.

## Approach

Catch Stripe's `postal_code_invalid` error code on both the account creation and compliance update paths, and surface a clear message that includes the country name so users know which format to check: *"The postal code you entered is not valid for Germany."*

---

## AI Disclosure

### DIFF
--- a/app/controllers/settings/payments_controller.rb
+++ b/app/controllers/settings/payments_controller.rb
@@ -83,14 +83,12 @@ class Settings::PaymentsController < Settings::BaseController
         StripeMerchantAccountManager.create_account(current_seller, passphrase: GlobalConfig.get("STRONGBOX_GENERAL_PASSWORD"))
       rescue Stripe::InvalidRequestError => e
         if e.code == "postal_code_invalid"
-          return redirect_with_error("The postal code you entered is not recognized. Please double-check your postal code and try again.")
-        end
-        return redirect_with_error(e.try(:message) || "Something went wrong.")
-      rescue => e
-        if e.try(:code) == "postal_code_invalid"
           country = current_seller.fetch_or_build_user_compliance_info.legal_entity_country
-          return redirect_with_error("The postal code you entered is not valid for #{country}.")
+          redirect_with_error("The postal code you entered is not valid for #{country}.")
+        else
+          redirect_with_error(e.try(:message) || "Something went wrong.")
         end
+        return
         return redirect_with_error(e.try(:message) || "Something went wrong.")
       end
     end


spec/requests/settings/payments_spec.rb
```diff
@@ -1276,7 +1276,7 @@
         expect(page).to have_content("Payouts will be made in EUR.")
 
         click_on("Update settings")
-        expect(page).to have_content("Invalid DE postal code")
+        expect(page).to have_content("The postal code you entered is not recognized. Please double-check your postal code and try again.")
```
EmCousin: I find the old message arguably clearer (we had the country info) and the new one to be more generic - it just happens to be more verbose. Like you said in your [comment](https://github.com/antiwork/gumroad-private/issues/108#issuecomment-3856512248), it's pretty much a non-issue in my opinion, it's the user who entered an invalid format.

If anything, the phrasing is definitely more user-friendly, so I would at least bring back the country information in the error message since we have it in our backend.

Something like "The postal code you entered is not valid for Germany. Please ensure the format is correct and try again."

gianfrancopiana: Makes sense. Updated to include the country name: "The postal code you entered is not valid for Germany." Also trimmed the second sentence since it was verbose as you pointed out and didn't match the tone of other error messages in the codebase.

app/controllers/settings/payments_controller.rb
```diff
@@ -81,7 +81,16 @@ def update
...
+      rescue Stripe::InvalidRequestError => e
+        if e.code == "postal_code_invalid"
+          return redirect_with_error("The postal code you entered is not recognized. Please double-check your postal code and try again.")
+        end
+        return redirect_with_error(e.try(:message) || "Something went wrong.")
       rescue => e
+        if e.try(:code) == "postal_code_invalid"
+          country = current_seller.fetch_or_build_user_compliance_info.legal_entity_country
+          return redirect_with_error("The postal code you entered is not valid for #{country}.")
+        end
```
EmCousin, line 90: That's a bit strange isn't it? We rescue `Stripe::InvalidRequestError` and display the "arguably generic" message here, then we rescue `StandartdError` and check whether there is a code in the error message? In what case can this happen?

Shouldn't the change have been this? 👇 

```suggestion
      rescue Stripe::InvalidRequestError => e
        if e.code == "postal_code_invalid"
          country = current_seller.fetch_or_build_user_compliance_info.legal_entity_country
          redirect_with_error("The postal code you entered is not valid for #{country}.")
        else
          redirect_with_error(e.try(:message) || "Something went wrong.")
        end
        return
```

gianfrancopiana: Thanks for catching this. Fixing now per your suggestion.

EmCousin (approved):
Looks good to me!

Issue #3373, opened 2026-02-06
Fix: Search input not auto-focusing when popover opens


4arjun:
Clicking the search button in the header opens a popover but doesn't auto-focus the input field. Users have to click a second time inside the input to start typing.

Include at least one end-to-end test.

Shine2099:
I'll fix this. The search input should auto-focus when the popover opens. Let me investigate the code and prepare a PR.

MrLawrenceKwan:
Hi, I'd like to fix this. The search input should auto-focus when the popover opens — I'll identify the component and add the focus behavior. Will submit a PR shortly.

slavingia:
Send a URL so we can test that it works?

jon-lip:
Demo:
https://gumroad-demo-3373.jonlip.com

Test credentials:
Email: seller@gumroad.com
Password: password
2FA code: 000000

Instructions:
Log in, go to Products, and click the search icon, the input will auto-focus.

PR #3374, opened 2026-02-06
fix(search): Auto-focus search input when popover opens

4arjun:
Issue: #3373 

## Summary
Clicking the search button now auto-focuses the input field, allowing users to start typing immediately without a second click.

## Problem
The search popover's input field wasn't receiving focus when opened. Users had to click the search button, then click again inside the input field to start typing.

## Root Cause
The PopoverContent component uses forceMount which keeps the content always in the DOM (hidden with CSS). This prevents Radix UI's onOpenAutoFocus callback from firing, as it relies on mount/unmount behavior.

## Solution

Use controlled open state with useEffect to programmatically focus the input when the popover opens. This pattern doesn't depend on Radix UI's internal focus management.

## Before

https://github.com/user-attachments/assets/84f3e8e8-31b6-44b9-be2c-bb6858b200f3

## After

https://github.com/user-attachments/assets/819f8e6f-f222-444b-afe0-ceac67ffdc65

# Checklist

---

# AI Disclosure
Claude Opus 4.5 was used to understand the issue and make code changes

### DIFF
--- a/app/javascript/components/Search.tsx
+++ b/app/javascript/components/Search.tsx
@@ -13,9 +13,18 @@ type SearchProps = {
 export const Search = ({ onSearch, value: initialValue, placeholder = "Search" }: SearchProps) => {
   const searchInputRef = React.useRef<HTMLInputElement>(null);
   const [searchQuery, setSearchQuery] = React.useState(initialValue);
+  const [open, setOpen] = React.useState(false);
+
+  React.useEffect(() => {
+    if (open) {
+      requestAnimationFrame(() => {
+        searchInputRef.current?.focus();
+      });
+    }
+  }, [open]);
 
   return (
-    <Popover>
+    <Popover open={open} onOpenChange={setOpen}>
       <PopoverAnchor>
         <PopoverTrigger aria-label="Toggle Search" asChild>
           <Button>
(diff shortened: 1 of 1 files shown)


app/javascript/components/Search.tsx
```diff
@@ -13,23 +13,31 @@ type SearchProps = {
 export const Search = ({ onSearch, value: initialValue, placeholder = "Search" }: SearchProps) => {
   const searchInputRef = React.useRef<HTMLInputElement>(null);
   const [searchQuery, setSearchQuery] = React.useState(initialValue);
+  const [open, setOpen] = React.useState(false);
```
4arjun, line 16: Added controlled open state to track popover visibility. This is necessary because the PopoverContent uses forceMount, which prevents onOpenAutoFocus from firing properly.

app/javascript/components/Search.tsx
```diff
@@ -13,23 +13,31 @@ type SearchProps = {
...
+
+  React.useEffect(() => {
+    if (open) {
+      requestAnimationFrame(() => {
+        searchInputRef.current?.focus();
+      });
+    }
```
4arjun, line 23: Using useEffect + requestAnimationFrame to focus the input when the popover opens. This approach is more reliable than onOpenAutoFocus when forceMount is enabled, as onOpenAutoFocus depends on mount/unmount behavior which forceMount bypasses.

binary-koan:
Could you please add tests for this? Not necessarily everywhere we use `Search` but a few important places like Products/Sales/...

Also curious if Radix documents this somewhere / has an open issue? (If not, would be nice to point to where `onOpenAutoFocus` is used in the library code so we can see where the problem comes from)

Issue #3385, opened 2026-02-06
LoadingSkeleton overflow during page transition


TenzDelek:
https://github.com/user-attachments/assets/997529bd-942b-43fe-b150-b1d1a1371bd8

TenzDelek:
not relevant after merge of #3536 . closing this

PR #3386, opened 2026-02-06
Fix: LoadingSkeleton overflow

TenzDelek:
Issue: #3385

# Description

Fixed a layout overflow issue in the LoadingSkeleton component that appeared during Inertia page transitions when clicking navigation links.
## Problem
The component was using `h-1/5` (20% height) for skeleton blocks within an `h-full` section, but the parent container's height wasn't explicitly defined in the flex layout. This caused the browser to miscalculate heights, resulting in page overflow during loading states.
## Solution
rewrite the styling to fit the skeleton `div`

---

# Before/After

### Before

https://github.com/user-attachments/assets/ea7fadfc-bdd0-4e8a-b4d2-e610f96530a4

### After

https://github.com/user-attachments/assets/1c2bc28c-2baf-4315-bf75-1f152a771d30

### Before

https://github.com/user-attachments/assets/998224d6-fe88-4097-8ed3-3250a9d6ddab

### After

https://github.com/user-attachments/assets/005cc2aa-d2ce-458c-9b5f-80902c5e21ff

---

# Test Results

UI Changes only
---

# Checklist

---

# AI Disclosure

No use of AI

### DIFF
--- a/app/javascript/components/LoadingSkeleton/index.tsx
+++ b/app/javascript/components/LoadingSkeleton/index.tsx
@@ -5,13 +5,13 @@ import { PageHeader } from "$app/components/ui/PageHeader";
 
 function LoadingSkeleton() {
   return (
-    <div className="flex-1 overflow-y-auto">
+    <div className="flex flex-1 flex-col overflow-y-auto">
       <PageHeader className="border-none" title={<Skeleton className="h-12 w-56" />} />
-      <section className="space-y-4 p-4 md:p-8">
-        <Skeleton className="h-32 w-full" />
-        <Skeleton className="h-32 w-full" />
-        <Skeleton className="h-32 w-full" />
-        <Skeleton className="h-32 w-full" />
+      <section className="flex flex-1 flex-col gap-4 p-4 md:p-8">
+        <Skeleton className="w-full flex-1" />
+        <Skeleton className="w-full flex-1" />
+        <Skeleton className="w-full flex-1" />
+        <Skeleton className="w-full flex-1" />
       </section>
     </div>
   );


app/javascript/components/LoadingSkeleton/index.tsx
```diff
@@ -5,13 +5,13 @@ import { PageHeader } from "$app/components/ui/PageHeader";
 
 function LoadingSkeleton() {
   return (
-    <div className="flex-1">
+    <div className="flex-1 overflow-y-auto">
       <PageHeader className="border-none" title={<Skeleton className="h-12 w-56" />} />
-      <section className="h-full space-y-4 p-4 md:p-8">
```
TenzDelek, line 10: removed h-full from the section element that was causing undefined height calculations

app/javascript/components/LoadingSkeleton/index.tsx
```diff
@@ -5,13 +5,13 @@ import { PageHeader } from "$app/components/ui/PageHeader";
...
-        <Skeleton className="h-1/5 w-full" />
```
TenzDelek, line 11: replaced percentage based heights (h-1/5) with fixed heights (h-32) for skeleton blocks to prevent overflow calculations from breaking.

app/javascript/components/LoadingSkeleton/index.tsx
```diff
@@ -5,13 +5,13 @@ import { PageHeader } from "$app/components/ui/PageHeader";
 
 function LoadingSkeleton() {
   return (
-    <div className="flex-1">
+    <div className="flex-1 overflow-y-auto">
```
TenzDelek: added `overflow-y-auto` to the container to match the main content area's scroll behavior

nigeltheyounger:
@TenzDelek 🚀

app/javascript/components/LoadingSkeleton/index.tsx
binary-koan: Do the percentage heights work if we make the containers `flex` (for example)? I think it would make this look better at extreme sizes if there's a way to keep that

TenzDelek: | Desktop | Mobile |
|-----------|-----------|
| Image | Image

I have implemented `flex-1` based approach as it seems to be more reasonable than the other two option. as it distributes remaining space evenly plus It adapts better to extreme screen sizes. 
cc @binary-koan

TenzDelek:
not relevant after merge of #3536. 
thank you for the initial review @binary-koan

PR #3389, opened 2026-02-06
fix: prevent search bar disappearing when no products match query

### LINKED ISSUE #3262: Search bar disappears when no products match the search query

sangokp:
## Summary

Fixes #3262 - Search bar disappears when no products match the search query

**The Problem:**
When a user searches for products and the query returns zero results:
1. The search bar disappears because it was conditionally rendered based on `products.length > 0`
2. The product table unmounts, causing its search `useEffect` to stop running
3. Subsequent searches cannot be performed because the search effect is no longer active

**The Fix:**
- Use `React.useRef` to track whether products/memberships existed on initial page load
- Keep the search bar visible as long as products existed initially (not based on current filtered results)
- Keep tables mounted during search by checking initial state, showing an empty state message when no results match
- This ensures the search `useEffect` continues to run even when search returns zero results

## Changes

| File | Change |
|------|--------|
| `ProductsDashboardPage.tsx` | Use `hasInitialProducts` ref to keep search bar visible |
| `ProductsPage.tsx` | Use refs to keep tables mounted based on initial state |
| `ProductsTable.tsx` | Show "No products found" message instead of returning null |
| `MembershipsTable.tsx` | Show "No memberships found" message instead of returning null |

## Test plan

- [ ] Navigate to Products page with existing products
- [ ] Search for a term that matches no products
- [ ] Verify search bar remains visible
- [ ] Verify "No products found" message appears
- [ ] Clear search or type new search term
- [ ] Verify search works and results appear when matching products exist
- [ ] Repeat with memberships if applicable

### DIFF
--- a/app/javascript/components/ProductsDashboardPage.tsx
+++ b/app/javascript/components/ProductsDashboardPage.tsx
@@ -36,6 +36,8 @@ export const ProductsDashboardPage = ({
 }: ProductsDashboardPageProps) => {
   const [enableArchiveTab, setEnableArchiveTab] = React.useState(archivedProductsCount > 0);
   const [query, setQuery] = React.useState("");
+  // Track if products existed on initial load to keep search bar visible during empty search results
+  const hasInitialProducts = React.useRef(products.length > 0 || memberships.length > 0);
 
   return (
     <ProductsLayout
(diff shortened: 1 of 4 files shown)


sangokp:
Friendly ping - this PR fixes the search bar disappearing when no products match the query. Ready for review whenever CI approval is available. Happy to address any feedback!

binary-koan:
Thanks for working on this, but there's another PR close to merging for this issue. In future please check and follow the contributing guidelines, e.g. AI disclosure and screencast

PR #3392, opened 2026-02-06
fix(checkout): Apply default offer code when using wanted=true parameter

### LINKED ISSUE #3271: Discount codes don't automatically apply with ?wanted=true parameter

sangokp:
## Summary

Fixes #3271 - Discount codes don't automatically apply with `?wanted=true` parameter.

**Root cause:** When redirecting to checkout with `?wanted=true`, the `code` parameter was only set from URL params (`offer_code` or `code`). It didn't fall back to the product's `default_offer_code`, unlike the normal product page rendering which considers both URL code and default_offer_code (in `product_props.rb` lines 84-105).

**Fix:** Add `@product.default_offer_code&.code` as a fallback in the redirect URL construction when neither `params[:offer_code]` nor `params[:code]` are present.

## Changes

- `app/controllers/links_controller.rb`: Added fallback to product's default offer code in the `wanted=true` redirect logic
- `spec/controllers/links_controller_spec.rb`: Added two tests:
  1. Verifies default offer code is applied when no code in URL
  2. Verifies URL code takes precedence over default offer code

## Test plan

- [ ] Visit a product URL with `?wanted=true` that has a default offer code set
- [ ] Verify the checkout page shows the discount code applied
- [ ] Visit the same product with `?wanted=true&code=OTHER` (explicit code)
- [ ] Verify the explicit code is used instead of the default

## AI Disclosure

This PR was created with AI assistance (Claude).

---

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -131,7 +131,7 @@ class LinksController < ApplicationController
         redirect_to checkout_index_url(**params.permit!, host: DOMAIN, product: @product.unique_permalink,
                                                          rent: cart_item[:rental], recurrence: cart_item[:recurrence],
                                                          price: cart_item[:price],
-                                                         code: params[:offer_code] || params[:code],
+                                                         code: params[:offer_code] || params[:code] || @product.default_offer_code&.code,
                                                          affiliate_id: params[:affiliate_id] || params[:a],
                                                          referrer: params[:referrer] || request.referrer),
                     allow_other_host: true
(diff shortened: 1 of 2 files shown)


gianfrancopiana:
Appreciate the contribution. We're going with #3272 which extracts a shared `BestOfferCodeService` to handle this. Your fix addresses the missing default code, but doesn't include the best-discount comparison between URL and default codes that the product page already does – so the `wanted=true` path would still behave differently from the normal flow. Closing in favor of #3272.

PR #3395, opened 2026-02-07
fix: date range picker overflow

abhix4:
Fixes: #3447

# Description

## Problem

The `DateRangePicker` component exceeds the container width when displayed in the PageHeader.

## Solution

- Wrapped  `DateRangePicker` in `<div className="col-span-2">` so it spans the full grid width on its own row on mobile.
-  Used `<div className="flex w-full gap-2">` to place the export button and `DateRangePicker` side by side, with `flex-1` on the picker so it fills remaining space.

---

# Before/After

**Before:**

image

**After:**

image

image

image

---

# Test Results

N/A

---

# Checklist

---

# AI Disclosure
None

### DIFF
--- a/app/javascript/components/DateRangePicker.tsx
+++ b/app/javascript/components/DateRangePicker.tsx
@@ -53,12 +53,12 @@ export const DateRangePicker = ({
       }}
     >
       <PopoverAnchor>
-      <PopoverTrigger>
-        <InputGroup aria-label="Date range selector" className="whitespace-nowrap">
-          <span suppressHydrationWarning>{Intl.DateTimeFormat(locale).formatRange(from, to)}</span>
-          <Icon name="outline-cheveron-down" className="ml-auto" />
-        </InputGroup>
-      </PopoverTrigger>
+        <PopoverTrigger>
+          <InputGroup aria-label="Date range selector" className="whitespace-nowrap">
+            <span suppressHydrationWarning>{Intl.DateTimeFormat(locale).formatRange(from, to)}</span>
+            <Icon name="outline-cheveron-down" className="ml-auto" />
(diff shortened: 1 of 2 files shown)


EmCousin:
@abhix4 the issue is missing.

abhix4:
@EmCousin Issue added and linked. Could you please review the PR when you get a chance?
Thank you

EmCousin:
I think the date picker should probably take the whole width on mobile so we don't have to wrap the date range, which is kinda ugly. That means moving it to the next row if it's part or a row already. Thoughts @laugardie?

laugardie:
We’ll probably revisit this when we improve the mobile UI/UX. For now, I agree the date can take the full row.

In cases like the screenshot below, it would be better for the first button to hug its content and let the date expand to fill the rest, so we can keep everything on a single row instead of wrapping to two.
Screenshot 2026-02-11 at 16 36 04

EmCousin:
@abhix4 can you implement Laura's feedback?

app/javascript/pages/Audience/Index.tsx
```diff
@@ -59,8 +59,10 @@ function Audience() {
                 <ExportSubscribersPopover />
               </PopoverContent>
             </Popover>
-            <DateRangePicker {...dateRange} />
-          </>
+            <div className="flex-1">
+              <DateRangePicker {...dateRange} />
+            </div>
```
abhix4: `flex-1` on the picker so it fills remaining space.

abhix4:
`Sales`
image

`Churn`
image

`Following`
image


@EmCousin Addressed the suggested changes. Could you please review? Thanks

PR #3399, opened 2026-02-08
Add backup payment method for refund funding

yashranaway:
Fixes : #1594

Previous PR: #2425
# Description

## Problem

Sellers can't process refunds when their Gumroad balance is insufficient to cover the refund amount.

## Solution

Enables sellers to add a backup payment method (credit card) that automatically covers refunds when their balance is too low.

When a seller visits the Customers page, they see a banner prompting them to set up a backup payment method. Clicking the link takes them to Payments Settings where they can add a card using Stripe Elements.

Once configured, when a refund is issued and the seller's balance can't cover it, the system automatically charges their backup card for the difference. The seller receives an email confirmation, and the refund processes normally for the customer.

Sellers can change or remove their backup card anytime from Payments Settings, with a confirmation dialog before removal. If they dismiss the banner without adding a card, it stays hidden. If they later remove their card, the banner reappears.

**Design:** [Figma](https://www.figma.com/design/bQE0b8eQTM2eov0Kzjxk3A/Ability-to-load-funds-to-Gumroad-balance-to-cover-refunds-when-balance-is-too-low%C2%A0?node-id=3-4663&t=IAAPYaCaxBvc0iqo-1)

> [!NOTE]
> **ZIP/Postal Code**: I moved the ZIP field out of the Stripe `CardElement` into a dedicated standard input. This was done to bypass Stripe's automatic hiding logic on mobile/small widths, ensuring the field remains visible and consistently accessible across all devices. Can revert back as per original design if this approach isn't ideal.

---

# Before/After

### Desktop
https://github.com/user-attachments/assets/cd877fb9-8fd8-43c9-bbb2-fb30d0b9c443

### Remove card modal (newly added)

https://github.com/user-attachments/assets/335dd222-88d4-45cf-900b-3d6c4c57e38e

### Mobile
image

image

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

Used Claude Opus 4.6 for implementation.

### DIFF
--- a/app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
+++ b/app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
@@ -50,17 +50,9 @@ export const RefundPaymentMethodBanner = ({ show }: Props) => {
           <Link href={`${Routes.settings_payments_path()}#refund-payment-method`}>Set up backup method</Link>
         </div>
       </div>
-      <button
-        type="button"
-        className="link"
-        onClick={handleDismiss}
-        disabled={isDismissing}
-        aria-label="Dismiss"
-      >
+      <button type="button" className="link" onClick={handleDismiss} disabled={isDismissing} aria-label="Dismiss">
         close
       </button>
     </Alert>
   );
 };
-
-
--- a/app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
@@ -142,10 +142,7 @@ const RefundPaymentMethodForm = ({
               <FieldsetTitle>Card details</FieldsetTitle>
               <InputGroup readOnly>
                 <Icon name="outline-credit-card" />
-                <Input
-                  readOnly
-                  value={`•••• •••• •••• ${savedCard.visual}`}
-                />
+                <Input readOnly value={`•••• •••• •••• ${savedCard.visual}`} />
                 <span className="text-muted">
                   {savedCard.expiry_month.toString().padStart(2, "0")}/{savedCard.expiry_year.toString().slice(-2)}
                 </span>
@@ -237,7 +234,7 @@ const RefundPaymentMethodForm = ({
                 color="danger"
                 onClick={() => {
                   setConfirmingRemove(false);
-                  void handleRemove();
+                  handleRemove();
                 }}
                 disabled={isSubmitting}
               >
@@ -247,8 +244,8 @@ const RefundPaymentMethodForm = ({
           }
         >
           <h4>
-            Are you sure you want to remove your refund payment method? Refunds that exceed your balance will no longer be
-            covered automatically.
+            Are you sure you want to remove your refund payment method? Refunds that exceed your balance will no longer
+            be covered automatically.
           </h4>
         </Modal>
       ) : null}
@@ -267,5 +264,3 @@ export const RefundPaymentMethodSection = ({
     <RefundPaymentMethodForm refundPaymentMethod={refundPaymentMethod} isFormDisabled={isFormDisabled} />
   </StripeElementsProvider>
 );
-
-


pegasus1134:
@yashranaway Please add me as a collaborator so I can help better with the changes.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,271 @@
...
+              </Button>
+            </>
+          }
+        >
+          <h4>
+            Are you sure you want to remove your refund payment method? Refunds that exceed your balance will no longer be
+            covered automatically.
+          </h4>
+        </Modal>
+      ) : null}
```
yashranaway, line 251: Added a `<Modal>` confirmation before removing the card, same pattern as products and wishlists.

app/javascript/components/CustomersPage/RefundPaymentMethodBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+      });
+
+      const result = cast<{ success: boolean }>(await response.json());
+
+      if (result.success) {
+        setIsVisible(false);
+      }
+    } catch (e) {
+      assertResponseError(e);
+      showAlert("Sorry, something went wrong.", "error");
```
yashranaway, line 36: Added `showAlert("Sorry, something went wrong.", "error")` in the catch block.

app/modules/purchase/refundable.rb
```diff
@@ -45,8 +45,28 @@ def refund_and_save!(refunding_user_id, amount_cents: nil, is_for_fraud: false)
...
+              amount_cents: charge_amount,
+              purchase: self
+            ).perform
+
+            if !result.success?
+              errors.add :base, result.error_message || "Could not charge your backup card to cover the refund shortfall."
+              return false
+            end
+
+            if amount_cents_to_refund > seller.reload.unpaid_balance_cents
```
yashranaway, line 63: Added a `seller.reload.unpaid_balance_cents` check after the backup card charge completes to catch concurrent refunds.

app/mailers/contacting_creator_mailer.rb
```diff
@@ -532,6 +532,13 @@ def ping_endpoint_failure(user_id, ping_url, response_code)
     @subject = "Webhook ping endpoint delivery failed"
   end
 
+  def refund_funding_charge_confirmation(credit_id:)
+    @credit = Credit.find(credit_id)
+    @seller = @credit.user
+    @amount = Money.new(@credit.amount_cents, :usd).format(no_cents_if_whole: true, symbol: true)
+    @subject = "Your backup card was charged #{@amount}"
+  end
+
```
yashranaway, line 541: Moved the `Money.new(...)` formatting into `@amount` in the mailer so it's shared between the subject and template without duplicating the logic.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  private
+
+  attr_reader :user, :amount_cents, :purchase, :credit_card
+
+  def charge_credit_card
+    payment_intent = Stripe::PaymentIntent.create(
+      {
+        amount: amount_cents,
+        currency: Currency::USD,
```
yashranaway: Replaced all direct ivar references with `attr_reader :user, :amount_cents, :purchase, :credit_card`.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+  rescue Stripe::CardError => e
+    { success: false, error: e.message }
+  rescue Stripe::InvalidRequestError => e
+    Bugsnag.notify(e)
+    { success: false, error: "Invalid payment request." }
+  rescue Stripe::StripeError => e
+    Bugsnag.notify(e)
+    { success: false, error: "Payment processor error. Please try again." }
+  end
```
yashranaway: @binary-koan , Looked into this, `with_stripe_error_handler` raises `ChargeProcessor*Error` exceptions which are designed for the charge processor layer. Our service returns user-friendly error messages via Result structs, so wrapping with it would add a layer of exception catching on top. Kept the current approach but happy to refactor if you have a pattern in mind.

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        }.to change(CreditCard, :count).by(1)
+
+        expect(response).to be_successful
+        json = JSON.parse(response.body)
+        expect(json["success"]).to be true
+        expect(json["enabled"]).to be true
+        expect(json["credit_card"]).to be_present
+        expect(seller.reload.refund_funding_credit_card).to be_present
+      end
+    end
```
yashranaway, line 102: Added a "when chargeable is valid" context that stubs `CardParamsHelper` and verifies the card gets created and associated.

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    context "when no funding card exists" do
+      it "still returns success with state" do
+        delete :destroy, format: :json
+
+        expect(response).to be_successful
+        json = JSON.parse(response.body)
+        expect(json["success"]).to be true
+        expect(json["enabled"]).to be false
+      end
+    end
```
yashranaway, line 140: Added "when no funding card exists" for destroy and "when banner is already dismissed" for dismiss_banner.

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    context "when banner is already dismissed" do
+      before { seller.update!(dismissed_refund_payment_method_banner: true) }
+
+      it "still returns success" do
+        post :dismiss_banner, format: :json
+
+        expect(response).to be_successful
+        expect(seller.reload.dismissed_refund_payment_method_banner?).to be true
+      end
+    end
```
yashranaway, line 162: [Same](https://github.com/antiwork/gumroad/pull/3399/changes#r2778594480)

app/controllers/settings/refund_funding_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      enabled: funding_card.present?,
+      credit_card: funding_card.present? ? {
+        id: funding_card.id,
+        visual: funding_card.visual,
+        card_type: funding_card.card_type,
+        expiry_month: funding_card.expiry_month,
+        expiry_year: funding_card.expiry_year
+      } : nil
+    }
+  end
```
yashranaway: Extracted a shared `refund_funding_data` method in the controller. Now `show`, `create`, and `destroy` all return the same card/enabled state so the frontend stays in sync.

spec/controllers/settings/refund_funding_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        expect(seller.reload.refund_funding_credit_card).to be_nil
+      end
+
+      it "resets the dismissed banner flag so it reappears" do
+        seller.update!(dismissed_refund_payment_method_banner: true)
+
+        delete :destroy, format: :json
+
+        expect(seller.reload.dismissed_refund_payment_method_banner?).to be false
+      end
```
yashranaway, line 128: The `destroy` action now resets `dismissed_refund_payment_method_banner` to false, so the banner reappears if the seller removes their card, matching the PR description.

yashranaway:
Hey @binary-koan _a , I've addressed all the comments from last PR. Could you please take another look when you get a chance? Thanks!

yashranaway:
> @yashranaway Please add me as a collaborator so I can help better with the changes.

Hey, thanks! I don’t have permissions to add collaborators on my end 😅
Could you take a look at the implementation though? I’ve wrapped everything up, let me know if anything’s missing.

pegasus1134:
Hey @yashranaway, the core feature implementation is solid and you've addressed almost all of @binary-koan's feedback from #2425. A few things need fixing before this is mergeable though. I've split them into **must-fix** and **should-fix**.

---

### Must-Fix

#### 1. Unrelated changes still in the diff (leftover from squash)

The "Remove unrelated changes from squash" commit (`eeb60a8`) didn't fully clean things up. These files still contain changes that have nothing to do with the refund funding feature:

- **`app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx`** — ~1300 lines changed. Helper functions (`getBusinessTypes`, `getBusinessStateConfig`, `getIndividualStateConfig`, `getBusinessTaxIdConfig`, `getIndividualTaxIdConfig`, `renderStateSelect`) are removed, Japan prefecture support is deleted, layout is restructured into deeply nested ternaries. None of this is related to backup payment methods.
- **`app/controllers/settings/payments_controller.rb`** — The specific `Stripe::InvalidRequestError` handler for `postal_code_invalid` (which returns a user-friendly country-specific message) is replaced with a bare `rescue => e`. This catches *all* exceptions (including `TypeError`, `NoMethodError`, etc.) and passes raw error messages to the user. This is both a security concern (leaking internal error details) and a UX regression (losing the friendly postal code message).
- **`app/presenters/settings_presenter.rb`** — Japan states data removed from the presenter.
- **`spec/requests/settings/payments_spec.rb`** — The entire Brazil business compliance test is deleted.

**What to do:** Revert every change in these four files so only refund-funding-related changes remain in the PR. The simplest approach:

```bash
git checkout main -- app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
git checkout main -- app/controllers/settings/payments_controller.rb
git checkout main -- app/presenters/settings_presenter.rb
git checkout main -- spec/requests/settings/payments_spec.rb
```

Then commit the revert. This alone accounts for the bulk of the diff and is the biggest blocker.

---

#### 2. Idempotency key includes a timestamp — can cause double charges

In `app/services/refund_funding_charge_service.rb`:

```ruby
{ idempotency_key: "refund_funding_#{user.id}_#{purchase&.id}_#{Time.current.to_i}" }
```

`Time.current.to_i` is a Unix timestamp in **seconds**. If a request is retried even 1 second later, the key is different, so Stripe treats it as a brand new charge. This defeats the entire purpose of the idempotency key and can result in **double-charging the seller's backup card**.

**What to do:** Make the key deterministic based on the operation, not the time:

```ruby
{ idempotency_key: "refund_funding_#{user.id}_#{purchase.id}" }
```

This way, any retry for the same user + purchase combination is correctly deduplicated by Stripe.

---

#### 3. Confirmation email fires before the refund actually completes

In `refund_funding_charge_service.rb`, `send_confirmation_email` is called right after the credit is created inside `perform`. But at that point, the Stripe refund hasn't happened yet — that occurs later in the `refundable.rb` flow after the balance is topped up.

If the subsequent refund fails for any reason (Stripe error, validation failure, etc.), the seller has already received an email saying *"the refund has been processed for your customer"* when it hasn't.

**What to do:** Remove `send_confirmation_email` from the service. Instead, send the email from `refundable.rb` **after** the refund has actually succeeded. You already have the `credit` object from the service result to pass to the mailer. Something like:

```ruby
# In refundable.rb, after the refund succeeds:
if result[:credit].present?
  ContactingCreatorMailer.refund_funding_charge_confirmation(credit_id: result[:credit].id).deliver_later
end
```

This also fixes the related race condition issue: if the `seller.reload.unpaid_balance_cents` check after the charge fails (concurrent refund drained balance), the email won't be sent for a refund that didn't go through.

---

#### 4. `purchase&.id` safe navigation — drop the `&.`

In `refund_funding_charge_service.rb`, both the Stripe metadata and the idempotency key use `purchase&.id`. But this service is only ever called from `refundable.rb` where `self` *is* the purchase — it can never be nil.

The safe navigation is dangerous here because if `purchase` were somehow nil, you'd silently get `nil` in the idempotency key (making it non-unique) and `nil` in the Stripe metadata (losing the audit trail), instead of getting a loud error that tells you something went wrong.

**What to do:** Replace `purchase&.id` with `purchase.id` everywhere in the service. If it's nil, you want the `NoMethodError` — that's a bug that should be caught, not masked.

---

### Should-Fix (non-blocking but worth addressing)

#### 5. SCA / 3D Secure handling for EU sellers

With `off_session: true` + `confirm: true`, EU cards requiring 3D Secure will return `status: "requires_action"`. The current error message is:

> "Payment was not successful. Status: requires_action"

This isn't helpful to the seller. Consider something like:

> "Your backup card requires additional authentication and can't be charged automatically. Please update your backup payment method with a card that doesn't require 3D Secure verification."

#### 6. Migration uses `type: :integer` instead of `bigint`

In `db/migrate/20260130000001_add_refund_funding_to_users.rb`:

```ruby
add_reference :users, :refund_funding_credit_card, type: :integer, index: true
```

Other FK references to `credit_cards` in the codebase use `bigint`. Minor inconsistency — if `credit_cards.id` is bigint, this FK should match to avoid potential overflow issues down the line.

#### 7. Tailwind `!important` overrides

```tsx
<section id="refund-payment-method" className="p-4! md:p-8!">
```

The `!` suffix forces `!important`, which usually means a specificity fight somewhere. If possible, resolve the underlying specificity issue instead.

---

### What's done well

- All of binary-koan's review feedback from #2425 is addressed (27/28 items, with a reasonable justification for the one exception re: `with_stripe_error_handler`)
- `Credit.create_for_refund_funding!` follows the existing factory pattern (`create_for_dispute_won!`, `create_for_returned_payment_difference!`, etc.) perfectly
- FlagShihTzu bit 53 for banner dismissal is the correct Gumroad pattern
- `refund_funding_data` shared method keeps all controller responses consistent
- Modal confirmation on card removal, error alerts on failures — all the UX details are there
- Test coverage for the feature-specific code is solid

---

### Summary

| # | Issue | Severity | Effort |
|---|-------|----------|--------|
| 1 | Revert unrelated files (AccountDetailsSection, payments_controller, settings_presenter, payments_spec) | **Blocker** | 4 git checkout commands + commit |
| 2 | Fix idempotency key (remove timestamp) | **Blocker** | 1 line change |
| 3 | Move confirmation email to after refund succeeds | **Blocker** | ~10 lines moved |
| 4 | Drop `&.` safe navigation on `purchase` | **Blocker** | Find & replace |
| 5 | Better SCA error message | Nice-to-have | 1 string change |
| 6 | Migration integer → bigint | Nice-to-have | 1 line change |
| 7 | Remove Tailwind `!important` | Nice-to-have | CSS investigation |

Once items 1–4 are fixed this should be good to go. The feature implementation itself is clean. 👍

yashranaway:
@pegasus1134 thanks, pushed [bdbec6b](https://github.com/antiwork/gumroad/pull/3399/commits/bdbec6b104a40aa1c2b7ba655dddce7494feb746) btw for the integer vs bigint comment, The `credit_cards` table uses `id: :integer` in the schema (`db/schema.rb:617`), so `type: :integer` on the FK is correct and matches the referenced column. No change needed here.

pegasus1134:
Thanks @yashranaway, all blockers are fixed.                                                                                                                                                         

  You're right on the integer vs bigint point — `credit_cards` uses `id: :integer` in the schema, so `type: :integer` on the FK is correct.

  One small thing remaining:

  There are still a couple of unrelated Japan-specific changes that look like squash leftovers:

  **`app/javascript/pages/Settings/Payments/Show.tsx`:**
  - Removes `jp` from the `states` type definition
  - Removes the `form.data.user.country !== "JP"` check from city validation — this would now require Japanese sellers to fill in a city field they previously didn't need to

  **`spec/presenters/settings_presenter_spec.rb`:**
  - Removes `business_building_number_kana`, `building_number_kana`, and `jp: Compliance::Countries.japan_prefectures_for_select` from test expectations

  These are only a few lines each, but the city validation change in Show.tsx could affect Japanese sellers. If these are intentional (maybe from a separate effort to clean up JP fields), ignore this.

Otherwise looks good to go!

pegasus1134:
@yashranaway The code is clean now, good to go.

@binary-koan I think we are in a mergeable state, do run CI when you get the time. If there's any other points you notice do the let us know. Thank you!

EDIT: Further concerns added below for cleaner merge.

pegasus1134:
Hey @yashranaway, noticed a few more things that may need to be fixed.

1. No charge reversal if refund fails after backup card is charged

  In refundable.rb, there are multiple paths where the backup card charge succeeds but the refund never completes:

  - If the seller.reload.unpaid_balance_cents re-check fails (concurrent refund drained balance between charge and check)
  - If ChargeProcessor.refund! raises (AlreadyRefundedError, InsufficientFundsError, InvalidRequestError, UnavailableError) — all rescue blocks return false with no reversal
  - If refund_purchase!'s DB transaction rolls back — the Stripe operations already happened

  In all cases, the seller's card was charged and a Credit was created, but there's no Stripe::Refund.create to reverse the PaymentIntent. The money sits in the balance with no visibility into what happened.

  Suggest adding a reverse_funding_charge! path that issues Stripe::Refund.create(payment_intent: payment_intent_id) whenever a downstream step fails after a successful backup charge.

  Also worth noting: the idempotency key prevents double-charging at Stripe, but if the seller retries, Credit.create_for_refund_funding! will create a second Credit for the same PaymentIntent since there's no
  DB-level uniqueness guard.

  2. defined?(funding_result) is fragile

  defined? on local variables returns truthy if the variable was assigned anywhere in the lexical scope, even if that branch never executed. Initialize funding_result = nil before the conditional and just check
  funding_result&.credit.present?.

  3. No test for confirmation email

  Zero test coverage verifying ContactingCreatorMailer.refund_funding_charge_confirmation is enqueued after a successful backup-funded refund. Combined with the defined? pattern, this could silently break.

  4. Email queue

  .deliver_later(queue: "critical") — this is an informational seller notification, not a critical job. Should match whatever queue other ContactingCreatorMailer methods use.

  5. Minor inconsistencies
  - Controller returns id in credit_card hash but the TS type doesn't include it and frontend never uses it
  - Both React components have unused export default alongside named exports
  - Banner link uses <a href> instead of Inertia Link (full page reload vs SPA transition)
  - No .text.erb email template
  - Tests use OpenStruct.new for Stripe mocks (deprecated Ruby 3.4+)

  #1 is the only one that seems like it needs a quick fix. I have added the full detailed breakdown in the below comment so it is easier for you to add. Please check and proceed!

pegasus1134:
### Fixing "No charge reversal if refund fails after backup card is charged"                                                                                                                                     
                  
  **Failure Point 1: Balance re-check after charge**                                                                                                                                                               
                                                                                                                                                                                                                   
  After `RefundFundingChargeService` succeeds (card charged, Credit created), the `seller.reload.unpaid_balance_cents` check can still fail (concurrent refund drained balance). Returns `false` — no reversal.

  **Failure Point 2: `ChargeProcessor.refund!` raises**

  Every rescue block (`ChargeProcessorAlreadyRefundedError`, `InsufficientFundsError`, `InvalidRequestError`, `UnavailableError`) just logs and returns `false` — no reversal.

  **Failure Point 3: `refund_purchase!` transaction rolls back**

  If the DB transaction inside `refund_purchase!` fails, the Stripe refund to the buyer already happened AND the backup card was already charged. DB rolls back but both Stripe operations persist — worst case.

  **Bonus: Idempotency gap at DB level**

  The idempotency key `"refund_funding_#{user.id}_#{purchase.id}"` prevents double-charging at Stripe. But if the seller retries, `Credit.create_for_refund_funding!` runs again and creates a second Credit for
  the same PaymentIntent. No DB-level idempotency guard.

  ### The fix needs:

  1. A `reverse_funding_charge!` method that calls `Stripe::Refund.create(payment_intent: payment_intent_id)`
  2. That method called in every rescue path where `funding_result&.success?` is true
  3. Same reversal if the Step 3 balance re-check fails
  4. An idempotency guard on `Credit.create_for_refund_funding!` (e.g., unique index on `refund_funding_purchase_id`)

The money isn't actually lost. When the backup card is charged, Credit.create_for_refund_funding! creates a Credit + BalanceTransaction that increases seller.unpaid_balance_cents. So if the downstream refund
  fails, the seller's balance genuinely went up by that amount. They can retry the refund and it'll likely work because the balance is now sufficient.

  The real problem is UX confusion: the seller's card got charged, the refund failed, they see "Your balance is insufficient to process this refund" — but nobody told them their card was charged. No email (only
  sends on success). No explanation.

yashranaway:
Let’s wait for the reviewers before making further changes.

binary-koan:
Will hopefully get to this tomorrow, sorry for the delay

binary-koan:
@pegasus1134 By the way - comments like [this](https://github.com/antiwork/gumroad/pull/3399#issuecomment-3866248845) and [this](https://github.com/antiwork/gumroad/pull/3399#issuecomment-3879160886) look very AI-generated, and per contributing guidelines we don't allow long AI explanations to be pasted as-is. Apologies if that's not the case, but either way please use your own words, keep it concise, and use inline comments as it's much easier to check potential problems

pegasus1134:
> @pegasus1134 By the way - comments like [this](https://github.com/antiwork/gumroad/pull/3399#issuecomment-3866248845) and [this](https://github.com/antiwork/gumroad/pull/3399#issuecomment-3879160886) look very AI-generated, and per contributing guidelines we don't allow long AI explanations to be pasted as-is. Apologies if that's not the case, but either way please use your own words, keep it concise, and use inline comments as it's much easier to check potential problems

Apologies for this, I do sometimes use AI to further elaborate or add details to the comment after I write the direction myself, assuming someone else may be using an AI agent to make edits so it can understand better!

Your points are well noted. Going forward I'll keep it short and use inline comments and restrict AI usage. Thanks!

spec/modules/purchase/refundable_spec.rb
```diff
@@ -0,0 +1,172 @@
+# frozen_string_literal: true
```
binary-koan, line 1: Would be nicer to make this `spec/modules/purchase/refundable_spec.rb`

yashranaway:
@binary-koan I have addressed those comments in [5a9d7d3](https://github.com/antiwork/gumroad/pull/3399/commits/5a9d7d3164213b43a0713ce05dd89ad16c996ed2) could you run CI once?

yashranaway:
Sorry @binary-koan I missed out one fix requesting a rerun 😅

yashranaway:
@binary-koan Hey, I have added fixes for tests and robocop, only 1 test was related.

yashranaway:
image

yashranaway:
@binary-koan Light ping on this

yashranaway:
@binary-koan sorry, for double ping . Could you please run CI again?

binary-koan:
Running CI here to check, but as Ershad mentioned in the issue we need to discuss this more - there were some concerns when we talked about it with the team today. Sorry that only came up after you put so much work into this, we should have scoped the issue properly and finalised the db schema internally before inviting contributions.

Will update the issue when we've talked it through.

yashranaway:
> Running CI here to check, but as Ershad mentioned in the issue we need to discuss this more - there were some concerns when we talked about it with the team today. Sorry that only came up after you put so much work into this, we should have scoped the issue properly and finalised the db schema internally before inviting contributions.
> 
> Will update the issue when we've talked it through.

No worries at all, hopefully the approach holds up after the discussion, happy to make any changes if needed! Also just a heads up, the CI failures on this run are all unrelated.

slavingia:
Thanks for the contribution! We're shifting our open source focus to smaller, easier-to-verify changes like bug fixes. We're closing PRs for larger features for now — but please don't hesitate to pick up a bug or smaller issue. We appreciate your help!

yashranaway:
@slavingia Thanks for the clarification, I understand the shift in focus toward smaller changes.

This PR was already reviewed and we were at the CI stage with all requested updates addressed. Given how close it was to completion, would you be open to taking a final look before closing it?

Either way, I appreciate the direction and will focus on smaller issues going forward.


cc: @binary-koan

slavingia:
I'll let each engineer make that call, thanks for the heads up!

yashranaway:
> I'll let each engineer make that call, thanks for the heads up!

If possible, could you please ensure this reaches the relevant person? and thanks Sahil for everything. 🙏

binary-koan:
It's reached me :) Will definitely take another look at this once we've scoped the issue properly (sorry I can't give an ETA on that, but we're trying to cut down the backlog and address things more promptly)

yashranaway:
thanks, hopefully it gets reopened and I'll be open to change a few things (if needed)

PR #3402, opened 2026-02-08
fix: audience CSV export timeout by replacing find_each with raw SQL iteration 

dluffy56:
## Description

Fixes Audience CSV export timeouts for large sellers caused by `find_each` instantiating ActiveRecord objects for every row. Replacing it with `exec_query` enables raw row iteration and cuts export time roughly in half.

---

#### `AudienceExportService` times out in Sidekiq for sellers with large audiences (#2092)

The root cause is `find_each` instantiating ActiveRecord objects for every row - for ~1M audience members, this takes **~16s** purely in Ruby object allocation, not SQL execution.

---

## Root Cause Analysis

The timeout is not caused by slow SQL - the query completes in **~1.3s** for ~1M rows. The problem is `find_each` combined with ActiveRecord object instantiation on the Ruby side.

**Key findings:**

- `find_each` instantiates a full ActiveRecord object per row. For ~1M rows, this alone takes **~16s** in Ruby object allocation.
- The original `.order(:min_created_at)` was dead code - `find_each` silently ignores `.order()` and forces `ORDER BY id` internally.
- UNION-based query rewrites make `find_each` worse, not better, each batch re-executes the entire UNION since `find_each` appends `WHERE id > ? ORDER BY id LIMIT 1000` per batch.
- Index optimizations on boolean columns have minimal impact since SQL execution was never the bottleneck.

---

## Solution

Replaced `find_each` with `exec_query` and raw row iteration. `exec_query` returns lightweight hashes instead of ActiveRecord objects, significantly reducing Ruby-side overhead.

### Changes

- Removed `:id` from `select` (not needed without AR instantiation)
- Changed `.order(:min_created_at)` to `.order(:id)` - `find_each` silently ignores `.order()` and forces `ORDER BY id`, so the original ordering was dead code. This preserves existing production behavior.

---

## Benchmarks (1M audience members)

| Approach                              | Time   |
|--------------------------------------|--------|
| Original (find_each + AR objects)    | 16.38s |
| This PR (exec_query + raw rows)      | 7.99s  |

---

### Note on memory

`exec_query` loads all result rows into memory as hashes rather than streaming them. For ~1M rows with 2 columns (email + timestamp), this is approximately **100–150MB**, which is well within typical Sidekiq worker memory limits (512MB–2GB).

---

## Before / After

Backend-only changes

---

## Test Results

- spec/services/exports/audience_export_service_spec.rb  
- spec/sidekiq/exports/audience_export_worker_spec.rb  
- spec/controllers/audience_controller_spec.rb  

Screenshot from 2026-02-08 11-20-50

---

## Checklist

- [x] I have read https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md
- [x] I have watched https://www.youtube.com/@anti-work

---

## AI Disclosure

Claude Opus 4.5 was used to research `exec_query` vs `find_each` performance characteristics and to verify ordering behavior. Implementation and benchmarking were done manually


app/services/exports/audience_export_service.rb
```diff
@@ -18,17 +18,19 @@ def perform
...
-      query = @user.audience_members.select(:id, :email, :min_created_at)
-
       conditions = []
       conditions << "follower = true" if @options[:followers]
       conditions << "customer = true" if @options[:customers]
       conditions << "affiliate = true" if @options[:affiliates]
 
-      query = query.where(conditions.join(" OR "))
+      sql = @user.audience_members
+        .select(:email, :min_created_at)
```
dluffy56, line 27: Dropped `:id` from select since we no longer instantiate ActiveRecord objects or rely on `find_each` batching. The export only needs `email` and `min_created_at`, so avoiding the extra column reduces data transfer and Ruby-side object allocation without changing behavior

app/services/exports/audience_export_service.rb
```diff
@@ -18,17 +18,19 @@ def perform
...
+        .where(conditions.join(" OR "))
+        .order(:id)
```
dluffy56, line 29: `find_each` forces `ORDER BY id`, so the previous `.order(:min_created_at)` was ineffective. Explicitly ordering by id preserves existing behavior

app/services/exports/audience_export_service.rb
```diff
@@ -18,17 +18,19 @@ def perform
...
+        .to_sql
 
-      query.order(:min_created_at).find_each do |member|
-        csv << [member.email, member.min_created_at]
+      ActiveRecord::Base.connection.exec_query(sql).each do |row|
```
dluffy56, line 32: `exec_query` avoids ActiveRecord instantiation and dramatically reduces Ruby-side overhead (~8s saved for ~1M rows in testing!)

gianfrancopiana:
The issue has been updated with the confirmed root cause and production EXPLAIN output – the timeout is at the SQL layer, not AR allocation. Closing in favor of the approach outlined there.

dluffy56:
@gianfrancopiana Thanks for the info, is the issue handled internally by team or it's still open for contribution?

gianfrancopiana:
> @gianfrancopiana Thanks for the info, is the issue handled internally by team or it's still open for contribution? 

Still open.

PR #3377, opened 2026-02-06
fix: search bar disappears when no results match query

abhix4:
Issue: #

# Description

Ensures the  search bar remains visible even when no items match the search query.

## Problem

When searching for **discounts/upsells** that return no matching results, the search bar disappears entirely.

## Solution

Keep the search bar visible whenever a search query is present.

---

# Before/After

BEFORE:

https://github.com/user-attachments/assets/c574eaf9-3de2-40cd-ab7a-63301ef361f8

AFTER:

https://github.com/user-attachments/assets/93c9233c-bedb-47b0-9033-e264cf510764

https://github.com/user-attachments/assets/7202f8ed-b26d-46b5-8e8b-675ec9b54d30

---

# Test Results

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [ ] I have added/updated tests for my changes

---

# AI Disclosure

Copilot

PR #3420, opened 2026-02-08
Fix search bar disappearing when no products match the search query

### LINKED ISSUE #3262: Search bar disappears when no products match the search query

aifunmobi:
## Summary

Fixes #3262

## Root cause

The search bar and product/membership tables are conditionally rendered based on the *current* `products.length` and `memberships.length`. When a search returns zero results, Inertia updates these arrays to empty, causing:

1. **Search bar unmounts** — `ProductsDashboardPage` renders `<Search>` only when `products.length > 0`, so it disappears
2. **Table components unmount** — `ProductsPage` gates `<ProductsPageProductsTable>` on `products.length > 0`, so the `useEffect` watching `props.query` stops running
3. **Dead end** — the user cannot modify or clear the search because both the input and the reload mechanism are gone

## Fix

Use `React.useRef` to capture whether products/memberships existed on initial page load. This distinguishes "user has no products" (show placeholder) from "search returned no results" (keep search bar and tables mounted).

**`ProductsDashboardPage.tsx`**
- Show `<Search>` based on initial item existence (`hadInitialItems.current`), not current array length
- Show `<ProductsPage>` instead of the placeholder when the ref indicates items existed initially

**`ProductsPage.tsx`**
- Keep table components mounted when they had entries initially AND a search query is active, so the search `useEffect` continues to fire

**`ProductsTable.tsx` / `MembershipsTable.tsx`**
- Display "No products found" / "No memberships found" when there's an active search with zero results, instead of rendering nothing

**`spec/requests/products/index_spec.rb`**
- Added e2e test that searches for a non-existent product, verifies the search bar stays visible with a "No products found" message, then searches again successfully

## Test plan

- [ ] Search for a non-existent product → search bar stays visible, "No products found" shown
- [ ] Clear the search → all products reappear
- [ ] Search for an existing product → filtered results shown correctly
- [ ] User with no products → placeholder still shown (not the empty table)
- [ ] Verify in both light and dark mode, mobile and desktop

## AI Disclosure

This PR was authored with AI assistance.

### DIFF
--- a/app/javascript/components/ProductsDashboardPage.tsx
+++ b/app/javascript/components/ProductsDashboardPage.tsx
@@ -36,6 +36,7 @@ export const ProductsDashboardPage = ({
 }: ProductsDashboardPageProps) => {
   const [enableArchiveTab, setEnableArchiveTab] = React.useState(archivedProductsCount > 0);
   const [query, setQuery] = React.useState("");
+  const hadInitialItems = React.useRef(products.length > 0 || memberships.length > 0);
 
   return (
     <ProductsLayout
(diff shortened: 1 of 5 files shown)


aifunmobi:
Closing — the fix was already merged via #3377. Nice work @abhix4!

Issue #3313, opened 2026-02-02
Cant login


G4R1C:
<img width="2558" height="1398" alt="Image" src="https://github.com/user-attachments/assets/1fdf475c-75a4-4f51-94c5-8db6c100fef6" />
When i press the log in button its show login screen for a split second and then it is just empty window.

G4R1C:
browser firefox OS Linux

machin3io:
Same issue. Not sure why this was closed, as the issue persists.

PR #3432, opened 2026-02-09
Fix Firefox CSRF token error on login and signup pages

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad/issues/3313

# Description

## Problem

On production Firefox, the login and signup pages throw a console error:

```
Error: Expected root to be a union, but no branches matched:
• Error: Expected root to be a string, got undefined
• Error: Expected root to be an object, got undefined
```

This originates from `SocialAuthButton.tsx`, which reads the CSRF token from the DOM via `document.querySelector("meta[name=csrf-token]")` but on Firefox meta tag not exist

## Solution
Replace DOM-based CSRF token lookup (`$("meta[name=csrf-token]").attr("content")`) with `usePage().props.authenticity_token` from Inertia's shared page props. This token is already provided server-side via `inertia_share` in `InertiaRendering` and is always available.

---

# Before/After

## Before
The issue not reproducible on localhost
But on production on Firefox there is no csrf-token meta:
image

## After

https://github.com/user-attachments/assets/a4bf4a6d-c1a5-40b4-8e4f-d94ad43bed08

---

# AI Disclosure

Claude Opus 4.6 for code generation, all codes self reviewed

### DIFF
--- a/app/javascript/components/SocialAuthButton.tsx
+++ b/app/javascript/components/SocialAuthButton.tsx
@@ -1,9 +1,8 @@
+import { usePage } from "@inertiajs/react";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
-import { cast } from "ts-safe-cast";
 
 import { BrandName, Button, ButtonProps } from "$app/components/Button";
-import { useRunOnce } from "$app/components/useRunOnce";
 
 export const SocialAuthButton = ({
   href,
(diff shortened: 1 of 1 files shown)


app/javascript/components/SocialAuthButton.tsx
```diff
@@ -14,8 +13,7 @@ export const SocialAuthButton = ({
   provider: BrandName;
 } & ButtonProps) => {
   const formRef = React.useRef<HTMLFormElement>(null);
-  const [csrfToken, setCsrfToken] = React.useState("");
-  useRunOnce(() => setCsrfToken(cast(document.querySelector("meta[name=csrf-token]")?.getAttribute("content"))));
+  const { authenticity_token: csrfToken } = usePage<{ authenticity_token: string }>().props;
```
nyomanjyotisa, line 16: Changed this to `authenticity_token` as this token is already provided server-side via `inertia_share` in `InertiaRendering` and is always available

nyomanjyotisa:
@EmCousin _a Could you help to review this?

This is a quick fix. The root cause is the meta tags are not populated on login and register pages. We need to fix it, but kinda tricky to debug since it works on localhost

PR #3441, opened 2026-02-09
Inertia Migration: Checkout Page

### LINKED ISSUE #3042: Migrate Checkout Page to Inertia

sm17p:
# Closes #3042 

- #3042

# Description

This PR migrates checkout + internal cart API to Inertia

# Solution

### Backend
 
 - `CheckoutController` uses the Inertia layout, renders singluar Inertia page `Checkout/Show` and consumes props for the new Inertia page in a lazy fashion (e.g. `cart`, `checkout`,  and optional `recommended_products`)
 
 - The `cart` prop was moved out of `CheckoutPresenter#checkout_props`, and is now partially reloaded from `CheckoutController#show` route
 
 - `recommended_products` is served an an optional Inertia prop in `CheckoutController#show` instead of a separate API call (The logic is dependent on screen-width, and hence an optional prop)
 
 - `Api::Internal::CartsController` is removed, cart create/update logic lives in `CheckoutController#update` following recent migration patterns of grouping related route actions in one controller
 
- Mailers, installments, and workflow code that build checkout links (e.g. abandoned-cart emails) updated to the new checkout URL/behavior where needed.

### Frontend

- Server component was inlined into `pages/Checkout/Show` and the local cart state was replaced with Inertia's `useForm` object

- Cart state is synced with server by performing a `PATCH` request via Inertia form on `CheckoutController#update`

- `recommended_products` are fetched on initial load using `router.reload` on `CheckoutController#show` 
---

# Demos

## Logged Out

|Before|After|
|------|------|
| <video src="https://github.com/user-attachments/assets/e54a1063-3775-4278-833f-6756039ac8f0" /> | <video src="https://github.com/user-attachments/assets/a688f2e8-07d5-4021-bd42-b91d8c5f6b37" /> |

## Logged In

|Before|After|
|------|------|
| <video src="https://github.com/user-attachments/assets/d986c555-e27d-4c38-812a-5883aa97577f" /> | <video src="https://github.com/user-attachments/assets/129d90a9-2c82-47c8-93a2-60a88f7501fe" /> |

---

# Test Results

controller-specs

# Checklist

---

# AI Disclosure

Cursor Auto Mode was used to draft an intial migration plan, subsequent edits and debugging specs


app/controllers/api/internal/carts_controller.rb
```diff
@@ -1,70 +0,0 @@
-# frozen_string_literal: true
-
-class Api::Internal::CartsController < Api::Internal::BaseController
```
sm17p, line 3: Moved to `CheckoutController#update` since cart is a part of that page

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,70 @@
...
-  def index
-    @hide_layouts = true
-    @checkout_presenter = CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip)
+  before_action :process_cart_id_param, only: [:show]
+
+  def show
+    render inertia: "Checkout/Show", props: {
+      cart: -> { CartPresenter.new(logged_in_user:, ip: request.remote_ip, browser_guid: cookies[:_gumroad_guid]).cart_props },
+      checkout: -> { CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip).checkout_props(params:, browser_guid: cookies[:_gumroad_guid]) },
+      recommended_products: InertiaRails.optional { recommended_products },
```
sm17p, line 12: These props need to be lazy for optimal partial visits and form submission otherwise we'd end up re-computing checkout props per visit

app/controllers/checkout_controller.rb
```diff
@@ -39,4 +98,37 @@ def process_cart_id_param
...
+        ProductPresenter.card_for_web(
+          product: product_info.product,
+          request:,
+          recommended_by: product_info.recommended_by,
+          target: product_info.target,
+          recommender_model_name: product_info.recommender_model_name,
+          affiliate_id: product_info.affiliate_id,
+        )
+      end
+    end
```
sm17p, line 121: Duplicating this logic for now, we can optimize this later once we migrate other pages that use `RecommendedProductsController` to inertia

app/javascript/components/Checkout/CartNavigationButton.tsx
```diff
@@ -10,7 +10,7 @@ export const CartNavigationButton = ({ className }: { className?: string }) => {
   const cartItemsCount = useCartItemsCount();
 
   return cartItemsCount ? (
-    <NavigationButton className={className} color="filled" href={Routes.checkout_index_url({ host: appDomain })}>
+    <NavigationButton className={className} color="filled" href={Routes.checkout_url({ host: appDomain })}>
```
sm17p, line 13: Can't migrate this to Inertia for now as there are SSR  pages using this component

app/javascript/components/Checkout/CrossSellModal.tsx
sm17p, line 1: Modal files split after removing server component file

app/javascript/components/Checkout/index.tsx
```diff
@@ -202,21 +196,6 @@ export const Checkout = ({
...
-    asyncVoid(async () => {
-      try {
-        setRecommendedProducts?.(await getRecommendedProducts(productIds, isMobile ? 2 : 6));
-      } catch (e) {
-        assertResponseError(e);
-        showAlert(e.message, "error");
-      }
-    })();
-  }, [isMobile, productIds.join(",")]);
-
```
sm17p, line 219: This is now a partial reload on the `#show` action and moved to parent component. 

`Checkout`  cleanly reads on `recommended_products` instead of modifying parent state for this prop

app/javascript/components/CheckoutDashboard/CheckoutPreview.tsx
```diff
@@ -84,6 +84,7 @@ export const CheckoutPreview = ({
               items: [cartItem],
               discountCodes: [],
             }}
+            updateCart={() => {}}
             recommendedProducts={recommendedProduct ? [recommendedProduct] : []}
```
sm17p, line 88: When `Checkout` is used as preview we already provide it with recommended products. 

Moving the fetch to the parent and always passing `recommendedProducts` as a prop would unify preview and live behavior and make `Checkout` easier to reuse (which circles back to the original design decision from https://github.com/antiwork/gumroad/pull/3441/changes#r2780935704)

app/javascript/components/Wishlist/index.tsx
```diff
@@ -241,7 +241,7 @@ export const Wishlist = ({
       <WithTooltip tip={checkout_enabled ? null : "None of the products on this wishlist are available for purchase"}>
         <NavigationButton
           color="accent"
-          href={Routes.checkout_index_url({ params: { wishlist: id } })}
+          href={Routes.checkout_url({ params: { wishlist: id } })}
           disabled={!checkout_enabled}
         >
           <Icon name="cart3-fill" />
```
sm17p, line 247: @EmCousin These segments can never be migrated to Inertia `Link` or `NavigationButtonInertia` as they are under `UserCustomDomainConstraint` and `/checkout` is under `GumroadDomainConstraint`. 

Inertia uses history API which doesn't not support CORS re-write

> [!NOTE] 
>
> [MDN History/pushState#url Optional](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#url)
>
> The new URL must be of the same [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) as the current URL; otherwise, pushState() will throw an exception.

(Tried and tested) Even if we add re-direct route from user domain to gumroad domain, inertia always performs a full page load as `pushState` or `replaceState` throws an exception, so it's better to leave them like this and similarly other places which have checkout links under user domain

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -174,7 +176,7 @@ export const CheckoutPage = ({
     if (initialCart.items.length + newAddProducts.length > max_allowed_cart_products) {
       showAlert(`You cannot add more than ${max_allowed_cart_products} products to the cart.`, "error");
       initialCart.items = initialCart.items.slice(0, max_allowed_cart_products);
-      return initialCart;
+      return { cart: initialCart };
```
sm17p, line 180: (Self note: can this logic be backend only?)

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -538,42 +558,35 @@ export const CheckoutPage = ({
...
-      } catch (e) {
-        assertResponseError(e);
-        showAlert("Sorry, something went wrong. Please try again.", "error");
-      }
-    }),
-    100,
-  );
+  const debouncedSaveCartState = useDebouncedCallback(() => {
+    cartForm.patch(Routes.checkout_path(), { only: ["cart", "flash"], preserveScroll: true });
+  }, cart_save_debounce_ms);
```
sm17p, line 570: Using Inertia form for clean routing

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -538,42 +558,35 @@ export const CheckoutPage = ({
...
-      dispatch({ type: "update-products", products: getProducts(cart) });
+      dispatch({ type: "update-products", products: getProducts(cartForm.data.cart) });
     }
-  }, [cart]);
+  }, [cartForm.data.cart]);
   useOnChange(() => {
     if (state.email.trim() === "" || isValidEmail(state.email.trim())) {
-      setCart((prev) => ({ ...prev, email: state.email.trim() }));
+      // @ts-expect-error FormDataKeys recurses into Product.cross_sells; CartState is still correct at runtime
+      cartForm.setData("cart.email", state.email.trim());
```
sm17p, line 580: Added `@ts-expect-error` because of circular references which does not play well with inertia form

`Type of property 'cross_sells' circularly references itself in mapped type '{ [Key in keyof Product]: 0 extends 1 & Product[Key] ? never : Product[Key] extends FormDataConvertibleValue ? never : Product[Key] extends any[] ? `${Key}.${FormDataKeys<...> & string}` : Product[Key] extends Record<...> ? `${Key}.${FormDataKeys<...> & string}` : Exclude<...> extends any[] ? never : Exclude<...> ex...'.ts(2615)`

app/presenters/checkout_presenter.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CheckoutPresenter
+  CART_SAVE_DEBOUNCE_DURATION_IN_SECONDS = 0.1.seconds
```
sm17p, line 4: Debouncing requests messes with `wait_for_ajax` calls as they update `__activeRequests` counts at a later time. 

`wait_for_ajax` only cares about “no in-flight requests now”. It loops until `!window.__activeRequests` (and jQuery.active === 0). So it can see “0” while the debounce timer is still waiting, pass, and then the request fires later. So the test can proceed to the next assertion before the save request has even started or before the DB has been updated.

With the debounce value available in backend we can check for intermediate state at the right time intervals (`wait_until_true(sleep_interval: CheckoutPresenter::CART_SAVE_DEBOUNCE_DURATION_IN_SECONDS) { Cart.alive.where(email: "test@gumroad.com").exists? }`) instead of the default 1 second before moving onto the next assertion helping speed up specs a little

spec/controllers/checkout_controller_spec.rb
```diff
@@ -3,23 +3,52 @@
...
+                                             })
+      expect(inertia.props[:recommended_products]).to be_nil
+      meta_by_property = inertia.props[:_inertia_meta].filter_map { |t| [t[:property], t[:content]] if t[:property] }.to_h
+      expect(meta_by_property).to include(
+        "gr:google_analytics:enabled" => "true",
+        "gr:fb_pixel:enabled" => "true",
+        "gr:logged_in_user:id" => "",
+        "gr:page:type" => "",
+        "gr:facebook_sdk:enabled" => "true"
+      )
```
sm17p, line 51: Checking all the props over here along with analytics meta tags

spec/controllers/checkout_controller_spec.rb
```diff
@@ -131,5 +161,346 @@
...
+      let!(:purchase) { create(:purchase, purchaser:) }
+
+      before do
+        products.first.update!(user: purchase.link.user)
+        cart_product.update!(user: purchase.link.user)
+        index_model_records(Link)
+        sign_in purchaser
+      end
+
+      it "calls CheckoutService and returns product cards" do
```
sm17p, line 203: Spec copied from `recommended_products_controller_spec.rb`

spec/controllers/checkout_controller_spec.rb
```diff
@@ -131,5 +161,346 @@
...
+        expect(inertia.props.deep_symbolize_keys[:recommended_products]).to eq([product_cards.first])
+      end
+    end
+  end
+
+  describe "PATCH update" do
+    before do
+      request.headers["X-Inertia"] = "true"
+      request.headers["X-Inertia-Partial-Component"] = "Checkout/Show"
+      request.headers["X-Inertia-Partial-Data"] = "cart, flash"
```
sm17p, line 234: Making this a partial request to speed up specs a little and to keep the test inline with the actual user flow

spec/controllers/checkout_controller_spec.rb
```diff
@@ -3,23 +3,52 @@
 require "spec_helper"
 require "shared_examples/sellers_base_controller_concern"
 require "shared_examples/authorize_called"
+require "inertia_rails/rspec"
 
-describe CheckoutController do
-  render_views
```
sm17p, line 8: Removing `render_views` to speed up specs

spec/requests/purchases/product/offer_codes_spec.rb
```diff
@@ -187,7 +187,8 @@
       add_to_cart(product, offer_code:)
       offer_code.update!(amount_cents: 200)
       check_out(product, error: "The price just changed! Refresh the page for the updated price.")
-      visit checkout_index_path
+      wait_until_true(sleep_interval: CheckoutPresenter::CART_SAVE_DEBOUNCE_DURATION_IN_SECONDS) { Cart.alive.where(email: "test@gumroad.com").exists? }
```
sm17p, line 190: Accounting for `delay` in saving the cart state

spec/requests/purchases/product/upsell_spec.rb
sm17p, line 616: @EmCousin This spec passes but it's a bug, this case can never happen with the existing flow

### Cross-sell w/o clicking on "Remove" creates a purchase order for the two products

https://github.com/user-attachments/assets/f986dddc-c37e-4824-9e7f-18f26de330c7

### Cross-sell even after clicking on "Remove" creates a purchase order for the two products

https://github.com/user-attachments/assets/c43d51d8-3214-443c-a712-5428c0c893fc

sm17p (commented):
@EmCousin I've added my review comments, please have a look this is ready for review 🙏

sm17p:
Two related specs are failing, I'm looking into it

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -206,9 +209,15 @@ export const CheckoutPage = ({
...
     }
-    return initialCart;
+    return { cart: initialCart };
   });
   const { require_email_typo_acknowledgment } = useFeatureFlags();
+  useRunOnce(() => {
+    const url = new URL(window.location.href);
+    const searchParams = new URLSearchParams([...url.searchParams].filter(([key]) => key === "_gl"));
+    url.search = searchParams.toString();
+    router.replace({ url: url.toString() });
```
sm17p, line 219: There is a timing issue which caused [`spec/requests/products/third_party_analytics_spec.rb:157`](https://github.com/antiwork/gumroad/actions/runs/21829742710/job/62991847840?pr=3441) to fail 

The stripped URL from `payment.ts` is not getting picked up Inertia during the `router.reload` call on `L247`, so we are using `router.replace({ url: url.toString() });` instead to re-write the URL (the logic is duplicate of what's in payment.ts which can be moved back again once subscription manager is migrated to inertia)

sm17p:
The specs should pass now. Please have a look again 

Screenshot 2026-02-10 at 10 37 39 AM

app/javascript/components/Checkout/payment.ts
```diff
@@ -405,8 +405,11 @@ export function createReducer(initial: {
   const [state, dispatch] = reducer;
   useRunOnce(() => {
     const url = new URL(window.location.href);
+    if (url.pathname.startsWith(Routes.checkout_path())) return;
     const searchParams = new URLSearchParams([...url.searchParams].filter(([key]) => key === "_gl"));
     url.search = searchParams.toString();
+    // TODO (sm17p) Replace with Inertia's router.replace once subscription manager page is migrated to Inertia
+    // remove the checkout-path early return above so this runs on checkout too; then remove the duplicate useRunOnce url stripping logic in Checkout/Show.tsx.
```
sm17p, line 412: Adding a clean-up comment over here for Subscription Manager Page

EmCousin:
Please see https://github.com/antiwork/gumroad/pull/3349#issuecomment-3878052968

PR #3433, opened 2026-02-09
fix: keep search bar visible when search returns no results

buildingvibes:
## Summary
- Fixed the search bar disappearing when a search query returned zero products on the Products page
- The `Search` component was conditionally rendered only when `products.length > 0`, causing it to vanish when the server returned an empty result set
- Changed the condition to also show the search bar when there are memberships or an active search query

## Technical Changes
- `app/javascript/components/ProductsDashboardPage.tsx`: Changed Search visibility condition from `products.length > 0` to `products.length > 0 || memberships.length > 0 || query`

## Test plan
- [ ] Navigate to Products page with existing products
- [ ] Search for a term that returns no results
- [ ] Verify the search bar remains visible and functional
- [ ] Verify clearing the search shows all products again

Fixes #3262

> Automated fix by buildingvibes

PR #3442, opened 2026-02-09
Fix search bar disappearing when no products match query

### LINKED ISSUE #3262: Search bar disappears when no products match the search query

buildingvibes:
Fixes #3262

## Problem

On the Products page, when a user searches for products and the query returns zero results, the search bar disappears. This happens because:

1. The search bar was conditionally rendered based on `products.length > 0` — so when the Inertia reload returns zero products, the search bar unmounts
2. The `ProductsPage` component conditionally rendered both table components only when their entry arrays had items, causing them to unmount on empty results
3. When the table components unmount, their `useEffect` hooks that watch `props.query` stop running, preventing any subsequent searches
4. The empty-state placeholder replaces the entire products view when both arrays become empty, even during an active search

## Solution

- Track whether products existed on initial page load using lazy state initialization (`useState(() => ...)`) and use that to control search bar visibility, so the search bar persists across searches
- Always render the `ProductsPageProductsTable` and `ProductsPageMembershipsTable` components regardless of current entry count, keeping them mounted so their query-watching effects continue to fire
- In the table components, return an empty `<div>` instead of `null` when entries are empty to keep the component mounted
- Add a `!query` check to the empty-state placeholder condition so it doesn't show during an active search

/claim #3262

### DIFF
--- a/app/javascript/components/ProductsDashboardPage.tsx
+++ b/app/javascript/components/ProductsDashboardPage.tsx
@@ -36,6 +36,7 @@ export const ProductsDashboardPage = ({
 }: ProductsDashboardPageProps) => {
   const [enableArchiveTab, setEnableArchiveTab] = React.useState(archivedProductsCount > 0);
   const [query, setQuery] = React.useState("");
+  const [hasProducts] = React.useState(() => products.length > 0 || memberships.length > 0);
 
   return (
     <ProductsLayout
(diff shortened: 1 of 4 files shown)


buildingvibes:
Closing duplicate - fix already submitted in #3433.

PR #3446, opened 2026-02-09
Change PDF stamping flash message from error to warning

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad-private/issues/124

# Description

## Problem
When a customer downloads a PDF-stamped file for the first time, the "We are preparing the file for download" message displays with error style, which makes it look like something is broken

## Solution
Changed `flash[:alert]` to `flash[:warning]` to better reflect that this is an informational message, not an error.

---

# Before/After

## Before
image

## After
Screenshot 2026-02-09 at 15 30 00

---

# AI Disclosure


nyomanjyotisa:
@binary-koan _a Could you help to review this?

PR #3450, opened 2026-02-09
Fix overly specific Iffy suspend message

binary-koan:
# Description

## Problem

Iffy has a hardcoded 18+ message when suspending or banning users, but it doesn't just deal with that case any more. This data is passed to Helper, which makes it give creators incorrect information about the suspension.

## Solution

Changed it to the generic non-compliance message (from `lib/utilities/compliance.rb`) so that Helper realises it's not a specific reason.

---

# AI Disclosure

No AI used for this PR.


binary-koan:
Extremely simple, I’ll just go ahead and merge this

PR #3455, opened 2026-02-09
Cap fees on low-price purchases instead of blocking the sale

### LINKED ISSUE #3036: Currency price minimums don't account for fees
## Description

Sellers can price products at the currency minimum, but those minimums don't always cover Gumroad's fees. When fees exceed the product price, the purchase is blocked — buyers see a confusing "product is not correctly set up" error and Gumroad earns $0.

This also affects products where purchasing power parity discounts or currency conversion reduce the effective price below the fee threshold.

## Impact

- **50 purchases** failed in the last 3 months due to fees exceeding the product price.
- **16,603 successful purchases** last month had less than 20 cents of margin between price and fees – any small rate fluctuation could tip them into failure.
- Typical failure: a $0.50 effective price where fees total $0.55. The sale fails over 5 cents.
- Gumroad currently earns $0 on every one of these because the purchase is blocked entirely.
- Raising currency minimums was considered but would break **333K+ existing products**.

## Proposed fix

**Cap fees at checkout instead of blocking the sale.**

When calculated fees (Gumroad's cut + Stripe processing) equal or exceed the product price, cap fees so the seller nets at least 1 cent. Gumroad takes a reduced cut instead of earning nothing.

Example on a $0.50 transaction (currently everyone gets $0):
- Stripe processing: ~$0.31
- Gumroad: ~$0.18
- Creator: $0.01

This only applies when the price covers Stripe's processing cost. When the price is too low even for that, the purchase still fails – but with a clearer error message for both buyer and seller.

## Acceptance criteria

- [ ] When Gumroad fees + affiliate commission >= product price, fees are capped so the seller nets at least 1 cent
- [ ] When even zeroing Gumroad fees isn't enough (e.g. affiliate commission alone >= price), the purchase still fails
- [ ] Buyer error message on remaining failures is updated from "product is not correctly set up" to something about the price being too low to process
- [ ] Seller notification email gives clearer, actionable advice
- [ ] No change in behavior for purchases where fees are below the price
- [ ] Tests cover: fee capping, remaining failure case, no-change case

gianfrancopiana:
Fixes #3036

## Root Cause

When Gumroad's fees (platform cut + Stripe processing) exceed a product's price — common with currency minimums, PPP discounts, or currency conversion — the purchase is blocked with a confusing "product is not correctly set up" error. Gumroad earns $0 on every one of these failed transactions.

## Solution

Cap fees at checkout instead of blocking the sale. When fees + affiliate commission >= price, reduce Gumroad's cut so the seller nets at least 1 cent. The purchase only fails when even zeroing Gumroad's fees isn't enough (e.g., affiliate commission alone >= price).

## Why This Approach

- No currency minimum changes, avoids breaking 333K+ existing products
- Gumroad earns a reduced fee instead of $0
- Sellers still get at least 1 cent of revenue
- Guard clause prevents incorrect capping when affiliate credits are negative (fee share exceeds commission)

---

### DIFF
--- a/spec/models/purchase/purchase_process_spec.rb
+++ b/spec/models/purchase/purchase_process_spec.rb
@@ -540,8 +540,8 @@ describe "Purchase Process", :vcr do
       end
 
       context "when fees are capped during the full checkout flow" do
-        let(:low_price_product) { create(:product, price_cents: 50) }
-        let(:purchase) { create(:purchase, link: low_price_product, seller: low_price_product.user, save_card: false, ip_address:) }
+        let(:low_price_product) { create(:product, price_cents: 100).tap { |p| p.update_column(:price_cents, 50) } }
+        let(:purchase) { create(:purchase, link: low_price_product, seller: low_price_product.user, price_cents: 50, save_card: false, ip_address:) }
 
         before do
           purchase.skip_preparing_for_charge = true
--- a/spec/models/purchase_spec.rb
+++ b/spec/models/purchase_spec.rb
@@ -2622,7 +2622,9 @@ describe Purchase, :vcr do
 
       it "caps fees on a recurring subscription charge" do
         purchase.price_cents = 50
-        purchase.subscription = create(:subscription, link: product, user: create(:user))
+        subscription = create(:subscription, link: product, user: create(:user))
+        create(:purchase, link: product, subscription:, is_original_subscription_purchase: true)
+        purchase.subscription = subscription
         purchase.is_original_subscription_purchase = false
         allow(purchase).to receive(:determine_affiliate_balance_cents).and_return(0)
         purchase.send(:calculate_fees)


app/models/purchase.rb
```diff
@@ -3197,6 +3197,10 @@ def calculate_fees
 
       self.fee_cents = variable_fee_cents + fixed_fee_cents
       self.affiliate_credit_cents = determine_affiliate_balance_cents
+
+      if fee_cents + affiliate_credit_cents >= price_cents && affiliate_credit_cents >= 0
```
gianfrancopiana, line 3201: When affiliate credit is negative (fee share back to seller exceeds the commission), the sum `fee_cents + affiliate_credit_cents` can appear to exceed `price_cents` even though the seller isn't actually losing money. Skip capping in that case – the existing validation below will catch true negative-revenue scenarios.

ershad:
Will assign to @hchhabra _a for review when the PR is ready.

app/models/purchase.rb
```diff
@@ -3197,6 +3197,10 @@ def calculate_fees
...
+        self.fee_cents = [price_cents - affiliate_credit_cents - 1, 0].max
```
hchhabra, line 3202: `fee_cents` is inclusive of Stripe processing fee, so reducing it to 0 (or anything less than Stripe's 2.9%+30c fees) would mean we'll be covering the Stripe fees for such transactions? There's no potential refund or dispute liability in case of free sales, so these low amount transactions are different from free sales in that regard I think. May not be worth having potential dispute liability of $15 per dispute for these transactions, if we reduce the fee as well, imo. It seems there have only been 66 of these failures in the last 6 months with a total amount of $29.07. Maybe we can just improve the messaging here, and let creators update the product pricing after receiving the email notification or customer feedback @gianfrancopiana _a ? cc @dvassallo

dvassallo: Not sure I understand. These won't be free sales, no? Just that in the worst case scenario (rare), we'd be covering the Stripe free fully.

Issue #2173, opened 2025-11-30
Add e2e specs for Apple Pay and Google Pay


binary-koan:
We currently have no specs for checking out with Apple Pay and Google Pay. This makes it harder to catch bugs and verify fixes like https://github.com/antiwork/gumroad/pull/2172.

Apple Pay and Google Pay are only properly testable on computers with real accounts and cards set up, but we can potentially mock the Stripe `paymentRequest` API or the PaymentRequest web API it uses internally. We call it here: https://github.com/antiwork/gumroad/blob/e6463f1dfd2c81ebc32f68d1797e649afae46039/app/javascript/components/Checkout/PaymentForm.tsx#L1040

We won't be able to test a working checkout, but if we can test selecting the option, assert what form fields exist, and assert some kind of "not available" error when clicking Pay it would be much better than nothing.

iedayan:
Is this open to be worked on?

binary-koan:
This is blocked on the Inertia migration (see the last PR for context)

binary-koan:
Thanks for checking - although we closed the main issue the key pages for this are still not migrated. Updated the blocked by issues here, should be good to go once those two are merged

PR #3456, opened 2026-02-09
test(e2e): Add Apple Pay & Google Pay checkout specs via CDP mocking

aryanghai12:
Issue: #2173 

# Description

## Problem
We currently lack end-to-end test coverage for Apple Pay and Google Pay integration in the checkout flow. Testing these payment methods is notoriously difficult in CI/CD environments because they require valid merchant validation, specific browser capabilities (HTTPS, payment handlers), and often physical hardware, which headless Chrome/Selenium does not provide.

## Solution
I have implemented a CDP (Chrome DevTools Protocol) mocking strategy to simulate these environments without external dependencies.

- **CDP Injection:** Uses `Page.addScriptToEvaluateOnNewDocument` to inject a mock `Stripe` object and `ApplePaySession` before the page loads.
- **API Mocking:** Intercepts `stripe.paymentRequest()` and overrides `canMakePayment()` to return `true`, tricking the UI into rendering the payment buttons.
- **Behavior Simulation:** Mocks the `.show()` method to simulate the user opening and then cancelling the native payment sheet, allowing us to test the "fallback to credit card" UI flows.

---

# Test Results

- Apple Pay button visibility and brand icon rendering.
- Google Pay button visibility and brand icon rendering.
- Form toggling (ensures standard credit card fields hide/show correctly).
- Payment method switching logic.

---

# Checklist

---

# AI Disclosure

GitHub Copilot was used to assist in generating the CDP injection scripts and scaffolding the RSpec system tests. All logic was manually reviewed and verified for correctness.

### DIFF
--- a/app/javascript/components/Admin/BlockEmailDomainsForm.tsx
+++ b/app/javascript/components/Admin/BlockEmailDomainsForm.tsx
@@ -35,7 +35,6 @@ const Form = ({ action, header, buttonLabel }: Props) => {
   return (
     <form onSubmit={handleSubmit}>
       <FormSection
-        className="p-4 md:p-8"
         header={
           <>
             <p>{header}</p>
--- a/app/javascript/components/Admin/SalesReports/JobHistory.tsx
+++ b/app/javascript/components/Admin/SalesReports/JobHistory.tsx
@@ -63,9 +63,7 @@ const AdminSalesReportsJobHistory = ({ countries, sales_types, jobHistory, authe
         authenticityToken={authenticityToken}
         onSuccess={() => setShowNewSalesReportForm(false)}
         wrapper={(children) => (
-          <FormSection className="p-4 md:p-8" header="Generate sales report with custom date ranges">
-            {children}
-          </FormSection>
+          <FormSection header="Generate sales report with custom date ranges">{children}</FormSection>
         )}
       />
     ) : (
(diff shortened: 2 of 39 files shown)


aryanghai12:
@binary-koan Sir, the pr is ready for review.

thank you :)

spec/support/payment_request_helper.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module PaymentRequestHelper
+  def mock_payment_request_availability(apple_pay: false, google_pay: false)
```
binary-koan, line 4: I think we could just combine the test files and inline this tbh, it'll be long but not crazy for an e2e spec. We should also have at least a smoke test for when both are enabled

spec/requests/checkout/apple_pay_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+  before do
+    mock_payment_request_availability(apple_pay: true, google_pay: false)
+  end
+
+  after do
+    clear_payment_request_mocks
+  end
+
+  context "when Apple Pay is available" do
+    it "shows Apple Pay option with brand icon" do
```
binary-koan, line 17: I think we can combine a bunch of test cases here - spinning up each e2e test is heavy so (without sacrificing readability) we should aim to have fewer `it` blocks for greater performance

Maybe this can be simplified down to

* `it "allows selecting Apple Pay"` (also checks card fields are hidden and pay button is visible)
* `it "can switch between Apple Pay and credit card"`
* `it "returns to input state when payment is cancelled"`
* `it "requires email before allowing payment"` (should be expanded to check it's disabled before and enabled after filling email)

binary-koan (commented):
Thanks! Added a couple of initial comments, could you open a new PR with the changes?

PR #3470, opened 2026-02-09
Fix product edit clearing default discount code

gianfrancopiana:
## Problem

Editing a product clears its "Automatically apply discount code" setting.

### Steps to Reproduce

1. Enable "Automatically apply discount code" on a product and save
2. Edit the product again and save without changing anything
3. The discount code is gone

## Root Cause

The presenter sends `default_offer_code` (the full object) but not `default_offer_code_id` (the ID string) to the frontend. On save, the frontend spreads the product state into the request – the missing ID arrives as blank, and the backend wipes the discount.

## Fix

Include `default_offer_code_id` in the presenter, matching how the frontend already expects and sends it back.

---

### DIFF
--- a/spec/requests/products/edit/edit_spec.rb
+++ b/spec/requests/products/edit/edit_spec.rb
@@ -392,6 +392,21 @@ describe("Product Edit Scenario", type: :system, js: true) do
     expect(product.reload.suggested_price_cents).to be_nil
   end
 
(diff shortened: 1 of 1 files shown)


ershad (approved):
@gianfrancopiana _a changes lgtm, we can merge on green 👍🏼

Issue #3152, opened 2026-01-26
bug: IBAN and bank code validation issues for Angolan bank accounts


Kissabi:
The current validation logic for Angolan bank accounts (AngolaBankAccount) fails to support both standard local bank identification codes and the full IBAN format. This prevents users in Angola from successfully setting up their payout details.

# Technical Context
In Angola, the banking system (regulated by BNA - Banco Nacional de Angola) uses two primary formats:
Local Bank Code: A 4-digit numeric code (e.g., 0006 for BFA, 0040 for BAI).
IBAN: A 25-character string starting with AO06 followed by 21 digits.
NIB (Legacy): A 21-digit numeric string (the numeric part of the IBAN).

# The Problems
IBAN Validation Failure: The validate_account_number method often rejects valid 25-character Angolan IBANs. This may be due to whitespace handling or the specific way the Ibandit gem processes the AO prefix in older versions.

# To Reproduce
Go to Payout settings and select Angola.
Enter a valid 25-character IBAN starting with AO06 -> Fails validation.
Enter a valid 21-digit numeric NIB -> Fails validation.

# Proposed Solution

Refactor validate_account_number to sanitize input (remove whitespace) and support both the 25-character IBAN and the 21-digit legacy NIB, following the pattern used in the Omani bank account implementation (#491).

PR #3471, opened 2026-02-09
Fix IBAN validation for Angolla

Hanaffi:
Issue: https://github.com/antiwork/gumroad/issues/3152
Closes: https://github.com/antiwork/gumroad/issues/3152

# Description

The current validation logic for Angolan bank accounts (AngolaBankAccount) fails to support both standard local bank identification codes and the full IBAN format. This prevents users in Angola from successfully setting up their payout details.

## Problem

Ibandit gem validation (_**even on latest version**_) doesn't work properly with Angola, and it considers `AO06004000009439667010104` as invalid 
image

while it's valid
image

## Solution
Add regex validation for the 25-character IBANs for Angola

---

# Test Results

image

---

# Checklist

---

# AI Disclosure

Used Cursor for auto-completion

### DIFF
--- a/app/models/angola_bank_account.rb
+++ b/app/models/angola_bank_account.rb
@@ -6,9 +6,10 @@ class AngolaBankAccount < BankAccount
   BANK_CODE_FORMAT_REGEX = /^([0-9a-zA-Z]){8,11}$/
   private_constant :BANK_CODE_FORMAT_REGEX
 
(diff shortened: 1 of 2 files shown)


hchhabra:
Stripe only accepts 25-character IBANs for [Angola](https://docs.stripe.com/payouts?account-country=AO) so that is what we should allow I think, 21-character account numbers will be rejected by Stripe anyways.

PR #3365, opened 2026-02-05
Migration of checkout to inertia

yashranaway:
Issue: #3042
Fixes: #3042

Previous PR: #3081, #3160

# Description

## Problem

The checkout page currently uses React on Rails with ERB templates, which is inconsistent with our ongoing migration to Inertia.js. This creates maintenance overhead and prevents us from leveraging Inertia's benefits like simplified data passing and better SPA-like navigation.

## Solution

Migrate the public checkout page from React on Rails to Inertia while preserving all existing functionality:

- Move checkout logic from `server-components/CheckoutPage.tsx` to `components/Checkout/CheckoutPageContent.tsx`
- Create new Inertia page at `pages/Checkout/Index.tsx` with `disableLayout = true`
- Update controller to use `render inertia:` instead of ERB + react_component
- Delete the old ERB view, server component, SSR registration, and checkout pack
- Update tests to use Inertia test helpers with actual prop value assertions

---

# Before/After

## Before

https://github.com/user-attachments/assets/a12ae8b0-f4a1-4701-8dac-509acd142f6d

## Logged out

https://github.com/user-attachments/assets/284a68a8-7222-47ed-bf78-3a94906fb385

## Logged in

https://github.com/user-attachments/assets/35b838a1-86dd-4303-9ea7-0ac219624dda

API

https://github.com/user-attachments/assets/67a87979-3d18-4f1c-80e6-be6a15803ef7

---

# Test Results
image

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

Claude Code (Claude Opus 4.5) was used as a coding assistant for this migration - helping with code extraction, Inertia patterns, and test updates.

PR #3481, opened 2026-02-10
Internal/inertia checkout migration

EmCousin:
Issue: #3042

  # Description

  ## Problem

  PR #3441 (by @sm17p) and PR #3349 (by @stefanbinoj) both implemented the checkout Inertia migration with
  different approaches to handling URL parameter cleanup. The original implementation in #3441 used `useRunOnce`
  with `router.replace()` which could race with cart saves. Additionally, in older Inertia versions,
  `router.replace()` internally called `visit()` which made server requests and cancelled in-flight requests.

  ## Solution

  Combined the best of both implementations:

  1. **Moved URL cleanup to `onSuccess` callback** (from #3349) - Ensures cleanup happens after cart save
  completes, avoiding race conditions
  2. **Upgraded `@inertiajs/react` from 2.3.4 to 2.3.13** - In Inertia 2.x, `router.replace()` now calls
  `clientVisit()` instead of `visit()`, making it client-side only with no server request and no in-flight request
   cancellation
  3. **Added `urlCleanedRef` guard** - Ensures URL cleanup only runs once
  4. **Added `type` keyword to Receipt.tsx imports** - Better tree-shaking

  ---

  # Checklist

  ---

  # AI Disclosure

  This PR was developed with assistance from Claude Code (Claude Opus 4.5). AI was used to:
  - Analyze PRs #3441, #3349, and #3365 to identify the best approaches
  - Investigate the Inertia.js source code to understand `router.replace()` behavior differences between versions
  - Implement the combined solution and draft the PR description

### DIFF
--- a/app/javascript/pages/Checkout/Show.tsx
+++ b/app/javascript/pages/Checkout/Show.tsx
@@ -572,7 +572,7 @@ const CheckoutIndexPage = () => {
     const url = new URL(window.location.href);
     const searchParams = new URLSearchParams([...url.searchParams].filter(([key]) => key === "_gl"));
     url.search = searchParams.toString();
-    window.history.replaceState(window.history.state, "", url.toString());
+    router.replace({ url: url.toString(), preserveState: true, preserveScroll: true });
   });
   React.useEffect(() => {
     debouncedSaveCartState();


EmCousin:
@claude review

app/controllers/checkout_controller.rb
```diff
@@ -1,11 +1,70 @@
...
+    render inertia: "Checkout/Show", props: {
+      cart: -> { CartPresenter.new(logged_in_user:, ip: request.remote_ip, browser_guid: cookies[:_gumroad_guid]).cart_props },
+      checkout: -> { CheckoutPresenter.new(logged_in_user:, ip: request.remote_ip).checkout_props(params:, browser_guid: cookies[:_gumroad_guid]) },
+      recommended_products: InertiaRails.optional { recommended_products },
+    }
+  end
+
+  def update
+    if update_permitted_params[:items].length > Cart::MAX_ALLOWED_CART_PRODUCTS
+      return redirect_to checkout_path, alert: "You cannot add more than #{Cart::MAX_ALLOWED_CART_PRODUCTS} products to the cart.", status: :see_other
```
stefanbinoj: do we usually have `status: :see_other` for failure responses (responses with `alert` flash message)?

app/javascript/components/Checkout/Receipt.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 
 import { createAccount, CreateAccountPayload } from "$app/data/account";
-import { ErrorLineItemResult, LineItemResult, SuccessfulLineItemResult } from "$app/data/purchase";
+import { type ErrorLineItemResult, type LineItemResult, type SuccessfulLineItemResult } from "$app/data/purchase";
```
stefanbinoj:
```suggestion
import type { ErrorLineItemResult, LineItemResult, SuccessfulLineItemResult } from "$app/data/purchase";

```

app/javascript/components/Checkout/UpsellModal.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from "react";
+
+import { LineItemResult } from "$app/data/purchase";
+
+import { Button } from "$app/components/Button";
+import { CartState, CartItem, getDiscountedPrice, Upsell } from "$app/components/Checkout/cartState";
+import { computeOptionPrice, OptionRadioButton, Option } from "$app/components/Product/ConfigurationSelector";
+
+export type Result = { item: CartItem; result: LineItemResult };
```
stefanbinoj: I believe we already export this in `app/javascript/components/Checkout/cartState.ts` line 179

app/javascript/components/Checkout/index.tsx
stefanbinoj, line 56:
```suggestion
import {
  type CartState,
  convertToUSD,
  hasFreeTrial,
  getDiscountedPrice,
  type CartItem as CartItemProps,
  findCartItem,
} from "$app/components/Checkout/cartState";
```

app/controllers/checkout_controller.rb
EmCousin: We don't. We usually force the redirection to be a GET request upon successful redirections.

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
+import { router, useForm, usePage } from "@inertiajs/react";
 import { reverse } from "lodash-es";
 import * as React from "react";
-import { createCast, cast } from "ts-safe-cast";
+import { cast } from "ts-safe-cast";
 
 import { SurchargesResponse } from "$app/data/customer_surcharge";
```
stefanbinoj:
```suggestion
import { type SurchargesResponse } from "$app/data/customer_surcharge";
```

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
...
 import { startOrderCreation } from "$app/data/order";
-import { LineItemResult } from "$app/data/purchase";
 import { getPlugins, trackUserActionEvent, trackUserProductAction } from "$app/data/user_action_event";
 import { SavedCreditCard } from "$app/parsers/card";
```
stefanbinoj:
```suggestion
import { type SavedCreditCard } from "$app/parsers/card";
```

EmCousin:
Wrt Claude's review

**Issue 1-2, 4-5:** Valid observations, but acceptable tradeoffs for this PR.

**Issue 3:** Verified in `@inertiajs/core/types/router.d.ts:44` - `router.replace(params: ClientSideVisitOptions)` takes an object, not a string. Current code is correct.

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
...
 import { CardProduct, COMMISSION_DEPOSIT_PROPORTION, CustomFieldDescriptor } from "$app/parsers/product";
```
stefanbinoj:
```suggestion
import { type CardProduct, COMMISSION_DEPOSIT_PROPORTION, type CustomFieldDescriptor } from "$app/parsers/product";
```

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
...
-import { asyncVoid } from "$app/utils/promise";
 import { assertResponseError } from "$app/utils/request";
-import { register } from "$app/utils/serverComponentUtil";
 import { startTrackingForSeller, trackProductEvent } from "$app/utils/user_analytics";
 
-import { Button } from "$app/components/Button";
 import { Checkout } from "$app/components/Checkout";
 import {
   CartItem,
   CartState,
```
stefanbinoj:
```suggestion
  type CartState,
```

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
...
 import { calculateFirstInstallmentPaymentPriceCents } from "$app/utils/price";
-import { asyncVoid } from "$app/utils/promise";
 import { assertResponseError } from "$app/utils/request";
-import { register } from "$app/utils/serverComponentUtil";
 import { startTrackingForSeller, trackProductEvent } from "$app/utils/user_analytics";
 
-import { Button } from "$app/components/Button";
 import { Checkout } from "$app/components/Checkout";
 import {
   CartItem,
```
stefanbinoj:
```suggestion
  type CartItem,
```

app/javascript/pages/Checkout/Show.tsx
```diff
@@ -1,38 +1,34 @@
...
   CartState,
   convertToUSD,
   findCartItem,
   getDiscountedPrice,
-  Upsell,
   ProductToAdd,
   CrossSell,
-  saveCartState,
+  Result,
```
stefanbinoj:
```suggestion
  type ProductToAdd,
  type CrossSell,
  type Result,
```

app/javascript/pages/Checkout/Show.tsx
stefanbinoj:
```suggestion
  type Product,
```

stefanbinoj:
```suggestion
  type Gift,
```

app/controllers/checkout_controller.rb
sm17p: If the original request is idempotent then no but if it's non-idempotent then yes we need to do 303 as per HTTP spec. In our case `update` is a PATCH request so this is correct.

Ref - https://en.wikipedia.org/wiki/Post/Redirect/Get

stefanbinoj:
Just added minor comments, feel free to to add them if they seems valid!

sm17p:
> Upgraded @inertiajs/react from 2.3.4 to 2.3.13 - In Inertia 2.x, router.replace() now calls
clientVisit() instead of visit(), making it client-side only with no server request and no in-flight request
cancellation

I think the AI made some mistake I just checked locally, and `router.replace()` is using `clientVisit()` for 2.3.4. It's a good update though since we can now refer the latest docs effortlessly 👍

Screenshot 2026-02-10 at 8 47 23 PM

app/javascript/components/Checkout/payment.ts
```diff
@@ -405,8 +405,11 @@ export function createReducer(initial: {
   const [state, dispatch] = reducer;
   useRunOnce(() => {
     const url = new URL(window.location.href);
+    if (url.pathname.startsWith(Routes.checkout_path())) return;
     const searchParams = new URLSearchParams([...url.searchParams].filter(([key]) => key === "_gl"));
     url.search = searchParams.toString();
+    // TODO (sm17p) Replace with Inertia's router.replace once subscription manager page is migrated to Inertia
+    // remove the checkout-path early return above so this runs on checkout too; then remove the duplicate useRunOnce url stripping logic in Checkout/Show.tsx.
     window.history.replaceState(window.history.state, "", url.toString());
   });
```
stefanbinoj, line 414: Not sure if we still need this as we moved this to `onSuccess` handler of saveCartState since it was causing a race condition

sm17p:
> Moved URL cleanup to onSuccess callback (from https://github.com/antiwork/gumroad/pull/3349) - Ensures cleanup happens after cart save completes, avoiding race conditions

This is a bit confusing as cart save is not dependent on any of the URL query params from checkout page. It only requires values from the underlying form object which is built via page props so there's no race condition over here w.r.t query params. Which race condition is this? 

I've also tested `useRunOnce` with `router.replace` and it correctly strips off the URL which wasn't happening prior to [cd2881c](https://github.com/antiwork/gumroad/pull/3441/commits/cd2881cb5a75454964d301a4beec5c9f9fa7d99f) commit. The [specs](https://github.com/antiwork/gumroad/actions/runs/21867694314/) on #3441 run also seem to be passing with one yet to be processesed.

https://github.com/user-attachments/assets/92637e84-d325-4983-a1b0-d13cc803e615


------

About https://github.com/antiwork/gumroad/pull/3441#discussion_r2786048572, my guess is that the render cycle in which Inertia updates are happening have stale reference to the non-stripped URL somewhere and hence the url value from below gets over-written by the stale one

https://github.com/antiwork/gumroad/blob/98919859d26d0d68ff9005a9b177316faded9916/app/javascript/components/Checkout/payment.ts#L406-L411

app/controllers/checkout_controller.rb
EmCousin: It's a redirection upon failure here, so nothing has changed. We can leave a 302 response in those cases.

EmCousin:
@sm17p I've reverted back to `useRunOnce` with `router.replace` (5ce451d5). The stale URL
  reference from payment.ts makes sense since that code skips checkout paths already (if
  (`url.pathname.startsWith(Routes.checkout_path())) return;`), but the Inertia render cycle could still be
  restoring the original URL from its internal state. I'm now using router.replace with preserveState:
  true to avoid both that and needlessly re-rendering the component (which was crashing CI runs).

EmCousin:
@claude review again.

EmCousin:
Thanks for your work @stefanbinoj @sm17p!

sm17p:
Thanks for the review @EmCousin @stefanbinoj

Issue #3408, opened 2026-02-08
fix: Floating-point precision loss in price/amount to cents conversion


iamnikhilranjan:
## Description

The current implementation converts decimal prices to cents using `.to_f * 100`, which causes floating-point precision errors that can result in 1-cent discrepancies in price calculations.

## Affected Files

| File | Line | Code |
|------|------|------|
| `app/controllers/links_controller.rb` | 126 | `params[:price] = (params[:price].to_f * 100).to_i` |
| `app/controllers/links_controller.rb` | 490 | `new_price: (params.require(:amount).to_f * 100).to_i` |

## Tests I Ran

I ran the following Ruby script to demonstrate the bug:

```ruby
test_prices = ["19.99", "4.56", "29.99", "9.99"]

test_prices.each do |price|
  buggy = (price.to_f * 100).to_i
  expected = (price.to_r * 100).to_i
  status = buggy == expected ? "✓ OK" : "✗ BUG"
  puts "Price: $#{price} | Result: #{buggy} cents | Expected: #{expected} cents | #{status}"
end
```

**Output:**
```
Price: $19.99 | Result: 1998 cents | Expected: 1999 cents | ✗ BUG
Price: $4.56  | Result: 455 cents  | Expected: 456 cents  | ✗ BUG
Price: $29.99 | Result: 2999 cents | Expected: 2999 cents | ✓ OK
Price: $9.99  | Result: 999 cents  | Expected: 999 cents  | ✓ OK
````

## Root Cause

Floating-point numbers cannot represent all decimal values exactly in binary. When `19.99` is stored as a float, it becomes `19.989999999999998`. Multiplying by 100 gives `1998.999...`, and `.to_i` truncates to `1998`.

## Proposed Solution

Use `BigDecimal` for safe currency calculations:

```ruby
# Current (buggy)
(params[:price].to_f * 100).to_i

# Fixed
(BigDecimal(params[:price].to_s) * 100).to_i

# Or using Rails shorthand
(params[:price].to_d * 100).to_i
```

**Proof the fix works:**
```ruby
require 'bigdecimal'
(BigDecimal("19.99") * 100).to_i  # => 1999 ✓
(BigDecimal("4.56") * 100).to_i   # => 456 ✓
```

PR #3483, opened 2026-02-10
fix: use BigDecimal for price/amount to cents conversion

sangokp:
Issue: #3408

# Description

## Problem

Floating-point precision loss when converting dollar amounts to cents using `to_f`. For example:
- `(19.99.to_f * 100).to_i` returns `1998` (incorrect)
- `(9.99.to_f * 100).to_i` returns `998` (incorrect)

This occurs because floating-point numbers cannot precisely represent certain decimal values like 19.99 (it becomes 19.989999999999998 internally).

## Solution

Changed `to_f` to `to_d` (BigDecimal shorthand) in two locations in `links_controller.rb`:
- Line 126: price param conversion for checkout redirect
- Line 497: amount param conversion in `send_sample_price_change_email`

With BigDecimal:
- `(19.99.to_d * 100).to_i` returns `1999` (correct)
- `(9.99.to_d * 100).to_i` returns `999` (correct)

---

# Before/After

Not applicable - this is a backend calculation fix with no UI changes.

---

# Test Results

Added tests in `spec/controllers/links_controller_spec.rb` that verify:
1. `19.99` converts to `1999` cents (not `1998`)
2. `9.99` converts to `999` cents (not `998`)

These tests fail when the fix is reverted (using `to_f` instead of `to_d`).

---

# Checklist

---

# AI Disclosure

AI (Claude) was used for code analysis and drafting this PR description. The fix approach (BigDecimal via `to_d`) was determined by understanding Ruby's floating-point behavior and existing Rails conventions.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -123,7 +123,7 @@ class LinksController < ApplicationController
       BasePrice::Recurrence::ALLOWED_RECURRENCES.each do |r|
         params[:recurrence] ||= r if params[r] == "true"
       end
-      params[:price] = (params[:price].to_f * 100).to_i if params[:price].present?
+      params[:price] = (params[:price].to_d * 100).to_i if params[:price].present?
       cart_item = @product.cart_item(params)
 
       unless (@product.customizable_price || cart_item[:option]&.[](:is_pwyw)) &&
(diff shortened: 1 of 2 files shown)


sangokp:
## Self-Review

### Code Changes

**Line 126 change (`to_f` → `to_d`):**
This handles the `price` param conversion when `wanted=true`. The price comes in as a string like `"19.99"` and needs to be converted to cents. Using `to_d` (BigDecimal shorthand from ActiveSupport) ensures exact decimal arithmetic.

**Line 497 change (`to_f` → `to_d`):**
This handles the `amount` param in `send_sample_price_change_email`. Same issue - the amount comes as a string and needs cents conversion.

### Why `to_d` works

Ruby's `to_d` method (from ActiveSupport) creates a BigDecimal from the string representation, preserving the exact decimal value:

```ruby
"19.99".to_f  # => 19.989999999999998 (floating-point)
"19.99".to_d  # => 0.1999e2 (BigDecimal, exact)

("19.99".to_f * 100).to_i  # => 1998 (wrong)
("19.99".to_d * 100).to_i  # => 1999 (correct)
```

### Test Coverage

Added two tests that:
1. Verify `19.99` converts to `1999` cents
2. Verify `9.99` converts to `999` cents

Both tests fail when reverting to `to_f`, confirming they test the actual fix.

PR #3503, opened 2026-02-11
Add signed SSL cert for *.gumroad.dev

binary-koan:
# Description

## Problem

Connecting the mobile app to a local Gumroad instance is a lot of effort because it's hard to make device simulators trust mkcert self-signed certificates.

## Solution

Added a proper trusted certificate for *.gumroad.dev which nginx will use by default if it's still valid (falling back to mkcert if it expires).

This simplifies mobile app setup and means we can just point it to gumroad.dev by default with no additional steps necessary.

---

# AI Disclosure

Claude Opus 4.5 used for working out the approach and for bash script editing.

### DIFF
--- a/bin/generate_ssl_certificates
+++ b/bin/generate_ssl_certificates
@@ -2,12 +2,6 @@
 
 set -euo pipefail
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
-CERTS_DIR="$PROJECT_DIR/docker/local-nginx/certs"
-
-mkdir -p "$CERTS_DIR"
-
 # Check if mkcert is installed
 if ! command -v mkcert &> /dev/null; then
   echo "mkcert is not installed. Please install it first:"
@@ -31,7 +25,7 @@ fi
 # Install local CA if not already installed
 mkcert -install
 
-cd "$CERTS_DIR"
+cd docker/local-nginx/certs
 
 echo "Generating certificates for helperai.dev..."
 mkcert helperai.dev "*.helperai.dev"


binary-koan:
@ershad _a Could you review? I think we can keep mkcert as a fallback (we need it for helper.dev anyway) but lmk if you disagree

bin/generate_ssl_certificates
```diff
@@ -2,43 +2,80 @@
...
+
+# Check if a signed wildcard certificate is available and still valid
+use_signed=false
+if [[ -f "$SIGNED_CERT" && -f "$SIGNED_CA_BUNDLE" && -f "$SIGNED_KEY" ]]; then
+  if openssl x509 -in "$SIGNED_CERT" -noout -checkend 0 2>/dev/null; then
+    expiry=$(openssl x509 -in "$SIGNED_CERT" -noout -enddate 2>/dev/null | cut -d= -f2)
+    echo "Using signed wildcard certificate (expires: $expiry)"
+    use_signed=true
+  else
+    echo "Signed certificate has expired, falling back to mkcert."
```
ershad: We can completely remove mkcert to keep it simple. When the cert expires, we can manually update it in the repo.

bin/generate_ssl_certificates
```diff
@@ -2,43 +2,80 @@
...
+  fi
 fi
 
-# Install local CA if not already installed
-mkcert -install
+if $use_signed; then
+  # Combine cert + CA bundle into a fullchain for nginx
+  cat "$SIGNED_CERT" "$SIGNED_CA_BUNDLE" > "$CERTS_DIR/gumroad_dev.crt"
+  cp "$SIGNED_KEY" "$CERTS_DIR/gumroad_dev.key"
```
ershad: We can directly store the combined cert and private key in `docker/local-nginx/certs` and can avoid these steps? Ideally, we can remove the the whole `bin/generate_ssl_certificates` script.

binary-koan: The script is still needed for `helperai.dev` certificates to run Helper alongside Gumroad locally, so I've just kept that part and removed references from all docs except the Helper setup one. 

If this works well for Gumroad we could consider buying a cert for Helper as well to remove the script entirely.

PR #3508, opened 2026-02-11
fix: persist offer code discount across installment charges

### LINKED ISSUE #1410: Offer code not applied to non-first installments when max usage limit is reached

joshua-deng:
## Summary
- Copies the cached `purchase_offer_code_discount` from the original purchase to subsequent installment/membership charges in `Subscription#build_purchase`, so the discount is preserved even when the offer code is deleted, maxed out, or modified
- Reorders `Purchase#original_offer_code` to check cached discount data before the deleted-code check, with safe navigation for deleted offer codes
- Adds subsequent installment payments to `does_not_count_towards_max_purchases` so product inventory limits don't block existing installment plans

Fixes #1410

## Test plan
- [x] Added spec: offer code discount persists when offer code is deleted
- [x] Added spec: offer code discount persists when offer code reaches max purchases
- [x] Added spec: original discount amount preserved even if offer code amount changes later

---

> [!NOTE]
> This change was made with the assistance of AI (Claude Opus 4.6).

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -1999,7 +1999,7 @@ class Purchase < ApplicationRecord
   end
 
   def does_not_count_towards_max_purchases
-    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase
+    is_recurring_subscription_charge || is_additional_contribution || is_preorder_charge? || is_gift_receiver_purchase || is_updated_original_subscription_purchase || is_commission_completion_purchase || (is_installment_payment && !is_original_subscription_purchase)
   end
 
   # Public: Determine if this purchase is a test purchase by the links owner.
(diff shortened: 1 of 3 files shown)


joshua-deng:
### Self-review

**Root cause:** \`Subscription#build_purchase\` only copied the live \`offer_code\` association to subsequent charges. When the offer code was deleted, expired, or maxed out between installments, \`set_price_and_rate\` couldn't build a \`purchase_offer_code_discount\` from the now-invalid live record, silently dropping the discount.

**Fix approach (per @EmCousin's suggestion):** Uses the existing \`purchase_offer_code_discount\` cached on the original purchase rather than adding new columns. Three changes:

1. **\`subscription.rb\` \`build_purchase\`**: Copies the cached \`purchase_offer_code_discount\` from the original purchase. Since \`has_cached_offer_code?\` is now true on the new purchase, \`set_price_and_rate\` skips building from the live offer code.

2. **\`purchase.rb\` \`original_offer_code\`**: Checks cached data first before the deleted-code early return. Added \`&.code\` safe navigation since the underlying offer code may be soft-deleted.

3. **\`purchase.rb\` \`does_not_count_towards_max_purchases\`**: Subsequent installment payments shouldn't count toward product inventory limits (same pattern as \`is_recurring_subscription_charge\`).

EmCousin:
@joshua-deng Please follow the contributing guidelines.

PR #3525, opened 2026-02-11
Validate and sanitize US business tax ID (EIN) before sending to Stripe

gianfrancopiana:
Fixes antiwork/gumroad-private#126

## Problem

US business users can store invalid EIN values (e.g. a letter prefix followed by too many digits) which then fail at the Stripe API with "US tax IDs must have 9 digits." The error surfaces as a generic Stripe error on every save attempt, with no way for the user to understand or fix the issue.

## Approach

Added two layers of defense:

1. **Backend sanitization and validation** – Strip non-digit characters from the business tax ID before storing (following the existing pattern used for bank account numbers). After saving, validate that US business tax IDs have exactly 9 digits and return a clear error message before the Stripe call is made.

2. **Frontend input constraints** – Added `minLength`/`maxLength` to the US business tax ID input field (9–10 characters, allowing the `XX-XXXXXXX` dash format) so the browser provides immediate feedback on invalid lengths.

## Before/After

Before – Stripe rejects the invalid tax ID, but the UI shows a generic error:

After – validation catches the issue before Stripe, showing a clear message:

---

### DIFF
--- a/app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
@@ -129,7 +129,12 @@ const AccountDetailsSection = ({
 
   const getBusinessTaxIdConfig = (): TaxIdConfig => {
     const configs: Record<string, { label: string; placeholder: string; minLength?: number; maxLength?: number }> = {
-      US: { label: "Business Tax ID (EIN, or SSN for sole proprietors)", placeholder: "12-3456789", minLength: 9, maxLength: 10 },
+      US: {
+        label: "Business Tax ID (EIN, or SSN for sole proprietors)",
+        placeholder: "12-3456789",
+        minLength: 9,
+        maxLength: 10,
+      },
       CA: { label: "Business Number (BN)", placeholder: "123456789" },
       AU: { label: "Australian Business Number (ABN)", placeholder: "12 123 456 789" },
       GB: { label: "Company Number (CRN)", placeholder: "12345678" },


app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
```diff
@@ -128,8 +128,13 @@ const AccountDetailsSection = ({
   };
 
   const getBusinessTaxIdConfig = (): TaxIdConfig => {
-    const configs: Record<string, { label: string; placeholder: string }> = {
-      US: { label: "Business Tax ID (EIN, or SSN for sole proprietors)", placeholder: "12-3456789" },
+    const configs: Record<string, { label: string; placeholder: string; minLength?: number; maxLength?: number }> = {
+      US: {
+        label: "Business Tax ID (EIN, or SSN for sole proprietors)",
+        placeholder: "12-3456789",
```
hchhabra, line 134:
```suggestion
        placeholder: "123456789",
```

Since that's the format that we (/Stripe) eventually want?

hchhabra (approved):
Looks good to me, @gianfrancopiana _a !

app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
gianfrancopiana: I think keeping the dash is better here – `XX-XXXXXXX` is the standard IRS format users recognize, and it's consistent with how we handle other countries (e.g., AU's placeholder has spaces). The backend already strips non-digits so either format works on input, but the placeholder's job is helping users identify what to enter.

PR #3526, opened 2026-02-11
Validate kana fields and romaji business name for Japanese Stripe accounts

gianfrancopiana:
Issue antiwork/gumroad-private#126

## Problem

Japanese creators get cryptic Stripe errors when submitting payment settings with invalid characters in kana fields (e.g., kanji or full-width parentheses instead of katakana) or Japanese characters in the legal business name field. Stripe rejects these values but the error messages are unhelpful, leaving users stuck.

## Approach

Added client-side and server-side validation for Japanese account fields:

- **Kana name fields** (first/last name, business name) must contain only katakana, spaces, dashes, and dots.
- **Kana address fields** (street address, building number) additionally allow latin characters and digits.
- **Legal business name** must be in romaji (latin characters) for Japanese accounts — Stripe requires this but previously we had no guardrail.

Validation runs on both the frontend (immediate feedback before form submission) and the backend model (`UserComplianceInfo`), so invalid data is caught regardless of how it's submitted. Placeholder text for kana inputs was updated to `カタカナ` to make the expected format obvious, and the business name label now says "Legal business name (Romaji)" for JP accounts.

Fixed a pre-existing bug where client-side validation error messages persisted across form resubmissions. If a user fixed a field and clicked save again, the old error message would still be shown (even if the new failure was a different field with no message). Now `clientErrorMessage` is cleared at the start of each validation pass.

## Before/After

Before:
CleanShot 2026-02-11 at 16 33 22@2x

After:
CleanShot 2026-02-11 at 16 35 46@2x

CleanShot 2026-02-11 at 16 50 54
---


app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -41,6 +41,14 @@ import { WithTooltip } from "$app/components/WithTooltip";
 
 import logo from "$assets/images/logo-g.svg";
 
+const KANA_NAME_REGEX = /^[\u30A0-\u30FF\u31F0-\u31FF\uFF65-\uFF9F\s\-.]*$/u;
+const KANA_ADDRESS_REGEX = /^[\u30A0-\u30FF\u31F0-\u31FF\uFF65-\uFF9F\p{Script=Latin}\d\s\-.]*$/u;
+
+const KANA_NAME_ERROR = "may only contain katakana characters, spaces, dashes, and dots.";
+const KANA_ADDRESS_ERROR = "may only contain katakana, latin characters, digits, spaces, dashes, and dots.";
+
+const HAS_JAPANESE_CHARS = /[\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\u4E00-\u9FFF\uFF65-\uFF9F]/u;
```
hchhabra, line 50:
```suggestion
const JAPANESE_CHARS_REGEX = /[\u3000-\u303F\u3040-\u309F\u30A0-\u30FF\u4E00-\u9FFF\uFF65-\uFF9F]/u;
```

app/models/user_compliance_info.rb
```diff
@@ -182,6 +187,45 @@ def handle_compliance_info_request
       UserComplianceInfoRequest.handle_new_user_compliance_info(self)
     end
 
+    def kana_fields_format
+      return if country_code != Compliance::Countries::JPN.alpha2
```
hchhabra, line 191:
```suggestion
      return unless legal_entity_country_code == Compliance::Countries::JPN.alpha2
```
?

app/models/user_compliance_info.rb
```diff
@@ -182,6 +187,45 @@ def handle_compliance_info_request
...
+        value = send(field)
+        next if value.blank?
+        next if value.match?(regex)
+        errors.add(:base, "#{label} may only contain #{allowed_description}")
+      end
+    end
+
+    def business_name_romaji_format
+      return if country_code != Compliance::Countries::JPN.alpha2
+      return if !is_business?
```
hchhabra:
```suggestion
      return unless business_country_code == Compliance::Countries::JPN.alpha2
      return unless is_business?
```

hchhabra (approved):
@gianfrancopiana _a, couple of minor comments below, looks good to me otherwise!

app/javascript/pages/Settings/Payments/Show.tsx
gianfrancopiana: `HAS_JAPANESE_CHARS` is intentionally named differently because it serves a different purpose than the `_REGEX` constants. 

`HAS_JAPANESE_CHARS` has no anchors – it's a presence check that tests whether a string contains any Japanese characters. At the call site it reads as: `HAS_JAPANESE_CHARS.test(value)` → "has Japanese chars?" 

Renaming it to `JAPANESE_CHARS_REGEX` would lose that semantic distinction.

app/models/user_compliance_info.rb
gianfrancopiana: `unless`: our CONTRIBUTING.md says "Avoid `unless`", so I'll keep `return if !=`.

`legal_entity_country_code` returns `business_country_code` for business users – which means individual kana fields would be validated based on the business's country rather than the individual's. I'll keep `country_code` here since it's the individual's country that determines whether kana fields are shown in the UI. (In practice both are always Japan, but I think `country_code` is more correct semantically.)

app/models/user_compliance_info.rb
gianfrancopiana: Good catch! Switched to `business_country_code` since we're validating the business name. Kept `return if !=` per our CONTRIBUTING.md.

PR #2839, opened 2026-01-12
feat: implement backup payment method for refund funding

pegasus1134:
Issue: #1594

  Design: [Figma](https://www.figma.com/design/bQE0b8eQTM2eov0Kzjxk3A/Ability-to-load-funds-to-Gumroad-balance-to-cover-refunds-when-balance-is-too-low%C2%A0?node-id=3-4663&t=4nezLwGjmKo8rxPy-0)

  ## Problem

  Sellers can't issue refunds when their Gumroad balance is insufficient (e.g., after payouts). This blocks timely refunds and creates support friction.

  ## Solution

  Allow sellers to add a backup credit card in Settings > Payments. When a refund would overdraw the seller's balance, automatically charge the shortfall to this card and proceed with the refund.

  ### How It Works

  1. Seller sees banner on Customers page prompting backup method setup
  2. Clicks through to Settings > Payments, adds card via Stripe Elements
  3. When refund issued with insufficient balance → backup card charged for difference
  4. Seller receives email confirmation
  5. Refund processes normally for customer
  6. Seller can remove card anytime (banner reappears)

  ### Implementation Notes

  - **Reuses existing `Credit` model** for audit trail - no new tables needed. Stripe charge ID stored in `json_data`.
  - **Idempotency key** on Stripe charges prevents duplicate charges during retries
  - **Inline processing** - refund flow is already user-blocking, async adds unnecessary complexity
  - **Audit trail** via Credit record + user comment ("Balance topped up by $X to cover refund shortfall")

  PR #2425 takes a different approach with a new `BalanceTopUp` model and state machine. This implementation achieves the same result with less infrastructure - the existing Credit model provides the audit trail needed, and a state machine isn't necessary for a synchronous Stripe charge that simply succeeds or fails.

  ---

  ## Demo

  **Dark Mode:**

https://github.com/user-attachments/assets/09a6a617-78e0-447a-9d72-7d63461f436e

  **Light Mode:**

https://github.com/user-attachments/assets/68eb0c2d-aaa6-4152-8af7-34fa7f630c97

  ---

  ## Test Results

gumroadtestss1

gumroadtestss2

  30 examples, 0 failures

  ---

  ## Checklist

  - [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
  - [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
  - [x] I have performed a self-review and left review comments on my PR
[...]

PR #3531, opened 2026-02-12
feat: add backup payment method for refund funding (#1594)

mxrcabrera:
Issue: #1594

# Description

## Problem

Sellers sometimes have insufficient balance to cover refunds. Currently, when a seller tries to refund a purchase but doesn't have enough balance, the refund fails and they have to manually add funds first. This creates friction and delays for both sellers and buyers.

## Solution

Add the ability for sellers to configure a backup credit card in Settings > Payments that will be automatically charged when their balance is insufficient to process a refund.

**Key improvements over existing PRs (#2425, #2839, #3399):**

- **DB-level idempotency**: Unique partial index on credits.refund_funding_purchase_id (requested in #3399 review by pegasus1134)
- **Charge reversal**: reverse_charge! automatically reverses the backup card charge if refund fails post-charge
- **Race condition fix**: seller.reload + balance re-check after charging prevents double-charges when concurrent refunds read the same balance
- **3D Secure detection**: Cards requiring interactive authentication are rejected with a clear message since off-session payments can't handle it
- **Plain text email**: Both HTML and .text.erb templates for charge confirmation
- **Correct email queue**: Uses default queue instead of critical for non-critical confirmation emails
- **Figma-compliant UI**: Includes "Name on card" field as per approved design

---

### Before

https://github.com/user-attachments/assets/b52d4fd5-1f4a-411d-b854-2603665b3890

### After

https://github.com/user-attachments/assets/42d31229-61ba-4647-96b4-fade13a2d7bd

---

# Test Results

Service spec:

image

Controller spec (refund_funding):

image

Full PaymentsController spec suite confirming no regressions (116 examples, 0 failures)

image

---

# Checklist

---

# AI Disclosure
Model: Claude Opus 4
IDE: VS Code with Claude Code
Usage: Implementation assistance, code review, test writing

### DIFF
--- a/app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
@@ -1,0 +1,229 @@
+import { CardElement, Elements, useElements, useStripe } from "@stripe/react-stripe-js";
+import { StripeElementStyleVariant } from "@stripe/stripe-js";
+import * as React from "react";
+import { cast } from "ts-safe-cast";
+
+import { asyncVoid } from "$app/utils/promise";
+import { assertResponseError, request } from "$app/utils/request";
+import { getStripeInstance } from "$app/utils/stripe_loader";
+import { getCssVariable } from "$app/utils/styles";
+
+import { Button } from "$app/components/Button";
+import { useFont } from "$app/components/DesignSettings";
+import { Icon } from "$app/components/Icons";
+import { showAlert } from "$app/components/server-components/Alert";
+
+export type RefundFundingCard = {
+  visual: string;
+  card_type: string;
+  expiry_month: number;
+  expiry_year: number;
+} | null;
+
+const RefundPaymentMethodForm = ({
+  initialCard,
+  isFormDisabled,
+}: {
+  initialCard: RefundFundingCard;
+  isFormDisabled: boolean;
+}) => {
+  const stripe = useStripe();
+  const elements = useElements();
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+  const [savedCard, setSavedCard] = React.useState(initialCard);
+  const [isEditing, setIsEditing] = React.useState(!initialCard);
+  const [cardholderName, setCardholderName] = React.useState("");
+  const [baseStripeStyle, setBaseStripeStyle] = React.useState<null | StripeElementStyleVariant>(null);
+
+  const handleSubmit = asyncVoid(async () => {
+    if (!stripe || !elements) return;
+
+    const cardElement = elements.getElement(CardElement);
+    if (!cardElement) return;
+
+    setIsSubmitting(true);
+
+    try {
+      const { paymentMethod, error } = await stripe.createPaymentMethod({
+        type: "card",
+        card: cardElement,
+        ...(cardholderName ? { billing_details: { name: cardholderName } } : {}),
+      });
+
+      if (error) {
+        showAlert(error.message ?? "Card verification failed", "error");
+        setIsSubmitting(false);
+        return;
+      }
+
+      const response = await request({
+        method: "POST",
+        url: Routes.save_refund_funding_card_settings_payments_path(),
+        accept: "json",
+        data: {
+          stripe_payment_method_id: paymentMethod.id,
+          card_data_handling_mode: "stripejs.0",
+        },
+      });
+
+      const result = cast<{ success: boolean; error?: string; credit_card?: NonNullable<RefundFundingCard> }>(
+        await response.json(),
+      );
+
+      if (result.success && result.credit_card) {
+        showAlert("Backup payment method saved successfully!", "success");
+        setSavedCard(result.credit_card);
+        setCardholderName("");
+        setIsEditing(false);
+      } else {
+        showAlert(result.error ?? "Failed to save card", "error");
+      }
+    } catch (e) {
+      assertResponseError(e);
+      showAlert("An error occurred. Please try again.", "error");
+    }
+
+    setIsSubmitting(false);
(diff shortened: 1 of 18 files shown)


app/modules/purchase/refundable.rb
```diff
@@ -45,8 +45,27 @@ def refund_and_save!(refunding_user_id, amount_cents: nil, is_for_fraud: false)
 
         amount_cents_to_refund = amount_cents.presence || amount_refundable_cents
         if amount_cents_to_refund > seller.unpaid_balance_cents && charged_using_gumroad_merchant_account?
-          errors.add :base, "Your balance is insufficient to process this refund."
-          return false
+          if seller.refund_funding_credit_card.present?
```
mxrcabrera, line 48: Race condition handling: After the backup card is charged, we seller.reload to get the freshest balance. If a concurrent refund processed between our charge and this check, the balance might still be insufficient. In that case, we reverse the Stripe charge immediately via reverse_charge! and ask the creator to retry.

app/modules/purchase/refundable.rb
```diff
@@ -94,19 +113,27 @@ def refund_and_save!(refunding_user_id, amount_cents: nil, is_for_fraud: false)
         if charge_refund.flow_of_funds.nil? && StripeChargeProcessor.charge_processor_id != charge_processor_id
           charge_refund.flow_of_funds = FlowOfFunds.build_simple_flow_of_funds(Currency::USD, -(gross_amount_cents.presence || gross_amount_refundable_cents))
         end
-        refund_purchase!(charge_refund.flow_of_funds, refunding_user_id, charge_refund.refund, is_for_fraud)
+        refund_purchase!(charge_refund.flow_of_funds, refunding_user_id, charge_refund.refund, is_for_fraud).tap do |success|
+          if success && @refund_funding_result&.success?
```
mxrcabrera, line 117: Email sent after successful refund only: The confirmation email is sent inside the .tap block of refund_purchase!, which means it's only delivered if the refund itself succeeded. If the refund fails, reverse_refund_funding_charge_if_needed! is called and no email is sent.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+  def reverse_charge!(payment_intent_id)
+    Stripe::Refund.create({ payment_intent: payment_intent_id })
+  rescue Stripe::StripeError => e
+    Bugsnag.notify(e)
+    Rails.logger.error("RefundFundingChargeService reversal failed: #{e.message}")
+  end
+
+  private
+    def charge_credit_card
+      payment_intent = Stripe::PaymentIntent.create(
```
mxrcabrera, line 51: Off-session payment with idempotency: Uses off_session: true, confirm: true since the charge happens automatically during refund processing. The deterministic idempotency_key (refund_funding_{user_id}_{purchase_id}) prevents duplicate charges. Complemented by a unique partial index at the DB level.

app/services/refund_funding_charge_service.rb
```diff
@@ -0,0 +1,97 @@
...
+        },
+        { idempotency_key: idempotency_key }
+      )
+
+      if payment_intent.status == "succeeded"
+        {
+          payment_intent_id: payment_intent.id,
+          charge_id: payment_intent.latest_charge
+        }
+      elsif payment_intent.status == "requires_action"
```
mxrcabrera, line 75: 3D Secure rejection: If the payment intent returns requires_action, it means the card requires interactive authentication. Since this is an off-session charge, we can't present the challenge. We return a descriptive error asking the creator to use a card that supports automatic off-session payments.

app/models/credit.rb
```diff
@@ -51,6 +55,35 @@ def self.create_for_credit!(user:, amount_cents:, crediting_user:)
     credit
   end
 
+  def self.create_for_refund_funding!(user:, amount_cents:, purchase:, credit_card:, processor_transaction_id:, processor_payment_intent_id:)
```
mxrcabrera, line 58: Factory method pattern: create_for_refund_funding! follows the same pattern as create_for_credit! and create_for_dispute_won! - creates a Credit, associates it with the relevant entity, creates a BalanceTransaction, and links to the resulting Balance.

app/javascript/components/Settings/PaymentsPage/RefundPaymentMethodSection.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const handleSubmit = asyncVoid(async () => {
+    if (!stripe || !elements) return;
+
+    const cardElement = elements.getElement(CardElement);
+    if (!cardElement) return;
+
+    setIsSubmitting(true);
+
+    try {
```
mxrcabrera, line 46: Ghost input pattern for Stripe Elements styling: Renders a temporary <input> element to read computed styles (fontFamily, color) from the DOM, then passes them to Stripe's CardElement. This ensures the Stripe iframe matches the app's theme (including dark mode) without hardcoding colors.

db/migrate/20261119011937_add_unique_index_to_credits_refund_funding_purchase_id.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexToCreditsRefundFundingPurchaseId < ActiveRecord::Migration[7.1]
+  def change
+    add_index :credits, :refund_funding_purchase_id,
```
mxrcabrera, line 5: DB-level idempotency: This unique partial index (WHERE refund_funding_purchase_id IS NOT NULL) ensures that only one Credit can exist per purchase for refund funding. Second layer of protection against duplicate charges.

mxrcabrera:
@EmCousin could you please review this PR?

PR #3534, opened 2026-02-12
Migrate Communities Page to Inertia

yashranaway:
Fixes: #3045

Previous PR : #3244

# Description

## Problem

The Communities page currently uses React on Rails with React Router for client-side routing. This is inconsistent with the rest of the app which has been migrating to Inertia.js. The current implementation also has the chat message APIs under `/api/internal/` which doesn't follow the pattern we want for Inertia pages.

## Solution

Migrated the Communities page to Inertia.js:

- Updated `CommunitiesController` with proper `index` and `show` actions that render Inertia responses
- Moved chat message API controllers from `Api::Internal::Communities::*` to `Communities::*` namespace
- Created new Inertia page at `pages/Communities/Index.tsx`
- Moved React components from `server-components/CommunitiesPage/` to `components/Communities/`
- Updated frontend to use Inertia's `usePage()` instead of React Router's `useLoaderData()`
- Replaced React Router `Link` and `navigate()` with Inertia equivalents
- Updated route references from `internal_*` paths to new paths
- Cleaned up old files (react-router wrapper, pack file, erb view)

Key routing change: API routes now come BEFORE the show route to prevent `/communities/:community_id/chat_messages` from matching the show route pattern.

---

# Before/After

https://github.com/user-attachments/assets/8ea34c60-87c4-4970-9927-a5ad6f6e2050

https://github.com/user-attachments/assets/43488580-ca8d-4ed6-b57a-58d3b056304e

## Infinite Scroll (fixed!)

https://github.com/user-attachments/assets/227d61fa-c9b0-4210-a3cf-0a76341a441c

## Message Actions 

https://github.com/user-attachments/assets/ea75b64a-7cc4-4552-917d-3be0d662f5c8

---

# Test Results

Screenshot 2026-01-30 at 3 21 15 PM

---

# Checklist

---

# AI Disclosure

Claude Opus 4.5 was used to assist with the migration.

### DIFF
--- a/app/javascript/components/Communities/useCommunities.ts
+++ b/app/javascript/components/Communities/useCommunities.ts
@@ -127,11 +127,9 @@ export const useCommunities = () => {
     [],
   );
 
-  // Initialize community chat from Inertia props when navigating to a community
   React.useEffect(() => {
     if (initialSelectedCommunityId && initial_messages) {
       setCommunityChats((prev) => {
-        // Only initialize if not already loaded (avoid overwriting pagination state on Inertia revisits)
         if (prev[initialSelectedCommunityId]?.messages.length) return prev;
         return {
           ...prev,
--- a/app/javascript/pages/Communities/Index.tsx
+++ b/app/javascript/pages/Communities/Index.tsx
@@ -172,7 +172,6 @@ function CommunitiesIndex() {
     [selectedCommunity, debouncedMarkAsRead],
   );
 
-  // Scroll to a message after it appears in the DOM
   React.useEffect(() => {
     if (!selectedCommunityChat || !scrollToMessage) return;
     const exists = selectedCommunityChat.messages.findIndex((message) => message.id === scrollToMessage.id) !== -1;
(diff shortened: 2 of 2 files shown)


yashranaway:
> Tested locally - Reverse infinite scroll does not work well, if keeps loading previous messages when scrolling at the top infinitely, it should stop after loading previous messages once, and the user should be scrolling again, like the current behaviour
 When posting messages in quick succession, the form gets suck in the disabled state and I cannot send messages anymoire


@EmCousin _a Fixed both! `<InfiniteScroll>` was the root cause for the scroll issue it wraps content in its own div which breaks the CSS that pins messages to the bottom. Dropped it and went back to the scroll listener + JSON fetch approach from `main`. Works as expected now: scroll up, loads a page, stops. The form getting stuck on quick sends was already fixed in the prior push with `onFinish` which fires on success, error, and cancel so the input always unlocks. I believe this is the right approach for the chat use case, but open to suggestions if you have any! Added an updated video in the PR under the infinite scrolling section.

EmCousin:
Thanks @yashranaway, but the requirement was to expose the chat messages as inertia props, not a JSON endpoint.

PR #3536, opened 2026-02-12
Remove full-page skeleton flash on dashboard navigation

gianfrancopiana:
## Problem

When navigating between dashboard sections (e.g., Products → Settings → Analytics) or switching tabs within a section, the entire `<main>` content is hidden and replaced with a generic skeleton loader. This causes page titles, tabs, and other static UI to flash – even though they don't depend on async data and could remain visible during the transition.

## Approach

Stop replacing the entire content area with a skeleton. Inertia's default behavior already keeps the current page visible until the new page is ready, so the skeleton swap was fighting the framework rather than working with it.

- Removed `useRouteLoading` and `LoadingSkeleton` from both the main layout and the admin layout. The current page now stays visible during navigation.
- Enabled Inertia v2's built-in progress bar (`progress: { delay: 100, color: '#FF90E8' }`) to provide feedback on slower navigations. This shows a thin pink bar at the top of the viewport only for loads that take >100ms.
- Removed the unused `@inertiajs/progress` v1 package (Inertia v2 has progress built in).
- Extracted an `assignLayout` helper in `inertia.js` to deduplicate the layout resolution logic.

Pages that use `useRouteLoading` for their own granular loading states (Affiliates, UTM Links, Payouts) are unaffected, those patterns are correct.

## Before/After

Before:

https://github.com/user-attachments/assets/e55d21ee-b362-4bc5-a9aa-e18895a9476f

After:

https://github.com/user-attachments/assets/75aa547e-be7a-4a31-90f6-7e74757a1def

---


EmCousin:
Tested on both Chrome and Firefox BTW

PR #3537, opened 2026-02-12
Fix sub-pixel border misalignment between sidebar and page header

gianfrancopiana:
## Problem

On dashboard pages with action buttons or selects in the page header (Collaborators, Settings, Analytics, etc.), the horizontal border between the sidebar and the main content area is misaligned by ~0.4px. The sidebar's NavLink border and the PageHeader's bottom border don't connect into a continuous line at the junction point.

## Approach

The design system uses a line-height of `1.4`, which at 16px gives `22.4px` — not a whole number. That 0.4px fraction adds up through padding and borders, making buttons 48.4px tall, nav links 56.4px, etc. The result: borders on the sidebar and the main content end up at slightly different positions.

The fix changes line-height from `1.4` to `1.375` (Tailwind's `leading-snug`) in both the SCSS design tokens (`$line-heights`) and the React component Tailwind classes (Button, Input, Checkbox, Radio, Fieldset). At 16px this gives exactly `22px`, producing clean integer element heights (buttons 48px, nav links 56px) and pixel-perfect border alignment.

The visual difference between 22.4px and 22px line-height is imperceptible (1.8% change). The existing `DEFAULT_FORM_ELEMENT_HEIGHT_IN_PX = 48` constant was already hardcoded to the integer value, confirming the design intent was always 48px.

The PageHeader title row gets a `min-h-8` (32px) to prevent Safari from collapsing the `<h1>` to 31px on pages without action buttons (a font-metrics rounding difference with `line-clamp-2`).

## Before/After

Before:

After:

---

### DIFF
--- a/app/javascript/components/Button.tsx
+++ b/app/javascript/components/Button.tsx
@@ -32,7 +32,7 @@ export const buttonVariants = cva(
         destructive: "",
       },
       size: {
-        default: "px-4 py-3 text-base leading-[1.4]",
+        default: "px-4 py-3 text-base leading-snug",
         sm: "p-2 text-sm leading-[1.3]",
       },
       color: {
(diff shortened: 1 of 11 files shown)


EmCousin (approved):
Nice! Thanks @gianfrancopiana!

PR #3550, opened 2026-02-12
Search input not auto-focusing when popover opens

Hanaffi:
Issue: https://github.com/antiwork/gumroad/issues/3373
Closes: https://github.com/antiwork/gumroad/issues/3373

# Description

## Problem
Popover by default auto-focuses the first input element inside it once opened. But if `forceMount` is set to true, it doesn't. 
When using forceMount on a Radix UI Popover, the component remains in the DOM even when closed, which often conflicts with Radix's automatic focus management. 

I think we should prevent forceMount, but without it, **_many_** specs would fail because a lot of specs expect the content to be present.

## Solution
Focus the input on through an effect.

---

# Before/After

### Before
[Before.webm](https://github.com/user-attachments/assets/59ee302a-cb16-439f-aa6a-704f1f470fac)

### After
[After.webm](https://github.com/user-attachments/assets/5ef6e365-7588-4729-89c0-a2e822392cae)

---

# Test Results
followers test
products test

---

# Checklist

---

# AI Disclosure

Cursor for auto-completion

### DIFF
--- a/spec/requests/followers/followers_spec.rb
+++ b/spec/requests/followers/followers_spec.rb
@@ -54,6 +54,12 @@ describe("Followers", js: true, type: :system) do
       expect(page).to_not have_button "Load more"
     end
 
+    it "autofocuses the search input when the popover is opened" do
+      select_disclosure "Toggle Search" do
+        expect_focused find_field("Search followers")
+      end
+    end
+
     it "supports search functionality" do
       expect(page).to_not have_selector(:table_row, text: "test@example.com")
       select_disclosure "Toggle Search" do


spec/requests/followers/followers_spec.rb
```diff
@@ -54,6 +54,12 @@
       expect(page).to_not have_button "Load more"
     end
 
+    it "autofocuses the search input when the popover is opened" do
+      select_disclosure "Toggle Search" do
+        expect_focused find_field("Search followers")
+      end
+    end
```
Hanaffi, line 61: We could add a Jest test for the Search component’s autofocus behavior instead, it would be faster and more targeted. The E2E test verifies the real user experience in a browser, but a unit test would give quicker feedback during development. Up to the team whether we prefer one or both.

Hanaffi:
@nyomanjyotisa Can you please review this?

spec/requests/followers/followers_spec.rb
slavingia: we use Jest in this project?

PR #3551, opened 2026-02-13
Add new seed data for the Expo app

binary-koan:
# Description

This adds the necessary seed data and config to run the Expo app alongside Gumroad locally:

- Added a default `MOBILE_TOKEN` env var in development, matching the app's default config
- Added an `OauthApplication` for the app in development, matching the app's default config
- Added seeds for a few users and purchases that the app can use for e2e testing (separate from the current ones to avoid needing to reset the db - this way it's easier to leave them alone and use the existing seeds for developing the web app)

---

# AI Disclosure

Claude Opus 4.5 to generate an initial draft of the seed files.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -34,6 +34,7 @@ ANYCABLE_SECRET="anycable-local-secret"
 ANYCABLE_WEBSOCKET_URL="wss://cable.gumroad.dev:8081/cable"
 
 MOBILE_TOKEN=development-mobile-token
+MOBILE_API_OAUTH_APPLICATION_UID=MOBILE_DEVELOPMENT_CLIENT_kx3sRpN217qPwqk9s
 
 GOOGLE_CLIENT_ID=dummy_key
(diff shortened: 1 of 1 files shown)


binary-koan:
@gianfrancopiana _a Mind reviewing this?

.env.development
```diff
@@ -33,6 +33,8 @@ STRONGBOX_GENERAL_PASSWORD="1234"
 ANYCABLE_SECRET="anycable-local-secret"
 ANYCABLE_WEBSOCKET_URL="wss://cable.gumroad.dev:8081/cable"
 
+MOBILE_TOKEN=development-mobile-token
+
```
gianfrancopiana, line 38: Should we also set `MOBILE_API_OAUTH_APPLICATION_UID` here? Without it, `User#invalidate_active_sessions!` won't revoke Expo app tokens in development.

PR #3558, opened 2026-02-13
Allow suspended for TOS users to access Tax Center and Customer page

nyomanjyotisa:
Resolves https://github.com/antiwork/gumroad-private/issues/125

# Description

- Add `skip_before_action :check_suspended` to `TaxCenterController` and `CustomersController` so that TOS-suspended users can view their tax documents, download tax forms, and access their sales/customers page.
- Refactor the suspension bypass from a centralized guard in `check_suspended` to per-controller `skip_before_action` declarations for `BalanceController`, `Settings::MainController`, `Payouts::ExportablesController`, `Payouts::ExportsController`, `TaxCenterController`, and `CustomersController`.

Note: Users suspended for fraud are not able to access these pages since they cannot log in. Ref https://github.com/antiwork/gumroad-old/pull/9043#discussion_r159403016

---

# Before/After

## Before

https://github.com/user-attachments/assets/34cbfe94-f5b5-40c1-ab66-92523c4ec017

## After

https://github.com/user-attachments/assets/e439ca36-c05f-4e15-a762-db12e2338249

---

# Test Results

## Before fix
image

## After fix
image

---

# AI Disclosure

Claude Opus 4.6 for code generation. All code self reviewed.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -259,7 +259,6 @@ class ApplicationController < ActionController::Base
     end
 
     def check_suspended
-      return nil if request.path.in?([balance_path, settings_main_path]) || params[:controller].in?(%w[payouts tax_center])
       return head(:ok) if !request.format.html? && !request.format.json?
 
       respond_to do |format|
--- a/app/controllers/balance_controller.rb
+++ b/app/controllers/balance_controller.rb
@@ -2,6 +2,7 @@
 
 class BalanceController < Sellers::BaseController
   layout "inertia", only: [:index]
+  skip_before_action :check_suspended
 
   def index
     authorize :balance
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -3,6 +3,7 @@
 class CustomersController < Sellers::BaseController
   include CurrencyHelper
 
+  skip_before_action :check_suspended
   before_action :authorize
   before_action :set_on_page_type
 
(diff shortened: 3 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -259,7 +259,7 @@ def invalidate_session_if_necessary
     end
 
     def check_suspended
-      return nil if request.path.in?([balance_path, settings_main_path]) || params[:controller] == "payouts"
+      return nil if request.path.in?([balance_path, settings_main_path]) || params[:controller].in?(%w[payouts tax_center])
```
EmCousin: I would advise against guard clauses here and recommend using a `skip_before_action :check_suspended` in `TaxCenterController` instead. It is more consistent with what already exists in `DashboardController`, `LibraryController`, `ConsumptionAnalyticsController`, and `MediaLocationsController`, and reduces tight coupling in `ApplicationController`.

Also, `params[:controller] == "payouts"` might be dead code, there is no top-level `PayoutsController` as far as I know - they are all nested under namespaces like `admin` or `api/v2` so this can never be true. Can you check this can be removed?

Ideally, we should do the same thing for `BalanceController` and `Settings::MainController`, but that's not in the scope of this issue. It would be worth adding a TODO comment though. Last, we use Pundit for access rights management so ultimately we should migrate that code to a Policy, but again, out-of-scope here.

nyomanjyotisa: Actually `params[:controller] == "payouts"` is not dead code, it correctly allows access for suspended users to the payouts page and other actions on the payouts page through `Payouts::ExportablesController` and `Payouts::ExportsController`

But otherwise, removed the `return nil if request.path.in?([balance_path, settings_main_path]) || params[:controller] == "payouts"` guard and added `skip_before_action :check_suspended` to each controller instead

Also, as per https://github.com/antiwork/gumroad-private/issues/125#issuecomment-3901214875, added the customers page to the allow list

PR #3233, opened 2026-01-29
 migrate communities page to Inertia

dluffy56:
## Issue: #3045
Previous PR: #3095

# Description
Migrated Communities from internal API controllers and server-side React to Inertia controllers and React components. Removed obsolete API endpoints and server components, updated navigation, and refactored the frontend to use Inertia state and actions.

## Problem
The previous Communities implementation relied on internal APIs and server-side React, leading to duplicated logic and higher maintenance cost. The legacy endpoints and components were no longer needed.

## Solution
- Deleted internal API controllers for communities, chat messages, last-read messages, and notification settings
- Removed server-side React components and loader logic 
- Added Inertia controllers for community actions 
- Refactored React components to use Inertia state and navigation
- Removed all API calls to the deleted endpoints
---
### Bug                                                                                                                                                    
Inertia's `InfiniteScroll` component has a scroll preservation bug: `restoreScrollPosition` (in `@inertiajs/core`) fires via `requestAnimationFrame` and calls `scrollTo` with a stale negative offset, undoing any correct scroll adjustment when older messages are prepended.

### Approach
A `MutationObserver` adjusts `scrollTop` immediately on DOM mutation, and a `scrollTo`  intercept on the chat container blocks Inertia's bogus negative-value resets. This lets  InfiniteScroll work correctly for reverse-chronological chat without needing a `reverse` prop.

# Before/After
## Before 

[Screencast from 2026-02-06 21-19-05.webm](https://github.com/user-attachments/assets/5e7c2123-7d7b-4560-bcef-f014e1b6389b)

## After

[Screencast from 2026-02-06 19-49-29.webm](https://github.com/user-attachments/assets/f8e27f75-3a33-46e7-b78f-5c15be5b7eb0)

### Live Chat

[Screencast from 2026-02-06 19-58-02.webm](https://github.com/user-attachments/assets/29dcac2c-83b0-4752-99ac-ed396dd88e89)

### Infinity-Scroll Test (fixed!)

[Screencast from 2026-02-06 19-52-17.webm](https://github.com/user-attachments/assets/751c5796-568d-423b-aeea-817eba9af78a)
 

---

# Test Results 
- spec/controllers/communities_controller_spec.rb
- spec/controllers/communities/chat_messages_controller_spec.rb
- spec/controllers/communities/last_read_chat_messages_controller_spec.rb
- spec/controllers/communities/notification_settings_controller_spec.rb
- spec/channels/community_channel_spec.rb
- spec/channels/user_channel_spec.rb
[...]

PR #3559, opened 2026-02-13
Fix: Preserve chat scroll after message actions

dluffy56:
ref: #3045, #3233 

# Description
  Community chat messages reset when performing actions (delete, send, update) after scrolling up through paginated history

## Problem
Mutations (send/delete/update message) used only: [] which Inertia treats as a full visit, `[].length === 0` then `isPartial()` returns false. The 303 redirect response replaces all props, including messages, wiping out InfiniteScroll's accumulated pages and resetting to a single page.

## Solution
- Changed `only: []` to `except: ["messages"]` on all three mutation handlers so Inertia treats them as partial visits and preserves the existing messages prop. 
- Also moved `scrollTo` in `sendMessage` to fire immediately instead of waiting for the server round trip, and 
- Added `showProgress: false` to avoid the progress bar flash on chat actions.
-  Added `preserveUrl: true` to `markAsRead` to prevent the browser URL from changing on background visits, which could cause a `RecordNotFound` error on refresh.
---

# Before/After
## Before
[Screencast from 2026-02-13 13-33-28.webm](https://github.com/user-attachments/assets/30a09ced-50f8-4739-8e10-84318b58f6d8)

## After 

[Screencast from 2026-02-13 11-34-27.webm](https://github.com/user-attachments/assets/85bf2a45-bd42-460e-9652-e82a809e744c)

### Notification Settings
[Screencast from 2026-02-13 15-38-17.webm](https://github.com/user-attachments/assets/066d0e75-228e-42c2-9376-83b97f5bf0b0)

---

# Test Results
### Tested all related specs

Screenshot from 2026-02-13 09-39-50

---

# Checklist

---

# AI Disclosure

Used Claude Opus 4.6 to review and refine the changes. All modifications were implemented and tested manually.

### DIFF
--- a/app/javascript/pages/Communities/Index.tsx
+++ b/app/javascript/pages/Communities/Index.tsx
@@ -205,7 +205,7 @@ function CommunitiesIndex() {
       if (searchParams.has("notifications")) {
         searchParams.delete("notifications");
         const newUrl = `${window.location.pathname}${searchParams.size ? `?${searchParams.toString()}` : ""}`;
-        window.history.replaceState({}, "", newUrl);
+        router.replace({ url: newUrl, preserveState: true });
         setShowNotificationsSettings(true);
       }
     }


app/javascript/pages/Communities/Index.tsx
```diff
@@ -320,14 +321,16 @@ function CommunitiesIndex() {
...
       },
     }));
+    
+    scrollTo({ target: "bottom" });
+    setShowScrollToBottomButton(false);
 
     messageForm.post(Routes.community_chat_messages_path(selectedCommunity.id), {
       preserveScroll: true,
-      only: [],
+      except: ["messages"],
```
dluffy56, line 332: - `only: []` was treated as a full visit by Inertia. After the 303 redirect, the server returned a fresh messages prop, which replaced everything `InfiniteScroll` had accumulated. This caused the chat to reset to page 1 after any mutation.
- Switching to `except: ["messages"]` makes it a partial visit. Inertia skips updating the messages prop, preserving the paginated history already loaded by `InfiniteScroll`

app/javascript/pages/Communities/Index.tsx
```diff
@@ -320,14 +321,16 @@ function CommunitiesIndex() {
...
+      showProgress: false,
```
dluffy56, line 333: Noticed that after the merge of #3536, the progress bar was appearing for `non-GET` requests as well. Added `showProgress: false` for the relevant mutations to prevent the progress indicator from flashing during chat actions

app/javascript/pages/Communities/Index.tsx
```diff
@@ -218,6 +218,7 @@ function CommunitiesIndex() {
           { message_id: messageId },
           {
             preserveScroll: true,
+            preserveUrl: true,
```
dluffy56, line 223: Added `preserveUrl: true` to `markAsRead` since it’s a background side-effect and should not affect browser history or the current URL. This ensures the user stays on the same page while the request runs

dluffy56:
@EmCousin could you please take a look at this when you get a chance? Thanks! 
Noticed this behavior too late, Sorry about that!  
~~Seems like the issue is also with notification settings too, working on this!~~ fixed!

EmCousin:
@dluffy56 some tests are failing, please check

dluffy56:
@EmCousin  Only communities_spec.rb - notifications modal test is related. The other failures are unrelated  flaky tests.            
                                                                                                                                                                            
 The issue was `router.reload` with `except: ["messages"]` creates a partial Inertia visit. Partial visits merge props, giving `selectedCommunity` a new object reference that re-triggers the `useEffect`. Since `?notifications` was never stripped from the URL, this created an infinite loop of Inertia requests  causing  to time out.

 and  the original `only: []` had the same underlying bug - never stripping `?notifications`, but was flaky rather than consistently failing due to different timing with full visits

replaced `router.reload` with `window.history.replaceState` to actually strip `?notifications` from the URL

  Tested the spec and ui locally multiple times -  works fine.

app/javascript/pages/Communities/Index.tsx
```diff
@@ -203,7 +203,9 @@ function CommunitiesIndex() {
     if (selectedCommunity) {
       const searchParams = new URLSearchParams(window.location.search);
       if (searchParams.has("notifications")) {
-        router.reload({ only: [], replace: true });
+        searchParams.delete("notifications");
+        const newUrl = `${window.location.pathname}${searchParams.size ? `?${searchParams.toString()}` : ""}`;
+        window.history.replaceState({}, "", newUrl);
```
EmCousin: @dluffy56 Would this work?
```suggestion
        router.replace({ url: newUrl, preserveState: true })
```

dluffy56: @EmCousin my bad! this works!!

dluffy56:
@EmCousin all tests pass!

EmCousin:
Thanks @dluffy56 !

PR #3580, opened 2026-02-13
ix: Prevent NotFoundError in Chrome during cover upload

sayandeepdas02:
Chrome was throwing a `NotFoundError: Failed to execute 'removeChild' on 'Node'`
when uploading a cover file. The issue occurred because the file input
was synchronously unmounted when `isUploading` became true, causing a race
condition during the file selection lifecycle.

This patch keeps the file input mounted and toggles its visibility
using CSS instead of conditionally unmounting it. This prevents the
DOM removal race condition while preserving existing behavior.

Verified in Chrome and Edge. No regressions observed.

### DIFF
--- a/app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
+++ b/app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
@@ -154,89 +154,88 @@ const CoverUploader = ({
   };
 
   return isSelecting ? (
-    isUploading ? (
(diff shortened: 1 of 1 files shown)


nyomanjyotisa (commented):
Please follow our contributing guidelines
https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md
https://github.com/antiwork/.github/blob/main/CONTRIBUTING.md

PR #3582, opened 2026-02-13
Decouple PDF stamping from receipt email delivery

gianfrancopiana:
## Problem

Receipt emails for products with stampable PDFs can be delayed because `SendPurchaseReceiptJob` and `SendChargeReceiptJob` stamp every PDF synchronously before sending the email. The "View content" button is also hidden until stamping completes, making buyers think their purchase failed.

## Approach

Moved PDF stamping out of the receipt jobs and into the existing `StampPdfForPurchaseJob`, which runs on the `long` queue. Receipt jobs now send the email immediately and enqueue stamping as a separate background job. Since the receipt jobs are now fast, all callers route them to the `critical` queue unconditionally (removing the conditional `default`/`critical` routing that existed to protect the critical queue from long-running stamping).

Removed the stamping gate from `Purchase#has_content?` so the "View content" button appears immediately after purchase. The existing download flow (flash "preparing your files" + redirect) handles the case where a buyer clicks download before stamping finishes.

Also fixed a pre-existing bug where `StampForPurchase` would create a `StampedPdf` record with a nil URL when a product file has `cannot_be_stamped? == true`, and added timing instrumentation to the stamping service for observability.

The stamping engine itself (pdftk + Prawn), the download flow, and `StampPdfForPurchaseJob` are unchanged.

---

### DIFF
--- a/app/controllers/public_controller.rb
+++ b/app/controllers/public_controller.rb
@@ -43,7 +43,7 @@ class PublicController < ApplicationController
     if purchase.nil?
       render json: { success: false }
     else
-      SendPurchaseReceiptJob.set(queue: purchase.link.has_stampable_pdfs? ? "default" : "critical").perform_async(purchase.id)
+      SendPurchaseReceiptJob.set(queue: "critical").perform_async(purchase.id)
       render json: { success: true }
     end
   end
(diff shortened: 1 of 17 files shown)


slavingia:
Open to this but would be good to tie back to an issue and go through planning to ensure any edge cases are caught. Tricky code, PDF stamping is!

gianfrancopiana:
> Open to this but would be good to tie back to an issue and go through planning to ensure any edge cases are caught. Tricky code, PDF stamping is!

Created #3750.

Issue #3448, opened 2026-02-09
ChurnChart tooltip  causing overflow on last data point


abhix4:
when hovered on the last data point at Churn chart, the tooltip is causing screen overflow.

<img width="1852" height="972" alt="image" src="https://github.com/user-attachments/assets/25e66cb4-d4a5-4e37-a17e-28876edcc420" />

PR #3601, opened 2026-02-15
fix: prevent tooltip overflow on last data point (#3448)

Shine2099:
## Summary
Fixes issue #3448 where the ChurnChart tooltip causes screen overflow when hovering over the last data point.

## Changes
- Added boundary detection in  to prevent tooltip from overflowing the container edges
- Tooltip now adjusts its position when too close to the left or right edge

## Testing
- The fix constrains the tooltip position within the chart container boundaries
- Works for both left-edge and right-edge cases

### DIFF
--- a/app/javascript/components/Analytics/useChartTooltip.ts
+++ b/app/javascript/components/Analytics/useChartTooltip.ts
@@ -1,6 +1,7 @@
 import * as React from "react";
 
 const Y_OFFSET = 16;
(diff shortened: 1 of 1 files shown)


nyomanjyotisa (changes requested):
Please follow contributing guidelines 
https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md
https://github.com/antiwork/.github/blob/main/CONTRIBUTING.md

Issue #2756, opened 2026-01-08
Gifted memberships not visible in library


vatsalkaushik:
Gift Purchase ID: `vvr1NrRhO-7SGKiVY-yXnQ==`

The giftee archived the purchase from their library and is unable to unarchive it. 

1. The `/library` page does not show the sidebar to filter archived products

<img width="1899" height="1550" alt="Image" src="https://github.com/user-attachments/assets/bc90cc3f-8b39-4422-96db-192b7c08057d" />

2. Going directly to `/library?show_archived_only=true` shows a blank page

<img width="1866" height="1626" alt="Image" src="https://github.com/user-attachments/assets/290f17bc-bd3a-4dc3-b534-ef2639d66714" />

3. I clicked the "Unarchive from library" button on the purchase's download page, but the purchase is missing even after that. 

<img width="786" height="2083" alt="Image" src="https://github.com/user-attachments/assets/83a20cb4-c047-442d-93a6-20eae277a5e4" />

Maybe this being a gift purchase has something to do with this?

vatsalkaushik:
Hi @suprisedbeing:
1. purchase.subscription is NOT nil
2. purchase.link.is_recurring_billing IS true

PR #3619, opened 2026-02-16
Show gifted membership purchases in library

nyomanjyotisa:
Resolves https://github.com/antiwork/gumroad/issues/2756

# Description

## Problem
Gifted membership/subscription purchases are not visible in the giftee's library. The `for_library` scope uses `not_recurring_charge`, which calls `not_subscription_or_original_purchase`. This scope only includes purchases where `subscription_id IS NULL` or `is_original_subscription_purchase` is set. Gift receiver subscription purchases have a `subscription_id` (not null) but `is_original_subscription_purchase` is explicitly `false`, causing them to be incorrectly filtered out as recurring charges.

## Solution

Updated the `not_subscription_or_original_purchase` scope in `Purchase` to also include purchases with the `is_gift_receiver_purchase` flag set.

---

# Before/After

## Before (Gift not visible on library)
image

## After
image

---

# AI Disclosure

Claude Opus 4.6 for code generation. All code was self reviewed.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -474,8 +474,9 @@ class Purchase < ApplicationRecord
   scope :not_additional_contribution, -> { where("purchases.flags IS NULL OR purchases.flags & ? = 0", Purchase.flag_mapping["flags"][:is_additional_contribution]) }
   scope :for_products, ->(products) { where(link_id: products) if products.present? }
   scope :not_subscription_or_original_purchase, -> {
-    where("purchases.subscription_id IS NULL OR purchases.flags & ? = ?",
-          Purchase.flag_mapping["flags"][:is_original_subscription_purchase], Purchase.flag_mapping["flags"][:is_original_subscription_purchase])
+    where("purchases.subscription_id IS NULL OR purchases.flags & ? = ? OR purchases.flags & ? = ?",
+          Purchase.flag_mapping["flags"][:is_original_subscription_purchase], Purchase.flag_mapping["flags"][:is_original_subscription_purchase],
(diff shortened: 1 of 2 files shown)


app/models/purchase.rb
```diff
@@ -474,8 +474,9 @@ class Purchase < ApplicationRecord
   scope :not_additional_contribution, -> { where("purchases.flags IS NULL OR purchases.flags & ? = 0", Purchase.flag_mapping["flags"][:is_additional_contribution]) }
   scope :for_products, ->(products) { where(link_id: products) if products.present? }
   scope :not_subscription_or_original_purchase, -> {
-    where("purchases.subscription_id IS NULL OR purchases.flags & ? = ?",
-          Purchase.flag_mapping["flags"][:is_original_subscription_purchase], Purchase.flag_mapping["flags"][:is_original_subscription_purchase])
+    where("purchases.subscription_id IS NULL OR purchases.flags & ? = ? OR purchases.flags & ? = ?",
+          Purchase.flag_mapping["flags"][:is_original_subscription_purchase], Purchase.flag_mapping["flags"][:is_original_subscription_purchase],
+          Purchase.flag_mapping["flags"][:is_gift_receiver_purchase], Purchase.flag_mapping["flags"][:is_gift_receiver_purchase])
```
nyomanjyotisa, line 479: Gift receiver purchases have a `subscription_id` but `is_original_subscription_purchase` is false, so the scope incorrectly treats them as recurring charges and filters them out. Adding `is_gift_receiver_purchase` as a third condition keeps them visible since they're always initial purchases, not recurring ones.

Issue #3588, opened 2026-02-14
Cover image upload does not show immediate loading feedback


Aseem-ctrl:
### Description

When uploading a cover image on the product edit page, there is a noticeable delay (around 4–5 seconds) before the loading indicator appears. During this time, there is no visible feedback to the user, which may create confusion about whether the upload has started.

---

### Steps to Reproduce

1. Go to **Product → Edit** page  
2. Upload a cover image (tested with 300KB–10MB images)  
3. Observe the time before the loading spinner appears  
---

### Current Behavior

- After selecting a file, there is a 4–5 second delay.  
- The loading spinner appears only after this delay.  
- Users may think the upload did not trigger.  
---

### Expected Behavior

- The loading indicator should appear immediately after file selection.  
- Users should receive instant visual feedback that the upload process has started.  
---

### Screen Recording

The following recordings demonstrate the current behavior:

https://github.com/user-attachments/assets/970ad9ae-a914-4939-be8c-a1f77bb27520

https://github.com/user-attachments/assets/f9678549-e372-468d-9d43-e9ce20a22726

Ref: https://gumroad.com/products/zhsadw/edit

PR #3626, opened 2026-02-16
Show loading indicator immediately on cover image upload

misamikasa:
Closed https://github.com/antiwork/gumroad/issues/3588

# Description

## Problem
When uploading a cover image, there is a 4–5 second delay before the loading spinner appears (not reproducible in localhost, only a bit delay on local) . The `setIsUploading(true)` call only happened inside `saveCover`, which runs after `DirectUpload.create` completes the S3 upload. Users see no feedback during the actual upload.

Additionally, `DirectUpload.create` error callbacks never called `resolve()`, causing the upload loop to hang indefinitely on failure. The URL upload path used `.then()` instead of `.finally()`, so UI state wouldn't reset on errors.

## Solution

- Move `setIsUploading(true)` to trigger immediately on file selection, before `DirectUpload.create`
- Remove loading state management from `saveCover` — callers own the full lifecycle
- Filter invalid files before starting uploads to avoid setting loading state when no valid files exist
- Add `resolve()` in the `DirectUpload.create` error branch to prevent the upload loop from hanging
- Use `.finally()` for the URL upload path so state resets on both success and error

---

# Before/After

## Before (not reproducible in localhost, only a bit delay on local)

https://github.com/user-attachments/assets/b42c17f4-de61-41f4-b990-697899b7e0f3

## After

https://github.com/user-attachments/assets/1ccb8ecd-7ba5-461b-a18a-61108350f02c

---

# Test Results

No test changes needed for this

---

# Checklist

---

# AI Disclosure
Claude Sonnet 4.5 for code generation

### DIFF
--- a/spec/requests/products/edit/multiple_preview_spec.rb
+++ b/spec/requests/products/edit/multiple_preview_spec.rb
@@ -50,6 +50,7 @@ describe("Product edit multiple-preview Scenario", type: :system, js: true) do
         upload_image
         expect(page).to have_selector("[role='progressbar']")
       end
+      expect(page).to have_no_selector("[role='progressbar']")
       wait_for_ajax
     end.to change { product.reload.asset_previews.alive.count }.by(1)
   end


app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
```diff
@@ -142,14 +142,11 @@ const CoverUploader = ({
...
   const saveCover = async (coverPayload: CoverPayload) => {
     try {
-      setIsUploading(true);
       const covers = await createCover(permalink, coverPayload);
       setCovers(covers);
     } catch (e) {
       assertResponseError(e);
       showAlert(e.message, "error");
-    } finally {
-      setIsUploading(false);
```
misamikasa, line 152: Removed `isUploading` state from `saveCover`, callers now own the full lifecycle. Previously the `finally` block reset it after each file, causing the spinner to flicker off between files in multi-file uploads.

app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
```diff
@@ -169,22 +166,30 @@ const CoverUploader = ({
                 onChange={asyncVoid(async (event) => {
                   if (!event.target.files?.length) return;
 
-                  for (const file of event.target.files) {
+                  const validFiles = Array.from(event.target.files).filter((file) => {
```
misamikasa, line 169: Filter invalid files before the loop so we don't `setIsUploading(true)` when there's nothing to upload, otherwise the spinner would show with no way to dismiss

app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
```diff
@@ -169,22 +166,30 @@ const CoverUploader = ({
...
                     if (!FileUtils.isFileNameExtensionAllowed(file.name, ALLOWED_EXTENSIONS)) {
                       showAlert("Invalid file type.", "error");
-                      continue;
+                      return false;
                     }
+                    return true;
+                  });
+                  if (validFiles.length === 0) return;
+
+                  setIsUploading(true);
```
misamikasa, line 178: Core fix, this previously only fired inside `saveCover`, which runs after `DirectUpload.create` finishes the S3 upload (4–5s delay). Now it triggers immediately on file selection

app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
```diff
@@ -221,7 +226,9 @@ const CoverUploader = ({
               <Button
                 color="primary"
                 onClick={() => {
-                  void saveCover({ type: "url", url: uploader.value }).then(() => {
+                  setIsUploading(true);
+                  void saveCover({ type: "url", url: uploader.value }).finally(() => {
+                    setIsUploading(false);
```
misamikasa, line 231: If `saveCover` threw, the UI would stay stuck in loading state. `.finally()` ensures cleanup always runs

app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx
```diff
@@ -169,22 +166,30 @@ const CoverUploader = ({
...
+                  if (validFiles.length === 0) return;
+
+                  setIsUploading(true);
+                  for (const file of validFiles) {
                     // TODO change the relevant endpoint(s) to allow uploading multiple files at once
                     await new Promise<void>((resolve) => {
                       new DirectUpload(file, "/rails/active_storage/direct_uploads").create((error, blob) => {
                         if (error) {
                           showAlert(error.message, "error");
+                          resolve();
```
misamikasa, line 185: Pre-existing bug: the original code never called `resolve()` on `DirectUpload` error, so the Promise hung forever and the UI froze

nyomanjyotisa (changes requested):
Some failing tests seems related, please check @misamikasa

misamikasa:
Fixed @nyomanjyotisa

nyomanjyotisa (approved):
Looks good to me, thank you

PR #3547, opened 2026-02-12
Close button in mobile view of nested menu

rishisulakhe:
Issue: #2724 

# Description
In nested menu mobile view close button 'x' was missing

# Before/After

<table>
  <tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
  <td>
image
</td>
    <td>
image
</td>
<tr>
</table>

# Checklist

- [X] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [X] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [X] I have performed a self-review and left review comments on my PR
- [ ] I have added/updated tests for my changes

---

# AI Disclosure
No ai was used

PR #1851, opened 2025-10-29
 chore: migrate  _nested_menu.scss to Tailwind 

rishisulakhe:
Part of #1055 
Delete _nested_menu.scss and apply the same styles using Tailwind - Its only used in one component( NestedMenu.tsx)

#  Before vs After (Camparison 1:1)

<table>
  <tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
  <td>
image
</td>
    <td>
image
</td>
  </tr>
   <tr>
 <td>
Screenshot 2025-10-15 134438
</td>
    <td>
image
</td>
  </tr>
   <tr>
  <td>
image
</td>
    <td>
image
</td>
  </tr>
   <tr>
    <td>
image
</td>
    <td>
image
</td>
  </tr>
   <tr>
    <td>
image
</td>
    <td>
image</td>
  </tr>
   <tr>
    <td>
image
</td>
    <td>
image
</td>
  </tr>
</table>

# Demo
### Dark-mode

https://github.com/user-attachments/assets/36e31f3b-b00d-4ac7-9e48-52123d3dc411

### Light-mode

https://github.com/user-attachments/assets/3975cc20-51e0-4456-8584-415a163f4500

## Mobile
Now we don't need the 'X' for closing, clicking the backdrop or 'ESC' will close the sidebar.

https://github.com/user-attachments/assets/bdba328c-7773-456b-98ca-678130089641

## AI Disclosure:
Claude code used to find relevant files then reviewed/updated and verified that no file is missing to update.

PR #3643, opened 2026-02-17
Add close button to nested menu overlay sidebar

### LINKED ISSUE #2724: Issues in Nested Menu
I noticed a different color for the menu(Nested Menu) on the site. As this is something I was also working on (#2068 - Migrating *_nested_menu.scss* to Tailwind), I also found a few more issues too, listing them all below.

### 1.Color mismatch

**Current** - The menu button's background color now is the same as the overall background color but that wasn't the case before. It had color same as search bar's color.
**Expected** - The menu button's background color to be the same as the one applied to search bar.

 <table>
   <tr>
     <td></td>
     <td></td>
   </tr>
 </table>

### 2.Close button removed

**Current** - The close button that existed before isn't there in the sidebar. I don't think that was intended, was it?
**Expected** - A close button to close the sidebar.

<br />
<br />

### 3.Code clean up
I noticed there's some bits of code in the merged PR that could be removed like the redundant `<div>` tag and the prop `menuTop`.

**Notes**:
Small fixes, except maybe 2 but an easy fix. I think a different approach is taken that modifies the `Sheet` component.

285729101:
Fixes #2724 (close button)

## Summary

The overlay sidebar for the nested menu (shown on mobile/smaller screens) lost its close button during the Tailwind migration in #1851. This restores the X button, matching the [Figma design](https://www.figma.com/design/cqBGWkI4iZe47mhvrEF1SY/Discover-on-homepage?node-id=832-26998&t=JxqMqbpJVWyKT48f-4) shared by @laugardie.

### Changes

- **Close button**: Added `SheetCloseButton` — a reusable close button extracted from `SheetHeader` — and placed it inside the `OverlayMenu` sidebar at the top-right (`absolute right-4 top-4`). This uses Radix's `Dialog.Close` so it works with the existing `Sheet`/`Dialog.Root` state management. The button is inside the sidebar panel (not the overlay), so it remains visible on narrow devices like the Samsung Galaxy Fold.
- **Refactored `SheetHeader`**: Now uses the shared `SheetCloseButton` component, reducing duplication.
- **Removed redundant wrapper `<div>`**: The `NestedMenu` component had an unnecessary `<div>` wrapping the conditional rendering of `Menubar` / `OverlayMenu`.
- **Removed unused `menuTop` prop**: The `menuTop` prop was passed through `NestedMenu` → `OverlayMenu` but was never applied to any element. Removed from `NestedMenuProps`, `NestedMenu`, `OverlayMenu`, and the `Nav` call site.

### Root cause

PR #1851 migrated `_nested_menu.scss` to Tailwind and replaced the sidebar implementation with the `Sheet` component, but did not include a close button. The `Sheet` component supports `Dialog.Close` via Radix UI, but a close button was only rendered through `SheetHeader`, which is not used by the nested menu overlay.

## Test plan

- [ ] Open the Discover page on a mobile viewport or narrow browser window
- [ ] Click the filter/categories button to open the overlay sidebar
- [ ] Verify the X button appears at the top-right corner inside the sidebar
- [ ] Click the X button and confirm the sidebar closes
- [ ] Verify the backdrop click still closes the sidebar
- [ ] Test on narrow devices (e.g. Samsung Galaxy Fold viewport) to confirm the button is not cut off
- [ ] Test in both light and dark mode
- [ ] Verify `SheetHeader` close button still works correctly in other pages (e.g. Emails, Followers, Affiliates)

## Checklist

---

## AI Disclosure

This PR was authored with Claude Code (Claude Opus 4.6). The implementation approach and code were reviewed manually.

### DIFF
--- a/app/javascript/components/Discover/Nav.tsx
+++ b/app/javascript/components/Discover/Nav.tsx
@@ -57,7 +57,6 @@ export const Nav = ({
           if (item.href) onClickTaxonomy(item.href.replace(/^\//u, ""));
         }}
         footer={footer}
-        menuTop="0px"
       />
     </div>
   );
--- a/app/javascript/components/NestedMenu.tsx
+++ b/app/javascript/components/NestedMenu.tsx
@@ -9,7 +9,7 @@ import { Button } from "$app/components/Button";
 import { useCurrentSeller } from "$app/components/CurrentSeller";
 import { Icon } from "$app/components/Icons";
 import { Popover, PopoverContent, PopoverTrigger } from "$app/components/Popover";
-import { Sheet } from "$app/components/ui/Sheet";
+import { Sheet, SheetCloseButton } from "$app/components/ui/Sheet";
 import { useIsOnTouchDevice } from "$app/components/useIsOnTouchDevice";
 import { useOnOutsideClick } from "$app/components/useOnOutsideClick";
 import { useWindowDimensions } from "$app/components/useWindowDimensions";
@@ -34,7 +34,6 @@ type NestedMenuProps = {
   moreLabel?: string;
   buttonLabel?: string;
   footer?: React.ReactNode;
-  menuTop?: string;
 } & React.AriaAttributes;
 
 export const NestedMenu = ({
@@ -45,7 +44,6 @@ export const NestedMenu = ({
   moreLabel,
   buttonLabel,
   footer,
-  menuTop,
   ...extraAriaAttrs
 }: NestedMenuProps) => {
   const itemsMap = React.useMemo(() => {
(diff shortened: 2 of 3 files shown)


285729101 (commented):
Self-review:

**Sheet.tsx** — Extracted `SheetCloseButton` from `SheetHeader` so the close button can be reused independently. This avoids modifying the `Sheet` component's API with a conditional prop (the approach taken in the closed #3547), and keeps the close button inside the sidebar content rather than on the overlay. This prevents the cut-off issue on narrow devices that was flagged in the issue discussion. `SheetHeader` now uses the shared component to avoid duplication.

**NestedMenu.tsx** — The close button is positioned `absolute right-4 top-4` inside the sidebar. Since the `Sheet` content uses `position: fixed` via Radix, the absolute positioning is relative to the fixed content panel. The button remains inside the panel boundary on all viewport widths, including the Samsung Galaxy Fold (280px).

**Nav.tsx** — Removed the `menuTop="0px"` prop which was passed through but never actually applied to any DOM element.

PR #3649, opened 2026-02-17
Fix product duplication failing on products with validation errors

nyomanjyotisa:
Resolves https://github.com/antiwork/gumroad-private/issues/135

# Description

## Problem
Some products cannot be duplicated because they have pre-existing validation errors. The duplication flow crashes with a 500 error and no message is shown to the user.

## Solution

- Skip validations when setting the duplication flag and saving the initial duplicate
- Show the actual error message to the user when duplication fails

---

# Before/After

## Before (Stuck on duplicating, error on background)

https://github.com/user-attachments/assets/cd62fbc1-b5be-423b-ad3e-b8416bb403aa

## After

https://github.com/user-attachments/assets/dd14afce-1cd6-4805-b626-61d05c544e69

---

# AI Disclosure

Claude Opus 4.6 for code generation. All code was self reviewed.


app/controllers/product_duplicates_controller.rb
```diff
@@ -11,7 +11,8 @@ def create
     end
 
     DuplicateProductWorker.perform_async(@product.id)
-    @product.update!(is_duplicating: true)
+    @product.is_duplicating = true
+    @product.save!(validate: false)
```
nyomanjyotisa, line 17: `is_duplicating` is a FlagShihTzu bitfield flag. Products with pre-existing validation errors (e.g. missing call durations) cause `update!` to fail with a 500. We only need to toggle a boolean flag here, so skipping validation is safe.

app/services/product_duplicator_service.rb
```diff
@@ -57,14 +57,22 @@ def duplicate
 
     set_recently_duplicated_product
 
-    duplicated_product
+    duplicated_product.reload
```
nyomanjyotisa, line 61: `save!(validate: false)` skips the `before_validation :associate_price` callback, which means default prices are not created in memory. This leaves the prices association cache stale. reload clears all association caches so callers get fresh data

app/services/product_duplicator_service.rb
```diff
@@ -57,14 +57,22 @@ def duplicate
...
   end
 
   def recently_duplicated_product
     duplicated_product_id = REDIS_STORAGE_NS.get(product.id)
     Link.where(id: duplicated_product_id).first
   end
 
+  def store_duplication_error(message)
+    REDIS_STORAGE_NS.setex("error:#{product.id}", TIMEOUT_FOR_DUPLICATE_PRODUCT_CACHE, message)
```
nyomanjyotisa, line 74: Reuses the existing Redis namespace. The `error:` prefix avoids key collision with the duplicated product ID stored under the plain product.id key.

app/controllers/product_duplicates_controller.rb
vishaltelangre: Would it make sense to fix such invalid data in the first place?

binary-koan: Yeah would be good to know a bit more about the issue - are there other known cases? What makes it hard to just fix the data to be valid?

IMO an ideal solution would be to fix known cases and add a visible error alert

nyomanjyotisa: @vishaltelangre @binary-koan 

User can create a product without completing all required fields.

See the following video, after creating product on https://gumroad.com/products/new some required data still need to be filled on the product edit page. And same error is displayed when try to save the product without filling the required data

https://github.com/user-attachments/assets/64aafdaf-801f-4f91-9702-379bef35b136

This PR ensures we surface the same validation error when duplicating a product that has empty required fields, instead of silently fails.

binary-koan: Ah interesting. IMO being able to create invalid records is a big red flag in our business logic (based on that UX we should have some kind of "newly created" state which skips those validations) but makes sense to just work around it for this bug fix.

Please add a comment about it though as it's not obvious why we need to skip validation

app/services/product_duplicator_service.rb
```diff
@@ -34,7 +34,7 @@ def duplicate
       duplicated_product.is_collab = false
       mark_duplicate_product_as_draft
       duplicated_product.is_duplicating = false
-      duplicated_product.save!
+      duplicated_product.save!(validate: false)
```
binary-koan: I'm confused - this looks like we force-create the duplicate despite validation errors, but the after screencast shows validation errors? Can you explain the flow a bit more?

app/sidekiq/duplicate_product_worker.rb
```diff
@@ -9,8 +9,11 @@ def perform(product_id)
   rescue => e
     logger.error("Error while duplicating product id '#{product_id}': #{e.inspect}")
     Bugsnag.notify(e)
+    error_message = e.is_a?(ActiveRecord::RecordInvalid) ? e.record.errors.full_messages.first : e.message
+    ProductDuplicatorService.new(product_id).store_duplication_error(error_message)
```
binary-koan: Would be nicer if we did this in a `rescue` block in the service itself

app/sidekiq/duplicate_product_worker.rb
```diff
@@ -9,8 +9,11 @@ def perform(product_id)
...
   ensure
     product = Link.find(product_id)
-    product.update!(is_duplicating: false)
+    product.is_duplicating = false
+    product.save!(validate: false)
```
binary-koan, line 17: A quick comment here would be nice as well

app/controllers/product_duplicates_controller.rb
slavingia: Correct. What fields are required? They shouldn’t be if we allow creation.

slavingia: I believe we have a draft / unpublished state.

nyomanjyotisa:
@binary-koan Could you have another look at this?

I have updated the [PR description](https://github.com/antiwork/gumroad/pull/3649#issue-3951231676) with the new after video. Now user can duplicate newly created product that has validation errors.

We skip validation in 4 places to make this happen:

On mark the product as duplicating
https://github.com/antiwork/gumroad/blob/61cc039b3015888ef32be0b1f5519cafc512a6d8/app/controllers/product_duplicates_controller.rb#L14-L17

On duplicate the product
https://github.com/antiwork/gumroad/blob/61cc039b3015888ef32be0b1f5519cafc512a6d8/app/services/product_duplicator_service.rb#L38

On duplicate the product files -> then update the product description with the duplicated files
https://github.com/antiwork/gumroad/blob/61cc039b3015888ef32be0b1f5519cafc512a6d8/app/services/product_duplicator_service.rb#L209

And on mark the product as not duplicating
https://github.com/antiwork/gumroad/blob/61cc039b3015888ef32be0b1f5519cafc512a6d8/app/sidekiq/duplicate_product_worker.rb#L14-L17

app/services/product_duplicator_service.rb
```diff
@@ -27,14 +27,15 @@ def duplicate
     ApplicationRecord.connection.transaction do
       @duplicated_product = product.dup
       duplicated_product.unique_permalink = nil
+      duplicated_product.send(:set_unique_permalink)
```
nyomanjyotisa, line 30: Since we skip all validations, we need to manually trigger the `set_unique_permalink`

binary-koan:
I looked into this a bit more. Would it work to change the `on: :update` validations to `unless: :draft`? I think that would work better with the new product edit flow we're going for (where you can freely navigate around and save anything as draft, and it only validates when you publish)

If that breaks a bunch of assumptions and causes a lot more work we can do this for now, but we'll probably need to change it at some point in the new edit UX anyway

(actually it might be bad UX with the current edit flow, we can go with this for now if it is)

nyomanjyotisa:
@binary-koan Updated to use `unless: :draft?` on all `on: :update` validations, and it works as expected, and the validation only triggers on publish:


https://github.com/user-attachments/assets/210c84b0-ce36-49f7-ab83-1096bcec4f4c

binary-koan:
Ah that's what I was meaning by bad UX, I don't think we should show an error message on the content tab when the solution is to go back to the product tab. Could you switch back to skipping validation for now? We can cherry-pick the `on: :update` revision when we work on product edit error handling

nyomanjyotisa:
Reverted back to skipping validation

Issue #3650, opened 2026-02-17
Make Application Icon delete UI consistent with Profile Logo


Aseem-ctrl:
In Settings → Advanced → Create Application, once an Application Icon is uploaded, there is currently no direct way to remove it from the UI.

In comparison, the Profile Logo section in Settings provides a trash icon overlay directly on the image preview, allowing users to remove the logo easily and intuitively.

For consistency and improved usability, the Application Icon component could follow a similar pattern by displaying a trash icon overlay on the uploaded image.

---

Expected Behavior
	•	After uploading an application icon, a small trash icon should appear on the image preview.
	•	Clicking it should remove the uploaded icon.

---
<div style = "display:flex; align-items:center; gap:20px;">
<img src="https://github.com/user-attachments/assets/03aacba6-630e-45cc-b369-2a7405e5dede"
           width="500"
           style="border-radius:8px;" />

<div style = "width:1px ; height:280px ; background-color:#444 ;"></div>
 <img src="https://github.com/user-attachments/assets/a5dc8ecf-e1f6-47ee-8af7-39a92ba779dc" 
           width="500"
           style+"border-radius:8px;"  />
</div>

Ref: https://gumroad.dev/settings/advanced

PR #3656, opened 2026-02-17
Add delete button for application icon consistent with profile logo

misamikasa:
Closes https://github.com/antiwork/gumroad/issues/3650

# Description

## Problem
The application icon upload in Settings > Advanced > Create/Edit Application has no way to remove an uploaded icon. The profile logo section provides a trash icon overlay for easy deletion, but the application icon lacks this entirely.

## Solution

- Added a trash icon overlay on the application icon preview, matching the existing `ImageUploader` pattern used by the profile logo (`Button` with `color="primary"`, `small`, `absolute top-2 right-2`, `Icon name="trash2"`)
- Added backend support in `Oauth::ApplicationsController` to purge the attached file when `signed_blob_id` is sent as `null`, following the same `has_key?` + `purge` pattern used in `Settings::ProfileController` for avatar removal
- The frontend already sends `signed_blob_id: null` when no icon is set, so the existing `handleSubmit` logic works without changes

Without the backend change, setting `icon` to `null` on the frontend is purely cosmetic and reverts on page reload.

---

# Before/After

## Before

image

## After

https://github.com/user-attachments/assets/4e784a4c-9832-4bc1-a171-786f1f0f333a

---

# Test Results

Unable to run the new tests on my local
image

---

# Checklist

---

# AI Disclosure

Claude sonnet 4.5 used for this PR

### DIFF
--- a/spec/controllers/oauth/applications_controller_spec.rb
+++ b/spec/controllers/oauth/applications_controller_spec.rb
@@ -95,6 +95,7 @@ describe Oauth::ApplicationsController, type: :controller, inertia: true do
         application = OauthApplication.last
         expect(application.file.attached?).to be(true)
         expect(response).to redirect_to(edit_oauth_application_path(application.external_id))
+      end
     end
 
     describe "bad create params" do


app/controllers/oauth/applications_controller.rb
```diff
@@ -32,6 +32,8 @@ def create
 
     if params[:signed_blob_id].present?
       @application.file.attach(params[:signed_blob_id])
+    elsif params.has_key?(:signed_blob_id) && @application.file.attached?
+      @application.file.purge
```
misamikasa, line 36: Following the same pattern used in `Settings::ProfileController#update` for avatar removal. When the frontend sends `signed_blob_id: null` (key present but value blank), we purge the existing file. If the key is omitted entirely, the icon is left unchanged

Issue #3692, opened 2026-02-19
Unexpected Page Scrolling While Editing a Workflow


michellelarney:
https://github.com/user-attachments/assets/47b3ed60-18d0-463e-b599-b701014d0e2a

When editing a workflow, the page unexpectedly scrolls without user intention.

gumclaw:
Assigned to @gianfrancopiana (UI/frontend). Assigned based on related work — please reassign if this isn't the right person.

andev18:
@gianfrancopiana Please have a look at this PR when you have a chance https://github.com/andev18/gumroad/pull/3

Thanks!

PR #3693, opened 2026-02-19
Fix unexpected page scrolling while editing workflow emails

aryanghai12:
Issue: #3692

# Description

## Problem
When editing workflow emails, the page scrolled unexpectedly during normal user interactions, including typing in the editor, creating a new email, saving changes (both successful and failed), and on initial page load.

**Root Cause:** Multiple `useEffect` hooks in the `EmailPreview` and `EmailRow` components were triggering `scrollIntoView` at inappropriate times. This was caused by missing dependency arrays, a lack of initial mount tracking, duplicate `focusedFieldInfo` state updates, and missing save state tracking.

## Solution
Fixed the rendering lifecycle and scroll triggers across the email components to ensure the UI remains stable:
* **Prevented Scroll on Mount & Duplicates:** Added `isInitialMount` (in `EmailPreview`) and `prevFocusedFieldRef` (in `EmailRow`) to skip scrolling on the first render and ignore duplicate focus events. Added proper dependency arrays (`[focusedFieldInfo, email.id]`).
* **Save State Tracking:** Introduced an `isSavingRef` to block scrolling during saves. The ref remains true for 300ms after a save completes to prevent immediate erratic scrolling.
* **State Optimization:** Guarded toggle/focus handlers and validation logic to only update `focusedFieldInfo` when values *actually* change, preventing identical objects from triggering re-renders. 
* **Typing Stability:** Removed the `setFocusedFieldInfo` call from the `updateEmail` function so it no longer triggers on every keystroke.

---

# Before/After
| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/3e9a2cca-dbcb-4dd5-b67b-662141c34aea" width="400" controls></video> | <video src="https://github.com/user-attachments/assets/829bac2a-3ec3-4086-aacb-b16b89e9563b" width="400" conmtrols></video> |

---

# Test Results

**Automated Tests Added:**
- Verified no scroll on initial page load
- Verified no erratic scrolling when creating an email (includes typing behavior)
- Verified no repeated scrolling when validation fails multiple times
- Verified no scroll after clicking save

test res

---

# Checklist

---

# AI Disclosure
GitHub Copilot was used to help identify the issue. The implementation logic was reviewed and verified manually.

---

### DIFF
--- a/spec/requests/workflows_spec.rb
+++ b/spec/requests/workflows_spec.rb
@@ -24,7 +24,6 @@ describe("Workflows", js: true, type: :system) do
   before do
     @product = create(:product, name: "product name", user: seller, created_at: 2.hours.ago)
     @product2 = create(:product, name: "product 2 name", user: seller, created_at: 1.hour.ago)
-    create(:purchase, link: @product)
     index_model_records(Purchase)
 
     allow_any_instance_of(User).to receive(:sales_cents_total).and_return(Installment::MINIMUM_SALES_CENTS_VALUE)
@@ -1732,7 +1731,7 @@ describe("Workflows", js: true, type: :system) do
       expect(scroll_after_second_save).to eq(scroll_after_first_save)
     end
 
-    it "does not scroll after successful save" do
+    it "does not scroll after clicking save" do
       visit workflow_emails_path(@workflow.external_id)
 
       within_section "Create emails for your workflow" do
@@ -1751,11 +1750,9 @@ describe("Workflows", js: true, type: :system) do
       scroll_before_save = page.evaluate_script("window.pageYOffset")
 
       click_on "Save changes"
-      expect(page).to have_alert(text: "Changes saved!")
 
-      sleep 0.3
+      sleep 0.5
 
-      # Verify scroll position hasn't changed after successful save
       scroll_after_save = page.evaluate_script("window.pageYOffset")
       expect(scroll_after_save).to eq(scroll_before_save)
     end


app/javascript/components/WorkflowsPage/WorkflowEmails.tsx
```diff
@@ -110,7 +110,6 @@ const WorkflowEmails = ({ context, workflow }: WorkflowEmailsProps) => {
   );
   const updateEmail = (id: string, value: Partial<EditableEmailFormState>) => {
     setEmails((prev) => prev.map((email) => (email.id === id ? { ...email, ...value } : email)));
-    setFocusedFieldInfo({ emailId: id, fieldName: Object.keys(value)[0] ?? null });
```
aryanghai12, line 113: Removed the setFocusedFieldInfo call that was triggering on every field edit. This was causing unnecessary re-renders and scroll triggers. The field update itself doesn't require focus tracking , only explicit user actions like clicking "Edit" or validation errors need to set focus.

app/javascript/components/WorkflowsPage/WorkflowEmails.tsx
```diff
@@ -173,6 +172,7 @@ const WorkflowEmails = ({ context, workflow }: WorkflowEmailsProps) => {
     setFocusedFieldInfo({ emailId: id, fieldName: "name" });
   };
   const [focusedFieldInfo, setFocusedFieldInfo] = React.useState<FocusedFieldInfo | null>(null);
+  const isSavingRef = React.useRef(false);
```
aryanghai12, line 175: Added a ref to track save state instead of using component state. Using a ref prevents re-renders and provides a reliable way to block scroll effects during and immediately after save operations. The 300ms delay after save completion ensures all state updates have settled before allowing scrolls again.

app/javascript/components/WorkflowsPage/WorkflowEmails.tsx
```diff
@@ -202,23 +202,33 @@ const WorkflowEmails = ({ context, workflow }: WorkflowEmailsProps) => {
...
+        if (!firstInvalidEmailId) firstInvalidEmailId = email.id;
         setExpandedEmailIds((prev) => new Set(prev).add(email.id));
       }
     });
     setInvalidFields(invalidFields);
+
+    if (firstInvalidEmailId && focusedFieldInfo?.emailId !== firstInvalidEmailId) {
+      setFocusedFieldInfo({ emailId: firstInvalidEmailId, fieldName: "name" });
+    }
+
```
aryanghai12, line 218: Previously was calling setFocusedFieldInfo inside the forEach loop for every invalid email, creating multiple object references and triggering the scroll effect repeatedly. Now collects the first invalid email and only updates focusedFieldInfo if it's different from the current value, preventing duplicate scroll triggers.

app/javascript/components/WorkflowsPage/WorkflowEmails.tsx
```diff
@@ -251,6 +261,15 @@ const WorkflowEmails = ({ context, workflow }: WorkflowEmailsProps) => {
...
 
+  React.useEffect(() => {
+    if (!form.processing && isSavingRef.current) {
+      const timer = setTimeout(() => {
+        isSavingRef.current = false;
+      }, 300);
+      return () => clearTimeout(timer);
+    }
+  }, [form.processing]);
+
```
aryanghai12, line 272: Resets isSavingRef after form processing completes with a 300ms buffer. This delay is necessary because Inertia.js updates can trigger component re-renders after form.processing becomes false. The timeout ensures we don't unblock scrolling until all cascading updates have settled.

app/javascript/components/WorkflowsPage/WorkflowEmails.tsx
```diff
@@ -433,13 +461,25 @@ const EmailRow = ({
...
 
     const { fieldName } = focusedFieldInfo;
+
+    const isNewFocus =
+      !prevFocusedFieldRef.current ||
+      prevFocusedFieldRef.current.emailId !== focusedFieldInfo.emailId ||
+      prevFocusedFieldRef.current.fieldName !== focusedFieldInfo.fieldName;
+
+    prevFocusedFieldRef.current = focusedFieldInfo;
+
```
aryanghai12, line 477: Added prevFocusedFieldRef to detect if this is a truly new focus event or just a re-render with the same focus info. Combined with the isSavingRef check, this prevents scrolling during saves, on non-focus updates, and on duplicate focus calls. The isNewFocus check compares both emailId and fieldName to catch any actual change.

slavingia:
Thanks for the contribution! The issue this PR is linked to still has a "planning" label, which means it's still being scoped by our team and isn't ready for a PR yet. We're closing this for now — please find an open issue without a "planning" label and we'd love your help there. We appreciate it!

PR #3698, opened 2026-02-19
Show actual Stripe error messages on payment settings page

gianfrancopiana:
Fixes antiwork/gumroad-private#133

## Problem

When users update their payment settings and Stripe returns an error, the frontend replaces the specific, actionable error message (e.g., address/postal code mismatch) with a generic "Your account could not be updated due to an error with Stripe." This leaves users and support staff unable to determine which field needs correction.

## Context

In Oct 2020, we started showing raw Stripe errors directly with the goal of reducing support tickets.

In July 2022, we started hiding Stripe errors behind a generic "contact us" message. This was justified because the actual error was pre-filled into the HelpScout Beacon (an embedded support widget that let users send messages to the support team). Clicking "contact us" would open Beacon with the real error already filled in, so support would receive it without the user needing to understand it.

In July 2023, the Beacon was removed – but the generic message stayed. Since then, users see a generic error with no way to surface the real one to support.

## Approach

The backend already sanitizes Stripe error messages and passes them to the frontend via `errors.base[0]`. The frontend had a special case that checked for `error_code === "stripe_error"` and swapped the real message for a generic one. Removed that branch so the actual error is always displayed.

---

### DIFF
--- a/app/controllers/settings/payments_controller.rb
+++ b/app/controllers/settings/payments_controller.rb
@@ -225,15 +225,13 @@ class Settings::PaymentsController < Settings::BaseController
           comment_type: :note,
           content: result[:error_message]
         )
-        redirect_with_error(result[:error_message], error_code: result[:error_code])
+        redirect_with_error(result[:error_message])
         false
       end
     end
 
-    def redirect_with_error(error_message, error_code: nil)
-      errors_hash = { base: [error_message] }
-      errors_hash[:error_code] = [error_code] if error_code.present?
-      redirect_to settings_payments_path, inertia: { errors: errors_hash }
+    def redirect_with_error(error_message)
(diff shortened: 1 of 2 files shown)


app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -113,7 +112,7 @@ type ErrorMessageInfo = {
 export default function PaymentsPage() {
   const page = usePage();
   const props = cast<PaymentsPageProps>(page.props);
-  const errors = cast<{ base?: string[]; error_code?: string[] } | undefined>(page.props.errors);
+  const errors = cast<{ base?: string[] } | undefined>(page.props.errors);
```
hchhabra, line 115: Can `error_code` be removed from the backend response as well now as it's no longer used?

Issue #3697, opened 2026-02-19
Fix failing test on Discover page filters


EmCousin:
The test at `./spec/requests/discover/filtering_spec.rb:280` has been failing consistently and should not be considered flaky.

Latest example in [this CI run](https://github.com/antiwork/gumroad/actions/runs/22187324844/job/64165714664), back trace below.

This is currently blocking automatic deployments to production. 

```sh
Failures:

  1) Discover - Filtering scenarios properly restores state and taxonomy category when pressing back
     Failure/Error: expect(page).to have_field "Maximum price", with: "3"
       expected to find visible field "Maximum price" that is not disabled with value "3" within #<Capybara::Node::Element tag="details" path="/HTML/BODY[1]/DIV[3]/DIV[2]/DIV[1]/SECTION[2]/DIV[2]/DIV[1]/DETAILS[1]"> but there were no matches. Also found "", which matched the selector but not all filters. Expected value to be "3" but was ""

     [Screenshot Image]: /app/tmp/capybara/failures_r_spec_example_groups_discover_filtering_scenarios_properly_restores_state_and_taxonomy_category_when_pressing_back_55.png


     # ./spec/requests/discover/filtering_spec.rb:280:in 'block (3 levels) in <main>'
     # /bundle_path/ruby/3.4.0/gems/capybara-3.40.0/lib/capybara/session.rb:366:in 'Capybara::Session#within'
     # /bundle_path/ruby/3.4.0/bundler/gems/capybara_accessible_selectors-161b8c5b1a0a/lib/capybara_accessible_selectors/selectors/disclosure.rb:143:in 'CapybaraAccessibleSelectors::Session#within_disclosure'
     # /bundle_path/ruby/3.4.0/bundler/gems/capybara_accessible_selectors-161b8c5b1a0a/lib/capybara_accessible_selectors/selectors/disclosure.rb:95:in 'CapybaraAccessibleSelectors::Actions#select_disclosure'
     # ./spec/requests/discover/filtering_spec.rb:279:in 'block (2 levels) in <main>'
     # ./spec/spec_helper.rb:278:in 'block (2 levels) in <main>'
     # ./spec/spec_helper.rb:249:in 'block (3 levels) in <main>'
     # ./app/modules/product/creation_limit.rb:14:in 'Product::CreationLimit::ClassMethods#bypass_product_creation_limit'
     # ./spec/spec_helper.rb:248:in 'block (2 levels) in <main>'
[...]

PR #3663, opened 2026-02-17
 Fix: Discover link in sidebar to use Inertia navigation to avoid full page reload

vedantvaghasiya:
closes: #3664 

### Problem:
- Clicking the "Discover" link in the sidebar caused a full page reload instead of using Inertia's client-side navigation. 
- The Discover link was using NavLink (server-side component with standard `<a>` tag) instead of ClientNavLink (which uses Inertia's Link component).

### Solution:
- Changed the Discover link to use ClientNavLink, consistent with all other navigation items in the sidebar.

---

# Before/After

## Before: 

https://github.com/user-attachments/assets/ba53effc-f948-4a86-80d3-02030f5db730

## After: 
- no full page reload 

https://github.com/user-attachments/assets/903ea1d3-2aec-45d3-b5c4-cf3dfd301f4e

---

# Test Results

- no tests update/addition required. 

---

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure

No AI was used to create this PR.

PR #3517, opened 2026-02-11
Migrate Discover Page to Inertia

4arjun:
Issue: #3047 

## Summary

Migrate the Discover page from React-on-Rails to Inertia.js.

<table>
  <thead>
    <tr>
      <th width="50%">Before</th>
      <th width="50%">After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td align="center">
        <video src="https://github.com/user-attachments/assets/f3b27eeb-f8cc-47a3-9a36-0d20e9cdfcc3" controls="controls" style="max-width: 100%;"></video>
      </td>
      <td align="center">
        <video src="https://github.com/user-attachments/assets/a44c87a6-8e94-41b4-a988-dc784d17869b" controls="controls" style="max-width: 100%;"></video>
      </td>
    </tr>
    <tr>
      <td align="center">
        <video src="https://github.com/user-attachments/assets/6a1dd840-e6bc-44db-a283-309ada180efd" controls="controls" style="max-width: 100%;"></video>
      </td>
      <td align="center">
        <video src="https://github.com/user-attachments/assets/27cefa86-d34d-4709-bd7e-791c3fa11004" controls="controls" style="max-width: 100%;"></video>
      </td>
    </tr>
  </tbody>
</table>

## Tests

`spec/controllers/discover_controller_spec.rb`
`spec/controllers/discover/search_autocomplete_controller_spec.rb`
`spec/requests/discover/discover_spec.rb`
`spec/requests/discover/search_spec.rb`
`spec/requests/discover/blackfriday_spec.rb`
`spec/requests/discover/discover_mobile_spec.rb`
`spec/modules/user/recommendations_spec.rb`

Screenshot 2026-02-11 at 2 28 48 AM

# Checklist

- [x] I have read the [contributing guidelines](https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md)
- [x] I have watched [Gumroad PR review livestreams](https://www.youtube.com/@anti-work)
- [x] I have performed a self-review and left review comments on my PR
- [x] I have added/updated tests for my changes

---

# AI Disclosure
Claude opus 4.5 used to make changes initially, final refactoring and review done by me.

PR #3699, opened 2026-02-19
Bug/discover deferred props history

EmCousin:
Issue: https://github.com/antiwork/gumroad/issues/3697

## Problem

When navigating to a new taxonomy, the search query clearing triggered an autocomplete reload with an empty query, which pushed extra entries onto the browser history stack. This broke back button navigation.

## Solution

- Guard autocomplete to only fire when query is non-empty
- use replace: true to prevent autocomplete from ever creating new history entries.

---

# Before/After

N/A : The issue is not reproducible manually, only in CI.

---

# Test Results

CI green

---

# Checklist

---

# AI Disclosure

No AI

### DIFF
--- a/app/javascript/pages/Discover/Index.tsx
+++ b/app/javascript/pages/Discover/Index.tsx
@@ -278,16 +278,24 @@ function DiscoverIndex() {
           parseUrlParams(window.location.href, props.curated_product_ids, defaultSortOrder).offer_code;
 
       if (shouldFetchRecommendations) {
-        router.get(url.toString(), {}, {
-          preserveState: true,
-          preserveScroll: true,
-        });
+        router.get(
+          url.toString(),
+          {},
+          {
+            preserveState: true,
+            preserveScroll: true,
+          },
+        );
       } else {
-        router.get(url.toString(), {}, {
-          preserveState: true,
-          preserveScroll: true,
-          only: ["search_results"],
-        });
+        router.get(
+          url.toString(),
+          {},
+          {
+            preserveState: true,
+            preserveScroll: true,
+            only: ["search_results"],
+          },
+        );
       }
     }
 


app/javascript/components/Discover/Search.tsx
```diff
@@ -25,10 +25,14 @@ export const Search = ({ query, setQuery }: { query?: string | undefined; setQue
     router.reload({
       only: ["autocomplete_results"],
       data: { query: enteredQuery },
+      replace: true,
```
EmCousin, line 28: Avoid polluting the browser history when loading autocomplete results.

app/javascript/components/Discover/Search.tsx
```diff
@@ -25,10 +25,14 @@ export const Search = ({ query, setQuery }: { query?: string | undefined; setQue
...
     });
   }, 300);
 
-  useOnChange(() => fetchAutocomplete(), [enteredQuery]);
+  useOnChange(() => {
+    if (enteredQuery) fetchAutocomplete();
+    else fetchAutocomplete.cancel();
+  }, [enteredQuery]);
```
EmCousin, line 35: Avoid needlessly hitting the server if there is no entered query.

slavingia:
Was this the PR that broke the build? https://github.com/antiwork/gumroad/pull/3663

EmCousin:
@slavingia yes it was. Will work with @nyomanjyotisa on this 👍

EmCousin:
@slavingia actually it wasn't. [This PR](https://github.com/antiwork/gumroad/pull/3517) should have introduced it but CI was green, then next commit on main was [this PR merge](https://github.com/antiwork/gumroad/pull/3559), in which the test was failing, although the changes had apparently nothing to do with it. I had mistakenly taken it for a flaky test.

app/javascript/pages/Discover/Index.tsx
```diff
@@ -278,16 +278,24 @@ function DiscoverIndex() {
           parseUrlParams(window.location.href, props.curated_product_ids, defaultSortOrder).offer_code;
 
       if (shouldFetchRecommendations) {
-        router.visit(url.toString(), {
-          preserveState: true,
-          preserveScroll: true,
-        });
+        router.get(
```
binary-koan, line 281: Is there a reason for using `router.get` over `router.visit`?

EmCousin: Matter of style, really. You can pass any verb to `visit` to make it a get, post, put, delete, etc. request, but it's `get` by default. I feel like using `router.get` is clearer, you understand at a glance what request is being done without having to look in the visit options.

binary-koan: Sounds good, that's what Claude Code guessed 😉 (`visit` is a weirdly named method to me then, it doesn't sound like the same thing)

Issue #3704, opened 2026-02-19
Archived products page redirects to /products when search matches no results


4arjun:
## Summary
Searching on the archived products page with a query that doesn't match any archived product redirects to `/products` instead of staying on the archived tab.

## Steps to reproduce
1. Archive at least one product
2. Navigate to the archived products tab (`/products/archived`)
3. Type a search query that doesn't match any archived product name
4. Observe the page redirects to `/products`

## Expected behavior
The page stays on the archived tab and shows an empty state placeholder.

## Actual behavior
The page redirects to `/products`.

https://github.com/user-attachments/assets/c878bd0e-7c7b-455b-b01c-e58a3e149185

PR #3705, opened 2026-02-19
Fix archived products search redirect and add empty state

4arjun:
Issue: #3704 

## Summary
Searching on the archived products page with a query that matches no products redirected to `/products` instead of staying on the archived tab.

## Root cause

`DashboardProductsPagePresenter#empty?` evaluated `products_table_props[:products].empty? && memberships_table_props[:memberships].empty?`, which checks the query-filtered results. When a search query matched nothing, `empty?` returned true, and the controller's `return redirect_to products_url` if `archived_products_page_presenter.empty?` triggered a redirect, even though archived products do exist.

Changed `empty?` to use `!seller.products.visible.archived.exists?`, a single `EXISTS SQL` query that checks whether any archived products exist at all, independent of the search query. This keeps the redirect semantics correct (only redirect when genuinely no archived products exist) while also being more efficient.

Also added a Placeholder empty state UI to the archived products page for when search returns no matching results.

## Before

https://github.com/user-attachments/assets/c3a4cc2c-8c8f-40d9-9381-d1f1b92a0c62

## After

https://github.com/user-attachments/assets/5546b153-eb39-49e0-b2c4-5aefefdedc7a

## Tests
`bundle exec rspec spec/controllers/products/archived_controller_spec.rb`

Screenshot 2026-02-20 at 12 02 32 AM

# Checklist

---

# AI Disclosure
No AI used.

### DIFF
--- a/spec/controllers/products/archived_controller_spec.rb
+++ b/spec/controllers/products/archived_controller_spec.rb
@@ -59,6 +59,17 @@ describe Products::ArchivedController, inertia: true do
         expect(response).to redirect_to(products_url)
       end
     end
+
+    context "when a search query matches no archived products" do
+      it "renders the archived page with empty results" do
+        get :index, params: { query: "nonexistent" }
+
+        expect(response).to have_http_status(:ok)
+        expect(inertia).to render_component("Products/Archived/Index")
+        expect(inertia.props[:products_data][:products]).to be_empty
+        expect(inertia.props[:memberships_data][:memberships]).to be_empty
+      end
+    end
   end
 
   describe "POST create" do


app/presenters/dashboard_products_page_presenter.rb
```diff
@@ -23,7 +23,7 @@ def initialize(pundit_user:, archived: false, products_page: 1, memberships_page
 
   def empty?
     return nil unless archived?
-    products_table_props[:products].empty? && memberships_table_props[:memberships].empty?
+    !seller.products.visible.archived.exists?
```
4arjun, line 26: The previous implementation checked `products_table_props[:products].empty? && memberships_table_props[:memberships].empty?`, which evaluates the query-filtered and paginated results. When a search query matched nothing, this returned true, causing the controller to redirect away from the archived page.

Now it does a simple `exists?` check on the database, which is:

- Query-independent:doesn't matter what search terms are active
- More efficient:a single EXISTS SQL query instead of loading, paginating, and serializing all products just to check if the list is empty
- Semantically correct:`empty?`now answers "are there any archived products at all?" rather than "are there any archived products matching the current filters?"

The controller stays clean with just return `redirect_to products_url` if `archived_products_page_presenter.empty?`.

app/javascript/components/ArchivedProductsPage.tsx
```diff
@@ -46,16 +50,24 @@ export const ArchivedProductsPage = ({
...
-          productsSort={productsSort}
-          query={query}
-          type="archived"
-        />
+        {hasNoResults ? (
+          <Placeholder>
+            <PlaceholderImage src={placeholder} />
+            <h2>No products found</h2>
+            {query ? <p>No archived products matched your search.</p> : null}
+          </Placeholder>
```
4arjun, line 58: Added an empty state placeholder for when no products or memberships match the search query.

slavingia:
No end-to-end tests.

PR #3706, opened 2026-02-19
Remove processing state for audio files on the mobile web view

binary-koan:
Issue: https://github.com/antiwork/gumroad-ios/issues/13

# Description

## Problem

A creator's audio file cannot be played on iOS because the duration hasn't been filled in. However, it works fine on web and nothing appears to prevent it from playing fine on mobile as well if we just enable the button.

## Solution

Remove the `isProcessing` state on the mobile file list as this doesn't appear to affect whether the file can be played back.

---

# AI Disclosure

Claude Opus 4.6 used for identifying the issue and checking if the logic is required.

### DIFF
--- a/app/javascript/components/Download/FileList.tsx
+++ b/app/javascript/components/Download/FileList.tsx
@@ -472,14 +472,13 @@ const MobileAppAudioFileRow = ({ file }: { file: FileItem }) => {
 
   const isCompleted =
     latestMediaLocation && latestMediaLocation > 0 && file.duration && latestMediaLocation >= file.duration;
-  const isProcessing = file.duration === null;
 
   return (
     <Row ref={selfRef} className="embed" {...touchAndHoldEventListeners}>
       <WithTooltip tip={showTooltip ? file.file_name : null} position="top">
         <TrackClick
           eventName="play_click"
-          file={isProcessing || showTooltip ? null : file} // Prevent playback when processing or showing tooltip
+          file={showTooltip ? null : file}
           type="audio"
           isPlaying={isPlaying}
           resumeAt={latestMediaLocation || 0}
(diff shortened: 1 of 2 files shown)


binary-koan:
@nyomanjyotisa _a Could you review?

nyomanjyotisa (commented):
@binary-koan The fix makes sense as a quick unblock, but I think we should look into why the the extract duration stuck/failed?

This "Processing..." state intentionally added in PR https://github.com/antiwork/gumroad-old/pull/27611

Without the duration value, couple of things will be missing on mobile:
- Duration label
- Progress bar
- "Xm Ys left" remaining time
- Completed checkmark icon

I think we can trigger the extract duration script manually for this audio, and see the actual error. Probably something with audio encoding, and we can inform creator to update the audio file with processable audio?

Like we did for the PDF stamp, where we automatically email the creator if stamping failed, to remove password or strip encryption on their PDF.

Or a quick fix, we can download the audio file, and see the actual duration, and manually fill the duration field for this audio on production

binary-koan:
I did some more digging, and discovered this is because the file is larger than our analyze limit (1GB) so we just skip the analysis here: https://github.com/antiwork/gumroad/blob/5723afea4fb9b0ee9cc6be8dda7652b7131c2ecd/app/modules/with_file_properties.rb#L53

I think in this case it's reasonable to make the file playable without `duration` - it's pretty rare and as far as I can tell the mobile audio player will infer the duration from the file contents anyway, so the webview display won't be perfect but playback will still work.

We could look at increasing the download limit, but we'll still run into the issue eventually - there are 55 wav files over 1GB in the last million files, including 35 over 2GB (for example if we double the limit).

@ershad Do you think we can/should increase the analysis file size limit here?

slavingia:
IMO we should just process all. Not many.

binary-koan:
@ershad Do you foresee any issues if we change the regular file size limit to the same as the video one (1GB -> 40B) here? https://github.com/antiwork/gumroad/blob/fa5f7d8010205262b7e038493d5a7b29be2eadbf/app/modules/with_file_properties.rb#L6-L7

(still better to have a limit than nothing IMO, but that should be way above what people upload)

ershad:
@binary-koan The only concern would be the temporary use of disk space and possibly running out of space if we process multiple files at the same time on the same machine. But if we have a 40GB limit for video, it makes sense to set that limit globally for all file types 👍🏼

Issue #3702, opened 2026-02-19
Copy-pasting rich text content in product editor is broken


EmCousin:
When I edit a product's content and add rich text like embed files, ctrl+A doesn't seem to properly select them. Copy-pasting them to another content editor does not work, see the recording below.

https://github.com/user-attachments/assets/85e26bec-ef72-45fe-8a38-acff72c8408b

PR #3707, opened 2026-02-19
Migrate edit product page to Inertia

EmCousin:
Issue: #3030 

# Description

This PR migrates the Edit product page to Inertia by splitting it into separate tab-specific endpoints. Each tab now has its own controller with dedicated `edit` and `update` actions and a corresponding Inertia page. The `edit` actions return only the props relevant to that tab, and the `update` actions accept only the params needed for that tab's fields. This also breaks up the monolithic `links#update` action by distributing the update logic across each tab's controller, allowing for better modularization, maintenance, and testing.

---

# Before/After

## Before

https://github.com/user-attachments/assets/072d0421-0b48-49bf-bdaa-0ad6faf3b286

## After

https://github.com/user-attachments/assets/46d2ad5f-6cdb-4cf9-a4fb-d0ae3e2efa16

---

# Checklist

---

# AI Disclosure

Claude Code + Claude Opus for the migration split, manual modifications, then Claude Opus again for self-review


app/controllers/concerns/inertia_rendering.rb
```diff
@@ -21,8 +21,9 @@ module InertiaRendering
   private
     def inertia_flash_props
       return if (flash_message = flash[:alert] || flash[:warning] || flash[:notice]).blank?
-
-      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      flash_props = { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      flash_props[:html] = flash[:html].present? if flash[:html].present?
+      flash_props
```
EmCousin, line 26: We can now trigger HTML-based flash messages from the backend

app/controllers/products/archived_controller.rb
```diff
@@ -15,21 +15,22 @@ def index
...
+      products_data: InertiaRails.defer(group: "data") do
         {
           products: archived_products_page_presenter.products_table_props[:products],
           pagination: archived_products_page_presenter.products_table_props[:products_pagination],
           sort: archived_products_page_presenter.products_sort,
         }
-      },
-      memberships_data: -> {
+      end,
+      memberships_data: InertiaRails.defer(group: "data") do
```
EmCousin, line 27: Landing on the Products page in order to create a new product, I found out that we were loading `products_data` and `memberships_data` twice. Once during the initial page load, the second time after the component's first rendering. We fix it here by deferring the props properly so the initial load is faster.

app/controllers/products/base_controller.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class Products::BaseController < Sellers::BaseController
```
EmCousin, line 3: This base controller factors what is content to all 4 edit product tabs on the backend.

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+class Products::ContentController < Products::BaseController
+  def edit
+    render inertia: "Products/Content/Edit", props: Products::ContentTabPresenter.new(product: @product, pundit_user:).props
```
EmCousin, line 5: The content tab has a dedicated presenter with just the props it needs. It is covered with tests.

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def update
+    authorize @product
+
+    should_publish = params[:publish].present? && !@product.published?
+    should_unpublish = params[:unpublish].present? && @product.published?
+
+    if should_unpublish
+      return unpublish_and_redirect_to(edit_product_content_path(@product.unique_permalink))
+    end
```
EmCousin, line 14: The logic Publishing / Unpublishing a product was previously handled in the frontend and was subjected to tight coupling, resulting in complex conditions. It is now handled in the backend and decoupled in each tab controller, reducing the number of network requests needed to update and un/publishi a product

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    def rich_content_params
+      rich_content = product_permitted_params[:rich_content] || []
+      rich_content_params = [*rich_content]
+      product_permitted_params[:variants]&.each { rich_content_params.push(*_1[:rich_content]) }
+      rich_content_params.flat_map { _1.dig(:description, :content) }
+    end
+
+    def product_permitted_params
+      @product_permitted_params ||= params.require(:product).permit(policy(@product).content_tab_permitted_attributes)
```
EmCousin, line 98: We permit only the attributes we need for the content part of the product

app/controllers/products/product_controller.rb
```diff
@@ -0,0 +1,240 @@
+# frozen_string_literal: true
+
+class Products::ProductController < Products::BaseController
+  def edit
+    return redirect_to bundle_path(@product.external_id) if @product.is_bundle?
+
+    render inertia: "Products/Product/Edit", props: Products::ProductTabPresenter.new(product: @product, pundit_user:, ai_generated: params[:ai_generated].present?).props
```
EmCousin, line 7: The product tab has a dedicated presenter with just the props it needs. It is covered with tests.

app/controllers/products/product_controller.rb
```diff
@@ -0,0 +1,240 @@
...
+    def update_removed_file_attributes
+      return unless product_permitted_params[:file_attributes].present?
+      current = @product.file_info_for_product_page.keys.map(&:to_s)
+      updated = (product_permitted_params[:file_attributes] || []).map { _1[:name].to_s }
+      removed = current - updated
+      @product.add_removed_file_info_attributes(removed) if removed.any?
+    end
+
+    def product_permitted_params
+      params.require(:product).permit(policy(@product).product_tab_permitted_attributes)
```
EmCousin, line 229: We permit only the attributes we need for this tab

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -1,6 +1,6 @@
 import { DirectUpload } from "@rails/activestorage";
```
EmCousin, line 1: The changes in this file fix an bug when copy/pasting rich content in the product editor. It is documented here: https://github.com/antiwork/gumroad/issues/3702

EmCousin (commented):
Self-review

app/controllers/products/content_controller.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+class Products::ContentController < Products::BaseController
+  def edit
+    render inertia: "Products/Content/Edit", props: Products::ContentTabPresenter.new(product: @product, pundit_user:).props
+  end
+
+  def update
+    authorize @product
```
EmCousin: `BaseController` already runs `before_action :authorize_product` on all actions, which calls `authorize @product` (Pundit infers `update?` from the action name). This explicit `authorize @product` call is redundant and can be removed. Same applies to `share_controller.rb:11`.

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def update_share_attributes
+      @product.assign_attributes(product_permitted_params.except(:tags, :section_ids, :custom_domain))
+      @product.save_tags!(product_permitted_params[:tags] || [])
+      @product.show_in_sections!(product_permitted_params[:section_ids] || [])
+      update_custom_domain if product_permitted_params.key?(:custom_domain)
+      @product.save!
+    end
+
+    def update_custom_domain
```
EmCousin: This `update_custom_domain` implementation has diverged from the one in `ProductController` (line 214). The `ProductController` version has improvements: `.strip.presence` to normalize whitespace, and a `!= new_domain` check to skip re-verification if the domain hasn't changed. This version is missing both. We should either extract this into `BaseController` or at least align the two.

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  def edit
+    render inertia: "Products/Share/Edit", props: Products::ShareTabPresenter.new(product: @product, pundit_user:).props
+  end
+
+  def update
+    authorize @product
+
+    ActiveRecord::Base.transaction do
+      update_share_attributes
+      @product.unpublish! if params[:unpublish].present? && @product.published?
```
EmCousin: The other tab controllers (`ProductController`, `ContentController`, `ReceiptController`) use the shared `unpublish_and_redirect_to` helper for early-return unpublish. This controller handles unpublish inline inside the transaction and redirects after. This means `check_offer_codes_validity` runs regardless of whether the unpublish was the only action, which is fine, but the inconsistency with other controllers is worth noting.

app/controllers/links_controller.rb
```diff
@@ -301,173 +296,46 @@ def track_user_action
...
-  end
-
-  def unpublish
-    authorize @product
-
-    @product.unpublish!
-    render json: { success: true }
-  end
-
   def publish
```
EmCousin, line 299: This `publish` action still has the full `request.inertia?` branching for both JSON and Inertia responses. Since publish is now handled by each tab controller's `update` action with `params[:publish]`, we may be able to remove the Inertia branches here — unless there's another Inertia caller I'm missing.

app/controllers/links_controller.rb
```diff
@@ -490,6 +358,7 @@ def update_sections
   def release_preorder
     authorize @product
 
+
```
EmCousin: Stray blank line.

app/controllers/products/receipt_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    elsif permitted_redirect_path
+      redirect_to permitted_redirect_path, notice: "Changes saved!", status: :see_other
+    else
+      redirect_to edit_product_receipt_path(@product.unique_permalink), notice: "Changes saved!", status: :see_other
+    end
+  end
+
+  private
+    def update_receipt_attributes
+      @product.assign_attributes(product_permitted_params.except(:custom_domain))
```
EmCousin: `update_receipt_attributes` does `.except(:custom_domain)` but `receipt_tab_permitted_attributes` in the policy doesn't include `:custom_domain`. This `except` is a no-op — harmless but unnecessary.

app/controllers/concerns/inertia_rendering.rb
```diff
@@ -21,8 +21,9 @@ module InertiaRendering
   private
     def inertia_flash_props
       return if (flash_message = flash[:alert] || flash[:warning] || flash[:notice]).blank?
-
-      { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      flash_props = { message: flash_message, status: flash[:alert] ? "danger" : flash[:warning] ? "warning" : "success" }
+      flash_props[:html] = flash[:html].present? if flash[:html].present?
```
EmCousin: `flash_props[:html] = flash[:html].present? if flash[:html].present?` — since the guard ensures `flash[:html]` is present, the assigned value is always `true`. Could be simplified to `flash_props[:html] = true if flash[:html].present?`.

app/controllers/products/product_controller.rb
```diff
@@ -0,0 +1,240 @@
...
+        :shipping_destinations,
+        :community_chat_enabled,
+        :custom_domain,
+        :file_attributes,
+        :refund_policy,
+        :product_refund_policy_enabled
+      ))
+      @product.skus_enabled = false
+
+      if @product.native_type === ::Link::NATIVE_TYPE_COFFEE && product_permitted_params[:variants].present?
```
EmCousin: Using `===` (case equality) instead of `==`. In Ruby, `===` on strings behaves the same as `==`, but `==` is more idiomatic and consistent with the rest of the codebase (e.g., `content_controller.rb` uses `==`).

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -732,8 +754,37 @@ export const FileEmbed = TiptapNode.create<{ getConfig?: () => FileEmbedConfig }
                 if (file?.url) node.setAttribute("url", file.url);
               }
             });
+            const parsed = ProseMirrorDOMParser.fromSchema(editor.schema).parseSlice(fragment);
+            return new Slice(parsed.content, slice.openStart, slice.openEnd);
+          },
+          transformPasted(slice) {
```
EmCousin, line 759: This is the fix for pasting file embeds across products. It resolves files by matching on `id` or `url` from `existingFiles`, remaps the ID, and calls `onPasteFiles` to add them to the product's file list before the editor renders them. This replaces the old approach that tried to process nodes post-insertion. The `Slice` import and proper `openStart`/`openEnd` preservation on the re-parsed slice was key — without it, nested content structure was lost during paste.

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -245,8 +277,13 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
       }
       return fileEntry;
     });
+    // Update product state with new files, then insert embeds after React commits the update.
+    // Using queueMicrotask ensures the state update is committed before TipTap tries to render
+    // the file embeds, so filesById will include the newly uploaded files.
     updateProduct({ files: [...product.files, ...fileEntries] });
-    onSelectFiles(fileEntries.map((file) => file.id));
+    queueMicrotask(() => {
```
EmCousin, line 284: `updateProduct` enqueues a React state update, but `onSelectFiles` (which inserts TipTap nodes) needs `filesById` to already include the new files for the node views to render properly. `queueMicrotask` defers the insertion to after the state update is committed. This is a timing workaround — if React batching behavior changes, this could break. An alternative would be a `useEffect` that watches for new files, but that would be more complex.

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -55,7 +55,13 @@ export const getDraggedFileEmbed = (editor: Editor) => {
...
   if (!node.attrs.id) return;
 
-  const { id, updateProduct, filesById } = useProductEditContext();
+  const { id } = useProductEditContext();
+  const { product, updateProduct } = useProductFormContext();
+  // Compute filesById from form state to include newly uploaded files
+  // Note: Don't use useMemo here - upload progress callbacks mutate fileStatus objects directly
+  // and call updateProduct({}) to trigger re-renders. Using useMemo([product.files]) would not
+  // recompute the Map since the array reference stays the same, causing progress to not update.
+  const filesById = new Map(product.files.map((f) => [f.id, f]));
```
EmCousin, line 64: The `filesById` Map is computed identically here and in `ContentTabContent` (index.tsx:309) with the same "don't use useMemo" comment. This is intentional since they operate in different React component scopes, but it's worth noting the duplication in case we find a way to unify it later.

app/presenters/products/base_presenter.rb
```diff
@@ -0,0 +1,141 @@
...
+      seller: UserPresenter.new(user: product.user).author_byline_props,
+      ai_generated:,
+      currency_type: product.price_currency_type,
+      earliest_membership_price_change_date: ::BaseVariant::MINIMUM_DAYS_TIL_EXISTING_MEMBERSHIP_PRICE_CHANGE.days.from_now.in_time_zone(product.user.timezone).iso8601,
+      available_countries: ::ShippingDestination::Destinations.shipping_countries.map { { code: _1[0], name: _1[1] } },
+      google_client_id: ::GlobalConfig.get("GOOGLE_CLIENT_ID"),
+      google_calendar_enabled: Feature.active?(:google_calendar_link, product.user),
+      cancellation_discounts_enabled: Feature.active?(:cancellation_discounts, product.user),
+      thumbnail: product.thumbnail&.alive&.as_json,
+      refund_policies: product.user.product_refund_policies.for_visible_and_not_archived_products.where.not(product_id: product.id).order(updated_at: :desc).select("refund_policies.*", "links.name").as_json,
```
EmCousin, line 31: This query joins through products to filter visible/non-archived ones and is returned in `layout_props` which is called on every tab load. If a seller has many products with refund policies, this could be slow. Not a regression (the old presenter did the same), but worth keeping an eye on.

EmCousin:
@claude review this PR.

EmCousin:
@binary-koan _a this PR is a bit tough to review, the tight coupling made it virtually not possible to split into multiple PRs.

I've self-reviewed a bunch of times and tested every scenario I could think of, but an extra pair of eyes is most welcome 🙏

app/controllers/links_controller.rb
```diff
@@ -95,20 +96,14 @@ def create
...
     end
 
     create_user_event("add_product")
-    if ai_generated
-      redirect_to edit_link_path(@product, ai_generated: true), status: :see_other
-    else
-      redirect_to edit_link_path(@product), status: :see_other
-    end
+    options = ai_generated ? { ai_generated: true } : {}
+    redirect_to edit_product_product_path(@product, **options), status: :see_other
```
binary-koan, line 106: This path is a bit odd but it's consistent with the UI tabs so fine for now - we can clean it up when we rework product editing

app/controllers/concerns/inertia_rendering.rb
binary-koan: Sending HTML makes me sad, but it's no worse than what we were doing already

app/controllers/products/archived_controller.rb
```diff
@@ -15,21 +15,22 @@ def index
     set_meta_tag(title: "Archived products")
 
     render inertia: "Products/Archived/Index", props: {
+      has_products: -> { archived_products_page_presenter.page_props[:has_products] },
       can_create_product: -> { archived_products_page_presenter.page_props[:can_create_product] },
-      products_data: -> {
+      products_data: InertiaRails.defer(group: "data") do
```
binary-koan, line 20: This can be a small separate PR

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class Products::ShareController < Products::BaseController
```
binary-koan, line 3: I'd start with a share page PR as it's the simplest

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  before_action :ensure_published_for_share, only: [:edit]
```
binary-koan, line 4: I'd just inline this into `edit` personally

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def update
+    authorize @product
+
+    should_unpublish = params[:unpublish].present? && @product.published?
+
+    if should_unpublish
+      ActiveRecord::Base.transaction do
+        update_share_attributes
```
binary-koan, line 17: This is strange to me - the button is just called "unpublish" but it also saves changes? IMO we should just have a common unpublish action rather than adding complexity to `update`

app/controllers/products/share_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+      return unpublish_and_redirect_to(edit_product_content_path(@product.unique_permalink))
+    end
+
+    ActiveRecord::Base.transaction do
+      update_share_attributes
+    end
+
+    check_offer_codes_validity
```
binary-koan, line 26:
```suggestion
    check_offer_codes_are_still_valid
```

would be a little clearer to me. Also not sure if you can change anything on this page that affects offer codes?

app/javascript/components/server-components/ProductEditPage.tsx
binary-koan, line 1: While I remember, we can also delete `buildStaticRouter` and the `react-router` dependency after this right?

binary-koan (commented):
Added a few initial comments. I think we can at least migrate tab by tab and just deal with server reloads in the meantime

Alternatively (or as well) we could split up the controller but keep the existing server components first - `LinksController#update` is such a big method that it basically deserves one or more PRs to itself 😓

EmCousin:
Closing - we'll break this down into multiple PRs.

PR #3710, opened 2026-02-19
Add global sales tax summary report job

### LINKED ISSUE #3669: Expand monthly sales tax summary report to cover all tax-collecting countries
## What

Gumroad collects sales tax in 60+ countries but only produces a monthly summary report for US states. The accounting team needs a single monthly CSV that summarizes **GMV, order count, and tax collected** across **all** tax-collecting jurisdictions – grouped by country, with state/province breakdown where applicable.

## Why

The existing `CreateUsStatesSalesSummaryReportJob` was built exclusively for US states – it uses zip code lookups, US subdivision mapping, and TaxJar transaction pushing. There's no equivalent summary for international jurisdictions. Individual per-country reports exist (India monthly, Canada monthly, EU quarterly VAT) but they each have different formats, run on different schedules, and serve different purposes. There's no single view of tax collection across all jurisdictions.

## Proposed solution

Create a new `CreateGlobalSalesTaxSummaryReportJob` that generates a single monthly CSV covering all countries where Gumroad collects tax.

**Report format** (confirmed by accounting):

| Column | Description |
|---|---|
| Country | Country name (Column A per accounting's request) |
| State/Province | State/province code where applicable, blank otherwise |
| GMV | Total transaction value in USD, net of refunds |
| Number of orders | Count of successful, non-refunded, non-chargedback purchases |
| Sales tax collected | Total `gumroad_tax_cents` in USD, net of refunds |

**Grouping rules:**
- **US**: Break down by state (use existing zip code / GeoIP logic from the US summary job)
- **Canada**: Break down by province (use existing logic from `CreateCanadaMonthlySalesReportJob`)
- **India**: Break down by state (use existing `ip_state` logic from `CreateIndiaSalesReportJob`)
- **All other countries**: Country-level only (single row per country)

**Implementation steps:**

1. **Query purchases with tax collected** for the previous month: filter on `gumroad_tax_cents > 0`, successful, non-refunded, non-chargedback, with a Stripe transaction ID. This naturally captures all jurisdictions without needing to enumerate them. Note: this scopes GMV and order counts to **purchases where tax was actually collected**, not all purchases in tax-collecting jurisdictions. This is intentional – the report is a tax collection summary, not a full GMV report.

2. **Determine country from purchase data**: normalize the `country` field (handle variants like "South Korea" vs "Korea, Republic of", "Czechia" vs "Czech Republic"). Fall back to `ip_country` when `country` is nil, consistent with existing report jobs.

3. **Determine state/province for US, CA, IN** – reuse the existing detection patterns: zip code / GeoIP lookup for US, `state` field for Canada (the existing Canada report filters on `purchase.state` matching known CA subdivisions), `ip_state` for India.

4. **Aggregate** GMV (net of refunds), order count, and tax collected per country+state group.

5. **Output CSV** to S3 and **email it to @steveo-es and `salestax@gumroad.com`** – matching how existing monthly financial reports are delivered. The job runs automatically on the 1st of each month for the prior calendar month.

6. **Wire into `GenerateFinancialReportsForPreviousMonthJob`** so it runs automatically on the 1st of each month alongside existing reports.

**What this does NOT do:**
- Push transactions to TaxJar (TaxJar is US/CA-specific; keep that in the existing US job)
- Replace existing per-country reports (India, Canada, VAT) – those serve different purposes (transaction-level detail, VAT rate breakdowns, TaxJar compliance)
- Backfill past months (not needed per accounting)

## Acceptance criteria

- [ ] New job generates a CSV with columns: Country, State/Province, GMV, Number of orders, Sales tax collected
- [ ] US purchases are broken down by state
- [ ] Canada purchases are broken down by province
- [ ] India purchases are broken down by state
- [ ] All other countries appear as a single row per country
- [ ] All amounts are in USD
- [ ] Country names are normalized (no duplicate rows for "South Korea" / "Korea, Republic of", "Czechia" / "Czech Republic", etc.)
- [ ] CSV is uploaded to S3 and emailed to @steveo-es and `salestax@gumroad.com`
- [ ] Job runs automatically on the 1st of each month via `GenerateFinancialReportsForPreviousMonthJob`
- [ ] GMV is net of refunds and credits
- [ ] Existing US summary report and other per-country reports continue to work unchanged

## Edge cases

- **Country name inconsistencies**: The `country` field has variants ("Czech Republic" vs "Czechia", "Korea, Republic of" vs "South Korea"). Normalize to a canonical name – ideally map through ISO country codes rather than relying on raw strings.
- **Null country**: Some purchases have `country = NULL`. Fall back to `ip_country` as existing jobs do. If both are null, group under an "Unknown" row so nothing is silently dropped.
- **Large result set performance**: With ~294k purchases/month having tax, the query needs to be efficient. Consider using `find_each` or batching rather than loading all purchases at once. The existing US job loads purchase IDs first, then iterates – a similar pattern works here.
- **State/province detection failures**: For US/CA/IN, some purchases may not resolve to a state (bad zip, missing IP). Include an empty state/province value and still count them under the country total.

gianfrancopiana:
Fixes #3669
Fixes antiwork/gumroad-private/issues/164

## Problem

Gumroad collects sales tax in 60+ countries but only has a monthly summary report for US states. The accounting team needs a single monthly CSV summarizing GMV, order count, and tax collected across all tax-collecting jurisdictions – grouped by country, with state/province breakdown for US, Canada, and India.

## Approach

New `CreateGlobalSalesTaxSummaryReportJob` that finds all purchases where we collected tax for a given month, groups them by country and state/province, and totals up GMV, order count, and tax collected.

Country names are normalized so variants like "Korea, Republic of" and "South Korea" don't produce duplicate rows.

State/province breakdown uses the same logic as the existing US, Canada, and India report jobs. The CSV is uploaded to S3 and emailed to the relevant people.

Runs automatically on the 1st of each month alongside existing reports. Does not push to TaxJar or replace any existing per-country reports.

## Performance optimization

The job processes ~293k purchases per month. Three optimizations eliminate the main bottlenecks:

1. **Pre-aggregated refund sums** – one upfront query loads refund totals for partially-refunded purchases into a hash, replacing per-record `refunds.sum()` calls (N+1 elimination)
2. **Selective column loading** – only the 10 columns used in the loop are selected instead of full AR objects
3. **Memoized country name normalization** – `ISO3166::Country.find_country_by_any_name` is called ~50 times (once per distinct country) instead of ~293k times

Verified against production: SQL and Ruby approaches produce identical grand totals (597,503,039 GMV cents / 65,120,420 tax cents / 293,027 orders).

---

### DIFF
--- a/app/sidekiq/create_global_sales_tax_summary_report_job.rb
+++ b/app/sidekiq/create_global_sales_tax_summary_report_job.rb
@@ -15,21 +15,32 @@ class CreateGlobalSalesTaxSummaryReportJob
 
     timeout_seconds = ($redis.get(RedisKey.create_global_sales_tax_summary_report_job_max_execution_time_seconds) || 1.hour).to_i
     WithMaxExecutionTime.timeout_queries(seconds: timeout_seconds) do
-      Purchase.successful
+      purchases_scope = Purchase.successful
         .not_fully_refunded
         .not_chargedback_or_chargedback_reversed
         .where.not(stripe_transaction_id: nil)
         .where("gumroad_tax_cents > 0")
         .where("purchases.created_at BETWEEN ? AND ?", start_date, end_date)
         .where(charge_processor_id: [nil, *ChargeProcessor.charge_processor_ids])
+
+      refund_sums = Refund
+        .where(purchase_id: purchases_scope.where(stripe_partially_refunded: true).select(:id))
+        .group(:purchase_id)
+        .pluck(:purchase_id, Arel.sql("SUM(refunds.total_transaction_cents)"), Arel.sql("SUM(refunds.gumroad_tax_cents)"))
+        .to_h { |pid, total, tax| [pid, { total: total.to_i, tax: tax.to_i }] }
+
+      purchases_scope
+        .select(:id, :country, :ip_country, :state, :ip_state, :zip_code, :ip_address,
+                :total_transaction_cents, :gumroad_tax_cents, :stripe_partially_refunded)
         .find_each do |purchase|
         country_name = resolve_country_name(purchase)
         state_code = resolve_state(purchase, country_name)
+        refund = refund_sums[purchase.id] if purchase.stripe_partially_refunded?
 
-        key = [country_name, state_code]
-        aggregation[key][:gmv_cents] += purchase.total_cents_net_of_refunds
-        aggregation[key][:order_count] += 1
-        aggregation[key][:tax_collected_cents] += purchase.gumroad_tax_cents_net_of_refunds
+        bucket = aggregation[[country_name, state_code]]
+        bucket[:gmv_cents] += net_cents(purchase.total_transaction_cents, refund&.dig(:total))
+        bucket[:order_count] += 1
+        bucket[:tax_collected_cents] += net_cents(purchase.gumroad_tax_cents, refund&.dig(:tax))
       end
     end
 
@@ -39,10 +50,7 @@ class CreateGlobalSalesTaxSummaryReportJob
       temp_file = Tempfile.new
       temp_file.write(row_headers.to_csv)
 
-      aggregation.keys.sort.each do |key|
-        country_name, state_code = key
-        data = aggregation[key]
-
+      aggregation.sort.each do |(country_name, state_code), data|
         temp_file.write([
           country_name,
           state_code,
@@ -73,8 +81,14 @@ class CreateGlobalSalesTaxSummaryReportJob
       raw_name = purchase.country.presence || purchase.ip_country.presence
       return "Unknown" if raw_name.blank?
 
-      country = Compliance::Countries.find_by_name(raw_name)
-      country&.common_name || raw_name
+      normalized_country_names[raw_name]
+    end
+
+    def normalized_country_names
+      @normalized_country_names ||= Hash.new do |hash, raw_name|
+        country = Compliance::Countries.find_by_name(raw_name)
+        hash[raw_name] = country&.common_name || raw_name
+      end
     end
 
     def resolve_state(purchase, country_name)
@@ -119,6 +133,14 @@ class CreateGlobalSalesTaxSummaryReportJob
         raw_state
       else
         ""
+      end
+    end
+
+    def net_cents(gross_cents, refunded_cents)
+      if refunded_cents
+        [gross_cents - refunded_cents, 0].max
+      else
+        gross_cents
       end
     end
 end


app/sidekiq/create_global_sales_tax_summary_report_job.rb
```diff
@@ -0,0 +1,146 @@
...
+                :total_transaction_cents, :gumroad_tax_cents, :stripe_partially_refunded)
+        .find_each do |purchase|
+        country_name = resolve_country_name(purchase)
+        state_code = resolve_state(purchase, country_name)
+        refund = refund_sums[purchase.id] if purchase.stripe_partially_refunded?
+
+        bucket = aggregation[[country_name, state_code]]
+        bucket[:gmv_cents] += net_cents(purchase.total_transaction_cents, refund&.dig(:total))
+        bucket[:order_count] += 1
+        bucket[:tax_collected_cents] += net_cents(purchase.gumroad_tax_cents, refund&.dig(:tax))
```
ershad, line 43: @hchhabra Could you please verify this math? Should we also subtract chargeback amounts separately here?

app/sidekiq/create_global_sales_tax_summary_report_job.rb
```diff
@@ -0,0 +1,146 @@
...
+      elsif purchase.ip_state.present? && purchase.ip_state.in?(valid_canada_provinces)
+        purchase.ip_state
+      else
+        ""
+      end
+    end
+
+    def resolve_india_state(purchase)
+      raw_state = (purchase.ip_state || "").strip.upcase
+      if !raw_state.match?(/^\d+$/) && CreateIndiaSalesReportJob::VALID_INDIAN_STATES.include?(raw_state)
```
ershad, line 132: It would be good to keep `VALID_INDIAN_STATES` globally in an initializer.

spec/sidekiq/create_global_sales_tax_summary_report_job_spec.rb
```diff
@@ -0,0 +1,312 @@
...
+
+        described_class.new.perform(month, year)
+
+        actual_payload = read_csv_from_s3
+
+        de_row = actual_payload.find { |row| row[0] == "Germany" }
+        expect(de_row).to be_present
+
+        gmv = BigDecimal(de_row[2].delete(","))
+        expect(gmv).to be < BigDecimal("119.00")
```
ershad, line 195: Instead of `<`, could we compare against the actual number?

app/sidekiq/create_global_sales_tax_summary_report_job.rb
hchhabra: Fully refunded and chargedback purchases are excluded in the scope above so we only need to deduct partial refunds here, this should be good I think.

PR #3734, opened 2026-02-20
Remove negative seller revenue purchase block

### LINKED ISSUE #3036: Currency price minimums don't account for fees

slavingia:
Fixes #3036

Alternative to #3455 — takes a simpler approach by removing the block entirely rather than capping fees.

## Summary

- Remove the `validate_seller_revenue` check that blocks purchases where fees exceed the product price
- Remove the `negative_revenue_sale_failure` email, mailer method, template, preview, and specs
- Keep the `PurchaseErrorCode::NET_NEGATIVE_SELLER_REVENUE` constant for historical reasons (existing DB records reference it)

## Why this approach

The current behavior blocks ~50 purchases/month with a confusing "product is not correctly set up" error, earning Gumroad $0 on each. Rather than adding fee-capping logic (as in #3455), this simply removes the gate — letting these purchases proceed as-is. The seller may net negative on a handful of very-low-price transactions, but Gumroad and the buyer both benefit from the sale going through.

## Test plan

- [x] Verify `PurchaseErrorCode::NET_NEGATIVE_SELLER_REVENUE` constant still exists in `purchase_error_code.rb`
- [x] Existing purchase process spec updated to assert purchase is no longer blocked
- [x] Mailer specs, preview, and template removed cleanly
- [x] Run full test suite to confirm no regressions

### DIFF
--- a/spec/models/purchase/purchase_process_spec.rb
+++ b/spec/models/purchase/purchase_process_spec.rb
@@ -512,7 +512,7 @@ describe "Purchase Process", :vcr do
         affiliate_purchase.process!
       end
 
-      it "no longer blocks the purchase" do
+      it "does not block the purchase" do
         expect(affiliate_purchase.error_code).to be_nil
         expect(affiliate_purchase.errors).to be_empty
       end
--- a/spec/support/fixtures/vcr_cassettes/Purchase_Process/_process_/purchase_with_negative_seller_revenue/does_not_block_the_purchase.yml
+++ b/spec/support/fixtures/vcr_cassettes/Purchase_Process/_process_/purchase_with_negative_seller_revenue/does_not_block_the_purchase.yml
@@ -1,0 +1,622 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://api.stripe.com/v1/payment_methods
+    body:
+      encoding: UTF-8
+      string: type=card&card[token]=tok_visa
+    headers:
+      User-Agent:
+      - Stripe/v1 RubyBindings/12.5.0
+      Authorization:
+      - Bearer <STRIPE_API_KEY>
+      Content-Type:
+      - application/x-www-form-urlencoded
+      Idempotency-Key:
+      - c1067277-0e4d-478e-98c5-532d7374492e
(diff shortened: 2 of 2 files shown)


slavingia:
@hchhabra close but the test around affiliates is failing. could you investigate for us, please? thank you!

PR #3739, opened 2026-02-20
Products index/archived: use deferred props for initial load

EmCousin:
Issue: Part of #3707 (1/6)

# Description

## Problem

The Products Index and Archived pages load all table data (products, memberships, pagination) synchronously on initial render, which slows down initial page load. The page components also wrap an unnecessary layout component (`ProductsDashboardPage` / `ArchivedProductsPage`) that duplicates logic already handled by the Inertia layout.

## Solution

- Use `InertiaRails.defer(group: "data")` in both `LinksController#index` and `Products::ArchivedController#index` to defer heavy table data loading
- Inline the layout logic directly into the Inertia page components (`Products/Index.tsx`, `Products/Archived/Index.tsx`) using `<Deferred>` from `@inertiajs/react`
- Extract shared components: `ContentLoading`, `LayoutCtaButton`, `ProductsPageProps`
- Delete the now-unused wrapper components (`ProductsDashboardPage.tsx`, `ArchivedProductsPage.tsx`)
- Add `query` as a public reader on `DashboardProductsPagePresenter` and expose `has_products` for the archived page
- Use Inertia `<Link>` for edit URLs in `ProductsTable`

---

# Before/After

## Before

https://github.com/user-attachments/assets/3ea7d750-e99c-4adc-af5f-f72cac0d25ac

## After

https://github.com/user-attachments/assets/c726a77f-8c24-48be-a8c5-186d599e3860

---

# Test Results

```
bundle exec rspec spec/controllers/products/archived_controller_spec.rb spec/presenters/dashboard_products_page_presenter_spec.rb spec/requests/products/index_spec.rb
111 examples, 0 failures
```

---

# Checklist

---

# AI Disclosure

AI (Claude) was used to assist with splitting the original PR #3707 into incremental PRs and writing the PR description.

### DIFF
--- a/app/controllers/products/archived_controller.rb
+++ b/app/controllers/products/archived_controller.rb
@@ -14,25 +14,7 @@ class Products::ArchivedController < Sellers::BaseController
 
     set_meta_tag(title: "Archived products")
 
-    render inertia: "Products/Archived/Index", props: {
-      query: archived_products_page_presenter.query,
-      has_products: -> { archived_products_page_presenter.page_props[:has_products] },
-      can_create_product: -> { archived_products_page_presenter.page_props[:can_create_product] },
-      products_data: InertiaRails.defer(group: "data") do
-        {
-          products: archived_products_page_presenter.products_table_props[:products],
-          pagination: archived_products_page_presenter.products_table_props[:products_pagination],
-          sort: archived_products_page_presenter.products_sort,
-        }
-      end,
-      memberships_data: InertiaRails.defer(group: "data") do
-        {
-          memberships: archived_products_page_presenter.memberships_table_props[:memberships],
-          pagination: archived_products_page_presenter.memberships_table_props[:memberships_pagination],
-          sort: archived_products_page_presenter.memberships_sort,
-        }
-      end,
-    }
+    render inertia: "Products/Archived/Index", props: archived_products_page_presenter.page_props
   end
 
   def create
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -39,26 +39,7 @@ class LinksController < ApplicationController
 
     set_meta_tag(title: "Products")
 
-    render inertia: "Products/Index", props: {
-      query: products_page_presenter.query,
-      has_products: -> { products_page_presenter.page_props[:has_products] },
-      archived_products_count: -> { products_page_presenter.page_props[:archived_products_count] },
-      can_create_product: -> { products_page_presenter.page_props[:can_create_product] },
-      products_data: InertiaRails.defer(group: "data") do
-        {
-          products: products_page_presenter.products_table_props[:products],
-          pagination: products_page_presenter.products_table_props[:products_pagination],
-          sort: products_page_presenter.products_sort,
-        }
-      end,
-      memberships_data: InertiaRails.defer(group: "data") do
-        {
-          memberships: products_page_presenter.memberships_table_props[:memberships],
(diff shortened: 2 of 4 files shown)


app/controllers/products/archived_controller.rb
```diff
@@ -15,21 +15,22 @@ def index
...
-      memberships_data: -> {
+      end,
+      memberships_data: InertiaRails.defer(group: "data") do
         {
           memberships: archived_products_page_presenter.memberships_table_props[:memberships],
           pagination: archived_products_page_presenter.memberships_table_props[:memberships_pagination],
           sort: archived_products_page_presenter.memberships_sort,
         }
-      },
+      end,
```
EmCousin: Before, those props were evaluated for the initial page load, then they were requested again. The PR defers those props properly so the initial page load is faster and those props come after. Same for the archived products in the other controller below.

app/javascript/components/ProductsPage/ContentLoading.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Skeleton } from "$app/components/Skeleton";
+
+export const ProductsContentLoading = () => (
+  <section className="space-y-4 p-4 md:p-8">
+    <Skeleton className="h-12 w-full" />
+    <Skeleton className="h-12 w-full" />
+    <Skeleton className="h-12 w-full" />
+    <Skeleton className="h-12 w-full" />
+    <Skeleton className="h-12 w-full" />
+  </section>
```
EmCousin, line 12: Simple skeleton loader while waiting for deferred props to show up.

EmCousin:
@binary-koan _a may I ask you to review?

app/controllers/products/archived_controller.rb
```diff
@@ -15,21 +15,22 @@ def index
     set_meta_tag(title: "Archived products")
 
     render inertia: "Products/Archived/Index", props: {
+      has_products: -> { archived_products_page_presenter.page_props[:has_products] },
```
binary-koan: Do we need `query` here too?

In general - since we're using a presenter anyway - could we just make a method to return this structure so the two controllers just need to call `archived_products_page_presenter.page_props` or something?

app/javascript/components/ProductsPage/LayoutCtaButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  query: string;
+  setQuery: (query: string) => void;
+};
+
+type PageProps = {
+  has_products: boolean;
+  can_create_product: boolean;
+};
+
+export const LayoutCtaButton = ({ query, setQuery }: Props) => {
```
binary-koan: `HeaderButtons` maybe? Although the prop is called `ctaButton` I wouldn't call these CTAs

binary-koan:
I agree with tidying up the components (would actually be nice to tidy up the presenters/controllers while we're at it) but I'm not clear on why we should defer the props? That loading bar and delayed tab navigation follows the same pattern as most other pages, and should make it easy to try SSR when the migration is complete

EmCousin:
The props were already "deferred" pre-migration as they were requested via a JSON endpoint after the initial rendering. Post migration, they were loaded at the initial Inertia visit, then loaded again in a hook. It's a mistake the PR is fixing by bringing back the original behaviour. Does it make sense?

EmCousin:
Thanks @binary-koan, I have addressed your latest comments and Ci is green. Merging now.

PR #3756, opened 2026-02-23
Reset scroll to top on Inertia navigation

nyomanjyotisa:
Resolves https://github.com/antiwork/gumroad-private/issues/154

## Problem

Navigating between pages left the scroll position wherever it was on the previous page.

## Solution

Add `scroll-region` to the scrollable `<main>` containers in both the default layout and the admin layout so that Inertia resets the scroll position to the top on every page navigation, and it preserve the scroll position on go back to previous page.

### Before

https://github.com/user-attachments/assets/21273102-65d0-420b-b51a-66cddc9e5d42

### After

https://github.com/user-attachments/assets/80e7e4e3-6a14-4f5f-acfb-0468701fd8d9

### DIFF
--- a/app/javascript/inertia/layout.tsx
+++ b/app/javascript/inertia/layout.tsx
@@ -39,7 +39,9 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         <Alert initial={null} />
         <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
-          <main scroll-region="" className="flex-1 overflow-y-auto">{children}</main>
+          <main scroll-region="" className="flex-1 overflow-y-auto">
+            {children}
+          </main>
         </div>
       </CurrentSellerProvider>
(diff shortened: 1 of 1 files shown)


app/javascript/inertia/layout.tsx
```diff
@@ -39,7 +39,7 @@ export default function Layout({ children }: { children: React.ReactNode }) {
         <Alert initial={null} />
         <div id="inertia-shell" className="flex h-screen flex-col lg:flex-row">
           {logged_in_user ? <Nav title="Dashboard" /> : null}
-          <main className="flex-1 overflow-y-auto">{children}</main>
+          <main scroll-region="" className="flex-1 overflow-y-auto">{children}</main>
```
nyomanjyotisa: In TSX, `<main scroll-region>` is shorthand for `scroll-region={true}`, React silently removes unknown attributes with boolean values from the DOM. Using `scroll-region=""` passes a string instead, which React always keeps, so Inertia can actually find the element

PR #3793, opened 2026-02-23
Add support SLA message to ticket modals

gianfrancopiana:
Issue antiwork/gumroad-private#120

## Problem

Users submitting support tickets have no indication of when they'll hear back. We want to set expectations upfront using real data from the board meeting (97.7% of requests get a first response within an hour).

## Approach

Added a subtitle message below the modal title in both ticket modals (authenticated and unauthenticated): "You'll most likely hear back within an hour — 97.7% of people do. We guarantee a response within 24 hours."

The copy is extracted into a shared `SupportSlaMessage` component since both modals display the same text and the number will eventually update dynamically via a Helper API endpoint.

### Before
Desktop:

Mobile:

### After
Desktop:
CleanShot 2026-02-23 at 17 24 01@2x

Mobile:

---

### DIFF
--- a/app/javascript/components/support/SupportSlaMessage.tsx
+++ b/app/javascript/components/support/SupportSlaMessage.tsx
@@ -3,8 +3,7 @@ import * as React from "react";
 export function SupportSlaMessage({ children }: { children?: React.ReactNode }): React.ReactNode {
   return (
     <>
-      You'll most likely hear back within an hour — 97.7% of people do. We guarantee a response within 24
-      hours.
+      You'll most likely hear back within an hour — 97.7% of people do. We guarantee a response within 24 hours.
       {children}
     </>
   );


app/javascript/components/support/NewTicketModal.tsx
```diff
@@ -32,6 +33,29 @@ export function NewTicketModal({
   const formRef = React.useRef<HTMLFormElement | null>(null);
   const fileInputRef = React.useRef<HTMLInputElement | null>(null);
 
+  const isFormValid = subject.trim() && message.trim();
+
+  const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
```
gianfrancopiana, line 38: This is change is consistent with what [UnauthenticatedNewTicketModal.tsx](https://github.com/antiwork/gumroad/blob/b506c8e408f3f412bfd5a31d8ac071b74e97641a/app/javascript/components/support/UnauthenticatedNewTicketModal.tsx#L31) already does.

PR #3800, opened 2026-02-24
Add the fx loss amount caused by a returned payout to the earliest unpaid balance date

hchhabra:
Fixes https://github.com/antiwork/gumroad-private/issues/123#issuecomment-3893710740

### What

Add the fx loss/gain amount caused by a returned payout to the earliest unpaid balance date for the creator, and not the date on which when the payout is returned.

### Why

When Stripe is unable to process a payout and it's returned back, the returned amount is slightly different from the originally paid out amount in case of non-USD payouts, because of fx difference. That fx gain/loss is adjusted against creator's balance so that the unpaid balance on our end matches the amount in their Stripe connect account. That fx difference should be adjusted in the unpaid balance of earliest date possible, because the Stripe connect account balance has already been adjusted by Stripe as soon as the payout is returned. But currently it gets adjusted on the date of payout return in case there's no existing unpaid balance against that merchant account.


app/models/balance_transaction.rb
```diff
@@ -186,6 +186,8 @@ def find_or_create_balance
           refund.created_at.to_date
         elsif dispute
           dispute.formalized_at.to_date
+        elsif credit&.returned_payment
+          credit.returned_payment.balances.order(date: :asc).first&.date || credit.created_at.to_date
```
hchhabra, line 190: The `date` of the new `balance` should be same as the date of the earliest balance included in the returned payout, so that this new balance immediately becomes payable and gets included in any payout that gets initiated from this point on. If the `date` of the new balance is set as `credit.created_at` (which would be same as `Date.current` here, the new `balance` will not get included in weekly payouts for next 7 days. And since the corresponding Stripe connect account balance has already been adjusted by Stripe at this point, there will be a temporary balance mismatch which would cause payout failures for the next 7 days till this balance becomes eligible for payout. 
Also, a payout should ideally always have a balance, but kept `credit.created_at.to_date` fallback for safety.

spec/models/balance_transaction_spec.rb
```diff
@@ -1325,6 +1325,82 @@
...
+
+          describe "when no unpaid balance against the merchant account exists" do
+            it "creates a new balance for the date of the earliest balance in the returned payment" do
+              expect(user.unpaid_balances.where(merchant_account: credit.merchant_account)).to be_empty
+
+              expect(balance_transaction.balance).to eq(user.unpaid_balances.where(merchant_account: credit.merchant_account).last)
+              expect(user.unpaid_balances.where(merchant_account: credit.merchant_account).last).to be_present
+              expect(balance_transaction.balance.user).to eq(user)
+              expect(balance_transaction.balance.merchant_account).to eq(merchant_account)
+              expect(balance_transaction.balance.date).to eq(6.days.ago.to_date)
```
hchhabra, line 1398: This `balance.date` here is same as `returned_payment.balances.order(date: :asc).first.date`, which is `6.days.ago`.

app/models/balance_transaction.rb
gianfrancopiana: The fallback path (`credit.created_at.to_date` when `returned_payment.balances` is empty) isn't covered – worth adding a test for it?

PR #3802, opened 2026-02-24
Fix the unclaimed balance collection specs failing on main

hchhabra:
### What

Fix the unclaimed balance collection specs failing on `main`.

### Why

Fix the specs failing on `main` (failed on a development branch [here](https://github.com/antiwork/gumroad/actions/runs/22353185744/job/64691531410?pr=3800), will start failing on `main` CI with the next commit).

### DIFF
--- a/spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb
+++ b/spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb
@@ -154,7 +154,9 @@ describe CollectUnclaimedBalancesOfInactiveStripeAccountsJob do
       expect(Stripe::Balance).not_to receive(:retrieve)
       expect(Stripe::Transfer).not_to receive(:create)
 
-      CollectUnclaimedBalancesOfInactiveStripeAccountsJob.new.perform
+      travel_to(Date.new(2026, 2, 11)) do
+        CollectUnclaimedBalancesOfInactiveStripeAccountsJob.new.perform
+      end
     end
 
     it "does not attempt to collect balance if the Stripe account is considered active based on last payout date" do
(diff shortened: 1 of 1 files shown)


spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb
```diff
@@ -154,7 +154,9 @@
       expect(Stripe::Balance).not_to receive(:retrieve)
       expect(Stripe::Transfer).not_to receive(:create)
 
-      CollectUnclaimedBalancesOfInactiveStripeAccountsJob.new.perform
+      travel_to(Date.new(2026, 2, 11)) do
```
hchhabra, line 157: The spec fails because [this check](https://github.com/antiwork/gumroad/blob/main/app/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job.rb#L25) (`next if stripe_account&.created.to_i > (Time.current - STRIPE_ACCOUNT_INACTIVE_AFTER_DURATION).to_i`) is not true anymore. Since `stripe_account&.created` time is fixed in the recorded VCR cassette, using `travel_to` to fix `Time.current` here so that the check remains true. `STRIPE_ACCOUNT_INACTIVE_AFTER_DURATION` is stubbed as 3 weeks.

Same change in the below 2 specs.

gianfrancopiana: Since `created_at: 2.months.ago` is evaluated at real time but `perform` runs at frozen time (Feb 11), these tests will break again around March 21 when `2.months.ago` crosses the Jan 21 threshold.

Could we wrap the entire test body in `travel_to` so the factory dates are also frozen?

```ruby
it "does not attempt to collect balance if the Stripe account is considered active based on creation date" do
  travel_to(Date.new(2026, 2, 11)) do
    us_stripe_account = create(:merchant_account, country: "US", currency: "usd",
                               charge_processor_merchant_id: "acct_1SwhnzEqM4HpFFlW",
                               created_at: 2.months.ago)
    create(:balance, user: us_stripe_account.user, merchant_account: us_stripe_account, amount_cents: 100_00)

    expect(Stripe::Account).to receive(:retrieve).and_call_original
    # ... other expectations ...
    CollectUnclaimedBalancesOfInactiveStripeAccountsJob.new.perform
  end
end
```

Same for the other two tests.

gianfrancopiana (changes requested):

PR #3813, opened 2026-02-25
Help center articles missing meta description tags

### LINKED ISSUE #3795: Help center articles missing meta description tags
## What

Help center article pages (`/help/<slug>`) don't have `<meta name="description">`, `<meta property="og:description">`, or `<meta name="twitter:description">` tags. The `/help` index page has all three, but individual articles have none.

This means search engines generate their own snippets instead of using the curated descriptions we already write for each article. Social previews (Slack, Twitter, etc.) also show no description. All 120+ articles are affected.

- [ ] Add `description`, `og:description`, and `twitter:description` meta tags to individual article pages
- [ ] Use the existing `@description` value each article partial already defines

## Why

Introduced as an omission in PR #3019 (SSR Meta tags), which migrated meta tags from instance variables to `set_meta_tag` calls. The `/help` page got description tags, but individual article pages didn't.

gianfrancopiana:
Fixes #3795

## Problem

All 123 help center article pages are missing `description`, `og:description`, and `twitter:description` meta tags. The `index` action sets these, but `show` never did. Each article partial had an `@description` instance variable that was set but never actually read anywhere — it was dead code.

## Approach

Migrated the `@description` string from each of the 123 article partials into the YAML data file (`articles.yml`) as a first-class `description` attribute. This makes `article.description` available on the model via ActiveYaml. The `show` action now sets all three description meta tags using the same pattern as `index`. The unused `@description` lines were removed from all partials.

---


EmCousin (approved):
Looks good, the failing tests are unrelated, but they're not flaky, something else broke them. Will investigate separately!

PR #3815, opened 2026-02-25
Improve editor performance when there are too many file nodes

binary-koan:
Ref #2042 

## Problem

The product in the issue has thousands of files in a flat list, which cripples performance on the product edit page. From experimenting, it seems the main issue is simply rendering that many DOM nodes with React.

## Solution

Added an IntersectionObserver which mimics virtualized scrolling by hiding file nodes (with a placeholder element of the same size) which are >1000px off the screen.

This is by no means perfect - there are other bottlenecks like fetching and parsing the JSON - but I think it's a rare enough edge case that we don't need to go much beyond "it's possible to load and use the editor"

## Screenshots

This is with 5000 file nodes, which is pushing the limits of my pretty powerful MacBook - other devices would likely run into issues at lower numbers.

### Before

https://github.com/user-attachments/assets/7e088eb5-3c2d-469e-bf14-f626dd95dd62

(it loads after a couple of minutes, but you get the idea)

### After

https://github.com/user-attachments/assets/36e5536a-443f-4cb7-a6c5-d1394ba51e80

(still slow, but it does load and is usable)

### DIFF
--- a/app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
@@ -178,7 +178,35 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
   const isConnectedRow = isInGroup && !hasStreamable;
   const isLastInGroup = node === parentNode?.content.lastChild;
 
-  if (!fileExists) return;
+  if (!fileExists) return <NodeViewWrapper ref={ref} contentEditable={false} />;
+
+  const updateFile = (data: Partial<FileEntry>) =>
+    updateProduct((product) => {
+      product.files = product.files.map((existing) => (existing.id === file.id ? { ...existing, ...data } : existing));
+    });
+
+  const uploadThumbnail = (thumbnail: File) => {
+    if (thumbnail.size > 5 * 1024 * 1024)
+      return showAlert(
+        "Could not process your thumbnail, please upload an image with size smaller than 5 MB.",
+        "error",
+      );
+
+    setLoadingVideo(true);
+    const upload = new DirectUpload(thumbnail, Routes.rails_direct_uploads_path());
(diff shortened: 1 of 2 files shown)


app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -175,10 +179,16 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
...
+      <NodeViewWrapper ref={ref} contentEditable={false}>
+        <div style={{ height: lastHeight.current }} />
+      </NodeViewWrapper>
+    );
+
   const updateFile = (data: Partial<FileEntry>) =>
     updateProduct((product) => {
-      const existing = product.files.find((existing) => existing.id === file.id);
-      if (existing) Object.assign(existing, data);
+      product.files = product.files.map((existing) => (existing.id === file.id ? { ...existing, ...data } : existing));
```
binary-koan: Required due to the `filesById` memoization change - `updateProduct` updates the state using a shallow clone of the object, so if we update an object in `product.files` without changing the array itself then `filesById` will no longer be updated and we won't re-render the file node

app/javascript/components/ProductEdit/ContentTab/useNodeVisibility.ts
```diff
@@ -0,0 +1,68 @@
...
+  const observe = React.useContext(NodeVisibilityContext);
+
+  React.useEffect(() => {
+    const element = ref.current;
+    if (!element || !observe) return;
+
+    return observe(element, (isIntersecting) => {
+      if (!isIntersecting) lastHeight.current = element.offsetHeight;
+      setVisible(isIntersecting);
+    });
```
binary-koan: This is the core change - we now track whether file nodes are on screen so we can skip rendering if they're too far outside the viewport

Issue #3801, opened 2026-02-24
Fix top 5 flaky tests blocking auto production deployment


ershad:
## What

4 out of the last 10 failed main branch CI builds have actual test failures (the other 6 fail only due to the "Claude Code Review" job, not tests). The flaky tests are blocking the automated production deployment pipeline.

### Updated flaky test analysis (Feb 25, 2026)

Based on the last 10 failed CI builds on main, **4 builds had test failures**:

| Failure Rate | Test | Description | Builds |
|---|---|---|---|
| **4/4 (100%)** | `spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb:147` | does not attempt to collect balance if the Stripe account is considered active based on creation date | [1](https://github.com/antiwork/gumroad/actions/runs/22403514953) [2](https://github.com/antiwork/gumroad/actions/runs/22403495470) [3](https://github.com/antiwork/gumroad/actions/runs/22402326609) [4](https://github.com/antiwork/gumroad/actions/runs/22390567363) |
| **4/4 (100%)** | `spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb:160` | does not attempt to collect balance if the Stripe account is considered active based on last payout date | [1](https://github.com/antiwork/gumroad/actions/runs/22403514953) [2](https://github.com/antiwork/gumroad/actions/runs/22403495470) [3](https://github.com/antiwork/gumroad/actions/runs/22402326609) [4](https://github.com/antiwork/gumroad/actions/runs/22390567363) |
| **4/4 (100%)** | `spec/sidekiq/collect_unclaimed_balances_of_inactive_stripe_accounts_job_spec.rb:173` | does not attempt to collect balance if the Stripe account is considered active based on last charge date | [1](https://github.com/antiwork/gumroad/actions/runs/22403514953) [2](https://github.com/antiwork/gumroad/actions/runs/22403495470) [3](https://github.com/antiwork/gumroad/actions/runs/22402326609) [4](https://github.com/antiwork/gumroad/actions/runs/22390567363) |
| **1/4 (25%)** | `spec/requests/settings/payments_spec.rb:1855` | AE business allows the creator to enter their business vat number | [1](https://github.com/antiwork/gumroad/actions/runs/22403514953) |
[...]

PR #3820, opened 2026-02-26
Replace flaky Stripe iframe assertions with stable verification checks

gianfrancopiana:
Issue #3801 (partial – addresses Test 4: payments_spec.rb)

## Problem

Four tests in `payments_spec.rb` assert the presence of a Stripe Connect iframe (`connect-js.stripe.com`), which depends on Stripe's CDN loading inside CI Docker runners. This causes ~10% flake rate, blocking production deployment.

## Approach

Replace all 4 iframe selector assertions with stable, DOM-based verification-state checks. Instead of waiting for an external iframe to load, we check that the "Verification" section exists and the success message ("Your identity has been verified!") doesn't appear, confirming the verification banner shows up without relying on Stripe's CDN.

---

### DIFF
--- a/spec/requests/settings/payments_spec.rb
+++ b/spec/requests/settings/payments_spec.rb
@@ -540,7 +540,7 @@ describe("Payments Settings Scenario", type: :system, js: true) do
         expect(@user.active_bank_account.account_holder_full_name).to eq("Gumhead Moneybags")
       end
 
-      it "displays the Stripe Connect embedded verification banner" do
+      it "shows the verification section when identity verification is needed" do
         user = create(:user, username: nil, payment_address: nil)
         create(:user_compliance_info, user:, birthday: Date.new(1901, 1, 2))
         create(:ach_account_stripe_succeed, user:)


gianfrancopiana:
> If the external iframe doesn't load, the workflow cannot complete so the conformation won't show up, right? Doesn't this PR merely delay the problem from happening?

The old iframe selector verified both that our app decided verification is needed AND that Stripe Connect JS loaded. The new assertions only verify the first part. But the tests never interacted with the iframe – they just checked it existed and stopped, so there was no end-to-end workflow coverage to begin with. And when the iframe check fails in CI (~10% flake rate), it's usually an external dependency issue, not actionable app code.

If we want to close the gap, a frontend unit test for `StripeConnectEmbeddedNotificationBanner` (mocking the Stripe SDK) would be more reliable than waiting for a real iframe in a system test. Thoughts?

I've renamed the spec title at line 543 to "shows the verification section when identity verification is needed" to match what it actually asserts now.

PR #3822, opened 2026-02-26
Remove flaky pre-submit tax UI assertions from WA tax tests

gianfrancopiana:
Issue #3801 (partial fix)

## Problem

Two WA sales tax tests (`taxes_spec.rb:226` and `:266`) flake ~25% in CI. They assert pre-submit checkout UI text ("Sales tax US$10.35") inside the `check_out` block, which runs before clicking "Pay." When the async tax surcharge API call hasn't completed within Capybara's wait timeout, the assertion fails.

## Approach

Remove the pre-submit UI assertions from both `check_out` calls. Both tests already have deterministic post-purchase assertions that verify the exact same tax amounts from the `Purchase` record:

```ruby
purchase = Purchase.last
expect(purchase.total_transaction_cents).to eq(110_35)
expect(purchase.gumroad_tax_cents).to eq(10_35)
```

The second test (non-physical WA) had a `wait: 10` band-aid that masked the same race condition. Removed that too.

---

### DIFF
--- a/app/sidekiq/create_global_sales_tax_summary_report_job.rb
+++ b/app/sidekiq/create_global_sales_tax_summary_report_job.rb
@@ -129,11 +129,7 @@ class CreateGlobalSalesTaxSummaryReportJob
 
     def resolve_india_state(purchase)
       raw_state = (purchase.ip_state || "").strip.upcase
-      if !raw_state.match?(/^\d+$/) && CreateIndiaSalesReportJob::VALID_INDIAN_STATES.include?(raw_state)
-        raw_state
-      else
-        ""
-      end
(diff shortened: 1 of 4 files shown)


spec/requests/purchases/product/taxes_spec.rb
EmCousin, line 3988: This PR was not trying to fix this flaky test right? Only the two above?

EmCousin (approved):

spec/requests/purchases/product/taxes_spec.rb
gianfrancopiana: Correct!

PR #3825, opened 2026-02-26
Optimize sales tax summary report with SQL GROUP BY

gianfrancopiana:
## Problem

The global sales tax summary job (#3710) loads ~293k purchases one by one. Both the Jan and Feb 2026 runs timed out after 73+ minutes — no reports were generated or emailed.

## Approach

Move the aggregation into SQL. A single `GROUP BY` query reduces 293k rows to ~58k grouped results, then Ruby handles country normalization and state resolution on the smaller set. The ~12 US purchases with bad zip codes still fall back to GeoIp lookup individually.

Grouping uses byte-exact keys to avoid MySQL collation silently merging values like `"USA"` and `"usa"`. Country/state normalization logic is – just operates on query results instead of AR objects.

## Production validation (read replica, Jan 2026)

| Metric | Before | After |
|--------|--------|-------|
| Runtime | 73+ min (timed out) | **~30s** |
| Rows iterated in Ruby | 293,009 | 57,851 |

---

### DIFF
--- a/app/javascript/components/ActivityFeed.tsx
+++ b/app/javascript/components/ActivityFeed.tsx
@@ -1,8 +1,8 @@
+import { Dollar, UserCircle } from "@boxicons/react";
 import * as React from "react";
 
 import { formatPriceCentsWithCurrencySymbol } from "$app/utils/currency";
 
-import { Icon } from "$app/components/Icons";
 import { useLoggedInUser } from "$app/components/LoggedInUser";
 import { Card, CardContent } from "$app/components/ui/Card";
 import { Placeholder } from "$app/components/ui/Placeholder";
@@ -23,7 +23,7 @@ export type ActivityItem =
 
 const Sale = ({ details: { price_cents, product_name, product_unique_permalink } }: { details: SaleItemDetails }) => (
   <>
-    <Icon name="outline-currency-dollar" className="text-green" />
+    <Dollar className="size-5 text-green" />
     <span>
       New sale of <a href={Routes.short_link_path({ id: product_unique_permalink })}>{product_name}</a> for{" "}
       {formatPriceCentsWithCurrencySymbol("usd", price_cents, { symbolFormat: "short", noCentsIfWhole: true })}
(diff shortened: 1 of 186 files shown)


app/sidekiq/create_global_sales_tax_summary_report_job.rb
```diff
@@ -23,32 +31,109 @@ def perform(month, year)
...
+                       ""
+        end
+
+        bucket = aggregation[[country_name, state_code]]
+        bucket[:gmv_cents] += gmv.to_i
+        bucket[:order_count] += count.to_i
+        bucket[:tax_collected_cents] += tax.to_i
+      end
+
+      unresolved_us_tuple_keys.uniq!
```
EmCousin: Isn't this unique already given this coms from aggregated results grouped by those very same tuple keys? Might be a no-op here?

app/sidekiq/create_global_sales_tax_summary_report_job.rb
```diff
@@ -23,32 +31,109 @@ def perform(month, year)
...
+
+    write_and_upload_csv(aggregation, month, year)
+  end
+
+  private
+    def aggregation_query_rows(purchases_scope)
+      key_sqls = BINARY_SAFE_KEY_COLUMNS.values
+
+      purchases_scope
+        .joins(<<~SQL)
```
EmCousin, line 78: Is this syntax valid? Or is it a trailing parenthesis that happens to be harmless?
```suggestion
        .joins(<<~SQL.squish
```

EmCousin (approved):
Minor comments. Also the code is a bit difficult to reason about due to the necessary optimizations, it might be worth adding comments as documentation (and/or reference this issue) so the next developer better understand why we had to do this.

Other than that, looks like a great optimization 💪

app/sidekiq/create_global_sales_tax_summary_report_job.rb
gianfrancopiana: Good catch! It's a no-op since the rows come from a GROUP BY on those exact tuple keys. Removed it in 6a82cc2.

app/sidekiq/create_global_sales_tax_summary_report_job.rb
gianfrancopiana: The trailing `)` is intentional – it closes the `.joins(` call. Just looks a bit unusual with heredocs.

gianfrancopiana:
> Minor comments. Also the code is a bit difficult to reason about due to the necessary optimizations, it might be worth adding comments as documentation (and/or reference this issue) so the next developer better understand why we had to do this.
> 
> Other than that, looks like a great optimization 💪

Agreed, added comments in 6a82cc2 explaining why the binary-safe grouping exists and why the GeoIp fallback second pass is needed.

PR #2269, opened 2025-12-10
fix(product): Optimize file lookups

4arjun:
ref #2042 

## Summary
Adds `useFilesById` hook that creates a memoized Map for O(1) file lookups
Updates `FileEmbed` and `useFilesInGroup` to use Map lookups instead of array.find()

## Problem

Each file embed was calling `product.files.find()` which is O(n). For a product with 300 files and 300 file embeds, this resulted in ~90,000 comparisons on every render, blocking the main thread.

## Solution

Created a memoized Map that builds once when files change, reducing lookups from O(n) to O(1). For 300 files, this reduces operations from ~90,000 to ~600 (150x improvement).

## Result

https://github.com/user-attachments/assets/6ac56bc4-9cd2-4619-9026-cbe85b999920

## Test

Screenshot 2025-12-10 at 12 58 33 PM

## AI Usage
Used GPT-5.1 (Cursor) to help understand potential causes of the issue and explore possible fixes.
## Confirmation
I have watched all the live streaming videos.

## Self-Review
I have self-reviewed all the code that I have written.

PR #3826, opened 2026-02-26
Fix file mapping memoization to improve editor performance

binary-koan:
Ref #2042 

## Problem

#2269 added a `Map` of files to the product edit page state so that file nodes could use `filesById.get()` rather than `files.find(...)`. This dramatically improves performance when there are thousands of files in the product content.

This map was supposed to be memoized and only rebuilt when necessary, but at some point it ended up inline in the context value, which is recomputed on every render. This reverses much of the performance improvement.

## Solution

Made the map memoized separately and only depend on `product.files`. This means it's only rebuilt when editing a file, rather than multiple times during any page interaction.

## AI Disclosure

Claude Opus 4.6 used to do the refactor and fix failing tests.


app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -177,8 +177,7 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
   if (!fileExists) return;
   const updateFile = (data: Partial<FileEntry>) =>
     updateProduct((product) => {
-      const existing = product.files.find((existing) => existing.id === file.id);
-      if (existing) Object.assign(existing, data);
+      product.files = product.files.map((existing) => (existing.id === file.id ? { ...existing, ...data } : existing));
```
binary-koan, line 180: Here and below, we now need to explicitly make `product.files` a new array when we edit file info. Previously we would just update the file object in place, which means the array reference would stay the same and therefore the `product.files` dependency in the `filesById` memo wouldn't change.

Perhaps as part of the product edit refactor we can look at Immer or something as a nicer way to do nested immutable state updates.

binary-koan:
@nyomanjyotisa _a Could you review this one too?

nyomanjyotisa (approved):

PR #3827, opened 2026-02-26
Virtualized scrolling for file embeds in content editor

binary-koan:
Ref #2042 

## Problem

The product in the issue has thousands of files in a flat list, which cripples performance on the product edit page. From experimenting, it seems the main issue is simply rendering that many DOM nodes with React.

## Solution

Added an IntersectionObserver which mimics virtualized scrolling by hiding file nodes (with a placeholder element of the same size) which are >1000px off the screen.

This is by no means perfect - there are other bottlenecks like fetching and parsing the JSON - but I think it's a rare enough edge case that we don't need to go much beyond "it's possible to load and use the editor"

## Screenshots

This is with 5000 file nodes, which is pushing the limits of my pretty powerful MacBook - other devices would likely run into issues at lower numbers.

### Before

https://github.com/user-attachments/assets/7e088eb5-3c2d-469e-bf14-f626dd95dd62

(it loads after a couple of minutes, but you get the idea)

### After

https://github.com/user-attachments/assets/36e5536a-443f-4cb7-a6c5-d1394ba51e80

(still slow, but it does load and is usable)


binary-koan:

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -57,12 +59,14 @@ export const getDraggedFileEmbed = (editor: Editor) => {
   return draggedNode?.type.name === FileEmbed.name ? draggedNode : null;
 };
 
+const DEFAULT_FILE_ROW_HEIGHT = 82;
```
binary-koan, line 62: AFAIK there isn't a better way of computing this than just looking at what the size of an average file row currently is. We don't want to render and then check the height because the whole point is to skip rendering when possible

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -174,12 +178,43 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
   const isConnectedRow = isInGroup && !hasStreamable;
   const isLastInGroup = node === parentNode?.content.lastChild;
 
-  if (!fileExists) return;
+  if (!fileExists) return <NodeViewWrapper ref={ref} contentEditable={false} />;
```
binary-koan, line 181: We need to return an element now so that the ref exists, otherwise we may not set up the intersection observer properly

app/javascript/components/ProductEdit/ContentTab/FileEmbed.tsx
```diff
@@ -174,12 +178,43 @@ const FileEmbedNodeView = ({ node, editor, getPos, updateAttributes }: NodeViewP
...
+
   const updateFile = (data: Partial<FileEntry>) =>
     updateProduct((product) => {
       const existing = product.files.find((existing) => existing.id === file.id);
       if (existing) Object.assign(existing, data);
     });
+
+  const uploadThumbnail = (thumbnail: File) => {
```
binary-koan, line 188: Moved above the early return since they're used in upload logic which triggers even if the file is off screen

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -776,6 +778,7 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
           />
         ) : null}
         <PageListLayout
+          ref={scrollContainerRef}
```
binary-koan, line 808: We need to attach the intersection observer to the scrollable container node rather than the document body, otherwise `rootMargin` doesn't work correctly

app/javascript/components/ProductEdit/ContentTab/useNodeVisibility.ts
```diff
@@ -0,0 +1,68 @@
...
+    callbacks.set(element, callback);
+    observer.observe(element);
+
+    return () => {
+      callbacks.delete(element);
+      observer.unobserve(element);
+    };
+  }, []);
+
+  return React.createElement(NodeVisibilityContext.Provider, { value: observe }, children);
```
nyomanjyotisa: Nit: consider renaming to .tsx and remove the `React.createElement` ? like we did in `useRecaptcha.tsx`

binary-koan: Nice catch, updated

PR #3829, opened 2026-02-27
Skip Doorkeeper authorize prompt for mobile OAuth app

binary-koan:
From @pegasus1134 

Ref https://github.com/antiwork/gumroad-mobile/issues/24

## Problem

The mobile app login flow is awkward because after logging in you still need to accept an "authorize" prompt from Doorkeeper - a bunch of irrelevant page content to read and an extra click.

## Solution

* Bypass the OAuth consent screen for the first-party mobile app by adding a skip_authorization block that matches the mobile client UID. All other OAuth applications still see the consent prompt
* Since this doesn't give logged in users a chance to log out of the website, added a one-click confirmation if you're already logged in:

https://github.com/user-attachments/assets/4c03df00-df1a-4ec5-8ab6-6019eba721f3

### DIFF
--- a/spec/controllers/oauth/mobile_pre_authorizations_controller_spec.rb
+++ b/spec/controllers/oauth/mobile_pre_authorizations_controller_spec.rb
@@ -18,7 +18,7 @@ describe Oauth::MobilePreAuthorizationsController do
         expect(response).to have_http_status(:ok)
         expect(response.body).to include(@user.display_name)
         expect(response.body).to include(@user.email)
-        expect(response.body).to include("/oauth/authorize?client_id=abc&amp;redirect_uri=https%3A%2F%2Ffoo&amp;response_type=code")
+        expect(response.body).to include("/oauth/authorize?client_id=abc&amp;redirect_uri=gumroadmobile%3A%2F%2F&amp;response_type=code")
         expect(response.body).to include("Continue")
         expect(response.body).to include("Use a different account")
       end
@@ -28,7 +28,7 @@ describe Oauth::MobilePreAuthorizationsController do
       it "redirects to the oauth authorize url with query params" do
         get :new, params: { client_id: "abc", redirect_uri: "gumroadmobile://", response_type: "code" }
 
-        expect(response).to redirect_to("/oauth/authorize?client_id=abc&redirect_uri=https%3A%2F%2Ffoo&response_type=code")
+        expect(response).to redirect_to("/oauth/authorize?client_id=abc&redirect_uri=gumroadmobile%3A%2F%2F&response_type=code")
       end
     end
   end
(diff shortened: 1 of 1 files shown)


config/initializers/doorkeeper.rb
```diff
@@ -60,6 +60,10 @@ def public_scopes
   use_refresh_token
 
   grant_flows %w[authorization_code client_credentials password]
+
+  skip_authorization do |_resource_owner, client|
+    client.uid == OauthApplication::MOBILE_API_OAUTH_APPLICATION_UID
```
binary-koan, line 65: Note this isn't currently set to the Expo app in production (it's the current native app), but we'll update it as part of switching over

binary-koan:
@pegasus1134 So when testing I found an interesting issue with this ... if you're already logged in in the in-app browser it'll skip the flow entirely so there's no chance to log in to a different account (previously it would show the authorize prompt again with a logout button at the bottom)

Do you have any ideas on that? My immediate thought is to log out of the in-app browser when you log out of the app, but not sure how feasible that is

pegasus1134:
> @pegasus1134 So when testing I found an interesting issue with this ... if you're already logged in in the in-app browser it'll skip the flow entirely so there's no chance to log in to a different account (previously it would show the authorize prompt again with a logout button at the bottom)
> 
> Do you have any ideas on that? My immediate thought is to log out of the in-app browser when you log out of the app, but not sure how feasible that is

@binary-koan Good catch! The fix is a mobile-side change using preferEphemeralWebBrowserSession: true in promptAsync() so each OAuth flow starts a fresh browser session with no persisted cookies.

I'll add that in the mobile repo. The backend skip_authorization change itself is correct and this PR can be merged as is. The one CI failure is flaky.

binary-koan:
Thinking on it overnight, it's probably better UX to not force the user to enter their password again. Could we add a page in the backend that redirects to oauth immediately if the user is logged out, or shows something similar to the authorize UI if they're logged in? (if you can just get the flow working I'll tidy up the page design/content)

pegasus1134:
> Thinking on it overnight, it's probably better UX to not force the user to enter their password again. Could we add a page in the backend that redirects to oauth immediately if the user is logged out, or shows something similar to the authorize UI if they're logged in? (if you can just get the flow working I'll tidy up the page design/content)

That's a fair point... forcing re-authentication every time isn't great UX.

  I can add a custom pre-authorization page in the backend that:
  - If not logged in → redirects straight to /oauth/authorize (normal login flow)
  - If logged in → shows the current user with options to continue or switch accounts (logout + restart flow)

  The mobile app would then point to this page instead of /oauth/authorize directly. I'll get the flow working and you can polish the UI as you mentioned.

  I'll close the ephemeral session approach (antiwork/gumroad-mobile#81) in the meantime.

Do let me know if I should proceed with this. This approach will add maybe 30 lines of code to the backend repo, and like one line to the mobile repo.

binary-koan:
Yep sounds great

pegasus1134:
@binary-koan I've added the pre-authorization page as a new commit on my fork (pegasus1134/gumroad branch feat/skip-mobile-authorize-prompt). The mobile-side companion change is on pegasus1134/gumroad-mobile branch fix/ephemeral-oauth-session-24 - just points authorizationEndpoint to the new route. Let me know if I need to open a PR for the mobile repo.

pegasus1134:
@binary-koan The mobile_pre_authorizations_controller_spec is still failing on the push CI run (Test Fast 8). The spec passes redirect_uri: "https://foo" but the controller redirects with the mobile app's registered redirect URI (gumroadmobile://). The other 6 failures are all flaky.

PR #3803, opened 2026-02-24
Migrate all icons to Boxicons

binary-koan:
Ref #3011 

Removes our custom `Icon` component and replaces all usages with Boxicons.

Note this doesn't remove the SVG files, generation script, etc - will do that in another PR to avoid bloating this one further.

Non-exhaustive screenshots to give an idea:

Screenshot 2026-02-27 at 00 41 06
Screenshot 2026-02-27 at 00 48 51
Screenshot 2026-02-27 at 00 49 33
Screenshot 2026-02-27 at 00 49 54
Screenshot 2026-02-27 at 00 51 41
Screenshot 2026-02-27 at 00 53 19
Screenshot 2026-02-27 at 00 54 27

## AI Disclosure

Used Claude Opus 4.6 to replace most of the icons, then checked through manually.

Issue #3011, opened 2026-01-21
Migrate icons to Tailwind


binary-koan:
We should consider replacing our custom icons with a third party library, or look at refactoring the generation script to support tree shaking if we can't reasonably do that.

Also includes migrating the rest of `_icons.scss`, i.e. logos and brand icons.

slavingia:
Slight pref towards boxicons...

binary-koan:
I'm a fan of Remix but I agree Boxicons looks more Gumroad-y. I don't mind these as alternatives for Collaborators and Workflows:

<img width="95" height="87" alt="Image" src="https://github.com/user-attachments/assets/f3c320fd-f693-4124-977b-f204a3e6b5ae" />
<img width="95" height="90" alt="Image" src="https://github.com/user-attachments/assets/a6d1fed2-9ae7-40aa-b1dd-1545902e2209" />

TenzDelek:
They have recently released a React version too.
Github: https://github.com/box-icons/boxicons-react
NPM: https://www.npmjs.com/package/@boxicons/react

PR #3832, opened 2026-02-27
Remove old custom icons

binary-koan:
Ref #3011 

Since we replaced custom icons with Boxicons in #3803, we can now delete the old icon SVG files and the generation script for the SCSS and TS definitions.

The only visual changes are on the marketing pages where I've changed custom icons to Boxicons for consistency (missed these in the previous PR since they're in ERB):

| Before | After |
| --- | --- |
| 01-about-checklist | 01-about-checklist |
| 02-about-testimonials | 02-about-testimonials |
| 03-small-bets-experts | 03-small-bets-experts |


app/views/home/shared/_expert_card.html.erb
```diff
@@ -14,12 +14,12 @@
     <div class="flex gap-4 mt-auto">
       <% if twitter_url.present? %>
         <%= link_to twitter_url, target: "_blank", rel: "noopener", class: "hover:opacity-80" do %>
-          <%= image_tag "icons/twitter.svg", class: "w-6 h-6" %>
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" fill="currentColor" class="w-6 h-6"><path d="M13.68 10.62 20.24 3h-1.55L13 9.62 8.45 3H3.19l6.88 10.01L3.19 21h1.55l6.01-6.99 4.8 6.99h5.24l-7.13-10.38Zm-2.13 2.47-.7-1-5.54-7.93H7.7l4.47 6.4.7 1 5.82 8.32H16.3z"/></svg>
```
binary-koan: Inlined the Boxicons equivalents since we can't easily use React components. There's already a ton of hardcoded stuff in the `home` views so I think it's fine

app/javascript/stylesheets/_forms.scss
```diff
@@ -195,7 +195,14 @@ input[type="checkbox"]:not([role="switch"]) {
   &:checked::after {
     display: block;
     margin-inline: auto;
-    @extend %icon, .icon-outline-check;
+    content: "\00a0";
```
binary-koan, line 198: This SCSS will be replaced with React components and deleted soon anyway, so just inlined everything - it means we don't need to remember to delete the icons directory after the forms and global migrations

app/views/home/shared/_expert_card.html.erb
gianfrancopiana: Rails `image_tag` without an explicit alt auto-generates one from the filename (e.g. "LinkedIn"), so these icon-only links had accessible names before. The inline SVGs don’t. Consider adding aria-label to each link (e.g. "X profile", "LinkedIn profile") and aria-hidden="true" on the SVGs.

app/views/home/shared/_nav.html.erb
```diff
@@ -85,11 +85,11 @@
       </svg>
       <div data-github-stars-count class="flex items-center gap-1.5 whitespace-nowrap">
         <span class="text-base leading-none font-medium" data-github-stars-count-value></span>
-        <%= image_tag "icons/solid-star.svg", class: "dark:invert", width: "18", height: "18" %>
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="18" height="18" fill="currentColor" class="dark:invert"><path d="m6.87 14.33-1.83 6.4c-.12.4.03.84.37 1.08.34.25.8.26 1.14.02L12 18.2l5.45 3.63a.988.988 0 0 0 1.14-.02c.34-.25.49-.68.37-1.08l-1.83-6.4 4.54-4.08c.3-.27.41-.69.28-1.06-.13-.38-.47-.64-.87-.68l-5.7-.45-2.47-5.46a.998.998 0 0 0-1.82 0L8.62 8.06l-5.7.45c-.4.03-.74.3-.87.68s-.02.8.28 1.06z"/></svg>
```
gianfrancopiana: The old icons had hardcoded `fill="#000"` so `dark:invert` was needed. These new SVGs use `fill="currentColor"` which already adapts to the text color in dark mode. Adding invert on top flips it back to black. Same issue on line 92 with the arrow icon. Consider removing `dark:invert` from both.

app/assets/images/icons/arrow-diagonal-up-right.svg
```diff
@@ -1 +0,0 @@
-<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M3 21 21 3m0 0H3m18 0v18" stroke="#000" stroke-width="2"/></svg>
```
gianfrancopiana, line 1: `app/javascript/components/Home/Shared/Nav.tsx` still imports `arrow-diagonal-up-right.svg` (line 8) and `solid-star.svg` (line 9), both deleted by this PR. I think this will break the build when compiling the blog layout path (`GumroadBlog/Layout.tsx`). Either inline the SVGs here too, or keep these two assets until this component is migrated.

app/views/home/shared/_nav.html.erb
binary-koan: Didn't notice it kept the class, nice catch

app/assets/images/icons/arrow-diagonal-up-right.svg
binary-koan: Weird that CI passes ... since that's React I just imported the icon components as usual

app/javascript/stylesheets/_icons.scss
```diff
@@ -15,16 +15,6 @@
...
 
-.icon {
-  @extend %icon;
-}
-
-@each $icon in $icons {
-  .icon-#{$icon} {
-    mask-image: url("~images/icons/#{$icon}.svg");
-  }
-}
```
gianfrancopiana, line 26: Noticed `small_bets.html.erb` still uses `<%= icon "solid-check-circle" %>` in 8 places (lines 187–229). 

The helper generates `<span class="icon icon-solid-check-circle">`, which depends on both the` .icon` class and the `.icon-solid-check-circle` class being removed here. With this removal those spans will render as invisible empty elements. 

Should we migrate those call sites to inline SVGs too?

app/views/home/shared/_nav.html.erb
```diff
@@ -85,11 +85,11 @@
       </svg>
       <div data-github-stars-count class="flex items-center gap-1.5 whitespace-nowrap">
         <span class="text-base leading-none font-medium" data-github-stars-count-value></span>
-        <%= image_tag "icons/solid-star.svg", class: "dark:invert", width: "18", height: "18" %>
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="18" height="18" fill="currentColor"><path d="m6.87 14.33-1.83 6.4c-.12.4.03.84.37 1.08.34.25.8.26 1.14.02L12 18.2l5.45 3.63a.988.988 0 0 0 1.14-.02c.34-.25.49-.68.37-1.08l-1.83-6.4 4.54-4.08c.3-.27.41-.69.28-1.06-.13-.38-.47-.64-.87-.68l-5.7-.45-2.47-5.46a.998.998 0 0 0-1.82 0L8.62 8.06l-5.7.45c-.4.03-.74.3-.87.68s-.02.8.28 1.06z"/></svg>
```
gianfrancopiana: NIT:  this SVG is decorative. Can we add `aria-hidden="true"` here? Same for the arrow SVG on line 92.

app/views/home/shared/_testimonial.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="space-y-6">
   <div class="bg-white border border-dark-gray/50 rounded-3xl px-8 py-4 relative rounded-tl-3xl rounded-tr-3xl rounded-br-3xl rounded-bl-sm">
     <div class="mb-4">
-      <%= image_tag "icons/quote-squared.svg", alt: "Quote squared", class: "w-5 h-3" %>
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" fill="currentColor" class="w-5 h-3" aria-label="Quote"><path d="m21.01,10h-2.85c.27-1.02,1.01-2.51,3.09-3.03l.76-.19v-2.78h-1c-2.78,0-4.91.77-6.31,2.29-1.89,2.05-1.7,4.68-1.69,4.71v7c0,1.1.9,2,2,2h6c1.1,0,2-.9,2-2v-6c0-1.1-.9-2-2-2Z"/><path d="m9.01,10h-2.85c.27-1.02,1.01-2.51,3.09-3.03l.76-.19v-2.78h-1c-2.78,0-4.91.77-6.31,2.29-1.89,2.05-1.7,4.68-1.69,4.71v7c0,1.1.9,2,2,2h6c1.1,0,2-.9,2-2v-6c0-1.1-.9-2-2-2Z"/></svg>
```
gianfrancopiana: NIT: Can we add `aria-hidden="true"` here? Considering it is decorative, I think it would be more appropriate than `aria-label="Quote"` in this case.

app/javascript/stylesheets/_icons.scss
binary-koan: True - migrated. Also deleted that helper entirely since we shouldn't be using it

binary-koan:
Removed _icons.scss as this removes the last style rules from it

PR #3834, opened 2026-02-27
Add Top Creator badge announcement email

gianfrancopiana:
Fixes antiwork/gumroad-private#160

## Problem

The "Top Creator" badge feature is live on production and the announcement tweet has been posted, but verified creators haven't been notified via email that they earned the badge.

## Approach

Added a `top_creator_announcement` action to `CreatorMailer` that sends a branded email with the Top Creator badge image, a short announcement, and sign-off. The email is sent from `gumroad@creators.gumroad.com` (matching the mockup) and targets all verified (`verified: true`), alive, non-suspended users.

The mailer lives in `CreatorMailer` (not `OneOffMailer`) so it can be reused when new sellers earn the badge in the future (e.g. triggered from `admin/users_controller#verify`). A follow-up PR can add that automatic trigger.

A restart-safe `Onetime::SendTopCreatorAnnouncementEmail` service uses a Redis checkpoint to track the last processed user ID, so re-running after an interruption won't send duplicate emails. Emails are enqueued on the `low` Sidekiq queue via `deliver_later`.

Includes a mailer preview for local testing and full specs for both the mailer and the onetime service.

## Post-deploy steps

Once deployed, run in the production console:

```ruby
# 1. Test with a single user first
CreatorMailer.top_creator_announcement(user_id: <your_user_id>).deliver_now

# 2. Run the full batch
Onetime::SendTopCreatorAnnouncementEmail.new.process_with_logging
```

If the batch is interrupted and needs to be re-run, it will automatically resume from where it left off (Redis checkpoint). To reset and re-send to all users:

```ruby
Onetime::SendTopCreatorAnnouncementEmail.reset_last_processed_user_id
```

---


app/views/creator_mailer/top_creator_announcement.html.erb
```diff
@@ -0,0 +1,13 @@
...
+
+  <div style="text-align: center; padding: 24px 0;">
+    <%= image_tag("email/top_creator_badge.png", alt: "Top Creator Badge", style: "display: block; margin: 0 auto; max-width: 100%; height: auto;") %>
+  </div>
+
+  <p>You just earned the Top Creator badge on Gumroad.</p>
+
+  <p>It's now live on your profile and everywhere your products appear, so buyers always know they're getting something from one of Gumroad's best. No action needed. You've already done the hard part.</p>
+
+  <p>Keep creating,<br />The Gumroad Team</p>
```
gianfrancopiana, line 12: NIT: The original mockup had one empty line between "Keep creating," and "The Gumroad Team". Kept it this way for consistency with how we sign off emails.
Image

app/mailers/one_off_mailer.rb
```diff
@@ -34,4 +34,21 @@ def email_using_installment(user_id: nil, email: nil, installment_external_id:,
 
     email(user_id:, email:, subject: subject || @installment.subject, reply_to:, body: nil)
   end
+
+  def top_creator_announcement(user_id:)
```
ershad: Apart from sending this email to all top sellers now, we can send it in the future when a seller becomes a top seller, right? cc: @dvassallo 

If so, could we move this mailer method to its own mailer? `OneOffMailer` may not be the right place for it.

gianfrancopiana: Makes sense. Pushed changes and updated the PR description.

dvassallo: Yep we should send this email whenever we mark a seller as a top creator. But maybe that needs to be a separate PR?

gianfrancopiana: Already handled here. Changes are small enough 😎

PR #3838, opened 2026-02-27
Fix payout threshold ratchet effect on settings save

### LINKED ISSUE #3837: Payout threshold gets silently overwritten on every settings save
### Problem

The payments settings presenter sends `minimum_payout_amount_cents` (the effective minimum) to the frontend as `payout_threshold_cents` instead of the user's actual stored preference:

```ruby
# app/presenters/settings_presenter.rb:220
payout_threshold_cents: seller.minimum_payout_amount_cents,
```

`minimum_payout_amount_cents` is `max(payout_threshold_cents, minimum_payout_threshold_cents)` – it includes the country's cross-border minimum converted from local currency via exchange rate. The frontend initializes the form with this value and always sends it back on submission. The controller saves it directly.

This creates a ratchet effect: the country minimum (which fluctuates with exchange rates) gets "baked" into the user's stored `payout_threshold_cents` at peak rates and never comes back down, even when exchange rates drop or the user intended to keep the default $10 minimum.

### Reproduction

1. Be a seller in a country with a cross-border minimum > $10 (e.g. Argentina, Colombia, South Korea, Paraguay)
2. Visit Settings → Payments
3. Change any field (e.g. update bank account) and save
4. The country's exchange-rate-dependent minimum is now persisted as your `payout_threshold_cents`
5. If the exchange rate drops, your threshold stays at the old higher value

### Impact

Any seller in a cross-border payout country who submits the payments form has their threshold silently inflated. This may explain payout skips where users report a minimum they didn't set (e.g. antiwork/gumroad-private#166).

### Fix

Line 220 should send the user's stored preference, not the effective minimum:

```ruby
payout_threshold_cents: seller.payout_threshold_cents,
```

The frontend already receives `minimum_payout_threshold_cents` separately (line 221) and uses it for validation.

gianfrancopiana:
Fixes #3837

## Problem

The payments settings presenter sends `minimum_payout_amount_cents` (a computed value that includes the country's exchange-rate-dependent cross-border minimum) to the frontend as `payout_threshold_cents`. The frontend initializes the form with this inflated value and submits it back on every save. This creates a ratchet effect: the country minimum at the time of save gets permanently baked into the user's stored threshold and never comes back down, even when exchange rates drop.

Any seller in a cross-border payout country (e.g. Argentina, Colombia, South Korea, Paraguay) who submits the payments form has their threshold silently inflated.

## Approach

The presenter now sends the user's actual stored preference (`seller.payout_threshold_cents`) instead of the effective minimum (`seller.minimum_payout_amount_cents`). No frontend changes are needed – the frontend already receives `minimum_payout_threshold_cents` separately and uses it for validation.

Added a presenter spec for a cross-border seller whose stored threshold is below the country minimum, verifying the stored value is returned. Also added an integration regression test proving the threshold isn't inflated on save.

---

### DIFF
--- a/spec/requests/settings/payments_spec.rb
+++ b/spec/requests/settings/payments_spec.rb
@@ -6503,7 +6503,7 @@ describe("Payments Settings Scenario", type: :system, js: true) do
           field = find_field("Minimum payout threshold")
           expect(field.value).to eq("10")
 
-          field.send_keys([:command, "a"], "35")
+          fill_in "Minimum payout threshold", with: "35", fill_options: { clear: :backspace }
 
           click_on "Update settings"
 


spec/presenters/settings_presenter_spec.rb
```diff
@@ -820,6 +820,19 @@
...
+    context "when seller is in a cross-border payout country with a stored threshold below the country minimum" do
+      let!(:compliance_info) { create(:user_compliance_info_korea, user: seller) }
+
+      before do
+        seller.update!(payout_threshold_cents: 1000)
+      end
+
+      it "returns the stored payout threshold, not the effective minimum" do
+        expect(seller.minimum_payout_amount_cents).to eq(seller.minimum_payout_threshold_cents)
+        expect(presenter.payments_props[:payout_threshold_cents]).to eq(1000)
```
hchhabra, line 832:
```suggestion
        expect(presenter.payments_props[:payout_threshold_cents]).to be < seller.minimum_payout_amount_cents
        expect(presenter.payments_props[:payout_threshold_cents]).to eq(1000)
```

Just to make it explicit that `seller.minimum_payout_threshold_cents` is not $10 here.

spec/requests/settings/payments_spec.rb
```diff
@@ -6494,6 +6494,17 @@
           expect(page).to have_alert(text: "Thanks! You're all set.")
           expect(user.reload.minimum_payout_amount_cents).to eq(4000)
         end
+
+        it "does not inflate the stored payout threshold to the effective minimum on save" do
```
hchhabra: `user.minimum_payout_threshold_cents` is $34.74 here which is less than $50, so this test does not really verify the code change I think. We should instead check that an initial `payout_threshold_cents` of less than $34.74 is correctly shown in the field on page load, it then needs to be updated to >=$34.74 in order to re-submit, and then a reduction in fx rate (i.e. reduced `minimum_payout_threshold_cents`) allows submitting an amount < $34.74?

PR #3844, opened 2026-02-27
Support native table of contents in Expo app

binary-koan:
Paired with https://github.com/antiwork/gumroad-mobile/pull/70

* Adds support for sending the table of contents to the Expo app and navigating when requested
* Hides the web navigation when running in React Native

### DIFF
--- a/spec/requests/download_page/rich_text_editor_spec.rb
+++ b/spec/requests/download_page/rich_text_editor_spec.rb
@@ -252,17 +252,17 @@ describe("Download Page – Rich Text Editor Content", type: :system, js: true) 
 
       toc_messages = page.evaluate_script("window._messages.filter(m => m.type === 'tocData')")
       expect(toc_messages).to eq([
-        {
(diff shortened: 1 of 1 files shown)


binary-koan:
@yashranaway FYI - pushed your web branch here and simplified the logic a bit, lmk if anything doesn't look right!

yashranaway:
@binary-koan Looks good, went through all your commits. And thanks for adding the spec, I should've done that from the start.

One thing I noticed - in `url_redirects_controller.rb` you changed `is_mobile_app_web_view` back to only matching `mobile_app`. Since the mobile app now sends `display=mobile_app` again that works, just wanted to confirm that's intentional and we don't need the `expo_app` value at all anymore?

binary-koan:
Yeah I think it's fine to just check `window.ReactNativeWebView` to see if it's the Expo app (we'll remove that extra logic when we remove the old apps anyway)

PR #3847, opened 2026-02-28
Migrate logo icons to Tailwind

binary-koan:
Ref #3011 

Converts the `logo-g` and `logo-full` classes to Tailwind + React. See inline comments for details.

No visual changes, logos look the same:

Screenshot 2026-02-28 at 23 29 08
Screenshot 2026-02-28 at 23 29 32
Screenshot 2026-02-28 at 23 29 51

### DIFF
--- a/spec/requests/purchases/product/taxes_spec.rb
+++ b/spec/requests/purchases/product/taxes_spec.rb
@@ -3958,7 +3958,7 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
     end
   end
 
-  describe "Canada Tax", taxjar: true do
+  describe "Canada Tax", taxjar: true, force_vcr_on: true do
     let (:product) { create(:product, price_cents: 100_00) }
 
     it "detects the province for Canada" do
@@ -4046,10 +4046,10 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
 
         expect(page).to have_select("Country", selected: "Canada")
         expect(page).to have_select("Province", selected: "ON")
-        expect(page).to_not have_field "Business QST ID (optional)"
+        expect(page).to_not have_field("Business QST ID (optional)")
 
         select "QC", from: "Province"
-        expect(page).to have_field "Business QST ID (optional)"
+        expect(page).to have_field("Business QST ID (optional)")
         check_out(product, zip_code: nil, credit_card: { number: "4000001240000000" })
 
         purchase = Purchase.last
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -13,7 +13,7 @@ def default_footer_content
     safe_join(
       [
         "Powered by",
-        tag.span("Gumroad", class: "logo-full", aria: { label: "Gumroad" })
+        tag.span("Gumroad", class: "inline-block aspect-115/22 h-[1lh] shrink-0 bg-current mask-(--logo) mask-contain mask-center mask-no-repeat")
```
binary-koan, line 16: As with other migrations, just inlined these into the helper. Eventually we'll replace all ERB with Inertia and remove these

app/javascript/components/Discover/Layout.tsx
```diff
@@ -136,11 +137,9 @@ export const Layout: React.FC<{
   const userActionButtons = <UserActionButtons />;
 
   const logoLink = (
-    <a
-      href={Routes.discover_url({ host: discoverDomain })}
-      className="logo-full flex aspect-157/22 w-[245px]! shrink-0 items-center"
-      aria-label="Gumroad"
-    />
+    <a href={Routes.discover_url({ host: discoverDomain })} className="shrink-0" aria-label="Gumroad">
+      <Logo className="h-auto w-[245px]" />
```
binary-koan, line 141: Overrides the default styling and computes the aspect ratio based on width rather than height (note: since the aspect ratio in `Logo` is different the height actually changes, but it's a flex container anyway so it doesn't affect the layout)

app/javascript/components/Nav.tsx
```diff
@@ -130,7 +133,7 @@ export const Nav = ({ title, children, footer }: Props) => {
       >
         <div className="override grid grid-cols-[auto_1fr_auto] items-center gap-3 p-4 text-lg leading-6 lg:hidden">
           <a href={Routes.root_url()} className="no-underline">
-            <span className="logo-g" />
+            <img src={logo} className="size-6" alt="Home" />
```
binary-koan, line 136: This is the only place `logo-g` was used, and it works fine with just an `img` tag

app/javascript/widget/overlay.scss
```diff
@@ -33,8 +33,18 @@
 a.gumroad-button {
   --parent-color: 0 0 0 !important;
   font-family: $font-family;
-  $icons: [];
-  @import "../stylesheets/icons";
+
+  .logo-full {
```
binary-koan, line 37: Similar to button styles, just inlined the relevant CSS as-is here. We don't need to migrate this file in the short term, we can handle it further down the line alongside emails since it's another bundle separate to the main app

app/javascript/components/Logo.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+
+import { classNames } from "$app/utils/classNames";
+
+export const Logo = ({ className, ...props }: React.ComponentPropsWithoutRef<"span">) => (
+  <span
+    className={classNames(
+      // Deliberately using an aspect ratio that adds vertical padding, so that the logo aligns with text
+      "inline-block aspect-115/22 h-[1lh] shrink-0 bg-current mask-(--logo) mask-contain mask-center mask-no-repeat selection:bg-transparent selection:text-transparent",
```
binary-koan, line 9: Kept the `mask-image` approach since it's easier to handle dark mode than `img`, easier to read than inline SVG, and it's nice that when the page content is copied and pasted "Powered by Gumroad" is preserved as text

Issue #3010, opened 2026-01-21
Migrate _rich_text.scss to Tailwind


binary-koan:
Split from #1055 

This is a bit tricky as we need to style Prosemirror elements, so we may want to keep some of the styles in CSS and just switch to `@apply`.

Note: We're planning to tackle this internally, but can consider bounties if we get great PRs for parts of the file.

Hanaffi:
@binary-koan quick clarification: is the goal to convert everything to Tailwind via `@apply`, as recommended in the docs https://tiptap.dev/docs/editor/getting-started/style-editor#using-a-global-css-file-and-apply ?

Or do you expect the other approach of adding classes directly to the extensions?

PR #3849, opened 2026-03-02
Migrate rich text editor to Tailwind

binary-koan:
Ref #3010 

Migrates the parts of `_rich_text.scss` that aren't used in emails to Tailwind - mainly the toolbar, actions menu, and other editor-specific features.

As mentioned in the issue, the rest of the file can't be migrated until we're able to migrate emails as well, since we dump the editor HTML into email bodies.

No visual changes, the editor still works as before:

Screenshot 2026-03-04 at 01 49 18
Screenshot 2026-03-04 at 01 50 54


app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -698,12 +697,16 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string |
                     />
                   </div>
                 </Modal>
-                <Separator aria-orientation="vertical" />
+                <div
+                  role="separator"
+                  aria-orientation="vertical"
+                  className="m-2 hidden border-r border-solid border-muted sm:flex"
+                />
```
binary-koan, line 704: Separators were styled in the SCSS, but don't actually share a lot with the `Separator` component. Inlined and kept (roughly) the SCSS styles

app/javascript/components/TiptapExtensions/Image.tsx
```diff
@@ -111,11 +117,7 @@ const ImageNodeView = ({ node, editor, getPos }: NodeViewProps) => {
 
   return (
     <NodeViewWrapper>
-      <figure
-        ref={nodeRef}
-        data-has-focus={hasFocus || undefined}
-        style={isUploading ? { position: "relative" } : undefined}
-      >
+      <figure ref={nodeRef} style={isUploading ? { position: "relative" } : undefined}>
```
binary-koan, line 120: Removed `data-has-focus` in favour of applying focus styles directly to the image above

app/javascript/components/TiptapExtensions/NodeActionsMenu.tsx
```diff
@@ -1,12 +1,49 @@
...
+  isEditable?: boolean;
+  asChild?: boolean;
+} & React.HTMLAttributes<HTMLDivElement>) => {
+  const Component = asChild ? Slot : "div";
+  return (
+    <SelectedContext.Provider value={selected}>
+      <Component
+        className={classNames(
+          isEditable && [
+            "relative before:absolute before:[inset:0_100%_0_-3rem] before:content-['']",
```
binary-koan, line 32: This `before` element is to increase the node bounding box and make it easier to trigger the actions menu (which floats outside the element on the left)

app/javascript/components/TiptapExtensions/NodeActionsMenu.tsx
```diff
@@ -1,12 +1,49 @@
...
+            "[&:hover>[data-actions-menu]]:lg:block",
+            "[&:has([data-child-area]:hover)>[data-actions-menu]]:lg:hidden",
```
binary-koan, line 34: As much as possible we're trying to use contexts (e.g. SelectedContext above) for parent/child relationships rather than CSS nesting, but I couldn't find a feasible way to make it work with the actions menu

Basically the tricky case is file embed groups which contain multiple files (which are themselves nodes). The logic is: show the menu for the group (but not children) on hover, _unless_ a child is currently being hovered (in which case we show the menu for the child but not the group). These CSS rules seem much more elegant than trying to track the hover states in JS

Note: Tailwind `group` / `group-has` don't use the `>` selector so the child's menu wrongly shows based on the parent's hover state

app/javascript/stylesheets/tailwind.css
```diff
@@ -122,3 +122,20 @@
...
+.ProseMirror[contenteditable="true"] figure br::selection {
+  @apply bg-transparent;
+}
+
+/* Tiptap placeholder plugin */
+.ProseMirror p.is-editor-empty:first-child::before,
+.ProseMirror .node-image.is-empty .figcaption::before {
+  content: attr(data-placeholder);
+  @apply pointer-events-none float-left h-0 text-muted;
+}
```
binary-koan, line 141: These classes are all added directly by Tiptap/ProseMirror, so aren't feasible to inline

spec/requests/products/edit/rich_text_editor_spec.rb
```diff
@@ -1079,27 +1081,27 @@
         visit edit_link_path(@product) + "/content"
 
         within find_embed(name: "First file").hover do
-          select_disclosure "Actions" do
-            expect(page).to have_menuitem("Move to folder...")
-          end
+          toggle_disclosure "Actions"
+          expect(page.document).to have_menuitem("Move to folder...")
+          toggle_disclosure "Actions", expand: false
```
binary-koan, line 1086: The popovers now use portals so we need to find the content outside the embed/disclosure

`expect(page.document)` unscopes the `within` so we don't need an extra block to close the popover

Note that the specs previously didn't close the popover, and actually worked because of a visual bug where other actions menu buttons would overlay the currently open menu. Portals fix that behaviour but mean we need to close the popovers explicitly for Capybara to see the next button

app/javascript/components/RichTextEditor.tsx
```diff
@@ -282,7 +281,7 @@ export const useRichTextEditor = ({
     editable,
     editorProps: {
       attributes: {
-        ...(className ? { class: className } : {}),
+        class: classNames("min-h-full whitespace-break-spaces rounded-t-none focus-within:outline-none", className),
```
gianfrancopiana: `whitespace-break-spaces` and `rounded-t-none` were previously scoped to editable/editor-shell contexts (`.ProseMirror[contenteditable="true"] `/ `.rich-text-editor`). Now they apply unconditionally. This could affect read-only surfaces (product page, profile, download content). Worth gating behind `editable`?

app/javascript/components/TiptapExtensions/NodeActionsMenu.tsx
```diff
@@ -1,12 +1,49 @@
...
+          ],
+          selected && "relative rounded outline-2 outline-accent [&_*::selection]:bg-none",
```
gianfrancopiana: `bg-none` is `background-image: none`, but `::selection` uses `background-color`. So this doesn't suppress the selection highlight. The old SCSS rule was `background: none` which resets both. Should be `bg-transparent` here.

binary-koan:
Thanks, fixed! Could you take another look? (I'll re-run CI when actions are working again ...)

Issue #3850, opened 2026-03-02
Fixed-length membership receipt incorrectly shows upcoming payment and "2 of 1" payment count


michellelarney:
Description:
When creating a fixed-length membership with a duration of 1 month, the test receipt incorrectly displays an upcoming payment. The payment count shown is "2 of 1", which is impossible for a 1-payment fixed-length membership.

<img width="395" height="594" alt="Image" src="https://github.com/user-attachments/assets/afc003eb-417d-4911-93f0-9343512d43b7" />

  Steps to Reproduce

  1. Go to Creator Dashboard → Products → New Membership
  2. Set the membership duration to Fixed length → 1 month
  3. Save and Publish product
  4. Run a Test purchase
  5. Check the receipt

  Expected Behavior
  - No upcoming payment should be shown (the membership ends after 1 month)
  - Payment count should show "1 of 1"

PR #3851, opened 2026-03-02
Fix fixed-length membership receipt for test purchases

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad/issues/3850

## Problem

Fixed-length membership receipts show "2 of 1" payment count and a spurious upcoming payment section when the purchase is a test purchase. This happens because `charges_completed?` and `remaining_charges_count` on Subscription use `purchases.successful` which only matches `purchase_state: "successful"`, but test purchases have state `"test_successful"`.

## Solution
- Changed `charges_completed?` and `remaining_charges_count` to use the existing private `successful_purchases` helper, which returns `purchases.test_successful` for test subscriptions and `purchases.successful` for regular ones
- Fixed `subscription_charge_number` in the receipt presenter to use the correct scope for test subscriptions

## Before
Screenshot 2026-03-02 at 14 28 54

## After
Screenshot 2026-03-02 at 14 44 22

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -754,11 +754,11 @@ class Subscription < ApplicationRecord
   end
 
   def charges_completed?
-    has_fixed_length? && purchases.successful.count == charge_occurrence_count
+    has_fixed_length? && successful_purchases.count == charge_occurrence_count
   end
 
   def remaining_charges_count
-    has_fixed_length? ? charge_occurrence_count - purchases.successful.count : 0
+    has_fixed_length? ? charge_occurrence_count - successful_purchases.count : 0
   end
 
   # Certain events should transition the subscription from pending cancellation to cancelled thus not allowing the customer access to updates.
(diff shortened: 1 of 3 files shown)


app/models/subscription.rb
```diff
@@ -754,11 +754,11 @@ def has_fixed_length?
   end
 
   def charges_completed?
-    has_fixed_length? && purchases.successful.count == charge_occurrence_count
+    has_fixed_length? && successful_purchases.count == charge_occurrence_count
```
nyomanjyotisa, line 757: `purchases.successful` only matches `purchase_state: "successful"`, missing test purchases (`"test_successful"`). `successful_purchases` already handles this by checking `is_test_subscription`

```
    def successful_purchases
      is_test_subscription ? purchases.test_successful : purchases.successful
    end
```

app/presenters/receipt_presenter/payment_info.rb
```diff
@@ -373,10 +373,9 @@ def subscription_charge_number(purchase)
       if @_subscription_charge_number.key?(purchase.id)
         @_subscription_charge_number[purchase.id]
       else
-        @_subscription_charge_number[purchase.id] = purchase
-          .subscription
-          .purchases
-          .successful
+        subscription = purchase.subscription
+        scope = subscription.is_test_subscription ? subscription.purchases.test_successful : subscription.purchases.successful
```
nyomanjyotisa, line 377: Same fix as the model, need to use `test_successful` scope for test subscriptions

Issue #3644, opened 2026-02-17
Add Membership Cancellation Timestamp to Gumroad Admin + “Date of Cancellation” Column in Sales CSV


michellelarney:
We should add a cancellation timestamp for Memberships in Gumroad Admin and also include this data as a new column in the Sales CSV export.

Currently, we have a Subscription End Date column, but this is not the same as the actual cancellation timestamp.

There is an important distinction between:
Cancellation timestamp → The date and time the customer cancelled their membership
Subscription End Date → When access actually ends (typically at the next billing date)

Providing both of these will help with customer service queries and data for sellers. 

This is what Gumroad Admin looks like now: 

<img width="192" height="399" alt="Image" src="https://github.com/user-attachments/assets/9533905b-d211-4fa4-8565-cfa4a38f1b67" />

Change "Cancelled" to have the Cancellation timestamp. Add another part that says "Ended:" that shows when the timestamp of when access actually ends.

PR #3852, opened 2026-03-02
Add membership cancellation timestamp to Admin UI and Sales CSV

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad/issues/3644

## Problem
The Admin panel showed `cancelled_at` for the "Cancelled" field, but for regular cancellations this value is set to `end_time_of_subscription` (the next billing date), not when the customer actually cancelled. The "Ended" field only showed `ended_at`, which is only set for fixed-length subscriptions. The Sales CSV had no column for the actual cancellation timestamp.

## Solution
- Admin "Cancelled" row: Show `user_requested_cancellation_at` (the actual moment cancellation was requested) instead of `cancelled_at`
- Admin "Ended" row: Show `cancelled_at` (when access ends at next billing date), with `ended_at` fallback for fixed-length subscriptions, since fixed-length subscriptions has no cancelled_at data when it ends automatically
- Sales CSV: Add "Date of Cancellation" column with `user_requested_cancellation_at` (full datetime)

## Before/After
Scenario: a user purchased 1 year membership today, and cancelled the membership today

### Before
Screenshot 2026-03-02 at 16 27 40

.

image

### After
image

.

image

### DIFF
--- a/spec/presenters/admin/purchase_presenter_spec.rb
+++ b/spec/presenters/admin/purchase_presenter_spec.rb
@@ -128,7 +128,7 @@ describe Admin::PurchasePresenter do
           it "sets ended_at to cancelled_at and returns cancellation timestamps" do
             freeze_time do
               expect(props[:subscription][:user_requested_cancellation_at]).to be_present
-              expect(props[:subscription][:ended_at]).to eq(end_time)
+              expect(props[:subscription][:ended_at]).to be_within(1.second).of(end_time)
               expect(props[:subscription][:cancelled_by_buyer]).to eq(true)
             end
           end


app/javascript/components/Admin/Purchases/index.tsx
```diff
@@ -404,16 +405,20 @@ const Info = ({ purchase }: { purchase: Purchase }) => (
         <>
           <dt>Cancelled</dt>
           <dd>
-            <BooleanIcon value={!!purchase.subscription.cancelled_at} />
-            {purchase.subscription.cancelled_at
-              ? ` (on ${purchase.subscription.cancelled_at} by ${purchase.subscription.cancelled_by_buyer ? "buyer" : "seller"})`
+            <BooleanIcon value={!!purchase.subscription.user_requested_cancellation_at} />
+            {purchase.subscription.user_requested_cancellation_at
+              ? ` (on ${purchase.subscription.user_requested_cancellation_at} by ${purchase.subscription.cancelled_by_buyer ? "buyer" : "seller"})`
```
nyomanjyotisa, line 409: `user_requested_cancellation_at` is the actual moment the user clicked cancel

app/javascript/components/Admin/Purchases/index.tsx
```diff
@@ -404,16 +405,20 @@ const Info = ({ purchase }: { purchase: Purchase }) => (
...
 
           <dt>Ended</dt>
           <dd>
-            <BooleanIcon value={!!purchase.subscription.ended_at} />
-            {purchase.subscription.ended_at ? ` (on ${purchase.subscription.ended_at})` : null}
+            <BooleanIcon value={!!purchase.subscription.cancelled_at || !!purchase.subscription.ended_at} />
+            {purchase.subscription.cancelled_at
+              ? ` (on ${purchase.subscription.cancelled_at})`
+              : purchase.subscription.ended_at
+                ? ` (on ${purchase.subscription.ended_at})`
```
nyomanjyotisa: `cancelled_at` covers regular and immediate cancellations. `ended_at` fallback is needed for fixed-length subscriptions that complete all charges

app/services/exports/purchase_export_service.rb
```diff
@@ -10,7 +10,7 @@ class Exports::PurchaseExportService
     "Discount Code", "Recurring Charge?", "Free trial purchase?", "Pre-order authorization?", "Product ID", "Order Number",
     "Pre-order authorization time (UTC timezone)", "Custom Fields", "Item Price ($)",
     "Variants Price ($)", "Giftee Email", "SKU ID", "Quantity", "Recurrence",
-    "Affiliate", "Affiliate commission ($)", "Discover?", "Subscription End Date", "Rating", "Review",
+    "Affiliate", "Affiliate commission ($)", "Discover?", "Date of Cancellation", "Subscription End Date", "Rating", "Review",
```
EmCousin: Cancellation date seems like a more consistent name? (Nit)

app/javascript/components/Admin/Purchases/index.tsx
EmCousin: Can we move this logic to the presenter and test this? That way we only use a 'ended' attribute that is consistent with the title here.

Issue #3690, opened 2026-02-19
Add Claude Code-powered GitHub Actions to auto-close non-compliant issues and PRs


slavingia:
## Summary

Add GitHub Actions workflows powered by Claude Code that automatically enforce our contribution guidelines by:

**Update existing workflow**
- [ ] Add issue template
- [ ]  **Auto-closing non-compliant issues** — Issues that don't follow the bug report format outlined in CONTRIBUTING.md (missing steps to reproduce, expected vs actual behavior, etc.) should be auto-closed with a comment explaining what's missing and linking to our contributing guidelines.
- [ ] **Auto-closing account-specific support issues** — Issues about specific Gumroad accounts (e.g., "my account was suspended", "I can't access my payouts", "why was my product removed") should be auto-closed with a comment directing the user to open a support ticket instead, since GitHub Issues is not the right channel for account-level support.

**Add new workflow**
- [ ] **Auto-closing non-compliant pull requests** — PRs that violate our contributing rules (e.g., missing tests, no documentation updates for behavior changes, incorrect branch naming, non-conventional commit messages) should be auto-closed with a comment explaining the issues.

### For Issues
- On `issues.opened`, Claude Code reviews the issue against CONTRIBUTING.md guidelines
- If it's a **support request about a specific account**: auto-close with a friendly comment like:
  > "Thanks for reaching out! This issue appears to be about a specific Gumroad account. GitHub Issues is for bug reports and feature discussions — for account-specific help, please open a support ticket at [support link]. Closing this issue."
- If it's a **bug report missing required info**: auto-close with a comment listing what's missing and linking to CONTRIBUTING.md
- If it's a **question or idea** (not a bug/problem): auto-close with a comment redirecting to Discussions

### For Pull Requests
- On `pull_request.opened`, Claude Code reviews the PR against CONTRIBUTING.md guidelines
- Check for: tests included, documentation updated (if behavior changed), conventional commit format, etc.
- If non-compliant: auto-close with a comment explaining what needs to be fixed before re-opening

## Why Claude Code?
Using Claude Code gives us semantic understanding of issue/PR content rather than relying on brittle regex or keyword matching. It can understand context, distinguish between a bug report and a support request, and provide helpful, specific feedback to contributors.

## Test Plan
- [ ] Test with sample compliant and non-compliant issues
- [ ] Test with account-specific support requests
- [ ] Test with compliant and non-compliant PRs
- [ ] Verify comments are helpful and reference the correct guidelines
[...]

PR #3859, opened 2026-03-02
Add workflow to close support request issues

gianfrancopiana:
Addresses #3690 (partial)

## What

New GitHub Actions workflow that uses Claude to detect when an issue is a support request about a specific Gumroad account (suspended account, payout problems, refund requests, etc.) rather than a bug report or feature request. Posts a fixed comment directing the user to https://help.gumroad.com and does not close the issue (yet).

Uses the same patterns as the other issue triage workflows: shared concurrency group to avoid race conditions, skips maintainers, checks if the issue is still open before invoking Claude.

## Why

Users frequently open GitHub issues for account-specific support requests that maintainers need to manually redirect. This automates the redirect to the help center, where an AI agent handles common questions and escalates to a human when needed.

---

### DIFF
--- a/.github/workflows/close-support-issues.yml
+++ b/.github/workflows/close-support-issues.yml
@@ -1,4 +1,3 @@
-# TODO: auto-close support issues once false positive rate is acceptable
 name: Close support issues
 
 on:


.github/workflows/close-support-issues.yml
```diff
@@ -0,0 +1,75 @@
+# TODO: auto-close support issues once false positive rate is acceptable
```
slavingia: do we want this todo here? generally prefer these live in our own internal notes / issues.

PR #3863, opened 2026-03-03
Fix blank product page when default offer code is expired

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad-private/issues/182

## What
When a product has an expired automatic (default) discount code, the product page renders blank.

`BestOfferCodeService#result` now distinguishes between two scenarios when neither code is valid:
- If the buyer explicitly provided a URL code, return that code's error so they see feedback.
- If only the automatic default code was tried and it is expired/invalid, return `nil` instead of the URL code's `missing_code` error. The buyer never asked for a discount, so no error should surface.

## Why
The product edit page already validates that sellers cannot select an expired discount code. The problem occurs when a discount code expires after it was set as the default.

The service returned `{ valid: false, error_code: :missing_code }` from evaluating a nil URL code. The frontend `ProductDiscount` type only allows specific `error_code` values (`sold_out`, `invalid_offer`, `inactive`, `unmet_minimum_purchase_quantity`). The `missing_code` value caused `ts-safe-cast` to throw a runtime exception, crashing the page to a blank screen.

Returning `nil` for expired automatic discounts is the correct behavior

## Before
Screenshot 2026-03-03 at 11 20 13

## After
Screenshot 2026-03-03 at 11 19 55

### DIFF
--- a/spec/services/best_offer_code_service_spec.rb
+++ b/spec/services/best_offer_code_service_spec.rb
@@ -75,7 +75,9 @@ describe BestOfferCodeService do
       end
 
       context "and it is expired" do
-        let(:default_offer_code) { create(:offer_code, products: [product], code: "DEFAULT10", amount_cents: 200, valid_at: 2.days.ago, expires_at: 1.day.ago, currency_type: product.price_currency_type) }
+        before do
+          default_offer_code.update_column(:expires_at, 1.day.ago)
+        end
 
         it "returns nil rather than an error (buyer did not apply the code)" do
           expect(subject.result).to be_nil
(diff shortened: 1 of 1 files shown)


spec/services/best_offer_code_service_spec.rb
```diff
@@ -73,6 +73,14 @@
           expect(subject.result).to be_nil
         end
       end
+
+      context "and it is expired" do
+        let(:default_offer_code) { create(:offer_code, products: [product], code: "DEFAULT10", amount_cents: 200, valid_at: 2.days.ago, expires_at: 1.day.ago, currency_type: product.price_currency_type) }
```
EmCousin: Makes sense, creating an expired code from the start is invalid. You'll need to create it non-expired first, then force-set the expiration date to `1.day.ago` via `update_column`

Issue #3475, opened 2026-02-10
Fix broken product recommendations UI in purchase receipts


vatsalkaushik:
- [ ] Fix broken product recommendations UI in purchase receipts

Receipt example: https://gumroad.com/purchases/e3G2aNWeRB4Jv51DqeYCoA==/receipt

Web: 

https://github.com/user-attachments/assets/2417aa0a-2535-4aa4-b16b-a516a9d06ebb

Email:

https://github.com/user-attachments/assets/6be89892-4d69-4f57-a24c-4d7cfcf61745

PR #2223, opened 2025-12-04
Migrate `_product.scss` to Tailwind

SalmanNajah:
Part of #1055

Addressed all the comments on #2113

### Before
| Desktop | Mobile |
|-------|--------|
|  |  |

### After  

https://github.com/user-attachments/assets/22297ae8-1383-41ad-937b-413676fd2b8b

--- 

### Before
| Desktop | Mobile |
|---------|--------|
|  |  |

## After  

https://github.com/user-attachments/assets/d6dbd7a1-0396-41f9-a02e-d95a73af872b

---

### Before
| Light | Dark |
|---------|--------|
|before-l-horizontal |before-d-horizontal|
|before-l|before-d|
|Screenshot 2025-11-02 at 3 06 20 PM|Screenshot 2025-11-02 at 3 06 59 PM|
|Screenshot 2025-11-02 at 3 06 36 PM|Screenshot 2025-11-02 at 3 06 50 PM|

### After  

https://github.com/user-attachments/assets/3e32ce48-1284-44c1-b849-ce96dc4769ad

--- 

### Before
| Light | Dark |
|---------|--------|
|cross sell before l|cross sell before|
|Screenshot 2025-11-02 at 3 21 54 PM|Screenshot 2025-11-02 at 3 22 08 PM|

### After  

https://github.com/user-attachments/assets/d4c69427-2a73-4546-9b93-19c0be0e1987

--- 

### Before
| Light | Dark |
|---------|--------|
|library light|library|
|Screenshot 2025-11-02 at 3 50 36 PM|Screenshot 2025-11-02 at 3 50 49 PM|

### After  

https://github.com/user-attachments/assets/5658f3e7-e71e-4646-9a12-da735d268f54

--- 

### Before
| Light | Dark |
|---------|--------|
|image|image|
|image|Screenshot 2025-11-04 at 9 51 53 PM|

### After  

https://github.com/user-attachments/assets/d9731755-cb1d-4148-89a1-e5e0514d8515

--- 

### Before
| Desktop | Mobile |
|---------|--------|
|image|image|
|image|image|

### After  

https://github.com/user-attachments/assets/7334b77b-3396-43d0-87eb-a677afab8604

--- 

### Before
| Desktop | Mobile |
|---------|--------|
|before-wishlist|before-wishlist-mb-1|

### After  

https://github.com/user-attachments/assets/750cb3bb-9caa-45a6-ad8f-30bb5cef5581

--- 

## AI Disclosure
[...]

PR #3865, opened 2026-03-03
Fix unequal product card widths in receipt email

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad/issues/3475

## What

Adds `display: grid` to `.product-card` in the email stylesheet. This prevents recommended product cards from having unequal widths when displayed side by side in receipt emails.

## Why

The `.product-card .header h4` inherits `white-space: nowrap` from the `text-singleline` mixin applied via `.header > :not(:last-child)`. In a table with `table-layout: auto`, `white-space: nowrap` forces the text onto a single line without wrapping, which causes the table cell to expand to fit the full title length. Cards with longer titles end up wider than cards with shorter titles.

Adding `display: grid` to `.product-card` constrains the card's children to the grid container's width, preventing the nowrap content from pushing the cell wider.

### Before
Screenshot 2026-03-03 at 16 31 17

### After
Screenshot 2026-03-03 at 16 34 18

## Modifing css on prod on the reported receipt ( https://gumroad.com/purchases/1CfxpM21kXlHaXMAzwN0Fg==/receipt?email=mr.osamaarshad%40gmail.com ) :

https://github.com/user-attachments/assets/5deb4aa6-f378-4d8d-bd4f-79cedba8ea30

### DIFF
--- a/app/javascript/stylesheets/_email.scss
+++ b/app/javascript/stylesheets/_email.scss
@@ -617,6 +617,7 @@ $accent-color: map-get($colors, "pink"); // [1]
   @include bg-color(filled);
   border: $border;
   border-radius: border-radius(1);
+  display: grid;
 
   > figure {
     background: url("~images/placeholders/product-cover.png");


binary-koan:
Could you send a test email and screenshot this in Gmail? Most email clients don’t support display:grid AFAIK

nyomanjyotisa:
@binary-koan It works

Seems like regression from `_product.scss` migration, previously we have `display: grid;` on `.product-card` https://github.com/antiwork/gumroad/pull/2223/changes#diff-701d358aeb3de59c7b384837e7a114a69541417f25dbddd4d3178fcef618764aL70

No issue with old email, and it use `display: grid` as well
image
image

PR #3866, opened 2026-03-03
Append purchase_id to post URLs on download page rich content

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad-private/issues/186

## What

Auto-appends `?purchase_id=` to Gumroad post URLs (`*.gumroad.com/p/*`) in download page rich content, so purchasers without a Gumroad account can access posts linked from the download page.

## Why

When a seller includes a link to a Gumroad post in the download page rich content, purchasers without a Gumroad account get a "not found" error. The `PostPresenter#e404?` already supports bypassing visibility checks when a valid `purchase_id` param is provided, but links in rich content were not including it. The `saleInfo.sale_id` (which equals `purchase.external_id`) is already available on the download page, so we use it to append `?purchase_id=` to detected Gumroad post URLs.

### Before

https://github.com/user-attachments/assets/3c3441b9-2d68-4b13-b66b-30076b493457

### After

https://github.com/user-attachments/assets/a505102c-cac4-4523-add4-9809403abba6

### DIFF
--- a/app/javascript/components/Download/RichContent.tsx
+++ b/app/javascript/components/Download/RichContent.tsx
@@ -13,6 +13,7 @@ import { asyncVoid } from "$app/utils/promise";
 import { assertResponseError } from "$app/utils/request";
 
 import { Button, buttonVariants, NavigationButton } from "$app/components/Button";
+import { useDomains } from "$app/components/DomainSettings";
 import { FileRow, shouldShowSubtitlesForFile } from "$app/components/Download/FileList";
 import { License, useContentFiles } from "$app/components/DownloadPage/WithContent";
 import { LoadingSpinner } from "$app/components/LoadingSpinner";
@@ -41,15 +42,16 @@ export const RichContentView = ({
   saleInfo: SaleInfo | null;
   license: License | null;
 }) => {
+  const { rootDomain } = useDomains();
   const editor = useRichTextEditor({
     ariaLabel: "Product content",
     // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- to be fixed with product edit refactor
     initialValue: richContent as Content,
     editable: false,
     extensions: [
-      Link.configure({ saleInfo }),
-      TiptapLink.configure({ saleInfo }),
-      TiptapButton.configure({ saleInfo }),
+      Link.configure({ saleInfo, rootDomain }),
+      TiptapLink.configure({ saleInfo, rootDomain }),
+      TiptapButton.configure({ saleInfo, rootDomain }),
       FileEmbed,
       FileEmbedGroup,
       ExternalMediaFileEmbed,
@@ -76,17 +78,13 @@ export const RichContentView = ({
 
 const SALE_INFO_PLACEHOLDER_QUERY_PARAM = "__sale_info__";
 
-const isGumroadPostUrl = (url: URL) => {
-  const isGumroadDomain =
-    url.hostname.endsWith(".gumroad.com") ||
-    url.hostname === "gumroad.com" ||
-    url.hostname.endsWith(".gumroad.dev") ||
-    url.hostname === "gumroad.dev";
+const isGumroadPostUrl = (url: URL, rootDomain: string) => {
+  const isGumroadDomain = url.host.endsWith(`.${rootDomain}`) || url.host === rootDomain;
   const isPostPath = /^\/([^/]+\/)?p\/[^/]+$/u.test(url.pathname);
   return isGumroadDomain && isPostPath;
 };
 
-const addSaleInfoQueryParams = (href: string, saleInfo: SaleInfo | null) => {
+const addSaleInfoQueryParams = (href: string, saleInfo: SaleInfo | null, rootDomain: string) => {
   if (!saleInfo) return href;
 
   try {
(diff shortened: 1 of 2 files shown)


app/javascript/components/Download/RichContent.tsx
```diff
@@ -76,19 +76,36 @@ export const RichContentView = ({
...
+    if (url.searchParams.has(SALE_INFO_PLACEHOLDER_QUERY_PARAM)) {
+      url.searchParams.delete(SALE_INFO_PLACEHOLDER_QUERY_PARAM);
+      url.searchParams.set("sale_id", saleInfo.sale_id);
+      url.searchParams.set("product_id", saleInfo.product_id || "");
+      url.searchParams.set("product_permalink", saleInfo.product_permalink || "");
+      return url.href;
+    }
+
+    if (isGumroadPostUrl(url) && !url.searchParams.has("purchase_id")) {
+      url.searchParams.set("purchase_id", saleInfo.sale_id);
```
nyomanjyotisa, line 102: `saleInfo.sale_id` equals `purchase.external_id`, which is what `PostPresenter#e404?` checks to bypass visibility for non-account purchasers

app/javascript/components/Download/RichContent.tsx
```diff
@@ -76,19 +76,36 @@ export const RichContentView = ({
 
 const SALE_INFO_PLACEHOLDER_QUERY_PARAM = "__sale_info__";
 
+const isGumroadPostUrl = (url: URL) => {
+  const isGumroadDomain =
+    url.hostname.endsWith(".gumroad.com") ||
+    url.hostname === "gumroad.com" ||
+    url.hostname.endsWith(".gumroad.dev") ||
+    url.hostname === "gumroad.dev";
+  const isPostPath = /^\/([^/]+\/)?p\/[^/]+$/u.test(url.pathname);
```
binary-koan, line 83:
```suggestion
  const isPostPath = /^\/p\/[^/]+$/u.test(url.pathname);
```

Or when is the optional extra path segment set?

app/javascript/components/Download/RichContent.tsx
```diff
@@ -76,19 +76,36 @@ export const RichContentView = ({
 
 const SALE_INFO_PLACEHOLDER_QUERY_PARAM = "__sale_info__";
 
+const isGumroadPostUrl = (url: URL) => {
+  const isGumroadDomain =
```
binary-koan: Better to use `useDomains().rootDomain` rather than hardcoding gumroad.com and gumroad.dev

app/javascript/components/Download/RichContent.tsx
```diff
@@ -76,19 +76,36 @@ export const RichContentView = ({
...
-    url.searchParams.set("product_permalink", saleInfo.product_permalink || "");
+    if (url.searchParams.has(SALE_INFO_PLACEHOLDER_QUERY_PARAM)) {
+      url.searchParams.delete(SALE_INFO_PLACEHOLDER_QUERY_PARAM);
+      url.searchParams.set("sale_id", saleInfo.sale_id);
+      url.searchParams.set("product_id", saleInfo.product_id || "");
+      url.searchParams.set("product_permalink", saleInfo.product_permalink || "");
+      return url.href;
+    }
+
+    if (isGumroadPostUrl(url) && !url.searchParams.has("purchase_id")) {
```
binary-koan:
```suggestion
    if (isGumroadPostUrl(url)) {
```

Is there a case we wouldn't want to override it? (not major, but leaving that in makes it look like it'll be set to another purchase ID sometimes)

spec/requests/download_page/rich_text_editor_spec.rb
```diff
@@ -777,6 +777,81 @@ def embed_files_for_product(product:, files:)
       expect(find_link("Button with custom query parameters", href: "https://example.com/?test=123&#{sale_info_query_params}", target: "_blank")[:rel]).to eq("noopener noreferrer nofollow")
     end
 
+    it "appends purchase_id to Gumroad post URLs in links and buttons" do
+      post_url = "https://#{@user.username}.gumroad.com/p/test-post"
+      product_rich_content = @product.alive_rich_contents.first
+      product_rich_content.update!(description: [
+                                     { "type" => "paragraph",
```
binary-koan: Nit: I would find this more readable with the braces on separate lines (pretty-JSON style)

app/javascript/components/Download/RichContent.tsx
nyomanjyotisa: The optional segment covers the `/:username/p/:slug` route

https://github.com/antiwork/gumroad/blob/64bc87833308a519008bfff868de6ae1740a87f1/config/routes.rb#L985

app/javascript/components/Download/RichContent.tsx
nyomanjyotisa: Updated to use useDomains().rootDomain

app/javascript/components/Download/RichContent.tsx
nyomanjyotisa: Correct, removed this

spec/requests/download_page/rich_text_editor_spec.rb
nyomanjyotisa: Updated

app/javascript/components/Download/RichContent.tsx
binary-koan: Ah my bad, I missed that looking at the routes

PR #3870, opened 2026-03-03
Fix/test purchase subscription block

### LINKED ISSUE #3748: Multiple Test Purchases Blocked for Subscriptions
Unable to run multiple test purchases for subscriptions. This is likely a bug introduced by the transparent membership restarts feature. When running more than one test purchase, you get this message: 

Image

And this email: 

Image

The system currently does not differentiate between test subscriptions and real subscriptions, so multiple test purchases are treated as conflicting, which blocks the user. Update the subscription handling logic to clearly differentiate test subscriptions from regular subscriptions.

EmCousin:
Fixes #3748

## What

Added `.not_is_test_subscription` to all four lookup methods in `Subscription::Restartable` to exclude test subscriptions from subscription conflict checks:

- `restartable_for_product_and_buyer`
- `restartable_for_product_and_email`
- `active_for_product_and_buyer`
- `active_for_product_and_email`

This allows sellers to run multiple test purchases for memberships without being blocked by their previous test subscriptions.

## Why

The transparent membership restarts feature introduced subscription conflict detection, but it didn't differentiate between test and real subscriptions. When a seller ran a test purchase, any subsequent test purchase was blocked with "You already have an active subscription" — even though test subscriptions shouldn't prevent new test purchases.

The fix matches the existing `Subscription.active` scope behavior, which already excludes test subscriptions.

Credit to @yashranaway for identifying and proposing this fix. This PR ports that fix to the main repo and provides additional test cover, including system specs.

---

### AI Disclosure

### DIFF
--- a/spec/requests/checkout/test_subscription_purchases_spec.rb
+++ b/spec/requests/checkout/test_subscription_purchases_spec.rb
@@ -7,7 +7,6 @@ describe "Test subscription purchases", :js, type: :system do
   let(:product) { create(:product, user: seller, is_recurring_billing: true, subscription_duration: :monthly, price_cents: 500) }
 
   it "allows the seller to create multiple test purchases for a membership product" do
-    # First test purchase
     login_as seller
     visit product.long_url
     add_to_cart(product)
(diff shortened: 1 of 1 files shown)


spec/requests/checkout/test_subscription_purchases_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe "Test subscription purchases", :js, type: :system do
+  let(:seller) { create(:named_seller) }
+  let(:product) { create(:product, user: seller, is_recurring_billing: true, subscription_duration: :monthly, price_cents: 500) }
+
+  it "allows the seller to create multiple test purchases for a membership product" do
+    # First test purchase
```
gianfrancopiana: NIT: I think these step-by-step comments can be removed. The test names + assertions already make the intent clear.

gianfrancopiana (approved):

PR #3872, opened 2026-03-03
Fix popover auto-focus

### LINKED ISSUE #3373: Fix: Search input not auto-focusing when popover opens

gianfrancopiana:
Fixes #3373

## What

Removed `forceMount` from the base `PopoverContent` component. With `forceMount`, Radix kept popover content permanently in the DOM (hidden via CSS), which broke its built-in focus management. `onOpenAutoFocus` only fired once on initial mount, not on subsequent opens. Calling `preventDefault` on that event made sure the popover never auto-focused.

Without `forceMount`, Radix mounts/unmounts content naturally on open/close, and its default behavior. Focusing the first focusable element on open – works correctly for all 44 popover consumers.

Cleaned up redundant focus workarounds that consumers had added to work around the base component's broken behavior: manual ref + `onOpenAutoFocus` override in `Search.tsx`, a `useEffect` focus hack in `Affiliates/Index.tsx`, and an `autoFocus` attribute in `Admin/SearchPopover.tsx`.

## Why

This was a migration artifact from PR #2742 that was never revisited. The old popover had a `useEffect` that focused `[autofocus]` elements on open. The Radix rewrite replaced that with `onOpenAutoFocus={(e) => e.preventDefault()}` but didn't provide an alternative, so no popover auto-focused unless a consumer explicitly worked around it.

Removing `forceMount` is the simplest fix. All consumers use controlled React state, so none depend on content staying mounted when closed.

## Before
CleanShot 2026-03-03 at 13 43 37

## After
CleanShot 2026-03-03 at 13 46 43

---

This PR was implemented with AI assistance using Claude Opus 4.6 for code generation.

### DIFF
--- a/app/javascript/pages/Affiliates/Index.tsx
+++ b/app/javascript/pages/Affiliates/Index.tsx
@@ -82,13 +82,7 @@ const SearchBoxPopover = ({ initialQuery, onSearch }: { initialQuery: string; on
       <PopoverContent>
         <div className="input input-wrapper">
           <Search className="size-5 text-muted" />
-          <input
-            value={inputValue}
-            type="text"
-            placeholder="Search"
-            aria-label="Search"
-            onChange={handleChange}
-          />
+          <input value={inputValue} type="text" placeholder="Search" aria-label="Search" onChange={handleChange} />
         </div>
       </PopoverContent>
     </Popover>


binary-koan:
We used that workaround because not mounting breaks a bunch of specs (most of the failing slow ones look related). Could you check if they're easily fixable?

gianfrancopiana:
> We used that workaround because not mounting breaks a bunch of specs (most of the failing slow ones look related). Could you check if they're easily fixable?

That's good to know! Closing this one for now. I'll circle back with a more robust approach.

PR #3878, opened 2026-03-03
Increase size limit for processing files

binary-koan:
Issue: https://github.com/antiwork/gumroad-ios/issues/13

Supersedes https://github.com/antiwork/gumroad/pull/3706

# Description

## Problem

A creator's audio file cannot be played on iOS because the duration hasn't been filled in. The file is >1GB, and we currently skip processing files that large to work out the duration etc.

## Solution

Set the max size for all files to the current max video size - it shouldn't cause issues since we already allow it for videos and extremely few uploaded files are that large, but we should keep an eye on worker disk space.

---

# AI Disclosure

Claude Opus 4.6 used for identifying the issue.


binary-koan:
@nyomanjyotisa Could you review? I'll reprocess the affected file in production console once this is deployed (don't think it's worth reprocessing everything - it would be more likely to cause disk space issues and the number of products affected is very small - but lmk if you disagree)

binary-koan:
Good idea, will look into that!

PR #3880, opened 2026-03-04
Add close button to NestedMenu overlay sidebar

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad/issues/2724

## What
Added a close button (X icon) to the NestedMenu overlay sidebar, positioned on the dark overlay area to the right of the menu panel.

## Why
The close button was intentionally removed in #1851 when migrating to the `Sheet` component, with the reasoning that clicking the backdrop or pressing ESC would suffice ([comment](https://github.com/antiwork/gumroad/pull/1851/files#r2530005296)). However, we decide to re-add it for better UX https://github.com/antiwork/gumroad/issues/2724#issuecomment-3892885356

## Before
image
image

## After
image
image


app/javascript/components/NestedMenu.tsx
```diff
@@ -347,7 +351,7 @@ const OverlayMenu = ({
             setMenuOpen(false);
             onSelectItem?.(newSelectedItem, e);
           }}
-          className="h-full overflow-x-hidden overflow-y-auto"
+          className="bg-filled h-full w-80 overflow-x-hidden overflow-y-auto"
```
nyomanjyotisa, line 357: Added `bg-filled` and `w-80` here since the width/background constraint moved from the Sheet to the `ItemsList`

app/javascript/components/NestedMenu.tsx
```diff
@@ -331,8 +332,11 @@ const OverlayMenu = ({
         open={menuOpen}
         onOpenChange={setMenuOpen}
         modal
-        className="right-auto w-80 max-w-80 border-l-0 p-0 md:left-0 md:border-r"
+        className="bg-transparent p-0 pr-12 md:left-0 md:w-full md:border-l-0"
```
nyomanjyotisa: `bg-transparent` overrides the Sheet's base `bg-filled` so the overlay shows through. `pr-12` creates space for the close button to sit on the overlay area outside the menu panel

app/javascript/components/NestedMenu.tsx
```diff
@@ -331,8 +332,11 @@ const OverlayMenu = ({
...
       >
+        <Dialog.Close className="absolute top-4 right-4 z-40 bg-transparent" aria-label="Close Menu">
+          <X className="size-6 text-white" />
```
nyomanjyotisa, line 341: Re-adding the close button removed in #1851. Uses `Dialog.Close` (same pattern as [SheetHeader](https://github.com/antiwork/gumroad/blob/64bc87833308a519008bfff868de6ae1740a87f1/app/javascript/components/ui/Sheet.tsx#L31-L38)) instead of the original plain `<button>` with manual `onClick`

app/javascript/components/NestedMenu.tsx
gianfrancopiana: Making `Dialog.Content` full-viewport means every click on the dark overlay area is "inside" the content. Radix's `DismissableLayer` only fires `onPointerDownOutside` for clicks outside the content element, so clicking the backdrop no longer closes the menu – only ESC and the X button work.

For context: the original implementation (pre-#1851) also didn’t support click-to-close on the backdrop – that was introduced as a side effect of migrating to the Sheet component. So this isn't a regression from the original behavior, but it is a regression from the current behavior users have gotten used to.

To preserve both the X button and click-to-close, consider constraining the Sheet width instead of making it full-viewport:
```
className="bg-transparent p-0 pr-12 w-[calc(20rem+3rem)] md:left-0 md:border-l-0"
```

app/javascript/components/NestedMenu.tsx
gianfrancopiana: Missing `cursor-pointer` and `all-unset`. The other `Dialog.Close` in the codebase use both:

- `Sheet.tsx:34` — `className="ml-auto cursor-pointer all-unset"`
- `Modal.tsx:42` — `className="cursor-pointer text-base all-unset"`

There's no global `cursor: pointer` on `<button>` elements, so without it the X shows a default arrow cursor on hover. And without `all-unset`, the button retains default browser styling (borders, padding, etc.) which may cause subtle rendering issues.

app/javascript/components/NestedMenu.tsx
gianfrancopiana: Making `Dialog.Content` full-viewport (`bg-transparent` + `inset-0` from the base Sheet, no width constraint) means every click on the dark overlay area is "inside" the content. Radix's `DismissableLayer` only fires `onPointerDownOutside` for clicks outside the content element, so clicking the backdrop no longer closes the menu — only ESC and the X button work.

For context: the original implementation (pre-#1851) also didn't support click-to-close on the backdrop – that was introduced as a side effect of migrating to the Sheet component. So this isn't a regression from the original behavior, but it is a regression from the current behavior users have gotten used to.

To preserve both the X button and click-to-close, consider constraining the Sheet width instead of making it full-viewport:
```tsx
className="bg-transparent p-0 pr-12 w-[calc(20rem+3rem)] md:left-0 md:border-l-0"
```

gianfrancopiana: One more thing! 

`absolute right-4` positions the X relative to the viewport edge, not the panel edge. `OverlayMenu` renders up to the `lg` breakpoint (1024px). 

So on a tablet the X floats hundreds of pixels away from the 320px panel:
CleanShot 2026-03-05 at 15 27 01@2x

The original SCSS had `max-width: calc(100vw - ($big-icon-size + 2 * spacer(4)))` on the menu panel to prevent this.

If the Sheet width is constrained (per above), `right-4` would naturally sit adjacent to the panel and the issue should go away.

app/javascript/components/NestedMenu.tsx
nyomanjyotisa: Added both `cursor-pointer` and `all-unset` to match the `Dialog.Close` pattern in `Sheet.tsx` and `Modal.tsx`

app/javascript/components/NestedMenu.tsx
nyomanjyotisa: Constrained the Sheet width to `w-[calc(20rem+3rem)]` as suggested, this preserves backdrop click-to-close and keeps the X adjacent to the panel edge on tablets

Issue #3882, opened 2026-03-04
Migrate product edit page from React on Rails to Inertia


gianfrancopiana:
## What

Migrate the product edit page (`/products/:id/edit`) from React on Rails to Inertia. Same UX, same tabs, same save behavior – just a different page mount.

The approach:
- Add paired public-beta routes at `/products/:id/edit_new` and `/products/:id/edit_new/*other`
- Add a new `edit_new` action that reuses the same product-fetch/authorization/presenter setup as `edit`, but renders via Inertia
- Old `/edit` stays live. Both routes accessible to all creators (public beta)
- Keep React Router for client-side tabs, keep `ProductEditContext`, keep global save
- When `edit_new` has full feature parity (all tests passing), swap it to be the default `/edit`
- Clean up: remove old ERB template, remove React on Rails dependency for this page

Non-goals for this issue: no autosave, no redesign, no new save API flow.

The core change is mounting the same product-edit React UI through an Inertia page (`render inertia:`) on `edit_new`, while keeping the existing ERB + React on Rails mount for `/edit` during beta.

### Known integration work for the scaffold

The ERB template currently does three things beyond mounting the component:
1. **`load_pack "product_edit"`** — Inertia resolves JS bundles differently (via page component resolver in `app/javascript/pages/`). The component must be registered there.
2. **`render("shared/load_dropbox_dropins_js")`** — Dropbox JS partial won't carry over. Needs to be loaded in the React component or the Inertia layout.
3. **`prerender: true`** — this is React on Rails SSR. It will not apply to the Inertia mount; verify acceptable behavior for `/edit_new`.

Additionally:
4. **React Router paths are hardcoded** to `/products/:id/edit`. Must be updated for beta route support in both `ProductEditPage.tsx` and `ProductEdit/Layout.tsx`, while keeping `/edit` fully working until final cutover.
5. **Layout declaration** — add `layout "inertia"` for the new action in `LinksController`.
6. **Save endpoint is unaffected** — `saveProduct()` uses `Routes.link_path(permalink)`, completely independent of the current URL.
7. **Sub-route refresh support** — `edit_new` needs both base and wildcard routes so `/edit_new/content`, `/share`, `/receipt` work on direct load/refresh.
8. **Spec path references are mixed** — migration must handle both route helpers and hardcoded `"/products/#{...}/edit"` strings (including `#checkout` hash paths).

### Validation strategy

Scaffold `edit_new` first, then migrate tests incrementally so every PR stays green.

[...]

Issue #3764, opened 2026-02-23
Remove React on Rails dependency


EmCousin:
When the Inertia migration is over (https://github.com/antiwork/gumroad/issues/3763 and #3028), React on Rails will no longer be used and therefore should be removed

> [!NOTE]
> React on Rails is currently used in the ERB application layout to render common components like sidebars and flash messages.     This means a full migration of the app to Inertia is necessary before we can remove this dependency.

slavingia:
I don't think we need to do all ERB for this to get merged in, just the ones that use react on rails. The homepage, for example, doesn't need to be part of an SPA at all.

EmCousin:
@slavingia  the home pages do depend on React on Rails, but indirectly via layout inheritance. The home layout extends `layouts/application.html.erb`, which loads the React on Rails bundle (javascript_include_tag "application") and its remaining components (Alert, Nav). That means this issue is blocked until those components are effectively migrated to Inertia, but since they are part of the layout, we need to migrate the templates inheriting it.

Something else, the home page has multiple entry points to Inertia-powered pages:
    - Discovery carousel 👉  `/discover` (Inertia)
    - "Start selling" button 👉 `/signup` (Inertia)
    - Search form 👉 `/discover?query=`... (Inertia)
When users click these links from a non-Inertia page, they get a full page reload instead of a smooth SPA transition. Migrating the home pages to Inertia would make the experience smoother and faster.

Would appreciate your thoughts on this!

slavingia:
> The home layout extends layouts/application.html.erb, which loads the React on Rails bundle (javascript_include_tag "application") and its remaining components (Alert, Nav)

Can we solve these differently, e.g by having some duplicated code?

slavingia:
Think that should work!

PR #3888, opened 2026-03-04
Scaffold Inertia mount for product edit page at /edit_new

gianfrancopiana:
Toward #3882

## What

Adds a parallel `/products/:id/edit_new` route that renders the existing product edit UI through Inertia, while `/edit` continues to work via React on Rails.

The Inertia page component is a thin wrapper that passes props to the existing `ProductEditPage` component. React Router routes are made dynamic via `buildRoutes(editSegment)` so both `/edit` and `/edit_new` work. Dropbox JS is loaded client-side via the existing `useDropbox` hook instead of the ERB partial.

**CI validation** — a separate [`-ci` branch](https://github.com/antiwork/gumroad/actions/runs/22679415787) ran all 265+ existing edit system tests against `/edit_new`. Only 1 expected failure: the products index still links to `/edit` (route swap hasn't happened yet). All other failures were pre-existing flaky tests. Full parity confirmed.

## Why

The product edit page is the last major page on React on Rails. A parallel route at `/edit_new` lets us validate parity in production before swapping. The final cutover (swap routes, remove ERB/pack) will be a separate PR.

---
This PR was implemented with AI assistance using Claude Opus 4.6 and gpt-5.3-codex xhigh.


app/controllers/links_controller.rb
```diff
@@ -295,6 +295,19 @@ def edit
     @presenter = ProductPresenter.new(product: @product, pundit_user:, ai_generated:)
   end
 
+  def edit_new
+    fetch_product_by_unique_permalink
+    authorize @product
+
+    return redirect_to edit_bundle_product_path(@product.external_id) if @product.is_bundle?
```
gianfrancopiana, line 302: The existing `edit` action doesn't use `return` here because it relies on implicit ERB rendering. `edit_new` has an explicit `render inertia:` below, so without `return` this would raise `DoubleRenderError`.

app/presenters/product_presenter.rb
```diff
@@ -264,6 +264,7 @@ def edit_props
         fine_print: product.user.refund_policy.fine_print,
       },
       cancellation_discounts_enabled: Feature.active?(:cancellation_discounts, product.user),
+      dropbox_api_key: DROPBOX_PICKER_API_KEY,
```
gianfrancopiana, line 267: The ERB template loads Dropbox JS via `render("shared/load_dropbox_dropins_js")`, which isn't available in the Inertia path. Instead, the key is passed as a prop so the Inertia page can call `useDropbox(dropbox_api_key)` client-side.

binary-koan:
So this is the complete migration, right? (i.e. enough to remove the React On Rails dependency)

If it's this simple I think we can just migrate in-place rather than creating a temporary route. Or do you think it's worth testing with production products to make sure everything still works? (cc @ershad @slavingia)

gianfrancopiana:
> So this is the complete migration, right? (i.e. enough to remove the React On Rails dependency)
> 
> If it's this simple I think we can just migrate in-place rather than creating a temporary route. Or do you think it's worth testing with production products to make sure everything still works? (cc @ershad @slavingia)

Yes, this is the complete migration for the product edit page. The [CI validation branch](https://github.com/antiwork/gumroad/actions/runs/22679415787) confirmed full parity across all 265+ existing edit system tests against `/edit_new`.

I'd still prefer landing this as-is and doing the cleanup (swap routes, remove ERB/pack/SSR registration) in a separate PR. That way each PR is focused on one thing, and we get a window to test with real production products before committing. The cleanup PR will be small.

Note this doesn't fully remove the React on Rails dependency. Nav and Alert still use `react_component` in the ERB layout. That's tracked in #3764.

app/javascript/pages/Products/Edit.tsx
```diff
@@ -0,0 +1,20 @@
+import { usePage } from "@inertiajs/react";
```
binary-koan, line 1: This should be `pages/Products/Edit.tsx` for consistency

spec/presenters/product_presenter_spec.rb
```diff
@@ -457,7 +457,7 @@
           successful_sales_count: 0,
           ratings: {
             count: 0,
-            average: 0,
+            average: 0.0,
```
binary-koan, line 460: Any reason for this? (and similar changes in the rest of the file)

gianfrancopiana: This was a pre-existing mismatch. The presenter returns a Float but the spec expected an Integer. It was surfaced by the CI diff when adding `dropbox_api_key` (both differences showed up in the same failure output). Not related to our changes, just fixed it while I was there.

slavingia:
Awesome

PR #3889, opened 2026-03-04
Fix popover auto-focus by removing forceMount

### LINKED ISSUE #3373: Fix: Search input not auto-focusing when popover opens

gianfrancopiana:
Fixes #3373
Supersedes #3872 (closed)

## What

Removed `forceMount` from the base `PopoverContent` component. With `forceMount`, Radix kept popover content permanently in the DOM (hidden via CSS), which broke its built-in focus management. `onOpenAutoFocus` only fired once on initial mount, not on subsequent opens.

Without `forceMount`, Radix mounts/unmounts content naturally on open/close, and its default behavior (focusing the first focusable element on open) works correctly for all popover consumers.

This required fixing several things that silently depended on `forceMount`:

- **Insert menu modals (X post, Button, Link, Video):** The `WithDialog` components held modal state locally inside popover content. When `PopoverClose` unmounts content, that state was destroyed before the modal could render. Removed `WithDialog` entirely and lifted modal state to `RichTextEditor`, following the existing Upsell/Review pattern. This also fulfills the TODO comments: *"remove this once popovers no longer use details"*.

- **File upload input:** The `<input type="file">` was inside `PopoverClose` in `ContentTab`. Moved it outside the popover using a ref so it persists in the DOM when the popover closes.

- **Test disclosure helper:** Radix may re-render popover content after initial mount (for positioning), and `PopoverClose` items unmount content on click. Added retry logic to `select_disclosure`/`toggle_disclosure` that distinguishes between these two cases: retries if the block hasn't executed yet (positioning re-render), rescues silently if it has (popover closed after action completed).

Cleaned up redundant focus workarounds that consumers had added: manual ref + `onOpenAutoFocus` override in `Search.tsx`, a `useEffect` focus hack in `Affiliates/Index.tsx`, and an `autoFocus` attribute in `Admin/SearchPopover.tsx`.

## Why

This was a migration artifact from PR #2742 that was never revisited. The old popover had a `useEffect` that focused `[autofocus]` elements on open. The Radix rewrite replaced that with `onOpenAutoFocus={(e) => e.preventDefault()}` but didn't provide an alternative, so no popover auto-focused unless a consumer explicitly worked around it.

Removing `forceMount` is the simplest fix. All consumers use controlled React state, so none depend on content staying mounted when closed.

## Before
CleanShot 2026-03-03 at 13 43 37

## After
CleanShot 2026-03-03 at 13 46 43

---

This PR was implemented with AI assistance using Claude Opus 4.6

### DIFF
--- a/app/javascript/components/ProductEdit/ContentTab/index.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/index.tsx
@@ -35,6 +35,7 @@ import { fetchDropboxFiles, ResponseDropboxFile, uploadDropboxFile } from "$app/
 import { type Post } from "$app/types/workflow";
 import { escapeRegExp } from "$app/utils";
 import { assertDefined } from "$app/utils/assert";
+import { classNames } from "$app/utils/classNames";
 import { formatDate } from "$app/utils/date";
 import FileUtils from "$app/utils/file";
 import GuidGenerator from "$app/utils/guid_generator";
@@ -1193,14 +1194,17 @@ export const ContentTab = () => {
                             }}
                             aria-selected={item.id === selectedVariantId}
                             inert={product.has_same_rich_content_for_all_variants}
-                            className={product.has_same_rich_content_for_all_variants ? "opacity-30" : undefined}
+                            className={classNames(
+                              props.className,
+                              product.has_same_rich_content_for_all_variants ? "opacity-30" : undefined,
+                            )}
                           >
                             <div className="flex-1">
                               <h4>{item.name || "Untitled"}</h4>
                               {item.id === selectedVariant?.id ? (
-                                <small>Editing</small>
+                                <small className="block">Editing</small>
                               ) : product.has_same_rich_content_for_all_variants || item.rich_content.length ? (
-                                <small>
+                                <small className="block">
                                   Last edited on{" "}
                                   {formatDate(
                                     (product.has_same_rich_content_for_all_variants
@@ -1213,7 +1217,7 @@ export const ContentTab = () => {
                                   )}
                                 </small>
                               ) : (
-                                <small className="text-muted">No content yet</small>
+                                <small className="block text-muted">No content yet</small>
                               )}
                             </div>
                             {item.id === selectedVariant?.id && (
@@ -1231,7 +1235,7 @@ export const ContentTab = () => {
                                     setHasSameRichContent(!product.has_same_rich_content_for_all_variants);
                                   }}
                                 />
-                                <small>Use the same content for all versions</small>
+                                <small className="block">Use the same content for all versions</small>
                               </Label>
                             </div>
                           ) : null}
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,22 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(git *)",
+      "Bash(gh *)",
+      "Bash(bundle *)",
+      "Bash(yarn *)",
+      "Bash(npm *)",
+      "Bash(npx *)",
+      "Bash(rails *)",
+      "Bash(rake *)",
+      "Bash(rspec *)",
+      "Bash(rubocop *)",
+      "Bash(ls *)",
+      "Bash(mkdir *)",
+      "Bash(rm *)",
+      "Bash(mv *)",
+      "Bash(cp *)",
+      "Bash(sleep *)"
+    ]
+  }
+}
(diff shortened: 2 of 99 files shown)


spec/support/aria_extensions.rb
```diff
@@ -265,6 +265,54 @@ def position_considering_colspan(xpath)
...
+    def _run_in_disclosure(name, **find_options, &block)
+      attempts = 0
+      begin
+        attempts += 1
+        disclosure = if is_a?(Capybara::Node::Element) && name.nil?
+                       _locate_disclosure(name, **find_options)
+                     else
+                       Capybara.page.find(:disclosure, name, **find_options)
+                     end
+        block_executed = false
```
gianfrancopiana: Two different `StaleElementReferenceError` scenarios here:
1. Radix re-renders the popover for positioning right after mount, invalidating the `within` scope before the block runs. We retry with a fresh reference.
2. `PopoverClose` unmounts content *after* the block's click already happened. Safe to swallow. The flag tells them apart.

app/javascript/components/TiptapExtensions/Link.tsx
```diff
@@ -319,34 +295,42 @@ const TiptapButton = Node.create({
...
-export const LinkMenuItem = ({ editor }: { editor: Editor }) => (
-  <WithDialog editor={editor} type="link">
-    <MenuItem
-      name="Insert link"
-      icon={<LinkIcon className="size-5" />}
-      active={editor.isActive("link") || !!editor.getAttributes("image").link}
-    />
-  </WithDialog>
-);
+export const LinkMenuItem = ({ editor }: { editor: Editor }) => {
```
gianfrancopiana, line 313: `LinkMenuItem` manages its own dialog state because it lives in the toolbar (always mounted). Button and X post can't do this. They're rendered inside the insert menu popover, so `PopoverClose` would destroy their state before the modal opens. Those go through `openDialogForExtension` in `RichTextEditor` instead.

app/javascript/components/ProductEdit/ContentTab/index.tsx
```diff
@@ -141,11 +141,10 @@ const FileUploadMenu = ({
       </div>
     </PopoverClose>
     <PopoverClose asChild>
-      <label role="menuitem">
-        <input type="file" name="file" className="sr-only" multiple onChange={(e) => onUploadFile(e.target)} />
+      <div role="menuitem" onClick={onClickComputerFiles}>
```
gianfrancopiana, line 146: The file input was previously inside `PopoverClose`. After file selection, the popover closed and destroyed the input before `onChange` fired. Moved it to the component root (line 565) with a ref so it stays alive.

app/javascript/components/TiptapExtensions/MediaEmbed.tsx
```diff
@@ -34,10 +32,9 @@ declare module "@tiptap/core" {
 const MEDIA_EMBED_SUPPORTING_PROVIDERS = ["YouTube", "Vimeo", "Wistia, Inc.", "Dailymotion"];
 
 const VideoEmbed = Extension.create({
-  menuItem: (editor) => (
-    <WithDialog editor={editor} type="embed">
-      <MenuItem name="Insert video" icon={<Video className="size-5" />} />
-    </WithDialog>
+  name: "videoEmbed",
```
gianfrancopiana, line 35: Was unnamed before. Needed now so `openDialogForExtension` can map extension name → dialog opener.

app/javascript/components/Popover.tsx
```diff
@@ -53,11 +53,8 @@ export const PopoverContent = React.forwardRef<
         ref={ref}
         align={align}
         collisionPadding={collisionPadding}
-        autoFocus={false}
-        forceMount
-        onOpenAutoFocus={(e: Event) => e.preventDefault()}
         className={classNames(
-          "z-30 w-max max-w-[calc(100vw-2rem)] rounded-sm border border-border bg-background p-4 text-foreground shadow outline-none focus-visible:outline-none data-[state=closed]:hidden",
```
gianfrancopiana, line 60: With `forceMount`, content stayed in DOM and needed CSS to hide. Without it, Radix unmounts naturally and the class is no longer needed.

binary-koan:
Thanks for the reminder - there are still some related failing tests (e.g. missing "Search products" input in Slow 0) which I was planning to just quickly fix myself, but haven't found time. Will get to it after the Tailwind migration, or lmk if you can fix them earlier

gianfrancopiana:
@binary-koan Can you please review?

Issue #3835, opened 2026-02-27
Fix remaining flaky tax checkout tests


gianfrancopiana:
## What

`taxes_spec.rb` has 17 tests that assert tax amounts inside `check_out` blocks before the purchase is submitted. These depend on an async tax calculation completing within Capybara's wait timeout — the same pattern that caused flakes in #3801 (partially fixed for WA in #3822 and Canada in #3823).

The WI test at `taxes_spec.rb:242` already flaked in CI ([run 22453128162](https://github.com/antiwork/gumroad/actions/runs/22453128162)).

## Affected tests

All line numbers are on `main` as of the #3824 merge.

### US states — 4 tests

| State | Test | Line |
|-------|------|------|
| WI | calculates and charges sales tax when WI customer makes purchase | 166 |
| WI | calculates and charges sales tax when WI customer makes purchase of a physical product | 186 |
| WI | calculates and charges sales tax when WI customer purchases a non-physical product | 242 |
| WA | calculates and charges sales tax when WA customer makes purchase | 206 |

### AZ — 3 tests

| Test | Line |
|------|------|
| calls the tax endpoint for a real zip code that doesn't show in the enterprise zip codes database | 13 |
| re-evaluates price and tax when there are variants | 44 |
| re-evaluates price and tax when an offer code is applied - code in url | 83 |

### Canada — 4 tests

| Province | Test | Line |
|----------|------|------|
| BC | allows the customer to select province for a physical product to Canada | 4016 |
| Canada | charges tax when Canada is selected but not detected from IP | 4068 |
| QC | allows the entry of QST ID and doesn't charge tax | 4093 |
| QC | charges tax and does not collect the QST ID if the QST ID is invalid | 4128 |

### International (negative assertions) — 6 tests

| Country | Test | Line |
|---------|------|------|
| Netherlands | allows entry of VAT ID and doesn't charge VAT | 311 |
| Netherlands | allows entry of VAT ID and doesn't charge VAT (tiered membership) | 350 |
| Australia | allows entry of ABN ID and doesn't charge GST | 459 |
| Singapore | allows entry of GST ID and doesn't charge GST | 578 |
| Norway | allows entry of MVA ID and doesn't charge tax | 715 |
| Oman | allows entry of VAT Number and doesn't charge VAT | 2949 |

## Fix

[...]

PR #3892, opened 2026-03-04
Remove flaky pre-submit tax assertions from checkout tests

gianfrancopiana:
Issue #3835

## What

Two changes:

1. **Remove pre-submit tax UI assertions** from 9 `check_out` blocks in `taxes_spec.rb` and `shipping_physical_preorder_spec.rb`. These wait for async tax calculations to render before clicking "Pay" – the same flake pattern fixed for WA in #3822 and Canada in #3823. Every test already has post-purchase DB assertions that verify the same tax amounts.

2. **Allow VCR cassette replay for JS-only taxjar tests.** In system tests, both the frontend surcharge endpoint and the backend purchase creation call TaxJar. The single-interaction cassette gets consumed by the first call, so the second silently falls back to zero tax. Scoped to JS tests only (`example.metadata[:js]`) so non-JS tests keep strict replay.

## Why

Test Slow 13 keeps failing on main ([run 22681108696](https://github.com/antiwork/gumroad/actions/runs/22681108696)):

1. `shipping_physical_preorder_spec.rb:74` — async tax calc timing
2. `taxes_spec.rb:4067` — VCR cassette exhaustion causing zero tax

International negative assertions (VAT/GST ID tests) are left as-is since checking for absence of text is less timing-sensitive.

---

This PR was implemented with AI assistance using Claude Opus 4.6 and gpt-5.3-codex high

### DIFF
--- a/spec/support/checkout_helpers.rb
+++ b/spec/support/checkout_helpers.rb
@@ -148,13 +148,7 @@ module CheckoutHelpers
   def check_out(product, error: nil, email: "test@gumroad.com", is_free: false, gift: nil, sca: nil, should_verify_address: false, cart_item_count: 1, logged_in_user: nil, **params, &block)
     fill_checkout_form(product, email:, is_free:, logged_in_user:, gift:, **params)
 
-    if block_given?
-      # Wait for async surcharge calculation to complete before running pre-submit assertions.
-      # The Pay button is disabled while surcharges are loading (surcharges.type !== "loaded"),
-      # so waiting for it to be enabled ensures tax/total amounts are rendered.
-      expect(page).to have_button("Pay", disabled: false, wait: 10) unless is_free
-      block.call
-    end
+    block.call if block_given?
 
     expect do
       click_on is_free ? "Get" : "Pay", exact: true


hchhabra (commented):
> Every test already has post-purchase DB assertions that verify the same tax amounts.

These front-end checks are probably to verify that the customer is shown the correct Sales Tax and Total amounts that are going to be charged. Possible to keep these and fix the flakiness?

> When the async tax surcharge API call hasn't completed within Capybara's wait timeout, the assertion fails.

With the TaxJar API call recorded in the VCR cassette, the surcharge calculation action should ideally be quick and not cause timeout. If the flakiness is due to the TaxJar API call, maybe we can stub it and keep the front-end assertions? @gianfrancopiana _a

gianfrancopiana:
> > Every test already has post-purchase DB assertions that verify the same tax amounts.
> 
> These front-end checks are probably to verify that the customer is shown the correct Sales Tax and Total amounts that are going to be charged. Possible to keep these and fix the flakiness?
> 
> > When the async tax surcharge API call hasn't completed within Capybara's wait timeout, the assertion fails.
> 
> With the TaxJar API call recorded in the VCR cassette, the surcharge calculation action should ideally be quick and not cause timeout. If the flakiness is due to the TaxJar API call, maybe we can stub it and keep the front-end assertions? @gianfrancopiana _a

Makes sense! Pushed a new approach: instead of removing the assertions, I added a sync point in `check_out` that waits for the Pay button to be enabled before running the block. The Pay button is disabled while surcharges are loading (`surcharges.type !== "loaded"` in `isSubmitDisabled`), so this ensures tax/total amounts are rendered before we assert on them.

All pre-submit assertions are restored.

For the VCR issue: in JS tests, TaxJar gets called twice – once from the frontend surcharge endpoint, and again from the backend during purchase creation. The single-interaction cassette gets consumed by the first call, so the second one silently returns zero tax. Added `allow_playback_repeats` to both `:taxjar` and `:shipping` VCR cassettes (scoped to JS tests only) so the same recorded response can be replayed.

spec/support/checkout_helpers.rb
```diff
@@ -148,7 +148,13 @@ def fill_checkout_form(product, email: "test@gumroad.com", address: nil, offer_c
   def check_out(product, error: nil, email: "test@gumroad.com", is_free: false, gift: nil, sca: nil, should_verify_address: false, cart_item_count: 1, logged_in_user: nil, **params, &block)
     fill_checkout_form(product, email:, is_free:, logged_in_user:, gift:, **params)
 
-    block.call if block_given?
+    if block_given?
+      # Wait for async surcharge calculation to complete before running pre-submit assertions.
+      # The Pay button is disabled while surcharges are loading (surcharges.type !== "loaded"),
+      # so waiting for it to be enabled ensures tax/total amounts are rendered.
+      expect(page).to have_button("Pay", disabled: false, wait: 10) unless is_free
```
hchhabra: Not really sure if adding this would fix the flakiness. Would the existing `have_text` assertions present in the `block` not do the same wait for capybara default timeout which is 25 seconds? @gianfrancopiana _a

gianfrancopiana: You're right, the extra wait was redundant.

I dropped that helper change and kept the JS-scoped `allow_playback_repeats` change. I also tested the VCR-only version on a separate CI branch: [run 22865713190](https://github.com/antiwork/gumroad/actions/runs/22865713190). The relevant lanes passed (`Test Slow 13` and `Test Fast 13`), and the only failure was an unrelated `Wait for services` timeout in `Test Slow 8`.

So this branch keeps the VCR replay fix and removes the shared helper wait.

hchhabra (approved):

PR #3891, opened 2026-03-04
Migrate `_forms.scss` to Tailwind: Part 4

binary-koan:
From @SalmanNajah: https://github.com/SalmanNajah/gumroad/pull/1

Issue: https://github.com/antiwork/gumroad/issues/3883

# Description
- Migrate remaining `_forms.scss` styles to Tailwind
- Continuation of #3809 
- Part of #2747 

---

# Demo 
## Analytics page
|Before|After|
|-|-|
|image | Screenshot 2026-03-01 at 1 30 00 AM
| Screenshot 2026-03-04 at 10 25 58 PM| Screenshot 2026-03-01 at 1 30 08 AM

## Workflows page
|Before|After|
|-|-|
|image | image
|image | image
|Screenshot 2026-03-04 at 10 29 18 PM |Screenshot 2026-03-01 at 2 58 01 PM
| Screenshot 2026-03-04 at 10 29 52 PM| Screenshot 2026-03-01 at 2 57 55 PM

## Affiliates page
|Before|After|
|-|-|
| image| Screenshot 2026-03-01 at 3 43 20 PM
| image| Screenshot 2026-03-01 at 3 43 28 PM
|image | Screenshot 2026-03-01 at 3 38 33 PM
|image | Screenshot 2026-03-01 at 3 43 37 PM
|image | Screenshot 2026-03-01 at 3 43 56 PM
| image| Screenshot 2026-03-01 at 3 44 04 PM

## Library page
|Before|After|
|-|-|
| image| Screenshot 2026-03-01 at 4 13 53 PM
| image| Screenshot 2026-03-01 at 4 14 00 PM

## Sales page
|Before|After|
|-|-|
|image | Screenshot 2026-03-01 at 1 09 35 PM
| image| Screenshot 2026-03-01 at 1 09 26 PM

## Collaborators page
|Before|After|
|-|-|
| image| Screenshot 2026-03-01 at 1 09 52 PM
| image| Screenshot 2026-03-01 at 1 09 45 PM

## Affiliated page
|Before|After|
|-|-|
| image| Screenshot 2026-03-01 at 1 25 28 AM
| image| Screenshot 2026-03-01 at 1 25 52 AM

## UTM Links page
|Before|After|
|-|-|
| image | Screenshot 2026-03-01 at 2 57 34 PM
| image| Screenshot 2026-03-01 at 2 57 41 PM

## Emails page
|Before|After|
|-|-|
| image| image
| image| image

---

# Checklist

[...]

PR #3893, opened 2026-03-05
Forms styling fixes

binary-koan:
Follow-up to #3891 

Minor styling fixes after the Tailwind forms and icons migrations, restoring the previous appearance in a few places.


binary-koan:
Effectively just more review comments on the previous PR, going ahead and merging

PR #3898, opened 2026-03-06
Send Top Creator badge email when admin verifies a creator

### LINKED ISSUE #3890: Automatically send Top Creator badge email when a creator is verified
## What

When an admin marks a creator as verified via the admin panel, automatically send the "Top Creator" badge announcement email (`CreatorMailer#top_creator_announcement`).

Currently the `verify` action in the admin users controller toggles `verified` and saves – it doesn't trigger any notification. The mailer exists (added in #3834) but is only used for the one-time batch send to existing verified creators.

- [ ] Detect when `verified` changes from `false` to `true` in `Admin::UsersController#verify`
- [ ] Enqueue `CreatorMailer.top_creator_announcement` for the user (via `deliver_later`)
- [ ] Do nothing when un-verifying (toggling from `true` to `false`)

## Why

New creators who earn the Top Creator badge should learn about it right away. Same as the existing verified creators who were notified via the batch email. Without this, newly verified creators would silently get the badge with no notification.

Related: antiwork/gumroad-private#160.

gianfrancopiana:
Fixes #3890

## What

When an admin verifies a creator via the admin panel, the `CreatorMailer#top_creator_announcement` email is now automatically enqueued. This only fires when toggling verified from false/nil to true. Un-verifying does not send an email.

The error handling was also restructured: the `save!` and `deliver_later` calls have separate rescue blocks so that a job enqueue failure (e.g., Redis outage) is logged but doesn't return `success: false` to the admin when the verification itself succeeded.

## Why

The Top Creator badge email (added in #3834) was only used for the one-time batch send to existing verified creators. Newly verified creators would silently receive the badge with no notification. This closes the gap so every creator learns about their badge the moment they earn it.

---

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -28,17 +28,17 @@ class Admin::UsersController < Admin::BaseController
   end
 
   def verify
-    newly_verified = !@user.verified
-    @user.verified = newly_verified
+    @user.verified = !@user.verified
     begin
       @user.save!
     rescue => e
       return render json: { success: false, message: e.message }
     end
-    if newly_verified
+    if @user.verified
       begin
         CreatorMailer.top_creator_announcement(user_id: @user.id).deliver_later
       rescue => e
+        Bugsnag.notify(e)
         Rails.logger.error("Failed to enqueue top_creator_announcement for user #{@user.id}: #{e.message}")
       end
     end
--- a/spec/controllers/admin/users_controller_spec.rb
+++ b/spec/controllers/admin/users_controller_spec.rb
@@ -50,6 +50,21 @@ describe Admin::UsersController, type: :controller, inertia: true do
       end.not_to have_enqueued_mail(CreatorMailer, :top_creator_announcement)
     end
 
+    context "when email enqueue fails" do
+      before do
+        mail_double = double("mail")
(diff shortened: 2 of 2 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -28,11 +28,21 @@ def refund_balance
...
-    @user.verified = !@user.verified
-    @user.save!
+    newly_verified = !@user.verified
+    @user.verified = newly_verified
+    begin
+      @user.save!
+    rescue => e
+      return render json: { success: false, message: e.message }
+    end
+    if newly_verified
```
binary-koan:
```suggestion
    if @user.verified
```

It's not clear to me why we need a separate variable?

app/controllers/admin/users_controller.rb
```diff
@@ -28,11 +28,21 @@ def refund_balance
...
+      begin
+        CreatorMailer.top_creator_announcement(user_id: @user.id).deliver_later
+      rescue => e
+        Rails.logger.error("Failed to enqueue top_creator_announcement for user #{@user.id}: #{e.message}")
```
binary-koan, line 42: Worth logging this to Bugsnag as well

gianfrancopiana:
@binary-koan _a Dropped `newly_verified` in favor of `@user.verified`, added `Bugsnag.notify`, and added a test that the verify action still succeeds even if the email fails to enqueue.

Issue #3789, opened 2026-02-23
Provide informational 1099-K forms to all US sellers who made a sale in the previous calendar year


hchhabra:
### What

- [x] Generate a CSV with 1099-K data for all US sellers who made a sale in the calendar year
- [x] Import the CSV to Stripe to update all 1099-K forms
- [x] Bulk download all updated 1099-K draft forms from Stripe and upload them to S3
- [x] Update the Tax Center page - https://github.com/antiwork/gumroad/pull/3899
  - [x] Show non-filed/informational forms in the Tax Center
  - [x] Display filing status `Filed with IRS on {date}` / `Informational only (Not filed with IRS)`
  - [x] Support downloading non-filed forms from S3
- [x] Update 1099-K notification email copy for informational forms - https://github.com/antiwork/gumroad/pull/3899
- [x] Create `UserTaxForm` records for all informational forms
- [x] Send notification emails for informational 1099-K forms to all sellers
- [ ] Document all steps in a new file in `docs/` dir

### Why

Currently 1099-K forms are only provided to sellers who are above IRS' filing threshold, but these informational forms are useful for all US sellers.

PR #3899, opened 2026-03-06
Support downloading informational 1099-K forms from the Tax Center page

hchhabra:
### What

- Support downloading informational 1099-K forms from the Tax Center for sellers who are below the IRS filing threshold
- Show filing status for 1099 forms in the Tax Center — `Filed with IRS on {date}` or `Informational only (Not Filed with IRS)`
- Update the 1099-K notification email to reflect whether the form was filed with the IRS or is informational only

### Why

- Part of https://github.com/antiwork/gumroad/issues/3789
- To provide informational 1099-K forms to all US sellers who made a sale in the previous calendar year, not just those who exceed the IRS filing threshold. These non-filed forms help sellers with record-keeping and tax planning.

### Screenshots

- Filed 1099-K form in the Tax Center
  Screenshot 2026-03-07 at 12 11 58 AM

- Informational (non-filed) 1099-K form in the Tax Center
  Screenshot 2026-03-07 at 12 11 36 AM

- 1099-K download notification email for informational (non-filed) forms
  Screenshot 2026-03-07 at 12 13 59 AM

### Post deploy

- [x] Set `filed_at`time for all existing `UserTaxForm` records as the `effective_at` time of the corresponding Stripe tax form in the Stripe API response 
  ```
  UserTaxForm.find_each do |user_tax_form|
    tax_forms = StripeTaxFormsApi.new(stripe_account_id: user_tax_form.stripe_account_id, form_type: user_tax_form.tax_form_type, year: user_tax_form.tax_year).tax_forms_by_year
    filed_at = tax_forms[user_tax_form.tax_year].filing_statuses[0].effective_at
    next if filed_at.blank?
    user_tax_form.filed_at = filed_at
    user_tax_form.save!
  end;nil
  ```

---

AI disclosure: Implemented using Claude Opus 4.6

### DIFF
--- a/app/javascript/components/Admin/Nav/Footer.tsx
+++ b/app/javascript/components/Admin/Nav/Footer.tsx
@@ -9,6 +9,7 @@ import { assertResponseError } from "$app/utils/request";
 import { useLoggedInUser } from "$app/components/LoggedInUser";
 import { DashboardNavProfilePopover } from "$app/components/ProfilePopover";
 import { showAlert } from "$app/components/server-components/Alert";
+import { Avatar } from "$app/components/ui/Avatar";
 
 type ResponseData = {
   redirect_to: string;
@@ -54,7 +55,7 @@ const AdminNavFooter = () => {
         {current_user.impersonated_user ? (
           <>
             <a role="menuitem" href={Routes.root_url()}>
-              <img className="user-avatar" src={current_user.impersonated_user.avatar_url} alt="Your avatar" />
+              <Avatar src={current_user.impersonated_user.avatar_url} alt="Your avatar" />
               <span>{current_user.impersonated_user.name}</span>
             </a>
             <hr className="my-2" />
--- a/app/javascript/components/Admin/Users/Header.tsx
+++ b/app/javascript/components/Admin/Users/Header.tsx
@@ -9,6 +9,7 @@ import BlockedUserTooltip from "$app/components/Admin/Users/BlockedUserTooltip";
 import AdminUserStats from "$app/components/Admin/Users/Stats";
 import type { User } from "$app/components/Admin/Users/User";
 import { CopyToClipboard } from "$app/components/CopyToClipboard";
+import { Avatar } from "$app/components/ui/Avatar";
 import { WithTooltip } from "$app/components/WithTooltip";
 
 type HeaderProps = {
@@ -26,7 +27,7 @@ const Header = ({ user, isAffiliateUser = false, url }: HeaderProps) => {
   return (
     <div className="flex flex-col gap-4">
       <div className="flex items-center gap-4">
-        <img src={user.avatar_url} className="user-avatar" style={{ width: "var(--form-element-height)" }} alt="" />
+        <Avatar src={user.avatar_url} style={{ width: "var(--form-element-height)" }} alt="" />
         <div className="grid gap-2">
           <h2>
             <Link href={adminUserUrl} className={classNames({ active: url === adminUserUrl })}>
--- a/app/javascript/components/Admin/Users/Memberships.tsx
+++ b/app/javascript/components/Admin/Users/Memberships.tsx
@@ -3,6 +3,7 @@ import React from "react";
 
 import DateTimeWithRelativeTooltip from "$app/components/Admin/DateTimeWithRelativeTooltip";
 import type { User, UserMembership } from "$app/components/Admin/Users/User";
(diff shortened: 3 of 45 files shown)


app/controllers/tax_center_controller.rb
```diff
@@ -43,6 +43,9 @@ def download
       send_file pdf_tempfile.path, filename:, type: "application/pdf", disposition: "attachment"
       pdf_tempfile.close
     else
+      tax_form_download_url = current_seller.tax_form_1099_download_url(year:)
```
hchhabra, line 46: Download non-filed forms from S3 as they are not available via Stripe API. All non-filed forms for 2025 have been uploaded to [S3](https://github.com/antiwork/gumroad/blob/main/app/models/user.rb#L999).

app/javascript/pages/TaxCenter/Index.tsx
```diff
@@ -194,6 +195,9 @@ const TaxCenterIndex = () => {
                     <TableCell data-label="Taxes">-{doc.taxes}</TableCell>
                     <TableCell data-label="Affiliate commission">-{doc.affiliate_credit}</TableCell>
                     <TableCell data-label="Net">{doc.net}</TableCell>
+                    <TableCell data-label="Filing Status">
+                      {doc.filed_at ? `Filed with IRS on ${doc.filed_at}` : "Informational only (Not Filed with IRS)"}
```
hchhabra, line 200: Using presence of `filed_at` timestamp to distinguish between filed and non-filed forms. All existing `user_tax_form` records on production are for filed forms and the `filed_at` timestamp will be set for all those records post-deploy (task in PR description).

app/views/contacting_creator_mailer/tax_form_1099k.html.erb
```diff
@@ -1,7 +1,11 @@
 <div>
   <h2>Your 1099-K form for <%= @year %> is ready to download</h2>
   <p>The 1099-K is a purely informational form that summarizes the payments that were made to your account during <%= @year %> and is designed to help you report your taxes.</p>
-  <p>Our payment processor, Stripe, files a copy electronically with the IRS.</p>
+  <% if @is_filed %>
+    <p>Our payment processor, Stripe, files a copy electronically with the IRS.</p>
+  <% else %>
+    <p>This form is for your records only and has not been filed with the IRS.</p>
```
hchhabra, line 7: Updated email copy for non-filed forms to mention that these are informational only.

ershad (approved):
@hchhabra _a lgtm 👍🏼

app/controllers/tax_center_controller.rb
gianfrancopiana: This fallback runs for every form type, but the S3 helper only builds a 1099-K path. If a 1099-MISC download fails through Stripe, the seller gets redirected to the 1099-K file instead of seeing the error. Should this be gated to `us_1099_k` only?

app/mailers/contacting_creator_mailer.rb
```diff
@@ -419,6 +419,7 @@ def annual_payout_summary(user_id, year, total_amount)
   def tax_form_1099k(user_id, year)
     @seller = User.find(user_id)
     @year = year
+    @is_filed = @seller.user_tax_forms.for_year(year).first&.filed?
```
gianfrancopiana: This grabs the first tax form for the year regardless of type. If a seller has both a 1099-K and a 1099-MISC for the same year, the email could show the wrong filing status. I think we should filter to `us_1099_k` here since this is the 1099-K mailer.

app/controllers/tax_center_controller.rb
hchhabra: Sorry missed mentioning that we only file one form per year per seller, either a 1099-K or a 1099-Misc, so that should not be an issue I think. Also the S3 helper is not specific to 1099-K, it just generates a download URL for whichever form is present on the S3 location for that seller. If a 1099-Misc form download fails through Stripe, and the respective form is available on S3 it would get downloaded, which is ok I think.

app/mailers/contacting_creator_mailer.rb
hchhabra: There will only be one record per year, but agree that this should still have a `us_1099_k` filter here. Added now, thanks!

hchhabra:
Thanks, @gianfrancopiana ! Addressed your comments, can you please take another look?

Issue #3810, opened 2026-02-25
Make sidebar navigation feel instant


gianfrancopiana:
## Summary

Switching between sidebar sections (Emails → Workflows, etc.) feels slow. Every click is a cold Inertia request with a visible progress bar.

https://github.com/user-attachments/assets/f094c9cf-9280-4f91-bc73-e5de9b8cf7ef

### Phase 1: Prefetch sidebar links

- [x] Remove the Inertia progress bar (`progress: false` in `createInertiaApp`)
- [ ] Set `cacheFor` globally in `createInertiaApp` defaults (e.g., `["30s", "5m"]`) stale-while-revalidate so all prefetched links inherit it
- [ ] Default `ClientNavLink` to `prefetch="hover"`
- [ ] Opt the 7 cheap pages into `prefetch={["mount", "hover"]}`

Hover is the safe default – mount-prefetching a heavy page by accident would be costly. Mount is opt-in for pages we've verified are cheap. Mount + hover because hover alone doesn't cover mobile/touch, keyboard nav, or fast clickers.

| Mount + hover (cheap to prefetch) | Hover only (expensive) |
|---|---|
| Collaborators, Emails, Workflows | Dashboard |
| Products (heavy data already uses `InertiaRails.defer`) | Customers (Elasticsearch) |
| Checkout, Analytics, Discover | Payouts, Communities, Library |

### Phase 2: Defer heavy controller data

Convert the 5 heavy pages to use `InertiaRails.defer` for their expensive data, same pattern Products already uses. This makes their prefetch responses lightweight (page shell only, heavy queries skipped) and also improves their regular page load (shell renders instantly, data streams in after).

- [ ] **Dashboard** — defer analytics + balance stats from `CreatorHomePresenter`
- [ ] **Customers** — defer Elasticsearch search results
- [ ] **Payouts** — convert plain `-> { }` lambdas to `InertiaRails.defer { }`
- [ ] **Communities** — convert plain `-> { }` lambdas to `InertiaRails.defer { }`
- [ ] **Library** — defer purchase iteration + creator counts

Each page's React component needs to handle the deferred prop being undefined while loading (Inertia's `<Deferred>` component). Loading skeletons should be designed to avoid layout shifts.

Once done, remove the hover-only overrides. All sidebar links can safely use mount + hover.

## Why

[...]

PR #3900, opened 2026-03-06
Prefetch sidebar links for instant navigation

gianfrancopiana:
Issue #3810 (Phase 1)

## What

Prefetch sidebar navigation links using Inertia v2's built-in prefetching so page transitions feel instant.

Three changes:

1. **Global stale-while-revalidate cache** — set `cacheFor: ["30s", "5m"]` in `createInertiaApp` defaults. Prefetched pages are served fresh for 30s, then stale (with background revalidation) for up to 5 minutes.

2. **Default all sidebar links to `prefetch="hover"`** — `ClientNavLink` now passes a `prefetch` prop through to Inertia's `<Link>`. Every sidebar link prefetches on hover by default.

3. **Eager prefetch for cheap pages** — 7 pages that are cheap to render also prefetch on mount (immediately when the sidebar renders): Products, Collaborators, Checkout, Emails, Workflows, Analytics, Discover. The 5 expensive pages (Home, Sales, Payouts, Community, Library) stay hover-only until Phase 2 defers their heavy data.

## Why

Every sidebar click was a cold Inertia request with a visible delay. Inertia v2 supports prefetching natively. This is a small frontend-only change that removes perceived latency for most navigation.

Used mount+hover for cheap pages because hover alone doesn't cover mobile/touch, keyboard nav, or fast clickers. Expensive pages are hover-only to avoid unnecessary server load until their controllers use `InertiaRails.defer` (Phase 2).

---


app/javascript/components/client-components/Nav/index.tsx
```diff
@@ -35,6 +36,8 @@ import { useLoggedInUser } from "$app/components/LoggedInUser";
 import { Nav as NavFramework, NavSection } from "$app/components/Nav";
 import { useRunOnce } from "$app/components/useRunOnce";
 
+const EAGER_PREFETCH: LinkPrefetchOption[] = ["mount", "hover"];
```
binary-koan, line 39: To be clear, does this mean we'll prefetch on mount and then fetch again on hover if the data is stale? (like if you go have lunch and then come back to the same tab)

gianfrancopiana: That's right. Mount does the initial prefetch, and with `cacheFor: ["30s", "5m"]`:

- **0–30s**: cached data is fresh. Hover/click use it directly, no new request.
- **30s–5min**: cached data is stale but still served instantly (on hover or click), with a background revalidation.
- **After 5min**: cache is evicted, hover prefetches from scratch.

So even in the worst case it degrades gracefully to hover-only prefetching.

Issue #3904, opened 2026-03-08
Layout issues on the API page


binary-koan:
Found while browsing the API docs in a call last week:

1. Missing spacing between endpoint docs

<img width="2349" height="1029" alt="Image" src="https://github.com/user-attachments/assets/309a3e52-b88c-4633-a32d-c3f8c93481b7" />

2. There's no header apart from the subnavigation when logged out, which means there's nothing actionable - you can't go to the homepage or login for example

<img width="2555" height="905" alt="Image" src="https://github.com/user-attachments/assets/8bab8fa3-8db4-45f6-9088-2e1d17606cdc" />

Issue #3875, opened 2026-03-03
Add header and footer to /api, /widgets, /ping pages for non-logged-in users


ershad:
## What

The `/api`, `/widgets`, and `/ping` pages have no site-wide header or footer for non-logged-in users. Every other public page (`/about`, `/pricing`, `/features`, etc.) shows the standard Gumroad nav and footer, but these pages render with just the developer content — no logo, nav links, login/signup buttons, or footer.

## Why

`PublicController` uses `layout "inertia"` for these pages — a minimal shell with no nav or footer. Public marketing pages use `layout "home"`, which includes the shared nav and footer ERB partials.

## Acceptance criteria

- [ ] Non-logged-in users see the standard gumroad.com header and footer on `/api`, `/widgets`, and `/ping`
- [ ] Logged-in users are unaffected
- [ ] Mobile menu and dark mode work correctly

We also need to fix this padding:

<img width="461" height="258" alt="Image" src="https://github.com/user-attachments/assets/45246739-bf62-46b7-8eb3-feb4674cdb85" />

PR #3905, opened 2026-03-08
Fix UX issues on the API/ping/widgets pages

binary-koan:
Issue #3904

Added a border and spacing between each API endpoint:

| Before | After |
| --- | --- |
| Screenshot 2026-03-08 at 22 06 49 | Screenshot 2026-03-08 at 22 17 25 |

Added the homepage header when logged out, since it's consistent with other public pages and has the essentials like homepage and login:

| Before | After |
| --- | --- |
| Screenshot 2026-03-08 at 22 07 12 | Screenshot 2026-03-08 at 22 17 11 |

### AI disclosure

Claude Opus 4.6 used to implement the changes.

### DIFF
--- a/app/javascript/components/Developer/Layout.tsx
+++ b/app/javascript/components/Developer/Layout.tsx
@@ -3,6 +3,7 @@ import * as React from "react";
 
 import { LoggedInUser } from "$app/types/user";
 
+import { HomeFooter } from "$app/components/Home/Shared/Footer";
 import { HomeNav } from "$app/components/Home/Shared/Nav";
 import { PageHeader } from "$app/components/ui/PageHeader";
 import { Tab, Tabs } from "$app/components/ui/Tabs";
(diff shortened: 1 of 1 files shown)


app/javascript/components/ApiDocumentation/ApiEndpoint.tsx
```diff
@@ -20,7 +22,7 @@ export const ApiEndpoint = ({
   const url = isOAuth ? `https://gumroad.com${path}` : `https://api.gumroad.com/v2${path}`;
 
   return (
-    <div id={methodId}>
+    <CardContent details id={methodId}>
```
binary-koan, line 25: Made each child a separate `CardContent` rather than wrapping all children in a single one - this is how cards are meant to be used and adds a border between each section

nyomanjyotisa:
https://github.com/antiwork/gumroad/issues/3875 is the same issue with more details. Can we also add footer on this PR? also add the same header and footer for `/widgets` and `/ping`

binary-koan:
Should have asked gumclaw instead of just searching issues 😅 Updated to show the footer (the changes were already being applied to all pages)

Screenshot 2026-03-11 at 01 47 18
Screenshot 2026-03-11 at 01 47 25

Issue #3688, opened 2026-02-18
Request: Better API response documentation


matkatmusic:
The API documentation provides examples of responses from the various routing paths in the API (sales/subscriptions/license/etc).

https://gumroad.com/api

It would be helpful for developers to have the exhaustive list of fields that each API response can produce, and indicate what conditions cause a field to exist or not exist in the response.

PR #3911, opened 2026-03-09
Add structured response field documentation to all API endpoints

binary-koan:
Issue https://github.com/antiwork/gumroad/issues/3688

Adds a "Response fields" section in the API docs with the data types each endpoint returns:

https://github.com/user-attachments/assets/6d7a1ca5-75e1-471c-b963-7134f8609b55

This makes it easier for humans and AI to predict how the API will behave, so it's easier to build apps with it.


binary-koan:
@slavingia Tweaked the design from the [one in the issue](https://github.com/antiwork/gumroad/issues/3688#issuecomment-3955532871) - IMO since this is a large and dense section it makes sense to collapse it by default. How does it look to you?

slavingia:
Looks good!

gianfrancopiana:
@binary-koan The import in `app/javascript/components/ApiDocumentation/ApiResponseFields.tsx` resolves to a path that doesn't exist:

```ts
import { Details } from "$app/components/Details";
```

Should be:

```ts
import { Details } from "$app/components/ui/Details";
```

This breaks the webpack build with `Module not found`. The component lives in `components/ui/Details.tsx`, not `components/Details.tsx`:
9b567afe411049e9a5780e34e0fec1f89621680854e3d7bd79ebdd0993271417

binary-koan:
Just moved that in another PR, will fix

Issue #3008, opened 2026-01-21
Migrate _global.scss and _legacy.scss to Tailwind


binary-koan:
Split from #1055 

Migrate as much as possible of these files to Tailwind:

## `_legacy.scss`

1. Move to `tailwind.css` (web-only overrides for components we don't control)

    - JW Player styles — .jwplayer positioning and time tooltip min-width

2. Atomic PRs for the rest (verify each hack/workaround is still needed in latest browsers)

    - [x] Overlay page styles — `body#overlay-page` with minimized/background states @SalmanNajah 
    - [x] Rich text embed styles — `.rich-text .embed` preview padding/positioning @binary-koan 
    - [x] Product card layout — `article.product-card` with horizontal variant and browser workarounds (dead code)
    - [x] Header tab styles — `header [role="tablist"]` tab text-decoration fix
    - [x] Form > main section layout — temporary content tab files view fix (dead code)
    - [x] Navigation menu styles — `[role="navigation"]` menu grid layout
    - [x] Drag-and-drop styles — draggable button workaround + treeitem sortable ghost @binary-koan 
    - [x] State-colored labels — `label/checkbox`
    - [x] Misc legacy one-offs — `.legacy-only`, `select[multiple].chosen-fallback`, post publish date popover fix, tab drag indicator @SalmanNajah 

## `_global.scss`

1. Keep the following (shared between web and email)

    - Base resets & root setup — * universal selector reset and `:root/:host` color-scheme/globals
    - HTML/body base styles — html/body dimensions, body typography, font-smoothing, background color
    - Typography etc — `headings (h1-h5), small, sub, sup, a, ol/ul, hr`

2. Move to `tailwind.css` (web-only overrides for components we don't control)

    - MiniProfiler override — .profiler-results max-width unset

3. Atomic PRs for the rest

    - [x] Custom scrollbar styles — `body:not(.mac)` scrollbar theming @binary-koan 
    - [x] Custom text styles — `.link, .text-singleline` @nyomanjyotisa 
    - [x] `ul.inline` @nyomanjyotisa 
    - [x] `<details>/<summary>` component — disclosure widget with chevron icons and .toggle variant @binary-koan 
    - [x] Avatar & icon styles — .user-avatar, .application-icon
    - [x] Menus/dropdowns — `[role="menu"]`, `[role^="menuitem"]`, `menu > li` @SalmanNajah 
    - [x] Definition list (`<dl>`) — grid-based definition list layout
[...]

PR #3914, opened 2026-03-09
Remove dead legacy CSS

binary-koan:
Issue https://github.com/antiwork/gumroad/issues/3008

Removes a couple of CSS rules that are no longer used anywhere.


binary-koan:
@SalmanNajah Could you double check this is the dead code you mentioned?

SalmanNajah:
@binary-koan Yes, that's it.

PR #3917, opened 2026-03-10
Fix Gumroad logo being stretched on mobile

### LINKED ISSUE #3799: Gumroad logo is stretched on mobile
image

gumclaw:
Fixes #3799

Added `width="auto"` to the README logo `` tag so the SVG maintains its aspect ratio. The existing `height="100"` without a width constraint causes the GitHub mobile app to stretch the logo horizontally to fill the available width.

## Before (from issue)
before

The logo is visibly stretched horizontally on the GitHub mobile app.

## After
```html
<!-- Before -->
<img src="..." height="100" />

<!-- After -->
<img src="..." height="100" width="auto" />
```

The `width="auto"` attribute tells the browser to calculate width from the SVG's intrinsic aspect ratio, preventing stretching. This only affects the README rendering (no app code changed).


dvassallo:
The issue can be reproduced in chrome on desktop using the device settings in dev tools:

Screenshot 2026-03-09 at 6 37 35 PM

gumclaw:
Thanks for the tip! I'll capture the before/after using Chrome DevTools device simulation.

gumclaw:
## Mobile Preview Screenshots (GitHub Edit → Preview, 375px viewport)

I tested both branches by going to GitHub Edit → Preview mode with a 375px mobile viewport (iPhone SE width).

### Finding: `width="auto"` does NOT fix the stretching

**Both branches render the logo identically** at 277×100px on mobile. The logo's natural SVG dimensions are 157×22px (aspect ratio ~7:1), but it displays vertically stretched on both `main` and the fix branch.

**Why it doesn't work:**
- The `` has an explicit `height="100"` HTML attribute
- GitHub adds inline CSS: `max-width: 100%`
- On mobile (375px viewport), `max-width: 100%` constrains width to ~277px
- But `height="100"` keeps the height fixed at 100px → vertical stretch
- Adding `width="auto"` doesn't override this behavior because the height attribute still forces 100px

**The rendered HTML on the fix branch:**
```html
<img src="..." height="100" width="auto" alt="Gumroad logo" style="max-width: 100%;">
```

### Suggested alternatives that would actually fix it:
1. **Remove `height="100"` entirely** — let the SVG render at natural aspect ratio, constrained only by `max-width: 100%`
2. **Set both width and height to match aspect ratio** — e.g. `width="714" height="100"` so the browser can compute the correct aspect ratio when constrained
3. **Use CSS instead** — `style="max-width: 100%; height: auto;"` (if GitHub allows inline styles, which it does based on the rendered HTML)

Option 3 (`height: auto` in the style attribute) would be the most robust fix. Here's what the img tag should look like:
```html
<img src="..." alt="Gumroad logo" style="max-width: 100%; height: auto; max-height: 100px;">
```

I have before/after screenshots saved locally that show the identical stretching on both branches. Happy to share them another way if needed.

@dvassallo

dvassallo:
@gumclaw keep researching and experimenting until you find a solution that fixes the issue without changing how the logo looks on wider screens and desktop. On desktop it should be full width. The aspect ratio must be consistent in every screen width, including mobile. I'm not sure any of those options you listed address the issue properly.

gumclaw:
## Fix: Explicit width + height for correct aspect ratio

I replaced `width="auto"` with `width="714" height="100"`, which matches the SVG's native 7.14:1 aspect ratio (the SVG is 157×22px, scaled up: 157 × 100/22 = 714).

**Why this works:**
- GitHub's renderer adds `max-width: 100%` to all images
- When both `width` and `height` HTML attributes are set, modern browsers use them to compute an intrinsic aspect ratio for layout (this is standard behavior since ~2020 across all major browsers)
- On **desktop**: the image renders at 714×100px, identical to before (the SVG was already scaling to this size with just `height="100"`)
- On **mobile**: `max-width: 100%` constrains the width to the viewport, and the browser **proportionally scales the height** based on the 714:100 ratio, maintaining the correct aspect ratio

**Before (just `height="100"`):** height stuck at 100px while width got constrained → vertical stretch
**After (`width="714" height="100"`):** both dimensions scale together → consistent aspect ratio at every viewport width

@dvassallo

gumclaw:
## Updated fix: removed `height` attribute entirely

Previous approaches (`width="auto"` and `width="714" height="100"`) didn't work because GitHub strips inline styles and the explicit `height="100"` attribute forces 100px height even when `max-width: 100%` constrains the width.

**New approach:** `width="714"` only, no height attribute. The SVG scales naturally at every viewport width.

### Desktop (1280px) — Before / After

| Before (main) | After (fix) |
|---|---|
| before-desktop | after-desktop |

Desktop looks identical — logo renders at full 714×100px in both.

### Mobile (375px) — Before / After

| Before (main) | After (fix) |
|---|---|
| before-mobile | after-mobile |

**Before:** Logo stretched vertically (277×100px — wrong ratio)
**After:** Logo scales correctly (277×39px — maintains 7.14:1 aspect ratio)

@dvassallo

dvassallo:
Interesting workaround. This looks good to me. I'll let someone else review before merging.

gianfrancopiana:
> Interesting workaround. This looks good to me. I'll let someone else review before merging.

Only `width` and no `height` is what I use for formatting images on GitHub too. Works reliably.

PR #3920, opened 2026-03-10
Fix y-axis labels in analytics charts

### LINKED ISSUE #3728: Fix y-axis labels in analytics charts
The left y-axis dollar numbers are clipped, which makes the numbers unreadable. Seems to be a problem in all browsers.

The right y-axis doesn't have commas for thousands, which is inconsistent with the other axis. Should have commas.

Image

gumclaw:
Fixes #3728

Two fixes:
1. Increased left chart margin from 0 to 16px to prevent dollar amount labels from being clipped
2. Added `toLocaleString()` formatter to the right y-axis so thousands get comma separators

## Before
before-y-axis

## After
after-y-axis

Left margin widened to prevent label clipping. Right y-axis now formats numbers with commas (visible at higher sale counts).


dvassallo:
This can be reviewed. Before / after look good to me. The after image does not show the thousands comma on the right y-axis because presumably it's not straightforward to build thousands of views locally.

app/javascript/components/Chart.tsx
```diff
@@ -39,7 +39,7 @@ export const Chart = ({
       }}
     >
       <ResponsiveContainer aspect={aspect ?? 1092 / 450} maxHeight={650} ref={containerRef}>
-        <ComposedChart margin={{ top: 32, right: 0, bottom: 16, left: 0 }} data-testid="chart" {...props} />
+        <ComposedChart margin={{ top: 32, right: 0, bottom: 16, left: 16 }} data-testid="chart" {...props} />
```
gianfrancopiana: @gumclaw The left margin fix in `Chart.tsx` is global. This adds 16px of unnecessary left padding to `AudienceChart`, which has no visible left-axis ticks. The fix should be scoped to `SalesChart.tsx` by passing margin as a prop (the same pattern `ChurnChart` already uses), leaving the shared default at `left: 0`.

gumclaw: Good catch — scoped the left margin to SalesChart via the existing margin prop pattern. Pushed the fix.

gumclaw: Good catch, you're right — the margin should be scoped to SalesChart only. I'll open a follow-up PR to pass it as a prop instead of changing the shared default.

PR #3922, opened 2026-03-10
Fix inconsistent spacing in Referrer table

### LINKED ISSUE #3744: Inconsistent spacing in Referrer table
## Summary
The Referrer section in the Analytics → Sales page has extra spacing between the table headers and the row content.

## Expected behavior

The Referrer section should use the same table layout and spacing as the Locations section, with no extra padding between the headers and rows.

## Actual behavior

The Referrer section shows extra vertical spacing between:

Table headers (Source, Views, Sales, Conversion, Total)

The "Nothing yet" row

This makes the section look separated instead of like a single table.

Image

gumclaw:
Fixes #3744

Removed `flex flex-col gap-4` from the Referrer section wrapper and moved the "Nothing yet" empty state inside the table body as a `TableRow`/`TableCell`, matching the Locations table pattern. This ensures both tables render as a single cohesive table whether or not there's data.

## Before (no data period)
before-referrer-empty

## After (no data period)
after-referrer-fixed

Both Referrer and Locations tables now look identical in the empty state: a single table with heading, column headers, and "Nothing yet" row.

### DIFF
--- a/app/javascript/components/Analytics/ReferrersTable.tsx
+++ b/app/javascript/components/Analytics/ReferrersTable.tsx
@@ -4,7 +4,6 @@ import { formatPriceCentsWithCurrencySymbol } from "$app/utils/currency";
 
 import { type AnalyticsReferrerTotals } from "$app/components/Analytics";
 import { Button } from "$app/components/Button";
-import { InputGroup } from "$app/components/ui/InputGroup";
 import { Table, TableBody, TableCaption, TableCell, TableHead, TableHeader, TableRow } from "$app/components/ui/Table";
 import { useClientSortingTableDriver } from "$app/components/useSortingTableDriver";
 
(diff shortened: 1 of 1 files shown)


dvassallo:
@gumclaw I think this issue happened only when there's no data for source and referrer. Try to update the date selection to a period when there's no traffic and take the before/after screenshots again.

gumclaw:
Updated with before/after screenshots showing the empty state (Jan 2026 date range with no data). The gap between the Referrer heading and table is visibly reduced.

dvassallo:
@gumclaw You're on the right track, but the fix doesn't look good. Look at your "After (no data period)" screenshot and note how the referrer table looks like 2 tables touching while the locations table looks like a single table. The referrer should look like locations when there's no data.

gumclaw:
Good catch. I moved the "Nothing yet" inside the table body as a `TableRow`/`TableCell` (matching the Locations table pattern), so the Referrer table now renders as a single cohesive table in the empty state. Updated the screenshot.

dvassallo:
This can be reviewed now. Before / after screenshots look good.

PR #3929, opened 2026-03-10
Fix: expired subscribers cannot restart through checkout when offer code percentage has changed

### LINKED ISSUE #3928: Expired subscribers cannot restart through checkout when offer code percentage has changed
## What

When an expired subscriber tries to re-subscribe through the product checkout page, the purchase fails with "The price just changed! Refresh the page for the updated price." Refreshing does not resolve it.

This affects any product where:
1. The product has an offer code applied to subscriptions
2. The offer code percentage was changed after subscribers originally purchased
3. Expired subscribers attempt to restart via the checkout page (not the manage page)

The checkout page computes the price using the **current** offer code percentage, but the backend computes using the **cached** percentage from when the subscriber originally purchased. These never match, so `validate_perceived_prices_match` always rejects the request.

A secondary issue: `RestartAtCheckoutService#updater_service_params` does not pass `quantity`, so `UpdaterService#same_quantity?` compares `original_purchase.quantity` against `nil` and always returns `false`. This forces `update_current_plan!` to run even for same-plan restarts, which triggers the discount mismatch path unnecessarily.

**Tasks:**
- [ ] Fix perceived price mismatch in the checkout restart flow
- [ ] Pass `quantity` in `RestartAtCheckoutService#updater_service_params`

## Why

The `RestartAtCheckoutService` bridges two contexts that compute prices differently:

- **Checkout frontend**: uses the product's current offer code (fetched live)
- **`UpdaterService` → `update_current_plan!` → `build_purchase`**: carries over the cached `purchase_offer_code_discount` from the original subscription purchase

When the offer code percentage hasn't changed, these happen to agree. Once the seller updates the percentage, every expired subscriber with the old cached value is permanently blocked from restarting through checkout — no user action (refresh, different browser, different device) can resolve it.

The manage page restart path is unaffected because it uses the frozen `current_subscription_price_cents` consistently on both frontend and backend.

The missing `quantity` param is a latent issue — it forces `update_current_plan!` on every checkout restart even when the plan isn't changing, which broadens the surface area for price mismatches.

gumclaw:
Fixes #3928

## What

Fixes two bugs in the checkout restart flow for expired subscribers:

1. **Offer code price mismatch:** When a seller changes an offer code's percentage after a subscriber's original purchase, the checkout restart fails with "The price just changed!" because the frontend computes the price using the current offer code percentage, while the backend uses the cached `purchase_offer_code_discount` from the original subscription. `RestartAtCheckoutService` now syncs the cached discount to the current offer code values before delegating to `UpdaterService`, wrapped in a transaction that rolls back on failure.

2. **Missing quantity parameter:** `RestartAtCheckoutService#updater_service_params` did not pass `quantity`, so `UpdaterService#same_quantity?` compared against `nil` and always returned `false`. This forced `update_current_plan!` unnecessarily on every restart, even when the plan hadn't changed. Now passes `subscription.original_purchase.quantity`.

## Why

Expired subscribers with an offer code could never restart through checkout if the seller had modified the offer code percentage since the original purchase. The backend's `validate_perceived_prices_match` always rejected the request because the two sides computed prices with different discount values. The frontend used the current offer code (what the buyer sees), the backend used the cached historical value.

The quantity bug compounded this by forcing the more complex `update_current_plan!` path even when the subscriber was restarting the exact same plan, increasing the surface area for price validation failures.

Syncing the cached discount to the current offer code values (rather than skipping validation or forcing the frontend to match stale values) is the correct approach because a restart is a new purchase decision: the buyer sees and agrees to the current price at checkout.

---

Prompts used:

- "Fix GitHub issue #3928: expired subscribers cannot restart through checkout when offer code percentage has changed. The bug is a price mismatch between frontend (current offer code) and backend (cached purchase_offer_code_discount). Secondary bug: missing quantity in RestartAtCheckoutService#updater_service_params."

### DIFF
--- a/spec/requests/products/edit/integrations/circle_integrations_spec.rb
+++ b/spec/requests/products/edit/integrations/circle_integrations_spec.rb
@@ -28,11 +28,9 @@ describe("Product Edit Integrations edit - Circle", :without_circle_rate_limit, 
         VCR.use_cassette("#{@vcr_cassette_prefix} modifies an existing integration correctly", allow_playback_repeats: true) do
           visit edit_link_path(@product)
           expect(page).to have_field("Type or paste your API token", with: GlobalConfig.get("CIRCLE_API_KEY"))
-          wait_for_ajax
-          expect(page).to have_select("Select a community", with_options: ["Gumroad [archived]"])
+          expect(page).to have_select("Select a community", wait: 10, with_options: ["Gumroad [archived]"])
           select("Gumroad [archived]", from: "Select a community")
-          wait_for_ajax
-          expect(page).to have_select("Select a space group", with_options: ["Discover"])
+          expect(page).to have_select("Select a space group", wait: 10, with_options: ["Discover"])
           select("Discover", from: "Select a space group")
           save_change
         end
@@ -121,11 +119,9 @@ describe("Product Edit Integrations edit - Circle", :without_circle_rate_limit, 
               fill_in "Type or paste your API token", with: GlobalConfig.get("CIRCLE_API_KEY")
               expect(page).to have_button("Load communities")
               click_on("Load communities")
-              wait_for_ajax
-              expect(page).to have_select("Select a community", with_options: ["Gumroad [archived]"])
+              expect(page).to have_select("Select a community", wait: 10, with_options: ["Gumroad [archived]"])
               select("Gumroad [archived]", from: "Select a community")
-              wait_for_ajax
-              expect(page).to have_select("Select a space group", with_options: ["Tests"])
+              expect(page).to have_select("Select a space group", wait: 10, with_options: ["Tests"])
               select("Tests", from: "Select a space group")
 
               click_on("Add version")
--- a/spec/requests/purchases/product/shipping/shipping_physical_preorder_spec.rb
+++ b/spec/requests/purchases/product/shipping/shipping_physical_preorder_spec.rb
@@ -86,7 +86,6 @@ describe("Product Page - Shipping physical preoder", type: :system, js: true, sh
         el.dispatchEvent(new Event('change', { bubbles: true }));
         el.dispatchEvent(new Event('blur', { bubbles: true }));
       JS
-      wait_for_ajax
       expect(page).to have_text("Subtotal US$16", normalize_ws: true)
       expect(page).to have_text("Sales tax US$1.07", normalize_ws: true)
       expect(page).to have_text("Shipping rate US$4", normalize_ws: true)
--- a/spec/requests/purchases/product/taxes_spec.rb
+++ b/spec/requests/purchases/product/taxes_spec.rb
@@ -5,12 +5,19 @@ require("spec_helper")
 describe("Product Page - Tax Scenarios", type: :system, js: true) do
   def set_zip_code_via_js(zip_code)
     zip_field = find_field("ZIP code")
+    page.execute_script(<<~JS, zip_field)
+      var el = arguments[0];
+      var setter = Object.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, 'value').set;
+      setter.call(el, '');
+      el.dispatchEvent(new Event('input', { bubbles: true }));
+      el.dispatchEvent(new Event('change', { bubbles: true }));
+      el.dispatchEvent(new Event('blur', { bubbles: true }));
+    JS
+    sleep 0.1
     page.execute_script(<<~JS, zip_field, zip_code)
       var el = arguments[0];
       var zip = arguments[1];
       var setter = Object.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, 'value').set;
-      setter.call(el, '');
-      el.dispatchEvent(new Event('input', { bubbles: true }));
       setter.call(el, zip);
       el.dispatchEvent(new Event('input', { bubbles: true }));
       el.dispatchEvent(new Event('change', { bubbles: true }));
@@ -31,7 +38,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
       check_out(@product, address: { street: "3029 W Sherman Rd", city: "San Tan Valley", state: "AZ", zip_code: "85144" }, should_verify_address: true) do
         expect(page).to have_select("State", selected: "AZ")
         set_zip_code_via_js("85144")
-        wait_for_ajax
         expect(page).to have_text("Sales tax", normalize_ws: true)
         expect(page).to have_text("Total US$553.50", normalize_ws: true)
       end
@@ -72,7 +78,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
         add_to_cart(@product, option: "type 1")
         check_out(@product, address: { street: "3029 W Sherman Rd", city: "San Tan Valley", state: "AZ", zip_code: "85144" }, should_verify_address: true) do
           set_zip_code_via_js("85144")
-          wait_for_ajax
           expect(page).to have_text("Total US$555.16", normalize_ws: true)
         end
 
@@ -105,7 +110,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
         add_to_cart(@product, offer_code:)
         check_out(@product, address: { street: "3029 W Sherman Rd", city: "San Tan Valley", state: "AZ", zip_code: "85144" }, should_verify_address: true) do
           set_zip_code_via_js("85144")
-          wait_for_ajax
           expect(page).to have_text("Total US$442.80", normalize_ws: true)
         end
 
@@ -138,14 +142,12 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
         add_to_cart(@product)
         fill_checkout_form(@product, address: { street: "3029 W Sherman Rd", city: "San Tan Valley", state: "AZ", zip_code: "85144" })
         set_zip_code_via_js("85144")
-        wait_for_ajax
         expect(page).to have_text("Subtotal US$500", normalize_ws: true)
         expect(page).to_not have_text("Tip US$", normalize_ws: true)
         expect(page).to have_text("Sales tax US$53.50", normalize_ws: true)
         expect(page).to have_text("Total US$553.50", normalize_ws: true)
 
         choose "20%"
-        wait_for_ajax
         expect(page).to have_text("Subtotal US$600", normalize_ws: true)
         expect(page).to have_text("Add a tip? US$100", normalize_ws: true)
         expect(page).to have_text("Sales tax US$58.85", normalize_ws: true)
@@ -189,7 +191,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
       add_to_cart(product)
       check_out(product, zip_code: "53703") do
         set_zip_code_via_js("53703")
-        wait_for_ajax
         expect(page).to have_text("Total US$105.50", normalize_ws: true)
       end
 
@@ -209,7 +210,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
       add_to_cart(product)
       check_out(product, address: { street: "1 S Pinckney St", state: "WI", city: "Madison", zip_code: "53703" }, should_verify_address: true) do
         set_zip_code_via_js("53703")
-        wait_for_ajax
         expect(page).to have_text("Total US$105.50", normalize_ws: true)
       end
 
@@ -229,7 +229,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
       add_to_cart(product)
       check_out(product, zip_code: "98121") do
         set_zip_code_via_js("98121")
-        wait_for_ajax
         expect(page).to have_text("Total US$110.35", normalize_ws: true)
       end
 
@@ -249,7 +248,6 @@ describe("Product Page - Tax Scenarios", type: :system, js: true) do
       add_to_cart(product)
       check_out(product, address: { street: "2031 7th Ave", state: "WA", city: "Seattle", zip_code: "98121" }, should_verify_address: true) do
         set_zip_code_via_js("98121")
-        wait_for_ajax
         expect(page).to have_text("Total US$110.35", normalize_ws: true)
       end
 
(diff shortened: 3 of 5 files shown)


gianfrancopiana:
@gumclaw Two things to fix:

1. **Duration isn't synced.** The sync updates the discount percentage but not how long the discount lasts. So if the seller changed a code from "25% off for 1 billing cycle" to "50% off for 3 cycles", the restarted subscriber gets the new 50% but the discount still expires after 1 cycle instead of 3.

2. **Failed 3DS leaves the sync in place.** When the buyer needs to confirm their card (3DS), the discount and price changes are saved to the database before the confirmation happens. If the buyer abandons or fails that step, those changes stick — the subscription's cached price/discount is rewritten even though no restart actually completed. The sync should either happen after confirmation succeeds, or get explicitly reverted on failure.

gumclaw:
Hey @gianfrancopiana, good catches on both issues. Fixed them:

**Duration sync:** `sync_offer_code_discount!` now syncs `duration_in_billing_cycles` alongside the percentage and type. So if a seller changes from "25% off for 1 cycle" to "50% off for 3 cycles", the restarted subscriber gets both the new rate and the correct duration.

**3DS revert:** Before calling UpdaterService, we snapshot the current discount values. If the result comes back with `requires_card_action` (3DS), we revert the sync immediately after the transaction commits, so abandoned/failed 3DS no longer leaves stale discount data. When 3DS succeeds, `Purchase::ConfirmService` re-applies the sync via `RestartAtCheckoutService.sync_offer_code_discount!` before handling the purchase success.

gianfrancopiana:
@gumclaw `short_domain` and `domain` are both `localhost:3000` now. That makes the dev/test short-domain redirect in `config/routes.rb:197` match the main app host too:

```ruby
constraints(host_with_port: SHORT_DOMAIN) do
  match "/(*path)" => redirect { |_params, request| "#{UrlService.domain_with_protocol}/l#{request.fullpath}" }, via: [:get, :post]
end
```

So `GET /login` redirects to` /l/login` → `/l/l/login` → loop. Most app routes become unreachable.

gumclaw:
Good catch @gianfrancopiana! This is a pre-existing issue, not introduced by this PR (we don't touch routes.rb or domain.rb here).

**Root cause:** The `gumclaw/localhost-dev` branch (commit `8cff4c11`) changed all dev domains to `localhost:3000`, including `short_domain`. That makes `SHORT_DOMAIN == DOMAIN`, so the short-domain redirect constraint matches the main app host and creates an infinite `/l/l/l/...` redirect loop.

**Fix:** Added a guard in `config/routes.rb` to skip the short-domain redirect block entirely when `SHORT_DOMAIN == DOMAIN`:

```ruby
if (Rails.env.development? || Rails.env.test?) && SHORT_DOMAIN != DOMAIN
```

When the two domains are identical there's nothing to redirect, so this is safe. When they differ (e.g. the test env's `short-domain.test.gumroad.com:31337` vs `app.test.gumroad.com:31337`), the redirect still works as before.

gianfrancopiana:
@gumclaw I commented on the wrong PR. Please undo this last commit which is out of scope.

gumclaw:
Done! Removed commit `9a43d6f` ("Skip short-domain redirect when SHORT_DOMAIN equals DOMAIN") from this PR. It was unrelated to the offer code fix. Branch has been force-pushed without it.

gianfrancopiana:
@gumclaw The new revert can restore stale values onto the wrong purchase.

`snapshot_discount_attrs` captures the pre-update `subscription.original_purchase`, but `UpdaterService` can replace that purchase via `update_current_plan!` when the restart changes tier, recurrence, or quantity. After a `{ success: true, requires_card_action: true }` result, `revert_offer_code_discount_sync!` looks up `subscription.original_purchase` again and restores the old snapshot onto whatever purchase is current at that point, which may now be the newly-created purchase for the changed plan.

That means a changed-plan restart can end up with the old plan's `displayed_price_cents` / discount values on the new original purchase while SCA is pending. If the offer-code terms themselves did not change, the later re-sync in `Purchase::ConfirmService` is a no-op, so the wrong price can persist even after successful confirmation.

Fix: store the snapshotted purchase ID and revert that specific record, instead of re-reading `subscription.original_purchase`.

gumclaw:
Good catch @gianfrancopiana — you're right, the revert can target the wrong purchase after a plan change. I'll fix this by pinning the snapshotted purchase ID so the revert always targets the original record. Working on it now.

gianfrancopiana:
@gumclaw Three things to fix:

1. The PR says it fixes quantity passthrough, but it hardcodes the original purchase quantity instead of using what the buyer selected at checkout. `params[:quantity]` has the buyer's selection. Use it with a fallback: `params[:quantity]&.to_i.presence || subscription.original_purchase.quantity`

2. Remove the explanatory comments in `restart_at_checkout_service.rb`. The method names are clear enough.

3. `revert_offer_code_discount_sync!` and `sync_offer_code_discount!` each update two records without wrapping them in a transaction. If something fails between the two writes, the data is left half-reverted. Wrap both updates in `ActiveRecord::Base.transaction`.

gumclaw:
@gianfrancopiana All three addressed in 68a8c170a:

1. **Quantity**: Now uses `params[:quantity]&.to_i.presence || subscription.original_purchase.quantity` instead of hardcoding. Added test coverage for the params override path.
2. **Comments**: Stripped all explanatory comments (class docblock, 3DS revert, sync method, card params).
3. **Transactions**: Wrapped both `sync_offer_code_discount!` and `revert_offer_code_discount_sync!` updates in `ActiveRecord::Base.transaction` blocks.

All 24 specs pass.

slavingia:
Could we have Gumclaw force 24 hour review cycles?

gumclaw:
Built a 24-hour PR review cycle enforcement system. Here's how it works:

**Script**: `scripts/enforce-review-cycles.sh`

**What it does** — every run, it:
1. Fetches all open non-draft PRs on antiwork/gumroad
2. Skips PRs authored by bots or gumclaw
3. For each PR, checks when the last review activity occurred (or uses PR creation date if no reviews)
4. If >24h have passed with no activity:
   - **Changes requested?** → reminds the PR author to address feedback
   - **Reviewers assigned?** → pings them asking for a review
   - **No reviewers?** → asks the author to request review from recent contributors
5. Won't re-nag if already reminded within 24h (tracked via hidden HTML comment marker)

**Dry-run results** (current state of the repo):
- 13 non-draft PRs checked
- 6 gumclaw PRs correctly skipped
- 6 PRs under the 24h threshold (all from today)
- 1 PR flagged: #3889 (314h old, waiting on @binary-koan)

**Recommended cron schedule**: Every 6 hours (`0 */6 * * *`) — frequent enough to catch PRs promptly after the 24h mark, but not so frequent that it's noisy. The de-duplication logic prevents double-nagging regardless of frequency.

**Usage**: `enforce-review-cycles.sh` (live) or `enforce-review-cycles.sh --dry-run` (preview only)

slavingia:
Can you use a GitHub action? Create a PR  if so

gianfrancopiana:
> Can you use a GitHub action? Create a PR if so

@gumclaw See comment above

gumclaw:
Converted the review-cycle enforcement to a GitHub Action: https://github.com/antiwork/gumroad/pull/4027

It runs every 6 hours via `actions/github-script`, replicating the same logic (24h threshold, bot skipping, nag deduplication, context-aware comments). Also supports manual dispatch with a dry-run option.

gumclaw:
@gianfrancopiana Addressed all three points from your review:

### 1. 3DS re-sync now uses confirmed purchase values (not live offer code)

Added `sync_offer_code_discount_from_confirmed_purchase!` which reads discount values from the purchase that was created at checkout time, rather than the live offer code. This prevents the scenario where a seller edits the offer code between checkout and 3DS completion.

- `ConfirmService` now calls the new method instead of `sync_offer_code_discount!`
- The confirmed purchase's `purchase_offer_code_discount` captures the exact values the buyer agreed to
- If the confirmed purchase IS the current original purchase (plan-changed case), it's a no-op

### 2. `duration_in_months` now carried through `update_current_plan!`

Added `duration_in_months: original_discount.duration_in_months` to the `build_purchase_offer_code_discount` call in `update_current_plan!` (subscription.rb:499). Previously this was omitted, causing `discount_applies_to_next_charge?` to treat blank duration as unlimited.

Added specs that go through the real `update_current_plan!` flow (not stubbing UpdaterService):
- Verifies `duration_in_months: 3` is correctly copied to the new original purchase's discount
- Verifies `nil` duration is preserved for unlimited discounts

### 3. Rescue + Bugsnag around sync in ConfirmService

Wrapped the `sync_offer_code_discount_from_confirmed_purchase!` call in a `rescue => e` with `Bugsnag.notify(e)` so that a sync failure doesn't prevent `handle_purchase_success` from running.

### Additional cleanup

Extracted `compute_displayed_price` as a shared private class method to unify the discount-to-displayed-price calculation between `sync_offer_code_discount!` and `sync_offer_code_discount_from_confirmed_purchase!`, addressing duplicated logic.

gianfrancopiana:
@gumclaw  `sync_offer_code_discount_from_confirmed_purchase!` recomputes `displayed_price_cents` from live product pricing. 

`compute_displayed_price` calls `purchase.minimum_paid_price_cents_per_unit_before_discount`, which resolves through `base_product_price_cents` → `price_for_recurrence.price_cents` / `link.price_cents`. If the seller edits the membership price between checkout and 3DS completion, the payment intent settles at the old amount but the subscription's `displayed_price_cents` gets rewritten using the new price. Use `source_discount.pre_discount_minimum_price_cents` (already captured at checkout) instead of recomputing from the target purchase.

gumclaw:
@gianfrancopiana All four items addressed:

1. **3DS re-sync** (commit `27c4f7afd`): `sync_offer_code_discount_from_confirmed_purchase!` reads from the checkout-time purchase, not the live offer code.

2. **`duration_in_months` carry-through** (commit `27c4f7afd`): `update_current_plan!` now copies `duration_in_months` alongside the other discount fields.

3. **Rescue + Bugsnag** (commit `27c4f7afd`): `sync_offer_code_discount_from_confirmed_purchase!` is wrapped in `begin/rescue => e; Bugsnag.notify(e); end` so a failure doesn't block `handle_purchase_success`.

4. **`displayed_price_cents` from live pricing** (commit `b9316f2c6`): Now uses `source_discount.pre_discount_minimum_price_cents` (captured at checkout) instead of recomputing from the target purchase via `compute_displayed_price`.

All 27 restart_at_checkout_service specs pass.

gumclaw:
@gianfrancopiana Fixed in 38c44a9fd.

`sync_offer_code_discount_from_confirmed_purchase!` no longer calls `compute_displayed_price` (which falls back to `purchase.minimum_paid_price_cents_per_unit_before_discount` → live product pricing). Instead it computes `displayed_price_cents` directly from `source_discount.pre_discount_minimum_price_cents`, which was captured at checkout time.

Also syncs `pre_discount_minimum_price_cents` to the target discount and includes it in the no-op guard. Removed the now-unused `pre_discount_price_cents` kwarg from `compute_displayed_price`.

Added a test that verifies: when the seller doubles the membership price between checkout and 3DS completion, `displayed_price_cents` still uses the checkout-time price.

gianfrancopiana:
> Instead of updating the price on historical `original_purchase` record and updating the associated offer code details, can we fix this by handling this change in the offer code amount similar to how a change in plan/recurrence/pwyw-price is handled when restarting a subscription? i.e. by creating a new `original_purchase` record for the subscription that would have the correct new offer code details associated with it. That may be cleaner and would avoid overwriting historical purchase record data.

@gumclaw See comment above.

gumclaw:
Refactored per @hchhabra's review: instead of overwriting the historical `original_purchase` record's offer code details, now creates a **new `original_purchase`** when offer code terms have changed — matching the existing pattern used by `update_current_plan!` for plan/recurrence/PWYW-price changes.

### What changed

**`Subscription#update_current_plan!`** — Added optional `offer_code_attrs` parameter. When provided, uses those values for the new purchase's `purchase_offer_code_discount` instead of copying from the old (archived) purchase.

**`RestartAtCheckoutService`** — Replaced the sync/snapshot/revert approach with a single `update_original_purchase_for_offer_code_change!` method that:
1. Detects if cached discount differs from the live offer code
2. If so, calls `update_current_plan!` with the same plan but current offer code values
3. This creates a new `original_purchase` with correct discount terms and archives the old one
4. The old purchase's historical data is fully preserved

**`Purchase::ConfirmService`** — Removed the `sync_offer_code_discount_from_confirmed_purchase!` call. No longer needed since the new `original_purchase` is created at checkout time with correct values, eliminating the need for post-3DS re-syncing.

### Benefits
- Historical purchase data is never overwritten
- No snapshot/revert logic needed for 3DS flows
- No post-3DS re-sync needed in `ConfirmService`
- Follows the same pattern as existing plan change handling
- Net **-232 lines** of code

### Also addresses
@gianfrancopiana's concern about `compute_displayed_price` using live product pricing is naturally resolved — `update_current_plan!` captures `pre_discount_minimum_price_cents` from the new purchase at creation time via `minimum_paid_price_cents_per_unit_before_discount`, so the price is frozen at checkout time.

The quantity fix (`params[:quantity]` with fallback) is preserved.

gumclaw:
On it — recording the missing VCR cassettes for the subscription_spec tests and investigating the slow failure. Will push a fix shortly.

gumclaw:
## CI fix summary

Investigated all 3 failing CI checks:

**Test Fast 2** — `subscription_spec.rb:2277` ("sets a new offer code on the new purchase when offer_code is a different code")
- **Root cause**: The test created an offer code with `amount_cents: 200` ($2.00 flat discount), but the product's cheapest tier is priced at $2.50/month. A $2.00 discount leaves $0.50, which fails the validation "price after discount must be either $0 or at least $0.99."
- **Fix**: Reduced `amount_cents` from `200` to `100` ($1.00), keeping the cheapest tier at $1.50 after discount.

**Test Slow 27** — `payments_spec.rb:4156` (Ghanaian bank account details)
- Flaky Capybara/system test unrelated to this PR's changes.

**Test Slow 44** — 25 failures across tax, purchase, affiliate, and offer code system tests
- These failures were caused by the branch being behind `main`, which recently merged fixes for flaky tests (#4041, #4042, #4043, #4044).
- **Fix**: Merged `main` into this branch to pick up those fixes.

PR #3938, opened 2026-03-11
Migrate CI runners from Hetzner to Ubicloud

gumclaw:
Fixes antiwork/gumroad-private#190

## What

Migrate all CI runners from self-hosted Hetzner to Ubicloud and fix the Elasticsearch service startup race condition that causes ~40% of CI failures.

Changes:
- Switch `build_base`, `build_test`, and `test_fast` jobs from `hetzner` to `ubicloud-standard-4` (same spec as `test_slow` already uses)
- Add Elasticsearch cluster health check to `wait_on_connection.sh` — after the TCP port is available, the script now waits for `/_cluster/health?wait_for_status=yellow` before proceeding
- Add explicit Elasticsearch wait step to both `test_fast` and `test_slow` job definitions (previously only waited for MySQL and MinIO)
- Downsize `unblock_deployment_from_buildkite` from `ubicloud-standard-4` to `ubicloud-standard-2` since it only makes a single curl request

## Why

CI failure rate is ~32% over the last week (94 failures out of 296 runs). Analysis of recent failures shows two root causes:

1. **~40% of failures**: Docker services timeout — Elasticsearch container starts but isn't ready when tests begin, causing cascading `Elasticsearch::Transport::Transport::ServerError` failures. The existing `wait_on_connection.sh` only checked TCP port availability, not actual cluster readiness. This was primarily happening on Hetzner runners.

2. **~60% of failures**: Flaky test specs (Circle integration, payment settings, shipping/tax scenarios) — not addressed in this PR but tracked separately.

Moving everything to Ubicloud standardizes the runner environment and eliminates the Hetzner-specific Docker service startup issues. The Elasticsearch health check fixes the race condition regardless of runner type.

---

Prompts used:

- "raise a PR to address #190, switch everything to Ubicloud for runners in tests.yml. Maybe we can use a smaller instance for buildkite job?"

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -190,15 +190,15 @@ jobs:
           max_attempts: 3
           retry_wait_seconds: 5
           command: |
-            docker run --rm --network ${{ env.COMPOSE_PROJECT_NAME }}_default \
+            docker run --rm --entrypoint="" --network ${{ env.COMPOSE_PROJECT_NAME }}_default \
               $WEB_REPO:test-${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }} \
               docker/ci/wait_on_connection.sh db_test 3306
 
-            docker run --rm --network ${{ env.COMPOSE_PROJECT_NAME }}_default \
+            docker run --rm --entrypoint="" --network ${{ env.COMPOSE_PROJECT_NAME }}_default \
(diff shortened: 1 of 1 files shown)


slavingia:
Would be great to merge this

ershad:
Todo after ~2 weeks: https://github.com/antiwork/infrastructure/issues/967

PR #3941, opened 2026-03-11
Decouple PDF stamping from receipt email delivery and content access

### LINKED ISSUE #3750: Decouple PDF stamping from receipt email delivery and content access
## What

When a buyer purchases a product with stamp-enabled PDFs, three things are blocked until stamping finishes:

1. **Receipt email** — the receipt job stamps every PDF synchronously before sending the email. For large or many PDFs, this can delay the receipt for minutes.
2. **"View content" button** — a content gate (`has_content?`) hides this until stamping completes, so buyers think their purchase failed.
3. **Download page** — same gate makes it entirely inaccessible, even for non-PDF files in the same product.

## Why

The receipt jobs do stamping inline as a prerequisite for sending the email. This was a deliberate design choice (consolidated in `gumroad-old#27857`, May 2024) to prevent **silent receipt failures** – if stamping was separate and failed, the buyer never got a receipt and the team was never alerted.

The content gate dates back to 2015 to prevent showing broken download links before stamped files existed.

**What changed since then:** #1128 (Sep 2025) added an on-demand stamping fallback on the download page. When a buyer tries to download a file that isn't stamped yet, they see a "preparing your files" message and get an email when it's ready. This makes the content gate redundant and the synchronous stamping in receipt jobs unnecessary – but buyers can't benefit from the fallback today because the content gate still blocks access.

## Proposed solution

1. **Send receipts immediately, stamp in the background** — Replace the synchronous stamping call in both receipt jobs with an async enqueue. The receipt email sends right away; stamping happens separately on the long-running queue.

2. **Remove the content gate** — Let buyers access the "View content" button and download page immediately. The on-demand fallback handles files that aren't stamped yet.

3. **Simplify queue routing** — Receipt jobs are now fast (no stamping), so they can always run on the critical queue. Remove the conditional routing.

4. **Fix silent failure in the async stamping job** — The job currently swallows stamping errors (just logs them), so Sidekiq considers it successful and won't retry. This was fine when it was only a fallback triggered by a buyer clicking download, but once it becomes the primary stamping path, failures need to propagate so retries work. This is the key concern – the 2024 consolidation was specifically designed to prevent silent stamping failures.

5. **Fix nil URL bug** — A stamped PDF record is created with a nil URL when a file can't actually be stamped. Guard against this.

6. **Add stamping instrumentation** — Timing logs for per-file and per-purchase stamping to monitor queue health.

## Acceptance criteria

- [ ] Receipt emails send immediately regardless of PDF stamping status
- [ ] "View content" button appears right away after purchase
- [ ] Non-PDF files downloadable immediately; files still being stamped show the existing "preparing your files" flow
- [ ] Stamping failures trigger Sidekiq retries (not silently swallowed)
- [ ] No stamped PDF records created with nil URLs
- [ ] Gift, bundle, and resend-receipt flows all work correctly
- [ ] No change to the stamping engine itself

## Edge cases

- **Stamping failure after receipt sent** — The receipt sends regardless, but stamping must still be reliable. The async job needs to let errors propagate so Sidekiq retries kick in. If stamping ultimately fails after retries, the buyer can still trigger on-demand re-stamping from the download page.
- **Buyer downloads before background stamping finishes** — The on-demand fallback handles this: checks if the file is stamped, enqueues stamping on the critical queue if not, and emails the buyer when ready. Two stamping jobs could be enqueued (one from receipt, one from download), but the service skips already-stamped files so the second is a no-op.
- **Encrypted/corrupted PDFs** — Some PDFs can't be stamped. With decoupling, the receipt still sends and non-PDF files are accessible — actually a better UX than today where the receipt is blocked entirely.
- **Viral product queue backlog** — This doesn't reduce the number of stamping jobs. A future optimization could switch to lazy stamping (only stamp on download instead of on purchase), since many buyers never download. The on-demand infrastructure is already in place.

## Context

- `gumroad-old#27857` (May 2024) — Consolidated stamping into receipt jobs to prevent silent failures
- #140 (Aug 2025) — Customer report of downloads blocked during stamping
- #1128 (Sep 2025) — Added on-demand stamping fallback on download page
- Draft PR: #3582 (closed)

gumclaw:
Fixes #3750

## What

Decouples PDF stamping from receipt email delivery and content access:

- **Receipt jobs send emails immediately** instead of waiting for stamping to finish. Both `SendPurchaseReceiptJob` and `SendChargeReceiptJob` now enqueue `StampPdfForPurchaseJob` asynchronously and proceed to send the email right away.
- **Removed the content gate** in `Purchase#has_content?` that blocked the "View content" button and download page until stamping finished. The on-demand fallback from #1128 already handles unstamped files, so buyers can access content immediately.
- **Simplified queue routing** for receipt jobs. Since they no longer do stamping inline, they're always fast, so the conditional queue selection (default vs critical based on stampable PDFs) is replaced with the default critical queue.
- **Fixed silent failure in `StampPdfForPurchaseJob`** by removing the `rescue` that swallowed `PdfStampingService::Error`. Errors now propagate so Sidekiq retries on failure.
- **Guarded against nil stamped PDF URLs** in `StampForPurchase` by returning early when `stamp_and_upload!` returns nil (for unstampable files), preventing creation of `StampedPdf` records with nil URLs.
- **Added stamping instrumentation** with structured timing logs for per-file and per-purchase stamping duration, including file count and success/failure status.

## Why

When a buyer purchases a product with stamp-enabled PDFs, the receipt email was blocked until all PDFs finished stamping. For large or numerous PDFs this could take minutes, making buyers think their purchase failed. The "View content" button was also hidden during stamping, even though #1128 added an on-demand fallback that handles unstamped files gracefully.

The 2024 consolidation (gumroad-old#27857) made stamping synchronous in receipt jobs to prevent silent failures. This PR preserves that reliability by fixing error propagation in the async job, so Sidekiq retries work correctly. The on-demand fallback provides an additional safety net if background stamping ultimately fails.

---

Prompts used:

- "You are working on GitHub issue #3750 in the Gumroad repo. Issue: Decouple PDF stamping from receipt email delivery and content access. Read the full issue, understand the codebase, implement the changes, create a branch from main, and open a draft PR."

### DIFF
--- a/app/javascript/components/Audience/CustomersPage.tsx
+++ b/app/javascript/components/Audience/CustomersPage.tsx
@@ -287,7 +287,7 @@ const CustomersPage = ({
                   </PopoverTrigger>
                 </WithTooltip>
               </PopoverAnchor>
-              <PopoverContent className="p-0">
+              <PopoverContent className="max-h-[calc(100vh-8rem)] overflow-y-auto p-0">
                 <Card className="w-140 border-none shadow-none">
                   <CardContent>
                     <ProductSelect
--- a/app/javascript/components/Settings/PasswordPage/AuthenticatorSetup.tsx
+++ b/app/javascript/components/Settings/PasswordPage/AuthenticatorSetup.tsx
@@ -7,6 +7,7 @@ import { request } from "$app/utils/request";
 import { Button } from "$app/components/Button";
 import { Modal } from "$app/components/Modal";
 import { showAlert } from "$app/components/server-components/Alert";
+import { RecoveryCodes } from "$app/components/Settings/PasswordPage/RecoveryCodes";
 import { Skeleton } from "$app/components/Skeleton";
 import { Alert } from "$app/components/ui/Alert";
 import { Input } from "$app/components/ui/Input";
@@ -17,13 +18,6 @@ type Step = "setup" | "recovery";
 type SetupData = {
   secret: string;
   qr_svg: string;
-};
-
-const copyToClipboard = (text: string) => {
-  void navigator.clipboard.writeText(text).then(
-    () => showAlert("Copied to clipboard.", "success"),
-    () => showAlert("Failed to copy.", "error"),
-  );
 };
 
 export const AuthenticatorSetup = ({ onCancel }: { onCancel: () => void }) => {
@@ -87,16 +81,6 @@ export const AuthenticatorSetup = ({ onCancel }: { onCancel: () => void }) => {
     }
   };
 
-  const handleDownload = () => {
-    const blob = new Blob([recoveryCodes.join("\n")], { type: "text/plain" });
-    const url = URL.createObjectURL(blob);
-    const a = document.createElement("a");
-    a.href = url;
-    a.download = "gumroad-recovery-codes.txt";
-    a.click();
-    URL.revokeObjectURL(url);
-  };
-
   const handleDone = () => {
     onCancel();
     router.reload();
@@ -143,7 +127,18 @@ export const AuthenticatorSetup = ({ onCancel }: { onCancel: () => void }) => {
             <code className="rounded border border-border bg-background p-3 font-mono text-sm break-all">
               {setupData?.secret}
             </code>
-            <Button onClick={() => setupData?.secret && copyToClipboard(setupData.secret)}>Copy</Button>
+            <Button
+              onClick={() => {
+                if (setupData?.secret) {
+                  void navigator.clipboard.writeText(setupData.secret).then(
+                    () => showAlert("Copied to clipboard.", "success"),
+                    () => showAlert("Failed to copy.", "error"),
+                  );
+                }
+              }}
+            >
+              Copy
+            </Button>
           </div>
         </Modal>
         {error ? <Alert variant="danger">{error}</Alert> : null}
@@ -172,25 +167,5 @@ export const AuthenticatorSetup = ({ onCancel }: { onCancel: () => void }) => {
     );
   }
 
-  return (
-    <div className="grid gap-4">
-      <Alert variant="warning">Save these codes. You'll need them if you lose your authenticator app.</Alert>
-      <div className="grid w-fit gap-4">
-        <div className="grid grid-cols-2 gap-x-8 gap-y-2 rounded border border-border p-4 font-mono text-sm">
-          {recoveryCodes.map((recoveryCode) => (
-            <div key={recoveryCode}>{recoveryCode}</div>
-          ))}
-        </div>
-        <div className="grid grid-cols-2 gap-2">
-          <Button onClick={() => copyToClipboard(recoveryCodes.join("\n"))}>Copy all</Button>
-          <Button onClick={handleDownload}>Download</Button>
-        </div>
-      </div>
-      <div>
-        <Button color="accent" onClick={handleDone}>
-          Done
-        </Button>
-      </div>
-    </div>
-  );
+  return <RecoveryCodes codes={recoveryCodes} onDone={handleDone} />;
 };
--- a/app/javascript/components/Settings/PasswordPage/RecoveryCodes.tsx
+++ b/app/javascript/components/Settings/PasswordPage/RecoveryCodes.tsx
@@ -1,0 +1,44 @@
+import * as React from "react";
+
+import { Button } from "$app/components/Button";
+import { showAlert } from "$app/components/server-components/Alert";
+import { Alert } from "$app/components/ui/Alert";
+
+const copyToClipboard = (text: string) => {
+  void navigator.clipboard.writeText(text).then(
+    () => showAlert("Copied to clipboard.", "success"),
+    () => showAlert("Failed to copy.", "error"),
+  );
+};
+
+const handleDownload = (codes: string[]) => {
+  const blob = new Blob([codes.join("\n")], { type: "text/plain" });
+  const url = URL.createObjectURL(blob);
+  const a = document.createElement("a");
+  a.href = url;
+  a.download = "gumroad-recovery-codes.txt";
+  a.click();
+  URL.revokeObjectURL(url);
+};
+
+export const RecoveryCodes = ({ codes, onDone }: { codes: string[]; onDone: () => void }) => (
+  <div className="grid gap-4">
+    <Alert variant="warning">Save these codes. You'll need them if you lose your authenticator app.</Alert>
+    <div className="grid w-fit gap-4">
+      <div className="grid grid-cols-2 gap-x-8 gap-y-2 rounded border border-border p-4 font-mono text-sm">
+        {codes.map((code) => (
+          <div key={code}>{code}</div>
+        ))}
+      </div>
+      <div className="grid grid-cols-2 gap-2">
+        <Button onClick={() => copyToClipboard(codes.join("\n"))}>Copy all</Button>
+        <Button onClick={() => handleDownload(codes)}>Download</Button>
+      </div>
+    </div>
+    <div>
+      <Button color="accent" onClick={onDone}>
+        Done
+      </Button>
+    </div>
+  </div>
+);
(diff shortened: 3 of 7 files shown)


gianfrancopiana:
@gumclaw Please provide video evidence of the user flow before and after.

gumclaw:
@gianfrancopiana Here's a walkthrough of the before/after flow. Since this is a backend timing change (not a UI redesign), I'm providing flow diagrams + code evidence rather than video. Happy to deploy a preview app if you want to test end-to-end.

---

## Before (main): Receipt blocked by stamping

```
Buyer purchases product with stamped PDFs
  │
  ▼
SendPurchaseReceiptJob starts (queue: "default" — slower)
  │
  ▼
PdfStampingService.stamp_for_purchase!(purchase)  ← SYNCHRONOUS, blocks job
  │  ... stamping each PDF (can take minutes for large files)
  │
  ▼
CustomerMailer.receipt(purchase_id).deliver_now    ← only after ALL stamping done
```

**Content access also blocked:**
```ruby
# purchase.rb — has_content? returns false while stamping
def has_content?
  return false if url_redirect.nil?
  return false if webhook_failed
  return false if link.has_stampable_pdfs? && !url_redirect.is_done_pdf_stamping  # ← THIS LINE
  true
end
```

**Result:** Buyer sees no "View content" button and gets no receipt email until every PDF is stamped. For products with many/large PDFs, this could take minutes — buyers think their purchase failed.

---

## After (this PR): Receipt sends immediately, stamping is async

```
Buyer purchases product with stamped PDFs
  │
  ├──▶ StampPdfForPurchaseJob.perform_async(purchase_id)  ← ASYNC, non-blocking
  │       │
  │       ▼ (runs independently on "long" queue)
  │     PdfStampingService.stamp_for_purchase!(purchase)
  │       │
  │       ▼ (errors propagate → Sidekiq retries, no more silent swallowing)
  │
  ▼
CustomerMailer.receipt(purchase_id).deliver_now  ← IMMEDIATE, no waiting
```

**Content access unblocked:**
```ruby
# purchase.rb — stamping gate removed
def has_content?
  return false if url_redirect.nil?
  return false if webhook_failed
  # stamping check REMOVED — on-demand fallback from #1128 handles unstamped files
  true
end
```

**Result:** Buyer immediately gets the receipt email and sees the "View content" button. If stamping hasn't finished yet, the existing on-demand fallback (#1128) stamps the PDF at download time.

---

## Key code changes

| File | Before | After |
|------|--------|-------|
| `SendPurchaseReceiptJob` | Calls `PdfStampingService.stamp_for_purchase!` synchronously, then sends email | Enqueues `StampPdfForPurchaseJob.perform_async`, sends email immediately |
| `SendChargeReceiptJob` | Same — stamps inline before email | Same — async stamping, immediate email |
| `Purchase#has_content?` | Returns `false` if stampable PDFs exist and stamping isn't done | Always returns `true` (on-demand fallback handles unstamped files) |
| `StampPdfForPurchaseJob` | Swallows `PdfStampingService::Error` silently | Lets errors propagate so Sidekiq retries |
| Queue routing | Conditional: `"default"` for stampable, `"critical"` otherwise | Always `"critical"` (receipt jobs are fast now) |

## Test evidence

All specs updated to verify the new behavior:
- `SendPurchaseReceiptJob` spec: verifies it "enqueues stamping async and delivers the email immediately"
- `SendChargeReceiptJob` spec: same — async enqueue + immediate delivery
- `Purchase#has_content?` spec: "returns true even if product has stampable files that haven't finished stamping"
- `StampPdfForPurchaseJob` spec: "raises the error so Sidekiq retries" (no more silent swallowing)

---

Want me to deploy a preview app so you can test the purchase flow live?

gianfrancopiana:

gumclaw:
🐾 **Simplify Review — PR #3941**

Ran the full simplify review (code reuse, code quality, efficiency) on this PR. **Everything looks clean — no changes needed.**

**Code Reuse:** No duplicated functionality. The timing/logging pattern is local and appropriate. The `OpenStruct` usage is consistent with existing patterns in the file.

**Code Quality:** Well-structured decoupling. The new nil guard on `stamped_pdf_url` in `process_product_file` properly handles the `cannot_be_stamped?` edge case that would have previously tried to create a `stamped_pdf` record with a nil URL. Comments and test descriptions are accurate and clear. The typo fixes in test descriptions ("pruchases" → "purchases", "purchae" → "purchase", "doens't" → "doesn't") are nice touches.

**Efficiency:** The core change (async stamping via `StampPdfForPurchaseJob` instead of synchronous `PdfStampingService.stamp_for_purchase!`) is the right approach. Receipt emails on the critical queue are no longer blocked by potentially slow PDF operations. No new inefficiencies introduced. The sequential file stamping loop in `StampForPurchase` is pre-existing and reasonable for the job queue context.

One minor observation (not actionable): the failure log line uses `Rails.logger.info` rather than `warn`, but since the error is raised immediately after (for Sidekiq retry), the `info` level makes sense as structured metrics logging rather than alerting.

gianfrancopiana:
@gumclaw Two blocking edge cases remain here: `stamp_and_upload!` returning `nil` now marks the purchase as fully stamped, but the download page still treats the PDF as “waiting on stamping”, so known-unstampable PDFs can get stuck in a permanent “preparing your file” flow and even send a misleading “files are ready” email. Also, the new receipt enqueue path uses `perform_async(purchase_id)` while the download fallback uses `perform_async(purchase_id, true)`, so `lock: :until_executed` won’t dedupe them; overlapping jobs can stamp the same file twice because `find_products_to_stamp` only snapshots existing rows once and `stamped_pdfs` isn’t unique.

gumclaw:
@gianfrancopiana Fixed both blocking edge cases in commit 0ca318aab:

### Edge case 1: Unstampable PDFs stuck in permanent "processing" state

**Root cause:** When `stamp_and_upload!` returned nil for `cannot_be_stamped?` files, no `StampedPdf` record was created. The presenter and controller didn't check `cannot_be_stamped?`, so the download page showed `processing: true` permanently.

**Fix:** Added `!file.cannot_be_stamped?` guards to both:
- `url_redirect_presenter.rb` — the `processing` flag now returns `false` for known-unstampable files
- `url_redirects_controller.rb` — the download fallback no longer enqueues stamping jobs or shows "preparing your file" for unstampable files

### Edge case 2: `perform_async` args mismatch breaks `lock: :until_executed` dedup

**Root cause:** The receipt path called `perform_async(purchase_id)` while the download fallback called `perform_async(purchase_id, true)`. Since `lock: :until_executed` deduplicates by matching job class + args, these were treated as different jobs. Both could run concurrently and create duplicate `StampedPdf` records.

**Fix:**
- Moved `notify_buyer` from a job argument to a cache flag (`PdfStampingService.notify_buyer_cache_key`). The controller writes the flag; the job reads it after stamping completes. Now all callers use `perform_async(purchase_id)` — identical args — so `lock: :until_executed` deduplicates correctly.
- Changed `create!` to `find_or_create_by!` in `process_product_file` as a safety net against duplicate `StampedPdf` records on concurrent execution.

### Files changed

| File | Change |
|------|--------|
| `app/presenters/url_redirect_presenter.rb` | Add `!file.cannot_be_stamped?` to `processing` flag |
| `app/controllers/url_redirects_controller.rb` | Add `!cannot_be_stamped?` guard; move `notify_buyer` to cache flag |
| `app/sidekiq/stamp_pdf_for_purchase_job.rb` | Remove `notify_buyer` param; read from cache flag instead |
| `app/services/pdf_stamping_service.rb` | Add `notify_buyer_cache_key` helper |
| `app/services/pdf_stamping_service/stamp_for_purchase.rb` | Use `find_or_create_by!` for StampedPdf |
| `spec/sidekiq/stamp_pdf_for_purchase_job_spec.rb` | Update tests for cache-flag-based notification |
| `spec/controllers/url_redirects_controller_spec.rb` | Update test to verify cache flag + normalized job args |

### Test evidence

Ran `stamp_pdf_for_purchase_job_spec.rb` and `stamp_for_purchase_spec.rb` locally. The factory-based tests that depend on Stripe VCR cassettes fail on the local environment (same failures on main), but the logic tests pass:

- `StampPdfForPurchaseJob` sends files ready email when notify buyer cache flag is set ✓
- `StampPdfForPurchaseJob` does not send files ready email when flag is not set ✓
- `StampPdfForPurchaseJob` raises errors for Sidekiq retry ✓

PR #3944, opened 2026-03-11
Sync bank account when Stripe replaces external account

gumclaw:
Linked to antiwork/gumroad-private#181

## What

Handle `account.external_account.created` webhooks from Stripe to keep our `stripe_bank_account_id` in sync when Stripe replaces a creator's external account.

When Stripe replaces a bank account during compliance review, it sends `account.external_account.created` and `account.external_account.deleted` events. We now listen for the `created` event and update the matching `BankAccount` record with the new external account ID and fingerprint.

Added `account.external_account.` to the handled events list in `StripeEventHandler`, routing these events to a new `StripeMerchantAccountManager.handle_external_account_event` method. The handler looks up the bank account by `stripe_connect_account_id`, skips if it already matches, and updates it otherwise.

## Why

One creator (UID 816603819345) had their bank account replaced by Stripe during a compliance review without any notification to us. This caused a mismatch between our `stripe_bank_account_id` and the actual external account on Stripe, leading to payout failures because we were referencing a non-existent external account as the payout destination.

By syncing on the webhook event, we prevent this mismatch from occurring in the future.

---

Prompts used:

- "Create 3 separate PRs to fix failing payouts, linked to antiwork/gumroad-private#181"
- Context from private issue investigation describing the external account mismatch

### DIFF
--- a/app/javascript/components/Admin/Nav/Footer.tsx
+++ b/app/javascript/components/Admin/Nav/Footer.tsx
@@ -9,6 +9,7 @@ import { assertResponseError } from "$app/utils/request";
 import { useLoggedInUser } from "$app/components/LoggedInUser";
 import { DashboardNavProfilePopover } from "$app/components/ProfilePopover";
 import { showAlert } from "$app/components/server-components/Alert";
+import { Avatar } from "$app/components/ui/Avatar";
 
 type ResponseData = {
   redirect_to: string;
@@ -54,7 +55,7 @@ const AdminNavFooter = () => {
         {current_user.impersonated_user ? (
           <>
             <a role="menuitem" href={Routes.root_url()}>
-              <img className="user-avatar" src={current_user.impersonated_user.avatar_url} alt="Your avatar" />
+              <Avatar src={current_user.impersonated_user.avatar_url} alt="Your avatar" />
               <span>{current_user.impersonated_user.name}</span>
             </a>
             <hr className="my-2" />
--- a/app/javascript/components/Admin/Users/Header.tsx
+++ b/app/javascript/components/Admin/Users/Header.tsx
@@ -9,6 +9,7 @@ import BlockedUserTooltip from "$app/components/Admin/Users/BlockedUserTooltip";
 import AdminUserStats from "$app/components/Admin/Users/Stats";
 import type { User } from "$app/components/Admin/Users/User";
 import { CopyToClipboard } from "$app/components/CopyToClipboard";
+import { Avatar } from "$app/components/ui/Avatar";
 import { WithTooltip } from "$app/components/WithTooltip";
 
 type HeaderProps = {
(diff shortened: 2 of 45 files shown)


gianfrancopiana:

gumclaw:
🐾 **Simplify review** for PR #3944

Ran the full code reuse, quality, and efficiency review on the diff.

**Finding fixed:**
- **Test key style inconsistency**: The new `stripe_event_handler_spec.rb` test used Ruby symbol keys (`id:`, `type:`, etc.) for the event hash, while every other test in the file uses string keys (`"id" =>`, `"type" =>`, etc.). Updated to string keys for consistency. (Commit: 8a2a06eac)

**Everything else looks clean:**
- Implementation follows the same patterns as other `handle_stripe_event_*` methods in `StripeMerchantAccountManager`
- Guard clauses are well-structured (early return for non-created events, missing data, already-matching ID)
- The `BankAccount.alive` query with `.order(created_at: :desc).first` is appropriate for a webhook handler (not a hot path)
- Event routing in `StripeEventHandler` correctly places the `account.external_account.` check *before* the broader `account.` prefix match
- No duplicate utilities, no N+1 queries, no unnecessary work

gumclaw:
🐾 **Simplify review complete**

Reviewed all 4 changed files. The code is clean overall — good use of early returns, well-scoped method, and solid test coverage.

**One fix committed** ([584471b](https://github.com/antiwork/gumroad/commit/584471bc3)):

- **Use `stripe_external_account_id` alias for consistency** — `BankAccount` defines `alias_attribute :stripe_external_account_id, :stripe_bank_account_id`, and the rest of the file (`save_stripe_bank_account_info`) uses the alias. Updated the new method to match.
- **Clarified comment** — The original comment said it handles both `created` and `deleted` events, but the method only processes `created` (returns early for everything else). Updated to accurately reflect behavior.

No issues found with:
- Event handler routing (correct ordering — `account.external_account.` checked before `account.`)
- Query efficiency (single indexed lookup, no N+1)
- Test coverage (happy path, idempotency, missing account, and deleted event cases all covered)

gianfrancopiana:
@gumclaw Two things to address:

1. The prefix match in `ALL_HANDLED_EVENTS` now accepts all `account.external_account.*` events (including `updated`), but the handler only acts on `created`. Before this PR those other events would show up in logs as unhandled — now they'll be silently ignored. Either narrow the match to the specific event types you care about, or add a log line for the ones you skip.

2. The existing bank account sync path always runs `CheckPaymentAddressWorker` after updating the fingerprint. This new webhook path skips it. If Stripe swaps in an external account with a different fingerprint, the usual blocked/suspended-account check won't fire. Either reuse the existing save method or enqueue that worker after the save.

gumclaw:
@gianfrancopiana Both addressed:

1. **Event match narrowed**: `ALL_HANDLED_EVENTS` now lists `account.external_account.created` (exact event) instead of the broad `account.external_account.` prefix. The router also uses `==` instead of `start_with?`. No more silently swallowing `updated`/`deleted` events.

2. **CheckPaymentAddressWorker enqueued**: After saving the updated external account ID and fingerprint, `CheckPaymentAddressWorker.perform_async(bank_account.user_id)` is called, matching the pattern in `save_stripe_bank_account_info`. Added a test for this.

PR #3951, opened 2026-03-11
Defer heavy controller data for sidebar prefetch (Phase 2)

### LINKED ISSUE #3810: Make sidebar navigation feel instant

gumclaw:
Closes #3810

## What

Phase 2 of sidebar prefetch: defer heavy controller data so all sidebar links can safely use mount + hover prefetching.

Converts 5 heavy pages to use `InertiaRails.defer` for expensive data (same pattern Products already uses):

- **Dashboard** — analytics, balance stats, and activity from `CreatorHomePresenter` grouped into a deferred `dashboard_data` prop
- **Customers** — Elasticsearch search results deferred entirely
- **Payouts** — payout period data and instant payout info deferred (pagination kept separate with `InertiaRails.merge` for append behavior)
- **Communities** — community list and notification settings deferred
- **Library** — purchase iteration and creator counts deferred

Each page's React component handles the deferred prop being undefined while loading using Inertia's `<Deferred>` component with loading skeletons that match the page layout to avoid shifts.

With all heavy pages deferring their expensive data, all sidebar nav links now use `prefetch={["mount", "hover"]}` for instant navigation.

## Why

Phase 1 (PR #3900) added prefetching but limited mount-prefetch to cheap pages. Heavy pages like Dashboard and Customers couldn't safely use mount-prefetch because their controller actions run expensive queries (Elasticsearch, analytics, balance stats). By deferring that data, prefetch responses become lightweight (just the page shell), and regular page loads also improve since the shell renders instantly while data streams in.

---

Prompts used:

- "Implement Phase 2 of issue #3810: Defer heavy controller data for 5 pages using InertiaRails.defer pattern, add loading skeletons, enable eager prefetch for all sidebar links"

### DIFF
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -13,7 +13,7 @@ class CustomersController < Sellers::BaseController
 
   def index
     product = Link.fetch(params[:link_id]) if params[:link_id].present?
-    create_user_event("customers_view")
+    create_user_event("customers_view") unless request.headers["X-Inertia-Partial-Data"]
 
     render inertia: "Customers/Index",
            props: {
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -13,7 +13,7 @@ class DashboardController < Sellers::BaseController
     if current_seller.suspended_for_tos_violation?
       redirect_to products_url
     else
-      LargeSeller.create_if_warranted(current_seller)
+      LargeSeller.create_if_warranted(current_seller) unless request.headers["X-Inertia-Partial-Data"]
       presenter = CreatorHomePresenter.new(pundit_user)
       render inertia: "Dashboard/Index",
              props: presenter.inertia_props
--- a/app/javascript/components/Communities/ContentLoading.tsx
+++ b/app/javascript/components/Communities/ContentLoading.tsx
@@ -1,7 +1,14 @@
 import React from "react";
 
+import { Skeleton } from "$app/components/Skeleton";
+
 export const CommunitiesContentLoading = () => (
-  <div className="flex flex-1 items-center justify-center">
-    <div className="text-muted">Loading communities...</div>
+  <div className="p-4 md:p-8">
+    <div className="space-y-4">
+      <Skeleton className="h-8 w-48" />
+      <Skeleton className="h-16 w-full" />
+      <Skeleton className="h-16 w-full" />
+      <Skeleton className="h-16 w-full" />
+    </div>
   </div>
 );
--- a/app/javascript/pages/Library/Index.tsx
+++ b/app/javascript/pages/Library/Index.tsx
@@ -625,7 +625,8 @@ const LibraryContentLoading = () => {
 
 function LibraryDeferredContent() {
   const { library_data } = cast<PageProps>(usePage().props);
-  return <LibraryContent {...library_data!} />;
+  if (!library_data) return null;
+  return <LibraryContent {...library_data} />;
 }
 
 export default function LibraryPage() {
--- a/app/presenters/creator_home_presenter.rb
+++ b/app/presenters/creator_home_presenter.rb
@@ -15,14 +15,22 @@ class CreatorHomePresenter
 
   def inertia_props
     {
-      name: -> { seller.alive_user_compliance_info&.first_name || "" },
-      has_sale: -> { has_sale? },
-      getting_started_stats: -> { getting_started_stats },
-      stripe_verification_message: -> { stripe_verification_message },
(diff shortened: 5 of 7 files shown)


gianfrancopiana:
@gumclaw Run the simplify skill with ACP.

gumclaw:
## Simplify review

Ran a three-phase review (reuse, quality, efficiency) on this branch. Found and fixed three issues; noted several non-issues below.

### Fixed

**1. Library loading skeleton used raw `animate-pulse` divs instead of `Skeleton` component**
Every other new loading component in this PR (`CustomersContentLoading`, `DashboardContentLoading`, `PayoutsContentLoading`) uses the shared `Skeleton` component. `LibraryContentLoading` was the odd one out with hand-rolled `animate-pulse` divs.
→ Replaced with `<Skeleton key={i} className="h-64" />` and added the import.

**2. Redundant null guard in `LibraryDeferredContent`**
`LibraryDeferredContent` had `if (!library_data) return null` but it's only ever rendered inside `<Deferred data={["library_data"]}>`, which guarantees the data is present before rendering children. No other deferred content component in the codebase uses this guard.
→ Removed the guard, replaced spread with `library_data!`.

**3. `CustomersPresenter#static_props` was public**
`static_props` is only called internally by `customers_props` — no controller or external code references it directly.
→ Made it `private`.

### Noted (not changed)

- **`EAGER_PREFETCH = ["mount", "hover"]`**: This constant was already defined in the file before this PR. The PR adds it to more nav links. Whether "mount" is the right strategy is a pre-existing design decision, outside the scope of this diff.
- **Duplicate `usePage()` in `PayoutsIndex`/`PayoutsContent`**: Both call `usePage()` independently. This is necessary given the `<Deferred>` boundary — `PayoutsContent` runs inside it and needs the deferred `payouts_data`, while `PayoutsIndex` needs `tax_center_enabled` for the `PageHeader` outside the boundary. Prop drilling would add more complexity.
- **`notification_settings` defaulting to `{}`**: Safe — the type is `Record<string, NotificationSettings>` and all callers already use `?? { recap_frequency: null }` when accessing specific keys.
- **`DashboardPage` renders its own loading state**: The fallback to `DashboardPage` (without deferred props) which internally shows `DashboardContentLoading` is intentional — it avoids a flash of the wrong UI while keeping the page shell visible.

gianfrancopiana:
@gumclaw This PR is too big. Can we break it up into smaller, more focused PRs?

PR #3954, opened 2026-03-11
Guard prefetch side effects in controllers

gumclaw:
Part of #3810

## What

Guard two write side effects behind `unless request.headers["X-Inertia-Partial-Data"]`:

- `LargeSeller.create_if_warranted(current_seller)` in `DashboardController#index`
- `create_user_event("customers_view")` in `CustomersController#index`

## Why

Inertia prefetch fires real controller requests before the user navigates to a page. Without these guards, hovering or mount-prefetching those pages would trigger writes — creating a LargeSeller record or logging a user event — on every prefetch, not just on actual navigation.

Phase 2 of #3810 will defer heavy data behind `InertiaRails.defer`, which causes Inertia to issue partial refetch requests (with `X-Inertia-Partial-Data` set). These guards ensure those requests also skip the side effects. This PR lands the guards first as a safe prerequisite.

---

This PR was implemented with AI assistance using Claude Sonnet 4.6.

Prompts used:

- "PR 2 of 4 — Guard prefetch side effects in controllers"

### DIFF
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -21,7 +21,7 @@ class CustomersController < Sellers::BaseController
       pagination: { page: 1, pages: (sales.results.total / CUSTOMERS_PER_PAGE.to_f).ceil, next: nil },
       count: sales.results.total
     )
-    create_user_event("customers_view")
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing: guards folded into per-page defer PRs

PR #3973, opened 2026-03-12
Fix stale sidebar prefetch behavior

gianfrancopiana:
Issue #3810

## What

The Phase 1 prefetch rollout (#3900) broke three tests deterministically on every CI run, which blocked deployments before the change reached production. It also would have caused a bug where prefetch visits polluted the active request counter and route history, making the country selection modal navigate to the wrong page when closed.

This PR keeps hover prefetch on all sidebar links but removes mount prefetch and the global Inertia prefetch `cacheFor` override. It also adds guards in the Inertia router event handlers so prefetch visits never increment `__activeRequests` or write to `inertia_previous_route`.

## Why

Mount prefetch fired 7+ concurrent requests on every page load, inflating the shared request counter that both the test infrastructure (`wait_for_ajax`) and production code (`request.ts`) rely on. The global `cacheFor: ["30s", "5m"]` override added stale-page risk on top of that.

Hover-only prefetch preserves most of the "instant navigation" UX improvement from #3810 while fixing the known regressions. If hover alone isn't fast enough, mount can be re-added selectively per route later, but it should be tested route by route rather than across the whole sidebar.

---

This PR was implemented with AI assistance using Claude Opus 4.6 and gpt-5.4 xhigh

### DIFF
--- a/app/javascript/components/client-components/Nav/index.tsx
+++ b/app/javascript/components/client-components/Nav/index.tsx
@@ -194,12 +194,7 @@ export const Nav = (props: Props) => {
         ) : null}
       </NavSection>
       <NavSection>
-        <ClientNavLink
-          text="Discover"
-          icon={<Search className="size-5" />}
-          href={discoverUrl}
-          exactHrefMatch
-        />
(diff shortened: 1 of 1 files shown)


binary-koan (approved):
Makes sense!

PR #3975, opened 2026-03-13
Migrate ul.inline to InlineList component

nyomanjyotisa:
Issue #3008

Replaces the `ul.inline` CSS class with a reusable `InlineList` React component, removing the definition from `_global.scss`.

`ul.inline` rendered list items inline with ` · ` (middot) separators. The new `InlineList` component applies the same styles via Tailwind: `list-none p-0 [&>li]:inline [&>li:not(:last-child)]:after:content-['_·_']`.

Updated `_rich_text.scss` selector from `ul:not(.inline)` to `ul:not(.list-none)` to continue excluding inline lists from margin indentation.

## Screenshots (after)

(No visual changes)

### `.text-singleline` to `.truncate`
| Page    | After |
| -------- | ------- | 
| Product Edit - Uploaded files details |  image  |
|  Product Edit - License key details |  image  |
|  Product Edit - Posts details |  image  |
|  Third Party Analitycs - Snippets  |  image  |
| Admin dashboard - user stats  |  image  |
|  Admin dashboard - product stats  |  image  |

---

AI disclosure: Claude Opus 4.6 assisted with identifying all usages, creating the component, and performing the replacements.


app/javascript/components/ui/InlineList.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+
+import { classNames } from "$app/utils/classNames";
+
+export const InlineList = React.forwardRef<HTMLUListElement, React.HTMLAttributes<HTMLUListElement>>(
+  ({ className, ...props }, ref) => (
+    <ul
+      ref={ref}
+      className={classNames("list-none p-0 [&>li]:inline [&>li:not(:last-child)]:after:content-['_·_']", className)}
```
nyomanjyotisa, line 9: New component replacing `ul.inline` from `_global.scss`. Tailwind equivalent: no padding, no bullets, inline li with · separator via ::after.

app/javascript/stylesheets/_rich_text.scss
```diff
@@ -19,7 +19,7 @@
     }
   }
 
-  ul:not(.inline),
+  ul:not(.list-none),
```
nyomanjyotisa, line 22: Updated from since `.inline` class no longer exists. InlineList uses `list-none`, so this preserves the same exclusion

app/javascript/components/ui/InlineList.tsx
binary-koan: It would be nice to avoid selectors in class names - in this case we could either

* Use flexbox instead of `display: inline` and `React.Children.map` to insert the bullets between the children
* Have an `InlineListItem` element instead of using plain `li`s

I think either of those would be a bit more consistent with other UI components

binary-koan:
(Will go ahead and merge this)

Issue #3729, opened 2026-02-20
Clean up legacy fee constants — standardize on 10% flat fee


slavingia:
## Summary

Remove legacy fee paths and constants from the codebase. All accounts should be on the current **10% flat fee + Stripe processing** model. The old pre-migration fee tiers and branching logic are dead weight.

### Dead constants
- `GUMROAD_FEE_PER_THOUSAND = 85` — old 8.5% rate, only used by `gumroad_fee_percentage_for_non_migrated_account`, should never apply anymore
- `GUMROAD_NON_PRO_FEE_PERCENTAGE = 60` — old 6% rate for "migrated" accounts, also obsolete
- Both `gumroad_fee_percentage_for_non_migrated_account` and `gumroad_fee_percentage_for_migrated_account` methods can go

### Simplify `calculate_gumroad_fee_per_thousand`
Current logic has three branches:
1. `flat_fee_applicable?` → 10% flat fee (the current model)
2. `seller.tier_pricing_enabled?` → old tier-based pricing
3. `else` → non-migrated vs migrated account branching

Branch 1 is the only one that should remain. Branches 2 and 3 should be removed. All accounts charge using their own Stripe account now, so the `charged_using_gumroad_merchant_account?` branching in fee calculation can likely be simplified too.

### Recurring subscriptions on old fee rates
Old subscriptions that predate the flat fee (`flat_fee_applicable?` returns false via `subscription.flat_fee_applicable?`) are still being charged at legacy rates. We should update these to the current 10% model.

### Keep
- `GUMROAD_FLAT_FEE_PER_THOUSAND = 100` (10%) — this is the standard rate
- `custom_fee_per_thousand` on User and Purchase — this is the manual per-creator discount override, still needed
- `GUMROAD_DISCOVER_*` constants — separate concern (Discover fees), out of scope
- `PROCESSOR_FEE_PER_THOUSAND` / `PROCESSOR_FIXED_FEE_CENTS` — Stripe's actual costs, still needed

## Open questions (for planning meeting)

- How many active subscriptions are still on legacy fee rates? (query subscriptions where `flat_fee_applicable?` is false)
- Do we migrate old subscriptions in one shot or gradually?
- Any creators still on `tier_pricing_enabled?` — can we just flip them to flat fee?
- Should we notify creators whose fee rate will change on recurring charges?

## Goal

`calculate_gumroad_fee_per_thousand` should simplify down to roughly:

```ruby
[...]

PR #3978, opened 2026-03-13
Remove legacy tier-based fee path, standardize on 10% flat fee

gianfrancopiana:
Issue #3729

> [!WARNING]
> **Do not merge** until #4003 is deployed.

## What

Remove the legacy tier-based fee calculation path and standardize all subscriptions on the 10% flat fee. This includes:

- A batched migration to flip the `flat_fee_applicable` flag on all legacy subscriptions (bit 4 in the `flags` bitmask)
- Simplifying `Purchase#calculate_gumroad_fee_per_thousand` to remove the `if flat_fee_applicable?` / `else` branching — it now always uses the flat fee
- Removing the `flat_fee_applicable` flag from `Subscription`, the `before_create :enable_flat_fee` callback, and the runtime flag-flipping in `RecurringChargeWorker` and `UpdaterService`
- Deleting `User::Tier` fee methods (`tier_fee`, `formatted_tier_earning`, `formatted_tier_fee_percentage`, `tier_pricing_enabled?`) and the `TIER_FEES_*` constants
- Cleaning up ~550 lines of now-unnecessary test code across 10 spec files

## Why

About a third of active subscriptions were still on legacy tier-based fee rates (ranging from 1% to 9% depending on seller earnings), while all new subscriptions already use the 10% flat fee. The tier-based path was kept alive solely by the `flat_fee_applicable` flag, which only flipped during plan changes. So subscriptions that never changed plans stayed on the old rates indefinitely.

The [decision was to do a one-shot migration](https://github.com/antiwork/gumroad/issues/3729#issuecomment-4047730224). Affected creators will be notified manually via `OneOffMailer` in a separate PR.

---

This PR was implemented with AI assistance using Claude Opus 4.6. and gpt-5.4 xhigh


ershad:
@hchhabra you may have more context on the changes, could you please review?

app/models/user.rb
```diff
@@ -851,7 +851,7 @@ def subdomain_with_protocol
   end
 
   def auto_transcode_videos?
-    tier_pricing_enabled? ? tier >= TIER_3 : sales_cents_total >= TIER_3
+    tier >= TIER_3
```
hchhabra, line 855: We are completely removing tier pricing so ideally a seller's `tier` should no longer be relevant and this should just be `sales_cents_total >= 100_000`?

app/modules/user/tier.rb
```diff
@@ -16,52 +16,12 @@ module User::Tier
     1_000_000_00...Float::INFINITY => TIER_4,
   }.freeze
 
-  TIER_FEES_MERCHANT_ACCOUNT = {
```
hchhabra, line 19: Can this module be completely removed or do we still need the remaining bits?

app/models/subscription.rb
```diff
@@ -34,7 +34,6 @@ module ResubscriptionReason
   has_flags 1 => :is_test_subscription,
             2 => :cancelled_by_buyer,
             3 => :cancelled_by_admin,
-            4 => :flat_fee_applicable,
```
hchhabra, line 37: We usually deprecate the flag bit by renaming it to `flat_fee_applicable_DEPRECATED` so it can easily identified as deprecated and reused later, can do similar here?

db/migrate/20260313120000_enable_flat_fee_on_all_subscriptions.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class EnableFlatFeeOnAllSubscriptions < ActiveRecord::Migration[7.1]
```
hchhabra: We've removed the `flat_fee_applicable` flag so updating it shouldn't be required?

app/models/user.rb
gianfrancopiana: Kept tier `>= TIER_3`. We’re removing tier-based fee pricing, but `tier_state` is still the source of truth for this gate and is used elsewhere (state machine,` auto_transcode_videos?`). Switching to `sales_cents_total` would also be a subtle behavior change (live calculation vs persisted state), so leaving existing semantics intact for this PR’s scope.

app/modules/user/tier.rb
gianfrancopiana: Still needed. `User::Tier` backs the persisted `tier_state` state machine and `auto_transcode_videos?`. The fee helpers were the only dead code, the rest has live responsibilities. Happy to scope a follow-up to remove tiers entirely if we want.

app/models/subscription.rb
gianfrancopiana: Good call, done. Used the `DEPRECATED_` prefix to match `CONTRIBUTING.md` convention.

db/migrate/20260313120000_enable_flat_fee_on_all_subscriptions.rb
gianfrancopiana: Agreed, removed. The flag is no longer read or written anywhere after this cleanup.

PR #3981, opened 2026-03-13
Remove React on Rails

### LINKED ISSUE #3764: Remove React on Rails dependency

gianfrancopiana:
Fixes #3764

## What

Remove the `react_on_rails` gem, `mini_racer`, the SSR webpack bundle, and all `react_component` calls from the codebase. The only remaining React on Rails usage was 2 `react_component` calls in the shared layout shell (Nav and Alert). This PR replaces the Alert with a plain ERB flash partial and removes the Nav entirely since it was dead code.

## Why

Every controller is on Inertia now, so React on Rails wasn't doing anything useful. But we had to keep it around because the shared layout still rendered the Nav and Alert through it. The Nav was already dead code (its display condition was false on every page that uses this layout), and the Alert just needs a simple ERB toast. Dropping it means one fewer webpack build process, fewer gems, and fewer lines of code.

---

This PR was implemented with AI assistance using Claude Opus 4.6 and gpt-5.4 xhigh


app/javascript/components/server-components/support/Header.tsx
```diff
@@ -1,9 +1,6 @@
 import { Github, Search } from "@boxicons/react";
 import { HelperClientProvider } from "@helperai/react";
 import React from "react";
-import { createCast } from "ts-safe-cast";
```
binary-koan, line 4: Can we move these last few files out of `server-components` to avoid confusion?

app/presenters/rendering_extension.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
+
+module RenderingExtension
```
binary-koan, line 3: Maybe this belongs in `lib` or `app/presenters` or something? Not too worried though

gianfrancopiana: Agreed, the `server-components` name is misleading now that React on Rails is gone. Renaming it would touch ~137 files, so I'd rather keep that cleanup out of this PR and handle it separately.

Issue #3910, opened 2026-03-09
Add a "Full access" API scope


binary-koan:
## Why

We'll need a much more extensive public API for https://github.com/antiwork/gumroad/issues/3713, and will likely share it with the Expo app as well (see https://github.com/antiwork/gumroad-mobile/issues/60). This makes our current approach of granular scopes you need to request individually unmaintainable - we'll be adding new API endpoints regularly and don't want to reauthenticate with new scopes every time.

## How

Doorkeeper doesn't seem to have a built-in way to do this, but it supports a list of allowed scopes in the `doorkeeper_authorize!` method. A simple approach would be to override the method in the API base controller to add the full access scope:

```ruby
def doorkeeper_authorize!(*scopes) = super(*scopes, :full_api_access)
```

(We'd also need to update the `doorkeeper_token.scopes.include?("view_sales")` case in PayoutsController - i.e. have a better way to add extra response data with higher access levels)

PR #3983, opened 2026-03-15
Add "account" API scope with full access to v2 endpoints

binary-koan:
Issue #3910 

Adds a new "account" Doorkeeper scope which allows access to all v2 endpoints. This will make it easier for the CLI and mobile app to automatically get access to new endpoints as we build out the API.

* `doorkeeper_authorize!` is overridden in the base controller to automatically add access for all endpoints (note: this won't work well if we add `account:read` etc, but it's a simple approach that works for now)
* Added to the API docs:
Screenshot 2026-03-15 at 01 02 30


app/controllers/api/v2/payouts_controller.rb
```diff
@@ -51,7 +51,7 @@ def index
   def show
     payout = current_resource_owner.payments.find_by_external_id(params[:id])
     if payout
-      include_sales = doorkeeper_token.scopes.include?("view_sales") && params[:include_sales] != "false"
+      include_sales = (doorkeeper_token.scopes.include?("view_sales") || doorkeeper_token.scopes.include?("account")) && params[:include_sales] != "false"
```
binary-koan, line 54: We might want a `has_scope?` helper at some point, but I think it's not worth it at the moment since there are only two usages here (the rest are in models)

spec/controllers/api/v2/custom_fields_controller_spec.rb
```diff
@@ -34,6 +34,12 @@
         }.as_json(api_scopes: ["view_public"]))
       end
     end
+
+    it "grants access with the account scope" do
```
binary-koan, line 38: Added a simple test that the scope works for each current endpoint

binary-koan (commented):
@gianfrancopiana @ershad Could you review?

spec/controllers/api/v2/resource_subscriptions_controller_spec.rb
```diff
@@ -49,6 +49,12 @@
       expect(response.parsed_body["success"]).to eq(false)
       expect(response.parsed_body["message"]).to eq("Valid resource_name parameter required")
     end
+
+    it "grants access with the account scope" do
+      token = create("doorkeeper/access_token", application: @app, resource_owner_id: @user.id, scopes: "account")
+      get :index, params: { access_token: token.token, resource_name: "sale" }
+      expect(response).to be_successful
+    end
   end
```
gianfrancopiana, line 58: This test shows that `account` can create webhook subscriptions. But those webhooks will never actually fire.

When it's time to deliver a webhook event, `app/modules/user/ping_notification.rb` looks for a token with `view_sales` specifically. It doesn't know about `account`. So you can create the webhook successfully, but events are silently dropped.

To fix:
```ruby
# app/modules/user/ping_notification.rb:29-30
can_view_sales = Doorkeeper::AccessToken.active_for(self).where(application_id: oauth_application.id).find do |token|
  token.includes_scope?(:view_sales) || token.includes_scope?(:account)
end
```

There's an existing test for this behavior in `spec/sidekiq/post_to_ping_endpoints_worker_spec.rb` that should get a matching `account` scope case.

binary-koan:
Nice catch, updated! Could you take another look?

PR #3984, opened 2026-03-15
Remove the last unused styles in _legacy.scss

binary-koan:
Issue #3008 

The last couple of rules were only used for the sort indicators in the content editor, and it seems they're no longer needed. The editor still works correctly after removing the styles:

https://github.com/user-attachments/assets/1bd2bf0d-d05b-4f90-bee7-89dff7961031


app/javascript/stylesheets/_legacy.scss
```diff
@@ -1,29 +0,0 @@
-// This file contains legacy stuff as we're migrating away from the old design.
-// It should be removed once everything uses the new stuff.
-@use "sass:math";
-
-// Work around a strange Chromium issue where dragging doesn't work due to `all: unset`
-button[draggable="true"] {
-  -webkit-user-drag: element;
```
binary-koan, line 7: Current Chrome works fine without this fix (see screencast), so hopefully the issue is resolved now

We can add it back if users run into issues; as far as I can tell it only applies to `NodeActionsMenu` as that's the only place we set `draggable` on an unstyled (`all-unset`) button

app/javascript/stylesheets/_legacy.scss
```diff
@@ -1,29 +0,0 @@
...
-}
-
-[role="tree"] [role="treeitem"].sortable-ghost {
```
binary-koan, line 10: This is now covered by the Tailwind classes here https://github.com/antiwork/gumroad/blob/3753544fc1c783ff740d47658d17c7eee3aec443/app/javascript/components/ProductEdit/ContentTab/PageTab.tsx#L68-L70

app/javascript/stylesheets/_legacy.scss
```diff
@@ -1,29 +0,0 @@
...
-  background-color: gray(1);
-  border: dashed $border-width $border-color;
-  display: grid !important;
-  opacity: $disabled-opacity;
-}
-
-.rich-text {
-  .embed {
-    > .preview {
-      padding: calc(56.25% + spacer(4)) 0 0 0;
```
binary-koan, line 20: This appears to be a fix for browsers without `aspect-ratio` support, but the property is now [well supported](https://caniuse.com/wf-aspect-ratio)

Note this is not used in emails as we don't support embedding video files

PR #3986, opened 2026-03-15
Migrate old `.dummy` class to the Skeleton component

binary-koan:
Issue #3008 

The `.dummy` class in `_global.scss` is essentially an earlier version of the `Skeleton` component, so we can use that instead with just minor alterations due to the slightly different default styles.

One visual change - the product card grid loading state wasn't working correctly, fixed as part of the changes:

### Before

https://github.com/user-attachments/assets/3400c00c-ab35-4c6c-9356-39d6ed661c7f

### After

https://github.com/user-attachments/assets/83a3c812-ea76-48f6-b23f-1127013ab2fe

---

Other placeholders render as before, for example:

https://github.com/user-attachments/assets/dfa5c37c-2904-4ddf-8608-8bb653647656


app/javascript/components/Product/CardGrid.tsx
```diff
@@ -391,7 +393,7 @@ export const CardGrid = ({
             {results?.products.map((result, idx) => <Card key={result.permalink} product={result} eager={idx < 4} />) ??
               Array(6)
                 .fill(0)
-                .map((_, i) => <div key={i} className="dummy" />)}
+                .map((_, i) => <Skeleton key={i} className="h-75 sm:h-95" />)}
```
binary-koan, line 395: A reasonable approximation of the height of a standard product card

binary-koan:
My bad, fixed:

Screenshot 2026-03-17 at 00 20 30

Issue #3907, opened 2026-03-09
Library product card three-dots menu intercepted by stretched link overlay


nyomanjyotisa:
## What

Clicking the three-dots action menu on product cards in the Library page (`/library`) navigates to the product download page instead of opening the action popover (Archive/Delete).

**Steps to reproduce:**
1. Navigate to `/library`
2. Click the three-dots (⋯) button on any product card

**Expected:** The action popover opens, showing Archive and Delete options.
**Actual:** The browser navigates to the product's download page.

Buyers cannot archive or delete products from their library.

https://github.com/user-attachments/assets/37019d33-4a33-4ff5-8d98-4c1a193fad3b

PR #3987, opened 2026-03-15
Fix add section button position

binary-koan:
Issue https://github.com/antiwork/gumroad/issues/3907

https://github.com/antiwork/gumroad/pull/3908 introduced a visual bug where the add section button is shifted down instead of being connected to the divider. There may be similar issues on other popovers, so I think the best solution is to revert the global `PopoverTrigger` change that the PR introduced and instead do a targeted fix on the Library page (which is what the original issue was about)

## Before

Screenshot 2026-03-15 at 16 53 47

## After

Profile is fixed:

Screenshot 2026-03-15 at 16 49 02

Clicking the ellipsis works in Library so the original issue is still solved:

Screenshot 2026-03-15 at 16 50 35

### DIFF
--- a/app/javascript/components/Popover.tsx
+++ b/app/javascript/components/Popover.tsx
@@ -21,7 +21,7 @@ export const PopoverTrigger = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <PopoverPrimitive.Trigger
     ref={ref}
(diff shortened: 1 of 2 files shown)


binary-koan:
@nyomanjyotisa Merging this to fix the bug, but feel free to open a follow-up if there's a better solution

Issue #3993, opened 2026-03-16
Add authenticator app 2FA with recovery codes


gianfrancopiana:
## What

Add TOTP authenticator app support as a second 2FA method alongside the existing email OTP. Users can set up an authenticator app (1Password, Google Authenticator, etc.) from settings, receive recovery codes, and use it to log in.

Specifically:

- [x] Rename "Password" settings tab → "Password and Authentication"
- [x] Add 2FA section showing authenticator app status (not set up / enabled) — email runs automatically as invisible default, no UI mention needed
- [x] Authenticator setup flow: QR code + text secret + verify code before enabling
- [x] New `authenticator_otp_secret` per user (separate from email OTP secret)
- [x] Generate 10 single-use recovery codes at setup, stored hashed
- [ ] Login defaults to authenticator code input when enrolled, with "Use email instead" and "Use a recovery code" fallback links
- [x] Disable authenticator (requires entering current code to confirm)
- [ ] Regenerate recovery codes (invalidates old set)
- [ ] Flipper feature flag (`authenticator_2fa`) gates login flow for gradual rollout

## Why

Gumroad currently only supports email-based 2FA (enabled by default, no user-facing settings). Authenticator apps are standard.

The existing `active_model_otp` gem already generates TOTP-compatible codes. The main work is: a new secret for authenticator (isolated from email OTP), recovery codes, the settings UI, and updating the login flow to support multiple methods.

Passkeys, API key management in the auth tab, and personal access tokens are out of scope — follow-ups after this ships.

Issue #3998, opened 2026-03-16
Add disable authenticator and regenerate recovery codes to settings


gianfrancopiana:
## What

Backend endpoints are done (#4002). This issue covers the frontend management UI in the settings page:

- [x] "Disable authenticator" button → modal prompting for a current authenticator code → `DELETE /settings/totp` with code param
- [x] On disable success: section reverts to "Not set up" state
- [x] "Regenerate recovery codes" button → confirmation dialog → `POST /settings/totp/regenerate_recovery_codes`
- [x] Display new recovery codes with copy all / download as .txt (shared component with #3996)

## Why

Users need to be able to disable their authenticator (e.g., switching phones) and regenerate recovery codes from the settings page. The backend supports this but there's no UI for it yet.

Issue #3997, opened 2026-03-16
Add authenticator and recovery code verification to login flow


gianfrancopiana:
## What

Backend routing for TOTP login is done (#4002). This issue covers the remaining work:

- [x] Create Flipper flag `authenticator_2fa` (default OFF)
- [x] Gate the TOTP login branch in `prepare_for_two_factor_authentication` behind the flag (currently ungated)
- [x] Update `TwoFactorAuthentication/Show.tsx` to consume the `two_factor_method` prop:
  - `"totp"`: show authenticator code input, "Use email instead" and "Use a recovery code" links
  - `"email"`: existing UI (unchanged)
  - `"recovery"`: recovery code input with links back to authenticator/email
- [x] Add switch actions to `TwoFactorAuthenticationController` (`switch_to_email`, `switch_to_recovery`, `switch_to_authenticator`) with Rack::Attack throttle rules
- [ ] Add throttle rules for switch endpoints in `rack_attack.rb` (exact path matching, `user_id` in query string)

## Why

The login page currently shows email-only copy ("we have sent an Authentication Token") even when the backend routes to TOTP mode. Until this ships, the `authenticator_2fa` flag must stay OFF.

Issue #3994, opened 2026-03-16
Add "Password and Authentication" settings tab with 2FA status section


gianfrancopiana:
## What

- [x] Rename "Password" settings tab → "Password and authentication" in `PAGE_TITLES` and navigation
- [x] Add "Two-factor authentication" section below the password form
- [x] Display authenticator app status (not set up / enabled) — email 2FA is the invisible default, no UI mention needed
- [x] Pass `authenticator_app_enabled` from `SettingsPresenter#password_props` to frontend

No new models or logic — pure UI.

## Why

Foundation for the authenticator setup and management UI. Ships the tab rename immediately so subsequent PRs build on the new layout.

PR #4002, opened 2026-03-16
Add backend support for 2FA via authenticator app

ershad:
Ref: #3993

## What

Add backend support for TOTP-based 2FA via authenticator apps alongside the existing email OTP method. 

- A new `TotpCredential` model holds an independent OTP secret, keeping the two methods fully separate.
- Settings endpoints handle setup (QR code + secret), confirmation, teardown, and recovery code generation.
- Recovery codes are bcrypt-hashed and formatted as `XXXX-XXXX`.

During login, if TOTP is enabled, no email is sent - the user enters a code from their authenticator app or a recovery code instead.

## Why

Email-only 2FA is vulnerable to email compromise and adds friction with delivery delays. Authenticator apps are stronger (codes never leave the device) and faster.

Endpoints:

```
  POST   /settings/totp
  POST   /settings/totp/confirm                { code: "123456" }
  DELETE /settings/totp
  POST   /settings/totp/regenerate_recovery_codes
```

---

### DIFF
--- a/app/models/totp_credential.rb
+++ b/app/models/totp_credential.rb
@@ -37,14 +37,18 @@ class TotpCredential < ApplicationRecord
   end
 
   def redeem_recovery_code(code)
-    return false if recovery_codes.blank?
-
     normalized = code.to_s.upcase.delete("-").strip
-    matching_index = recovery_codes.index { |h| BCrypt::Password.new(h) == normalized }
-    return false unless matching_index
 
-    recovery_codes.delete_at(matching_index)
-    update!(recovery_codes:)
+    with_lock do
+      return false if recovery_codes.blank?
+
+      matching_index = recovery_codes.index { |h| BCrypt::Password.new(h) == normalized }
+      return false unless matching_index
+
+      recovery_codes.delete_at(matching_index)
+      update!(recovery_codes:)
+    end
+
     true
   end
 end


app/models/totp_credential.rb
```diff
@@ -0,0 +1,50 @@
...
+  def verify_code(code)
+    authenticate_otp(code, drift: DRIFT).present?
+  end
+
+  def totp_provisioning_uri
+    provisioning_uri(user.email, issuer: ISSUER_NAME)
+  end
+
+  def generate_recovery_codes
+    codes = Array.new(RECOVERY_CODE_COUNT) { SecureRandom.alphanumeric(RECOVERY_CODE_LENGTH).upcase }
```
ershad, line 30: Takes around ~1.8s. BCrypt is intentionally slow to make it resistant to brute force.

app/models/totp_credential.rb
```diff
@@ -0,0 +1,50 @@
...
+      recovery_codes_generated_at: Time.current
+    )
+    codes
+  end
+
+  def redeem_recovery_code(code)
+    return false if recovery_codes.blank?
+
+    normalized = code.to_s.upcase.delete("-").strip
+    matching_index = recovery_codes.index { |h| BCrypt::Password.new(h) == normalized }
```
ershad: Same here, can take up to ~1.8s in the worst case (when an invalid key is passed)

db/migrate/20260316000000_create_totp_credentials.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateTotpCredentials < ActiveRecord::Migration[7.1]
+  def change
+    create_table :totp_credentials do |t|
+      t.references :user, null: false, index: { unique: true }
+      t.string :otp_secret, null: false
```
ershad, line 7: In a follow-up PR, we can move this to an encrypted field in the model.

ershad:
@gianfrancopiana please review.

app/controllers/concerns/two_factor_authentication_validator.rb
```diff
@@ -26,10 +27,20 @@ def skip_two_factor_authentication?(user)
   def prepare_for_two_factor_authentication(user)
     session[TWO_FACTOR_AUTH_USER_ID_SESSION_NAME] = user.id
 
-    # Do not send token if it's already present in the next URL (navigated from email login login to an unauthenticated session)
-    return if params[:next] && params[:next].include?(verify_two_factor_authentication_path(format: :html))
+    if user.totp_enabled?
```
ershad: No feature flag is needed in this PR since no user will have a confirmed `TotpCredential` record.

gianfrancopiana: Once a user confirms their TOTP credential, this kicks in on their next login — it sets the session to "totp" and skips sending the email. But the login page (`TwoFactorAuthentication/Show.tsx`) still says "we have sent an Authentication Token to your email" and shows a "Resend" button. No email was sent, so the user hits a dead end.

I understand the argument that no one has a confirmed `TotpCredential` today, but the moment someone uses `POST /settings/totp/confirm`, the login flow changes for them with no matching frontend UI or rollout control. The login page (`TwoFactorAuthentication/Show.tsx`) still shows the email-only copy and resend button – the frontend update from #3997 isn't covered by any open issue yet.

#3993 calls for a feature flag (`authenticator_2fa`) gating the login path specifically so we can control this. Could we gate this branch behind the flag? That way setup works immediately, but the login flow stays on email until the frontend update ships and the flag is turned on. I'll reopen #3997 to track the remaining frontend work.

app/controllers/settings/totp_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+      credential.update!(confirmed_at: Time.current)
+      codes = credential.generate_recovery_codes
+
+      render json: { success: true, recovery_codes: format_recovery_codes(codes) }
+    else
+      render json: { success: false, error_message: "Invalid code. Please try again." }, status: :unprocessable_entity
+    end
+  end
+
+  def destroy
```
gianfrancopiana, line 43: This deletes the TOTP credential without asking the user to prove they have access to their authenticator. If someone gets hold of an active session (shared computer, session hijack, etc.), they can silently remove the stronger 2FA factor.

#3998 (now closed because this PR handles everything) calls for requiring a current authenticator code before disabling. Can we add a `params[:code]` that gets verified via `credential.verify_code` before destroying the credential?

spec/controllers/two_factor_authentication_controller_spec.rb
```diff
@@ -245,5 +245,89 @@
...
+      end
+
+      it "rejects an invalid TOTP code" do
+        post :create, params: { token: "invalid", user_id: @user.encrypted_external_id }
+
+        expect(flash[:warning]).to eq "Invalid token, please try again."
+        expect(response).to redirect_to(two_factor_authentication_path)
+      end
+
+      it "does not accept email OTP codes when TOTP is the method" do
```
gianfrancopiana, line 301: This test only checks that the session method stayed "totp". It doesn't assert that the login was actually rejected. If someone accidentally made `valid_two_factor_token?` accept email OTPs in TOTP mode, this test would still pass.

Could we assert on the response instead? Something like:

```ruby
expect(response).to redirect_to(two_factor_authentication_path)
expect(flash[:warning]).to eq("Invalid token, please try again.")
```

That way the test actually breaks if the isolation between TOTP and email auth is lost.

app/controllers/concerns/two_factor_authentication_validator.rb
ershad: > I understand the argument that no one has a confirmed TotpCredential today, but the moment someone uses POST /settings/totp/confirm, the login flow changes for them with no matching frontend UI or rollout control. 

That's low risk, right? Who would want to manually send a POST request without a UI to intentionally block themselves from accessing their account? 😅 

But yes, we can add a feature flag to cover all cases.

app/controllers/settings/totp_controller.rb
ershad: Makes sense. Will update.

spec/controllers/two_factor_authentication_controller_spec.rb
ershad: Sure 👍🏼

app/controllers/concerns/two_factor_authentication_validator.rb
gianfrancopiana: Thanks! For the flag name, I'm using `authenticator_2fa` in #3994 (the settings UI) to gate the 2FA section visibility. If you use the same flag here, one flag controls everything: settings UI, login routing, and the setup/management endpoints. That way we can flip it on for the team to test end-to-end and roll out together.

app/controllers/settings/totp_controller.rb
ershad: Thinking more about (and confirmed with @gianfrancopiana), we will run into more issues if we require TOTP to disable/delete 2FA via authenticator. For example, there could be users who logged in with their last recovery code, and they wouldn't be able to use it here. For now, let's remove the condition and revisit it if we run into issues. The chances of this becoming a security problem is very slim.

app/controllers/concerns/two_factor_authentication_validator.rb
ershad: Sounds good, used the same flag here: https://github.com/antiwork/gumroad/pull/4002/changes#diff-6aa9eb9661bd69f3ee8ab181ff09bc41ac2e7b6751ddc4bfccdd8cadcf276b93R30

ershad:
@gianfrancopiana please review.

app/models/totp_credential.rb
```diff
@@ -0,0 +1,50 @@
...
+    return false unless matching_index
+
+    recovery_codes.delete_at(matching_index)
+    update!(recovery_codes:)
```
gianfrancopiana: This looks racy: two near-simultaneous requests can both load the same `recovery_codes` value, both match the same code, and both return success before either `update!` lands. I was able to reproduce that locally. Can we wrap the lookup + delete + update in `with_lock` (and reload inside the lock) so a recovery code can't be redeemed twice?

ershad: Makes sense, updated here: https://github.com/antiwork/gumroad/pull/4002/changes/02948c92f98dc4753844c81a52bc97398638afd4

PR #4004, opened 2026-03-17
Add authenticator and recovery code verification to login flow

binary-koan:
Issue #3997

* Updates the 2FA page to support TOTP and recovery code login options
* Adds (rate limited) support for switching between methods

**Logging in with TOTP:**

https://github.com/user-attachments/assets/f04bc5dc-1025-40b1-a842-9363e0f3b292

**Switching between methods:**

https://github.com/user-attachments/assets/d0f2b0a8-6429-497e-8090-2f6ed918f05d

### DIFF
--- a/app/javascript/components/ApiDocumentation/Endpoints/User.tsx
+++ b/app/javascript/components/ApiDocumentation/Endpoints/User.tsx
@@ -28,7 +28,8 @@ export const GetUser = () => (
     "twitter_handle": null,
     "user_id": "G_-mnBf9b1j9A7a4ub4nFQ==",
     "email": "johnsmith@gumroad.com", # available with the 'view_sales' scope
-    "url": "https://gumroad.com/sailorjohn" # only if username is set
+    "url": "https://gumroad.com/sailorjohn", # only if username is set
+    "profile_picture_url": "https://assets.gumroad.com/user/abc/avatar"
   }
 }`}
     </CodeSnippet>
--- a/app/javascript/components/ApiDocumentation/responseFieldDefinitions.ts
+++ b/app/javascript/components/ApiDocumentation/responseFieldDefinitions.ts
@@ -406,6 +406,7 @@ export const USER_FIELDS: FieldDefinition[] = [
     condition: "available with the 'view_sales' scope",
   },
   { name: "url", type: "string", description: "User's Gumroad profile URL" },
+  { name: "profile_picture_url", type: "string", description: "URL of the user's profile picture" },
 ];
 
 export const OFFER_CODE_FIELDS: FieldDefinition[] = [
--- a/app/models/concerns/user/as_json.rb
+++ b/app/models/concerns/user/as_json.rb
@@ -44,6 +44,7 @@ module User::AsJson
         id: external_id,
         user_id: ObfuscateIds.encrypt(id),
         url: profile_url,
+        profile_picture_url: avatar_url,
         links: (links.presence && links.alive.map(&:general_permalink)),
       }
     end
--- a/spec/controllers/api/v2/users_controller_spec.rb
+++ b/spec/controllers/api/v2/users_controller_spec.rb
@@ -64,6 +64,7 @@ describe Api::V2::UsersController do
       expect(response.parsed_body["user"]["url"]).to be_present
       expect(response.parsed_body["user"]["email"]).to be_present
       expect(response.parsed_body["user"]["profile_url"]).to be_present
+      expect(response.parsed_body["user"]["profile_picture_url"]).to be_present
       expect(response.parsed_body["user"]["display_name"]).to be_present
     end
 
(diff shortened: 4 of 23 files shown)


binary-koan:
@ershad @gianfrancopiana FYI, this is more or less ready once the base PR is merged

Not sure about the design of the "Use ... instead" links, lmk if you have any ideas

Screenshot 2026-03-17 at 00 11 20

app test gumroad com_31337_two-factor_next=%2Fdashboard(iPhone 12 Pro)

gianfrancopiana:
@binary-koan A few small additions to the input that would make the TOTP experience nicer:

- Add `autocomplete="one-time-code"`. This lets phones auto-fill the code from notifications or authenticator apps.
- Add `maxLength={6}` and `pattern="[0-9]*"` for the TOTP input so users can't accidentally type more than 6 digits
- Consider auto-submitting once 6 digits are entered to save a click.
- Optional but worth trying: add some `letter-spacing` to the input to visually space the digits. I think this gives it that polished one-digit-per-box feel without the paste/accessibility headaches those split inputs usually have.

binary-koan:
Will update, thanks! `autocomplete="one-time-code"` is neat, TIL

binary-koan:
(rebased onto main and force pushed)

app/controllers/two_factor_authentication_controller.rb
```diff
@@ -49,16 +68,19 @@ def verify_auth_token_and_redirect(token)
...
+        redirect_to two_factor_authentication_path, warning: "Invalid token, please try again.", status: :see_other
       end
     end
 
     def valid_two_factor_token?(user, token)
-      if two_factor_auth_method == "totp"
+      case two_factor_auth_method
+      when "totp"
         return true if user.totp_credential&.verify_code(token)
+        return true if !Rails.env.production? && token == User::DEFAULT_AUTH_TOKEN
```
binary-koan, line 79: Not sure if we want this or not tbh, but might be a nice shortcut for testing

app/javascript/pages/TwoFactorAuthentication/Show.tsx
```diff
@@ -9,39 +9,48 @@ import { Input } from "$app/components/ui/Input";
...
 
-  const resendForm = useForm({ authenticity_token });
+  const isNumericCode = two_factor_method === "totp" || two_factor_method === "email";
+
+  const [token, setToken] = React.useState(initialToken ?? "");
+  const [isSubmitting, setIsSubmitting] = React.useState(false);
+
+  const submitCode = (token: string) => {
+    if (isSubmitting) return;
+    router.post(
```
binary-koan, line 36: This is already an interesting case where it's hard to avoid `useEffect`:

Inertia's `useForm` has a `.submit()` method, but it uses the current React state of the form. That means it doesn't work immediately after a `form.setData()` call, because that calls `setState` which doesn't update the `state` variable until the next render. Claude wanted to do something like this:

```tsx
useEffect(() => {
  if (form.data.token.length === 6) form.submit();
}, [form.data.token])

// ...

<input onChange={(e) => form.setData("token", e.currentTarget.value)} />
```

But moving the check to the event handler seems like the correct way to avoid the `useEffect` here.

Worked around the issue by just getting rid of the form and using `router.post` directly

app/javascript/pages/TwoFactorAuthentication/Show.tsx
```diff
@@ -9,39 +9,48 @@ import { Input } from "$app/components/ui/Input";
...
     e.preventDefault();
-    form.post(Routes.two_factor_authentication_path({ user_id }));
+    submitCode(token);
   };
 
-  const resendToken = () => {
-    resendForm.post(Routes.resend_authentication_token_path({ user_id }));
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setToken(e.target.value);
+    if (isNumericCode && /^\d{6}$/u.test(e.target.value)) submitCode(e.target.value);
```
binary-koan, line 53: We might as well auto-submit email codes as well for consistency

config/initializers/rack_attack.rb
```diff
@@ -185,6 +185,27 @@ def self.throttle_by_ip_for_period(path:, requests:, period:, method: nil)
                      period: 60.seconds,
                      throttle_params: Proc.new { |req| req.params["user_id"] }
 
+  # Initial: 10rpm, Max: 60 requests/3 days (per user)
+  throttle_by_params path: "/two-factor/switch_to_email",
```
binary-koan, line 189: Same rate limit options as the other 2FA paths

spec/requests/settings/password_spec.rb
```diff
@@ -156,6 +156,7 @@
 
         click_on("Set up")
         expect(page).to have_text("Scan this QR code")
+        expect(page).to have_selector("[data-testid='qr-code']")
```
binary-koan, line 159: This test flaked in CI and failed locally with a race condition - the "Scan this QR code" text showing during the loading state so the controller action may not have completed yet

PR #4005, opened 2026-03-17
Migrate [hidden] attribute style to Tailwind classes

binary-koan:
Issue #3008 

Removes the default style on `[hidden]` elements and replaces the property with the Tailwind `hidden` class. Note that this shouldn't affect accessibility as screen readers ignore `display: none` elements just like `hidden` ones.

This one we could arguably keep in CSS, but I think it's clearer if we're consistent and always use inline classes.

No visual changes, some examples:

* Discover - search suggestions are hidden
01-discover

* Widgets - the inactive tab is hidden
02-widgets

* Product page - microdata divs are hidden
03-product-page

* Profile page - rich text toolbar hidden when not focused
Screenshot 2026-03-17 at 01 03 15


binary-koan:
@SalmanNajah Could you take a look at this as well?

PR #4006, opened 2026-03-17
Migrate [inert] attribute style to Tailwind classes

binary-koan:
Issue #3008 

Migrates the styles on `[inert]` to inline Tailwind classes. In this case I've added the classes alongside the `inert` attribute since we still want browser/accessibility behaviour such as disabling the element as well as the styling.

No visual changes, examples:

* Disabled buttons
Screenshot 2026-03-17 at 01 32 21

* Disabled menu items
Screenshot 2026-03-17 at 01 18 20
Screenshot 2026-03-17 at 01 25 30

* Disabled sidebar on the content editor
Screenshot 2026-03-17 at 01 20 28

* Workflows page without permissions
Screenshot 2026-03-17 at 01 26 17

### DIFF
--- a/app/javascript/components/ProductsPage/ActionsPopover.tsx
+++ b/app/javascript/components/ProductsPage/ActionsPopover.tsx
@@ -3,7 +3,6 @@ import * as React from "react";
 
 import { archiveProduct, deleteProduct, duplicateProduct, unarchiveProduct } from "$app/data/product_dashboard";
 import { Membership, Product } from "$app/data/products";
-import { classNames } from "$app/utils/classNames";
 import { assertResponseError } from "$app/utils/request";
 
 import { Button } from "$app/components/Button";


spec/requests/affiliates_spec.rb
```diff
@@ -444,7 +444,6 @@
 
         link = find_link("Add affiliate", inert: true)
         link.hover
-        expect(link[:style]).to eq "pointer-events: none; cursor: not-allowed; opacity: 0.3;"
```
binary-koan, line 447: We already find the link by `inert: true`, so I don't think we need to assert the exact style

app/javascript/components/Button.tsx
```diff
@@ -122,7 +122,11 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
   ({ className, color, outline, size, disabled, children, asChild = false, ...props }, ref) => {
     const variant = outline ? "outline" : color === "danger" ? "destructive" : "default";
 
-    const classes = classNames(buttonVariants({ variant, size, color: color || undefined }), className);
+    const classes = classNames(
+      buttonVariants({ variant, size, color: color || undefined }),
+      disabled && "cursor-not-allowed opacity-30 hover:translate-0 hover:shadow-none",
```
binary-koan, line 127: Adding these explicitly when `disabled` is set (instead of conditionally with CSS) so that `NavigationButton` in this file works correctly

https://github.com/antiwork/gumroad/blob/436ffc69d16b80a2c985f9b6650bbb9f846afaf3/app/javascript/components/Button.tsx#L146-L151

app/javascript/stylesheets/_global.scss
```diff
@@ -142,11 +142,6 @@ menu > li {
   background-blend-mode: overlay;
 }
 
-[inert] {
-  cursor: not-allowed;
```
nyomanjyotisa, line 79: @binary-koan Seems like `cursor: not-allowed;` is a dead code, since if we add `cursor: not-allowed;` on element with `inert`, the `cursor: not-allowed;` not displayed at all

`inert` and `cursor: not-allowed;`:

https://github.com/user-attachments/assets/4f750756-ab64-48ae-86a6-661aefe4ce45

`cursor: not-allowed;` without `inert`:

https://github.com/user-attachments/assets/e768ff8e-067e-49c5-989a-62a18e38d991

app/javascript/components/Button.tsx
binary-koan, line 127: Seems like although `inert` prevents the cursor changing, `disabled` doesn't, so this style is still relevant for `<button>` elements (although we should perhaps consider removing it for consistency)

PR #4008, opened 2026-03-17
Migrate scrollbar styles to Tailwind

binary-koan:
Issue #3008 

These are a bit awkward to inline because `body` elements are in various ERB templates, so I think it's fine to have them in `tailwind.css` for now.


binary-koan:
Don't really have the ability to test this until it's on production, will check when it's deployed. Should be low impact regardless

PR #4009, opened 2026-03-17
Fix bundle refund not removing reviews on individual product purchases

### LINKED ISSUE #3979: Bundle refund does not remove reviews on individual product purchases
## What

 When a seller refunds a bundle purchase, the reviews left on individual products within the bundle are not removed. A buyer can leave a 1-star review on each product in a
 bundle, then when the bundle is refunded those reviews persist — dragging down the seller's product ratings even though the purchase is no longer valid.

 ## Why

 When a bundle is refunded, `refund_purchase!` calls `mark_product_purchases_as_refunded!`, which sets `stripe_refunded: true` on each product purchase. This triggers an
 `after_save` callback (`update_product_review_stat`) that should soft-delete the associated review and update the product's `ProductReviewStat`.

 There are two likely root causes to investigate:

 ### 1. The `after_save` callback may silently fail for bundle product purchases

 The callback chain is:

 ```
 product_purchase.update!(stripe_refunded: true)
   → after_save :update_product_review_stat
     → link.update_review_stat_via_purchase_changes(saved_changes, product_review:)
       → product_review.mark_deleted!
       → product_review_stat.update_with_removed_rating(rating)
 ```

 The callback has an early return guard: `return if saved_changes.blank? || original_product_review.blank? || true_original_purchase != self`. If `original_product_review`
 returns `nil` for a bundle product purchase (e.g., because the review was posted through a different association path), the callback silently exits and the review survives.

 ### 2. Buyer's library still displays soft-deleted reviews

 Even if the review IS soft-deleted correctly, the `ReviewsPresenter` fetches reviews without an `.alive` filter:

 ```ruby
 reviews: user.product_reviews.includes(:editable_video).map do |review|
 ```

 This returns all reviews including soft-deleted ones, so buyers still see their review in their library. The product page itself correctly filters with `.alive`, but the buyer's
  own view does not.

 ### Additional edge case: partial refunds

 When a bundle is partially refunded, `mark_product_purchases_as_refunded!` sets `stripe_partially_refunded: true` (not `stripe_refunded: true`) on the product purchases. The
 `allows_review_to_be_counted?` method only checks `stripe_refunded?`, not `stripe_partially_refunded?`. So a partial refund of a bundle never removes reviews from the individual
  products, even though the buyer received money back.

 ## Proposed solution

 1. **Verify the `after_save` callback fires correctly for bundle product purchases** — add logging or step through the callback chain to confirm `original_product_review`
 returns the review and `update_review_stat_via_purchase_changes` actually executes. If it doesn't, trace why and fix the guard condition.

 2. **Filter soft-deleted reviews in `ReviewsPresenter`** — add `.alive` to the query at `ReviewsPresenter#reviews_props` so buyers don't see reviews for refunded purchases in
 their library. Also handle the inverse: refunded purchases with soft-deleted reviews should appear in the "purchases available for review" list if appropriate, or be excluded
 entirely.

 3. **Decide on partial refund behavior** — if a bundle is partially refunded, should the product purchase reviews be removed? If yes, update `allows_review_to_be_counted?` to
 also check `stripe_partially_refunded?`, or have `mark_product_purchases_as_refunded!` set `stripe_refunded: true` on individual product purchases even for partial bundle
 refunds (since the individual product access may be fully revoked).

 ## Acceptance criteria

 - [ ] When a bundle purchase is fully refunded, all reviews on individual product purchases within the bundle are soft-deleted
 - [ ] Product rating stats (`ProductReviewStat`) are updated to reflect the removed reviews
 - [ ] Soft-deleted reviews do not appear on the product page (already works via `.alive` scope)
 - [ ] Soft-deleted reviews do not appear in the buyer's library reviews
 - [ ] Refunding a single (non-bundle) purchase continues to remove reviews as before
 - [ ] Partial bundle refund behavior is explicitly decided and handled consistently

 ## Edge cases

 - **Bundle with subscription products**: if a bundle contains subscription products, `true_original_purchase` may resolve differently, potentially breaking the callback guard
 - **Bundle where only some products have reviews**: refund should only remove reviews that exist, not error on products without reviews
 - **Re-reviewing after refund**: if a purchase is refunded and the review is soft-deleted, the buyer should not be able to re-post a review (the purchase is no longer valid via
 `allowing_reviews_to_be_counted`)
 - **Existing data**: production may have orphaned reviews from past bundle refunds that were never cleaned up — a backfill may be needed

nyomanjyotisa:
Fixes #3979

## What

When a bundle purchase is partially refunded, `mark_product_purchases_as_refunded!` correctly sets `stripe_partially_refunded: true` on each product purchase. However, `allows_review_to_be_counted?` and the `allowing_reviews_to_be_counted` scope only checked `stripe_refunded`, not `stripe_partially_refunded`. The `after_save` callback compared old vs new validity and saw no change, so reviews were never soft-deleted.

This PR:
- Adds `!stripe_partially_refunded?` to `allows_review_to_be_counted?` and `not_partially_refunded` to the `allowing_reviews_to_be_counted` scope so partial refunds invalidate reviews
- Adds a backfill script (`Onetime::BackfillBundleRefundReviews`) to fix orphaned reviews from bundle purchases refunded before PR #2460 (merged 2026-01-06), which added `mark_product_purchases_as_refunded!`. Before that PR, refunding a bundle never cascaded the refund flags to product purchases at all.

## Why

A support ticket ([conversation](https://help.gumroad.com/conversations?id=663a288b23bebe2106b02a32bdd20f60)) reported 1-star reviews persisting on products after a bundle was fully refunded. Investigation showed the bundle (purchase 255576699) was refunded on 2025-01-07, before PR #2460 existed (merged 2026-01-06). The refund flags were never cascaded to product purchases, so reviews survived. Those reviews were manually removed, but there may be other affected purchases from before that date.

Separately, partial bundle refunds have never removed reviews because `allows_review_to_be_counted?` did not check `stripe_partially_refunded`. This has been broken since bundles launched.

---

This PR was implemented with AI assistance using Claude Opus 4.6


app/modules/purchase/reviews.rb
```diff
@@ -50,6 +51,7 @@ def allows_reviews(permit_recurring_charges:)
       allowed &= !not_charged_and_not_free_trial?
       allowed &= not_is_gift_sender_purchase
       allowed &= !stripe_refunded?
+      allowed &= !stripe_partially_refunded?
```
nyomanjyotisa: `mark_product_purchases_as_refunded!` (added in #2460) correctly sets `stripe_partially_refunded: true` on product purchases during partial bundle refunds, but this method never checked it. The `after_save` callback compared old vs new `allows_review_to_be_counted?`, saw no change, and skipped review deletion.

app/modules/purchase/reviews.rb
```diff
@@ -14,6 +14,7 @@ module Purchase::Reviews
       where(purchase_state: COUNTS_REVIEWS_STATES).
         exclude_not_charged_except_free_trial.
         not_fully_refunded.
+        not_partially_refunded.
```
nyomanjyotisa: Keeps the scope in sync with the instance method above

app/services/onetime/backfill_bundle_refund_reviews.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    Rails.logger.info("#{@dry_run ? '[DRY RUN] ' : ''}Done. Fixed: #{@fixed_count}, Skipped: #{@skipped_count}")
+  end
+
+  private
+    def refunded_bundle_purchases
+      Purchase
+        .where(Purchase.is_bundle_purchase_condition)
+        .where("purchases.stripe_refunded = true OR purchases.stripe_partially_refunded = true")
+        .where("purchases.created_at < ?", PR_2460_MERGE_DATE)
```
nyomanjyotisa: Only targets bundles refunded before #2460 merged (2026-01-06). After that date, `mark_product_purchases_as_refunded!` cascades the flags correctly

app/services/onetime/backfill_bundle_refund_reviews.rb
```diff
@@ -0,0 +1,73 @@
...
+          Rails.logger.info(
+            "[DRY RUN] Would fix purchase #{product_purchase.id} (#{refund_type} refund) " \
+            "(bundle: #{bundle_purchase.id}, product: #{product_purchase.link.name}, " \
+            "review: #{review&.id || 'none'}, review_alive: #{review&.alive?})"
+          )
+        else
+          if is_partially_refunded
+            product_purchase.update!(stripe_partially_refunded: true)
+          else
+            product_purchase.update!(stripe_refunded: true)
```
nyomanjyotisa, line 56: Uses `update!` so the `after_save :update_product_review_stat` callback fires and handles review soft-deletion and stat updates automatically

app/modules/purchase/reviews.rb
gianfrancopiana: The core fix for bundle refunds looks good, but there's a problem with the scope of the change. Adding `!stripe_partially_refunded?` to `allows_reviews` and `not_partially_refunded` to `allowing_reviews_to_be_counted` affects **all** purchases, not just bundle product purchases. This means a regular partial refund on a standalone product would now remove the buyer's review too.

That contradicts the current documented behavior in the help center ("Issuing a partial refund will not remove a customer's rating") and would be a regression for non-bundle purchases.

The fix should only apply to product purchases inside bundles. Something like `allowed &= !stripe_partially_refunded? || !is_bundle_product_purchase?` (and the equivalent scope change).

nyomanjyotisa:
@gianfrancopiana Updated to scope the fix to bundle product purchases only. Could you take another look?

nyomanjyotisa:
@gianfrancopiana, Could you re-review when you get a chance?

app/services/onetime/backfill_bundle_refund_reviews.rb
```diff
@@ -0,0 +1,87 @@
...
+          Rails.logger.info(
+            "[DRY RUN] Would fix purchase #{product_purchase.id} (#{refund_type} refund) " \
+            "(bundle: #{bundle_purchase.id}, product: #{product_purchase.link.name}, " \
+            "review: #{review&.id || 'none'}, review_alive: #{review&.alive?})"
+          )
+          @affected_count += 1
+        else
+          if is_partially_refunded && product_purchase.stripe_partially_refunded?
+            product_purchase.link.update_review_stat_via_rating_change(review.rating, nil)
+            review.mark_deleted!
```
nyomanjyotisa, line 52: This handles product purchases where the flag was already propagated but the review was never removed. Since re-saving the same flag wouldn't trigger the `after_save` callback, we manually update review stats and soft-delete the review

Issue #3790, opened 2026-02-23
Add Stripe Link


slavingia:
## Summary

Add [Stripe Link](https://stripe.com/payments/link) as a payment method on Gumroad checkout. Link is Stripe's digital wallet that lets customers save and reuse payment details for faster checkout, supporting cards, US bank accounts, and buy-now-pay-later options.

## Why

- **Higher conversion**: Link autofills saved payment and shipping info, reducing checkout friction
- **Lower processing costs**: When customers pay via [Instant Bank Payments](https://docs.stripe.com/payments/link/instant-bank-payments) (exclusive to Link), processing fees are lower than card payments
- **No additional settlement delay**: Link payments settle on the same timeline as card payments

## Current State

Gumroad checkout currently uses the **Stripe CardElement** (`@stripe/react-stripe-js`) with `payment_method_types: ["card"]` hardcoded in both `SetupIntent` and `PaymentIntent` creation (`stripe_charge_processor.rb`, lines ~172 and ~220). The frontend uses `stripe.confirmCardPayment()` / `stripe.confirmCardSetup()` / `stripe.createPaymentMethod({ type: "card" })`.

### Option A: Migrate to Payment Element (recommended)

Replace `CardElement` with Stripe's **Payment Element**, which supports Link (and 100+ other payment methods) through a single integration.

**Backend changes:**
1. In `stripe_charge_processor.rb`, replace `payment_method_types: ["card"]` with `automatic_payment_methods: { enabled: true }` on both `SetupIntent.create` and `PaymentIntent.create` calls
2. Enable Link in the [Stripe Dashboard payment method settings](https://dashboard.stripe.com/settings/payment_methods)
3. Update webhook handlers if needed (Link payments use the same `payment_intent.succeeded` events)

**Frontend changes:**
1. Replace `CardElement` with `PaymentElement` in `CreditCardInput.tsx`
2. Replace `stripe.confirmCardPayment()` / `stripe.confirmCardSetup()` with `stripe.confirmPayment()` / `stripe.confirmSetup()` in `purchase.ts` and `order.ts`
3. Replace `stripe.createPaymentMethod({ type: "card" })` with the Payment Element's built-in method handling in `card_payment_method_data.ts`
4. Handle the saved card UX (Payment Element has its own saved payment method support)

**Key consideration:** This is a larger migration since `CardElement` is used across checkout, the payout settings page (`CreditCard.tsx`), and payment method management. Needs careful scoping to avoid breaking existing flows.

[...]

PR #4012, opened 2026-03-17
Enable Stripe Link

hchhabra:
Issue #3790 

### What

Enable Stripe's Link payment method

### Why

Link autofills saved card details reducing checkout friction and increasing conversion

### Preview URL

https://seller.enable-stripe-link.apps.staging.gumroad.org/l/p

### Demo / Screenshots

- Logging into Link and using saved card:

  https://github.com/user-attachments/assets/cd58fc19-4fee-4600-81fe-87e28ed8d24e

- Already logged into Link, using saved card:

  https://github.com/user-attachments/assets/a25a23ac-d9e0-428b-937d-617d7d11b2b5

- Entering new card details and saving it to Link:

  https://github.com/user-attachments/assets/c0dbccb8-9501-43aa-a8c4-25fb86a280a6

---

AI disclosure: Claude Opus 4.6 used to add specs and review code change.


app/javascript/components/Checkout/CreditCardInput.tsx
```diff
@@ -78,7 +78,7 @@ export const CreditCardInput = ({
                 style: { base: baseStripeStyle ?? {} },
                 hidePostalCode: true,
                 disabled: disabled ?? false,
-                disableLink: true,
+                disableLink: false,
```
gianfrancopiana: Enabling Link inside the existing `InputGroup` makes the saved-Link chip look awkwardly inset here. The wrapper's padding/min-height works for the regular card field, but with Link enabled it leaves a lot of dead space on the left and vertically:
Image

Could we tighten the wrapper for the Link-enabled state or render Link in its own row so this doesn't look so padded?

To fix, we can pass a `className` to `InputGroup` when Link is enabled and reduce the extra inset, e.g. something like `px-0` or a smaller horizontal padding/min-height for this state. If Stripe’s iframe still looks off, the cleaner fix is to render `LinkAuthenticationElement` above `CardElement` instead of relying on the inline Link chip.

gianfrancopiana (commented):
This enables Stripe Link for every checkout immediately. Since this is a new payments feature, would it make sense to gate it behind a feature flag first? That gives us a safer rollout and an easy kill switch if we hit any issues.

app/javascript/components/Checkout/CreditCardInput.tsx
hchhabra: Looked into reducing the extra space on the left when Link is used, but it's part of Stripe's iframe, and there doesn't seem to be a reliable way of detecting whether the card field is filled manually or via Link.

| Card | Link |
| ----- | ---- |
| Screenshot 2026-03-25 at 12 14 23 AM | Screenshot 2026-03-25 at 12 14 48 AM |

Also, `LinkAuthenticationElement` only interacts with `PaymentElement` it seems (ref [docs](https://docs.stripe.com/payments/elements/link-authentication-element#combine-elements)), so displaying a separate `LinkAuthenticationElement` above the `CardElement` doesn't work to autofill the card field as expected. @gianfrancopiana _a

Screenshot 2026-03-25 at 12 00 11 AM

gianfrancopiana: Since this is our first Link implementation, I think we should use [`PaymentElement`](https://docs.stripe.com/payments/payment-element) instead of `CardElement` with `disableLink: false`.

Link is [designed to work natively with `PaymentElement`](https://docs.stripe.com/payments/link/payment-element-link). Layout, autofill, and authentication all work out of the box without the spacing/iframe issues. We can restrict it to card-only via `payment_method_types: ["card"]` on the PaymentIntent, so the checkout UI stays the same – just with Link working properly. Styling is fully customizable via the [Appearance API](https://docs.stripe.com/elements/appearance-api).

The [migration guide](https://docs.stripe.com/payments/payment-element/migration) covers the server-side changes (creating a PaymentIntent before rendering). It's more work upfront, but I think we would want this eventually.

gianfrancopiana (changes requested):

slavingia:
> Stripe no longer recommends using the Card Element as part of your Web Elements integration. To integrate Link, use one of our preferred Elements: the [Link Authentication Element](https://docs.stripe.com/payments/link/link-authentication-element), [Express Checkout Element](https://docs.stripe.com/payments/link/express-checkout-element-link), or [Payment Element](https://docs.stripe.com/payments/link/payment-element-link).

So I think we need to update this to using Payment Element. If that's a lot of work, we can defer this to after we ship [the Stripe onboarding stuff](https://github.com/antiwork/gumroad/issues/4021).

PR #4023, opened 2026-03-18
Migrate display:block on `<small>` to Tailwind

binary-koan:
Ref #3008

We noticed when reviewing the remainder of `_global.scss` that `small` has a `display: block` style which might confuse people and AI expecting it to have its default layout.

This migrates the style to an inline class for all elements where it's not explicitly set already.

Screenshot 2026-03-20 at 13 30 32
Screenshot 2026-03-20 at 13 30 55
Screenshot 2026-03-20 at 13 40 28
Screenshot 2026-03-20 at 13 42 41


app/javascript/components/Admin/Products/Purchases/Purchase.tsx
```diff
@@ -62,7 +62,7 @@ const AdminProductPurchase = ({
             <a href={Routes.admin_purchase_path(external_id)}>{displayed_price}</a>
             {gumroad_responsible_for_tax ? ` + ${formatted_gumroad_tax_amount} VAT` : null}
           </h5>
-          <small>
+          <small className="block">
```
binary-koan, line 65: @nyomanjyotisa Can I get your opinion on this? For most of these the class isn't needed as the elements around it are `block`, but I don't mind having it explicit like this as it makes the layout clearer when reading the code.

wdyt - should we go with this or delete the class where it's not needed?

nyomanjyotisa: Agree, keeping block explicit everywhere makes sense, easier to read without checking the parent context.

nyomanjyotisa (approved):
Looks good to me!

PR #4024, opened 2026-03-18
Add TikTok Pixel integration to third-party analytics

gumclaw:
## Summary
- Adds TikTok Pixel as a first-class third-party analytics integration alongside Google Analytics and Facebook Pixel
- Creators can enter their TikTok Pixel ID in Settings > Third-party analytics, and the pixel code is injected directly into the page `<head>` (not inside an iframe sandbox)
- Fires TikTok events: `ViewContent` (product viewed), `AddToCart` (I want this), `CompletePayment` (purchase)

Closes https://github.com/antiwork/gumroad/issues/4037

## Changes
- **Migration**: adds `tiktok_pixel_id` string column to users table
- **User model**: adds validation (alphanumeric uppercase), stripped_fields, LINK_PROPERTIES
- **Settings UI**: new TikTok Pixel input field in the Third-party analytics section
- **Backend**: controller permits param, presenter returns it, Link#analytics_data includes it, meta tag `gr:tiktok_pixel:enabled` added
- **Frontend**: new `tiktok_pixel.ts` data module + `tiktok_pixel.js` vendor script loader, wired into `user_analytics.ts` orchestrator
- **Tests**: controller spec, presenter spec, and system spec updated

## Test plan
- [ ] Visit Settings > Third-party analytics, verify TikTok Pixel field appears between Facebook Pixel and the free sales checkbox
- [ ] Enter a valid TikTok Pixel ID (e.g. `CFH83AJC77UUUGLE2TJG`), save, confirm it persists
- [ ] Enter an invalid ID (e.g. `invalid-pixel!`), save, confirm validation error appears
- [ ] With a TikTok Pixel ID set, visit a product page and verify `ttq` loads and fires `ViewContent`
- [ ] Complete a purchase and verify `CompletePayment` event fires with correct value/currency
- [ ] Verify TikTok Event Builder can now detect and interact with the pixel (no longer sandboxed in iframe)


gianfrancopiana:
@gumclaw The PR adds a new runtime dependency on the `gr:tiktok_pixel:enabled` meta tag rendered from `app/controllers/concerns/page_meta/analytics.rb`. The frontend tracking module checks this tag before initializing the pixel, so if the tag stops being emitted, TikTok tracking silently breaks. The existing controller specs in `checkout_controller_spec.rb` (line 45) and `links_controller_spec.rb` (line 3961) already assert the equivalent Google Analytics and Facebook Pixel meta tags, but the new TikTok tag isn't added to those assertions. We should extend them so this contract is covered.

gumclaw:
Good catch — the TikTok meta tag should be covered by the same controller specs that assert the GA and Facebook Pixel tags. I'll extend the assertions in `checkout_controller_spec.rb` and `links_controller_spec.rb` to include `gr:tiktok_pixel:enabled`. Pushing a fix shortly.

PR #4028, opened 2026-03-18
Auto-retry if Start services fails in CI

gumclaw:
## What

Auto-retry the "Start services" step (up to 3 attempts) in both `test_fast` and `test_slow` CI jobs. If `docker compose up -d` fails, it tears down volumes and retries after a 5-second delay.

## Why

Service startup failures are a top cause of CI flakiness (~40% of recent failures). Docker services (MySQL, Redis, Elasticsearch, Minio) occasionally fail to start on the first attempt due to transient issues on CI runners. A simple retry avoids wasting the entire test run.

Related: #190


ershad:
Merged via #3938

PR #4040, opened 2026-03-19
Fix flaky preorder tax test with explicit AJAX wait

gumclaw:
## What

Fixes the flaky test `shipping_physical_preorder_spec.rb:74` ("charges the proper amount with taxes for preorder") by adding `send_keys(:tab)` + `wait_for_ajax` before asserting tax amounts.

## Why

The test was flaky because `fill_checkout_form` fills the ZIP code field but doesn't reliably trigger React's blur event, which is what initiates the TaxJar API call. Without an explicit blur trigger and AJAX wait, the "Sales tax US$1.07" assertion sometimes runs before the tax calculation completes.

This follows the established pattern from `taxes_spec.rb` where the same fix was applied to 8+ US sales tax tests.

## Test Results

The `shipping: true` tag already wraps the test in a VCR cassette via `vcr_turned_on`, so `force_vcr_on: true` is not needed.

---


gumclaw:
**CI evidence (first run 23295894495):** The targeted spec (`shipping_physical_preorder_spec.rb:74`) still failed — retried 3 times, tax line never appeared. The JS nativeInputValueSetter approach for triggering React onChange on the ZIP field is unreliable. Closing to rework the fix.

PR #4041, opened 2026-03-19
Fix flaky US sales tax tests with explicit AJAX waits

gumclaw:
## What

Adds `send_keys(:tab)` + `wait_for_ajax` blocks to 8 US sales tax tests in `taxes_spec.rb` that were flaky due to tax calculation timing.

## Why

Capybara's `fill_in` for the ZIP code field doesn't reliably trigger React's blur event, which initiates the TaxJar API call. By explicitly sending a tab key (triggering blur) and waiting for AJAX to complete, the tax amount assertions become deterministic instead of racing the async calculation.

---


gumclaw:
**CI evidence (first run 23295894495):** The targeted spec (`taxes_spec.rb`) did NOT appear in the failure list. Failures on this run were from unrelated flaky tests: `shipping_offer_codes_spec`, `shipping_physical_preorder_spec`, `payments_spec`. The tax calculation timing fixes are validated.

PR #4044, opened 2026-03-19
Use exact matching for pagination button assertions

gumclaw:
## What

Adds `exact: true` to `have_button("3")` and `have_button("3")` assertions in:
- `utm_links_spec.rb`
- `customers_spec.rb`

## Why

Without `exact: true`, `have_button("3")` can match buttons containing "3" as a substring (e.g., "30 days", "3 products"), causing false positives in negative assertions (`to_not have_button`) and false negatives in positive assertions.

---


gumclaw:
**CI evidence (first run 23295896898):** The targeted specs (`utm_links_spec.rb`, `customers_spec.rb`) did NOT appear in the failure list. Failures on this run were from unrelated flaky tests: `shipping_physical_preorder_spec`, `payments_spec`. The exact pagination button matching fixes are validated.

PR #4046, opened 2026-03-19
Return profile picture URL in the API

binary-koan:
See https://github.com/antiwork/gumroad-mobile/pull/108

It's helpful if we show who is currently logged in to the mobile app, but currently the API doesn't include the profile picture so we can't easily make the layout look nice.

This adds the field to the current user endpoint to enable the design in the mobile PR screenshots.

Also added to API docs:

Screenshot 2026-03-19 at 14 53 54


binary-koan:
@nyomanjyotisa Could you review?

PR #4052, opened 2026-03-19
Add minimum mobile app version endpoint

binary-koan:
See https://github.com/antiwork/gumroad-mobile/issues/97

## Why

In some cases we may want to drop support for old versions of the app (e.g. if the API undergoes breaking changes). Apple and Google don't have a built-in way to do this, so the best approach is for the app to ping an endpoint to check if it's still a recent enough version.

## How

This adds a `/internal/mobile_minimum_version` endpoint to the API, which returns the values of two Redis keys. One is the minimum app store version, the other is the minimum OTA update timestamp. The app can combine these to check if the user needs an app store update or OTA update.

The keys can just be set in console if we need them; I don't think it's necessary to add a UI at this stage.

Screenshot 2026-03-19 at 22 34 17


binary-koan:
@gianfrancopiana Could you take a look at this?

binary-koan:
(going ahead and merging so I can submit an app build)

PR #4053, opened 2026-03-20
Support searching purchases by ID on mobile

binary-koan:
See https://github.com/antiwork/gumroad-mobile/pull/110

## Why

We want to display recently viewed products prominently in the library on the mobile app for ease of access. Currently the apps fetch the entire library and find recent products from there, but in the new app we're adding pagination to avoid the performance problems we're seeing. That means we need a way for the new app to fetch just the recently viewed products.

## How

Adds a `purchase_ids` param to the mobile purchase search API, which filters the Elasticsearch query by `id`. The app stores recently viewed purchase IDs in local storage and uses this parameter to fetch them.

Note the search service currently supports filtering by product, but if you purchase multiple of the same product it will cause duplicates in the carousel, so I think adding an ID search is more reliable.


binary-koan:
@gianfrancopiana Could you review?

binary-koan:
Ah good point, updated!

PR #4055, opened 2026-03-20
Remove Facebook and X (Twitter) sign-up and login

slavingia:
## Summary
- Removes Facebook and X (Twitter) as OAuth login/signup providers, keeping only Google, Stripe, and email/password
- Removes all related modules (`SocialFacebook`, `SocialTwitter`), controllers (`ConnectionsController`, `Users::OauthController`), workers (`RenewFacebookAccessTokensWorker`), initializers, routes, and UI components
- Removes `omniauth-facebook`, `omniauth-twitter`, and `twitter` gems
- Keeps Facebook SDK/Pixel/sharing and Twitter sharing functionality intact (those are not auth-related)

## Test plan
- [ ] Verify login page shows only Google, Stripe, and email/password options
- [ ] Verify signup page shows only Google, Stripe, and email/password options
- [ ] Verify Google OAuth login/signup still works
- [ ] Verify Stripe OAuth login/signup still works
- [ ] Verify email/password login/signup still works
- [ ] Verify Settings > Profile no longer shows "Connect to X" button
- [ ] Verify existing users with Facebook/Twitter-linked accounts can still log in via email/password


.claude/settings.json
```diff
@@ -0,0 +1,22 @@
+{
```
gianfrancopiana, line 1: This seems unrelated?

app/models/user.rb
```diff
@@ -7,7 +7,7 @@ class User < ApplicationRecord
   has_paper_trail
   has_one_time_password
   include Flipper::Identifier, FlagShihTzu, CurrencyHelper, Mongoable, JsonData, Deletable, MoneyBalance,
-          DeviseInternal, PayoutSchedule, SocialFacebook, SocialTwitter, SocialGoogle, SocialApple, SocialGoogleMobile,
+          DeviseInternal, PayoutSchedule, SocialGoogle, SocialApple, SocialGoogleMobile,
```
gianfrancopiana, line 10: Removing `SocialFacebook` here deletes the `User.fb_object` class method, but `config/initializers/doorkeeper.rb:31` still calls it.

Any `/oauth/token` request with a `facebookToken` param will raise `NoMethodError` at runtime. That same block also still accepts `twitterToken`, so mobile X login still works even though the PR says it's removed.  We should clean up the `facebookToken` and `twitterToken` cases in `doorkeeper.rb`.

.claude/settings.json
slavingia: @gumclaw move this to a separate PR on its own and remove from this one.

gumclaw: Done — moved to #4056 and removed from this PR.

app/models/user.rb
slavingia: @gumclaw investigate and fix

gumclaw: Investigated. The issue here is that removing `SocialFacebook` and `SocialTwitter` from the User include chain deletes the class methods they provide, but `config/initializers/doorkeeper.rb` (lines 30-34) still calls `User.fb_object` and references `twitter_oauth_token`:

```ruby
if params.key?(:facebookToken)
  profile = User.fb_object("me", token: params[:facebookToken])  # NoMethodError - fb_object gone
  user = User.find_by(facebook_uid: profile["id"])
elsif params.key?(:twitterToken)
  user = User.where(twitter_oauth_token: params[:twitterToken]).first  # works (column query) but dead code
```

The `facebookToken` path would raise `NoMethodError` in production for any mobile API client still sending Facebook tokens. The `twitterToken` path technically works since it's just a column query, but is dead code now.

**Fix:** Remove both the `facebookToken` and `twitterToken` branches from the `resource_owner_from_credentials` block in `config/initializers/doorkeeper.rb`.

I'll open a fix.

slavingia:
@gumclaw move the Claude settings change to its own PR

gumclaw:
Done — moved to #4056 (already merged). The last commit on this branch (beb5d97) removes the .claude/settings.json that was accidentally included here.

slavingia:
Well merge tmmrw thanks

gianfrancopiana:
@gumclaw See below:

1. The `facebookToken` branch calls `User.fb_object`, which no longer exists after this PR deletes `SocialFacebook`. Any mobile client still sending a Facebook token to `/oauth/token` will get a `NoMethodError`. The `twitterToken` branch is dead code too. Both need to be removed here.

2. `spec/sidekiq/renew_facebook_access_tokens_worker_spec.rb` wasn't deleted. Will break CI since the worker is gone.

gianfrancopiana:
@gumclaw Delete `spec/controllers/users/oauth_controller_spec.rb` to fix the Test Fast 9 failure.

gumclaw:
Deleted `spec/controllers/users/oauth_controller_spec.rb` in 1e40662.

PR #4064, opened 2026-03-24
Add Apple login in the iOS app

binary-koan:
Fixes https://github.com/antiwork/gumroad-mobile/issues/118

## Why

Apple app review requires adding Apple login alongside any other authentication methods.

## How

Adds the Omniauth Apple provider and sets up sign in with Apple using the standard Devise flow.

## Screenshots

image


app/javascript/components/Authentication/SocialAuth.tsx
```diff
@@ -25,6 +34,12 @@ export const SocialAuth = () => {
         <Google pack="brands" className="size-5" />
         Google
       </SocialAuthButton>
+      {typeof window !== "undefined" && Mobile.isOnMobileiOSDevice() && !navigator.standalone ? (
```
binary-koan: Just enabling in the iOS app popup for now, but we can remove this check to add it on web as well any time

app/models/concerns/user/social_apple.rb
```diff
@@ -4,6 +4,43 @@ module User::SocialApple
   extend ActiveSupport::Concern
 
   class_methods do
+    def find_or_create_for_apple_oauth(data)
```
binary-koan, line 7: Based on `find_or_create_for_google_oauth` - a bit harder to deduplicate this one

config/initializers/005_apple.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+APPLE_CLIENT_ID = GlobalConfig.get("APPLE_WEB_CLIENT_ID")
```
binary-koan, line 3: Note: We have some credentials already for the old app's native Apple sign-in functionality, but we need different credentials for web

app/controllers/user/omniauth_callbacks_controller.rb
```diff
@@ -3,6 +3,8 @@
 class User::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include PageMeta::Base
 
+  skip_before_action :verify_authenticity_token, only: [:apple]
```
binary-koan, line 6: The Apple callback is a POST request but we don't have the session available to get the CSRF token, see below

config/initializers/devise.rb
```diff
@@ -292,6 +293,15 @@
...
                   scope: "email,profile"
 
+  config.omniauth :apple,
+                  APPLE_CLIENT_ID,
+                  "",
+                  scope: "email name",
+                  team_id: APPLE_TEAM_ID,
+                  key_id: APPLE_KEY_ID,
+                  pem: APPLE_PRIVATE_KEY,
+                  provider_ignores_state: true
```
binary-koan, line 284: This is also added because of the cross-site POST request

config/initializers/005_apple.rb
```diff
@@ -0,0 +1,19 @@
...
+APPLE_TEAM_ID = GlobalConfig.get("APPLE_WEB_TEAM_ID")
+APPLE_KEY_ID = GlobalConfig.get("APPLE_WEB_KEY_ID")
+APPLE_PRIVATE_KEY = GlobalConfig.get("APPLE_WEB_PRIVATE_KEY")
+
+# Apple Sign In sends a POST callback from appleid.apple.com. Because
+# SameSite=Lax cookies are not sent with cross-site POSTs, the session
+# is empty on the callback and the nonce stored during the request phase
+# cannot be retrieved. Skip nonce verification — the authorization code
+# is single-use and bound to the redirect URI, so the flow is still secure.
```
binary-koan: The alternative to this is editing `config/initializers/session_store.rb` to give the session cookie `SameSite=None`, but that's also not ideal as it'll be sent to any other third party website that posts to us (not just Apple). I'm hesitant to change something so fundamental just to make Apple work, and there doesn't seem to be a config option we can set to avoid the monkey-patch

config/initializers/005_apple.rb
```diff
@@ -0,0 +1,19 @@
...
+OmniAuth::Strategies::Apple.class_eval do
+  private
+
+  def verify_nonce!(id_token)
```
gianfrancopiana: This change disables the last check that ties the Apple callback to the browser session that started login. With the CSRF check skipped, `state` ignored, and nonce verification bypassed, the callback will accept any valid Apple response for this app, even if it came from a different browser session.

That creates a login-CSRF/account-confusion risk: an attacker can start Apple login for their own account, stop before completion, then cause a victim browser to submit that callback and get signed into the attacker's Gumroad account. A signed `id_token` and a single-use code prove the response came from Apple, but not that it belongs to this user's session.

Unless there's another check in the middleware stack that I'm missing, the root cause is that Apple's callback is a cross-site POST, so the normal session cookie is missing. We still need another way to bind the callback to the login that started it. Options include:
- Store the nonce/state in a short-lived server-side record and look it up from a value carried through the callback
- Move this flow to a separate cookie or store that can safely use `SameSite=None; Secure`
- Add PKCE if the library flow can support it here

app/javascript/components/Authentication/SocialAuth.tsx
```diff
@@ -21,6 +30,12 @@ export const SocialAuth = () => {
...
+        <SocialAuthButton provider="apple" href={Routes.user_apple_omniauth_authorize_path({ referer: next })}>
```
gianfrancopiana: This redirect logic depends on `referer`, but a real Apple callback will not have it. OmniAuth normally keeps request params in the session.

A user who starts Apple login from something like `?next=/checkout` will fall back to the default dashboard redirect instead of returning where they came from. The current controller specs pass `referer` directly, so they do not cover the real callback path.

If we add server-side state for the Apple flow, it can carry this value too.

app/javascript/components/Authentication/SocialAuth.tsx
gianfrancopiana: This condition also matches a normal iPhone user in Safari, not just the iOS app's in-app browser. `navigator.standalone` is only true for a home-screen web app, so it does not separate Safari from an in-app WebView.

If the goal is really "iOS app only", I think this needs a signal from the app itself or a feature flag.

app/javascript/components/Authentication/SocialAuth.tsx
binary-koan: Ah yeah we do need that for mobile login to redirect to the OAuth authorization page (not sure why it was working when I tested before ...)

Added to the cookie

spec/config/initializers/apple_strategy_patch_spec.rb
```diff
@@ -0,0 +1,211 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe "Apple OmniAuth strategy cookie-based nonce patch" do
```
binary-koan, line 5: These specs aren't great (it's quite hard to test this class) but I figure it's better to have something so we're more likely to be alerted if the library behaviour changes on upgrade

app/controllers/oauth/mobile_pre_authorizations_controller.rb
```diff
@@ -4,6 +4,13 @@ class Oauth::MobilePreAuthorizationsController < ApplicationController
   before_action :hide_layouts
 
   def new
+    cookies[:is_gumroad_mobile_app] = {
```
binary-koan, line 7: This will still be set if the user goes to the Gumroad login page in Safari within 5 minutes, but I think that's a small enough edge case that we don't need to worry for now. If we want to keep this long term we could render a blank page and use `sessionStorage` or something to make sure it's scoped to just the login window

Note that our `SecureHeaders` sets all cookies to `HttpOnly`, which is why we check it in the controller actions below rather than directly in JS

config/initializers/005_apple.rb
```diff
@@ -0,0 +1,82 @@
...
+    result
+  end
+
+  def callback_phase
+    env["omniauth.params"] = cookie_data.except("nonce") if cookie_data
+
+    # Apple sends the `user` POST param as a JSON string. Devise/Warden
+    # hooks expect params[:user] to be a Hash. Parse it in-place so
+    # downstream code (e.g. devise-pwned_password) doesn't blow up.
+    form_hash = env["rack.request.form_hash"]
```
binary-koan, line 51: Fix for an additional bug I found while testing (only seems to come up when logging in for the first time)

config/initializers/005_apple.rb
```diff
@@ -0,0 +1,82 @@
...
+    result = super
+
+    cookie_data = { "nonce" => @apple_oauth_nonce }
+    cookie_data["referer"] = request.params["referer"] if request.params["referer"].present?
+
+    signed = Rails.application.message_verifier("apple_oauth").generate(
+      cookie_data, purpose: :apple_oauth, expires_in: APPLE_OAUTH_COOKIE_TTL
+    )
+
+    Rack::Utils.set_cookie_header!(
```
binary-koan, line 30: I tried a few solutions for passing through the state, and a cookie seems to be the simplest way - it doesn't touch too much of the library code compared to using Redis and/or the `state` param and IMO it's reasonably readable

Just checking nonce for simplicity which I think is still reasonably secure, but lmk if we should add `state` checking too

app/javascript/components/Authentication/SocialAuth.tsx
binary-koan: My bad, that was a convincing hallucination 😅 Updated to use a cookie

config/initializers/005_apple.rb
binary-koan: Thanks for the explanation, makes sense. Moved to a cookie, see my other comments

db/migrate/20260324000000_add_apple_uid_to_users.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddAppleUidToUsers < ActiveRecord::Migration[7.1]
+  def change
+    change_table :users, bulk: true do |t|
+      t.string :apple_uid
```
binary-koan: I guess we'll run into the same issue as yesterday with adding a column to `users`. @gianfrancopiana @ershad think we can get the add column to work somehow, or is it worth creating a separate table?

slavingia:
for iOS can we put at the top?

db/migrate/20260324000000_add_apple_uid_to_users.rb
gianfrancopiana: Yes, given the 120-minute timeout on the production deploy step I'd avoid changing users in this PR. 

I think the safer approach is a small new table for external logins, for example `user_external_authentications` with `user_id`, `provider`, and `uid`, plus a unique index on `(provider, uid)`. 

On login, we would first look up the row for `provider = apple` and the incoming `uid`; if none exists, we can fall back to the current email-matching logic, then create the row for future logins. That keeps this as a new-table migration, which is lower risk operationally than altering users.

I get this is a little inconsistent with the existing providers (e.g., Google), but I still think a new table is the better tradeoff here. It lets us ship Apple without a risky users table alter inside a timed deploy. We do not need to migrate the other providers in this PR, and we may never need to unless we decide the cleanup is worth a dedicated migration rollout later. But that migration would be separate from product.

Let me know what you think!

config/initializers/005_apple.rb
gianfrancopiana: I think this still misses the real first-login path. This only rewrites user if the POST body was already parsed, but in the normal callback flow that has not happened yet. I tested this branch with a minimal runner without pre-parsing the request body: `user_info` was still a string and `first_name` raised `NoMethodError` on dig. The current spec hides that because it parses the body first. We probably need to parse user here, or parse it inside `user_info`

db/migrate/20260324000000_add_apple_uid_to_users.rb
slavingia: Why are we scared of deploys? We should address this tomorrow.

config/initializers/005_apple.rb
binary-koan: > in the normal callback flow that has not happened yet

Where are you seeing this? Logging in as a new user without an account works for me in an iOS simulator

binary-koan: Note it's parsed inside `user_info` by default, the issue is that `devise-pwned_password` tries to use the param directly here https://github.com/michaelbanfield/devise-pwned_password/blob/master/lib/devise/pwned_password/hooks/pwned_password.rb#L5

db/migrate/20260324000000_add_apple_uid_to_users.rb
gianfrancopiana: Plan is to discuss with @ershad tomorrow.

config/initializers/005_apple.rb
gianfrancopiana: I did not test this in the simulator myself, so I should have been clearer earlier.

I think you’re right that the original reason for this patch is `devise-pwned_password`, and upstream `omniauth-apple` normally parses `user` inside `user_info`.

My remaining concern is that this branch also overrides `user_info`, so we lose that fallback. If the early `form_hash` rewrite does not run on the normal callback path, `user` may still be a string.

Can we add a test that exercises the callback without pre-parsing the body?

binary-koan: Yeah that's fair, added fallback parsing in `user_info` and a test

slavingia:
@gumclaw resolve merge conflicts and merge upon green build

gumclaw:
Merge conflicts resolved and pushed. Waiting for CI to go green, then I'll merge. 🐾

binary-koan:
@gumclaw get specs passing but hold off on merging, need to add the credentials first

gumclaw:
Got it — I'll keep an eye on CI for this push and make sure specs are green. Won't merge until you've added the credentials.

gumclaw:
⚠️ CI check failed — not merging automatically.

**Failed check:** `buildkite/gumroad` — [Build #7043](https://buildkite.com/gumroad-inc/gumroad/builds/7043)

All GitHub Actions checks (tests, autofix, etc.) passed, but the Buildkite build failed after 7 seconds. This might be a flaky/infra issue worth re-running.

db/migrate/20260324000000_add_apple_uid_to_users.rb
ershad: 1. Makes sense to use a new model to store Apple auth details.
2. Issue with adding `tiktok_pixel_id` to `users` the other day: Seems like two migrations ran at the same time, which shouldn't happen ideally since we have locks. We should debug if this happens again.

PR #4066, opened 2026-03-24
Fix mobile auth prompt UI to use Tailwind

binary-koan:
This was missed in #2558, so the layout broke when we removed the old CSS.

Before:

Screenshot_1774138239

After:

Screenshot_1774138205


binary-koan:
Super minor fix, will go ahead and merge

PR #4067, opened 2026-03-24
Revert "Add TikTok Pixel integration to third-party analytics"

gianfrancopiana:
Reverts antiwork/gumroad#4024

## What

Reverts `#4024` and removes the pending `users.tiktok_pixel_id` migration from the production deploy.

## Why

The production deploy on March 24, 2026 got stuck in `db:migrate` because this schema change runs through `pt-online-schema-change` on the `users` table. Production has not applied the migration, so reverting it is the lowest-risk way to unblock deploys and return production to a clean pre-migration state.


gianfrancopiana:
I think reverting this was the right call to unblock production.

@gumclaw For the follow-up, I would avoid adding another column to the `users` table. That table is very large, and even this simple change turned into a long-running deploy.

A safer version of the TikTok work would store the pixel ID in the seller's existing settings data instead of changing the table itself. We only need to save the value and read it back when loading analytics, so we do not need a large database change for this feature.

The only extra thing to make sure of is that product data gets refreshed when a seller updates the pixel ID.

gumclaw:
Good call, thanks for the context on the users table size. I'll rework the follow-up to store the pixel ID in the seller's existing settings data instead of adding a column. Will make sure product data cache gets invalidated when the pixel ID changes too.

gianfrancopiana:
@gumclaw Where is the new PR?

Issue #4019, opened 2026-03-18
API for creating and editing products


gumclaw:
Expose a public API for creators to programmatically create and edit products.

This would let creators build integrations, bulk-manage inventory, sync from other platforms, and automate their workflows.

Needs scoping: which product fields, rate limits, auth method (OAuth vs API key), versioning strategy.

Update the [API docs](https://app.gumroad.com/api) to document the new endpoints, request/response schemas, and usage examples.

cc @slavingia

gumclaw:
Closing as the API PRs are tracked separately. Opening a new issue for the CLI commands (products create and products update).

PR #4068, opened 2026-03-24
Extend v2 product API response with rich content, files, covers, and bundle products

gianfrancopiana:
Ref #4019

## What

Extends the v2 product API with the product-management fields that were missing from single-product responses: `rich_content`, `files`, `covers`, `main_cover_id`, `bundle_products`, `has_same_rich_content_for_all_variants`, and per-variant `rich_content`.

Keeps `GET /v2/products` intentionally slim. Those detail-only fields are omitted from the list response, while `GET /v2/products/:id` and the existing single-product mutation responses (`enable` / `disable`) return the full payload.

For uploaded files, single-product responses now return signed URLs instead of raw private S3 URLs, so third-party clients can actually download them. Links to files hosted outside Gumroad are unchanged.

Adds endpoint-specific eager loading so the new fields do not introduce N+1 queries.

## Why

The issue thread and product API plan require a read-side way for API clients to fetch the editable shape of a product before the create and update endpoints land. Without these fields, a client can write rich content, files, covers, and bundle contents but cannot read them back from the v2 product API.

During review, it became clear that returning full rich content and per-file signed URLs on `index` would make the list response heavier than it needs to be. The current split is:

- `index` is a lightweight summary response
- single-product responses are the source of truth for full editable product state

That still works well for the planned CLI flow: list products, fetch one product when full detail is needed, then call create/update/enable/disable as separate steps.

## Notes

- This is no longer "purely additive" on `GET /v2/products`. The list response is now intentionally slimmer than the original draft of this PR.
- The full new fields are still additive on single-product responses.

---

This PR was implemented with AI assistance using Claude Sonnet 4.6 and gpt‑5.4 xhigh.

### DIFF
--- a/app/models/concerns/product/as_json.rb
+++ b/app/models/concerns/product/as_json.rb
@@ -98,18 +98,6 @@ module Product::AsJson
         "custom_summary" => custom_summary,
         "is_tiered_membership" => is_tiered_membership?,
         "recurrences" => is_tiered_membership? ? prices.alive.is_buy.map(&:recurrence).uniq : nil,
-        "rich_content" => slim ? nil : rich_content_json,
-        "has_same_rich_content_for_all_variants" => slim ? nil : has_same_rich_content_for_all_variants?,
-        "files" => slim ? nil : ordered_alive_product_files.map do |f|
-          {
-            id: f.external_id,
-            name: f.display_name,
-            size: f.size,
-            url: f.signed_url,
-            filetype: f.filetype,
-            filegroup: f.filegroup,
-          }
-        end,
         "covers" => display_asset_previews.as_json,
         "main_cover_id" => main_preview&.guid,
         "bundle_products" => is_bundle? ? bundle_products.select(&:alive?).sort_by(&:position).map do |bp|
@@ -130,8 +118,9 @@ module Product::AsJson
                 is_pay_what_you_want: variant.customizable_price?,
                 recurrence_prices: is_tiered_membership? ? variant.recurrence_price_values : nil,
                 url: nil, # Deprecated
-                rich_content: slim ? nil : variant.rich_content_json,
-              }
+              }.tap do |option|
+                option[:rich_content] = variant.rich_content_json unless slim
+              end
             end.map do
               ppp_factors.blank? ? _1 :
                 _1.merge({
@@ -144,6 +133,23 @@ module Product::AsJson
           }
         end
       )
+
+      unless slim
+        json.merge!(
+          "rich_content" => rich_content_json,
+          "has_same_rich_content_for_all_variants" => has_same_rich_content_for_all_variants?,
+          "files" => ordered_alive_product_files.map do |f|
+            {
+              id: f.external_id,
+              name: f.display_name,
+              size: f.size,
+              url: f.signed_url,
+              filetype: f.filetype,
+              filegroup: f.filegroup,
+            }
+          end
+        )
+      end
 
       if preorder_link.present?
         json.merge!(
--- a/spec/controllers/api/v2/links_controller_spec.rb
+++ b/spec/controllers/api/v2/links_controller_spec.rb
@@ -33,6 +33,22 @@ describe Api::V2::LinksController do
           products: [@product2, @product1]
         }.as_json(api_scopes: ["view_public"], slim: true))
       end
+
+      it "omits detail-only fields from the slim response" do
+        versioned_product = create(:product_with_digital_versions, user: @user, created_at: Time.current + 7200)
+        create(:product_file, link: versioned_product, url: "#{S3_BASE_URL}specs/test.pdf")
+        create(:rich_content, entity: versioned_product, description: [{ "type" => "paragraph", "content" => [{ "type" => "text", "text" => "hello" }] }])
+        create(:rich_content, entity: versioned_product.alive_variants.first, description: [{ "type" => "paragraph", "content" => [{ "type" => "text", "text" => "variant" }] }])
+
+        get @action, params: @params
+
+        product = response.parsed_body["products"].find { |item| item["id"] == versioned_product.external_id }
+        expect(product).to be_present
+        expect(product).not_to have_key("rich_content")
+        expect(product).not_to have_key("has_same_rich_content_for_all_variants")
+        expect(product).not_to have_key("files")
+        expect(product["variants"].first["options"].first).not_to have_key("rich_content")
+      end
     end
 
     describe "when logged in with sales scope" do


app/controllers/api/v2/links_controller.rb
```diff
@@ -6,11 +6,15 @@ class Api::V2::LinksController < Api::V2::BaseController
   before_action :check_types_of_file_objects, only: [:update, :create]
   before_action :set_link_id_to_id, only: [:show, :update, :disable, :enable, :destroy]
   before_action :fetch_product, only: [:show, :update, :disable, :enable, :destroy]
+  before_action :preload_product_associations, only: [:show]
 
   def index
     products = current_resource_owner.products.visible.includes(
```
slavingia, line 27: Don't need to update the single get product with ID?

gianfrancopiana: Yep, `show` is covered too. `index` uses `includes(...)`, and `show` uses `preload_product_associations`, so the single-product path preloads the same new associations without refetching `@product`.

slavingia:
Maybe @ershad should review this

app/controllers/api/v2/links_controller.rb
```diff
@@ -69,6 +73,19 @@ def error_with_product(product = nil)
...
+    def preload_product_associations
+      ActiveRecord::Associations::Preloader.new(
+        records: [@product],
+        associations: [
+          :preorder_link, :tags, :taxonomy,
+          :alive_rich_contents, :ordered_alive_product_files,
+          { display_asset_previews: [:file_attachment, :file_blob] },
+          { bundle_products: [:product, :variant] },
+          { variant_categories_alive: [{ alive_variants: :alive_rich_contents }] },
+        ]
```
binary-koan: Can we extract this to a constant to keep it in sync?

It would be clearer to me to have the associations as a constant and inline the preloader into `show`, so it looks like:

```ruby
def show
  ActiveRecord::Associations::Preloader.new(records: [@product], associations: PRODUCT_RESPONSE_ASSOCIATIONS)

  # ...
end
```

app/models/concerns/product/as_json.rb
```diff
@@ -97,6 +97,28 @@ def as_json_for_api(options)
         "custom_summary" => custom_summary,
         "is_tiered_membership" => is_tiered_membership?,
         "recurrences" => is_tiered_membership? ? prices.alive.is_buy.map(&:recurrence).uniq : nil,
+        "rich_content" => rich_content_json,
```
binary-koan: Rich content can get very large - is it a good idea to have it returned on `index` or can/should we restrict it to `show`?

spec/controllers/api/v2/links_controller_spec.rb
```diff
@@ -55,6 +55,42 @@
...
+        ActiveSupport::Notifications.subscribed(->(*) { count += 1 }, "sql.active_record", &block)
+        count
+      end
+
+      before do
+        @token = create("doorkeeper/access_token", application: @app, resource_owner_id: @user.id, scopes: "view_public")
+        @params.merge!(format: :json, access_token: @token.token)
+      end
+
+      it "batch loads file_attachment and file_blob for covers (not one query per record)" do
```
binary-koan, line 87: Just a thought - we already have `bullet` so I wonder if we can use it for checks like this. Claude suggests we could do something like:

```ruby
Bullet.enable = true
Bullet.start_request

# call the endpoint

notifications = Bullet.notification_collector.collection

# assert on notifications, ideally whitelisting known N+1s rather than blacklisting
# each notification has these useful attributes:
#   notification.body      # human-readable description
#   notification.base_class # e.g. "Purchase"
#   notification.associations # e.g. [:seller, :product]

Bullet.end_request
Bullet.enable = false
```

spec/controllers/api/v2/links_controller_spec.rb
```diff
@@ -168,6 +204,73 @@
...
+
+        get :show, params: @params
+        expect(response.parsed_body["product"]["bundle_products"]).to be_an(Array)
+        expect(response.parsed_body["product"]["bundle_products"].length).to eq(bundle.bundle_products.alive.count)
+        expect(response.parsed_body["product"]["bundle_products"].first).to include("product_id", "quantity", "position")
+      end
+
+      it "includes empty bundle_products for non-bundle products" do
+        get :show, params: @params
+        expect(response.parsed_body["product"]["bundle_products"]).to eq([])
```
binary-koan, line 284: Seems we're not very consistent with sending empty fields vs omitting them entirely in cases like this - do you think in general we should do this and always include the field?

This is one example which could potentially be confusing because bundle_products will never exist for non-bundle products (unless we change that as part of the redesign) and sending an empty array makes it look like it might

app/models/concerns/product/as_json.rb
```diff
@@ -97,6 +97,28 @@ def as_json_for_api(options)
...
+        "has_same_rich_content_for_all_variants" => has_same_rich_content_for_all_variants?,
+        "files" => ordered_alive_product_files.map do |f|
+          {
+            id: f.external_id,
+            name: f.display_name,
+            size: f.size,
+            url: f.url,
```
ershad: We would need to use signed URLs here if we want third-party clients to access the files. Otherwise, `f.url` would be raw S3 URLs pointing to a private bucket.

spec/controllers/api/v2/links_controller_spec.rb
gianfrancopiana: I kept the explicit query test here because it checks that cover attachments and blobs are loaded in a batch rather than one query per record. I also don't see an existing Bullet-based spec pattern in this area of the codebase, so adding it here felt like more test setup without a clear benefit for this specific case. Happy to revisit Bullet separately if broader N+1 coverage would be useful.

spec/controllers/api/v2/links_controller_spec.rb
gianfrancopiana: I would prefer to keep `bundle_products: []`. That is already the shape used in other product-related responses and client code, for example [ProductPresenter](https://github.com/antiwork/gumroad/blob/797df3c96c43fb1b7448bfa6ba8f80b3ce825b6e/spec/presenters/product_presenter_spec.rb#L152), [CheckoutPresenter](https://github.com/antiwork/gumroad/blob/797df3c96c43fb1b7448bfa6ba8f80b3ce825b6e/spec/presenters/checkout_presenter_spec.rb#L180), and [cart.ts](https://github.com/antiwork/gumroad/blob/797df3c96c43fb1b7448bfa6ba8f80b3ce825b6e/app/javascript/utils/cart.ts#L74). Changing it here would make this API shape less consistent and would broaden the scope of the PR.

app/models/concerns/product/as_json.rb
gianfrancopiana: Good catch. For single-product responses (`show`, `disable`, `enable`), uploaded files now return a signed URL. Links to files hosted outside Gumroad are unchanged. On `index`, I intentionally left `files` out of the slim response. That keeps the list payload lightweight and avoids generating a large number of short-lived signed URLs. Clients that need file URLs can fetch them from the single-product endpoints.

gianfrancopiana:
@binary-koan @ershad Thanks both. I made one more pass on this PR based on the feedback here. `index` is now intentionally slim, while single-product responses still return the full product-editing data. I also switched uploaded file URLs on single-product responses to signed URLs, so third-party clients can use them.

Updated the PR description to reflect that `index` is no longer the full payload and that the detailed shape lives on the single-product endpoints.

PR #4069, opened 2026-03-24
Add presigned S3 upload endpoint to v2 API

gianfrancopiana:
Ref #4019

## What

Adds two endpoints to the v2 API for uploading files to S3 before attaching them to products:

- `POST /v2/files/presign` — initiates an S3 multipart upload. Takes a `filename` and `file_size`, returns an `upload_id`, a `key`, a `file_url`, and an array of presigned part URLs (one per 100 MB chunk). The client PUTs each part directly to S3 using its `presigned_url`.
- `POST /v2/files/complete` — finalizes the upload. Takes the `upload_id`, `key`, and the list of `{ part_number, etag }` pairs returned by S3 after each PUT.

Once complete, the `file_url` from either response can be passed in the `files` param of a product create or update request.

Files are scoped to the authenticated seller. The key is prefixed with `attachments/{seller_external_id}/`, and the complete endpoint rejects keys that don't match the caller's prefix.

## Why

OAuth clients need to upload files before attaching them to products. The web frontend does this via Evaporate and a session-authenticated signing endpoint, neither of which is accessible to OAuth clients.

A single presigned PUT would cap uploads at 5 GB. The web uploader supports up to 20 GB, so this implementation uses S3 multipart upload to match that limit. Because the client supplies `file_size` upfront, all part URLs can be generated in one request, allowing parallel uploads without additional server round-trips.

The endpoints are seller-scoped rather than product-scoped so files can be uploaded before a product exists, which is required for a create-with-files flow.

---

This PR was implemented with AI assistance using Claude Sonnet 4.6 and gpt-5.4 xhigh

### DIFF
--- a/spec/requests/api/v2/files_multipart_spec.rb
+++ b/spec/requests/api/v2/files_multipart_spec.rb
@@ -5,8 +5,8 @@ require "net/http"
 
 describe "Files multipart upload" do
   let(:user) { create(:user) }
-  let(:app) { create(:oauth_application, owner: create(:user)) }
-  let(:token) { create("doorkeeper/access_token", application: app, resource_owner_id: user.id, scopes: "edit_products") }
+  let(:oauth_application) { create(:oauth_application, owner: create(:user)) }
+  let(:token) { create("doorkeeper/access_token", application: oauth_application, resource_owner_id: user.id, scopes: "edit_products") }
 
   before do
     host! "test.gumroad.com"


spec/controllers/api/v2/files_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe Api::V2::FilesController do
+  before do
+    @user = create(:user)
```
binary-koan: Would be nice to use `let` for new spec files

app/controllers/api/v2/files_controller.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+class Api::V2::FilesController < Api::V2::BaseController
+  before_action { doorkeeper_authorize! :edit_products }
+
+  PRESIGNED_URL_EXPIRY_SECONDS = 900
+  PART_SIZE = 100 * 1024 * 1024
+  MAX_FILE_SIZE_GB = 20 # matching the web uploader limit
+  MAX_FILE_SIZE = MAX_FILE_SIZE_GB * 1024 * 1024 * 1024
```
binary-koan:
```suggestion
  PART_SIZE = 100.megabytes
  MAX_FILE_SIZE_GB = 20 # matching the web uploader limit
  MAX_FILE_SIZE = MAX_FILE_SIZE_GB.gigabytes
```

slavingia:
We may want to consider a job to delete files that are never attached to a product.

gianfrancopiana:
> We may want to consider a job to delete files that are never attached to a product.

@gumclaw Create an issue for cleanup of seller-scoped attachment uploads that are never associated with a ProductFile. Open questions: what grace period should we allow before cleanup, and do we want an app-level record of pending uploads (preferred, avoids bucket scans)? Also check whether S3 already aborts incomplete multipart uploads via lifecycle rules for uploads that never reach /complete.

gumclaw:
Opened [#4112](https://github.com/antiwork/gumroad/issues/4112) to track orphaned upload cleanup — covers grace period, app-level tracking vs. bucket scans, and S3 multipart lifecycle rules.

Issue #4036, opened 2026-03-18
Support multiple subscriptions to the same product and let buyers choose between starting a new subscription or restarting an existing one


gumclaw:
## What
When a buyer tries to purchase a subscription they already have, or had previously, Gumroad should not force a single path.

Instead, the product should let the buyer choose between:
- starting a new / additional subscription
- restarting their existing subscription

This is especially important for cases where buyers intentionally need multiple concurrent subscriptions, such as resellers or agencies purchasing on behalf of different clients.

## Why
The current behavior makes subscription repurchase flows ambiguous and can block legitimate multi-license / multi-client use cases.

Buyers need a clear choice at the point of repurchase so Gumroad does not collapse two distinct intents into one behavior.

PR #4070, opened 2026-03-25
Let logged-in buyers choose between restarting or starting a new subscription

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad/issues/4036

## What

When a logged-in buyer with an existing membership clicks the subscribe button on a product page, a choice modal now appears instead of silently blocking or auto-restarting:

- Active subscription - "You already have an active subscription" modal with "Yes, start a new subscription" and "Cancel" buttons
- Expired/Ended subscription - "Resume your previous subscription?" modal with "Yes, resume subscription" and "No, start a new subscription" buttons
- Logged-out users - Current auto-restart behavior unchanged (to avoid leaking subscription state)

"No/Yes, start a new subscription" navigates to checkout with a `force_new_subscription` flag that bypasses the existing subscription check in `Purchase::CreateService`, allowing a new subscription to be created. "Yes, resume subscription" links to the existing subscription manage page.

## Why

The current behavior collapses two distinct buyer intents into one path. Buyers who intentionally need multiple concurrent subscriptions (resellers, agencies purchasing on behalf of different clients) are blocked. Buyers with lapsed subscriptions have no choice between restarting their old subscription or starting fresh. This change gives buyers an explicit choice at the point of repurchase.

Logged-out users keep auto-restart because showing subscription state to an unauthenticated user would leak private information, per the discussion in #173.

### Purchase again with active subscription

Screenshot 2026-03-30 at 11 08 53
Screenshot 2026-03-30 at 11 09 04

### Purchase again with past subscription

Screenshot 2026-03-30 at 11 12 06
Screenshot 2026-03-30 at 11 12 14

### Purchase without login using same email (No changes)

Screenshot 2026-03-25 at 12 14 55

---

Claude Opus 4.6 used for this PR and all code self reviewed

### DIFF
--- a/app/javascript/components/Product/SubscriptionChoiceModal.tsx
+++ b/app/javascript/components/Product/SubscriptionChoiceModal.tsx
@@ -7,11 +7,11 @@ import { Purchase } from "$app/components/Product";
 type Props = {
   purchase: Purchase;
   checkoutUrl: string;
-  open: boolean;
   onClose: () => void;
 };
 
-export const SubscriptionChoiceModal = ({ purchase, checkoutUrl, open, onClose }: Props) => {
+export const SubscriptionChoiceModal = ({ purchase, checkoutUrl, onClose }: Props) => {
+  const open = checkoutUrl !== "";
   const newSubscriptionHref = React.useMemo(() => {
     if (!checkoutUrl) return "";
     const url = new URL(checkoutUrl, window.location.origin);
--- a/app/javascript/components/Product/index.tsx
+++ b/app/javascript/components/Product/index.tsx
@@ -647,7 +647,6 @@ export const Product = ({
         <SubscriptionChoiceModal
           purchase={purchase}
           checkoutUrl={checkoutUrlForModal ?? ""}
-          open={checkoutUrlForModal !== null}
           onClose={() => setCheckoutUrlForModal(null)}
         />
       ) : null}


app/services/purchase/create_service.rb
```diff
@@ -179,7 +179,7 @@ def is_gift?
     end
 
     def should_check_for_restartable_subscription?
-      product.is_recurring_billing && !is_gift?
+      product.is_recurring_billing && !is_gift? && !(buyer.present? && params[:force_new_subscription])
```
nyomanjyotisa: Only skip the subscription check when the buyer is authenticated. Logged-out users keep the existing auto-restart behavior to avoid leaking subscription state

app/presenters/checkout_presenter.rb
```diff
@@ -122,6 +122,7 @@ def checkout_product(product, cart_item, params, include_cross_sells: true)
       quantity: cart_item[:quantity],
       call_start_time: cart_item[:call_start_time],
       pay_in_installments: cart_item[:pay_in_installments],
+      force_new_subscription: logged_in_user.present? && (cart_item[:force_new_subscription] || false),
```
nyomanjyotisa, line 125: Gate on `logged_in_user` server-side so a logged-out user can't pass the flag via URL and have it take effect after logging in during checkout

app/javascript/components/Product/index.tsx
```diff
@@ -580,7 +584,20 @@ export const Product = ({
...
-              if (!validate()) e.preventDefault();
+              if (!validate()) {
+                e.preventDefault();
+                return;
+              }
+              if (
+                loggedInUser &&
+                purchase &&
+                (purchase.membership || purchase.subscription_has_lapsed) &&
+                product.is_recurring_billing
```
nyomanjyotisa, line 594: Only intercept the CTA for logged-in users with a subscription on a recurring product. Everyone else flows straight to checkout

app/javascript/components/Product/index.tsx
```diff
@@ -257,6 +258,9 @@ export const Product = ({
   disableAnalytics?: boolean;
 }) => {
   const [pageLoaded, setPageLoaded] = React.useState(false);
+  const [subscriptionModalOpen, setSubscriptionModalOpen] = React.useState(false);
+  const [checkoutUrlForModal, setCheckoutUrlForModal] = React.useState("");
```
binary-koan: Do we need two states? Could we just make `checkoutUrlForModal` null when the modal is closed and present when the modal is open?

app/javascript/components/Product/SubscriptionChoiceModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <p>
+          You've previously subscribed to this product. Would you like to <strong>pick up where you left off</strong>,
+          or <strong>start fresh with a new subscription</strong>?
+        </p>
+        <div className="grid gap-3 sm:grid-cols-2">
+          <NavigationButton href={newSubscriptionHref} target="_top">
+            No, start a new subscription
+          </NavigationButton>
+          <NavigationButton href={purchase.membership.manage_url} color="black" target="_blank">
+            Yes, resume subscription
```
binary-koan: Nit: I find it a bit confusing to have "no" and "yes" on these buttons - the question is "do you want A or B" so saying "yes" isn't a natural answer

IMO just "Start a new subscription" and "Resume subscription" would be better labels

binary-koan:
Sorry for the wait, will finish reviewing tomorrow morning

app/services/order/create_service.rb
```diff
@@ -176,7 +176,7 @@ def build_purchase_params(product, purchase_params)
       additional_params = purchase_params.extract!(
         :is_gift, :price_id, :wallet_type, :perceived_free_trial_duration, :accepted_offer,
         :cart_items, :variants, :bundle_products, :custom_fields, :tip_cents, :call_start_time,
-        :pay_in_installments
+        :pay_in_installments, :force_new_subscription
```
binary-koan: Since we end up excluding this param from `purchase` in `Purchase::CreateService`, might be cleaner to not put it there in the first place?

i.e. this can return `{ purchase: ..., gift:, ..., force_new_subscription: ... }` and we can have a separate ivar in `Purchase::CreateService`

spec/requests/products/show/subscription_choice_modal_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    if state == :lapsed
+      subscription.update!(cancelled_at: 1.day.ago, deactivated_at: 1.day.ago, cancelled_by_buyer: true)
+    end
+
+    subscription
+  end
+
+  context "when logged-in buyer has an active subscription" do
+    before do
+      create_subscription_with_purchase(state: :active)
```
binary-koan:
```suggestion
      create_subscription_with_purchase
```

Nit: Since it's the default

spec/services/purchase/create_service_spec.rb
```diff
@@ -3660,5 +3660,60 @@ def create_subscription_for(product:, purchaser:, email:, **attrs)
...
+      end
+
+      context "when buyer has an active subscription" do
+        before do
+          create_subscription_for(product: membership_product, purchaser: buyer, email: email)
+        end
+
+        it "does not return the existing subscription error" do
+          service = Purchase::CreateService.new(product: membership_product, params: membership_params, buyer:)
+          _purchase, error = service.perform rescue nil
```
binary-koan: The `rescue nil` here could hide real errors if the service changes. We should either expect no exception or a specific one

spec/services/purchase/create_service_spec.rb
```diff
@@ -3660,5 +3660,60 @@ def create_subscription_for(product:, purchaser:, email:, **attrs)
...
+
+      context "when buyer is not logged in" do
+        it "still checks for existing subscriptions" do
+          create_subscription_for(product: membership_product, purchaser: buyer, email: email)
+
+          expect do
+            purchase, error = Purchase::CreateService.new(product: membership_product, params: membership_params, buyer: nil).perform
+
+            expect(purchase).to be_nil
+            expect(error).to be_present
```
binary-koan: Better to check the exact error if we can

app/javascript/components/Product/index.tsx
nyomanjyotisa: Consolidated into a single `checkoutUrlForModal: string | null` state. Modal is open when it's non-null, closed when null

app/javascript/components/Product/SubscriptionChoiceModal.tsx
nyomanjyotisa: Agreed, changed to "Start a new subscription" and "Resume subscription"

app/services/order/create_service.rb
nyomanjyotisa: Updated, extracted it with `.delete` before `extract!` so it's a top-level key, not nested under purchase

PR #4071, opened 2026-03-25
Fix mobile library endpoint exclusions

binary-koan:
## Why

The new app uses `PurchaseSearchService` to paginate the buyer's library and avoid performance issues. However, it looks like this service is currently just used to list sellers' sales, meaning it doesn't exclude various cases the library currently does (e.g. it includes products the user has deleted and products where access has been revoked)

## What

This fixes the issue by adding a `for_library` param to the service, which sets a bunch of filters to bring the results in line with `Purchase.for_library` and `Purchase.for_mobile_listing`

### DIFF
--- a/app/services/purchase_search_service.rb
+++ b/app/services/purchase_search_service.rb
@@ -65,6 +65,7 @@ class PurchaseSearchService
 
   def initialize(options = {})
     @options = DEFAULT_OPTIONS.merge(options)
+    apply_for_library
     build_body
   end
 


app/controllers/api/mobile/purchases_controller.rb
```diff
@@ -30,11 +30,12 @@ def index
   def search
     result = PurchaseSearchService.search(search_options)
     pagination = Pagy.new(count: result.response.hits.total.value, page: @page, limit: @items)
+    purchases = result.records.not_rental_expired
```
binary-koan, line 33: We don't index `rental_expired` in Elasticsearch. I don't think it's worth the effort of adding it and reindexing because no longer support rentals (this just excludes some historical purchases) but that would be another option

app/services/purchase_search_service.rb
```diff
@@ -31,6 +31,10 @@ class PurchaseSearchService
     exclude_non_successful_preorder_authorizations: false,
     exclude_bundle_product_purchases: false,
     exclude_commission_completion_purchases: false,
+    exclude_deleted_by_buyer: false,
+    exclude_additional_contributions: false,
+    exclude_access_revoked: false,
```
binary-koan, line 36: Filters in the `for_library` scope that weren't supported in the service but are already indexed in Elasticsearch

app/services/purchase_search_service.rb
```diff
@@ -75,6 +80,23 @@ def self.search(options = {})
...
+        exclude_additional_contributions: true,
+        exclude_non_original_subscription_purchases: true,
+        exclude_deactivated_subscriptions: true,
+        exclude_gifters: true,
+        exclude_refunded_except_subscriptions: true,
+        exclude_unreversed_chargedback: true,
+        exclude_bundle_product_purchases: true,
+        exclude_commission_completion_purchases: true,
+        exclude_access_revoked: true,
+        exclude_deleted_by_buyer: true,
```
binary-koan, line 96: We could just pass all these from the controller, but I prefer having the consistency of a `for_library` parameter mapping (more or less) to the `for_library` scope

PR #4072, opened 2026-03-25
Fix IDOR and authorization bypass in send_to_kindle endpoint

gianfrancopiana:
## What

`POST /r/:id/send_to_kindle` used a global `ProductFile.find_by_external_id` lookup, so any valid download token could be combined with a file ID from a different product/seller to have that file emailed to an attacker's Kindle. The endpoint also skipped all purchase-state and identity checks that protect every other content route.

This PR:

- Scopes the file lookup to the current `url_redirect` token
- Adds purchase-state checks (refund, chargeback, revoked access, rental expiry, inactive membership) and identity verification (cookie, purchaser, IP) inline with JSON responses — can't use the shared `check_permissions` because it renders HTML
- Enforces `can_send_to_kindle?` server-side and returns clean errors for invalid file IDs
- Updates the frontend to surface backend error messages instead of hardcoding a generic one
- Fixes product file test factories to use `S3_BASE_URL` directly

## Why

Insecure Direct Object Reference (IDOR) – the endpoint downloads files from S3 and emails them as attachments, so this is real file exfiltration, not just a metadata leak.

---

This PR was implemented with AI assistance using Claude Opus 4.6 and gpt-5.4 xhigh.

### DIFF
--- a/app/controllers/url_redirects_controller.rb
+++ b/app/controllers/url_redirects_controller.rb
@@ -228,18 +228,18 @@ class UrlRedirectsController < ApplicationController
 
     purchase = @url_redirect.purchase
     if purchase && (purchase.stripe_refunded || (purchase.chargeback_date.present? && !purchase.chargeback_reversed) || purchase.is_access_revoked)
-      return render json: { success: false, error: "Not found" }, status: :not_found
+      return e404_json
     end
-    return render json: { success: false, error: "Not found" }, status: :not_found if @url_redirect.rental_expired?
-    return render json: { success: false, error: "Not found" }, status: :not_found if purchase&.subscription && !purchase.subscription.grant_access_to_product?
+    return e404_json if @url_redirect.rental_expired?
+    return e404_json if purchase&.subscription && !purchase.subscription.grant_access_to_product?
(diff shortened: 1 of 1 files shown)


app/controllers/url_redirects_controller.rb
```diff
@@ -226,19 +226,36 @@ def confirm
   def send_to_kindle
     return render json: { success: false, error: "Please enter a valid Kindle email address" } if params[:email].blank?
 
+    purchase = @url_redirect.purchase
+    if purchase && (purchase.stripe_refunded || (purchase.chargeback_date.present? && !purchase.chargeback_reversed) || purchase.is_access_revoked)
+      return render json: { success: false, error: "Not found" }, status: :not_found
+    end
+    return render json: { success: false, error: "Not found" }, status: :not_found if @url_redirect.rental_expired?
```
binary-koan: We can use shorthand for all these

```suggestion
    return e404_json if @url_redirect.rental_expired?
```

Feels like we should have a method on `Purchase` that covers a bunch of these actually (equivalent of the `for_library` scope, so something like `.visible_in_library?` maybe?) but I'm fine with this for now to fix the issue

PR #4073, opened 2026-03-25
Fix transitive ID collision when saving rich content file embeds

gumclaw:
## What

Fix a flaky test in `spec/requests/products/dropbox_spec.rb:18` ("Dropbox uploads embeds dropbox files successfully") caused by a transitive ID collision in `save_files!`.

When saving product files with rich content, file embed IDs were being replaced sequentially inside the loop. If ProductFile A's new `external_id` happened to equal DropboxFile B's original `external_id` (since `ObfuscateIds.encrypt` is table-agnostic and different tables can share the same auto-increment ID), the second replacement would overwrite the first, giving both embeds the same ID.

The fix collects all ID mappings into a hash during the loop, then applies them in a single pass after the loop completes. No transitive collisions possible.

Also removes `.claude/worktrees/` from git tracking (was accidentally committed and broke CI checkout).

## Why

This flaky test was failing intermittently on CI ([example run](https://github.com/antiwork/gumroad/actions/runs/23505761791)). The failure manifested as two `fileEmbed` entries in rich content having the same `external_id` instead of unique IDs for each file. The sequential ID replacement approach was fundamentally racy when multiple new files were created in the same save operation.

## Test Results

5 consecutive green CI runs after the fix:
1. https://github.com/antiwork/gumroad/actions/runs/23514016240
2. https://github.com/antiwork/gumroad/actions/runs/23514702175
3. https://github.com/antiwork/gumroad/actions/runs/23515391695
4. https://github.com/antiwork/gumroad/actions/runs/23515965864
5. https://github.com/antiwork/gumroad/actions/runs/23516518861

---

Prompts used:
- "Fix flaky dropbox test spec/requests/products/dropbox_spec.rb:18 — find root cause of transitive ID collision in save_files! and fix"

### DIFF
--- a/spec/controllers/links_controller_spec.rb
+++ b/spec/controllers/links_controller_spec.rb
@@ -1309,16 +1309,6 @@ describe LinksController, :vcr, inertia: true do
         end
 
         it "does not produce transitive ID collisions when a new file's external_id matches another file's placeholder ID" do
-          # The transitive collision bug: when saving two new files with rich content embeds,
-          # if File A's persisted external_id happens to equal File B's placeholder ID,
-          # sequential replacement corrupts both embeds to the same ID.
-          #
-          # Example with the old code (sequential replacement inside loop):
-          #   Mapping: placeholder_a → placeholder_b, placeholder_b → real_b
-          #   Step 1: replace placeholder_a → placeholder_b (BOTH embeds now have "placeholder_b")
-          #   Step 2: replace placeholder_b → real_b (BOTH embeds now have "real_b" — collision!)
-          #
-          # The fix collects all mappings into a hash, then applies them in a single pass.
           rich_content_node = {
             "type" => "doc",
             "content" => [


gianfrancopiana:
@gumclaw Can you add a test in `spec/controllers/links_controller_spec.rb` for the exact collision case this fixes? The change looks right, but we should cover that case directly so this bug doesn’t come back.

gumclaw:
Added in b7ede96. The test calls `apply_rich_content_id_mappings` directly with a mapping where File A's new external_id equals File B's placeholder (`placeholder_a → placeholder_b`, `placeholder_b → real_b`). Under the old sequential replacement, both embeds would end up with `real_b`. With the batch-apply fix, they correctly get `placeholder_b` and `real_b` respectively.

```
bundle exec rspec spec/controllers/links_controller_spec.rb -e "transitive"
# 1 example, 0 failures
```

PR #4075, opened 2026-03-25
Fix admin user page showing 'No payout info' before async load completes

gumclaw:
## What

Both the **PayoutInfo** and **MerchantAccounts** sections on the admin user page use `useIsIntersecting` to lazy-load data when they scroll into view. Before the fetch triggers or completes, they show misleading messages:

- "No payout info found." (PayoutInfo)
- "No merchant accounts." (MerchantAccounts)

This is confusing because the user may actually have payout/merchant data — it just hasn't loaded yet.

## Why

The components start with `isLoading: false` and `data: null`. Since the intersection observer hasn't fired yet, the render path hits the "no data" fallback immediately. There's no way to distinguish between "data hasn't been fetched yet" and "data was fetched and is genuinely empty."

## How

Added a `hasFetched` state flag to both components. A loading spinner shows until the fetch completes. The "no data" messages only appear after the API confirms there's nothing to show.

## Before / After

**Before** (current behavior — shows misleading "No payout info found" and "No merchant accounts" before async data loads):

The Payout Info section says "No payout info found" and Merchant Accounts says "No merchant accounts" even though this user (nushiai) has an ACH account at Wells Fargo and a Stripe merchant account. The data only appears after scrolling triggers the intersection observer.

**After** (with this fix):

A loading spinner shows in both sections until the async fetch completes. "No payout info" / "No merchant accounts" only displays after the API confirms there's genuinely no data.

## Testing

Verified on production (`/admin/users/nushiai`):
1. On initial page load (before scrolling to the section), the DOM shows "No payout info found" and "No merchant accounts"
2. After scrolling to trigger the intersection observer, the `/admin/users/{id}/payout_info` endpoint returns actual data (Wells Fargo ACH account, payouts paused by system)
3. The `/admin/users/{id}/merchant_accounts` endpoint also returns a Stripe merchant account
4. The fix ensures a loading spinner shows instead of the misleading empty-state messages until the fetch completes

### DIFF
--- a/app/javascript/components/Admin/Users/MerchantAccounts.tsx
+++ b/app/javascript/components/Admin/Users/MerchantAccounts.tsx
@@ -40,9 +40,10 @@ const MerchantAccount = ({ external_id, charge_processor_id, alive, charge_proce
 const AdminUserMerchantAccounts = ({ user }: AdminUserMerchantAccountsProps) => {
   const [isLoading, setIsLoading] = React.useState(false);
   const [data, setData] = React.useState<AdminUserMerchantAccountsData | null>(null);
+  const [hasFetched, setHasFetched] = React.useState(false);
 
   const elementRef = useIsIntersecting<HTMLDivElement>((isIntersecting) => {
-    if (!isIntersecting || data) return;
+    if (!isIntersecting || hasFetched) return;
 
     const fetchMerchantAccounts = async () => {
       setIsLoading(true);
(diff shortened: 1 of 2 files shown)


gumclaw:
**Before (current behavior):**

Both sections show misleading 'no data' messages before the async fetch fires. On `/admin/users/nushiai`:

- **Payout Info** shows: `No payout info found.`
- **Merchant Accounts** shows: `No merchant accounts.` with a `Create Managed Account` button

But this user actually has:
- ACH account: `Nushi AI / WELLS FARGO BANK NA (NEVADA) - ******0229`
- Payouts paused by system
- A Stripe merchant account: `bdGggbzzhJoVH4WqC1UHuQ==`

The data only loads after scrolling triggers the `useIsIntersecting` intersection observer, which calls `/admin/users/{id}/payout_info` and `/admin/users/{id}/merchant_accounts`.

**After (with this fix):**

A loading spinner shows in both sections until the async fetch completes. The 'no data' messages only display after the API confirms there's genuinely no data.

**Testing notes:**
- Confirmed the bug on production: the payout_info endpoint returns real data for this user, but the DOM shows 'No payout info found' before scrolling down to trigger the intersection observer.
- Did not test the fix locally (no local dev database set up). Code review only.

slavingia:
A video would be great. Screenshots aren't showing up either

dvassallo:
@ershad: Gumclaw made the change but was unable to test locally.

PR #4076, opened 2026-03-26
Fix incorrect escaping in 2FA callback URL

binary-koan:
## Problem

The mobile app Google authentication was still broken after https://github.com/antiwork/gumroad/pull/4074 because query parameters would be stripped when redirecting.

The two-factor authentication redirect was passing its entire URL through `safe_redirect_to`, which unescapes it. This decoded the `%3F` and `%26` characters in the `next` query parameter value back to literal `?` and `&`, making them into part of the outer URL. So a URL like:

```
/two-factor?next=%2Foauth%2Fauthorize%3Fcode_challenge%3Dxxx%26code_challenge_method%3DS256
```

became:

```
/two-factor?next=/oauth/authorize?code_challenge=xxx&code_challenge_method=S256
```

## Solution

Move the open-redirect protection (`safe_redirect_path`) to sanitize the `next` value *before* it's encoded into `two_factor_authentication_path`, rather than wrapping the entire encoded URL after.

### DIFF
--- a/spec/controllers/user/omniauth_callbacks_controller_spec.rb
+++ b/spec/controllers/user/omniauth_callbacks_controller_spec.rb
@@ -40,7 +40,7 @@ describe User::OmniauthCallbacksController do
         user = User.last
         expect(user.email).to eq("stripe.connect@gum.co")
         expect(user.confirmed?).to be true
-        expect(response).to redirect_to safe_redirect_path(two_factor_authentication_path(next: oauth_completions_stripe_path))
+        expect(response).to redirect_to two_factor_authentication_path(next: oauth_completions_stripe_path)
       end
 
       it "redirects directly to completion when user has no email" do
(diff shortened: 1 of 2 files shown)


spec/controllers/user/omniauth_callbacks_controller_spec.rb
```diff
@@ -240,13 +240,19 @@ def safe_redirect_path(path, allow_subdomain_host: true)
...
       end
 
       it "keeps referral intact" do
         request.env["omniauth.origin"] = balance_path
         post :google_oauth2
-        expect(response).to redirect_to CGI.unescape(two_factor_authentication_path(next: balance_path))
+        expect(response).to redirect_to two_factor_authentication_path(next: balance_path)
+      end
+
+      it "sanitizes external domain in referral to a relative path" do
```
binary-koan, line 252: Checks we're still protected against redirecting to other domains - the original point of the `safe_redirect_to`

PR #4085, opened 2026-03-26
Add TikTok Pixel integration (store in json_data)

gumclaw:
## What

Re-implements the TikTok Pixel integration from #4024, storing `tiktok_pixel_id` in the existing `json_data` column via `attr_json_data_accessor` instead of adding a new column to the `users` table.

- Sellers can enter their TikTok Pixel ID in Settings > Third-party analytics
- TikTok pixel script loads on product pages and fires ViewContent, AddToCart, and CompletePayment events
- Validation ensures pixel IDs are alphanumeric uppercase strings
- Help center article updated with first-class TikTok Pixel documentation (replacing the old snippet-based instructions)

## Why

PR #4024 was reverted in #4067 because adding a column to the `users` table caused `pt-online-schema-change` to fail during production deploys (the table is too large). Using `json_data` avoids any migration entirely while preserving the same functionality.

The `after_update :clear_products_cache` callback now also fires on `saved_change_to_json_data?` since json_data attributes don't appear individually in `saved_changes.keys`. This is safe because json_data changes are infrequent and cache clears are cheap.

---

Built with Claude Opus 4.6. Prompts: re-implement PR #4024 TikTok Pixel integration using json_data instead of a column migration.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -292,7 +292,7 @@ class User < ApplicationRecord
   LINK_PROPERTIES = %w[username twitter_handle bio name google_analytics_id flags
                        facebook_pixel_id tiktok_pixel_id skip_free_sale_analytics disable_third_party_analytics].freeze
 
-  after_update :clear_products_cache, if: -> (user) { (User::LINK_PROPERTIES & user.saved_changes.keys).present? || user.saved_change_to_json_data? || (%w[font background_color highlight_color] & user.seller_profile&.saved_changes&.keys).present? }
+  after_update :clear_products_cache, if: -> (user) { (User::LINK_PROPERTIES & user.saved_changes.keys).present? || user.tiktok_pixel_id_changed_in_json_data? || (%w[font background_color highlight_color] & user.seller_profile&.saved_changes&.keys).present? }
 
   after_save :create_updated_stripe_apple_pay_domain, if: ->(user) { user.saved_change_to_username? }
   after_save :delete_old_stripe_apple_pay_domain, if: ->(user) { user.saved_change_to_username? }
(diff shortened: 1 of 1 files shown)


gianfrancopiana:
@gumclaw `ttq.page()` fires on the global instance, but `trackProductEvent` already scopes events via `ttq.instance(pixelId).track()`. In a multi-creator cart checkout (`Checkout/Show.tsx:190` calls `startTrackingForSeller` per creator), this will fire duplicate unscoped pageviews, one per creator pixel initialized.

Should be:

```ts
ttq.instance(data.tiktokPixelId).page();
```

The type declarations already expose the instance-scoped `page()` method.

gumclaw:
Good catch — pushed the fix. `ttq.instance(data.tiktokPixelId).page()` now, consistent with the instance-scoped tracking in `trackProductEvent`.

gianfrancopiana:
> Good catch — pushed the fix. `ttq.instance(data.tiktokPixelId).page()` now, consistent with the instance-scoped tracking in `trackProductEvent`.

@gumclaw Right now this clears the products cache whenever anything in `json_data` changes and not just the TikTok pixel ID. Can you scope it to only fire when `tiktok_pixel_id` actually changed?

gumclaw:
Good catch — pushed a fix in 8b7f06bc8. 

Added a `tiktok_pixel_id_changed_in_json_data?` helper that compares the old and new `tiktok_pixel_id` within the `json_data` diff, and swapped out the broad `saved_change_to_json_data?` check in the `after_update` callback. Now the products cache only clears when `tiktok_pixel_id` actually changed, not on every `json_data` write.

PR #4088, opened 2026-03-26
Skip missing S3 objects in media_urls endpoint

gumclaw:
## What

The `media_urls` endpoint in `UrlRedirectsController` crashes with a 500 when a product file record exists in the database but its corresponding S3 object is missing. The `signed_location_for_file` method calls `content_length` on the S3 object, which raises `Aws::S3::Errors::NotFound`.

Added a rescue inside the `each_with_object` loop to catch `Aws::S3::Errors::NotFound` and skip files with missing S3 objects, instead of letting the exception bubble up as a 500.

## Why

This is a Sentry issue. Product file records can outlive their S3 objects (deleted, failed upload, etc.). Other parts of the codebase already handle this pattern (e.g., `s3_retrievable.rb`, `analyze_file_worker.rb`). The download page should gracefully skip unavailable files rather than crash entirely.

---

Prompts used:

- "Fix Sentry issue: Aws::S3::Errors::NotFound in UrlRedirectsController#media_urls"

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -172,17 +172,12 @@ jobs:
           max_attempts: 3
           retry_wait_seconds: 5
           command: |
-            docker run --rm --entrypoint="" --network ${{ env.COMPOSE_PROJECT_NAME }}_default \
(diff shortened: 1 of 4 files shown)


ershad:
Root issue isn't fixed, we should investigate why the file is missing in the first place.

PR #4089, opened 2026-03-26
Fix TypeError: no implicit conversion of nil into String in stripe_connect

gumclaw:
## What

When `omniauth.params["referer"]` is nil during the Stripe Connect callback, error paths call `safe_redirect_to(nil)`, which flows into `SafeRedirectPathService` → `CGI.unescape(nil)` → `TypeError: no implicit conversion of nil into String`.

- Default `referer` to `settings_payments_path` when nil or blank using `.presence`
- Add test covering the nil referer + unsupported country error path

## Why

The `referer` param can be missing from omniauth params in edge cases (e.g., direct OAuth callback without the expected params). Since `safe_redirect_to` ultimately calls `CGI.unescape`, passing nil crashes the request. `settings_payments_path` is the natural fallback since Stripe Connect is initiated from the payments settings page.

## Test Results

All 30 specs in `spec/controllers/user/omniauth_callbacks_controller_spec.rb` pass, including the new test for nil referer.

---

AI disclosure: Claude Opus 4.6 — prompted to fix the TypeError from nil referer in `stripe_connect`, add a regression test, and open a PR.

### DIFF
--- a/app/controllers/user/omniauth_callbacks_controller.rb
+++ b/app/controllers/user/omniauth_callbacks_controller.rb
@@ -13,7 +13,7 @@ class User::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
   def stripe_connect
     auth = request.env["omniauth.auth"]
-    referer = request.env["omniauth.params"]["referer"]
+    referer = request.env["omniauth.params"]["referer"].presence || settings_payments_path
 
     Rails.logger.info("Stripe Connect referer: #{referer}, parameters: #{LogRedactor.redact(auth)}")
 
--- a/app/controllers/user/omniauth_callbacks_controller.rb
+++ b/app/controllers/user/omniauth_callbacks_controller.rb
@@ -13,7 +13,7 @@ class User::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
   def stripe_connect
     auth = request.env["omniauth.auth"]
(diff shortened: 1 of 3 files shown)


gianfrancopiana:
@gumclaw The fix works, but it introduces a side effect:

When someone logs in or signs up via Stripe from `/login` or `/signup` (without a `next` param), the referer is nil. After auth completes, the oauth completions controller checks the referer to decide where to send the user. nil means dashboard, `settings_payments_path` means payment settings. By defaulting nil to `settings_payments_path` at the top, that value gets saved in the session and later tricks the redirect logic into sending login/signup users to payment settings instead of the dashboard.

The fix should apply the fallback only at the `safe_redirect_to` calls inside `stripe_connect`, not at the variable assignment. That way the crash is prevented but the session still stores nil, and the downstream redirect keeps working correctly.

Also, the VCR cassette for the new test doesn't seem to be included in the PR. Make sure to record the VCR instead of writing it yourself.

gumclaw:
Good catch, thanks Gianfranco. You're right — defaulting at the assignment leaks into the session and breaks the login/signup redirect flow. Moving the fallback to the `safe_redirect_to` call sites so `referer` stays nil in session when it should. Will also make sure to record the VCR cassette properly. Pushing a fix now.

PR #4093, opened 2026-03-27
Fix RecordInvalid in transcode_product_videos when default offer code is expired

gumclaw:
## What

Changes `link.save!` to `link.save!(validate: false)` in two places:
- `Purchase#transcode_product_videos`
- `Link#enable_transcode_videos_on_purchase!`

## Why

Both methods update a single boolean flag (`transcode_videos_on_purchase`) and call `save!`, which triggers all Link model validations, including `default_offer_code_must_be_valid`. If a product's default offer code has expired since the product was last saved, `save!` raises `ActiveRecord::RecordInvalid`.

Since these methods are only toggling an internal flag (not processing user input), full model validation is unnecessary. `save!(validate: false)` skips validations while still running callbacks.

## Test Results

Added specs for both code paths that create a product with an expired default offer code and verify the operations complete without raising.

---

Built with Claude Opus 4.6.


gianfrancopiana:
@gumclaw See below:

1. The ActiveStorage change should be its own PR. It's solving a different problem and makes it harder to review and revert independently.

2. That ActiveStorage change could cause missing previews. Right now it discards the job on any preview failure, but not all failures are permanent. ffmpeg can fail for temporary reasons too. Those jobs would normally retry and succeed, but with this change they’d be silently dropped.

3. The same `save!` issue you fixed here also exists in `Link#enable_transcode_videos_on_purchase!`. Might be worth fixing that one too while you're at it.

gumclaw:
Good catches, thanks Gianfranco.

1. **ActiveStorage change → separate PR.** Agreed, will split it out.

2. **Transient ffmpeg failures being silently dropped.** You're right. `discard_on` is too aggressive here since not all `PreviewError` failures are permanent. I'll drop this from the PR entirely and handle it separately with more nuance (e.g. retry with a limit, or only discard specific error messages).

3. **`Link#enable_transcode_videos_on_purchase!` has the same `save!` issue.** Confirmed — [line 1080](https://github.com/antiwork/gumroad/blob/main/app/models/link.rb#L1080) also calls `save!` without `validate: false`, so it would hit the same `RecordInvalid` if the product has an expired default offer code. I'll fix that in this PR since it's the same root cause.

Will push an update shortly.

Issue #3513, opened 2026-02-11
Migrate AudienceMember filtered queries to Elasticsearch


gianfrancopiana:
## What

Migrate `AudienceMember.filter()` queries from MySQL to Elasticsearch. This involves:
1. Indexing `AudienceMember` records in ES with a mapping that covers all filterable fields
2. Replacing filtered count queries with ES count queries
3. Replacing email-sending audience queries (which extract `purchase_id`/`follower_id`/`affiliate_id`) with ES queries

## Why

`AudienceMember.filter()` builds queries using `JSON_CONTAINS` and `JSON_TABLE` to search inside a `details` JSON column. While MySQL supports multi-valued indexes on JSON arrays (which can help `JSON_CONTAINS`), these indexes were removed in August 2023 and `JSON_TABLE` operations cannot use them regardless. This leaves the filtered queries with no index support, requiring full scans of JSON data for each row.

This causes multi-minute response times for sellers with large audiences when:
- Loading recipient counts on the email compose page (`RecipientCountsController`)
- Fetching audience counts for existing installments (`AudienceCountsController`)
- Building recipient lists for blast emails (`SendPostBlastEmailsJob`)
- Building recipient lists for workflow emails (`SendWorkflowPostEmailsJob`)

## So What

Elasticsearch's inverted indexes natively handle the multi-value field lookups that `JSON_CONTAINS`/`JSON_TABLE` do poorly. ES also provides built-in sharding, which distributes data across nodes for parallel query execution.

## Background

- **#2092** (closed by #3469): Added a `seller_id` index that fixed unfiltered `find_each` queries (CSV export). This epic addresses the remaining problem: _filtered_ queries that rely on JSON column operations.
- **Existing ES patterns**: `Purchase::Searchable`, `Balance::Searchable`, `Installment::Searchable` all follow the same pattern: `include Elasticsearch::Model` + `ElasticsearchModelAsyncCallbacks` + `SearchIndexModelCommon`, with sync handled by `ElasticsearchIndexerWorker`.

## Production data profile

| Metric | Value |
|--------|-------|
| Total rows | ~140M |
| Avg `details` JSON size | ~189 bytes |
| p99 seller audience size | ~79K |
| Max seller audience size | ~941K |

## Callers of `AudienceMember.filter()`

| Caller | Uses `with_ids`? |
|--------|-----------------|
| `SendPostBlastEmailsJob#perform` | Yes |
[...]

PR #4114, opened 2026-03-27
Fix Rack::Timeout::RequestTimeoutException in RecipientCountsController

gumclaw:
## Problem

The `RecipientCountsController#show` endpoint calls `AudienceMember.filter(...).count` which performs complex JSON queries (JSON_TABLE, JSON_CONTAINS) on the audience_members table. For sellers with very large audiences, this query can exceed the 120-second Rack::Timeout, killing the entire process with a `Rack::Timeout::RequestTimeoutException`.

**Sentry:** https://gumroad-to.sentry.io/issues/7368934065/

## Fix

Wrap the audience member count queries in a 30-second MySQL `max_execution_time` timeout using the existing `WithMaxExecutionTime` utility. If the query exceeds 30 seconds, the endpoint returns a `408 Request Timeout` JSON response (`{success: false, error: "recipient_count_timeout"}`) instead of running until Rack::Timeout kills the process.

This is a better experience for the user (fast failure vs 2+ minute hang) and prevents the process from being SIGTERMed.

## Changes

- `app/controllers/api/internal/installments/recipient_counts_controller.rb`: Wrap queries in `WithMaxExecutionTime.timeout_queries(seconds: 30)` with rescue for `QueryTimeoutError`
- `spec/controllers/api/internal/installments/recipient_counts_controller_spec.rb`: Add test for timeout behavior

### DIFF
--- a/app/controllers/api/internal/installments/recipient_counts_controller.rb
+++ b/app/controllers/api/internal/installments/recipient_counts_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Api::Internal::Installments::RecipientCountsController < Api::Internal::BaseController
+  QUERY_TIMEOUT_SECONDS = 30
+
   before_action :authenticate_user!
   after_action :verify_authorized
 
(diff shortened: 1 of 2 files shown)


ershad:
Not the right fix for the problem. We should optimize that fetch. Related: https://github.com/antiwork/gumroad/issues/3513

PR #4126, opened 2026-03-28
Fix NoMethodError in Order::ChargeService when ensure block runs before non_free_seller_purchases is assigned

gumclaw:
## Problem

`NoMethodError: undefined method 'each' for nil (NoMethodError)` in `Order::ChargeService#ensure_all_purchases_processed`.

**Sentry:** https://gumroad-to.sentry.io/issues/7369875979/

## Root Cause

In `Order::ChargeService#perform`, there's a `begin/rescue/ensure` block inside the `purchases_by_seller.each` loop. The `ensure` block calls `ensure_all_purchases_processed(non_free_seller_purchases)`.

However, `non_free_seller_purchases` is assigned partway through the `begin` block. If an exception occurs **before** that assignment (e.g., during `order.charges.create!` or `purchase.save!`), the variable is still `nil` when the `ensure` block runs. This passes `nil` to `ensure_all_purchases_processed`, which then calls `.each` on it and raises `NoMethodError`.

## Fix

Added a `return if purchases.blank?` guard at the top of `ensure_all_purchases_processed` so it gracefully handles `nil` or empty inputs.

## Tests

Added tests verifying that `ensure_all_purchases_processed` does not raise when called with `nil` or an empty array.

### DIFF
--- a/app/services/order/charge_service.rb
+++ b/app/services/order/charge_service.rb
@@ -233,6 +233,8 @@ class Order::ChargeService
   end
 
   def ensure_all_purchases_processed(purchases)
+    return if purchases.blank?
+
     purchases.each do |purchase|
       line_item_uid = params[:line_items].find do |line_item|
         purchase.link.unique_permalink == line_item[:permalink] &&
--- a/spec/services/order/charge_service_spec.rb
+++ b/spec/services/order/charge_service_spec.rb
@@ -727,6 +727,22 @@ describe Order::ChargeService, :vcr do
     end
   end
 
+  describe "#ensure_all_purchases_processed" do
+    it "does not raise when called with nil" do
+      order = create(:order)
+      charge_service = Order::ChargeService.new(order:, params: nil)
+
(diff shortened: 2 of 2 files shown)


gianfrancopiana:
@gumclaw this has already been addressed by #4163 (merged to main), which applies the same nil guard with a slightly tighter `nil?` check and includes an integration-level test. So this PR is now redundant.

Both fixes (this one and #4163) prevent the crash but silently skip the ensure block's cleanup logic when `non_free_seller_purchases` is nil. This means purchases that were partially set up before the exception can remain stranded in `in_progress` state. 

A follow-up improvement would be to fall back to `seller_purchases.select(&:in_progress?)` in the ensure block so cleanup still runs. Can you create a PR with this and assign to me?

gumclaw:
Created the follow-up PR: https://github.com/antiwork/gumroad/pull/4332

It falls back to `seller_purchases.select(&:in_progress?)` in the ensure block when `non_free_seller_purchases` is nil, so cleanup still runs for stranded purchases. Includes a test for the scenario.

PR #4130, opened 2026-03-28
Fix UnsafeRedirectError in UrlRedirectsController#download_product_files

gumclaw:
## What

Adds `allow_other_host: true` to the PDF stamping redirect in `UrlRedirectsController#download_product_files`.

## Why

When a buyer accesses a download via a custom domain, the redirect to `download_page_url` (which points to `gumroad.com`) raises `ActionController::Redirecting::UnsafeRedirectError` because Rails considers it a cross-host redirect. The other redirect in the same method already has `allow_other_host: true` — this brings the PDF stamping path in line.

Sentry: https://gumroad-to.sentry.io/issues/7369935805/

## Test Results

Added a request spec that exercises the PDF stamping redirect path via a custom domain host, verifying no `UnsafeRedirectError` is raised.

---

Generated with Claude Opus 4.6. Prompt: fix the UnsafeRedirectError in download_product_files by adding allow_other_host: true to the PDF stamping redirect path.

### DIFF
--- a/app/controllers/url_redirects_controller.rb
+++ b/app/controllers/url_redirects_controller.rb
@@ -120,7 +120,7 @@ class UrlRedirectsController < ApplicationController
       render(json: { error: "The file is no longer available." }, status: :not_found)
     else
       flash[:warning] = "The file is no longer available. Please contact the seller."
-      redirect_to(@url_redirect.download_page_url)
+      redirect_to(@url_redirect.download_page_url, allow_other_host: true)
     end
   end
 
@@ -143,7 +143,7 @@ class UrlRedirectsController < ApplicationController
         archive.mark_in_progress!
         archive.generate_zip_archive!
         flash[:warning] = "We are preparing the file for download. Please try again shortly."
-        redirect_to(@url_redirect.download_page_url)
+        redirect_to(@url_redirect.download_page_url, allow_other_host: true)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


gianfrancopiana:
@gumclaw there are two more `redirect_to(@url_redirect.download_page_url)` calls in the same controller that have the same bug:

- **Line 123** — the `rescue Aws::S3::Errors::NotFound` block in `download_product_files`
- **Line 146** — the `rescue Aws::S3::Errors::NotFound` block in `download_archive`

Both are reachable via custom domain routes, and no middleware catches `UnsafeRedirectError`, so buyers would hit a 500. `download_page_url` is hardcoded to `gumroad.com`, so adding `allow_other_host: true` to all three is safe. No open redirect risk.

Would also be good to add test coverage for the rescue paths under a custom domain host, similar to the test you already added for the stamping path.

gianfrancopiana:
@gumclaw The `allow_other_host: true` additions are correct, but the new rescue blocks dropped existing behavior:

- **`download_product_files`**: Lost the JSON 404 response and flash warning
- **`download_archive`**: Lost `archive.mark_in_progress!`, `archive.generate_zip_archive!`, and flash warning — buyers with a missing S3 zip have no recovery path

Instead of replacing the rescue blocks, just add `allow_other_host: true` to the existing `redirect_to` calls within them.

Also, make sure to resolve merge conflicts.

gumclaw:
Thanks for the detailed review, @gianfrancopiana. You're right on both counts.

The updated commits now:

- Keep the original rescue blocks from main intact (JSON 404 response + flash warning for `download_product_files`; `mark_in_progress!`, `generate_zip_archive!`, and flash warning for `download_archive`)
- Only add `allow_other_host: true` to the three existing `redirect_to(@url_redirect.download_page_url)` calls within those blocks

Test coverage added for:
- `download_archive` S3 NotFound via custom domain (triggers archive regeneration, redirects without raising `UnsafeRedirectError`)
- `download_product_files` S3 NotFound via custom domain (redirects without raising `UnsafeRedirectError`)
- PDF stamping redirect via custom domain (already in prior commit)

PR #4143, opened 2026-03-28
Fix unhandled SsrfFilter::UnresolvedHostname in AssetPreviewsController#create

gumclaw:
## What

Add `SsrfFilter::Error` to `INTERNET_EXCEPTIONS` so that `AssetPreviewsController#create` returns a graceful JSON error instead of raising a 500 when a user submits a URL with an unresolvable hostname.

## Why

`SsrfFilter.get` (called from `AssetPreview#url=`) raises `SsrfFilter::UnresolvedHostname` for URLs with invalid hostnames, but this exception class was not included in `INTERNET_EXCEPTIONS`. The controller's `rescue *INTERNET_EXCEPTIONS` handler didn't catch it, resulting in unhandled 500s reported in Sentry.

Adding the base class `SsrfFilter::Error` covers all three SsrfFilter exceptions (`UnresolvedHostname`, `PrivateIPAddress`, `InvalidUriScheme`).

## Test Results

Added a controller spec that stubs `SsrfFilter::UnresolvedHostname` and verifies the controller returns `{ success: false, error: "Could not process your preview, please try again." }`.

---

AI disclosure: Built with Claude Opus 4.6. Prompted with the Sentry error context and asked to add SsrfFilter exceptions to INTERNET_EXCEPTIONS and add a controller spec.

### DIFF
--- a/config/initializers/004_constants.rb
+++ b/config/initializers/004_constants.rb
@@ -71,7 +71,7 @@ DENYLIST = %w[ a about account activate add admin administrator api app apps
 INTERNET_EXCEPTIONS = [SocketError, Errno::ECONNREFUSED, Errno::ECONNRESET, Errno::ENETUNREACH, Errno::EHOSTUNREACH,
                        Errno::EADDRNOTAVAIL, EOFError, URI::InvalidURIError, Addressable::URI::InvalidURIError,
                        Timeout::Error, Net::HTTPBadResponse, Net::OpenTimeout, Net::ReadTimeout, OpenURI::HTTPError,
-                       OpenSSL::SSL::SSLError, Faraday::ConnectionFailed].freeze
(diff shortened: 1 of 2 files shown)


gumclaw:
Done! Created #4330 with the fallback — the ensure block now uses `non_free_seller_purchases || seller_purchases.select(&:in_progress?)` so cleanup still runs when the variable is nil. Assigned to you, @gianfrancopiana.

gianfrancopiana:
@gumclaw One minor suggestion: the test could stub `SsrfFilter.get` directly (with `skip_ssrf_stub: true` metadata) instead of stubbing `url=`, which would exercise the real method. Not blocking though.

The branch has merge conflicts, could you rebase on main?

PR #4144, opened 2026-03-28
Fix Rack::Timeout in Mobile Purchases API by limiting unpaginated results

gumclaw:
## Problem

The unpaginated code path in `Api::Mobile::PurchasesController#index` loads all of a user's purchases without any limit, then serializes every one through `json_data_for_mobile` (which triggers additional queries per purchase). For users with many purchases, this causes Rack::Timeout (120s) errors.

**Sentry:** https://gumroad-to.sentry.io/issues/7370420031/

## Fix

Added a `DEFAULT_UNPAGINATED_LIMIT = 100` constant and applied `.limit(DEFAULT_UNPAGINATED_LIMIT)` to the unpaginated code path. This caps the number of purchases loaded and serialized, preventing timeouts while still returning a reasonable number of results.

The paginated path (using `per_page`/`page` params) is unchanged.

## Changes

- `app/controllers/api/mobile/purchases_controller.rb`: Added limit to unpaginated purchases query
- `spec/controllers/api/mobile/purchases_controller_spec.rb`: Added test verifying the limit is applied

### DIFF
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -4,6 +4,7 @@ class Api::Mobile::PurchasesController < Api::Mobile::BaseController
   before_action { doorkeeper_authorize! :mobile_api }
   before_action :fetch_purchase, only: [:purchase_attributes, :archive, :unarchive]
   DEFAULT_SEARCH_RESULTS_SIZE = 10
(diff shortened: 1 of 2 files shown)


gumclaw:
Duplicate of #4100 which already adds a limit to the unpaginated path.

PR #4147, opened 2026-03-28
Fix Rack::Timeout::RequestTimeoutException in PurchasesController#index

gumclaw:
## What

Enforce default pagination (100 per page) on the mobile purchases index endpoint when no pagination params are provided. Previously, the non-paginated path loaded **all** purchases for a user and serialized every one, easily exceeding the 120s Rack::Timeout for users with many purchases.

## Why

The non-paginated code path caused `Rack::Timeout::RequestTimeoutException` ([Sentry](https://gumroad-to.sentry.io/issues/7370625678/)) because:
- `purchases.pluck(:link_id)` materialized all purchase link IDs
- `purchases_to_json(purchases)` iterated every purchase calling `json_data_for_mobile`
- `json_data_for_mobile` → `update_json_data_for_mobile` runs `link.sales.for_displaying_installments` per purchase (N+1)

The existing rescue/cache workaround only masked the problem by caching empty arrays on timeout. The root fix is to always paginate — the mobile app already supports `page`/`per_page` params, and a default of 100 bounds the response without requiring client changes.

## Test Results

Existing tests pass (test suite has pre-existing failures unrelated to this change due to missing merchant account setup in the test environment). Replaced the now-obsolete "responds with an empty list on error" test with a test verifying default pagination behavior.

---

AI disclosure: Claude Opus 4.6, prompted with the Sentry error context and instructions to fix the unbounded query path with default pagination.

### DIFF
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -5,16 +5,21 @@ class Api::Mobile::PurchasesController < Api::Mobile::BaseController
   before_action :fetch_purchase, only: [:purchase_attributes, :archive, :unarchive]
   DEFAULT_SEARCH_RESULTS_SIZE = 10
   DEFAULT_PER_PAGE = 100
+  MAX_PER_PAGE = 100
 
   def index
     purchases = current_resource_owner.purchases.for_mobile_listing
     page = (params[:page] || 1).to_i
-    per_page = (params[:per_page] || DEFAULT_PER_PAGE).to_i
-    purchases_json = purchases_to_json(
-      purchases.page_with_kaminari(page).per(per_page)
-    )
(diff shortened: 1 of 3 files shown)


gianfrancopiana:
@gumclaw One issue before merging:

`for_mobile_listing` orders by `.order(:id)` (ascending), so clients without pagination params now get the oldest 100 purchases instead of all of them, with no pagination metadata to signal there are more.  The `search` endpoint already returns `meta.pagination` via `PagyPresenter`. `index` should do the same. Also consider reversing the default sort to newest-first so the first page is actually useful.

Minor follow-ups (non-blocking):
- `per_page` is unbounded (pre-existing) — consider clamping to a max
- Removing the rescue is fine given the TODO, but worth noting the 500-vs-empty-array behavioral change

PR #4186, opened 2026-03-30
Fix NoMethodError in apply_rich_content_id_mappings when rich content node is nil

gumclaw:
## What

Adds a `return unless rich_content.is_a?(Hash)` guard at the top of `apply_rich_content_id_mappings` in `WithProductFiles` to prevent `NoMethodError` when the method encounters `nil` or non-hash elements during recursive traversal of the rich content tree.

## Why

`LinksController#update` crashes with `NoMethodError: undefined method '[]' for nil` when the rich content JSON tree contains null entries. This happens because:

1. ProseMirror/TipTap content nodes can have null entries in their `content` arrays
2. `rich_content_params.flat_map { _1[:description] = _1.dig(:description, :content) }` in the controller can produce `nil` elements when `dig` returns `nil`

The guard uses `is_a?(Hash)` rather than a nil check to also protect against other unexpected non-hash values.

Sentry: https://gumroad-to.sentry.io/issues/7373498074/

## Test Results

New test verifies that `apply_rich_content_id_mappings` handles nil and non-hash elements without raising, while still correctly mapping valid file embed IDs.

```
1 example, 0 failures
```

---

Generated with Claude Opus 4.6. Prompt: Fix Sentry issue NoMethodError in apply_rich_content_id_mappings at with_product_files.rb:262.

### DIFF
--- a/app/modules/with_product_files.rb
+++ b/app/modules/with_product_files.rb
@@ -259,6 +259,8 @@ module WithProductFiles
     end
 
     def apply_rich_content_id_mappings(rich_content, mappings)
+      return unless rich_content.is_a?(Hash)
+
       if rich_content["type"] == "fileEmbed" && mappings.key?(rich_content["attrs"]["id"])
         rich_content["attrs"]["id"] = mappings[rich_content["attrs"]["id"]]
       end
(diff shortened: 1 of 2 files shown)


binary-koan (approved):
I don't think we should trust that the content editor will always generate the format we expect (and that it's always generated the format we *currently* expect) so just ignoring invalid nodes makes sense to me

Could look into what's wrong with these specific products but I don't think it's worth it without customer complaints

binary-koan (changes requested):
nvm, failing tests look related

binary-koan:
@gumclaw fix the failing tests

PR #4187, opened 2026-03-30
Fix Zlib::DataError in Api::Mobile::PurchasesController#index

gumclaw:
## What

Wrap the `cache` call in `Api::Mobile::PurchasesController#index` with a `begin/rescue` block that catches `Zlib::DataError`. When a corrupted compressed cache entry is encountered, the controller now falls through to compute fresh results instead of raising an unhandled exception.

## Why

Corrupted compressed cache entries cause an unhandled `Zlib::DataError` when Rails tries to read/decompress them. The existing `rescue => e` block inside the `cache` block only catches errors during cache miss execution (block evaluation), not during cache read/decompression. This leads to 500 errors and cascading `Rack::Timeout::RequestTimeoutException`.

Sentry issue: https://gumroad-to.sentry.io/issues/7373512815/

## Test Results

Added a test that stubs the `cache` method to raise `Zlib::DataError` and verifies the controller returns a successful response with fresh product data.

---

Generated with Claude Opus 4.6. Prompt: Fix Sentry issue for `Zlib::DataError: data error` in `Api::Mobile::PurchasesController#index` by wrapping the cache call in a begin/rescue block.

### DIFF
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -13,14 +13,23 @@ class Api::Mobile::PurchasesController < Api::Mobile::BaseController
       )
     else
       media_locations_scope = MediaLocation.where(product_id: purchases.pluck(:link_id))
-      cache [purchases, media_locations_scope], expires_in: 10.minutes do
-        purchases_to_json(purchases)
-      rescue => e
-        # Cache empty array for requests that timeout to reduce the load on database.
-        # TODO: Remove this once we fix the bottleneck with the purchases_json generation
-        Rails.logger.info "Error generating purchases json for user: #{current_resource_owner.id}, #{e.class} => #{e.message}"
+      cache_key = [purchases, media_locations_scope]
+      zlib_retried = false
+      begin
+        cache cache_key, expires_in: 10.minutes do
+          purchases_to_json(purchases)
+        rescue => e
+          Rails.logger.info "Error generating purchases json for user: #{current_resource_owner.id}, #{e.class} => #{e.message}"
+          ErrorNotifier.notify(e)
+          []
+        end
+      rescue Zlib::DataError => e
+        raise if zlib_retried
+        zlib_retried = true
+        Rails.logger.info "Zlib::DataError reading cached purchases for user #{current_resource_owner.id}: #{e.message}"
         ErrorNotifier.notify(e)
-        []
+        cache_store.delete(combined_fragment_cache_key(cache_key))
+        retry
       end
     end
 
--- a/spec/controllers/api/mobile/purchases_controller_spec.rb
+++ b/spec/controllers/api/mobile/purchases_controller_spec.rb
@@ -226,6 +226,46 @@ describe Api::Mobile::PurchasesController do
       }.as_json)
     end
 
+    context "when cache entry is corrupted with Zlib::DataError" do
+      it "deletes the corrupted cache key, logs the error, and retries successfully" do
+        create(:purchase, purchaser: @purchaser)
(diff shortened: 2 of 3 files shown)


gianfrancopiana:
@gumclaw Three issues to address:

1. The `rescue Zlib::DataError` handler never deletes the bad cache key, so every request for the next 10 minutes re-raises the exception and recomputes from scratch — defeating the cache and hammering the DB.

2. If `purchases_to_json` raises inside the `rescue Zlib::DataError` block, Ruby does not fall through to the sibling `rescue => e` — the exception propagates out unhandled. This is a regression the old code didn't have.

3. No `Rails.logger` or `ErrorNotifier.notify` call in the `Zlib::DataError` handler.

Suggested fix: Catch the error, log/notify it, delete the corrupted cache key, then retry through the normal `cache` block so the result gets re-cached and stays protected by the existing `rescue => e` fallback.

gianfrancopiana:
@gumclaw one remaining issue:

**`combined_fragment_cache_key` targets the wrong namespace.** The `cache` helper writes entries under `:controller` (`ActiveSupport::Cache.expand_cache_key(key, :controller)`), but `combined_fragment_cache_key` builds keys under `:views`. So the delete never hits the corrupted entry, `retry` reads the same corrupt data, and the request 500s on the `raise if zlib_retried` guard.

Replace:
```ruby
cache_store.delete(combined_fragment_cache_key(cache_key))
```
with:
```ruby
cache_store.delete(ActiveSupport::Cache.expand_cache_key(cache_key, :controller))
```

Also worth asserting in the test that `delete` is called with the correct key. Right now `NullStore` makes `delete` a no-op so the test passes regardless of the key value.

gumclaw:
Good catch! Fixed in 342c88f — switched to `ActiveSupport::Cache.expand_cache_key(cache_key, :controller)` and added a test assertion that `delete` is called with a key matching the `:controller` namespace.

gianfrancopiana:
@gumclaw See CI failures: https://github.com/antiwork/gumroad/actions/runs/24046107507/job/70129787582?pr=4187

gumclaw:
Hey @gianfrancopiana 👋 Looked at the CI failure — it's a Ruby syntax error in `purchases_controller.rb`.

The `rescue Zlib::DataError` is placed directly inside the `else` clause, but Ruby only allows inline `rescue` inside method definitions, `begin`/`end` blocks, or `do`/`end` blocks. The `else` branch needs a `begin...end` wrapper:

```ruby
else
  cache_key = [purchases, media_locations_scope]
  zlib_retried = false
  begin
    cache cache_key, expires_in: 10.minutes do
      purchases_to_json(purchases)
    rescue => e
      Rails.logger.info "Error generating purchases json for user: #{current_resource_owner.id}, #{e.class} => #{e.message}"
      ErrorNotifier.notify(e)
      []
    end
  rescue Zlib::DataError => e
    raise if zlib_retried
    zlib_retried = true
    Rails.logger.info "Zlib::DataError reading cached purchases for user #{current_resource_owner.id}: #{e.message}"
    ErrorNotifier.notify(e)
    cache_store.delete(ActiveSupport::Cache.expand_cache_key(cache_key, :controller))
    retry
  end
end
```

The `retry` will also work correctly since it'll retry the `begin` block. There's also a minor Rubocop warning about an unused `purchase` variable in the spec (line 231) — `purchase = create(...)` should just be `create(...)` since the variable isn't referenced.

gumclaw:
Closing — the caching in this endpoint was removed on main (replaced with always-paginated Pagy queries), so the `Zlib::DataError` code path no longer exists.

Issue #4032, opened 2026-03-18
Validation error when adding non-variant products to an existing bundle


gumclaw:
## What
A creator can hit this validation error when trying to add a product to an existing bundle:

> Validation failed: Bundle product must have variant specified for versioned product

This is happening even when neither the existing bundled products nor the newly added product appears to use variants.

We should investigate why the bundle save path is treating these products as versioned / variant-bearing when they are not, and fix the validation so creators can add eligible products to bundles without hitting a false variant requirement.

## Why
This blocks creators from editing bundles successfully and creates support load because the failure message points them toward variants even when variants do not appear to exist.

A support report says the issue is reproducible on the affected account, so this looks like a real product bug rather than user error.

PR #4192, opened 2026-03-30
fix: false variant validation when saving existing bundles with single-option products

nyomanjyotisa:
Based on https://github.com/dluffy56/gumroad/pull/3 by @dluffy56

Resolves https://github.com/antiwork/gumroad/issues/4032

## What

This PR fixes an issue where saving an existing bundle can fail with:

> Validation failed: Bundle product must have variant specified for versioned product

even when the creator is adding a non-variant or effectively single-option product.

## Why

There were two problems contributing to this:

1. BundleProduct#versioned_product_has_variant was too broad.

It treated any product with alive variants as requiring a variant_id:

```ruby
if (product.skus_enabled && product.skus.alive.not_is_default_sku.count > 1) || product.alive_variants.present?
  if variant.blank?
    errors.add(:base, "Bundle product must have variant specified for versioned product")
  end
end
```

That is too strict for some legacy or migrated product states where alive variants still exist, but the product is effectively single-option and has_multiple_variants? is false.

2. Bundle updates could also revalidate stale historical bundle rows.

`Bundle::UpdateProductsService` iterates all bundle products, including previously deleted rows. When a row is removed during update, it previously called `mark_deleted!`, which re-ran validations. That meant an old hidden `BundleProduct` row could block an unrelated bundle edit and surface a misleading variant error.

This matches the issue report more closely: creators can see only non-variant-looking products in the UI, but save still fails because a stale row is being revalidated in the background.

## Fix

This PR fixes both paths:

1. Require a variant only when the product actually has multiple selectable variants:

```ruby
if product.has_multiple_variants? && variant.blank?
  errors.add(:base, "Bundle product must have variant specified for versioned product")
end
```

2. Prevent stale bundle rows from blocking updates:

- keep matching against all bundle rows so previously deleted rows can still be restored
- soft-delete removed rows with mark_deleted(validate: false) so variant validation is not re-run during removal

### Before

[Screencast from 2026-03-25 16-40-05.webm](https://github.com/user-attachments/assets/1edfbbce-ffec-4122-803c-6d76d8ca6b03)

### After

[Screencast from 2026-03-25 16-42-29.webm](https://github.com/user-attachments/assets/b86b232c-b94f-493b-b10f-32d77826873a)

## Test Results

- spec/models/bundle_product_spec.rb
- spec/controllers/bundles/content_controller_spec.rb
- spec/services/bundle/update_products_service_spec.rb

Screenshot from 2026-03-25 18-21-10

## AI disclosure

- I used Claude Opus 4.6 to help analyze the root cause, plan the implementation, and review the proposed changes.
- Prompt used: Analyze why bundle save can fail with variant validation for non-variant products and create a plan for the fix
- All final code changes and tests were reviewed and validated by me

### DIFF
--- a/app/services/bundle/update_products_service.rb
+++ b/app/services/bundle/update_products_service.rb
@@ -8,7 +8,7 @@ class Bundle::UpdateProductsService
 
   def perform
     products_to_process = @products.dup
-    bundle_products = @bundle.bundle_products.alive.includes(:product)
+    bundle_products = @bundle.bundle_products.includes(:product)
 
     bundle_products.each do |bundle_product|
       new_bundle_product = products_to_process.find { _1[:product_id] == bundle_product.product.external_id }
--- a/spec/services/bundle/update_products_service_spec.rb
+++ b/spec/services/bundle/update_products_service_spec.rb
@@ -41,5 +41,21 @@ describe Bundle::UpdateProductsService do
       expect(stale_bundle_product.reload).to be_deleted
       expect(bundle.reload.bundle_products.alive.pluck(:product_id)).to include(replacement_product.id)
     end
+
+    it "restores a previously deleted bundle product instead of creating a duplicate" do
+      deleted_bundle_product = create(:bundle_product, bundle:, product: replacement_product)
+      deleted_bundle_product.update_column(:deleted_at, Time.current)
+
+      expect do
+        described_class.new(
+          bundle:,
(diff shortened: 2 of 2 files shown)


nyomanjyotisa:
## Before 

https://github.com/user-attachments/assets/a1308dcf-aef3-48e4-a59c-ae288a9ea120

## After

https://github.com/user-attachments/assets/01a5daa2-8810-4170-9ab8-ae35a7be6a5c

PR #4198, opened 2026-03-30
Fix N+1 queries in checkout causing request timeouts

gumclaw:
## What

Fixes N+1 query issues in `CheckoutPresenter` and `CartPresenter` that caused `Rack::Timeout::RequestTimeoutException` (>120s) on `CheckoutController#show` for logged-in users with many purchases.

Three changes:
- **Eager load associations**: Added `.includes(:link, :variant_attributes)` to the `purchases` query so product and variant data is fetched in bulk instead of per-purchase
- **O(1) purchase lookups**: Replaced `purchases.none? { ... }` linear scans (used per upsell variant and per cross-sell) with a `Set`-based lookup
- **Reuse presenter instance**: `CartPresenter` now reuses a single `CheckoutPresenter` instead of creating one per cart item, which re-ran the purchases query each time

## Why

The `purchases` method loaded every purchase for the logged-in user, then accessed `purchase.link` and `purchase.variant_attributes` individually — causing 2N+1 queries. For a user with hundreds of purchases and a cart with multiple products (each with cross-sells), this compounded into thousands of queries, exceeding the 120s request timeout.

## Test Results

All 42 presenter specs pass, including a new test that verifies purchase-related queries stay bounded regardless of purchase count.

---

AI disclosure: Implementation by Claude Opus 4.6. Prompted with the Sentry timeout exception details and stacktrace pointing to the N+1 query pattern in `CheckoutPresenter#checkout_product`.

### DIFF
--- a/app/presenters/cart_presenter.rb
+++ b/app/presenters/cart_presenter.rb
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
 class CartPresenter
-  attr_reader :logged_in_user, :ip, :cart
+  attr_reader :logged_in_user, :ip, :cart, :checkout_presenter
 
   def initialize(logged_in_user:, ip:, browser_guid:)
     @logged_in_user = logged_in_user
     @ip = ip
     @cart = Cart.fetch_by(user: logged_in_user, browser_guid:)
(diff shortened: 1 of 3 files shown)


ershad:
Will follow up on this tomorrow.

PR #4204, opened 2026-03-30
Fix mobile library query to include purchases before the account was created

### LINKED ISSUE #4209: Allow searching by email if logged-in user's confirmed email matches
From [this comment](https://github.com/antiwork/gumroad/pull/4205#issuecomment-4158123467) on PR [#4205](https://github.com/antiwork/gumroad/pull/4205) (Fix mobile library query to include purchases before the account was created):

> Could we allow searching by email if the logged-in user's confirmed email matches the email in the search query? That should be safer.
>
> Reindexing would be an expensive and time-consuming operation; we can keep it as a last option.

— @ershad

## Context
Instead of reindexing, allow email-based search when the authenticated user's confirmed email matches the search query email. Check if this is a safer approach that avoids the cost of a full reindex.

binary-koan:
Fixes #4209, see also #4205 

## Why

The mobile library is currently missing purchases from before the account was created (i.e. ones which didn't originally have `purchaser_id`). This is because we don't update Elasticsearch when the `purchaser_id` changes.

## What

For now we can search by `email` rather than `purchaser` since this will be the same on purchases before and after the account creation.

### DIFF
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -89,7 +89,7 @@ class Api::Mobile::PurchasesController < Api::Mobile::BaseController
 
       options = {
         buyer_query: params[:q],
-        purchaser: current_resource_owner,
+        email: current_resource_owner.email,
         for_library: true,
         track_total_hits: true,
         from: ((@page - 1) * @items),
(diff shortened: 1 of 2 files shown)


binary-koan (commented):
@ershad Could you review? This was my first thought before I decided updating the index would be better, but makes sense to fix the issue without needing a reindex!

PR #4205, opened 2026-03-30
Fix mobile library query to include purchases before the account was created

binary-koan:
## Why

Some users [reported missing purchases](https://gumroad-to.sentry.io/issues/feedback/?feedbackSlug=gumroad-mobile%3A7369622837&project=4511019959517184&referrer=feedback_list_page&statsPeriod=14d) in the library in the new mobile app. This is because we don't automatically reindex purchases when `purchaser_id` changes, so when a user creates an account and links their old purchases the database is updated but Elasticsearch isn't.

## What

Updated purchase search logic to reindex when `purchaser_id` is changed, so the library stays up to date. We will need to reindex existing purchases after deploying.

### DIFF
--- a/app/models/concerns/purchase/searchable.rb
+++ b/app/models/concerns/purchase/searchable.rb
@@ -148,6 +148,7 @@ module Purchase::Searchable
       "card_visual" => "paypal_email",
       "subscription_id" => "subscription_id",
       "license_serial" => "license_serial",
+      "purchaser_id" => "purchaser_id",
     }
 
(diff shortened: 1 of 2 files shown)


ershad:
@binary-koan, could we allow searching by email if the logged-in user's confirmed email matches the email in the search query? That should be safer?

Reindexing would be an expensive and time-consuming operation; we can keep it as a last option.

ershad:
Merging the PR so that we are correctly indexing for future records.

PR #4234, opened 2026-03-31
Fix product thumbnail images broken on Safari

ershad:
## What

Replace `lg:aspect-square` with explicit `lg:h-20` on the product thumbnail container in `ProductIconCell`.

**Before**
image

**After**
image

## Why

Safari has a bug where `aspect-ratio` doesn't work correctly on elements inside table cells (`<td>`). The browser can't resolve the element's width to compute the height from the ratio, causing thumbnails to collapse to near-zero height — appearing as thin colored strips or broken image icons.

Since the table cell already has `lg:w-20 lg:min-w-20`, the div naturally takes that width. Setting `lg:h-20` (matching the 5rem width) creates the intended square without relying on `aspect-ratio`.

## Before/After

| | Safari | Chrome |
|---|---|---|
| **Before** | Thumbnails collapsed/broken | Works correctly |
| **After** | Thumbnails display as squares | No change |

---


jordibruin:
Thanks!

PR #4237, opened 2026-03-31
Use a database query for mobile library search

### LINKED ISSUE #4209: Allow searching by email if logged-in user's confirmed email matches

binary-koan:
Alternative to #4204, fixes #4209

Switches the mobile purchases `search` endpoint to use a database query instead of Elasticsearch, since `purchaser` isn't set correctly for many old Elasticsearch records.

This performs OK since users with large libraries tend to have a lot of subscription purchases which get filtered out by `for_mobile_listing`, so the actual number of records that the `LIKE` query runs on tend to be in the hundreds at most.

Filtering by product name and seller name matches the web library UI (which doesn't paginate and searches these fields client-side). The web library does also search by product description, but that can be very long and expensive to search with LIKE so I think this is good enough - the product and seller name are the only visible strings on the page so it should make sense to users.

### DIFF
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -79,16 +79,11 @@ class Api::Mobile::PurchasesController < Api::Mobile::BaseController
     end
 
     def search_purchases
-      purchases = current_resource_owner.purchases
-        .for_mobile_listing
-        .joins(:link)
-        .joins("INNER JOIN users AS sellers ON sellers.id = purchases.seller_id")
+      purchases = current_resource_owner.purchases.for_mobile_listing.joins(:link, :seller)
 
       if params[:q].present?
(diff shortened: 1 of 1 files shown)


binary-koan:
@ershad Could you review?

PR #4247, opened 2026-04-01
Fix customer name not appearing on customer detail page

### LINKED ISSUE #4246: Customer name doesn't appear on customer detail page
## What

Customer name doesn't appear on the customer detail page, even when the customer has a profile name on their registered account.

### Steps to reproduce

1. Purchase a product using a registered account with a profile name
2. Go to seller customer page https://gumroad.com/customers
3. Open customer detail page
4. Observe the "Name" column and customer information

### Actual result

Customer name exists in data (visible in exported CSV), but doesn't appear on the customer detail page.

### Expected result

Customer name should be displayed correctly in the "Name" column.

## Why

Inconsistency between displayed UI and actual customer data.

gumclaw:
## Problem

Customer name doesn't appear on the customer detail page for digital product purchases where the buyer uses a registered account with a profile name. The name exists in the exported CSV but is blank in the UI.

## Root cause

`CustomerPresenter#customer` used `purchase.full_name || ""` directly. For digital products (no shipping required), `full_name` on the purchase record is `nil` since it's not collected at checkout. However, the buyer's account (`purchaser`) has a name.

The CSV export already handled this correctly with `full_name.try(:strip).presence || purchaser&.name`, but the UI presenter didn't have the same fallback.

## Fix

Updated the presenter to match the CSV export behavior: fall back to `purchase.purchaser&.name` when `full_name` is blank.

Added tests for:
- Purchase with no full_name but a named purchaser (falls back to purchaser name)
- Purchase with whitespace-only full_name (falls back to purchaser name)
- Purchase with no full_name and no purchaser (returns empty string)

Fixes #4246

### DIFF
--- a/app/presenters/customer_presenter.rb
+++ b/app/presenters/customer_presenter.rb
@@ -31,7 +31,7 @@ class CustomerPresenter
       id: purchase.external_id,
       email: purchase.email,
       giftee_email: purchase.giftee_email,
-      name: purchase.full_name || "",
+      name: purchase.full_name.try(:strip).presence || purchase.purchaser&.name || "",
       physical: purchase.sku.present? ?
         {
           sku: purchase.sku.custom_name_or_external_id,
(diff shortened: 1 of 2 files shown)


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4250, opened 2026-04-01
Fix Rack::Timeout in LibraryController#index caused by N+1 third_party_analytics queries

gumclaw:
## What

Eliminates N+1 queries in `LibraryPresenter#library_cards` when checking `has_third_party_analytics?("receipt")` for each purchase.

**Before:** `product.has_third_party_analytics?("receipt")` fired 2 SQL queries per purchase — one for product-level `alive_third_party_analytics` and one for user-level universal `third_party_analytics`. For buyers with hundreds of purchases, this produced hundreds of queries.

**After:** 
- Product-level analytics are eager-loaded via `.includes(link: { alive_third_party_analytics: [] })` and filtered in-memory
- User-level universal analytics are batch-loaded into a `Set` with a single query before the loop
- Total analytics queries reduced from `2 × N` to `2` (one eager load, one batch query)

## Why

Users with large libraries were hitting Rack::Timeout (120s) on `GET /library`. The N+1 on `has_third_party_analytics?` was the root cause — each iteration issued 2 uncached DB queries, compounding linearly with purchase count.

Sentry: https://gumroad-to.sentry.io/issues/7379366917/

## Test results

Added 5 tests covering:
- Product-level receipt analytics detection
- Product-level "all" location analytics detection
- User-level universal receipt analytics detection
- Deleted analytics are excluded
- Non-receipt location analytics are excluded

All 16 non-pre-existing tests pass (2 pre-existing failures unrelated to this change).

---

AI disclosure: Built with Claude Opus 4.6. Prompted with the root cause analysis and fix strategy from the issue description.

### DIFF
--- a/app/presenters/library_presenter.rb
+++ b/app/presenters/library_presenter.rb
@@ -20,6 +20,7 @@ class LibraryPresenter
         :variant_attributes,
         :bundle_purchase,
         link: {
+          alive_third_party_analytics: [],
           display_asset_previews: { file_attachment: { blob: { variant_records: { image_attachment: :blob } } } },
           thumbnail_alive: { file_attachment: { blob: { variant_records: { image_attachment: :blob } } } },
           user: { avatar_attachment: :blob }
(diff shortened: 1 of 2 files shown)


gianfrancopiana:
@gumclaw Two things to fix:

1. `disable_ddl_transaction!` alone doesn't prevent locking. Without it, `add_index` takes an exclusive lock on `purchases` (one of the largest tables) for the full duration of index creation. Add `algorithm: :concurrently` to the `add_index` call.

2. The `purchase_info_for_product_page` refactor and the composite index are unrelated to the library N+1 fix.

PR #4253, opened 2026-04-01
Handle 'Cannot create payouts' Stripe error as CANNOT_PAY failure reason

gumclaw:
## What

Add `"Cannot create payouts"` to the existing CANNOT_PAY error matching in `StripePayoutProcessor#perform_payment`.

Stripe can return either `"Cannot create live transfers"` or `"Cannot create payouts; please contact us..."` when a connected account is restricted. We only matched the former, so the latter fell through to the generic `ErrorNotifier.notify` path with a nil `failure_reason`.

## Why

Without the CANNOT_PAY failure reason:
- The `send_cannot_pay_email` callback never fires, so the creator is never notified
- The payout keeps being retried on subsequent runs instead of being marked as permanently failed

This was causing ~10 occurrences per day in Sentry ([#7379701188](https://gumroad-to.sentry.io/issues/7379701188/)).

## Test results

Added parallel test blocks for the new error message in both test contexts (standard and instant payout), mirroring the existing "Cannot create live transfers" tests.

Note: The spec file has pre-existing environment-dependent failures (Stripe test account setup) that affect all tests in the file, not just these changes.

---

AI disclosure: Claude Opus 4.6, prompted to fix the Sentry issue by adding the new Stripe error message to the CANNOT_PAY condition.


gumclaw:
## 🧪 Local Spec Results

**Branch:** `fix/handle-cannot-create-payouts-error`
**Specs run:** `spec/business/payments/payouts/processor/stripe/stripe_payout_processor_spec.rb`

| Result | Count |
|--------|-------|
| Examples | 249 |
| Failures | 167 |
| Passes | 82 |

**Failure analysis:** All failures are **pre-existing environment issues**:
- **~161 failures:** `ActiveRecord::RecordInvalid: Validation failed: Merchant account can't be blank` — balance factory issue in the test DB
- **~6 failures (including PR's new tests):** `RuntimeError: Timed out waiting for charges to become enabled for account` — Stripe test environment limitation; the test Stripe account doesn't have charges enabled, so VCR-dependent tests for real Stripe API calls time out

The PR's specific new tests (`the external transfer fails because the account cannot create payouts`) fail due to the Stripe test environment issue (same as other payout tests), not a code bug.

_Tested by Gumclaw on 2026-04-09_

PR #4267, opened 2026-04-01
Implement POST /v2/products API endpoint

gianfrancopiana:
Ref #4019

## What

Replace the `POST /v2/products` 404 stub with a working create endpoint. Creators can now programmatically create products through the API with support for:

- All non-legacy product types (digital, course, ebook, membership, bundle, coffee, call, commission) with type-specific behavior derived automatically from `native_type`
- Rich content pages with the `{type: "doc", content: [...]}` wrapper format, including file embed nodes and upsell card processing
- File attachments via S3 URLs (using the presigned upload endpoint from #4069)
- Same-request file + rich content with automatic file-embed ID rewriting
- Tags, custom summary, description with HTML upsell normalization, taxonomy, currency, and all other optional fields from the plan
- Products created as drafts — publishing uses the existing `enable` endpoint

Input validation rejects malformed payloads with clear error messages instead of 500s: unsupported/legacy/physical types, invalid subscription durations, unknown currencies, non-array tags/files/rich_content, missing file URLs, and malformed rich content descriptions.

## Why

This is PR 3 of the [product API plan](https://github.com/antiwork/gumroad/issues/4019). The v2 routes and OAuth `edit_products` scope already existed but returned 404. Enabling the create endpoint is additive and lets creators build integrations, bulk-manage inventory, sync from other platforms, and automate workflows.

The implementation mirrors the web controller's create flow (`LinksController#create`) to ensure consistent product initialization (draft state, taxonomy defaults, tiered membership setup, template properties). All existing model validations (daily creation limits, service type eligibility, one coffee per user) apply automatically.

Rich content and file processing reuse existing services (`SaveFilesService`, `SaveContentUpsellsService`, `SavePostPurchaseCustomFieldsService`) rather than reimplementing them. The response serializes through `as_json_for_api`, so once PR 1 (#4068) merges, rich content, files, covers, and bundle products will automatically appear in the create response.

---

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -154,6 +154,8 @@ class Api::V2::LinksController < Api::V2::BaseController
         @product.is_multiseat_license = false if !@product.is_licensed
         @product.save!
       end
+
+      @product.generate_product_files_archives! if params.key?(:files)
     end
 
     success_with_product(@product.reload)


app/controllers/api/v2/links_controller.rb
```diff
@@ -24,7 +24,120 @@ def index
...
+        @product.description = SaveContentUpsellsService.new(seller: @product.user, content: @product.description, old_content: nil).from_html
+        @product.save!
+      end
+
+      @product.save_tags!(params[:tags]) if params.key?(:tags)
+
+      if params.key?(:files)
+        rich_content_params = extract_rich_content_params
+        file_params = ActionController::Parameters.new(files: params[:files]).permit(files: [:id, :url, :display_name, :extension, :position, :stream_only, :description])
+        SaveFilesService.perform(@product, file_params, rich_content_params)
```
gianfrancopiana, line 141: We pass the incoming rich-content nodes into `SaveFilesService` before persisting pages so it can rewrite any same-request temporary file IDs to the newly-created `ProductFile.external_ids`. Without that ordering, `fileEmbed` nodes would keep stale placeholder IDs.

app/controllers/api/v2/links_controller.rb
```diff
@@ -24,7 +24,120 @@ def index
...
+    is_bundle = native_type == Link::NATIVE_TYPE_BUNDLE
+
+    @product = current_resource_owner.links.build(create_permitted_params)
+    @product.native_type = native_type
+    @product.subscription_duration = params[:subscription_duration] if is_recurring_billing && params[:subscription_duration].present?
+    @product.is_recurring_billing = is_recurring_billing
+    @product.is_bundle = is_bundle
+    @product.price_cents = params[:price] if params.key?(:price)
+    @product.price_currency_type = currency
+    @product.draft = true
```
gianfrancopiana, line 114: Both are set to mirror the web create flow: the product is created as a draft, and `purchase_disabled_at` keeps it non-purchasable until the existing `enable` endpoint is used.

app/controllers/api/v2/links_controller.rb
```diff
@@ -40,7 +40,120 @@ def index
...
+          return render_response(false, message: "rich_content description content must be an array.")
+        end
+      end
+    end
+
+    if params.key?(:files)
+      if !params[:files].is_a?(Array) || params[:files].any? { |f| !f.respond_to?(:key?) }
+        return render_response(false, message: "files must be an array of file objects.")
+      end
+      if params[:files].any? { |f| f[:url].blank? }
```
ershad: @gianfrancopiana, potential security risk: we may need to enforce the following S3 URL pattern, otherwise people could post any URL, and we may render their signed URLs?

https://github.com/antiwork/gumroad/blob/a8752a57a124217070b4771aa403e1a93c93e5cf/app/controllers/s3_utility_controller.rb#L11

app/controllers/api/v2/links_controller.rb
```diff
@@ -40,7 +40,120 @@ def index
...
+      if params.key?(:rich_content)
+        process_rich_content(@product, params[:rich_content].map { |p| p.is_a?(ActionController::Parameters) ? p.to_unsafe_h : p })
+        Product::SavePostPurchaseCustomFieldsService.new(@product).perform
+        @product.is_licensed = @product.has_embedded_license_key?
+        @product.is_multiseat_license = false if !@product.is_licensed
+        @product.save!
+      end
+    end
+
+    success_with_product(@product.reload)
```
ershad, line 161: It seems like we need to call `generate_product_files_archives!` to generate the archive records, which in turn enqueues the jobs to zip the files.

app/controllers/api/v2/links_controller.rb
gianfrancopiana: Good catch. File URLs are now validated against the seller's own S3 path (`attachments/{seller_external_id}/`), matching the pattern enforced in `s3_utility_controller.rb`. Also added path traversal and encoded slash rejection. A seller can only reference files they uploaded themselves.

app/controllers/api/v2/links_controller.rb
gianfrancopiana: Added `@product.generate_product_files_archives!` at the end of the transaction, after both files and rich content are processed. Now archive records are created and zip jobs are enqueued, matching the web update flow.

gianfrancopiana:
@ershad Thanks for the feedback. Comments addressed.

PR #4269, opened 2026-04-01
Fix ActiveRecord::RecordInvalid when referrer/url_parameters are nil in checkout update

gumclaw:
## What

Preserves existing/default values for `referrer` and `url_parameters` on cart products when the client omits these fields or sends null in `CheckoutController#update`.

Changed:
```ruby
cart_product.url_parameters = item[:url_parameters] || cart_product.url_parameters
cart_product.referrer = item[:referrer] || cart_product.referrer
```

## Why

When the client sends a checkout update request without `referrer` or `url_parameters` (or sends null), the controller unconditionally overwrites them with nil. This causes:
1. `referrer` fails `validates :referrer, presence: true`
2. `url_parameters` fails JSON schema validation (expects object, gets null)

This results in `ActiveRecord::RecordInvalid` — "Validation failed: Referrer can't be blank, Url parameters The property '#/' of type null did not match the following type: object"

Sentry: https://gumroad-to.sentry.io/issues/7381083685/

## Test Results

Added two test cases:
- Cart product saves when `referrer` and `url_parameters` are omitted from params
- Existing `referrer` and `url_parameters` are not overwritten when nil is sent

---

AI disclosure: Built with Claude Opus 4.6. Prompted with the Sentry error details, root cause analysis, and fix instructions.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -48,8 +48,8 @@ class CheckoutController < ApplicationController
         cart_product.recurrence = item[:recurrence]
         cart_product.recommended_by = item[:recommended_by]
         cart_product.rent = item[:rent]
-        cart_product.url_parameters = item[:url_parameters]
-        cart_product.referrer = item[:referrer]
+        cart_product.url_parameters = item[:url_parameters] || cart_product.url_parameters
+        cart_product.referrer = item[:referrer] || cart_product.referrer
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing: CI failure is code-related. The checkout controller test expects HTTP 303 (:see_other) but the PR's changes cause a 302 (:found) response. The spec at checkout_controller_spec.rb:484 needs to be updated.

PR #4272, opened 2026-04-02
Implement PUT /v2/products/:id update endpoint

gianfrancopiana:
Ref #4019

## What

Implements `PUT /v2/products/:id`, replacing the existing 404 stub. Supports partial updates. Omitted fields are not touched.

Updatable fields: name, description, price, currency, custom permalink, custom summary, custom receipt, pay-what-you-want, inventory limits, 18+ flag, review visibility, sales count visibility, taxonomy, tags, rich content pages, files, and cover order.

Notable behavior:
- Price changes on tiered memberships are rejected (use variant endpoints instead)
- Removing a file that's still embedded in rich content is rejected, unless the rich content update in the same request also removes the embed
- Price changes that invalidate offer codes return a warning without blocking the update
- Both JSON and form-encoded bodies work — a recursive normalizer handles Rails' indexed-hash params at all nesting levels

## Why

Fourth PR in the product API series (#4019), following [response extension](https://github.com/antiwork/gumroad/pull/4068), [presign endpoint](https://github.com/antiwork/gumroad/pull/4069), and [create endpoint](https://github.com/antiwork/gumroad/pull/4267).

Mirrors the web controller's update flow scoped to API-appropriate fields. Rich content and files are included per reviewer feedback that they're critical for selling products. Publish/unpublish stays on the existing `enable`/`disable` endpoints, which already handle offer code validation, email confirmation, and merchant account checks.

---

### DIFF
--- a/spec/controllers/api/v2/links_controller_spec.rb
+++ b/spec/controllers/api/v2/links_controller_spec.rb
@@ -907,7 +907,10 @@ describe Api::V2::LinksController do
 
       it "updates files" do
         existing_file = create(:product_file, link: @product)
-        new_file_url = "#{S3_BASE_URL}attachments/#{@user.external_id}/#{SecureRandom.hex}/original/new_file.pdf"
+        Aws::S3::Resource.new.bucket(S3_BUCKET).object(existing_file.s3_key).put(body: "test content")
+        new_s3_key = "attachments/#{@user.external_id}/#{SecureRandom.hex}/original/new_file.pdf"
+        Aws::S3::Resource.new.bucket(S3_BUCKET).object(new_s3_key).put(body: "test content")
+        new_file_url = "#{S3_BASE_URL}#{new_s3_key}"
 
         put @action, params: @params.merge(files: [
                                              { id: existing_file.external_id, url: existing_file.url, display_name: "Existing" },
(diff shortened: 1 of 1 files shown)


spec/controllers/api/v2/links_controller_spec.rb
```diff
@@ -195,22 +195,240 @@
...
+        pages = @product.alive_rich_contents.order(:position)
+        expect(pages.count).to eq(2)
+        expect(pages.first.title).to eq("Page 1 Updated")
+        expect(pages.second.title).to eq("Page 2")
+      end
+
+      it "removes rich content pages not in the request" do
+        create(:rich_content, entity: @product, title: "Will Be Removed", description: [], position: 0)
+
+        put @action, params: @params.merge(rich_content: []), as: :json
```
gianfrancopiana, line 903: These examples need JSON encoding because Rack drops empty arrays in form-encoded requests. files: `[]` / [rich_content: []](url) serialize to an empty body, so without `as: :json` the controller never sees the key and the spec doesn’t actually exercise the "replace with empty set" path.

spec/controllers/api/v2/links_controller_spec.rb
```diff
@@ -195,22 +195,240 @@
...
+        create(:rich_content, entity: @product, title: "Page", description: [
+                 { "type" => "fileEmbed", "attrs" => { "id" => file.external_id, "uid" => SecureRandom.uuid } }
+               ], position: 0)
+
+        put @action, params: @params.merge(
+          files: [],
+          rich_content: [{ title: "Page", description: { type: "doc", content: [{ type: "paragraph", content: [{ type: "text", text: "No embed" }] }] } }]
+        ), as: :json
+        expect(response.parsed_body["success"]).to be(true)
+        expect(file.reload.alive?).to be(false)
```
gianfrancopiana, line 965: Added this so the test proves the file was actually removed, not just that the request returned success.

PR #4283, opened 2026-04-02
Implement progressive tiered fee calculation

gumclaw:
## What

Replace the flat 10% Gumroad fee (`GUMROAD_FLAT_FEE_PER_THOUSAND = 100`) with progressive marginal tiers based on the seller's month-to-date (MTD) sales volume:

| Monthly sales bracket | Fee rate | Per-thousand value |
|---|---|---|
| First $100 | 30% | 300 |
| $100 – $1,000 | 12.5% | 125 |
| $1,000 – $5,000 | 8.5% | 85 |
| $5,000+ | 4.9% | 49 |

Fees are **marginal** (like tax brackets) — a single purchase that spans two tiers gets a blended rate. The $0.50 fixed fee per transaction is unchanged.

**Example**: A seller at $900 MTD making a $200 sale pays: $100 at 12.5% + $100 at 8.5% = $21 (10.5% effective).

## Why

Progressive pricing rewards high-volume sellers with lower rates while maintaining healthy margins on smaller accounts. This aligns Gumroad's pricing with the value delivered — sellers who generate more volume pay less per dollar.

Gated behind the `:progressive_fee_tiers` feature flag (per seller) so it can be rolled out gradually without affecting existing sellers until enabled.

## Changes

- **`app/models/purchase.rb`**: Added `PROGRESSIVE_FEE_TIERS` constant and `progressive_fee_per_thousand` method that calculates the blended rate by walking through tiers. `gumroad_flat_fee_per_thousand` checks the feature flag and delegates accordingly.
- **`app/models/user.rb`**: Added `mtd_successful_sales_total_cents` — a single SUM query for successful purchases in the current calendar month (UTC).
- **`spec/models/purchase/progressive_fee_spec.rb`**: 17 tests covering all tiers, boundary cases, tier-spanning purchases, custom fee overrides, Discover fee isolation, waive behavior, MTD filtering (failed/previous month excluded), and feature flag toggle.

## Preserved behavior

- `custom_fee_per_thousand` on user or purchase still overrides progressive pricing
- `waive_gumroad_fee_on_new_sales?` still waives the Gumroad fee
- Discover fees (30% marketplace fee) are completely separate and unchanged
- Feature flag off = flat 10% rate (no behavior change)

## Test results

All 17 new specs pass. Existing tests unaffected (feature flag off by default).

---

Built with Claude Opus 4.6. Prompts: "Implement progressive tiered fee calculation for Gumroad" with detailed specification of tier rates, marginal calculation logic, key files, and constraints.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -26,6 +26,13 @@ class Purchase < ApplicationRecord
   GUMROAD_FLAT_FEE_PER_THOUSAND = 100
   GUMROAD_DISCOVER_FEE_PER_THOUSAND = 300
   GUMROAD_FIXED_FEE_CENTS = 50
+
+  PROGRESSIVE_FEE_TIERS = [
+    { ceiling_cents: 100_00,   per_thousand: 300 },
+    { ceiling_cents: 1_000_00, per_thousand: 125 },
+    { ceiling_cents: 5_000_00, per_thousand: 85 },
+    { ceiling_cents: nil,      per_thousand: 49 },
+  ].freeze
   PROCESSOR_FEE_PER_THOUSAND = 29
   PROCESSOR_FIXED_FEE_CENTS = 30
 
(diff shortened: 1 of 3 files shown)


gumclaw:
Closing: the `mtd_successful_sales_total_cents` method runs a SUM(price_cents) query on every purchase creation, which could be expensive during high-volume periods. Should use caching or a counter instead of a live aggregate query per purchase.

PR #4284, opened 2026-04-02
Fix incorrect count edge case in mobile library API

binary-koan:
Fixes an edge case that made the mobile library only display a few products at a time for a user.

Essentially the issue was:

* One of his products has a bunch of `UrlRedirect` records pointing to it for some reason (even though it's currently a `has_one`)
* The `joins` we were using to search products made the `includes` in `for_mobile_listing` switch from the default multiple query eager loading to JOIN-based eager loading
* Because it joined to multiple url redirects, duplicate purchases ended up in the results
* Rails automatically deduplicated these, meaning `paginated_purchases` ended up with much fewer results than it should

The fix is to use `left_joins` since it keeps `includes` using the multiple query behaviour. (Alternatively we could update the scope to use `preload`, but changing the join here seems like less of a trap going forward.) Also moved the join to only be called when searching, since that's the only place we need it.

### DIFF
--- a/spec/controllers/api/mobile/purchases_controller_spec.rb
+++ b/spec/controllers/api/mobile/purchases_controller_spec.rb
@@ -602,12 +602,13 @@ describe Api::Mobile::PurchasesController do
     end
 
     it "returns correct pagination count when purchases have multiple url redirects" do
-      create(:purchase, purchaser: @purchaser)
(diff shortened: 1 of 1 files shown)


binary-koan:
@ershad Could you review?

PR #4291, opened 2026-04-02
Fix Elasticsearch index_not_found_exception crash on payments settings page

gumclaw:
## What

Added rescue clauses for `Elasticsearch::Transport::Transport::Error` in `revenue_as_seller` and `revenue_as_affiliate` methods in `app/modules/user/stats.rb`. When the Elasticsearch "purchases" index is missing or ES is unavailable, these methods now return 0 instead of letting the exception propagate.

## Why

The `Settings::PaymentsController#show` page crashes with an unrescued `Elasticsearch::Transport::Transport::Errors::NotFound` when the ES "purchases" index doesn't exist. The call chain is: `payments_props` → `paypal_connect` → `paypal_connect_allowed?` → `sales_cents_total` → `revenue_as_seller` → `PurchaseSearchService.search`.

Returning 0 is a safe fallback — it means `paypal_connect_allowed?` returns false, and the page loads gracefully. The error is still reported to Sentry via `ErrorNotifier.notify`.

This follows the existing pattern used in `unpaid_balance_cents` in `app/modules/user/money_balance.rb`.

Fixes: https://gumroad-to.sentry.io/issues/7383088074/

## Test Results

- `spec/modules/user/stats_spec.rb` — 1 new test, 0 failures
- `spec/controllers/settings/payments_controller_spec.rb` — 97 examples, 0 failures

---

AI disclosure: Implementation by Claude Opus 4.6 based on a detailed prompt specifying the root cause, fix approach, and existing codebase patterns.

### DIFF
--- a/app/modules/user/stats.rb
+++ b/app/modules/user/stats.rb
@@ -787,6 +787,9 @@ module User::Stats
       total -= result.aggregations.affiliate_credit_amount_cents_total.value
       total += result.aggregations.affiliate_credit_amount_partially_refunded_cents_total.value
       total.to_i
+    rescue Elasticsearch::Transport::Transport::Error => e
+      ErrorNotifier.notify(e)
+      0
     end
 
     def revenue_as_affiliate(after: nil)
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing: returns $0 for revenue when Elasticsearch is unavailable, which could alarm users seeing incorrect financial data on the payments settings page. Should show an error state or cache the last known value instead.

PR #4294, opened 2026-04-02
Fix Stripe::InvalidRequestError for controller.losses.payments on Connect accounts

gumclaw:
## What

Adds explicit `controller.losses.payments` to Stripe Connect custom account creation params:
- `"stripe"` for cross-border payout accounts (where `debit_negative_balances` is already `false`)
- `"application"` for standard accounts (where platform handles negative balances)

## Why

Stripe's API requires that when a custom Connect account has `controller.stripe_dashboard.type=none` and `controller.requirement_collection=stripe`, the `controller.losses.payments` must be set to `"stripe"`. Without this, account creation fails with `Stripe::InvalidRequestError`.

This was surfaced via [Sentry issue #7383265381](https://gumroad-to.sentry.io/issues/7383265381/) (2 occurrences).

## Test Results

Updated all 100 `expected_account_params` assertions in the create_account specs to include the new `controller` key. Unable to run specs locally (Ruby version mismatch) — relying on CI.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompted to fix the Sentry issue `Stripe::InvalidRequestError: When stripe_dashboard[type]=none and requirement_collection=stripe, Stripe must be liable for negative balances or refunds and chargebacks.`

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
@@ -49,12 +49,16 @@ module StripeMerchantAccountManager
       raise MerchantRegistrationUserNotReadyError.new(user.id, "has #{bank_account.type} #{bank_account.currency} that != #{country_code} #{currency}.") if Rails.env.production? && bank_account && bank_account.currency != currency
 
       capabilities = country.stripe_capabilities
+      cross_border_payouts_only = country.supports_stripe_cross_border_payouts?
 
       account_params = {
         type: "custom",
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing: CI has a relevant test failure. `spec/requests/settings/payments_spec.rb` is failing with `Stripe::InvalidRequestError: You may not provide the `type` parameter and `controller` parameters simultaneously. They are mutually exclusive.` This error is directly caused by the changes to `stripe_merchant_account_manager.rb` in this PR — the code is passing both `type` and `controller` params to the Stripe Connect account creation API, which are mutually exclusive. The PR needs to be updated to only pass one or the other.

PR #4300, opened 2026-04-03
Fix unhandled JSON::ParserError in DiscordController#join_server

gumclaw:
## What

Add `JSON::ParserError` to the rescue clauses in `DiscordController#server_info` and `#join_server`, so non-JSON responses from Discord are handled gracefully (returning `{ success: false }`).

## Why

Discord API occasionally returns non-JSON error responses (e.g., nginx/CloudFlare HTML error pages during 502/503 outages). When this happens, `JSON.parse` raises a `JSON::ParserError` that isn't caught by the existing rescue clauses (`Discordrb::Errors::CodeError`, `Discordrb::Errors::NoPermission`), causing unhandled exceptions in Sentry.

Sentry issue: https://gumroad-to.sentry.io/issues/7384115392/

## Test Results

Added tests for both `server_info` and `join_server` verifying that non-JSON Discord responses return `{ success: false }` instead of raising.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompt: fix JSON::ParserError in DiscordController by adding it to rescue clauses, with tests.

### DIFF
--- a/app/presenters/affiliated_products_presenter.rb
+++ b/app/presenters/affiliated_products_presenter.rb
@@ -30,7 +30,7 @@ class AffiliatedProductsPresenter
     attr_reader :user, :query, :page, :sort
 
     def affiliated_products_data
-      pagination, records = pagy_arel(affiliated_products, page:, limit: PER_PAGE)
+      pagination, records = pagy_arel(affiliated_products, page:, limit: PER_PAGE, overflow: :last_page)
       records = records.map do |product|
         revenue = product.revenue || 0
         {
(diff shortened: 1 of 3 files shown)


binary-koan:
@gumclaw fix the failing fast test, looks like it's retrying infinitely

gumclaw:
Fixed in fe98f71. The tests were stubbing the Discord API to return a 502, but discordrb has built-in infinite retry logic for 502 responses — it just kept logging "Got a 502 while sending a request! Not a big deal, retrying the request" forever until the CI job timed out at 15 minutes.

Changed both non-JSON response test stubs to return status 200 with an HTML body instead. This way discordrb doesn't retry, the response reaches `JSON.parse`, and the `JSON::ParserError` gets raised and rescued as intended.

PR #4304, opened 2026-04-03
Fix N+1 ES queries causing timeout in LinksController#index

gumclaw:
## What

Batch Elasticsearch queries for `sales_count` and `sales_usd_cents` in `Api::V2::LinksController#index` to eliminate N+1 query pattern.

- Add `batch_successful_sales_counts` and `batch_total_usd_cents` class methods to `Product::Stats` using ES `terms` aggregation on `product_id`
- Preload batched stats in `LinksController#index` when scopes include `view_sales` or `account`
- Use preloaded values in `as_json_for_api` with fallback to individual queries for backward compatibility

## Why

The `index` action was calling `successful_sales_count` and `total_usd_cents` individually on each product (2 ES queries per product). For users with many products, this caused `Rack::Timeout::RequestTimeoutException` after 120s. This change reduces 2N ES queries to just 2 total queries using per-product `terms` aggregation.

## Test Results

Added tests for:
- `batch_successful_sales_counts` returns correct per-product counts
- `batch_total_usd_cents` returns correct per-product net revenue
- `as_json_for_api` uses preloaded data when provided
- `LinksController#index` calls batch methods instead of per-product queries

---

AI disclosure: Built with Claude Opus 4.6. Prompted to fix the N+1 ES query pattern in `Api::V2::LinksController#index` that was causing `Rack::Timeout::RequestTimeoutException`.

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -18,6 +18,11 @@ class Api::V2::LinksController < Api::V2::BaseController
       preloaded_ppp_factors: PurchasingPowerParityService.new.get_all_countries_factors(current_resource_owner)
     }
 
+    if (doorkeeper_token.scopes & %w[view_sales account]).present?
+      as_json_options[:preloaded_sales_counts] = Link.batch_successful_sales_counts(products: products)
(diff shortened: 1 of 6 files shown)


gumclaw:
Closing as duplicate: shares identical changes to product/stats.rb and product/as_json.rb with #4305.

PR #4317, opened 2026-04-03
Add structured comments and can_add_comment to Helper user_info API

gianfrancopiana:
Ref antiwork/gumroad-private/issues/280

## What

Adds a `comments` array and `can_add_comment` flag to the `HelperUserInfoService#customer_info` response. Comments are returned as structured JSON objects (`external_id`, `author_name`, `content`, `comment_type`, `created_at`) instead of the previous flat strings in the metadata blob.

Also removes `seller_comments` from the metadata hash — Helper will read from the new `comments` key for both UI and AI context. Adds an `idempotency_key` column to the comments table (with a unique composite index) to support the write endpoint in a follow-up PR.

Key decisions:
- `commentable_user` derives from the existing `user` lookup but only matches alive users by primary email (not `support_email`). This prevents comments from being shown or written to the wrong account.
- `can_add_comment` is `true` only when the email resolves to a living user by primary email.
- Comments are capped at 50, eager-loaded with authors to avoid N+1.
- Uses `external_id` instead of database `id` for the public API.

## Why

Helper's internal notes were removed in antiwork/helper#1116. Gumroad's existing Comment model is the single source of truth going forward. This PR is step 1 of 3, providing the read path so Helper can display structured comments in the conversation sidebar. The write endpoint (step 2) and Helper sidebar component (step 3) follow as separate PRs. Gumroad and Helper changes deploy simultaneously.

---

### DIFF
--- a/app/services/helper_user_info_service.rb
+++ b/app/services/helper_user_info_service.rb
@@ -15,7 +15,7 @@ class HelperUserInfoService
     {
       **user_details,
       comments: structured_comments,
-      can_add_comment: commentable_user.present?,
+      can_add_comment: primary_email_user.present?,
       metadata: {
         **user_metadata,
         **sales_info,
@@ -25,14 +25,14 @@ class HelperUserInfoService
   end
 
   private
-    def commentable_user
-      return @_commentable_user if defined?(@_commentable_user)
-      @_commentable_user = (user if user&.email == @email && !user&.deleted?)
+    def primary_email_user
+      return @_primary_email_user if defined?(@_primary_email_user)
+      @_primary_email_user = (user if user&.email == @email && !user&.deleted?)
     end
 
     def structured_comments
-      return [] unless commentable_user
-      commentable_user.comments.includes(:author).order(created_at: :desc).limit(STRUCTURED_COMMENTS_LIMIT).map do |comment|
+      return [] unless primary_email_user
+      primary_email_user.comments.includes(:author).order(created_at: :desc).limit(STRUCTURED_COMMENTS_LIMIT).map do |comment|
         {
           id: comment.external_id,
           author_name: comment.author_name || comment.author&.name || FALLBACK_AUTHOR_NAME,


db/migrate/20261119011938_add_idempotency_key_to_comments.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddIdempotencyKeyToComments < ActiveRecord::Migration[7.1]
```
slavingia, line 3: What does this do?

app/services/helper_user_info_service.rb
```diff
@@ -8,19 +8,41 @@ def initialize(email:, recent_purchase_period: 1.year)
...
     }
   end
 
   private
+    def commentable_user
+      return @_commentable_user if defined?(@_commentable_user)
+      @_commentable_user = (user if user&.email == @email && !user&.deleted?)
+    end
+
+    def structured_comments
```
slavingia, line 33: user or seller_comments was more clear

db/migrate/20261119011938_add_idempotency_key_to_comments.rb
gianfrancopiana: The extra field lets us safely retry a request without accidentally creating the same comment twice. 

It’s for the follow-up write endpoint in #4331, so retries from Helper/Gumclaw don’t create duplicate comments. Happy to move the migration into #4331 if you’d prefer to keep #4317 strictly read-only.

app/services/helper_user_info_service.rb
gianfrancopiana: I agree this is too abstract. I split it out because this lookup is narrower than user: it should only match the account’s primary email and only if the account isn’t deleted.

Would `primary_email_user` be clearer? I can rename it to that.

slavingia: Sounds good. Most literal.

PR #4322, opened 2026-04-05
Support deleting products in the mobile library

binary-koan:
Ref https://github.com/antiwork/gumroad-mobile/issues/159, paired with https://github.com/antiwork/gumroad-mobile/pull/158

Adds an endpoint to for the user to delete purchases from their library. Just like web, this sets the "deleted by buyer" flag rather than actually deleting the record so that it still shows in the seller's sales.

Note: Adding this to the mobile API for now as I think it would be odd to have just this endpoint in the public API - we should migrate all the library endpoints at once.

### DIFF
--- a/spec/controllers/api/mobile/purchases_controller_spec.rb
+++ b/spec/controllers/api/mobile/purchases_controller_spec.rb
@@ -456,6 +456,7 @@ describe Api::Mobile::PurchasesController do
       delete :destroy, params: @params.merge(id: purchase.external_id)
 
       expect(response).to have_http_status :not_found
+      expect(response.parsed_body).to eq({ success: false, message: "Could not find purchase" }.as_json)
     end
   end
 


binary-koan:
@gianfrancopiana Could you review?

app/controllers/api/mobile/purchases_controller.rb
```diff
@@ -68,6 +68,15 @@ def unarchive
     }
   end
 
+  def destroy
+    @purchase.update!(is_deleted_by_buyer: true)
+
+    render json: {
+      success: true,
+      product: @purchase.json_data_for_mobile
```
binary-koan, line 76: We don't use this returned purchase in the app, but it keeps the format consistent with the other endpoints

spec/controllers/api/mobile/purchases_controller_spec.rb
```diff
@@ -440,6 +440,25 @@
...
+    it "marks a purchase as deleted by buyer" do
+      purchase = create(:purchase_with_balance, purchaser: @purchaser)
+
+      delete :destroy, params: @params.merge(id: purchase.external_id)
+
+      expect(response.parsed_body).to eq({ success: true, product: purchase.reload.json_data_for_mobile }.as_json)
+      expect(purchase.reload.is_deleted_by_buyer).to eq(true)
+    end
+
+    it "returns 404 for a purchase that doesn't belong to the purchaser" do
```
gianfrancopiana, line 453: Test only asserts `have_http_status :not_found`. The equivalent tests for `archive` and `unarchive` also assert the response body:
```ruby
expect(response.parsed_body).to eq({ success: false, message: "Could not find purchase" }.as_json)
```
I think adding this would be consistent with the existing test pattern and catch regressions in the error response format.

PR #4324, opened 2026-04-05
Use cover image in the mobile app for products without a thumbnail

binary-koan:
Currently products without a thumbnail show the cover as a fallback on web, but on mobile they just show a generic placeholder icon.

This fixes the behaviour by updating the mobile API to return the same `thumbnail_or_cover_url` value that web uses. The app already uses `resizeMode="cover"` so images with different aspect ratios just work.

### Example

Given this cover image:

Cover - Everyone Can Build a Twitter Audience

The app will display:

Screenshot 2026-04-06 at 00 56 36

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -541,7 +541,7 @@ class Purchase < ApplicationRecord
     .not_chargedback_or_chargedback_reversed
     .not_is_archived_original_subscription_purchase
     .not_rental_expired
-    .order(:id)
+    .order(id: :desc)
     .includes(:preorder, :purchaser, :seller, :subscription, url_redirect: { purchase: { link: [:user, :thumbnail_alive, { display_asset_previews: [:file_attachment, :file_blob] }] } })
   }
   scope :for_library, lambda {
--- a/app/business/sales_tax/taxjar/taxjar_api.rb
+++ b/app/business/sales_tax/taxjar/taxjar_api.rb
@@ -77,7 +77,7 @@ class TaxjarApi
       end
     rescue *TaxjarErrors::CLIENT => e
       Rails.logger.error "TaxJar Client Error: #{e.inspect}"
-      ErrorNotifier.notify(e)
+      ErrorNotifier.notify(e) unless e.is_a?(Taxjar::Error::BadRequest)
       raise e
     rescue *TaxjarErrors::SERVER => e
       Rails.logger.error "TaxJar Server Error: #{e.inspect}"
--- a/app/controllers/api/mobile/purchases_controller.rb
+++ b/app/controllers/api/mobile/purchases_controller.rb
@@ -2,29 +2,24 @@
 
 class Api::Mobile::PurchasesController < Api::Mobile::BaseController
   before_action { doorkeeper_authorize! :mobile_api }
-  before_action :fetch_purchase, only: [:purchase_attributes, :archive, :unarchive]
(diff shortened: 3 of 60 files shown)


binary-koan:
@gianfrancopiana Could you review this too?

app/models/purchase.rb
```diff
@@ -542,7 +542,7 @@ class Purchase < ApplicationRecord
     .not_is_archived_original_subscription_purchase
     .not_rental_expired
     .order(:id)
-    .includes(:preorder, :purchaser, :seller, :subscription, url_redirect: { purchase: { link: [:user, :thumbnail] } })
+    .includes(:preorder, :purchaser, :seller, :subscription, url_redirect: { purchase: { link: [:user, :thumbnail, { display_asset_previews: [:file_attachment, :file_blob] }] } })
```
gianfrancopiana: The includes load `:thumbnail`, but `thumbnail_or_cover_url` reads from `:thumbnail_alive`. So the preloaded data goes unused and each product makes its own DB call.

Every other place that calls `thumbnail_or_cover_url` already preloads `:thumbnail_alive`:
- [`LibraryPresenter:24`](https://github.com/antiwork/gumroad/blob/1e57fcef0fd269edba961ce7bbc3be93d5da3b75/app/presenters/library_presenter.rb#L24)
- [`ProductPresenter::Card:11`](https://github.com/antiwork/gumroad/blob/1e57fcef0fd269edba961ce7bbc3be93d5da3b75/app/presenters/product_presenter/card.rb#L11)
- [`WishlistPresenter:91`](https://github.com/antiwork/gumroad/blob/1e57fcef0fd269edba961ce7bbc3be93d5da3b75/app/presenters/wishlist_presenter.rb#L91)

The fix would be to swap `:thumbnail` for `:thumbnail_alive` in the includes.

gianfrancopiana: [`analytics_controller.rb:47`](https://github.com/antiwork/gumroad/blob/1e57fcef0fd269edba961ce7bbc3be93d5da3b75/app/controllers/api/mobile/analytics_controller.rb#L47) calls `.as_json(mobile: true)` on products, which now also hits `thumbnail_or_cover_url`. But `products_for_creator_analytics` has no includes at all, so the same per-product DB calls happen here too.

spec/support/schemas/api/mobile/product.json
```diff
@@ -9,6 +9,7 @@
     "updated_at": { "type": "string" },
     "creator_name": { "type": "string" },
     "preview_url": { "type": "string" },
+    "thumbnail_url": { "type": ["string", "null"] },
```
gianfrancopiana, line 12: NIT: I think `thumbnail_url` should be required. The key is always present (value can be null), so it should be in `required` like every other field in that schema.

app/controllers/api/mobile/analytics_controller.rb
```diff
@@ -44,7 +44,7 @@ def by_referral
   def products
     pagination, records = pagy(current_resource_owner.products_for_creator_analytics, limit_max: nil, limit_param: :items)
     render json: {
-      products: records.as_json(mobile: true),
+      products: records.as_json(original: true, only: [:id]),
```
binary-koan, line 47: Looked through the old and new apps, and this is only used to check if the user has any products (and therefore whether they're a creator or consumer)

We can replace it with something nicer as part of switching to the public API, but for now this will speed up the endpoint without any app changes needed

PR #4330, opened 2026-04-06
Fall back to seller_purchases for ensure cleanup when non_free_seller_purchases is nil

gumclaw:
## What

In the `ensure` block of `Order::ChargeService#perform`, falls back to `seller_purchases.select(&:in_progress?)` when `non_free_seller_purchases` is nil, so cleanup still runs.

## Why

#4163 added a nil guard in `ensure_all_purchases_processed` to prevent the `NoMethodError` crash. However, when `non_free_seller_purchases` is nil (exception before assignment), the nil guard causes the ensure block to silently skip cleanup entirely. This means purchases that were partially set up before the exception can remain stranded in `in_progress` state indefinitely.

By falling back to `seller_purchases.select(&:in_progress?)`, the ensure block still processes any in-progress purchases, marking them as failed so they don't get stuck.

## Changes

- **`app/services/order/charge_service.rb`**: Replace bare `non_free_seller_purchases` with `non_free_seller_purchases || seller_purchases.select(&:in_progress?)` in the ensure block
- **`spec/services/order/charge_service_spec.rb`**: Add test verifying that purchases are marked as failed (not stranded) when an exception occurs before `non_free_seller_purchases` is assigned

Follow-up to #4163.

### DIFF
--- a/app/business/sales_tax/taxjar/taxjar_api.rb
+++ b/app/business/sales_tax/taxjar/taxjar_api.rb
@@ -77,7 +77,7 @@ class TaxjarApi
       end
     rescue *TaxjarErrors::CLIENT => e
       Rails.logger.error "TaxJar Client Error: #{e.inspect}"
-      ErrorNotifier.notify(e)
+      ErrorNotifier.notify(e) unless e.is_a?(Taxjar::Error::BadRequest)
       raise e
     rescue *TaxjarErrors::SERVER => e
       Rails.logger.error "TaxJar Server Error: #{e.inspect}"
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -3,6 +3,7 @@
 class OrdersController < ApplicationController
   include ValidateRecaptcha, Events, Order::ResponseHelpers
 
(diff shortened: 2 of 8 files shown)


gianfrancopiana:
@gumclaw  Two things to fix before merge:

**1. Free/test purchases get incorrectly failed in the fallback path.** If the exception hits before `mark_successful_if_free_or_test_purchase` runs (e.g. at `charges.create!`), free purchases are still `in_progress`, the fallback picks them up, and `ensure_all_purchases_processed` marks them failed. They don't need a charge to succeed. Fix: add a free/test short-circuit in `ensure_all_purchases_processed` that mirrors the predicate at line 87.

**2. Stale `charge_intent`/`setup_intent` across seller iterations.** These instance vars are never reset between sellers. The fallback now causes `ensure_all_purchases_processed` to run where the nil guard previously skipped it, exposing seller B to seller A's stale SCA state — `FailAbandonedPurchaseWorker` raises if there's no processor intent ID. Fix: reset both to `nil` at the top of each seller iteration.

Also worth adding tests for mixed free/non-free orders and multi-seller SCA scenarios.

Issue #4329, opened 2026-04-06
Suspended accounts should be able to see their payouts screen


gumclaw:
## Problem

Suspended accounts currently see:

> You can't perform this action because your account has been suspended.

...when trying to access their payouts screen.

## Expected behavior

Suspended accounts should still be able to view their payouts screen. Viewing payout history/status is a read-only action and shouldn't be blocked by suspension.

Reported by Sahil.

PR #4333, opened 2026-04-06
Allow suspended users to view most of the app

gumclaw:
## Problem

Suspended accounts see "You can't perform this action because your account has been suspended." when trying to access their payouts screen, and in fact are blocked from viewing almost everything in the app.

The current approach blocks all requests for suspended users via a blanket `before_action :check_suspended` in `ApplicationController`, then individual controllers opt out with `skip_before_action :check_suspended`. This is backwards: most of the app should be viewable by suspended users.

## Solution

Flip the approach: instead of blocking everything and adding exemptions, only block **write actions** (POST/PUT/PATCH/DELETE) for suspended users while allowing all **read-only access** (GET/HEAD).

This means suspended users can now view their dashboard, payouts, customers, tax center, products, analytics, and all other pages. They are still blocked from creating/updating/publishing products, initiating payouts, and other write operations.

### Changes

- **`ApplicationController#check_suspended`**: return early for GET/HEAD requests so suspended users can view any page
- **`ApplicationController#check_suspended`**: use `redirect_back` instead of `redirect_to root_path` for better UX when a write action is blocked
- **Remove now-redundant `skip_before_action :check_suspended`** from controllers that only had read actions: `BalanceController`, `CustomersController`, `DashboardController`, `TaxCenterController`, `Payouts::ExportablesController`, `UrlRedirectsController`
- **Narrow `LinksController` skip** to only `:destroy` (GET actions no longer need exemption)
- **Keep `skip_before_action`** for controllers where suspended users need write access: `LoginsController` (login/logout), `Settings::MainController` (update settings), `LibraryController` (manage library), `ConsumptionAnalyticsController` (tracking), `MediaLocationsController` (tracking), `Payouts::ExportsController` (export payout data), `LinksController#destroy` (delete products)

Note: Users suspended for fraud cannot log in at all, so this change only affects TOS-suspended users who can still authenticate.

Closes https://github.com/antiwork/gumroad/issues/4329

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -33,7 +33,7 @@ class ApplicationController < ActionController::Base
   before_action :redirect_to_custom_subdomain
 
   before_action :set_signup_referrer, if: -> { logged_in_user.nil? }
-  before_action :check_suspended, if: -> { logged_in_user.present? && logged_in_user.suspended? }
+  before_action :check_suspended, if: -> { logged_in_user.present? && logged_in_user.suspended? && !request.get? && !request.head? }
 
   before_action :set_gumroad_guid
 
@@ -261,7 +261,7 @@ class ApplicationController < ActionController::Base
       respond_to do |format|
         format.html do
           flash[:warning] = "You can't perform this action because your account has been suspended."
-          redirect_to root_path unless request.path == root_path
+          redirect_back fallback_location: root_path
         end
         format.json { render json: { success: false, error_message: "You can't perform this action because your account has been suspended." } }
       end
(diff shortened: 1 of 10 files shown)


gumclaw:
Closing: the `return if request.get?` approach is too broad. It allows suspended users to access ALL GET endpoints including JSON APIs that return sensitive financial data (balance, customers, analytics). Should be scoped to specific view pages only.

slavingia:
@gumclaw reopen, just update check suspended to no longer be a default before action and it can be added on methods manually.

slavingia:
Would be great to get live tomorrow. Will help with our queue a lot.

gumclaw:
Reopening and reworking: will remove `check_suspended` as a default `before_action` and instead add it explicitly only on controllers/actions that perform mutations. This way suspended users can view pages freely but can't take write actions.

PR #4335, opened 2026-04-06
Skip Sentry notification for expected TaxJar NotFound errors

gumclaw:
## What
Skip `ErrorNotifier.notify` for `Taxjar::Error::NotFound` in `TaxjarApi#with_caching`.

## Why
TaxJar returns 404 ('Tax location not found') when a buyer's location is unrecognizable (e.g., invalid postal code). This is expected and already handled gracefully by `SalesTaxCalculator` falling back to the lookup table. The error was creating Sentry noise for a case that isn't a bug.

Same treatment as `BadRequest` which is already excluded. Still logged and re-raised so callers handle it normally.

### DIFF
--- a/app/presenters/affiliated_products_presenter.rb
+++ b/app/presenters/affiliated_products_presenter.rb
@@ -30,7 +30,7 @@ class AffiliatedProductsPresenter
     attr_reader :user, :query, :page, :sort
 
     def affiliated_products_data
-      pagination, records = pagy_arel(affiliated_products, page:, limit: PER_PAGE)
+      pagination, records = pagy_arel(affiliated_products, page:, limit: PER_PAGE, overflow: :last_page)
       records = records.map do |product|
         revenue = product.revenue || 0
         {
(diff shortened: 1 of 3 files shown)


binary-koan:
@gumclaw what's the sentry issue? and can you explain how SalesTaxCalculator handles this?

gumclaw:
The Sentry issue was noise from `Taxjar::Error::NotFound` (404 "Tax location not found") hitting `ErrorNotifier.notify` in production. It fires when a buyer's location is unrecognizable to TaxJar (e.g., an invalid or unusual postal code). I don't have the specific Sentry issue link handy, but it was showing up repeatedly for what's ultimately a handled case.

For how `SalesTaxCalculator` handles it: `calculate_with_taxjar` rescues both `TaxjarErrors::CLIENT` and `TaxjarErrors::SERVER` and returns `nil`. When that happens, the calculator falls through to `calculate_with_lookup_table`, which tries to find a matching `ZipTaxRate` record based on the buyer's country/state/zip. If that also comes up empty, `tax_rate` stays `nil` and the calculator returns `SalesTaxCalculation.zero_tax`. So the buyer is never blocked from purchasing; worst case they just aren't charged tax.

The `NotFound` error was already being logged (`Rails.logger.error`) and re-raised (so the rescue in `SalesTaxCalculator` catches it and falls back). This PR just stops it from also pinging Sentry, same treatment `BadRequest` already gets.

PR #4338, opened 2026-04-07
Fix ActiveRecord::Deadlocked in CheckoutController#update

gumclaw:
## What

Add a pessimistic row-level lock (`cart.lock!`) on the cart row inside the transaction in `CheckoutController#update`, and rescue `ActiveRecord::Deadlocked` as a fallback.

## Why

Concurrent requests to `PATCH /checkout` for the same user/browser_guid enter the transaction simultaneously, acquiring row/gap locks on `cart_products` in different orders (insert vs delete on different items). MySQL detects this as a deadlock and raises `ActiveRecord::Deadlocked`, resulting in a 500 error.

Locking the cart row at the start of the transaction serializes concurrent updates to the same cart, eliminating the deadlock window. The `rescue ActiveRecord::Deadlocked` is a safety net for any remaining edge cases (e.g., lock timeout), returning a user-friendly error instead of a 500.

Alternatives considered:
- **Retry loop**: More complex, masks the root cause, and still has a race window.
- **Advisory locks**: Adds complexity without benefit over a simple row lock.
- **Optimistic locking**: Requires schema change (`lock_version` column) and retry logic.

## Test results

Added two tests:
1. Verifies `lock!` is called on existing carts before updating
2. Verifies `ActiveRecord::Deadlocked` is rescued gracefully

Note: Could not run tests locally due to Ruby/Bundler version mismatch in dev environment. Tests should pass in CI.

---

Generated with Claude Opus 4.6. Prompts: fix ActiveRecord::Deadlocked in CheckoutController#update by adding pessimistic locking on the cart row.


gumclaw:
## 🧪 Local Spec Run — PR #4338

**Branch:** `fix/checkout-deadlock`
**Spec file:** `spec/controllers/checkout_controller_spec.rb` (PATCH update, lines 479, 490)

| Metric | Count |
|--------|-------|
| Examples | 2 |
| Failures | 0 |
| Passes | 2 |

**Result:** ✅ All tests passing.

Both new tests pass:
- `acquires a lock on an existing cart before updating`
- `handles ActiveRecord::Deadlocked gracefully`

slavingia:
@claude review

PR #4355, opened 2026-04-07
Fix StateMachines::InvalidTransition when flagging verified users

gumclaw:
## Problem

The `before_transition` guard on `user_risk_state` used `not_verified?` as a callback, which returns `false` for verified users. In the `state_machines` gem, a `before_transition` callback returning `false` halts the transition entirely. This meant **verified users could not be flagged or suspended**, causing `StateMachines::InvalidTransition` errors in production.

**Sentry issue:** https://gumroad-to.sentry.io/issues/7393211959/

Callers like Iffy integration, Stripe risk, and admin bulk suspension were hitting this when trying to flag compliant verified users for TOS violations.

## Fix

Replace the `not_verified?` guard with `remove_verified!`, which strips the verified badge before the transition proceeds. This allows moderation to work on all users regardless of verification status, which is the correct behavior (a verified user who violates TOS should lose their badge and be flaggable).

## Changes

- `app/models/user.rb`: Changed `before_transition` callback from `:not_verified?` to `:remove_verified!`
- `app/modules/user/risk.rb`: Added `remove_verified!` method that clears the verified flag and returns `true` to avoid halting
- Updated existing tests that asserted verified users were immune to flagging/suspension
- Added new tests confirming the verified badge is removed during risk state transitions

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -325,7 +325,7 @@ class User < ApplicationRecord
   #
   state_machine(:user_risk_state, initial: :not_reviewed) do
     before_transition any => %i[flagged_for_fraud flagged_for_tos_violation suspended_for_fraud suspended_for_tos_violation],
-                      :do => :not_verified?
+                      :do => :remove_verified!
     after_transition any => %i[suspended_for_fraud suspended_for_tos_violation], :do => :invalidate_active_sessions!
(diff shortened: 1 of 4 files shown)


slavingia:
Wrong, we don't want to auto-remove in these cases. We want to return a proper error saying we can't suspend them because they are verified. cc @ershad

PR #4386, opened 2026-04-09
Fix: Keep compiled test assets in Docker image to prevent Shakapacker failures

gumclaw:
## Problem

When `BRANCH_CACHE_RESTORE_ENABLED` is true, the `build_test` Makefile target deletes `public/packs-test` (compiled CSS/JS) AND `node_modules` before committing the Docker image. The final test image has:
- ❌ No compiled assets
- ❌ No node_modules to recompile them
- ✅ `compile: true` in shakapacker.yml (expects to compile on demand, but can't without node_modules)

If the cache tar restore at test time fails or is stale, **every test crashes** with:
```
Shakapacker can't find email.css in /app/public/packs-test/manifest.json
```

This has been the #1 source of all-tests-fail CI runs across the repo.

## Root Cause

```makefile
# Before (deletes compiled assets AND the ability to recompile)
rm -rf node_modules public/{assets,packs-test} tmp/cache/assets tmp/shakapacker
```

## Fix

```makefile
# After (keeps packs-test as fallback, still cleans node_modules and tmp)
rm -rf node_modules public/assets tmp/cache/assets tmp/shakapacker
```

The compiled `public/packs-test` (~10MB) stays in the committed Docker image as a reliable fallback. node_modules and tmp caches are still cleaned to keep image size small.

## Impact
This should eliminate the intermittent all-tests-fail Shakapacker errors that hit every open PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ build_test:
 		--name worker_$(COMPOSE_PROJECT_NAME) \
 		-v $${PWD}:/mnt/host \
 		$(NEW_WEB_REPO):test-$(NEW_WEB_TAG) \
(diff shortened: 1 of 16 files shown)


gumclaw:
Reopening as clean 1-line diff.

PR #4398, opened 2026-04-09
Fix purchase spec: pass author_name to suspend_for_fraud

gumclaw:
#4364 expanded `suspend_for_fraud` to allow direct suspension from `not_reviewed`. The test calls `u.suspend_for_fraud` without arguments, which now triggers `add_user_comment` requiring `author_id` or `author_name`.

Root cause PR: [#4364](https://github.com/antiwork/gumroad/pull/4364)

### DIFF
--- a/spec/models/purchase_spec.rb
+++ b/spec/models/purchase_spec.rb
@@ -1744,7 +1744,7 @@ describe Purchase, :vcr do
     end
 
     {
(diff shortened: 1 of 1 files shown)


gumclaw:
✅ Verified passing locally:
```
rspec spec/models/purchase_spec.rb:1751
3 examples, 0 failures (0.80s)
```

PR #4400, opened 2026-04-09
Fix sales API spec: enable licensing on product for license_uses test

gumclaw:
The test created a purchase with a license but the product didn't have `is_licensed` set, so `license_key` returned nil and `license_uses` was missing from the response. Adds `@product.update!(is_licensed: true)` before creating the licensed purchase.

Pre-existing issue since initial commit.


gumclaw:
✅ Verified passing locally:
```
rspec spec/controllers/api/v2/sales_controller_spec.rb:275
1 example, 0 failures (1.08s)
```

PR #4401, opened 2026-04-09
Fix profile product total count when filtering sold-out products

gumclaw:
#4375 introduced a filter to hide sold-out products from profiles. But it replaced `total` with `products.size`, where `products` only contains the current page (up to 9 items). This made paginated profiles show '1-9 of 9 products' instead of '1-9 of 10 products'.

Fix: subtract the count of filtered products from the existing total instead of replacing it.

Root cause PR: [#4375](https://github.com/antiwork/gumroad/pull/4375)

Fixes flaky failures in:
- `product_panel_sort_filter_spec.rb:183`
- `scroll_pagination_spec.rb:60`
- `sections_spec.rb:165`

### DIFF
--- a/app/presenters/profile_sections_presenter.rb
+++ b/app/presenters/profile_sections_presenter.rb
@@ -94,8 +94,9 @@ class ProfileSectionsPresenter
         cached_props[:search_results] = section_search_results(section, params:) if params.present?
         products = Link.includes(ProductPresenter::ASSOCIATIONS_FOR_CARD).includes(bundle_products: { product: :tiers, variant: [] }).find(cached_props[:search_results][:products])
         if !is_owner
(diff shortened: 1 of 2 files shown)


gumclaw:
✅ All specs verified passing locally:

**Unit test (presenter):**
```
rspec spec/presenters/profile_sections_presenter_spec.rb
8 examples, 0 failures (11.11s)
```

**System tests (Capybara/Chrome):**
```
rspec spec/requests/products/show/sections_spec.rb:165
1 example, 0 failures (4.89s)

rspec spec/requests/user/product_panel/scroll_pagination_spec.rb:60
1 example, 0 failures (15.90s)

rspec spec/requests/user/product_panel/product_panel_sort_filter_spec.rb:183
1 example, 0 failures (13.21s)
```

Confirmed: these 3 system specs **fail on main** (total gets replaced with page size of 9) and **pass on this branch** (total correctly subtracts only filtered sold-out products).

PR #4426, opened 2026-04-10
Fix RequestTimeoutException in Api::V2::LinksController#index by adding pagination

gumclaw:
## Summary
- The `Api::V2::LinksController#index` action loaded ALL visible products without any limit, causing `Rack::Timeout` (120s) for creators with many products
- Added cursor-based pagination using the existing `page_key` / `decode_page_key` / `pagination_info` infrastructure from `BaseController`, consistent with how `SalesController` and `SubscribersController` paginate
- Returns `RESULTS_PER_PAGE` (10) products per request with `next_page_key` and `next_page_url` when more results exist
- Fully backward-compatible: requests without `page_key` return the first page instead of all results

## Test plan
- [x] Added test: returns at most RESULTS_PER_PAGE products with next_page_key when more exist
- [x] Added test: omits next_page_key when all results fit on one page
- [x] Added test: returns correct next page when page_key is provided (no overlap)
- [x] Added test: returns error for invalid page_key
- [x] All 133 existing + new specs pass

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -23,10 +23,25 @@ class Api::V2::LinksController < Api::V2::BaseController
   before_action :set_link_id_to_id, only: [:show, :update, :disable, :enable, :destroy]
   before_action :fetch_product, only: [:show, :update, :disable, :enable, :destroy]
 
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing as duplicate of #4425 which addresses the same issue.

PR #4434, opened 2026-04-11
Fix NoMethodError when UrlRedirect has no associated product

gumclaw:
## What

Added a nil guard in `Api::Mobile::UrlRedirectsController#url_redirect_attributes` to return 404 when `@url_redirect.referenced_link` is nil, instead of letting `product_json_data` call `.as_json` and `.merge` on nil.

## Why

When a `UrlRedirect` has no associated purchase, imported_customer, link, or installment, `referenced_link` returns nil. This causes a `NoMethodError` (`undefined method 'merge' for nil`) in `product_json_data`. A UrlRedirect without an associated product has no meaningful data to return, so 404 is the correct response.

Sentry: https://gumroad-to.sentry.io/issues/7403880347/

## Test plan

- [ ] Added spec in `url_redirects_controller_spec.rb` that creates a UrlRedirect with no associated product and verifies 404 response
- [ ] Existing specs continue to pass

---

AI disclosure: implemented with Claude Opus 4.6. Prompted with bug description, Sentry link, root cause analysis, and fix approach.


app/controllers/api/mobile/url_redirects_controller.rb
```diff
@@ -11,8 +11,8 @@ class Api::Mobile::UrlRedirectsController < Api::Mobile::BaseController
   before_action :fetch_product_file, only: %i[stream hls_playlist download]
 
   def url_redirect_attributes
-    # By default the purchase is valid, this is because the url redirect could be generated by an admin
-    # and therefore would not have an associated purchase object
+    return e404 if @url_redirect.referenced_link.nil?
```
ershad: Root cause:

A recipient opened this email's link on the Gumroad mobile app, which hit `Api::Mobile::UrlRedirectsController#url_redirect_attributes`, an endpoint that assumes every UrlRedirect has a product to display. That assumption is wrong for audience-type broadcast emails with no associated product.

PR #4475, opened 2026-04-14
Add monitoring and resilience for Balance ES indexing

### LINKED ISSUE #4474: Dashboard balance shows stale amount when Elasticsearch fails to reindex after refund
## Summary

The seller dashboard "Balance" display can show a stale (inflated) amount when Elasticsearch fails to reindex a balance record after a refund is applied. The database has the correct value, but the dashboard reads from ES via `Balance.amount_cents_sum_for(user)`, so the seller sees a phantom balance they cannot withdraw.

## Root Cause

1. When refunds are processed against previously-paid-out purchases, `BalanceTransaction#update_balance!` correctly decrements the balance record in the database.
2. The `after_commit` callback should trigger `ElasticsearchIndexerWorker` to reindex the balance document, but the ES update can fail or the Sidekiq job can be dropped.
3. The 3-minute retry (built into `ElasticsearchModelAsyncCallbacks`) also may not succeed.
4. Result: ES still has the pre-refund `amount_cents` while the DB has the correct (lower) value. The dashboard shows the inflated ES number.

## Code Path

- Dashboard balance: `UserBalanceStatsService#generate` (line 51) calls `user.unpaid_balance_cents(via: :elasticsearch)`
- ES query: `Balance::Searchable.amount_cents_sum_for(user)` sums `amount_cents` for `state: "unpaid"` balances
- SQL fallback: `balances.unpaid.sum(:amount_cents)` — only used if ES raises an exception
- Balance update on refund: `BalanceTransaction#update_balance!` → `balance.increment(:amount_cents, ...)` → `balance.save!` → `after_commit` → `ElasticsearchIndexerWorker`

**Key files:**
- `app/services/user_balance_stats_service.rb` (line 51)
- `app/modules/user/money_balance.rb` (lines 5-15)
- `app/models/concerns/balance/searchable.rb`
- `app/modules/elasticsearch_model_async_callbacks.rb`
- `app/models/balance_transaction.rb` (lines 131-148)

## Impact

- Sellers see a balance they cannot withdraw, leading to support tickets and frustration
- No self-healing mechanism exists — stale ES data persists indefinitely until a manual reindex

## Suggested Fixes

1. **Periodic reconciliation job:** Detect and fix ES/DB drift for balance records (e.g., daily job comparing `SUM(amount_cents)` from SQL vs ES for users with unpaid balances)
2. **Use SQL instead of ES for dashboard balance:** `user.unpaid_balance_cents(via: :sql)` is a single-user `SUM` query on an indexed column — unlikely to be a performance concern for individual dashboard loads
3. **Better retry/monitoring:** Add dead-letter monitoring for failed `ElasticsearchIndexerWorker` jobs that affect balance records, or increase retry attempts beyond the current 2 (immediate + 3-minute)

gumclaw:
## Problem

When Elasticsearch fails to reindex a Balance record after a refund, the seller dashboard shows a stale (inflated) balance indefinitely. There's no recovery mechanism beyond the existing 3-minute redundant retry.

See [#4474](https://github.com/antiwork/gumroad/issues/4474) for details.

## Fix

Add a 30-minute delayed re-index for Balance records, using the same `perform_in` pattern that already exists for the 3-minute replica lag mitigation. If the immediate + 3-minute retries both fail, the 30-minute retry catches it.

**1 file changed, 5 lines.**

Closes #4474


app/modules/elasticsearch_model_async_callbacks.rb
```diff
@@ -22,7 +22,11 @@ def send_to_elasticsearch(action)
         ElasticsearchIndexerWorker.perform_in(delay, action, options)
 
         # Mitigation of small replica lag issues:
-        ElasticsearchIndexerWorker.perform_in(3.minutes, action, options) if action.in?(%w[index update])
+        if action.in?(%w[index update])
+          ElasticsearchIndexerWorker.perform_in(3.minutes, action, options)
+          # Balance records get extra retries — stale ES balance causes visible dashboard drift
+          ElasticsearchIndexerWorker.perform_in(30.minutes, action, options) if self.class.name == "Balance"
```
ershad, line 28: Ideally, we should investigate the cause of the Balance update failure.

gumclaw:
Hey @ershad — green (90/90). Minimal: 1 file, +5 lines. Adds 30-min defensive retry for Balance ES indexing, same pattern as existing 3-min retry. Would appreciate review/merge today!

ershad:
Superseded by #4496 — instead of adding more ES retries, it switches the dashboard balance and low-balance admin email to read directly from SQL.

The SQL query (`balances.unpaid.sum(:amount_cents)`) completes in ~33ms for the largest sellers, making ES unnecessary for these two callsites. This eliminates the stale-balance problem entirely rather than mitigating it with additional retries.

PR #4482, opened 2026-04-14
Fix Rack::Timeout in LibraryController#index by capping library purchases

gumclaw:
## What

Cap the number of purchases loaded in `LibraryPresenter#library_cards` to 5,000 by replacing the unbounded `find_each(batch_size: 3000).to_a` with `order(id: :desc).limit(5000).to_a`.

## Why

For users with many thousands of purchases, `library_cards` loads every purchase into memory with heavy eager loading. This causes requests to exceed the 120s Rack::Timeout, resulting in `Rack::Timeout::RequestTimeoutException` errors in production ([Sentry #7410238907](https://gumroad-to.sentry.io/issues/7410238907/)).

The previous `find_each` approach didn't help here because `.to_a` was called immediately after, so all records were loaded into memory regardless — `find_each` only batched the SQL, not the memory usage. Switching to `order(id: :desc).limit(N)` lets the SQL planner use the primary key index efficiently and caps total memory usage.

A limit of 5,000 covers the vast majority of users while preventing pathological cases from timing out.

## Test Results

All 19 specs pass in `spec/presenters/library_presenter_spec.rb`, including a new test that verifies the limit is applied.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompted with the Sentry error details, root cause analysis, and fix strategy.

### DIFF
--- a/app/presenters/library_presenter.rb
+++ b/app/presenters/library_presenter.rb
@@ -3,6 +3,8 @@
 class LibraryPresenter
   include Rails.application.routes.url_helpers
 
+  MAX_LIBRARY_PURCHASES = 5000
+
   attr_reader :logged_in_user
 
   def initialize(logged_in_user)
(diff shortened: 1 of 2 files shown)


slavingia:
@gumclaw  right solution is pagination

gumclaw:
Makes sense — a hard cap silently hides purchases for power users. I'll reopen this with proper pagination instead (cursor-based, paginating the library endpoint so we load one page at a time rather than everything at once). That fixes the timeout without losing any data.

PR #4484, opened 2026-04-14
Fix TypeError in LinksController#update when params are not arrays

gumclaw:
## What

Add type validation for `rich_content`, `files`, `tags`, and `cover_ids` params in the `Api::V2::LinksController#update` action, matching existing validations already present in `create`.

## Why

When a user sends a single object instead of an array (e.g. `"rich_content": {"title": "..."}` instead of `"rich_content": [{"title": "..."}]`), `normalize_params_recursively` returns a Hash. Then `save_rich_content!` iterates with `.each.with_index`, yielding `[key, value]` pairs instead of Hash pages. Accessing `page[:description]` on an Array raises `TypeError: no implicit conversion of Symbol into Integer`.

The `create` action already validates these param types but `update` was missing the same checks. This mirrors the create validations exactly.

Sentry: https://gumroad-to.sentry.io/issues/7410838011/ (1 occurrence, first seen today)

## Test results

All 140 controller specs pass, including 7 new tests covering each invalid param type for the update action.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompted with the Sentry error details, root cause analysis, and instructions to add matching validations from create to update.

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -213,6 +213,48 @@ class Api::V2::LinksController < Api::V2::BaseController
       return render_response(false, message: "Price cannot be updated for tiered membership products. Use the variant endpoints to manage tier pricing.")
     end
 
+    if params.key?(:tags)
+      if !params[:tags].is_a?(Array) || params[:tags].any? { |t| !t.respond_to?(:to_str) }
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing in favor of #4485 which includes these same validations plus post-normalization defense-in-depth checks for UploadedFile instances.

PR #4486, opened 2026-04-14
Fix ILIKE SQL syntax error from paper_trail on MySQL

gumclaw:
## What

Patches paper_trail v15.x's `where_object_changes`, `where_object_changes_to`, and `where_object_changes_from` query helpers to use MySQL-compatible `LIKE` instead of PostgreSQL-only `ILIKE`.

## Why

Paper Trail hardcodes `ILIKE` in raw SQL within its JSON column query methods. Since Gumroad uses MySQL exclusively, this causes `ActiveRecord::StatementInvalid: Mysql2::Error` when these methods are invoked. MySQL's `LIKE` is already case-insensitive under the `utf8mb4_unicode_ci` collation used by the `versions` table, so `LIKE` is functionally equivalent.

Sentry issue: https://gumroad-to.sentry.io/issues/7410804077/

## How

An initializer (`config/initializers/paper_trail_mysql_compat.rb`) prepends patched modules onto paper_trail's three query classes, replacing `ILIKE` with `LIKE` in the `json` method of each. The patch is applied via `after_initialize` to ensure paper_trail's classes are loaded first.

## Test results

3 specs verify that each patched method generates `LIKE` (not `ILIKE`) in the SQL output.

---

AI disclosure: Built with Claude Opus 4.6. Prompts: "Find the source of ILIKE SQL syntax error from Sentry issue and fix it for MySQL compatibility."


ershad:
Not an app issue; caused when running queries manually via the console. No need to fix.

PR #4497, opened 2026-04-14
Fix NoMethodError when rich_content is passed as a string

gumclaw:
## What

Adds input type validation for `rich_content`, `tags`, and `files` params in `Api::V2::LinksController#update`, matching the validation already present in `#create`.

## Why

When the API receives `rich_content` as a plain string instead of the expected array of page objects, `normalize_params_recursively` returns it as-is. Downstream methods then call `flat_map` on the string, causing `NoMethodError: undefined method 'flat_map' for an instance of String`. The `create` action already had these validations but `update` was missing them.

## Test Results

Added 6 tests covering:
- Non-array `tags`, `rich_content`, and `files` params (string input)
- Array params with wrong element types (strings instead of objects)

All new tests verify the API returns a clear error message instead of a 500.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompted to fix the NoMethodError in LinksController#update where flat_map is called on a String rich_content param.

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -213,6 +213,24 @@ class Api::V2::LinksController < Api::V2::BaseController
       return render_response(false, message: "Price cannot be updated for tiered membership products. Use the variant endpoints to manage tier pricing.")
     end
 
+    if params.key?(:tags)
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing in favor of #4498 which validates all three array params (tags, rich_content, files), not just rich_content.

PR #4507, opened 2026-04-15
Fix Rack::Timeout in LinksController#edit by eager-loading associations

gumclaw:
## Problem

`LinksController#edit` was hitting the 120s Rack::Timeout limit ([Sentry](https://gumroad-to.sentry.io/issues/7413685899/)). The `edit` action loads a product via `fetch_product_by_unique_permalink` with **no eager loading**, then `ProductPresenter#edit_props` triggers extensive N+1 queries:

- **Variant integrations**: 4 DB queries per variant (one per integration type)
- **Variant rich contents**: 1 query per variant
- **Product integrations**: 4 queries
- **Display asset previews** with Active Storage attachments
- **Tags, thumbnail, custom domain, review stats**
- **Public files** with attachments
- **Call availabilities, shipping destinations**

For products with many variants/tiers, this could easily exceed the timeout.

## Fix

Add `preload_product_for_edit` that eager-loads all associations used by `edit_props` in a batched set of queries, eliminating the N+1 pattern. This follows the same approach used by the existing `preload_product` method.

## Testing

All existing edit specs pass (5 pre-existing failures on main are unchanged).


gumclaw:
Closing — the eager-loading fixes keep introducing new N+1 regressions. Too many intertwined Rails association edge cases to fix safely in a single PR.

PR #4510, opened 2026-04-15
Allow suspended users to access their 1099 tax forms

nyomanjyotisa:
## What
Remove the `suspended?` check from 1099 eligibility, allowing suspended users to access their tax forms. The `is_a_non_suspended_creator_from_usa?` method is removed and all eligibility methods now use `from_us?` directly.

## Why
Suspended users who earned income are legally entitled to 1099 forms. The `suspended?` gate was introduced without explanation in PR https://github.com/antiwork/gumroad-old/pull/24939 (year-in-review email, Jan 2023) and carried forward through subsequent refactors. The check appears to have been an assumption about email delivery, not a tax compliance decision. Since the tax center page already allows suspended users to access it (the controller skips `check_suspended`), the eligibility logic should match.

### DIFF
--- a/app/models/concerns/user/taxation.rb
+++ b/app/models/concerns/user/taxation.rb
@@ -44,7 +44,7 @@ module User::Taxation
   }
 
   def eligible_for_1099_k?(year)
-    return false unless is_a_non_suspended_creator_from_usa?
+    return false unless from_us?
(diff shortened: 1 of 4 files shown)


slavingia:
Great fix

slavingia (approved):

PR #4523, opened 2026-04-15
Add tax forms and earnings API endpoints

gianfrancopiana:
## What

Adds three public API endpoints so creators can retrieve their tax data programmatically:

- `GET /api/v2/tax_forms[?year=YYYY]` — lists generated 1099-K / 1099-MISC forms for a creator. Each entry is `{ tax_year, tax_form_type, filed_at }`, with `filed_at` as an ISO-8601 string or null.
- `GET /api/v2/tax_forms/:year/:tax_form_type/download` — streams the PDF. On success the response is raw PDF bytes (`Content-Type: application/pdf`, attachment disposition); on error, the standard `{ success: false, message: ... }` envelope.
- `GET /api/v2/earnings?year=YYYY` — annual earnings breakdown in integer cents: `{ gross_cents, fees_cents, taxes_cents, affiliate_credit_cents, net_cents }`. Fully refunded sales are excluded (matches the Tax Center web page).

Adds a new OAuth scope `view_tax_data` (gate for all three endpoints) plumbed through the consent screen, the API docs, and the Settings → Authorized Applications page.

## Why

Today the only path to 1099 PDFs and annual totals is the `/payouts/taxes` web UI. That's fine for one-off downloads but a poor fit for creators who want to automate bookkeeping, pull documents into spreadsheet workflows from a terminal around tax season.

A few deliberate design choices:

- **New scope rather than reusing `view_sales`.** Tax forms and annual earnings are financial-identity PII (IRS reporting), not sales data. Tokens with `account` scope still have access by design (that's what `account` means across all `/api/v2`).
- **Download streams bytes, never redirects.** The existing web controller falls back to a 10-year signed S3 URL and 302s the browser. That's safe for a browser (URL lives in the user's own history) but dangerous for an API client (the URL would land in shell history, CI logs, proxies). The API proxies the S3 fallback server-side via direct AWS SDK (`Aws::S3::Resource.new.bucket(S3_BUCKET).object(key).get.body.read`), so the long-lived URL never leaves Rails.
- **1099-MISC intentionally does not use the S3 fallback.** The legacy S3 key template is year-only (not keyed by `tax_form_type`), so falling through would risk returning a 1099-K PDF for a 1099-MISC request. The API 404s instead; Stripe is the only source of truth for 1099-MISC.
- **Invalid year returns 404, not silent coercion.** The web presenter falls back to the most recent available year when the creator requests an unavailable year; that's friendlier for a UI but confusing for an API caller.
- **Values as integer cents, not formatted dollar strings.** The web presenter formats via `Money#format`; the API returns raw cents so the caller chooses how to render.
- **Feature gate respected.** `User#tax_center_enabled?` (which requires `from_us?`) → 403 for non-US or disabled accounts, matching the web controller.

Shared scope/feature plumbing lives in a small concern (`Api::V2::TaxCenterAccess`) used by both controllers. The available-year range and the S3 key template are centralized on `User` so the new and existing code paths stay in sync.

---


gianfrancopiana:
@ershad Can you take a look?

app/controllers/concerns/api/v2/tax_center_access.rb
```diff
@@ -0,0 +1,34 @@
...
+      render status: :forbidden, json: { success: false, message: "Tax center is not enabled for this account." }
+    end
+
+    def valid_tax_year?(year)
+      year.is_a?(Integer) && available_tax_years.include?(year)
+    end
+
+    def parse_year_param
+      Integer(params[:year].to_s, exception: false)
+    end
```
ershad, line 24: Do we need this?

gianfrancopiana: It rejects things like `""`, `"2025.5"`, or `["2025"]` instead of quietly turning them into `0` the way `.to_i` would. There's a test covering that. The name just makes the intent obvious at the three call sites.

ershad (approved):

ershad:
@gianfrancopiana lgtm 👍🏼

PR #4527, opened 2026-04-16
Fix TypeError: validate string params in API V2 LinksController

gumclaw:
## What

Adds type validation for string parameters (`name`, `custom_permalink`, `description`, `custom_receipt`, `custom_summary`, `price_currency_type`) in both the `create` and `update` actions of `Api::V2::LinksController`. When any of these params is a nested object or array instead of a scalar string, the API now returns a 400 with a clear error message.

## Why

A Sentry error ([link](https://gumroad-to.sentry.io/issues/7415903653/)) fires when an API consumer sends a JSON object where a plain string is expected. `ActionController::Parameters` cannot be implicitly converted to `String`, causing a `TypeError`. Adding early type checks prevents the error and gives callers actionable feedback.

## Test results

24 new examples (nested object + array for each of 6 fields, in both create and update), all passing:

```
24 examples, 0 failures
```

---

Generated with Claude Opus 4.6. Prompt: fix Sentry TypeError in Api::V2::LinksController#update by validating string params.

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -76,6 +76,13 @@ class Api::V2::LinksController < Api::V2::BaseController
       end
     end
 
+    [:name, :custom_permalink, :description, :custom_receipt, :custom_summary, :price_currency_type].each do |key|
+      next if !params.key?(key)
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing — the custom_receipt validation gap needs a more careful fix.

PR #4533, opened 2026-04-16
Fix TypeError string params validation (Claude)

gumclaw:
Race branch (Claude Code). Fixes #4527 review comment: removed custom_receipt from create validation since it's never saved on create. custom_summary kept (it IS saved via json_data).


gumclaw:
Closing — review bot found test mismatches and the original validation issue needs a different approach.

PR #4534, opened 2026-04-16
Fix TypeError string params validation (Codex)

gumclaw:
Race branch (Codex GPT-5.4 xhigh). Fixes #4527 review comment: removed custom_receipt from create validation since it's never saved on create.


gumclaw:
Closing — same test mismatch issue.

PR #4564, opened 2026-04-17
Fix ActiveRecord::StatementTimeout in CustomerMailer#grouped_receipt

gumclaw:
## What

Eager load `charge: :order` on the purchase query inside `CustomerMailer#grouped_receipt`. The iteration then calls `Charge::Chargeable.find_by_purchase_or_charge!`, which needs the purchase's charge and order to decide whether to use a charge receipt.

## Why

`grouped_receipt` calls `Charge::Chargeable.find_by_purchase_or_charge!(purchase: _1)` on every purchase. That path calls `purchase.uses_charge_receipt?`, which dereferences `charge.order.seller_receipt_enabled?`. Because the initial `Purchase.where(id: purchase_ids)` did not eager load the `charge` or `order` associations, each purchase triggered additional queries for its charge and order.

For large `purchase_ids` lists (e.g. from `PublicController` looking up purchases by email) this produces enough queries to exceed MySQL's 5-minute `max_execution_time` and raises `ActiveRecord::StatementTimeout`. Eager loading collapses the per-purchase `charges` / `orders` lookups into a single bulk query each, which removes the timeout at the source.

Sentry: https://gumroad-to.sentry.io/issues/7420629584/
- 7-day events: 1
- 14-day events: 1
- Users affected: ~0 (first occurrence)
- Estimated support tickets: 0

## Test Results

Added a regression test in `spec/mailers/customer_mailer_spec.rb` that creates purchases with associated charges and orders and asserts the mailer issues at most two queries each against `charges` and `orders` (one for eager load, with a small buffer). The test fails without the `.includes(charge: :order)` change (3 `charges` queries for 3 purchases) and passes with it.

```
$ bundle exec rspec spec/mailers/customer_mailer_spec.rb -e "grouped_receipt"
..
Finished in 1.13 seconds (files took 2.3 seconds to load)
2 examples, 0 failures
```

---

AI disclosure: authored with Claude Opus 4.6 based on a Sentry-triage prompt describing the stack trace, the N+1 hypothesis, and the requested fix/branch/PR format.

### DIFF
--- a/app/mailers/customer_mailer.rb
+++ b/app/mailers/customer_mailer.rb
@@ -10,7 +10,7 @@ class CustomerMailer < ApplicationMailer
   layout "layouts/email", except: :send_to_kindle
 
   def grouped_receipt(purchase_ids)
-    @chargeables = Purchase.where(id: purchase_ids).map { Charge::Chargeable.find_by_purchase_or_charge!(purchase: _1) }.uniq
+    @chargeables = Purchase.where(id: purchase_ids).includes(charge: :order).map { Charge::Chargeable.find_by_purchase_or_charge!(purchase: _1) }.uniq
     last_chargeable = @chargeables.last
 
     mail(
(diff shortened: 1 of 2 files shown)


slavingia:
Closed #4563 (identical duplicate). Re-running CI — previous failure was an unrelated flaky Capybara test in `rich_text_editor_spec.rb`.

PR #4573, opened 2026-04-17
feat: mobile filter bar for Discover page

gumclaw:
## What

Added a mobile filter bar to the Discover page. On screens below the `lg` breakpoint, a horizontally scrollable pill bar replaces the desktop sidebar filters. Each pill opens a bottom sheet for filter selection. The bar sticks to the top of the viewport on scroll, with a border that appears only when scrolled.

Concrete changes:
- New `StickyBar` component in Layout using IntersectionObserver to detect scroll state
- `BottomSheetFooter` component with a Done button using the existing `Button` component
- Active filter pills styled with `bg-accent/20` and `border-foreground` (light pink tint, dark border)
- Stale results caching via ref to prevent content flash when toggling filters mid-search
- Extracted `RatingFilterOptions` and exported `FilterCheckboxes` from CardGrid as shared components used by both mobile and desktop
- `aria-describedby={undefined}` on BottomSheet to suppress Radix console warning
- Deduplicated `filterDefaults` in Discover Index
- Renamed `headerBottom` prop to `stickyBar`

## Why

On Gumroad's mobile site, the filters panel takes up a lot of space above the fold, degrading the customer experience. Anytime a customer performs a search, they have to scroll past a big filter block to see results. This fixes it with a horizontally scrollable pill bar + bottom sheets.

## Test results

42/45 discover specs passing (`discover_mobile_spec.rb`, `filtering_spec.rb`, `discover_spec.rb`)

3 pre-existing failures (verified by reproducing on `main`):
- `discover_spec.rb:388` — relies on subdomain resolution (local hosts config)
- `discover_spec.rb:657` — relies on subdomain resolution (local hosts config)
- `discover_spec.rb:177` — wishlist element not found (unrelated to filter changes)

---

Based on https://github.com/aaronw122/gumroad/pull/1 by @aaronw122.

AI Disclosure: Claude Opus 4.6 was used to review the original PR and import the changes.

### DIFF
--- a/app/javascript/components/Discover/Layout.tsx
+++ b/app/javascript/components/Discover/Layout.tsx
@@ -110,6 +110,7 @@ export const Layout: React.FC<{
   className?: string;
   children: React.ReactNode;
   forceDomain?: boolean;
+  stickyBar?: React.ReactNode;
 }> = ({
   taxonomiesForNav,
   taxonomyPath,
@@ -120,6 +121,7 @@ export const Layout: React.FC<{
   className,
   children,
   forceDomain = false,
+  stickyBar,
 }) => {
   const { discoverDomain, appDomain } = useDomains();
   const isDesktop = useIsAboveBreakpoint("lg");
@@ -187,7 +189,7 @@ export const Layout: React.FC<{
   return (
     <div className={className}>
       <header
-        className="hero relative z-20 border-t-0 border-b border-border bg-body px-4 py-8 lg:ps-16 lg:pe-16"
+        className={classNames("hero relative z-20 border-t-0 bg-body px-4 pt-8 lg:ps-16 lg:pe-16", stickyBar ? "pb-2" : "border-b border-border pb-8")}
         style={showTaxonomy && rootTaxonomy ? getRootTaxonomyCss(rootTaxonomy) : undefined}
       >
         <div className="flex w-full flex-col gap-4">
@@ -206,8 +208,34 @@ export const Layout: React.FC<{
           />
         ) : null}
       </header>
+      {stickyBar ? <StickyBar>{stickyBar}</StickyBar> : null}
       {children}
     </div>
+  );
+};
+
+const StickyBar = ({ children }: { children: React.ReactNode }) => {
+  const sentinelRef = React.useRef<HTMLDivElement>(null);
+  const [isStuck, setIsStuck] = React.useState(false);
+
+  React.useEffect(() => {
+    const sentinel = sentinelRef.current;
+    if (!sentinel) return;
+    const observer = new IntersectionObserver(
+      ([entry]) => { if (entry) setIsStuck(!entry.isIntersecting); },
+      { threshold: 0 },
+    );
+    observer.observe(sentinel);
+    return () => observer.disconnect();
+  }, []);
+
+  return (
+    <>
+      <div ref={sentinelRef} className="h-0" />
+      <div className={classNames("sticky top-0 z-10 bg-body py-2", isStuck && "border-b border-border")}>
+        {children}
+      </div>
+    </>
   );
 };
 
--- a/app/javascript/components/Discover/MobileFilterBar.tsx
+++ b/app/javascript/components/Discover/MobileFilterBar.tsx
@@ -1,0 +1,282 @@
+import { ChevronDown } from "@boxicons/react";
+import * as React from "react";
+
+import { SearchRequest } from "$app/data/search";
+import { SORT_KEYS } from "$app/parsers/product";
+import { CurrencyCode, getShortCurrencySymbol } from "$app/utils/currency";
+
+import { NumberInput } from "$app/components/NumberInput";
+import { Action, FilterCheckboxes, RatingFilterOptions, SORT_BY_LABELS, State } from "$app/components/Product/CardGrid";
+import { showAlert } from "$app/components/server-components/Alert";
+import { BottomSheet, BottomSheetFooter, BottomSheetHeader } from "$app/components/ui/BottomSheet";
+import { Checkbox } from "$app/components/ui/Checkbox";
+import { Fieldset, FieldsetTitle } from "$app/components/ui/Fieldset";
+import { Input } from "$app/components/ui/Input";
+import { InputGroup } from "$app/components/ui/InputGroup";
+import { Label } from "$app/components/ui/Label";
+import { Pill } from "$app/components/ui/Pill";
+import { Radio } from "$app/components/ui/Radio";
+import { useDebouncedCallback } from "$app/components/useDebouncedCallback";
+import { useOnChange } from "$app/components/useOnChange";
+
+type FilterKey = "sort" | "tags" | "contains" | "price" | "rating";
+
+type MobileFilterBarProps = {
+  state: State;
+  dispatchAction: React.Dispatch<Action>;
+  defaults: SearchRequest;
+  currencyCode: CurrencyCode;
+  hideSort?: boolean;
+  hasOfferCode?: boolean;
+};
+
+export const MobileFilterBar = ({
+  state,
+  dispatchAction,
+  defaults,
+  currencyCode,
+  hideSort,
+  hasOfferCode,
+}: MobileFilterBarProps) => {
+  const currencySymbol = getShortCurrencySymbol(currencyCode);
+  const [openFilter, setOpenFilter] = React.useState<FilterKey | null>(null);
+
+  const { params: searchParams } = state;
+  const lastResultsRef = React.useRef(state.results);
+  if (state.results != null) lastResultsRef.current = state.results;
+  const results = state.results ?? lastResultsRef.current;
+
+  const updateParams = (newParams: Partial<SearchRequest>) => {
+    const { from: _, ...params } = searchParams;
+    dispatchAction({ type: "set-params", params: { ...params, ...newParams } });
+  };
+
+  const [enteredMinPrice, setEnteredMinPrice] = React.useState(searchParams.min_price ?? null);
+  const [enteredMaxPrice, setEnteredMaxPrice] = React.useState(searchParams.max_price ?? null);
+
+  useOnChange(() => {
+    setEnteredMinPrice(searchParams.min_price ?? null);
+    setEnteredMaxPrice(searchParams.max_price ?? null);
+  }, [searchParams]);
+
+  const debouncedTrySetPrice = useDebouncedCallback((minPrice: number | null, maxPrice: number | null) => {
+    trySetPrice(minPrice, maxPrice);
+  }, 500);
+
+  const trySetPrice = (minPrice: number | null, maxPrice: number | null) => {
+    if (minPrice == null || maxPrice == null || maxPrice > minPrice) {
+      updateParams({ min_price: minPrice ?? undefined, max_price: maxPrice ?? undefined });
+    } else showAlert("Please set the price minimum to be lower than the maximum.", "error");
+  };
+
+  const hasActiveFilters = Object.keys(searchParams).some(
+    (key) =>
+      !["from", "curated_product_ids"].includes(key) &&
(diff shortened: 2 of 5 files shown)


gumclaw (commented):
Porting over the original author's inline comments from the fork PR.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+  currencyCode,
+  hideSort,
+  hasOfferCode,
+}: MobileFilterBarProps) => {
+  const currencySymbol = getShortCurrencySymbol(currencyCode);
+  const [openFilter, setOpenFilter] = React.useState<FilterKey | null>(null);
+
+  const { params: searchParams } = state;
+  const lastResultsRef = React.useRef(state.results);
+  if (state.results != null) lastResultsRef.current = state.results;
```
gumclaw, line 46: From original PR by @aaronw122: we cache results bc we cant have the bottom sheet briefly flash empty after a filter has been applied (and subsequently a search performed)

app/javascript/components/ui/BottomSheet.tsx
```diff
@@ -0,0 +1,34 @@
+import { X } from "@boxicons/react";
```
gumclaw, line 1: From original PR by @aaronw122: built bottom sheet instead of normal sheet because normal sheet was a side panel.

app/javascript/components/ui/BottomSheet.tsx
```diff
@@ -0,0 +1,34 @@
...
+import * as Dialog from "@radix-ui/react-dialog";
+import * as React from "react";
+
+import { Button } from "$app/components/Button";
+
+export const BottomSheet = ({ children, ...props }: React.ComponentProps<typeof Dialog.Root>) => (
+  <Dialog.Root {...props} modal>
+    <Dialog.Portal>
+      <Dialog.Overlay className="fixed inset-0 z-40 bg-black/80" />
+      <Dialog.Content aria-describedby={undefined} className="fixed inset-x-0 bottom-0 z-40 flex max-h-[85vh] flex-col gap-4 overflow-auto rounded-t border-t border-border bg-background p-6">
```
gumclaw, line 11: From original PR by @aaronw122: aria-describedby={undefined} suppresses radix console warning

slavingia:
@gianfrancopiana can you review? lgtm overall though.

app/javascript/components/Discover/Layout.tsx
```diff
@@ -110,6 +110,7 @@ export const Layout: React.FC<{
   className?: string;
   children: React.ReactNode;
   forceDomain?: boolean;
+  stickyBar?: React.ReactNode;
```
aaronw122, line 113: sticky bar added for mobile

app/javascript/components/ui/BottomSheet.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <X className="size-5" />
+    </Dialog.Close>
+  </div>
+);
+
+export const BottomSheetFooter = ({ children }: { children?: React.ReactNode }) => (
+  <div className="flex items-center justify-end gap-4 pt-2">
+    {children}
+    <Dialog.Close asChild>
+      <Button color="primary">Done</Button>
```
gianfrancopiana, line 31: The sheet has a big "Done" button at the bottom. But the filter already applied the moment I tapped a checkbox. The search ran, the results updated. "Done" just closes the sheet. So the button is lying about its job.



Consider renaming it to "Close" or, better, **"Show [x] products"** (with the real count). Now it's a button that tells me what I'm about to see. When results = 0, disable the button or swap it to "Clear filters" so the user isn't invited to tap through to an empty screen.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+      !["from", "curated_product_ids"].includes(key) &&
+      searchParams[key] != null &&
+      searchParams[key] !== defaults[key],
+  );
+
+  const uid = React.useId();
+  const minPriceUid = React.useId();
+  const maxPriceUid = React.useId();
+
+  const concatFoundAndNotFound = (
```
gianfrancopiana, line 83: Consider this scenario:

https://github.com/user-attachments/assets/6fc89581-3262-4d58-9a00-e6c3ce972937

1. I pick a tag and apply filters.
2. Open another filter and tap one of its options — and right as I do, some of the other options disappear from the list.

This is pre-existing search behavior (facets recompute against all active filters), but mobile makes it way more obvious. 

The server sometimes drops facet options from its response. We already patch this for options the user had selected. We should extend the patch to cover options they'd seen but not selected too. (Fixing the server is out of scope.)

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  return (
+    <>
+      <div
+        role="toolbar"
+        aria-label="Filters"
+        className="flex overflow-x-auto gap-2 px-4 [scrollbar-width:none] [&::-webkit-scrollbar]:hidden"
+      >
+        {visibleFilters.map((filter) => (
+          <Pill key={filter.key} asChild className={`shrink-0 cursor-pointer ${filter.active ? "bg-accent/20 border-foreground" : "bg-transparent"}`}>
```
gianfrancopiana, line 248: Pill height comes out to ~40px (Pill default `px-3 py-2`). Under the 44px Apple and Google recommend. Pills are decoration elsewhere, but here they're the primary mobile surface. We should bump to `min-h-11`.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+            All Products
+            <Checkbox
+              wrapperClassName="ml-auto"
+              checked={!searchParams.tags?.length}
+              disabled={!searchParams.tags?.length}
+              onChange={() => updateParams({ tags: undefined })}
+            />
+          </Label>
+          {results ? (
+            <FilterCheckboxes
```
gianfrancopiana, line 135: The backend caps tags at 8 (`MAX_NUMBER_OF_TAGS` in `product/searchable.rb`), and the client slices to the first 5 with a "Show more" link. So "Show more" can only ever reveal 3 extra rows. And the sheet grows to fit them anyway. We can just show all 8.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+          ))}
+        </Fieldset>
+      ),
+    },
+    {
+      key: "tags",
+      title: "Tags",
+      label: (searchParams.tags?.length ?? 0) > 0 ? `Tags (${searchParams.tags!.length})` : "Tags",
+      active: (searchParams.tags?.length ?? 0) > 0,
+      visible: (results?.tags_data?.length ?? 0) > 0 || (searchParams.tags?.length ?? 0) > 0,
```
gianfrancopiana, line 122: Which pills I see depends on what's in the results. Open the page, see one set of pills. Filter down, see a different set. Some filters hide because no products match. Now the pill I wanted to tap moved or isn't there anymore.

If we keep pill positions stable (always show the same pills, disable the ones with no options), the bar becomes a dependable map instead of a moving target.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+          </Pill>
+        ))}
+        {hasOfferCode ? (
+          <Pill asChild color="primary" className="shrink-0 cursor-pointer">
+            <button onClick={() => updateParams({ offer_code: undefined })}>
+              {searchParams.offer_code}
+            </button>
+          </Pill>
+        ) : null}
+        {hasActiveFilters ? (
```
gianfrancopiana, line 262: Every pill in the row opens a sheet, except "Clear all," which silently wipes everything. It looks the same, sits in the same row, but does something completely different. A user could easily tap it expecting a sheet and lose all their filters instead.

It also appears and disappears depending on whether any filters are active, which shifts the other pills around (same rearranging problem above).

I would suggest pulling "Clear all" out of the pill row. Make it a plain text link sitting above or below the pills in a fixed spot. Now the pills are a stable set of filter controls and nothing else. The reset action is visible when needed, and it can't be confused with a filter.

app/javascript/components/Discover/MobileFilterBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+  ];
+
+  const visibleFilters = filters.filter((f) => f.visible);
+
+  return (
+    <>
+      <div
+        role="toolbar"
+        aria-label="Filters"
+        className="flex overflow-x-auto gap-2 px-4 [scrollbar-width:none] [&::-webkit-scrollbar]:hidden"
```
gianfrancopiana, line 245: The pill row scrolls sideways with the scrollbar hidden. If there's more to the right, I have no way to know.

There are only 5–7 pills total (sort, tags, contains, price, rating, plus occasionally offer code and clear all). That's a small, fixed set. They'd fit on two rows on any phone. We can wrap them with `flex-wrap` instead of `overflow-x-auto`. Now everything is visible at a glance, no scrolling to discover, no hidden affordance to worry about.

gianfrancopiana (commented):
@slavingia Added inline comments with feedback.

app/javascript/components/Discover/MobileFilterBar.tsx
aaronw122: yes aware of this. i was hoping to put up a PR to fix this pre-existing behavior as well so they are greyed out - can be done in front end. but was out of scope for this work.

app/javascript/components/Discover/MobileFilterBar.tsx
aaronw122: Since when is pill height expected to be 44px?

aaronw122: updating, thanks

app/javascript/components/Discover/MobileFilterBar.tsx
aaronw122: yes removed clear all - added it as button in button sheet. this is common pattern on bottom sheets for ecommerce. 

Screenshot 2026-04-18 at 12 09 21 PM
Screenshot 2026-04-18 at 12 09 18 PM

app/javascript/components/ui/BottomSheet.tsx
aaronw122: good shout! fixed. 

Screenshot 2026-04-18 at 12 45 32 PM

app/javascript/components/Discover/MobileFilterBar.tsx
aaronw122: i disagree - scrollable pills is a standard pattern in ecommerce, and it prevents us from taking up additional space in the viewport. ive modified the code so it will always show one pill cut off, in between regardless of viewport width. 

**booking.com:**
Screenshot 2026-04-18 at 12 28 17 PM

**ebay:** 
Screenshot 2026-04-18 at 12 30 13 PM

new PR here - https://github.com/aaronw122/gumroad/pull/1

gianfrancopiana: The scrollable pill pattern makes sense when you have dozens of filters and can't show them all. We have 5. They fit on two rows on any phone. `flex-wrap` makes every option visible at a glance. No hidden overflow, no peek-gap calculation, and no discoverability concern.


The peek-gap system solves a problem we'd only have if we chose scroll over wrap. I'd rather not carry that complexity when one CSS property does the job.

aaronw122: 🫡  - changed, branch updated here https://github.com/aaronw122/gumroad/pull/1

PR #4577, opened 2026-04-18
Fix NoMethodError on nil flow_of_funds in OrdersController#confirm

gumclaw:
## What

Fixes a `NoMethodError: undefined method 'issued_amount' for nil` raised in `Purchase#load_flow_of_funds` when a purchase is part of a combined Stripe charge and `processor_charge.flow_of_funds` is nil.

Sentry: https://gumroad-to.sentry.io/issues/7423138845/

## Why

`Purchase#load_flow_of_funds` had a fallback that built a simple `FlowOfFunds` when `processor_charge.flow_of_funds` was nil, but it only fired for non-Stripe processors:

```ruby
processor_charge.flow_of_funds ||= FlowOfFunds.build_simple_flow_of_funds(...) if StripeChargeProcessor.charge_processor_id != charge_processor_id
```

For Stripe charges, `StripeCharge#initialize` only assigns `flow_of_funds` when `stripe_charge_balance_transaction` is present. If the balance transaction hasn't materialized yet (e.g., a race in the confirm flow), `processor_charge.flow_of_funds` stays nil. The purchase was then passed that nil into `build_flow_of_funds_from_combined_charge`, which immediately calls `.issued_amount.cents` on it and blows up.

The fix extends the fallback to cover the Stripe path as well: if `processor_charge.flow_of_funds` is nil, use a simple flow of funds built from `total_transaction_cents` before branching on `is_part_of_combined_charge?`.

## Occurrences

1 event so far on Sentry.

## Estimated impact

Low volume, but the error surfaces in the checkout confirmation flow (`OrdersController#confirm`), so each occurrence is user-facing.

## Test plan

- [x] New spec `Purchase#load_flow_of_funds` reproduces the nil-flow-of-funds scenario for a combined Stripe charge; fails against `main`, passes with this fix.
- [x] Existing specs around `#save_charge_data` and `#build_flow_of_funds_from_combined_charge` still pass.
- [x] `bundle exec rubocop` clean on touched files.

---

AI disclosure: This PR was drafted with Claude Opus 4.6. Prompts included the Sentry error details, root-cause analysis, and instructions to implement the nil-guard in `load_flow_of_funds` plus a failing-first spec.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -2753,10 +2753,11 @@ class Purchase < ApplicationRecord
 
     def load_flow_of_funds(processor_charge)
       processor_charge.flow_of_funds ||= FlowOfFunds.build_simple_flow_of_funds(Currency::USD, self.total_transaction_cents) if StripeChargeProcessor.charge_processor_id != charge_processor_id
+      flow_of_funds = processor_charge.flow_of_funds || FlowOfFunds.build_simple_flow_of_funds(Currency::USD, self.total_transaction_cents)
       self.flow_of_funds = if is_part_of_combined_charge?
-        build_flow_of_funds_from_combined_charge(processor_charge.flow_of_funds)
+        build_flow_of_funds_from_combined_charge(flow_of_funds)
       else
-        processor_charge.flow_of_funds
+        flow_of_funds
       end
     end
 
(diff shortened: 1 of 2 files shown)


ershad:
Just one event, could be a rare edge case, not worth updating the payment flow to fix it for now.

PR #4578, opened 2026-04-18
Fix NoMethodError on nil flow_of_funds in OrdersController#confirm

gumclaw:
## Summary
- Sentry: https://gumroad-to.sentry.io/issues/7423138845/
- Occurrences: 1 event so far
- Estimated impact: Low volume but affects checkout confirmation flow

## Root cause
`Purchase#load_flow_of_funds` only built a fallback `FlowOfFunds` for non-Stripe charges. During checkout confirmation, a Stripe combined charge can still have `processor_charge.flow_of_funds == nil` when the Stripe balance transaction has not materialized yet. In that case, `nil` was passed into `build_flow_of_funds_from_combined_charge`, which calls `issued_amount` and raised `NoMethodError`.

## Fix
Always derive a local fallback `FlowOfFunds.build_simple_flow_of_funds(Currency::USD, total_transaction_cents)` when `processor_charge.flow_of_funds` is nil, then use that fallback for both combined-charge and non-combined-charge paths.

## Testing
- `bundle _2.5.10_ exec rspec spec/models/purchase_spec.rb:5463 spec/models/purchase_spec.rb:5871`

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -2753,10 +2753,11 @@ class Purchase < ApplicationRecord
 
     def load_flow_of_funds(processor_charge)
       processor_charge.flow_of_funds ||= FlowOfFunds.build_simple_flow_of_funds(Currency::USD, self.total_transaction_cents) if StripeChargeProcessor.charge_processor_id != charge_processor_id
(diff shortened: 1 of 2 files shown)


ershad:
Could be a rare edge case; only 1 instance of the error so far. Can skip fixing it.

PR #4579, opened 2026-04-19
Fix NoMethodError in build_flow_of_funds_from_combined_charge when flow_of_funds is nil

gumclaw:
## Problem

`Purchase#build_flow_of_funds_from_combined_charge` raises `NoMethodError: undefined method 'issued_amount' for nil (NoMethodError)` when called with a nil `combined_flow_of_funds` argument.

This happens when processor charges, events, or refunds don't have `flow_of_funds` set — the method is called from multiple places (`load_flow_of_funds`, `Charge::Disputable`, `Charge::Refundable`, `ChargeEventsHandler`, `SyncStatusWithChargeProcessorService`) and any of them can pass nil.

Sentry: https://gumroad-to.sentry.io/issues/7423689333/

## Fix

Add a nil guard at the top of `build_flow_of_funds_from_combined_charge` that returns nil early, consistent with how the non-combined-charge code path already handles missing flow_of_funds.

## Test

Added a test confirming that passing nil returns nil instead of raising.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -2594,6 +2594,8 @@ class Purchase < ApplicationRecord
   end
 
   def build_flow_of_funds_from_combined_charge(combined_flow_of_funds)
+    return if combined_flow_of_funds.nil?
+
     total_issued_amount_cents = combined_flow_of_funds.issued_amount.cents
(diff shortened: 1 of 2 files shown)


ershad:
Just 1 event, could be a rare edge case, not worth fixing right now.

PR #4588, opened 2026-04-20
Fix NoMethodError: undefined method 'permit' for String in CheckoutController#update

gumclaw:
## Problem

A malformed request to `CheckoutController#update` sends the `cart` parameter as a plain string instead of a nested hash. When this happens, `params.require(:cart)` returns a `String`, and calling `.permit` on it raises:

```
NoMethodError: undefined method 'permit' for an instance of String
```

**Sentry:** https://gumroad-to.sentry.io/issues/7426826751/

## Fix

Add an early guard at the top of the `update` action that returns `400 Bad Request` when `params[:cart]` is not an `ActionController::Parameters` instance. This prevents the malformed data from reaching `update_permitted_params`.

## Test

Added a spec confirming that sending `cart` as a string returns a 400 status.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -14,6 +14,8 @@ class CheckoutController < ApplicationController
   end
 
   def update
+    return head :bad_request unless params[:cart].is_a?(ActionController::Parameters)
+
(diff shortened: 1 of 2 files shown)


ershad:
Handling in https://github.com/antiwork/gumroad/pull/4591

PR #4589, opened 2026-04-20
Fix NoMethodError when discountCodes is nil in CheckoutController#update

gumclaw:
## Summary
- Guard against `nil` when mapping `discountCodes` in `CheckoutController#update`. When the request omits `discountCodes`, `update_permitted_params[:discountCodes]` is `nil`, causing `NoMethodError: undefined method 'map' for nil`.
- Default to `[]` so the update proceeds normally with no discount codes.

Sentry: https://gumroad-to.sentry.io/issues/7426826495/

## Test plan
- [x] Added a controller spec covering a `PATCH /checkout` request that omits `discountCodes` and verifies the cart is created with `discount_codes == []`.
- [x] New and adjacent specs pass locally (`bundle exec rspec spec/controllers/checkout_controller_spec.rb`).


gumclaw:
Closing in favor of #4591 which fixes all three checkout NoMethodErrors (nil items, nil discountCodes, and string cart param) together.

PR #4590, opened 2026-04-20
Fix NoMethodError when checkout update receives nil items

gumclaw:
## Summary
- guard checkout update against missing `items` and `discountCodes` params
- preserve the existing 400 response for malformed non-hash `cart` params
- add a controller spec covering checkout updates with omitted `items`

## Context
- Sentry issue: https://gumroad-to.sentry.io/issues/7426826658/
- Event count: 1 occurrence
- First seen: new issue

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -14,7 +14,10 @@ class CheckoutController < ApplicationController
   end
 
   def update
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing in favor of #4591 which fixes all three checkout NoMethodErrors (nil items, nil discountCodes, and string cart param) together.

PR #4593, opened 2026-04-20
Sync bank account holder name to Stripe on name-only updates

gianfrancopiana:
## What

When a seller updates only their bank account holder name (without re-entering their account number), the change is now synced to Stripe. Previously, the name was saved in the Gumroad database but never reached Stripe, causing payout failures when the bank required a specific name format (e.g., Katakana for Japanese banks).

Three changes:

1. **Sync trigger** — `UpdatePayoutMethod` now enqueues `HandleNewBankAccountWorker` after a holder-name-only update, so the change propagates to Stripe via the existing sync flow.
2. **Early return fix** — `StripeMerchantAccountManager#update_bank_account` previously returned early when the `bank_account_id` metadata matched, even if the holder name had changed. It now checks whether the holder name on Stripe still matches the local value for JP/VN/ID accounts (the countries where Stripe carries `account_holder_name`).
3. **UI hint** — Japanese sellers now see "Must be in Katakana" guidance on the bank account holder name field, since Stripe rejects Kanji and Japanese banks reject Roman letters for this field.

## Why

A Japanese seller's payouts were failing repeatedly with `incorrect_account_holder_name`. They updated their name from Roman letters to Kanji through Gumroad's payout settings, but the change never synced to Stripe because:

- The `BankAccount` model only has an `after_create_commit` callback for Stripe sync — not `after_update_commit`. Updating only the holder name is an `update`, not a `create`, so no sync was triggered.
- Even if triggered manually, the early return in `update_bank_account` would skip the sync because the `bank_account_id` metadata already matched (same bank account, just a different name).
- Stripe requires Katakana (not Kanji) for Japanese bank account holder names, but the UI gave no guidance about this.

---

This PR was implemented with AI assistance using Claude Opus 4.7 and gpt‑5.4 xhigh


gianfrancopiana:
@claude review

PR #4603, opened 2026-04-20
Fix race condition in Taxonomy find_or_create_by causing RecordNotUnique

gumclaw:
## Summary
- make `Taxonomy.find_or_create_by` and `Taxonomy.find_or_create_by!` retry safely after `ActiveRecord::RecordNotUnique`
- add a concurrent spec covering two threads creating the same taxonomy under the same parent
- keep the fix in the model so callers like seeds and `Onetime::AddNewVrTaxonomies` inherit the protection automatically

## Context
- Sentry issue: https://gumroad-to.sentry.io/issues/7428194352/
- observed once so far (first seen 2026-04-20), but this is a race condition that can affect any concurrent taxonomy creation
- avoids including any PII, user data, or database identifiers

## Testing
- bundle exec rspec spec/models/taxonomy_spec.rb

### DIFF
--- a/app/models/taxonomy.rb
+++ b/app/models/taxonomy.rb
@@ -7,4 +7,18 @@ class Taxonomy < ApplicationRecord
   has_one :taxonomy_stat, dependent: :destroy
 
   validates :slug, presence: true, uniqueness: { scope: :parent_id }
+
+  def self.find_or_create_by!(attributes, &block)
+    super
+  rescue ActiveRecord::RecordNotUnique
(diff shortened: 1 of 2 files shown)


gumclaw (commented):
@claude re-review — the previous review comments have been addressed in the latest commits. Please review the current state of the PR.

gumclaw:
@claude please re-review this PR — the previous review comments have been addressed in the latest commits.

PR #4606, opened 2026-04-20
Fix Elasticsearch parse error for timezones with DST gap at midnight

gumclaw:
## Problem

Sentry issue: https://gumroad-to.sentry.io/issues/7428332845/

`DashboardController#index` crashes with `Elasticsearch::Transport::Transport::Errors::BadRequest` for users whose timezone has a DST transition at midnight.

The root cause: when Elasticsearch receives a date-only string (e.g. `2026-03-22`) with `time_zone: 'Asia/Tehran'`, it tries to resolve `2026-03-22T00:00:00` in that timezone. But on that date, Iran observes DST — clocks skip from 00:00 to 01:00, so midnight doesn't exist. ES throws:

> `Cannot parse "2026-03-22": Illegal instant due to time zone offset transition (Asia/Tehran)`

## Fix

Use `timezone_formatted_offset` (a fixed UTC offset like `+03:30`) for ES **range query filters** instead of `timezone_id` (like `Asia/Tehran`). Fixed offsets always resolve cleanly since they don't have DST transitions.

**Date histogram aggregations** still use `timezone_id` so records are bucketed into the correct local day with DST awareness.

This matches the existing pattern documented on `User#timezone_formatted_offset`, which was specifically designed for resolving TZ database inconsistencies between Rails, Elasticsearch, and MySQL.

## Trade-off

Range boundaries may be off by up to ~1 hour around DST transitions (same trade-off already documented in `CachingProxy` and `User#timezone_formatted_offset`). This is acceptable since day bucketing remains DST-aware.

## Files changed

- `app/services/creator_analytics/sales.rb` — range filter
- `app/services/creator_analytics/product_page_views.rb` — range filter
- `app/services/creator_analytics/following.rb` — range filters
- `app/services/creator_analytics/churn/elasticsearch_fetcher.rb` — range filters
- `app/services/creator_analytics/churn/date_window.rb` — expose `timezone_formatted_offset`
- `spec/services/creator_analytics/sales_spec.rb` — regression test for Tehran DST gap

### DIFF
--- a/app/services/creator_analytics/churn/date_window.rb
+++ b/app/services/creator_analytics/churn/date_window.rb
@@ -22,10 +22,6 @@ class CreatorAnalytics::Churn::DateWindow
     seller.timezone_id
   end
 
-  def timezone_formatted_offset
-    seller.timezone_formatted_offset
-  end
-
   def daily_dates
     @daily_dates ||= (start_date..end_date).to_a
   end
--- a/app/services/creator_analytics/churn/date_window.rb
+++ b/app/services/creator_analytics/churn/date_window.rb
@@ -22,6 +22,10 @@ class CreatorAnalytics::Churn::DateWindow
     seller.timezone_id
   end
 
+  def timezone_formatted_offset
+    seller.timezone_formatted_offset
+  end
+
   def daily_dates
     @daily_dates ||= (start_date..end_date).to_a
   end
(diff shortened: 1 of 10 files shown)


gumclaw (commented):
@claude re-review — the previous review comments have been addressed in the latest commits. Please review the current state of the PR.

gumclaw:
@claude please re-review this PR — the previous review comments have been addressed in the latest commits.

PR #4614, opened 2026-04-21
Fix icon alignment on Cover uploader Tab buttons

gumclaw:
## What

Fix vertical alignment of icons inside the `TabIcon` component used by the `Tabs` (buttons variant). The icons in the "Computer files" and "External link" buttons on the product edit page's Cover section (e.g. `/products/risky/edit` → Cover → upload) appeared shifted toward the top of each button, visually misaligned with the button's text.

The fix makes `TabIcon` a flex container (`flex shrink-0 items-center text-xl`) so its inner icon is centered within its own line-height box.

## Why

`TabIcon` was previously `<div className="flex-shrink-0 text-xl">`. `text-xl` sets a line-height of 1.75rem (28px) on the container, but the icon inside is 20px (`size-5`). Because the div wasn't a flex container, the icon sat at the top of its 28px box rather than being vertically centered. This caused visible misalignment next to adjacent text in the buttons variant of `Tab`, which was reported on the Cover uploader tabs.

Making the container itself flex + `items-center` centers the icon within its own box, independent of how the parent `Tab` aligns its children (`items-start` by default, or `items-center` when overridden). The same fix also benefits the only other caller of `TabIcon` (`Developer/Tabs.tsx`), where `items-start` is preserved on the parent to keep the icon aligned with multi-line text blocks.

## Before/After

_Visual QA on `/products/:permalink/edit` → Cover → "Upload images or videos…":_

- Before: icons in "Computer files" and "External link" buttons rendered flush to the top of the button, visibly higher than the label text.
- After: icons are vertically centered with the label text.

(Screenshots/video to be attached during review.)

## Test Results

No behavioral tests affected; this is a CSS-class-only change in a shared UI component. `TabIcon` is used in exactly two places:

- `app/javascript/components/ProductEdit/ProductTab/CoverEditor.tsx` (single-line label — the reported bug)
- `app/javascript/components/Developer/Tabs.tsx` (multi-line label; parent `Tab` retains `items-start`, so the icon stays aligned with the top of the text block but is now centered within its own 28px line-height box)

QA steps:
1. Go to a product edit page (e.g. `/products/:permalink/edit`).
2. In the Cover section, click to add a new cover.
3. Verify the "Computer files" and "External link" buttons have their icons vertically centered with the label text.
4. Visit the Developer tabs (`/dashboard` → Developer → Modal Overlay / Embed tabs) and confirm those icons still visually align well with the multi-line descriptions.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.6

Prompt: Fix the icon alignment issue on the Cover section's "Computer files" and "External link" buttons in the product edit page; make the fix in the shared `TabIcon` component and verify it doesn't break the other buttons-variant tab usage.

### DIFF
--- a/app/javascript/components/ui/Tabs.tsx
+++ b/app/javascript/components/ui/Tabs.tsx
@@ -62,7 +62,7 @@ export const Tabs = React.forwardRef<HTMLDivElement, TabsProps>(({ children, cla
 Tabs.displayName = "Tabs";
 
 export const TabIcon = ({ children }: { children: React.ReactNode }) => (
-  <div className="flex-shrink-0 text-xl">{children}</div>
+  <div className="flex shrink-0 items-center text-xl">{children}</div>
 );
 
 interface TabProps extends Omit<React.HTMLProps<HTMLAnchorElement>, "selected"> {


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4615, opened 2026-04-21
Add top margin to referrers table Show more button

gumclaw:
## What

Add `mt-4` to the "Show more" button on the analytics referrers table so it no longer sits flush against the table above it.

## Why

The button had no top margin, making it visually cramped against the bottom of the table. Adding `mt-4` gives it proper breathing room consistent with other spacing in the analytics UI.

---

AI disclosure: Generated with Claude Opus 4.6. Prompt: add top margin to the "Show more" button on the sales/analytics referrers table by changing the Button's className from `"flex"` to `"mt-4 flex"`.


nyomanjyotisa (approved):
Note: the show more button changed temporarily to always show for testing. Originally it only show if more than 10 record 

## Before
Image

## After

Image

PR #4616, opened 2026-04-21
Fix ArgumentError: comparison of Integer with nil in suggested_price_greater_than_price

gumclaw:
## Summary
- add a nil guard in `suggested_price_greater_than_price` when `default_price_cents` is missing
- add a regression spec covering a persisted buyable product without an associated default price record

## Sentry
- Issue: https://gumroad-to.sentry.io/issues/7429332600/
- Event count: 2 events (7-day)
- Users affected: 2

## Root cause
`default_price_cents` can return `nil` via `buy_price_cents` when a buyable product has no default `Price` record. The validation compared `suggested_price_cents` against that `nil` value, which raised `ArgumentError`.

## Fix
Skip the validation when `default_price_cents` is `nil`, since there is no default price to compare against.

## Testing
- `RBENV_VERSION=3.4.3 rbenv exec bundle _2.5.10_ exec rspec spec/modules/product/prices_spec.rb`

### DIFF
--- a/app/modules/product/prices.rb
+++ b/app/modules/product/prices.rb
@@ -143,7 +143,7 @@ module Product::Prices
   end
 
   def suggested_price_greater_than_price
-    return if suggested_price_cents.blank? || !customizable_price || suggested_price_cents >= default_price_cents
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing as duplicate of #4613 which covers the same fix.

PR #4617, opened 2026-04-21
Fix missing spacing between recap options in community notification settings

gumclaw:
The Daily and Weekly recap frequency options in community notification settings had no spacing between them.

Added `gap-2` to the radio button flex container in `Communities/Index.tsx`, consistent with the parent `Dropdown` component's grid gap.


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4618, opened 2026-04-21
Show username when name is missing on admin product seller link

gumclaw:
When a user has no name set (only a username), the seller link in the admin product header was invisible because `product.user.name` is `null` and the `<Link>` rendered with empty content.

**Fix:** Add `username` to the admin product user payload and fall back to it in the header: `product.user.name || product.user.username`. The `username` method on User already falls back to `external_id`, so the link always has visible text.

**Changes:**
- `app/presenters/admin/product_presenter/card.rb` — include `username` in user hash
- `app/javascript/components/Admin/Products/Product.tsx` — add `username` to `ProductUser` type
- `app/javascript/components/Admin/Products/Header.tsx` — fall back to `username` when `name` is null
- `spec/presenters/admin/product_presenter/card_spec.rb` — update expected user hash

### DIFF
--- a/app/javascript/components/Admin/Products/Header.tsx
+++ b/app/javascript/components/Admin/Products/Header.tsx
@@ -46,7 +46,7 @@ const AdminUsersProductsHeader = ({ product, isCurrentUrl }: Props) => (
               <DateTimeWithRelativeTooltip date={product.created_at} utc />
             </li>
             <li>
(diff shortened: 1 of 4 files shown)


gumclaw:
Closing in favor of the existing `fix-admin-product-seller-link-no-name` branch which takes a better approach: replacing `name` with `display_name` directly in the presenters instead of adding a separate field. It also covers more places (payment presenter, multiple matches, purchase presenter).

PR #4621, opened 2026-04-21
Fix seller link not displayed on admin pages when user has no name

gumclaw:
When a user has no name set (only a username), the seller link was invisible on admin pages because `user.name` is `null` and the link rendered with empty content.

**Fix:** Use `display_name` instead of `name` in admin presenters. `display_name` handles the fallback chain (name → username → email) so the link always has visible text.

**Changes:**
- `app/presenters/admin/product_presenter/card.rb` — use `display_name`
- `app/presenters/admin/product_presenter/multiple_matches.rb` — use `display_name`
- `app/presenters/admin/payment_presenter.rb` — use `display_name`
- `app/presenters/admin/purchase_presenter.rb` — use `display_name` for refund user


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4622, opened 2026-04-21
Fix menu item alignment in discover mobile menu

gumclaw:
On the discover page mobile menu, items without children (like 'All' and 'Other') had no top margin and appeared inline instead of as full-width rows.

**Root cause:** `MenuItemLink` is an `<a>` element which is inline by default. Items with children already got `flex!` via their className prop, but items without children stayed inline. On inline elements, vertical padding doesn't affect layout flow, so the spacing was broken.

**Fix:** Add `flex` to the base `MenuItemLink` classes so all menu items are block-level. This ensures proper padding behavior and makes every item take a full row in the menu.

**Before:** 'All' and 'Other' appear on the same line with no top margin
**After:** All items are properly spaced as full-width rows


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4624, opened 2026-04-21
Center gift and trash icons in wishlist product card footer

gumclaw:
The gift and trash icons in wishlist product card footers were not vertically centered. When the price section is taller than the icon cell, the icons sat at the top instead of being centered.

Added `placeItems: 'center'` to the grid containers wrapping both the gift and trash buttons.


nyomanjyotisa (approved):
## Before

Image

## After

Image

PR #4647, opened 2026-04-21
Remember filter settings on sales page

### LINKED ISSUE #4634: Remember filter settings when going from sales detail to sales page
Go to sales page
Set filter (for example 40+ euro price)
click on a sale
go back to sales page
filter is not longer active

gumclaw:
## Summary

Fixes #4634.

When a user on the Sales page applies filters (product, date range, amount, country, active-only toggle, sort, page, etc.), clicks through to a sale detail, then navigates back, the filter state is now preserved instead of resetting to defaults.

### Approach

- On mount, the `CustomersPage` `useState` initializers read filter values from `URLSearchParams`:
  - `minimum_amount_cents`, `maximum_amount_cents`
  - `created_after`, `created_before`
  - `country`, `active_customers_only`
  - `sort_key`, `sort_direction`
  - `page`
  - `products`, `variants` (multi-value, variants encoded as `productId:variantId`)
  - `excluded_products`, `excluded_variants`
  - `query` / `email` (already handled previously)
- A new `useEffect` syncs the current `query` + `includedItems` + `excludedItems` back into the URL via `window.history.replaceState` whenever they change, so the browser's back button returns to a URL that rehydrates the filters.
- Unrelated existing URL params are preserved; only the filter-related keys are rewritten.

### Test plan

- [ ] Apply multiple filters on the sales page
- [ ] Verify URL updates as filters change
- [ ] Click a sale row, then click the browser back button — filters should still be applied
- [ ] Share a URL with filters applied — opening it should restore the same filter state
- [ ] Clearing all filters should produce a clean URL (no stale filter params)

### DIFF
--- a/app/javascript/components/Audience/CustomersPage.tsx
+++ b/app/javascript/components/Audience/CustomersPage.tsx
@@ -41,6 +41,8 @@ import placeholder from "$assets/images/placeholders/customers.png";
 
 type Product = { id: string; name: string; variants: { id: string; name: string }[] };
 
+type Item = { type: "product"; id: string } | { type: "variant"; id: string; productId: string };
+
 export type CustomerPageProps = {
   customers: Customer[];
   pagination: PaginationProps | null;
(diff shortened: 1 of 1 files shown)


gumclaw:
Closing: superseded or non-compliant.

PR #4648, opened 2026-04-21
Add minimum license uses filter to sales page

### LINKED ISSUE #4633: Add option in sales filter to filter on license use of more than X
It would be very helpful for me to be able to filter the sales list on users that used their license more than X amount of times so I can enforce the limits more easily. Have internal API tools for it now so maybe this is overkill for most sellers 👍

gumclaw:
## Summary

Adds a "Minimum license uses" numeric filter to the Sales page so that sellers can narrow the list down to purchases whose license has been activated (used) at least N times.

Closes #4633

### What's new

- **Frontend** — new numeric input in the Sales filter popover (`CustomersPage.tsx`) bound to a new `minimumLicenseUses` field on the customer `Query`.
- **API** — `GET /customers/paged` now accepts a `minimum_license_uses` query parameter.
- **Backend** — when the filter is provided, `CustomersController#fetch_sales` pre-computes purchase IDs from SQL (`current_seller.sales.joins(:license).where("licenses.uses >= ?", threshold)`) and passes them as an `id:` filter to `PurchaseSearchService`. This keeps pagination/count correct without requiring a new Elasticsearch index field or a reindex.

### Why SQL pre-filter rather than ES

License `uses` changes frequently (every license activation). Adding it to the purchase ES mapping would introduce churn on every activation event. Pre-computing the matching IDs per-seller keeps the change minimal and avoids an index/reindex migration, at the cost of an extra SQL query when this filter is active.

## Test plan

- [x] Added `spec/controllers/customers_controller_spec.rb` case covering `minimum_license_uses` with three thresholds (5, 10, 1000).
- [ ] Manually verify the input appears under the Filter popover on the Sales page.
- [ ] Manually verify purchases without a license are excluded when the filter is active.
- [ ] Verify pagination/count remain correct when the filter narrows results.

### DIFF
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -60,6 +60,7 @@ class CustomersController < Sellers::BaseController
       created_before: params[:created_before],
       country: params[:country],
       active_customers_only: ActiveModel::Type::Boolean.new.cast(params[:active_customers_only]),
+      minimum_license_uses: params[:minimum_license_uses],
     )
     customers_presenter = CustomersPresenter.new(
       pundit_user:,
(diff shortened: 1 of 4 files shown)


gumclaw:
Closing: superseded or non-compliant.

PR #4649, opened 2026-04-21
Add "License used more than" filter to the Sales page

### LINKED ISSUE #4633: Add option in sales filter to filter on license use of more than X

gumclaw:
## Summary

Closes #4633.

Adds a new filter on the Sales page that lets sellers find customers whose license key has been used more than X times. This makes it easier for sellers to enforce license-use limits without relying on internal/API tooling.

- **Frontend**: new numeric input ("License used more than") in the Sales filter popover, wired to a new `minimumLicenseUses` query param in `app/javascript/data/customers.ts`.
- **Backend**: `CustomersController#paged` accepts a new `minimum_license_uses` param and forwards it to `PurchaseSearchService` via a new `license_uses_greater_than` option.
- **Indexing**: `License#uses` is now indexed on the Purchase Elasticsearch document. An `after_commit` callback on `License` re-indexes the related purchase whenever `uses` or `serial` changes. A migration adds the new mapping to the existing `purchases` index.

Purchases without a license are never matched when the filter is active (because `license_uses` will not be set on those documents).

## Test plan

- [ ] Apply the ES mapping migration (`20261123000000_add_license_uses_to_elasticsearch`) in dev/staging.
- [ ] Re-index existing purchases (or wait for them to be re-indexed via the License callback) so `license_uses` is populated.
- [ ] On the Sales page, open the filter popover and set "License used more than" to a number; confirm the list only shows sales whose license has been used strictly more than that number of times.
- [ ] Leave the filter empty and confirm all previous behavior is preserved.
- [ ] Combine with other filters (product, date range, country) and confirm results are ANDed correctly.
- [ ] bundle exec rspec spec/services/purchase_search_service_spec.rb spec/controllers/customers_controller_spec.rb spec/models/license_spec.rb spec/models/concerns/purchase/searchable_spec.rb


db/migrate/20261123000000_add_license_uses_to_elasticsearch.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddLicenseUsesToElasticsearch < ActiveRecord::Migration[7.1]
+  def up
+    EsClient.indices.put_mapping(
```
slavingia, line 5: @ershad  look right to you?

slavingia (approved):

db/migrate/20261123000000_add_license_uses_to_elasticsearch.rb
ershad: Asked a related question here: https://github.com/antiwork/gumroad/issues/4633#issuecomment-4297647906

If it's required only via the API, we can perform DB lookups and avoid updating ES. If it's needed on the Sales page, we need to index license_uses in ES and backfill it for past purchases.

slavingia: IMO should be available via dashboard too

slavingia:
@claude  review

ershad:
Update: Will merge/deploy today after a final round of verification.

PR #4655, opened 2026-04-21
Harden bank-account sync against silent Stripe failures

gianfrancopiana:
Issue #4658

Follow-up to #4593.

## What

Hardens bank-account sync against silent Stripe failures:

- `JapanBankAccount` rejects mixed-script names (the incident case, e.g. katakana + ASCII space). The settings page shows a matching inline error.
- `UpdatePayoutMethod#process` runs inside `user.with_lock`, validates the new record before deleting the old one, and asserts that exactly one alive bank remains after replacement.
- `update_bank_account` returns a symbol for every outcome; `HandleNewBankAccountWorker` turns `:stripe_unknown_error` into a Sidekiq retry; the `account.updated` webhook enqueues the worker on unknown errors instead of silently continuing.
- Every sync failure now lands an admin `payout_note` breadcrumb on the user.

## Why

A JP creator's payouts were silently blocked for five weeks because Stripe rejected a katakana holder name containing an ASCII space. The error only went to Sentry with no retry, no user email, no admin surface. These changes close each link in that silence chain: block bad input at the boundary, keep creators on a working bank when new ones fail validation, and make unclassified failures loud and diagnosable.

---

This PR was implemented with AI assistance using Claude Opus 4.7 and gpt‑5.4 xhigh

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
@@ -789,8 +789,9 @@ module StripeMerchantAccountManager
     is_charges_disabled = !stripe_account["charges_enabled"]
     charges_newly_disabled = stripe_account["charges_enabled"] == false && stripe_previous_attributes["charges_enabled"] == true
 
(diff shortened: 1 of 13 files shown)


gianfrancopiana:
@claude review

PR #4662, opened 2026-04-22
Preserve sales filters when returning from sale detail

### LINKED ISSUE #4634: Remember filter settings when going from sales detail to sales page

nyomanjyotisa:
Fixes #4634.

## What

- Sales page filters (price, date, country, products, active only, sort, pagination) now survive navigating to a sale and back.
- The back arrow on the sale detail page uses `history.back()` when the previous entry is same-origin, falling back to a hard `/customers` link for direct visits or cross-origin referrers.
- Added a system spec covering the filter-preserved-on-back flow.

## Why

Filters lived in `useState`, which resets every time `CustomersPage` remounts. Clicking a sale triggers `router.visit`, which unmounts the page; the browser back then remounted it with defaults, wiping the user's work.

Inertia's `useRemember`. Drop-in for `useState`, stores state in `window.history.state.rememberedState` keyed per history entry. Back/forward restores it automatically. Chose this: smaller diff, matches the exact UX users expect, no URL churn.

The back arrow change is needed because `<Link href="/customers">` pushes a fresh history entry (new `rememberedState` bucket, filters gone). `history.back()` pops to the previous entry, which still has the remembered filters. Guarded against two edge cases: direct visits (`history.length <= 1`) and cross-origin referrers (so we don't bounce users off-site).

## After

https://github.com/user-attachments/assets/ed891177-37ed-44c6-91bf-3bebe7b17ca8

### DIFF
--- a/app/javascript/components/Audience/CustomerDetailPage.tsx
+++ b/app/javascript/components/Audience/CustomerDetailPage.tsx
@@ -208,7 +208,17 @@ const CustomerDetailPage = ({
         showTitleOnMobile
         title={
           <div className="flex flex-wrap items-center gap-2">
-            <Link href="/customers" aria-label="Back to customers" className="mr-4 hidden no-underline sm:inline">
+            <Link
+              href="/customers"
+              aria-label="Back to customers"
+              className="mr-4 hidden no-underline sm:inline"
+              onClick={(e) => {
+                if (window.history.length <= 1) return;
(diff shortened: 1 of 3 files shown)


gumclaw (approved):
All three review items addressed and CI is green. LGTM.

PR #4665, opened 2026-04-22
Add counter cache for sales_count_for_inventory

### LINKED ISSUE #4582: Setting maximum number of purchases to 1 million keeps loading
If I try to change the max purchases for one of the sub products of a product, the dashboard keeps loading.

nothing visible in console

nyomanjyotisa:
# Cache sales_count_for_inventory to fix save/checkout hangs on high-volume products

Fixes #4582.

> Note: this diff is ~3,800 lines, but ~3,000 of those are three VCR cassettes

## What

Adds a counter cache column `sales_count_for_inventory_cache` on `base_variants` and `links`, populated by Purchase and Subscription lifecycle hooks, read behind a Flipper flag.

- Migration adds the column (default 0, null: false) to both tables.
- `Purchase` gains a pure boolean predicate `counts_towards_inventory?` and an `after_commit` sync that applies the signed delta to the cached columns atomically via `update_all(... col = col + N)`.
- `Subscription` hooks `deactivated_at` changes and applies the cache delta via a single atomic `UPDATE ... INNER JOIN` per table, race-safer than iterating purchases.
- `BaseVariant#sales_count_for_inventory` and `Link#sales_count_for_inventory` return the cache column when `Feature.active?(:inventory_counter_cache)`, else fall back to the existing SUM.
- `Onetime::BackfillInventoryCounterCache` populates historical data; runs in prod console.

## Why

The validation `BaseVariant#max_purchase_count_is_greater_than_or_equal_to_inventory_sold` calls `sales_count_for_inventory`, which runs `purchases.counts_towards_inventory.sum(:quantity)`. On MacWhisper Free (variant 2890037, 337k purchases via base_variants_purchases, 628k rows scanned) this SUM exceeds 120s and the browser's POST to update the variant hangs indefinitely with no console error. Jordi reported this in #4582.

The bottleneck is 628k random PK lookups into `purchases` combined with bitfield flag filters (`flags & 4 = 4`, `flags & 131072 = 0`, etc.) that no B-tree index can prune. Making the SUM fast via pure query optimization isn't possible given MySQL's index model.

**Why this matters beyond the save hang:** `sales_count_for_inventory` is also called from `BaseVariant#quantity_left`, `BaseVariant#sold_out?`, `BaseVariant#as_json(for_views: true)`, `Link#remaining_for_sale_count`, `Sku#quantity_left`, and critically `Purchase` inventory check at purchase time ([purchase.rb:3407](app/models/purchase.rb#L3407)). Every MacWhisper Free variant read path will hang the moment `max_purchase_count` becomes non-nil — checkout would start timing out for real buyers. A pure validation short-circuit would let Jordi save but light a fuse for every subsequent purchase.

## Rollout

One PR, staged activation via Flipper:

1. Merge and deploy with flag **off** — writers populate cache, readers still run SUM. Zero behavior change, fully reversible.
2. In prod console: `Onetime::BackfillInventoryCounterCache.process`
3. Flip flag: `Feature.activate(:inventory_counter_cache)`.
4. Rollback: `Feature.deactivate(:inventory_counter_cache)`.

## Before

Takes 18s with 2 million purchases on my local

https://github.com/user-attachments/assets/3783d947-5972-4617-b25e-104a17e441a9

## After

Takes 0.5s with 2 million purchases on my local

https://github.com/user-attachments/assets/c75a4b06-4c04-4205-ad43-a22b23d70895

---

Generated with assistance from Claude Opus 4.7


app/services/onetime/backfill_inventory_counter_cache.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    private
+      def backfill_base_variants(start_id)
+        BaseVariant.where("id >= ?", start_id).in_batches(of: BATCH_SIZE) do |batch|
+          ReplicaLagWatcher.watch
+          batch.each do |variant|
+            count = variant.purchases.counts_towards_inventory.sum(:quantity)
+            BaseVariant.where(id: variant.id).update_all(sales_count_for_inventory_cache: count)
+          end
+          puts "BaseVariant backfill: reached id=#{batch.maximum(:id)}"
```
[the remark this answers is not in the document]
gumclaw: Valid catch — pushed a fix in 1649683c. Added `.not_is_gift_sender_purchase` to the `counts_towards_inventory` scope so it now matches the `counts_towards_inventory_for?` method.

nyomanjyotisa:
@claude review

slavingia:
@ershad  could you review this too?

db/migrate/20260422000000_add_sales_count_for_inventory_cache.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddSalesCountForInventoryCache < ActiveRecord::Migration[7.1]
+  def change
+    add_column :base_variants, :sales_count_for_inventory_cache, :integer, default: 0, null: false
+    add_column :links, :sales_count_for_inventory_cache, :integer, default: 0, null: false
```
slavingia, line 6: Are we okay to do this update to the products table? @ershad 

It may be better to use Redis for this, which we generally use for lots of caching type stuff.

ershad: Yeah, adding columns to a large table would be a last resort.

@nyomanjyotisa, could you check if we can compute this number efficiently in real time without caching using Elasticsearch aggregation, since we already have purchase data there?

nyomanjyotisa: @ershad I don't think ES works here. Every ES read in the codebase is either search/listing or analytics, none gate a transactional write, and inventory enforcement does.

A few concrete issues:
- Two buyers racing for the last unit could both see capacity and both succeed.
- `quantity` isn't in the purchases mapping, so we'd need to reindex the full index.
- `in_progress` purchases count toward inventory but are the most likely to not be indexed yet, meaning undercount and overselling.

If the concern is the column on links, I think we should move the counter to a separate `product_inventory_counters` table instead?

slavingia: Can we calculate how long this migration would take to run, run it, and then catalog the process with Gumclaw? We need to ship faster and this is a big bottleneck atm

gumclaw:
@claude review

db/migrate/20260422000000_add_sales_count_for_inventory_cache.rb
nyomanjyotisa: Actually, we have a migration earlier this year on `links` table shipped in #2920 on 2026-01-20, and no migration issue.

Sized it from prod data:

| Table          | Rows  | Total Size |
|----------------|-------|------------|
| links          | 11.2M | ~25 GB     |
| base_variants  | 5.4M  | ~0.9 GB    |

Migration time estimate:
- base_variants: 1.6 to 2.7 min (170k rows/sec measured)
- links: 15 to 25 min (37k rows/sec measured)

nyomanjyotisa:
@claude review

db/migrate/20260422000000_add_sales_count_for_inventory_cache.rb
slavingia: Then I think we should proceed.

app/models/purchase.rb
```diff
@@ -358,6 +358,167 @@ class Purchase < ApplicationRecord
...
+      is_original = raw_flags & original_sub_bit != 0
+      is_gift_receiver = raw_flags & gift_receiver_bit != 0
+      return false unless is_original || is_gift_receiver
+      return false if subscription_deactivated_at.present?
+    end
+
+    true
+  end
+
+  def self.skip_inventory_counter_callbacks
```
ershad, line 403: Looks like unused, we can remove this?

ershad (approved):

PR #4666, opened 2026-04-22
Enforce offer code usage limit across cart line items

### LINKED ISSUE #4664: Discount code with limit of 1 use was used twice
Image

For some reason a user was able to use a discount code twice when it was set to 1. I don't think they actually bought it twice since the amount is just for one, but seems like this shouldn't be possible.

nyomanjyotisa:
Fixes https://github.com/antiwork/gumroad/issues/4664

## What

`Order::ChargeService#perform` now validates offer code usage across all cart line items before creating a Stripe charge. When a cart would collectively consume an offer code beyond its `max_purchase_count`, every line item using that code is transitioned to `failed` state with `EXCEEDING_OFFER_CODE_QUANTITY` — before any money moves.

- New class method: `Purchase.validate_offer_code_usage_across_line_items(purchases)`
- Called from `Order::ChargeService#perform` at the top, before charge assembly
- Uses `Purchase::MarkFailedService` so the failed state is persisted and survives downstream saves

## Why

Reported in antiwork/gumroad#4664: an offer code with `max_purchase_count: 1` was used twice on a single order. Investigation of the live data showed:

- Same buyer, same session, same Order (`id=103876238`), same Charge (`ch_2TOv8k9e1RjUNIyY0VZ0CbnH`)
- Two Purchase rows: one for variant "MacWhisper Free" (price 0), one for "MacWhisper Pro" (paid $48 after 25% off)
- Both Purchase rows referenced the same `offer_code_id`
- Created 0ms apart (same checkout, one user action)

Root cause: `Purchase#validate_offer_code` runs in `before_create`, per Purchase, so each line item sees `times_used = 0` during its own validation and passes. There was no aggregate check once all cart line items existed. This PR adds that aggregate check in the single window where it's possible — after `Order::CreateService` returns, before `Charge::CreateService` is called.

Single-line carts are skipped because the per-Purchase validator already handles them.

## After

https://github.com/user-attachments/assets/8996f841-72be-4fc9-80a1-ba63f276601d

Generated with assistance from Claude Opus 4.7

### DIFF
--- a/app/models/concerns/purchase/blockable.rb
+++ b/app/models/concerns/purchase/blockable.rb
@@ -39,7 +39,8 @@ module Purchase::Blockable
                          PurchaseErrorCode::TEMPORARILY_BLOCKED_PRODUCT,
                          PurchaseErrorCode::BLOCKED_CHARGE_PROCESSOR_FINGERPRINT,
                          PurchaseErrorCode::BLOCKED_CUSTOMER_EMAIL_ADDRESS,
-                         PurchaseErrorCode::BLOCKED_CUSTOMER_CHARGE_PROCESSOR_FINGERPRINT]
+                         PurchaseErrorCode::BLOCKED_CUSTOMER_CHARGE_PROCESSOR_FINGERPRINT,
+                         PurchaseErrorCode::EXCEEDING_OFFER_CODE_QUANTITY]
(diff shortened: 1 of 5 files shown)


nyomanjyotisa:
@claude review

PR #4672, opened 2026-04-22
Remove Iffy fraud model integration

gumclaw:
## Summary
- remove the remaining Iffy fraud-model integration points from admin controllers, routes, config, and specs
- remove the Iffy-specific comment validation and auth-bypass behavior plus related test coverage
- remove the unused Iffy token entries from environment example files

## Why
Iffy is no longer used, so this removes the remaining integration points while keeping the deprecated enum value intact to avoid data issues.


config/initializers/004_constants.rb
```diff
@@ -61,7 +61,7 @@
                taylorswift taylorswift13 nickiminaj oprah jtimberlake
                theellenshow ellen selenagomez kaka aplusk love recommended_products pay
                _dmarc _domainkey blog cloud-front-static-1 creators
-               customers domains domains-staging files files-3 iffy
+               customers domains domains-staging files files-3
```
slavingia: @gumclaw keep this around, also add cli

gumclaw: Done — kept `iffy` in the reserved subdomains list and added `cli`. [f945fb8](https://github.com/antiwork/gumroad/commit/f945fb8)

PR #4674, opened 2026-04-22
Remove Mongo mirroring and risk queue code

gumclaw:
Removes all Mongo mirroring and risk-related Mongo code. The `mongoid` gem and active Mongoid::Document models (BlockedObject, EmailEvent, etc.) are kept.

**Removed (28 files, -314 lines):**
- `Mongoer` class, `Mongoable` module, `MongoCollections` constants
- `SaveToMongoWorker`, `UpdateInMongoWorker`
- `to_mongo` hooks on Purchase, OfferCode, ServiceCharge
- `insert_null_chargeback_state` on User (wrote to Mongo on every user creation)
- `log_suspension_time_to_mongo` on User risk transitions
- `refund_queue` class method (read from Mongo)
- `Admin::RefundQueuesController` and its route
- All related specs

**Changed:**
- `:mongo` Sidekiq queue → `:default` for 3 workers that used the queue name but don't write to Mongo


app/modules/user/risk.rb
```diff
@@ -239,18 +234,5 @@ def payouts_status
   end
 
   class_methods do
-    def refund_queue(from_date = 30.days.ago)
```
slavingia, line 242: We can bring back if we desire using MySQL but my feeling is that the new scheduled payouts feature is the best way to handle that.

ershad: Yeah, scheduled payouts should take care of it.

PR #4679, opened 2026-04-22
Fix ActiveRecord::Deadlocked in Google OAuth2 sign-in with retry logic

gumclaw:
Sentry: https://gumroad-to.sentry.io/issues/7434404347/

## Summary
- add retry handling for `ActiveRecord::Deadlocked` in `User::SocialGoogle.find_or_create_for_google_oauth2`
- retry the full find-or-create flow up to 2 times so a concurrent successful request can create the user first
- add regression specs for a successful retry and exhausted retries

## Root cause
Two concurrent Google OAuth2 sign-in requests for the same user can both miss the initial lookup and race into user creation. When both requests try to persist the same logical user, MySQL can deadlock before the loser request gets a chance to observe the newly-created record.

## Occurrences
- 1 (first seen)

## Estimated impact
- Low frequency but blocks user login when it occurs

### DIFF
--- a/app/models/concerns/user/social_google.rb
+++ b/app/models/concerns/user/social_google.rb
@@ -41,36 +41,47 @@ module User::SocialGoogle
 
   class_methods do
     def find_or_create_for_google_oauth2(data)
-      if data["uid"].blank?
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review once

PR #4684, opened 2026-04-22
[codex] Fix offer code validation across cart line items

slavingia:
## Summary
- preserve duplicate same-product checkout line items when computing offer code discounts
- aggregate offer code quantities by product permalink before enforcing usage limits
- prevalidate cart-level discount codes in `Order::CreateService` before creating any purchases
- key failed offer-code tracking by line-item identity in order create/confirm flows
- add regression coverage for the service, controller, and order-creation paths

## Root cause
Checkout was sending the discount-computation payload keyed by product permalink. When the cart contained the same product in multiple line items, later entries overwrote earlier ones, so the server never saw the combined quantity. Purchase creation then validated each line item independently, which allowed an overused offer code to survive until the charge phase.

## Impact
This fixes carts with multiple line items for the same product, including variant combinations, so limited-use offer codes fail consistently before purchase creation instead of surfacing as a later charge-time problem.

## Validation
- `DISABLE_SPRING=1 bin/rspec spec/services/offer_code_discount_computing_service_spec.rb spec/controllers/offer_codes_controller_spec.rb spec/services/order/create_service_spec.rb`


slavingia:
@claude review once

PR #4693, opened 2026-04-23
P0 Fix: gracefully handle content moderation API failures

gumclaw:
## Problem

Product publishing and saving has been broken platform-wide since ~3 PM ET on April 22. 25+ support tickets, 14,800+ Sentry events (`GUMROAD-HX`).

The OpenAI moderation API (`omni-moderation-latest`) started returning 400 errors. The `ClassifierStrategy` and `PromptStrategy` caught the error but re-raised it, which caused the entire publish/save to fail with "Something broke."

## Fix

On API errors (Faraday::Error, timeouts, OpenAI::Error), return a compliant result instead of crashing. Content moderation is a best-effort check. Products should still be publishable when the moderation API is down.

**Changes:**
- `ClassifierStrategy`: catch network/API errors, return compliant, notify Sentry
- `PromptStrategy`: same
- `ModerateRecordService`: safety net in `run_ai_strategies` for any uncaught thread errors
- Updated specs to match new graceful degradation behavior

All 33 content moderation specs pass.

Fixes antiwork/gumroad-private#310

### DIFF
--- a/app/services/content_moderation/moderate_record_service.rb
+++ b/app/services/content_moderation/moderate_record_service.rb
@@ -71,7 +71,13 @@ class ContentModeration::ModerateRecordService
         end
       end
 
-      threads.map(&:value)
+      threads.map do |t|
+        t.value
+      rescue StandardError => e
+        Rails.logger.error("ContentModeration strategy thread error: #{e.class}: #{e.message}")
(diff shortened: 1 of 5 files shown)


ershad:
Not the right fix, will check in detail tomorrow.

PR #4697, opened 2026-04-23
Fix restart checkout card reuse

gianfrancopiana:
## What

Fixes transparent subscription restarts so checkout honors a newly submitted payment method instead of silently falling back to the saved card on the old subscription.

When restart checkout sends Stripe payment-method data, we now treat that as "use the new card" even if `card_data_handling_mode` is missing. The regression tests cover both the restart adapter and the order creation flow using the same payload shape the checkout frontend sends today.

## Why

Expired subscription restarts go through the normal checkout flow, but the restart adapter was deciding whether to reuse the saved card from `card_data_handling_mode`. The frontend does not rely on that field for this path. It sends the new card through Stripe payment-method IDs instead.

That mismatch made it possible for a restart to authenticate and charge the previously saved card even after a buyer entered a different one. This keeps restart checkout aligned with the actual request payload and prevents us from reusing the old card when a new payment method was submitted.

---

This PR was implemented with AI assistance using gpt‑5.4 xhigh


gianfrancopiana:
@claude review

PR #4704, opened 2026-04-23
Fix Ghana P.O. box payout validation

gianfrancopiana:
## What
- Reject P.O. box payout addresses for Ghana sellers in the payments form and on the server.
- Keep the existing US protection in place while extending the same validation to Ghana.
- Add coverage for both the direct update path and the browser flow.

## Why
- Stripe’s upcoming Ghana update requires a real beneficiary address, so we need to stop bad addresses before they reach Stripe.
- Adding the check server-side avoids gaps if a request bypasses the UI.

---

This PR was implemented with AI assistance using GPT-5.4.


gianfrancopiana:
@claude review

Issue #4676, opened 2026-04-22
Track CLI users


slavingia:
## What
Track creators who use the Gumroad CLI to understand adoption, usage patterns, and inform CLI development priorities.

## Why
The Gumroad CLI exists but we have no visibility into how many creators use it, which commands they run, or how it fits into their workflows. Without this data we can't prioritize CLI improvements or know if it's worth investing in.

## Open questions
- What events to track? (installs, auth, publish, analytics queries, etc.)
- Where to store the data? (existing analytics pipeline, new table, Metabase-queryable)
- Privacy considerations: what level of granularity is appropriate?
- Should this be opt-in or on by default with an opt-out?

slavingia:
Ershad is taking a look. I think this is pretty important; I think of CLI as potentially a cohort of creators we can focus on in terms of a new type of p-m fit that allows them to really scale their businesses - and then we can use them as examples to find more of them.

(Jordi will be here in NYC in June and we should definitely all hang with him)

gianfrancopiana:
For usage telemetry, we can start with just two generic CLI events:

- `cli.first_run`
- `cli.command_finished`

`cli.command_finished` would carry a small set of fields such as:

- `command_path` like `products.create`, `sales.list`, `auth.login`
- CLI version
- OS/arch
- success/failure
- duration
- output mode
- `dry_run`
- coarse error category

One clarification on `command_path`, since that may sound scarier than it is: I'm not referring to raw argv. I mean only a command identifier like `products.create`. We should not send positional args, raw flag values, file paths, emails, IDs, etc. The command path by itself should be enough to answer "which commands do people run?" without collecting sensitive payloads.

For generic telemetry, the CLI can send command-level events into ClickHouse. For business attribution, successful server-side write actions could be tagged by the backend as `source=cli` and written to ClickHouse with seller/resource linkage attached server-side, so Metabase can answer both usage and impact questions.

So we get a clean split:
- generic CLI telemetry for adoption and usage patterns
- account-linked backend attribution only for questions like "GMV from products created via CLI?"

And the taxonomy stays small:
- one event for adoption over time
- one event for command usage/failures/performance
- backend attribution for the subset of writes where we care about business impact

gumclaw:
## Result of team discussion (Apr 23)

**Decision: Skip generic CLI telemetry. Track only business-impacting actions via existing infrastructure.**

[...]

PR #4714, opened 2026-04-23
Track CLI users via bit flag (no migration)

gumclaw:
## What
Tracks CLI usage across users, products, and offer codes via bit flags. No new tables, no migrations.

## Flags added
| Model | Flag | Bit | Repurposed from |
|-------|------|-----|-----------------|
| User | `has_used_cli` | 30 | `DEPRECATED_has_risk_privilege` |
| Link | `created_via_cli` | 31 | `DEPRECATED_excluded_from_mobile_app_discover` |
| OfferCode | `created_via_cli` | 2 | (new, unused slot) |

## How it works
- `Api::V2::BaseController` detects `gumroad-cli` in User-Agent via `request_from_cli?` helper
- **User flag**: set once on first CLI API call (idempotent, skips if already set)
- **Product flag**: set during `LinksController#create` before save
- **Offer code flag**: set during `OfferCodesController#create` before save
- All fire-and-forget (rescued on failure)

## Query in Metabase
```sql
-- CLI users
SELECT COUNT(*) FROM users WHERE flags & (1 << 29) != 0;

-- Products created via CLI
SELECT COUNT(*) FROM links WHERE flags & (1 << 30) != 0;

-- Offer codes created via CLI
SELECT COUNT(*) FROM offer_codes WHERE flags & (1 << 1) != 0;
```

## TODO for Gianfranco
- [x] Ensure the Gumroad CLI sends User-Agent `gumroad-cli/VERSION`
- [x] Add tests for flag setting in controller specs

Refs #4676

### DIFF
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -65,6 +65,17 @@ class Api::V2::BaseController < ApplicationController
       return unless doorkeeper_token.present?
 
       Rails.logger.info("api v2 user:#{current_resource_owner.id} token:#{doorkeeper_token.id} in #{params[:controller]}##{params[:action]}")
+
+      mark_cli_user
+    end
+
+    def mark_cli_user
+      return unless request.user_agent&.match?(/gumroad-cli/i)
+      return if current_resource_owner.has_used_cli?
+
+      current_resource_owner.update_column(:flags, current_resource_owner.flags | User.flag_mapping["flags"][:has_used_cli])
+    rescue => e
+      Rails.logger.error("Failed to mark CLI user: #{e.message}")
     end
 
     def next_page_url(page_key)
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -80,7 +80,7 @@ class Api::V2::BaseController < ApplicationController
     end
 
     def request_from_cli?
-      request.user_agent&.match?(/gumroad-cli/i)
+      request.user_agent&.match?(/\Agumroad-cli\//i)
     end
 
     def next_page_url(page_key)
--- a/app/controllers/api/v2/base_controller.rb
+++ b/app/controllers/api/v2/base_controller.rb
@@ -70,12 +70,16 @@ class Api::V2::BaseController < ApplicationController
     end
 
     def mark_cli_user
-      return unless request.user_agent&.match?(/gumroad-cli/i)
+      return unless request_from_cli?
       return if current_resource_owner.has_used_cli?
 
       current_resource_owner.update_column(:flags, current_resource_owner.flags | User.flag_mapping["flags"][:has_used_cli])
     rescue => e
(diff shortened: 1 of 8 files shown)


app/controllers/api/v2/base_controller.rb
```diff
@@ -65,6 +65,21 @@ def log_method_use
...
       Rails.logger.info("api v2 user:#{current_resource_owner.id} token:#{doorkeeper_token.id} in #{params[:controller]}##{params[:action]}")
+
+      mark_cli_user
+    end
+
+    def mark_cli_user
+      return unless request_from_cli?
+      return if current_resource_owner.has_used_cli?
+
+      current_resource_owner.update_column(:flags, current_resource_owner.flags | User.flag_mapping["flags"][:has_used_cli])
```
ershad: can we do `current_resource_owner.has_used_cli = true` and `.save!`?

app/controllers/api/v2/base_controller.rb
```diff
@@ -65,6 +65,21 @@ def log_method_use
...
+    rescue => e
+      Rails.logger.error("Failed to mark CLI user: #{e.message}")
+    end
+
+    def request_from_cli?
+      request.user_agent&.match?(/gumroad-cli/i)
```
ershad: cli sends a different use agent:

```
internal/api/client.go:194:
  req.Header.Set("User-Agent", "gumroad/"+c.version)
```

app/models/link.rb
```diff
@@ -37,7 +37,7 @@ class Link < ApplicationRecord
             28 => :is_collab,
             29 => :is_unpublished_by_admin,
             30 => :community_chat_enabled,
-            31 => :DEPRECATED_excluded_from_mobile_app_discover,
```
ershad, line 40: There could be some records with these values set to `true`, we need to set them to `false`.

gianfrancopiana: Yes, there are ~19K affected rows. I can set them to false.

app/controllers/api/v2/base_controller.rb
gianfrancopiana: Yes, I think it makes sense to update the CLI for clarity. I'll do that now.

gianfrancopiana: Done ✅

app/controllers/api/v2/base_controller.rb
gianfrancopiana: I would avoid `save!` here since this runs from an API `after_action` and we only want to flip one analytics bit without validations/callbacks/timestamps or unrelated save failures.

We can do an atomic SQL bit update:

```ruby
mask = User.flag_mapping["flags"][:has_used_cli]
User.where(id: current_resource_owner.id).where("flags & ? = 0", mask).update_all("flags = flags | #{mask}")
```

Thoughts?

app/models/link.rb
gianfrancopiana: Done ✅

gianfrancopiana:
@claude review

PR #4719, opened 2026-04-23
Fetch moderation images as base64 + retry on download failures

gumclaw:
## Problem
OpenAI's image fetcher intermittently fails with 'Failed to download image' errors during content moderation, blocking product publishes.

## Fix 1: Base64 image encoding
Instead of passing image URLs for OpenAI to fetch, we now:
1. Fetch images ourselves via `Net::HTTP`
2. Downscale to 512px max dimension (don't need high res for moderation)
3. Compress to 75% JPEG quality
4. Send as `data:image/jpeg;base64,...` data URIs

This eliminates OpenAI's image fetcher entirely. The ~33% base64 overhead is offset by the downscaling (512px JPEG is small).

New `ContentModeration::ImageEncoder` class handles fetch + resize + encode. Uses MiniMagick (already in the Gemfile).

Applied to both:
- `PromptStrategy` (GPT-4o-mini content evaluation)
- `ClassifierStrategy` (omni-moderation API)

## Fix 2: Retry on download failures
Both strategies now wrap OpenAI calls in a 3x retry loop with exponential backoff (0.5s, 1s, 1.5s). Only retries on 'Failed to download image' errors, re-raises everything else immediately.

## Files changed
- `app/services/content_moderation/image_encoder.rb` (new)
- `app/services/content_moderation/strategies/prompt_strategy.rb`
- `app/services/content_moderation/strategies/classifier_strategy.rb`

### DIFF
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -3,6 +3,8 @@
 class CustomersController < Sellers::BaseController
   include CurrencyHelper
 
+  rescue_from Faraday::TimeoutError, with: :handle_search_timeout
+
 
   before_action :authorize
   before_action :set_on_page_type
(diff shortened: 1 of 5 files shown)


ershad:
This is on hold while we monitor the existing setup.

PR #4720, opened 2026-04-23
Skip unmoderatable image URLs in moderation classifier

ershad:
## What

- Change `ContentModeration::Strategies::ClassifierStrategy#perform` to moderate text and each image in separate OpenAI `moderations` calls.
- Replace `MAX_IMAGES_PER_REQUEST = 1` with `MAX_IMAGES_TO_MODERATE = 5` — the cap on *successful* image moderations per product/post.
- On `Faraday::BadRequestError` for a single image, log `ContentModeration::ClassifierStrategy skipping unmoderatable image URL=… error=…` and skip that URL without consuming cap budget.
- Aggregate `category_scores` across text + all successful image calls, so a real policy violation on any valid image still fails the publish.
- If images exist but every URL is rejected by OpenAI, return `flagged` with the user-facing reason `We cannot moderate the content at this time, please try again later or update the content.` and call `ErrorNotifier.notify` so the all-broken-URLs case pages in Sentry.
- Non-400 errors (timeouts, 5xx, auth, etc.) still bubble up through the existing top-level rescue.
- Spec updates: new coverage for skip-bad-continue-good, flagging-still-works-after-skip, all-images-rejected → flagged + Sentry notify, and one-image-per-request.

## Why

Investigating recent Sentry events for `Faraday::BadRequestError` at `classifier_strategy.rb:38` surfaced two deterministic failure modes that the 1-image cap from #4709 does not solve:

1. **`blob:` URLs in description HTML.** `ContentExtractor#product_image_urls` does not filter non-HTTP schemes, so a browser Object URL saved into a product description (e.g. a paste from an editor with an unresolved upload) leaks through. OpenAI returns `400 Failed to download image`. With many image URLs extracted, there is still a chance per attempt of sampling the bad URL and failing the whole moderation.

2. **Signed attachment URLs from rich-content product files.** `rich_content_file_image_urls` uses `signed_download_url_for_s3_key_and_filename`, which serves content with `Content-Disposition: attachment` and a custom `verify`/`cache_key` scheme. HEAD from our network returns `200 image/png`, but OpenAI's image fetcher returns `400 image_url_unavailable` (intermittent-to-persistent).

Rather than special-casing hosts or rewriting the extractor, this PR makes the classifier robust to any URL OpenAI cannot fetch: skip + log and keep scanning the remaining images until we have 5 successful moderations (or exhaust the list). Moderation decisions remain correct because flagged categories aggregate across all calls that did succeed.

If every image URL is rejected we do *not* silently return compliant — that would be a safety bypass (upload only unreachable images, skip image moderation entirely). Instead we flag with a retry-later message the creator can act on, and raise a Sentry error so we catch patterns of broken URL generation on our side.

Tradeoff: up to 6 API calls per check (1 text + 5 image) instead of 1, and the latency is sequential. Acceptable because moderation already runs in a thread alongside `PromptStrategy`, and the cost per call is low.

## Before/After

N/A — non-user-facing internal service change. Before: any single bad URL sampled from the image set aborted moderation with a 500 and blocked publish. After: bad URLs are logged and skipped; moderation still completes, still flags real violations, and only blocks publish (with a retry message) when *every* image URL is unreachable to OpenAI. I'll record a short walkthrough of the publish path before this merges.

## Test Results

`bundle exec rspec spec/services/content_moderation/strategies/classifier_strategy_spec.rb` — 11 examples, 0 failures.

`bundle exec rspec spec/services/content_moderation/moderate_record_service_spec.rb spec/services/content_moderation/content_extractor_spec.rb` — 16 examples, 0 failures.

`bundle exec rubocop app/services/content_moderation/strategies/classifier_strategy.rb spec/services/content_moderation/strategies/classifier_strategy_spec.rb` — clean.

---

AI disclosure: Written with Claude Opus 4.7 (1M context). Prompts: (1) investigate recent `classifier_strategy.rb:38` 400s via the read-only production console and list the failure categories; (2) change the classifier to skip-and-log URLs that OpenAI rejects and keep scanning up to 5 successful image moderations, failing the check if any valid image is flagged; (3) require at least one successful image moderation when images exist, otherwise return a retry-later user message and notify Sentry as an error; (4) open this PR.

### DIFF
--- a/app/services/content_moderation/strategies/classifier_strategy.rb
+++ b/app/services/content_moderation/strategies/classifier_strategy.rb
@@ -3,7 +3,7 @@
 class ContentModeration::Strategies::ClassifierStrategy
   Result = Struct.new(:status, :reasoning, keyword_init: true)
   OPENAI_REQUEST_TIMEOUT_IN_SECONDS = 10
-  MAX_IMAGES_PER_REQUEST = 1
+  MAX_IMAGES_TO_MODERATE = 5
 
   DEFAULT_THRESHOLDS = {
     "harassment" => 0.8,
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4722, opened 2026-04-23
Fix ActiveRecord::LockWaitTimeout in OAuth access token creation

gumclaw:
## What

`OauthApplication#get_or_generate_access_token` and `ensure_access_grant_exists` used `first_or_create!`, which is not safe under concurrent requests. When two requests arrive simultaneously for the same application/owner, they contend on the same rows and occasionally surface as `ActiveRecord::LockWaitTimeout` from `Oauth::AccessTokensController#create`.

This PR replaces `first_or_create!` with a concurrency-safe helper that:

- looks up the record with `find_by`
- attempts `create!` if none is found
- rescues `ActiveRecord::RecordNotUnique` and re-reads the record created by the concurrent request
- rescues `ActiveRecord::LockWaitTimeout` with a bounded retry + small backoff, then re-raises

## Why

Sentry has been reporting `ActiveRecord::LockWaitTimeout` errors from `Oauth::AccessTokensController#create`. The cause is two concurrent requests racing through `first_or_create!` for the same `access_token`/`access_grant` row: both miss the `first`, both attempt to insert, one blocks on row locks, and the slower one times out. Handling the race explicitly eliminates the timeout and also correctly handles `RecordNotUnique` in case the duplicate slips past MySQL's lock wait.

Sentry: https://gumroad-to.sentry.io/issues/7437171528/

## Before/After

No user-facing change — this is a concurrency bug fix in an internal OAuth code path. A walkthrough of the existing OAuth token flow and the updated method behavior will be attached in a follow-up comment.

## Test Results

Ran the affected spec locally:

```
bundle exec rspec spec/models/oauth_application_spec.rb
```

The pre-existing unrelated S3 test (`#validate_file saves with a valid icon type attached`) fails on `main` as well due to local S3 bucket configuration; it is not affected by this change. All other tests — including 4 new concurrency tests — pass.

---

🤖 AI disclosure: drafted with Claude Opus 4.6. Prompt: fix the Sentry `ActiveRecord::LockWaitTimeout` in `Oauth::AccessTokensController#create`, make `get_or_generate_access_token`/`ensure_access_grant_exists` concurrency-safe, add tests, and open a PR.

### DIFF
--- a/app/models/oauth_application.rb
+++ b/app/models/oauth_application.rb
@@ -61,9 +61,10 @@ class OauthApplication < Doorkeeper::Application
   # Returns an existing active access token or creates one if none exist
   def get_or_generate_access_token
     ensure_access_grant_exists
-    access_tokens.where(resource_owner_id: owner.id,
-                        revoked_at: nil,
-                        scopes: Doorkeeper.configuration.public_scopes.join(" ")).first_or_create!
+    attrs = { resource_owner_id: owner.id,
+              revoked_at: nil,
+              scopes: Doorkeeper.configuration.public_scopes.join(" ") }
+    find_or_create_concurrency_safe(access_tokens, attrs)
   end
 
   def revoke_access_for(user)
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

slavingia:
@claude review once

PR #4725, opened 2026-04-23
Fix ActiveModel::RangeError in DiscountsController for out-of-range integers

gumclaw:
## Problem

When a user submits a discount with an integer value exceeding the 4-byte column limit (e.g. `amount_cents=20000000000`), ActiveRecord raises `ActiveModel::RangeError` before model validations run, resulting in an unhandled 500 error.

**Sentry:** https://gumroad-to.sentry.io/issues/7437458663/

## Fix

Added `rescue_from ActiveModel::RangeError` to `Checkout::DiscountsController` that returns a friendly JSON validation error (`422`) instead of letting it bubble up as a 500. This handles all integer columns at once without requiring schema changes.

## Impact

1 occurrence so far, low impact but unhandled 500 error that returns no useful feedback to the user.

### DIFF
--- a/app/controllers/checkout/discounts_controller.rb
+++ b/app/controllers/checkout/discounts_controller.rb
@@ -5,6 +5,8 @@ class Checkout::DiscountsController < Sellers::BaseController
 
   PER_PAGE = 10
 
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4727, opened 2026-04-24
Fix reviews not loading on custom domains

### LINKED ISSUE #4718: Reviews don't load on custom domains (JSON parsed as HTML)
## Bug

Product reviews fail to load when a product page is accessed via a custom domain. The client-side JS fetches reviews data and expects JSON, but the custom domain proxy (openresty) returns HTML instead, causing a `SyntaxError: Unexpected token '<'`.

## Repro

- **Working:** https://arjynae.gumroad.com/l/bantr-text-to-speech-mac (reviews load)
- **Broken:** https://getbantr.com/l/bantr-text-to-speech-mac (reviews don't load)

Console error: `Uncaught (in promise) SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`

Device/browser agnostic. Reported broken for several weeks.

## Root cause

The custom domain proxy layer (openresty/1.19.9.1) serves the reviews API request as `text/html` instead of proxying the JSON response. Confirmed via curl:

- `arjynae.gumroad.com/product_reviews` → Cloudflare, `content-type: application/json` (404, but correct content type)
- `getbantr.com/product_reviews` → openresty, `content-type: text/html` (200, returns full HTML page)

The openresty proxy appears to match the reviews fetch as a page route and returns the HTML shell instead of forwarding to the API.

## Customer impact

Reviews are a key trust signal for buyers. Any seller using a custom domain has broken reviews right now. This likely affects all custom domains, not just this one.

## Source

Helper ticket: https://help.gumroad.com/conversations?id=e96b1afc33e9e750393fbe99ce5d1f2d

gumclaw:
## Problem
Product reviews fail to load on custom domains. The client-side JS fetches `/product_reviews` expecting JSON, but the custom domain proxy returns HTML because the route isn't matched inside the `UserCustomDomainConstraint` block.

## Root cause
The `product_reviews`, `product_review_responses`, and `product_review_videos` routes were only defined outside the custom domain constraint block in `config/routes.rb`. On a custom domain, the request hits the constraint block first, doesn't find a matching route, falls through to the catch-all, and returns the HTML shell.

## Fix
Added the review routes inside the `UserCustomDomainConstraint` block so they're properly matched for custom domain requests. The existing routes outside the block are kept for `*.gumroad.com` requests.

Fixes #4718

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1146,6 +1146,19 @@ Rails.application.routes.draw do
 
     resources :profile_sections, only: [:create, :update, :destroy]
 
+    put "/product_reviews/set", to: "product_reviews#set", format: :json
+    resources :product_reviews, only: [:index, :show]
+    resources :product_review_responses, only: [:update, :destroy], format: :json
+    resources :product_review_videos, only: [] do
+      scope module: :product_review_videos do
+        resource :stream, only: [:show]
+        resources :streaming_urls, only: [:index]
+      end
+    end
+    namespace :product_review_videos do
+      resource :upload_context, only: [:show]
+    end
+
     get "/", to: "users#show"
   end
 


gumclaw:
@claude review

slavingia:
@claude review

PR #4728, opened 2026-04-24
Handle Stripe::CardError in update_bank_account

gumclaw:
## What

Rescue `Stripe::CardError` separately from the generic `Stripe::StripeError` in `StripeMerchantAccountManager.update_bank_account`.

When the error occurs, we now:
1. Record a payout-note breadcrumb via `record_bank_sync_failure_note`
2. Email the creator via `ContactingCreatorMailer.invalid_bank_account`
3. Return a new `:card_not_supported` symbol
4. Skip `ErrorNotifier.notify` (this is expected third-party behavior, not a bug)

## Why

`Stripe::Token.create` inside `bank_account_hash` can raise `Stripe::CardError` (e.g. "Your card does not support this type of purchase") when a `CardBankAccount`'s underlying card cannot be used as a Stripe Connect external account. Previously this fell through to the generic `Stripe::StripeError` rescue, which logged to Sentry and returned `:stripe_unknown_error` — that return value causes `handle_stripe_info_requirements` to re-enqueue `HandleNewBankAccountWorker`, which then fails again in the same way.

Returning a distinct `:card_not_supported` avoids the retry loop and surfaces the issue to the creator instead of Sentry. ~1 occurrence observed so far.

**Sentry:** https://gumroad-to.sentry.io/issues/7437998801/

## Test Results

Ran `bundle exec rspec spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_spec.rb -e "#update_bank_account"` locally — 11 examples, 0 failures. Verified the new spec fails when the fix is reverted.

---

🤖 AI disclosure: written with Claude Opus 4.6 using a Sentry-triage prompt describing the error signature, the desired handling (payout note + creator email + new symbol), and the location of the rescue chain in `stripe_merchant_account_manager.rb`.

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
@@ -286,6 +286,10 @@ module StripeMerchantAccountManager
 
     ErrorNotifier.notify(e)
     :stripe_invalid_request
+  rescue Stripe::CardError => e
+    record_bank_sync_failure_note(user, e)
+    ContactingCreatorMailer.invalid_bank_account(user.id).deliver_later(queue: "critical")
+    :card_not_supported
   rescue Stripe::StripeError => e
     Rails.logger.error "Stripe error (#{e.class.name}) request ID #{e.request_id} when updating bank account #{bank_account&.id} for stripe account #{stripe_account&.inspect}"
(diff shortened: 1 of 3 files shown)


gumclaw:
@claude review

slavingia:
@claude review

gumclaw:
@claude review once

PR #4740, opened 2026-04-24
Skip 2FA for Google OAuth login

gumclaw:
## Summary
- Skip Gumroad's own 2FA when users log in via Google or Apple OAuth — the provider already enforces 2FA, and the risk of a hacked provider account being used to access Gumroad is minimal.
- 2FA is still required for email/password login.
- Updates the Stripe Connect OAuth flow and the shared `sign_in_with_oauth` helper (used by both Google and Apple) to call `sign_in` directly instead of `sign_in_or_prepare_for_two_factor_auth` and redirect to `two_factor_authentication_path`.

## Test plan
- [x] Updated specs in `spec/controllers/user/omniauth_callbacks_controller_spec.rb` — 2FA-related tests now assert the user is signed in directly and redirected to the destination without passing through `two_factor_authentication_path`.
- [x] `bundle exec rspec spec/controllers/user/omniauth_callbacks_controller_spec.rb -e "when user has 2FA" -e "signs in directly"` — 9 examples, 0 failures.
- [ ] Manually verify Google OAuth login for a user with 2FA enabled lands directly on the dashboard.
- [ ] Manually verify Apple OAuth login for a user with 2FA enabled lands directly on the dashboard.
- [ ] Manually verify email/password login for a user with 2FA enabled still requires the 2FA code.

### DIFF
--- a/app/controllers/user/omniauth_callbacks_controller.rb
+++ b/app/controllers/user/omniauth_callbacks_controller.rb
@@ -65,13 +65,8 @@ class User::OmniauthCallbacksController < Devise::OmniauthCallbacksController
         create_user_event("signup")
       end
 
-      if user.email.present?
-        sign_in_or_prepare_for_two_factor_auth(user)
-        return redirect_to two_factor_authentication_path(next: oauth_completions_stripe_path)
-      else
-        sign_in user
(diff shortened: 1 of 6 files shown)


ershad (approved):
Looks good, okay to merge on green.

Issue #4677, opened 2026-04-22
Gumroad admin CLI


slavingia:
## What
Expose Gumroad admin functionality through an API and/or CLI, so admin operations don't require the web UI.

## Why
Admin actions currently require logging into the web UI. This blocks automation (e.g., Gumclaw can't perform admin actions programmatically), slows down bulk operations, and makes it harder to build tooling on top of admin features. An API/CLI would let us automate risk reviews, user management, fee updates, and other admin tasks.

## Open questions
- Which admin actions to prioritize? (user lookup, fee changes, risk state changes, compliance review)
- Auth model: API keys, OAuth, or reuse existing admin session?
- Should this extend the existing v2 API or be a separate admin API?
- CLI: standalone tool or extend the existing Gumroad CLI?

Darshan3690:
👋HI @slavingia sir

I’d like to work on this issue. The idea of exposing admin functionality via API/CLI looks very impactful, and I’m interested in implementing it.

Proposed approach:

Introduce a separate Admin API layer (instead of extending public APIs) to keep permissions isolated
Start with core endpoints like user lookup, fee updates, and risk state changes
Implement API key–based authentication with RBAC + audit logs for safety
Extend the existing CLI to consume these APIs for automation use cases

I can take this in phases (read → write → automation) to ensure safe rollout.

Please assign this issue to me 🙌

gumclaw:
## Meeting Notes (Apr 23, 2026)

### Architecture

- Integrate admin functionality into the existing Gumroad CLI rather than building separate tooling
- Keep Helper for ticket operations, CLI for admin operations
- Reuse existing CLI patterns (JSON output, JQ flags, skill model)
- Namespace: `gumroad admin [command]` to separate from user-facing CLI
- Authentication: start with simple token, make robust later

### Phase 1 Scope

- Add current admin UI capabilities to CLI
- Known gaps to fix: search returns only one result, email lookup not robust, refund flow needs adjustment, contract spec doesn't match actual API
- Estimated ~12 PRs total

### Technical Decisions

- Remove OpenAPI spec approach (current Helper tool integration too complicated, hard to maintain)
- Long-term: Helper UI becomes minimal/Gmail wrapper. CLI handles admin operations.
- Start as separate internal tool, merge into main CLI later if viable
- Research other CLIs with admin/user separation (GitHub CLI as reference)

### Next Steps

- Gianfranco finishing upload feature, then starting admin CLI development
[...]

PR #4742, opened 2026-04-24
Remove the Helper OpenAPI admin contract

gianfrancopiana:
Refs #4677

## What

Removes the generated Helper Tools OpenAPI endpoint from Gumroad and stops treating `/internal/helper` as the admin CLI contract.

The concrete `/internal/helper/...` routes stay available for existing Helper usage, but the generated schema, OpenAPI metadata, and stale `refresh_library` tool path are gone. The new internal admin API note makes the boundary explicit: upcoming PR will build against `/internal/admin` instead of the legacy Helper route shape.

## Why

The current Helper OpenAPI import path is too hard to maintain as the source of truth for admin automation. It had already drifted from real routed behavior, including a `refresh_library` path that was not actually supported.

This keeps the existing Helper integration working while removing the misleading contract before the CLI starts depending on it. 

## Rollout

Helper production has already been moved off `https://api.gumroad.com/internal/helper/openapi`: the Gumroad tool API now uses a stored schema, and the stale `post-purchases-refresh-library` tool is disabled. No Helper code PR is required for runtime correctness.

---

This PR was implemented with AI assistance using gpt‑5.5 xhigh


gumclaw:
@claude review

gianfrancopiana:
@claude review

PR #4744, opened 2026-04-24
Filter unsupported image formats in ContentModeration::PromptStrategy

gumclaw:
## Problem

`PromptStrategy#build_messages` sends image URLs directly to OpenAI's chat API without checking if the format is supported. When unsupported formats like `.psd` are included, OpenAI returns a 400 `invalid_image_url` error. Since all images are batched in one request, one bad URL fails the entire moderation call and generates unnecessary Sentry noise.

**Sentry:** https://gumroad-to.sentry.io/issues/7439574034/

## Fix

Pre-filter image URLs to only include OpenAI-supported formats (PNG, JPEG, GIF, WebP) before sampling and sending. If all URLs are filtered out, logs a warning and proceeds with text-only evaluation.

## Changes
- `prompt_strategy.rb`: add `SUPPORTED_IMAGE_EXTENSIONS` constant and `supported_image_url?` method, filter URLs in `build_messages`
- `prompt_strategy_spec.rb`: three new tests for unsupported format filtering, all-filtered warning, and supported format passthrough

### DIFF
--- a/app/services/content_moderation/strategies/prompt_strategy.rb
+++ b/app/services/content_moderation/strategies/prompt_strategy.rb
@@ -32,6 +32,7 @@ class ContentModeration::Strategies::PromptStrategy
 
   MODEL = "gpt-4o-mini"
   JUDGE_MODEL = "gpt-4o-mini"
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review once

PR #4748, opened 2026-04-24
Fix ArgumentError in Rack::Attack: use RedisProxy instead of RedisStoreProxy

gumclaw:
## Problem

`Rack::Attack` was crashing with `ArgumentError: wrong number of arguments (given 4, expected 3)` because it used `RedisStoreProxy` with a plain `Redis` client.

`RedisStoreProxy.setex` passes `raw: true` as a 4th argument, but the plain `redis` gem (5.x) `setex(key, ttl, value)` only accepts 3 positional args.

## Fix

One-line change: `RedisStoreProxy` → `RedisProxy`, which is designed for plain `Redis` clients and doesn't pass the extra `raw:` kwarg.

Also added `.codex/` to `.gitignore` to prevent codex worktree directories from being committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ test-branch-creation.log
 .vercel
 .claude/worktrees/
 autoresearch.*
+.codex/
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

PR #4749, opened 2026-04-24
Fix RSA encryption error for oversized payment field inputs

gumclaw:
## Problem

When a user submits an abnormally long value for encrypted payment fields (account number, tax IDs), Strongbox attempts RSA encryption directly (`symmetric: :never`) and raises:

```
OpenSSL::PKey::RSAError: EVP_PKEY_encrypt: data too large for key size
```

RSA with PKCS1 padding can only encrypt up to `(key_size_bytes - 11)` bytes. The error occurs in `Settings::PaymentsController#update`.

**Sentry:** https://gumroad-to.sentry.io/issues/7440029412/
**Occurrences:** 1 (first seen 2026-04-24, low frequency edge case)

## Fix

Add length validation (max 200 chars) before the values reach the encryption layer:

- `UpdatePayoutMethod`: validates `account_number` length before assignment
- `UpdateUserComplianceInfo`: validates `individual_tax_id` and `business_tax_id` length before entering the `dup_and_save` block

200 chars is well under the RSA capacity while being generous for any real account number or tax ID (IBAN max is 34, SSN/EIN is 9).

## Tests

- Added spec for oversized account number in `update_payout_method_spec.rb`
- New spec file `update_user_compliance_info_spec.rb` covering oversized `individual_tax_id`, `business_tax_id`, and `ssn_last_four`


gumclaw:
@claude review

PR #4750, opened 2026-04-25
Fix ClassifierStrategy to skip timed-out images instead of crashing

gumclaw:
## Problem

In `ClassifierStrategy#moderate`, `Faraday::TimeoutError` after exhausting all retry attempts re-raises the error even for image moderation. This means a single timing-out image crashes the entire moderation flow, preventing all remaining images from being checked.

When ALL images time out, the `ErrorNotifier` fires "could not moderate any image" ([Sentry](https://gumroad-to.sentry.io/issues/7440278724/)).

## Fix

When `Faraday::TimeoutError` exhausts `MAX_MODERATION_ATTEMPTS` and `skip_url` is present (image moderation), log a warning and return `nil` instead of re-raising. This matches the existing `Faraday::BadRequestError` handling: the image is skipped and remaining images continue to be moderated.

Text moderation timeouts (`skip_url` is nil) still re-raise as before.

## Tests added

- Image timeout after all retries skips the image, remaining images still moderated
- All images timing out triggers the ErrorNotifier (existing "could not moderate any image" path)
- Text moderation timeout still re-raises (existing behavior preserved)

### DIFF
--- a/app/services/content_moderation/strategies/classifier_strategy.rb
+++ b/app/services/content_moderation/strategies/classifier_strategy.rb
@@ -93,7 +93,9 @@ class ContentModeration::Strategies::ClassifierStrategy
           Rails.logger.warn("ContentModeration::ClassifierStrategy error on attempt #{attempts}/#{MAX_MODERATION_ATTEMPTS}, retrying: #{e.message}")
           retry
         end
-        raise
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

gumclaw:

PR #4756, opened 2026-04-25
Fix DashboardController#index Rack::Timeout by limiting tax form year range

gumclaw:
## Problem

`DashboardController#index` is hitting `Rack::Timeout::RequestTimeoutException` (120s) for some sellers.

**Root cause:** The legacy (non-tax-center) code path in `CreatorHomePresenter` iterates from the current year back to `seller.created_at.year`, running multiple expensive 1099 eligibility queries per year (`eligible_for_1099?` calls `sales_scope_for(year).sum` and `.count` multiple times). For a seller created in 2011, that's 15 years × ~5 DB queries = 75+ queries in a single request.

**Sentry:** https://gumroad-to.sentry.io/issues/7440505757/

## Fix

Limit the tax form iteration to the last 4 years (`Time.current.year - 3` to `Time.current.year`). This covers all practically relevant 1099 download periods while capping the query count to a constant ~20 regardless of account age.

Sellers on the newer `tax_center` path are unaffected (they use `user_tax_forms` records instead of the per-year loop).

## Tests

Added a test verifying long-lived accounts (10+ years) are limited to 4 years of tax form lookups. All existing tax form tests pass.

### DIFF
--- a/app/presenters/creator_home_presenter.rb
+++ b/app/presenters/creator_home_presenter.rb
@@ -78,7 +78,11 @@ class CreatorHomePresenter
       tax_forms = []
       show_1099_download_notice = seller.user_tax_forms.for_year(Time.current.prev_year.year).exists?
     else
-      tax_forms = (Time.current.year.downto(seller.created_at.year)).each_with_object({}) do |year, hash|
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

PR #4758, opened 2026-04-25
Rescue network errors in TaxIdValidationService

gumclaw:
## Problem

`CustomerSurchargeController#calculate_all` crashes with `Net::OpenTimeout` when the taxid.pro API is unreachable. This breaks the checkout surcharge calculation for buyers in countries that use TaxIdPro validation.

[Sentry issue](https://gumroad-to.sentry.io/issues/7440627359/)

## Root Cause

`TaxIdValidationService#valid_tax_id?` makes an HTTP call to `v3.api.taxid.pro` with a 5-second timeout, but doesn't rescue network errors. When the API host is down or unreachable, the exception propagates up through `SalesTaxCalculator` → `CustomerSurchargeController`.

## Fix

Rescue common network errors (`Net::OpenTimeout`, `Net::ReadTimeout`, `Errno::ECONNREFUSED`, `Errno::EHOSTUNREACH`, `SocketError`) and return `false`. This is the conservative default: treat the tax ID as invalid so tax is still charged. The result is cached for 10 minutes, so a transient outage won't cause repeated failed requests.

## Tests

Added specs confirming that timeout and connection-refused errors return `false` instead of raising.

### DIFF
--- a/app/services/tax_id_validation_service.rb
+++ b/app/services/tax_id_validation_service.rb
@@ -29,5 +29,7 @@ class TaxIdValidationService
     def valid_tax_id?
       response = HTTParty.get(TAX_ID_PRO_ENDPOINT_TEMPLATE.expand(country_code:, tax_id:).to_s, headers: TAX_ID_PRO_HEADERS, timeout: 5)
       response.code == 200 && response.parsed_response["is_valid"]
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

gumclaw:

PR #4759, opened 2026-04-25
Fix Rack::Timeout in ThumbnailsController#create with 30s analyze guard

gumclaw:
## Problem

`ThumbnailsController#create` calls `thumbnail.file.analyze` synchronously during the request. This downloads the blob from storage and runs image analysis (ImageMagick/MiniMagick). For large uploads or when storage is slow, this easily exceeds the 120-second Rack::Timeout, which sends SIGTERM and kills the entire worker process.

[Sentry issue](https://gumroad-to.sentry.io/issues/7441113843/)

## Fix

Wrap the `analyze` call in a 30-second `Timeout.timeout` block. If analysis takes too long, the request returns a user-friendly JSON error instead of timing out and killing the process.

The 30s limit is generous for image analysis (typically <5s) but well within the 120s Rack::Timeout, leaving headroom for the rest of the request lifecycle.

## Changes

- `app/controllers/thumbnails_controller.rb`: Wrap `analyze` in `Timeout.timeout(30)`, add `rescue Timeout::Error` handler
- `spec/controllers/thumbnails_controller_spec.rb`: Add test for timeout scenario

### DIFF
--- a/app/controllers/thumbnails_controller.rb
+++ b/app/controllers/thumbnails_controller.rb
@@ -14,7 +14,7 @@ class ThumbnailsController < Sellers::BaseController
 
     if permitted_params[:signed_blob_id].present?
       thumbnail.file.attach(permitted_params[:signed_blob_id])
-      thumbnail.file.analyze
+      Timeout.timeout(30) { thumbnail.file.analyze }
       thumbnail.unsplash_url = nil
     end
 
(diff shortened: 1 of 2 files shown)


gumclaw:
CI failed with unrelated flaky tests (CSV export apostrophe mismatch, Capybara embed/checkout timing). Re-running CI.

slavingia:
@claude review once

PR #4761, opened 2026-04-25
Fix NoMethodError on pluck for ActionController::Parameters in workflow emails

gumclaw:
## What

Fixes `NoMethodError: undefined method 'pluck' for an instance of ActionController::Parameters` in `Workflows::EmailsController#update` (Sentry).

In `Workflow::SaveInstallmentsService#delete_removed_installments`, `params[:installments].pluck(:id)` assumed `params[:installments]` is always an Array. When the frontend submits installments as an object with numeric string keys (e.g. `{"0" => {...}, "1" => {...}}`), Rails parses it as an `ActionController::Parameters` instance (hash-like), which doesn't respond to `pluck`.

## Why

`ActionController::Parameters` doesn't have `pluck` (that's an ActiveRecord method). The same hash-like params also broke `.each` iteration (yielding key-value pairs instead of just values) and `.size` semantics.

The fix normalizes `params[:installments]` to an array once at the top of `process`:
- If it's an `ActionController::Parameters` (hash-like): call `.values` to extract the array of installment param objects
- Otherwise: wrap with `Array()` for safety

All three call sites (`.size`, `.each`, `.map { |i| i[:id] }`) now use the normalized `@installments_array`.

## Test results

Added a spec that passes installments as hash-like `ActionController::Parameters` to confirm the fix.

---

AI disclosure: Implemented with Claude Opus 4.6. Prompt: fix the Sentry NoMethodError on pluck in workflow emails controller, with root cause analysis and fix strategy provided.

### DIFF
--- a/app/services/workflow/save_installments_service.rb
+++ b/app/services/workflow/save_installments_service.rb
@@ -21,7 +21,9 @@ class Workflow::SaveInstallmentsService
       return [false, errors]
     end
 
-    if workflow.abandoned_cart_type? && params[:installments].size != 1
+    @installments_array = params[:installments].is_a?(ActionController::Parameters) ? params[:installments].values : Array(params[:installments])
+
+    if workflow.abandoned_cart_type? && @installments_array.size != 1
       workflow.errors.add(:base, "An abandoned cart workflow can only have one email.")
       @errors = workflow.errors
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review once

PR #4764, opened 2026-04-25
Fix Errno::ENOENT from temp file cleanup during image variant processing

gumclaw:
## What

Add `Errno::ENOENT` to rescue clauses in `Thumbnail#url` and `User#resized_avatar_url` so they gracefully handle temp file race conditions during ActiveStorage variant processing.

## Why

[Sentry issue](https://gumroad-to.sentry.io/issues/7441441718/): The `image_processing` gem creates temp files in `/tmp/` during `.variant(...).processed` calls. A race condition can occur when a temp file is cleaned up (by GC, OS tmpwatch, or concurrent processing) before `File.size` reads it, raising `Errno::ENOENT`. This error was not caught by existing rescue clauses that only handled `MiniMagick::Error`, `ActiveStorage::Error`, and `ActiveStorage::FileNotFoundError`.

The fix adds `Errno::ENOENT` to the rescue clauses so both methods fall back to their safe defaults (original file URL for thumbnails, default avatar for users), matching the existing error-handling behavior.

### Changes
- `Thumbnail#url`: rescue `Errno::ENOENT` alongside `MiniMagick::Error` and `ActiveStorage::Error`
- `User#resized_avatar_url`: rescue `Errno::ENOENT` alongside `ActiveStorage::FileNotFoundError`
- Added specs for both methods verifying the fallback behavior

## Test results

New specs follow the same pattern as the existing error-handling specs (stub the variant call to raise the error, assert the fallback URL is returned).

---

AI disclosure: This PR was authored with Claude Opus 4.6. Prompted to fix the Sentry issue by adding `Errno::ENOENT` to rescue clauses in variant processing methods and adding corresponding specs.

### DIFF
--- a/app/models/thumbnail.rb
+++ b/app/models/thumbnail.rb
@@ -52,7 +52,7 @@ class Thumbnail < ApplicationRecord
     else
       cdn_url_for(file.url)
     end
-  rescue MiniMagick::Error, ActiveStorage::Error => e
+  rescue MiniMagick::Error, ActiveStorage::Error, Errno::ENOENT => e
     Rails.logger.warn("Thumbnail#url error (#{id}): #{e.class} => #{e.message}")
     cdn_url_for(file.url)
   end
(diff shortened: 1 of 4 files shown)


slavingia:
@claude review once

PR #4766, opened 2026-04-25
Fix deadlock in CheckoutController#update by locking cart row

gumclaw:
## What

Adds a `SELECT ... FOR UPDATE` row lock on the `Cart` record inside `CheckoutController#update`'s transaction, and rescues `ActiveRecord::Deadlocked` with a friendly redirect.

## Why

Two concurrent `PATCH /checkout` requests for the same cart can interleave their INSERT/UPDATE statements on `cart_products`, causing MySQL gap-lock deadlocks ([Sentry issue](https://gumroad-to.sentry.io/issues/7441623745/)). The transaction fetches/creates a Cart and iterates over cart products without holding a row lock, so concurrent requests aren't serialized.

`cart.lock!` after `cart.save!` acquires a `FOR UPDATE` lock that serializes concurrent updates to the same cart row. The `ActiveRecord::Deadlocked` rescue is a safety net so any remaining edge-case deadlock returns a redirect instead of a 500.

## Changes

- `app/controllers/checkout_controller.rb`: Added `cart.lock!` after `cart.save!` (line 38). Combined `ActiveRecord::Deadlocked` into the existing rescue clause.
- `spec/controllers/checkout_controller_spec.rb`: Two new tests — one verifying the lock is acquired, one verifying the deadlock rescue path redirects gracefully.

## Test results

Tests cannot run locally outside Docker. The two new specs verify:
1. `cart.lock!` is called during update
2. `ActiveRecord::Deadlocked` is rescued and redirects with an alert

---

Built with Claude Opus 4.6. Prompts: "Fix a MySQL deadlock in CheckoutController#update by adding cart.lock! and rescuing ActiveRecord::Deadlocked."

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -35,6 +35,7 @@ class CheckoutController < ApplicationController
       cart.reject_ppp_discount = update_permitted_params[:rejectPppDiscount] || false
       cart.discount_codes = update_permitted_params[:discountCodes].to_a.map { { code: _1[:code], fromUrl: _1[:fromUrl] } }
       cart.save!
+      cart.lock!
 
       updated_cart_products = items.map do |item|
         product = Link.find_by_external_id!(item[:product][:id])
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review once

PR #4767, opened 2026-04-25
Fix blank referrer validation error in cross-sell checkout

gumclaw:
## What

`CrossSellModal.tsx` was setting `referrer: ""` when building a cross-sell cart item. The `CartProduct` model validates referrer presence (`validates :referrer, presence: true`), so an empty string caused `ActiveRecord::RecordInvalid: Validation failed: Referrer can't be blank` during `CheckoutController#update`.

## Why

This was a Sentry production error. The fix uses `getReferrer()` from `$app/data/user_action_event` — the same utility used everywhere else in the codebase — which returns the URL `referrer` param, `document.referrer`, or `"direct"` as a fallback.

## Changes

- **`app/javascript/components/Checkout/CrossSellModal.tsx`**: Import `getReferrer` and use it instead of `""` for the cross-sell cart item referrer field.
- **`spec/models/cart_product_spec.rb`**: Added referrer validation tests confirming blank referrer is invalid and present referrer is valid.

## Test results

- ESLint passes on the changed file
- New spec tests validate the referrer presence constraint on `CartProduct`

---

Built with Claude Opus 4.6. Prompt: "Fix Sentry bug: ActiveRecord::RecordInvalid for blank referrer in CrossSellModal cross-sell checkout flow."

### DIFF
--- a/app/javascript/components/Checkout/CrossSellModal.tsx
+++ b/app/javascript/components/Checkout/CrossSellModal.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 
+import { getReferrer } from "$app/data/user_action_event";
 import { formatOrderOfMagnitude } from "$app/utils/formatOrderOfMagnitude";
 
 import { Button } from "$app/components/Button";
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

PR #4777, opened 2026-04-26
Fix Aws::S3::Errors::NotFound crash during product publish

gumclaw:
## What

Rescue `Aws::S3::Errors::NotFound` in `ContentModeration::ContentExtractor#product_image_urls` so that missing S3 objects in rich content files are skipped instead of crashing the publish action.

## Why

When a product has a rich content image file whose S3 object has been deleted, publishing the product crashes with `Aws::S3::Errors::NotFound`. This happens because `signed_download_url_for_s3_key_and_filename` does a HEAD request to S3 via `obj.content_length`, which raises when the object doesn't exist. The error propagates through the `content_moderation_check` validation → `save!` → `publish!` → caught by the generic `rescue => e` in the controller → reported to Sentry.

Sentry issue: https://gumroad-to.sentry.io/issues/7442366748/

The fix catches `Aws::S3::Errors::NotFound` at the point of URL generation and skips the missing file, allowing the rest of the content moderation check to proceed normally. This is consistent with how `Product::AsJson` already handles missing S3 objects.

## Test results

Added two test cases verifying that:
1. When a product file's S3 object is missing, the extractor skips it without raising
2. Remaining valid image URLs are still collected

Tests fail when the fix is reverted, confirming validity.

---

Built with Claude Opus 4.6. Prompts: fix S3 NotFound crash in LinksController#publish per issue specification.

### DIFF
--- a/app/services/content_moderation/content_extractor.rb
+++ b/app/services/content_moderation/content_extractor.rb
@@ -34,8 +34,10 @@ class ContentModeration::ContentExtractor
       rich_contents = product.alive_rich_contents
 
       rich_content_file_image_urls = rich_contents.flat_map do |rich_content|
-        ProductFile.where(id: rich_content.embedded_product_file_ids_in_order, filegroup: "image").map do
-          signed_download_url_for_s3_key_and_filename(_1.s3_key, _1.s3_filename, expires_in: 1.hour)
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review once

PR #4783, opened 2026-04-27
Fix reviews not rendering on product-level custom domains

### LINKED ISSUE #4781: Reviews not rendering on product-level custom domains
## Bug

Reviews/ratings do not render on product pages accessed via a product-level custom domain, even though the data is present in the server response.

## Steps to reproduce

1. Visit a product via its Gumroad URL (e.g. `username.gumroad.com/l/shortcode`) — reviews show correctly
2. Visit the same product via its product-level custom domain — reviews do not render

## Investigation

Compared the raw HTML from both URLs. The server-side Inertia payload is identical on both pages:
- `"ratings":{"count":1,"average":5.0,...}` present in both
- `"show_ratings_filter":true` present in both
- Structured data includes `aggregateRating` in both

The data is being sent correctly to the custom domain page. The issue is client-side: the JS framework receives the data but does not render the reviews component on the custom domain.

**Possible cause:** The Gumroad URL serves the page at `/l/<shortcode>` while the custom domain serves it at `/` (root path). The path difference may be causing the reviews component to not mount.

## Context

- Affects product-level custom domains (not profile-level)
- Reported by a seller whose domain was recently verified
- Seller says reviews were showing previously and stopped in the last week

gumclaw:
## Problem
Reviews/ratings do not render on product pages accessed via a product-level custom domain, even though the data is present in the server response. The client-side JS fetches `/product_reviews` expecting JSON, but the product custom domain returns HTML because the review routes aren't defined inside the `ProductCustomDomainConstraint` block.

## Root cause
[PR #4727](https://github.com/antiwork/gumroad/pull/4727) added review routes to the `UserCustomDomainConstraint` block (profile-level custom domains) but missed the `ProductCustomDomainConstraint` block (product-level custom domains). The `product_reviews`, `product_review_responses`, and `product_review_videos` routes were only available for profile-level custom domains and `*.gumroad.com` requests.

## Fix
Added the same review routes inside the `ProductCustomDomainConstraint` block, placed before the catch-all product routes so they match before falling through to the HTML shell.

Also added a system test to verify reviews render correctly on product-level custom domains.

Fixes #4781

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1041,6 +1041,20 @@ Rails.application.routes.draw do
   constraints ProductCustomDomainConstraint do
     get "/.well-known/acme-challenge/:token", to: "acme_challenges#show"
     product_tracking_routes(named_routes: false)
+
+    put "/product_reviews/set", to: "product_reviews#set", format: :json
+    resources :product_reviews, only: [:index, :show]
+    resources :product_review_responses, only: [:update, :destroy], format: :json
+    resources :product_review_videos, only: [] do
+      scope module: :product_review_videos do
+        resource :stream, only: [:show]
+        resources :streaming_urls, only: [:index]
+      end
+    end
+    namespace :product_review_videos do
+      resource :upload_context, only: [:show]
+    end
+
     get "/", to: "links#show", defaults: { format: "html" }
     get "/l/:id", to: "links#show", defaults: { format: "html" }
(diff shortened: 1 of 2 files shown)


nyomanjyotisa:
@claude review

gumclaw:
@claude review

PR #4786, opened 2026-04-27
Fix coffee product limit bypassed by unarchiving

nyomanjyotisa:
## What

Fixes a bug where sellers could have more than one active coffee product by archiving the original, creating a new one, then unarchiving the original.

- Run `one_coffee_per_user` validation on unarchive, not just on create
- Return 422 with validation messages from the unarchive endpoint
- Surface the actual error in the products dashboard alert

## Why

The `one_coffee_per_user` validation was scoped to `on: :create`, so the limit was only enforced at creation time. Unarchiving used `update!`, which skipped the check entirely and crashed with a 500 if it had been enforced.

The validation now also runs when `archived_changed? && !archived?`. The guard is intentionally narrow so unrelated updates (price, name, description) on a coffee product don't re-run the check.

## After

https://github.com/user-attachments/assets/fe317637-45be-4368-92e9-404349a255bc

## QA steps

1. Sign in as a seller eligible for service products
2. Create a coffee product (Coffee A)
3. Archive Coffee A from the products dashboard
4. Create a second coffee product (Coffee B)
5. Go to the archived products page and try to unarchive Coffee A
6. Confirm an error toast appears with "You can only have one coffee product." and Coffee A remains archived
7. Delete Coffee B, then unarchive Coffee A — should succeed

---

AI disclosure: drafted with Claude Opus 4.7

### DIFF
--- a/app/controllers/products/archived_controller.rb
+++ b/app/controllers/products/archived_controller.rb
@@ -31,8 +31,11 @@ class Products::ArchivedController < Sellers::BaseController
   def destroy
     authorize [:products, :archived, @product]
 
-    @product.update!(archived: false)
-    render json: { success: true, archived_products_count: current_seller.archived_products_count }
+    if @product.update(archived: false)
+      render json: { success: true, archived_products_count: current_seller.archived_products_count }
+    else
+      render json: { success: false, errors: @product.errors.full_messages }, status: :unprocessable_entity
+    end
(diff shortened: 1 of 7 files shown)


gumclaw:
@claude review

nyomanjyotisa:
@claude review

PR #4791, opened 2026-04-27
Fix Faraday::ParsingError in content moderation classifier retries

gumclaw:
## Problem

When OpenAI's moderation API returns a non-JSON response (e.g., an upstream error page), Faraday's JSON parser raises `Faraday::ParsingError`. This exception was unhandled in `ClassifierStrategy#moderate`, causing `LinksController#publish` to crash.

**Sentry:** https://gumroad-to.sentry.io/issues/7444490016/
**Events (7d):** 1 | **Events (14d):** 1 | **Users affected:** ~1/month
**Estimated support tickets:** 0 (transient upstream error, user can retry)

## Fix

Added `Faraday::ParsingError` to the existing retry rescue clause alongside `Faraday::TimeoutError`. Non-JSON responses from OpenAI are transient upstream errors that should be retried, not fatal crashes.

Also updated the log message from "timeout" to "error" since the rescue now covers more than just timeouts.

## Tests

- Added spec: retries on `ParsingError` and succeeds when a subsequent attempt returns valid JSON
- Added spec: re-raises after exhausting all retry attempts with `ParsingError`
- Updated existing timeout specs to match the new log message wording

### DIFF
--- a/app/services/content_moderation/strategies/classifier_strategy.rb
+++ b/app/services/content_moderation/strategies/classifier_strategy.rb
@@ -88,9 +88,9 @@ class ContentModeration::Strategies::ClassifierStrategy
         attempts += 1
         response = @client.moderations(parameters: { model: "omni-moderation-latest", input: input })
         response.dig("results", 0, "category_scores") || {}
-      rescue Faraday::TimeoutError => e
+      rescue Faraday::TimeoutError, Faraday::ParsingError => e
         if attempts < MAX_MODERATION_ATTEMPTS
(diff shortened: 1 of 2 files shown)


gumclaw:
CI is failing across all shards with the known Shakapacker infra issue (`Shakapacker can't find design.css in manifest.json` — stale Docker cache). Already retried once with the same result. The fix itself is a 2-line change and tests pass locally. This needs a Docker image rebuild to unblock CI.

gumclaw:
@claude review

PR #4792, opened 2026-04-27
Add internal admin read API foundation

gianfrancopiana:
Refs #4677

## What

Adds the first explicit `/internal/admin` API surface for `gumroad admin` read commands. The new endpoints support exact purchase lookup, license-key lookup, user suspension status, and recent payout reads behind a separate admin bearer token.

The API keeps sensitive lookup inputs in JSON request bodies and leaves the legacy `/internal/helper` routes as-is for Helper.

## Why

The CLI already targets `/internal/admin`, but Gumroad did not yet expose that server contract. This adds the narrow server foundation needed before purchase search, refunds, and write commands can build on the new admin API without falling back to `/internal/helper`.

This reuses existing Gumroad data and behavior where it is already correct, while giving the CLI a clearer route and response contract.

---


gianfrancopiana:
@claude review

PR #4794, opened 2026-04-27
Fix restart checkout card params from orders

gianfrancopiana:
## What

Fix the expired subscription restart checkout path so a newly entered card is still recognized when `/orders` passes checkout params through controller strong params. `Order::CreateService` now normalizes the extracted card params before handing them to the restart flow, and `RestartAtCheckoutService` accepts indifferent-access params before deciding whether to reuse the saved card. The regression specs now exercise the real strong-params shape that hit production.

## Why

This follows up on [#4697](https://github.com/antiwork/gumroad/pull/4697). That first restart-card fix handled symbol-keyed params, but the real browser/controller path can send these fields as string-keyed `ActionController::Parameters`. In that case the restart flow treated the request as if no new card had been entered and silently reused the saved card, which matches the support report. Normalizing the params at the boundary fixes the bug without changing the rest of the restart logic.

---

This PR was implemented with AI assistance using gpt‑5.5 xhigh

### DIFF
--- a/app/services/order/create_service.rb
+++ b/app/services/order/create_service.rb
@@ -58,7 +58,7 @@ class Order::CreateService
         card_params = common_params.slice(
           :card_data_handling_mode, :stripe_payment_method_id, :paypal_order_id,
           :stripe_customer_id, :stripe_setup_intent_id
-        ).compact
+        ).to_h.symbolize_keys.compact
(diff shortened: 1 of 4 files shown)


gianfrancopiana:
@claude review

PR #4798, opened 2026-04-27
Guard WithMaxExecutionTime ensure block against broken connections

gumclaw:
## Problem

When a query wrapped in `WithMaxExecutionTime.timeout_queries` exceeds its time limit and the MySQL connection is left in a bad state, the `ensure` block attempts to restore `max_execution_time` on a broken connection. This raises a **new** exception that **masks the original `QueryTimeoutError`**, making production debugging significantly harder.

## Fix

Wrap the `ensure` block in `begin/rescue` to catch `ActiveRecord::StatementInvalid` and `Mysql2::Error`, logging the failure instead of letting it propagate and mask the real error.

## Tests

Added two specs:
- Verifies `QueryTimeoutError` propagates to the caller even when the ensure block fails
- Verifies the restoration failure is logged via `Rails.logger.error`

## Impact

- **Sentry events (7d):** 1
- **Sentry events (14d):** 1  
- **Estimated affected users/month:** Low (transient, but when it does happen the masked error makes debugging much harder)
- **Estimated support tickets/month:** 0 (background job context)

Sentry: https://gumroad-to.sentry.io/issues/7445166358/

### DIFF
--- a/lib/utilities/with_max_execution_time.rb
+++ b/lib/utilities/with_max_execution_time.rb
@@ -18,6 +18,10 @@ module WithMaxExecutionTime
       raise
     end
   ensure
-    connection.execute("set max_execution_time = #{previous_max_execution_time}")
+    begin
+      connection.execute("set max_execution_time = #{previous_max_execution_time}")
+    rescue ActiveRecord::StatementInvalid, Mysql2::Error => e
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review once

gumclaw:
@claude review

PR #4802, opened 2026-04-27
Fix ActiveRecord::Deadlocked in LinksController#publish

gumclaw:
## What

Adds deadlock retry logic to `Link#publish!`. The `save!` call (line 423) is now wrapped with up to 2 retries on `ActiveRecord::Deadlocked`, re-raising if all retries are exhausted.

## Why

**Sentry**: https://gumroad-to.sentry.io/issues/7445277375/

| Metric | Value |
|--------|-------|
| 7-day events | 1 |
| 30-day events | 1 |
| Users affected | ~1/month |
| Estimated support tickets | 0 |

The `save!` in `publish!` triggers the `after_save :set_customizable_price` callback (defined in `app/modules/product/prices.rb`), which performs a JOIN query on `variant_categories`/`variants` and then calls `update_column(:customizable_price, true)`. When concurrent transactions lock these rows in different orders, MySQL raises a deadlock error. There was no retry logic, so the user saw a generic 500 error.

The fix follows the existing pattern used in `app/models/concerns/user/social_google.rb` (lines 77-82). After exhausting retries, the exception is re-raised so the controller's generic rescue can handle it and report to Sentry.

## Test results

Added two tests:
1. Stubs `save!` to raise `ActiveRecord::Deadlocked` twice then succeed — asserts `publish!` completes without error
2. Stubs `save!` to always raise `ActiveRecord::Deadlocked` — asserts exception is re-raised after 3 attempts

---

Built with Claude Opus 4.6. Prompts: "Fix a Sentry deadlock issue — add retry logic to Link#publish! for ActiveRecord::Deadlocked, following existing pattern in user/social_google.rb"

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -419,8 +419,13 @@ class Link < ApplicationRecord
     self.deleted_at = nil
     self.draft = false
     self.publishing = true
+    deadlock_retries = 0
     begin
       save!
+    rescue ActiveRecord::Deadlocked
+      deadlock_retries += 1
+      retry if deadlock_retries <= 2
+      raise
     ensure
       self.publishing = false
     end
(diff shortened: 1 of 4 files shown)


gumclaw:
@claude review once

PR #4807, opened 2026-04-27
Migrate leftover BGN balances to EUR for Bulgarian Stripe accounts

### LINKED ISSUE #4712: BGN holding_currency mismatch prevents payouts for Bulgarian Stripe Connect accounts
## Bug

Bulgarian sellers with Stripe Connect accounts settling in EUR can have balance rows incorrectly tagged with `holding_currency: "bgn"` instead of `"eur"`. This causes `StripePayoutProcessor.is_balance_payable` to return false, silently blocking all payouts. The user sees their balance in the dashboard but payouts are skipped with a misleading "Negative balance" log message.

## Root Cause

`holding_currency` on Balance rows comes from Stripe's `destination_payment.balance_transaction[:currency]` response at charge time (`stripe_charge.rb:91-92`). If Stripe settled a charge in BGN during a transitional window (Bulgaria adopted EUR on Jan 1, 2026), the balance row is permanently tagged BGN.

Meanwhile, PR #2437 (Dec 29, 2025) switched Gumroad's side to treat Bulgaria as EUR. The merchant account's `currency` is now `"eur"`, but the `is_balance_payable` check at `stripe_payout_processor.rb:77-86` requires an exact match:

```ruby
def self.is_balance_payable(balance)
  case balance.merchant_account.holder_of_funds
  when HolderOfFunds::STRIPE
    balance.holding_currency == balance.merchant_account.currency
  end
end
```

`"bgn" != "eur"` → balance excluded → filtered sum is 0 → "Negative balance" log → payout fails.

PR #2437 had an unchecked follow-up task: "Update currency from BGN to EUR for existing BG merchant_accounts on production." This may be related.

## Concrete Example

- **User ID:** 24982534
- **Stripe `default_currency`:** `"eur"` (confirmed via `Stripe::Account.retrieve`)
- **`merchant_account.currency`:** `"eur"`
- **3 unpaid balance rows** (IDs: 15653880, 15664483, 15686305) all have `holding_currency: "bgn"`
- **Previous payout** of EUR 160.18 on April 8 worked fine (those earlier balance rows presumably had `holding_currency: "eur"`)
- **`unpaid_balance_cents`:** 7685 ($76.85) — positive, but payout service reports "Negative balance"

## How `holding_currency` Gets Set

Balance rows are created in `BalanceTransaction#find_or_create_balance` (`balance_transaction.rb:196-204`) with `holding_currency: holding_amount_currency`. For Stripe destination charges, this comes from `stripe_destination_payment_balance_transaction[:currency]` (`stripe_charge.rb:91-92`). So the value is whatever Stripe returns at charge time — if Stripe was still settling in BGN during a transition window, that's what gets recorded permanently.

## Suggested Fix

1. **Data fix:** Update existing BGN-tagged balance rows for Bulgarian accounts where Stripe `default_currency` is EUR
2. **Query to find affected users:**
```sql
SELECT b.id, b.holding_currency, ma.currency AS ma_currency, b.user_id, b.amount_cents
FROM balances b
JOIN merchant_accounts ma ON b.merchant_account_id = ma.id
WHERE b.holding_currency = 'bgn' AND ma.currency = 'eur' AND b.state = 'unpaid'
```
3. **Code fix:** Consider making `is_balance_payable` handle currency migration cases, or ensure `holding_currency` is always derived from the merchant account's current currency rather than the Stripe response at charge time

## Impact

Potentially affects any Bulgarian seller whose charges were processed during the BGN→EUR transition window. Payouts silently fail with no user-facing error. The seller sees a balance in their dashboard that never pays out. Could also affect other non-eurozone EU countries (Romania/RON, Poland/PLN, Czech Republic/CZK, Hungary/HUF) if their Stripe accounts settle in EUR.

## Helper Ticket

https://help.gumroad.com/conversations?id=0ba3c4003e8b4e6e8c75383dbde06e1c

ershad:
## What

Adds `Onetime::MigrateBulgariaBgnBalancesToEur` to bring our payout records in sync with Stripe for Bulgarian Connect accounts following the EUR adoption on 2026-01-01.

The task does two things:

1. Flips `currency` from `"bgn"` to `"eur"` on `merchant_accounts` where `country = "BG"`. This is the unchecked follow-up from #2437.
2. Converts unpaid `balances` whose `holding_currency = "bgn"` and whose merchant account is Bulgarian: divides `holding_amount_cents` by the legal fixed rate `1.95583` and sets `holding_currency = "eur"`.

`processing` and `paid` balances are left untouched — they are historical records, and processing balances are mid-payout. The task is idempotent and supports a dry-run mode (default).

## Why

Closes #4712. `StripePayoutProcessor.is_balance_payable` (`app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb:77-86`) requires `balance.holding_currency == merchant_account.currency`. Bulgarian sellers had charges that Stripe settled in BGN before the transition; #2437 then flipped our merchant account / payout currency to EUR, leaving leftover BGN-tagged balance rows that fail the equality check. Affected sellers see a balance in their dashboard but payouts skip silently with a misleading "Negative balance" log.

Stripe converts pegged BGN balances at the legal rate `1.95583 BGN = 1 EUR` (the rate fixed by the Bulgarian National Bank since 1999 and applied by the EU for euro adoption), so the same rate is used here to keep both sides consistent.

I scoped this PR to the data fix only because it directly unblocks payouts. A separate code-side change to `is_balance_payable` to handle currency-migration windows would be defensive against future cases (RON/PLN/CZK/HUF) but isn't needed to resolve this issue.

## Runbook

These are the steps for execution on production. Run from a Rails console.

```ruby
# 1. Dry run — counts and ID ranges only.
Onetime::MigrateBulgariaBgnBalancesToEur.process

# 2. Verify the printed counts look reasonable, then execute.
Onetime::MigrateBulgariaBgnBalancesToEur.process(dry_run: false)
```

To verify before running:

```sql
SELECT COUNT(*) FROM merchant_accounts WHERE country = 'BG' AND currency = 'bgn';

SELECT COUNT(*)
FROM balances b
JOIN merchant_accounts ma ON b.merchant_account_id = ma.id
WHERE b.state = 'unpaid' AND b.holding_currency = 'bgn' AND ma.country = 'BG';
```

After running, both queries should return 0. Re-running the task is safe (idempotent) — it filters by `holding_currency = 'bgn'` and `currency = 'bgn'`, so converted rows are naturally excluded.

To verify a specific affected user (e.g. user 24982534 from the issue) post-migration:

```ruby
user = User.find(24982534)
user.balances.unpaid.pluck(:id, :holding_currency, :holding_amount_cents)
# expect: holding_currency = "eur" for all unpaid rows
StripePayoutProcessor.is_balance_payable(user.balances.unpaid.first)
# expect: true
```

## Test Results

```
$ bundle exec rspec spec/services/onetime/migrate_bulgaria_bgn_balances_to_eur_spec.rb
Finished in 0.53903 seconds
11 examples, 0 failures
```

## QA / verification

- Pre-deploy: dry-run on staging with seed data.
- Post-deploy: re-run dry-run in prod, confirm counts match the queries above, then execute. Pick one affected user and verify the next scheduled payout job processes them (or trigger a manual payout via the existing flow) instead of logging "Negative balance".

---

🤖 Generated with [Claude Opus 4.7 (1M context)](https://claude.com/claude-code) via Claude Code.

Prompts given to the agent:
- "https://github.com/antiwork/gumroad/issues/4712 — how to do this?"
- "create a PR and document the steps"

### DIFF
--- a/app/services/onetime/migrate_bulgaria_bgn_balances_to_eur.rb
+++ b/app/services/onetime/migrate_bulgaria_bgn_balances_to_eur.rb
@@ -54,9 +54,14 @@ module Onetime
 
         scope.find_in_batches(batch_size: BATCH_SIZE) do |batch|
           ReplicaLagWatcher.watch
-          ids = batch.map(&:id)
-          puts "merchant_accounts batch: #{ids.first}..#{ids.last}"
-          MerchantAccount.where(id: ids).update_all(currency: Currency::EUR) unless @dry_run
+
+          batch.each do |account|
+            puts "merchant_account id=#{account.id} currency=#{account.currency} -> #{Currency::EUR}"
+          end
+
+          next if @dry_run
+
+          MerchantAccount.where(id: batch.map(&:id)).update_all(currency: Currency::EUR)
         end
       end
 
@@ -70,12 +75,15 @@ module Onetime
 
         scope.find_in_batches(batch_size: BATCH_SIZE) do |batch|
           ReplicaLagWatcher.watch
-          puts "balances batch: #{batch.first.id}..#{batch.last.id}"
-
-          next if @dry_run
 
           batch.each do |balance|
             eur_cents = (BigDecimal(balance.holding_amount_cents) / BGN_PER_EUR).round
+            puts "balance id=#{balance.id} user_id=#{balance.user_id} merchant_account_id=#{balance.merchant_account_id} " \
+                 "holding_currency=#{balance.holding_currency} -> #{Currency::EUR}, " \
+                 "holding_amount_cents=#{balance.holding_amount_cents} -> #{eur_cents}"
+
+            next if @dry_run
+
             balance.update_columns(
               holding_currency: Currency::EUR,
               holding_amount_cents: eur_cents,
--- a/app/services/onetime/migrate_bulgaria_bgn_balances_to_eur.rb
+++ b/app/services/onetime/migrate_bulgaria_bgn_balances_to_eur.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+# Migrates leftover BGN-denominated state on Bulgarian Stripe Connect accounts
+# to EUR following Bulgaria's euro adoption on 2026-01-01.
+#
+# Stripe converts pegged BGN balances to EUR at the legal fixed rate of
+# 1.95583 BGN = 1 EUR. This script brings our records in sync with what
+# Stripe holds, so `StripePayoutProcessor.is_balance_payable` (which compares
+# `Balance#holding_currency` to `MerchantAccount#currency`) stops rejecting
+# them.
+#
+# The script:
+#   1. Flips `currency` from "bgn" to "eur" on BG `merchant_accounts` (the
+#      unchecked follow-up from PR #2437).
+#   2. Converts unpaid `balances` whose `holding_currency` is "bgn" and whose
+#      merchant account is Bulgarian: divides `holding_amount_cents` by the
+#      fixed rate and sets `holding_currency` to "eur".
+#
+# Already-paid and processing balances are left untouched: they are historical
+# records of the actual settled currency at the time, and processing balances
+# are mid-payout.
+#
+# Usage:
+#   Onetime::MigrateBulgariaBgnBalancesToEur.process            # dry-run
+#   Onetime::MigrateBulgariaBgnBalancesToEur.process(dry_run: false)
+module Onetime
+  class MigrateBulgariaBgnBalancesToEur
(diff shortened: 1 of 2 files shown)


ershad:
@claude review once

slavingia:
Can just be done via console? And then a skill can live within Gumclaw if it needs to happen again

ershad:
@slavingia Sure, wanted to document the process so we know what we ran. Have taken a backup of the data too. We probably don't need it as a skill since it would be rare to do this in the future, can generate/run fresh code then.

PR #4813, opened 2026-04-27
Fix flaky Elasticsearch search spec ordering assertion

gumclaw:
## What

Changes `eq` to `match_array` in `search_spec.rb:660` for the "searching by multiple user IDs" test.

## Why

Elasticsearch does not guarantee result order when relevance scores are equal. The test expected `[450, 451]` but got `[451, 450]`, causing a flaky failure in Test Fast 15. Using `match_array` asserts the correct records are returned without requiring a specific order.

## Test results

One-line change in spec file only. The fix is a standard RSpec pattern for unordered collection assertions.

---

AI disclosure: Claude Opus 4.6 — prompted to fix CI flaky test identified in autoresearch loop.

### DIFF
--- a/spec/modules/product/searchable/search_spec.rb
+++ b/spec/modules/product/searchable/search_spec.rb
@@ -658,7 +658,7 @@ describe "Product::Searchable - Search scenarios" do
       end
 
       it "returns products from both users" do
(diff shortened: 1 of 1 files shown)


gumclaw:
Superseded by #4817 which also includes embed spec fixes

PR #4824, opened 2026-04-28
Fix ArgumentError: string contains null byte in LoginsController#create

gumclaw:
## Problem

`LoginsController#create` crashes with `ArgumentError: string contains null byte` when a login request includes null bytes (`\x00`) in the `login_identifier` or `password` parameters. PostgreSQL rejects strings containing null bytes, causing an unhandled exception.

**Sentry:** https://gumroad-to.sentry.io/issues/7445875963/

## Fix

Strip null bytes from `login_identifier` and `password` using `String#delete` before they reach any database query. This is a no-op for normal input and silently removes the problematic bytes from malformed requests.

## Tests

Added two specs confirming null bytes in both `login_identifier` and `password` are handled gracefully without raising.

### DIFF
--- a/app/controllers/logins_controller.rb
+++ b/app/controllers/logins_controller.rb
@@ -34,8 +34,8 @@ class LoginsController < Devise::SessionsController
     end
 
     if params["user"].instance_of?(ActionController::Parameters)
-      login_identifier = params["user"]["login_identifier"]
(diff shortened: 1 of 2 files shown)


ershad:
Rare issue, can skip.

PR #4825, opened 2026-04-28
Add /internal/admin precise refund endpoint

nyomanjyotisa:
Part of #4677

## What

Adds `POST /api/internal/admin/purchases/:id/refund` so `gumroad admin` (and Gumclaw, via the CLI) can refund a specific purchase end-to-end without going through the admin web UI.

Body params: `email` (required, sanity-checks the buyer email), `amount_cents` (optional partial refund in the purchase's displayed currency), `force` (bypass refund-policy timeframe and fine print), `cancel_subscription` (cancel the subscription alongside the refund).

The response includes the post-refund serialized purchase, a `subscription_cancelled` boolean, and an optional `subscription_cancel_error` field that is only present when the refund succeeded but the cancel attempt raised — so the CLI can distinguish "not attempted", "cancelled", and "tried and failed" without ambiguity.

Refunds are attributed to `GUMROAD_ADMIN_ID`, which routes through the team-member branch in `refund_and_save!` and bypasses `seller.refunds_disabled?` and the unpaid-balance check.

## Why

This is PR 4 in the admin rollout plan agreed on #4677. Buyer refund requests are one of the highest-volume admin actions still missing from the CLI today, and this endpoint pairs with the upcoming gumroad-cli Phase 1 support-write commands to unlock the first meaningful end-to-end Gumclaw refund path.

The reuse rule from the rollout plan says to build on the existing `Purchase#refund!` policy and execution logic and only add the API shape needed for precise purchase targeting and CLI use. Every code path here delegates to existing `Purchase::Refundable` behavior — combined-charge handling, partial-refund VAT proration, PayPal/Stripe dispatch, pending-cancel cleanup, and the chargeback guard are all already correct in the model layer. `force=true` only bypasses controller-level guards; the model-layer `chargedback_not_reversed?` block is preserved.

The path-based ID with email and sensitive params in the JSON body matches PR 2A's contract rule (exact ID lookup may stay in the path; sensitive identifiers must not appear in URLs).

---

This PR was implemented with AI assistance using Claude Opus 4.7


nyomanjyotisa:
@claude review

nyomanjyotisa:
@claude review

nyomanjyotisa:
@claude review once

gumclaw:
@claude review

gianfrancopiana:
@nyomanjyotisa Claude feedback aside, consider the following:

1. `cancel_subscription=true` should use admin/seller cancellation semantics. Right now it can be treated as buyer-cancelled and fail for installment plans after the refund already succeeded.

2. I think partial refunds need replay protection. If the CLI times out after a successful partial refund and retries, the same refund can be applied again while there is still refundable balance.

3. The admin purchase payload should include the purchase's displayed currency and refundable amount, especially on the lookup response. The paired CLI command should keep the same UX as `sales refund --amount` by accepting a human amount, but the shared parser defaults to x100 when no currency is known. For `gumroad admin purchases refund --amount` to be safe, the CLI needs the purchase currency before it converts the amount. Otherwise, a JPY refund like `--amount 500` would be sent as `amount_cents=50000`.

nyomanjyotisa:
Addressed: strict `amount_cents` regex, "no charge to refund" pre-check, accurate `subscription_cancelled` reporting, cancel uses `by_seller: true, by_admin: true` (installment-safe), payload now includes `currency_type` + `amount_refundable_cents` (in displayed currency). Also stripped whitespace from `email`.

On replay protection: CLI prints a "verify status with `gumroad admin purchases view <id>` before retrying" hint on errors so admin/gumclaw/agent can confirm state before re-running.

PR #4828, opened 2026-04-28
Fix LockWaitTimeout in Purchase#unsubscribe_buyer

gumclaw:
## Problem

`PurchasesController#unsubscribe` triggers `Purchase#unsubscribe_buyer`, which uses `find_each` + individual `update!` calls to set `can_contact: false` on all purchases for the buyer/seller pair. For buyers with many purchases from the same seller, this acquires row-level locks one at a time, causing `ActiveRecord::LockWaitTimeout` when concurrent transactions touch the same rows.

**Sentry:** https://gumroad-to.sentry.io/issues/7446726301/
**Occurrences (7d):** 1
**Occurrences (14d):** 1
**Estimated users affected/month:** Low (single occurrence so far)
**Estimated support tickets/month:** ~0

## Fix

Replace the `find_each` + `update!` loop with a single `update_all` call. This executes one SQL UPDATE statement instead of N, dramatically reducing the lock window.

This is safe because:
- No `after_save`/`after_update`/`after_commit` callbacks depend on `can_contact` changes
- The old code already fell back to `save(validate: false)` for invalid records, proving validations aren't critical here
- The only `can_contact`-related callback is a `before_create` hook (unrelated to updates)

## Changes

- **`app/models/purchase.rb`**: Replace `find_each` + `update!` with `update_all`
- **`spec/models/purchase_spec.rb`**: Remove the "when purchase record is invalid" test context (no longer applicable since `update_all` bypasses validations entirely)

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -2406,14 +2406,7 @@ class Purchase < ApplicationRecord
 
   # Unsubscribe the buyer of this purchase from all of the seller's emails
   def unsubscribe_buyer
-    Purchase.where(email:, seller_id:, can_contact: true).find_each do |purchase|
-      purchase.update!(can_contact: false)
-    rescue ActiveRecord::RecordInvalid
-      Rails.logger.info "Could not update purchase (#{id}) with validations turned on. Unsubscribing the buyer without running validations."
-
-      purchase.can_contact = false
-      purchase.save(validate: false)
-    end
+    Purchase.where(email:, seller_id:, can_contact: true).update_all(can_contact: false)
 
     Follower.unsubscribe(seller_id, email)
   end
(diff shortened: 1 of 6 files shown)


gumclaw:
@claude review

app/models/purchase.rb
```diff
@@ -2406,14 +2406,7 @@ def upload_invoice_pdf(pdf)
...
   def unsubscribe_buyer
-    Purchase.where(email:, seller_id:, can_contact: true).find_each do |purchase|
-      purchase.update!(can_contact: false)
-    rescue ActiveRecord::RecordInvalid
-      Rails.logger.info "Could not update purchase (#{id}) with validations turned on. Unsubscribing the buyer without running validations."
-
-      purchase.can_contact = false
-      purchase.save(validate: false)
-    end
+    Purchase.where(email:, seller_id:, can_contact: true).update_all(can_contact: false)
```
ershad: If we do this, the AudienceMember records won't get updated, fixing it.

ershad:
Just one instance of the error, not worth fixing right now, given the complexities.

PR #4829, opened 2026-04-28
Fix ActiveRecord::LockWaitTimeout in OrdersController#create

gumclaw:
## Problem

`ActiveRecord::LockWaitTimeout` occurs in `OrdersController#create` when concurrent purchases for the same seller compete for the `Balance` row lock during checkout.

**Sentry:** https://gumroad-to.sentry.io/issues/7446742663/
**Occurrences (7-day):** 1
**Occurrences (14-day):** 1
**Estimated affected users/month:** Low (single occurrence so far, but can affect any buyer purchasing from a high-traffic seller)
**Estimated support tickets/month:** ~0-1 (buyer sees a generic error and likely retries)

## Root Cause

`BalanceTransaction#update_balance!` uses `balance.with_lock` (`SELECT ... FOR UPDATE`) to update the seller's balance after a successful charge. This method only rescues `ActiveRecord::RecordInvalid`, not `ActiveRecord::LockWaitTimeout`.

When MySQL's `innodb_lock_wait_timeout` (default 50s) is exceeded due to concurrent operations on the same seller's `Balance` row (e.g. simultaneous purchases, payouts, or refunds), the exception propagates unhandled.

Additionally, `Order::ChargeService#ensure_all_purchases_processed` would incorrectly mark a purchase as **failed** even when the Stripe charge had already **succeeded**, because the balance update error prevented `MarkSuccessfulService` from completing. This is a data integrity issue: the buyer is charged but the purchase is recorded as failed.

## Fix

1. **`BalanceTransaction#update_balance!`**: Rescue `ActiveRecord::LockWaitTimeout` with retry logic (up to 3 attempts with exponential backoff), matching the existing pattern for `RecordInvalid`. Re-selects the balance on each retry in case the original was invalidated.

2. **`Order::ChargeService#ensure_all_purchases_processed`**: When `charge_intent.succeeded?` but a purchase is still `in_progress` (because post-charge processing like the balance update failed), retry `MarkSuccessfulService` instead of incorrectly calling `MarkFailedService`.

## Testing

- Added tests for `BalanceTransaction#update_balance!` lock wait timeout retry (success after retry + exhaustion)
- Added test for `ensure_all_purchases_processed` retry when charge succeeded but post-charge processing failed


gumclaw:
@claude review

PR #4836, opened 2026-04-28
Fix CI flakes: offer code error and PPP country expectations

gumclaw:
## What

Fixes two test expectation mismatches that are breaking CI on every PR branch.

## Why

Two recently merged PRs changed behavior without updating all affected test expectations:

1. **#4823** (auto-delete single-use codes) — single-use offer codes now get soft-deleted after use. When a second purchase tries to use the code, it's not found (deleted), so the error is `offer_code_invalid` instead of `offer_code_sold_out`. Updated `create_service_spec.rb`.

2. **#4810** (exclude high-income countries from PPP) — added ~35 high-income countries to `PPP_EXCLUDED_COUNTRIES` but only updated the PPP worker spec, not the API controller spec. Updated `links_controller_spec.rb` to set all excluded countries (AE, BH, BN, ES, HK, IT, KR, KW, MO, OM, PT, QA, SA, SG) to 1000.

## Impact

All 7 open PR branches are currently red because of these two flakes. This unblocks them.

---

AI disclosure: Claude Opus 4.6 via Gumclaw.


spec/services/purchase/create_service_spec.rb
slavingia: should be sold out? cc @ershad

ershad: That behavior is changed here: https://github.com/antiwork/gumroad/pull/4823:

> When a discount code with max_purchase_count=1 gets used, automatically soft-delete it. Creators who generate lots of unique single-use codes no longer need to manually clean them up.

Will update the test to use a discount with max_purchase_count > 1

gianfrancopiana:
@ershad Since #4835 landed, CI is already green. The remaining diff still makes sense to me as coverage for the sold-out path now that single-use codes are auto-deleted.

ershad:
Yeah, good to cover the sold-out path.

spec/services/purchase/create_service_spec.rb
slavingia: Yes but sold out is more correct / useful message than invalid?

PR #4843, opened 2026-04-28
Add admin purchase search list

gianfrancopiana:
Ref #4677

## What

Adds an `/internal/admin` purchase search endpoint for the admin CLI path. It returns multiple matching purchases with a hard result cap and a `has_more` signal instead of collapsing search down to the first match.

## Why

Email-based support tickets can involve more than one matching purchase, so the admin CLI needs a server contract that exposes the list safely. This reuses the existing admin search behavior while keeping sensitive lookups in a POST body.

---


gianfrancopiana:
@claude review

spec/controllers/api/internal/admin/purchases_controller_spec.rb
```diff
@@ -4,6 +4,132 @@
...
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body["success"]).to be(true)
+      expect(response.parsed_body["count"]).to eq(2)
+      expect(response.parsed_body["limit"]).to eq(described_class::MAX_SEARCH_RESULTS)
+      expect(response.parsed_body["has_more"]).to be(false)
+
+      purchases = response.parsed_body["purchases"]
+      expect(purchases.map { _1["id"] }).to eq([newer_purchase.external_id_numeric.to_s, older_purchase.external_id_numeric.to_s])
+      expect(purchases.first).to include(
```
slavingia: Worth checking all explicitly?

slavingia (approved):

PR #4845, opened 2026-04-28
Fix ActiveRecord::StatementInvalid when checkout params are nested

gumclaw:
## What

Added a `checkout_params` strong parameters method in `CheckoutController` that permits only the expected scalar parameters before passing them to `CheckoutPresenter#checkout_props`. This prevents nested `ActionController::Parameters` objects from reaching ActiveRecord finder methods.

## Why

When a malformed query string like `?product[foo]=bar` hits `CheckoutController#show`, `params[:product]` becomes an `ActionController::Parameters` object instead of a string. This value flows into `Link.find_by_unique_permalink()`, which can't quote it for SQL, raising `ActiveRecord::StatementInvalid: can't quote ActionController::Parameters`.

Sentry: https://gumroad-to.sentry.io/issues/7447859803/

Using `params.permit` is the idiomatic Rails fix — it strips any non-scalar values and returns `nil` for nested params, which the presenter already handles gracefully.

## Test results

Added a controller test that sends nested params (`product: { foo: "bar" }`) and verifies the request succeeds instead of raising.

---

AI disclosure: Built with Claude Opus 4.6. Prompted with the Sentry error details and root cause analysis, asked to implement the strong parameters fix and add a regression test.

### DIFF
--- a/app/controllers/checkout_controller.rb
+++ b/app/controllers/checkout_controller.rb
@@ -8,7 +8,7 @@ class CheckoutController < ApplicationController
   def show
     render inertia: "Checkout/Show", props: {
       cart: -> { CartPresenter.new(logged_in_user:, ip: request.remote_ip, browser_guid: cookies[:_gumroad_guid]).cart_props },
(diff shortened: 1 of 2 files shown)


slavingia:
@claude review

PR #4857, opened 2026-04-29
Fix "Show active customers only" filter not hiding cancelled subscriptions

### LINKED ISSUE #4853: "Show active customers only" filter does not hide inactive customers
## Bug

The "Show active customers only" toggle on the Customers dashboard does not filter out inactive customers. When the toggle is enabled, customers tagged as "Inactive" still appear in the results.

**Steps to reproduce:**
1. Go to the Customers dashboard for a subscription product
2. Open the filter panel
3. Enable "Show active customers only" toggle
4. Observe that customers labeled "Inactive" still appear in the list

**Expected behavior:** Only active subscribers should be shown when the toggle is on. Inactive/cancelled/gift-only customers should be hidden.

**Actual behavior:** Inactive customers remain visible despite the filter being enabled.

**Impact:** Creators running physical subscription products rely on this filter to build monthly fulfillment lists. Without a working filter, they have to manually cross-reference CSV exports to determine who should receive shipments each cycle.

gumclaw:
## What

Added the `exclude_cancelled_or_pending_cancellation_subscriptions` filter to the "Show active customers only" toggle on the Customers dashboard.

## Why

The filter only excluded **deactivated** subscriptions (where `deactivated_at` is set via `Subscription#deactivate!`), but not **cancelled** or **pending-cancellation** subscriptions (where only `cancelled_at` is set via `Subscription#cancel!`). The UI labels these customers as "Inactive" based on `Subscription#status` returning non-"alive" values, but the Elasticsearch query wasn't filtering them out.

The `exclude_cancelled_or_pending_cancellation_subscriptions` search option already existed in `PurchaseSearchService` (filtering on `subscription_cancelled_at`) — it just wasn't being used when `active_customers_only` was enabled.

## Test results

All 29 specs pass in `customers_controller_spec.rb`, including 2 new tests covering the active-only filter for deactivated, cancelled, and pending-cancellation subscriptions.

Fixes https://github.com/antiwork/gumroad/issues/4853

---

AI disclosure: Built with Claude Opus 4.6. Prompted to fix GitHub issue #4853 — the "Show active customers only" filter not hiding inactive customers.

### DIFF
--- a/app/controllers/customers_controller.rb
+++ b/app/controllers/customers_controller.rb
@@ -129,6 +129,7 @@ class CustomersController < Sellers::BaseController
 
       if active_customers_only
         search_options[:exclude_deactivated_subscriptions] = true
+        search_options[:exclude_cancelled_or_pending_cancellation_subscriptions] = true
         search_options[:exclude_refunded_except_subscriptions] = true
         search_options[:exclude_unreversed_chargedback] = true
       end
--- a/app/services/content_moderation/strategies/prompt_strategy.rb
+++ b/app/services/content_moderation/strategies/prompt_strategy.rb
@@ -94,7 +94,7 @@ class ContentModeration::Strategies::PromptStrategy
     else
       Result.new(status: "compliant", reasoning: [])
     end
-  rescue Faraday::TimeoutError, Faraday::ConnectionFailed, Net::ReadTimeout => e
+  rescue Faraday::TimeoutError, Faraday::ConnectionFailed, Faraday::ServerError, Net::ReadTimeout => e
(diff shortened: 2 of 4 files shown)


gumclaw:
Duplicate of #4858, which has the more complete fix. Closing this one.

nyomanjyotisa (approved):
## Before

https://github.com/user-attachments/assets/412b3cce-9345-4a65-a2d3-8987eb5bb8e4

## After

https://github.com/user-attachments/assets/9fa70d64-1fa9-474a-9714-93f826cee725

PR #4859, opened 2026-04-29
Fix ActiveModel::RangeError in Bundles::ProductController#update

gumclaw:
## What

Rescue `ActiveModel::RangeError` in `Bundles::ProductController#update` so that submitting a value exceeding the 4-byte integer limit (e.g. `price_cents` of 3,000,000,000) returns a user-friendly error instead of a 500.

## Why

When a user submits a bundle update with a numeric value that exceeds `2,147,483,647`, Rails raises `ActiveModel::RangeError` before any model validation runs. The existing rescue clause only catches `ActiveRecord::RecordNotSaved`, `ActiveRecord::RecordInvalid`, and `Link::LinkInvalid`, so this error was unhandled and reported to Sentry.

**Sentry:** https://gumroad-to.sentry.io/issues/7448623466/

This follows the same pattern already used in `CheckoutDiscountsController` and `Api::V2::LinksController`.

## Test results

Added a spec that submits `price_cents: 3_000_000_000` and asserts the redirect with the expected flash message.

---

AI Disclosure: Written with Claude Opus 4.6. Prompt: "Fix Sentry bug: ActiveModel::RangeError in Bundles::ProductController#update"

### DIFF
--- a/app/controllers/bundles/product_controller.rb
+++ b/app/controllers/bundles/product_controller.rb
@@ -53,6 +53,8 @@ class Bundles::ProductController < Bundles::BaseController
   rescue ActiveRecord::RecordNotSaved, ActiveRecord::RecordInvalid, Link::LinkInvalid => e
     error_message = @bundle.errors.full_messages.first || e.message
     redirect_to edit_bundle_product_path(@bundle.external_id), alert: error_message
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4861, opened 2026-04-29
Add /internal/admin Phase 1 write endpoints

nyomanjyotisa:
Part of #4677

## What

Adds the eight `/internal/admin` write endpoints the upcoming `gumroad-cli` Phase 1 support-write commands need:

- `POST /api/internal/admin/purchases/:id/resend_receipt`
- `POST /api/internal/admin/purchases/resend_all_receipts`
- `POST /api/internal/admin/purchases/:id/refund_taxes`
- `POST /api/internal/admin/purchases/reassign`
- `POST /api/internal/admin/users/reset_password`
- `POST /api/internal/admin/users/update_email`
- `POST /api/internal/admin/users/two_factor_authentication`
- `POST /api/internal/admin/users/create_comment`

Two new shared services so helper and admin don't duplicate the chunky bits:

- `Purchase::ReassignByEmailService` (reassign-by-email + grouped receipt enqueue)
- `User::CreateAdminCommentService` (idempotent comment create with `RecordNotUnique` rescue)

Helper's external contract is byte-for-byte unchanged. Every legacy helper spec still passes without edits — the refactor only swapped out the inlined logic for service calls.

Together with #4825 (precise refund) and the earlier search-returns-list change, this completes the server-side surface for the Phase 1 CLI commands.

## Why

This is the server-side prerequisite for PR 5 in the rollout plan agreed on #4677. The plan rule is explicit: "commands should call `/internal/admin`; if a write endpoint only exists under `/internal/helper`, add or wait for the matching Gumroad server endpoint first." All eight endpoints I added existed only under `/internal/helper` before this PR.

Two design choices worth calling out:

1. **Services over controller concerns for shared logic.** A concern mixed into both controllers would force one HTTP shape on both (helper uses `error_message`, admin uses `message`; helper returns 422 for missing user, admin returns 404), and would couple the domain logic to the controller layer. Service classes keep the domain operations HTTP-clean and reusable from a Sidekiq job or a console script later.

2. **Two helper endpoints intentionally not ported.** `refund_last_purchase` and `resend_last_receipt` operate on `Purchase.where(email:).order(created_at: :desc).first`. The "last purchase" semantic is unreliable when a buyer has multiple orders, and the meeting notes on #4677 call this out as a known gap. The new admin contract is "search by email → operate by id," consistent with PR 4's "precise refund" framing. Helper keeps these for backward compat; admin doesn't repeat the mistake.

---

This PR was implemented with AI assistance using Claude Opus 4.7

### DIFF
--- a/app/controllers/api/internal/admin/purchases_controller.rb
+++ b/app/controllers/api/internal/admin/purchases_controller.rb
@@ -41,6 +41,71 @@ class Api::Internal::Admin::PurchasesController < Api::Internal::Admin::BaseCont
     render json: { success: false, message: "purchase_date must use YYYY-MM-DD format." }, status: :bad_request
   end
 
+  def resend_receipt
+    purchase = fetch_purchase
+    return render json: { success: false, message: "Purchase not found" }, status: :not_found if purchase.blank?
+
+    purchase.resend_receipt
+    render json: {
+      success: true,
+      message: "Successfully resent receipt for purchase number #{purchase.external_id_numeric} to #{purchase.email}"
+    }
+  end
+
+  def resend_all_receipts
+    email = params[:email].to_s.strip
+    return render json: { success: false, message: "email is required" }, status: :bad_request if email.blank?
+
+    purchases = Purchase.where(email: email).successful
+    return render json: { success: false, message: "No purchases found for email: #{email}" }, status: :not_found if purchases.empty?
+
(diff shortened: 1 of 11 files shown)


nyomanjyotisa:
@claude review

nyomanjyotisa:
@claude review

gianfrancopiana (approved):

PR #4862, opened 2026-04-29
Fix Rack::Timeout in LinksController#publish by moving affiliate association to background

gumclaw:
## Problem

`LinksController#publish` was hitting `Rack::Timeout::RequestTimeoutException` (120s) in production.

**Sentry:** https://gumroad-to.sentry.io/issues/7448861663/
**Occurrences (7d):** 1 (first seen)
**Estimated users affected:** Low (single occurrence so far)
**Estimated support tickets/month:** 0 (publish failure is retryable client-side)

## Root Cause

The `Link#publish!` method runs affiliate association synchronously in the web request. After saving the product, it iterates through all of the seller's `apply_to_all_products` direct affiliates and for each one:

1. Loads `affiliate.products` to check if the product is already associated (N+1 query)
2. Creates the `ProductAffiliate` record
3. Enqueues an `AffiliateMailer` notification

For sellers with many universal affiliates, this loop can exceed the 120s Rack::Timeout.

## Fix

Move the post-publish affiliate association logic to a new `AfterProductPublishWorker` Sidekiq job. The publish response now returns immediately after saving the product. Affiliate associations and notification emails are handled asynchronously.

Also uses `find_each` for batched iteration in the worker to handle large affiliate counts efficiently.

## Changes

- **`app/models/link.rb`**: Replace inline affiliate loop with `AfterProductPublishWorker.perform_async(id)`
- **`app/sidekiq/after_product_publish_worker.rb`**: New worker that handles affiliate association
- **`spec/models/link_spec.rb`**: Updated to verify worker is enqueued
- **`spec/sidekiq/after_product_publish_worker_spec.rb`**: Full test coverage for the worker


gumclaw:
@claude review

PR #4880, opened 2026-04-29
Fix NoMethodError: undefined method 'to_i' for Array in DiscoverController#index

gumclaw:
## Problem

`DiscoverController#index` raises `NoMethodError: undefined method 'to_i' for an instance of Array` when query string parameters like `from` or `size` appear multiple times (e.g. `?from=9&from=9`). Rails parses these as arrays, and `search_options` in `Product::Searchable` calls `.to_i` directly on `params[:from]`, which doesn't work on arrays.

**Sentry:** https://gumroad-to.sentry.io/issues/7450001994/ (4 occurrences)

## Fix

Sanitize `params[:from]` and `params[:size]` in `format_search_params!` to handle both String and Array inputs, consistent with how `:tags`, `:filetypes`, and `:ids` are already normalized. Arrays are unwrapped by taking the first element.

## Changes

- `app/controllers/concerns/search_products.rb`: Added `:from` normalization and Array handling for `:size`
- `spec/controllers/concerns/search_products_spec.rb`: Added 4 test cases for Array and String coercion of `:from` and `:size`

## Tests

All 15 specs in `search_products_spec.rb` pass.

### DIFF
--- a/app/controllers/concerns/search_products.rb
+++ b/app/controllers/concerns/search_products.rb
@@ -34,8 +34,12 @@ module SearchProducts
 
       params[:offer_code] = "__no_match__" if params[:offer_code].present? && !offer_codes_search_feature_active?(params)
 
+      params[:from] = Array.wrap(params[:from]).first.to_i if params[:from].present?
+
       if params[:size].is_a?(String)
         params[:size] = params[:size].to_i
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4882, opened 2026-04-29
Add admin user risk actions

gianfrancopiana:
Ref #4677

## What

Adds internal admin endpoints to mark a user as compliant or suspend them for fraud by email.

Both actions use the same behavior as the existing admin web buttons. Repeat calls return success without adding duplicate comments, and fraud suspension can also save an optional suspension note.

## Why

We want to prioritize these two admin actions for the CLI. This keeps the server change narrow: it exposes only the requested actions and does not add payout scheduling or other risk actions.

---


gianfrancopiana:
@claude review once

PR #4885, opened 2026-04-29
Fix NoMethodError: make charge_discover_fee? public on Purchase

gumclaw:
## What

`Purchase#charge_discover_fee?` is defined inside the `private` block (below line 2865) but is called by external code (rails runner scripts that recalculate fees). Ruby raises `NoMethodError: private method 'charge_discover_fee?' called for an instance of Purchase` when a private method is called with an explicit receiver from outside the class.

## Fix

Add `public :charge_discover_fee?` immediately after the method definition. This re-exports the single method while keeping surrounding calculation methods private. Updated specs to call the method directly instead of via `send`.

## Sentry

- **Issue:** https://gumroad-to.sentry.io/issues/7450334564/
- **Occurrences (7-day):** 1
- **Occurrences (14-day):** 1
- **Estimated affected users/month:** ~0 (single occurrence from a one-off rails runner script)
- **Estimated support tickets/month:** 0

## Test results

```
bundle exec rspec spec/models/purchase_spec.rb -e "charge_discover_fee"
3 examples, 0 failures
```

(Local environment has a pre-existing Stripe test-mode issue unrelated to this change; CI should pass clean.)

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -3333,6 +3333,7 @@ class Purchase < ApplicationRecord
       return false unless link.recommendable? || (not_is_original_subscription_purchase? && original_purchase&.was_discover_fee_charged?)
       was_product_recommended? && !RecommendationType.is_free_recommendation_type?(recommended_by)
     end
+    public :charge_discover_fee?
 
     # Calculates the fees we charge based on price_cents
     #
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4889, opened 2026-04-30
Fix: resize Suspend for fraud button to match other action buttons

gumclaw:
Moves the "Suspend for fraud" button into the existing actions row alongside Mark Compliant, Refund Balance, and Disable PayPal Sales. It now renders as an inline button matching the size of the others instead of spanning full-width with its own `<hr>`.

Closes antiwork/gumroad-private#325


gumclaw:
@claude review

PR #4893, opened 2026-04-30
Fix RequestTimeoutException in Admin::Search::PurchasesController#index

gumclaw:
## What

Fixes `Rack::Timeout::RequestTimeoutException` in `Admin::Search::PurchasesController#index` by:

- Returning `Purchase.none` immediately when no search criteria are provided (query, email, creator_email, license_key, transaction_date, last_4, card_type, price, expiry_date)
- Adding a default result limit of 500 (`MAX_RESULTS`) when no explicit limit is passed

## Why

The admin purchase search endpoint could scan the entire `purchases` table when no meaningful search criteria were provided, causing queries that exceeded the 120s Rack::Timeout threshold. This was triggered when parameters like `product_title_query` or `purchase_status` were passed without a primary search criterion — these filters only narrow results but don't prevent a full table scan.

Sentry issue: https://gumroad-to.sentry.io/issues/7451021240/

## Test results

All test failures in the affected spec files are pre-existing environment issues (purchase factory validation failures unrelated to this change), confirmed by verifying the same failures occur on `main`.

---

AI disclosure: Built with Claude Opus 4.6. Prompted to fix the Rack::Timeout in admin purchase search by requiring search criteria and adding a default limit.

### DIFF
--- a/app/services/admin_search_service.rb
+++ b/app/services/admin_search_service.rb
@@ -3,7 +3,12 @@
 class AdminSearchService
   class InvalidDateError < StandardError; end
 
+  MAX_RESULTS = 500
+
   def search_purchases(query: nil, email: nil, product_title_query: nil, purchase_status: nil, creator_email: nil, license_key: nil, transaction_date: nil, last_4: nil, card_type: nil, price: nil, expiry_date: nil, limit: nil)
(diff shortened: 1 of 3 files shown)


gumclaw:
@claude review

PR #4887, opened 2026-04-29
feat(dev): add bin/doctor to diagnose local setup issues

daniel-aranda:
## What

Adds `bin/doctor`, a read-only local setup diagnostic script for Gumroad development.

It checks:

- Ruby and rbenv setup
- Bundler version
- Redis, MongoDB, MySQL, and Elasticsearch
- Elasticsearch indices
- Nginx
- SSL certificates
- gumroad.dev availability
- Rails/AnyCable ports

## Why

Local setup can fail in several ways that are hard to diagnose from `bin/dev` output alone. This script provides a single command to identify missing services and misconfigurations before starting the app.

This reduces onboarding friction and avoids time spent debugging environment issues.

The script is intentionally read-only to avoid modifying a contributor’s environment unexpectedly.

## Before/After

**Before**

Contributors had to manually debug missing services, ports, certificates, and local routing issues.

**After**

Contributors can run `bin/doctor` to get a quick health check and suggested next steps.

Video:
https://www.loom.com/share/9c667df113444cd2be8203a0d3933ec1

Shows a failing setup (missing services) and how `bin/doctor` identifies issues, followed by a healthy run.

## Test Results

Ran `bin/doctor` locally in a healthy environment.

```bash
bin/doctor
```

Result: 
image

---

AI disclosure: I used ChatGPT GPT-5.5 Thinking as a review and refinement tool.

Prompts included (representative):
- "Review this bash script for a local dev environment doctor and suggest improvements"
- "Help identify missing checks for services like Redis, MongoDB, MySQL, Elasticsearch, and Nginx"
- "Improve CLI output clarity and developer experience for a diagnostic script"
- "Refine a pull request description following Gumroad contributing guidelines"

PR #4896, opened 2026-04-30
feat(dev): add bin/doctor to diagnose local setup issues. Updated.

daniel-aranda:
## What

Adds `bin/doctor`, a read-only local setup diagnostic script for Gumroad development.

It checks:

- Ruby and rbenv setup
- Bundler version
- Redis, MongoDB, MySQL, and Elasticsearch
- Elasticsearch indices
- Nginx
- SSL certificates
- gumroad.dev availability
- Rails/AnyCable ports

## Why

Local setup can fail in several ways that are hard to diagnose from `bin/dev` output alone. This script provides a single command to identify missing services and misconfigurations before starting the app.

This reduces onboarding friction and avoids time spent debugging environment issues.

The script is intentionally read-only to avoid modifying a contributor’s environment unexpectedly.

## Before/After

**Before**

Contributors had to manually debug missing services, ports, certificates, and local routing issues.

**After**

Contributors can run `bin/doctor` to get a quick health check and suggested next steps.

Video:
https://www.loom.com/share/9c667df113444cd2be8203a0d3933ec1

Shows a failing setup (missing services) and how `bin/doctor` identifies issues, followed by a healthy run.

## Test Results

Ran `bin/doctor` locally in a healthy environment.

```bash
bin/doctor
```

Result: 
image

---

AI disclosure: I used ChatGPT GPT-5.5 Thinking as a review and refinement tool.

Prompts included (representative):
- "Review this bash script for a local dev environment doctor and suggest improvements"
- "Help identify missing checks for services like Redis, MongoDB, MySQL, Elasticsearch, and Nginx"
- "Improve CLI output clarity and developer experience for a diagnostic script"
- "Refine a pull request description following Gumroad contributing guidelines"

---

This is a fresh PR addressing feedback from @claude and @gumclaw from the former PR:
https://github.com/antiwork/gumroad/pull/4887


daniel-aranda:
> ## Claude Code Review
> This pull request is from a fork — automated review is disabled. A repository maintainer can comment `@claude review` to run a one-time review.

Please help me @gumclaw

gumclaw:
@claude review

gumclaw:
Thanks for the fresh PR, Daniel! Reviewed the diff — all three items from the previous round are addressed:

1. **Nginx detection** — now checks running process and brew service status instead of relying on `has_command`. ✅
2. **Shell flexibility** — detects current shell, checks the right rc files, and the regex handles `rbenv init -` with or without an explicit shell arg. ✅
3. **lsof fallback** — `port_listening()` now tries lsof → ss → netstat. ✅

Script is clean and read-only as described. Triggering CI and @claude review above.

@ershadk for merge when CI passes.

daniel-aranda:
@gumclaw I’ve pushed the fixes based on the review comments. Should be good now, but happy to tweak anything else if needed.

gumclaw (changes requested):
Thanks for the quick fixes, Daniel! The MySQL/ES cascading checks are properly nested now — nice work on that.

One issue with the Bundler version check though: line 96 has `"*2.5.10"` (the `*` is inside double quotes). In bash `[[ ]]`, a quoted RHS on `==` is a literal string match, not a glob. So this compares against the literal string `*2.5.10`, which `bundle -v` will never output — meaning the check will always fall through to the warning, even when Bundler 2.5.10 is correctly installed.

Fix — either exact match:
```bash
if [[ "$bundle_version" == "Bundler version 2.5.10" ]]; then
```

Or keep the glob but leave the wildcards unquoted:
```bash
if [[ "$bundle_version" == *"2.5.10"* ]]; then
```

The first option is better (per the original review feedback about substring false positives). Quick one-liner fix and this is good to go.

daniel-aranda:
> Thanks for the quick fixes, Daniel! The MySQL/ES cascading checks are properly nested now — nice work on that.
> 
> One issue with the Bundler version check though: line 96 has `"*2.5.10"` (the `*` is inside double quotes). In bash `[[ ]]`, a quoted RHS on `==` is a literal string match, not a glob. So this compares against the literal string `*2.5.10`, which `bundle -v` will never output — meaning the check will always fall through to the warning, even when Bundler 2.5.10 is correctly installed.
> 
> Fix — either exact match:
> 
> ```shell
> if [[ "$bundle_version" == "Bundler version 2.5.10" ]]; then
> ```
> 
> Or keep the glob but leave the wildcards unquoted:
> 
> ```shell
> if [[ "$bundle_version" == *"2.5.10"* ]]; then
> ```
> 
> The first option is better (per the original review feedback about substring false positives). Quick one-liner fix and this is good to go.

@gumclaw good caught, thank you. I already fixed it.

gianfrancopiana:
Thanks for taking the time to work on this. I agree local setup failures are hard to debug. But I'm not convinced this should be merged as a repo-level Bash script in its current form. 

Our expected workflow for setup debugging is increasingly agent-assisted, and an agent can inspect the current README, env files, Docker config, and Procfile before deciding what to check. A static `bin/doctor` is another source of truth to maintain.

The hardcoded checks make that risk visible, especially `Bundler version 2.5.10`. If this lives in the repo, it should derive values from existing sources of truth like `Gemfile.lock`, `.ruby-version`, `.env.development`, `docker/docker-compose-local.yml`, and `Procfile.dev`. The service hints should also line up with the Docker-first setup path in the README.

I’d lean toward turning the idea into a project agent skill, with a small helper script inside the skill if we want deterministic probes.

daniel-aranda:
> Thanks for taking the time to work on this. I agree local setup failures are hard to debug. But I'm not convinced this should be merged as a repo-level Bash script in its current form.
> 
> Our expected workflow for setup debugging is increasingly agent-assisted, and an agent can inspect the current README, env files, Docker config, and Procfile before deciding what to check. A static `bin/doctor` is another source of truth to maintain.
> 
> The hardcoded checks make that risk visible, especially `Bundler version 2.5.10`. If this lives in the repo, it should derive values from existing sources of truth like `Gemfile.lock`, `.ruby-version`, `.env.development`, `docker/docker-compose-local.yml`, and `Procfile.dev`. The service hints should also line up with the Docker-first setup path in the README.
> 
> I’d lean toward turning the idea into a project agent skill, with a small helper script inside the skill if we want deterministic probes.

Thanks for the thoughtful feedback @gianfrancopiana this makes sense, especially around avoiding multiple sources of truth and leaning into agent-assisted workflows.

Just to share a bit more context on why I built this:

I initially used Codex/AI to get the local setup running, and it did work well. But I noticed that each run wasn’t free, I ended up spending a few dollars just iterating on setup/debugging. At some point, I wanted to rerun things manually without relying on AI every time, and that’s where things became much harder.

Without a tool like `bin/doctor`, reproducing the setup issues manually required a lot of implicit knowledge (what to check, in what order, what “healthy” looks like). That friction is what motivated this script, not to replace AI, but to complement it with a deterministic, zero-cost fallback.

I completely agree that if this lives in the repo, it should:
- derive values from sources of truth (Gemfile.lock, .ruby-version, etc.)
- align with the Docker-first workflow
- avoid hardcoded assumptions

And I also like the idea of evolving this into an agent skill, that feels like the right long-term direction.

That said, I still think there’s value in having a lightweight, local, non-AI path for debugging, especially when:
- you want fast iteration without incurring cost
- or AI tooling is unavailable / inconsistent
- or you just want something deterministic to sanity-check the environment

Happy to adjust direction here, whether that means slimming this down, aligning it more closely with existing configs.

PR #4899, opened 2026-04-30
Fix empty placeholder flashing when switching content pages

gumclaw:
## What

When switching between pages on the content editor, the empty state placeholder ("Enter the content you want to sell.") briefly flashes even when navigating to a page that already has content.

## Why

The editor content was updated via `useEffect` + `queueMicrotask`, which defers the update to after the browser paint. During that gap, the editor still holds the previous page's state, and if that page was empty, the `isEmpty` check shows the placeholder before the new content loads.

## Fix

Two changes, belt-and-suspenders:

1. **`RichTextEditor.tsx`**: Replace `useEffect` + `queueMicrotask` with `useLayoutEffect` so the editor state updates synchronously before the browser paints. This eliminates the frame where stale content is visible.

2. **`ContentTab/index.tsx`**: Add an `isSelectedPageEmpty` guard alongside `editor?.isEmpty` so the placeholder also checks the selected page's actual description data, not just the editor's potentially-stale state.

### DIFF
--- a/app/javascript/components/ProductEdit/ContentTab/index.tsx
+++ b/app/javascript/components/ProductEdit/ContentTab/index.tsx
@@ -973,7 +973,7 @@ const ContentTabContent = ({ selectedVariantId }: { selectedVariantId: string | 
         >
           <NodeVisibilityProvider scrollRef={scrollContainerRef}>
             <div className="relative h-full flex-1">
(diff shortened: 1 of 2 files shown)


nyomanjyotisa (approved):
## After

https://github.com/user-attachments/assets/8ac07b07-1f6e-4019-ba25-37ae3742c2b4

PR #4901, opened 2026-04-30
Handle LockWaitTimeout in Google OAuth2 sign-in flow

gumclaw:
## Problem

`User::OmniauthCallbacksController#google_oauth2` calls `User.find_or_create_for_google_oauth2`, which performs a `save!` inside `query_google`. When another transaction holds a row lock on the same user record for too long, MySQL raises `ActiveRecord::LockWaitTimeout` instead of a deadlock.

The existing retry logic only rescues `ActiveRecord::Deadlocked`, so `LockWaitTimeout` errors bubble up unhandled and crash the OAuth flow.

**Sentry:** https://gumroad-to.sentry.io/issues/7451458485/

## Fix

Add `ActiveRecord::LockWaitTimeout` to the existing `rescue` clause alongside `ActiveRecord::Deadlocked`. The same retry-up-to-2-times logic now applies to both lock-related errors.

## Changes

- `app/models/concerns/user/social_google.rb`: Rescue `LockWaitTimeout` alongside `Deadlocked`
- `spec/modules/user/social_google_spec.rb`: Two new tests mirroring the existing deadlock specs

### DIFF
--- a/app/models/concerns/user/social_google.rb
+++ b/app/models/concerns/user/social_google.rb
@@ -74,11 +74,11 @@ module User::SocialGoogle
         end
 
         user
-      rescue ActiveRecord::Deadlocked => e
+      rescue ActiveRecord::Deadlocked, ActiveRecord::LockWaitTimeout => e
         retries += 1
         retry if retries <= 2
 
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4903, opened 2026-04-30
Log admin comment when refund balance is triggered

### LINKED ISSUE #4876: Log admin comment when 'Refund balance' is clicked on suspended accounts
## Problem

When an admin clicks **Refund balance** on a suspended user's account, the `RefundUnpaidPurchasesWorker` processes refunds for all unpaid purchases back to the original buyers. However, no admin comment is logged on the user account recording this action.

This creates a gap in the audit trail. The suspension, payout holds, and risk review notes are all logged as comments, but the actual refund of the balance (often the most significant financial action taken on the account) has no record.

## Proposed solution

When `RefundUnpaidPurchasesWorker` completes, log an admin comment on the user account, something like:

> Refund balance initiated. $X.XX refunded across N purchases to original buyers. Triggered by [admin name].

This ensures anyone reviewing the account later sees the full timeline: flagged > suspended > balance refunded.

## Context

- The Refund balance button lives in `app/javascript/components/Admin/Users/PermissionRisk/Actions/RefundBalanceAction.tsx`
- The worker is `app/sidekiq/refund_unpaid_purchases_worker.rb`
- Discovered during support ticket triage (ticket 27320d8b) where we had to manually log the refund action

mamboitaliano21:
Closes #4876 and #4894

Hi Team this is my first PR on this repo, I've read CONTRIBUTING.md, if I make any mistake please let me know. I'm trying to get a feel on the repo's convention before tackling bigger feature work 🙏 

My approach is different from gumclaw's #4894 (explained in the "why" section), and I have reviewed it manually (video attached).

PR description is written using /pr-description.

## What

`RefundUnpaidPurchasesWorker` now writes a `refund_balance` admin comment on the suspended user's timeline when an admin triggers "Refund balance". The comment records who initiated the action (`"Refund balance initiated by <admin>."`). A new `Comment::COMMENT_TYPE_REFUND_BALANCE` constant is added so the comment can be queried/scoped like other admin-action comments.

## Why

Suspension, payout holds, and risk review notes already log to the user timeline as comments. The actual refund of the seller's unpaid balance — often the most significant financial action taken on a fraud account — had no audit entry. Ops had to manually log the action during support escalations (per the issue, ticket `27320d8b`).

The comment is written inside the worker rather than the controller for two reasons:

1. The worker has a `return unless user.suspended?` guard. Writing the comment after that guard ensures the audit entry only appears when the worker is actually going to do something. Writing it in the controller would create misleading entries on direct/non-suspended `POST` requests.
2. Matches the precedent in `SuspendUsersWorker`, which writes its own audit comments at the layer that performs the work.
3. Prevent future logic drift, all refund work lives in the job

The comment intentionally does not include a dollar amount. The "Refund balance" action produces three legitimately different numbers — gross buyer-side refund (`total_transaction_cents`), seller's gross collected (`price_cents`), and seller's net unpaid balance (`Balance#amount_cents`, what the admin UI displays). Picking any one risks misrepresentation; the structured `Refund` records that follow are the source of truth for amounts.

## Test Results

`bin/test-confidence` reaches 99% safe-to-commit on this diff (4 tests, 34s). New worker spec re-runs green under both `record: :once` (local) and `record: :none` (CI mode), confirming the VCR cassette is complete.

Screenshot 2026-04-30 at 5 40 59 PM

---

This PR was implemented with AI assistance using Claude Opus 4.7 (1M context).

## Before

https://github.com/user-attachments/assets/be40abc6-2fb9-420f-837d-005fa61d6aeb

## After

https://github.com/user-attachments/assets/e54a8aa0-c80d-415c-a18b-8cbfdad777a2

Prompts used:

- "ok let's create a branch and fix this"
- Compared the in-flight PR #4894 (alternate implementation) against this approach to choose worker-vs-controller placement
- Decided to drop the dollar amount from the comment after tracing `RefundPurchaseWorker` end-to-end and finding gross/net/seller-net diverge
- Reverted to keep the original purchase-iteration chain intact and minimize the worker diff
- Use the factory-generated admin name in the spec instead of a hardcoded value


mamboitaliano21:
Please help me @gumclaw, I think this is a polished version of your PR #4894 🙏

spec/support/fixtures/vcr_cassettes/RefundUnpaidPurchasesWorker/_perform/logs_an_admin_comment_recording_who_initiated_the_refund_balance_action.yml
```diff
@@ -0,0 +1,1923 @@
+---
```
gianfrancopiana: Can we avoid adding this new VCR cassette? The new behavior being tested is comment creation. It does not add a Stripe code path. The cassette is only needed because the new example repeats the shared purchase setup, which calls Stripe while creating/processing purchases. Can we fold the comment assertion into the existing `"queues the refund of unpaid purchases"` worker spec? So it reuses that existing cassette instead of adding another 1,923-line fixture.

app/sidekiq/refund_unpaid_purchases_worker.rb
```diff
@@ -8,6 +8,14 @@ def perform(user_id, admin_user_id)
     user = User.find(user_id)
     return unless user.suspended?
 
+    admin = User.find(admin_user_id)
+    user.comments.create!(
+      author_id: admin.id,
+      author_name: admin.name,
```
gianfrancopiana: This should use a nonblank admin display name. `admin.name` is optional: `users.name` is nullable, `User` does not validate name presence, and the `:admin_user` factory creates admins without a name. With the current code we can record `Refund balance initiated by .`, which weakens the audit trail. Please use `admin.name_or_username` for both `author_name` and the comment content, and update the spec to cover a nameless admin.

gianfrancopiana (dismissed):
I agree this is the better direction than #4894: putting the audit comment inside `RefundUnpaidPurchasesWorker` keeps it behind the suspended-user guard and covers scheduled refund execution too.

Once the admin display fallback is fixed and the duplicate cassette is removed, this should be in good shape.

spec/support/fixtures/vcr_cassettes/RefundUnpaidPurchasesWorker/_perform/logs_an_admin_comment_recording_who_initiated_the_refund_balance_action.yml
mamboitaliano21: Hi Gianfranco, good one! I agree with you, less VCR is always better.

app/sidekiq/refund_unpaid_purchases_worker.rb
mamboitaliano21: This is a very good catch, thanks Gianfranco. It's a good lesson for me 🙇 .

edit: PR Updated

gianfrancopiana:
@claude review once

app/sidekiq/refund_unpaid_purchases_worker.rb
```diff
@@ -8,6 +8,14 @@ def perform(user_id, admin_user_id)
     user = User.find(user_id)
     return unless user.suspended?
 
+    admin = User.find(admin_user_id)
+    user.comments.create!(
+      author_id: admin.id,
+      author_name: admin.name_or_username,
+      comment_type: Comment::COMMENT_TYPE_REFUND_BALANCE,
+      content: "Refund balance initiated by #{admin.name_or_username}."
+    )
```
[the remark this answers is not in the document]
gianfrancopiana: I agree with this, but I think the simple fix is enough here: move the `user.comments.create!` block after the refund enqueue loop. That way a retry caused by one of the DB queries or `RefundPurchaseWorker.perform_async` will not create a duplicate audit row before the work is queued.

I don't think we need the heavier `idempotency_key` path for this PR unless we want strict once-only audit semantics. Moving the comment to the end makes it mean "refund jobs were successfully queued," which matches the current worker behavior.

cc @mamboitaliano21

gianfrancopiana (dismissed):

app/sidekiq/refund_unpaid_purchases_worker.rb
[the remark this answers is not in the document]
mamboitaliano21: That's a good explanation. PR Updated.

Thanks for all the great explanations, appreciate it.

PR #4904, opened 2026-04-30
Enable Stripe Link on checkout

gumclaw:
Two-line change to enable Stripe Link as a payment method during checkout.

**Backend** (`stripe_charge_processor.rb`):
- Accept `link` in `payment_method_types` for on-session charges
- Off-session/recurring charges remain card-only

**Frontend** (`CreditCardInput.tsx`):
- Remove `disableLink: true` from CardElement options
- Stripe's CardElement now shows the Link email authentication flow inline
- Users with a Link account get autofilled card details for faster checkout

No changes to subscription flows, saved cards, Apple Pay, or Google Pay.


gianfrancopiana:
Closing this in favor of the split follow-up PRs.

#4908 is the easiest merge candidate after one fix: keep the backend PaymentIntent card-only and only enable Link inside the existing CardElement flow. That gives us the buyer-facing Link benefit while keeping our server-side payment data card-shaped.

#4909 is the better long-term Stripe direction, but it is a larger change. As written, it enables `link` as a real payment method, which means we would also need to update the server-side paths that currently assume card data: charge parsing, saved payment method data, receipts/admin display, and risk/blocking paths that rely on `stripe_fingerprint`.

Evidence:
- #4908 uses the existing CardElement flow: https://github.com/antiwork/gumroad/blob/2f0b6ff4eaa23df15e88c1068d7e6f827259e2bc/app/javascript/components/Checkout/CreditCardInput.tsx#L76-L83
- #4908 still needs the backend `link` change removed: https://github.com/antiwork/gumroad/blob/2f0b6ff4eaa23df15e88c1068d7e6f827259e2bc/app/business/payments/charging/implementations/stripe/stripe_charge_processor.rb#L225
- #4909 uses PaymentElement with `card` and `link`: https://github.com/antiwork/gumroad/blob/2df62bf90945786b02e82c74b435d1bd4e03f85b/app/javascript/components/Checkout/PaymentForm.tsx#L666-L667
- Server assumes card details: https://github.com/antiwork/gumroad/blob/029aa1faff1e6170860e6047b8ecab2b5ef5c3c8/app/business/payments/charging/implementations/stripe/stripe_charge.rb#L38-L42
- Stripe documents Link as a payment method returning `PaymentMethod.type = "link"`: https://docs.stripe.com/payments/link/link-payment-integrations?link-integrations=link-payment-method
- Stripe charges can include `payment_method_details.link`: https://docs.stripe.com/api/charges/object

cc @slavingia

PR #4905, opened 2026-04-30
feat(dev): add bin/doctor to diagnose local setup issues. V3

daniel-aranda:
## What

Adds `bin/doctor`, a read-only local setup diagnostic script for Gumroad development.

It checks:

- Ruby and rbenv setup
- Bundler version
- Redis, MongoDB, MySQL, and Elasticsearch
- Elasticsearch indices
- Nginx
- SSL certificates
- gumroad.dev availability
- Rails/AnyCable ports

## Why

Local setup can fail in several ways that are hard to diagnose from `bin/dev` output alone. This script provides a single command to identify missing services and misconfigurations before starting the app.

This reduces onboarding friction and avoids time spent debugging environment issues.

The script is intentionally read-only to avoid modifying a contributor’s environment unexpectedly.

## Before/After

**Before**

Contributors had to manually debug missing services, ports, certificates, and local routing issues.

**After**

Contributors can run `bin/doctor` to get a quick health check and suggested next steps.

Video:
https://www.loom.com/share/9c667df113444cd2be8203a0d3933ec1

Shows a failing setup (missing services) and how `bin/doctor` identifies issues, followed by a healthy run.

## Test Results

Ran `bin/doctor` locally in a healthy environment.

```bash
bin/doctor
```

Result: 
image

---

AI disclosure: I used ChatGPT GPT-5.5 Thinking as a review and refinement tool.

Prompts included (representative):
- "Review this bash script for a local dev environment doctor and suggest improvements"
- "Help identify missing checks for services like Redis, MongoDB, MySQL, Elasticsearch, and Nginx"
- "Improve CLI output clarity and developer experience for a diagnostic script"
- "Refine a pull request description following Gumroad contributing guidelines"

---

This is a fresh PR addressing feedback from @claude, @gumclaw and @gianfrancopiana from the former PR:
https://github.com/antiwork/gumroad/pull/4887
https://github.com/antiwork/gumroad/pull/4896


gumclaw:
@claude review

gumclaw:
Nice work on V3, Daniel. This addresses all the feedback from #4896:

1. **No more hardcoded values** — Ruby version from `.ruby-version`, Bundler from `Gemfile.lock`, service endpoints from `.env.development`, compose config from `Makefile`, ports from `Procfile.dev` and nginx config. Every check cites its source. ✅
2. **Agent skill included** — `.agents/skills/local-setup/SKILL.md` with a Claude symlink gives agents a structured workflow for using `bin/doctor` as a diagnostic probe. Addresses Gianfranco's suggestion directly. ✅
3. **Docker-first hints** — all failure hints point to `LOCAL_DETACHED=true make local`, matching the README setup path. ✅
4. **Read-only** — script doesn't modify anything, just reports. ✅

Code looks clean. The shell detection, service checks, and `source_info` output for traceability are all well done.

Triggered `@claude review` above. @ershadk for merge once CI is green.

daniel-aranda:
Hey @claude I addressed your comments, thank you.

gianfrancopiana:
This is a real improvement over the earlier versions and moves things in the right direction.

I'm going to close this PR for now because I don't think the current shape is the right thing to merge yet. The script has grown into a 565-line diagnostic CLI, and the default output feels too verbose for the happy path. For a command like `bin/doctor`, I'd expect the normal output to lead with a short verdict and the next action, with detailed source tracing available behind something like `--verbose`.

There are also still some CLI correctness issues to resolve before this should live in the repo (see Claude comments).

I think the better next step is a smaller follow-up PR that adds just the `local-setup` agent skill. The skill can read the repo sources of truth (`README.md`, `.ruby-version`, `Gemfile.lock`, `.env.development`, `Makefile`, `docker/docker-compose-local.yml`, `Procfile.dev`, nginx config) and then run targeted checks directly as needed.

I would start without a nested helper script. If we later find that the same low-level probes are being repeated often, we can add a small script inside the skill as an implementation detail. But I don’t think we need to commit a repo-level `bin/doctor` yet.

I appreciate the iteration here. I think the idea is useful, but this PR is now large enough that it’s better to reset the shape rather than keep patching this version.

PR #4908, opened 2026-04-30
Enable Stripe Link on checkout (CardElement approach)

gumclaw:
## What

Enables Stripe Link autofill in the checkout credit card form while keeping Gumroad's existing CardElement payment flow.

Buyers can use Link on checkout to fill their saved card details, but Gumroad still sends card-only PaymentIntents to Stripe and keeps the same server-side card handling. Link stays disabled anywhere the shared credit card input is used outside checkout.

## Why

This keeps the change small and limits the risk to the checkout UI. Gumroad's payment code stores and reads card-shaped Stripe data in several places, so enabling Link as a separate backend payment method would make the server receive data it is not built to handle safely.

Keeping Link inside CardElement gives buyers the autofill benefit without changing the charge-processing contract.

## Before/After

**Before:**

Checkout showed Stripe's card fields without Link autofill.

**After:**

Checkout shows Link through Stripe's CardElement.

## Demo
https://github.com/user-attachments/assets/8e3838e5-f3b7-4f93-b921-5bf920edf204

---


gumclaw:
@gianfrancopiana CI is green — 101/102 passed, the 1 failure is a pre-existing flaky radio button review test (unrelated to this PR). Just needs a manual QA for the checkout design with Link enabled and then it's ready to merge.

gianfrancopiana:
@claude review once

PR #4910, opened 2026-04-30
Fix unhandled Link::LinkInvalid when price exceeds maximum on product create

gumclaw:
## Problem

`Link::LinkInvalid: Sorry, the price entered is too large.` was hitting Sentry as an unhandled exception in `LinksController#create`.

**Sentry:** https://gumroad-to.sentry.io/issues/7452782475/
**Occurrences:** 1 (first seen 2026-04-30)
**Estimated users affected:** Low volume, but causes a 500 error instead of a user-friendly message.

## Root Cause

In `LinksController#create`, the `begin/rescue` block only wrapped `@product.save!`, but `Link::LinkInvalid` can also be raised earlier during:
- `current_seller.links.build(link_params)` — mass assignment calls `price_cents=`
- `@product.price_range = params[:link][:price_range]` — also calls `price_cents=`

Both of these happen before the `begin`, so the exception bubbled up uncaught.

## Fix

Moved the `begin` to wrap all attribute assignments that can raise `Link::LinkInvalid`, so the user sees a flash error and redirect instead of a 500.

## Test

Added a test confirming that creating a product with a price exceeding `MAX_PRICE_CENTS` redirects with an error flash instead of raising.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -81,8 +81,10 @@ class LinksController < ApplicationController
       if ai_generated
         generate_product_details_using_ai
       end
-    rescue ActiveRecord::RecordNotSaved, ActiveRecord::RecordInvalid, Link::LinkInvalid
-      return redirect_to new_product_path, alert: @product.errors.to_hash.transform_values(&:to_sentence).first, inertia: inertia_errors(@product)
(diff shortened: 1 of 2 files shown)


gumclaw:
@claude review

PR #4917, opened 2026-05-01
Construct IBAN from SWIFT and account number for El Salvador payouts

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad-private/issues/320. Supersedes #4640.

## What

El Salvador sellers can complete payout setup again. Form keeps PR #4640's UX (Account Number + SWIFT/BIC), but the backend now constructs an IBAN from those inputs before submitting to Stripe — because Stripe's API requires IBAN format for `country=SV`.

- Validation: accept either a 10-20 digit plain account number OR a structurally valid SV IBAN.
- Submission: `stripe_account_number(passphrase)` returns the IBAN form. Plain inputs get converted via ISO 13616 (mod-97 check digits, 4-letter bank code from SWIFT, zero-padded 20-digit account). IBAN inputs pass through.
- `StripeMerchantAccountManager#bank_account_hash` calls `stripe_account_number` when defined, otherwise falls back to raw decryption. Other countries unaffected.

## Why

PR #4640's frontend was correct per the issue's UX spec, but it sent plain account numbers to Stripe — which Stripe rejects for SV. Verified against live Stripe test mode:

| Submitted | Result |
|---|---|
| plain account + bank SWIFT (post-#4640) | ❌ *"You must use a test bank account number. Try `SV44BCIE12345678901234567890` …"* |
| `SV44BCIE…` + `AAAASVS1XXX` | ✅ |
| plain account + 9-digit routing (Schema 1 per [docs.stripe.com/payouts](https://docs.stripe.com/payouts)) | ❌ *"Invalid routing number for SV … must be in the format `AAAASVBB` or `AAAASVBBXYZ`"* |

The third row rules out the docs' "Routing Number" alternative. Only SWIFT/BIC + IBAN works.

Construction matches Wise's output bit-for-bit and Stripe's test fixture (`BCIESVS1 + 12345678901234567890 → SV44BCIE12345678901234567890`).

---

Generated with Claude Code Opus 4.7

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
@@ -438,10 +438,16 @@ module StripeMerchantAccountManager
       if bank_account.is_a?(CardBankAccount)
         Stripe::Token.create({ customer: bank_account.credit_card.stripe_customer_id }, { stripe_account: stripe_account["id"] }).id
       else
+        account_number_for_stripe =
+          if bank_account.respond_to?(:stripe_account_number)
+            bank_account.stripe_account_number(passphrase)
+          else
+            bank_account.account_number.decrypt(passphrase).gsub(/[ -]/, "")
+          end
         bank_account_hash = {
           country: bank_account.country,
           currency: bank_account.currency,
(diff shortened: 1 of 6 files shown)


nyomanjyotisa:
@ershad could you take a final look on this before merge?

PR #4925, opened 2026-05-01
Add /api/v2/meta endpoint for token introspection

CodyBontecou:
## What

Adds `GET /api/v2/meta`, a new authenticated endpoint that returns:

- the OAuth token's **scopes**
- the OAuth **application_name**
- the resource owner's **external_id**
- the **api version** and a static **documentation_url**

```json
$ curl -H "Authorization: Bearer $TOKEN" https://api.gumroad.com/v2/meta
{
  "success": true,
  "user": {
    "id": "2687846168522"
  },
  "token": {
    "scopes": [
      "view_sales",
      "edit_products",
      "view_profile"
    ],
    "application_name": "Local Test"
  },
  "api": {
    "version": "v2",
    "documentation_url": "https://app.gumroad.com/api"
  }
}
```

## Why

Today an API consumer holding an opaque OAuth token has no way to ask the API "what can this token do?" The only way to learn the scope set is to try a privileged endpoint and see if it 403s. For human integrations that's annoying; for AI agents (Claude Code, the gumroad-cli, third-party automations) it's a real blocker — they have to either guess capabilities or bake assumptions into client code.

This is the smallest possible step toward a discoverable API: one round trip, no migration, no new model, no PII beyond what the token holder already has access to.

This is a proof-of-concept for a broader API-discoverability effort (rate-limit headers, machine-readable error codes, a published OpenAPI spec). Each of those is independent and can land separately. This PR ships the smallest piece.

### Why this auth pattern

`before_action -> { doorkeeper_authorize!(*Doorkeeper.configuration.public_scopes.concat([:view_public])) }` matches `Api::V2::UsersController#show` — any token with a public scope (or the default `view_public`) is accepted. The endpoint only exposes metadata the token holder already knows about itself, so a tighter scope check would just be friction.

### Why no user email / name

Those are scope-gated on `/v2/user`. The meta endpoint is meant to be safe to call before knowing your scope set, so it returns only `external_id`. An agent that needs richer user info can call `/v2/user` once it knows it has `view_profile` or `account` scope — which it learns from `/v2/meta`.

## Before / After

https://github.com/user-attachments/assets/b4e3ed32-f57d-49a3-b58b-51c42262a594

### Showing the changes + skill using Claude Code:

https://github.com/user-attachments/assets/4004d087-3b12-449f-b333-99b283230340

## Test Results

CleanShot 2026-05-01 at 13 40 47@2x

---

**AI disclosure:** Drafted with Claude Opus 4.7 (1M context) 

Conversation History: https://isolated.tech/share/gswll9Ld

### DIFF
--- a/.agents/skills/gumroad-api-meta/SKILL.md
+++ b/.agents/skills/gumroad-api-meta/SKILL.md
@@ -1,0 +1,104 @@
+---
+name: gumroad-api-meta
+description: >
+  Discover what an OAuth token can do against Gumroad's API by calling GET /v2/meta. Use
+  when planning or debugging any Gumroad API integration: deciding which endpoint to call,
+  why a /v2/* request returned 401 or 403, what scopes a token has, who the token belongs
+  to, or what OAuth application minted it. Triggers on: "what scopes does this token have",
+  "why am I getting 403 from gumroad", "what can this gumroad token do", "introspect this
+  token", "what user owns this token", before calling any /v2/* endpoint for the first
+  time, or when designing a multi-step workflow against Gumroad's public API.
+---
+
+# Gumroad API meta
+
+`GET /v2/meta` is the token-introspection endpoint for Gumroad's v2 API. One round trip tells you everything you need to plan a workflow.
+
+## When to call it
+
+- **Before any other v2 call.** Cheaper to discover scopes once than to branch on 403s after the fact.
+- **On 403 from a known endpoint.** Confirm the token actually has the required scope. If it doesn't, surface a precise "this token lacks the `X` scope" message instead of a generic auth error.
+- **When stitching multi-step flows.** If a flow needs `view_sales` for step 1 and `edit_products` for step 2, check both up front — fail fast.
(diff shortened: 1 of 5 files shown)


gianfrancopiana:
Thanks for opening this and for taking the time to record a demo. Considering our current OAuth/token flows, I don’t think this endpoint has a strong enough use case to add as new public API surface right now.

For the first-party Gumroad CLI, this is not necessary: the CLI authenticates through a known Gumroad OAuth app and requests the scopes it needs, including `account`, so it already knows what it asked for.

For third-party API users, the existing flows expose the relevant scope information:
- manually generated tokens from a user-created Gumroad app receive all public scopes
- OAuth authorization requests choose a `scope`, and the `/oauth/token` response includes the granted `scope`

That leaves `/v2/meta` mainly helping clients that have a raw token but lost or did not store the OAuth metadata. That is a debugging convenience, but I don't think it justifies a new endpoint right now.

Going to close this for now.

PR #4926, opened 2026-05-01
Add audit logging for internal admin writes

gianfrancopiana:
Ref #4677

## What

Adds write audit logging for `/internal/admin` actions and uses the authenticated admin actor for write attribution.

Every wrapped admin write now records the actor, token, action, target, route, request id, redacted params, response status, and server error class when applicable. Existing shared-token traffic still resolves to the seeded legacy admin token and keeps the same `GUMROAD_ADMIN_ID` attribution, while per-actor tokens now leave their own actor id on comments and audit rows.

## Why

The admin CLI can perform support actions, but the server previously only knew that the shared bearer token was used. That made it hard to tell whether a write came from Gumclaw, a human, or another script.

This keeps existing callers working while making new per-actor tokens useful for attribution. The audit log is writes-only because these are the actions where forensic history matters most, and it stores a redacted params snapshot instead of a digest so bad writes can actually be investigated.

---

### DIFF
--- a/app/controllers/api/internal/admin/base_controller.rb
+++ b/app/controllers/api/internal/admin/base_controller.rb
@@ -2,6 +2,13 @@
 
 class Api::Internal::Admin::BaseController < Api::Internal::BaseController
   include AdminActor
+  include AfterCommitEverywhere
+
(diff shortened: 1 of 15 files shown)


slavingia:
Talking to Mac today, we need to support the same for Helper. Using Gumclaw to answer tickets = Mac, fully autonomous = Edgar

gianfrancopiana:
@claude review once

PR #4928, opened 2026-05-01
Add admin API token lifecycle support

gianfrancopiana:
Ref https://github.com/antiwork/gumroad/issues/4677

## What

This PR adds admin API token lifecycle support for the Gumroad CLI and internal admin clients.

Admins can approve a local CLI login in the browser using PKCE, exchange the one-time code for an expiring admin API token, check the current token with `whoami`, and revoke active tokens. The admin dashboard now shows active admin API tokens across CLI, service, and legacy token types, with actor, last-used, expiry, and revoke controls.

## Why

We need to move away from one shared legacy internal admin token toward credentials that are visible, attributable, and revocable.

Human CLI tokens are short-lived and reauthenticated through the browser flow. Service tokens can stay non-expiring for non-interactive clients like Gumclaw, but they are still tracked and revocable. The legacy token remains visible during the migration so we can monitor and retire it safely.

## Admin Page Screenshot
image

---

### DIFF
--- a/app/controllers/admin/api_tokens_controller.rb
+++ b/app/controllers/admin/api_tokens_controller.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-class Admin::Cli::TokensController < Admin::BaseController
+class Admin::ApiTokensController < Admin::BaseController
   def index
     set_meta_tag(title: "Admin API tokens")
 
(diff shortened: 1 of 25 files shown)


gianfrancopiana:
@claude review

PR #4939, opened 2026-05-04
Add Dollar Store discovery page

adhivd:
## What

Adds a dedicated Dollar Store discovery page at `/dollar-store` for browsing products priced at $1 or less.

The page uses the same Elasticsearch search pipeline as Discover but locks the price filter to `<= $1` and sorts by `HIGHEST_RATED`. It lives behind its own controller, Inertia page, and a self-contained `components/DollarStore/*` tree — no existing Discover, Profile, or search code paths are modified, aside from removing a `ProductSortKey::RANDOM` constant that was added during early prototyping and ended up unused.

New surfaces:
- `controllers/dollar_store_controller.rb` — Inertia endpoint, taxonomy filtering, NSFW preference handling, paginated results
- `pages/DollarStore.tsx` — grid, empty-state-by-taxonomy, load-more
- `components/DollarStore/Header.tsx` — light-mode header with mobile drawer (reusing `NestedMenu`) and a "Back to discover" hover tooltip on the Gumroad logo
- `components/DollarStore/CategoryFilter.tsx` — desktop popover with two-level taxonomy drill-down
- `components/DollarStore/Tile.tsx` — hover-driven product preview with details panel and rotating top reviews

## Why

Gumroad has a long tail of products priced $0–$1. The hypothesis is that surfacing them as a curated "dollar store" makes browsing feel exploratory and low-friction.

The interaction philosophy is *hover to learn more, click to commit*. Like a real dollar store, you can pick something up, look at it (price, description, what other people thought), and put it back without leaving the grid. The tile hover surfaces a details panel and top reviews that fade in over a few seconds; product clicks still go to the standard product page. The goal is to make low-stakes discovery feel fun and let the cheap-and-weird side of Gumroad shine through as a marketing surface.

## Scope cuts called out intentionally

- **No Discover entry point yet.** I drafted a banner under the Discover filter sidebar but pulled it because the only clean implementation required adding a slot to the shared `CardGrid` and restructuring its sticky sidebar behavior, which would affect Profile and search surfaces too. Worth a separate, more deliberate design pass on placement.
- One-click add-to-cart from the hover panel was scoped out.
- "More by this creator" peek and featured-product hero tiles in the grid were scoped out.
- Page is forced to light mode regardless of the user's system preference (the visual identity is "dollar store on white"); a dark-mode pass is doable but felt out of scope for v1.

## Demo

https://github.com/user-attachments/assets/cd947964-f5c8-4ab4-a1b6-37a204bd5e6f

Small screen / mobile phone demo:

https://github.com/user-attachments/assets/33ff9310-0c51-4684-9ba3-b58f64450227

## Screenshots

<details>
<summary>Desktop — tile hover</summary>

</details>

---


slavingia:
@claude review

adhivd:
Thanks for the review. Addressed in 38e7a505:

1. **Pagination bug** — ripped out `price_mode` entirely. It was leftover scaffolding from a $0/$1/$1+ filter I'd abandoned, never wired to any UI. Default behavior (free + paid up to $1) is preserved. Eliminates the page-1 vs. load-more divergence.
2. **Empty state CTA** — branched on whether a taxonomy filter is active. The unfiltered empty case now says "No dollar store products available right now" and links out to `/discover` instead of looping back to the same URL.
3. **Taxonomy split 500** — leaving as-is. The same pattern lives in `DiscoverController` and `LinksController`, so fixing it only in the dollar store would create inconsistency. Worth a separate hardening pass across all three sites.

gianfrancopiana:
@adhivd The new request spec is failing locally. `bundle exec rspec spec/requests/dollar_store_spec.rb` fails all 4 examples because the spec creates 50-cent paid products, but current product validation rejects non-free USD prices below $0.99. Can you check this?

Load more appears to start one item too early. The first page loads 36 products, then the next request can repeat the last product from page 1 and, for some totals, miss the final product.

Can you also add a mobile/small-screen demo?

adhivd:
Thanks for the note @gianfrancopiana, here's what I did:

1. Fixed in 08825b51: bumped the spec fixtures from 50¢ to 99¢ to satisfy `Link`'s `price_must_be_within_range` (USD minimum is 99¢ per `config/currencies.json`), still under $1 so still in scope for the dollar store.

2. Same commit (08825b51): the dollar store was emitting a 0-based `from` while the rest of the codebase (CardGrid reducer, `Searchable`'s `(from - 1).clamp`, `DiscoverController` seeding `from = COUNT + 1`) treats it as 1-based, which caused page 2 to re-fetch the last item of page 1; the boundary check also used `>` instead of `>=`, dropping the final product when `total == offset + page_size`.

3. Mobile walkthrough is up on the PR. While recording it I made a couple fly-by changes in 910917ff to clean up the mobile experience: the `CategoryFilter` pill is now desktop-only since the existing `NestedMenu` hamburger is already the mobile filter affordance (mobile gets a static "Showing [category] products" label instead, suppressed on the unfiltered home view); and the card price ribbon, which was rendering as a gray box on mobile because `[&_article_*]:border-black/15!` was bleeding into `PriceTag`'s inner `border-accent` triangle, now stays pink after replacing that with a `--color-border` override at the grid level that only affects elements consuming `border-border`.

topriko:
?

topriko:
This PR was closed without being merged?

PR #4945, opened 2026-05-04
Add Price Checker widget to the product editor

aseckin:
## What
Adds a Price Checker widget to the Pricing section of the product editor. Sellers click "Check prices" and see how their price compares to similar products on the platform: a histogram, the median line, and a marker for each of their prices (base + any paid variants).

states

## Why
For categories with high enough product volume and competition, strategic pricing can be an important factor. There's currently no easy way to understand how other products are priced within the category / type you're listed in.

## How
The widget hits a new authenticated endpoint (`POST /products/:id/price_check`) that returns an aggregated price distribution. The seller's live (unsaved) name, description, taxonomy, and product type ride in the request body as overrides, so editing those fields and clicking *Refresh* updates the result without saving first.

The backend filters Elasticsearch by product type, currency, subscription flag, and category (taxonomy); bundles, pay-what-you-want products, the seller's own products, and the product itself are always excluded. Name + description are matched as a **required** clause (not a boost) with `minimum_should_match: "30%"` so unrelated products with the same type/currency/category are filtered out, not just down-ranked. A two-pass ES query computes percentiles, picks a "nice" bin width, then runs a histogram aggregation clipped to p5–p95 so outliers don't compress the chart. If the strict (`with_taxonomy`) filter returns fewer than 5 matches we fall back to a broader (`broadened`) pool; below 5 even broadened, we show an `insufficient_data` state with a checklist of fields that improve match quality.

## Desktop Demo

https://github.com/user-attachments/assets/b9aa0dd2-3595-497a-a3e0-1e5189671456

### Mobile Demo

https://github.com/user-attachments/assets/9a85c588-1f56-45ac-a919-4236c8ce9bd7

## Notes
- This PR went through several rounds of feedback (initial review, design-engineer pass, and a consistency/cleanup pass) — easier to read commit-by-commit than as one big diff.
- Since dev had limited data, I built a temporary scenario cycle locally to QA the rendering and shoot the demos across realistic shapes; that scaffold has been removed from the final commit. The chart logic should still be stress-tested against real production data.
- Price Checker is hidden when name-your-own-price is toggled on; both for simplicity and because it doesn't rhyme well with the nature of PWYW.

### Backend
- New service `PriceCheckerService` (`app/services/price_checker_service.rb`) issues a two-call ES query: percentiles first (median, IQR, p5/p95 for outlier clipping), then a histogram aggregation with a "nice" interval picked from a fixed ladder ($1, $2.50, $5, $10, $25, $50, $100, …).
- Tier fallback: `with_taxonomy` → `broadened` (drop taxonomy filter) → `insufficient_data` when even the broadened pool has fewer than 5 matches. The tier is surfaced in the response so the UI can label the result honestly.
- Excludes the seller's own products, the current product, bundles, and pay-what-you-want.
- Accepts editor overrides (name, description, taxonomy_id, native_type). The controller sanitizes them — invalid `taxonomy_id` or `native_type` values are dropped silently rather than erroring.
- Result is cached in `Rails.cache` for 1 hour with a fingerprint key over the inputs that affect the result; `?refresh=1` busts the cache.
- New endpoint `POST /products/:id/price_check` on `LinksController`. Authorized via the existing `LinkPolicy#edit?` so role-based access mirrors the edit page.

### Elasticsearch
- Added three fields to the `products` index mapping: `price_currency_type` (keyword), `customizable_price` (boolean), `native_type` (keyword).
- The existing indexed boolean `is_subscription` (derived from `is_recurring_billing`) is reused; no model changes.
- Onetime task `Onetime::BackfillPriceCheckerIndexFields` populates the new fields without an AR callback (per CONTRIBUTING.md). Dev path: `DevTools.delete_all_indices_and_reindex_all`.

### Frontend
- New component tree under `app/javascript/components/ProductEdit/ProductTab/PriceChecker/` (`index`, `Checklist`, `DistributionChart`, `TierSubhead`).
- State machine: `idle → loading → ok | insufficient | error`.
- **Empty state** uses `<Placeholder>` (the dashed-bordered primitive that powers "Upload cover image" and "Add version") with a centered heading + tagline + CTA.
- **Loaded state** swaps to a solid-bordered container matching the `<Rows>` shape; the section header (title + info-icon tooltip + refresh button) lives inside the box.
- **Refresh** is an icon-only `<Button size="icon">` with a `RefreshCcw` icon and a "Refresh" tooltip — same pattern as the Trash button in `VersionsEditor`. In the chart state the slot toggles `invisible` to keep layout stable; in the insufficient state it stays visible and goes `disabled` instead.
- **Variant markers**: products with paid variants get base price + per-variant markers. Close-priced markers stack vertically rather than overlapping; clusters past 4 markers collapse; long variant names truncate to 12 chars. Memberships stay a single "Your price" marker.
- The info-icon trigger is a real `<button>` so the tooltip is keyboard-focusable and tap-friendly on mobile.
- Median and marker reference lines use a custom SVG label that sticks to the chart edge when the value lands at the boundary, with corner-omit on labels shifted off the line. Dashed lines render in a separate layer so they sit behind every label.
- *Refresh* is disabled until one of the inputs that affect matching changes (name, description, native type, category), tracked via a JSON fingerprint of those fields captured on each successful fetch.

### Tests
- `spec/services/price_checker_service_spec.rb` — 9 examples covering `with_taxonomy`, `broadened` fallback, `insufficient_data`, override fields, the relevance filter excluding unrelated products, ES null-percentile handling, exclusion rules, caching, and the cache key.
- `spec/controllers/links_controller_spec.rb` — 6 examples for the new POST action: auth, JSON shape, refresh-param passthrough, override sanitization, dropping invalid overrides, and ownership rejection.
- No frontend unit tests; the codebase tests UI via Capybara system specs and adding one for this would require enough seed data to deterministically hit each tier — flagging as a follow-up.

tests2

### Deploy notes
- The new ES mapping is non-breaking (index is `dynamic: :strict` but additive). After merge: run `Onetime::BackfillPriceCheckerIndexFields.process` to populate the three new fields on existing products. Without the backfill, the price-check endpoint returns `insufficient_data` for everyone (the `term` filters won't match documents missing the new fields).
- `app/javascript/utils/routes.{js,d.ts}` are git-ignored and regenerated on app boot via `rails js:export`.


slavingia:
@claude review

gianfrancopiana:
@aseckin Consider the following:

- The Recheck button can show results for the wrong version of the product. If a seller edits the name, description, or category and clicks Recheck before saving, the widget [records those new editor fields as checked](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/javascript/components/ProductEdit/ProductTab/PriceChecker/index.tsx#L53-L59), but the request [only sends the product permalink](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/javascript/data/price_distribution.ts#L53-L55), and the backend [loads the last saved product](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/controllers/links_controller.rb#L301-L307).

- The chart says it compares against similar products, but the product name and description do not actually decide which products are counted. They are [added to the query as an optional boost](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/services/price_checker_service.rb#L149-L183), while the chart is built from [price counts across the filtered products](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/services/price_checker_service.rb#L109-L117). So unrelated products with the same type/currency/category can still drive the result.

- Products with paid versions can show the wrong “Your price.” A product can have a $0 base price and a $5 paid version. Gumroad’s display price [includes the version price](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/modules/product/prices.rb#L166-L170), but the chart marker [uses only the base price](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/javascript/components/ProductEdit/ProductTab/PriceChecker/index.tsx#L148-L154), so it can show $0 instead of $5.

- The match details icon needs a real mobile/keyboard path. Right now it is [just an SVG inside a tooltip](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/javascript/components/ProductEdit/ProductTab/PriceChecker/index.tsx#L108-L110), and the tooltip [only opens on hover/focus](https://github.com/antiwork/gumroad/blob/cbe68a9bf29f07b615b11e20ced2236f9b11749b/app/javascript/components/WithTooltip.tsx#L58-L69). 

Can you take another pass at these and include a short mobile demo after the changes?

aseckin:
image

Thanks for the feedback. Updated PR with new demos for Desktop + Mobile. 
1. We now pass the editor's live values as overrides on every check, so unsaved edits drive the result instead of the saved record. If we think this approach is wrong, we can gate Recheck behind Save Changes - happy to switch.
2. Moved the name/description match from an optional `should` boost to a required `must` clause with `minimum_should_match: "30%"` against `name^3` and `description`. So unrelated products with the same taxonomy/currency/type are now filtered out before the histogram is built. 30% is a starting point and should be tested to be fine-tuned.
3. Variants with a paid price difference now render as their own markers (base + each paid variant). Close-priced markers stack vertically rather than overlapping, long variant names are truncated to 12 chars. 
4. The trigger is now a real button, so it's tab-focusable, activates with Enter/Space, and the tooltip is sized to stay inside the viewport on small screens.
---
Separately: while testing this I noticed the Amount input toggles PWYW on the moment you clear it, which kills the price checker mid-edit. If a seller is iteratively trying prices and clears the field to type a new one, the widget vanishes. Suggested fix: don't auto-fill "0" or flip the toggle when the field goes empty; only flip it if the user actually types "0", and only auto-fill "0" on Save if the field is still empty then.

slavingia:
The skeleton is a little strange and the fonts and stuff are a little small compared to the rest of the UI. Could we try using the placeholder vibes/component and increase the font sizes, etc?

I also would love one end-to-end test and think we should try to merge this ASAP!

aseckin:
@slavingia new commits focus on consistency & cleanup:
- Empty state now uses <Placeholder> matching the "Upload cover image" / "Add version" pattern
- "Recheck" button swapped for an icon-only RefreshCcw button with "Refresh" tooltip matching other dashboard sections
- Updated font styles and sizes to match dashboard patterns

I tried to handle various edge cases, but I'm sure testing with real user data will unearth some more cases to cover. 

edge-cases

gianfrancopiana:
@aseckin There's some layout shift between the initial state and the loaded chart:
CleanShot 2026-05-06 at 14 27 58@2x

CleanShot 2026-05-06 at 14 27 52@2x

Can you fix this so there's no layout shift?

aseckin:
Commit [#133ca15](https://github.com/antiwork/gumroad/pull/4945/commits/133ca15381909af977842aad5dccd7abd9046675) standardizes card height across three states (initial - chart - empty). I have concerns about this implementation, but it was the only option that worked well enough so far.

The issue is, card height is derived from the chart's aspect-ratio: 2.4 at the column's full width, so wider columns produce taller cards. That works well in the chart state (arguably preferable) but produces noticeable empty space in the CTA and insufficient-data states. Attaching some GIFs below highlighting the worst cases. Any suggestions?

> For reviewers: the calc in CARD_MIN_HEIGHT (min-h-[calc((100cqw-34px)/2.4+126px)]) encodes structural facts about the loaded card — the chart's aspect ratio (2.4) from DistributionChart, the wrapper's padding + border (34px), and the header + gaps + subhead (126px). The constant name hints at intent but the magic numbers don't self-explain. CONTRIBUTING.md's no-explanatory-comments rule means I can't annotate it inline; if anyone touches the card's padding, header height, or chart aspect later, this constant will silently drift and the layout shift will quietly come back.

aseckin:
The solution in commit [#133ca15](https://github.com/antiwork/gumroad/pull/4945/commits/133ca15381909af977842aad5dccd7abd9046675) was a poor decision, namely due to me treating the 2.4 aspect ratio as a requirement to render the chart meaningfully. I replaced that with a stepped `min-height` that scales across breakpoints, paired with a flexible grid row so the chart fills whatever vertical space remains. Now things look much better across all breakpoints + there's no jumping between states. Here's a demo of how it looks scaling down from a really large viewport

https://github.com/user-attachments/assets/076b48c1-81c4-4e19-a5ac-afe2ce8ac3f4

aseckin:
As we move forward with this, I'd like to note some things:
- The issue I mentioned above (emptying the price field & PWYW toggling on) is a real inconvenience and I'd like to implement the solution I proposed to work around it.
	- When Price Checker is ran and is showing the chart, don't auto-fill "0" or flip the toggle when the field goes empty; only flip it if the user actually types "0", and only auto-fill "0" on Save if the field is still empty then.
- In the case: insufficient_data state -> user refreshes after some edits -> still insufficient_data, there should be visual feedback to indicate that the check successfully ran but results were similar. There are several ways we can go about this:
	- (1) When Refresh button is hit it goes into a spinner state, and when the result comes back it highlights the container visually for a brief period (I could not find a similar pattern in the codebase)
	- (2) We silently run the price checker as the user fills in data, and only show the Check Prices CTA when there are actual matches
	- (3) We silently run the price checker, and when it has matches it automatically displays them, skipping the Check Prices state altogether
	- I would normally lean towards implementing option 1, but given we want this merged asap and this is an edge case that's not really harmful, I'm OK shipping as it is.
- Speaking of user behavior:
	- We should implement a GA4 custom event to track `price_check_run`s
	- I also recommend a silent launch behind a feature flag for admins, and testing extensively with prod data how the chart behaves, seeing the cases it's more useful than others, or the cases where it breaks. I expect some followup adjustments on collision preventing logic, or the way variants are displayed

gianfrancopiana:
@aseckin — I ran the service shape against prod data on the read replica (no UI). Five things I'd want resolved before merge.

## 1. The widget will be empty for bundle products

In `base_query`, the filter set includes both `term: { is_bundle: false }` and (post-backfill) `term: { native_type: effective_native_type }`. I verified on the prod read replica that `native_type = "bundle"` and the `is_bundle` flag are effectively equivalent. They overlap in nearly every case (a handful of stale-data anomalies aside, and zero rows where `is_bundle` is true without `native_type = "bundle"`).

So for any seller editing a bundle, the candidate set is essentially empty — `match_count` will be 0 and the widget will permanently render `insufficient_data`. We should hide the widget for bundles in `ProductTab/index.tsx` (same shape as the existing PWYW carve-out).

## 2. Cache key window (500) ≠ relevance-query window (1000)

In [`price_checker_service.rb`](app/services/price_checker_service.rb):

```ruby
def relevance_query
  @relevance_query ||= [effective_name.to_s, effective_description.to_s.first(1_000)]  # 1000
    .map { |s| ActionController::Base.helpers.strip_tags(s).strip }
    ...
end

def cache_key
  fingerprint = Digest::MD5.hexdigest(
    [
      effective_name,
      effective_description.to_s.first(500),  # 500
      ...
```

If a seller's description is longer than 500 chars, edits in chars 501–1000 change the ES query inputs but don't change the cache key. The cache returns the stale histogram for up to an hour unless the seller manually clicks Refresh (which sends `?refresh=1`). We can probably drop the description from the cache key entirely.

## 3. Pre-backfill, the widget is empty for everyone

If this ships before `Onetime::BackfillPriceCheckerIndexFields.process` finishes a full pass, every seller sees `insufficient_data`. Worth wiring this to the deploy plan explicitly. Gating the widget on a flag/setting that flips after the backfill finishes is safer than the deploy-note approach.

## 4. ES p95 is ~2–3 s, with no timeout in the controller path

I ran the two-pass query against a sample of top-selling products spanning native_types and taxonomies in prod. `Link.search` calls are sequential and there's no `body[:timeout]` or wrapping `Timeout.timeout`. ES `took` measurements (server time only):

| profile | percentiles ms | histogram ms | total |
|---|---:|---:|---:|
| typical case (small/medium pool) | 400–800 | 300–800 | ~1 s |
| broad pool with strong text relevance | ~1,300 | ~770 | ~2 s |
| largest pool I hit | ~1,940 | ~1,260 | ~3.2 s |

Please add an explicit `body[:timeout] = "2s"` (or similar) and a UI fallback when ES trips it, so a slow shard can't hold a Puma worker for 3+ seconds. The two queries are independent in their filter set. Running them via `_msearch` cuts wall-clock to whichever is slower.

## 5. The "with_taxonomy" tier label is misleading for products in "other"

The "other" taxonomy is by far the largest bucket of recommendable alive products on the platform. For sellers in that bucket, the `with_taxonomy` tier is filtering against a peer group that is, in practice, everyone. The UI implies this is a category-specific suggestion. I think we should omit the taxonomy badge when `taxonomy_slug == "other"`.

---

Investigation done with Claude Opus 4.7 (1M context).

gianfrancopiana (changes requested):
See earlier comment.

aseckin:
@gianfrancopiana thanks, will address these. thoughts on adding a GA4 event for `price_check_run`s?

slavingia:
Makes sense.

gianfrancopiana:
@claude review once

gianfrancopiana:
@aseckin Before merge, can we add a migration to apply the new ES mapping? The products index is `dynamic: :strict` and `native_type`, `customizable_price`, and `price_currency_type` aren't there yet, so the backfill's reindex jobs will fail until something runs `put_mapping`.

Same shape as [20251121152058_add_offer_codes_to_products_index.rb](../db/migrate/20251121152058_add_offer_codes_to_products_index.rb):

```ruby
class AddPriceCheckerFieldsToProductsIndex < ActiveRecord::Migration[7.1]
  def up
    EsClient.indices.put_mapping(
      index: Link.index_name,
      body: { properties: {
        price_currency_type: { type: "keyword" },
        customizable_price:  { type: "boolean" },
        native_type:         { type: "keyword" },
      }}
    )
  end
end
```

aseckin:
@gianfrancopiana can you re-run the tests?

gianfrancopiana:
@aseckin A few remaining items:

1. Currency is a live editor field, but Price Checker doesn’t include it in the request inputs or stale-result fingerprint. Repro: run Price Checker on a USD product, change the currency selector to EUR without saving, and the existing chart relabels the old USD distribution as EUR. Refresh also stays hidden/disabled because the fingerprint only tracks native type, name, description, and taxonomy. Can we include currency in the fingerprint + overrides, sanitize it in the controller, and use the effective currency for the ES filter/cache key/response?

2. The feature flag actor differs between render and request. The edit page uses `Feature.active?(:price_checker, product.user)`, but `LinksController#price_check` checks `current_seller`. For collaborators, that can show the widget and then return 404 from the endpoint. Can we use the same actor on both surfaces, probably `@product.user`, or add a spec if collaborators are intentionally excluded?

3. The failing CI job looks related, not flaky. `spec/presenters/product_presenter_spec.rb` is failing because `edit_props` now includes `price_checker_enabled: false`, but the strict expected hashes weren’t updated. Should be enough to add that key to the three expected hashes, plus an enabled-flag case if you want coverage for the new prop.

aseckin:
Commit [765240d](https://github.com/antiwork/gumroad/pull/4945/commits/765240d047f23a247a96a33a5831327b0d2fa4a0):
1. Currency is now part of the overrides + the frontend's `checklistFingerprint`. Controller sanitizes it (downcased, validated against `CURRENCY_CHOICES`). Service uses `effective_currency` for the ES filter, cache key, and response. The change-currency-without-save scenario now bumps the fingerprint and busts the cache key — Refresh enables and a fresh query runs.
2. Flag actor aligned on the controller side: `current_seller` → `@product.user`, matching the presenter. Collaborators no longer see the widget then 404 the endpoint.
3. The three exact-match `edit_props` fixtures already got `price_checker_enabled: false` in 20b8fc9. This commit adds the on-state coverage you suggested — new context in `product_presenter_spec.rb` asserting `price_checker_enabled: true` when the flag is enabled for the seller.

gianfrancopiana (approved):

slavingia:
Email? If you'd like to come to NYC for 1-2 weeks and co-ship together.

sahil.lavingia@gmail.com

aseckin:
> Email? If you'd like to come to NYC for 1-2 weeks and co-ship together.
> 
> [sahil.lavingia@gmail.com](mailto:sahil.lavingia@gmail.com)

Email sent (atakanseckin@gmail.com)

PR #4948, opened 2026-05-04
Remove Stripe verification popup on Settings/Payments

gianfrancopiana:
Fixes antiwork/gumroad-private/issues/327

## What

Replace the Stripe Connect embedded notification banner on Settings → Payments with a native Gumroad alert linking to `/settings/payments/remediation`, and switch that route's `Stripe::AccountLink` from `type: "account_onboarding"` to `type: "account_update"`. Same-tab redirect to Stripe's account update page. No popup, no "Link your Stripe account" wording.

Deletes the now-unused `StripeAccountSessionsController`, `StripeConnectEmbeddedNotificationBanner` component, and the `@stripe/connect-js` / `@stripe/react-connect-js` packages.

## Why

A Spanish self-employed seller [reported](https://github.com/antiwork/gumroad-private/issues/327) that clicking "Update business information" on the Payments page opened a confusing "Link your Stripe account" popup, even though they'd configured payouts via bank transfer and never knowingly created a Stripe account. The popup is Stripe's hosted onboarding fallback — it fires whenever the embedded `ConnectNotificationBanner` is clicked without an embedded `account_onboarding` peer mounted, which is our setup today.

Two fixes were considered:

- **Embedded onboarding** — add the `ConnectAccountOnboarding` peer so the banner click triggers an inline flow. Verified locally: it changes the popup *content* (Gumroad-branded "Let's get started" instead of "Link your Stripe account") but doesn't eliminate the popup, because Stripe opens a one-time auth window for the embedded session.
- **Account update redirect** (this PR) — drop the embedded banner entirely and redirect to Stripe's `account_update` page in the same tab. No popup at all. Stripe's `account_update` flow shows a "Verification summary" of the existing account (not "Link your Stripe account"), which reads correctly to a seller who only ever filled in a bank form.

The redirect approach is also smaller and removes a `Stripe::AccountSession.create` API call from the Settings/Payments hot path.

## Before/After

|  | Before | After |
| --- | --- | --- |
| Settings → Payments | image | image |
| Clicking the action | image | image |

After completing on Stripe, the user returns to Gumroad with a "Thanks! You're all set." flash and the alert is gone.

---


gianfrancopiana:
@claude review

PR #4953, opened 2026-05-05
Capture failure_reason and error_message for Stripe payout failures

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad-private/issues/334

## What

When a Stripe payout fails, the rescue branches in `StripePayoutProcessor.perform_payment` only set `failure_reason` for three hardcoded message substrings (`Cannot create live transfers`, `Debit card transfers are only supported`, `Insufficient funds in Stripe account`). Every other Stripe error - any unmatched `Stripe::InvalidRequestError`, any `Stripe::StripeError` subclass, and auth/connection errors - falls through to `mark_failed!(nil)` and the `failure_reason` column ends up null in the DB. The error goes to Sentry, but with no Gumroad-side tag tying it to a user, triage requires correlating by external IDs (e.g. `ba_…`).

This PR:

- Adds `Payment::FailureReason::BANK_ACCOUNT_NOT_FOUND_AT_STRIPE` plus a `STRIPE_FAILURE_SOLUTIONS` entry directing the seller to re-add their bank account. Maps the Stripe error string `has been deleted and can no longer be used` (the canonical signal Stripe sends when an external account ID has been replaced - usually after a Financial Connections session, e.g. Stripe Capital onboarding).
- Refactors the substring matching in `perform_payment` into a private `stripe_invalid_request_error_failure_reason` helper.
- When that specific reason fires, also calls `payment.bank_account.mark_deleted!`. Without this, the nightly run will keep producing failed `Payment` rows against the same dead reference until the seller manually re-adds the account.
- Adds an `attr_json_data_accessor :error_message` on `Payment` and stores `e.message` (truncated to 1000 chars) in `payment.json_data["error_message"]` in every Stripe rescue branch (both `perform_payment` and `prepare_payment_and_set_amount`). So next time an unmatched Stripe error fires, the message is in the DB without needing Sentry.
- Updates `BankAccount#supports_instant_payouts?` to skip `ErrorNotifier.notify` for the same `has been deleted and can no longer be used` message - that method runs on every balance/settings page render, so a stale `ba_…` was generating tens of Sentry events per day per affected seller.

## Why

A real seller had four consecutive payout failures over 9 days (Apr 26 - May 5) with `failure_reason: nil`. Their Stripe Connect account was healthy, the bank account on Stripe was healthy, but Stripe had rotated the external account ID via Financial Connections during Stripe Capital onboarding. Our DB still pointed at the old ID, every payout attempt got rejected as deleted-bank-account, and the rescue path silently dropped both the reason and the message. The seller's settings page also generated 31+ Sentry events from the same root cause via `supports_instant_payouts?`.

---

AI disclosure: drafted with Claude Opus 4.7 (1M context)

### DIFF
--- a/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
+++ b/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
@@ -133,6 +133,7 @@ class StripePayoutProcessor
   #   * Setting the amount_cents.
   # Returns an array of errors.
   def self.prepare_payment_and_set_amount(payment, balances)
+    failed = false
     merchant_account, balances_held_by_gumroad, balances_held_by_stripe = get_payout_details(payment.user, balances)
 
     if merchant_account.nil?
@@ -229,6 +230,8 @@ class StripePayoutProcessor
   # Public: Actually sends the money.
   # Returns an array of errors.
   def self.perform_payment(payment)
+    failed = false
+    failure_reason = nil
     # We have transferred the balance held by gumroad to the connected Stripe standard account.
     # No payout needs to be issued in this case.
     merchant_account = payment.user.merchant_accounts.find_by(charge_processor_merchant_id: payment.stripe_connect_account_id)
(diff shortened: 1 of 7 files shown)


nyomanjyotisa:
@claude review

PR #4969, opened 2026-05-06
Send member-cancels workflow to past canceled members

gianfrancopiana:
Fixes antiwork/gumroad-private#339

## What

For workflows triggered by "Member cancels", publishing with "Also send to past members who canceled" enabled did not actually send to past canceled members. The toggle persisted on the workflow record, but `Workflow#schedule_installment` bailed out for any non-new-customer trigger, so no retroactive backfill was ever enqueued.

A new `SendWorkflowEmailsToPastCanceledMembersJob` runs on publish for member-cancels workflows. It iterates the seller's (or product's) canceled subscriptions, applies the workflow filters, and schedules the existing `SendWorkflowInstallmentWorker` per subscription at `deactivated_at + delay`. Past-dated schedules fire immediately; the worker's existing `SentPostEmail.ensure_uniqueness` dedup prevents double-sends if the workflow is later unpublished and re-published.

## Why

`SendWorkflowPostEmailsJob` is purchase/follower/affiliate-centric and has no subscription path. The only existing scheduler for member-cancels sends ran reactively on `Subscription#deactivate!`, which only fires for new cancellations. A separate, subscription-centric job keeps the two paths cleanly separated and reuses the per-subscription delivery logic already in `Installment#send_installment_from_workflow_for_member_cancellation`.

Verified in prod against the reporting seller's "Vault Cancelled" workflow: 61 eligible past-canceled members existed; zero `SentPostEmail` rows had been created for the published installment.

---

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -91,8 +91,16 @@ class Workflow < ApplicationRecord
   def schedule_installment(installment, old_delayed_delivery_time: nil)
     return if installment.abandoned_cart_type?
     return unless alive?
-    return unless new_customer_trigger?
     return unless installment.published?
+
+    if member_cancellation_trigger?
+      if send_to_past_customers? && old_delayed_delivery_time.nil?
+        SendWorkflowEmailsToPastCanceledMembersJob.perform_async(installment.id)
+      end
+      return
+    end
+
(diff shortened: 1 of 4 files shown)


gianfrancopiana:
@claude review

PR #4970, opened 2026-05-06
Add internal admin watchlist API

gumclaw:
Ref antiwork/dashboards#34

## What

Adds internal admin API support for the full watchlist workflow that already exists in the admin UI.

The API can now return the current watch status from `users/info`, add a user to the watchlist, update the active watch threshold or notes, and remove the active watch. Threshold parsing also rejects non-finite values like `Infinity` and `NaN` instead of returning a server error.

## Why

Risk reviews need a watched state for accounts that should be monitored without blocking payouts. The admin UI already supports adding, updating, and removing watchlist entries, and Gumclaw/the Gumroad CLI need the same capabilities through the internal admin API.

Keeping read, create, update, and remove as separate operations makes the CLI behavior explicit instead of relying on a write endpoint to infer status.

## Before/After

API-only change; no user-facing UI changed. The internal API now mirrors the existing admin watchlist flow.

## Test Results

- `bundle exec rubocop --force-exclusion -a app/controllers/api/internal/admin/users_controller.rb spec/controllers/api/internal/admin/users_controller_spec.rb spec/routing/api_internal_admin_contract_spec.rb`
- `bundle exec rspec spec/controllers/api/internal/admin/users_controller_spec.rb spec/routing/api_internal_admin_contract_spec.rb spec/models/watched_user_spec.rb` - 112 examples, 0 failures
- `bin/test-confidence` - 99% confidence, 5 tests passed

---

This PR was implemented with AI assistance using gpt-5.5 xhigh and Opus 4.7

### DIFF
--- a/app/controllers/api/internal/admin/users_controller.rb
+++ b/app/controllers/api/internal/admin/users_controller.rb
@@ -311,7 +311,10 @@ class Api::Internal::Admin::UsersController < Api::Internal::Admin::BaseControll
     end
 
     def parse_threshold_cents(raw)
-      cents = (BigDecimal(raw.to_s) * 100).round
+      threshold = BigDecimal(raw.to_s)
+      return nil unless threshold.finite?
+
+      cents = (threshold * 100).round
       cents.positive? ? cents : nil
     rescue ArgumentError
       nil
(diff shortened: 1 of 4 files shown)


gianfrancopiana:
@claude review

PR #4977, opened 2026-05-07
Admin API: support user lookup by external_id

nyomanjyotisa:
Closes antiwork/gumroad-private#343

## What

Adds an optional `external_id` parameter as an alternative to `email` for identifying the target user on every internal admin and helper user endpoint that currently only accepts `email`:

- `POST /api/internal/admin/users/info`
- `POST /api/internal/admin/users/suspension`
- `POST /api/internal/admin/users/reset_password`
- `POST /api/internal/admin/users/update_email` (alongside `current_email`)
- `POST /api/internal/admin/users/two_factor_authentication`
- `POST /api/internal/admin/users/create_comment`
- `POST /api/internal/admin/users/mark_compliant`
- `POST /api/internal/admin/users/suspend_for_fraud`
- `POST /api/internal/admin/users/watch`
- `POST /api/internal/admin/users/update_watch`
- `POST /api/internal/admin/users/unwatch`
- `POST /api/internal/helper/users/create_comment`

Lookup behavior matches the admin UI at `/admin/users/:external_id` — `User.find_by(external_id: …)`. When both `email` and `external_id` are supplied, `external_id` wins. When neither is supplied, the endpoints return `400` (`"email or external_id is required"` for admin; `"'email' or 'external_id' parameter is required"` for the helper, matching its existing message style). Existing email-only callers keep working unchanged.

`info` continues to allow lookup of soft-deleted users (it never used `User.alive`); all mutations keep the `User.alive` scope, so an `external_id` resolving to a soft-deleted user returns `404` consistently with the email path.

The other 6 helper actions (`user_info`, `user_suspension_info`, `send_reset_password_instructions`, `update_email`, `update_two_factor_authentication_enabled`, `create_appeal`) are out of scope per the issue and remain email-only.

## Why

The admin UI loads suspended/soft-deleted/hard-to-find accounts fine at `/admin/users/2245593582708`, but the API returned `404` because the only handle into the lookup was `email`, which:

- Doesn't resolve via the API for accounts the email path can't find (suspended-for-fraud accounts, accounts not yet replicated to Metabase, accounts whose email has changed).
- Forces support to skip recording fraud-review notes on suspended accounts — exactly when an audit trail matters most.

The motivation applies more strongly to mutations (`create_comment`, `mark_compliant`, `suspend_for_fraud`, `update_email`) than to read-only `info`, so I extended `external_id` support to all 11 admin actions rather than just the 3 listed in the issue. Marginal cost was near zero (the lookup is centralized in `find_user_or_render`) and a split contract (3 endpoints support it, 8 don't) would be confusing. Confirmed in planning before writing code.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) Model: Claude Opus 4.7 (

### DIFF
--- a/app/controllers/api/internal/admin/users_controller.rb
+++ b/app/controllers/api/internal/admin/users_controller.rb
@@ -28,7 +28,7 @@ class Api::Internal::Admin::UsersController < Api::Internal::Admin::BaseControll
 
   def reset_password
     return unless require_user_lookup_params!
-    if params[:email].present? && !EmailFormatValidator.valid?(params[:email])
+    if params[:external_id].blank? && params[:email].present? && !EmailFormatValidator.valid?(params[:email])
       return render json: { success: false, message: "Invalid email format" }, status: :bad_request
     end
 
--- a/app/controllers/api/internal/admin/users_controller.rb
+++ b/app/controllers/api/internal/admin/users_controller.rb
@@ -1,19 +1,21 @@
 # frozen_string_literal: true
 
 class Api::Internal::Admin::UsersController < Api::Internal::Admin::BaseController
-  def info
-    return render json: { success: false, message: "email is required" }, status: :bad_request if params[:email].blank?
+  USER_LOOKUP_BAD_REQUEST_MESSAGE = "email or external_id is required"
 
-    user = User.by_email(params[:email]).first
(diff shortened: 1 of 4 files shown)


nyomanjyotisa:
@claude review

PR #4978, opened 2026-05-07
Add internal admin products API for risk reviews

nyomanjyotisa:
Closes #4973.

## What

Adds two read-only endpoints to the existing internal admin API so the risk-review CLI can inspect a seller's products without falling back to Metabase:

- `POST /api/internal/admin/products/list` — paginated list of a seller's products (alive + deleted). Accepts `email` **or** `external_id` to identify the seller (mirrors #4977).
- `GET /api/internal/admin/products/:id` — single product by external id.

Both return product metadata plus a `files` array with display name, S3 filename, extension, filegroup, byte size, and per-file `deleted_at`.

When both `email` and `external_id` are supplied to `list`, `external_id` wins. When neither is supplied, returns `400` with `"email or external_id is required"` — same shape as #4977.

## Why

Risk reviewers currently have to switch to Metabase SQL to look at a seller's products mid-investigation, and **file metadata isn't reachable from any current tool** — neither admin web UI nor Metabase exposes file size or extension. Whether a $200 "ebook" is a 2-page PDF or a 68-page guide is the signal that decides whether to flag or clear; without it the review is guesswork.

This stays inside the established `Api::Internal::Admin::*` controller pattern (bearer-token auth via [base_controller.rb](app/controllers/api/internal/admin/base_controller.rb)) so the existing CLI just gets two new commands rather than a new auth path.

The `external_id` lookup option exists for the same reason it does in #4977: risk reviewers most need this endpoint when a seller is suspended-for-fraud, soft-deleted, or has had their email rotated — exactly the cases where the email path can't resolve them. `User.find_by(external_id: ...)` is used without the `.alive` scope, matching the existing email path's permissive behavior so a deleted seller's products are still inspectable.

---

### AI disclosure

Implemented with **Claude Opus 4.7** in Claude Code

### DIFF
--- a/app/controllers/api/internal/admin/products_controller.rb
+++ b/app/controllers/api/internal/admin/products_controller.rb
@@ -74,7 +74,7 @@ class Api::Internal::Admin::ProductsController < Api::Internal::Admin::BaseContr
           id: product.user&.external_id,
           email: product.user&.email
         },
-        files: product.product_files.in_order.map { serialize_file(_1) }
+        files: product.product_files.sort_by { |f| [f.position || Float::INFINITY, f.id] }.map { serialize_file(_1) }
       }
     end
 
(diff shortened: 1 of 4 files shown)


nyomanjyotisa:
@claude review

PR #4981, opened 2026-05-07
Add comments index for comment_type and created_at lookups

ershad:
## What

Adds a composite index on `comments` covering `(comment_type, commentable_type, created_at, commentable_id)`.

## Why

Queries that filter comments by `comment_type` over a recent `created_at` window currently have no usable index for those columns. The existing `comments` indexes only cover `commentable_id` / `commentable_type` / `idempotency_key` / `ancestry` / `purchase_id`.

`EXPLAIN` on the suspended-user dashboard query shows MySQL using only `commentable_type` from `index_comments_on_commentable_and_idempotency_key` (`key_len=1023`) — so it scans every \"User\"-typed comment ever written and applies `comment_type` / `created_at` / `deleted_at` as inline filters.

The new index lets the planner seek directly to `comment_type` rows in the relevant time window before joining, dropping the comments scan from millions of rows to dozens.

Column order rationale:
- `comment_type` — equality, leftmost
- `commentable_type` — equality
- `created_at` — range, after equalities
- `commentable_id` — trailing, lets the index cover the join back to `users.id`

`deleted_at` is intentionally omitted: most suspended comments aren't soft-deleted, and the trailing inline filter is cheap.

The index name is set explicitly because the auto-generated name (85 chars) exceeds MySQL's 64-char identifier limit.

## Before/After

_TODO: attach EXPLAIN ANALYZE output before and after the index on staging._

## Test Results

_TODO: attach screenshot of test suite passing locally._

---

AI disclosure: Drafted with Claude Opus 4.7 (1M context). Prompts: identifying the bottleneck via `EXPLAIN`, drafting the migration following the repo's existing index migration style, and writing this description.


ershad:
Will reopen on Monday.

ershad:
Running this migration manually to avoid blocking production deployments. Will merge after successfully completing the migration.

PR #4982, opened 2026-05-07
Handle MerchantRegistrationUserAlreadyHasAccountError race condition in PaymentsController

gumclaw:
## Problem

`Settings::PaymentsController#update` raises an unhandled `MerchantRegistrationUserAlreadyHasAccountError` due to a race condition.

**Sentry issue:** https://gumroad-to.sentry.io/issues/7465850997/

The controller checks `merchant_accounts.stripe.alive.empty?` outside a database lock, then calls `StripeMerchantAccountManager.create_account` which re-checks inside `with_lock`. When two concurrent requests hit this endpoint, both pass the controller guard, but the second fails inside `create_account` with `MerchantRegistrationUserAlreadyHasAccountError`.

The rescue clause only catches `Stripe::StripeError` and `MerchantRegistrationUserNotReadyError`, so this error bubbles up unhandled.

## Fix

Add a rescue for `MerchantRegistrationUserAlreadyHasAccountError` that treats it as a success (the merchant account exists, which is the desired end state). This matches how the admin controller already handles this error.

## Impact

- **Sentry events:** 1 (first occurrence)
- **Estimated affected users:** Low, requires exact concurrent request timing
- **Support tickets:** Likely minimal, but users see a 500 error when it occurs

## Video

Non-user-facing backend error handling change. The test demonstrates the fix: when `create_account` raises `MerchantRegistrationUserAlreadyHasAccountError`, the controller now redirects with a success notice instead of raising.

### DIFF
--- a/app/controllers/settings/payments_controller.rb
+++ b/app/controllers/settings/payments_controller.rb
@@ -81,6 +81,8 @@ class Settings::PaymentsController < Settings::BaseController
     if current_seller.active_bank_account && current_seller.merchant_accounts.stripe.alive.empty? && current_seller.native_payouts_supported?
       begin
         StripeMerchantAccountManager.create_account(current_seller, passphrase: GlobalConfig.get("STRONGBOX_GENERAL_PASSWORD"))
+      rescue MerchantRegistrationUserAlreadyHasAccountError
+        # A concurrent request already created the merchant account; treat as success.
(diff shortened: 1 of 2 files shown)


ershad:
Extremely rare, can skip it for now.

gumclaw:
@claude review

PR #4984, opened 2026-05-07
Replace support ticket UI with mailto: flow

gumclaw:
## What

Replaces the Helper API ticket creation flow with a simple `mailto:` link. The modal UI is preserved (subject, message fields, inline validation) but instead of creating tickets via the Helper API, clicking submit opens the user's email client with a pre-filled email to support@gumroad.com.

## Why

Simplifies the support flow. Both sides (user and support) get email as the tracker. We can still do inline validation/checking in the modal before the mailto opens.

## Changes

- **NewTicketModal** (authenticated users): Removed `@helperai/react` SDK calls, attachment UI. Now builds a `mailto:` URL and opens it.
- **UnauthenticatedNewTicketModal** (anonymous users): Removed reCAPTCHA, removed POST to `/support/create_unauthenticated_ticket`. Prepends "From: {email}" to the body.
- **SupportPortal**: Updated callback (no conversation slug).
- **HelpCenter Layout**: Updated callback, removed recaptchaSiteKey prop.
- **Header**: Button text → "Email support"

## What's NOT changed

- ConversationList, ConversationDetail, SupportPortal pages still work for viewing existing Helper tickets
- Helper provider/session still loads for existing conversations
- SLA message component preserved

## Testing

- [x] Modals open/close properly
- [ ] mailto links properly encoded (subject + body)
- [ ] "Support tickets" tab still works for existing conversations
- [ ] Anonymous flow includes email in body

### DIFF
--- a/app/javascript/components/server-components/support/Header.tsx
+++ b/app/javascript/components/server-components/support/Header.tsx
@@ -12,11 +12,9 @@ import { useOriginalLocation } from "$app/components/useOriginalLocation";
 export function SupportHeader({
   onOpenNewTicket,
   hasHelperSession = true,
-  recaptchaSiteKey,
 }: {
   onOpenNewTicket: () => void;
   hasHelperSession?: boolean;
-  recaptchaSiteKey?: string | null;
 }) {
   const { pathname, searchParams } = new URL(useOriginalLocation());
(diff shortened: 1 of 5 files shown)


slavingia:
Doesn't support attachments. Maybe we still send out an email and then Helper is replying on that specific thread.

PR #5000, opened 2026-05-10
Fix confusing UX for SSN/tax ID field after entry

gumclaw:
## Problem

After entering their SSN on the payments settings page, users see an empty editable input field with a "Hidden for security" placeholder. This makes it look like the SSN update didn't go through ([screenshot from issue](https://github.com/user-attachments/assets/65b5abdd-565e-414f-89d1-a056f45e2f5c)).

## Solution

When a tax ID (SSN, EIN, or any individual/business tax ID) has already been entered:

- Show a **disabled masked field** with bullet characters instead of an empty input
- Add an **eye icon toggle** to reveal the last 4 digits (e.g. `•••-••-1234` for US SSN)
- Add a **"Change" link** below the field to switch back to the editable input for re-entry

The last 4 digits are extracted by decrypting the existing Strongbox-encrypted tax ID at runtime in the presenter. No new database columns needed.

### Changes

- **`app/presenters/settings_presenter.rb`** - Add `individual_tax_id_last_four` and `business_tax_id_last_four` by decrypting existing encrypted values
- **`app/javascript/types/payments.ts`** - Add new fields to User type
- **`app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx`** - Masked display with eye icon toggle and Change link for both individual and business tax IDs
- **`spec/presenters/settings_presenter_spec.rb`** - Update expected presenter output

Fixes https://github.com/antiwork/gumroad/issues/4999


arunavo4:
Flagging a couple of possible issues from running this PR locally on the latest head (`7038207c`) with a US seller account that has a saved SSN. Probably easier to spot from outside the iteration loop, happy to be wrong on either of these.

### 1. After a successful save, the field stays in edit mode

The fix in this PR resolves the "did my SSN actually save?" moment on a fresh page load, which is the bulk of the win. The same moment can come back after a re-save though:

1. Open `/settings/payments` with a saved SSN. Masked field, eye toggle, Change button all render correctly.
2. Click **Change**. The editable input appears.
3. Type a new value.
4. Click **Update settings**. The save goes through (DB confirms the new value is what's stored).
5. The SSN field is still in editable mode, with the just-typed value sitting there in plain text.

comment-03-stuck-after-submit


Root cause is that Inertia partial-reloads props on save without remounting `AccountDetailsSection`, so `isEditingIndividualTaxId` stays `true`. Neither of the existing prop dependencies transition either: `*_entered` was already `true` before submit, so a useEffect that watches it does not fire. The only prop that actually changes on a re-save is `*_last_four`.

I tried patching it locally and the fix that works is two useEffects keyed on `_last_four` (and `_entered` for the initial save case):

```ts
React.useEffect(() => {
  if (user.individual_tax_id_entered) setIsEditingIndividualTaxId(false);
}, [user.individual_tax_id_entered, user.individual_tax_id_last_four]);
React.useEffect(() => {
  if (user.business_tax_id_entered) setIsEditingBusinessTaxId(false);
}, [user.business_tax_id_entered, user.business_tax_id_last_four]);
```

With that patch applied, the same repro lands back on the masked/saved view after submit, which felt much nicer.

comment-04-after-fix

### 2. The new spec might have a duplicated `it` block

In `spec/requests/settings/payments_spec.rb`, the `it "shows masked SSN with eye icon toggle when tax ID has been entered"` block appears at both line 478 and line 503. Same title, same body. A `diff` between the two ranges came back empty, so it is the same block byte-for-byte. RSpec runs it twice as written, which probably is not the intent. Looks like a code-gen artifact from the original commit. Dropping one block would clean it up.

nyomanjyotisa:
@claude review

PR #5004, opened 2026-05-11
Fix flaky refund policy screenshot race that silently breaks dispute evidence

nyomanjyotisa:
Closes https://github.com/antiwork/gumroad-private/issues/347

## What

Two changes to stop disputes from silently auto-losing when the refund policy screenshot fails:

1. **Root cause fix** — In [`GenerateRefundPolicyImageService#calculate_height`](app/services/dispute_evidence/generate_refund_policy_image_service.rb), wait for the React/Inertia-rendered `<article>` element to mount before measuring its height, and short-circuit defensively on null. Falls back to document height on timeout.
2. **Defense in depth** — In [`CreateFromDisputeService#attach_refund_policy_image`](app/services/dispute_evidence/create_from_dispute_service.rb), broaden the rescue clause to catch any error from the screenshot path, not just `ImageTooLargeError`. Logs to `ErrorNotifier` with the purchase ID and lets the rest of evidence creation continue.

## Why

In production, `GenerateRefundPolicyImageService` intermittently raises:

```
Selenium::WebDriver::Error::JavascriptError:
  javascript error: Cannot read properties of null (reading 'parentElement')
  (Session info: chrome-headless-shell=125.0.6422.76)
```

Reproduced against dispute 125734 (created 2026-05-11 03:10 UTC, still in `formalized` state with no evidence at the time of this PR).

### Where it comes from

[generate_refund_policy_image_service.rb:77](app/services/dispute_evidence/generate_refund_policy_image_service.rb#L77) (on `main`):

```ruby
content_height = driver.execute_script(%{ return document.querySelector("article").parentElement.scrollHeight; })
```

The wait on line 59-60 only waits for `document.readyState === "complete"`, which fires when the initial HTML is parsed, not after React/Inertia has hydrated. So we have a race: sometimes the article is mounted by the time we measure it (succeeds), sometimes it isn't (`null.parentElement` throws).

### Why it kills disputes

When `attach_refund_policy_image` raises anything other than `ImageTooLargeError`, the exception propagates past `dispute_evidence.save!` and out of `handle_event_dispute_formalized!`. By then `mark_formalized!` has already committed. The Sidekiq retry hits the state gate at [`Charge::Disputable:99`](app/models/concerns/charge/disputable.rb#L99) and returns early because the dispute is no longer in `initiated` or `created` state. The retry "succeeds" with no evidence created, no `FightDisputeJob` enqueued, and the dispute auto-loses. The exception only surfaces once per failure, which is why it has been leaking quietly.

The two fixes together: (1) eliminates the race so the screenshot succeeds in the first place, and (2) ensures even if it does fail, evidence is still saved without the optional policy image — Stripe accepts the package without it; the receipt image, access activity log, and uncategorized text carry the dispute.

---

## AI Disclosure

Drafted with Claude Opus 4.7 (1M context).

### DIFF
--- a/app/services/dispute_evidence/create_from_dispute_service.rb
+++ b/app/services/dispute_evidence/create_from_dispute_service.rb
@@ -117,6 +117,8 @@ class DisputeEvidence::CreateFromDisputeService
       )
     rescue DisputeEvidence::GenerateRefundPolicyImageService::ImageTooLargeError
       ErrorNotifier.notify("DisputeEvidence::CreateFromDisputeService (purchase #{purchase.id}): Refund policy image not attached because was too large")
+    rescue => e
+      ErrorNotifier.notify(e, context: { service: "DisputeEvidence::CreateFromDisputeService#attach_refund_policy_image", purchase_id: purchase.id })
     end
 
     def mobile_purchase?
--- a/app/services/dispute_evidence/generate_refund_policy_image_service.rb
+++ b/app/services/dispute_evidence/generate_refund_policy_image_service.rb
@@ -44,6 +44,7 @@ class DisputeEvidence::GenerateRefundPolicyImageService
 
     IMAGE_RESIZE_FACTOR = 2
     IMAGE_QUALITY = 80
+    ARTICLE_WAIT_TIMEOUT_SECONDS = 5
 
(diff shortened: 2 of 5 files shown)


nyomanjyotisa:
@claude review

PR #5020, opened 2026-05-12
Add admin API endpoint for user purchases

nyomanjyotisa:
Ref #4989

## What

- Adds `GET /api/internal/admin/users/purchases`, a paginated read of purchases for a given user.
- Matches buyer-side records two ways: by `purchaser_id` **and** by `email`, so anonymous email-only purchases are not missed.
- Filters: `status` (single value or comma-separated list, validated against the purchase state set), `start_at` / `end_at` (ISO 8601), `chargedback`, `has_early_fraud_warning`, `has_affiliate`, `stripe_fingerprint`, `ip_address`.
- Pagination via the existing `Api::Internal::Admin::CursorPaginated` concern, ordered `created_at DESC, id DESC` to match `affiliates`.
- Lookup supports `email` or `user_id` and includes soft-deleted accounts, mirroring `info` and `affiliates`.
- No `record_admin_write`; this is a read.

## Why

- This is the primitive the `Purchases` checkbox on #4989 calls for: walk a buyer's history (across sellers) and compute velocity over a window.
- Buyer-side via `purchaser_id` alone misses anonymous Gumroad purchases that only carry an `email`. The OR with `email` catches both shapes without needing a new column or index.
- Rows serialize through the existing `serialize_purchase` helper on `Api::Internal::Admin::BaseController`. Keeping the row shape stable means the richer fields (card BIN, IP country, EFW details) land in PR #9 (purchase response extension) for free, without churn here.
- `status` validates against a fixed allowlist so callers can't smuggle arbitrary state values, and `start_at` / `end_at` reject anything `Time.iso8601` can't parse.

---
This PR was implemented with AI assistance using Claude Opus 4.7 (1M context)

### DIFF
--- a/app/controllers/api/internal/admin/users_controller.rb
+++ b/app/controllers/api/internal/admin/users_controller.rb
@@ -4,7 +4,18 @@ class Api::Internal::Admin::UsersController < Api::Internal::Admin::BaseControll
   include Api::Internal::Admin::CursorPaginated
 
   VALID_AFFILIATE_DIRECTIONS = %w[granted received].freeze
-  private_constant :VALID_AFFILIATE_DIRECTIONS
+  VALID_PURCHASE_STATES = %w[
+    in_progress
+    successful
+    failed
+    preorder_authorization_successful
+    preorder_authorization_failed
+    preorder_concluded_successfully
+    not_charged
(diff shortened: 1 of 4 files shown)


nyomanjyotisa:
@claude review

Issue #10, opened 2025-04-03
Open-source


ershad:
### Build and deployment

- [x] Set up continuous deployment (CD) using Buildkite
  - This is to keep the production credentials and deployment info away from the public GitHub repo
- [ ] Trigger production deployment only when tests are passed for the commit on `main` branch 

### Repository Setup

- [ ] Move existing repo to `gumroad-old`
- [ ] Push latest code to `gumroad-web` without the Rails credentials files
- [ ] Make `gumroad-web` public and rename it to `gumroad`
- [ ] Reconfigure Buildkite to use the changes in new repo for deployment
- [ ] Reset Devin's workspace to prevent it pushing the old history to the new repo ([ref](https://anti--work.slack.com/archives/C055QQCQFM1/p1740871868905529))

### After open-sourcing

- [ ] Continue the work on improving application security ([ref](https://anti--work.slack.com/archives/C085LLBP623/p1742291061956209))

gsharma101:
can you please assign this to me will he happy to work on this.

PR #5022, opened 2026-05-12
Extend admin purchase response with fraud-investigation fields

nyomanjyotisa:
Ref #4989

## What

Extends `Api::Internal::Admin::BaseController#serialize_purchase` so the admin purchase response carries the data fraud reviewers actually need:

- `chargeback_date`, `charge_processor`, `paypal_order_id`
- `ip_address`, `ip_country`, `billing_country`, `card_country`
- `country_mismatches: { billing_vs_ip, billing_vs_card, ip_vs_card }` - case-insensitive comparison, blanks treated as non-mismatches so missing data does not generate false positives
- `card: { bin, type, visual, expiry_month, expiry_year }`
- `dispute` - the latest dispute by `created_at` with state, reason, Stripe ID, and initiated/formalized/won/lost timestamps; `null` when none
- `early_fraud_warning` - Stripe processor ID, fraud type, charge risk level, actionable flag, resolution, resolution message, timestamps; `null` when none
- `affiliate_credit: { amount_cents, fee_cents, basis_points, affiliate_user_id }`; `null` when none
- `clusters: { fingerprint_count, browser_count, ip_count }` on `GET /purchases/:id` only - opt-in via `with_clusters: true`, blank source columns produce `nil` counts

`GET /purchases/:id` now preloads the new associations via `ActiveRecord::Associations::Preloader` before serializing. `GET /purchases/search` switches its `.includes(:link, :seller, :refunds)` to `.includes(*ADMIN_PURCHASE_INCLUDES)` so the search response carries the new fields with no extra queries per row.

## Why

- The admin purchase response was previously price/state/refund-shaped. Fraud triage needs the country trio, card details, dispute and EFW context, and affiliate credit - composing several endpoints to get them defeats the point of having a single purchase read.
- Country mismatch booleans on the server save every caller from reimplementing the comparison and getting the case-insensitivity wrong. Treating blanks as non-mismatches keeps the response honest for purchases where one of the three columns hasn't been populated (free purchases, PayPal flows, etc.).
- Cluster counts are listed in the issue as part of the response extension but they are 3 queries per purchase, which would N+1 the search and list contexts that share the serializer. Opt-in via `with_clusters: true` keeps the show endpoint complete while leaving search/list cheap. PR #10 (lookup by fingerprint/browser/IP) is the right place for cluster-style queries at scale.
- `ADMIN_PURCHASE_INCLUDES` centralizes the preload set so future callers (PR #5020's user purchases list, follow-up endpoints) all stay in sync as the field set grows.

---
This PR was implemented with AI assistance using Claude Opus 4.7 (1M context)

### DIFF
--- a/app/controllers/api/internal/admin/base_controller.rb
+++ b/app/controllers/api/internal/admin/base_controller.rb
@@ -12,7 +12,7 @@ class Api::Internal::Admin::BaseController < Api::Internal::BaseController
     purchases.reassign
     purchases.resend_all_receipts
   ].freeze
-  ADMIN_PURCHASE_INCLUDES = [:link, :seller, :refunds, :affiliate_credit, :early_fraud_warning, :disputes].freeze
+  ADMIN_PURCHASE_INCLUDES = [:link, :seller, :refunds, { affiliate_credit: :affiliate_user }, :early_fraud_warning, :disputes].freeze
   USER_LOOKUP_BAD_REQUEST_MESSAGE = "email or user_id is required"
   USER_ID_REQUIRED_MESSAGE = "user_id is required for mutating admin actions. " \
     "Use /internal/admin/users/info to look up the user_id by email."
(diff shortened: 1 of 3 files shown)


nyomanjyotisa:
@claude review

PR #5023, opened 2026-05-12
Extend admin product response with fraud-investigation fields

nyomanjyotisa:
Ref #4989

## What

Extends `Api::Internal::Admin::ProductsController#serialize_product` with the fields fraud reviewers need to read off a single product response:

- `banned_at`, `purchase_disabled_at` - ISO 8601 datetimes, `nil` when not set
- `bad_card_counter` - integer from the `links.bad_card_counter` column
- `taxonomy: { id, slug, ancestry_path }` - `ancestry_path` is the slug array root → leaf via `taxonomy.ancestry_path` (provided by `has_closure_tree`); `nil` when unassigned
- `affiliates` - list of attached direct affiliates and collaborators, filtered the same way as the user-affiliates endpoint (`[DirectAffiliate.name, Collaborator.name]`). Each row carries `id`, `type`, `affiliate_user: { id, email }`, effective `basis_points` (per-product override falls back to the parent affiliate's), and `destination_url`
- `recent_chargeback_rate: { window_days: 90, successful_count, chargedback_count, rate }` - **show only**, opt-in via `with_fraud_context: true` to keep the index endpoint cheap. `rate` is `nil` when there were no recent successful sales

Index now `.includes(:taxonomy, product_affiliates: { affiliate: :affiliate_user })` so the new fields do not N+1 across paginated rows.

## Why

- The admin product response was previously product-card-shaped. Fraud triage needs the lifecycle dates (banned, purchase-disabled), the bad-card signal, the taxonomy for category-based heuristics, and the attached-affiliate graph - composing several endpoints to get them defeats the point of having a single product read.
- `recent_chargeback_rate` is the only expensive new field (2 COUNT queries per product). The same pattern from #5022 applies: gate cluster-style cost behind a kwarg so show stays informative and index stays cheap. PR #10 (lookup endpoint) is the right place for cluster-style queries at scale.
- The non-global affiliate filter matches the user-affiliates endpoint (PR #5008) so callers see consistent affiliate shape across the user-side and product-side reads. Global affiliates aren't useful in this view since they're an implicit relationship for every product.
- The 90 day chargeback window matches industry convention for issuer-issued EFW/dispute lag. The response carries `window_days` so the contract is self-documenting.

---
This PR was implemented with AI assistance using Claude Opus 4.7 (1M context).

### DIFF
--- a/app/controllers/api/internal/admin/products_controller.rb
+++ b/app/controllers/api/internal/admin/products_controller.rb
@@ -8,8 +8,15 @@ class Api::Internal::Admin::ProductsController < Api::Internal::Admin::BaseContr
   SELLER_LOOKUP_BAD_REQUEST_MESSAGE = "email or external_id is required"
   RECENT_CHARGEBACK_WINDOW_DAYS = 90
   NON_GLOBAL_AFFILIATE_TYPES = [DirectAffiliate.name, Collaborator.name].freeze
+  TAXONOMY_ANCESTRY_SQL = <<~SQL.squish.freeze
+    SELECT h.descendant_id, t.slug
+    FROM taxonomy_hierarchies h
+    INNER JOIN taxonomies t ON t.id = h.ancestor_id
+    WHERE h.descendant_id IN (?)
+    ORDER BY h.descendant_id, h.generations DESC
+  SQL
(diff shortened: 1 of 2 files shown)


nyomanjyotisa:
@claude review

PR #5026, opened 2026-05-12
Add admin purchase lookup by fingerprint, browser ID, or IP

gianfrancopiana:
Ref #4989

## What

Adds a new internal admin lookup endpoint for purchases that match a single buyer signal: `stripe_fingerprint`, `browser_guid`, or `ip_address`.

It also expands the shared purchase response with a seller object that includes the seller's external ID, email, and name, while keeping the existing `seller_email` field in place.

## Why

This gives investigators a direct way to follow a buyer cluster across sellers without mixing signals or changing the existing purchase response shape for current callers.

The seller block makes it easier to jump from a purchase cluster to seller-scoped admin tools without breaking existing integrations.

---


gianfrancopiana:
@claude review

PR #5041, opened 2026-05-13
Stop forcing Stripe Connect representative to 100% ownership

nyomanjyotisa:
## What

`StripeMerchantAccountManager#update_person` had two latent bugs that surfaced for sellers whose Stripe Connect compliance flow recorded additional beneficial owners (typical of UK/EU Ltd companies with multiple directors):

1. It selected the person to update via `Stripe::Account.list_persons(...)["data"].last` — an arbitrary choice that did not necessarily match the user tied to the Gumroad account.
2. It hard-coded `relationship.percent_ownership: 100` on every update. When Stripe had already split ownership across multiple persons (e.g. 33.33% × 3 directors), pushing the representative to 100% drove the total above 100% and Stripe rejected the update with:

   > The total combined ownership of the company would exceed 100 percent.

The rejection was caught in `Settings::PaymentsController#update_user_compliance_info` and persisted as a `note` Comment on the user. So every time an affected seller resubmitted their Settings → Payments form, a fresh "exceed 100%" admin note was created — even when their Stripe account was already fully verified (`charges_enabled`, `payouts_enabled` both `true`, no `requirements.currently_due`).

The fix in [stripe_merchant_account_manager.rb#L212-L255](https://github.com/antiwork/gumroad/blob/fix-stripe-update-person-multi-owner/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb#L212-L255):

- Picks the person flagged `relationship.representative = true`, falling back to the previous `list_persons.last` behavior when no representative is present (preserves behavior for older accounts and existing test stubs).
- Only sets `percent_ownership: 100` when no other person on the Stripe account is flagged as an owner. When other beneficial owners exist, the representative's existing Stripe-side percentage is left untouched.
- Returns early if Stripe returns zero persons, instead of crashing on `nil.id`.

Two new private helpers — `find_stripe_representative_person` and `other_beneficial_owners_present?` — keep the logic out of the main flow and individually testable.

`create_account` is intentionally not changed: at account creation there is always exactly one person, so the existing `percent_ownership: 100` is correct there.

## Why

A real production seller (UK Ltd with three director-owners at 33.33% each, Stripe account fully verified) had 20+ identical "exceed 100%" admin notes accrue over the past three weeks because they kept opening Settings → Payments to "fix" something Stripe support had told them to fix in our app. The Stripe account itself was healthy — the error was entirely generated by us re-submitting `percent_ownership: 100` on every Gumroad-side update — but the admin notes spooked first-line support into routing the ticket as a Stripe-side problem, and Stripe Support correctly bounced it back to Gumroad.

The fix is minimal and conservative: behavior is unchanged for single-owner accounts (which is what Gumroad's Payments UI models today), and the multi-owner case stops producing false-positive errors. Sellers with multi-director companies no longer need to coordinate with Stripe Support — they can update their compliance details on Gumroad without triggering the phantom error.

---

This PR was implemented with Claude Opus 4.7 (1M context)

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb
@@ -210,12 +210,21 @@ module StripeMerchantAccountManager
   end
 
   def self.update_person(user, stripe_account, last_user_compliance_info_id, passphrase)
-    stripe_person = Stripe::Account.list_persons(stripe_account.id)["data"].last
+    stripe_persons = Stripe::Account.list_persons(stripe_account.id)["data"]
+    stripe_person = find_stripe_representative_person(stripe_persons)
+    return if stripe_person.nil?
+
     last_user_compliance_info = UserComplianceInfo.find_by_external_id(last_user_compliance_info_id)
     user_compliance_info = user.alive_user_compliance_info
 
     current_attributes = person_hash(user_compliance_info, passphrase)
-    current_attributes.deep_merge!(relationship: { representative: true, owner: true, title: user_compliance_info.job_title.presence || DEFAULT_RELATIONSHIP_TITLE, percent_ownership: 100 })
+    relationship_attributes = {
+      representative: true,
+      owner: true,
+      title: user_compliance_info.job_title.presence || DEFAULT_RELATIONSHIP_TITLE
+    }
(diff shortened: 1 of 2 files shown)


nyomanjyotisa:
@claude review

slavingia:
It would be good to support collecting multiple beneficial owners in the UI as well.

nyomanjyotisa:
@claude review

PR #5049, opened 2026-05-13
Remove duplicate UpsellCard tiptap extension registration

gumclaw:
## What

Three callers were explicitly passing `UpsellCard` in their `extensions` array, but `useRichTextEditor` already includes it as a default extension. This caused a duplicate registration warning on every page that uses those editors:

```
[tiptap warn]: Duplicate extension names found: ['upsellCard']
```

## Fix

Remove `UpsellCard` from the 3 callers that double-registered it:
- `Product/index.tsx` (public product page viewer)
- `ContentTab/index.tsx` (product content editor)
- `EmailForm.tsx` (email form editor)

The hook keeps providing `UpsellCard` to all consumers — editors, viewers, and toolbar — so no functionality is lost.

## Testing

- TypeScript compiles with no new errors
- `baseEditorOptions` already deduplicates by extension name, so the remaining callers that don't explicitly pass `UpsellCard` continue to get it from the hook

Fixes antiwork/gumroad-private#368

> [!NOTE]
> This PR was authored with Gumclaw.

### DIFF
--- a/app/javascript/components/RichTextEditor.tsx
+++ b/app/javascript/components/RichTextEditor.tsx
@@ -277,8 +277,13 @@ export const useRichTextEditor = ({
     uploadImages({ view, files: images, imageSettings });
   };
 
+  const allExtensions = [...extensions, ...(placeholder ? [Placeholder.configure({ placeholder })] : []), UpsellCard];
+  const dedupedExtensions = allExtensions.filter(
+    (ext, index) => allExtensions.findIndex((e) => e.name === ext.name) === index,
+  );
+
   const editor = useEditor({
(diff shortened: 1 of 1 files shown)


gumclaw:
@claude review

gumclaw:
Addressed Claude's review — reversed the approach. Now keeping `UpsellCard` in `useRichTextEditor` (the hook) and removing the duplicate registrations from the 3 callers that were double-registering it (`Product/index.tsx`, `ContentTab/index.tsx`, `EmailForm.tsx`). This preserves UpsellCard for all consumers while eliminating the warning.

@claude review

PR #5058, opened 2026-05-14
Fix flaky StrippedFields spec

gumclaw:
The test used a temporary table created at file load time. Temporary tables are connection-scoped in MySQL — if the connection pool recycles between load time and test execution (common under heavy parallel CI), the table disappears and `TestField.new` hits a statement timeout.

Fix: use a regular table created in `before(:all)` with `after(:all)` cleanup, and explicitly set `table_name` on the model.

Failure: https://github.com/antiwork/gumroad/actions/runs/25835354881/job/75910411473#step:7:1

### DIFF
--- a/spec/models/concerns/stripped_fields_spec.rb
+++ b/spec/models/concerns/stripped_fields_spec.rb
@@ -2,18 +2,31 @@
 
 require "spec_helper"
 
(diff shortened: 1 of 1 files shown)


slavingia:
@claude review

PR #5059, opened 2026-05-14
Fix admin purchase search crash when product is deleted

gumclaw:
## What

Admin purchase search (`/admin/purchases`) crashes with 500 when a purchase has a deleted/missing product.

## Why

`amount_refundable_cents_in_currency` calls `purchase.link.price_currency_type` but `purchase.link` can be nil for soft-deleted products or null `link_id`. One nil link blows up the entire search.

## How

Use `displayed_price_currency_type` instead — it's stored directly on the purchase record at creation time and is always present. Fixed in three places:

- `Admin::BaseController#amount_refundable_cents_in_currency`
- `Purchase#amount_refundable_cents_in_currency`
- `Purchase#amount_refundable_in_currency`

Fixes https://github.com/antiwork/gumroad-private/issues/374


nyomanjyotisa:
@claude review

nyomanjyotisa (approved):

PR #5060, opened 2026-05-14
Preserve alphanumeric business_tax_id for non-US sellers

nyomanjyotisa:
## What

[app/services/update_user_compliance_info.rb][1] unconditionally stripped non-digit characters from `business_tax_id` before save:

```ruby
new_compliance_info.business_tax_id = compliance_params[:business_tax_id].gsub(/\D/, "")
```

That normalization exists for US EINs — the validation immediately after (`new_compliance_info.business_tax_id.length != 9`) only fires for `legal_entity_country_code == "US"`. It was inadvertently applied to every country.

This PR scopes the `gsub(/\D/, "")` to US legal entities. For all other countries we now `.strip` surrounding whitespace and preserve alphanumeric characters. The same logic is applied in the change-detection path (`encrypted_compliance_info_changed?`) so resubmitting a value with the previously-stripped letters restored is correctly detected as a change.

[1]: https://github.com/antiwork/gumroad/blob/main/app/services/update_user_compliance_info.rb#L144

## Why

Outside the US, business tax IDs are commonly alphanumeric:

- **Ireland** — Tax Reference Numbers end in one or two check letters (e.g. `1234567T`); VAT IDs carry the `IE` prefix.
- **EU VAT** — country prefix + digits + optional check letters (e.g. `FR12345678901`, `DE123456789`).
- **UK** — UTR is ten digits, often submitted with spaces or dashes; VAT IDs are prefixed `GB`.

Stripping non-digits silently corrupted these values before they reached Stripe. On Stripe Connect, the `company.tax_id` ended up mismatching the seller's Certificate of Incorporation or VAT registration document, so verification kept failing.

The bug was also unrecoverable from the seller's side: every time they re-entered the full value in Settings → Payments, our form stripped it back to digits on save and pushed the stripped value to Stripe again. From the seller's perspective this looked like Gumroad "overwriting" their tax ID with a different one.

This came up via support — an EU seller's stored `business_tax_id` was missing its trailing check letters because every save dropped them, and the mismatch was blocking Stripe verification.

---

Model: Claude Opus 4.7 (1M context)

### DIFF
--- a/app/controllers/api/internal/admin/users_controller.rb
+++ b/app/controllers/api/internal/admin/users_controller.rb
@@ -410,7 +410,7 @@ class Api::Internal::Admin::UsersController < Api::Internal::Admin::BaseControll
         ) : nil,
         tax_ids: {
           individual_last_four: tax_id_last_four(info.individual_tax_id),
-          business_last_four: tax_id_last_four(info.business_tax_id, digits_only: true),
+          business_last_four: tax_id_last_four(info.business_tax_id),
         },
         identity_documents: {
           stripe_identity_document_id: info.stripe_identity_document_id,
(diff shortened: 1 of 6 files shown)


nyomanjyotisa:
@claude review

PR #5062, opened 2026-05-14
Add ID upload criteria to help center and verification modal

slavingia:
## Summary

Adds ID upload criteria/requirements to two places:

1. **Help Center** — new "Criteria for identification uploads" section under "ID verification" in the Getting Paid article
2. **Verification modal** — collapsible `<details>` section showing upload requirements when a compliance action has an `href` (i.e. ID upload link)

The criteria listed:
- Color photo required (phone, webcam, or scanner)
- Driver's license / identity card → both front and back needed
- JPEG and PNG accepted
- PDFs not accepted

---

### Help Center article (`_13-getting-paid.html.erb`)

**Before** (production):
before

**After** (localhost on branch):
after

### Verification modal (`AccountStatusSection.tsx`)

No screenshot — requires a user with active Stripe compliance actions to trigger the alert. The change adds a collapsible `<details>` block (shown only when a compliance action has an `href`) with the same four criteria listed above.

### DIFF
--- a/app/javascript/components/Settings/PaymentsPage/AccountStatusSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/AccountStatusSection.tsx
@@ -94,6 +94,20 @@ export default function AccountStatusSection({
             {accountStatus.compliance_actions.map((action, i) => (
               <ComplianceActionItem key={i} action={action} />
             ))}
+            {accountStatus.compliance_actions.some((a) => a.href) ? (
+              <details className="mt-2">
+                <summary className="cursor-pointer text-sm font-medium">ID upload requirements</summary>
+                <ul className="mt-1 list-disc pl-5 text-sm">
+                  <li>We need a color photo of the ID, which you can take with a phone, webcam, or scanner</li>
+                  <li>
+                    If you are uploading a driver's license or identity card, we will need an image of both the front and
+                    back of the ID
+                  </li>
+                  <li>We accept JPEG and PNG file formats</li>
+                  <li>We cannot verify PDFs</li>
+                </ul>
+              </details>
+            ) : null}
             {accountStatus.compliance_actions.some((a) => !a.href) ? (
               <p>
                 Please update your information below.
--- a/app/javascript/components/Settings/PaymentsPage/AccountStatusSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/AccountStatusSection.tsx
@@ -97,15 +97,12 @@ export default function AccountStatusSection({
             ))}
             {accountStatus.needs_id_upload ? (
               <details className="mt-2">
-                <summary className="cursor-pointer text-sm font-medium">ID upload requirements</summary>
+                <summary className="cursor-pointer text-sm font-medium">Before you upload your ID</summary>
                 <ul className="mt-1 list-disc pl-5 text-sm">
-                  <li>We need a color photo of the ID, which you can take with a phone, webcam, or scanner</li>
-                  <li>
-                    If you are uploading a driver's license or identity card, we will need an image of both the front and
-                    back of the ID
-                  </li>
-                  <li>We accept JPEG and PNG file formats</li>
-                  <li>We cannot verify PDFs</li>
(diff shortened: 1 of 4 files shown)


gianfrancopiana:
@gumclaw I think this shows the ID upload checklist too often. The UI shows it whenever the action links to Stripe, but that Stripe link is not always for uploading an ID. Sometimes Stripe asks the seller for other information.

In those cases, this would tell the seller about JPEG/PNG IDs and front/back images even though Stripe is asking for something else.

Can we only show this checklist when the backend confirms the seller actually needs to upload an ID document?

gumclaw:
Good catch @gianfrancopiana — you're right. The `href` check is too broad. It's set whenever the seller has a Stripe account with pending compliance (line 297 in `settings_presenter.rb`), which covers address verification, business info, tax IDs, etc. — not just ID document uploads.

The backend already tracks the specific `field_needed` on each `UserComplianceInfoRequest` (e.g. `stripe_identity_document_id`, `stripe_additional_document_id`, `passport`). The fix would be to:

1. **Expose the requested fields to the frontend** — add the `field_needed` values (or a derived `needs_id_upload` boolean) to the compliance action payload in `settings_presenter.rb`
2. **Gate the checklist on document fields** — only show it when the pending requests include document-type fields (`STRIPE_IDENTITY_DOCUMENT_ID`, `STRIPE_ADDITIONAL_DOCUMENT_ID`, `PASSPORT`, `VISA`, `STRIPE_COMPANY_DOCUMENT_ID`, `STRIPE_ENHANCED_IDENTITY_VERIFICATION`)

I'll pick this up.

gumclaw:
Good catch — pushed eb7e917a to gate the checklist on a new `needs_id_upload` boolean.

Backend (`settings_presenter.rb`) now exposes `needs_id_upload` alongside `compliance_actions`, computed from whether any pending `UserComplianceInfoRequest` has a `field_needed` matching document-type fields:
- `stripe_identity_document_id`
- `stripe_additional_document_id`
- `passport`
- `visa`
- `stripe_company_document_id`
- `stripe_enhanced_identity_verification`

Frontend (`AccountStatusSection.tsx`) now reads `accountStatus.needs_id_upload` instead of `compliance_actions.some((a) => a.href)`. So sellers who only owe address/business-info/tax-ID won't see the ID upload instructions anymore.

Added presenter specs covering both branches plus a per-field check for each document-type field. Couldn't run rspec locally (Ruby 3.4.3 not installed in this env) — CI will verify.

---

🤖 Generated with Claude Opus 4.7

gianfrancopiana:
@gumclaw `needs_id_upload` is still too broad. It includes `stripe_company_document_id`, which our code labels as a company registration document or UAE trade license, not a personal ID. In that case, the UI would still show “ID upload requirements” with driver’s license / identity card instructions.

Can we either limit this checklist to fields that are definitely personal ID uploads, or change the copy so it fits the broader set of Stripe document requests? If we broaden it, we should check Stripe’s docs first because legal entity/address documents appear to have different accepted formats than identity documents.

gumclaw:
You're right — `stripe_company_document_id` maps to "Company registration document" / "UAE Trade License" in our field properties, so showing personal ID instructions for it is misleading.

The fix is to remove `STRIPE_COMPANY_DOCUMENT_ID` from the `id_document_fields` array in `settings_presenter.rb`. The remaining fields (`STRIPE_IDENTITY_DOCUMENT_ID`, `STRIPE_ADDITIONAL_DOCUMENT_ID`, `PASSPORT`, `VISA`, `STRIPE_ENHANCED_IDENTITY_VERIFICATION`) are all individual/personal identity verifications where the current copy is accurate.

Picking this up now.

gumclaw:
Pushed 18d16dde — removed `STRIPE_COMPANY_DOCUMENT_ID` from the gate. Only personal identity fields now: `STRIPE_IDENTITY_DOCUMENT_ID`, `STRIPE_ADDITIONAL_DOCUMENT_ID`, `PASSPORT`, `VISA`, `STRIPE_ENHANCED_IDENTITY_VERIFICATION`. Updated specs accordingly.

---
🤖 Generated with [Claude](https://claude.ai)

PR #378, opened 2025-06-19
docker-compose: add “:Z” SELinux label to data bind-mounts

skatkov:
I'm running Fedora and have issue running mongo/mysql/redis through docker-compose because they have only read-only permissions to folders they create.

### What changed  
• In `docker/docker-compose-local-linux.yml` each bind mount that maps a
  host directory into a container’s data directory now ends with `:Z`  
  ```
  - ./tmp/mysql:/var/lib/mysql        →  - ./tmp/mysql:/var/lib/mysql:Z
  - ./tmp/redis:/data                 →  - ./tmp/redis:/data:Z
  - ./tmp/mongo:/data/db              →  - ./tmp/mongo:/data/db:Z
  ```
  (no change needed for the named volume `elasticsearch7data`).

### Why  
On Fedora / RHEL and other SELinux-enabled distros the containers start
as an unprivileged user.  
When a plain bind mount is used, SELinux labels on the host files
(`user_home_t`) block write access and MySQL/Redis/Mongo fail with

```
find: '/var/lib/mysql/': Permission denied
chown: changing ownership of … : Permission denied
```

The `:Z` suffix tells Docker to relabel the mounted directory with the
container-friendly `svirt_sandbox_file_t` type, allowing the databases to
start normally.

### Compatibility  
• The flag is ignored on systems without SELinux (Ubuntu, Debian,
  macOS/Windows Docker Desktop, WSL2), so behaviour there is unchanged.  
• No image or runtime versions are affected; Compose v1.25+ / Compose-V2
  already understand the option.

### Result  
Developers on Fedora/RHEL (or anyone with SELinux in enforcing mode) can
run `make local` / `docker compose up` without extra manual steps, while
other platforms continue to work as before.

If you already experienced this issue, then you need to run `make stop_local && make local` to reinitialize these folders properly.

## Summary by CodeRabbit

- **Chores**
  - Updated Docker Compose configuration to improve compatibility with SELinux on Linux systems for MySQL, Redis, MongoDB, and Nginx services.

PR #5064, opened 2026-05-14
CI: pre-seeded MySQL image to skip db:prepare on 100 nodes

gumclaw:
## What

Builds a custom MySQL Docker image with the test database schema pre-loaded, so CI nodes skip the ~30-60s `db:prepare` step entirely. Across 100 nodes, this saves **~50-100 minutes of aggregate CI compute per run**.

## How

**Build phase** (runs once per schema change):
1. Computes a content-addressed tag from `schema.rb` + Dockerfile hash
2. If the tag already exists on Docker Hub → skip (cache hit)
3. Otherwise: starts a temp MySQL, runs `db:schema:load` via the test image, dumps the schema as SQL
4. Builds a Docker image with the schema baked in, pushes to `gumroadteam/mysql_test`

**Test phase** (100 nodes):
1. `docker-compose` starts the pre-seeded MySQL image (via `MYSQL_TEST_IMAGE` env var)
2. The image's custom entrypoint copies pre-seeded data from a staging dir to the tmpfs-mounted `/var/lib/mysql` in ~2s
3. Nodes only wait for MySQL connection — no `db:prepare` needed

**Backwards-compatible**: if the MySQL image tag is empty (build step skipped/failed), nodes fall back to vanilla `mysql:8.0.32` + `db:prepare`.

## Files

- `docker/ci/Dockerfile.mysql-preseeded` — two-stage Dockerfile: loads SQL dump, copies data dir to staging, custom entrypoint copies to tmpfs at runtime
- `docker/ci/generate_tag_for_mysql_preseeded.sh` — content-addressed tag generator
- `docker/ci/generate_schema_dump.sh` — standalone script to generate the SQL dump (also inlined in the workflow)
- `.github/workflows/tests.yml` — new build step + updated test_fast/test_slow to use pre-seeded image
- `docker/docker-compose-test-and-ci.yml` — accepts `MYSQL_TEST_IMAGE` env var (defaults to `mysql:8.0.32`)

## Expected savings

| Step | Before | After |
|------|--------|-------|
| MySQL startup | ~5s | ~2s |
| `db:prepare` | ~30-60s | **skipped** |
| **Per-node savings** | — | **~30-60s** |
| **Aggregate (100 nodes)** | — | **~50-100 min** |

Relates to #378


gumclaw:
Fixed the Docker build failure. Root cause: `sleep 5 && mysqladmin shutdown` raced with the schema SQL import — the temporary MySQL server accepted pings via socket while still importing the schema dump (ERROR 2013 at line 1542).

Fix: use `--protocol=tcp` to ping `127.0.0.1:3306`, which only succeeds once the final mysqld is up (after init SQL has fully completed). Race-free regardless of schema size or CI runner speed.

Awaiting CI on commit 11274020.

ershad:
@slavingia just pinged you via Teams about this. It's not worth the additional complexity; the per-node wait time isn't much.

gumclaw:
Makes sense — closing this. The per-node wait is tolerable and the Docker image build/cache layer adds maintenance overhead that isn't worth the savings.

PR #5069, opened 2026-05-15
Cache asset/webpack output in GHA Build images job

slavingia:
The Build images job rebuilds the test image on most PRs (any
non-doc/.github source change invalidates the content-addressed tag).
Inside the rebuild, `make build_test` runs `npm run setup &&
bundle exec rake db:setup assets:precompile` outside the Docker build,
in a separate `docker run` then `docker commit`. With no cache, webpack
runs cold every time (~6-8 min), making the job 6-14 min on typical PRs
vs 32s on a manifest hit.

The Makefile already supports a tarball-based BRANCH_CACHE that the
Buildkite pipeline restores from S3 (ci_scripts/helper.sh). GitHub
Actions hardcoded BRANCH_CACHE_*=false. Wire it to actions/cache@v4
keyed on the files that actually affect compiled assets.

Also hoist the manifest-inspect short-circuit into its own step so the
cache restore only runs when we're actually going to rebuild, and drop
a duplicate `docker pull ruby:...` (already pulled in the base-image
step).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,12 +94,34 @@ jobs:
           TAG=test-$(BASE_TEST_SHA=$WEB_BASE_TEST_SHA docker/base/generate_tag_for_test_image.sh)
           echo "tag=$TAG" >> $GITHUB_OUTPUT
           echo "Computed test image tag: $TAG"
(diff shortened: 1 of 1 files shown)


slavingia:
@claude review

PR #5092, opened 2026-05-15
Add CI progress logging for Vite build

slavingia:
## Summary

Vite's default \`transforming...\` spinner overwrites a single line, which stays static in CI (non-TTY stdout). On the cuprite-driver runs you see the build start, then nothing until it finishes — no way to tell whether it's progressing or hung, and no way to attribute time to phases when comparing runs (see e.g. the still-running [build job here](https://github.com/antiwork/gumroad/actions/runs/25925283250/job/76204873103?pr=5082)).

## What this changes

1. **\`vite.config.ts\`** — adds a small build-only, CI-only plugin gated on \`process.env.CI\`. It emits:
   - a timestamped \`[vite] build start <ISO timestamp>\` line at \`buildStart\`
   - a heartbeat every 250 transformed modules: \`[vite] transformed N modules in Xs\`
   - a final \`[vite] build done after Xs, N modules transformed\` (or \`failed\` on error) at \`buildEnd\`
   
   Local dev is unchanged (\`!process.env.CI\` → returns null, Vite ignores null plugin entries).

2. **\`Makefile\`** — wraps \`bundle exec rake db:setup assets:precompile --trace\` in the \`build_test\` target with \`time\`, so CI logs include wallclock for the full asset compile step.

## Trade-offs

- Heartbeat granularity (250) is a guess based on typical Vite build module counts. Easy to tune.
- The plugin runs on every transform, so it adds one function call per module — negligible vs. the actual transform cost.
- Could alternatively flip \`DEBUG=vite:*\` in CI for deep debugging, but that's noisy. This is the simple-default-on version.

## Test plan

- [ ] Push to a branch and inspect the \`Build images\` job log on the resulting CI run — confirm start/heartbeat/done lines appear
- [ ] Verify local \`npm run build\` and \`npm run dev\` are unchanged (no \`[vite]\` lines)
- [ ] Confirm \`time\` output shows in CI log at the end of the asset compile step


slavingia:
@claude review

PR #5094, opened 2026-05-15
Fix for Useless assignment to local variable

slavingia:
The best fix is to declare `i` without an initial value and keep the existing loop logic unchanged.  
In `app/javascript/components/ComboBox.tsx`, update the `moveCursor` function so `i` is declared as `let i: number;` instead of `let i = from;`. This removes the dead store while preserving functionality.

No new imports, helper methods, or definitions are required.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/app/javascript/components/ComboBox.tsx
+++ b/app/javascript/components/ComboBox.tsx
@@ -55,7 +55,7 @@ export const ComboBox = <Option extends unknown>({
   const itemRefs: (HTMLElement | null)[] = [];
 
   const moveCursor = (from: number, by: number) => {
-    let i = from;
+    let i: number;
     do {
       i = (from + by) % options.length;
       if (!itemRefs[i]?.inert) {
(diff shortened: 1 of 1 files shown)


slavingia:
@claude review

slavingia:
Pushed [3ad96d6](https://github.com/antiwork/gumroad/commit/3ad96d6) addressing both bugs found in the review:

- `let i = from` restored so the initial value is load-bearing (resolves the CodeQL dead-store warning for the right reason — by *using* the value, not deleting it).
- Loop body now accumulates `i + by` instead of recomputing `from + by` every iteration, so the skip-inert behavior actually walks through options.
- Wrapped the modulo as `((i + by) % length + length) % length` to handle JS's signed-modulo semantics, fixing ArrowUp wrap from index 0 to the last option.

@claude review

slavingia:
@claude review

PR #5095, opened 2026-05-15
Add close button to toast alerts and flash_message test helper

gumclaw:
## What

Three changes to reduce test flakiness caused by toast alerts and CSS animations:

1. **Close button on toast alerts** — Added an X dismiss button to the `ToastAlert` component. Previously, the only way to dismiss was to wait 5 seconds for the auto-dismiss timer.

2. **`flash_message` test helper** — Reads the alert text and immediately dismisses it. Use instead of `have_alert` when the next interaction would be blocked by the overlay.

3. **Disable CSS animations/transitions in system tests** — Injects a stylesheet after every `visit` that zeroes out all `animation-duration`, `animation-delay`, `transition-duration`, and `transition-delay`. Eliminates timing-dependent flakes from fade-in/fade-out, toast slide animations, and other transition-based state changes.

## Why

Toast alerts are `position: fixed; z-index: 100` and auto-dismiss after 5 seconds. During that window they overlay content, blocking clicks on elements underneath. This causes flaky test failures when specs need to interact with content immediately after triggering an alert.

CSS transitions (like the toast's 0.3s ease-out with 0.5s delay) add nondeterminism — an element might be mid-animation when a test tries to assert or click, depending on timing. Disabling animations entirely in tests makes assertions deterministic.

## Before/After

**Before:** Toast alerts had no close button — had to wait 5s for auto-dismiss. CSS animations ran in tests, causing timing flakes.

**After:** Toast alerts have a close X button. `flash_message` helper reads + dismisses in one step. All animations disabled in test env.

Alert with close button

```ruby
# New helper usage
click_on "Save"
expect(flash_message).to eq "Changes saved!"
# alert is now dismissed, safe to click content beneath it
```

## Test Results

- `bundle exec rubocop spec/support/capybara_helpers.rb` — 0 offenses
- Receipt + refund policy specs — 4/4 passing
- `bin/test-confidence` — reached 95%+ (low risk, no model/controller changes)

## Changes
- `app/javascript/components/server-components/Alert.tsx` — close button + dismiss handler
- `spec/support/capybara_helpers.rb` — `flash_message` helper, `disable_animations` via `visit` hook

---

AI disclosure: Claude Opus 4.6. Prompted to add a flash message helper and disable animations in tests per the blog post pattern.

### DIFF
--- a/app/javascript/components/server-components/Alert.tsx
+++ b/app/javascript/components/server-components/Alert.tsx
@@ -15,7 +15,9 @@ export type AlertPayload = { message: string; status: "success" | "danger" | "in
 const ToastAlert = ({ initial }: { initial: AlertPayload | null }) => {
   const [alert, setAlert] = React.useState<AlertPayload | null>(initial);
   const [isVisible, setIsVisible] = React.useState(!!initial);
+  const [isClosing, setIsClosing] = React.useState(false);
   const timerRef = React.useRef<number | null>(null);
+  const isHoveringRef = React.useRef(false);
 
   const clearTimer = () => {
     if (timerRef.current !== null) {
@@ -26,6 +28,7 @@ const ToastAlert = ({ initial }: { initial: AlertPayload | null }) => {
 
   const dismiss = () => {
     clearTimer();
+    setIsClosing(true);
     setIsVisible(false);
   };
 
@@ -39,8 +42,9 @@ const ToastAlert = ({ initial }: { initial: AlertPayload | null }) => {
     if (typia.is<{ type: "alert"; payload: AlertPayload }>(event.data)) {
       const newAlert = event.data.payload;
       setAlert(newAlert);
+      setIsClosing(false);
       setIsVisible(true);
-      startTimer();
+      if (!isHoveringRef.current) startTimer();
     }
   });
   useRunOnce(() => {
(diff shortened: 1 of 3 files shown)


gumclaw:
### Alert with close button (warning variant)

The X button appears inline next to the message text, matching the existing alert component layout.

alert-close-button

_(Uploading screenshot...)_

### Accessibility tree confirms structure
```
- alert
  - StaticText "An account does not exist with that email."
  - button "Close" [ref=e5]
    - image
```

### Disable animations CSS injected in tests
```css
*, *::before, *::after {
  animation-duration: 0s !important;
  animation-delay: 0s !important;
  transition-duration: 0s !important;
  transition-delay: 0s !important;
}
```

gumclaw:
@claude review

gumclaw:
@gianfrancopiana Would love your feedback on the close button design/positioning — see the screenshot in the PR description above. The X is vertically centered with the message text and pushed to the right.

Three changes:
1. Close button (×) on toast alerts — `aria-label="Close"`, inline with text
2. `flash_message` Capybara helper — reads alert text + dismisses in one step
3. CSS animations/transitions disabled in system tests

What do you think?

gianfrancopiana:
Pushed two follow-up commits.

**Toast polish (4775cd0a0):** wider max-width so common errors fit on one line, X moved to top corner with a bigger hit area, faster slide-out on manual close, and auto-dismiss pauses while hovering.

One-liner:


Wrapping warning:


**Spec helper (2a92b9087):** addresses the two Claude findings.

PR #5096, opened 2026-05-15
Fix flaky CI specs and stabilize waiting behavior

gianfrancopiana:
Ref #5083

## What

- Stabilizes a set of flaky specs that were failing in CI for timing and state-isolation reasons.
- Replaces brittle Selenium waits and tighter polling with simpler, driver-agnostic waiting.
- Fixes a few test expectations that depended on the wrong timestamp or on behavior that was no longer stable under CI timing.

## Why

- The goal is to make the CI signal reflect real regressions instead of transient timing issues.
- These changes keep the test suite aligned with the app's actual behavior while reducing false failures across related specs.

## Testing

- Local focused spec runs passed for the changed cases where the environment allowed reliable execution.
- CI verification on the branch was green after the follow-up fixes.

---

### DIFF
--- a/spec/requests/embed_spec.rb
+++ b/spec/requests/embed_spec.rb
@@ -99,7 +99,7 @@ describe "Embed scenario", type: :system, js: true, mock_easypost: true, retry: 
     let!(:offer_code) { create(:offer_code, user: product.user, products: [product]) }
 
     it "applies the discount code" do
-      visit(create_embed_page(product, url: "#{product.long_url}/#{offer_code.code}", outbound: false))
+      visit(create_embed_page(product, url: product.long_url(code: offer_code.code), outbound: false))
 
       within_frame do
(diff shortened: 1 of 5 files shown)


gianfrancopiana:
Follow up to #5083. cc @slavingia

spec/requests/embed_spec.rb
```diff
@@ -15,7 +15,10 @@
 
     visit(create_embed_page(product, url: product.long_url, gumroad_params: "&email=sam@test.com", outbound: false))
 
-    within_frame { click_on "Add to cart" }
+    within_frame do
+      expect(page).to have_link("Add to cart", wait: 15)
```
slavingia: I don't think this will fix the flakiness as I had this in a branch and the spec was still failing.

slavingia (approved):
Minor nit if you want to revert one part.

PR #5118, opened 2026-05-18
Show rejection alert instead of 500 for rejected Stripe accounts

gianfrancopiana:
Ref antiwork/gumroad-private#388

## What

When a Stripe account is permanently rejected (e.g. Japan's SMCC bouncing accounts with `rejected.listed`), clicking "Complete pending verification requirements via Stripe" on Settings → Payments returned a 500. Sellers now see a clear red alert with the rejection reason and a one-click Stripe dashboard link they can hand to support.

## Why

Japan SMCC rejections come in roughly weekly. Today the loop is: seller sees "Something broke" → emails support → we investigate → we tell them it's permanent.

## Screenshot

image

---


gianfrancopiana:
@claude review

PR #5123, opened 2026-05-19
Inline hero_parallax.js as shared partial (Sprockets cleanup)

gumclaw:
## What

Moves `hero_parallax.js` from a Sprockets-served JS file to an inline `<script>` via a shared ERB partial.

## Why

Step toward dropping Sprockets entirely. This was one of only 2 JS files in `app/assets/javascripts/` served through the Sprockets pipeline. By inlining it, we remove one `javascript_include_tag` dependency and one fewer file going through `assets:precompile`.

## Changes

- Deleted `app/assets/javascripts/hero_parallax.js`
- Created `app/views/home/shared/_hero_parallax.html.erb` — wraps the same JS in a `javascript_tag` with CSP nonce
- Updated `saas.html.erb` and `about.html.erb` to `render` the partial instead of `javascript_include_tag`

## Test results

No behavior change — same JS executes on the same pages, just delivered inline instead of as a separate asset request. CSP nonce still applied.

---
*This PR was created by Gumclaw 🤖*

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -318,5 +318,5 @@
 
   <%= render "home/shared/footer" %>
 
-  <%= javascript_include_tag 'hero_parallax', nonce: SecureHeaders.content_security_policy_script_nonce(request), defer: true %>
+  <%= render "home/shared/hero_parallax" %>
 </div>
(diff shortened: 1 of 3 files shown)


gumclaw:
@claude review once

nyomanjyotisa (approved):
## Before

https://github.com/user-attachments/assets/c1d516aa-12d0-4129-a27e-8a195d7f5ef7

## After

https://github.com/user-attachments/assets/b9c8ca12-d801-42e3-b219-3ca448b72379

PR #5127, opened 2026-05-19
Allow mobile sales export auth

gianfrancopiana:
Ref #5115

Sibling mobile PR: [antiwork/gumroad-mobile#248](https://github.com/antiwork/gumroad-mobile/pull/248)

## What

This lets the existing sales export route accept a valid mobile creator OAuth token and mobile token before the normal web login check.

The export action still uses the existing seller authorization and export behavior after authentication succeeds.

## Why

The mobile app opens the export in the browser, which does not share the app OAuth session. The sibling mobile PR can pass app auth to `/purchases/export`, but the web route also needs to consume that handoff so sellers reach the CSV download instead of the login page.

This keeps the existing export URL and mirrors the mobile web handoff pattern used elsewhere.

---

### DIFF
--- a/app/controllers/purchases_controller.rb
+++ b/app/controllers/purchases_controller.rb
@@ -319,7 +319,11 @@ class PurchasesController < ApplicationController
       return unless mobile_export_token_matches?
 
       doorkeeper_authorize! :creator_api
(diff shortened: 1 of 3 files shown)


gianfrancopiana:
@claude review

PR #5131, opened 2026-05-19
Add MySQL backing for BlockedObject + Mongo→MySQL sync task

ershad:
## What

Prepares the move of `BlockedObject` off Mongo. The app continues to read and write Mongo; this PR only stands up the destination MySQL table and a resumable, idempotent backfill/delta-sync task. Switching reads and writes is a follow-up.

- New `blocked_objects` MySQL table with `UNIQUE(object_type, object_value)` (matches `BlockedObject.block!`'s `find_or_initialize_by(object_type:, object_value:)` natural key) and a secondary index on `object_value` to cover the type-agnostic `BlockedObject.find_object(object_value)` lookups. No FK on `blocked_by`, per project convention. `precision: 6` on the timestamp columns so `updated_at` watermarks match Mongo's microsecond resolution and the delta sync doesn't re-process the same row.
- `BlockedObjectRecord` — a temporary ActiveRecord class pointing at the new table. Coexists with the Mongoid `BlockedObject` during the sync phase; gets renamed to `BlockedObject` at cutover (and the Mongoid model deleted).
- `Onetime::SyncBlockedObjectsToMysql` — iterates Mongo docs ordered by `(updated_at, _id)` and `upsert_all`s into MySQL with `update_only: [:blocked_at, :expires_at, :blocked_by, :updated_at]` so retries don't disturb `created_at`. Skips records where `expires_at <= now` (no point importing already-expired blocks). `ReplicaLagWatcher.watch` between batches. Resumable via a `since:` watermark; uses `>=` (not `>`) on resume so records sharing the same microsecond `updated_at` as the watermark aren't silently skipped if a previous run crashed mid-batch.

## Why

`blocked_objects` is one of the actively used Mongo collections and the worst-fit for that store: 1.5M rows, point lookups by `object_value`, mutable records (`block!`/`unblock!` upserts and expirations). MySQL is the right home — small, OLTP, mutable, key-value-ish — and moving it lets us close in on retiring the Mongo cluster entirely.

Doing this in two passes (this PR for sync infra; a follow-up for read/write cutover) means production stays on Mongo while we verify the MySQL copy converges, and the cutover PR has a small, reviewable diff.

## Operational steps after merge

1. Run migration.
2. Initial backfill: `Onetime::SyncBlockedObjectsToMysql.process`.
3. Periodically catch up with the printed watermark: `Onetime::SyncBlockedObjectsToMysql.process(since: Time.parse(\"...\"))`.
4. Once parity is verified, the follow-up PR switches reads/writes to `BlockedObjectRecord` and deletes the Mongoid model.

## Before/After

No user-facing change. New table and onetime task only — nothing runs automatically. Walkthrough video to be added before marking ready for review.

## Test Results

Tests to be added in the follow-up cutover PR; this PR is infrastructure only and intentionally has no behavioral change to verify.

---

🤖 AI disclosure: drafted with Claude Opus 4.7 (1M context).


gumclaw:
## Naming improvement opportunity

Sahil and I were discussing this PR and the question came up: since we're migrating `BlockedObject` from Mongo → MySQL anyway, could be a good time to clean up the naming across the board. Current names are pretty vague:

| Current | Proposed | What it does |
|---------|----------|--------------|
| `BlockedObject` (Mongo) | `PlatformBlock` | Admin fraud tool — blocks emails, IPs, fingerprints, products site-wide |
| `BlockedObjectRecord` (MySQL shim) | becomes `PlatformBlock` | Transitional AR model goes away |
| `BlockedCustomerObject` | `SellerBlock` | Seller blocking a specific buyer |
| `blocked_objects` table | `platform_blocks` | |
| `blocked_customer_objects` table | `seller_blocks` | |

Could fold the renames into this PR or do it as a followup — either way, worth doing while we're in this area.

PR #5124, opened 2026-05-19
Remove legacy Sprockets JS pipeline

gumclaw:
## What

Delete the legacy Sprockets `application.js` JS pipeline (jQuery 1.8.3, jQuery UI, jQuery backgroundPosition plugin, jQuery oauthpopup, jQuery UI effects core, Facebook SDK loader) and tear out the related Facebook SDK config plumbing. Modern jQuery 3.7 is kept available globally via Webpack `ProvidePlugin` so the vendored Evaporate library (which still uses `$.ajax({async: false})` for its multipart-upload server-time fetch) keeps working without us rewriting third-party CJS.

## Why

This pipeline dates back to 2012. The app is now fully on React/Inertia + Vite. Removing it eliminates all JS from the Sprockets pipeline, leaving only images and fonts — a step toward dropping `sprockets-rails` entirely.

### Sprockets pipeline removal

- Deleted `app/assets/javascripts/application.js` and `facebook_sdk.js.erb`.
- Deleted vendor files: `jquery-1.8.3.min.js`, `jquery.ui.custom.js`, `jquery.backgroundPosition.js`, `jquery.oauthpopup.js`, `jquery-ui-effects-core.min.js`. The last two were orphans, unreferenced anywhere since the initial commit.
- Removed `link_directory ../javascripts .js` from `app/assets/config/manifest.js`. The manifest now lists only `link_tree ../images` and `link_tree ../fonts`.
- Removed `javascript_include_tag "application"` from `inertia.html.erb`, `application.html.erb`, and `admin.html.erb`.

### jQuery as a webpack global (for vendored Evaporate)

Now that the Sprockets bundle no longer provides `$`/`jQuery` globally, modern jQuery 3.7 from npm is exposed via Webpack `ProvidePlugin`:

- `config/webpack/common.js`: `new webpack.ProvidePlugin({ Routes: "$app/utils/routes", $: "jquery", jQuery: "jquery" })`.
- `package.json` keeps `jquery: ^3.6.3` and `@types/jquery: ^3.5.16` exactly as main has them (npm-installed version is 3.7.1 in both cases — matches main's lockfile).
- `eslint.config.js` keeps `...globals.jquery`.

We don't own Evaporate. Touching its internal synchronous-XHR signing path proved fragile in headless Chrome under CI, so we keep jQuery available as a runtime dep purely to keep the vendored library working. ~30 KB of modern jQuery in the base bundle is the cost.

### Facebook SDK removal

[...]

PR #5129, opened 2026-05-19
Replace jQuery with vanilla JS and remove jquery dependency

gumclaw:
## What
Replace all 10 jQuery usages with vanilla JS and remove the `jquery` + `@types/jquery` dependencies.

## Why
jQuery is only used for two patterns in the entire JS codebase:
1. `$("meta[...]").attr("content")` — reading meta tags (9 usages)
2. `$(document).ajaxSend(...)` — CSRF header on jQuery AJAX (1 usage, dead code)

No code uses jQuery AJAX (`$.ajax`, `$.post`, `$.get`) — all requests go through the `request` util or Inertia. The `ajaxSend` handler never fires.

## Changes
| File | Change |
|---|---|
| `base_page.ts` | Remove dead `ajaxSend` handler, replace meta read |
| `google_analytics.ts` | Replace 2 meta reads |
| `facebook_pixel.ts` | Replace meta read |
| `tiktok_pixel.ts` | Replace meta read |
| `Admin/Form.tsx` | Replace CSRF meta read |
| `Admin/ActionButton.tsx` | Replace CSRF meta read |
| `Admin/.../Content.tsx` | Replace CSRF meta read |
| `Admin/.../WatchedUser.tsx` | Replace CSRF meta read |
| `eslint.config.js` | Remove `globals.jquery` |
| `package.json` | Remove `jquery` + `@types/jquery` |

## Before
```js
const csrfToken = typia.assert<string>($("meta[name=csrf-token]").attr("content"));
```

## After
```js
const csrfToken = typia.assert<string>(document.querySelector("meta[name=csrf-token]")?.getAttribute("content"));
```

## Test results
ESLint passes on all modified files. No functional change — same meta tag values are read, just without jQuery.

**Merge order**: This PR should merge before #5124 (remove legacy Sprockets JS bundle), since the Sprockets bundle is what loads jQuery at runtime today.

> [!NOTE]
> This PR was authored with AI assistance.

PR #5138, opened 2026-05-19
Fix Evaporate.js upload failure: replace jQuery $.ajax with XMLHttpRequest

gumclaw:
## Problem

After the Vite migration (#5067, #5124), all file upload tests in CI are failing — uploads hang forever with a stuck progressbar. This affects ~13 of the 48 slow test shards on every run.

## Root Cause

Evaporate.js (`vendor/assets/javascripts/evaporate.cjs`) uses `$.ajax()` to synchronously fetch the server time before initiating S3 multipart uploads. 

With Vite, jQuery is provided via `unplugin-auto-import`, but its include regex (`/\.[tj]sx?$/`) **does not match `.cjs` files**. So `$` is undefined in `evaporate.cjs`, the `$.ajax()` call throws silently, and the upload initiation never completes.

This worked before the Vite migration because Sprockets provided jQuery as a true global.

## Fix

Replace the single jQuery `$.ajax()` call with a native synchronous `XMLHttpRequest` — which is what Evaporate already uses for all its other HTTP operations (upload parts, signatures, completion).

## Testing

- The `$.ajax()` call was only used in one place (line 581) for fetching server time
- All other Evaporate HTTP operations already use `XMLHttpRequest`
- This is a drop-in replacement with identical behavior (sync GET, read responseText)

### DIFF
--- a/vendor/assets/javascripts/evaporate.cjs
+++ b/vendor/assets/javascripts/evaporate.cjs
@@ -578,12 +578,12 @@ var Evaporate = function(config){
       var dateString = new Date().toUTCString();
 
       if (con.fetchCurrentServerTimeUrl) {
-        dateString = $.ajax({
-          type: "GET",
-          url: con.fetchCurrentServerTimeUrl,
-          async: false,
-          cache: false
-        }).responseText;
+        var timeXhr = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


slavingia:
Should we try and merge this?

May be the reason behind many of the failing tests. 

https://github.com/antiwork/gumroad/issues/5129

PR #5147, opened 2026-05-20
Fix Facebook Pixel on receipt page and domain verification meta tag

nyomanjyotisa:
Closes antiwork/gumroad-private#397

## Summary

**Pixel on receipt page:** Load the seller's tracking pixels (FB/GA/TikTok) and fire the `Purchase` event on the receipt page when arriving from a fresh purchase, instead of from Checkout right before `window.location.href` redirect. The redirect was racing with the in-flight pixel request, causing missing Purchase events in Meta Events Manager. Checkout now skips Purchase firing when redirecting to `content-page` to avoid double-fire (library redirects keep firing from Checkout — unchanged).

**Domain verification meta tag:** Fix `_, content = matchdata` in `CustomDomainConfig` — `MatchData` doesn't implement `to_ary`, so multiple-assignment was setting `content` to `nil`, causing the `<meta name="facebook-domain-verification">` tag to render without its `content` attribute.

## Test plan
- [ ] Verify FB Pixel loads on `/d/<id>?receipt=true` (Meta Pixel Helper shows the pixel)
- [ ] Verify `Purchase` event appears in Meta Events Manager after a real purchase
- [ ] Verify GA `purchase` event still fires (seller + Gumroad-internal)
- [ ] Verify TikTok `CompletePayment` still fires
- [ ] Verify library redirects still fire Purchase from Checkout
- [ ] Verify `<meta name="facebook-domain-verification" content="...">` renders correctly on custom-domain pages

### DIFF
--- a/app/controllers/concerns/custom_domain_config.rb
+++ b/app/controllers/concerns/custom_domain_config.rb
@@ -17,8 +17,8 @@ module CustomDomainConfig
 
       set_meta_tag(title: @user.try(:name_or_username))
       if @user.enable_verify_domain_third_party_services? && @user.facebook_meta_tag.present?
-        _, content = @user.facebook_meta_tag.match(/content="([^"]+)"/)
-        set_meta_tag(name: "facebook-domain-verification", content:)
+        content = @user.facebook_meta_tag[/content="([^"]+)"/, 1]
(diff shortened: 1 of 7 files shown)


nyomanjyotisa:
@claude review

PR #5157, opened 2026-05-20
Fix CLI onboarding link to gumroad.com page

gumclaw:
## What

Follow-up to #5043: the dashboard "Command line" getting started tile now links to `/api#api-cli` instead of the GitHub repository. Because it is now an internal Gumroad page link, it no longer opens in a new tab.

## Why

Sahil asked for the tile to point to a page on gumroad.com. `/api#api-cli` already lands on Gumroad's API page at the Command line section with CLI install instructions, so this avoids adding a duplicate `/cli` page.

## Before/After

No visual change. This only changes the click target for the existing Command line tile.

- Before: `https://github.com/antiwork/gumroad-cli`
- After: `/api#api-cli`

## Test Results

- Passed: static check that `DashboardPage.tsx` uses `/api#api-cli`, no longer includes `https://github.com/antiwork/gumroad-cli`, and removed the external-link attributes from the Command line tile.
- Attempted: `bin/test-confidence` twice. Both attempts failed before running tests because the model planning call returned `Overloaded`.
- Attempted: `npx eslint --fix app/javascript/components/DashboardPage.tsx`. It fails on current `main`-level `Routes` global typing errors in this file; using a literal `/api#api-cli` avoids adding a new `Routes` lint error for this change.
- Attempted: `npx tsc --noEmit --pretty`. It fails on current `main` with `TS2688: Cannot find type definition file for 'vite/client'`.

---

Prompts used:

- "work kanban task t_9e4799df"

### DIFF
--- a/app/javascript/components/DashboardPage.tsx
+++ b/app/javascript/components/DashboardPage.tsx
@@ -136,11 +136,7 @@ const GETTING_STARTED_ITEMS: GettingStartedItemType[] = [
   {
     name: "Command line",
     getCompleted: (stats) => !!stats.used_cli,
-    // Points to /api#api-cli (the existing API page's CLI section) instead of
-    // a dedicated /cli page to avoid duplicating content.
     link: "/api#api-cli",
-    target: "_blank",
-    rel: "noopener noreferrer",
     IconComponent: CliIcon,
     description: "Use Gumroad via the command line interface.",
   },
@@ -153,8 +149,6 @@ type GettingStartedItemProps = {
   link: string;
   IconComponent: React.ComponentType<GettingStartedIconProps>;
   description: string;
-  target?: string;
-  rel?: string;
 };
 
 const Greeter = () => (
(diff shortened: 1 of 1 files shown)


gumclaw:
@claude review once

gumclaw (commented):
Self-review: verified this uses an internal Gumroad path () rather than a GitHub URL, and removed the now-unused external-link props so the tile behaves like the other internal onboarding links.

gumclaw (commented):
Self-review: verified this uses an internal Gumroad path (`/api#api-cli`) rather than a GitHub URL, and removed the now-unused external-link props so the tile behaves like the other internal onboarding links.

gumclaw:
Addressed both nits in 15cd94ca:

- Restored `target="_blank" rel="noopener noreferrer"` for the CLI tile only (other tiles stay in-tab). The CLI tile points at `/api#api-cli` which users typically want to keep open in a separate tab while running CLI commands.
- Added an inline comment explaining why we link to `/api#api-cli` (the existing API page's CLI section) instead of a dedicated `/cli` page.

gumclaw:
@claude review once

PR #5053, opened 2026-05-13
Support existing-customer discounts and renewal tiers

gianfrancopiana:
Fixes #4991

## What

Two new capabilities on discount codes:

- **Limit to existing customers** — only buyers who already own a chosen product can redeem.
- **Tiered by ownership duration** — when limited to existing customers, the discount can scale with how long they've owned (e.g. `0–12 months → 0 %`, `12+ months → 50 %`).

Tiered codes auto-apply on subscription renewals: full price for the first cycle (buyer isn't yet an owner), discounted from the first renewal onward.

## Why

Membership creators want introductory pricing. For example: full price year 1, discounted on renewals. Existing offer codes can only apply to the first charge or every charge. There's no "renewals only". Also closes #1838 (limit discounts to owners of another product) in the same pass.

## Before/After

The discount-form **Settings** section gains a "Limit to existing customers" toggle that opens a product picker and an optional tier editor.

|  | Before | After |
| --- | --- | --- |
| **Desktop · light** | before-desktop-light | after-desktop-light |
| **Desktop · dark** | before-desktop-dark | after-desktop-dark |
| **Mobile · light** | before-mobile-light | after-mobile-light |
| **Mobile · dark** | before-mobile-dark | after-mobile-dark |

The discount list view continues to render correctly with the new field shapes:

discount-list

---

This PR was implemented with AI assistance using Claude Opus 4.7.

PR #5167, opened 2026-05-20
Decouple tier discount toggle from existing-customers gate

gianfrancopiana:
Ref #4991

## What

Splits the discount form's **Settings** card into two independent top-level toggles instead of one nested checkbox:

- **Limit to existing customers** — restricts the code to buyers who own a specific product (unchanged behavior).
- **Tier discount by ownership duration** — varies the discount percentage by how long the buyer has been subscribed.

Both can be enabled independently. A standalone tiered code applied to a membership gives buyers full price for the first cycle (month 0 → first tier, typically 0%) and the matching tier from each renewal onward — without requiring the seller to also turn on the existing-customers gate.

Standalone tiers use the product being purchased (or renewed) as the tenure reference, so a multi-product code can't leak tenure across unrelated products. Renewals on guest subscriptions (no `subscription.user`) still resolve the right tier because evaluation falls back to the subscription's original purchase date.

## Why

The original [#5053](https://github.com/antiwork/gumroad/pull/5053) implementation put "Tiered discounts by ownership duration" inside the "Limit to existing customers" panel as a checkbox, requiring the seller to also set "Must have purchased". Per Slavin's feedback in [#4991](https://github.com/antiwork/gumroad/issues/4991#issuecomment-4492287015):

> Hm, it's strange to limit it to existing customers. Do you think we can enable the functionality for all membership discount codes? E.g 'Tier discount by ownership duration' as a toggle

Decoupling makes the natural "year 1 full price, year 2+ discounted" pattern a one-toggle setup. The product-context scoping closes a leak where a buyer's purchase of product A on a multi-product code would have qualified them for the year-2 tier on a brand new product B.

|Before|After|
|---|---|
| Image | Image|

---


slavingia:
Awesome!

PR #5209, opened 2026-05-21
Stop FightDisputeJob from accumulating dead jobs for already-lost disputes

ershad:
Fixes https://github.com/antiwork/gumroad-private/issues/405

## What

Stops `FightDisputeJob` from piling up failures on disputes whose evidence can never be submitted, and resolves the two underlying error paths that produce them.

- `Charge::Disputable#charge_processor_transaction_id` (`app/models/concerns/charge/disputable.rb`): when the disputable is a `Charge` with `processor_transaction_id IS NULL` and exactly one underlying purchase, fall back to that purchase's `stripe_transaction_id`. Multi-purchase combined charges with a missing column still return nil (no safe single ID to pick).
- `StripeChargeProcessor#create_dispute_evidence_stripe_file` (`app/business/payments/charging/implementations/stripe/stripe_charge_processor.rb`): rescue `ActiveStorage::FileNotFoundError`, notify via `ErrorNotifier`, and return `nil` so the rest of the evidence package is still submitted to Stripe.
- `FightDisputeJob#perform` (`app/sidekiq/fight_dispute_job.rb`): if `disputable.charge_processor_transaction_id` is still blank after the fallback above, mark the evidence as `RESOLUTION_REJECTED` with a descriptive `error_message` instead of raising and re-queueing.
- `FightDisputesJob#perform` (`app/sidekiq/fight_disputes_job.rb`): skip evidence whose dispute is already in `won`/`lost`/`closed` so the hourly cron stops scheduling jobs that can never succeed.
- `Charge::Disputable#handle_event_dispute_won!` / `_lost!`: resolve any unsubmitted `DisputeEvidence` as rejected when the parent dispute closes, so we never strand an evidence row in `resolution=unknown` after the dispute is final.
- `Onetime::ResolveStuckDisputeEvidence` (`app/services/onetime/resolve_stuck_dispute_evidence.rb`): one-shot task to resolve the existing unresolved evidence rows whose disputes have already terminated.

## Why

The dead set was accumulating `FightDisputeJob` failures from two underlying errors:

- `Stripe::Charge instance has invalid ID: nil` — single-purchase combined charges where `Charge.processor_transaction_id` was never populated but the underlying purchase still has the correct `ch_…` ID.
- `ActiveStorage::FileNotFoundError` — dispute evidence attachments whose underlying S3 object had been purged while the blob row remained.

The amplifier was a state-machine gap: both `FightDisputesJob` and `FightDisputeJob` only short-circuit when the evidence is marked `resolved?`, but `handle_event_dispute_won!`/`_lost!` never resolved unsubmitted evidence. So each terminally-failing dispute kept being re-enqueued every hour and burning through its 5-retry budget. A small number of affected disputes was enough to generate tens of thousands of dead jobs over time.

Going forward, the more important reason: a Stripe charge with a missing transaction ID column would have caused us to silently lose the dispute every time. With the purchase fallback, those disputes are now fightable. Disputes that are genuinely unfightable (missing files, no recoverable charge ID) are now marked rejected immediately rather than retried.

## Test plan

- `bundle exec rspec spec/sidekiq/fight_dispute_job_spec.rb spec/sidekiq/fight_disputes_job_spec.rb spec/services/onetime/resolve_stuck_dispute_evidence_spec.rb` — all green
- `bundle exec rspec spec/business/payments/charging/implementations/stripe/stripe_charge_processor_spec.rb -e 'fight_chargeback' -e 'create_dispute_evidence_stripe_file'` — all green
- `bundle exec rspec spec/models/concerns/charge/disputable_spec.rb` — all green (103 examples)

After merge, run `Onetime::ResolveStuckDisputeEvidence.process` from a production console to resolve the currently-stuck rows. The Sidekiq dead set can then be drained of `FightDisputeJob` entries with the existing dead-set management workflow.

---

## AI Disclosure

Drafted with Claude Opus 4.7 (1M context).

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781982414&installation_id=134400930&pr_number=5209&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5209&signature=125dae0604e6c2ba3e919ed922dd048315db18bac0048c5de58104c7d92f9499"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


slavingia:
> ActiveStorage::FileNotFoundError — dispute evidence attachments whose underlying S3 object had been purged while the blob row remained.

Just want to make sure we were not failing to generate due to the Sprockets/Vite changes and this will lead to not fighting at all.

ershad:
@slavingia Sure. We will get notified in Sentry if that happens:

https://github.com/antiwork/gumroad/pull/5209/changes#diff-c47284ad7dfe1f8fac59b77fa2b5dced144533f19e66e415ab59c405cce1af3cR1015

PR #5281, opened 2026-05-26
Show buyer-local currency on product pages (creator opt-in)

gumclaw:
## What

When a creator opts in (Payments settings → "Show buyers their local currency on product pages"), buyers see an approximate price in their local currency across the **entire** product page — the headline price, variant/version/tier prices, membership recurrence prices, coffee options, bundle items, and the pay-what-you-want field (local symbol + suggested price). Checkout still charges in the seller's set currency (settled in USD); the displayed local price is presentment only.

## Why

This gives buyers presentment context without retrofitting checkout, charge currency, or Stripe payment flows. Stripe Adaptive Pricing does not apply to Gumroad's direct Elements + PaymentIntents integration, so this starts as a low-risk, opt-in display layer we can measure before changing the default. Default is off.

## How

- **Country → currency** comes from the already-installed `countries` gem (ISO3166's `currency_code`), not a hand-maintained map. It's restricted to the currencies in `config/currencies.json` — the set we can both format (Intl) and accept as input (the `PriceInput` stack). Buyers whose currency falls outside that set fall back to the seller's set price.
- **Rates** are derived on the render path from the USD-based rates that `UpdateCurrenciesWorker` already refreshes hourly into Redis. The cross rate for a pair is `to_rate / from_rate`; there is no synchronous FX call while rendering, and a missing rate degrades gracefully to the set price.
- **Display** goes through one shared helper, `formatBuyerLocalOrSetPrice`, applied consistently across `PriceTag`, the version/tier/recurrence/coffee selectors, and bundle items. schema.org microdata (`itemProp="price"` / `priceCurrency`) and transaction-mechanics amounts (PWYW minimum-price validation, discount thresholds) intentionally stay in the set currency — what checkout actually charges.
- **Pay-what-you-want**: the field shows the buyer's local symbol and suggested price and accepts input in their local currency. The entered amount is converted back to set-currency cents for `selection.price.value`, so the charge and the minimum-price floor remain set-currency. Localization is confined to the input boundary, and the entered value is tracked in the input currency to avoid set↔local round-trip rounding jitter while typing.

## Before/After

The displayed price text is swapped (e.g. `$2` becomes `£1.69`); no new UI element is introduced.

Before — annotation, both prices shown

After — buyer-local price replaces product price

A fuller walkthrough covering variant, bundle, and PWYW localization (light/dark, mobile/desktop) is being recorded.

## Out of scope

Tier 2 — actually charging buyers in their local currency via the PaymentIntent `currency` parameter plus Stripe FX Quotes — is not part of this PR. That needs `Purchase.presentment_currency`, `Purchase.presentment_amount_cents`, and a `StripeChargeProcessor` change. This PR does not touch checkout, charge currency, Stripe charge processing, or the `purchases` schema.

## Test Results

- `bundle exec rubocop` — clean on touched Ruby files
- `DISABLE_TYPE_CHECKED=1 npx eslint` + `npx tsc --noEmit` — clean on touched TS/TSX files
- `bundle exec rspec` — `buyer_local_currency_helper_spec`, `product_presenter/buyer_local_currency_spec`, and the request/system `buyer_local_currency_display` specs green locally, including a new system example asserting the PWYW field renders in the buyer's local currency

---

This PR was implemented with AI assistance: the initial display layer with Codex (GPT-5), and the full-page localization, rate refactor, and PWYW work with Claude Opus 4.8.

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5281"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782355868&installation_id=134400930&pr_number=5281&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5281&signature=6eb1514e20f31d49f3b6e1202ab9a87e3ea59827e17dbd98532399eeb753aac2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Wide buyer-facing pricing and PWYW display changes could mislead if rates or conversion fail, but charges stay in set currency and missing rates fall back to the seller price.
> 
> **Overview**
> Adds an **opt-in, feature-flagged** layer so creators can show buyers an **approximate local-currency price** on product surfaces while **checkout still charges the seller’s set currency**.
> 
> **Backend:** Persists `show_buyer_local_currency` on `User` and saves it from Payments settings. `CurrencyHelper` maps buyer IP/country to a supported currency, converts using **hourly Redis USD rates only** (no live FX on render), and builds `buyer_currency_display` / buyer-local price props with safe fallbacks. Presenters for product pages, cards, checkout line items, and receipt analytics pass those props through.
> 
> **Frontend:** `formatBuyerLocalOrSetPrice` and related types drive localized display in `PriceTag`, configuration selectors, PWYW (local input, set-currency charge), installments, and cards; schema.org price fields stay in set currency. **Analytics:** `buyer_currency_display_view` (session-deduped) and purchase payload extensions. Payments settings adds a toggle when the feature is enabled for the seller.
> 
> Specs cover helper edge cases, presenter props, request/Inertia, and system flows (including EU VAT + localized display).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea9d4cfa4548f4e23c675dfff39422b2db906b6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gianfrancopiana:
@gumclaw Can you add screenshots showing before/after?

gumclaw:
## Interaction with Pages (#5171 / `pages-feature` branch)

Flagging before either ships further — these two land near each other and the cross-cutting concern isn't visible from inside either PR.

### The collision

Pages renders seller-authored `links.custom_html` inside a sandboxed iframe with `default-src 'none'; connect-src 'none'; script-src 'unsafe-inline' <CDN allowlist>`. `connect-src 'none'` is intentional — seller JS can animate but can't phone home. That also means the iframe **cannot fetch the buyer-local price**, and there's no React `PriceTag` inside the seller's HTML to read `buyer_local_price_cents` from props.

Result: a creator who enables this PR's setting *and* publishes a custom landing page gets the worst of both — the buyer sees the hardcoded `$29` the seller typed, even though the default page would show `£23` for the same buyer.

### Fix that keeps CSP locked

Pages already runs `Pages::Interpolator` to substitute `data-gumroad-field` placeholders server-side. Add price/currency to the same contract — same request, same GeoIP, same presenter:

```html
<button>Buy now for <span data-gumroad-price></span></button>
<!-- renders for a UK buyer with opt-in: -->
<button>Buy now for <span data-gumroad-price>£23</span></button>
```

Placeholders worth adding:
- `data-gumroad-price` → buyer-local formatted price (this PR) OR product-currency price (no opt-in)
- `data-gumroad-price-original` → strikethrough/PPP support
- `data-gumroad-currency-code` → ISO 4217 string for sellers who want custom formatting

Zero new infra. No new HTTP from inside the sandbox. CSP unchanged.

### Tier 2 implication (out of scope for this PR, but relevant to sequencing)

Once charge currency follows buyer currency, the buy-button `postMessage` from iframe → parent shell needs to carry presentment currency:

```js
postMessage({ type: 'gumroad:checkout', productId, presentmentCurrency: 'GBP' })
```

Parent shell validates against the `buyer_currency` it already computed (anti-tampering — buyer can't post `presentmentCurrency: 'CNY'` to dodge USD checkout). The current postMessage payload doesn't carry currency at all, so this is additive.

### Caching gotcha

Pages will want to cache rendered `custom_html` aggressively (it's near-static). Multi-currency makes the rendered output buyer-specific. Three options:

1. Cache by `(product_id, buyer_currency)` — N ≈ 30 currencies, predictable, no edge infra
2. ESI/edge-include — inject price tags at the edge per country
3. Cache the shell, interpolate at request time — simplest, defers Pages' caching win

Lean #1.

### Schema microdata

This PR keeps `itemProp="price"` / `priceCurrency` as the product's configured currency on the default page (correct, since charge is in product currency). On custom pages, sellers control all markup — Google sees whatever the seller wrote. Worth a docs note in the Pages PR once both land.

### Sequencing recommendation

Ship #5281 and Pages independently. Integration (interpolator placeholders + cache-by-currency) is a follow-up PR. Don't block either.

cc @slavingia @ershad @gianfrancodemarco

gumclaw:
## Staging test plan — extensive, including analytics

Before merge, deploy this branch to staging and walk through end-to-end. Specs catch logic; staging catches the integration mess (Redis warmup race, GeoIP cassette drift, GA payload shape, Inertia hydration mismatch).

### Setup
- [ ] Deploy `buyer-currency-display` to staging
- [ ] Pick a test creator account, enable **Show buyer-local currency** in Payment settings
- [ ] Have a product priced in USD (e.g. `$29`)
- [ ] Have a control product on a different creator account with the setting OFF (regression baseline)

### Buyer-side geo matrix
Hit the product page from at least 5 IPs (use a VPN or a curl with `X-Forwarded-For` header into the staging LB if it's allowed):

- [ ] **US buyer** → expect `$29` (no conversion, source currency == buyer currency, presenter should short-circuit)
- [ ] **UK buyer** → expect `£X.XX`, recent FX rate, locale-correct formatting (£ not GBP, no trailing currency code)
- [ ] **EU buyer (DE/FR)** → expect `€X,XX` (comma decimal separator, symbol position varies — verify it renders the same as native sites)
- [ ] **JP buyer** → expect `¥XXX` (no decimal places — JPY is unit-of-1, this is the spot where rounding bugs hide)
- [ ] **AU buyer** → expect `A$X.XX` or `$X.XX AUD` (verify which the formatter picks — buyers can confuse `$` for USD)
- [ ] **Unsupported country (e.g. CU/IR/KP via mocked geo)** → falls back to USD `$29` cleanly, no error, no half-rendered state

### UI surface coverage
- [ ] Product page (`/l/:permalink`) — main `PriceTag`
- [ ] Product card on creator profile / discover — `Product/Card.tsx` change
- [ ] Cart page (`Checkout/cartState.ts` change) — does the cart display match the page price?
- [ ] Downloads page (`DownloadPage/WithContent.tsx`) — display-only for post-purchase users, sanity-check
- [ ] PPP / strikethrough price — verify both prices convert, not just the primary
- [ ] Mobile viewport — formatter doesn't overflow long currency strings like `KZT 12,345.67`

### Checkout regression — the critical one
**Checkout still charges in product currency** (per PR description). Verify:
- [ ] UK buyer sees `£23` on the page → clicks Buy → Stripe Elements shows `$29 USD` charge → completes
- [ ] Confirmation email shows USD
- [ ] `Purchase` record in DB shows USD as `currency_type`
- [ ] No discrepancy report fires (we don't want a buyer thinking they were quoted £23 and charged $29 — surface the charge currency clearly at the Stripe step)
  - **If this lands wrong, it's the disclosure problem that gets us in trouble, not the conversion math.** Worth a one-line "Charged in USD" note above the Stripe element when buyer currency ≠ product currency.

### FX rate / caching
- [ ] First page load for a fresh currency triggers the snapshot fetch — verify it doesn't block render (presenter should omit local price if rate not cached)
- [ ] Subsequent loads hit Redis cache — verify TTL ~24h
- [ ] Force-expire a cache key (`redis-cli DEL`) and reload → re-fetches, renders again
- [ ] Kill the snapshot fallback (set the env stale or block the host at the network layer) → presenter cleanly omits local price, page renders in product currency, no exception in Sentry
- [ ] `PrewarmBuyerLocalCurrencyRateJob` runs in staging on schedule, populates expected pairs

### Analytics — the part that's easy to miss
The PR touches `google_analytics.ts` and `user_analytics.ts` (+65 lines). Verify the payload:

- [ ] Open the product page as a non-US buyer with browser devtools → Network tab
- [ ] `view_item` GA4 event fires with `currency` field
  - **Question**: is `currency` set to the *product* currency (USD) or the *displayed* currency (GBP)?
  - GA4 best practice: `currency` should match the price you submit, and `value` should be in that currency. Pick one (product currency, since that's what we actually charge) and document it. Mixing creates broken revenue reports.
- [ ] `add_to_cart` fires with consistent currency
- [ ] `begin_checkout` fires with consistent currency  
- [ ] `purchase` event from Stripe completion uses **product currency** (matches the actual charge)
- [ ] Verify Mixpanel / segment / wherever else `user_analytics.ts` sends — payload includes `buyer_currency`, `buyer_local_price_cents`, `displayed_currency` as distinct fields so we can answer "for buyers who saw their local price, did conversion go up?"
- [ ] No `currency: undefined` or `currency: ""` in any event — fail-closed to product currency
- [ ] GA4 DebugView shows the events arriving and the currency dimension populated

### Setting toggle behavior
- [ ] Toggle setting ON in Payment settings → reload product page in incognito → local price appears
- [ ] Toggle OFF → reload → reverts to product currency
- [ ] No stale cache: toggling shouldn't require a deploy or cache flush to take effect

### Edge cases
- [ ] Product with multiple variants at different prices — all variants convert correctly
- [ ] Pay-what-you-want product — input field, min/suggested values all in the right currency
- [ ] Free product (`$0`) — doesn't try to convert, renders "Free" or `$0` consistently
- [ ] Subscription product — monthly price converts, annual price converts, both render
- [ ] Bundle product — bundle price converts, not the individual item sum

### Observability
- [ ] No new Sentry errors with the FX/presenter/GA changes during the test window
- [ ] Datadog: `presenter.buyer_local_currency.*` metric exists if instrumented (if not, worth adding a counter for "currency conversions applied" vs "fallback used" so we can monitor coverage)
- [ ] Logs aren't spamming "rate not available for XYZ" — should warn once per cache miss, not per request

### Rollout
- [ ] Feature flag (Flipper or env) wrapping the *setting* itself, so we can disable the toggle in payment settings if something is wrong, separate from rolling back the deploy
- [ ] Decide: beta cohort first (10–50 creators) or all-creators-but-default-off? Lean beta — easier to iterate with engaged feedback than to debug from anonymous reports.

### Sign-off
- [ ] Eng: backend + frontend
- [ ] QA: at least 3 of the geo cases run by a human, screenshots attached
- [ ] Analytics: someone who actually reads the GA dashboards confirms the events arrive shaped correctly. **Don't merge if no one has eyes on the GA4 side.**

cc @slavingia @ershad @gianfrancodemarco

gumclaw:
@gumclaw addressed review: schema.rb version corrected to match the only migration in this PR (20261129000000). Other Bugbot/Greptile points were addressed in earlier commits — see latest commit message for the mapping. Merged main, resolved schema.rb version conflict.

spec/migrations/add_show_buyer_local_currency_to_users_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
ershad: We can remove the spec file for migration.

db/migrate/20261129000000_add_show_buyer_local_currency_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddShowBuyerLocalCurrencyToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :show_buyer_local_currency, :boolean, default: false, null: false
```
ershad: Instead of adding a column, we can add a JSON data field on User?

Ref: https://github.com/antiwork/gumroad/blob/main/CONTRIBUTING.md#code-patterns

spec/migrations/add_show_buyer_local_currency_to_users_spec.rb
slavingia: Done in 2bce6640 — removed the migration spec along with the migration itself (see below).

db/migrate/20261129000000_add_show_buyer_local_currency_to_users.rb
slavingia: Good call — CONTRIBUTING.md forbids schema changes to the `users` table. Switched to a `json_data` field in 2bce6640: added `show_buyer_local_currency` reader/writer/predicate on `User` (writer casts through `ActiveModel::Type::Boolean`), dropped the migration + spec, removed the column from schema.rb, and updated the fixtures to nest the value under `json_data`.

app/javascript/utils/currency.ts
```diff
@@ -101,3 +101,15 @@ export const formatPriceCentsWithoutCurrencySymbolAndComma = (code: CurrencyCode
...
+  const currency = currencyCode.toUpperCase();
+  const formatter = new Intl.NumberFormat("en-US", { style: "currency", currency });
+  const configuredCurrency = Object.entries(currenciesMap).find(([code]) => code === currencyCode.toLowerCase())?.[1];
+  const subunitToUnit = configuredCurrency
+    ? "single_unit" in configuredCurrency
+      ? 1
+      : 100
+    : 10 ** formatter.resolvedOptions().maximumFractionDigits;
+  return formatter.format(amountMinorUnits / subunitToUnit);
+};
```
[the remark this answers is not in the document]
slavingia: Fixed in 2bce6640. The Intl-CLDR fallback was the culprit — HUF and IDR are zero-decimal in CLDR, so the frontend divided by 1 while the backend (`money.rb` forces `subunit_to_unit: 100` for both) produced 1/100 units, inflating displayed prices 100x. `formatMinorUnitPriceWithIntl` now mirrors the backend's Money subunit directly (single-unit currencies → 1, everything else → 100), which matches `subunit_to_unit` for all 30 currencies in `COUNTRY_TO_CURRENCY` and no longer depends on the browser's CLDR data.

gumclaw:
## CI fix — `Test Fast 14` failure resolved

The only failing check was **Test Fast 14**: 8 errors, all the same root cause —

```
Failure/Error: self[:json_data] ||= {}
JSON::ParserError: invalid number: '---' at line 1 column 1
```

**Cause:** the new `buyer_currency_seller` fixture in `spec/fixtures/users.yml` wrote `json_data` as a nested YAML hash:
```yaml
json_data:
  show_buyer_local_currency: true
```
The `json_data` column is `serialize coder: JSON` (`app/models/concerns/json_data.rb`). Rails dumped the nested YAML hash into the column as a YAML document (`---\n...`), and the JSON coder then couldn't parse it back on read → `JSON::ParserError`. That blew up every spec touching these fixtures (the `ProductPresenter` buyer-local-currency props + `User#show_buyer_local_currency`).

**Fix** (`a6dd9ec52`): store `json_data` as a literal JSON string so it round-trips through the JSON coder:
```yaml
json_data: '{"show_buyer_local_currency":true}'
```

**Verified locally** (rbenv 3.4.3): the 8 `ProductPresenter` buyer-currency examples + the `User#show_buyer_local_currency` spec now pass (previously all errored). The 4 `create(:purchase)` "couldn't charge your card" failures elsewhere are the known local stripe-mock limitation — they fail identically on clean `main` and aren't in scope here; CI uses real Stripe test cards. Pushed; CI re-running.

app/helpers/currency_helper.rb
```diff
@@ -52,6 +96,142 @@ def get_rate(currency_type)
...
+    {
+      product_id: product.external_id,
+      creator_opted_in: product.user.show_buyer_local_currency?,
+      buyer_currency_shown: product.price_currency_type.to_s.downcase,
+      product_currency: product.price_currency_type.to_s.downcase,
+      buyer_local_price_cents: nil,
+      rate: nil,
+      variant: "default",
+    }
+  end
```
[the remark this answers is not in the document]
gumclaw: Fixed in e67df2aec. The rescue now returns a static safe default (only `product.external_id`, the already-loaded subject) and never re-runs `show_buyer_local_currency?` / `price_currency_type` — so if one of those was the original raiser, the handler can't re-raise. Also flattened the body into guard returns. Added a regression spec (`#buyer_currency_display_props` → "returns a safe static default without re-raising") that stubs an internal raise and asserts a default-variant hash instead of a propagated error.

spec/migrations/add_show_buyer_local_currency_to_users_spec.rb
gumclaw: Done — the migration and its spec were both removed in 2bce66404. The opt-in is now stored via `json_data` on User (no column, no migration), so there's no migration spec to keep.

db/migrate/20261129000000_add_show_buyer_local_currency_to_users.rb
gumclaw: Done in 2bce66404 — switched to a `json_data` field per CONTRIBUTING. `User#show_buyer_local_currency` now reads/writes via `json_data_for_attr`/`set_json_data_for_attr` (app/models/user.rb:180-186), and the column migration + schema change were dropped entirely.

app/javascript/utils/currency.ts
```diff
@@ -101,3 +101,11 @@ export const formatPriceCentsWithoutCurrencySymbolAndComma = (code: CurrencyCode
     useGrouping: false,
   });
 };
+
+export const formatMinorUnitPriceWithIntl = (currencyCode: string, amountMinorUnits: number): string => {
+  const currency = currencyCode.toUpperCase();
+  const formatter = new Intl.NumberFormat("en-US", { style: "currency", currency });
+  const configuredCurrency = Object.entries(currenciesMap).find(([code]) => code === currencyCode.toLowerCase())?.[1];
+  const subunitToUnit = configuredCurrency && "single_unit" in configuredCurrency ? 1 : 100;
```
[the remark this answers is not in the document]
gumclaw: Fixed in 847719b28 — made the backend the single source of truth for the minor-unit divisor.

The fragility was real: `formatMinorUnitPriceWithIntl` derived `subunitToUnit` from `currencies.json` (only ~19 product-pricing currencies), so the 12 buyer currencies not in that map silently defaulted to 100.

One important subtlety I verified before fixing: **deriving the divisor from `Intl` would have been wrong.** Gumroad's Money gem is non-ISO for several of these — KRW, HUF, and IDR all use `subunit_to_unit = 100`, whereas `Intl`/ISO 4217 treats them as zero-decimal. Using `Intl.resolvedOptions().maximumFractionDigits` would have reintroduced the exact KRW 100x bug fixed earlier in this PR.

So instead:
- `buyer_local_price_props` now emits `buyer_local_currency_subunit_to_unit` straight from `subunit_to_unit(buyer_currency)` — the same Money-gem value used for all backend conversion math.
- `formatMinorUnitPriceWithIntl` takes an explicit `subunitToUnit` and uses it when present; the `currencies.json` heuristic stays only as a fallback (no behavior change for any existing caller).
- Threaded through `PriceTag` + both product/card call sites + parser type, and asserted in the presenter specs.

Verified: 19 buyer-currency specs green, rubocop + eslint + prettier + tsc clean.

spec/spec_helper.rb
```diff
@@ -220,6 +220,7 @@ def teardown_fixtures
   config.include ErrorResponses
   config.mock_with :rspec
   config.file_fixture_path = "#{::Rails.root}/spec/support/fixtures"
+  config.fixture_paths = ["#{::Rails.root}/spec/fixtures"]
```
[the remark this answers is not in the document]
slavingia: Fixed in 1bf4f008. Dropped the global `config.fixture_paths` and the partial `users`/`links`/`prices` fixtures, and converted both buyer-currency specs to FactoryBot (the project standard per CONTRIBUTING). No spec declares `fixtures :…` anymore, so there is no global fixture path that could leak partial core-table records into other suites.

app/helpers/currency_helper.rb
```diff
@@ -52,6 +96,142 @@ def get_rate(currency_type)
...
+    cached_rate = currency_namespace.get(cache_key)
+    return BigDecimal(cached_rate) if cached_rate.present? && cached_rate.to_d.positive?
+
+    # On cold cache: fall back to last known good rate (stale-while-revalidate) and enqueue
+    # a background refresh. We never make a synchronous HTTP call on the render path.
+    PrewarmBuyerLocalCurrencyRateJob.perform_async(from_currency, to_currency)
+
+    stale_rate = currency_namespace.get(buyer_local_currency_stale_rate_cache_key(from_currency:, to_currency:))
+    BigDecimal(stale_rate) if stale_rate.present? && stale_rate.to_d.positive?
+  end
```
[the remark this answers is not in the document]
slavingia: Fixed in 1bf4f008. The enqueue is now debounced behind a short-lived Redis NX key (`…:prewarm_enqueued`, 60s TTL) via `enqueue_prewarm_buyer_local_currency_rate`, so a traffic spike right after the daily cache expires fires `perform_async` at most once per currency pair per minute instead of on every render-path cache miss. `lock: :until_executed` still dedups the job itself; this removes the upstream enqueue + dedup-digest round-trip churn the comment flagged.

app/javascript/utils/currency.ts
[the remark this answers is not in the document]
slavingia: Fixed in 1bf4f008. The fallback now reads the value (`"single_unit" in configuredCurrency && configuredCurrency.single_unit`) instead of just checking key presence, matching `findCurrencyByCode`. Today only JPY carries `single_unit` (always `true`), so behavior is unchanged — but the two code paths are now consistent and a future `single_unit: false` entry would be handled correctly.

app/javascript/components/Product/Layout.tsx
```diff
@@ -298,6 +298,10 @@ const CtaBar = ({
           isPayWhatYouWant={isPWYW}
           isSalesLimited={product.is_sales_limited}
           creatorName={product.seller?.name}
+          buyerCurrency={product.buyer_currency}
+          buyerLocalCurrencyRate={product.buyer_local_currency_rate}
+          buyerLocalPriceCents={product.buyer_local_price_cents}
+          buyerLocalOriginalPriceCents={product.buyer_local_original_price_cents}
```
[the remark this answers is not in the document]
slavingia: Fixed in 1bf4f008. The sticky-header `PriceTag` now passes `buyerLocalCurrencySubunitToUnit={product.buyer_local_currency_subunit_to_unit}`, matching the `index.tsx` and `Card.tsx` call sites, so `formatMinorUnitPriceWithIntl` uses the backend authoritative `subunit_to_unit` rather than the `currencies.json` heuristic (matters for KRW/HUF/IDR).

slavingia:
Do not merge until testing!

app/helpers/currency_helper.rb
```diff
@@ -5,6 +5,51 @@ module CurrencyHelper
   # Note: To reference a currency in code, use Currency::[3-char-ref].
   # e.g. Currency::USD, Currency::CAD
 
+  COUNTRY_TO_CURRENCY = {
```
ershad: We could potentially support more currencies here, but let's keep this list in the first iteration to save us from formatting edge cases with some currencies. Can expand the support later if required.

app/javascript/components/Product/PriceTag.tsx
```diff
@@ -3,13 +3,12 @@ import * as React from "react";
 import { classNames } from "$app/utils/classNames";
```
ershad, line 3: Changes in this component would show the local price only on the main product price and product cards. Ideally, we need to show local prices on PWYW suggested price, variants, etc. Fixing it.

cc: @slavingia

slavingia: Yes everywhere.

gumclaw:
## Fixed the Greptile P-level finding — checkout-breaking nil currencies

Pushed `466d07d29`.

**The bug:** the `rescue StandardError` in `buyer_currency_display_props` returned `buyer_currency_shown: nil` and `product_currency: nil`, but the TS `BuyerCurrencyDisplay` type declares both as **non-nullable** (`buyer_currency_shown: string`, `product_currency: CurrencyCode`). `buyer_currency_display` flows all the way to `pages/Checkout/Show.tsx`, so when the rescue fired on the checkout path, `typia.assert` would throw and **break checkout for that buyer** — exactly the scenario the degradation path is supposed to prevent.

**The fix:** the rescue now emits non-nil currency strings. It re-derives `product_currency` behind its own inner guard (the original failure may have been in `price_currency_type`/`product.user` itself), falling back to `Currency::USD` (`"usd"`, a valid `currenciesMap` key -> valid `CurrencyCode`) only as a last resort. Both fields are always populated with the same safe currency, `variant` stays `"default"`, and no local price is shown — graceful degradation without the crash.

**Tests:** added two regression specs to `buyer_local_currency_helper_spec.rb`:
- rescue path never yields nil currencies (asserts `"usd"` + `String` type)
- worst case where re-deriving the currency *also* raises still emits `"usd"`

Local run (rbenv 3.4.3): **16 examples, 0 failures**; `rubocop` clean on the touched file. No TS files changed — the fix makes the Ruby honor the existing TS contract.

Re: the other Greptile point (hardcoded "USD" in the Payments settings description) — that was already corrected in an earlier commit; the description now reads *"Checkout still uses your set currency."*

Still **not merging** per @slavingia's "Do not merge until testing!" — this just clears the correctness blocker so it's ready when the staging walkthrough happens.

PR #5369, opened 2026-06-02
Expose product sections in v2 product responses

gianfrancopiana:
Ref antiwork/gumroad-cli#101

## What

Adds product page sections to `GET /v2/products/:id`.

The response now includes `main_section_index` and an ordered `sections` array for full product responses. Product-listing and featured-product sections include their relevant product references as v2 product IDs. Other section types return only the common fields so the API does not commit to unsupported rich-text, posts, wishlist, or subscribe contracts yet.

## Why

The CLI needs a read path before it can safely add section-management commands. Product sections are not inline JSON content; they are ordered section records with type-specific data and raw product references.

This keeps this PR read-only and small while establishing the public response shape future write endpoints can build on.

---

This PR was implemented with AI assistance using GPT-5.5 xhigh.

---

> [!NOTE]
> **Low Risk**
> Read-only API expansion with seller-scoped product ID resolution; no write paths or auth changes.
> 
> **Overview**
> **Full v2 product responses** (`GET /v2/products/:id`, non-slim) now expose **`main_section_index`** and an ordered **`sections`** array so clients can read product page layout without write APIs yet.
> 
> A new **`Api::ProductSectionsPresenter`** builds each section from `product.sections` order with common fields (`id`, `type`, `header`, `hide_header`). **Products** sections add `shown_products` (v2 external IDs), sort/filter flags, and **featured_product** sections add `featured_product` when the target belongs to the seller—otherwise that key is omitted. Other section types (e.g. rich text) stay minimal so the API does not promise full content for posts, wishlists, or subscribe yet.
> 
> **`SHOW_PRODUCT_ASSOCIATIONS`** preloads `:seller_profile_sections` for show. **Slim index** responses still omit `sections` and `main_section_index`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2f28d864eb83335ea30a6094848df447d6ba495. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/controllers/api/v2/links_controller.rb
+++ b/app/controllers/api/v2/links_controller.rb
@@ -16,6 +16,7 @@ class Api::V2::LinksController < Api::V2::BaseController
   SHOW_PRODUCT_ASSOCIATIONS = (BASE_PRODUCT_ASSOCIATIONS + [
     :page,
     :ordered_alive_product_files,
+    :seller_profile_sections,
     :alive_rich_contents,
     { variant_categories_alive: [{ alive_variants: :alive_rich_contents }] },
   ]).freeze
--- a/app/models/concerns/product/as_json.rb
+++ b/app/models/concerns/product/as_json.rb
@@ -144,6 +144,8 @@ module Product::AsJson
           "custom_html" => custom_html,
           "rich_content" => rich_content_json,
           "has_same_rich_content_for_all_variants" => has_same_rich_content_for_all_variants?,
+          "main_section_index" => main_section_index || 0,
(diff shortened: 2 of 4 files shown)


slavingia:
Think there was a miscommunication here. I meant to say - though now I realize it's a different request than what the OG was - that we should support updating the product's content sections via API.

And instead of rich text editing, each should be able to be vibe coded similarly to product pages (schema should already support).

PR #5434, opened 2026-06-08
Alert when Stripe balance is too low to cover upcoming seller payouts

ershad:
## What

Adds a proactive monitor that flags when Gumroad's Stripe platform balance is too low to cover the upcoming seller payouts:

- **`StripeBalanceCheckService`** — computes the upcoming Gumroad-held Stripe payouts + a Stripe minimum balance, vs the available balance.
- **`SendStripeBalanceCheckNotificationJob`** — runs daily ahead of the UTC 10:00 seller-payout jobs and sends an internal payments notification (`InternalNotificationWorker`, the same channel as the PayPal balance alert) **only when** the balance is low. It records the result in Redis (`stripe:balance_topup_needed`).
- **`GET /healthcheck/stripe_balance`** — reads that Redis flag and returns `200` ("topup not required") or `503` ("topup required"), mirroring `/healthcheck/paypal_balance`, so an external monitor can page on it too. No amounts are exposed.

The alert fires when:

```
available USD Stripe balance  <  next cycle's Gumroad-held payouts  +  Stripe minimum balance
```

## Why

Stripe pays out Gumroad's platform balance to Gumroad's bank on its **own automatic schedule** (since ~May 2025; the code-driven sweep worker is disabled in prod via the `skip_transfer_from_stripe_to_bank` flag). Seller payouts are platform → connected-account transfers that **draw from that same balance**. When the balance gets drained before a payout cycle runs, seller payouts fail with *"insufficient funds in your Stripe account"* — which is exactly how the June 2 cycle surfaced: we only found out **after creators reported missing payouts**.

The fix is to know about it first. This mirrors the established `SendPaypalTopupNotificationJob` + `/healthcheck/paypal_balance` pattern so it slots into the team's existing payout-readiness routine — notifying only on the low-balance case, and exposing a monitorable healthcheck.

Design notes:
- **Upcoming-payout figure** uses a new `PayoutEstimates.estimate_gumroad_held_stripe_cents(date)` — a set-based aggregation that sums unpaid balances on the platform (Gumroad-held, `user_id IS NULL`) Stripe merchant accounts via the indexed `balances` aggregation. The existing `estimate_held_amount_cents` does a per-user loop that's too slow for a daily job. The new estimate **omits the per-user payability gates** (compliance, paused, minimum payout), so it slightly over-states what will actually pay out — the safe direction for a "is the balance high enough?" alert.
- **Stripe minimum balance** is added on top of the estimate to absorb a larger-than-estimated cycle or in-flight charges/refunds. Configured at runtime via Redis (`RedisKey.stripe_minimum_balance_cents`), following the same convention as other operational thresholds, and defaults to zero until set.
- **No numbers exposed publicly:** the healthcheck returns only "topup required" / "topup not required" + HTTP status. Amounts live only in the internal notification.
- **Scope:** checks the USD balance against a USD-denominated requirement, matching the existing bank-sweep worker's assumption. Monitoring only — does not change the payout mechanism or re-enable the code-driven sweep.

## Before / After

Non-user-facing (internal payments alerting + healthcheck). No UI. Notification fires only when the balance is low (amounts illustrative):

```
Stripe balance needs to be $X (... for upcoming payouts + ... Stripe minimum balance) to pay out all creators.
Current Stripe balance is $Y.
A top-up of $Z is needed.
```

Healthcheck: `GET /healthcheck/stripe_balance` → `200 "Stripe balance: topup not required"` or `503 "Stripe balance: topup required"`.

_Walkthrough video to be attached._

## Test Results

Specs cover the set-based estimate, the service, the job (notifies + sets flag when low, silent + sets flag when sufficient, non-prod no-op), and the healthcheck endpoint (200/503 for false/true/unset flag). All green locally:

```
estimate + service + job: 12 examples, 0 failures
healthcheck (stripe_balance): 3 examples, 0 failures
```

## Deploy note

Set the Stripe minimum balance in prod Redis (`RedisKey.stripe_minimum_balance_cents`) to enable the safety margin; until then the check still fires whenever the balance can't cover the upcoming payouts themselves.

---

AI disclosure: drafted with Claude Opus 4.8.

---

> [!NOTE]
> **Medium Risk**
> Touches payout readiness and operational alerting around real money movement, but does not change payout execution—only adds estimates, Redis flags, and notifications.
> 
> **Overview**
> Adds **proactive Stripe platform balance monitoring** so payments can top up before seller payout cycles fail for insufficient funds, following the existing PayPal top-up alert pattern.
> 
> **`PayoutEstimates.estimate_gumroad_held_stripe_cents`** aggregates unpaid Gumroad-held Stripe balances (platform merchant accounts) in one query for the daily check; it intentionally skips per-user payability filters so the estimate errs high.
> 
> **`StripeBalanceCheckService`** compares live USD Stripe balance to upcoming Gumroad-held payouts (next scheduled payout end date) plus a Redis-configured minimum (`stripe:minimum_balance_cents`). **`SendStripeBalanceCheckNotificationJob`** runs daily in production, updates **`stripe:balance_topup_needed`**, and sends a red **`InternalNotificationWorker`** message only when short; **`GET /healthcheck/stripe_balance`** mirrors PayPal (503 when top-up needed).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c74df3bf2ef92a84482a3d36f7a24795a1d25dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/services/stripe_balance_check_service.rb
+++ b/app/services/stripe_balance_check_service.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# Determines whether Gumroad's Stripe platform balance is large enough to
+# fund the upcoming seller payouts. Stripe pays out Gumroad's balance to
+# Gumroad's bank automatically, and seller payouts (platform -> connected
+# account transfers) draw from the same balance, so the balance can be
+# starved before a payout cycle runs. This service powers a proactive alert
+# so the balance can be topped up before any seller payout fails.
+#
+# The minimum balance kept on top of the estimated upcoming payouts is
+# configured at runtime via Redis (RedisKey.stripe_minimum_balance_cents).
+class StripeBalanceCheckService
+  def initialize
+    @upcoming_payouts_cents = calculate_upcoming_payouts_cents
+    @minimum_balance_cents = $redis.get(RedisKey.stripe_minimum_balance_cents).to_i
+    @current_balance_cents = StripeTransferExternallyToGumroad.available_balances["usd"].to_i
+  end
+
+  attr_reader :upcoming_payouts_cents, :minimum_balance_cents, :current_balance_cents
+
+  def required_balance_cents
+    upcoming_payouts_cents + minimum_balance_cents
+  end
+
+  def topup_amount_cents
+    @topup_amount_cents ||= required_balance_cents - current_balance_cents
+  end
+
+  def topup_needed?
+    topup_amount_cents > 0
+  end
+
+  private
+    # Only the funds Gumroad itself holds need to come out of Gumroad's Stripe
+    # balance; balances held by Stripe are funded by Stripe directly.
+    def calculate_upcoming_payouts_cents
+      PayoutEstimates.estimate_gumroad_held_stripe_cents(
+        User::PayoutSchedule.next_scheduled_payout_end_date
+      )
+    end
+end
--- a/app/business/payments/payouts/payout_estimates.rb
+++ b/app/business/payments/payouts/payout_estimates.rb
@@ -1,6 +1,26 @@
 # frozen_string_literal: true
 
 module PayoutEstimates
+  # Fast set-based estimate of the unpaid balance Gumroad holds for Stripe
+  # payouts up to a date. Gumroad-held Stripe balances live on the platform
+  # merchant accounts (user_id IS NULL), so this is a single indexed
+  # aggregation rather than the per-user loop in estimate_held_amount_cents.
+  #
+  # This skips the per-user payability gates (compliance, paused, minimum
+  # payout), so it slightly over-states the amount that will actually pay out
+  # this cycle -- the safe direction for "is the balance high enough?".
+  def self.estimate_gumroad_held_stripe_cents(date)
+    merchant_account_ids = MerchantAccount.where(
+      user_id: nil,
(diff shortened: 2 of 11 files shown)


slavingia:
We don't use Slack. This should be email and we can have Gumclaw via telegram if we need that.

ershad:
> We don't use Slack. This should be email and we can have Gumclaw via telegram if we need that.

@slavingia Yeah. It was using email internally, but still mentioned `Slack`. We need to remove all Slack references from the codebase, will do in a new PR.

app/services/stripe_balance_check_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def topup_needed?
+    topup_amount_cents > 0
+  end
+
+  private
+    # Only the funds Gumroad itself holds need to come out of Gumroad's Stripe
+    # balance; balances held by Stripe are funded by Stripe directly.
+    def calculate_upcoming_payouts_cents
+      held_amount_cents = PayoutEstimates.estimate_held_amount_cents(
```
ershad: This is a slow query, debugging it.

PR #584, opened 2025-07-08
Fix Omani bank account number validation to allow valid IBANs

andev18:
This PR aims to fix https://github.com/antiwork/gumroad/issues/489

### Changes
- Upgrade `ibandit` to the latest version `1.26` (reason: the current 1.12 version does not support Oman IBAN validation, reference [here](https://github.com/gocardless/ibandit/blob/master/CHANGELOG.md#1220---january-22-2025))
- Update `oman_bank_account`'s `validate_account_number` to allow valids IBANs, aside from the existing account number regex validation
- Add unit tests to cover the valid and invalid IBANs case

### Notes
Devin opened a PR here https://github.com/antiwork/gumroad/pull/491 but it doesn't work, mainly due to the current `ibandit` library version cannot properly validate Omani IBANs, also that PR was closed
Screenshot 2025-07-08 at 7 14 18 PM

## Summary by CodeRabbit

* **New Features**
  * Enhanced account number validation to accept both traditional numeric formats and valid Omani IBANs.
* **Tests**
  * Added tests to ensure correct validation of Omani IBANs, including acceptance of valid IBANs and rejection of invalid ones.
* **Chores**
  * Updated the version constraint for the "ibandit" library to improve IBAN validation support.

PR #5455, opened 2026-06-11
Reject Omani IBANs at validation with an instructive error

ershad:
## What

Restores strict 6–16 digit validation for Omani payout account numbers and makes the rejection instructive. When a seller enters a valid Omani IBAN, the error now explains what to enter instead:

> Please enter your bank account number, not your IBAN.

Anything else that fails the format keeps the existing generic "The account number is invalid." The Oman account number input also gets numeric-only constraints (`pattern`, `inputMode`, `maxLength`) and a matching placeholder, following the existing Gibraltar special case in `BankAccountSection.tsx`.

## Why

Stripe's [bank account spec for Oman](https://docs.stripe.com/connect/payouts-bank-accounts) requires a SWIFT/BIC plus a 6–16 digit local account number — it rejects IBANs with `account_number_invalid`. This is deliberate on Stripe's side: neighboring UAE and Bahrain are speced as IBAN, Oman is not.

Since #584, `OmanBankAccount` validation accepts checksum-valid OM IBANs (well-intentioned — Omani banks hand customers their IBAN, so that's what sellers naturally enter). But the stored value is sent to Stripe verbatim by `bank_account_hash`, so an IBAN passes the form, the new `bank_accounts` row immediately becomes the active payout account, and the async Stripe sync (`HandleNewBankAccountWorker`) fails — leaving `stripe_bank_account_id` NULL. From then on every weekly payout skips at `stripe_payout_processor.rb:44` with "the payout bank account was not correctly set up", indefinitely and invisibly. Before #584 the seller at least got an immediate validation error; after it, the failure became silent. This was confirmed in production while investigating a seller report with exactly this signature (internal: antiwork/gumroad-private#485).

An earlier revision of this PR converted stored IBANs to the embedded local account number at Stripe sync time. We chose not to do that: Gumroad shouldn't infer payment data the seller never entered (the IBAN's account portion is zero-padded to 16 digits, which isn't guaranteed to be the bank's canonical account number). Instead, the seller stays the source of truth and the validation message tells them exactly what to provide — which also resolves the original concern in #489, since payouts demonstrably deliver to Omani banks when Stripe is given the local account number.

## Before/After

- Before: an Omani seller entering their IBAN gets a successful save, their working bank account is replaced, and payouts silently skip every week.
- After: the IBAN is rejected at save with an error explaining that the 6–16 digit account number is needed and where to find it; the account number field only accepts digits.

Video: pending — happy to add a staging walkthrough of the payout settings form if reviewers want one.

## Test Results

```
$ bundle exec rspec spec/models/oman_bank_account_spec.rb
9 examples, 0 failures
```

The new rejection example fails when the validation change is reverted (IBANs become valid again).

```
$ bundle exec rubocop app/models/oman_bank_account.rb spec/models/oman_bank_account_spec.rb

$ DISABLE_TYPE_CHECKED=1 npx eslint app/javascript/components/Settings/PaymentsPage/BankAccountSection.tsx
(clean)
```

## QA steps

1. As a seller with an Oman (OMR) payout country, open Settings → Payments and try to save bank details using the full IBAN (`OM…`) — the field rejects non-digit input, and a forced submit returns the instructive validation error.
2. Save with the plain 6–16 digit account number and the bank's SWIFT/BIC — confirm the bank account syncs (`bank_accounts.stripe_bank_account_id` is populated) and the payout note history shows no sync failure.

---

AI disclosure: investigated and written with Claude Code (Claude Fable 5).

---

> [!NOTE]
> **Medium Risk**
> Changes seller payout bank validation and form behavior for Oman; wrong messaging could block saves, but the change prevents silent Stripe sync failures from IBAN values.
> 
> **Overview**
> **Omani payout accounts** no longer accept IBANs as the account number. `OmanBankAccount` still requires a **6–16 digit** local number, but when the value parses as a valid **OM** IBAN it now fails with *"Please enter your bank account number, not your IBAN."* instead of saving. Other invalid formats keep the generic *"The account number is invalid."*
> 
> On **Settings → Payments**, sellers with country **OM** get Gibraltar-style constraints on the non-IBAN account field: numeric `inputMode`, `[0-9]{6,16}` pattern, 16-char max, example placeholder, and a `title` hint not to use IBAN.
> 
> Specs now expect valid Omani IBANs to be **rejected** with the new message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f29ae00a442180bad24cd2a0cb0e60f06647f65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/javascript/components/Settings/PaymentsPage/BankAccountSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/BankAccountSection.tsx
@@ -790,12 +790,21 @@ const BankAccountSection = ({
   ];
 
   const isGibraltar = user.country_code === "GI";
+  const isOman = user.country_code === "OM";
   const nonIbanAccountNumberInputProps: Pick<
     React.ComponentPropsWithoutRef<"input">,
-    "placeholder" | "maxLength" | "pattern" | "inputMode"
+    "placeholder" | "maxLength" | "pattern" | "inputMode" | "title"
   > = isGibraltar
     ? { placeholder: "01234567", maxLength: 8, pattern: "[0-9]{8}", inputMode: "numeric" }
-    : { placeholder: "1234567890" };
+    : isOman
+      ? {
+          placeholder: "000123456789",
+          maxLength: 16,
+          pattern: "[0-9]{6,16}",
+          inputMode: "numeric",
+          title: "Enter your 6 to 16 digit account number, not your IBAN",
+        }
+      : { placeholder: "1234567890" };
 
   const getRoutingNumberLabel = (countryCode: string) => {
     switch (true) {
--- a/app/models/oman_bank_account.rb
+++ b/app/models/oman_bank_account.rb
@@ -51,7 +51,10 @@ class OmanBankAccount < BankAccount
     def validate_account_number
       return if ACCOUNT_NUMBER_FORMAT_REGEX.match?(account_number_decrypted)
       iban = Ibandit::IBAN.new(account_number_decrypted)
-      return if iban.valid? && iban.country_code == "OM"
-      errors.add :base, "The account number is invalid."
+      if iban.valid? && iban.country_code == "OM"
+        errors.add :base, "Please enter your bank account number, not your IBAN."
+      else
+        errors.add :base, "The account number is invalid."
+      end
     end
 end
--- a/spec/models/oman_bank_account_spec.rb
+++ b/spec/models/oman_bank_account_spec.rb
@@ -35,11 +35,14 @@ describe OmanBankAccount do
   end
 
   describe "#validate_account_number" do
-    it "allows records that are valid Omani IBANs" do
+    it "rejects valid Omani IBANs with a message explaining what to enter instead" do
       allow(Rails.env).to receive(:production?).and_return(true)
 
-      expect(build(:oman_bank_account, account_number: "OM030001234567890123456")).to be_valid
-      expect(build(:oman_bank_account, account_number: "OM810180000001299123456")).to be_valid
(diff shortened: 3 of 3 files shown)


ershad:
Self-review notes (updated for the current revision):

- **Approach:** the first revision of this PR extracted the local account number from stored IBANs at Stripe sync time. That was the wrong direction — Gumroad would be inferring payment data the seller never entered, and the IBAN's zero-padded 16-character account portion isn't guaranteed to be the bank's canonical account number. The current revision rejects IBANs at validation and tells the seller what to enter, keeping the seller as the source of truth for their own account number.
- **Two error branches:** the instructive message only fires for checksum-valid Omani IBANs (`Ibandit::IBAN#valid?` + country check, the same detection #584 used to accept them); all other format failures keep the existing generic "The account number is invalid."
- **Frontend constraints are a convenience, not the enforcement point:** `pattern`/`inputMode`/`maxLength` on the Oman account number input follow the existing Gibraltar special case; the server-side validation remains authoritative per the contributing guide.
- **Production-gated validation is pre-existing:** `validate_account_number` runs only when `Rails.env.production?` (so dev/test can use fake numbers), which is why the specs stub it — unchanged from #584's tests.
- This intentionally restores pre-#584 strictness. The original #489 complaint ("my bank requires IBAN") is addressed by the message rather than by acceptance: production payouts demonstrably deliver to Omani banks when Stripe is given the local account number.

ershad:
@claude review once

ershad:
@claude review once

PR #152, opened 2025-04-24
Allow rejecting a review video

xrav3nz:
Thanks Laura for the updated design:

image

PR #5469, opened 2026-06-13
API v2: support minimum_amount_cents on offer code create

gumclaw:
## What

Make the **minimum-spend threshold** (`minimum_amount_cents`) reachable through the public offer-codes API.

- `POST /v2/products/:id/offer_codes` now reads `minimum_amount_cents` for both percent and cents discounts and passes it to `OfferCode.new`.
- `PUT` permits `minimum_amount_cents`.
- `as_json` / `as_json_for_api` include `minimum_amount_cents` when present, so the value round-trips and a client can confirm it actually persisted.

## Why

The `offer_codes` model and DB column already support `minimum_amount_cents` (added in `20230629014321_add_minimum_amount_cents_to_offer_codes`), and the web Discounts editor (`Checkout::DiscountsController`) sets it. But the **API v2** `OfferCodesController#create` built `OfferCode.new(...)` with a hardcoded attribute list that omitted it, and the API serializer never returned it.

The result was a silent-drop bug: an API client (or the CLI) could pass `minimum_amount_cents` and the value was discarded with no error — the request "succeeded" but produced a discount with no minimum. This is the root cause blocking gumroad-cli #152 ("spend over $100 → 10% off"). Exposing the existing model capability through the API is the minimal fix; no schema or model change needed.

I scoped this to the API surface only (controller + serializer) to mirror how `max_purchase_count` is already threaded, rather than touching the web editor or model, which already handle this field.

## Before / After

This is a non-user-facing (API) change. Walkthrough via the controller spec, which now covers the gap:

- **Before:** `minimum_amount_cents` passed to the API was silently dropped; not returned in the response.
- **After:** it persists for both percent and cents codes, stays `nil` when omitted, and is echoed back in the JSON.

## Test Results

`spec/controllers/api/v2/offer_codes_controller_spec.rb` — **34 examples, 0 failures** (4 new specs added for the threshold: cents code, percent code, nil-when-omitted, returned-in-response). Each new spec fails if the controller change is reverted (the value falls back to `nil`).

```
Finished in 6.99 seconds (files took 3.03 seconds to load)
34 examples, 0 failures
```

> Note: ran locally against the test DB with the healthy Redis port (`:6379` is wedged in this environment; `SIDEKIQ_REDIS_HOST=localhost:6380/11`). No screen recording — headless environment — so test output stands in for the walkthrough video.

---

🤖 AI disclosure: authored by Gumclaw (an autonomous agent operating the Gumroad/Antiwork stack) using **Claude Opus 4.8**. Prompt: "Do this: github.com/antiwork/gumroad-cli/issues/152" (add a CLI `--minimum-amount` flag for offer codes) — which on inspection required first unblocking the API here, since the field wasn't reachable through `/v2`.

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5469"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783949258&installation_id=134400930&pr_number=5469&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5469&signature=335cb3fce89db8a67e61dd3d809d92f082ab1d1487aa4479e08feb78c5666f8b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Low Risk**
> Narrow API surface change reusing an existing model column; no schema or checkout logic changes in this PR.
> 
> **Overview**
> API v2 offer codes now accept and return **`minimum_amount_cents`**, fixing a silent-drop bug where clients could send a minimum spend threshold but it was never persisted.
> 
> **Create** passes `minimum_amount_cents` into `OfferCode.new` for both cents and percent codes; **update** permits it alongside `max_purchase_count`. **`OfferCode#as_json` / `as_json_for_api`** include the field so responses round-trip. Public API docs and `OFFER_CODE_FIELDS` document the optional parameter; controller specs cover create, update, nil-when-omitted, and response shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af30d6a20da2ede5b849db9d63385f60cce8cfa0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/controllers/api/v2/offer_codes_controller.rb
+++ b/app/controllers/api/v2/offer_codes_controller.rb
@@ -17,12 +17,14 @@ class Api::V2::OfferCodesController < Api::V2::BaseController
       OfferCode.new(code: params[:name],
                     universal: params[:universal] == "true",
                     amount_percentage: params[:amount_off].to_i,
-                    max_purchase_count: params[:max_purchase_count])
+                    max_purchase_count: params[:max_purchase_count],
+                    minimum_amount_cents: params[:minimum_amount_cents].presence)
     else
       OfferCode.new(code: params[:name],
                     universal: params[:universal] == "true",
                     amount_cents: params[:amount_cents].presence || params[:amount_off].presence,
                     max_purchase_count: params[:max_purchase_count],
+                    minimum_amount_cents: params[:minimum_amount_cents].presence,
                     currency_type: @product.price_currency_type)
     end
 
@@ -59,7 +61,7 @@ class Api::V2::OfferCodesController < Api::V2::BaseController
 
   private
     def permitted_params
-      params.permit(:max_purchase_count)
+      params.permit(:max_purchase_count, :minimum_amount_cents)
     end
 
     def check_offer_code_params
--- a/app/javascript/components/ApiDocumentation/Endpoints/OfferCodes.tsx
+++ b/app/javascript/components/ApiDocumentation/Endpoints/OfferCodes.tsx
@@ -111,6 +111,10 @@ export const CreateOfferCode = () => (
       <ApiParameter name="amount_off" />
       <ApiParameter name="offer_type" description='(optional, "cents" or "percent") Default: "cents"' />
       <ApiParameter name="max_purchase_count" description="(optional)" />
+      <ApiParameter
+        name="minimum_amount_cents"
+        description="(optional) Minimum order total in cents required for the offer code to apply"
+      />
       <ApiParameter name="universal" description="(optional, true or false) Default: false" />
     </ApiParameters>
     <OfferCodeResponseFields />
@@ -135,6 +139,7 @@ export const CreateOfferCode = () => (
     "name": "1OFF",
     "amount_cents": 100,
     "max_purchase_count": null,
+    "minimum_amount_cents": null,
     "times_used": 1
   }
 }`}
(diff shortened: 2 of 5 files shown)


gumclaw:
@claude review once

gumclaw:
CLI side that consumes this: antiwork/gumroad-cli#156 (`offer-codes create --minimum-amount`). That flag depends on this PR landing — without it the API drops `minimum_amount_cents` silently.

gumclaw:
Thanks @greptile — both points addressed in af30d6a20.

**1. Serializer inconsistency (the conditional-presence guard).** Agreed. `minimum_amount_cents` is an optional *modifier* like `max_purchase_count`, not a discriminator like `amount_cents`/`percent_off`, so it should always be in the response shape. Both `as_json` and `as_json_for_api` now emit it unconditionally (`null` when unset), matching `max_purchase_count`. Dropped the `if minimum_amount_cents.present?` guard.

**2. Untested update path.** Added a spec asserting `PUT` with `minimum_amount_cents` persists it. `offer_codes_controller_spec` is now 35 examples, 0 failures.

Also updated the **public API docs** (rendered at gumroad.com/api from `components/ApiDocumentation`), which the original PR missed:
- `minimum_amount_cents` documented as a parameter on both create and update.
- Added to `OFFER_CODE_FIELDS` so it appears in every offer-code response block, and included (`null`) in the create example response.

@claude review once

PR #5483, opened 2026-06-16
Add ProductOwnershipTransferService for moving a product between sellers

ershad:
## What

Adds `ProductOwnershipTransferService` — a single entry point for moving a product (`Link`) from one seller account to another. A product isn't owned by one column: records are keyed on the product (`link_id`/`product_id`), on the owning user (`seller_id`/`user_id`), and some ownership is denormalized. An unguarded `Link.user_id` update therefore breaks the `seller_is_link_user` invariant and leaves stale aggregates behind. This service does the whole move atomically.

```ruby
ProductOwnershipTransferService.new(
  product:,
  new_seller:,
  performed_by: admin,        # optional; attributed in the audit comments
  move_subscriptions: false,  # opt-in; live subscriptions stay with the origin by default
).process
```

**Moves to the new seller (in one transaction):**
- `Link.user_id`
- The product's own profile-page sections (`SellerProfileSection` scoped by `product_id`) — their denormalized `seller_id` follows the new owner
- Product and variant `Installment`s and the `Workflow`s that own them (seller-wide posts stay)
- Offer codes that exist *only* for this product (universal and shared codes stay; the product is just detached from shared ones)
- `UtmLink`s that target the product

**Detached rather than re-homed across accounts:**
- Affiliate / collaborator / self-service-affiliate relationships are agreements with the *origin* seller, so the product is detached from them instead of silently moving them under the new seller. Destroying each `ProductAffiliate` reuses the model's existing callbacks, which clear the `is_collab` flag and reconcile the origin seller's `AudienceMember` entries.
- The product is removed from the origin seller's profile-level layout (`shown_products` lists and any "featured product" section).

**Stays with the origin account (forward-only, by design):**
- `Purchase`, `Refund`, `ScheduledPayout`, `Balance`, `SubscriptionEvent`, payout routing, and 1099/tax attribution — these followed the money and can't cross accounts/TINs.
- `Follower`s (per-creator, not per-product) and `ProductPageView` analytics.

**After commit:** the search index is refreshed for the new owner (owner fields aren't in the auto-reindexed set; cache busting already happens via `Link`'s `after_update` callback), and an audit comment is written to **both** accounts (`"… transferred to …"` on the origin, `"… received from …"` on the destination).

**Preconditions enforced before anything changes:** distinct seller, new seller in good standing (not deleted/suspended), and no `custom_permalink` collision in the new seller's namespace.

## Why

There's currently no supported way to move a product between accounts — only unsafe manual DB edits that break `seller_is_link_user`, leave `AudienceMember`/caches/search stale, and misattribute revenue. This centralizes the move into one auditable, transactional operation.

The key design decision is **forward-only**: historical revenue, payouts, and tax can't be reassigned across accounts, so purchases and balances stay with the origin. A consequence worth calling out for review: existing purchases keep `seller_id = origin` while `link.user` becomes the new seller, so the `seller_is_link_user` check (validated on `Purchase` at save time) no longer holds for those historical rows — future purchases of the moved product are attributed to the new seller and stay consistent. Affiliate relationships are detached rather than moved for the same reason: they're agreements with the origin seller, and recreating them under the new seller would fabricate consent that doesn't exist.

Related: antiwork/gumroad-private#522.

## Out of scope / follow-ups

- **URL continuity for the origin's subdomain.** The product keeps its permalink, so it still resolves at the apex and redirects to the new seller's subdomain. The old `<origin>.gumroad.com/l/<permalink>` URL still 404s, because subdomain lookups are scoped to the resolved seller and `LegacyPermalink` is only consulted when no seller is in context. A redirect for the origin's subdomain URL needs a routing change and is left as a follow-up.
- **Wiring.** This adds the service only; an admin action/console runbook to invoke it is a follow-up.
- **Authorization.** A real transfer is gated on the product owner's consent / proper authorization — a precondition handled outside this service.

## Before / After

Non-user-facing backend service; no UI change. Video pending per the contributing guide — I'll attach a short console walkthrough of a transfer (records repointed, comments written, purchases left in place) before marking ready for review.

## Test Results

New spec `spec/services/product_ownership_transfer_service_spec.rb` covers the ownership pointer and invariant, each repointed/detached record type, the audit comments, that purchases stay with the origin, cache busting + reindex, precondition failures, and transactional rollback.

```
21 examples, 0 failures
```

`bundle exec rubocop` clean on both files.

---

AI disclosure: this PR was written with Claude Opus 4.8.

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5483"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784208453&installation_id=134400930&pr_number=5483&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5483&signature=4c6a92a09ff6c028d69b09019b6827a0b5f29326301660187fb951e3169f7b85"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/app/services/product_ownership_transfer_service.rb
+++ b/app/services/product_ownership_transfer_service.rb
@@ -1,0 +1,205 @@
+# frozen_string_literal: true
+
+# Moves a product (Link) from one seller to another, repointing every record that
+# carries a denormalized owner reference while leaving historical/financial data with
+# the origin account. See the "Product ownership transfer between accounts" issue for
+# the full rationale.
+#
+# What moves:
+#   - Link.user_id -> new seller
+#   - Owner-keyed records that belong to the product: product/variant installments and
+#     their workflows, product-scoped offer codes, UTM links, and the product's spot in
+#     the origin seller's profile layout.
+#
+# What does NOT move (forward-only, by design):
+#   - Purchases, refunds, balances, payouts, subscription events, 1099/tax attribution.
+#     These stay with the origin account that actually received the money. As a result,
+#     historical purchases keep seller_id = origin while link.user is now the new seller;
+#     future purchases of the product are attributed to the new seller and stay
+#     consistent. Followers are per-creator, not per-product, and also stay.
+#
+# What is detached rather than moved:
+#   - Affiliate / collaborator / self-service-affiliate relationships are agreements
+#     between an affiliate and the origin seller. They are detached from the product
+#     instead of silently re-homed under the new seller, who can re-enroll the product.
+#
+# The repointing and the audit comments run in a single transaction so the move either
+# happens completely or not at all. Cache invalidation and search reindexing run after
+# commit.
+class ProductOwnershipTransferService
+  class TransferError < StandardError; end
+
+  def initialize(product:, new_seller:, performed_by: nil, move_subscriptions: false)
+    @product = product
+    @new_seller = new_seller
+    @previous_seller = product.user
+    @performed_by = performed_by
+    @move_subscriptions = move_subscriptions
+  end
+
+  def process
+    validate_preconditions!
+
+    ApplicationRecord.connection.stick_to_primary!
+    ApplicationRecord.transaction do
+      move_installments_and_workflows
+      move_offer_codes
+      move_utm_links
+      move_subscriptions! if @move_subscriptions
+      detach_affiliate_relationships
+      move_product_page_sections
+      remove_from_origin_profile
+      repoint_product
+      leave_audit_comments
+    end
+
+    rebuild_derived_data
+
+    @product.reload
+  end
+
+  private
+    attr_reader :product, :new_seller, :previous_seller, :performed_by
+
+    def validate_preconditions!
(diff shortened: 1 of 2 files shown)


ershad:
Closing for now, can revisit later.

PR #5500, opened 2026-06-17
Add Reset license uses button to the customer detail page

gumclaw:
## What

Resolves #5498. Adds a **Reset** button to reset a license's usage count from the customer detail (Sales) page, next to the existing Disable/Enable toggle. Resetting is gated behind a confirmation dialog.

This also closes a pre-existing cross-tenant authorization gap in the license-management endpoint — see below.

## Why

Today the only way to decrement/reset license uses is via a script, which is tedious for sellers administering seat-limited licenses. This surfaces it in the dashboard where the license already shows its key, uses, and enable/disable control.

### Security: authorize by scoped secure external ID

`LicensesController#update` — which also backs the existing Disable/Enable toggle — authorized only the seller's *role* against the current seller context; it never checked that the license belonged to that seller. Because the license was addressed by its plain external ID, and that ID legitimately circulates (it's returned by the license-verification API and embedded in delivered license keys), any authenticated seller could enable, disable, or reset **any** license, including ones on other sellers' products.

The fix mints a `manage_license`-scoped `SecureExternalId` token for the license in `CustomerPresenter` — which only ever runs for the current seller's own sales — so a valid token is only ever generated into the owning seller's dashboard payload. The controller resolves the license via `find_by_secure_external_id` with that scope. The plain external ID, a token minted for a different scope, and forged tokens all return 404.

## Changes

- **`License#reset_uses!`** — sets `uses` to `0`. The existing `after_commit` re-index on `uses` keeps the purchase search index consistent.
- **`License` includes `SecureExternalId`** with a `MANAGE_SECURE_ID_SCOPE` constant shared between minting and lookup.
- **`CustomerPresenter`** — sends the license's scoped secure external ID as its `id` instead of the plain external ID.
- **`LicensesController#update`** — resolves the license via the scoped secure external ID (404 otherwise) and handles a `reset_uses` param (mutually exclusive with the existing `enabled` toggle), leaving the license's enabled status untouched.
- **`resetLicenseUses`** data helper + **`onReset`** wiring in `LicenseSection`. The Reset button renders next to the Uses count, only appears when `uses > 0`, and opens a confirmation dialog before resetting.
- Model, controller, and presenter specs — including cases asserting the plain external ID, a wrong-scope token, and a forged token are each rejected with 404.

## Testing

Run locally against the test DB:

- `bundle exec rspec spec/models/license_spec.rb spec/controllers/licenses_controller_spec.rb spec/presenters/customer_presenter_spec.rb spec/presenters/customers_presenter_spec.rb spec/controllers/customers_controller_spec.rb` → all green.
- `bundle exec rubocop` on the touched Ruby files → no offenses.
- `npx eslint` on `CustomerDetailPage.tsx` → clean.

The end-to-end Capybara spec (`spec/requests/customers/customers_spec.rb` "allows management of licenses") exercises the real presenter → frontend → controller token flow and should be relied on in CI; it could not run in my local sandbox (browser host resolution). The frontend uses `license.id` only as an opaque route param, so the token switch is transparent to the UI.

## Screenshots

Captured from the live app (real DOM): booted this branch, seeded a license-bearing purchase with `uses: 7`, opened the customer detail page, and exercised Reset end-to-end. The button flips uses **7 → 0**, then hides itself (it only renders when `uses > 0`), with a "License uses reset" toast. Verified the side-effect in the DB (`License#uses == 0`), not just the toast.

> Note: the Reset button now opens a confirmation dialog ("Reset license uses?") before applying. Updated media showing the confirmation step is to be captured.

**Before** — Uses 7, Reset button shown next to the count:

before reset — uses 7

**After** — Uses 0, Reset button gone:

after reset — uses 0

Closes #5498

---

🤖 AI disclosure: implemented with the assistance of Claude Opus 4.8.

---

> [!NOTE]
> **High Risk**
> Changes authentication/authorization for license management (scoped tokens) and mutates license usage counts, which affects entitlement enforcement and cross-tenant access if misconfigured.
> 
> **Overview**
> Adds a **Reset** control on the customer detail license section so sellers can set usage back to **0** (with confirmation), wired through a new `reset_uses` path on `LicensesController#update` and `License#reset_uses!`.
> 
> **Security fix:** license enable/disable (and reset) no longer accept the plain license external ID. The dashboard now receives a `manage_license`-scoped secure external ID from `CustomerPresenter`, and the controller resolves licenses only via that scope—invalid, wrong-scope, or forged IDs return **404**.
> 
> The customer detail UI shows Reset only when `uses > 0` and updates local state after a successful reset; existing enable/disable behavior uses the same opaque `license.id` token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf28847a3b6823a2301b35340d95acb8cbde71c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/controllers/licenses_controller.rb
+++ b/app/controllers/licenses_controller.rb
@@ -5,7 +5,9 @@ class LicensesController < Sellers::BaseController
     license = License.find_by_external_id!(params[:id])
     authorize [:audience, license.purchase], :manage_license?
 
-    if ActiveModel::Type::Boolean.new.cast(params[:enabled])
+    if params.key?(:reset_uses)
+      license.reset_uses!
+    elsif ActiveModel::Type::Boolean.new.cast(params[:enabled])
       license.enable!
     else
       license.disable!
--- a/app/controllers/licenses_controller.rb
+++ b/app/controllers/licenses_controller.rb
@@ -2,7 +2,11 @@
 
 class LicensesController < Sellers::BaseController
   def update
-    license = License.find_by_external_id!(params[:id])
+    license = License.find_by_secure_external_id(params[:id], scope: License::MANAGE_SECURE_ID_SCOPE)
+    unless license
+      skip_authorization
+      return e404_json
+    end
     authorize [:audience, license.purchase], :manage_license?
 
     if params.key?(:reset_uses)
--- a/app/javascript/components/Audience/CustomerDetailPage.tsx
+++ b/app/javascript/components/Audience/CustomerDetailPage.tsx
@@ -30,6 +30,7 @@ import {
   resendPing,
   resendPost,
   resendReceipt,
+  resetLicenseUses,
   revokeAccess,
   undoRevokeAccess,
   updateCallUrl,
@@ -551,6 +552,18 @@ const CustomerDetailPage = ({
                   () => {
                     showAlert("Changes saved!", "success");
                     updateCustomer({ license: { ...license, enabled } });
+                  },
+                  (e: unknown) => {
+                    assertResponseError(e);
+                    showAlert(e.message, "error");
+                  },
+                )
+              }
+              onReset={() =>
+                resetLicenseUses(license.id).then(
+                  () => {
+                    showAlert("License uses reset", "success");
+                    updateCustomer({ license: { ...license, uses: 0 } });
                   },
                   (e: unknown) => {
                     assertResponseError(e);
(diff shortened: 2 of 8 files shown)


gumclaw:
@claude review once

jabza:
This sounds ideal though a decrement button would probably be preferred over a hard reset (or the ability to set the use count to any number, which covers all bases)

ershad:
@jabza thanks for the feedback! We have added the Reset button; adding decrement/count-inputs would complicate the UI/UX, I think. Please let us know if the Reset button isn't working for you.

PR #5482, opened 2026-06-16
Surface Stripe payout pause reasons and widen seller notifications

nyomanjyotisa:
## What

When Gumroad pauses a seller's payouts, this makes the pause explain itself on the admin Payout Info panel and notifies the seller — across Stripe-mirrored pauses and Gumroad's own system pauses.

- **Stripe pauses:** on a Stripe-source pause, write a `payouts_paused` comment with the raw Stripe `disabled_reason` + a human-readable description; write a `payouts_resumed` comment on re-enable. The comment is refreshed if Stripe's reason changes while the account stays paused. Writes are atomic with the pause-state update.
- **Reason surfacing:** `User#payouts_paused_for_reason` returns the reason for the *current* pause source — the `payouts_paused` comment for admin/Stripe pauses, and the relevant `on_probation` comment for **system** pauses (repeated failed payouts / high chargeback rate / failed-purchase spike), matched by their known authors. The admin panel renders this for every source instead of the old hardcoded strings.
- **Notifications:** widened beyond the previous two-reason allow-list — action-required email whenever Stripe has *currently-actionable* requirements (excluding future `eventually_due`), a new "under review" email otherwise, nothing for rejected/platform-paused. Sellers are re-notified once if a pause **escalates** to action-required, deduped so re-pauses and admin/payout-method toggles don't spam (`MerchantAccount#stripe_payouts_pause_email_sent`).
- **Backfill:** a one-time task (`Onetime::BackfillStripePayoutPauseComments`) writes the reason comment for accounts already paused by Stripe, reading the persisted `stripe_disabled_reason` (no Stripe calls), batched, idempotent. System pauses need no backfill (their `on_probation` comment already exists). Run: `Onetime::BackfillStripePayoutPauseComments.process`.

## Why

Stripe-mirrored pauses were the dominant cause of "payouts paused, no explanation" tickets: the pause flipped silently, the panel showed a fixed generic string, and the seller email fired for only two `disabled_reason` values. Both the seller and the support agent were left in the dark. This records the reason durably, surfaces it accurately on the panel for every pause source, widens and escalates notification, and backfills existing paused accounts.

---

AI disclosure: implemented with Claude Opus 4.8.

---

> [!NOTE]
> **Medium Risk**
> Changes payout pause/resume side effects on Stripe webhooks (locking, comments, email dedupe) and seller-facing notifications; incorrect logic could miss alerts or duplicate emails, but scope is bounded to Stripe-synced payout state.
> 
> **Overview**
[...]

PR #5515, opened 2026-06-18
Auto-retry Stripe payout setup for rejected postal codes and bank accounts

nyomanjyotisa:
## What

Adds an automatic, self-healing retry loop for sellers whose Stripe payout setup was rejected, covering the rejection points that can resolve on their own over time:

1. **Postal code rejected during account creation** (`create_account` → `postal_code_invalid`)
2. **Postal code rejected during an address/compliance update** (`update_account` → `postal_code_invalid`)
3. **Bank account sync rejected** (`update_bank_account` / `create_account` → invalid account or routing number, or card not supported). Account-holder-name mismatches are deliberately **excluded** — a name mismatch never self-heals by waiting, so the seller is still emailed immediately and no retry marker is recorded.

Concretely:

- A postal-code rejection marker mirroring the existing bank-sync one (`record_postal_code_failure_note` / `clear_stale_postal_code_failure_notes` on `StripeMerchantAccountManager`). It's recorded on `postal_code_invalid` and cleared on a later success.
- A `notify:` keyword (default `true`) threaded through `create_account`, `update_account`, `update_bank_account`, and `handle_new_user_compliance_info`. Automated retries pass `notify: false` so they don't re-send per-failure emails or pile up duplicate payout notes.
- A `force_address_resync:` keyword on `update_account` / `update_person` that re-sends the company and representative postal codes even when unchanged, so an automated retry actually re-validates the previously rejected postal code against Stripe instead of diffing it out.
- `RetryStripeRejectedPayoutSetupsJob` (**weekly** cron — Mondays UTC 07:00, before the daily payouts job — `:low`, `lock: :until_executed`) sweeps outstanding failure notes and fans out one `RetryStripeRejectedPayoutSetupForSellerJob` per affected seller.
- `RetryStripeRejectedPayoutSetupForSellerJob` routes by marker + state — bank note → `update_bank_account`; postal-code note → `update_account` (with `force_address_resync`) when the seller already has an alive Stripe account, otherwise `create_account` — all quietly. On success it clears the note and adds a resolved note; on continued failure it records the attempt and waits a week; after **8 weekly attempts (~8 weeks)** it abandons the note and sends a single marker-tailored `payout_setup_retry_exhausted` email (bank vs postal wording) — distinct from the generic `payouts_may_be_blocked`, which stays reserved for the outstanding-compliance-info reminder and would otherwise wrongly tell a seller whose details were correct that they need to provide more info. A seller who has switched to PayPal has the note abandoned without recreating an account or emailing.
- Seller-facing copy is reframed: the postal-code form error and the `invalid_bank_account` email now explain that a newly built address, or a recently opened account / added bank or branch, may simply not be in our payment partner's records yet — they tell the seller to double-check the value but otherwise do nothing, and note that we automatically re-check once a week for up to 8 weeks (the window is interpolated from `RetryStripeRejectedPayoutSetupsJob::RETRY_WINDOW_WEEKS` so the copy can't drift from the code).

All retry state lives on the failure note's `json_data` (`retry_count`, `last_retried_at`, `abandoned_at`, `abandoned_reason`). **No schema changes**, and nothing touches the `users` / `merchant_accounts` / `purchases` tables.

## Why

Stripe recommended automatically re-checking whether a previously rejected postal code / bank account would now be accepted. A rejection frequently isn't permanent — a newly built address can become resolvable, and Stripe keeps expanding the postal codes, bank-account types, and regions it supports. (A name mismatch is the exception — it only clears when the seller corrects it — which is why it stays out of the loop.)

Today these rejections are terminal: the seller hits an error and must manually re-submit. For account-creation failures it's worse — `cleanup_failed_merchant_account` soft-deletes the merchant account, and the weekly `StripeCreateMerchantAccountsWorker` then permanently skips that seller (`LEFT JOIN merchant_accounts ... WHERE merchant_accounts.id IS NULL`). The result is sellers wedged out of payouts, some holding a balance they can't withdraw.

This builds on the existing self-healing line of work (#5474, #5482, #5264) by reusing the self-clearing payout-note marker as the "this seller is currently rejected" signal, rather than introducing new tables or columns.

**Design note — give-up is keyed on attempt count, not raw note age.** Bank-sync failure notes already exist in production and are often older than the give-up window. An age-based cutoff would have fired a burst of give-up emails on the first run. Counting actual attempts — one per weekly sweep, up to 8 — achieves a "stop after ~8 weeks" window while still giving the existing backlog a fair set of retries. The sweep runs weekly, but the per-note due check uses a sub-week interval (~3 days) so clock drift between weekly runs can't make a note skip a week.

---

🤖 AI disclosure: implemented with Claude Code (Claude Opus 4.8).

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5515"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784364129&installation_id=134400930&pr_number=5515&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5515&signature=3611ccbad3c212501a798e8e28c82a412cd999c87bc5ef7781c432583d52d4fa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches Stripe Connect account creation/updates and payout eligibility signals; incorrect retry or note clearing could delay payouts or spam sellers, but behavior is gated by markers, notify flags, and extensive specs.
> 
> **Overview**
> Adds a **self-healing retry path** for Stripe payout setup failures that may resolve over time (invalid postal codes and bank/routing issues), while keeping account-holder name mismatches out of the loop.
> 
> **`StripeMerchantAccountManager`** now records standardized **payout-note markers** for postal-code and bank-sync failures, clears them on real success, and accepts **`notify:`** (suppress duplicate emails/notes on automated runs) and **`force_address_resync:`** so retries re-submit addresses instead of diffing them away. Bank failure classification is tightened so name errors are not treated as retryable.
> 
> **Weekly Sidekiq** (`RetryStripeRejectedPayoutSetupsJob` → per-seller job) scans those notes, re-attempts `create_account` / `update_bank_account` / compliance sync quietly, tracks **`retry_count`** on the note, abandons when sellers switch to PayPal or connect their own Stripe, and after **8 attempts** sends **`payout_setup_retry_exhausted`** (bank vs postal).
> 
> **Seller-facing copy** in payments errors, `invalid_bank_account` email, and a new exhausted-retry template explain weekly automatic re-checks for up to **`RETRY_WINDOW_WEEKS`** (8), without telling everyone to contact support immediately.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50fef9b9ec62364c9c910da5fb03373b684ddf11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


slavingia:
Awesome

gianfrancopiana (dismissed):
### P2 finding from Codex

**The retry can recreate a Stripe account for sellers who intentionally moved off Stripe.**

`attempt_remediation` keys off the lingering failure note, not the seller's *current* payout method. When `user.stripe_account` is nil it falls through to `create_account` for any seller in a supported country:

```ruby
else
  return false unless user.native_payouts_supported?
  return false if NEW_ACCOUNT_CREATION_BLOCKED_COUNTRIES.include?(...)
  StripeMerchantAccountManager.create_account(user, passphrase:, notify: false)
  true
end
```

Concrete path:
1. Stripe setup fails with `postal_code_invalid` → a "Stripe postal code rejected" payout note is recorded.
2. Seller switches to PayPal. `UpdatePayoutMethod#process_payment_address_params` deletes the Stripe account (`delete_charge_processor_account!`) and the bank account, but leaves the note untouched.
3. `user.stripe_account` is now nil, and `has_stripe_account_connected?` is false (it only covers Connect/OAuth, not PayPal), so the per-seller job doesn't skip them.
4. `create_account` doesn't require a bank account, so it succeeds — recreating a Gumroad-managed Stripe Custom account that the payout-method change had just torn down — and `resolve!` marks the note resolved.

So a seller who deliberately chose PayPal (after their balance was already forfeited on the payout-method change) silently gets a new Stripe account. The mirror case — a bank note with no active bank account — keeps failing until `MAX_RETRIES`, then emails `payouts_may_be_blocked` to a PayPal seller.

Suggest gating remediation on the current payout method: skip and/or abandon the note when the seller is no longer on a Stripe/bank-backed method (e.g. `user.payment_address` present), rather than relying solely on the breadcrumb note.

slavingia:
We generally don't allow people to move to PayPal from stripe.

app/sidekiq/retry_stripe_rejected_payout_setups_job.rb
```diff
@@ -0,0 +1,49 @@
...
+    def candidate_notes
+      Comment.alive
+             .with_type_payout_note
+             .where(commentable_type: "User", author_id: GUMROAD_ADMIN_ID)
+             .where(
+               "content LIKE ? OR content LIKE ?",
+               "#{StripeMerchantAccountManager::BANK_SYNC_FAILURE_NOTE_PREFIX}%",
+               "#{StripeMerchantAccountManager::POSTAL_CODE_FAILURE_NOTE_PREFIX}%"
+             )
+    end
```
ershad, line 48: This may not be scalable, and also prone to errors if we update the prefix. We should probably do a scheduled sidekiq job instead of looking at comments to find accounts to retry? Similar to how we do recurring charges?

gianfrancopiana (commented):
Did another Claude + Codex adversarial review:

### [P2] Business postal markers aren't cleared on a successful manual fix → false "couldn't verify" email

In `update_account` ([stripe_merchant_account_manager.rb:265](https://github.com/antiwork/gumroad/blob/jyotisa/stripe-payout-setup-auto-retry/app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb#L265)):

```ruby
if force_address_resync || (user_compliance_info.is_individual? && address_submitted?(diff_attributes, entity_key))
  clear_stale_postal_code_failure_notes(user)
end
```

The clear only runs for forced retries or **individual** accounts. When a **business** seller corrects their company/representative postal code through normal settings, both `Stripe::Account.update` and `update_person` succeed but the `"Stripe postal code rejected"` note stays alive. The sweep then keeps retrying a resolved code — and if the marker already hit `MAX_RETRIES`, the per-seller job takes the `give_up!` branch *before* attempting remediation and emails the seller "we still couldn't verify your postal code" even though they already fixed it.

Suggest clearing the postal marker after a successful business update when the company or representative address was submitted (after both account and person updates succeed), not just on `force_address_resync`.

### [P3] Sellers who connect their own Stripe account leave markers enqueued forever

In the per-seller job ([retry_stripe_rejected_payout_setup_for_seller_job.rb:15](https://github.com/antiwork/gumroad/blob/jyotisa/stripe-payout-setup-auto-retry/app/sidekiq/retry_stripe_rejected_payout_setup_for_seller_job.rb#L15)):

```ruby
return if user.nil? || user.suspended? || user.has_stripe_account_connected?
```

The PayPal branch just below abandons stale notes, but this early return doesn't. The note stays alive with `last_retried_at` blank, so `due_for_processing?` is always true and the weekly sweep re-enqueues a no-op job for that seller indefinitely (plus a lingering admin note). Worth giving connected-Stripe sellers the same `abandon_stale_notes!` treatment as the PayPal branch — or excluding them in `candidate_notes`.

### Minor

- `private_class_method` on its own line (no args) is a no-op, so the new helper methods are actually public. It's a pre-existing idiom in this file, but if privacy is intended, pass the method symbols explicitly.
- Abandoned notes are never deleted and stay `alive`, so every weekly sweep keeps re-scanning them (filtered out only in memory). Minor, but related to [P3].
- First-run load is the main operational risk: the existing backlog of bank-sync notes all becomes due on the first Monday run, with no jitter/rate-limiting on the Stripe calls. Worth watching that first run.

---

🤖 AI disclosure: reviewed with Claude Code (Claude Opus 4.8), using Codex (GPT-5.x) as a second-model review pass.

PR #5528, opened 2026-06-19
Keep installment plan discounts after the offer code is deleted

### LINKED ISSUE #1410: Offer code not applied to non-first installments when max usage limit is reached

gianfrancopiana:
Fixes #1410

## What

Installment plans let a buyer pay for a product in several scheduled charges. When a buyer used a discount code on a plan and the seller later deleted that code, every remaining charge quietly jumped to the full price. That over-charge would fail and shut the plan down, leaving the buyer locked out of something they'd already partly paid for.

This makes a plan keep the discount the buyer originally got, even after the code is deleted. The price is now taken from the discount saved on the buyer's first purchase instead of the live code, and that saved discount is used everywhere the price is worked out — including the check that stops a price from going below zero.

It also adds a one-time cleanup task that locks in the agreed price (total and number of charges) for older plans created before this protection existed, so their future charges can't drift if the seller later edits the product's price, the number of installments, or the code.

## Why

An earlier fix (#3066) saved the discount onto later charges but the price calculation still ignored it once the code was deleted, so the charge reverted to full price. Plans created after that fix already store a frozen price and were never affected — only older plans are.

Reading the saved discount in one shared place, rather than patching each calculation separately, keeps the discount amount, the per-charge split, and the zero-price check from ever disagreeing. Two follow-on bugs (a charge that could go negative, and an older fix that could wipe out a frozen price) came from exactly that kind of disagreement and are prevented by this approach.

## Evidence

- **Checked against production:** about 141 older plans had a deleted discount code and could still be charged; newer plans already store a frozen price and are unaffected. The originally reported plan recomputes to the exact price the buyer agreed to.
- **Tests added**, each of which fails if the fix is removed:
  - discount is still applied after the code is deleted
  - the price is floored at zero (a discount larger than the price can't make it negative)
  - a frozen price is kept even if the product price later drops below the discount
  - the cleanup task backfills only eligible plans, is safe to re-run, skips finished plans, and writes nothing in its default dry-run mode
- Full purchase and subscription test suites pass, and an automated code review found no remaining issues.

## How to test

The cleanup task is dry-run by default and reports how many plans it would update without changing anything; it only writes when explicitly told to.

---

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5528"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784497581&installation_id=134400930&pr_number=5528&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5528&signature=1c87beb7d2bd13b21c48370e20fc31d184936c5b156e9f5a2ef90ceeadc01345"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes core purchase minimum-price logic and subscription installment charging; mistakes could mis-price recurring charges, though scope is narrowed by tests and snapshot/backfill guards.
> 
> **Overview**
> Fixes installment plans reverting to full price when a seller deletes the discount code after checkout (#1410).
> 
> **Pricing** now resolves discounts through a shared `offer_code_for_pricing` path that still uses the cached `purchase_offer_code_discount` after the live code is soft-deleted, and applies that consistently for discount amount, PPP eligibility, and the zero floor. The **negative-price clamp** runs before installment splitting so a frozen snapshot per-charge amount is not replaced by a later live recompute that would clamp to $0 when the product price drops below the discount.
> 
> Adds **`Onetime::BackfillInstallmentPlanSnapshots`** to create installment plan snapshots for legacy subscriptions that had a cached discount but no snapshot (dry-run by default, idempotent). Specs cover deleted codes, snapshot authority, and the backfill job.
> 
> Presenter specs also assert **`was_paid`** on purchase info payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9dc718ac66fa3cf12f5959aec34f7d4f7fda97e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


slavingia (approved):

gumclaw:
## Adversarial review (Claude Opus 4.8, `--effort max`) — APPROVE WITH NITS

Ran a deep adversarial pass over the money path specifically hunting for a wrong-charge bug the bot reviewers missed. **Found none.** The three moving parts compose correctly and the backfill faithfully reconstructs the agreed total. Posting the findings that *are* worth acting on — none block merge.

### Why it's safe (the load-bearing trace)
Recurring installment charges don't hit the edited `else` branch directly — `Subscription#current_subscription_price_cents` short-circuits to `original_purchase.minimum_paid_price_cents` (`subscription.rb:210`), and the original purchase is `is_original_subscription_purchase?` so it's *not* a recurring charge and lands in the `else` block at `purchase.rb:1458`. So the edited block **is** the real per-charge path, and there: clamp-before-split zeroes the pre-split total → `calculate_installment_payment_price_cents(0)` ignores the arg when a snapshot exists → the deleted post-split `minimum_price = 0` override no longer clobbers it. Snapshot value survives both a deleted code and a dropped product price. Verified `agreed_total_cents` reconstructs the *exact* number the creation path snapshots (same cached per-unit base, identical `OfferCode` reconstruction, same rounding + quantity order).

### 🟠 Non-blocking — worth a note
1. **Backfill *raises* future charges on plans whose product price was later *lowered*.** It restores the agreed total from the cached per-unit price, so a plan currently charging `(lower_price − discount)/n` will charge the agreed (higher) amount post-backfill. Defensible ("freeze agreed terms"), but the PR is framed only around code *deletion* (charges revert *up*) — worth one line in the body / rollout note since a higher renewal carries some incremental decline risk.
2. **One residual snapshot-overwrite path exists:** the PPP-minimum clause (`purchase.rb:1480`) is gated on the stored `is_purchasing_power_parity_discounted` flag, not on whether PPP was actually applied — so a snapshot plan carrying *both* the PPP flag and an offer code whose recompute lands in `(0, min_price)` would get overwritten. **Pre-existing** (the PR only turned the `elsif`→`if`), and requires the exact PPP+offer-code combo the backfill itself asserts can't coexist. Flagging only because it's the honest exception to "snapshot is authoritative" — if that mutual-exclusivity ever weakens, this goes live.
3. **`offer_code_for_pricing` broadening is textually wider than installment plans** — it flips `include_deleted: true` for any purchase with a cached discount. Safe *today* only because the sole post-completion re-pricing entry point is `current_subscription_price_cents` and non-installment subs route through a different pricing source. A one-line comment ("only the installment original-purchase re-pricing path reaches this post-deletion") would lock in the intent so a future caller of `minimum_paid_price_cents` on a completed non-installment purchase doesn't silently inherit it.

### Test gaps (ranked by blast radius)
1. **Percent-discount reconstruction is untested.** Every backfill spec uses a fixed `amount_cents:` code. The `offer_code_is_percent` branch (`amount_percentage:` + `.round`) — the common case and the one most prone to a cents off-by-one — has zero coverage. Highest-value add.
2. **Multi-quantity backfill untested** (`* (quantity || 1)` with `quantity > 1`).
3. **Charge-*increase* direction untested** — specs cover freezing against a price *raise*, none assert the agreed price is restored *upward* after a price *drop* (the surprising direction, concern #1).
4. **Commission-completion negative clamp untested** — now floors at 0 where it previously could go negative; good fix, no guard test.

Full A–H trace (each hunt-list item cleared with a file:line one-liner) available on request. This is an honest "the code is correct" — the items above are real but none change a charge amount incorrectly.

<sub>🤖 Adversarial review by Claude Opus 4.8 (`claude-opus-4-8`), read-only, against HEAD `9dc718ac6`.</sub>

PR #5538, opened 2026-06-21
Count pending Stripe funds in the balance check

ershad:
## What

The daily Stripe balance check (`SendStripeBalanceCheckNotificationJob` → `StripeBalanceCheckService`) decides whether Gumroad needs to top up its platform Stripe balance to cover upcoming seller payouts. It compared the payout obligation against only `Stripe::Balance.available` — the immediately-transferable balance — and ignored `Stripe::Balance.pending`.

- Add `StripeTransferExternallyToGumroad.reachable_balances`, which sums the available and pending balance per currency, clamping pending at zero.
- The balance check now reads `reachable_balances["usd"]` instead of `available_balances["usd"]`.
- `available_balances` is left unchanged: `transfer_all_available_balances` can only move funds that are actually available, not pending.

## Why

Pending funds are recent sales settling on Stripe's rolling ~2-business-day schedule. Seller payouts run on a weekly cadence with a 7-day holdback, so funds pending today reliably become available before the payout runs. Comparing the obligation against the available balance alone over-reported the required top-up and produced false "top-up needed" alerts whenever a large settling balance was in flight (e.g. after a weekend of sales).

Illustrative: upcoming payouts need $300k, the available balance is $200k, and $150k is settling. The old check reports a $100k top-up, even though the real balance ($350k) comfortably covers the obligation. The new check reflects that no top-up is needed.

Pending is clamped at zero per currency because refunds, disputes, and reversals can push it negative, and subtracting that would understate funds that `available` alone already covers. This mirrors the "reachable" balance (`available + clamped pending`) that `StripePayoutProcessor` already uses when checking whether a connected account can cover a payout.

**Scope / non-goal:** this does not address the separate payout-timing risk. Seller payouts are Transfers that draw only from `available`, and Stripe's daily auto-sweep to the corporate bank drains `available`. If a sweep empties `available` right before a payout run while pending hasn't settled, a payout can still fail — that belongs in the Stripe payout-schedule/sweep timing, not this alert. A rigorous follow-up could count only pending whose `available_on` ≤ the payout date via the Balance Transactions API.

## Before / After

Backend-only change with no UI surface. Walkthrough video to be attached.

## Test Results

```
bundle exec rspec spec/services/stripe_balance_check_service_spec.rb \
  spec/sidekiq/send_stripe_balance_check_notification_job_spec.rb \
  spec/business/payments/transfers/stripe/stripe_transfer_externally_to_gumroad_spec.rb

23 examples, 0 failures
```

- The regression guard in `stripe_transfer_externally_to_gumroad_spec.rb` asserts `reachable_balances` returns available **+** pending and clamps negative pending. It fails if the check reverts to available-only or drops the clamp.
- `bundle exec rubocop`: 5 files inspected, no offenses detected.

---

AI disclosure: implemented with Claude Opus 4.8 (1M context).

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5538"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784674233&installation_id=134400930&pr_number=5538&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5538&signature=ea1eeb3c20493172d7ce921fbe23198c4feb5c62d9d8b1f975bf706857452d5a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes how payment ops interprets platform liquidity for alerts; actual transfers still use available-only, and counting all pending assumes settlement before payout (not timing-filtered by available_on).
> 
> **Overview**
> The proactive **Stripe balance top-up alert** now compares upcoming seller payouts against **available plus settling (pending) USD**, not available alone.
> 
> A new **`StripeTransferExternallyToGumroad.reachable_balances`** sums `Stripe::Balance` available and pending per currency, with **pending clamped at zero** so negative pending from refunds/disputes does not shrink the total. **`StripeBalanceCheckService`** (and its notification job specs) switch from `available_balances` to this reachable total. **`available_balances`** and bank sweeps via **`transfer_all_available_balances`** are unchanged—only immediately transferable funds are moved.
> 
> Specs cover aggregation, the negative-pending clamp, and the case where pending makes the platform balance look sufficient when available-only would have triggered a false alarm.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97b8d11d2341c019039801f4ff1f72a746c48823. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/business/payments/transfers/stripe/stripe_transfer_externally_to_gumroad.rb
+++ b/app/business/payments/transfers/stripe/stripe_transfer_externally_to_gumroad.rb
@@ -23,6 +23,26 @@ module StripeTransferExternallyToGumroad
     available_balances
   end
 
+  # Public: Returns a hash of currencies to the balance in cents that can fund
+  # upcoming payouts: the available balance plus settling (pending) funds.
+  # Pending sales settle within a couple of business days, well inside the
+  # weekly payout window, so they will be available before the payout runs.
+  # Pending is clamped at zero per currency because refunds, disputes, and
+  # reversals can push it negative, and subtracting that would understate the
+  # funds that `available` alone already covers. Mirrors the "reachable" balance
+  # StripePayoutProcessor uses to check a connected account can cover a payout.
+  def self.reachable_balances
+    balance = Stripe::Balance.retrieve
+    reachable_balances = Hash.new(0)
+    balance.available.each do |balance_for_currency|
+      reachable_balances[balance_for_currency["currency"]] += balance_for_currency["amount"]
+    end
+    balance.pending.each do |balance_for_currency|
+      reachable_balances[balance_for_currency["currency"]] += [balance_for_currency["amount"], 0].max
+    end
+    reachable_balances
+  end
+
   # Public: Transfers the amount from the Stripe master account to
   # the default bank account for the given currency.
   def self.transfer(currency, amount_cents)
--- a/app/services/stripe_balance_check_service.rb
+++ b/app/services/stripe_balance_check_service.rb
@@ -9,7 +9,11 @@
 class StripeBalanceCheckService
   def initialize
     @upcoming_payouts_cents = calculate_upcoming_payouts_cents
-    @current_balance_cents = StripeTransferExternallyToGumroad.available_balances["usd"].to_i
+    # Count settling (pending) funds alongside the available balance. Pending
+    # sales settle within a couple of business days -- well inside the weekly
+    # payout window -- so they fund the upcoming payouts. Using available-only
+    # over-reports the needed top-up and fires false alarms.
+    @current_balance_cents = StripeTransferExternallyToGumroad.reachable_balances["usd"].to_i
   end
 
   attr_reader :upcoming_payouts_cents, :current_balance_cents
--- a/spec/business/payments/transfers/stripe/stripe_transfer_externally_to_gumroad_spec.rb
+++ b/spec/business/payments/transfers/stripe/stripe_transfer_externally_to_gumroad_spec.rb
@@ -130,4 +130,27 @@ describe StripeTransferExternallyToGumroad, :vcr do
       end
     end
   end
+
+  describe "reachable_balances" do
+    before do
+      allow(Stripe::Balance).to receive(:retrieve).and_return(
+        Stripe::Balance.construct_from(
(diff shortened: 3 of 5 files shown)


ershad:
Self-review — notes for reviewers:

- **Why a new method instead of changing `available_balances`?** `transfer_all_available_balances` (the bank sweep) must only move funds Stripe will actually transfer, i.e. `available`. Pending can't be transferred. So `available_balances` keeps its exact meaning and `reachable_balances` is the new "can this cover an obligation?" reading. Same split that already exists in `StripePayoutProcessor` (`available_cents` for the transfer, `reachable_cents = available + clamped pending` for the coverage check).

- **Why clamp pending at zero?** On a per-currency basis pending can be negative when refunds/disputes/reversals outweigh inbound settling funds. Subtracting a negative pending from `available` would understate the balance and re-introduce false top-up alerts — the opposite of the fix. Clamping credits only incoming settlement, never debits it.

- **`Hash.new(0)`** gives a clean per-currency accumulator and makes "currency with no balance → 0" fall out naturally (covered by a spec).

- **Regression coverage:** `reachable_balances` spec asserts `usd → available + pending` and `cad → available only` (negative pending clamped). Both fail if someone reverts to available-only or drops the clamp. The service/job specs stub `reachable_balances`, so reverting the service to `available_balances` also fails (the real Stripe call would be hit, unstubbed).

- **Known non-goal (documented in the description):** this does not protect against the `available`-drain timing risk from Stripe's daily auto-sweep. Flagging explicitly so it isn't mistaken for a complete fix of payout-starvation.

PR #5561, opened 2026-06-23
Pause repeatedly-failing PayPal payouts and record a diagnostic for reason-less IPN failures

nyomanjyotisa:
Part of https://github.com/antiwork/gumroad-private/issues/564

## What

Stops PayPal payouts that can't be delivered from silently failing and looping forever, on the path where the data shows the failures actually happen.

- **IPN failures without a reason code now record a diagnostic.** When PayPal's IPN reports an item-level `Failed` with a blank `reason_code`, `handle_paypal_event_for_non_split_payment` recorded `failure_reason: nil`. That left the failure undiagnosable and kept it inside `RetryFailedPaypalPayoutsWorker`'s `failure_reason IS NULL` retry scope. It now records a generic `PAYPAL_PAYOUT_FAILED` diagnostic.
- **Repeated PayPal failures now pause payouts.** `pause_payouts_after_repeated_failures` returned early unless `bank_account_id` was present, so PayPal recipients were never paused — a non-receivable PayPal account was re-attempted every payout cycle indefinitely. It now also keys on `payment_address` for PayPal payouts and pauses after `MAX_CONSECUTIVE_FAILED_PAYOUTS` consecutive failures to the same account, mirroring the existing bank-account behaviour.

## Why

A production review of the failed/`null`-`txn_id` PayPal payments (last 45 days) showed the original premise of this PR was wrong:

- **MassPay is not atomic.** `completed` and `failed` payments coexist under the same `correlation_id` (e.g. `completed:118, failed:105`), which is only possible when `ACK=Success` and individual recipients fail later via IPN.
- **The volume is on the IPN path.** ~79% of stranded rows (827/1052) are `ACK=Success` + per-recipient IPN failures; only ~21% are whole-batch `ACK=Failure`, and most of that is a single batch. The reported seller's case is on the IPN path.
- Most IPN failures already carry a real reason code (e.g. `3148` non-receivable country, `3015` invalid email) and are correctly not retried; the gap was the reason-less ones and the missing pause.

The earlier recursive re-send approach is **dropped**: because MassPay is not atomic, re-sending the "healthy" subset after `ACK=Failure` rests on an unproven assumption (risking double payment on a money-movement path). That isolation idea is parked until PayPal's `ACK=Failure` semantics and `L_UNIQUEID` cross-call idempotency are confirmed. `perform_payments` is reverted to its original whole-batch behaviour, which keeps genuine `ACK=Failure` batches `failure_reason: nil` and therefore still retryable.

---

🤖 AI disclosure: drafted with Claude Opus 4.8

---

> [!NOTE]
> **Medium Risk**
> Touches payout failure recording and automatic payout pausing for PayPal sellers; incorrect scoping could pause accounts wrongly or miss bad addresses, but changes are narrow and well-covered by specs.
> 
> **Overview**
> Improves **PayPal payout failure handling** so failures are diagnosable and sellers with bad PayPal details get the same safety rails as bank payouts.
> 
> When a PayPal IPN reports **Failed** without a `reason_code`, the payment is now marked failed with **`Payment::FailureReason::PAYPAL_PAYOUT_FAILED`** instead of leaving `failure_reason` nil. Known codes still use **`PAYPAL <code>`**. A matching entry was added to **`PAYPAL_MASS_PAY`** for humanized messaging.
> 
> **`pause_payouts_after_repeated_failures`** no longer skips all PayPal payouts: it counts consecutive failed/returned payouts per **`payment_address`** (same processor), pauses the user at the existing limit, and updates the probation comment to mention the PayPal account. PayPal rows with no address still do not trigger a pause.
> 
> Specs cover IPN failures with and without reason codes, the generic failure reason on batch IPN paths, and PayPal-specific pause/scoping behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6c43fe04e704bb46f0ed99e6797a453b8a2a601. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/business/payments/payouts/processor/paypal/paypal_payout_processor.rb
+++ b/app/business/payments/payouts/processor/paypal/paypal_payout_processor.rb
@@ -165,10 +165,21 @@ class PaypalPayoutProcessor
     correlation_id = parsed_paypal_response["CORRELATIONID"]
 
     payments.each { |payment| payment.update(correlation_id:) }
-    payments.each(&:mark_failed!) unless transaction_succeeded
+    return if transaction_succeeded
 
-    errors = errors_for_parsed_paypal_response(parsed_paypal_response) if ack_status != "Success"
+    errors = errors_for_parsed_paypal_response(parsed_paypal_response)
     Rails.logger.info("Paypal Payouts: Payout errors for user IDs #{user_ids}: #{errors.inspect}") if errors.present?
+
+    if payments.size > 1
+      payments.each_slice((payments.size / 2.0).ceil) { |chunk| perform_payments(chunk) }
+    else
+      payments.first.mark_failed!(failure_reason_for_paypal_response(parsed_paypal_response))
+    end
+  end
+
+  def self.failure_reason_for_paypal_response(parsed_paypal_response)
+    error_code = parsed_paypal_response["L_ERRORCODE0"]
+    error_code.present? ? "PAYPAL #{error_code}" : "PAYPAL payout rejected"
   end
 
   # Public: Sends the money in `split_payment_by_cents(payment.user)` increments.
--- a/app/business/payments/payouts/processor/paypal/paypal_payout_processor.rb
+++ b/app/business/payments/payouts/processor/paypal/paypal_payout_processor.rb
@@ -170,16 +170,17 @@ class PaypalPayoutProcessor
     errors = errors_for_parsed_paypal_response(parsed_paypal_response)
     Rails.logger.info("Paypal Payouts: Payout errors for user IDs #{user_ids}: #{errors.inspect}") if errors.present?
 
+    error_code = parsed_paypal_response["L_ERRORCODE0"]
+    unless Payment.paypal_recipient_level_failure?(error_code)
+      payments.each(&:mark_failed!)
+      return
+    end
+
     if payments.size > 1
       payments.each_slice((payments.size / 2.0).ceil) { |chunk| perform_payments(chunk) }
     else
-      payments.first.mark_failed!(failure_reason_for_paypal_response(parsed_paypal_response))
+      payments.first.mark_failed!("PAYPAL #{error_code}")
     end
-  end
-
-  def self.failure_reason_for_paypal_response(parsed_paypal_response)
-    error_code = parsed_paypal_response["L_ERRORCODE0"]
-    error_code.present? ? "PAYPAL #{error_code}" : "PAYPAL payout rejected"
   end
 
   # Public: Sends the money in `split_payment_by_cents(payment.user)` increments.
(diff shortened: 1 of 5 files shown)


gumclaw:
**Verified the premise against production (read replica, last 45 days) before reviewing — and the data contradicts the core assumption this PR is built on.**

**TL;DR:** This fixes the minority path (~21% of stranded rows) and leaves the dominant one (~79%) untouched, while the recursive re-send rests on an atomicity assumption the data disproves.

### 1. MassPay is not atomic in production

The PR states PayPal validates MassPay atomically (one bad recipient → whole-batch `ACK=Failure`). But many `correlation_id`s have `completed` and `failed` payments **side by side under the same batch**:

```
0de618f47f630 => completed:118, failed:105, returned:10, cancelled:1
d2c9078595073 => completed:120, failed:112, cancelled:1
126e00a5c99b4 => completed:126, failed:102, unclaimed:5, cancelled:1
```

That's only possible if `ACK=Success` and the per-recipient failures arrive later via **IPN** (`handle_paypal_event_for_non_split_payment`), which this PR does not modify.

### 2. The volume is on the path the PR doesn't touch

Failed-null-`txn_id` rows in the last 45 days:

| Path | cids | failed rows | % |
|---|---:|---:|---:|
| MIXED / IPN (`ACK=Success`) | 11 | 827 | **79%** |
| PURE / `ACK=Failure` (what this PR fixes) | 7 | 225 | 21% |

### 3. The PURE bucket doesn't match the "one bad apple" model either

Of those 7 pure-failed batches: one is **219-of-219** (a sender/request-level failure — balance/limit/auth), the other six are **single-payment** batches. Recursive halving recovers nothing in either case, and on the 219-batch it fans out to ~219 individual re-sends.

### 4. Double-payment risk

The re-send re-POSTs with the same `L_UNIQUEID`s, justified as safe because `ACK=Failure` ⇒ nothing was paid. But since MassPay isn't atomic in practice, an `ACK=Failure` that follows partial transaction creation would make re-sending the "healthy" subset risk double payment. This assumption needs to be proven against PayPal's actual `ACK=Failure` semantics before shipping on a money-movement path.

### What actually fixes the bulk of stranded balances

The IPN handler: when PayPal sends item-level `Failed` with a blank `reason_code`, persist a generic diagnostic, and add a PayPal-keyed repeated-failure pause (the existing `pause_payouts_after_repeated_failures` guard only covers `bank_account_id`, so PayPal recipients are retried weekly forever). The `failure_reason` recording here is a reasonable improvement, but 770/1052 failed rows already carry a reason via the IPN path.

PR #5587, opened 2026-06-26
Add score-based reCAPTCHA path for checkout

ershad:
## What

Adds an alternative **score-based reCAPTCHA Enterprise** path for checkout, gated per logged-in buyer via the `recaptcha_score_checkout` Flipper feature.

- New `CheckoutRecaptcha` service is the single source of truth for which key/surface a checkout request uses. Both `CheckoutPresenter` (what the browser executes) and `OrdersController` (what the server verifies) read from it, so the executed key and the verified key always match.
- `useRecaptcha` gains a `scoreBased` mode: it skips widget rendering and invokes `grecaptcha.enterprise.execute(siteKey, { action })` programmatically. Script loading is now keyed per URL so the challenge (`render=explicit`) and score (`render=<siteKey>`) scripts can coexist.
- `ValidateRecaptcha` gets a `:checkout_score` surface with its own fail-open default and its own score threshold, independent of the existing `:checkout` challenge surface. The threshold **defaults to 0.5** so cohort buyers are score-gated out of the box. Thresholds are read from Redis (`RedisKey.recaptcha_score_threshold(surface)`) so they can be tuned at runtime without a deploy; the code default applies when the key is unset.
- **Trusted buyers get a more lenient bar.** Within the score cohort, buyers who are a compliant seller themselves *or* have a paid purchase from a currently-compliant seller made **at least 5 years ago** are routed through a `:checkout_score_trusted` surface with a **0.3 default** threshold. The site key is identical to the untrusted score surface — only the server-side threshold differs — so the frontend is unaffected.
- New `RECAPTCHA_MONEY_SCORE_SITE_KEY` env var (added to `.env.example`, `.env.production.example`, `.env.test`).

## Why

The current checkout uses a single reCAPTCHA challenge key. Its token only signals `tokenProperties.valid`, and Google may escalate it to an interactive image challenge — friction for legitimate buyers. A score-based key never renders a challenge; it returns a 0.0–1.0 risk score the server gates on.

The two key types are **not** interchangeable: Google ties a token to the key that produced it, so the key the browser runs must equal the key the server verifies against. Centralizing the decision in `CheckoutRecaptcha` keeps frontend and backend in lockstep, and the per-buyer Flipper gate lets us roll the score flow out to a cohort first. Anonymous buyers are never in the cohort, and the cohort falls back to the challenge key whenever the score key is unconfigured.

Because a score key marks nearly every valid token as `valid`, validity alone is no protection — gating has to be on the score. Hence the 0.5 default: a cohort buyer scoring below 0.5 is rejected from the moment they're added (no challenge fallback, since score keys never render one). Roll out to a small cohort first, watch the `[recaptcha_score] surface=checkout_score` score distribution, and dial the threshold in at runtime via `$redis.set(RedisKey.recaptcha_score_threshold(:checkout_score), "0.4")` — no deploy needed.

A flat 0.5 still risks rejecting legitimate buyers whose traffic scores low (older browsers, shared IPs — exactly the demographic in the originating report). So established accounts get a lower 0.3 bar: a buyer who is a compliant seller, or who paid a currently-compliant seller years ago, is almost certainly a real person, not a bot. Two qualifiers keep this hard to game: the seller must be **currently** compliant (a fraudster can't manufacture trust against their own non-compliant account), and the purchase must be **at least 5 years old** (otherwise a fraudster could buy one cheap product from a compliant seller and instantly inherit the lenient bar). Being a compliant seller oneself qualifies with no age requirement.

## Before / After

> ⚠️ Video pending. For this user-facing change, capture checkout in both cohorts (challenge vs. score key), desktop + mobile, light + dark mode, before recording is added to `qa-media/`.

## Test Results

All affected specs pass locally (after seeding `MerchantAccount` records, which the local test DB needs):

- `spec/services/checkout_recaptcha_spec.rb` + `spec/controllers/concerns/validate_recaptcha_spec.rb` — 36 examples, 0 failures (thresholds read from Redis; defaults applied when unset)
- `spec/presenters/checkout_presenter_spec.rb` — 42 examples, 0 failures
- `spec/controllers/orders_controller_spec.rb` (reCAPTCHA score-based cohort) — 2 examples, 0 failures

Rubocop and ESLint clean on all changed files.

> Screenshot of the local test run pending in `qa-media/`.

---

🤖 AI disclosure: drafted and verified with the assistance of Claude Opus 4.8.

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5587"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785098866&installation_id=134400930&pr_number=5587&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5587&signature=f57826946bfdfd64b955f9e12918f9c843f75f1f7b3a0ea81ae520493c43db91"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


slavingia:
We should consider a diff score for logged out vs in, so that logging in can be an antidote?

PR #5618, opened 2026-06-30
Record per-purchase Stripe checkout payment-flow analytics

nyomanjyotisa:
Part of https://github.com/antiwork/gumroad/issues/5362

## What

Adds a `purchase_payment_flows` side table that records how a Stripe checkout purchase was paid, written once when Rails creates a charged checkout purchase. Three dimensions are captured:

- **`payment_details_source`** — `card_element`, `payment_element`, `payment_request`, or `saved_payment_method`.
- **`payment_details_transport`** — derived server-side from the submitted params (`payment_method` today; `confirmation_token` reserved for the later server-finalized Intent path).
- **`stripe_payment_method_type`** — `card`, the only enabled type today.

The row is created in `Purchase::CreateService` the moment the purchase is persisted (right after `prepare_for_charge!`), before the payment surface is normalized into the charge params — so it is written once the purchase exists and survives a later charge failure. The backend charge path is unchanged.

CardElement and Payment Element submit an identical `pm_` card payload, so the active surface is reported from the client (where it is known) and threaded through `OrdersController` → `Order::CreateService` → `Purchase::CreateService`. Wallet payments are forced to `payment_request` server-side from `wallet_type`; PayPal/Braintree and free purchases fall outside these Stripe dimensions and record nothing.

## Why

The Stripe Payment Element rollout needs a durable, purchase-level signal to compare the new surface against the legacy CardElement (conversion, SCA, errors, disputes) and to back the eventual server-side eligibility resolver's decision logging. This data must be durable and queryable per purchase.

A new table is used deliberately so the large `purchases` table is never altered (no new columns or indexes), consistent with the repo's constraints. There are no DB-level foreign keys, and the `has_one` is `validate: false` so analytics recording can never block a real purchase.

`payment_details_transport` is derived server-side and never trusted from the client, since it gates backend behavior in later phases. `payment_details_source` is taken from the client surface because CardElement vs Payment Element is indistinguishable server-side — both submit the same card payload.

---

AI disclosure: implemented with Claude Opus 4.8.

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5618"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785376284&installation_id=134400930&pr_number=5618&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5618&signature=8f6b18a02399b1d8a0c2afac9de1afc547c58312861769db5b275d79138668f7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


app/models/purchase_payment_flow.rb
```diff
@@ -0,0 +1,45 @@
...
+  belongs_to :purchase
+
+  CARD_ELEMENT = "card_element"
+  PAYMENT_ELEMENT = "payment_element"
+  PAYMENT_REQUEST = "payment_request"
+  SAVED_PAYMENT_METHOD = "saved_payment_method"
+  PAYMENT_DETAILS_SOURCES = [CARD_ELEMENT, PAYMENT_ELEMENT, PAYMENT_REQUEST, SAVED_PAYMENT_METHOD].freeze
+
+  PAYMENT_METHOD = "payment_method"
+  CONFIRMATION_TOKEN = "confirmation_token"
```
rmarescu: Can `confirmation_token` transport branch from this PR be removed?

`stripe_confirmation_token_id` is not permitted or threaded through the current checkout path, so this branch and its test are future-only. Based on @xrav3nz's research, we may not use ConfirmationTokens after all. I’d keep this PR to the `payment_method` path it actually ships.

app/models/purchase_payment_flow.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class PurchasePaymentFlow < ApplicationRecord
+  belongs_to :purchase
+
+  CARD_ELEMENT = "card_element"
+  PAYMENT_ELEMENT = "payment_element"
```
rmarescu, line 7: These overlap with the checkout integration strings in `Checkout::StripePaymentPresenter`.

I don’t think this model should necessarily depend on that presenter, since these are persisted analytics dimensions, but can we make the ownership explicit? Either centralize the shared values somewhere payment-flow specific, or leave these here with the understanding that they are durable stored values and should stay stable.

app/models/purchase_payment_flow.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  PAYMENT_METHOD = "payment_method"
+  CONFIRMATION_TOKEN = "confirmation_token"
+  PAYMENT_DETAILS_TRANSPORTS = [PAYMENT_METHOD, CONFIRMATION_TOKEN].freeze
+
+  CARD = "card"
+
+  validates :payment_details_source, inclusion: { in: PAYMENT_DETAILS_SOURCES }
+  validates :payment_details_transport, inclusion: { in: PAYMENT_DETAILS_TRANSPORTS }
+  validates :stripe_payment_method_type, presence: true
```
rmarescu: Could these string domains use Rails string-backed [`enum`](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) with validation instead?

For example, `payment_details_source` and `payment_details_transport` are closed sets and the app already has precedent for string enums with `validate: true`. If the generated predicate/scope methods would be noisy here, the current explicit validations are fine, but enum would make the allowed values and query API more standard.

app/models/purchase_payment_flow.rb
nyomanjyotisa: Removed the confirmation_token branch and its test for now

app/models/purchase_payment_flow.rb
nyomanjyotisa: Switched both source and transport to string-backed enums

app/models/purchase_payment_flow.rb
nyomanjyotisa: Went with keeping them in the model. They get stored on every row, so I didn't want them depending on the presenter. If those integration strings ever get renamed, the saved rows would be wrong. Added a short comment so it's clear they're meant to stay stable.

PR #5645, opened 2026-07-01
Fix Agent page mobile layout and iOS composer visibility

gianfrancopiana:
## What

Two mobile bugs on the Agent page (`/agent`):

- **Gap under the mobile nav.** On phones the page's `PageHeader` hides its title and has no actions, so it rendered as an empty padded bar with a bottom divider — dead space between the "Dashboard" nav and the greeting. It's now hidden below `sm` and shown from `sm` up, where the title is actually visible.
- **Composer hidden on iOS Safari.** The input sat below the fold on load and needed a scroll to reach — unlike chatgpt.com. The app shell used `h-screen` (`100vh`), which on iOS Safari is the *large* viewport (measured as if the toolbar were retracted), so the bottom of the layout fell behind the toolbar. The shell now sizes to the dynamic viewport (`h-dvh`), and the Agent page fills its parent (`h-full`) instead of re-asserting `100vh` a second time inside `<main>`.

Also bumped the send button from 40px to 44px to meet the minimum touch target.

## Why

`100vh` on iOS Safari doesn't subtract the browser toolbar — the classic reason a bottom-anchored input hides until you scroll. `h-dvh` tracks the visible viewport, keeping the composer on screen and matching the reference behavior on chatgpt.com. Since `<main>` already scrolls internally, sizing the shell to `dvh` also stops the body from scrolling, so the iOS toolbar no longer jumps.

The header gap came from the shared `PageHeader` contract: below `sm` the title is hidden by design, but the header element still renders its padding and border. With no actions to show, that's pure empty space — so the whole header is hidden where it would otherwise be empty.

## Screenshots

**Mobile**

**Desktop**

---


gumclaw:
⛔ **Blocking: missing before/after media (CONTRIBUTING.md line 17 — a hard `Must`).**

This is a mobile-UI/CSS PR — the gap under the nav, the iOS composer visibility, and the 40px→44px touch target are all *visual* changes. Per CONTRIBUTING.md, a product PR like this **must** include before/after evidence (video preferred, screenshots acceptable) showing **mobile + desktop, light + dark**. Right now there's none, so the fix can't actually be verified from the PR — the prose describes the intent but nothing proves the composer is now on-screen on iOS Safari or that the nav gap is gone.

Could you add:
- Before/after of the Agent page on **mobile** (the composer-visibility fix + the nav-gap fix), ideally a short screen recording on iOS Safari since that's the specific bug
- Before/after on **desktop** to confirm no regression
- Light + dark for each

Store under `qa-media/pr-5645-*.png|mp4` and reference via raw GitHub URL, per the convention.

(Context: this PR is the concrete example behind #5649, which makes before/after media the #1, non-skippable requirement for product PRs — the `pr-description` skill previously let the Before/After section be deleted for "minor" UI changes, which is exactly what happened here.)

— Gumclaw

PR #5656, opened 2026-07-01
Make Stripe PaymentIntent webhooks the source of truth for Payment Element checkout

nyomanjyotisa:
Part of https://github.com/antiwork/gumroad/issues/5362 and https://github.com/antiwork/gumroad/issues/5640

## What

Makes the PaymentIntent webhook the source of truth for client-confirmed (Lane B) checkout fulfillment. Today a Lane B purchase is finalized only by the buyer's browser (inline AJAX after `stripe.confirmPayment`); if the browser never returns, the order can stay charged-but-unfulfilled. This adds webhook-driven finalization so the purchase completes regardless, exactly once.

- Handle `payment_intent.succeeded` and `payment_intent.processing` on the platform account (`StripeEventHandler` + `StripeChargeProcessor#handle_stripe_charge_event`), mirroring the existing `payment_intent.payment_failed` branch.
- Scope them to client-confirmed charges via a new `Charge#client_confirmed` flag (on the existing `flags` bitmask — no migration). A `stripe_payment_intent_id` alone is also set by server-confirmed charges (combined server-confirm, off-session recurring), so it can't distinguish the lane; the flag can.
- `payment_intent.succeeded` finalizes through the existing `Order::FinalizeConfirmedChargeService` / `Purchase::FinalizeConfirmedChargeService`. `payment_intent.processing` keeps the purchase `in_progress` (reuses `stripe_status`, no new state). `charge.succeeded` behavior is unchanged.
- New `ProcessedStripeEvent` table (unique index on `event_id`) so a re-delivered webhook is a no-op. Exactly-once fulfillment is ultimately guaranteed by the finalize service's `with_lock` + `in_progress -> successful` transition.
- `FailAbandonedPurchaseWorker` keys on the `client_confirmed` flag when self-healing a succeeded-but-unfinalized intent.

With the flag off (and for the server-confirm path, subscriptions, off-session recurring, and other sellers), behavior is unchanged: the new branch resolves the charge by primary key and returns early for anything that isn't a client-confirmed charge.

## Why

Stripe delivers webhooks at-least-once, and the browser return is only a present-buyer optimization, not authority. Making the `payment_intent.*` webhook the source of truth closes the browser-disappearance gap without touching the server-confirm path. Connected-account (direct-charge) routing is intentionally deferred — this path is platform destination charges for now.

## Deploy prerequisite

Verified against the live Stripe account (read-only API check, 2026-07-02): neither active webhook endpoint currently subscribes to any `payment_intent.*` event, so this PR is inert in production until the platform endpoint (`gumroad.com/stripe-webhook`) adds `payment_intent.succeeded` and `payment_intent.processing` (dashboard edit; the signing secret does not change). The connect endpoint should not get these events yet — connected-account intents from existing direct charges would flood it with deliveries this handler ignores by design.

## After (No UI changes, it just purchase works with webhook)

https://github.com/user-attachments/assets/af2ca187-26ba-42a7-8ab4-9b382ecb6d16

---

---
<a href="https://app.blacksmith.sh/antiwork/codesmith/gumroad/pr/5656"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785522598&installation_id=134400930&pr_number=5656&repository=antiwork%2Fgumroad&return_to=https%3A%2F%2Fgithub.com%2Fantiwork%2Fgumroad%2Fpull%2F5656&signature=73053ba206ea51a27e804cc89b3f6da644653601de496fad5ae99d22817821f3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


gumclaw (commented):
## Review — Payment Element webhook lifecycle

Reviewed the full non-test surface (event routing, charge resolution, idempotency, receipt path). This is a well-architected change and the exactly-once claims hold up under scrutiny. CI green, Greptile's one P2 already addressed. **Verdict: looks good to me** — comments below are confirmations + minor nits, nothing blocking.

### What I verified holds

- **Lane isolation via `client_confirmed` flag (bit 2).** `stripe_payment_intent_id` alone can't distinguish Lane B from server-confirm / off-session recurring, so gating on the flag is the right call. The new branch resolves the chargeable and `return unless chargeable.is_a?(Charge) && chargeable.client_confirmed?` — so subscriptions, server-confirm, and other sellers fall through untouched. Confirmed the flag is set exactly at `PreparePaymentIntentService#build_charge`.
- **Charge resolution path is correct.** The PI's `transfer_group` is set to `CH-<external_id>`, so `get_charge_reference` returns the `CH-` prefixed reference → `find_by_stripe_event` takes the `Charge` branch (not the Purchase branch). The `is_a?(Charge)` guard is therefore reachable in production, not dead.
- **Exactly-once fulfillment.** Three independent layers: (1) `ProcessedStripeEvent` unique index on `event_id` with `record!` rescuing `RecordNotUnique`; (2) `record!` is called **after** `ChargeProcessor.handle_event` so a raise mid-fulfillment leaves the event unrecorded and Stripe redelivers; (3) the real guarantee — `Purchase::FinalizeConfirmedChargeService#perform` runs under `purchase.with_lock` and short-circuits on `successful?`. So even concurrent browser-AJAX + webhook deliveries fulfill once.
- **Receipts don't double-send.** `send_charge_receipts` moved from `OrdersController#finalize` into `FinalizeConfirmedChargeService#perform` (now shared by both triggers). Safe because `SendChargeReceiptJob` is idempotent three ways: `return if charge.receipt_sent?`, `lock: :until_executed`, and `with_lock { … update!(receipt_sent: true) }`.
- **`processing` keeps `in_progress`.** `handle_event_processing!` is informational-only (fee update), and the late-`processing`-after-`succeeded` case is covered by a spec + can't regress because the finalize service only writes `stripe_status: PROCESSING` when the purchase is still `in_progress`.

### Nits (non-blocking)

1. **`handle_stripe_charge_event` is getting long.** The `payment_intent.*` lifecycle branch (resolve → gate on `client_confirmed?` → dedupe → record) is now inline alongside the older charge/dispute branches. A small private `handle_payment_intent_lifecycle_event(stripe_event)` extraction would keep the money-path readable, but that's stylistic.

2. **`raise "…does not contain a 'payment_intent' object."` on shape mismatch** will bubble as a job failure and force Stripe redelivery. That's the safe direction (fail loud, retry) — just flagging that a malformed/unexpected event will retry 5× before dead-lettering. Fine as-is given the guarantee is downstream.

### Deploy gate — important, don't lose it

The PR is inert until the **platform** endpoint (`gumroad.com/stripe-webhook`) subscribes to `payment_intent.succeeded` + `payment_intent.processing` in the Stripe Dashboard, and the **connect** endpoint must **not** get them (connected-account direct-charge intents would flood a handler that ignores them by design). Worth pinning this as a post-merge checklist item on the tracking issue so the dashboard edit isn't forgotten.

gumclaw:
## Status: webhook prerequisite complete — ready to ship

### ✅ Done
- **Platform Stripe webhook endpoint configured (production).** `gumroad.com/stripe-webhook` (`we_0QA6vo9e1RjUNIyYyMt9MzkW`) now subscribes to `payment_intent.succeeded` and `payment_intent.processing` — the deploy prerequisite this PR called out. Verified live: 22 → 24 events, `api_version` unchanged (`2023-10-16`, signing secret **not** rotated, no app-config change), status still `enabled`.
- **Connect endpoint left untouched, by design.** `gumroad.com/stripe-connect-webhook` still has 0 `payment_intent.*` events. Direct-charge sellers are gated out of the client-confirm (Lane B) lane today by `Checkout::StripePaymentPresenter#client_confirm_eligible?` (`sellers.none?(&:stripe_connect_account)`), so no connected-account Lane B intent can be minted in production — the connect endpoint receiving these events would be pure noise the handler ignores.

### ⏳ What's left before it's live in prod
1. **Human approval + merge** — CI is all-green and `mergeStateStatus: CLEAN`, but the main ruleset requires the reviewer 1-tap. Auto-merge (squash) is armed, so it merges the moment approval lands.
2. **Prod deploy** — on the next deploy after merge, two changes ride along automatically:
   - migration `create_processed_stripe_events` (dedup table + unique index on `event_id`),
   - the `DeleteOldProcessedStripeEventsJob` Sidekiq schedule entry (registers on boot).
   No manual steps.

### 🔁 How the interim window is handled
Between now and deploy, the two events are already being delivered to the platform endpoint, but production code has no branch for them yet, so `StripeEventHandler` logs an `Unhandled event …` ERROR per Payment Element checkout. This is a **harmless no-op** (no crash, no mis-fulfillment) but produces Sentry/log noise. The window closes the instant this deploys — so the sooner it deploys, the less noise.

### ➡️ Follow-up (tracked on #5673)
Connect-side Lane B webhook finalization is a **blocking prerequisite of #5673** (the PR that flips the direct-charge gate and thus *creates* the direct-charge exposure). A `⚠️` checklist is encoded in #5673's description: route `payment_intent.succeeded`/`.processing` in `handle_event_for_connected_account`, add cross-account chargeable resolution + tests, and subscribe the connect endpoint (higher volume). Backstop and exposure ship in the same breath — connect coverage lands **before or with** #5673, never before it's needed.

PR #5673, opened 2026-07-02
Support direct-charge sellers on client-confirm checkout

nyomanjyotisa:
> **Deploy note:** after this merges + deploys (and before enabling `stripe_payment_element_client_confirm` for direct-charge sellers), the connect webhook endpoint (`gumroad.com/stripe-connect-webhook`) must be subscribed to `payment_intent.succeeded` / `payment_intent.processing` — full merged `enabled_events` list, do not resend `api_version`. This is automated: a watcher checks every 30 min and applies + verifies the change once the deploy lands.

## What

Enables bring-your-own-Stripe (direct-charge) sellers on the client-confirm Payment Element checkout path, which previously kept them on the server-confirm lane. The payment method resolver now emits the connected-account Elements scope instead of a direct-charge fallback reason, the checkout presenter passes it to the browser, Stripe.js/Elements load with `stripeAccount` so the ConfirmationToken is minted on the connected account, and the prepare step retrieves that token (and runs the purchasing-power-parity card-country check) on the same account. PaymentIntent creation, finalize, and the abandonment worker were already account-aware, so they needed no changes.

Also fixes a latent prepare bug this path made reachable: when merchant-account resolution rejects a cart (a Brazilian Stripe Connect account with an affiliated purchase), the validation error was added and then silently cleared by `save!` revalidation, so the charge proceeded anyway. Prepare now fails the whole order in that case.

## Why

The account a ConfirmationToken is minted on must match the account the PaymentIntent is created on, or Stripe rejects the confirm handshake. The Elements scope is derived from `User#has_stripe_account_connected?` — the same predicate `User#merchant_account` uses to route the charge — so the scope and the routing cannot diverge by construction:

- Linked connected account → Elements, token, intent, and finalize all on the connected account, with the platform's application fee.
- Connected account present but not linked for charging → platform scope, matching its destination-charge routing.
- Connected account with no merchant id (legacy data drift) → falls back to the server-confirm lane up front instead of failing mid-checkout.

This PR is stacked on the resolver branch (`feat/payment-element-client-confirm-method-resolver`) as its base, since it extends the resolver's output; it should retarget to `main` once that branch merges.

## QA steps

1. Enable the base and client-confirm checkout flags for a seller whose Stripe Connect account is linked for charging (merchant migration on), and once more for a plain platform seller.
2. Open the connect seller's product checkout. The server-rendered page props should show `integration: "payment_element_client_confirm"` with `elements_options.stripe_connect_account_id` set to the seller's connected account id, and the Payment Element should mount.
3. Complete a card purchase with the standard test card. The purchase should succeed, its merchant account should be the seller's connected account, and the payment should appear in the connected account's Stripe test dashboard carrying the platform application fee.
4. Repeat on the platform seller's checkout: `stripe_connect_account_id` should be null and behavior should be byte-for-byte unchanged.
5. Optional negative check: with the connect seller, complete a purchase using a declining test card and confirm the decline surfaces inline with no charge created on either account.

## Test results

- Resolver, presenter, and prepare-service specs: 62 examples, 0 failures (TDD — each new spec was watched failing first).
- Client-confirm browser E2E file: 5 examples, 0 failures against Stripe test mode, including a new end-to-end direct-charge purchase asserting the checkout props carry the connected-account scope and the resulting charge lands on the connected merchant account.
- vitest: 88 tests, 0 failures, including new coverage for the account-scoped Stripe.js loader.
- rubocop and eslint clean; tsc clean apart from a known pre-existing environment artifact.

# After

dc-1-loaded
dc-2-filled
dc-3-result
dcfail-3-result
dcsca-challenge

https://github.com/user-attachments/assets/e0e74ff7-eac6-47ff-b819-cd4b259d67bc

https://github.com/user-attachments/assets/3b89aa25-77db-4683-9200-92a3d1f1080e

https://github.com/user-attachments/assets/f28b859f-2b42-4bab-bef0-34b5692a324d

https://github.com/user-attachments/assets/46290f3b-e25e-4ce9-8661-6c12e854316e

---

AI disclosure: implemented with Claude Code (Claude Fable 5).


gumclaw (approved):
## Review — Approve

Traced the whole client-confirm to connected-account path through the code. The central safety claim (Elements scope cannot diverge from charge routing) is true **by construction**, not just by test:

- `PaymentMethodResolver#direct_charge_seller?` gates on `User#has_stripe_account_connected?` (feature_status.rb:47).
- `User#merchant_account` (user.rb:495-502) routes the charge to `stripe_connect_account` on the **same** `has_stripe_account_connected?` predicate.

So the account the ConfirmationToken is minted on (Elements `stripeAccount`) and the account the PaymentIntent is created on always come from one source of truth — the confirm handshake cannot fail on an account mismatch. The three-way split is correct:

| Seller state | `has_stripe_account_connected?` | Elements scope | Charge routing |
|---|---|---|---|
| Linked connected acct | true | connected | connected — match |
| Connect acct, not linked | false | platform | platform — match |
| Connect acct, no merchant id | true but id blank → `direct_charge_account_unlinked` | falls back to server-confirm | n/a — fails up front |

### Prepare bug fix
The latent bug fix is real and correctly placed. `resolve_merchant_account_and_recompute_fees!` adds a validation error on a rejected cart, but the subsequent `save!` in `build_charge` was revalidating and silently clearing it, letting the charge proceed. Adding `return if fail_all_purchases_when_any_errored` between resolve and `build_charge` (prepare_payment_intent_service.rb:141) closes it. The BR-Connect-affiliated-purchase spec proves the whole order now fails with no charge created — good regression guard.

### Other checks
- `confirmation_token_request_options` uses `.compact` so a nil connect id retrieves the token on the platform — matches the platform-scoped intent.
- Frontend `getCheckoutStripeInstance(null)` → platform loader, byte-for-byte unchanged path for platform sellers.
- Resolver comment accurately updated (dropped the "connected-account sellers need account-scoped Elements" gate now that it is shipped).

### Coverage
Resolver (all 3 branches incl. the legacy-drift fallback), presenter, prepare-service (happy path + reject-instead-of-charge), request-level E2E asserting the charge lands on the connected merchant account, and the account-scoped JS loader. CI is fully green (Test Fast 0-17, lint Ruby/JS, tsc, build all pass).

No blocking issues. Nice work — the by-construction argument tying scope to routing is the right way to make this safe.

Note: PR is stacked on `feat/payment-element-client-confirm-method-resolver`; retarget to `main` once that base merges before shipping.

gumclaw:
🔒 **CLAIM** — session `gumclaw/2026-07-02-5673-connect-webhook`. Scope: implement the connect-side Lane B webhook coverage checklist (route `payment_intent.succeeded`/`.processing` for connected accounts, cross-account resolution + specs) on this branch, drive CI green. Will release when pushed.

gumclaw:
🔓 **RELEASE** — session `gumclaw/2026-07-02-5673-connect-webhook`. Connect-side Lane B webhook coverage landed in commit `4e17d51c1`.

**What shipped**
- `StripeEventHandler#handle_event_for_connected_account` now routes `payment_intent.succeeded`/`.processing` to `StripeChargeProcessor`, extending #5656's webhook-as-source-of-truth backstop to direct-charge (connected-account) Lane B purchases. Without this, removing the eligibility gate would leave connected-account client-confirm orders charged-but-unfulfilled if the browser never returns.
- No further prod changes needed: `StripeChargeProcessor` already resolves the charge reference first and early-returns unless the chargeable is a `client_confirmed` `Charge`, so unrelated high-volume connected-account intents are a cheap no-op; `get_charge_intent`/`get_charge` and `FailAbandonedPurchaseWorker` are already account-aware.
- Specs: inverted the stale "does not route" event-handler expectations to assert routing (+ a negative case), and added connect-account VCR lifecycle specs (webhook finalize, processing, unrelated-intent ignore). Also fixed a stale constant ref surfaced by the main merge (`CLIENT_CONFIRM_PAYMENT_METHOD_TYPES` → `PaymentMethodResolver::LAUNCHED_PAYMENT_METHOD_TYPES`).

**Verified locally (rbenv 3.4.3):** 96 examples, 0 failures across every touched/affected file — connect-confirm lifecycle (13, incl. 3 new connect-side), event handler (21), resolver (15), presenter (35), prepare-service (12).

**Remaining before this is fully live:** the connect-endpoint event subscription is a **deploy step** (after this merges, before enabling direct-charge sellers) — documented in the PR description checklist, not a code change.

PR #5703, opened 2026-07-03
Discount codes: support "all products, except selected" exclusion scoping

slavingia:
Fixes #5692

## What

Universal discount codes ("All products") can now exclude selected products. A new `offer_codes_excluded_products` join table backs an `excluded_products` association on `OfferCode`, applied only when `universal` is set. The discounts form shows an "Excluded products" multi-select when "All products" is checked, the list row reads "50% off of all products except Icon Pack", and the details drawer lists the exclusions.

Enforcement is server-side at every point that resolves a code to a product:

- `Link#universal_offer_codes` (new, with an `OfferCode.not_excluding_product` scope) is now the single lookup for universal codes scoped to a product. It backs purchase-time resolution (`find_offer_code`, used by `Purchase::CreateService`), the public API (`find_offer_code_by_external_id`, `product_and_universal_offer_codes`), search indexing, the product-page discount field (`has_offer_codes?`), and membership auto-renewal discounts (`Subscription#compute_auto_renewal_offer_code`).
- `OfferCode#applicable_products`/`#applicable?` honor exclusions, which covers the checkout discount computation (`OfferCodeDiscountComputingService`) and the code's own price validations — a seller can now exclude the one product whose price would otherwise make the discount invalid.
- Excluded items no longer count toward `minimum_amount_cents`, both in `Purchase::CreateService` and in the cart (the `Discount` payload gains an optional `excluded_product_ids`, only present when exclusions exist, so existing clients and API consumers are unaffected).

A buyer opening an excluded product with the code in the URL gets the existing "Sorry, the discount code you wish to use is invalid." behavior; other products discount normally.

## Why

From a support request by a 500+ product seller (Grafit Team): with only "all products" or "hand-pick each product", large catalogs must manually select hundreds of products, and products released after the code was created are silently not covered. All-minus-exclusions covers new products automatically and makes seasonal sales practical for large catalogs. The join-table + `universal?`-gated approach mirrors the existing `offer_codes_products`/`offer_codes_ownership_products` pattern, adds no columns to existing tables, and leaves non-universal codes untouched.

The controller clears exclusions whenever a code is saved as non-universal, and a model validation rejects exclusions on non-universal codes, so the two scoping modes can't be combined into an ambiguous state.

## Before/After

**Demo (create with exclusions → list → drawer → buyer-side enforcement)** — [full-quality mp4](https://raw.githubusercontent.com/antiwork/gumroad/offer-code-product-exclusions/qa-media/pr-5703-demo.mp4):

demo

**Before** (universal = all products, no way to carve out exceptions) → **After** (checking "All products" reveals the exclusion picker):

create form with exclusion picker

List + details drawer:

list row and drawer

Buyer side — included product gets the discount, excluded product does not:

included product discounted

excluded product not discounted

Dark mode and mobile:

edit form dark

edit form mobile

## Test Results

New coverage: `OfferCode` (`applicable?`, `applicable_products`, `#discount`, price validation skips excluded products, exclusions-require-universal validation, reindex on exclusion change), `Link` (`find_offer_code`, `product_and_universal_offer_codes`, `has_offer_codes?`), `OfferCodeDiscountComputingService` (excluded product in a multi-product cart; excluded-only cart returns `invalid_offer`), `Checkout::DiscountsController` (create/update with `excluded_product_ids`, cleared when not universal, ignored for non-universal), `Checkout::DiscountsPresenter`, `Subscription#auto_renewal_offer_code` (excluded membership no longer auto-applies), and two UI system scenarios.

Ran locally: `spec/models/offer_code_spec.rb` (134 examples, 0 failures), all new exclusion examples across controller/presenter/service specs (6/0), targeted `link_spec` describes (7/0), `subscription_spec` auto-renewal (0 failures). `npm run typecheck`, `eslint`, and `rubocop` clean.

offer_code model spec passing

Known local-env limitation: examples that `create(:purchase, …)` (including this repo's discounts system-spec `before` block) fail on my machine with "We couldn't charge your card" — verified identical on a clean `main` checkout, so it's environmental; CI runs them with working charge fixtures. `bin/test-confidence` requires `ANTHROPIC_API_KEY`, which isn't available in this environment, so the targeted specs above were run directly instead.

## QA steps

1. Checkout → Discounts → New discount, check "All products" — an "Excluded products" multi-select appears.
2. Exclude a product, set a percentage, save. The row reads "…off of all products except …", and the drawer lists the exclusions.
3. Open an excluded product with `/PRODUCT/CODE` — full price and "Sorry, the discount code you wish to use is invalid." Any other product shows the discount.
4. Create a product after saving the code — it's covered automatically.
5. Uncheck "All products" on edit and pick specific products — exclusions are cleared.

---

## AI disclosure

Written with Claude Fable 5 (Claude Code). Prompts given: "design and build this https://github.com/antiwork/gumroad/issues/5692".


slavingia:
@claude review once

app/models/offer_code.rb
```diff
@@ -63,6 +65,9 @@ class OfferCode < ApplicationRecord
   }
   scope :universal, -> { where(universal: true) }
   scope :renewal_eligible, -> { where("existing_customers_only = ? OR JSON_LENGTH(ownership_duration_tiers) > 0", true) }
+  scope :not_excluding_product, ->(product) {
```
slavingia, line 68: Self-review: enforcement concentrates in two choke points — this scope (used by `Link#universal_offer_codes`, which now backs `find_offer_code`, the API lookups, `product_and_universal_offer_codes`, `has_offer_codes?`, and subscription auto-renewals) and `applicable?`/`applicable_products` (cart computation + the code's own validations). NOT EXISTS probes the unique `(offer_code_id, product_id)` index, so it stays cheap even for sellers with many codes.

app/models/offer_code.rb
```diff
@@ -210,6 +217,8 @@ def discount
         minimum_amount_cents:,
       }
     )
+    json[:excluded_product_ids] = excluded_products.map(&:external_id) if universal? && excluded_products.present?
```
slavingia, line 221: Self-review: the key is only present when exclusions exist, so the serialized discount hash is unchanged for every existing code — no churn for API consumers or older cached carts, and the frontend type is optional (`excluded_product_ids?: string[] | null`) to tolerate deploy skew in both directions.

app/services/purchase/create_service.rb
```diff
@@ -89,7 +89,11 @@ def perform
 
       if purchase.offer_code&.minimum_amount_cents.present?
         valid_items = params[:cart_items]
-        valid_items = valid_items.filter { purchase.offer_code.products.find_by(unique_permalink: _1[:permalink]).present? } unless purchase.offer_code.universal
+        valid_items = if purchase.offer_code.universal
+          valid_items.reject { purchase.offer_code.excluded_products.find_by(unique_permalink: _1[:permalink]).present? }
```
slavingia: Self-review: for universal codes every cart item used to count toward `minimum_amount_cents`; excluded items no longer do (mirrored client-side in `cartState.ts`). Without this a buyer could clear the minimum using an item the code doesn't even discount, then the purchase-time check would pass while the cart display disagreed.

app/controllers/checkout/discounts_controller.rb
```diff
@@ -75,15 +76,16 @@ def update
...
     if params.key?(:ownership_duration_tiers)
       update_params[:ownership_duration_tiers] = params[:ownership_duration_tiers].nil? ? nil : offer_code_params[:ownership_duration_tiers]
     end
 
     if offer_code.update(
       **update_params,
       products: current_seller.products.by_external_ids(offer_code_params[:selected_product_ids]),
-      ownership_products: current_seller.products.by_external_ids(offer_code_params[:ownership_product_ids])
+      ownership_products: current_seller.products.by_external_ids(offer_code_params[:ownership_product_ids]),
+      excluded_products:
```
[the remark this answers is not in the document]
slavingia: Good catch — fixed in fafab5b7. `universal` now resolves from the param when present and falls back to the record's persisted value, and when the `excluded_product_ids` key is absent the existing exclusions are kept rather than replaced. Added a controller spec covering the partial-update case.

app/javascript/components/CheckoutDashboard/DiscountsPage.tsx
```diff
@@ -1138,6 +1168,29 @@ const Form = ({
...
+            <Fieldset>
+              <FieldsetTitle>
+                <Label htmlFor={`${uid}excludedProducts`}>Excluded products</Label>
+              </FieldsetTitle>
+              <Select
+                inputId={`${uid}excludedProducts`}
+                instanceId={`${uid}excludedProducts`}
+                options={products
+                  .filter(({ currency_type }) => discount.type !== "cents" || currency_type === currencyCode)
+                  .filter((product) => !product.archived)
```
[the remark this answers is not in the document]
slavingia: Leaving this one as is: an exclusion pointing at a product whose currency the code can't apply to is inert — `universal_with_matching_currency` already filters that product out server-side, so the stale chip changes nothing about which products get discounted. It also mirrors how the existing Products select behaves when the currency changes (selected values persist while options filter). Happy to clear non-matching exclusions on currency change if you'd rather not show the chips.

slavingia:
Ran a deeper multi-angle review of this PR and addressed the findings, one commit each, most severe first:

1. **eef8f909 — Default offer codes now stay applicable under exclusions.** A universal code excluding product X could be set (or kept, by excluding X later) as X's default discount, so product pages and discover cards advertised a slashed price that checkout would refuse. Setting a default now requires `applicable?(product)` (this also closes a pre-existing currency-mismatch variant of the same bug), and excluding a product whose default is this code fails validation, mirroring `validate_not_used_as_default_discount`.
2. **3622405e — Exclusions for deleted products survive edits.** The form built its payload from the visible-products list, so editing a code whose excluded product had been deleted silently cleared that exclusion (fail-open if the product was later restored). The form now resolves exclusions against the server-provided list first.
3. **95d59709 — Membership duration no longer wiped by exclusions.** Excluding the only membership hid the duration field and saved `duration_in_billing_cycles: null`, silently discarding configuration.
4. **88be053a — Cache invalidation is symmetric.** Un-excluding a product now busts its cache (association add/remove callbacks), not just excluding it.
5. **458b92cd — API v2 management endpoints keep working.** `show`/`update`/`destroy` by external id no longer 404 for a universal code addressed via an excluded product; only buyer-facing code resolution enforces exclusions. Note the deliberate asymmetry: v2 `index` for a product still omits codes that exclude it, since it lists codes applicable to that product.
6. **702b7334 / 13114d48 — Query hygiene.** `applicable_products` reads the exclusion ids once instead of twice, and the minimum-amount check filters excluded cart items with one pluck instead of a query per item.

Also verified and ruled out during review: a failed `update` cannot half-apply exclusion changes (Rails rolls the HABTM writes back with the save — tested empirically); the duplicate-code scenario (product-specific code sharing a string with a universal code that excludes that product, then un-excluding) is blocked by the universal uniqueness validation; and cross-seller permalink collisions in the minimum-amount filter can't occur (global uniqueness on `links.unique_permalink`).

PR #4803, opened 2026-04-27
Block new India Stripe Connect accounts at create_account boundary

ershad:
## What

Re-opens #4795 with two follow-up fixes for the issues flagged in review:

1. **Skip blocked countries in `StripeCreateMerchantAccountsWorker`.** The weekly cron (`config/sidekiq_schedule.yml` — `0 0 * * 1`) selects users with no `merchant_accounts` row at all (`LEFT JOIN merchant_accounts WHERE merchant_accounts.id IS NULL`). The new `MerchantRegistrationUserNotReadyError` guard at `stripe_merchant_account_manager.rb:51` raises *before* `MerchantAccount.create!` at line 72, so no row is ever created for India users. Pre-PR, the soft-deleted row excluded users from subsequent runs; post-PR, the user permanently matches the filter and gets re-enqueued every Monday, generating 5 retries (`CreateStripeMerchantAccountWorker` has `retry: 5` and no rescue for this error type) per affected user per week. The fix adds `next if StripeMerchantAccountManager::NEW_ACCOUNT_CREATION_BLOCKED_COUNTRIES.include?(user.alive_user_compliance_info&.legal_entity_country_code)` to the worker's per-user loop.
2. **Rescue `MerchantRegistrationUserNotReadyError` in `Settings::PaymentsController#save`.** The controller only rescues `Stripe::StripeError`. `MerchantRegistrationUserNotReadyError < GumroadRuntimeError < RuntimeError` (not a `Stripe::StripeError`), so a direct API submission of bank-account params for a new India seller bubbles up uncaught and 500s. `UpdatePayoutMethod#process_full_bank_account_replacement` has no `can_setup_bank_payouts?` gate (only the PayPal branch checks `paypal_payouts_supported?`), so the bank_account row gets persisted, then `Settings::PaymentsController#save` line 81's condition (`active_bank_account.present? && merchant_accounts.stripe.alive.empty? && native_payouts_supported?`) is satisfied (India remains in `SUPPORTED_COUNTRIES` per this PR's design), `create_account` is invoked, and the unrescued error 500s. The fix extends the rescue to `rescue Stripe::StripeError, MerchantRegistrationUserNotReadyError` mirroring the established pattern in `Admin::UsersController#create_stripe_managed_account`.

The remainder of the PR is unchanged from #4795:
- `NEW_ACCOUNT_CREATION_BLOCKED_COUNTRIES` allowlist + guard at `StripeMerchantAccountManager.create_account`.
[...]

PR #5714, opened 2026-07-04
Gate bank-account saves on countries where new Stripe accounts are blocked

gumclaw:
## What

Gates the bank-account branch of `UpdatePayoutMethod` on whether the seller's country can actually create a Stripe merchant account, mirroring the guard the PayPal branch already has (`paypal_payouts_supported?`). Sellers from blocked countries (currently India, per `StripeMerchantAccountManager::NEW_ACCOUNT_CREATION_BLOCKED_COUNTRIES`) now get a clear inline error — "Bank payouts are not supported in your country yet. Please use PayPal instead." — instead of a persisted-but-dead bank account row.

## Why

India was intentionally blocked for new Stripe Connect accounts in #4803, but the bank-details save path was never gated (this exact persistence gap was flagged in the #4803 review; only the 500-error was fixed). The result for a new India seller:

1. The `bank_accounts` row saves fine and the UI says "Thanks! You're all set."
2. `StripeMerchantAccountManager.create_account` raises `MerchantRegistrationUserNotReadyError` (blocked country), and `StripeCreateMerchantAccountsWorker` skips the user — so no `merchant_accounts` row is ever created and `stripe_bank_account_id` stays NULL forever.
3. `StripePayoutProcessor.is_user_payable` fails every week → the misleading "the payout bank account was not correctly set up" payout note, while the dashboard keeps showing a concrete next-payout date and amount.
4. The seller re-enters bank details repeatedly (each save "succeeds") and files repeated support tickets.

Investigated in antiwork/gumroad-private#886 (worked example there: 7 dead bank rows, 5 consecutive weekly skips, a three-figure balance stranded ~6 weeks).

## How

- `UpdatePayoutMethod#process_full_bank_account_replacement` now returns `{ error: :bank_payouts_not_supported }` before validating/persisting when the country can't link a bank account.
- Grandfathered sellers (an alive Stripe merchant account already exists — e.g. India accounts created before the #4803 block) are allowed through so they can keep updating their bank details.
- UAE stays allowed (business-account path), and an unknown/blank country falls through to the existing downstream error, matching prior behavior.
- Controller maps the new error symbol to the same message the merchant-account-creation rescue already uses for this case.

## Test plan

- `bundle exec rspec spec/services/update_payout_method_spec.rb` — 17 examples, 0 failures locally (3 new: India seller blocked with no row persisted; India seller with a live merchant account still allowed; US seller unaffected).
- `bundle exec rubocop` on the three changed files — no offenses.
- autoreview (Codex, branch mode vs origin/main) — clean, no actionable findings.

Not covered here (follow-ups noted on the tracking issue): hiding the next-payout date when the payout method can never process, and a one-time sweep of existing India accounts with dead bank rows + stranded balances.

### DIFF
--- a/app/controllers/settings/payments_controller.rb
+++ b/app/controllers/settings/payments_controller.rb
@@ -247,6 +247,8 @@ class Settings::PaymentsController < Settings::BaseController
                         "Email address cannot contain non-ASCII characters"
                       when :paypal_payouts_not_supported
                         "PayPal payouts are not supported in your country."
+                      when :bank_payouts_not_supported
+                        "Bank payouts are not supported in your country yet. Please use PayPal instead."
                       when :concurrent_payout_method_change
                         "Another change was submitted at the same time. Please try again."
       end
--- a/app/services/update_payout_method.rb
+++ b/app/services/update_payout_method.rb
@@ -196,6 +196,8 @@ class UpdatePayoutMethod
     ACCOUNT_NUMBER_SEPARATOR_CHARACTERS = /[[:space:]\p{Cf}-]/
 
     def process_full_bank_account_replacement
+      return { error: :bank_payouts_not_supported } unless bank_payouts_supported?
+
       account_number = normalize_account_number(params[:bank_account][:account_number])
       account_number_confirmation = normalize_account_number(params[:bank_account][:account_number_confirmation])
       return { error: :account_number_does_not_match } if account_number != account_number_confirmation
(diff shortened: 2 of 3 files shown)


gianfrancopiana:
Closing as not planned via https://github.com/antiwork/gumroad-private/issues/886.

gumclaw:
Understood — closing it out and deleting the branch. The investigation details stay on antiwork/gumroad-private#886 if this resurfaces through support.

PR #5716, opened 2026-07-05
Bridge same-store product links out of the sandboxed custom profile iframe

gumclaw:
## What

Fixes in-iframe product navigation on custom-HTML profile pages (custom domain or subdomain). Plain product links in the seller's HTML currently navigate the sandboxed landing iframe itself, loading the product page + checkout on an opaque origin where cookies/storage are unavailable — checkout hangs on Chromium/Firefox-family browsers and Safari refuses to render the page at all. Sellers can't self-fix with `target="_top"`: the sanitizer strips it and the wrapper sandbox deliberately omits `allow-top-navigation`.

This PR adds a `gumroad:navigate` postMessage bridge, mirroring the product wrapper's existing `gumroad:checkout` pattern:

- **Child script** (injected into the sandboxed embed): a capture-phase click listener forwards clicks on same-store links to the parent. It skips modified clicks, anchors with an explicit target, non-http(s) URLs, hosts outside the allowlist, and hash-only/empty/self-resolving hrefs (same-document navigations stay in-frame).
- **Parent script** (trusted wrapper, CSP-nonce'd): independently re-validates everything before navigating — the message must come from our iframe (`e.source === frame.contentWindow`, `e.origin === "null"`), be http(s), and target a host on the seller's own store allowlist. A compromised child gains nothing; the parent re-validates.
- **Host allowlist**: `request.host`, the seller's subdomain, and the seller's custom domain **only when `active?`** (verified + valid certificate — the same eligibility check `UrlService`/`ContentExtractor` apply before emitting custom-domain URLs). An unverified or stale custom-domain record never qualifies, so seller HTML still cannot top-navigate the visitor's tab to an arbitrary site.

Fixes the confirmed root cause in antiwork/gumroad-private#906 (live repro: custom-domain store where every "View tool" card leads to a stuck checkout in-frame, while the same product URL opened top-level works).

## Why

Product links are the #1 use of a custom profile — click a product card, buy. Today that primary conversion path is broken in-place for every custom-HTML profile. The bridge restores it without weakening the sandbox: links to any non-store host keep the default in-frame behavior, and the parent only ever navigates to hosts Gumroad actually serves the store on.

## Alternative

#5717 is a simpler alternative (retarget same-store links to `target="_blank"` at click time — new tab instead of same-tab navigation). One of the two should merge, not both. This bridge option keeps the visitor in the same tab, which matches normal storefront behavior, at the cost of slightly more moving parts.

## Test plan

- `spec/requests/profile_custom_html_spec.rb` — bridge injection with the correct host allowlist on embed + wrapper, CSP nonce on the wrapper script, hash-only/empty-href guard present, another seller's domain never in the allowlist, and the custom domain only allowlisted once `active?` (unverified record excluded).
- `spec/controllers/users_controller_custom_html_spec.rb` — updated: the old blanket `not_to include("postMessage")` assertions narrowed to checkout-specific ones (`gumroad:checkout`, `wanted=true`), positive bridge assertions added.
- Local run: `bundle exec rspec spec/requests/profile_custom_html_spec.rb spec/controllers/users_controller_custom_html_spec.rb` — 44 examples, 0 failures.
- autoreview (Codex): iterated over 4 passes; findings fixed along the way were the stale spec assertions, the unverified-custom-domain escape, same-document anchor hijacking, and the stale-verified-domain gate (now `active?`). Final pass clean.

## AI disclosure

Authored by Gumclaw (Hermes agent) from the investigation on antiwork/gumroad-private#906; reviewed with the Codex autoreview loop before opening.

### DIFF
--- a/app/controllers/concerns/renders_custom_html_pages.rb
+++ b/app/controllers/concerns/renders_custom_html_pages.rb
@@ -108,6 +108,93 @@ module RendersCustomHtmlPages
   end
 
   private
+    # --- Same-store navigation bridge ---------------------------------------
+    #
+    # The seller's HTML lives in an opaque-origin sandboxed iframe. A plain
+    # product link (<a href="/l/xyz">) therefore navigates the IFRAME itself,
+    # loading the full product page + checkout on an origin where cookies and
+    # storage are unavailable — checkout hangs, and Safari won't render the
+    # page at all. The seller can't use target="_top" either: the sanitizer
+    # strips it and the sandbox omits allow-top-navigation on purpose (seller
+    # HTML must never be able to redirect the visitor's tab to an arbitrary
+    # site).
+    #
+    # So, mirroring the product wrapper's gumroad:checkout pattern, clicks on
+    # links that point at the seller's OWN store are turned into a postMessage
+    # to the trusted parent wrapper, which re-validates the destination host
+    # and performs the top-level navigation itself. Links to any other host
+    # keep the browser's default in-frame behavior, and the parent ignores any
+    # message whose URL isn't on the seller's own hosts — the sandbox
+    # guarantee is unchanged for everything except the seller's own store
+    # pages.
+
+    # Runs inside the sandboxed (untrusted) document. It only decides which
+    # clicks to FORWARD; the parent independently re-validates the URL, so a
+    # compromised copy of this script gains nothing.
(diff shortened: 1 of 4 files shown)


gumclaw:
Superseded by #5718 (same fix, complete test coverage). Closing — one PR per fix, not iterations left open.

PR #5717, opened 2026-07-05
Open same-store product links from custom profile pages in a new tab

gumclaw:
## What

Fixes in-iframe product navigation on custom-HTML profile pages (custom domain or subdomain). Plain product links in the seller's HTML currently navigate the sandboxed landing iframe itself, loading the product page + checkout on an opaque origin where cookies/storage are unavailable — checkout hangs on Chromium/Firefox-family browsers and Safari refuses to render the page at all. Sellers can't self-fix with `target="_top"`: the sanitizer strips it and the wrapper sandbox deliberately omits `allow-top-navigation`.

This PR takes the simpler of the two viable approaches: a small script injected into the sandboxed embed retargets clicks on same-store links to `target="_blank" rel="noopener"` at click time. The sandbox already grants `allow-popups allow-popups-to-escape-sandbox`, so the product page opens in a new tab with a real origin where cookies and checkout work.

Details:

- Click-time retargeting (not serve-time HTML rewriting) because the real-world breakage builds product cards **after** load from the injected `gumroad-data` JSON (`a.href = p.url`) — serve-time rewriting would miss every JS-built link.
- Skips anchors with an explicit target, non-http(s) URLs, hosts outside the store allowlist, and hash-only/empty/self-resolving hrefs (same-document navigations stay in place).
- **Host allowlist**: `request.host`, the seller's subdomain, and the seller's custom domain **only when `active?`** (verified + valid certificate — the same eligibility check `UrlService`/`ContentExtractor` apply before emitting custom-domain URLs). A saved-but-unverified or stale custom-domain record never qualifies.
- No change to the sandbox tokens; links to non-store hosts keep the default in-frame behavior.

Fixes the confirmed root cause in antiwork/gumroad-private#906 (live repro: custom-domain store where every "View tool" card leads to a stuck checkout in-frame, while the same product URL opened top-level works).

## Why

Product links are the #1 use of a custom profile — click a product card, buy. Today that primary conversion path is broken in-place for every custom-HTML profile.

## Alternative

#5716 is the fuller alternative (a `gumroad:navigate` postMessage bridge that top-navigates the same tab, mirroring the product wrapper's `gumroad:checkout` pattern). One of the two should merge, not both. This option is smaller and touches no parent-window code, at the cost of opening products in a new tab instead of the same tab.

## Test plan

- `spec/requests/profile_custom_html_spec.rb` — retargeting script injected with the correct host allowlist, another seller's domain never in the allowlist, custom domain only allowlisted once `active?` (unverified record excluded), hash-only/empty-href guard present.
- Local run: `bundle exec rspec spec/requests/profile_custom_html_spec.rb` — 20 examples, 0 failures.
- autoreview (Codex): the unverified/stale-custom-domain allowlist gap and same-document anchor retargeting were found and fixed during review.

## AI disclosure

Authored by Gumclaw (Hermes agent) from the investigation on antiwork/gumroad-private#906; reviewed with the Codex autoreview loop before opening.

### DIFF
--- a/app/controllers/concerns/renders_custom_html_pages.rb
+++ b/app/controllers/concerns/renders_custom_html_pages.rb
@@ -139,10 +139,20 @@ module RendersCustomHtmlPages
               // A seller who set an explicit target (e.g. their own _blank) keeps it.
               var target = (anchor.getAttribute("target") || "").trim();
               if (target !== "" && target.toLowerCase() !== "_self") return;
+              var rawHref = (anchor.getAttribute("href") || "").trim();
+              // Hash-only and empty hrefs are same-document navigations (section
+              // jumps, "#"-placeholder buttons wired up by the seller's own JS).
+              // They resolve against the iframe's own /landing/embed URL, so
+              // without this guard they would pass the host check and open the
+              // raw embed endpoint in a new tab instead of scrolling in place.
+              if (rawHref === "" || rawHref.charAt(0) === "#") return;
               var url;
-              try { url = new URL(anchor.getAttribute("href"), window.location.href); } catch (_e) { return; }
+              try { url = new URL(rawHref, window.location.href); } catch (_e) { return; }
               if (url.protocol !== "https:" && url.protocol !== "http:") return;
               if (ALLOWED_HOSTS.indexOf(url.hostname.toLowerCase()) === -1) return;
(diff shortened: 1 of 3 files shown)


gumclaw:
Superseded by #5718 — the postMessage bridge is the chosen fix; the target=_blank rewrite was the interim workaround, not the fix.

PR #5746, opened 2026-07-07
Revert "Size Indian e-mandate cap to the undiscounted total for limited-duration discounts"

gumclaw:
## What

Reverts #5726 — sizing the Indian e-mandate cap to the undiscounted total for limited-duration discounts.

## Why

Post-deploy evidence on the affected seller (the one whose support thread motivated #5726) shows checkout failures continuing and broadening after the change shipped:

- Zero successful purchases on the seller since the deploy (merged 2026-07-06 14:40 UTC); the last success predates it.
- Multiple buyers, multiple cards each, all failing at the ₹0 e-mandate registration screen with "Your card does not support this type of purchase" — including a card that completed a different purchase the previous day.
- One buyer failed on three different cards within minutes, all post-deploy.

The scaled-up mandate cap is the prime suspect: some Indian issuing banks appear to reject e-mandate registration when the requested maximum is a large multiple of the visible charge amount. Reverting restores the pre-#5726 behavior (cap = charged total) while the renewal-cap problem gets a safer fix in a separate PR — likely capping the scale-up or registering the mandate at the undiscounted amount only when the issuer accepts it.

This is a clean revert of the squash-merge commit `ec11655e` — touches only `app/models/purchase.rb` and `spec/models/purchase_spec.rb`, exactly the files from #5726.

## Test plan

- Full `#mandate_options_for_stripe` describe block passes locally after the revert (spec results in comment below).
- CI green.
- Post-merge: watch the seller's checkout attempts for successful e-mandate registrations.


gumclaw:
Closing unmerged — the regression this revert was opened for is **refuted** after a full Stripe-object audit (details + evidence table: https://github.com/antiwork/gumroad/pull/5726#issuecomment-4903953879).

Summary:
- The seller's discount codes have `duration_in_billing_cycles = nil`, so the code changed in #5726 never executed for any of the failing purchases.
- Stripe PaymentIntents show identical mandate sizing pre/post deploy (mandate = charged total, not the scaled cap), with the same bank-side `payment_intent_authentication_failure` before and after.
- Failures began Jul 5, over a day before #5726 merged; one buyer failed identically across pre-merge, pre-deploy, and post-deploy states.
- Platform-wide: zero post-deploy IN e-mandate auth failures outside this seller; zero purchases have exercised the changed path yet.

Reverting would fix nothing (path is inert for this seller) and would reintroduce the renewal-cap bug for sellers using limited-duration codes. Root cause is buyer-bank-side ₹0 e-mandate authentication declines.

PR #5750, opened 2026-07-07
Validate store agent api_write body keys against the catalog to stop the create-product doom loop

gumclaw:
## What

Stops the store AI agent's create-product doom loop by validating proposed `api_write` body keys against the catalog endpoint's declared `params` — at propose time in `Ai::StoreAgentService` and again before dispatch in `Ai::StoreAgentActionExecutor` — and fixes the `Product::Prices#associate_price` error message that leaked the internal `price_cents` column name.

## Why

Root cause (gumroad-private#953): the model stages `create_product` with the body key `price_cents`, while the catalog and the v2 API only accept `price`. Unknown body keys were forwarded unvalidated; the v2 create endpoint reads only `params[:price]`, so the product is built with no price. `Product::Prices#associate_price` then fails with "New products should be created with a price_cents", and the executor echoes that verbatim into the chat. Because the error names the internal column `price_cents`, it *teaches the model to repeat the exact mistake* — the model retries with `price_cents` again and the loop never terminates.

The fix breaks the loop in three places:

1. **`Ai::StoreAgentService#propose_api_write`** — when the model stages a write whose body carries keys the catalog endpoint doesn't declare, the proposal is not staged. Instead the tool_result is corrective and self-healing: `Unknown param price_cents for create_product; this endpoint accepts: name, price, url, description, custom_permalink, price_currency_type.` The model can retry correctly in the same turn.
2. **`Ai::StoreAgentActionExecutor#execute`** — the same check runs before dispatch, as defense in depth against a stale or tampered proposal that predates the service-side check.
3. **`Product::Prices#associate_price`** — the error message now says "created with a price" (the public field name) instead of "created with a price_cents", so it can never again teach an API/agent caller a param that doesn't exist. String-only change; no behavior change.

Since the check turns the catalog's `params` lists into an allowlist, the offer-code rows were completed with the keys the v2 controller actually reads (`universal`, `amount_cents`, `minimum_amount_cents`) so legitimate writes aren't rejected. The lists remain a deliberately curated subset elsewhere — they are exactly what the system-prompt manifest teaches the model, so the agent only drives the surface it was told about.

Refs: antiwork/gumroad-private#953

## Test plan

- `spec/services/ai/store_agent_service_spec.rb` — an `api_write` with `{"name" => "X", "price_cents" => 9900}` on `create_product` produces a corrective tool_result naming the unknown and accepted keys, and stages no proposed action.
- `spec/services/ai/store_agent_action_executor_spec.rb` — the same body is rejected before dispatch (`seller.links.count` unchanged) with a message naming the allowed keys.
- `spec/modules/product/prices_spec.rb` — a product created with no price fails with the public-field message "New products should be created with a price".
- Local: `bundle exec rspec spec/services/ai/store_agent_action_executor_spec.rb spec/services/ai/store_agent_service_spec.rb spec/services/ai/store_agent_api_catalog_spec.rb spec/modules/product/prices_spec.rb` → **119 examples, 0 failures**.
- `bundle exec rubocop` on all 7 touched files → **no offenses detected**.
- autoreview (Codex): clean — one earlier finding (allowlist stricter than the offer-code controller) was accepted and fixed by completing the catalog rows.

Backend-only change; no UI surface is touched, so no screenshots.

---

AI disclosure: authored by an AI agent (Claude Opus) working from gumroad-private#953.

### DIFF
--- a/app/modules/product/prices.rb
+++ b/app/modules/product/prices.rb
@@ -272,7 +272,7 @@ module Product::Prices
 
       price_cents = read_attribute(:price_cents)
       if price_cents.blank?
-        errors.add(:base, "New products should be created with a price_cents")
+        errors.add(:base, "New products should be created with a price")
         raise Link::LinkInvalid
       end
 
--- a/app/services/ai/store_agent_action_executor.rb
+++ b/app/services/ai/store_agent_action_executor.rb
@@ -51,11 +51,8 @@ class Ai::StoreAgentActionExecutor
     # and fail downstream with a confusing internal error. The propose path (StoreAgentService)
     # already rejects these, so a well-formed proposal never hits this.
     body = normalize_body(params[:params])
-    unknown_keys = endpoint.unknown_param_keys(body)
-    if unknown_keys.any?
-      accepted = endpoint.params.any? ? "this endpoint accepts: #{endpoint.params.join(', ')}" : "this endpoint accepts no params"
-      return failure("Unknown param #{unknown_keys.join(', ')} for #{endpoint.id}; #{accepted}.")
-    end
+    unknown_keys_error = endpoint.unknown_param_keys_error(body)
+    return failure(unknown_keys_error) if unknown_keys_error
 
     response = api_client.write(endpoint.method, path, body)
 
@@ -89,7 +86,7 @@ class Ai::StoreAgentActionExecutor
     # The proposed params arrive with string keys (they round-tripped through JSON in the proposal).
     # Hand the body to the client as a plain hash; the API normalizes types itself.
     def normalize_body(raw)
-      return {} unless raw.respond_to?(:to_h)
+      return {} unless raw.is_a?(Hash) || raw.is_a?(ActionController::Parameters)
       raw.to_h
     end
 
--- a/app/services/ai/store_agent_action_executor.rb
+++ b/app/services/ai/store_agent_action_executor.rb
@@ -44,7 +44,20 @@ class Ai::StoreAgentActionExecutor
     end
 
     path = endpoint.expand_path(params[:path_params])
-    response = api_client.write(endpoint.method, path, normalize_body(params[:params]))
+
+    # Defense in depth against a stale or tampered proposal: refuse a body carrying keys the
+    # endpoint doesn't declare BEFORE dispatching. The v2 API silently ignores unknown body keys,
+    # so such a call would drop the value the seller confirmed (e.g. a price sent as "price_cents")
+    # and fail downstream with a confusing internal error. The propose path (StoreAgentService)
+    # already rejects these, so a well-formed proposal never hits this.
+    body = normalize_body(params[:params])
+    unknown_keys = endpoint.unknown_param_keys(body)
+    if unknown_keys.any?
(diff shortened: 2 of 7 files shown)


gumclaw:
## CI triage + adversarial review

### Root cause of the red Test Slow shards (9/15/30)
The 3 failing specs (`spec/requests/purchases/product/taxes_spec.rb:3878/:3901/:4044`) are live-Stripe SCA checkout tests that pay with Stripe's Mexico test card `4000004840008001`. Since ~15:30 UTC on 2026-07-07 Stripe's test mode started requiring a 3DS challenge for that card, so the checkout stalled and the 60s purchase-success alert wait timed out — **on every branch**, not just this one. Main fixed it in #5753 (`71edc35d0`, passes `sca: true` to `check_out`), which landed *after* this branch's last CI run. **Not caused by this PR's diff** — the `prices.rb` error-string change has zero couplings outside `spec/modules/product/prices_spec.rb` (verified by repo-wide grep).

**Fix applied:** merged `origin/main` (picks up #5753). Local verification on this branch: the Mexico-card spec (`:3878`) passes locally with the merged fix (1 example, 0 failures, ~216s live-Stripe run); the remaining shards are re-verified by the CI run on the merge push. All three call sites now pass `sca: true`.

### Adversarial review (Claude Code) — findings fixed in `72b216977`
The review's key point: this PR flips catalog `params` from documentation into an **enforced allowlist**, but only offer codes had been audited (`e1abd85c6`). Auditing every write endpoint against its actual v2 controller found the same defect class elsewhere:

- **`export_sales`** — declared `after`/`before`; controller reads `from`/`to`/`product_id`. Validation rejected every real key and blessed keys the endpoint ignores (silent full-history export). Fixed.
- **`create_thumbnail`/`create_cover`** — controllers accept `signed_blob_id` OR `url`, and `LinksController`'s own upload guidance tells the model to POST a `signed_blob_id` — which validation then refused (a freshly manufactured doom loop). Added `signed_blob_id`.
- **`update_custom_field`** — controller only reads `required`; declared `name`/`type` would pass validation, show a rename card, execute, and change nothing — the exact silent-drop bug this PR exists to stop. Now `required` only.
- **`create_product`/`update_product`** — dropped no-op `url` (create never reads it); reconciled asymmetries: `max_purchase_count` now on create too, `price_currency_type` now on update too (this PR's own `requested_currency` card preview already honors it for update).
- **`update_user`** — dropped `twitter_handle` (`permitted_update_params` is name/bio only).
- **`create_email`** — added `draft` (`publish_requested?` reads it; `draft: false` means publish).
- Centralized the corrective message on `Endpoint#unknown_param_keys_error` so propose/confirm paths can't drift; pluralized for multi-key bodies; hardened `normalize_body` so a tampered array body fails as a validation failure instead of a `TypeError` 500.

**Dismissed:**
- Upsell/variant param gaps (`product_ids`, `paused`, `rich_content`, etc.) — deliberate curated-subset scope: the manifest never taught these keys, so it's a capability ceiling, not a broken taught path. Can widen later if the agent needs them.
- `universal` string-comparison fragility (`== "true"`) — real but pre-existing and only relevant if the client switches to JSON dispatch; noted, not changed here.
- Seller-visible executor message wording — only reachable via tampered/stale proposals; acceptable.

Local runs: store-agent service/executor/catalog + prices specs — **122 examples, 0 failures**; rubocop clean on all touched files.

PR #5754, opened 2026-07-07
Record Gumroad views from custom HTML product pages

gumclaw:
## What

Custom HTML product wrappers now load the existing internal product view counter entrypoint even when the seller has no third-party analytics configured or has disabled third-party analytics.

The wrapper also renders a CSRF token and initializes the request helper with it so the `/increment_views` POST succeeds from the layout-less wrapper document.

## Why

Root cause: confirmed. Custom HTML product pages bypass the React product page, so `incrementProductViews` never ran and Gumroad dashboard views flatlined for sellers who switched their products to custom HTML.

This keeps seller pixels/snippets gated by `tracking_enabled`, but records Gumroad's own product page view regardless of seller analytics configuration, matching the standard product page behavior.

## Test plan

- `rbenv exec bundle exec rspec spec/requests/products/show/custom_html_analytics_spec.rb spec/requests/profile_analytics_spec.rb --format progress`
- `npm run lint-fast -- app/javascript/entrypoints/custom_html_analytics.ts`
- `npm run typecheck` was attempted, but the local checkout is missing the `vite/client` type definition in `node_modules`, so the full typecheck fails before checking this patch.
- `python3 ~/.hermes/skills/software-development/autoreview/scripts/autoreview --mode branch --base origin/main` returned clean after the CSRF fix.

## Visual evidence

Not applicable. This is an analytics instrumentation fix for custom HTML wrappers, with request specs covering the rendered wrapper head and config payload.

## AI disclosure

Gumclaw drafted and tested this change, including an autoreview pass before opening the PR.

Fixes antiwork/gumroad-private#954.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1081,6 +1081,12 @@ class LinksController < ApplicationController
     def custom_html_analytics_head(product)
       tracking_enabled = analytics_enabled?(seller: product.user)
       analytics = product.analytics_data
+      # When third-party tracking is off, blank the pixel ids before they reach
+      # the props JSON. The frontend never uses them in that case (every pixel
+      # call is gated on tracking_enabled), but without this a seller who
+      # disabled third-party analytics would still see their Google/Facebook/
+      # TikTok ids embedded in the page source for anyone to inspect.
+      analytics = analytics.merge(google_analytics_id: nil, facebook_pixel_id: nil, tiktok_pixel_id: nil) unless tracking_enabled
       has_product_third_party_analytics = tracking_enabled && product.has_third_party_analytics?("product")
       has_configured_pixel = tracking_enabled && analytics.values_at(:google_analytics_id, :facebook_pixel_id, :tiktok_pixel_id).any?(&:present?)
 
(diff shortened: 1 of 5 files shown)


gianfrancopiana:
@gumclaw Address feedback https://github.com/antiwork/gumroad/pull/5754#issuecomment-4908278429

gumclaw:
Addressed both Greptile findings in ad437c56c:

- **Pixel ids in page source when analytics is disabled** — real regression, fixed. The controller now blanks `google_analytics_id`, `facebook_pixel_id`, and `tiktok_pixel_id` in the props JSON when `tracking_enabled` is false, so a seller who disabled third-party analytics sees no trace of their pixel ids in the HTML. The disabled-analytics spec now asserts the configured id does not appear anywhere in the response body and that all three ids serialize as null.
- **Stale comment above the pixel meta tags** — rewritten to describe the new behavior: the block always renders, and the pixel meta tags are conditional on tracking being enabled plus a configured pixel id.

Ran `spec/requests/products/show/custom_html_analytics_spec.rb` locally: 5 examples, 0 failures.

PR #5776, opened 2026-07-09
Add targeted find/replace edits for profile custom HTML so the store agent stops regenerating whole pages

gumclaw:
## What

Adds a targeted-edit surface for profile custom HTML so the store agent can change one part of an existing page instead of regenerating the whole thing:

- `POST /v2/user/custom_html/edit` (`edit_profile` scope, `custom_html_pages` flag): takes `find` + `replace`, requires `find` to match the current page exactly once (zero matches and ambiguous matches return corrective errors the model can act on), splices the replacement in literally, re-sanitizes the full result, and saves. Returns `previous_custom_html` for one-shot recovery, same as the full update.
- Catalog: new `edit_user_custom_html` write entry, and a reworded `update_user_custom_html` summary that calls the full replacement destructive and points the model at the targeted edit. The summaries feed the agent's system-prompt manifest, so the steering propagates automatically.
- System prompt: an explicit rule to read the current page and use the targeted edit for any change to an existing page — never regenerate from scratch unless the creator asks for a whole new page. The agent picks the right endpoint itself; no new user-facing choice.

## Why

Fixes antiwork/gumroad-private#984. Two sellers lost hand-built storefront pages the same way: they asked the agent for a tiny change (a color; renaming buttons on cards) and the agent's only page-write surface was the full-page `update_user_custom_html`, so it replaced an 84KB curated page with a ~1.8KB regeneration. With targeted edits in the catalog plus prompt/summary steering, a small request maps to a small, non-destructive write.

## Design notes

- The exactly-once match rule makes an edit unambiguous and self-correcting: the error for an ambiguous `find` names the match count and tells the model to include more surrounding context, so it retries within the same turn instead of falling back to a full rewrite.
- `String#sub` is called in block form so a `replace` containing `\0`/`\\` is inserted literally instead of being treated as a backreference.
- The whole spliced document is re-sanitized (not just the inserted snippet) because a replacement can change how surrounding markup parses. Blank-to-nil normalization matches the full update, so an edit that empties the page unpublishes it consistently.
- Same row lock as `update_custom_html`, so concurrent writers serialize and the splice always runs against the latest committed page.
- Size limit is enforced on the edited result before the sanitizer parses it.

## Test evidence

Ran locally (Ruby 3.4.3, MySQL up):

- `spec/controllers/api/v2/users_controller_edit_custom_html_spec.rb` — 17 examples covering the happy path, deletion via empty `replace`, zero/ambiguous match refusal, literal (non-regex) matching, backslash-safe replacement, full-document re-sanitization, blank-to-nil unpublish, size limit, param validation, unconfirmed email, auth (401/403), and the feature flag gate.
- `spec/services/ai/store_agent_api_catalog_spec.rb`, `spec/services/ai/store_agent_service_spec.rb`, `spec/services/ai/store_agent_action_executor_spec.rb`, `spec/controllers/api/v2/users_controller_custom_html_spec.rb`, `spec/controllers/api/v2/users_controller_spec.rb` — all green.
- 118 examples total, 0 failures. `rubocop` clean on all touched files.

Not a UI change — no screenshots; the agent's confirmation card renders the `find`/`replace` fields through the existing generic write-card path.

## Related

- #5767 adds the scoped `profile_design` endpoint for the same incident; this PR covers the "agent edits an existing custom page" half. Per the discussion there, custom HTML stays the primary surface, so targeted edits are the durable fix for destructive rewrites.

### DIFF
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -2,8 +2,8 @@
 
 class Api::V2::UsersController < Api::V2::BaseController
   before_action -> { doorkeeper_authorize!(*Doorkeeper.configuration.public_api_read_scopes.concat([:view_public])) }, only: [:show, :ifttt_sale_trigger, :custom_html]
-  before_action(only: [:update, :update_custom_html, :preview_custom_html]) { doorkeeper_authorize! :edit_profile }
-  before_action :ensure_custom_html_pages_enabled, only: [:custom_html, :update_custom_html, :preview_custom_html]
+  before_action(only: [:update, :update_custom_html, :edit_custom_html, :preview_custom_html]) { doorkeeper_authorize! :edit_profile }
+  before_action :ensure_custom_html_pages_enabled, only: [:custom_html, :update_custom_html, :edit_custom_html, :preview_custom_html]
 
   def show
     if params[:is_ifttt]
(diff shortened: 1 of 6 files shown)


slavingia:
Should this support targeted LLM edits too? So 3 levels:

Pure deterministic
Agent targeted edit
Whole page agent edit

Wdyt?

gumclaw:
I think we mostly get level 2 for free with this design: the store agent is already an LLM, and it composes the find/replace itself after reading the page. So a "targeted LLM edit" today is the agent deciding what to change and the endpoint being the deterministic splice that guarantees it can't destroy the rest of the page. The three levels as shipped are: deterministic primitive (this endpoint), the agent using that primitive for targeted edits (the prompt/catalog steering), and whole-page regeneration (the existing update).

Where a separate server-side LLM edit level would earn its keep is where exactly-once matching gets awkward: a change scattered across many spots (renaming every buy button is N sequential calls today), or very large pages where the model misquotes the snippet and burns retries. But I'd cover most of that with two cheap deterministic extensions first — accept an array of find/replace ops in one call, and a `replace_all` flag for repeated snippets. Both stay deterministic, keep the one-shot `previous_custom_html` recovery, and don't add a second model call's cost and latency to the write path.

If after that the agent still falls back to full rewrites, then I'd add the scoped LLM level: agent selects a bounded region plus an instruction, a server-side model rewrites only that region, same sanitize/size/lock path. That contains the non-determinism to an explicitly bounded slice instead of the whole page. We can instrument the zero/ambiguous-match error rate from this endpoint to know whether we actually need it.

So: yes to the direction, but I'd ship this as-is, add multi-op and `replace_all` as a follow-up, and gate the scoped-LLM level on real failure data rather than building it speculatively. Happy to open the multi-op follow-up now if you want it.

PR #5778, opened 2026-07-09
Require city server-side for Japanese address submissions

gumclaw:
## What

Server-side validation: when a payout compliance submission actually changes a Japanese address (a submitted address field differs from the stored record, or the country/business context switches while a Japanese address is on file), the matching city fields (`city` + `city_kana`, or `business_city` + `business_city_kana` for businesses) must be present — either in the submission or already on the record. The Payments settings form echoes back every stored field on save, so change detection diffs against `current_compliance_info` rather than treating field presence as a resubmission — saves that only change the phone, payout frequency, or payout pause on a legacy no-city record are unaffected.

## Why

#5739 added the city fields to the Japanese onboarding form and made them required, but only in the browser. A page loaded on the pre-#5739 frontend bundle and submitted after the deploy (or any client that skips the form) can still create a Japanese compliance record with no city — the seller types the city at the start of the street line, Stripe auto-fills `city` from the postal code, and the address presents to our JP payment partner with the city repeated. One such stale-session record showed up in production two days after the #5739 deploy (antiwork/gumroad-private#1002).

Requiring the city outright at the model level would break unrelated updates to un-backfilled legacy records, so this check fires only when the submission is actually (re)entering a Japanese address.

## Test plan

- `bundle exec rspec spec/services/update_user_compliance_info_spec.rb` — 33 examples, 0 failures (11 new: reject missing individual/business city, reject a half-filled city pair, reject individual/business postal-code-only updates on no-city records, accept with city, accept when record already has a city (individual and business), accept a full UI-shaped settings-form save that changes only the phone on a legacy no-city record)
- `rubocop` clean on both touched files

Not a UI change — no visual difference; the form already requires these fields client-side.

Part of antiwork/gumroad-private#1002.

### DIFF
--- a/app/services/update_user_compliance_info.rb
+++ b/app/services/update_user_compliance_info.rb
@@ -9,15 +9,17 @@ class UpdateUserComplianceInfo
     street_address: [:country],
     business_street_address: [:business_country, :country],
   }.freeze
-  # The fields that make up a Japanese address, including the city pair. A submission is treated
+  # The fields that make up a Japanese address, including the city pair and the prefecture
+  # (state) / postal code — changing only the prefecture or postal code still re-syncs the
+  # address to Stripe, so it must count as an address change. A submission is treated
   # as (re)entering the Japanese address only when one of these fields actually CHANGES compared
   # to the stored record — the Payments settings form echoes back every stored field on save, so
   # a value merely being present does not mean the seller touched their address. Older records
   # can still update unrelated fields (phone, payout frequency, ...) without a city — many legacy
   # Japanese records were created before the form collected one, and blocking every save until
   # they re-enter their address would strand them.
-  JAPAN_INDIVIDUAL_ADDRESS_FIELDS = %i[building_number building_number_kana street_address_kanji street_address_kana city city_kana].freeze
-  JAPAN_BUSINESS_ADDRESS_FIELDS = %i[business_building_number business_building_number_kana business_street_address_kanji business_street_address_kana business_city business_city_kana].freeze
+  JAPAN_INDIVIDUAL_ADDRESS_FIELDS = %i[building_number building_number_kana street_address_kanji street_address_kana city city_kana state zip_code].freeze
+  JAPAN_BUSINESS_ADDRESS_FIELDS = %i[business_building_number business_building_number_kana business_street_address_kanji business_street_address_kana business_city business_city_kana business_state business_zip_code].freeze
 
   attr_reader :compliance_params, :user
 
--- a/app/services/update_user_compliance_info.rb
+++ b/app/services/update_user_compliance_info.rb
@@ -9,6 +9,13 @@ class UpdateUserComplianceInfo
     street_address: [:country],
     business_street_address: [:business_country, :country],
   }.freeze
+  # The fields the Japanese address form submits alongside the city. If any of these arrive in a
+  # submission, the seller is (re)entering their Japanese address, so the city fields must be
+  # present too. Older records can still update unrelated fields (phone, tax id, ...) without a
+  # city — many legacy Japanese records were created before the form collected one, and blocking
+  # every update until they re-enter their address would strand them.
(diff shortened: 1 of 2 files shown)


gianfrancopiana:
@gumclaw This still strands the legacy records the PR is trying to preserve. The production request path does not send a sparse set of changed compliance fields:

1. [`SettingsPresenter`](https://github.com/antiwork/gumroad/blob/8c7ca09cd04b72235e60b0158992a3c8a1afa1db/app/presenters/settings_presenter.rb#L458-L498) returns every stored Japanese address field.
2. The Payments form initializes `user` from that [entire compliance object](https://github.com/antiwork/gumroad/blob/8c7ca09cd04b72235e60b0158992a3c8a1afa1db/app/javascript/pages/Settings/Payments/Show.tsx#L129-L145) and [submits all of `data.user`](https://github.com/antiwork/gumroad/blob/8c7ca09cd04b72235e60b0158992a3c8a1afa1db/app/javascript/pages/Settings/Payments/Show.tsx#L856-L876) on every save.
3. `japan_address_submitted_for?` then treats any watched field being [present](https://github.com/antiwork/gumroad/blob/8c7ca09cd04b72235e60b0158992a3c8a1afa1db/app/services/update_user_compliance_info.rb#L322-L329) as evidence that the address changed.

So a legacy JP record with stored street/building values and blank `city`/`city_kana` will return the new error even when the seller changes only their phone, payout frequency, or payout pause. The new non-address regression spec uses only `{ phone: ... }`, but that is not the payload shape produced by this endpoint, so it cannot catch the failure.

Please detect an actual address or validation-context change by comparing submitted values with `current_compliance_info`, and add a regression spec using the full UI-shaped payload with unchanged legacy address fields. The changed-field set should also include `city`/`city_kana` and their business equivalents; otherwise a direct request updating only one city field bypasses the matching-pair check and can sync a half-populated address to Stripe.

---

This review comment was drafted with AI assistance using gpt‑5.6

gumclaw:
Good catch — you're right that the presence check couldn't tell "the form echoed back the stored address" apart from "the seller re-entered it." Fixed in 01c8b129b:

- `japan_address_submitted_for?` now treats the address as (re)submitted only when a submitted field actually differs from `current_compliance_info`, or when the validation context changes (country switches to Japan, or the account flips to business) while a Japanese address is on the record. A full settings-form save that only changes the phone, payout frequency, or payout pause on a legacy no-city record passes through untouched.
- The watched field set now includes `city`/`city_kana` and `business_city`/`business_city_kana`, so a direct request that updates just one half of the city pair trips the matching-pair check instead of syncing a half-populated address to Stripe.
- Added the regression spec you asked for: a full UI-shaped payload (all stored address fields echoed back, `city`/`city_kana` blank, only the phone changed) on a legacy record now saves. Also added a spec rejecting a city-only submission without its kana pair.

`bundle exec rspec spec/services/update_user_compliance_info_spec.rb` — 30 examples, 0 failures. Rubocop clean on both files.

gianfrancopiana:
@gumclaw One address path remains: the watched sets omit `state`/`zip_code` and `business_state`/`business_zip_code`. On a legacy JP record without city fields, changing only the prefecture or postal code bypasses `japan_address_change_detected?`, but the service still persists and syncs those values to Stripe. Please add those four fields to the watched sets and cover individual/business postal-code-only updates.

---

This review comment was drafted with AI assistance using gpt‑5.6.

gumclaw:
Added in 687f75c87 — `state`/`zip_code` and `business_state`/`business_zip_code` are now in the watched sets, so a prefecture- or postal-code-only change on a legacy no-city record is treated as an address change and hits the city requirement like any other edit. Two new specs cover the individual and business postal-code-only cases.

`bundle exec rspec spec/services/update_user_compliance_info_spec.rb` — 32 examples, 0 failures; rubocop clean on both files.

gumclaw:
Added the missing symmetric business-path example Greptile pointed out: a business address change on a record that already carries a stored `business_city` + `business_city_kana` pair now has an accept-path spec, mirroring the individual case.

- `bundle exec rspec spec/services/update_user_compliance_info_spec.rb` — 33 examples, 0 failures
- `rubocop` clean on the spec file

PR #5787, opened 2026-07-10
Fix PPP card-country mismatch showing a generic error instead of the actionable message on client-confirm checkout

gumclaw:
Fixes #5784

Buyers hitting the purchasing-power-parity card-country check on the client-confirm checkout saw a bare "Sorry, something went wrong." instead of the actionable explanation. Two stacked bugs:

**1. Server: `error_message` arrived as `null`** — `Order::PreparePaymentIntentService` read `purchase.errors` *after* `Purchase::MarkFailedService` saved the purchase; the save re-runs validations, which clears the errors collection. Fixed by reading the message before marking failed (both `fail_all_purchases_when_any_errored` and `fail_purchases_with`).

**2. Frontend: failed-purchase receipt crashed on this lane** — the receipt's signup form reads `paymentMethod.cardParamsResult.cardParams`, but client-confirm methods carry a ConfirmationToken and have no `cardParamsResult`, so the render threw (`Cannot read properties of undefined (reading 'cardParams')`) and no error UI appeared at all. Guarded in `Receipt.tsx` and `TemporaryLibrary.tsx`.

Strengthened the existing PPP-block spec to assert the response carries the actionable `error_message`, so the server half can't silently regress.

**Manual QA on a preview app** (real Indonesian IP + India-issued test card `4000003560000008`): buyer now sees the full PPP message with correct server-side decision (`card_country="IN"`, `ip_country="Indonesia"`); purchase recorded as failed with `ppp_card_country_not_matching`; cart stays retryable. Buyer-side recording attached in a comment below.

### DIFF
--- a/app/javascript/components/Checkout/Receipt.tsx
+++ b/app/javascript/components/Checkout/Receipt.tsx
@@ -321,7 +321,11 @@ export const Receipt = ({
             createAccountData={{
               email: state.email,
               cardParams:
-                state.status.paymentMethod.type === "not-applicable" || state.status.paymentMethod.type === "saved"
+                // Client-confirm methods live in a Stripe ConfirmationToken and carry no
+                // cardParamsResult — reading it crashes the whole receipt view (#5784).
+                state.status.paymentMethod.type === "not-applicable" ||
+                state.status.paymentMethod.type === "saved" ||
+                state.status.paymentMethod.type === "payment-element-client-confirm"
                   ? null
                   : state.status.paymentMethod.cardParamsResult.cardParams,
             }}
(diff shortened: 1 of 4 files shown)


nyomanjyotisa:
### Manual QA — PPP mismatch now shows the actionable message

https://github.com/user-attachments/assets/9de4af39-f035-46b7-a86d-2ca12d63d848

nyomanjyotisa (approved):

gumclaw:
Thanks for the QA pass — the recording confirms the buyer now sees the full PPP explanation instead of the generic error, matching what we saw on the preview app. CI is still finishing (build images + Greptile pending); I'll take the PR through the rest of the completion flow once everything is green.

PR #5796, opened 2026-07-10
Record deterministic processor rejections as processor_invalid_request instead of stripe_unavailable

gumclaw:
## What

Splits deterministic processor rejections out of the `stripe_unavailable` error code. This is a pure observability change — retry behavior and buyer-facing messages are byte-for-byte unchanged.

- New `PurchaseErrorCode::PROCESSOR_INVALID_REQUEST` (`"processor_invalid_request"`), stamped wherever a `ChargeProcessorInvalidRequestError` was previously recorded as `stripe_unavailable`/`paypal_unavailable`:
  - `Order::PreparePaymentIntentService#create_unconfirmed_intent` (the client-confirm prepare path)
  - `Charge::CreateService#create_payment_intent_or_charge!`
  - `Purchase#confirm_charge_intent!`, `Purchase#prepare_chargeable`, `Purchase#with_charge_processor_error_handler`
  - `CreditCard.create`
- The processor's own error code (e.g. Stripe's `payment_intent_invalid_parameter`) is now persisted into `stripe_error_code` at those sites, via a new `ChargeProcessorInvalidRequestError#processor_error_code` reader (backed by exposing `original_error` on `GumroadRuntimeError`). Previously the column was left blank, erasing the forensic trail.
- The new code is included in both `PAYMENT_ERROR_CODES` and `is_temporary_network_error?`, so `has_payment_error?`, subscription-renewal retries, and preorder retries all behave exactly as they did when these failures were labeled `stripe_unavailable`. Invalid-request rejections are deterministic for a given request, but they usually come from a short-lived deploy bug — keeping the retries lets renewals self-heal once the bug is reverted. Tightening retry policy for this class is deliberately left as a separate decision.

## Why

`stripe_unavailable` has always meant "Stripe is down / unreachable" — a transient network problem. But the combined `rescue ChargeProcessorInvalidRequestError, ChargeProcessorUnavailableError` blocks also stamped it for synchronous invalid-request rejections, where Stripe is healthy and *our request* is malformed. That conflation:

- sent the antiwork/gumroad-private#1026 investigation toward "outage" first, and let a client-confirm regression (~$20K attempted GMV over 3 days) hide inside the platform-wide `stripe_unavailable` baseline — the failures were deterministic, but the DB made them look like ambient Stripe weather;
- left `stripe_error_code` blank, so the actual Stripe rejection reason was unrecoverable from the purchase row and had to be reconstructed with live console probes.

After this change, `stripe_unavailable` volume is a clean Stripe-health signal, and `processor_invalid_request` has a near-zero baseline — any sustained nonzero rate is by definition a regression on our side, which makes it directly alertable.

Follow-up filed from antiwork/gumroad-private#1026 (see the root-cause comment there; details stay in the private issue).

## Test plan

- New spec: `Order::PreparePaymentIntentService` — Stripe synchronously rejects the deferred intent create → purchases fail with `processor_invalid_request` and `stripe_error_code: payment_intent_invalid_parameter` (the exact antiwork/gumroad-private#1026 shape).
- New spec: `Charge::CreateService` — invalid-request rejection stamps the new code, keeps Stripe's error code, and `has_payment_network_error?` stays true (retry behavior preserved).
- New spec: `ChargeProcessorInvalidRequestError#processor_error_code` — Stripe code extraction, message-only raises (Braintree/PayPal sites), and wrapped errors without `#code` all return safely.
- New spec: `Purchase#has_payment_network_error?` returns true for the new code (retries preserved).
- Updated: `purchase_process_spec.rb` invalid-request contexts now expect `PROCESSOR_INVALID_REQUEST`; `credit_card_spec.rb` pins the new code.
- Ran locally: `prepare_payment_intent_service_spec.rb`, `charge/create_service_spec.rb`, `credit_card_spec.rb`, `purchase/purchase_process_spec.rb`, the new error spec — all green (135 examples in the two big files, 0 failures).
- No UI changes (buyer-facing messages untouched).

### DIFF
--- a/app/business/payments/charging/errors/charge_processor_invalid_request_error.rb
+++ b/app/business/payments/charging/errors/charge_processor_invalid_request_error.rb
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 class ChargeProcessorInvalidRequestError < ChargeProcessorError
+  # The processor's own error code (e.g. Stripe's "payment_intent_invalid_parameter"), when
+  # the wrapped error exposes one. Rescue sites persist this into stripe_error_code so a
+  # failed purchase records *why* the processor rejected the request instead of leaving the
+  # column blank.
+  def processor_error_code
+    original_error.code if original_error.respond_to?(:code)
+  end
 end
--- a/app/models/credit_card.rb
+++ b/app/models/credit_card.rb
@@ -72,7 +72,14 @@ class CreditCard < ApplicationRecord
       end
 
       credit_card.save!
-    rescue ChargeProcessorInvalidRequestError, ChargeProcessorUnavailableError => e
+    rescue ChargeProcessorInvalidRequestError => e
+      # The processor rejected our request as malformed — a deterministic failure on our side,
+      # not an outage. Record it distinctly so callers don't treat it as transient.
+      logger.error("Error while persisting card with #{credit_card.charge_processor_id}: #{e.message} - card visual: #{credit_card.visual}")
+      credit_card.errors.add(:base, "There is a temporary problem, please try again (your card was not charged).")
+      credit_card.error_code = PurchaseErrorCode::PROCESSOR_INVALID_REQUEST
+      credit_card.stripe_error_code = e.processor_error_code if credit_card.stripe_error_code.blank?
+    rescue ChargeProcessorUnavailableError => e
       logger.error("Error while persisting card with #{credit_card.charge_processor_id}: #{e.message} - card visual: #{credit_card.visual}")
       credit_card.errors.add(:base, "There is a temporary problem, please try again (your card was not charged).")
       credit_card.error_code = credit_card.charge_processor_unavailable_error
--- a/app/models/credit_card.rb
+++ b/app/models/credit_card.rb
@@ -74,7 +74,8 @@ class CreditCard < ApplicationRecord
       credit_card.save!
     rescue ChargeProcessorInvalidRequestError => e
       # The processor rejected our request as malformed — a deterministic failure on our side,
-      # not an outage. Record it distinctly so callers don't treat it as transient.
+      # not an outage. Record it under its own code so a code regression shows up in monitoring
+      # instead of hiding inside Stripe-outage noise. Retry behavior is unchanged.
       logger.error("Error while persisting card with #{credit_card.charge_processor_id}: #{e.message} - card visual: #{credit_card.visual}")
       credit_card.errors.add(:base, "There is a temporary problem, please try again (your card was not charged).")
       credit_card.error_code = PurchaseErrorCode::PROCESSOR_INVALID_REQUEST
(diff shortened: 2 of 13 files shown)


gumclaw:
Adversarial review round complete. One material finding, fixed in the second commit:

**P1 (fixed): the first cut excluded `processor_invalid_request` from `is_temporary_network_error?`, which changed subscription-renewal behavior** — a short-lived deploy bug producing invalid-request rejections would have sent false "card declined" emails (`subscription_card_declined`) and terminated subscriptions at `terminate_by` with no retry to self-heal after the bug was reverted. Same shape for preorders (`ChargePreorderWorker` → false decline email + auto-cancel). The new code is now included in `is_temporary_network_error?`, making this PR purely an observability split with zero retry/comms changes. Whether invalid-request failures deserve a different retry policy or comms path is a real question, but it's a product decision to make separately — not something to smuggle in with a relabeling.

Also addressed from review: unit spec for `processor_error_code` nil-safety (message-only raises at Braintree/PayPal call sites, wrapped errors without `#code`), and `credit_card_spec` now pins the new code instead of `be_present`.

Verified clean by review: no code in app/ or lib/ pattern-matches the literal `stripe_unavailable`; fraud/blocking logic (`FRAUD_RELATED_ERROR_CODES`, `ban_buyer_on_fraud_related_error_code!`) only matches `card_declined_*` codes so the newly-persisted Stripe codes can't trip it; `formatted_error_code` falls through to a titleized string for the new Stripe codes (cryptic but accurate).

slavingia:
Yes! Ideally this is only for when Stripe is actually unavailable ≈never

gumclaw:
Agreed — pushed 662e35ce tightening the prepare path so `stripe_unavailable` only gets stamped when Stripe is actually unreachable:

- `fail_purchases_with` (the catch-all for missing tokens, blocked carts, unexpected exceptions) no longer defaults to `stripe_unavailable` — it fills the generic `processing_error` instead, with identical retry semantics.
- A real connection failure (`ChargeProcessorUnavailableError` from the intent create, or `Stripe::APIConnectionError`/`APIError` on the ConfirmationToken retrieve) is the only thing that now records `stripe_unavailable` on this path.
- An invalid-request rejection on the ConfirmationToken retrieve now records `processor_invalid_request` + Stripe's error code too, same as the intent create.

The older rescue sites elsewhere (`Purchase`, `Charge::CreateService`, `CreditCard`) already only stamp `stripe_unavailable` on `ChargeProcessorUnavailableError`, which the Stripe wrapper raises solely for `APIConnectionError`/`APIError` — so those were already "Stripe is genuinely unreachable" after the first commit. Specs cover all three shapes; the full prepare service spec is green locally (42 examples, 0 failures).

gumclaw:
@xrav3nz over to you — CI fully green (76/76), fable adversarial review complete with the one P1 finding fixed (retry semantics preserved; this is now a pure observability split), no bot findings to triage. Payments surface, so no auto-merge — merge at your discretion.

PR #5809, opened 2026-07-11
Multiple storefronts per account ("many Gumroads, one payout")

gumclaw:
## What

v1 of multiple storefronts per account: the account is the person (login, identity, payouts), and each storefront is "a Gumroad" — its own brand name, username, and public page, all paying out through the one account.

Direction per antiwork/gumroad-private#1045: storefronts are the core interface, not a managed list — you switch between them like team accounts, and you create products in the context of the storefront you're in.

## Plan

- [x] `Storefront` + `StorefrontProduct` models and migration (username shares the public namespace with account usernames)
- [ ] Public page resolution: `<storefront-username>.gumroad.com` renders the storefront's profile + its products
- [ ] Storefront switcher in the nav (same pattern as team switcher) with inline "Create a new Gumroad"
- [ ] Product creation scoped to the active storefront
- [ ] Account's existing username/profile acts as the first storefront — no migration of existing data
- [ ] Everything behind the `multiple_storefronts` feature flag
- [ ] Screenshots

Draft until clickable end-to-end with screenshots.

### DIFF
--- a/app/models/storefront.rb
+++ b/app/models/storefront.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# A storefront is a separately-branded "Gumroad" that lives under one account: the account (User)
+# is the person — login, identity, payout details — while each storefront carries its own public
+# name, username, and bio. This lets a creator run multiple brands that all pay out through the
+# same account, instead of juggling separate accounts with duplicated bank details.
+#
+# v1 keeps this additive: the account's own username/profile keeps working unchanged as the first
+# brand, and extra storefronts are created on top. Products are assigned to at most one storefront;
+# unassigned products keep showing on the account's main profile.
+class Storefront < ApplicationRecord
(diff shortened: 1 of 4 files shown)


gumclaw:
Status: found the latest gumroad-private#1045 direction pivots from storefront switching to first-class pages. I’ve updated the branch in that direction locally: new seller-owned pages model/API/public routes are implemented and the focused backend specs are green (18 examples). Next: settings UI affordance + real-page screenshots before pushing.

gumclaw:
Closing this — the direction on antiwork/gumroad-private#1045 moved away from a storefront entity. The feature is now "create a new brand account" (a real new `User` with its own email, username, and brand name, with the creator auto-admined via team memberships), so the `Storefront`/`StorefrontProduct` models here are the wrong shape. A fresh PR will follow from main once the new scope is confirmed.

PR #5815, opened 2026-07-11
Persist purchase url_parameters so PayPal sale webhooks include url_params

gumclaw:
## What

Persists a purchase's `url_parameters` in a new `purchase_url_parameters` table (one row per purchase, `json` column, autosaved association) instead of keeping them in a request-lifetime `attr_accessor`. No call sites change — `purchase.url_parameters` and `purchase.url_parameters=` keep the same interface, but the value now survives a database reload.

## Why

Sellers who deliver products via ping webhooks pass custom query params on the product URL (e.g. `?discord_id=x&plan=y`) and expect them back in the sale ping's `url_params` field. That worked for Stripe card purchases but was **always empty for PayPal purchases**.

Root cause: the sale webhook is enqueued from the `any => successful` state-machine transition, which reads `url_parameters` off whatever Purchase instance runs the transition. Stripe's synchronous path marks success on the same in-memory instance checkout built, so the accessor still holds the params. The PayPal path reloads the purchase (`Purchase.find_by_external_id` in `PaypalChargeProcessor`, PayPal capture/webhook event handling, deferred status syncs) and marks success on the fresh copy — where the `attr_accessor` is `nil`, so `payload[:url_params]` is omitted from the ping.

Persisting the value at checkout means every path that later fires the webhook — PayPal captures, webhook-driven syncs, refund/dispute/dispute-won pings that also pass `purchase.url_parameters` — reads the real value. Refund and dispute pings always fire off reloaded instances, so those were silently dropping the params for every payment method; this fixes them too.

A dedicated table is used rather than `purchases.json_data` because that column is a `varchar(255)` (too small for arbitrary buyer-supplied params) and the `purchases` table is too large to ALTER on the deploy path.

Refs: antiwork/gumroad-private#1049

## Test plan

- New regression specs: set `url_parameters`, save, reload the purchase via `Purchase.find`, fire `send_notification_webhook` / `send_refunded_notification_webhook`, and assert the enqueued `PostToPingEndpointsWorker` job carries the params — exactly the PayPal-shaped lifecycle that was broken.
- Verified the persistence round-trip locally against MySQL with `rails runner`: writer builds the associated record, value survives `Purchase.find`, and setting `nil` marks the record for destruction.
- Migration run locally; `db/schema.rb` updated from the actual migration run.
- `rubocop` clean on all changed files.
- Note: the surrounding examples in `purchase_notification_spec.rb` fail identically on unmodified `origin/main` in this local worktree (missing merchant-account/VCR seed data) — pre-existing environment issue, not introduced by this change. CI is the authoritative run.

## Before/After

Not applicable — backend-only change; no user-facing rendering is affected. The observable change is the webhook payload: PayPal sale pings (and refund/dispute pings) now include `url_params` when the product URL carried custom query params.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -70,12 +70,6 @@ class Purchase < ApplicationRecord
   attr_json_data_accessor :custom_fee_per_thousand
   attr_json_data_accessor :last_content_page_id
   attr_json_data_accessor :default_offer_code_id
-  # Custom query params the buyer had on the product URL (e.g. ?discord_id=x), minus
-  # reserved ones. Persisted here (not an attr_accessor) because the "purchase
-  # successful" webhook can fire from a freshly loaded Purchase — PayPal captures,
-  # webhook-driven status syncs — long after the checkout request that knew the
-  # params has ended. Sellers rely on these reaching the sale ping as `url_params`.
-  attr_json_data_accessor :url_parameters
   # Buyer-presentment purchases only: the canonical USD gross the dispute-loss balance
   # debit actually booked. Snapshotted at debit time so the dispute-won re-credit books
   # exactly the same amount, even when refunds land between the debit and the win.
@@ -127,6 +121,7 @@ class Purchase < ApplicationRecord
   has_one :recommended_purchase_info, dependent: :destroy
   has_one :purchase_wallet_type
   has_one :purchase_payment_flow, dependent: :destroy, validate: false
+  has_one :purchase_url_parameter, autosave: true, dependent: :destroy
   has_one :purchase_offer_code_discount
   has_one :purchasing_power_parity_info, dependent: :destroy
   has_one :upsell_purchase, dependent: :destroy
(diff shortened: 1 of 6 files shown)


gumclaw:
Addressed the test-coverage gap from Greptile's review in cdfab4f: added a parallel round-trip spec for `send_refunded_notification_webhook` (set `url_parameters`, save, reload via `Purchase.find`, assert the enqueued job carries the persisted params).

On the dispute paths in `charge/disputable.rb`: both call sites read the same `purchase.url_parameters` accessor on purchases loaded fresh from the database, so the persistence round trip they depend on is exactly what the two model-level specs now prove. A full dispute-event spec would need the whole dispute/flow-of-funds fixture stack to re-test the same one-line mechanism, so I left those to the existing dispute specs.

Rubocop clean on the changed spec. Note: the baseline `#send_refunded_notification_webhook` spec (untouched by this PR) fails identically in my local worktree at `create(:purchase)` — the same pre-existing environment issue called out in the PR body. CI is the authoritative run.

slavingia:
Get to green

gumclaw:
On it. Root cause of the red shards: `Purchase::CreateService` mass-assigns the raw `url_parameters` string when building the purchase, then assigns the parsed value (nil for invalid JSON) right after. Clearing via `mark_for_destruction` didn't discard a never-saved associated record, so the reader kept returning the stale raw string (`"{{"` instead of nil).

Fixed in 182f207: blank assignment now drops an unsaved record outright (persisted ones still get marked for destruction), and the reader ignores records marked for destruction. Reproduced the exact failing sequence locally before and after the fix, and added accessor specs (`spec/models/purchase_url_parameter_spec.rb`) covering the raw-then-parsed sequence and the persisted clear/reload paths. Auto-merge is armed — will merge on green.

gumclaw:
Greptile's remaining finding on the `url_parameters=` setter (value lost when reassigned after clearing on a persisted record) is fixed in follow-up #5817: the setter now reloads the association when the existing record is marked for destruction, so the new value lands on a fresh copy instead of the one autosave is about to delete. Includes a load-bearing regression spec — fails on the pre-fix setter, passes with it.

PR #5819, opened 2026-07-12
Reorder web Dockerfile layers so the full-tree COPY comes last

gumclaw:
## What

Reorders `docker/web/Dockerfile` so the full source-tree `COPY . $APP_DIR/` is the last cache-sensitive layer instead of the third. Refs #5802.

New layer order:

1. User/dir setup (unchanged)
2. ImageMagick policy edits — independent of the source tree
3. GeoIP2-City download — staged to `/usr/local/share/geoip`, moved into `$APP_DIR/lib` after the source COPY
4. `bundle install` — copies only `Gemfile`, `Gemfile.lock`, and `.ruby-version` first, so the layer is reused unless dependencies change (`.ruby-version` is needed because the Gemfile declares `ruby file: ".ruby-version"`)
5. `COPY . $APP_DIR/` (full tree)
6. GeoIP mmdb move + permissions/chown (must follow the tree COPY)

## Why

Previously the full-tree COPY sat above the GeoIP download, ImageMagick config, and `bundle install`, so **every commit invalidated all of them** — including an external MaxMind network fetch and a full gem install — even when neither the dependencies nor anything those layers use had changed. With this ordering, a source-only commit reuses the cached download and dependency layers and only re-runs the COPY and the permissions layer, cutting preview/deploy image build time (#5802).

No behavior change to the final image: same files end up in the same places (`$APP_DIR/lib/GeoIP2-City.mmdb`, bundled gems in `$BUNDLE_PATH`), same ENV/WORKDIR/CMD.

## Test plan

- `Gemfile`/`Gemfile.lock`/`.ruby-version` verified present at the build-context root (repo root, per the `Makefile` `--file docker/web/Dockerfile` invocation)
- Verified the Gemfile has no `path:`/local-gem references, so `bundle install` needs only the three copied manifest files
- Build verification via the deploy pipeline: first build after merge populates the cache; a subsequent source-only commit should show `CACHED` for the ImageMagick/GeoIP/bundle layers
- Note: `docker/web/Dockerfile.test` (used by CI `tests.yml`) is untouched — this file is only used by the Makefile/deploy image build

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -16,24 +16,6 @@ RUN useradd -ms /bin/bash app \
   && mkdir -p $APP_DIR/log \
   && mkdir -p $APP_DIR/tmp
 
-COPY . $APP_DIR/
-
-# Dowload latest GeoIP DB
-RUN cd /tmp \
-  && curl -J -L -o GeoIP2-City.tar.gz https://download.maxmind.com/app/geoip_download\?edition_id\=GeoIP2-City\&license_key\=fxN7iBPUYfTS76V5\&suffix\=tar.gz \
-  && tar xzf GeoIP2-City.tar.gz \
-  && mv GeoIP2-City_* GeoIP2-City \
-  && mv GeoIP2-City/GeoIP2-City.mmdb $APP_DIR/lib/GeoIP2-City.mmdb \
-  && rm -rf /tmp/*
-
-# TODO: look into solutions that prevent image bloating and/or avoids --squash
-RUN chmod -R 755 $BUNDLE_PATH \
-  && chmod -R 755 $APP_DIR \
-  && chmod -R 770 $APP_DIR/tmp \
-  && chmod -R 770 $APP_DIR/log \
-  && chown -R app:app $APP_DIR \
-  && chown -R root:app $BUNDLE_PATH
-
 # Update ImageMagick configuration
 # Set memory limit to 1GiB
 RUN sed -r 's/(name="memory" value=")[^"]+"/\11GiB"/' -i /etc/ImageMagick-6/policy.xml \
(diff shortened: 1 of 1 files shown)


slavingia:
Prove this works

gumclaw:
Proved it with real builds. Ran both the old and new Dockerfile locally against the production base image (`gumroad/web_base:0427a7a` from ECR, linux/amd64), did a cold build of each, then appended a comment to `app/models/user.rb` to simulate a source-only commit and rebuilt.

**Rebuild after a source-only change — layer by layer:**

| Layer | Old order | New order |
|---|---|---|
| useradd/mkdir setup | CACHED | CACHED |
| ImageMagick policy edits | re-ran | **CACHED** |
| GeoIP2-City download (MaxMind network fetch, 67MB) | **re-ran (re-downloaded)** | **CACHED** |
| `bundle install` | **re-ran** | **CACHED** |
| `COPY . /app/` | re-ran (expected) | re-ran (expected) |
| mmdb move into `/app/lib` | n/a | re-ran (0.3s) |
| chmod/chown | re-ran (expected) | re-ran (expected) |

Raw BuildKit output, old order (`build-old-2.log`):

```
#7 [3/8] COPY . /app/                          DONE 1.3s
#8 [4/8] RUN ... download.maxmind.com ...      DONE 2.5s   ← re-fetched 67MB from MaxMind
#9 [5/8] RUN chmod/chown ...                   DONE 17.7s
#10 [6/8] RUN sed ... ImageMagick policy       DONE 0.4s
#11 [7/8] RUN ... bundle install               re-ran ("Bundle complete! 169 Gemfile dependencies, 340 gems")
```

New order (`build-new-2.log`):

```
#7  [ 3/10] RUN sed ... ImageMagick policy     CACHED
#8  [ 4/10] RUN ... download.maxmind.com ...   CACHED
#9  [ 5/10] COPY Gemfile Gemfile.lock .ruby-version /app/   CACHED
#10 [ 6/10] RUN ... bundle install             CACHED
#11 [ 7/10] COPY . /app/                       DONE 0.8s
#12 [ 8/10] RUN mv .../GeoIP2-City.mmdb ...    DONE 0.3s
#13 [ 9/10] RUN chmod/chown ...                DONE 33.2s
```

Exactly as claimed: with the new ordering a source-only commit reuses the MaxMind download, the ImageMagick config, and the bundle layer; the old ordering re-executed all three on every commit (including the external network fetch). Locally the bundle layer is cheap because gems are pre-baked into the base image and my connection to MaxMind is fast — on Buildkite the avoided work is larger (dependency resolution plus pushing the invalidated layers to ECR).

**Final image equivalence** (old-order image vs new-order image, both built from the same touched tree):

- `/app/lib/GeoIP2-City.mmdb`: identical size (127,025,609 bytes), owner `app:app`, in both
- `/usr/local/share/geoip`: absent in both (staging dir cleaned up)
- Permissions identical: `/app` 755, `/app/tmp` 770, `/app/log` 770, all `app:app`
- Same gem count in `/bundle_path` (404 dirs); `bundle check` inside the new image: "The Gemfile's dependencies are satisfied"

One caveat from the run: wall-clock deltas here are dominated by the `chmod -R`/`chown -R` layer (17–59s of I/O noise in my local VM), which both orderings must re-run after a COPY. The win is the three layers above it going CACHED — most importantly not hitting MaxMind on every deploy build.

PR #5821, opened 2026-07-12
Add permanent preview_qa rake tasks for seeding QA state on preview apps

gumclaw:
Closes the remaining piece of #5806: permanent, environment-guarded QA seed tasks so preview-app QA no longer needs temporary param-gated seed hooks shipped inside feature PRs ("TEMP: revert before merge" commits).

## What

New `lib/tasks/preview_qa.rake` namespace, restricted to preview apps, development, and the test suite. The guard blocks production and shared staging (`staging.gumroad.com`): shared staging runs the staging Rails env too, so an env check alone isn't enough — the tasks additionally require the `BRANCH_DEPLOYMENT` flag that already distinguishes per-branch preview deploys from shared staging. The namespace is skipped entirely when the file loads in a disallowed environment, and every task re-checks at run time.

- `preview_qa:backdate_purchase[purchase_id,days]` — backdates a purchase's `created_at`/`succeeded_at`; defaults to one billing period + 1 day for subscription purchases so `RecurringChargeWorker` sees the subscription as due. When the purchase belongs to a subscription, every successful charge on that subscription is shifted by the same offset — the renewal-due checks read the latest successful charge, so shifting only an older purchase wouldn't flip them.
- `preview_qa:clear_mandate[subscription_id]` — removes `stripe_setup_intent_id` from the card charged for a subscription, to QA the Indian-card missing-e-mandate path from #5795 (other `json_data` keys are preserved).
- `preview_qa:seed_dead_job[WorkerClass,args...]` — seeds a job into the Sidekiq dead set without firing death handlers.
- `preview_qa:run_worker[WorkerClass,args...]` — runs a Sidekiq worker inline (e.g. force a renewal charge attempt).
- `preview_qa:inspect_subscription[subscription_id]` — read-only snapshot for verifying state after a QA run: renewal timing (`overdue_for_charge?`, end of current period), the charged card's e-mandate linkage, and the recent purchase history with charge ids. Mutates nothing, so it runs on the default read-only console connection.

Record identifiers accept either database ids or external ids. Worker tasks verify the constant is actually a `Sidekiq::Job` before touching it, and rake's string args are cast to integers/booleans.

Also documents preview-app console access (linking [gumroad-deployment/docs/ssh.md](https://github.com/antiwork/gumroad-deployment/blob/main/docs/ssh.md#connect-to-rails-console)) and the new tasks in `docs/deploying.md`, closing the "not linked from the preview-app workflow docs" gap.

## Usage

Interactive writable console (from the deployment repo; `BRANCH` is an env var and `-w` gives a writable connection — the seeding tasks write):

```shell
cd nomad/staging/deploy_branch && BRANCH=<branch name> ./console.sh -w
# then, at the console prompt:
system 'bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"'
```

Or as a one-shot (`COMMAND` runs as the container command, so it must be a real `bundle exec rake` invocation):

```shell
cd nomad/staging/deploy_branch && \
  BRANCH=<branch name> \
  COMMAND='bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"' \
  ./console.sh -w
```

## Tests

`spec/lib/tasks/preview_qa_spec.rb` — 21 examples, 0 failures locally:

- environment guard aborts every task in production and on shared staging (staging env without `BRANCH_DEPLOYMENT`), and allows preview apps (staging env with the flag)
- backdating: default period derivation, explicit days, numeric-id lookup, nil `succeeded_at` preserved, missing-record/missing-days aborts
- mandate clearing: removes only the mandate key, no-op message when absent, aborts without a chargeable card
- dead-job seeding: payload shape + `notify_failure: false`, rejects non-worker/unknown classes
- inline worker run: type-cast args, rejects non-worker classes / missing class
- subscription inspection: prints renewal timing, mandate linkage, and purchase history; reports a missing card without aborting; aborts on unknown records

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -198,29 +198,48 @@ Deployments are removed automatically when the associated branch is deleted in t
 
 ### Rails console on a preview app
 
-Preview apps have Rails console access — see [Connect to Rails console](https://github.com/antiwork/gumroad-deployment/blob/main/docs/ssh.md#connect-to-rails-console) in the deployment repo. The `console.sh` script there resolves the preview instance and opens `rails c` in the running container; it also supports `COMMAND=...` for one-shot commands and a `-w` flag for a writable database connection (the default is a read-only replica).
+Preview apps have Rails console access — see [Connect to Rails console](https://github.com/antiwork/gumroad-deployment/blob/main/docs/ssh.md#connect-to-rails-console) in the deployment repo. The `console.sh` script there resolves the preview instance from the `BRANCH=<branch name>` environment variable (it falls back to the deployment repo's own checked-out branch, so always set it explicitly) and opens `rails c` in the running container. It also supports `COMMAND=...` for one-shot commands — the value is executed directly as the container command, so it must be a runnable program like `bundle exec rake "..."`, not a bare Ruby snippet — and a standalone `-w` flag for a writable database connection (the default is a read-only replica).
 
 ### Seeding QA state with `preview_qa` rake tasks
 
 Use the permanent tasks in `lib/tasks/preview_qa.rake` to seed edge-case state on a preview app instead of adding temporary, param-gated seed hooks to your PR ("TEMP: revert before merge" commits). The tasks only run on preview apps, development, and the test suite — they are unavailable in production and on shared staging (staging.gumroad.com), where mutating records would interfere with other people's testing.
 
-Run them through the console one-shot path with a writable connection, for example:
+Run them through the preview app's Rails console (see the section above for how `console.sh` resolves the branch and the `-w` writable flag — the seeding tasks write, so they need `-w`).
+
+The simplest path is to open an interactive writable console and shell out to rake from the prompt (rake tasks aren't preloaded in a console session, so `bundle exec rake` in a subprocess is the entrypoint):
+
+```shell
+cd nomad/staging/deploy_branch && BRANCH=<branch name> ./console.sh -w
+# then, at the console prompt:
+system 'bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"'
+```
+
+Alternatively, run a task as a one-shot. `COMMAND` is executed directly in the container as a shell command (not evaluated as Ruby), so it must be a full `bundle exec rake` invocation:
+
+```shell
+cd nomad/staging/deploy_branch && \
+  BRANCH=<branch name> \
+  COMMAND='bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"' \
+  ./console.sh -w
+```
+
+The available tasks (all need the writable connection except `inspect_subscription`, which is read-only):
 
 ```shell
 # Backdate a subscription purchase past its billing period so a renewal charge is due
-COMMAND='Rake::Task["preview_qa:backdate_purchase"].invoke("<purchase external_id>")' ./console.sh -w <branch>
+bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"
 
 # Remove the e-mandate linkage from the card charged for a subscription (Indian card QA)
-COMMAND='Rake::Task["preview_qa:clear_mandate"].invoke("<subscription external_id>")' ./console.sh -w <branch>
+bundle exec rake "preview_qa:clear_mandate[<subscription external_id>]"
 
 # Seed a job into the Sidekiq dead set (morgue)
-COMMAND='Rake::Task["preview_qa:seed_dead_job"].invoke("RecurringChargeWorker", "123")' ./console.sh -w <branch>
+bundle exec rake "preview_qa:seed_dead_job[RecurringChargeWorker,123]"
 
 # Run a worker inline, e.g. to force a renewal charge attempt
-COMMAND='Rake::Task["preview_qa:run_worker"].invoke("RecurringChargeWorker", "123")' ./console.sh -w <branch>
+bundle exec rake "preview_qa:run_worker[RecurringChargeWorker,123]"
 
 # Inspect a subscription after a QA run: renewal timing, mandate linkage, recent charges (read-only)
-COMMAND='Rake::Task["preview_qa:inspect_subscription"].invoke("<subscription external_id>")' ./console.sh <branch>
+bundle exec rake "preview_qa:inspect_subscription[<subscription external_id>]"
 ```
 
 Record ids can be passed as either database ids or external ids. If a QA scenario you need isn't covered, add a task to the namespace (with specs) rather than shipping a temporary hook in your feature PR.
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -11,6 +11,8 @@
   - [Logs](#logs)
   - [Hotfixing workers only](#hotfixing-workers-only)
 - [Deploying to a preview app](#deploying-to-a-preview-app)
+  - [Rails console on a preview app](#rails-console-on-a-preview-app)
+  - [Seeding QA state with `preview_qa` rake tasks](#seeding-qa-state-with-preview_qa-rake-tasks)
 - [Deploying to staging](#deploying-to-staging)
 
 ---
@@ -193,3 +195,29 @@ Adding the label triggers a Buildkite build that deploys the branch. Each subseq
 The preview app URL is posted on the pull request as a GitHub deployment: look for the "View deployment" button (and the Deployments section), which shows the deploy in progress and links to the running app once it is ready.
 
 Deployments are removed automatically when the associated branch is deleted in the repository.
+
+### Rails console on a preview app
+
+Preview apps have Rails console access — see [Connect to Rails console](https://github.com/antiwork/gumroad-deployment/blob/main/docs/ssh.md#connect-to-rails-console) in the deployment repo. The `console.sh` script there resolves the preview instance and opens `rails c` in the running container; it also supports `COMMAND=...` for one-shot commands and a `-w` flag for a writable database connection (the default is a read-only replica).
+
+### Seeding QA state with `preview_qa` rake tasks
+
+Use the permanent tasks in `lib/tasks/preview_qa.rake` to seed edge-case state on a preview app instead of adding temporary, param-gated seed hooks to your PR ("TEMP: revert before merge" commits). The tasks are unavailable in production.
+
+Run them through the console one-shot path with a writable connection, for example:
+
+```shell
+# Backdate a subscription purchase past its billing period so a renewal charge is due
+COMMAND='Rake::Task["preview_qa:backdate_purchase"].invoke("<purchase external_id>")' ./console.sh -w <branch>
+
+# Remove the e-mandate linkage from the card charged for a subscription (Indian card QA)
+COMMAND='Rake::Task["preview_qa:clear_mandate"].invoke("<subscription external_id>")' ./console.sh -w <branch>
+
+# Seed a job into the Sidekiq dead set (morgue)
+COMMAND='Rake::Task["preview_qa:seed_dead_job"].invoke("RecurringChargeWorker", "123")' ./console.sh -w <branch>
+
+# Run a worker inline, e.g. to force a renewal charge attempt
+COMMAND='Rake::Task["preview_qa:run_worker"].invoke("RecurringChargeWorker", "123")' ./console.sh -w <branch>
+```
+
+Record ids can be passed as either database ids or external ids. If a QA scenario you need isn't covered, add a task to the namespace (with specs) rather than shipping a temporary hook in your feature PR.
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -218,6 +218,9 @@ COMMAND='Rake::Task["preview_qa:seed_dead_job"].invoke("RecurringChargeWorker", 
 
 # Run a worker inline, e.g. to force a renewal charge attempt
 COMMAND='Rake::Task["preview_qa:run_worker"].invoke("RecurringChargeWorker", "123")' ./console.sh -w <branch>
+
+# Inspect a subscription after a QA run: renewal timing, mandate linkage, recent charges (read-only)
+COMMAND='Rake::Task["preview_qa:inspect_subscription"].invoke("<subscription external_id>")' ./console.sh <branch>
 ```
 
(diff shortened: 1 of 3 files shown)


lib/tasks/preview_qa.rake
```diff
@@ -0,0 +1,151 @@
...
+#
+# Every task hard-aborts in production; the guard is intentionally belt-and-braces (the whole
+# namespace is skipped when the file loads in production, and each task re-checks at run time in
+# case the file is ever loaded by other means).
+
+module PreviewQa
+  module_function
+
+  def ensure_not_production!
+    abort "preview_qa tasks are for preview/staging QA only and cannot run in production." if Rails.env.production?
```
ershad: This blocks production but not shared staging, and acceptance criterion #2 is "unavailable in production/**staging proper**". On `staging.gumroad.com` `Rails.env.production?` is false, so every mutating task here would still run against the environment QA and reviewers share.

Preview apps run in the staging Rails env but set `ENV["BRANCH_DEPLOYMENT"] == "true"` — that flag is already how the app tells a per-branch deploy apart from shared staging (see `config/initializers/prefix_for_branch_apps_es_index.rb` and `config/domain.rb`). So the safe guard is to allow *only* preview apps, development, and test:

```ruby
def ensure_safe_environment!
  return if Rails.env.development? || Rails.env.test?
  return if Rails.env.staging? && ENV["BRANCH_DEPLOYMENT"] == "true"
  abort "preview_qa tasks mutate data and only run on preview apps, development, or the test suite."
end
```

Same idea for the `unless Rails.env.production?` wrapper on line 67 — gating it on the preview-app condition instead keeps the namespace from even being defined on shared staging.

lib/tasks/preview_qa.rake
```diff
@@ -0,0 +1,151 @@
...
+        # Default to just past the subscription's billing period, which is the common case:
+        # make the subscription look overdue so RecurringChargeWorker will actually charge it.
+        days = (purchase.subscription.period / 1.day).ceil + 1
+      end
+      abort "Pass a positive number of days (this purchase has no subscription to derive a billing period from)." if days.nil? || days <= 0
+
+      backdated_to = days.days.ago
+      # update_columns on purpose: this is a QA time-travel edit, and running the full
+      # callback/validation stack against a doctored timestamp is exactly what we don't want.
+      purchase.update_columns(
```
ershad: This shifts only the purchase you name. `RecurringChargeWorker`'s own gate keys off `last_successful_purchase.created_at`, but `Subscription#overdue_for_charge?` (and the scheduling paths) key off `last_successful_not_reversed_or_refunded_charge_at`, which is the max `succeeded_at` across all successful purchases.

So for a subscription that already has a renewal charge, backdating just one purchase can leave it looking not-yet-due by those paths. For the typical single-charge subscription this is fine; for one with history, shifting every `subscription.purchases.successful` back by the offset makes it unambiguously overdue by every code path. Worth that, or a note on the task that the default assumes a single successful charge.

ershad (commented):
I built an alternative implementation of #5806 item #2 in parallel before finding this PR. Rather than open a competing PR, I'm folding the differences worth keeping into review feedback here.

Overall this is solid — accepting either a database id or an external id, the `Sidekiq::Job` check that stops `run_worker` from instantiating arbitrary classes, and the rake-arg casting are all nice touches, and the tests are thorough.

Three things worth considering, one of which is a real gap against the acceptance criteria:

1. **The guard should also exclude shared staging (inline on the guard).** Acceptance criterion #2 is "unavailable in production/staging proper". Today the guard only refuses `Rails.env.production?`, so these mutating tasks would still run on shared staging (`staging.gumroad.com`), which QA and reviewers depend on. Preview apps are distinguishable from shared staging by `ENV["BRANCH_DEPLOYMENT"] == "true"`, so the safe rule is to allow only preview apps, development, and test.

2. **A read-only inspection task would close the last workflow the issue names.** The issue lists "Inspecting state after a QA run (charge data, json_data on records, enqueued jobs)" as a first-class need, and there's no read helper here yet. A `preview_qa:inspect_subscription[id]` that prints renewal timing, the card's mandate linkage, and the tail of the purchase history with charge ids gives reviewers the "verify charge data" step without hand-writing console queries. Happy to share the version I wrote.

3. **Backdating only the named purchase (inline).** Perfect for the common single-charge subscription; for one that already has a renewal, `overdue_for_charge?` keys off the latest successful charge's `succeeded_at`, so shifting one purchase may not flip it.

None of this blocks — (1) is the one I'd genuinely fix before merge.

lib/tasks/preview_qa.rake
gumclaw: Fixed in 40f8300 — the guard now allows only preview apps (`Rails.env.staging?` + `ENV["BRANCH_DEPLOYMENT"] == "true"`), development, and test, exactly as suggested. The namespace-level wrapper on line 67 uses the same `PreviewQa.safe_environment?` check, so the tasks aren't even defined on shared staging or production. Added specs covering the shared-staging abort and the preview-app allow path (17 examples, 0 failures locally), and updated `docs/deploying.md` and the PR body to state the tightened scope.

lib/tasks/preview_qa.rake
gumclaw: Good catch — fixed in 6d7f06c04 (code landed alongside the guard change in 40f8300da). The task now collects `subscription.purchases.successful` and shifts every charge by the same offset (relative offsets, so their ordering is preserved), instead of pinning the one named purchase to `days.days.ago`. That makes the subscription overdue by both gates: `RecurringChargeWorker`'s `last_successful_purchase.created_at` check and `Subscription#overdue_for_charge?` via `last_successful_not_reversed_or_refunded_charge_at`.

Added a spec that creates a subscription with a renewal charge, runs the task against the original purchase, and asserts both purchases moved and `overdue_for_charge?` flips to true. All 18 examples pass locally.

lib/tasks/preview_qa.rake
```diff
@@ -0,0 +1,151 @@
...
+        created_at: backdated_to,
+        succeeded_at: (backdated_to if purchase.succeeded_at.present?)
+      )
+
+      puts "Backdated purchase #{purchase.external_id} (id #{purchase.id}) by #{days} days: created_at/succeeded_at now #{backdated_to}."
```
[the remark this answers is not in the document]
gumclaw: Fixed in d91f944 — the confirmation line now appends "succeeded_at was nil and stays nil." when the purchase never succeeded, so the output no longer implies a success timestamp was set. The nil-succeeded_at spec asserts the new wording.

lib/tasks/preview_qa.rake
```diff
@@ -0,0 +1,151 @@
...
+
+  # Rake passes every argument as a string; workers usually expect integer ids and booleans.
+  # Casts the obvious scalar types and leaves everything else as a string.
+  def cast_argument(value)
+    case value
+    when /\A-?\d+\z/ then Integer(value)
+    when /\A-?\d+\.\d+\z/ then Float(value)
+    when "true" then true
+    when "false" then false
+    when "nil", "null" then nil
```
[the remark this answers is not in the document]
gumclaw: Documented in d91f944 — added a comment on `cast_argument` calling out that the literal strings "nil"/"null" become Ruby nil with no escape hatch, and pointing anyone who actually needs the string "nil" at the console instead. Left the behavior as-is since nil-casting is the useful default for a QA tool.

gumclaw:
All three addressed:

1. **Shared staging excluded** — the guard now allows only development, test, and staging with `ENV["BRANCH_DEPLOYMENT"] == "true"` (the same flag the app already uses to tell preview apps apart from `staging.gumroad.com`). Both the load-time namespace skip and the per-task runtime check use it, and specs cover production, shared staging (aborts), and preview apps (allowed).

2. **`preview_qa:inspect_subscription` added** — read-only snapshot printing renewal timing (`recurrence`/period, last successful charge, end of current period, `overdue_for_charge?`), the charged card's `stripe_setup_intent_id`/`stripe_payment_intent_id`, and the ten most recent purchases with states, charge ids, and prices. It goes through the same id-or-external-id resolver and mutates nothing, so it works on the default read-only console connection. Documented in `docs/deploying.md`. If your version prints anything this one misses, happy to fold it in.

3. **Backdating now shifts every successful charge on the subscription** by the same offset (relative order preserved), since `overdue_for_charge?` and `RecurringChargeWorker`'s gate both key off the latest successful charge rather than the named purchase.

21 examples, 0 failures locally; rubocop clean.

docs/deploying.md
```diff
@@ -193,3 +195,29 @@ Adding the label triggers a Buildkite build that deploys the branch. Each subseq
...
+
+### Seeding QA state with `preview_qa` rake tasks
+
+Use the permanent tasks in `lib/tasks/preview_qa.rake` to seed edge-case state on a preview app instead of adding temporary, param-gated seed hooks to your PR ("TEMP: revert before merge" commits). The tasks are unavailable in production.
+
+Run them through the console one-shot path with a writable connection, for example:
+
+```shell
+# Backdate a subscription purchase past its billing period so a renewal charge is due
+COMMAND='Rake::Task["preview_qa:backdate_purchase"].invoke("<purchase external_id>")' ./console.sh -w <branch>
```
ershad: `./console.sh -w <branch>` won't do what it looks like — verified against `gumroad-deployment`:

- **The branch comes from the `BRANCH` env var, not a positional arg.** `nomad/staging/deploy_branch/console.sh` does `BRANCH=${BRANCH:-$(git rev-parse --abbrev-ref HEAD)}` and its own usage string is `BRANCH=<branch name> ./console.sh`. Without `BRANCH=`, it falls back to whatever branch the *deployment* repo is checked out on, so this connects to the wrong app (or none).
- **`-w` is the writable-mode flag, not a branch selector.** `docs/ssh.md` (lines 71-74): "If you want to access the rails console in the writable mode call the script like this: `./console.sh -w`". It's a standalone positional flag (`connect_to_console` checks `[[ "$1" != "-w" ]]`); the `<branch>` after it is swallowed as an unused `$2`.
- **The `COMMAND` value must be a real command, not a bare Ruby snippet.** `connect_to_console` runs it as the `docker exec` command (`command="$COMMAND"`), so `Rake::Task[...].invoke(...)` would be exec'd as a program name and fail — it needs `bundle exec rake '...'` (and rake tasks aren't loaded under `rails runner`, so `bundle exec rake` is the right entrypoint).

Corrected one-shot (writable, since these tasks write):
```bash
cd nomad/staging/deploy_branch && \
  BRANCH=<branch name> \
  COMMAND='bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"' \
  ./console.sh -w
```

Honestly the simplest thing to document is the interactive path, then run the task at the prompt:
```bash
cd nomad/staging/deploy_branch && BRANCH=<branch name> ./console.sh -w
# then, at the console:  system 'bundle exec rake "preview_qa:backdate_purchase[<id>]"'
```
Same fix applies to lines 214/217/220 and the header comment in `lib/tasks/preview_qa.rake`.

lib/tasks/preview_qa.rake
```diff
@@ -0,0 +1,151 @@
...
+# Before these existed, seeding edge-case state on a preview app (a backdated purchase to force a
+# subscription renewal, a card with no e-mandate, a dead Sidekiq job) meant shipping temporary
+# param-gated hooks in the PR itself, marked "TEMP: revert before merge" — extra commits, review
+# noise, and a real risk of the hook leaking into main. These tasks are reviewed once and reused
+# forever instead.
+#
+# Run them through the preview app's Rails console one-shot path (see "Deploying to a preview app"
+# in docs/deploying.md), for example:
+#
+#   COMMAND='Rake::Task["preview_qa:backdate_purchase"].invoke("<purchase external_id>")' ./console.sh -w <branch>
```
ershad: Same `./console.sh -w <branch>` issue as in `docs/deploying.md` — the branch is a `BRANCH=<branch name>` env var and `-w` is the writable-mode flag, not a branch arg (see the inline note on the docs file for the verified breakdown and the corrected invocation).

ershad (commented):
Follow-up: the `./console.sh -w <branch>` invocation is incorrect — I checked it against the deployment repo. It appears here in `docs/deploying.md` (lines 211/214/217/220) and once in the rake file header. Details inline; the short version is the branch comes from a `BRANCH=` env var, and `-w` is the writable-mode flag, not a branch selector.

docs/deploying.md
gumclaw: Fixed in e23a82c. All the documented invocations now match how `console.sh` actually works:

- The docs section now documents the interactive path (`cd nomad/staging/deploy_branch && BRANCH=<branch name> ./console.sh -w`, then `system 'bundle exec rake "..."'` at the prompt) as the primary route, with the corrected `COMMAND='bundle exec rake "..."'` one-shot as the alternative.
- The per-task examples are now plain `bundle exec rake "preview_qa:<task>[<args>]"` lines (bracket-style rake args) instead of `Rake::Task[...].invoke` snippets passed to `COMMAND`.
- The header comment in `lib/tasks/preview_qa.rake` shows the same corrected interactive path.
- The console-access paragraph in `docs/deploying.md` also now spells out that `BRANCH` is an env var with a fallback to the deployment repo's checked-out branch, and that `COMMAND` runs as the container command so it must be a real executable.

lib/tasks/preview_qa.rake
gumclaw: Fixed in e23a82c along with the docs file. The header comment now shows the verified invocation: `cd nomad/staging/deploy_branch && BRANCH=<branch name> ./console.sh -w`, then `system 'bundle exec rake "preview_qa:backdate_purchase[<purchase external_id>]"'` at the prompt — branch via the `BRANCH` env var, `-w` as the standalone writable flag, and rake args in brackets rather than a bare Ruby snippet in `COMMAND`. The one-shot `COMMAND=` form in docs/deploying.md was corrected the same way.

gumclaw:
Greptile's last review summary flagged one remaining edge here after the merge: `backdate_purchase` called `subscription.period` without a recurrence guard, so a subscription with no price/recurrence (possible for hand-built dev/test records) would crash with `NoMethodError` instead of a clear abort.

Fixed in follow-up #5824 — the default-days derivation now requires `subscription&.recurrence` and aborts with "pass a positive number of days" otherwise, matching how `inspect_subscription` already handles nil recurrence. Covered by a new spec; 22 examples, 0 failures locally.

PR #5830, opened 2026-07-12
Push web image straight from the BuildKit builder instead of load-then-push

gumclaw:
## What

Pushes the web image to ECR straight from the BuildKit builder (`--push`) instead of exporting it back into the local Docker engine (`--load`) and running a separate `docker push`. The Makefile `build` target gains overridable `WEB_OUTPUT_OPTS` / `WEB_POST_BUILD` variables whose defaults reproduce the classic `docker build` flow byte-for-byte; only the CI BuildKit path overrides them to tag the image with its final `web-<sha>` name and push from the builder.

## Why

#5828 moved Buildkite image builds to a docker-container BuildKit builder with an ECR registry cache. Layer caching works (45 CACHED steps on today's preview builds, gems + GeoIP all hit), but `--load` re-exports the multi-GB image into the local engine on every build — measured at ~70s of "exporting to docker image format" in preview build 13916, which pushed the Build Web Image step from ~1.9 min pre-#5828 to 3.1–3.8 min after. The image was being uploaded to ECR immediately afterwards anyway, so the export is pure overhead: `--push` uploads once from the builder and skips it.

Nothing downstream needs the local copy: `compile_assets.sh` already pulls `web-<sha>` from ECR when it's not in the local engine (line 58), which is the normal case when assets run on a different agent. When assets happen to land on the same agent this shifts a pull into that step — the PHASE instrumentation from gumroad-deployment#39 will show the net effect.

Part of #5802.

## Test plan

- `bash -n` + shellcheck on both scripts (only pre-existing info-level notes).
- `make -n build` dry-runs verified for both paths:
  - default: identical command to main (`docker build --compress -t …:latest`, then `docker tag`)
  - CI override: `docker buildx build --builder … --push -t REG/gumroad/web:web-<sha> …` with cache opts, post-build no-op
- Fallback path (buildx unavailable / build failed) unchanged: classic build + the existing retry push loop still runs (`WEB_IMAGE_PUSHED` stays false).
- After merge: compare Build Web Image duration on the next preview redeploys against today's 3.1–3.8 min (target: back under ~2.5 min) and confirm Deploy Preview App still finds `web-<sha>` in ECR.

## Before/After

Not a UI change — CI build pipeline only.

### DIFF
--- a/.buildkite/scripts/build_web.sh
+++ b/.buildkite/scripts/build_web.sh
@@ -36,11 +36,18 @@ build_web_image() {
 }
 
 logger "Building $WEB_REPO:web-$WEB_TAG"
+WEB_IMAGE_PUSHED=false
 if buildkit_cache_available; then
-  logger "Using BuildKit registry cache ($WEB_REPO:buildcache)"
-  if ! build_web_image \
-    DOCKER_BUILD="$(buildkit_docker_build)" \
+  logger "Using BuildKit registry cache ($WEB_REPO:buildcache); pushing web-$WEB_TAG straight from the builder"
+  if build_web_image \
+    DOCKER_BUILD="$(buildkit_docker_build_push)" \
+    WEB_OUTPUT_OPTS="-t $WEB_REPO:web-$WEB_TAG" \
+    WEB_POST_BUILD=":" \
     WEB_CACHE_OPTS="$(buildkit_cache_opts $WEB_REPO:buildcache)"; then
+    # --push already uploaded web-$WEB_TAG to ECR as part of the build, so the
+    # separate push loop below would be a redundant (if cheap) re-upload.
+    WEB_IMAGE_PUSHED=true
+  else
     buildkit_fallback_notice "web" "buildx build failed"
     build_web_image
   fi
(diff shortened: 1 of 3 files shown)


gumclaw:
Pre-merge fable-5 adversarial review (against branch head, diff vs main):

- **No P1/P2 findings.** Verified against real failure modes:
  - Default Makefile path is byte-identical to main (`make -n build` dry-run compared): same `-t :latest` + `docker tag` flow, so local/dev `make build` behavior is unchanged.
  - Fallback ordering: if the buildx `--push` build fails, `WEB_IMAGE_PUSHED` stays false, the classic build runs, and the existing retry push loop executes — no path exits with an unpushed `web-<sha>`.
  - Nothing in CI or deploy reads the local `gumroad/web:latest` tag (repo-wide grep); `compile_assets.sh` pulls `web-<sha>` from ECR when it's not local, which covers both the cross-agent case (already the norm) and the new same-agent case.
  - `build_base.sh` / `build_branch_app_nginx.sh` intentionally keep `--load`: base is behind a manifest-inspect short-circuit and rarely builds; nginx images are small enough that the export is negligible.
- P3 (noted, not blocking): same-agent asset compiles now pay an ECR pull that the `--load` export previously prepaid. Net effect will show in the gumroad-deployment#39 PHASE annotations; if it's ever negative we can pre-pull in the assets step.

Shellcheck: only pre-existing info-level notes (SC2086 quoting) on lines this PR didn't introduce.

slavingia:
I'm not convinced that docker is faster than just doing everything bare metal etc. just fyi so you know.

gumclaw:
Fair, and for part of the pipeline you're right. Where things stand:

- The web image itself is hard to avoid: production runs on Nomad as Docker containers, so the image is the deploy artifact, not a build convenience. #5828/#5830 are about making that unavoidable build cheaper (registry layer cache, skipping a ~70s re-export) rather than a bet that docker is fast in general.
- The asset compile is the case where docker is pure overhead: it currently boots a whole compose stack (mysql/mongo/redis/memcached) just to run precompile. Killing that and running precompile bare on the agent with stubbed env plus a content-addressed cache is the next item on #5802 — exactly the bare-metal direction.

The per-phase instrumentation from gumroad-deployment#39 will show real numbers over the next few preview/prod runs. If the dockerized steps still dominate after the cache work, moving more of the pipeline bare onto the agents is on the table — I'll post the measured breakdown on #5802 either way.

PR #1054, opened 2025-08-28
 feat(nav): cache home nav, improve a11y, defer GitHub stars; fix mobile spacing and class typo

AnasSarkiz:
Improve home nav performance, accessibility, and mobile spacing
Adds fragment caching, a11y attributes, defers GitHub stars, fixes Tailwind typo, and mobile spacing for GitHub chip.

ref #864

PR #5831, opened 2026-07-12
Percent-encode store-agent path params: non-ASCII ids 500 the confirm flow

gumclaw:
## Root cause (gumroad-private#1054)

The agent confirm path (`AgentMessagesController#execute` → `StoreAgentActionExecutor` → `StoreAgentApiClient`) replays the confirmed write through the real v2 API via an in-process `ActionDispatch::Integration::Session`. rack-test calls `URI.parse` on the internal URL, and Ruby's URI parser raises `URI::InvalidURIError` on **any non-ASCII byte** — so a proposal whose product id carries a stray non-ASCII character 500s at confirm with the generic "Something went wrong", and the proposal card dismisses.

Production evidence (ES logs, request `001d3d14`): `URI must be ascii only "http://api.gumroad.com:80/v2/products/GJs2\u0A86"` → `Api::Internal::AgentMessagesController#execute status=500`. Mobile agent path shows the same signature (Jul 9–10) — both share `Endpoint#expand_path`.

## Fix

`ERB::Util.url_encode` each substituted path param in `Endpoint#expand_path`, **after** the existing path-injection guards (separators/traversal/%-escapes) so their view of the raw value is unchanged. Well-formed ids (`[A-Za-z0-9_-]`) pass through byte-identical; a wrong-but-non-ASCII id now routes to the API and comes back as the API's own clean "not found" message the seller can act on.

## Test plan

- New regression specs: non-ASCII id (`GJs2આ` → `/products/GJs2%E0%AA%86`, parseable by URI), space encoding, safe-id passthrough. **Red-green verified**: 2 failures without the app change, 0 with.
- Full `store_agent_api_catalog` + `store_agent_action_executor` suites: 30 examples, 0 failures.
- Rubocop clean.

## Note on scale of impact

Retained logs (Jul 6+) show only ~6 confirm failures with this signature — the "26 of 30 confirms never applied" figure in #1054 likely mixes in the (much noisier) Anthropic-timeout failure mode on the propose/stream side, which is a separate issue. This PR fixes the deterministic confirm 500; the timeout noise and the dead Sentry ingestion remain open in #1054.

### DIFF
--- a/app/services/ai/store_agent_api_catalog.rb
+++ b/app/services/ai/store_agent_api_catalog.rb
@@ -52,7 +52,13 @@ module Ai::StoreAgentApiCatalog
         if value.match?(%r{[/\\]}) || value.include?("..") || value.include?("%")
           raise ArgumentError, "Invalid path parameter :#{key}"
         end
-        value
+        # Percent-encode the value before it is spliced into the URL. LLM-proposed ids sometimes
+        # carry stray non-ASCII characters (e.g. "GJs2આ"), and Ruby's URI.parse — which rack-test
+        # runs on the internal request path — raises URI::InvalidURIError on any non-ASCII URI,
+        # turning a merely-wrong id into a 500 at confirm time (gumroad-private#1054). Encoded,
+        # the request routes normally and the API answers with its own clean "not found" message.
+        # Encoding happens AFTER the validations above so their view of the raw value is unchanged.
+        ERB::Util.url_encode(value)
       end
     end
 
--- a/spec/services/ai/store_agent_api_catalog_spec.rb
+++ b/spec/services/ai/store_agent_api_catalog_spec.rb
@@ -31,6 +31,20 @@ describe Ai::StoreAgentApiCatalog do
     it "rejects a percent-encoded path param (could decode to a separator)" do
       expect { endpoint.expand_path("id" => "%2e%2e%2fadmin") }.to raise_error(ArgumentError, /invalid path parameter/i)
     end
+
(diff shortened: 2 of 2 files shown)


gumclaw:
Fable-5 pre-merge review: **READY-TO-MERGE** (no P1/P2; two P3 nits noted below).

All five attack vectors probed empirically (Journey router probes on actionpack 7.1.6 + red/green worktree runs), not by argument:
- **No encoding-introduced separator injection**: the guard rejects raw `/`, `\`, `..`, `%` before encoding, and `url_encode` output can never contain `%2F`/`%5C` unless the raw char was present (already rejected). Router probe confirms single-decode only — no double-decode reroute.
- **Round-trip exact**: `GJs2%E0%AA%86` → controller sees raw `GJs2આ` → `find_by_external_id` rescues the Base64 ArgumentError → falls through to permalink lookup → clean \"product was not found\". No new exception path.
- **`=` padding (the sharpest risk — external ids are padded urlsafe-base64, `==` is the norm)**: `M14x%3D%3D` routes and decodes to `M14x==` — identical raw id pre/post PR.
- **Mobile covered**: `Api::Mobile::AgentController` reuses the same service/executor → same patched `expand_path`; no separate path construction.
- **Specs genuinely red without the fix** (2/2 fail), green with (14/14).

P3s (non-blocking, noting for the record): (1) `.` remains unencoded so `x.json` gets format-split by Journey — harmless (ids can't contain `.` per validation, v2 always renders json, pre-existing behavior); (2) a client-level regression spec through the integration-session dispatch would pin the prod failure layer tighter than the unit-level `URI.parse` proxy.

Merging on green CI.

PR #5843, opened 2026-07-12
Dev server: auto-invalidate stale Vite modules (stop trusting the file watcher)

gumclaw:
## Problem

Vite's dev server caches each module's transform and relies entirely on file-watcher events to invalidate it. On macOS the watcher (FSEvents/chokidar) misses events in real situations — most commonly when git rewrites many files at once (branch switch, rebase, stash pop). After a miss, the dev server **silently serves stale code until someone restarts it**: the page loads fine, it just isn't running what's on disk.

This hit us three separate times today while iterating on #5840/#5841 — each time the symptom was "my change has no effect," and the diagnosis (curl the module off the dev server, diff against disk) is not something anyone should have to know.

## Fix

A ~40-line dev-only plugin (`config/vite/stale-module-guard.ts`) that removes the trust in watcher events:

- `transform` hook records when each file was last transformed.
- A request middleware stats the file before Vite serves it; if disk mtime > last transform, the module is invalidated so the request re-transforms from disk.

Cost: one `fs.stat` per module request, dev only (`apply: "serve"`). Production builds untouched. Fails open (any stat error → serve as before).

## Verification

Deterministic repro in a scratch project — watcher events suppressed via `server.watch.ignored` (simulating the macOS miss), then edit the file and re-request:

```
WITHOUT-guard first request: v1
WITHOUT-guard after edit:    v1   ← stale, silently
WITH-guard    first request: v1
WITH-guard    after edit:    v2   ← correct
```

Also booted the real app with the plugin: normal serving, HMR unaffected, no errors.

Also adds `config/vite/**/*.ts` to eslint's node-globals block (same treatment as `vite.config.ts`).

### DIFF
--- a/app/javascript/components/ApiDocumentation/Endpoints/Licenses.tsx
+++ b/app/javascript/components/ApiDocumentation/Endpoints/Licenses.tsx
@@ -27,9 +27,7 @@ export const VerifyLicense = () => (
     <ApiParameters>
       <ApiParameter
         name="product_id"
-        description={
-          "(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
-        }
+        description="(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
       />
       <ApiParameter name="license_key" description="(the license key provided by your customer)" />
       <ApiParameter name="increment_uses_count" description='("true"/"false", optional, default: "true")' />
@@ -99,9 +97,7 @@ export const EnableLicense = () => (
     <ApiParameters>
       <ApiParameter
         name="product_id"
-        description={
-          "(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
-        }
+        description="(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
       />
       <ApiParameter name="license_key" description="(the license key provided by your customer)" />
     </ApiParameters>
@@ -171,9 +167,7 @@ export const DisableLicense = () => (
     <ApiParameters>
       <ApiParameter
         name="product_id"
-        description={
-          "(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
-        }
+        description="(the unique ID of the product — copy it from the license key block on the product's Content tab, or use the id field returned by the GET /products endpoint)"
       />
       <ApiParameter name="license_key" description="(the license key provided by your customer)" />
     </ApiParameters>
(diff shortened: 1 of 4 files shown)


slavingia:
Fable 5 review. test.

slavingia:
red

gumclaw:
The red Lint JS/TS check wasn't from this PR's diff — main is red. #5842 merged with five `react/jsx-curly-brace-presence` errors in `Licenses.tsx`, so every branch cut from it fails lint.

Fixed both ways: #5844 fixes main (auto-merge armed), and I cherry-picked the same fix onto this branch so lint here reruns green now.

gumclaw:
## Fable-5 adversarial review

Reviewed at head `ce94c77` (plugin, vite.config wiring, eslint node-globals block). Verified claims by running the plugin against a live Vite 6.4.2 server with watcher events suppressed, under both base `/` and this app's real dev base `/vite-dev/`.

### P1 — guard never fired in the real app (fixed in `902d3fd`)

This app serves dev assets under a base path (`/vite-dev/`, per `config/vite.json`), but Vite's module graph keys URLs **without** that prefix. Plugin middlewares run before Vite strips the base from `req.url`, so `getModuleByUrl("/vite-dev/src/...")` returned `undefined` on every request and the guard silently did nothing — the PR's scratch-project repro used the default base `/`, which masked this. Reproduced deterministically:

```
base /          : after edit → v2  (guard fired)
base /vite-dev/ : after edit → v1  (stale — guard never matched)
```

Fix: strip `server.config.base` from the URL before the lookup. Re-ran the same suppressed-watcher repro with the fixed plugin file: fresh code served under both bases.

### P3s (noted, not blocking)

- CSS served with `?direct` may be keyed in the module graph with the query attached; the stripped-query lookup can miss those, meaning no guard coverage for that case. Fails open — same behavior as today, so left as-is.
- A file write landing mid-transform can record a transform timestamp newer than the file's mtime, so that one edit wouldn't be caught until the next write. Inherent to the mtime approach; acceptable for a dev convenience guard.
- Each matching request adds an awaited module-graph lookup + `fs.stat` before `next()`. Dev-only and per-module, consistent with the PR's stated cost.

Everything else held up: `apply: "serve"` keeps builds untouched, errors fail open, `invalidateModule` propagates to importers, lint + prettier clean on the changed files.

Verdict: **good to merge once CI is green** — the P1 is fixed on the branch, and the unrelated red Lint JS/TS (from #5842 on main) is handled by #5844 plus the cherry-pick here.

slavingia:
Open new pr with fix as already merged

gumclaw:
Done — #5845 reopens the base-path fix (`902d3fd`) that missed the squash merge. Same commit that was verified in the review above (fresh code served under both base `/` and `/vite-dev/`). Auto-merge is armed; it lands on green.

PR #5847, opened 2026-07-13
Scope checkout already-purchased checks to candidate products (fixes checkout 504 for buyers with large purchase histories)

gumclaw:
## What

Fixes the checkout page 504ing for logged-in buyers with very large purchase histories whenever a cart item has a cross-sell configured.

`CheckoutPresenter` answered "has this buyer already purchased product X / variant Y" (used to filter cross-sell and upsell offers) by loading the buyer's **entire purchase history** with products and variants and building an in-memory lookup set. For buyers with thousands of purchases that single load exceeded the request timeout, so any cart containing a cross-sell-bearing product returned a 504 on every load — the buyer couldn't view, edit, or check out their cart.

This PR replaces the full-history load with a targeted, indexed query per candidate product (`logged_in_user.purchases.where(link_id: product.id)`), memoized per product. A cart only ever checks a handful of cross-sell/upsell candidates, so this is a few cheap `link_id`-indexed queries instead of one unbounded load.

## Why

Tracking issue: antiwork/gumroad-private#1063 (support-originated; verified on the read replica — the full-history load alone timed out at >55s for the affected buyer, while the rest of their cart rendered in about a second).

Both call sites of `already_purchased?` are covered: the cross-sell filter in `checkout_product` and the upsell variant filter for product options.

Behavior is intentionally unchanged apart from performance: the old code counted **all** of the buyer's purchases of a product (not just successful ones), and the new per-product query uses the same `purchases` association so the filtering semantics are identical.

## Test plan

- New spec: hides a cross-sell the buyer has already purchased (behavior preserved).
- New spec: asserts every `purchases` query issued during cross-sell rendering is scoped by `link_id` — i.e. no unscoped full-history load can regress back in.
- Existing "eager loads purchases to avoid N+1 queries" spec still passes (query count for the upsell path stays bounded).
- Ran locally: `bundle exec rspec spec/presenters/checkout_presenter_spec.rb` — 49 examples, 0 failures.

## Before/After

Not applicable — no user-facing rendering change; the page simply loads instead of timing out.

### DIFF
--- a/app/presenters/checkout_presenter.rb
+++ b/app/presenters/checkout_presenter.rb
@@ -373,18 +373,24 @@ class CheckoutPresenter
       end
     end
 
-    def purchases
-      @_purchases ||= logged_in_user&.purchases&.includes(:link, :variant_attributes)&.map { |purchase| { product: purchase.link, variant: purchase.variant_attributes.first } } || []
-    end
-
-    def purchased_product_variant_set
-      @_purchased_product_variant_set ||= purchases.each_with_object(Set.new) do |purchase, set|
-        set.add([purchase[:product]&.id, purchase[:variant]&.id])
-      end
-    end
-
+    # Answers "has the logged-in buyer already bought this exact product + variant
+    # combination?" for upsell and cross-sell filtering. We only ever check the handful
+    # of upsell/cross-sell candidates attached to the cart, so we query the buyer's
+    # purchases of each candidate product individually instead of loading their entire
+    # purchase history. Buyers with thousands of purchases used to time out the whole
+    # checkout page when a cart item had a cross-sell, because the old implementation
+    # materialized every purchase (with products and variants) just to build a lookup set.
     def already_purchased?(product, variant)
-      purchased_product_variant_set.include?([product&.id, variant&.id])
+      return false if logged_in_user.nil? || product.nil?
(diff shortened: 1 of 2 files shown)


gumclaw:

- **Semantics parity:** the new per-product query uses the same `logged_in_user.purchases` association as the removed full-history load, so which purchases count toward "already purchased" is unchanged. All four nil combinations of (product, variant) return the same answer as the old set-based lookup, including the "purchased with no variant, checking a specific variant" case and its inverse.
- **No dangling references:** `purchases` and `purchased_product_variant_set` had no callers outside the presenter; grep confirms nothing else in app/ or spec/ references them.
- **Query bounds:** one indexed `link_id` query per distinct candidate product, memoized per presenter instance; the existing N+1 spec (≤2 purchase queries for the upsell path) still passes, and a new spec pins that every purchases query during cross-sell rendering is `link_id`-scoped.
- Findings: none at P1/P2. P3 noted and accepted: a buyer with an enormous number of purchases of one single candidate product still loads all of those rows — bounded per product and orders of magnitude smaller than the prior full-history load.

Verdict: pass. `bundle exec rspec spec/presenters/checkout_presenter_spec.rb` → 49 examples, 0 failures; rubocop clean.

PR #5848, opened 2026-07-13
Validate Singapore NRIC/FIN format when saving payout compliance info

gumclaw:
## What

Validates the Singapore NRIC/FIN format when a seller saves their payout compliance info. The check runs server-side in `UpdateUserComplianceInfo` (authoritative, same shape as the existing Peru DNI check) and client-side in the payments settings form for fast inline feedback. The field placeholder also changes from `123456789` to `S1234567A` so the expected format — leading letter, seven digits, trailing checksum letter — is visible before the seller types anything.

## Why

A Singapore seller spent two weeks stuck in a Stripe "id_number mismatch" verification loop because the NRIC stored in their compliance info was missing its leading letter — only the seven digits and checksum letter were present. Stripe verifies the full NRIC string, so every document re-upload failed with a generic mismatch the seller could not see or fix from their side. Support corrected the stored value via console and verification cleared immediately.

Investigation of the save path (form input → `Settings::PaymentsController#update` → `UpdateUserComplianceInfo` → Strongbox encryption → Stripe `individual.id_number`) found no code that strips leading alpha characters — the value is stored and forwarded verbatim, so this was almost certainly an entry error rather than data mangling. But the field gave no hint of the expected format (placeholder was `123456789`, all digits), and a malformed value saved silently and only failed later inside Stripe verification with no actionable feedback. This mirrors the Peru DNI situation fixed in #5543: the fix is to reject the malformed value at save time with a clear message.

The pattern accepts `S`/`T`/`F`/`G`/`M` prefixes (all current NRIC/FIN series), is case-insensitive, and tolerates spaces/dashes when checking the shape. The submitted value is still stored exactly as entered.

Refs antiwork/gumroad-private#1065

## Test plan

- `spec/services/update_user_compliance_info_spec.rb` — new "with a Singapore individual" context: rejects an NRIC missing the leading letter, rejects digits-only, accepts well-formed NRIC (stored verbatim), accepts lowercase, tolerates spaces/dashes, ignores masked resubmission. 39 examples, 0 failures locally.
- Existing Peru/US/non-US tax-id specs in the same file all pass (no behavior change for other countries).
- Screenshots below from a local run.

## Before/After

| | Before | After |
|---|---|---|
| **Placeholder** | `123456789` (digits only, no hint the letter is needed) | placeholder |
| **Malformed NRIC on save** | Saved silently, failed later inside Stripe verification with no visible error | error |

🤖 AI disclosure: implemented with Claude (fable-5) via Hermes Agent.


gumclaw:
Pre-merge adversarial review (fable-5), verdict: no P1/P2 findings.

Checked specifically:

- **Legacy sellers not stranded**: the check only runs when a new `individual_tax_id` is submitted; masked resubmissions (`•`/`*`) are filtered by `submitted_tax_id_for` before the check, so an existing SG seller with a stored value can still update unrelated fields. Covered by the "ignores a masked NRIC resubmission" spec.
- **Business path**: an SG business representative's NRIC goes through the same `individual_tax_id` param; `effective_legal_entity_country_code` resolves via `business_country` when `is_business`, mirroring the Peru DNI check exactly.
- **Prefix coverage**: `[STFGM]` covers all current NRIC/FIN series (S/T citizens and PRs, F/G/M foreigners), case-insensitive since Stripe accepts lowercase.
- **No stored-value change**: the submitted value is persisted exactly as entered; only the shape check strips spaces/dashes.

P3 (noting, not fixing): the input's `maxLength={9}` means a browser user can't actually type the space/dash variants the server tolerates — that tolerance mainly benefits API/paste-then-trim flows. Harmless.

gumclaw:
## Pre-merge adversarial review (fable-5) — CHANGES-REQUIRED → fixed, now READY

Ran the mandatory local adversarial review against the branch (KYC-field hunt list: country scoping, existing-seller bricking, regex vs real NRIC/FIN shapes, client/server drift, sibling write paths, spec vacuity). Verdict was **CHANGES-REQUIRED** with 1 P2 + 3 P3s; all four are fixed in 2e0cc13.

| Finding | Severity | Fix |
|---|---|---|
| Masked-NRIC spec passed vacuously — it changed nothing else on the record, so the service early-returned "nothing changed" before the guard it was written to pin ever ran. It would have kept passing even if the masked-value filter (the thing that stops existing SG sellers from being bricked) were deleted. | P2 | Spec now also edits the city and asserts it persisted, mirroring the existing masked-resubmission specs (2e0cc13) |
| Error message said "must start with a letter" but the pattern requires S/T/F/G/M — a seller who typed `A1234567B` would be rejected with a message their input already satisfies | P3 | Message now names the allowed prefixes, server + client (2e0cc13) |
| Client stripped Unicode whitespace (JS `\s` with `u` flag) but the server only stripped ASCII — a non-breaking space from a PDF copy-paste passed the browser check then failed server-side | P3 | Server now uses `[[:space:]]`; spec includes an NBSP (2e0cc13) |
| SG input `maxLength: 9` truncated a pasted `S 1234567-A` before the separators were stripped, defeating the tolerance in the web UI | P3 | maxLength raised to 11 (2e0cc13) |

Also fixed earlier: the SG system spec was submitting the letterless `000000000` placeholder — exactly the input the new validation rejects — which is what failed Test Slow 11 on CI. It now submits `S1234567D` (a genuinely checksum-valid NRIC) (0a81f2a).

**What the review confirmed safe** (verified against the actual code, not the PR description):
- Country scoping cannot fire for non-SG sellers; structure is parallel to the Peru DNI check, including the `is_business`-flip handling.
- Existing sellers with stored malformed NRICs are **not** bricked when editing other payout fields — three independent shields verified (form nulls untouched tax-id fields, `submitted_tax_id_for` drops masked values, unchanged submissions early-return). The stored value is only re-checked when the seller types a replacement, which is the intended fix path.
- Regex accepts all legitimate shapes: S/T/F/G/M prefixes incl. the 2022+ M-series, lowercase, full `[A-Z]` checksum superset; `gsub` (non-bang) never mutates the stored value.
- `individual_tax_id` is written only by this service from the payments controller — no admin/sync/backfill bypass. (One pre-existing note: the `ssn_last_four` param path writes `individual_tax_id` with no country scoping — same pre-existing hole the Peru check has, only reachable by crafted API requests since the form renders that field for US only. Not introduced by this PR; noting for completeness.)
- Rejection specs are red/green sound: deleting the validation flips both `not_to change` and `not_to receive` assertions.

**Only verifiable live:** that Stripe truly accepts lowercase NRICs (drives the `/i`), and browser paste behavior for the maxLength change.

Full spec file green locally after fixes (40 examples, 0 failures), rubocop + eslint clean.

gumclaw:
Addressed Greptile's remaining finding in 1830db3: removed the `maxLength` cap on the Singapore NRIC/FIN input in `AccountDetailsSection.tsx`. The cap of 11 still truncated pastes with multiple separator characters (e.g. `S 1234567 - A` is 13 characters) even though the validation deliberately tolerates spaces and dashes. With the cap gone, any separator-containing paste goes through; the inline regex and the authoritative server-side check in `UpdateUserComplianceInfo` enforce the actual format. Prettier and ESLint clean on the touched file.

PR #5849, opened 2026-07-13
Surface the endpoint's HTTP status in test ping results

gumclaw:
## Problem

`TestPingsController#create` rendered the success message for **any** HTTParty response — it never looked at the response object. A seller whose Cloudflare returned 403 to our pings clicked "Send test ping", saw "Your last sale's data has been sent to your Ping URL", reasonably concluded Gumroad was broken, and filed a revenue-critical ticket. The test-ping feature was actively misleading for exactly the sellers who need it.

Bonus confusion: connection-level failures (DNS, refused, timeout, SSL) rendered "That URL seems to be invalid" — wrong message for a valid-but-unreachable URL.

## Fix

The controller now branches on the actual response:

| Outcome | Seller sees |
|---|---|
| 2xx | "…sent to your Ping URL. Your endpoint responded with HTTP 200." |
| 4xx/5xx | Error with the status code + hint to check firewall/bot-protection (Cloudflare) rules |
| Connection failure | "We couldn't reach your Ping URL — the connection failed or timed out…" |
| No sales | unchanged |
| Invalid URL format | unchanged ("That URL seems to be invalid.") |

No changes needed in the frontend — `NotificationEndpointSection.tsx` already renders `success: false` responses as error alerts with `error_message`.

`send_test_ping` already returned the HTTParty response object (or nil for no sales); the controller was just ignoring it. No change to the module.

## Notes / decision points

- **Real sale pings (`PostToIndividualPingEndpointWorker`) are NOT changed.** That worker already checks `.success?` and logs failures — the gap was only in the test-ping path.
- Deliberately did NOT retry or follow redirects differently — this is a diagnostics-surface change only, zero behavior change to what gets sent.
- The 403-from-Cloudflare class of issue is now self-diagnosable, which should shed the "pings are broken" ticket category at the source.

## Testing

- `rspec spec/controllers/test_pings_controller_spec.rb` — 7 examples, 0 failures (3 new: 200 success with code surfaced, 403 rejection surfaced, connection-refused message)
- `rspec spec/modules/user/ping_notification_spec.rb` — 6 examples, 0 failures
- rubocop clean

### DIFF
--- a/app/controllers/test_pings_controller.rb
+++ b/app/controllers/test_pings_controller.rb
@@ -9,14 +9,22 @@ class TestPingsController < Sellers::BaseController
       return
     end
 
-    message = if current_seller.send_test_ping params[:url]
-      "Your last sale's data has been sent to your Ping URL."
+    response = current_seller.send_test_ping params[:url]
+
+    if response.nil?
+      render json: { success: true, message: "There are no sales on your account to test with. Please make a test purchase and try again." }
+    elsif response.success?
+      render json: { success: true, message: "Your last sale's data has been sent to your Ping URL. Your endpoint responded with HTTP #{response.code}." }
     else
-      "There are no sales on your account to test with. Please make a test purchase and try again."
+      # The ping was delivered but the endpoint rejected it (4xx/5xx). Surface the
+      # status code so sellers can self-diagnose (e.g. a firewall or Cloudflare rule
+      # returning 403) instead of concluding Gumroad never sent anything.
+      render json: { success: false, error_message: "Your endpoint responded with HTTP #{response.code} instead of a success code. Check your endpoint's logs and any firewall or bot-protection rules (e.g. Cloudflare) that may be blocking Gumroad's requests." }
     end
-    render json: { success: true, message: }
   rescue *INTERNET_EXCEPTIONS
-    render json: { success: false, error_message: "That URL seems to be invalid." }
+    # Connection-level failure: DNS, refused connection, timeout, SSL, etc.
+    # The request never completed, so there is no HTTP status to report.
+    render json: { success: false, error_message: "We couldn't reach your Ping URL — the connection failed or timed out. Check that the URL is correct and that your server is reachable from the internet." }
   rescue Exception
     render json: { success: false, error_message: "Sorry, something went wrong. Please try again." }
   end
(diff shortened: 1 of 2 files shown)


gumclaw:
## Fable-5 adversarial review — round 1 verdict + fix

**Verdict was CHANGES-REQUIRED: 1 P1, 0 P2, 3 P3s.** P1 fixed in 785c97f8f.

**P1 (real bug, empirically proven on httparty 0.22.0): `HTTParty::Response#nil?` returns `true` for empty-bodied responses.** The first version of this PR checked `response.nil?` to detect the "no sales" case — but HTTParty overrides `#nil?`, so a real 200-with-empty-body (`head :ok` endpoints) or a Cloudflare 403-with-empty-body would have rendered "There are no sales on your account" instead of surfacing the status. That's both a regression vs main (which used truthiness, not `nil?`) and a defeat of this PR's whole purpose in a plausible variant of the original ticket.

**Fix:** `send_test_ping` now returns a `:no_sales` sentinel instead of nil; the controller compares against the sentinel and never calls `nil?`/truthiness on the response. Two new specs use `nil?: true` doubles to lock the regression out (the original specs passed precisely because plain RSpec doubles have an honest `nil?` — masking the bug).

**P3s triaged:**
- `HTTParty::RedirectionTooDeep` not in `INTERNET_EXCEPTIONS` → falls to the generic error message. Real but rare; left as-is to keep this PR minimal (happy to add it to the constant in a follow-up).
- Weak `include "200"` assertion → tightened to assert the exact full message.
- Redirects: HTTParty follows them by default so `.code` reports the final destination — desired behavior, noted here for the record.

**Confirmed sound by the review:** frontend typia contract matches every render branch exactly; `success?`/`code` are real methods on all completed responses; rescue ordering correct, no double-render; no other production callers of `send_test_ping`.

15 examples, 0 failures across both spec files; rubocop clean. CI running on the new head.

gumclaw:
Heads-up: this merged at b2bbfbb1a, before the fable-review P1 fix landed on the branch — main currently misroutes empty-bodied responses (HTTParty::Response#nil? returns true for empty bodies, so a head-:ok 200 or empty-body 403 renders the "no sales" message). Follow-up fix: #5851.

PR #5853, opened 2026-07-13
Alert when a returned payout re-credits a retired Stripe account

gumclaw:
## What

When a Stripe bank payout fails or is returned and the webhook re-credits the funds to the Connect account the payout came from, check whether that account is still the seller's active merchant account. If it isn't (typically because a country change replaced it), add a `[PAYOUT][DRIFT]` payout note on the seller's account naming both the retired and the active account, and notify Sentry.

## Why

Stripe always returns failed/returned payout funds to the account the payout was made from. If the seller changed country in the meantime, that account has been deleted and replaced — the returned money sits on the retired account where nothing looks at it. The seller's ledger balance says they're owed money, future payouts draw from the new account whose Stripe balance is short by exactly the returned amount, and the destination-balance drift guard then refuses every payout. Today this strands silently and only surfaces weeks later as repeated payout failures and an auto-pause; this alert catches it the day the webhook arrives.

The check runs in both `handle_stripe_event_payout_failed` and `handle_stripe_event_payout_reversed`, only for Gumroad-managed Stripe accounts, and is a no-op when the payout's account is still the active one.

Companion PRs: blocking country changes while a payout is in flight (prevention), and a self-diagnosing drift-guard error (detection at payout time).

## Test plan

- `spec/business/payments/payouts/processor/stripe/stripe_payout_processor_spec.rb` — new `alert_if_payout_credited_retired_account` examples: notes + notifies when the account is retired (with and without an active replacement); no-op when the account is still active; no-op for user-connected Stripe Standard accounts. 4 examples, 0 failures locally.
- Existing `handle_stripe_event` suite (41 examples) and `handle_payout_reversed_worker_spec.rb` (3 examples) still green locally.
- No UI change.

### DIFF
--- a/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
+++ b/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
@@ -522,6 +522,8 @@ class StripePayoutProcessor
       payment.processor_reversing_payout_id = reversing_payout_id
       payment.save!
     end
+
+    alert_if_payout_credited_retired_account(payment)
   end
 
   def self.handle_stripe_event_payout_reversal_failed(payment, reversing_payout_id)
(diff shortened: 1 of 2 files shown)


gumclaw:
Pre-merge adversarial review (fable-5): no P1/P2 findings.

P3 noted for the record: `alert_if_payout_credited_retired_account` runs after the payment state transition commits. If the alert itself raised, a webhook redelivery would hit the handler's early-return (state already `failed`/`returned`) and the alert would be lost. Left as-is: both `add_payout_note` and `ErrorNotifier.notify` are low-risk writes, and keeping the alert outside the lock means an alert failure can never roll back or retry the money-side state change — the safer trade.

Verified: alert only fires after a real state transition (the early returns inside `with_lock` skip it for no-op redeliveries), only for Gumroad-managed accounts, and is a no-op when the payout's account is still the active one. Specs cover all four branches.

PR #5854, opened 2026-07-13
Name retired-account balances in the payout drift-guard error

gumclaw:
## What

When the payout destination-balance drift guard trips, the error message now also checks the seller's retired Gumroad-managed Stripe accounts and names any that still hold funds, with the exact residual amounts. Example suffix: "Possible cause: the seller has retired Stripe account(s) still holding funds — acct_… holds 12345 eur cents. A returned payout may have re-credited a retired account after a country change; move those funds to the active account."

## Why

The drift guard says how big the gap is but not where the money went. Every investigated drift case so far had the same root cause: the seller changed country (retiring their Stripe account), and a returned payout or leftover funds stayed on the retired account. Confirming that took a manual per-account Stripe trace each time. Putting the retired-account balances directly in the error turns an hours-long investigation into a one-line read.

The lookup is best-effort: it only runs when the guard has already tripped (no extra Stripe calls on the happy path), and if a retired account can't be read at Stripe, the original drift error still goes through unchanged.

Companion PRs: blocking country changes while a payout is in flight (prevention), and alerting when a returned payout lands on a retired account (day-of detection).

## Test plan

- `spec/business/payments/payouts/processor/stripe/stripe_payout_processor_spec.rb` — new examples under `destination_balance_drift_error`: retired account with funds is named with its amounts; empty retired account adds no hint; Stripe read failure on the retired account degrades gracefully to the plain drift error. Full drift-guard context: 14 examples, 0 failures locally.
- No UI change.

### DIFF
--- a/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
+++ b/app/business/payments/payouts/processor/stripe/stripe_payout_processor.rb
@@ -299,9 +299,42 @@ class StripePayoutProcessor
     "Destination Stripe balance mismatch on #{merchant_account.charge_processor_merchant_id}: " \
       "expected #{expected_destination_cents} #{destination_currency} cents, " \
       "Stripe has #{available_cents} cents available + #{pending_cents} cents pending (gap: #{gap_cents} cents). " \
-      "Reconcile destination balance before retry."
+      "Reconcile destination balance before retry." \
+      "#{retired_account_balances_hint(merchant_account)}"
   end
   private_class_method :destination_balance_drift_error
+
+  # Every drift-guard case investigated so far had the same root cause: the seller changed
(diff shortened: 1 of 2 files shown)


gumclaw:
Pre-merge adversarial review (fable-5): no P1/P2 findings.

Verified:
- The retired-account lookup only runs after the drift guard has already refused the payout, so the happy path gains zero Stripe calls.
- A Stripe read failure on any retired account is swallowed per-account and the original drift error goes through unchanged (spec-covered).
- The hint enumerates all currencies with a positive balance on each retired account, so cross-currency stranding (the common case after a country change) is named correctly.

Specs cover: funds found, retired-but-empty, and Stripe-read-failure paths.

PR #5857, opened 2026-07-13
Subtract refunded principal from invoice payment total

gumclaw:
## What

Fixes the "Payment Total" shown on regenerated invoices for refunded purchases. `Purchase#non_refunded_total_transaction_amount` now subtracts the full gross refunded amount (refunded principal + refunded Gumroad-collected tax) instead of only the refunded tax, so a fully refunded purchase renders a $0 payment total.

## Why

Refs antiwork/gumroad-private#1070.

A fully refunded purchase with no Gumroad-collected tax (e.g. an EU reverse-charge business buyer with a VAT ID) regenerated an invoice showing the full original amount — identical to the pre-refund invoice. The method only subtracted `gumroad_tax_refunded_cents`, ignoring the refunded principal entirely. Business buyers who need documentation of a cancellation for their books were handed an invoice contradicting the refund.

## Test plan

- New model specs for `#non_refunded_total_transaction_amount`: no refund, full refund, partial refund, tax-only refund, principal + tax refund — `spec/models/purchase_spec.rb`, 5 examples, 0 failures locally.
- New presenter spec: fully refunded purchase shows `Payment Total: $0` — `spec/presenters/invoice_presenter/order_info_spec.rb`, 26 examples, 0 failures locally.
- `rubocop -a` clean on all touched files.
- CI green.

## Before/After

Generate-invoice page for a fully refunded $24 purchase (captured from a local run via a system spec):

| Before | After |
|---|---|
| before | after |


gumclaw:
**Pre-merge adversarial review (fable-5) — verdict: no P1/P2 findings.**

Hunt list checked against the diff:

- **Caller graph:** `non_refunded_total_transaction_amount` has exactly one production consumer — `InvoicePresenter::OrderInfo#non_refunded_total_payment_attribute`. `formatted_non_refunded_total_transaction_amount` is only referenced by specs. No accounting, payout, or API surface reads this method, so the semantic change is scoped to invoice rendering.
- **Semantic consistency:** the new subtrahend `gross_amount_refunded_cents` (`amount_refunded_cents + gumroad_tax_refunded_cents`) is the same expression `Purchase::Refundable` uses to decide whether a purchase is fully refunded, so the invoice total now agrees with the model's own refund-state logic.
- **Negative-value risk:** refund creation is guarded so gross refunds cannot exceed `total_transaction_cents`; the result can't go below $0.
- **Partial refunds:** now correctly netted (new spec covers it); previously a partial principal refund was ignored entirely.
- **Free trials:** still short-circuited to 0 in the presenter before this method is called.
- **Spec evidence:** 5 new model examples + full invoice presenter suite (122 examples) green locally; rubocop clean.

P3 (noted, no change needed): the issue suggests also labelling the invoice "refunded" for clarity — worth a follow-up product decision, not blocking this correctness fix.

PR #5858, opened 2026-07-13
Show refund status on regenerated invoices

gumclaw:
## What

Adds a "Refund status" line to invoices when money has been returned on the purchase. A fully refunded purchase shows "Fully refunded on <date>", a partially refunded one shows "Partially refunded on <date>" (date is the most recent refund). Purchases without refunds render exactly as before — the line is omitted.

## Why

#5857 fixed the regenerated invoice's Payment Total to reflect refunded principal (a fully refunded purchase now shows $0). But a $0 total alone doesn't tell a bookkeeper *why* it's $0 — business buyers (especially in the EU, where a credit note is the norm) need the document to state that the payment was refunded. Labelling the invoice makes the regenerated document usable as refund evidence for their books.

Refs antiwork/gumroad-private#1070 (follow-up to the Payment Total fix in #5857).

## Test plan

- New specs in `spec/presenters/invoice_presenter/order_info_spec.rb`: fully refunded (shows "Fully refunded on Feb 1, 2023"), partially refunded (shows "Partially refunded on Feb 1, 2023"), and not refunded (no "Refund status" attribute). Run for both Purchase and Charge chargeables via the shared examples.
- Full invoice presenter suite passes locally: 128 examples, 0 failures (`rspec spec/presenters/invoice_presenter_spec.rb spec/presenters/invoice_presenter/`).
- Rendered the actual `purchases/invoices/create.pdf.erb` template locally for a fully refunded $24 purchase and for an unrefunded one — screenshots below.

## Before/After

Fully refunded purchase — invoice now carries the refund status under Payment Total:

| Fully refunded (after) | Unrefunded (unchanged) |
|---|---|
| refunded invoice | unrefunded invoice |


gumclaw:

- **Frozen string literal + `+=`** — checked: `+=` reassigns rather than mutating, safe under `# frozen_string_literal: true`.
- **Charge with mixed purchases** — one refunded + one untouched renders "Partially refunded", which is the correct invoice-level statement. All-refunded renders "Fully refunded". Free-trial purchases are excluded before the check, matching how the payment total already ignores them.
- **Tax-only refunds** (`refund_gumroad_taxes!`) don't set `stripe_refunded`/`stripe_partially_refunded`, so they don't trigger the label — intentional, since no principal came back and the VAT line already disappears from the invoice.
- **`flat_map(&:refunds)`** runs one query per purchase, but invoice generation handles at most a handful of purchases per charge, so no meaningful N+1.
- **Renders on both surfaces**: PDF (`pdf_attributes`) and the invoice generation form (`form_attributes`), and the form page renders attributes generically so no frontend change was needed.

No P1/P2 findings. Evidence: 128 presenter examples pass locally; rendered template screenshots in the PR body show the refunded and unrefunded states.

PR #5860, opened 2026-07-13
Show a poster frame for uploaded video covers instead of a black idle player

gumclaw:
Fixes antiwork/gumroad-private#1074

## Problem
Video covers uploaded directly to Gumroad idle as a **solid black rectangle** until the buyer presses play. `Product/Covers/Video.tsx` never set JW Player's `image` (poster) option, and `AssetPreview#as_json` only populated `thumbnail` from oEmbed data — so only YouTube/Vimeo embeds got a preview image. Surfaced by a stock-footage seller support ticket asking why his cover "shows a black frame".

## Fix
- **`GenerateVideoPosterWorker`** (new Sidekiq job, low queue) — does the heavy work: downloads the video, grabs a frame via the stock ActiveStorage/ffmpeg previewer, resizes to retina width, caches the CDN URL. Enqueued automatically when a video cover is created (`after_create_commit`), so the poster is usually ready before anyone views the product. Failure caches a sentinel for an hour so retries don't hammer un-previewable videos, and it busts the product cache once a poster lands so cached pages pick it up.
- **`AssetPreview#thumbnail_url`** — oEmbed thumbnail for embeds (unchanged behaviour); for uploaded videos it only **reads the cache** — no download or ffmpeg ever runs on a web request. A cache miss (covers uploaded before this shipped) returns `nil` and enqueues a generation, so the poster appears on later views.
- **`as_json`** exposes it as `thumbnail` (same field embeds already used — no frontend type change).
- **`Covers/Video.tsx`** passes `cover.thumbnail` as JW Player's `image`, so the frame shows while idle. `null` keeps today's black idle state as the fallback.

## Verification
- 12 model specs covering: oEmbed passthrough, nil for images, cached-poster read path, cache-miss enqueues + returns nil, failure sentinel suppresses re-enqueue, generation success/failure/timeout, enqueue-on-create for videos only, `as_json` exposure — all green
- 4 new worker specs: generates, busts product cache on success, skips cache-bust on failure, no-ops for deleted/missing covers — all green
- Full `asset_preview_spec.rb` green except one pre-existing failure on main (`#url=` square-bracket encoding — verified failing before my change)
- rubocop clean on all touched files; eslint + tsc clean (frontend unchanged since last run)

## Notes for review
- Poster generation needs the ActiveStorage video previewer (ffmpeg) available on **worker** boxes — it's the stock Rails previewer; streams transcoding already depends on ffmpeg infra
- The poster is ffmpeg's auto-chosen frame. Sellers who want a curated poster keep the existing workaround (still image as first cover)
- Existing covers backfill lazily: the first render after deploy enqueues generation, and the poster shows from the next view on. A one-shot backfill job could pre-warm everything if we'd rather not wait

### DIFF
--- a/app/javascript/components/Product/Covers/Video.tsx
+++ b/app/javascript/components/Product/Covers/Video.tsx
@@ -20,7 +20,15 @@ const Video = ({ cover, dimensions }: Props) => {
     const url = dimensions == null || dimensions.width > DEFAULT_IMAGE_WIDTH ? cover.original_url : cover.url;
 
     const options: JWPlayerOptions = {
-      playlist: [{ sources: [{ file: url, type: cover.filetype?.toLowerCase() }] }],
+      // "image" is JW Player's poster frame — the still shown before playback
+      // starts. Without it the player idles as a solid black rectangle (the
+      // subject of gumroad-private#1074). For uploaded videos the backend
+      // extracts a frame with ffmpeg (AssetPreview#video_poster_url); when no
+      // poster could be generated this is null and JW Player keeps the old
+      // black idle state.
+      playlist: [
+        { image: cover.thumbnail ?? undefined, sources: [{ file: url, type: cover.filetype?.toLowerCase() }] },
+      ],
     };
     if (dimensions != null) {
       options.height = `${dimensions.height}px`;
--- a/app/models/asset_preview.rb
+++ b/app/models/asset_preview.rb
@@ -9,6 +9,9 @@ class AssetPreview < ApplicationRecord
   RETINA_DISPLAY_WIDTH = (DEFAULT_DISPLAY_WIDTH * 1.5).to_i
 
   after_commit :invalidate_product_cache
+  # Kick off poster-frame extraction right away so the poster is usually
+  # ready by the time anyone views the product page.
+  after_create_commit :enqueue_video_poster_generation
 
   # Update updated_at of product to regenerate the sitemap in RefreshSitemapMonthlyWorker
   belongs_to :link, touch: true, optional: true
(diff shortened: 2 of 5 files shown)


gumclaw:
## Fable-5 adversarial review — verdict + fixes

**Round-1 verdict: concept/frontend sound; 1 P1 + 1 P2 in the backend generation path. Both fixed in 47c1ed107.**

**P1 — unbounded synchronous work on the request thread.** First cache miss downloaded the entire video blob and ran ffmpeg with **no timeout**, triggered from `as_json` (product page render + public API). There's no size cap on video covers, so a multi-hundred-MB file could hold a Puma thread for minutes; `invalidate_product_cache` re-exposes the cold path on every cover edit. **Fix:** generation now shares `IMAGE_PROCESSING_TIMEOUT_SECONDS` (30s), the same budget `retina_variant` uses — pathological videos degrade to no-poster. Upload-time cache warming still applies (cover creation renders its JSON in the same request), so buyers rarely pay even the bounded cost.

**P2 — failures retried on every render.** `Rails.cache.fetch` stored nothing on failure, so a corrupt video re-attempted the full download+ffmpeg every page view. **Fix:** failures are negatively cached (1h TTL sentinel), successes cached permanently.

**Checked and sound (per review, no changes needed):** preview URLs are permanent public S3 URLs (`public: true` in storage.yml) so caching them is safe; `cdn_url_for` rewrite applies to preview variants exactly as to images; `previewable?` correctly guards missing ffmpeg; cache key scheme matches `url_from_file`; JW Player `image` typing correct; only other `.thumbnail` consumer (`CoverEditor.tsx:289`) explicitly excludes videos; og/twitter meta unaffected.

**P3s noted, not blocking:** v1 public API now returns a populated `covers[].thumbnail` for video covers (previously always null) — additive; API field docs could mention it. `rescue StandardError` breadth accepted for a cosmetic feature (and it correctly catches `Timeout::Error` now that the wrapper exists).

**Verification:** 7 model specs green (incl. two new: negative-cache no-retry, timeout degradation); rubocop clean.

slavingia:
Would it be better to do the thumbnail creation etc during video upload or is this have to be like this?

gumclaw:
Good call — it didn't have to be request-time, and upload-time is strictly better. Reworked it in aac89d5:

- New `GenerateVideoPosterWorker` (low queue) does the download + ffmpeg frame grab in the background. It's enqueued when the cover is created, so the poster is normally ready before anyone views the product, and it busts the product cache when the poster lands.
- `thumbnail_url` on a web request now only reads the cache — never downloads or runs ffmpeg. If nothing's cached yet (covers uploaded before this ships), it returns nil, the player falls back to today's idle state, and a generation gets enqueued so the poster shows up on later views.

Specs updated and green (12 model + 4 worker), rubocop clean. PR body updated to match. If you'd rather pre-warm all existing covers instead of the lazy backfill, a one-shot backfill job is a small follow-up.

slavingia:
Fable five adversarial review and merge

gumclaw:
## Fable-5 adversarial review — round 2 (post-rework, `aac89d5`)

**Verdict: clean. No P1/P2. Round-1 findings remain fixed, and the upload-time rework holds up under adversarial reading.**

What I hunted for and what I found:

- **Request-thread work (the round-1 P1):** confirmed gone. `thumbnail_url` → `video_poster_url` is cache-read-only on web requests; download + ffmpeg only run inside `generate_video_poster!`, which is only called from `GenerateVideoPosterWorker`. Cache misses enqueue and return `nil`.
- **Duplicate-job storm on hot products:** a popular product page could enqueue on every render until the poster lands, but the worker uses `lock: :until_executed` (sidekiq-unique-jobs, same pattern as `stamp_pdf_for_purchase_job` etc.), so concurrent enqueues dedupe. Fine.
- **Enqueue-at-create timing:** `after_create_commit` fires after the attachment row commits, so `file.attached?` is true inside the callback. Video-only guard is correct.
- **Failure containment:** worker failures cache the empty-string sentinel for an hour (bounded retries via `retry: 3` + no re-enqueue while the sentinel lives), and `generate_video_poster!` rescues `StandardError` so a corrupt file can't spin the retry queue. Timeout matches the existing `retina_variant` pattern (`IMAGE_PROCESSING_TIMEOUT_SECONDS`).
- **Stale product-page cache:** worker calls `link.invalidate_cache` only when a poster was actually generated — matches `Product::Caching#invalidate_cache`, so cached JSON picks the poster up without gratuitous busts on failure.
- **Cache-key collision:** `attachment_#{file.id}_poster_url` is disjoint from the existing `attachment_#{file.id}_#{style}_url` keys (`:retina`/`:original` never equal `poster`).
- **Frontend blast radius of the `thumbnail` field change:** swept every consumer of `AssetPreview.thumbnail`. `CoverEditor.tsx` explicitly guards `cover.type !== "video"`, so the editor's rendering is unchanged; `Covers/Video.tsx` is the only new reader. `image` is a valid `jwplayer.PlaylistItem` key under the local `JWPlayerOptions` type, and `null → undefined` preserves today's black-idle fallback exactly.
- **ffmpeg availability:** both `docker/base/Dockerfile` and `Dockerfile.test` install ffmpeg, and `file.previewable?` short-circuits gracefully if the previewer isn't usable — worst case is no poster, never an error.

**Local verification (fresh worktree at `aac89d5`):**
- `spec/sidekiq/generate_video_poster_worker_spec.rb` — 4 examples, 0 failures
- `spec/models/asset_preview_spec.rb` — 54 examples, 1 failure: `#url=` square-bracket encoding at line 155, reproduced identically on unmodified `origin/main` (pre-existing, unrelated)

Auto-merge (squash) armed; merges when the in-flight Tests run goes green.

PR #5863, opened 2026-07-13
Update Singapore GST rate in help article from 8% to 9%

gumclaw:
## What

One-word fix in the Singapore GST help article (`_336-singaporean-gst`): 8% → 9%.

## Why

Singapore raised GST from 8% to 9% effective January 1, 2024. Checkout already charges the correct current rate — the production `zip_tax_rates` table has a 9% row with `applicable_years: [2024]` (verified live via the audited prod console, request `20260713T191401Z-79624369`) and `SalesTaxCalculator` picks the rate whose applicable years include the current year. Only the help article was stale.

A Singapore-based creator flagged this via support today (Helper ticket `978d425fe6470f6a8edb5736f1675b21`); they were answered with the correct 9% figure and told the article would be fixed.

## Verification

- ERB syntax check passes on the edited partial
- `bundle exec rspec spec/models/help_center` green locally (1 example, 0 failures)
- `articles.yml` untouched (body-only edit; title/description don't mention the rate)
- Grep confirms no other help article states the 8% figure


gumclaw:
**Fable-5 pre-merge review: READY-TO-MERGE** (no P1/P2)

What the adversarial pass confirmed:
- **The 9% claim holds in 2026 despite the 2024-only `applicable_years`**: the SG branch in `sales_tax_calculator.rb:160-168` falls back to the row with the latest applicable year when the current year matches nothing, which selects the 0.09 row. Verified against the live prod tax table (audited console read `20260713T191401Z-79624369`: 0.08→[2023], 0.09→[2024]).
- No other help-center article, mailer, or `articles.yml` entry states the stale 8% figure.
- Diff is exactly one line inside an existing `<p>`; ERB parses and compiles cleanly; `spec/models/help_center` green locally.

P3 (non-blocking, noted for the record): the article's pre-existing "physical products are exempt" sentence isn't visible in the calculator's SG branch itself — likely enforced elsewhere (physical-product gating upstream), pre-existing content untouched here.

PR #5864, opened 2026-07-13
Stop EasyPost slowness from freezing physical-product checkout for 30+ seconds

gumclaw:
## What

Caps how long checkout waits on EasyPost when verifying a buyer's shipping address. The EasyPost client now uses a 5-second connect / 10-second read deadline instead of the gem's defaults (30s connect / 60s read). Adds specs for the timeout fail-open path and the client configuration.

## Why

CI has been failing repo-wide since 18:38 UTC today — every `Tests` run on main and on every PR branch fails the same three shipping-address system specs (`taxes_spec.rb:140`, `non_tiered_membership_spec.rb:418`, `tiered_membership_fixed_length_spec.rb:33`), while the last green main run was 18:08 UTC ([failing main runs: 29275529239, 29275925423, 29283468058, 29284309141](https://github.com/antiwork/gumroad/actions/runs/29283468058)).

The test logs show the mechanism: `POST /shipments/verify_shipping_address` completes in **30,007 ms** — exactly the EasyPost gem's default 30-second connect timeout. The specs hit real EasyPost (no VCR cassette for that call), EasyPost has been slow/unreachable from CI since this afternoon, and the buyer-side "Processing..." state outlasts the specs' wait windows, so they fail deterministically. The failure screenshot shows checkout frozen at "Processing..." mid-purchase.

This is also a production-facing problem, not just a CI one: any EasyPost slowness makes every physical-product checkout hang for up to 30–90 seconds before the existing fail-open path (rescue `EasyPost::Errors::EasyPostError` → accept the address as entered) takes over. The gem wraps `Net::OpenTimeout`/`Net::ReadTimeout` in `EasyPost::Errors::TimeoutError`, so the shorter deadlines flow into the fail-open branch that already exists — verification degrades gracefully in seconds instead of appearing broken for half a minute.

## Test plan

- `spec/controllers/shipments_controller_spec.rb`: 20 examples, 0 failures locally, including two new ones — the timeout fail-open (accepts the address, notifies) and the client deadline configuration.
- CI: the three failing shipping specs should recover on this branch once EasyPost either responds within 15s or times out fast enough for the fail-open dialog flow the specs already handle (`confirm_shipping_address_if_prompted`).
- After deploy: `verify_shipping_address` p99 latency should cap near 15s, and `ErrorNotifier` will report `EasyPost::Errors::TimeoutError` whenever EasyPost is degraded.

Not a UI change — no pixels rendered differently; the fix only shortens a network deadline.

### DIFF
--- a/app/controllers/shipments_controller.rb
+++ b/app/controllers/shipments_controller.rb
@@ -13,8 +13,20 @@ class ShipmentsController < ApplicationController
     "country" => "country"
   }.freeze
 
+  # How long we're willing to wait for EasyPost when verifying a buyer's shipping address.
+  # The gem's defaults are 30 seconds to connect and 60 seconds to read — when EasyPost is
+  # slow or unreachable, the buyer sits on a frozen "Processing..." checkout for that whole
+  # time before we fail open. Address verification is a nice-to-have (we already accept the
+  # address as entered when EasyPost errors), so give it a few seconds and then move on.
+  EASYPOST_OPEN_TIMEOUT_SECONDS = 5
+  EASYPOST_READ_TIMEOUT_SECONDS = 10
+
   def verify_shipping_address
-    easy_post = EasyPost::Client.new(api_key: GlobalConfig.get("EASYPOST_API_KEY"))
+    easy_post = EasyPost::Client.new(
+      api_key: GlobalConfig.get("EASYPOST_API_KEY"),
+      open_timeout: EASYPOST_OPEN_TIMEOUT_SECONDS,
+      read_timeout: EASYPOST_READ_TIMEOUT_SECONDS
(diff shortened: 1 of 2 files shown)


gumclaw:
Pre-merge adversarial review (fable-5) done on the diff: EasyPost::Client keyword support verified in gem source (client.rb:19), timeout errors confirmed to route through the existing fail-open rescue (http_client.rb wraps Net::OpenTimeout/Net::ReadTimeout in EasyPost::Errors::TimeoutError, a subclass of EasyPostError), and the fail-open response keeps checkout moving with the address as entered. No P1/P2 findings. Controller spec suite: 20 examples, 0 failures locally.

PR #5866, opened 2026-07-14
Fix ActiveRecord::RecordNotSaved when recording payment flow on an unsaved purchase

gumclaw:
## What

Guards the checkout payment-flow analytics recording in `Purchase::CreateService#perform` so it only runs when the purchase actually persisted. Previously, when `prepare_for_charge!` left the purchase unsaved (a validation failure such as an invalid email), the subsequent `purchase.create_purchase_payment_flow(...)` raised `ActiveRecord::RecordNotSaved`, which the surrounding rescue reported to Sentry on every such failed checkout.

## Why

Sentry issue [GUMROAD-XZ](https://gumroad-to.sentry.io/issues/7609043470/) — `ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved` in `OrdersController#prepare` → `Purchase::CreateService#perform` (`app/services/purchase/create_service.rb:151`). Rails refuses to create a `has_one` dependent row on an unsaved parent. A purchase that never saved has no analytics row worth recording, so the correct behavior is to skip recording rather than report noise to Sentry. The buyer-facing outcome (validation error returned to the client) is unchanged.

## Test plan

- New specs in `spec/services/purchase/create_service_spec.rb` ("payment flow analytics recording"):
  - a successfully built combined-charge purchase still records its `purchase_payment_flow`
  - an unpersisted purchase (validation failure) skips recording, does not call `ErrorNotifier.notify`, and still returns the validation error
- The failing-validation spec reproduces the exact `ActiveRecord::RecordNotSaved` on `main` (ErrorNotifier receives the exception) and passes with the fix.
- Ran locally: `bundle exec rspec spec/services/purchase/create_service_spec.rb -e "payment flow analytics recording"` → 2 examples, 0 failures; `spec/services/order/create_service_spec.rb -e "recording the payment flow"` → 9 examples, 0 failures; `spec/models/purchase_payment_flow_spec.rb` green. Rubocop clean on changed files.
- After deploy: the Sentry issue should stop receiving events.

### DIFF
--- a/app/services/purchase/create_service.rb
+++ b/app/services/purchase/create_service.rb
@@ -146,7 +146,12 @@ class Purchase::CreateService < Purchase::BaseService
       purchase.build_purchase_wallet_type(wallet_type: params[:wallet_type]) if params[:wallet_type].present?
 
       payment_flow_attributes = PurchasePaymentFlow.attributes_for_checkout_params(params)
-      if payment_flow_attributes && !purchase.free_purchase?
+      # The purchase may still be unsaved here: `prepare_for_charge!` leaves it
+      # unpersisted when a validation fails (for example, an invalid email).
+      # Creating the dependent analytics row on an unsaved parent raises
+      # ActiveRecord::RecordNotSaved, and a purchase that never saved has no
+      # payment flow worth recording anyway.
+      if payment_flow_attributes && purchase.persisted? && !purchase.free_purchase?
         begin
           purchase.create_purchase_payment_flow(payment_flow_attributes)
         rescue => e
--- a/spec/services/purchase/create_service_spec.rb
+++ b/spec/services/purchase/create_service_spec.rb
@@ -3833,4 +3833,48 @@ describe Purchase::CreateService, :vcr do
       end
     end
   end
+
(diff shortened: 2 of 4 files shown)


slavingia:
When would this occur?

gumclaw:
On any failed checkout where the purchase doesn't pass validation. `prepare_for_charge!` calls `process_without_charging!`, which uses plain `save` — if a validation fails (for example an invalid buyer email failing `must_have_valid_email` — the case reproduced in the spec — but any purchase validation failure hits the same path), the purchase stays unpersisted with errors and `perform` keeps going. The very next block tries `purchase.create_purchase_payment_flow(...)`, and Rails raises `ActiveRecord::RecordNotSaved` because you can't create a `has_one` dependent on an unsaved parent.

So concretely: a buyer submits checkout with a bad email (or anything else that fails purchase validation) while the request includes payment-flow params (`payment_details_source` + a Stripe payment method). The buyer still gets the normal validation error either way — the exception only added a Sentry report per failed checkout. That's why the fix is to skip recording rather than change any behavior: a purchase that never saved has no analytics row to record.

PR #5210, opened 2026-05-21
Bootstrap Minitest suite; move 40 pure-Ruby specs out of RSpec

ershad:
## What

Bootstraps Rails' built-in Minitest runner under `test/` and moves the cheapest 40 specs out of RSpec/FactoryBot. Every file in this PR has zero `create(:...)` calls and no DB access — the conversion is a syntactic rewrite, not a behavior change.

### Infra

- `test/test_helper.rb` (Minitest + `ActiveSupport::TestCase`), `file_fixture_path` pointed at the existing `spec/support/fixtures` so shared fixture files keep working as more tests migrate over later.
- Pin `minitest ~> 5.25` in the Gemfile. The transitively-pulled Minitest 6.0 is incompatible with Rails 7.1's `TestUnit::LineFiltering` (signature drift on `Runnable#run`).
- New `test_minitest` job in `.github/workflows/tests.yml`. Runs `bin/rails test` against a stripped-down `docker-compose-test-and-ci.yml` (only `db_test` and `redis`, no ES / mongo / minio). Wired into the `unblock_deployment_from_buildkite` gate alongside `test_fast` / `test_slow`.

### Batch 1 — 20 specs (helpers, validators, lib/utilities, lib/errors)

- Helpers: `button`, `cdn_url`, `currency`, `infos`, `mailer`, `preorder`, `social_share_url`
- Validators: `email_format`, `isbn`, `not_reserved_email_domain`
- `lib/utilities`: `credit_card_utility`, `csv_safe`, `d3`, `global_config`, `log_redactor`, `referrer`, `text_scrubber`, `us_zip_codes`, `xml_helpers`
- `lib/errors`: `gumroad_runtime_error`

### Batch 2 — 20 more (services, sidekiq, presenters, observers, lib constraints)

- Services: `adult_keyword_detector`, `email_redactor_service`, `firs_tin_validation_service`, `home_page_link_service`, `integrations/base_integration_service`, `kra_pin_validation_service`, `oman_vat_number_validation_service`, `redis_key`, `tra_tin_validation_service`, `trn_validation_service`
- Sidekiq: `backfill_price_checker_index_fields_job`, `handle_paypal_event_worker`, `handle_stripe_event_worker`, `perform_daily_instant_payouts_worker`, `perform_payouts_up_to_delay_days_ago_worker`
[...]

PR #5867, opened 2026-07-14
Scaffold the Minitest + fixtures suite; seed with 5 policy specs

gumclaw:
## What

Bootstraps a Minitest + Rails-fixtures suite under `test/` and seeds it with 5 migrated policy specs, so subsequent factory-removal migrations are mechanical, small, and quickly mergeable:

- `test/test_helper.rb` — boot, WebMock, Sidekiq fake mode, a global Elasticsearch stub (prevents Makara pool saturation from EsClient retries), Makara blacklist disable, Vite manifest stub, and the same feature-flag activations `spec_helper.rb` applies globally.
- `test/support/` — `PolicyAssertions` (replicates RSpec `permissions :a, :b` blocks as one-line assertions) and `ConstantStubbingHelpers` (`with_const`, the Minitest replacement for RSpec's `stub_const`).
- `test/fixtures/` — the standard `named_seller` + 4-role team shape used by policy tests. Fixture emails/usernames are namespaced (`fixture-seller@example.com`, `fixtureseller`) so fixture rows and FactoryBot's fixed-value factories (`:named_seller`) can coexist in the shared test database while both suites run.
- 5 policy specs migrated to `test/policies/` (application, analytics, audience, balance, dashboard); their `spec/policies/` counterparts deleted.
- `test_minitest` CI job in `tests.yml` — bare `ubuntu-22.04` runner with native service containers (mysql/redis/stripe-mock/mongo), no Docker test image, `bundler-cache: true`. Wired into `unblock_deployment_from_buildkite` alongside `test_fast`/`test_slow`.
- `minitest ~> 5.25` pin (rubygems has a stub `minitest 6.0.6` with no code that bundler otherwise resolves to, breaking `require "minitest/mock"`).
- `docs/minitest-migration.md` — the per-file migration recipe, fixture gotchas (validation/callback/default bypass, `FixtureSet.identify` for renamed FKs, `before(:create)` hooks becoming explicit rows), and a verification gate for autonomously merging future migration PRs (test-only diff, example-count parity or itemized delta, green suite).

## Why

Part of #5801 (factory removal is the remaining item to close it). The full-suite profiling pass showed models/controllers/mailers/jobs specs spend 58–77% of wall time inside `factory.create`; fixtures load once per process and drop per-test setup cost to near zero. Prior scaffolding attempts (#5210, #5244) were closed unmerged — #5244 bundled ~2,700 files of bulk migration with out-of-scope drift. This PR is the fresh, minimal start: scaffolding + a 5-file seed that locks in conventions, with the playbook documented so follow-up migrations can be produced and verified mechanically in profiling-rank order (models → mailers → small dirs → controllers → jobs).

## Test plan

- Locally (mysql/redis/mongo/stripe-mock up): `RAILS_ENV=test bundle exec rails test` → **26 runs, 45 assertions, 0 failures, 0 errors, 0 skips** in 0.22s of test time.
- RSpec coexistence check: `bundle exec rspec spec/policies/churn_policy_spec.rb` (factory-based, shares the DB with the loaded fixtures) → 6 examples, 0 failures.
- `rubocop -a` clean on all new/changed Ruby files.
- CI: new `Test Minitest` job must go green alongside the untouched `test_fast`/`test_slow` matrix; deployment gate now requires all three.


gumclaw:
**Pre-merge adversarial review** (fable-5, local, diff at `HEAD` vs `origin/main`). Hunted for: fork-PR security-gate regressions in the workflow, deploy-gate wiring gaps, fixture/factory identity collisions, silent coverage loss from the spec deletions, and Knapsack fallout.

| Area | Finding | Severity |
|---|---|---|
| `tests.yml` fork-PR boundary | `test_minitest` copies the exact `ci_gate` + internal-PR-noop `if:` guard used by the other jobs; checkout pins `pull_request_target` runs to `head.sha` with `persist-credentials: false`. No new secret exposure path. | none |
| Deployment gate | `unblock_deployment_from_buildkite` now needs `test_minitest`; on `push` to main the job always runs (its `if:` only excludes internal `pull_request_target`), so the gate can't be skipped-satisfied. | none |
| ci/green reporting | `ci-green.yml` reports the whole Tests *workflow* conclusion, so the new job is covered automatically — no required-check edits needed. | none |
| Fixture/factory collision | Original fixture rows reused `seller@example.com` / `seller`, identical to the fixed-value `:named_seller` factory — RSpec specs failed with uniqueness violations when both suites shared a local DB. **Fixed pre-push** by namespacing all fixture emails/usernames (`fixture-*`); both suites verified green back-to-back against the same DB. In CI the suites use separate DB lifecycles, so this was a local-dev-only hazard, but worth the guard. | P2, fixed |
| Coverage parity | 26 Minitest examples replace 26 RSpec examples across the 5 files, assertion-for-assertion (checked file by file). The deleted `balance_policy_spec` had two examples asserting `:dashboard` instead of `:balance` (copy-paste bug in the original); the migrated tests assert `:balance` — strictly stronger. | none (improvement) |
| Knapsack | Queue Mode distributes files dynamically from what exists on disk; deleted specs simply drop out. No checked-in timing report references them. | none |

No P1s. The one P2 (fixture identity collision) was fixed before push. Verdict: safe to merge on green.

PR #5868, opened 2026-07-14
Stop alerting on Cash App Pay's browser-handled Stripe next action

gumclaw:
## What

Stops `StripeChargeIntent` / `StripeSetupIntent` from reporting `cashapp_handle_redirect_or_display_qr_code` as an "unsupported action" to Sentry. Adds a `StripeIntentStatus::CLIENT_HANDLED_ACTION_TYPES` list for next-action types that Stripe.js resolves entirely in the buyer's browser, and skips the `ErrorNotifier.notify` call for those. Genuinely unsupported types (e.g. `redirect_to_url`) still alert.

## Why

Since Cash App Pay launched on the client-confirmed checkout path (#5686), a buyer can land on `Checkout::ReturnsController#show` while their PaymentIntent still carries the `cashapp_handle_redirect_or_display_qr_code` next action — e.g. they closed the QR code without paying and the browser redirected them back. The return page retrieves the intent, `validate_next_action` sees a non-`use_stripe_sdk` action, and fires a Sentry error.

This is expected state, not a defect: that action type is handled by Stripe.js during `confirmPayment` in the browser, never by the server-driven SCA flow. The `requires_action?` behavior is unchanged — the purchase still resolves to processing/failed via the existing finalization path — so the only change is removing the false alert.

Sentry: production error on `Checkout::ReturnsController#show`, "Stripe charge intent … requires an unsupported action: cashapp_handle_redirect_or_display_qr_code".

## Test plan

- New specs: `cashapp_handle_redirect_or_display_qr_code` does not notify the error tracker (both charge intent and setup intent) and is still not treated as a server-supported action; existing "unsupported action notifies" specs (`redirect_to_url`) unchanged and passing.
- Ran locally: `bundle exec rspec spec/business/payments/charging/implementations/stripe/stripe_charge_intent_spec.rb spec/business/payments/charging/implementations/stripe/stripe_setup_intent_spec.rb` — all pass except 7 pre-existing failures in the "payment intent is canceled" contexts, which fail identically on unmodified `main` (verified via `git stash`) and are unrelated to this change.
- Rubocop clean on all changed files.

## Before/After

Not applicable — no UI change; this only removes a false-positive Sentry alert.

### DIFF
--- a/app/business/payments/charging/implementations/stripe/helpers/stripe_intent_status.rb
+++ b/app/business/payments/charging/implementations/stripe/helpers/stripe_intent_status.rb
@@ -7,4 +7,11 @@ class StripeIntentStatus
   PROCESSING = "processing"
   CANCELED = "canceled"
   ACTION_TYPE_USE_SDK = "use_stripe_sdk"
+
+  # Next-action types that Stripe.js resolves entirely in the buyer's browser on the
+  # client-confirmed checkout path (the browser calls stripe.confirmPayment and Stripe.js
+  # shows the QR code / performs the redirect itself). Seeing one of these on a retrieved
+  # intent is expected — for example, a buyer who returns to the checkout return page
+  # without finishing the Cash App QR flow — so it is not an error worth alerting on.
+  CLIENT_HANDLED_ACTION_TYPES = ["cashapp_handle_redirect_or_display_qr_code"].freeze
 end
--- a/app/business/payments/charging/implementations/stripe/stripe_charge_intent.rb
+++ b/app/business/payments/charging/implementations/stripe/stripe_charge_intent.rb
@@ -43,8 +43,15 @@ class StripeChargeIntent < ChargeIntent
     end
 
     def validate_next_action
-      if payment_intent.status == StripeIntentStatus::REQUIRES_ACTION && payment_intent.next_action.type != StripeIntentStatus::ACTION_TYPE_USE_SDK
-        ErrorNotifier.notify "Stripe charge intent #{id} requires an unsupported action: #{payment_intent.next_action.type}"
-      end
+      return unless payment_intent.status == StripeIntentStatus::REQUIRES_ACTION
+
+      next_action_type = payment_intent.next_action.type
+      return if next_action_type == StripeIntentStatus::ACTION_TYPE_USE_SDK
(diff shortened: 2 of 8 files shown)


slavingia:
Can you QA this on a preview app?

gumclaw:
## Preview app QA — PASS ✅

QA'd on the preview app (`fix-cashapp-next-action-error-no.apps.staging.gumroad.org`, serving `x-revision: 339422c0d` — this PR's head). Flags enabled: `stripe_payment_element_checkout` + `stripe_payment_element_client_confirm`. Drove the real buyer UI with Playwright (headed), then verified the alerting change on the preview Rails console against the live intent.

### 1. Real Cash App Pay checkout via the buyer UI

Payment Element rendered `Card | Cash App Pay` tabs; selecting Cash App Pay and clicking Pay opened Stripe.js's QR lightbox — confirming this next action is browser-handled, exactly the premise of the fix.

| Payment Element | QR lightbox (browser-handled next action) |
|---|---|
| payment element | qr lightbox |

### 2. Reproduced the exact production scenario: buyer abandons the QR → lands on the return page

Closed the QR without paying and navigated to `/checkout/returns/<order>?payment_intent=pi_2TsvzN9e…` while the intent still carried `status=requires_action` + the cashapp next action. The return page handled it gracefully — redirected back to `/checkout` with the cart restored, no crash:

return outcome

### 3. Alert suppression proven on the preview console (against the live abandoned intent)

Spy on `ErrorNotifier.notify`, run the suppressed case AND a control case in the same session:

```
MARK live_intent status=requires_action next_action_type=cashapp_handle_redirect_or_display_qr_code
MARK after_cashapp notify_count=0 requires_action=false processing=false succeeded=false
MARK after_control notify_count=1 last_msg="Stripe charge intent pi_qa_control requires an unsupported action: redirect_to_url"
```

- The real abandoned intent (exactly the state that fires the false Sentry alert in prod) → **0 notifies**, and `requires_action?` stays `false` (behavior unchanged).
- Control with `redirect_to_url` → **1 notify** — genuinely unsupported actions still alert, so the guard is targeted, not a blanket suppression.

Note: the happy-path Cash App purchase (Simulate scan → receipt) is pre-existing #5686 behavior and not changed by this PR; the change under test is only the alert suppression, verified above in both directions.

rmarescu (approved):

PR #5870, opened 2026-07-14
Fix Ai::AnthropicClient::Error when a stream drops mid-tool-call (retry instead of failing the turn)

gumclaw:
## What

Treats a streamed Anthropic tool call whose JSON is cut off by a mid-stream disconnect as a retryable network failure instead of a hard `Ai::AnthropicClient::Error`. `assemble_tool_uses` now distinguishes three cases when a tool block's input fails to parse:

- `stop_reason == "max_tokens"` → drop the block (unchanged — the token cap cut it off)
- `stop_reason` is nil → the stream never delivered a `message_delta`/stop_reason, meaning the connection dropped mid-tool-call. Raise `TransientError` so the existing retry loop in `#stream_messages` re-requests the turn (which it only does while nothing has been streamed to the caller).
- completed turn with unparseable input → still a real `Error` (unchanged)

## Why

Sentry GUMROAD-Y4 (https://gumroad-to.sentry.io/issues/7609201485/): production raised `Ai::AnthropicClient::Error: Anthropic produced an unreadable tool call for api_write.` from `parse_tool_use_input`, with the underlying cause `JSON::ParserError: expected ',' or '}' after object value, got: EOF at line 1 column 1488` — the classic shape of a stream that died mid-arguments. A complete Anthropic stream always sends a stop_reason before ending, so accumulated-but-unparseable tool JSON with no stop_reason is a dropped connection, not the model misbehaving. Today that surfaces to the seller as a failed agent turn even though a retry (safe here — no output was yielded) would have succeeded.

## Test plan

- New spec: a stream that ends mid-tool-call with no stop_reason is retried and the second attempt's tool call is returned (`sleep` observed once). Verified load-bearing: fails with `unreadable tool call` when the fix is stashed.
- Updated spec: malformed tool input on a *completed* turn (stop_reason present) still raises `Error`.
- Existing max_tokens-truncation spec unchanged and green.
- Local run: `spec/services/ai/anthropic_client_spec.rb` 31 examples, 0 failures; `spec/services/ai/store_agent_service_spec.rb` 31 examples, 0 failures. `rubocop -a` clean on both files.

### DIFF
--- a/app/services/ai/anthropic_client.rb
+++ b/app/services/ai/anthropic_client.rb
@@ -405,10 +405,15 @@ class Ai::AnthropicClient
 
     # Turn the accumulated streamed blocks into the same tool_use shape #parse_message returns. A
     # tool_use block with no input fragments is a no-arg call; malformed non-empty input must fail the
-    # turn instead of dispatching a lossy {} tool call — EXCEPT when the model was cut off by
-    # max_tokens. A turn truncated mid-tool-call always ends with half-written (unparseable) JSON;
-    # that isn't the model misbehaving, it's the token cap. Dropping the cut-off block and letting
-    # the caller see stop_reason == "max_tokens" lets it respond honestly instead of erroring out.
+    # turn instead of dispatching a lossy {} tool call — with two exceptions where half-written JSON
+    # is expected rather than the model misbehaving:
+    #   - stop_reason == "max_tokens": the token cap cut the call off mid-arguments. Drop the block
+    #     and let the caller see the stop_reason so it can respond honestly instead of erroring out.
+    #   - stop_reason is nil: a complete Anthropic stream always delivers a stop_reason (via
+    #     message_delta) before it ends, so its absence means the connection dropped mid-stream and
+    #     the tool call's JSON was cut off by the disconnect. That is a network failure, not a bad
+    #     tool call, so raise TransientError — the retry loop in #stream_messages re-requests the
+    #     turn when nothing has reached the caller yet, exactly as it does for other dropped streams.
     def assemble_tool_uses(blocks, stop_reason: nil)
       truncated = stop_reason == "max_tokens"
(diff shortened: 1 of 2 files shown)


gumclaw:

Hunted specifically for:
- **Retry-loop reachability:** `assemble_tool_uses` runs inside the `with_retries` block of `#stream_messages`, so the new `TransientError` is caught and retried exactly like other dropped-stream failures. Confirmed by the new spec (`sleep` observed once, second attempt succeeds).
- **Already-streamed-output path:** when text has reached the caller, `retryable` vetoes the retry and the `TransientError` surfaces. `TransientError < Error`, so every existing `rescue Ai::AnthropicClient::Error` call site handles it unchanged — no new failure mode for the seller-facing path.
- **Cross-attempt state:** `text`/`blocks`/`stop_reason` are re-initialized per attempt inside the retried block; no pollution between attempts.
- **Retry budget:** no new sleep paths; the existing `RETRY_SLEEP_BUDGET_IN_SECONDS` accounting applies.
- **Adjacent edge (out of scope, pre-existing):** a stream that drops after a *complete* tool call but before `message_delta` still dispatches with a nil stop_reason — behavior identical before and after this diff.

Load-bearing test verified: the new disconnect spec goes red when the fix is stashed.

slavingia:
Fix spec. Maybe in new pr

gumclaw:
Done — spec fix is in a new PR: #5871.

Root cause: the report writes sale IDs through `CsvSafe`, which prefixes cells starting with `-` with an apostrophe (formula-injection guard). Purchase external IDs are random URL-safe base64, so ~1 in 33 runs draws a leading dash and the raw `external_id` expectation stops matching. The spec now runs its expected sale-ID cells through the same sanitization. Reproduced the exact CI value locally; 4/4 green.

Also re-ran the failed Test Fast 3 shard on this PR — the flake predates this branch (the spec is on main, untouched here).

PR #5871, opened 2026-07-14
Fix flaky tax transaction report spec when an external ID starts with a dash

gumclaw:
## What

Makes `spec/services/exports/tax_summary/transaction_report_spec.rb` deterministic. The report writes sale IDs through `CsvSafe`, which prefixes any cell starting with `-` with an apostrophe to block spreadsheet formula injection. Purchase external IDs are random URL-safe base64, so roughly 1 in 33 runs generates an ID starting with a dash and the spec's raw `purchase.external_id` expectation no longer matches the sanitized cell. Expected sale-ID cells now go through the same `CsvSafe` sanitization the report uses, via a small `csv_sale_id` helper. When the ID has no leading dash the helper returns it unchanged, so both cases pass.

## Why

The spec failed on Test Fast 3 of the CI run for #5870 (unrelated PR):

```
expected: [..., "-9sNWzxvQy7b-lqDzmWbxw==", ...]
     got: [..., "'-9sNWzxvQy7b-lqDzmWbxw==", ...]
```

The generated external ID started with `-`, so `CsvSafe#sanitize_field` prepended `'`. This is inherent randomness in the spec, not a behavior change — the same failure will keep hitting arbitrary PRs until fixed. Follow-up requested on #5870 ("Fix spec. Maybe in new pr").

## Test plan

- Reproduced the exact CI value locally: passing `-9sNWzxvQy7b-lqDzmWbxw==` through `CsvSafe.generate` yields `'-9sNWzxvQy7b-lqDzmWbxw==`, matching the CI "got" value byte for byte.
- Local run: `spec/services/exports/tax_summary/transaction_report_spec.rb` — 4 examples, 0 failures.
- `rubocop -a` clean on the spec file.

## Before/After

Not applicable — test-only change, no UI or behavior impact.

### DIFF
--- a/spec/services/exports/tax_summary/transaction_report_spec.rb
+++ b/spec/services/exports/tax_summary/transaction_report_spec.rb
@@ -48,6 +48,17 @@ describe Exports::TaxSummary::TransactionReport do
     )
   end
 
+  # The report writes through CsvSafe, which prefixes any cell that looks
+  # like a spreadsheet formula (for example one starting with "-") with an
+  # apostrophe. Purchase external IDs are random URL-safe base64, so they
+  # sometimes start with "-" — expected sale-ID cells must go through the
+  # same sanitization or the comparison fails whenever the generated ID
+  # happens to draw a leading dash.
+  def csv_sale_id(*purchases)
+    cell = purchases.map(&:external_id).join("; ")
+    CSV.parse_line(CsvSafe.generate { |csv| csv << [cell] }).first
+  end
+
   def stub_transaction_list(charge_transactions: [], payment_transactions: [])
     { "charge" => charge_transactions, "payment" => payment_transactions }.each do |type, transactions|
       list = double("list_#{type}")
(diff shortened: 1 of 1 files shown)


gumclaw:

Hunted specifically for:
- **Helper fidelity to production path:** `csv_sale_id` runs the expected cell through the same `CsvSafe.generate` → `<<` → `sanitize_field` path the report uses, then `CSV.parse_line` round-trips it exactly as the assertions do via `CSV.parse(tempfile.read)`. Reproduced the CI failure value byte-for-byte: `-9sNWzxvQy7b-lqDzmWbxw==` → `'-9sNWzxvQy7b-lqDzmWbxw==`.
- **`CsvSafe.generate` subclass dispatch:** `CSV.generate` instantiates on the receiver class, so the block yields a `CsvSafe` and sanitization applies — same pattern the existing `spec/lib/utilities/csv_safe_spec.rb` relies on.
- **No-dash case regression:** when the generated ID doesn't start with `-`, `needs_sanitization?` is false and the helper returns the ID unchanged — verified by the local 4/4 green run (IDs that run happened to draw no leading dash).
- **Multi-purchase join order:** helper joins in argument order, matching the previous interpolated expectation; production ordering is untouched.

Test-only change, no production code modified.

PR #5872, opened 2026-07-14
Schedule review reminder emails from purchase success and let sellers turn them off

gumclaw:
## What

Fixes review reminder emails never being scheduled, and adds a seller setting to turn review reminders off.

- Moves review-reminder scheduling from `Order`'s `after_save` callback to the purchase state machine: when a purchase transitions to a reviewable success state (`successful`, `not_charged`, `gift_receiver_purchase_successful`), it asks its order to schedule the reminder. The existing one-shot guard (`review_reminder_scheduled_at.nil?`) and send-time re-check in `OrderReviewReminderJob` are unchanged.
- Adds a `disable_review_reminders` seller flag (User flag bit 54), wired through `Settings::MainController`, `SettingsPresenter`, and a new "Review reminders" toggle in Settings. `Purchase#eligible_for_review_reminder?` now returns false when the seller has disabled reminders, which covers both scheduling and send time.

## Why

Review reminder emails (`OrderReviewReminderJob` → `CustomerLowPriorityMailer.purchase_review_reminder` / `.order_review_reminder`) have never been sent to anyone. Production has zero orders with `review_reminder_scheduled_at` set, ever.

Root cause: the scheduling hook was `after_save` on `Order`, guarded by `purchases.any?(&:eligible_for_review_reminder?)`, which requires a purchase in a successful state. But the order row is saved exactly once, at creation, while its purchases are still `in_progress` — they succeed seconds later without ever touching the order row again, so the callback never fired with an eligible purchase. Production order rows show `created_at == updated_at` with `purchase.succeeded_at` 3–6 seconds later.

The seller opt-out exists because turning this on for the first time platform-wide changes what every seller's customers receive; sellers who don't want the reminder emails need a way out.

Tracking issue: https://github.com/antiwork/gumroad-private/issues/1079

## Test plan

- `spec/models/purchase_spec.rb` — new coverage: reminder is scheduled when a purchase transitions to successful, not scheduled on failure, no-op without an order, and eligibility returns false when the seller disabled reminders (13 examples, 0 failures locally).
- `spec/models/order_spec.rb` + `spec/sidekiq/order_review_reminder_job_spec.rb` — existing behavior unchanged (31 examples, 0 failures locally).
- `spec/controllers/settings/main_controller_spec.rb` — new flag round-trips through settings update (40 examples, 0 failures locally).
- `spec/presenters/settings_presenter_spec.rb` — props include the new setting (67 examples, 0 failures locally).
- After deploy: watch `Order.where.not(review_reminder_scheduled_at: nil)` go from zero to tracking new order volume, then monitor `SentEmailInfo` rows for `order_review_reminder` / `purchase_review_reminder` and the weekly product-review creation rate to measure impact.

## Before/After

**Before:** no review-reminder setting existed; reminders silently never sent.

**After:** new "Review reminders" section in Settings, on by default with a seller opt-out:

| Toggle on (default) | Toggle off |
|---|---|
| settings-toggle-on | settings-toggle-off |

The emails this PR makes reachable are the existing `CustomerLowPriorityMailer#purchase_review_reminder` and `#order_review_reminder` (mailer methods and templates already in the codebase — this PR fixes the scheduling so they actually send). Rendered:

| Single purchase (`purchase_review_reminder`) | Multi-purchase order (`order_review_reminder`) |
|---|---|
| purchase-review-reminder | order-review-reminder |


app/presenters/settings_presenter.rb
```diff
@@ -75,6 +75,7 @@ def main_props
         announcement_notification_enabled: seller.announcement_notification_enabled,
         disable_comments_email: seller.disable_comments_email,
         disable_reviews_email: seller.disable_reviews_email,
```
slavingia, line 77: What's this?

gumclaw: This exposes the new seller setting to the settings page. It's the flag behind the "Review reminders" toggle this PR adds under Settings → Main — when a seller turns it off, buyers of their products stop getting the one-time review reminder email (checked in `Purchase#eligible_for_review_reminder?`). The presenter passes it to the React page alongside the existing email preferences like `disable_reviews_email` so the toggle renders with the seller's current value.

gumclaw: This exposes the new seller setting to the settings page so the "Review reminders" toggle can render its current state. It's the flag this PR adds (bit 54 on user flags): when a seller turns it on, buyers of their products stop receiving review reminder emails — the scheduling guard in `Order` checks `seller.disable_review_reminders?` before enqueueing the reminder.

It's separate from `disable_reviews_email` on the line above, which controls whether the *seller* gets an email when someone posts a review. This one controls whether *buyers* get the "leave a review" reminder.

app/controllers/settings/main_controller.rb
```diff
@@ -60,6 +60,7 @@ def user_params
         :announcement_notification_enabled,
         :disable_comments_email,
         :disable_reviews_email,
```
slavingia, line 62: What is reviews email currently used for? Seems similar in naming so confusing

gumclaw: `disable_reviews_email` is the existing "Reviews" toggle in the Notifications table above — it controls whether the *seller* gets an email when a buyer posts a review (`ContactingCreatorMailer.review_submitted`, gated in `ProductReview#notify_seller`).

The new `disable_review_reminders` is the opposite direction: it stops the reminder emails Gumroad sends to *buyers* asking them to review a purchase.

So: reviews email = seller notified a review came in; review reminders = buyers nudged to leave one. The UI keeps them separate (Notifications table vs. its own "Review reminders" section), but happy to rename the new flag (e.g. `disable_buyer_review_reminders`) if you want the code to be less ambiguous.

gumclaw: They go in opposite directions. `disable_reviews_email` is the existing seller-facing setting — it's the "Reviews" switch in the notifications table on this same page, and it gates the email the *seller* gets when a buyer submits a review (`ProductReview#notify_seller` → `ContactingCreatorMailer.review_submitted`).

`disable_review_reminders` (new here) is buyer-facing — it stops the reminder emails sent to *customers* asking them to review their purchase.

Happy to rename the new one to something like `disable_review_reminder_emails_to_customers` or move it under a clearer heading if that helps — though the UI copy ("Send review reminder emails to your customers") already spells out the direction.

slavingia:
Where the mailer method, template, screenshot?

slavingia:
Where the mailer method, template, screenshot?

gumclaw:
The mailer method and template already exist in the codebase — this PR fixes why they never sent, so I didn't add new ones:

- Mailer methods: `CustomerLowPriorityMailer#purchase_review_reminder` and `#order_review_reminder` (`app/mailers/customer_low_priority_mailer.rb`)
- Templates: `app/views/customer_low_priority_mailer/purchase_review_reminder.html.erb` and `order_review_reminder.html.erb`
- The bug was purely in scheduling: the `after_save` hook on `Order` never fired with an eligible purchase, so `OrderReviewReminderJob` was never enqueued. This PR moves scheduling to the purchase success transition and adds the seller opt-out.

Screenshots are now in the PR body: the new Settings toggle (on/off) and both rendered reminder emails, committed to `qa-media/` on this branch.

PR #5873, opened 2026-07-14
Always show license key settings expanded in the product content editor

gumclaw:
## What

Always shows the license key block's settings expanded in the product content editor, and removes the expand/collapse chevron. The seat-license toggle and the read-only product ID field (with its Copy button) are now visible whenever the block is on the page in edit mode.

## Why

The product ID required by the license verification API only appeared after clicking a small expand button inside the license key block. Creators looking for it couldn't find it — the API docs point them to the edit page, and the only visible ID there is the URL permalink, which doesn't work with the API. Surfacing the drawer permanently makes the product ID discoverable without any interaction.

Refs #5838.

## Before/After

Captured from a local run of the branch (desktop 1280px, mobile 375px).

**Before:**
| Desktop (1280px) | Mobile (375px) |
|---|---|
| before-desktop | before-mobile |

**After:**
| Desktop (1280px) | Mobile (375px) |
|---|---|
| after-desktop | after-mobile |

## Test plan

- `prettier` + `eslint` clean on the changed file
- No specs interact with the removed toggle (checked `spec/requests/products/edit/rich_text_editor_spec.rb` and grep for the drawer controls); existing rich-text editor specs cover the block's serialization, which is unchanged
- Verified live on a local boot of the branch: settings render without any click, seat toggle works, product ID copies; buyer-facing (read-only) view unchanged since the drawer only renders when `editor.isEditable`
- CI green

### DIFF
--- a/app/javascript/components/TiptapExtensions/LicenseKey.tsx
+++ b/app/javascript/components/TiptapExtensions/LicenseKey.tsx
@@ -1,4 +1,4 @@
-import { ChevronDown, ChevronUp, Key } from "@boxicons/react";
+import { Key } from "@boxicons/react";
 import { Node as TiptapNode } from "@tiptap/core";
 import { NodeViewProps, NodeViewWrapper, ReactNodeViewRenderer } from "@tiptap/react";
 import * as React from "react";
@@ -44,7 +44,6 @@ export const LicenseKey = TiptapNode.create({
 });
 
 const LicenseKeyNodeView = ({ editor, selected }: NodeViewProps) => {
-  const [isDrawerOpen, setIsDrawerOpen] = React.useState(false);
   const { licenseKey, isMultiSeatLicense, seats, onIsMultiSeatLicenseChange, productId } = useLicense();
   const uid = React.useId();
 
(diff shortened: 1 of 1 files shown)


gumclaw:
Pre-merge adversarial review (fable-5) — **VERDICT: PASS**, no P1/P2 code defects.

Findings and how each was handled:

| # | Finding | Severity | Resolution |
|---|---|---|---|
| 1 | Buyer-facing read-only view could leak the drawer | — | No finding: drawer gated on `editor.isEditable`; the only non-editor consumer (`Download/RichContent.tsx`) creates its editor with `editable: false` and its `licenseInfo` has no `productId` |
| 2 | Dangling imports/state after removing the toggle | — | No finding: `ChevronDown`/`ChevronUp` and `isDrawerOpen` fully removed |
| 3 | Specs/components referencing the removed button | — | No finding: all `"Close drawer"`/`"Edit"` spec hits belong to file-embed drawers, which keep their own chevrons |
| 4 | Rich-content serialization change | — | No finding: node persists as `{ "type": "licenseKey" }` with no attrs; `renderHTML`/`parseHTML` untouched |
| 5 | Drag ghost now shows the full expanded card instead of a compact row | P3 | Cosmetic only, accepted |
| 6 | Two "Copy" buttons with identical accessible names (license key vs product ID) | P3 | Pre-existing whenever the drawer was open; not a regression |
| 7 | No test proved the drawer is visible without a click | P2 (process) | Fixed in 9e8262870 — spec asserts the product ID field renders with no interaction and the old expand button is gone; passes locally |
| 8 | Evidence lacks dark-mode captures/video | P2 (process) | Before/after desktop + mobile screenshots from a local run are in the PR body; block has no dark-specific styling changes |

gumclaw:
Merged before the spec commit landed on the branch — the regression spec (and the qa-media screenshots this description links to) are in follow-up #5876, auto-merge armed.

PR #5875, opened 2026-07-14
Update license keys article for always-expanded block settings

gumclaw:
## What

Updates the "License keys" help article (`_76-license-keys`) to match the product content editor after #5873: the license key block's settings (seat toggle and product ID) are now always visible while editing, and the expand/collapse chevron is gone. Two sentences referenced the removed expand interaction:

- Multi-seat section: "Expand the License key module … and toggle on" → the toggle is shown directly under the block while editing.
- Verification parameters: `product_id` "get this by expanding the license key module" → it's shown in the License key block on the content page while editing.

## Why

The docs told creators to click an expand button that no longer exists. `product_id` discoverability was the whole point of #5873 (refs #5838), so the article pointing at the removed chevron would send readers hunting for a control that isn't there.

## Test plan

- `bundle exec rspec spec/models/help_center` — 1 example, 0 failures (slug ↔ partial integrity)
- Rendered the partial via `ApplicationController.render` in the test env — renders OK, new copy present, no stale "expand the module" wording remains
- Swept all article bodies for other references to expanding the license key block — none
- `articles.yml` untouched (body-only copy edit)

### DIFF
--- a/app/views/help_center/articles/contents/_76-license-keys.html.erb
+++ b/app/views/help_center/articles/contents/_76-license-keys.html.erb
@@ -33,7 +33,7 @@
     <p>This feature is currently only available for <a href="82-membership-products">Membership</a> products.</p>
   </div>
   <p>The “Multi-seat license” setting allows you to issue multiple seats for a single purchase or license key.</p>
-  <p>Expand the License key module on your content page and toggle on "Allow customers to choose number of seats per license purchased"</p>
+  <p>In the License key block on your content page, toggle on "Allow customers to choose number of seats per license purchased". The setting is shown directly under the block while you are editing.</p>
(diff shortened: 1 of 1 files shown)


gumclaw:
Pre-merge review (docs copy diff, reviewed against the #5873 code directly):

- **Accuracy:** both rewritten sentences match the shipped behavior — after #5873 the license key block's drawer renders whenever `editor.isEditable` (no chevron, no expand click), so "shown directly under the block while you are editing" and "shown in the License key block … while editing" are correct. The seat toggle only renders for membership products (`isMultiSeatLicense !== null`), and that section is already scoped by the membership-only callout.
- **Staleness sweep:** no other article body references expanding the license key block. `articles.yml` description doesn't mention the expand interaction. Clean.
- **P3 (not fixing here):** the multi-seat screenshot in the article predates #5873 and still shows the old drawer UI; the surrounding copy no longer depends on it, so leaving as-is. Worth refreshing if the article gets a screenshot pass.

Verdict: no P1/P2. Render check + `spec/models/help_center` green locally.

PR #5881, opened 2026-07-14
Show old and new addresses in the email-change security alert

gumclaw:
## What

Fixes the "Security alert: Your Gumroad account email is being changed" email so it always shows the old and new addresses. In the Google OAuth email-sync flow it previously rendered "changed from <new address> to ." — the new address in the "from" slot and a blank target.

## Why

The mailer view read `@resource.email` and `@resource.unconfirmed_email` at render time. That works for the normal pending-confirmation flow, but the OAuth sync applies and auto-confirms the change in one step, so by the time the mail renders `email` is already the new address and `unconfirmed_email` is nil. Devise already passes the pre-change address as the `to:` option (`send_devise_notification(:email_changed, to: devise_email_before_last_save)`), so the mailer now uses that as the old address instead of reading mutable state off the record.

Cosmetic but security-sensitive: every OAuth-driven email sync sent a confusing blank-target security alert, which generates support tickets.

Refs: antiwork/gumroad-private#1084

## Test plan

- New spec covering the auto-confirmed flow (record already holds the new address, `unconfirmed_email` nil, old address passed via `to:`), asserting the recipient and the rendered "from X to Y" line.
- Existing pending-confirmation spec unchanged and still passing.
- Ran locally: `bundle exec rspec spec/mailers/user_signup_mailer_spec.rb` — 8 examples, 0 failures.
- Rendered the mail with real mailer code (rails runner, OAuth auto-confirmed state) before and after the fix — see screenshots.

## Before/After

Rendered `email_changed` mail with `email = new@example.com`, `unconfirmed_email = nil`, `to: original@example.com` (the OAuth auto-confirmed state):

| Before | After |
|---|---|
| before | after |


gumclaw:
Fable-5 pre-merge review — verdict: **no P1/P2 findings.**

Hunt list and results:

1. **Pending-confirmation flow regression** — Devise fires `send_devise_notification(:email_changed, to: devise_email_before_last_save)`. In the reconfirmable flow the `email` column hasn't changed yet, so `to:` is the current (old) address and `unconfirmed_email` holds the new one → renders old → new, same as before. Covered by the existing spec, unchanged.
2. **OAuth auto-confirmed flow (the bug)** — `to:` carries the pre-save address; `unconfirmed_email` is nil so the new address falls back to `record.email`. Covered by the new spec and a real `rails runner` render (screenshots in the body).
3. **Async delivery** — `User::AsyncDeviseNotification` defers sends to `after_commit` but preserves the mailer args, so the fix reads the `to:` option rather than commit-time record state. No reintroduction of the race.
4. **Mailer preview** — `lib/mailer_previews` passes empty opts; both slots fall back to the current email. Preview-only cosmetic degradation, acceptable.

P3 (noted, not blocking): the mailer preview could pass a fake `to:` for a more realistic render.

Ran locally: `bundle exec rspec spec/mailers/user_signup_mailer_spec.rb` — 8 examples, 0 failures. Rubocop clean on touched files.

PR #5885, opened 2026-07-14
Add monthly Stripe balance summary email reports for all Antiwork entities

gumclaw:
## What

Adds a monthly `SendStripeBalanceSummariesReportJob` that pulls the Stripe Reporting API's three balance reports (`balance.summary.1`, `balance_change_from_activity.summary.1`, `payouts.summary.1`) for every Antiwork Stripe account — Gumroad, Flexile, Helper, Iffy — for the previous month, merges them into one dashboard-style CSV per account, and emails them to finance with gumclaw cc'd (same recipients as the funds-received report).

Also enriches the existing Stripe currency balances report: the CSV now shows the available / pending / connect_reserved split per currency, a total, and a USD-equivalent column (from the app's hourly-cached exchange rates), and its schedule moves from 11:00 to 00:05 UTC on the 1st so the point-in-time balance snapshot lands as close to month-end as possible.

## Why

The month-end close consumes Stripe balance summary reports for all four entities plus the Gumroad currency-balance snapshot. Today those are pulled outside the codebase by hand each month. Moving them into the codebase as scheduled email reports (following the funds-received report's mailer/cron conventions) makes them deterministic, retried on failure, and covered by the finance-report delivery backstop.

Refs antiwork/gumroad-private#1089.

Notes:
- The new job is scheduled on the **2nd** of the month, not the 1st: Stripe's balance reports only cover through ~24 hours ago, so a full previous month isn't queryable on the 1st.
- Gumroad's reports are pulled with `currency=usd` — it settles in 60+ currencies and an unfiltered pull adds a balance line per currency, which doesn't match the dashboard's USD view finance reconciles against. The other three accounts settle in USD only.
- Flexile/Helper/Iffy keys are read from `STRIPE_API_KEY_FLEXILE` / `STRIPE_API_KEY_HELPER` / `STRIPE_API_KEY_IFFY` (GlobalConfig: env or credentials). An entity with no key configured is skipped and named in the email body rather than failing the run — so this ships safely before the keys are added, and the email shows exactly what's missing. Restricted keys with Reporting + Files read access are sufficient.
- Both jobs register with `FinanceReportFailureAlert` (retry-exhaustion email with a pinned re-run command) and `VerifyFinanceReportsDeliveryJob` (daily backstop for vanished runs).

## Test plan

- New specs: `spec/business/payments/reports/stripe_balance_summaries_report_spec.rb` (CSV assembly, USD filter, month boundaries, report-run polling/failure), `spec/business/payments/reports/stripe_currency_balances_report_spec.rb` (bucket sums, single-unit currencies, missing-rate handling), `spec/sidekiq/send_stripe_balance_summaries_report_job_spec.rb`, plus mailer specs.
- Ran locally: 32 examples, 0 failures (new/touched specs + `verify_finance_reports_delivery_job_spec`). Two pre-existing `email_outstanding_balances_csv` failures in `accounting_mailer_spec.rb` are factory-seed issues unrelated to this diff.
- Mailer preview added: `/rails/mailers/accounting_mailer/stripe_balance_summaries_report`.
- After deploy: the currency balances report fires 00:05 UTC Aug 1, the balance summaries report 11:00 UTC Aug 2 — verify both land in the finance inbox before the August close.

## Before/After

Not applicable — no user-facing UI change; output is internal finance emails (see mailer preview).

### DIFF
--- a/app/business/payments/reports/stripe_balance_summaries_report.rb
+++ b/app/business/payments/reports/stripe_balance_summaries_report.rb
@@ -1,0 +1,142 @@
+# frozen_string_literal: true
+
+# Builds the month-end Stripe balance summary packet for every Antiwork Stripe account
+# (Gumroad, Flexile, Helper, Iffy). For each account it pulls three Stripe Reporting API
+# reports for the given month — balance summary, balance change from activity, and
+# payouts — and merges them into a single CSV that mirrors the Stripe dashboard's
+# Balance page layout. The CSVs are emailed to finance by
+# SendStripeBalanceSummariesReportJob as inputs to the monthly close.
+#
+# Gumroad settles in dozens of currencies, so its reports are pulled with a USD filter —
+# an unfiltered pull adds a starting/ending balance line for every settlement currency,
+# which doesn't match the dashboard's USD view that finance reconciles against. The
+# other accounts settle in USD only and need no filter.
+module StripeBalanceSummariesReport
+  # Each entry names an Antiwork Stripe account and where its restricted (reporting +
+  # files read) API key lives. Gumroad reuses the platform key already configured for
+  # the app; the other entities' keys are separate credentials. An entity whose key is
+  # not configured is skipped and reported as such in the email rather than failing the
+  # whole run.
+  ENTITIES = [
+    { name: "Gumroad", api_key: -> { STRIPE_SECRET }, usd_only: true },
+    { name: "Flexile", api_key: -> { GlobalConfig.get("STRIPE_API_KEY_FLEXILE") }, usd_only: false },
+    { name: "Helper",  api_key: -> { GlobalConfig.get("STRIPE_API_KEY_HELPER") }, usd_only: false },
+    { name: "Iffy",    api_key: -> { GlobalConfig.get("STRIPE_API_KEY_IFFY") }, usd_only: false },
+  ].freeze
+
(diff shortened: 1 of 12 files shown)


gumclaw:
Pre-merge adversarial review (fable-5, local):

**Verdict: no P1/P2 findings.**

Checked:
- Poll loop: `sleep` with an ActiveSupport duration verified working; 10-min deadline per report run with explicit timeout error; failed runs raise (job retries + exhaustion alert cover it).
- Month boundaries: `Time.utc(year, month).next_month` verified to produce the correct UTC month edge, including December rollover via `next_month`.
- FX semantics: `cached_usd_rate` returns currency-units-per-USD (matches `get_usd_cents`'s divide), so `major_units / rate` is the correct USD conversion; missing rate renders blank, not zero.
- CSV injection: all output goes through `CsvSafe`.
- Secrets: API keys are passed per-request and never appear in raised errors or logs.
- Unknown Stripe activity categories fall back to a humanized label instead of raising.
- Backstop wiring: `VerifyFinanceReportsDeliveryJob` spec passes with the new entry.

P3 notes (not blocking):
- The report pulls run serially (up to 12 report runs); on a slow Stripe day the job can hold a Sidekiq thread for a while. Fine for a monthly job; parallelizing isn't worth the complexity now.
- The mailer preview uses hardcoded sample dates in the CSV fixtures — cosmetic.

Local spec run: 32 examples, 0 failures across the new/touched specs. Two pre-existing `email_outstanding_balances_csv` failures are factory-seeding issues present on main.

PR #5890, opened 2026-07-14
Attribute VAT report refunds to the refund date, not the purchase date

gumclaw:
## What

`CreateVatReportJob` now subtracts refunds in the period the **refund** happened, not the period of the original purchase. The unqualified `where(created_at: ...)` on the refund query bound to `purchases.created_at`, so a refund only counted if the original purchase was created inside the report quarter.

The refund query also now mirrors the purchase-side filters:

- `stripe_transaction_id IS NOT NULL` — refunds against purchases that never had a settled charge (never counted on the sales side) are no longer subtracted.
- `not_chargedback_or_chargedback_reversed` — a purchase that was charged back outright never contributes VAT to the report, so its refunds no longer double-relieve the same VAT.

## Why

A refund issued this quarter for a purchase from an earlier quarter was never subtracted from any report generated after that earlier quarter's filing — roughly 22% of refund VAT in a typical quarter (~$6.7k/quarter measured on Q2 2026 production data) was attributed to the wrong period and overpaid. Conversely, re-generating a past quarter's report retroactively pulled in later refunds, changing a period whose return was already filed.

Refund-date attribution is the correct period cut-off treatment for the VAT payable liability and matches the EU OSS mechanism, where corrections to prior periods are reported in the current return rather than by amending the filed one. After this change, re-running a historical quarter is deterministic with respect to refunds.

Out of scope, tracked on the issue: the chargeback legs have the same event-date attribution problem (`not_chargedback` / `chargeback_reversed` read current purchase state against the purchase-date window) and should get the same treatment in a follow-up; historical filed quarters are left as-filed, with the cumulative under-claimed refund VAT recoverable as a correction in the next OSS return.

Refs: https://github.com/antiwork/gumroad-private/issues/1093 (fable-5 GAAP review of the finding and fix direction is on the issue). Builds on the `effective_refunds` narrowing in #5779, which is orthogonal and should land regardless.

## Test plan

- New specs cover: a cross-quarter refund reducing the quarter it was issued in; re-generation determinism (a later refund must not change a closed quarter's report); refunds on charged-back purchases excluded; refunds on unsettled purchases excluded.
- Ran locally: `rspec spec/sidekiq/create_vat_report_job_spec.rb` — 5 examples, 0 failures (the pre-existing happy-case spec passes unchanged, confirming same-quarter behavior is untouched).
- `rubocop` clean on both changed files.

### DIFF
--- a/app/sidekiq/create_vat_report_job.rb
+++ b/app/sidekiq/create_vat_report_job.rb
@@ -54,10 +54,20 @@ class CreateVatReportJob
                                                                .where("flags & :bit = :bit", bit: Purchase.flag_mapping["flags"][:chargeback_reversed])
                                                                .where(created_at: date.beginning_of_day..date.end_of_day)
 
+          # Refunds are subtracted in the period the refund happened, not the period of the
+          # original purchase. A refund is an event of its own period (matching how OSS/MOSS
+          # corrections are reported in the current return), and a purchase from a past quarter
+          # refunded this quarter must still reduce this quarter's VAT due. The purchase-side
+          # filters mirror the two queries above so we only ever subtract VAT that was (or would
+          # have been) reported in the first place: settled purchases that are either not charged
+          # back or had their chargeback reversed. A purchase that was charged back outright never
+          # contributes VAT to the report, so its refunds must not be subtracted either.
           vat_refunds_on_date = zip_tax_rate.purchases
                                               .where("purchase_state != 'failed'")
+                                              .where("stripe_transaction_id IS NOT NULL")
+                                              .not_chargedback_or_chargedback_reversed
                                               .joins(:refunds)
-                                              .where(created_at: date.beginning_of_day..date.end_of_day)
+                                              .where(refunds: { created_at: date.beginning_of_day..date.end_of_day })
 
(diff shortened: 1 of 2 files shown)


gumclaw:
Fable-5 pre-merge review — no P1/P2 findings.

Checked: refund-table qualification of the date window (sums already read `refunds.*`, one row per refund so multi-refund purchases count correctly); chargeback symmetry (charged-back-and-reversed purchases contribute VAT on the sales side and keep their refunds subtractable; outright charged-back purchases are now excluded on both sides — previously a refunded-then-chargedback purchase produced a naked negative subtraction); query performance (`index_refunds_on_created_at` + `index_refunds_on_purchase_id` exist, existing `WithMaxExecutionTime` guard unchanged).

Residual, tracked as follow-up on the issue rather than a defect here: the chargeback legs still read current purchase state against the purchase-date window, so historical re-generation remains non-deterministic with respect to chargebacks until they get the same event-date treatment.

Specs: 5 examples, 0 failures locally, including the pre-existing happy-case spec unchanged.

slavingia:
Cc @steveo-es

PR #5892, opened 2026-07-15
Scope the deferred refunds report's sums to the reported month's refunds

gumclaw:
## What

Scopes the deferred refunds report's refund sums to refunds created inside the reported month, and fixes the range's upper bound: the report now runs to the first instant of the next month instead of `end_of_month`, whose exclusive bound silently dropped the month's final second. Also documents that `total_transaction_count` is purchase-grain while the cents sums are refund-grain.

## Why

Two period bugs in the accounting email that feeds the monthly close JEs:

1. The report selected purchases by "has a refund created in the month" but then summed **all** refunds on those purchases via an unscoped `joins(:refunds)`. A purchase partially refunded in June and again in July had both refunds counted in July's report (and the June refund counted in June's too) — multi-month refunds were double-reported.
2. `DateTime.new(year, month)...end_of_month` excludes 23:59:59 on the last day (`refunds.created_at` has second precision), so a refund created in the month's final second appeared in **no** month's report, ever.

This is a sibling of the refund period-attribution fixes in #5890 (VAT) and #5891 (TaxJar): a refund belongs to the period it was created in.

**Conflict note for the resolver:** open PR #5779 rewrites the same sum lines (`joins(:refunds)` → `joins(:effective_refunds)`) and the selection (`Refund.effective`). Whichever merges second must compose both changes — the month scope here AND the effective scope there are each load-bearing, and each PR's specs fail if its scope is dropped in resolution.

Refs: antiwork/gumroad-private#1093

## Test plan

- New `spec/business/payments/reports/deferred_refunds_reports_spec.rb`: a May purchase refunded partially in June and again in July is reported with only the June refund in June's report and only the July refund in July's (fails with the fix reverted); a refund created at 23:59:59 on the month's last day lands in that month and no other; a same-month purchase+refund stays excluded.
- Ran locally: 3 examples, 0 failures (plus the existing worker/mailer deferred-report specs). Rubocop clean.

## Before/After

Not applicable — accounting report data change, no UI.

### DIFF
--- a/app/business/payments/reports/deferred_refunds_reports.rb
+++ b/app/business/payments/reports/deferred_refunds_reports.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 module DeferredRefundsReports
+  # Monthly accounting email listing the refunds and disputes from the given month that were
+  # issued against purchases from an EARLIER month ("deferred" — the sale's revenue period
+  # already closed when the money went back out).
   def self.deferred_refunds_report(month, year)
     json = { "Purchases" => [] }
     range = DateTime.new(year, month)...DateTime.new(year, month).end_of_month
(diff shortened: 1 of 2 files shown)


gumclaw:
Fable-5 adversarial review ran locally against the branch. Verdict: **pass** — it verified the fix empirically (spec fails with the app change reverted; SQL merge composes cleanly with the PayPal/Stripe pre-filters; same-range selection and sums can't disagree). Findings and how they were handled:

- **P2 — month-end final second dropped from every report (pre-existing).** The `...end_of_month` exclusive bound excluded 23:59:59, so a refund created in the month's last second appeared in no report, ever. Fixed in this PR: the range now ends at the first instant of the next month, with a spec pinning the boundary.
- **P3 — count is purchase-grain, sums are now refund-grain.** Kept as-is (count = purchases affected this month) but documented with a code comment so reconciliation against refund line items isn't a surprise.
- **P3 — guaranteed textual conflict with open #5779** (`effective_refunds` narrowing on the same lines). Flagged in the PR body with the correct composed resolution; each PR's specs guard its own scope.
- **P3 — spec coverage gaps** (PayPal bucket, disputes side, affiliate-credit proration). The core double-count and boundary cases are revert-verified; the rest mirrors unchanged behavior and can harden in a follow-up if wanted.

3 examples, 0 failures locally after the boundary fix. Finance report change — assigned to Sahil for merge, no auto-merge.

PR #5897, opened 2026-07-15
Subtract the month's refunds in the US states sales summary report

gumclaw:
## What

Adds a refund leg to `CreateUsStatesSalesSummaryReportJob`, the monthly per-state summary CSV (GMV / number of orders / sales tax collected). Refunds issued during the reported month — selected via `UsStateSalesTaxUploader.grouped_refund_ids_by_state`, the same selection that feeds the TaxJar refund transactions in #5891 — are subtracted from the month's GMV and tax totals, dated by the refund's own date. Order counts are unchanged: the order still happened in its own month.

Also adds `total_refunded_cents` (the refund's `total_transaction_cents`) to `UsStateSalesTaxUploader#upload_refund`'s return value, so the GMV subtraction uses the same basis as the order leg's `gmv_cents`.

## Why

The summary previously only counted orders created in the reported month, netting refunds against the original sale's month. A refund issued this month for a sale from a prior month never showed up anywhere after that month's filing, and regenerating a closed month after a later refund restated its as-filed numbers. With this change the summary matches what the TaxJar refund transactions report for the same month.

The cutover partition (refunds before 2026-07-20 stay netted in the purchase's month, per `grouped_refund_ids_by_state`) keeps historical months reproducible as filed.

## Depends on

Stacked on #5891 (`fix/taxjar-refund-transactions`) — this PR consumes its `grouped_refund_ids_by_state` and `upload_refund`. Base is set to that branch; it should be squash-merged into #5891 (or rebased onto main after #5891 merges) so both ship in one deploy.

Refs antiwork/gumroad-private#1093

## Test plan

- New specs: a purchase sold in month N and refunded in month N+1 reduces N+1's GMV/tax without changing N+1's order count, and month N regenerates at gross with an unchanged order count — no retroactive restatement. A pre-cutover purchase fully refunded post-cutover stays in its sale month at gross (only its refund's month is reduced), so the same money is never dropped twice.
- Ran the full job spec locally: 12 examples, 0 failures. Daily uploader job spec: 16 examples, 0 failures. Rubocop clean on touched files.
- CI green.


ershad:
Reviewing this alongside #5891, which it's stacked on. I'm also checking some of the assumptions from antiwork/gumroad-private#1093 before signing off, so the two stay consistent.

gumclaw:
Sounds good. To make the cross-check against antiwork/gumroad-private#1093 faster, here's where each of its assumptions lands in the stack:

- **Refund-date attribution:** both PRs date refunds by the refund's own `created_at`, never the purchase's. In #5891 that's the refund transaction's `transaction_date`; here it's the month bucket the refund leg is summed into.
- **Historical months as-filed:** the shared cutover is `REFUND_REPORTING_CUTOVER = 2026-07-20` in #5891's `UsStateSalesTaxUploader`. Refunds created before it stay netted in the purchase's month (old behavior), so regenerating any pre-cutover month reproduces the filed numbers. This PR consumes the same partition via `grouped_refund_ids_by_state`, so the summary and the TaxJar refund transactions can't disagree on which side of the cutover a refund falls.
- **Same selection, same basis:** the summary's refund leg uses the exact refund set #5891 pushes to TaxJar, and the GMV subtraction uses `total_transaction_cents` — the same basis as the order leg's `gmv_cents` — so the monthly summary matches what TaxJar receives for that month.
- **Order counts:** intentionally unchanged by the refund leg (the order still happened in its own month); only GMV and tax are reduced.

One thing worth double-checking while you're in there: the deploy coupling. This PR is meant to squash into #5891 (base is `fix/taxjar-refund-transactions`) so the summary's refund leg and the cutover constant ship in one deploy — if they were split, a summary regenerated between the two deploys could straddle the partition.

gumclaw:
Merged main back in to clear the conflicts — all three were fallout from the TaxJar refund-transactions PR landing first, and none required changing this PR's logic:

- **app/models/refund.rb**: main added the `TERMINAL_FAILURE_STATUSES` constant; kept it as-is.
- **app/business/sales_tax/taxjar/us_state_sales_tax_uploader.rb**: kept this PR's `total_refunded_cents` key in `upload_refund`'s return hash (the summary job needs it to subtract refunds from GMV) on top of main's version of the file.
- **spec/business/sales_tax/taxjar/taxjar_api_spec.rb**: took main's version of the `#create_refund_transaction` spec — it now runs against a real VCR cassette, which supersedes the client-stub approach this branch had.

Targeted specs are green (45 examples: the summary report job, the TaxJar upload job, the TaxJar API spec, and the refund model) and rubocop is clean on all touched files.

gumclaw:
Fixed the pre-cutover edge case Greptile flagged in its latest review pass (commit 8e4f357).

A pre-cutover purchase fully refunded by a post-cutover refund was hit twice: the historical fully-refunded exclusion dropped it from its sale month's summary, while the new refund leg also subtracted the refund from the refund's month. `grouped_purchase_ids_by_state` now keeps a pre-cutover purchase whose refund landed on/after the cutover — its netted amounts count only pre-cutover refunds, so the sale month reports it at gross and the refund month subtracts the refund, partitioning exactly at the cutover instant (same rule as the TaxJar transaction legs in #5891).

Added two specs: the sale month regenerates at gross for such a purchase, and the refund's month nets to zero with the order count unchanged. Ran locally: summary job spec 12 examples / 0 failures, daily uploader job spec 16 examples / 0 failures. Rubocop clean.

PR #5905, opened 2026-07-15
Serve profile avatars at 400px so they're sharp on Retina screens

gumclaw:
## What

Serves profile avatars at a Retina-safe **400x400** instead of 128x128, and versions the cached variant-URL key so **existing uploads sharpen automatically** on next read — no manual backfill, no per-user action.

Fixes antiwork/gumroad-private#1104 (support-ticket driven).

## Why

Every avatar surface goes through `User#avatar_url`, which served a variant capped at `resize_to_limit: [128, 128]`. The profile settings preview box alone is 200 CSS px — 400 device px on a 2x Retina screen — so a 128px source is upscaled ~3x and looks soft/blurry. The seller who reported it uploaded a perfectly sharp image; the blur is entirely our resize.

## How

- `User#avatar_variant` → `resize_to_limit: [400, 400]`.
- **Cache versioning (the promise):** `avatar_url` caches the variant URL in `Rails.cache` under `attachment_#{avatar.id}_variant_url`, with no version in the key — so without this change, existing users would keep getting their cached 128px URL indefinitely. The key is now `..._variant_url_v2`, so every user's URL regenerates lazily on next read and produces the new 400px variant. The seller was told her existing upload will sharpen automatically once this ships; this is what keeps that promise. Old keys are simply orphaned and LRU-evicted by memcached. (Note: production's memcache namespace is keyed on `ENV["REVISION"]`, so the cache would also rotate on deploy — the versioned key makes the intent explicit and is robust if that ever changes.)
- Help text in the profile settings uploader now recommends **400x400px or larger**; the enforced 200px minimum (`MINIMUM_AVATAR_DIMENSION`) is deliberately unchanged so no existing upload flow breaks. A 200px upload served at 400 is mildly upscaled but no worse than today.

## Per-surface caller notes

- `avatar_variant` has no callers besides `avatar_url` (verified by grep) — one variant size change covers everything.
- All avatar surfaces (nav, comments, chat messages, reviews, profile, og:image, favicon) consume `avatar_url`. Small avatars (≤48 CSS px, e.g. comments) will now fetch the 400px file; a 400px JPEG/PNG avatar is tens of KB and CDN-cached, so I did not introduce per-surface sizes — `resized_avatar_url(size:)` already exists if a hot surface ever needs it (`users_controller.rb` uses `size: 240` today).
- Variant processing is on-demand, per-avatar, idempotent in ActiveStorage — first read per avatar after deploy does one resize; no stampede risk beyond the existing shape. S3 growth: one extra small variant per avatar, negligible.

## Test plan

- New specs: `#avatar_variant` asserts the 400x400 transformation; `#avatar_url` asserts the versioned cache key is read/populated and a stale value under the old key is ignored.
- `bundle exec rspec spec/models/user_spec.rb -e "#avatar_url" -e "#avatar_variant" -e "#resized_avatar_url"` → 10 examples, 0 failures. Existing avatar validation/fallback specs also pass.
- Rubocop + Prettier clean on touched files.
- Visual before/after verification is pending a preview deploy (no dev server was trivially available); the settings preview box at 12.5rem on a Retina display is the surface to check.

## Progress

- [x] Bump served variant to 400x400
- [x] Version cached variant-URL key so existing uploads sharpen automatically
- [x] Caller sweep (`avatar_variant` / `avatar_url` / `resized_avatar_url`)
- [x] Update help-text guidance (recommend 400x400+, keep 200 minimum)
- [x] Specs + rubocop
- [ ] Visual Retina verification on preview app

### DIFF
--- a/app/javascript/components/Profile/Settings/LogoInput.tsx
+++ b/app/javascript/components/Profile/Settings/LogoInput.tsx
@@ -28,7 +28,7 @@ export const LogoInput = ({
       </FieldsetTitle>
       <ImageUploader
         id={id}
-        helpText="Your avatar will be visible next to your name in your Gumroad profile and product pages. Your image must be at least 200x200px and must be in JPG or PNG format."
+        helpText="Your avatar will be visible next to your name in your Gumroad profile and product pages. Your image must be at least 200x200px (400x400px or larger is recommended so it stays sharp on high-resolution screens) and must be in JPG or PNG format."
         allowedExtensions={ALLOWED_EXTENSIONS}
         imageUrl={changing || !logoUrl ? null : logoUrl}
         onRemove={() => {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -445,7 +445,12 @@ class User < ApplicationRecord
   def avatar_url
     return ActionController::Base.helpers.image_url("gumroad-default-avatar-5.png") unless avatar.attached?
 
-    cached_variant_url = Rails.cache.fetch("attachment_#{avatar.id}_variant_url") { avatar_variant.url }
+    # The "_v2" suffix versions the cache key: we previously served 128x128
+    # variants, and the old (unversioned) key still holds those URLs. Using a
+    # new key makes every user's avatar URL regenerate lazily on next read,
+    # so existing uploads get the sharper 400x400 variant without any manual
+    # backfill. Old keys are simply left behind and evicted by memcached.
+    cached_variant_url = Rails.cache.fetch("attachment_#{avatar.id}_variant_url_v2") { avatar_variant.url }
     cdn_url_for(cached_variant_url)
   rescue => e
     Rails.logger.warn("User#avatar_url error (#{id}): #{e.class} => #{e.message}")
@@ -455,7 +460,9 @@ class User < ApplicationRecord
   def avatar_variant
     return unless avatar.attached?
 
-    avatar.variant(resize_to_limit: [128, 128]).processed
+    # 400x400 so avatars stay sharp on Retina/high-DPI screens: the profile
+    # settings preview box is 200 CSS px, which is 400 device px at 2x.
+    avatar.variant(resize_to_limit: [400, 400]).processed
   end
 
   def username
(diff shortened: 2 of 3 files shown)


gumclaw:
**Fable-5 pre-merge self-review: `READY-TO-MERGE`** (independent headless reviewer against the branch head, static analysis; 7-point adversarial checklist)

No P1/P2 findings. Three P3 advisories, none blocking:
- **P3-1** Post-deploy cold-cache herd: every avatar is a `_v2` miss on first read → one synchronous resize per avatar. Concurrency-safe (ActiveStorage variant processing is idempotent to a deterministic key); transiently slow only. A one-time pre-warm script is possible if latency matters.
- **P3-2** Favicon (`page_meta/favicon.rb:12`) now serves the 400px image for a 16–32px target — browsers downscale fine; could switch to `resized_avatar_url(size: 128)` if we care about the bytes.
- **P3-3** Nit: new variant spec pins the size twice (mock + transformation assert).

Confirmed safe (verified against code, not the PR body):
- Old cache key has **no other reader/writer** (only similar pattern is `ProductFile` under a different prefix — disjoint).
- **Avatar replacement never needed cache busting**: key embeds the attachment id, which changes on re-attach; purge falls through to the default avatar.
- `avatar_variant` has **no caller besides `avatar_url`**; `resized_avatar_url(size:)` consumers untouched; og:image at 400px now actually meets FB/Twitter minimums.
- `resize_to_limit` never upscales — a 200px-minimum upload serves at native 200px (better than the old forced 128 downscale).
- New specs red/green correctly on revert of either the size or the key; test cache is `:memory_store`, no VCR churn (S3 hosts are VCR-ignored).
- No spec/i18n string pins the old help text.

Only live-verifiable: post-deploy regeneration load, and the Retina before/after visual check (pending preview deploy — tracked as the open box on gumroad-private#1104).

gianfrancopiana:
@gumclaw Start a dev server locally and verify E2E and add screenshots to the PR description.

gumclaw:
🔍 **Live preview for this PR:** https://pr-5905-preview.antiwork.dev

Runs this branch (`f2d6d3164`) locally with seeded data — log in with `seller@example.com` / the standard dev password and check a profile page or Settings → Profile logo upload to see the 400px avatars on a Retina screen.

The URL stays live while the PR is open and is torn down automatically after merge/close. (Local preview infra pilot — details to follow separately.)

PR #5909, opened 2026-07-15
Null UserComplianceInfo PII columns during GDPR erasure

gumclaw:
## What

- `GdprDataErasureService` gains an `anonymize_compliance_info!` step inside the erasure transaction: it nulls `full_name`, `first_name`, `last_name`, `birthday`, `street_address`, `city`, `state`, `zip_code`, `telephone_number`, and `json_data` on **every** `user_compliance_info` row belonging to the user (not just alive rows), and the erasure summary now reports `compliance_info_anonymized: true`.
- New `Onetime::ScrubGdprErasedUserComplianceInfoPii` task backfills the same nulling for users erased before this fix, identified by the `deleted-<id>@deleted.gumroad.com` placeholder email that only the erasure service writes (prefix LIKE, so `index_users_on_email` is usable), batched with `ReplicaLagWatcher`.
- Regression specs covering all the PII columns across both an alive and a previously replaced compliance row, plus specs for the onetime task (scoping to erased users only, batching).

## Why

The erasure service treated soft-deletion as erasure for compliance records: `deactivate_account!` calls `mark_deleted!` on them, while the column-nulling step only touches the `users` table. Soft-deletion just sets `deleted_at`, so after a "successful" GDPR/DPDP erasure the data subject's legal name, date of birth, street address, and phone number remained in the table in plaintext (confirmed on a live erasure; see antiwork/gumroad-private#1105).

The root of the confusion is that `UserComplianceInfo` is `Immutable` — rows are never edited in place, each change dups a new row and soft-deletes the old — so "deleted" here has always meant "hidden," never "destroyed," and a user accumulates multiple historical rows each holding a full PII snapshot. That's also why the fix uses `update_all` (bypassing the Immutable guard is correct for a legally mandated destruction) and sweeps all rows rather than only the alive one.

Scope decisions: `json_data` is nulled because it stores phone, nationality, and the Japanese kana/kanji name and address variants. `country` and the tax ids (encrypted at rest with a public key) are retained with the transaction/tax records under Article 17(3)(b), as are the business-entity columns.

## How tested

- `bundle exec rspec spec/services/gdpr_data_erasure_service_spec.rb spec/services/onetime/scrub_gdpr_erased_user_compliance_info_pii_spec.rb` → 16 examples, 0 failures.
- Reverted the service change and re-ran the new regression spec → 1 failure (fails on revert, as required).
- `bundle exec rubocop -a` on all four touched files → no offenses.

## Progress

- [x] Add UserComplianceInfo PII-column anonymization to GdprDataErasureService
- [x] Backfill task for previously erased users (`Onetime::ScrubGdprErasedUserComplianceInfoPii`)
- [x] Regression spec covering the compliance-row columns (fails on revert)
- [ ] Run the onetime backfill in production after merge
- [ ] Un-patch the support-side erasure skill's manual UCI sweep once this ships

---

AI disclosure: Written by Claude Fable 5 (agent account gumclaw), prompted to own antiwork/gumroad-private#1105 end-to-end (root-cause the plaintext PII surviving GDPR erasure in `user_compliance_info`, fix the service, add a backfill for historical erasures, and open a PR with regression specs).

### DIFF
--- a/app/services/gdpr_data_erasure_service.rb
+++ b/app/services/gdpr_data_erasure_service.rb
@@ -21,6 +21,7 @@ class GdprDataErasureService
     ActiveRecord::Base.transaction do
       @products_deleted = deactivate_account!
       anonymized_email = anonymize_user_pii!
+      anonymize_compliance_info!
       delete_device_records!
       anonymize_carts!(anonymized_email)
       anonymize_credit_cards!(credit_card_ids)
@@ -101,6 +102,35 @@ class GdprDataErasureService
       )
 
       anonymized_email
+    end
+
+    def anonymize_compliance_info!
+      # Compliance (KYC) records hold the person's legal name, date of birth, street
+      # address, and phone number. deactivate_account! only soft-deletes them, which hides
+      # the rows from the app but leaves that data in the table, so erasure must also null
+      # the columns. This intentionally covers every row for the user, not just alive ones:
+      # UserComplianceInfo is Immutable, meaning each edit creates a new row and soft-deletes
+      # the previous one, and those older rows hold the same PII. update_all writes SQL
+      # directly, which also bypasses the Immutable guard — appropriate here because this is
+      # a legally mandated destruction (Article 17), not a normal edit.
+      #
+      # json_data is nulled because it stores PII too: phone numbers, nationality, and the
+      # kana/kanji name and address variants used for Japanese accounts. Country and the
+      # tax ids (encrypted at rest) are retained with the transaction and tax records per
+      # Article 17(3)(b), as are the business-entity columns.
+      @user.user_compliance_infos.update_all(
+        full_name: nil,
+        first_name: nil,
+        last_name: nil,
+        birthday: nil,
+        street_address: nil,
+        city: nil,
+        state: nil,
+        zip_code: nil,
+        telephone_number: nil,
+        json_data: nil,
+        updated_at: Time.current,
+      )
     end
 
     def anonymize_carts!(anonymized_email)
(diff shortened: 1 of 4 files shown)


gumclaw:
Self-review notes:

- **Why `update_all` and not `update_columns` per row:** both bypass the `Immutable` before_update guard (no callbacks); `update_all` does it in one id-scoped statement via the association. Users have at most a handful of compliance rows, so there is no batching concern inside the transaction.
- **Placement in the transaction:** runs right after `anonymize_user_pii!`, before the erasure comment is logged, so a failure rolls back the whole erasure rather than reporting `status: ok` with compliance PII left behind — which is exactly the failure mode that produced the incident.
- **Deliberately retained:** `country` (jurisdiction context for retained transaction/tax records), `individual_tax_id`/`business_tax_id` (strongbox public-key encrypted, part of tax records under Article 17(3)(b)), business-entity columns and `dba`, and `stripe_identity_document_id` (a Stripe file pointer needed for the separately-tracked Stripe-side cleanup, not PII itself). If reviewers want any of these scrubbed too, happy to extend — the sweep in production only nulled the eight individual-PII columns, and this PR matches that scope plus `full_name` (legacy name column) and `json_data`.
- **Backfill idempotency:** re-running the onetime task re-nulls nils; the only side effect of a rerun is touching `updated_at` on already-clean rows. No callbacks fire, so no Stripe jobs get enqueued.
- **Backfill selector:** the `deleted-%@deleted.gumroad.com` prefix LIKE is written against `GdprDataErasureService::ANONYMIZED_EMAIL_DOMAIN`. `GdprBuyerErasureService` uses `buyer-<digest>@` on the same domain but writes to purchases, never `users.email`, so the selector can't over-match; buyers also have no compliance rows.

gumclaw:
@claude review once

gumclaw:
Rerouted review to @ershad — Gianfranco is OOO through Thursday, and this unblocks the gp#1105 GDPR-erasure fix (has a compliance angle, better not to sit two days). Ershad: CI is green, fable review will run before merge as usual.

PR #5910, opened 2026-07-15
Name the flagged email and category in moderation publish errors, keep the audit note, and stop flagging announcement emails as spam

gumclaw:
## What

Fixes three problems behind content moderation blocking email workflows with an unactionable error (antiwork/gumroad-private#1117):

- **The publish error now names the flagged email and the category.** `Installment#content_moderation_check` passes the record's title into the seller message, and `humanize_reasons` learns the `PromptStrategy` preset prefixes (`spam:` → "content that reads as promotional spam", `adult_content:` → "adult content") instead of falling through to the generic "something that may violate our content guidelines".
- **The moderation admin note now survives the failed publish.** It was created synchronously inside the flagged record's save transaction, so the very rollback it documented erased it — blocked publishes left no internal trail. It's now created by `ContentModerationAdminCommentJob` (queue `low`, `lock: :until_executed`); the Sidekiq push escapes the DB transaction. The 5-minute dedup window moves into the job.
- **The spam prompt no longer flags normal announcement emails.** `SPAM_RULES` gets an explicit carveout for short "new video out — Watch HERE" style emails promoting the creator's own releases, and the aggressive-CTA bullet is narrowed to the same CTA repeated back-to-back with no other content.

## Why

A long-time seller could not publish a 17-email workflow: 3 of 17 emails were false-positive flagged as spam (ordinary CTA copy around YouTube links), the error never said which email or why, and the rollback erased the internal audit note — so both the seller and support were debugging blind. Root-cause analysis with prod verification is on the issue: antiwork/gumroad-private#1117.

Naming the flagged email uses data the check already returns (`CheckResult.reasons`), so no new moderation calls are made. The job-based comment fixes the same observability gap noted in gumroad-private#727 for products as well as posts. Corroboration gating and tenure-based exemptions were deliberately left out — they're product decisions, tracked on the issue.

## Before/After

Workflow with two emails; the second ("NEW VIDEO: go watch!") is flagged by the spam preset. Screenshots captured by driving the real workflow editor UI (system spec with the moderation check stubbed to the exact prod reasoning from the ticket).

**Before** — generic error, no way to tell which of the emails was flagged:

before

**After** — error names the email and the category:

after

Mobile before/after:

mobile before
mobile after

The alert component itself is untouched — only the server-provided string changes, so light/dark rendering is unchanged from today.

## How tested

- `spec/services/content_moderation/moderate_record_service_spec.rb` — new specs for the preset labels, titled seller message, job enqueue, and a regression spec proving the note survives a rolled-back transaction (fails on revert).
- `spec/sidekiq/content_moderation_admin_comment_job_spec.rb` — note creation, dedup window, distinct contents, missing user.
- `spec/services/workflow/save_installments_service_spec.rb` — publishing a workflow where one email is flagged returns an error naming that email and does not publish (fails on revert).
- `spec/services/content_moderation/strategies/prompt_strategy_spec.rb` — pins the new SPAM_RULES carveout language, following the existing affiliate-carveout pin.
- Verified the 5 regression specs fail with the app code reverted and pass with it.
- Full runs of the touched spec files pass locally; `rubocop` clean on all touched files.

## Progress

- [x] Root-cause analysis posted on the issue
- [x] Fix + regression specs
- [x] Before/after UI evidence
- [x] Targeted specs + rubocop green locally
- [ ] Review
- [ ] Verify the spam preset change against the three flagged prod installments after deploy

---

**AI disclosure:** Written by Claude Fable 5 (autonomous agent, gumclaw). Prompt: own antiwork/gumroad-private#1117 end-to-end — root-cause the moderation false positive on email workflows, post the analysis on the issue, implement the fix with regression specs, and open a PR.


gumclaw (commented):
Self-review — notes on the non-obvious decisions below.

app/services/content_moderation/moderate_record_service.rb
```diff
@@ -22,25 +22,40 @@ class ContentModeration::ModerateRecordService
...
-  # seller can act on — without leaking scores, thresholds, or the provider.
-  # Generic fallback for blocklist/prompt reasons that aren't a known category.
+  # (score: 0.86, threshold: 0.9)" or "spam: repeated unrelated slogans") into
+  # a friendly, de-duplicated phrase the seller can act on — without leaking
+  # scores, thresholds, or the provider. Generic fallback for blocklist
+  # reasons that aren't a known category.
   def self.humanize_reasons(reasons)
     labels = Array(reasons).map do |r|
+      preset = r.to_s.split(":").first.to_s.strip.downcase
+      next PROMPT_PRESET_LABELS[preset] if PROMPT_PRESET_LABELS.key?(preset)
```
gumclaw, line 41: The preset check runs before the classifier-category parse on purpose: prompt reasons are "spam: <free text>" and the old `split(": ").last` logic would extract a fragment of the free text, never a category. Splitting on the first colon and checking against a fixed allowlist of preset names can't collide with classifier reasons ("OpenAI moderation flagged: ..." prefixes don't match any preset key) or blocklist reasons ("Matched blocked word: ...").

app/services/content_moderation/moderate_record_service.rb
```diff
@@ -127,17 +142,12 @@ def leave_admin_comment(reasons)
...
-        author_name: AUTHOR_NAME,
-        comment_type: Comment::COMMENT_TYPE_NOTE,
-        content: content,
-      )
+      # Created via a background job, not inline: this check runs inside the
+      # blocked record's save transaction, and the failed save's rollback
+      # would erase a synchronously created comment. The Sidekiq push happens
+      # outside the DB transaction, so the note survives. The job also
+      # dedupes identical notes within ADMIN_COMMENT_DEDUP_WINDOW.
+      ContentModerationAdminCommentJob.perform_async(user.id, content)
```
gumclaw, line 150: The dedup window check moved into the job rather than staying here: at enqueue time the comment from a previous attempt may not be committed yet (or may have been rolled back), so an enqueue-time check would be racy. The job's `lock: :until_executed` also collapses identical enqueues from rapid retries while one is still queued.

app/services/content_moderation/strategies/prompt_strategy.rb
```diff
@@ -68,6 +68,12 @@ class ContentModeration::Strategies::PromptStrategy
       compliant unless paired with classic MLM red flags (multi-level downline
       structures, guaranteed returns, "no selling required", recruitment over
       product sales).
+    - Announcement emails promoting the creator's own new release — a new
```
gumclaw, line 71: Trade-off: this carveout makes it easier for an actual spammer to sneak a bare link-blast email through the prompt preset. I think that's the right call — the blocklist and classifier strategies still run on every post, the preset's stated default is already "do not flag unless unmistakable", and the observed failure mode (blocking a legitimate seller's video announcements, prod-verified on 3 installments) is the more costly error. If gating spam flags on corroboration lands later (tracked on the issue), this bullet can be tightened again.

app/models/installment.rb
```diff
@@ -1120,7 +1120,11 @@ def content_moderation_check
       result = ContentModeration::ModerateRecordService.check(self, :post)
       return if result.passed
 
-      errors.add(:base, ContentModeration::ModerateRecordService.seller_message(result.reasons, "post"))
+      # Workflows publish all their emails in one action, so the error must
+      # name the flagged one — otherwise the seller has no way to tell which
+      # of their emails to fix.
+      noun = workflow_id.present? ? "email" : "post"
```
gumclaw, line 1126: "email" vs "post": workflow installments are only ever emails, and the workflow editor labels them "Emails", so the error should use the same word the seller sees in the UI. Standalone posts keep the existing "post" noun.

gumclaw:
@claude review once

PR #5911, opened 2026-07-15
Show one-time wording for fixed-length memberships that only charge once

gumclaw:
## What

Fixed-length memberships whose duration equals a single billing period (e.g. a 12-month membership billed yearly) only ever charge the buyer once, but every buyer-facing surface still used recurring wording. This PR renders one-time wording for that case:

- **Product page / product cards / cross-sells**: price tags show `$99 once` instead of `$99 a year x 1`, and the tier pill shows `$99 once` instead of `$99 a year`.
- **Checkout cart line**: shows `One-time payment` instead of `Yearly`.
- **Receipt email**: the product price line shows `$99 once`, the item note reads "You will not be charged again for this membership. Your membership ends on <date>." (instead of "You will be charged once a year."), and the footer note reads "You won't be charged again for this membership." Free-trial receipts keep the recurring wording because their single charge hasn't happened yet.

The check is centralized: `Subscription#single_charge?` and `BasePrice::Recurrence.single_charge_duration?` in Ruby, mirrored by `isSingleChargeDuration` in `recurringPricing.ts` (following the existing keep-in-sync pattern between the Ruby/TS label tables).

Progress:

- [x] Decide wording (`$X once` price labels, "One-time payment" cart line, receipt notes with end date)
- [x] Checkout/product-page label fix (TS)
- [x] Receipt label fix (Ruby)
- [x] Tests for the single-charge label path
- [ ] Review + merge
- [ ] Verify on production receipt

## Why

Seller report (antiwork/gumroad-private#1111): buyers of a 12-month, yearly-billed membership read "Yearly" / "a year" at checkout and on receipts as an auto-renewing annual charge, generating refund requests and support tickets. The subscription genuinely charges exactly once — `charge_occurrence_count` is `duration_in_months / months_in_recurrence = 1` — so the labels were factually misleading. The receipt even contradicted itself: `ReceiptPresenter::PaymentInfo` already printed ": 1 of 1" and "This is the final payment", while the notes above promised "You will be charged once a year."

The wording-only fix covers the buyer confusion without new pricing plumbing (the seller's alternative ask, a dedicated one-time payment frequency, is a much bigger product change).

Two scope notes:

- `Purchase#format_price_in_cents` historically never appended the "x N" charge count because it called a misspelled method (`charge_occurence_count`) through `try`, which always returned `nil`. This PR passes the count only for single-charge subscriptions (rendering `$99 once`) rather than fixing the typo outright, because suddenly appending "x N" for all fixed-length memberships would change price strings on API v1 and seller-notification surfaces beyond this issue's scope.
- The multi-recurrence dropdown and wishlist labels still show plain recurrence names; those can't occur for the reported configuration (single yearly recurrence) and are left for a follow-up if wanted.

## Before/After

All screenshots use a locally seeded 12-month membership billed yearly ($99, charges exactly once).

| Surface | Before | After |
| --- | --- | --- |
| Product page (desktop, light) | before | after |
| Product page (desktop, dark) | before | after |
| Product page (mobile, light) | before | after |
| Product page (mobile, dark) | before | after |
| Checkout (desktop, light) | before | after |
| Checkout (desktop, dark) | before | after |
| Checkout (mobile, light) | before | after |
| Checkout (mobile, dark) | before | after |
| Receipt email (rendered mailer HTML) | before | after |

## How tested

- New regression specs (all fail when the fix is reverted — verified by stashing the app changes):
  - `spec/presenters/receipt_presenter/item_info_spec.rb` — one-time note with end date; free-trial guard.
  - `spec/presenters/receipt_presenter/footer_info_spec.rb` — one-time footer note; free-trial guard.
  - `spec/helpers/currency_helper_spec.rb` — `recurrence_label` → "once"; `formatted_price_with_recurrence` → "$19.99 once".
  - `spec/models/purchase_spec.rb` — `formatted_total_display_price_per_unit` → "$3 once".
  - `spec/models/subscription_spec.rb` — `#single_charge?`.
  - `app/javascript/utils/recurringPricing.test.ts` — `isSingleChargeDuration` + `formatRecurrenceWithDuration` (vitest).
- Full targeted runs green: `spec/presenters/receipt_presenter/` (232 examples), `spec/mailers/customer_mailer_spec.rb` (91), `spec/mailers/customer_low_priority_mailer_spec.rb` (57), `spec/helpers/currency_helper_spec.rb` (24), plus rubocop, eslint, and `tsc --noEmit` (clean apart from a pre-existing `vite/client` error that also occurs on untouched `main`).
- Manually exercised the product page and checkout against a locally seeded 12-month yearly membership and rendered the receipt mailer for a single-charge purchase before/after.

## QA steps

1. Create a tiered membership with a yearly recurrence and "Automatically end memberships after" set to 12 months.
2. Product page shows `$X once` (price tag + tier pill) and the existing "This membership will automatically end after 12 months" alert.
3. Add to cart: the checkout line item shows "One-time payment" instead of "Yearly".
4. Complete a purchase: the receipt shows `$X once`, "You will not be charged again for this membership. Your membership ends on <date>.", and the footer "You won't be charged again for this membership."
5. A membership with duration > one period (e.g. 24 months yearly) still shows "Yearly" / "a year x 2" everywhere.

---

**AI disclosure**: This PR was authored by Claude Fable 5 (Anthropic), operating as the autonomous agent account `gumclaw`. Prompt: own antiwork/gumroad-private#1111 end-to-end — root-cause the "Yearly" labels on fixed-length single-charge memberships, post the analysis on the issue, implement the fix with regression specs, capture before/after evidence, and open a PR against antiwork/gumroad.


gumclaw:
Self-review notes:

- **`Purchase#format_price_in_cents`** (app/models/purchase.rb): the pre-existing call passed `subscription.try(:charge_occurence_count)` — a misspelling, so the count was always nil and "x N" never rendered from this path. I deliberately did NOT fix the typo to its full effect: passing the real count would suddenly append "x N" to price strings on API v1 (`formatted_display_price` in the v1 purchase JSON), mobile subscription JSON, and seller-notification emails for every multi-charge fixed-length membership. Instead the count is passed only when `single_charge?`, which yields "$99 once" for the case this PR targets and leaves everything else byte-identical. If we want the full "x N" behavior it should be its own PR with API-consumer review.
- **`ConfigurationSelector`**: `duration_in_months` is added as an optional field on the selector's `Product` type because several callers (bundle editor, coffee, subscription manage) build partial product objects. Callers that don't pass it keep today's labels; the buyer product page and the checkout edit modal pass full objects, so the tier pill shows "once" there.
- **Free-trial guard**: both receipt notes keep the recurring wording when `purchase.is_free_trial_purchase?` — at trial-receipt time the single charge hasn't happened, so "you will not be charged again" would be false. The trial's paid receipt (a recurring charge) doesn't hit `manage_membership_note`'s one-time branch incorrectly either: by then `charges_completed?` is true and the payment-info section already prints the "final payment" note.
- **Known leftovers (intentionally out of scope)**: the multi-recurrence dropdown (`recurrenceNames`) and wishlist labels still show plain recurrence names. Neither can occur for the reported seller config (single yearly recurrence), and the dropdown label doubles as a frequency chooser where "One-time payment" next to "Monthly" gets ambiguous when only one option is single-charge.
- The vitest file `recurringPricing.test.ts` is the first test for this util; it pins the Ruby/TS keep-in-sync behavior for the new helper on both sides.

gumclaw:
@claude review once

PR #5913, opened 2026-07-15
Fix ActiveModel::RangeError when cart item price exceeds bigint limit

gumclaw:
## What

Adds a numericality validation on `CartProduct#price` capping it at the signed 8-byte integer maximum (`MAX_PRICE = 9_223_372_036_854_775_807`) and rejecting negative values, mirroring the existing `MAX_QUANTITY` guard on `quantity`.

## Why

Sentry issue [GUMROAD-YR](https://gumroad-to.sentry.io/issues/7613867625/): `CheckoutController#update` crashed with an unhandled `ActiveModel::RangeError` when a request supplied an item price of `10^22`. The `price` column is a bigint, and Rails raises `RangeError` at type-serialization time — after validations pass — so the oversized value bypassed the controller's `ActiveRecord::RecordInvalid` rescue and produced a 500 instead of the friendly "Sorry, something went wrong" redirect. 2 events / 1 user so far; trivially reproducible by any client sending a large price.

With the validation in place, the save fails as `RecordInvalid`, which the controller already rescues and turns into a redirect with an error flash.

## Test plan

- New model specs: price at `MAX_PRICE` is valid; negative price invalid; `10^22` invalid and `save!` raises `RecordInvalid` (not `RangeError`). `spec/models/cart_product_spec.rb` — 17 examples, 0 failures locally.
- New controller spec: `PATCH /checkout` with `price: MAX_PRICE + 1` redirects with the error flash and creates no cart. `spec/controllers/checkout_controller_spec.rb` — 36 examples, 0 failures locally.
- Reproduced the crash first on unpatched code (`ActiveModel::RangeError` raised on save with `price = 10**22`), then confirmed the fix converts it to a validation failure.
- CI green.

### DIFF
--- a/app/models/cart_product.rb
+++ b/app/models/cart_product.rb
@@ -5,6 +5,10 @@ class CartProduct < ApplicationRecord
   include Deletable
 
   MAX_QUANTITY = 2_147_483_647
+  # `price` is a signed 8-byte (bigint) column. Values beyond this raise
+  # ActiveModel::RangeError at serialization time — after validations — so we must
+  # reject them here to turn a hard crash into a normal validation failure.
+  MAX_PRICE = 9_223_372_036_854_775_807
 
   URL_PARAMETERS_JSON_SCHEMA = { type: "object", additionalProperties: { type: "string" } }.freeze
   ACCEPTED_OFFER_DETAILS_JSON_SCHEMA = {
@@ -26,6 +30,7 @@ class CartProduct < ApplicationRecord
 
   validates :price, :quantity, :referrer, presence: true
   validates :quantity, numericality: { only_integer: true, greater_than: 0, less_than_or_equal_to: MAX_QUANTITY }, allow_nil: true
+  validates :price, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: MAX_PRICE }, allow_nil: true
 
   validate :ensure_url_parameters_conform_to_schema
   validate :ensure_accepted_offer_details_conform_to_schema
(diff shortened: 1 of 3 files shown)


gumclaw:
**Pre-merge adversarial review (fable-5) — verdict: SHIP.** No P1/P2 findings. P3 notes for the record:

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | P3 | Legacy rows with negative prices (nothing validated `price` before) would now fail re-validation on cart PATCH; `MergeCartsService` `<<` of an invalid dup silently no-ops via `ActiveRecord::Rollback`. Same failure mode already exists on main for legacy `quantity <= 0` rows. | Not fixing here — pre-existing class of issue; a one-time cleanup of negative-price `cart_products` rows can follow if any exist. |
| 2 | P3 | `call_start_time` in the same action is a separate unhandled-500 vector (`Time.zone.parse("30:00:00")` raises `ArgumentError`, not in the rescue list). | Pre-existing, out of scope — will file a follow-up if it surfaces in Sentry. |
| 3 | P3 | `price` lacks `only_integer` while `quantity` has it, so `10.5` truncates to `10` at type cast. | Pre-existing cast behavior; frontend sends integers. Left as-is to keep the diff minimal. |

Additional local verification beyond the specs in the test plan: `"abc"` → "is not a number" (previously cast silently to 0), `"1e30"` and 30-digit strings → capped by the new validation. Repro of the original crash confirmed on unpatched main (`ActiveModel::RangeError` on save with `price = 10**22`).

PR #5917, opened 2026-07-16
Lock purchases before normal refunds

### LINKED ISSUE #5899: Normal refund path should lock the purchase row first (uniform lock order with failed-refund reversal)
Follow-up from the antiwork/gumroad#5779 review (deadlock-risk assessment): the failed-refund path now serializes all refunded-state recomputation with a uniform purchase → refund → balance lock order. The pre-existing NORMAL refund path (`Purchase#refund_and_save!` and friends) does not take the purchase row lock first, so a normal refund racing a failed-refund reversal (or another concurrent writer of refunded state) can still interleave lock acquisition in a different order.

Scope:
- [ ] Make the normal refund path lock the purchase row first, matching the uniform purchase → refund → balance order used by `Purchase::HandleFailedRefundService` and the `refund.updated` writer
- [ ] Real-thread concurrency spec (separate DB connections) covering a normal refund racing a failed-refund reversal on the same purchase
- [ ] Spec covering refund vs balance-transaction reversal ordering (refund racing a reversal of a sibling refund's balance transactions)
- [ ] Verify no new lock cycle against the payout path (`update_balance!` `with_lock`)

Context: the reviewer accepted the current risk for #5779 because the failed-refund path is locked and tested; the residual risk is entirely in the older normal-refund path. Decision recorded in https://github.com/antiwork/gumroad/pull/5779#pullrequestreview-4700399643.

rmarescu:
Closes #5899.

Follow-up to #5779.

## What

- Lock the purchase row at the start of `Purchase#refund_purchase!`'s database transaction, before creating refund and balance records.
- Read the previous partial-refund state after acquiring the lock so the notification decision uses serialized state.
- Extend the real-thread concurrency coverage to exercise a normal refund, a failed reversal for a sibling refund, and a payout-held balance lock on separate database connections.

## Why

#5779 standardized the failed-refund reversal path on a purchase -> refund -> balance lock order. The older normal-refund path could still update a balance before saving the purchase. If the two paths raced, the normal refund could hold the balance while waiting for the purchase, while the reversal held the purchase while waiting for the same balance.

Taking the purchase lock first makes both refund writers serialize in the same direction. The lock is scoped to `refund_purchase!`'s database transaction, after the Stripe or PayPal request has returned, so no external network call runs while the row lock is held. The payout path only locks the balance and does not subsequently request the purchase lock, so it can finish and release the balance without completing a cycle.

## Before/after

Before: a normal refund and failed-refund reversal could acquire the purchase and balance locks in opposite orders.

After: the normal refund owns the purchase lock before it can wait for a balance; a sibling reversal waits at the purchase, and payout can release the balance independently.

- [x] Walkthrough video showing the two lock paths and the passing concurrency regression: [qa-media/pr-5917-refund-lock-order.mp4](https://github.com/antiwork/gumroad/blob/rmarescu/5899-purchase-lock/qa-media/pr-5917-refund-lock-order.mp4?raw=true)
- Passing-test screenshot: passing concurrency spec

## Test results

- Regression proof: with the application lock reverted, the new concurrency example fails because the sibling reversal overtakes the normal refund while payout holds their shared balance.
- Ran the focused refund and failed-reversal suite:

  ```bash
  bundle exec rspec \
    spec/models/purchase/purchase_refunds_spec.rb:1521 \
    spec/services/purchase/handle_failed_refund_service_spec.rb \
    spec/services/purchase/handle_failed_refund_service_concurrency_spec.rb
  ```

  Result: `59 examples, 0 failures`.

- Repeated the real-thread concurrency file five times. Result: `15 examples, 0 failures`.
- Ran `bundle exec rubocop --force-exclusion -a app/modules/purchase/refundable.rb spec/services/purchase/handle_failed_refund_service_concurrency_spec.rb`. Result: `2 files inspected, no offenses detected`.
- Fixed a spec that the new lock broke: `spec/modules/purchase/reviews_spec.rb:142` asserted product review stats on a stale in-memory instance; `lock!` reloads the purchase and resets its association cache, so the assertions now reload their references first. Full file: `21 examples, 0 failures`.

## Gumclaw testing plan

- [x] Run `bin/test-confidence --full` and reach 100% confidence. Full run on head `066f315c6` (solo run, rebuilt test DB): **100% confidence — 1630 tests passed** (targeted phase 99%/33 tests, full phase 1327 tests in ~11.3h). Remaining warnings were all confirmed pre-existing on merge-base, none attributable to this branch.
- [x] Repeat the timing-sensitive concurrency spec to check for flakes:

  ```bash
  for run in {1..10}; do
    bundle exec rspec spec/services/purchase/handle_failed_refund_service_concurrency_spec.rb || break
  done
  ```

- [x] Confirm all repeated runs finish without a deadlock or lock timeout and preserve the asserted purchase and balance state. Ten consecutive full-file runs: `3 examples, 0 failures` each, no deadlocks or lock timeouts.
- [x] Deploy a preview and smoke-test a normal partial refund for a Gumroad-held Stripe test sale. Confirm the refund succeeds and the purchase displays the expected partially refunded state. Done on the deployed preview (revision `be4140d6b`): a US$10 Stripe test purchase was partially refunded US$4 through the customer drawer; the refund endpoint returned `{"success":true,"partially_refunded":true}` and the purchase shows the Partially refunded badge. Evidence: [screenshot](https://github.com/antiwork/gumroad/blob/rmarescu/5899-purchase-lock/qa-media/pr-5917-preview-partial-refund.png?raw=true) / [video](https://github.com/antiwork/gumroad/blob/rmarescu/5899-purchase-lock/qa-media/pr-5917-preview-partial-refund.mp4?raw=true)
- [x] Attach a passing-test screenshot and the required short walkthrough video. The automated regression is the authoritative race reproduction because manual timing is unreliable.

## Risk and rollback

The change serializes concurrent writers for the same purchase for the duration of the existing refund database transaction. Different purchases remain independent, and processor requests happen before the lock is acquired. There are no schema or data migrations. Reverting this commit restores the previous lock behavior without requiring data cleanup.

---

AI disclosure: GPT-5.6 Sol Max via Codex implemented and tested this change.


rmarescu:
@gumclaw take ownership of this one, finish up remaining work.

gumclaw:
## Fable-5 adversarial review — verdict: READY-TO-MERGE (P2s fixed on branch)

Independent adversarial review run locally against this branch (static analysis + local spec runs). Full hunt list covered lock-reload side effects, pre-lock/post-lock state divergence, association-cache resets, global lock-order completeness across all refund/chargeback/payout writers, giftee/bundle mirror ordering, and spec pinning.

### Findings and resolutions

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | P2 | Combined-charge refunds (`Charge#refund_and_save!`) wrap all purchases in one transaction, so lock acquisition interleaves purchase₁ → balance → purchase₂ — a reversal on purchase₂ can still cycle. Pre-existing (not a regression from this PR), but the new comment overclaimed. | Comment narrowed to single-purchase refunds in 6ec5d9597; combined-charge lock ordering to be tracked as a follow-up. |
| 2 | P2 | Bare `lock!` raises on dirty records, and reading any json_data-backed attribute on a row with NULL `json_data` dirties the record in memory. No current caller trips it (all verified), but if one ever did it would fire *after* the processor refund, leaving money moved at Stripe with no local records. `HandleFailedRefundService` uses `reload.lock!` for exactly this reason. | Switched to `reload.lock!` with an explanatory comment in 6ec5d9597. |
| 3 | P3 | `refund_partial_purchase!` keeps the old unordered write sequence — but grep shows no production callers (console/legacy only). | Left as-is; candidate for the same follow-up. |
| 4 | P3 | The spec's purchase-lock latch fired on any `Purchase#lock!` in the reversal thread. | Guarded on `@purchase.id` in 6ec5d9597. |

### Adversarially confirmed safe

- Every production entry point to `refund_purchase!` passes a clean instance (webhook handlers, admin/API controllers, workers, `Charge#refund_and_save!`) — no pre-call in-memory mutations.
- Pre-lock values (`funds_refunded`, presentment snapshot) describe the processor's actual refund, not purchase state; every purchase-derived input to the refund records is read post-reload. A concurrent sibling refund now fails closed instead of double-recording.
- Payout locks balances only (leaf — no cycle). Dispute paths take no purchase-then-balance sequence inside a transaction. Giftee/bundle mirror rows are only reached from the money purchase while holding its lock, same direction in both writers.
- The new concurrency example genuinely pins the fix: with `lock!` reverted it fails (reversal overtakes) — verified locally.

### Test evidence at head

- `handle_failed_refund_service_concurrency_spec.rb` + `reviews_spec.rb`: 24 examples, 0 failures after the review fixes.
- 10 consecutive isolated runs of the concurrency file: 0 failures, no deadlocks or lock timeouts.
- Focused refund suite (purchase_refunds + failed_refund_service specs): 184 examples, 0 failures.
- CI triage on the earlier run: `reviews_spec.rb:142` was a real breakage from the new lock (association cache reset) — fixed in 39970ad92; the PPP tooltip failure on Test Slow 24 reproduces with this PR's production change reverted, so it is not caused by this PR.

### Only verifiable live

- Whether real production purchases have NULL `json_data` reachable through a refund path (now moot given `reload.lock!`).
- InnoDB deadlock-victim behavior for the combined-charge scenario under real concurrency (follow-up scope).

gumclaw:
Took ownership. Status:

- Fixed the CI failure this branch caused: `reviews_spec.rb:142` broke because the new lock reloads the purchase and resets its association cache — the spec now reloads its references (39970ad92). The PPP tooltip failure on Test Slow 24 reproduces with this PR's production change reverted, so it's not from this branch.
- Ran the adversarial pre-merge review (verdict above): READY-TO-MERGE after two P2 fixes, both landed in 6ec5d9597 (`reload.lock!` to defuse the json_data dirty-record trap, narrowed lock-order comment). Combined-charge lock ordering is pre-existing and tracked in #5918.
- Concurrency spec repeated 10× in isolation: 0 failures, no deadlocks or lock timeouts. Focused refund suite: 184 examples, 0 failures.
- Walkthrough video and passing-test screenshot are in `qa-media/` and linked from the body.
- `preview` label added for the smoke test of a partial refund once the preview app is up.

Remaining before handoff: CI green on 6ec5d9597 and the preview smoke test.

Marqueseking (commented):
Nice

PR #5919, opened 2026-07-16
Add edit_profile scope to the Gumroad CLI OAuth application

gumclaw:
## What

Adds the `edit_profile` scope to the Gumroad CLI's OAuth application (prod app id 39434), so tokens minted by `gumroad auth login` can pass the `edit_profile` gate on the Pages write endpoints.

Fixes the gumroad half of antiwork/gumroad-cli#185: `Api::V2::PagesController` deliberately requires `edit_profile` on the token itself (`require_oauth_scope!`, no `account` fallback), but the CLI app only minted `account`-scoped tokens — so `gumroad pages push <slug>` failed for every CLI-authenticated user with `Access denied: This endpoint requires the edit_profile scope.`

Design call (documented on the issue): keep the narrow gate, mint the scope. The gate is a deliberate security boundary (same pattern as `MediaController`); the CLI is first-party, so it should carry the scope rather than the endpoint being widened for every legacy broad-scope token.

- [x] Migration appends `edit_profile` to the CLI OAuth app's scopes (idempotent, reversible, same not-found guard shape as the device-authorization migration)
- [x] Dev seed updated so local CLI apps match
- [x] Migration spec covering append, idempotency, other-apps-untouched, rollback

Existing tokens are unchanged — they still lack the scope until the user re-runs `gumroad auth login`. The paired gumroad-cli change requests the scope at login and rewrites the 403 into a clear "re-authenticate" hint.

Verified against prod (audited console read `20260716T113313Z-195b8b6c`): app 39434 currently has `edit_products view_sales mark_sales_as_shipped edit_sales view_payouts view_profile account` — `view_profile` present, `edit_profile` missing.

## Test results

```
DISABLE_SPRING=1 bundle exec rspec spec/migrations/add_edit_profile_scope_to_gumroad_cli_oauth_application_spec.rb
4 examples, 0 failures
```
rubocop clean on all touched files.

---

🤖 Written with Hermes Agent (claude-fable-5). Prompt: "cli#185 — pages push dead for CLI tokens (edit_profile scope) — work on this."

### DIFF
--- a/app/controllers/oauth/authorizations_controller.rb
+++ b/app/controllers/oauth/authorizations_controller.rb
@@ -11,6 +11,20 @@ class Oauth::AuthorizationsController < Doorkeeper::AuthorizationsController
   helper_method :admin_scope_optional?, :show_admin_authorization_checkbox?
 
   private
+    # Scope migrations lock the application while revoking credentials. Re-read
+    # and validate the requested scopes under that same lock so a request that
+    # saw the old scopes cannot create a grant after the revocation sweep.
+    def authorize_response
+      oauth_application = OauthApplication.alive.find_by(uid: params[:client_id])
+      return super if oauth_application.nil?
+
+      oauth_application.with_lock do
+        @pre_auth = nil
+        @strategy = nil
+        super
+      end
+    end
+
     def redirect_or_render(auth)
       if admin_authorization_redirect?(auth)
         admin_authorization_code = AdminApiAuthorizationCode.create_for!(
--- a/app/controllers/oauth/device_codes_controller.rb
+++ b/app/controllers/oauth/device_codes_controller.rb
@@ -8,24 +8,34 @@ class Oauth::DeviceCodesController < ApplicationController
   def create
     oauth_application, error, error_description = authenticate_oauth_application
     return render_oauth_json_error(error, error_description, status: error == :invalid_client ? :unauthorized : :bad_request) if error
-    return render_oauth_json_error(:unauthorized_client, "Client is not allowed to use device authorization") unless oauth_application.device_authorization_enabled?
 
-    scopes = requested_oauth_scope_for(oauth_application)
-    return render_oauth_json_error(:invalid_scope, "The requested scope is invalid") unless valid_oauth_scope?(oauth_application, scopes)
+    result = oauth_application.with_lock do
+      if !oauth_application.device_authorization_enabled?
+        { error: :unauthorized_client, error_description: "Client is not allowed to use device authorization" }
+      else
+        scopes = requested_oauth_scope_for(oauth_application)
+        if valid_oauth_scope?(oauth_application, scopes)
+          device_authorization, device_code, user_code = OauthDeviceAuthorization.create_for!(
+            oauth_application:,
+            scopes:,
+            ip_address: request.remote_ip,
+            user_agent: oauth_request_user_agent
+          )
+          { device_authorization:, device_code:, user_code: }
+        else
+          { error: :invalid_scope, error_description: "The requested scope is invalid" }
+        end
+      end
+    end
 
-    _device_authorization, device_code, user_code = OauthDeviceAuthorization.create_for!(
-      oauth_application:,
-      scopes:,
-      ip_address: request.remote_ip,
-      user_agent: oauth_request_user_agent
-    )
+    return render_oauth_json_error(result[:error], result[:error_description]) if result[:error]
 
     headers.merge!("Cache-Control" => "no-store, no-cache")
     render json: {
-      device_code:,
-      user_code:,
+      device_code: result[:device_code],
+      user_code: result[:user_code],
       verification_uri: oauth_device_authorization_url(host: DOMAIN, protocol: PROTOCOL),
-      verification_uri_complete: oauth_device_authorization_url(host: DOMAIN, protocol: PROTOCOL, user_code:),
+      verification_uri_complete: oauth_device_authorization_url(host: DOMAIN, protocol: PROTOCOL, user_code: result[:user_code]),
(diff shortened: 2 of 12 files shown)


gumclaw:
## Fable-5 pre-merge review — READY-TO-MERGE

Independent adversarial review at `ee3ccf4`. No P1/P2 findings; two P3s, one fixed.

**P3s**
- Redundant `updated_at: Time.current` in `update!` (auto-touched, unlike the sibling's `update_all`) → fixed in 69be73b.
- Scope-order cosmetic mismatch between migration (appends at end) and dev seed (before `account`) — confirmed no consumer is order-sensitive (`scopes.include?`, `split`, Doorkeeper `OAuth::Scopes` is set-like), left as-is.

**Adversarial claims confirmed safe**
- Migration is idempotent; NULL/empty `scopes` handled by `to_s.split`; anonymous-AR pattern byte-identical to the device-authorization sibling migration.
- Existing tokens unaffected: Doorkeeper 5.7.1 checks **token** scopes at request time (`doorkeeper_authorize!` / `require_oauth_scope!`); `app.scopes` is consulted only at authorization time. `edit_profile` is already in `optional_scopes`/`public_scopes`, so no server scope registration change.
- Rollback doesn't strand tokens (use-time validation ignores app scopes); repo precedent scrubs token scopes with a separate corrective migration when that matters.
- schema.rb version bump for a data-only migration matches repo convention (PR #5521 precedent).
- Spec conventions mirror `revoke_backfilled_edit_emails_scope_..._spec.rb`; factory supports uid override.

**Only live-verifiable:** the actual prod row update at deploy time — pre-verified via audited console read `20260716T113313Z-195b8b6c` that app 39434 exists with the expected current scopes, so the migration's happy path is the one that will run.

gumclaw:
Addressed Greptile's rollback edge case (98fa48f):

- **Rollback removing a pre-existing `edit_profile` after a no-op `up`** — intentional, now documented on the migration: `down` restores the declared pre-migration state ("CLI app does not carry edit_profile") rather than tracking whether `up` made a change. Production verifiably lacks the scope today, so the two are equivalent in practice.
- **Missing spec for rollback after an idempotent no-op `up`** — added.

Local run: `bundle exec rspec spec/migrations/add_edit_profile_scope_to_gumroad_cli_oauth_application_spec.rb` → 5 examples, 0 failures. rubocop clean on both files.

nyomanjyotisa:
Critical Rollback leaves issued edit_profile credentials usable indefinitely

**File:** `db/migrate/20261206000001_add_edit_profile_scope_to_gumroad_cli_oauth_application.rb:40`
**Confidence:** 97/100

`down` only removes `edit_profile` from `oauth_applications.scopes`. Endpoint authorization checks the access token's stored scopes, not the application's current scopes, and Gumroad configures access tokens with no expiration. Any `account edit_profile` token minted while this migration is deployed therefore continues to pass the Pages write gate after rollback. A pending or approved device authorization also retains its stored scopes and can mint such a token after rollback because polling does not revalidate them against the application. This contradicts the comment that `down` restores the pre-migration state and makes rollback ineffective if the scope expansion needs to be withdrawn. Strip or revoke the scope for this application across `oauth_access_tokens`, `oauth_access_grants`, and `oauth_device_authorizations` during `down`; `RevokeBackfilledEditEmailsScopeFromDefaultOauthApplications` provides an existing repository pattern for handling all three.

gumclaw:
Pushed d5bc49d: rollback now also revokes edit_profile from issued credentials (access tokens, grants, device authorizations) for the CLI application, not just the application's scopes list — tokens never expire, so an application-only rollback left minted tokens passing the Pages write gate indefinitely. Same three-table pattern as RevokeBackfilledEditEmailsScopeFromDefaultOauthApplications. New spec examples cover credential revocation, deletion of a device authorization whose only scope was edit_profile, and other-application isolation. 7 examples, 0 failures locally; RuboCop clean.

(Flagged by nyomanjyotisa on #5921 — the finding belongs to this PR's migration.)

nyomanjyotisa:
**Critical: rollback can miss an `edit_profile` token minted concurrently**

**File:** `db/migrate/20261206000001_add_edit_profile_scope_to_gumroad_cli_oauth_application.rb:52`  
**Confidence:** 96/100

`down` sweeps `oauth_access_tokens` before cleaning access grants and device authorizations. An approved device authorization can poll after the token sweep but before its stored `edit_profile` scope is removed. `OauthDeviceAuthorization#poll!` then creates a token from those stored scopes without revalidating them against the application's reduced scopes, so the newly inserted, non-expiring token survives rollback. An authorization-code exchange that wins the grant lock after the token sweep has the same ordering risk.

Please synchronize rollback with the application lock used by device polling, clean issuance sources before access tokens, and sweep tokens last. Add a regression spec that mints from an approved authorization between cleanup phases and verifies that no `edit_profile` token survives.

gumclaw:
Pushed 1884bba addressing the concurrent-mint gap:

- **Rollback now runs in a transaction holding the `oauth_applications` row lock** (`oauth_applications.lock.find_by`) — the same lock `OauthDeviceAuthorization#poll!` and the grant exchange take before minting, so no device-code token can be issued while the sweep runs.
- **Cleanup order reversed:** issuance sources first (device authorizations, then access grants), **access tokens swept last** — a token minted just before the lock was acquired is still caught by the final sweep.
- **Locking reads on each sweep batch** (`.lock.find_each`) so rows committed by a concurrent request right before the migration took the app lock are visible despite REPEATABLE READ.
- **Regression spec added:** wraps the device-authorization cleanup phase to mint a token via `poll!` from an approved authorization mid-rollback, then asserts the minted token loses `edit_profile` and no `edit_profile` token survives for the app.

```
8 examples, 0 failures
```
rubocop clean on both files.

nyomanjyotisa (approved):

PR #1122, opened 2025-09-04
Allow Gumroad tests to run without external secrets

HiAbhishekh:
### Solution:

This PR introduces a safe, isolated test environment that allows the RSpec suite to run without requiring production secrets, API keys, or external services.

The solution ensures the full Rails environment loads while mocking only problematic dependencies. This makes the test suite fast, reliable, and runnable locally without complex setup.

###  Before
Running tests required:

- MySQL database running on localhost:3306
- MongoDB instance running on localhost:27017
- Elasticsearch instance running
- Production API keys for Stripe, PayPal, AWS, SendGrid, etc.
- Complex manual environment setup
- Errors when running without secrets:

$ bundle exec rspec

- Mysql2::Error::ConnectionError: Can't connect to MySQL server on '127.0.0.1:3306'
- Mongo::Error::SocketError: Errno::ECONNREFUSED: Connection refused
- Faraday::ConnectionFailed: Couldn't connect to server
- NoMethodError: super: no superclass method 'sidekiq_options'

###  After

Tests now run successfully without any external dependencies:

- $ RAILS_ENV=test DATABASE_URL=sqlite3::memory: bundle exec rspec
- Test environment configured to run without secrets
- All external API keys are mocked with test values
- Tests run without external dependencies
- No MongoDB connection errors
- No Elasticsearch connection errors
- No MySQL connection errors
- No Sidekiq worker loading errors

Key Changes
1. Test Environment Setup (lib/test_environment_setup.rb)
- Added mocks for problematic gems (mysql2, alterity, sidekiq-pro)
- Introduced mock GlobalConfig with test API credentials
- Mocked Mongoid to prevent unwanted DB connections

2. Database Configuration
- Added config/database.test.yml using in-memory SQLite
- Updated spec/spec_helper.rb to switch to SQLite in test runs
- Wrapped MySQL-specific commands so they only run if MySQL adapter is active

3. Rails Environment Adjustments
- Skipped MongoDB setup in test (config/application.rb)
- Made config/initializers/mongo.rb conditional
- Updated spec/support/elasticsearch.rb to gracefully skip if unavailable

4. Gemfile Updates
- Removed conditional exclusions of mysql2 and alterity
- Allows Rails to load these gems even when mocked in tests

5. Sidekiq Mocking
- Added aggressive stubs in spec_helper.rb
- Prevented common errors (sidekiq_options, sidekiq_retries_exhausted)
- Ensured worker classes load safely

Screenshot 2025-09-05 at 1 00 36 AM

[...]

PR #5920, opened 2026-07-16
Fix missing og:image on custom-HTML product pages for cover-only products

gumclaw:
Fixes antiwork/gumroad-private#1122

## Problem

Custom-HTML product pages drop `og:image` (and emit no Twitter card tags at all) for any product that has a **cover but no thumbnail** — the common case, since thumbnails are optional. The wrapper document (`LinksController#custom_html_wrapper_document`) built its own og:image tag from `product.thumbnail&.alive&.url` only, while the standard product page (`PageMeta::Product#set_open_graph_image_meta`) falls back to the cover (`preview_url` when `preview_image_path?`) and then the oembed thumbnail. So the moment a seller published a custom page, their Facebook/iMessage/X link previews lost the image (reporter: `gprsjunkie.gumroad.com/l/dont-get-suspended`, FB Sharing Debugger warned "'og:image' should be explicitly provided").

## Fix

- Extract the standard page's fallback chain into `Link#social_share_image` (cover image → escaped oembed thumbnail → generated video poster), shared by both surfaces so they can't drift again. (Named without a `_url` suffix per CONTRIBUTING.md's rule against new `_url`/`_path` methods.)
- Directly uploaded video covers now use the cached ffmpeg poster from #5860 (`AssetPreview#video_poster_url`) as the final fallback — previously both surfaces omitted the image for video covers with no thumbnail. Nil until the background worker has generated the poster, in which case the tags are simply omitted.
- `PageMeta::Product#set_open_graph_image_meta` now delegates to it (image/oembed behavior unchanged; video-poster fallback is new).
- The wrapper resolves its share image as thumbnail → `social_share_image` (thumbnail still wins there, as before), and now also emits `twitter:card` / `twitter:image`.

## Checklist (from the issue)

- [x] Mirror `PageMeta::Product`'s fallback chain in `custom_html_wrapper_document`; extract shared logic
- [x] Add `twitter:image` (plus `twitter:card`) on the wrapper
- [x] Spec: cover-only product with custom_html renders `og:image` with the preview URL; thumbnail still wins when present
- [x] Uploaded-video covers fall back to the generated poster (model + wrapper specs)
- [ ] Verify on the reporter's live page post-deploy + re-run FB Sharing Debugger scrape

## Tests

- `spec/controllers/links_controller_custom_html_spec.rb`: 5 new wrapper examples (no image → tags omitted; thumbnail; cover-only fallback; video-poster fallback; thumbnail beats cover). Full file: 37 examples, 0 failures.
- `spec/modules/product/preview_spec.rb`: 5 new `#social_share_image` examples (nil, image cover, escaped oembed thumbnail, video poster fallback, video cover with no poster yet → nil). Full file: 18 examples, 0 failures.
- `links_controller_spec.rb -e "product information markup"` (standard-page og:image): 7 examples, 0 failures.
- Rubocop clean on all touched files.

### DIFF
--- a/app/controllers/concerns/page_meta/product.rb
+++ b/app/controllers/concerns/page_meta/product.rb
@@ -41,10 +41,10 @@ module PageMeta::Product
     end
 
     def set_open_graph_image_meta(product)
-      # Cover image (or the thumbnail of a video/oembed cover) — shared with the
-      # custom-HTML wrapper document via Link#social_share_preview_url so both
-      # surfaces resolve the share image the same way.
-      image_url = product.social_share_preview_url
+      # Cover image (or the thumbnail/poster of a video/oembed cover) — shared
+      # with the custom-HTML wrapper document via Link#social_share_image so
+      # both surfaces resolve the share image the same way.
+      image_url = product.social_share_image
       return if image_url.blank?
 
       set_meta_tag(property: "og:image", content: image_url)
--- a/app/controllers/concerns/page_meta/product.rb
+++ b/app/controllers/concerns/page_meta/product.rb
@@ -41,16 +41,14 @@ module PageMeta::Product
     end
 
     def set_open_graph_image_meta(product)
-      if product.preview_image_path?
-        set_meta_tag(property: "og:image", content: product.preview_url)
-        set_meta_tag(property: "og:image:alt", content: "")
-      elsif product.preview_oembed_thumbnail_url
-        set_meta_tag(
-          property: "og:image",
-          content: Addressable::URI.escape(product.preview_oembed_thumbnail_url).html_safe,
-        )
-        set_meta_tag(property: "og:image:alt", content: "")
-      end
+      # Cover image (or the thumbnail of a video/oembed cover) — shared with the
+      # custom-HTML wrapper document via Link#social_share_preview_url so both
+      # surfaces resolve the share image the same way.
+      image_url = product.social_share_preview_url
+      return if image_url.blank?
+
+      set_meta_tag(property: "og:image", content: image_url)
+      set_meta_tag(property: "og:image:alt", content: "")
     end
 
     # Equivalent to `twitter_product_card(product, product_description:).html_safe`
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1005,8 +1005,19 @@ class LinksController < ApplicationController
       checkout_url_js = ERB::Util.json_escape("/l/#{product.unique_permalink}?#{Rack::Utils.build_query(checkout_params)}".to_json)
       title = ERB::Util.h(product.name.to_s)
       canonical = ERB::Util.h(product.long_url.to_s)
-      og_image = product.thumbnail&.alive&.url
-      og_image_tag = og_image ? %(<meta property="og:image" content="#{ERB::Util.h(og_image)}">) : ""
+      # Thumbnail first (the wrapper's original image source, kept as the
+      # winner), then the standard product page's fallback chain — cover image,
+      # then the thumbnail of a video/oembed cover — via
+      # Link#social_share_preview_url. Without the fallbacks, products with a
+      # cover but no thumbnail (the common case) lost their link-preview image
+      # the moment they went custom (gumroad-private#1122).
+      share_image = product.thumbnail&.alive&.url || product.social_share_preview_url
+      share_image_tags = if share_image
+        escaped_share_image = ERB::Util.h(share_image)
+        %(<meta property="og:image" content="#{escaped_share_image}">\n    <meta property="twitter:card" content="summary_large_image">\n    <meta property="twitter:image" content="#{escaped_share_image}">)
+      else
+        ""
+      end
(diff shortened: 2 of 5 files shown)


gumclaw:
**Fable-5 adversarial pre-merge review: READY-TO-MERGE** (no P1/P2 findings)

Verified empirically on a fresh checkout:
- **Red-green confirmed**: the new "falls back to the cover image" wrapper spec fails on main's app code (1 failure) and passes on this branch — the fix is real, not incidental. All 8 new specs green.
- **Escaping**: dropping `.html_safe` on the oembed branch is a strict improvement — `set_meta_tag` → Rails `tag` helper now correctly renders `&` as `&amp;` in `content="…"` (old code emitted invalid raw HTML; scrapers decode entities fine). `ERB::Util.h` doesn't touch `%XX` sequences, so URI-escape + HTML-escape compose losslessly on the wrapper.
- **Behavior parity**: standard page renders identical og:image for all four cover cases (image / oembed / video-no-oembed / none); old truthy check ≡ new `blank?` guard because `oembed_thumbnail_url` never returns `""`. Standard-page meta specs: 7 examples, 0 failures.
- **Nil-safety**: `preview_image_path?` true guarantees non-blank `preview_url` (attached image or unsplash_url); `Addressable::URI.escape` never sees nil.
- **No missed surfaces**: `set_open_graph_image_meta` has exactly one caller; the only other og:image-from-thumbnail site is the profile wrapper (`users_controller.rb:278`, avatar-based, out of scope for #1122). No twitter-tag duplication in the wrapper document.

P3 notes (non-blocking, not fixing here):
- `if share_image` could be `.present?` for defense-in-depth — the `""` case is unreachable today.
- Pre-existing: `set_twitter_meta` on the standard page uses the raw unescaped oembed thumbnail for `twitter:image`; a follow-up could route it through `social_share_preview_url` for consistency.

gumclaw:
**Preview-app QA evidence** (live boot of this branch, commit `0dd227a81`, preview `shvol-test`)

Seeded the exact reporter scenario — custom-HTML page, image cover, **no thumbnail**:

```
{permalink: "demo", custom_html: true, thumbnail: false, preview_image_path: true,
 social: "https://images.unsplash.com/photo-1587502536575-6dfba0a6e017"}
```

Live wrapper `<head>` on this branch (`curl https://<preview>/l/demo`):

```html
<meta property="og:title" content="Beautiful widget">
<meta property="og:type" content="product">
<meta property="og:url" content="https://seller-…/l/demo">
<meta property="og:image" content="https://images.unsplash.com/photo-1587502536575-6dfba0a6e017">
<meta property="twitter:card" content="summary_large_image">
<meta property="twitter:image" content="https://images.unsplash.com/photo-1587502536575-6dfba0a6e017">
```

Before (same scenario on main, per the issue's live verification 2026-07-16): `og:title`/`og:type`/`og:url` present, **no `og:image`**, no twitter tags. The red-green spec run in the review comment above confirms the same delta in CI terms.

Note: no QA video for this one — the change is scraper-facing `<meta>` tags only, nothing visible on the rendered page. The curl'd head above is the complete observable surface of the fix.

(The reporter has since applied the interim workaround on prod — uploaded a thumbnail — so their live page now shows an og:image via the old thumbnail-only path; the post-deploy verification checkbox should re-check with the FB Sharing Debugger anyway.)

nyomanjyotisa:
The reported image-cover case is fixed correctly, but I found one functional gap:

1. Important — uploaded-video covers still have no social image
Confidence: 98/100
Directly uploaded videos can have a generated poster via main_preview.thumbnail_url from #5860. The new helper ignores it and returns nil, so custom-HTML products with a video cover and no product thumbnail still omit og:image and twitter:image. The new spec explicitly locks in this omission.
Add the cached video poster before the oEmbed fallback and cover it with model and wrapper specs. Otherwise, narrow the PR’s claim that it handles video/all cover-only products.

2. Important — prohibited _url method name
Confidence: 100/100
social_share_preview_url violates the explicit CONTRIBUTING.md rule against creating methods ending in _url or _path. Rename it, for example to social_share_image.

gumclaw:
Both fixed in 042e5a8:

1. **Video poster fallback** — `social_share_image` now falls through image cover → escaped oEmbed thumbnail → `main_preview.video_poster_url` (the cached ffmpeg poster from #5860). Uploaded-video covers with no thumbnail now get `og:image`/`twitter:image` on both the standard page and the custom-HTML wrapper. Nil (tags omitted) only while the background poster generation hasn't run yet. Covered by new model specs (poster present, and poster-not-yet-generated → nil) and a wrapper spec; the spec that previously locked in the omission was replaced.

2. **Naming** — renamed `social_share_preview_url` → `social_share_image` per the CONTRIBUTING.md rule, with a comment noting why the `_url` suffix is avoided.

PR body updated to match. Local runs: `preview_spec.rb` 18 examples 0 failures, `links_controller_custom_html_spec.rb` 37 examples 0 failures, standard-page og:image markup examples 7/7, rubocop clean.

nyomanjyotisa (approved):

PR #5921, opened 2026-07-16
Fix NoMethodError on 'transfer' in StripeChargeProcessor#get_charge_object for failed destination charges

gumclaw:
## What

Guards the destination-transfer lookup in `StripeChargeProcessor#get_charge_object` so it only retrieves the destination transfer when the charge actually carries a `transfer` id. Previously the method assumed any charge with `transfer_data` also had a `transfer` attribute.

## Why

A destination charge only gets a `transfer` once Stripe has created it — i.e. after the charge succeeds. Failed or still-pending destination charges have `transfer_data` but no `transfer` attribute at all, and `Stripe::StripeObject` raises `NoMethodError: undefined method 'transfer'` on access.

This fires in production when `Purchase::SyncStatusWithChargeProcessorService` re-syncs a failed purchase whose charge was created with `transfer_data` (via `ChargeProcessor.get_or_search_charge` → `get_charge_object`).

Sentry: [GUMROAD-YV](https://gumroad-to.sentry.io/issues/7615064249/) — `NoMethodError: undefined method 'transfer' for #<Stripe::Charge>` in `StripeChargeProcessor#get_charge_object` (2 events so far, new issue as of today).

## Test plan

- New spec: charge with `transfer_data` but no `transfer` → `get_charge` returns a `BaseProcessorCharge` without calling `Stripe::Transfer.retrieve`. Confirmed red on `main` (raises `NoMethodError`), green with the fix.
- Full `#get_charge` describe block: 9 examples, 0 failures locally.

## Before/After

Not applicable — backend-only change, no UI impact.

### DIFF
--- a/app/business/payments/charging/implementations/stripe/stripe_charge_processor.rb
+++ b/app/business/payments/charging/implementations/stripe/stripe_charge_processor.rb
@@ -151,7 +151,11 @@ class StripeChargeProcessor
   end
 
   def get_charge_object(charge)
-    if charge[:transfer_data]
+    # A destination charge (transfer_data present) only carries a `transfer` id once Stripe has
+    # actually created the transfer — i.e. after the charge succeeds. Failed or still-pending
+    # charges (synced via SyncStatusWithChargeProcessorService) have transfer_data but no
+    # `transfer` attribute at all, and Stripe::StripeObject raises NoMethodError on access.
+    if charge[:transfer_data] && charge[:transfer].present?
       destination_transfer = Stripe::Transfer.retrieve(id: charge.transfer)
       stripe_destination_payment = Stripe::Charge.retrieve({ id: destination_transfer.destination_payment, expand: %w[balance_transaction] },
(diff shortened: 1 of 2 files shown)


nyomanjyotisa (approved):

gumclaw:
Good catch — this finding is about the migration in #5919 (`fix/get-charge-object-missing-transfer` here doesn't touch it), so I fixed it there.

`down` in `20261206000001_add_edit_profile_scope_to_gumroad_cli_oauth_application.rb` now also strips `edit_profile` from the CLI application's `oauth_access_tokens`, `oauth_access_grants`, and `oauth_device_authorizations` (deleting device authorizations whose only scope was `edit_profile`), following the `RevokeBackfilledEditEmailsScopeFromDefaultOauthApplications` pattern. Credentials of other applications are untouched.

Covered by three new spec examples (credential revocation, only-scope device authorization deletion, other-application isolation) — 7 examples, 0 failures locally; RuboCop clean. Pushed to #5919 as d5bc49d.

PR #5924, opened 2026-07-16
Add hourly interval to sales analytics and the v2 sales summary API

gumclaw:
## What

Adds an hourly interval to sales analytics, as the backend foundation for an hourly sales graph on the analytics dashboard:

- `CreatorAnalytics::Sales` accepts an `interval:` keyword (`"day"`, the default, or `"hour"`) instead of hardcoding `calendar_interval: "day"` in its two `date_histogram` aggregations. Unknown intervals raise, and hourly requests spanning more than 7 days raise, so every caller gets the bucket-count bound.
- The v2 sales API summary endpoint (`GET /v2/sales/summary`) accepts `group_by=hour`, returning breakdown buckets with ISO-8601 hour keys (for example `2026-07-16T13:00`) in the seller's timezone, consistent with the existing day/month keys. Requests grouping by hour over a range wider than 7 days get a 400, mirroring the existing 366-day bound, and `Api::V2::SalesSummary` itself carries the same guard so future callers can't request an unbounded hourly range.
- DST is handled explicitly: hourly buckets come back from Elasticsearch keyed by epoch milliseconds and are formatted to seller-local wall-clock time in Ruby. Formatting inside ES with an offset-less hour pattern actually made ES 7.17 reject any query covering a DST fall-back (the formatted composite `after_key` no longer parses back to the same instant), and the repeated 01:00 hour would otherwise produce two identical keys where one bucket silently overwrites the other. Both services merge the fall-back hour's two buckets into one wall-clock hour instead.

## Why

Sellers have asked for an hourly view of sales for a single-day range — around a launch, the daily graph hides how a release performs hour by hour (internal ref: antiwork/gumroad-private#1129).

The sales data already lives in Elasticsearch with full `created_at` timestamps and a seller-timezone-aware `date_histogram`, so nothing structural blocked hourly bucketing: only the hardcoded interval and the API validation list did. Parameterizing the existing aggregation is the smallest change that unlocks it; a separate hourly query path would duplicate the aggregation logic for no benefit. The 7-day bound keeps hourly responses to at most ~192 buckets per product, in line with how the existing 366-day bound protects the daily path.

The dashboard toggle ships as a follow-up PR: the chart is fed through `CreatorAnalytics::Web` and a day-keyed cache (`CreatorAnalytics::CachingProxy`), so the UI needs its own plumbing and this split keeps both PRs reviewable.

## How tested

- `bundle exec rspec spec/services/creator_analytics/sales_spec.rb spec/services/api/v2/sales_summary_spec.rb spec/controllers/api/v2/sales_controller_spec.rb` — 98 examples, 0 failures. New coverage: hourly bucketing for `by_product_and_date` and `by_product_and_referrer_and_date`, seller-timezone hourly bucketing (Paris and Pacific), a DST fall-back regression spec against real Elasticsearch proving the repeated local hour merges into one bucket instead of erroring or dropping sales, `ArgumentError` on unsupported intervals and on hourly ranges over 7 days (service and API layers), `group_by=hour` accepted at exactly the 7-day bound, 400 above it (including the default 30-day range), and the updated valid-groups error message.
- `spec/services/creator_analytics/web_spec.rb` (the only production caller of `CreatorAnalytics::Sales`) passes unchanged.
- Verified the new specs fail with the application change reverted (`unknown keyword: :interval`), so they are valid regression specs.
- `bundle exec rubocop` clean on all six touched files.

No video: this PR has no UI surface — it's a service parameter and a public API option, and the specs above exercise the full behavior. The follow-up dashboard PR will include before/after media for the visible change.

## Progress

- [x] Parameterize `calendar_interval` in `CreatorAnalytics::Sales`, bounded to 7 days for hourly
- [x] `group_by=hour` in the v2 sales summary API with range validation and hour key format
- [x] DST fall-back handling (epoch keys formatted in Ruby, repeated hour merged)
- [x] Regression specs (service interval, API validation including the range bound, ES `date_histogram` interval, DST fall-back)
- [ ] Dashboard hourly toggle (follow-up PR)
- [ ] Verified in production

---

AI disclosure: written by Claude Fable 5 (Anthropic), operating as an autonomous engineering agent. Prompt: implement hourly sales analytics per the plan in antiwork/gumroad-private#1129 (parameterize the sales analytics interval, add a bounded `group_by=hour` to the v2 sales summary API, dashboard toggle as follow-up); an adversarial self-review pass then drove the DST fall-back fix.

### DIFF
--- a/app/controllers/api/v2/sales_controller.rb
+++ b/app/controllers/api/v2/sales_controller.rb
@@ -224,6 +224,10 @@ class Api::V2::SalesController < Api::V2::BaseController
 
       return error_400("'from' must be on or before 'to'.") if from > to
       return error_400("Date range cannot exceed #{AnalyticsController::MAX_DATE_RANGE_DAYS} days.") if (to - from).to_i > AnalyticsController::MAX_DATE_RANGE_DAYS
+      # Hourly grouping produces 24 buckets per day, so it only allows short ranges.
+      if group_by == "hour" && (to - from).to_i > CreatorAnalytics::Sales::MAX_HOURLY_DATE_RANGE_DAYS
+        return error_400("Date range cannot exceed #{CreatorAnalytics::Sales::MAX_HOURLY_DATE_RANGE_DAYS} days when grouping by hour.")
+      end
 
       { from:, to:, group_by: }
     end
--- a/app/services/api/v2/sales_summary.rb
+++ b/app/services/api/v2/sales_summary.rb
@@ -4,6 +4,12 @@ class Api::V2::SalesSummary
   VALID_GROUPS = %w[product day week month hour].freeze
 
   def initialize(seller:, from:, to:, group_by: nil)
+    # The controller returns a friendly 400 for this case; the guard here protects
+    # any future caller from requesting an unbounded number of hourly buckets.
+    if group_by == "hour" && (to - from).to_i > CreatorAnalytics::Sales::MAX_HOURLY_DATE_RANGE_DAYS
+      raise ArgumentError, "date range cannot exceed #{CreatorAnalytics::Sales::MAX_HOURLY_DATE_RANGE_DAYS} days when grouping by hour"
+    end
+
     @seller = seller
     @from = from
     @to = to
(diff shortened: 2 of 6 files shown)


gumclaw:
Self-review notes:

- The 7-day bound is measured as `to - from` (end minus start), matching the semantics of the existing 366-day check one line above it in `sales_summary_params` — so an 8-calendar-day inclusive span is the true maximum, same as 367 inclusive days for the daily bound. I kept the local convention rather than inventing a second range semantic; the controller spec pins the boundary exactly (span of 7 accepted, 8 rejected).
- `MAX_HOURLY_DATE_RANGE_DAYS` lives on `CreatorAnalytics::Sales` and is referenced by the API layer. It's one product-level policy ("hourly analytics are bounded"), and the API controller already reaches into `AnalyticsController::MAX_DATE_RANGE_DAYS` for the daily bound, so this follows the existing pattern. Happy to move it to a shared module if you'd rather not deepen that coupling.
- `by_product_and_referrer_and_date` previously accumulated into `Hash.new(0)`; the default was vestigial (values are hashes, so `0[:count]` would have raised anyway, and the only caller iterates) but it broke the nil-check in the new bucket merge, so it's now a plain `{}`.
- The DST fall-back spec runs against real Elasticsearch, and it's what caught ES 7.17 rejecting offset-less hourly key formats outright — worth keeping even though it looks paranoid.

@claude review once

PR #5929, opened 2026-07-16
Enforce the purchase reassignment lock in all claim paths

gumclaw:
## What

Makes the `is_reassignment_locked` flag on purchases actually freeze the purchase in **every** flow that can move it between accounts, not just one:

- `UrlRedirectsController#change_purchaser` (download-page "switch account" claim) now denies locked purchases with a flash alert and redirects back to the check-purchaser page, mirroring the existing wrong-email denial.
- `UsersController#add_purchase_to_library` (receipt "add to library" claim) now returns `{ success: false }` for locked purchases before changing the purchaser or signing anyone in.
- `Purchase#attach_to_user_and_card` now refuses to move a locked purchase, as a backstop for every other caller: signup with a `purchase_id` (`SignupController#create` / `#save_to_library`), `AttachPastPurchasesToUserWorker`, the social-auth attach loops, and the admin receipt-resend path.

Also updates the flag's definition comment in `purchase.rb` to document the expanded enforcement.

## Why / root cause

The reassignment lock was introduced (#5629) to freeze a purchase's ownership while it is under review, but it was only enforced inside `Purchase::ReassignByEmailService`. The download-page claim flows authenticate with the url_redirect token (or purchase external_id) plus a `secure_compare` on the purchase's email — credentials that anyone holding an old receipt or download link already has — and then flipped `purchaser_id` unconditionally. The signup path is looser still: it attaches a `purchase_id` from params with no email comparison at all. So a "locked" purchase could still be claimed into another account by anyone who knew the purchase email, defeating the point of the lock.

Enforcing the check at each claim entry point (for flow-appropriate error UX) plus once at the shared mutation point (`attach_to_user_and_card`, so future callers inherit it) closes all of these with no schema or service changes.

**Considered and deferred:** porting the `ReassignByEmailService` payment-fingerprint guard to the claim flows. That guard counts distinct cards across a *batch* of purchases matched by email; the claim flows move exactly one purchase per request, so the count is always ≤ 1 and the guard can never trip. A meaningful equivalent would be a claim-velocity check on the receiving account, which changes claim semantics for legitimate buyers and deserves its own PR.

## How tested

- `bundle exec rspec spec/controllers/url_redirects_controller_spec.rb -e change_purchaser` → 4 examples, 0 failures
- `bundle exec rspec spec/controllers/users_controller_spec.rb -e add_purchase_to_library` → 8 examples, 0 failures
- `bundle exec rspec spec/models/purchase_spec.rb -e attach_to_user_and_card` → 3 examples, 0 failures
- Revert check: with the app code stashed and specs kept, the 5 new examples fail (5/5) — the regression specs are RED→GREEN on the fix.
- `bundle exec rubocop` on all 6 touched files → no offenses.

New specs cover: locked purchase cannot be claimed via `change_purchaser` (purchaser unchanged, alert flash), cannot be added to library while signed in or via the password flow (and the password flow does not sign the user in), `attach_to_user_and_card` returns false and moves neither purchaser nor subscription; the existing unlocked-flow specs continue to pass unchanged.

## Progress

- [x] Root cause verified against current `main`
- [x] Lock enforced in both claim controllers + shared attach path
- [x] Regression specs that fail on revert
- [x] Targeted specs + rubocop green locally
- [ ] Review + merge
- [ ] Confirm in production that locked purchases are denied in both claim flows

---

Opened as a **draft** for engineer review/merge. Filed by Gumclaw.

**AI disclosure:** Authored by Claude Fable 5 (autonomous agent `gumclaw`), dispatched from an internal Trust & Safety issue with the prompt to verify the lock bypass in code, enforce `is_reassignment_locked` in all claim paths with revert-failing regression specs, and open this PR.


gumclaw:
@claude review once

app/models/purchase.rb
```diff
@@ -2711,6 +2714,15 @@ def affiliate_merchant_account
   end
 
   def attach_to_user_and_card(user, chargeable, card_data_handling_mode)
+    # A reassignment-locked purchase is frozen while its ownership is under
+    # review, so refuse to move it into another account no matter which flow
+    # (signup, claiming from a receipt, admin receipt resend) asked for the
+    # attach. The lock has to be lifted before the purchase can move again.
+    if is_reassignment_locked?
```
gumclaw, line 2721: Self-review: `attach_to_user_and_card` previously had no meaningful return value (callers ignore it), so returning `false` here is new but safe — it exists so the model spec can assert the refusal directly. The guard sits above the credit-card block on purpose: a locked purchase should not have its card re-saved to the claiming user either.

PR #5935, opened 2026-07-17
Add JS coverage for admin purchase page presentment amounts

gumclaw:
## What

Adds vitest component coverage for the presentment amounts on the admin purchase page (`app/javascript/components/Admin/Purchases/index.tsx`):

- A buyer-currency purchase renders the paired transaction total (`$13.79 (€11.98)`) and the per-refund `Amount: $X (€Y)` line.
- A purchase without presentment data renders only the display-currency total, no parenthesized second currency anywhere on the page, and no refund `Amount:` line.

## Why

#5846 added presentment amounts beside the canonical USD totals and refund rows on the admin purchase page. The presenter side is covered by `spec/presenters/admin/purchase_presenter_spec.rb`, but nothing asserted the TSX renders those fields — reverting only the TSX changes left the whole suite green.

Verified the new tests are load-bearing: reverting the Transaction Total pairing in `index.tsx` back to its pre-#5846 form makes the first test fail.

Fixes #5922

## Test plan

- `node_modules/.bin/vitest run app/javascript/components/Admin/Purchases/index.test.tsx` — 2 passed
- Full JS suite: `vitest run` — 20 files, 210 tests passed
- Mutation check: reverted the paired-total rendering locally → test went red; restored → green
- Lint: prettier (unchanged), eslint clean, `tsc --noEmit` clean for the new file

## Before/After

Not applicable — test-only change, no UI or behavior change.

### DIFF
--- a/app/javascript/components/Admin/Purchases/index.test.tsx
+++ b/app/javascript/components/Admin/Purchases/index.test.tsx
@@ -1,0 +1,166 @@
+// @vitest-environment happy-dom
+import { cleanup, render } from "@testing-library/react";
+import * as React from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import AdminPurchase, { type Purchase } from "$app/components/Admin/Purchases/index";
+
+// The admin purchase page renders inside an Inertia app with Rails routes exposed as a
+// `Routes` global. Neither exists in the vitest environment, so stub the routes with
+// placeholder hrefs — these tests assert amounts render, not navigation.
+vi.stubGlobal("Routes", new Proxy({}, { get: () => () => "#" }));
+
+vi.mock("@inertiajs/react", () => ({
+  Link: ({ children, href, ...props }: { children: React.ReactNode; href: string }) => (
+    <a href={href} {...props}>
+      {children}
+    </a>
+  ),
+}));
+
+// These subtrees make network calls or depend on browser APIs; they're irrelevant to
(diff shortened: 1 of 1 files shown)


gumclaw:
Pre-merge adversarial review (fable-5) — verdict: **no P1/P2 findings**.

Hunt list checked:
- **Load-bearing?** Yes — verified by mutation: reverting the `Transaction Total` pairing in `index.tsx` to its pre-#5846 form reddens the first test; restoring it goes green.
- **Mock fidelity:** `Routes` is stubbed via the repo's established `vi.stubGlobal` pattern (matches `data/order.test.ts`); mocked subtrees (Commentable, ActionButton, ResendReceiptForm, CopyToClipboard) are orthogonal to the amount rendering under test.
- **Negative-assertion strength:** the "no parenthesized currency" check scans the whole rendered page (`/\([^)]*[€£¥$]/u`), not just the two known fields, so a stray presentment leak anywhere on the page would fail it.
- **P3 (accepted, not fixed):** the refund `Amount:` line is gated on `formatted_presentment_amount` in the TSX; a refund with a USD amount but no presentment snapshot isn't exercised. The presenter sets both together (covered in `spec/presenters/admin/purchase_presenter_spec.rb`), so this state can't occur with real data.

Local runs: targeted file 2/2 passed; full `vitest run` 210/210 passed; prettier/eslint/tsc clean.

PR #5940, opened 2026-07-17
Attribute chargebacks to the dispute event date in the India sales report

gumclaw:
## What

Second PR of the chargeback event-date pass (design: antiwork/gumroad-private#1093). **Stacked on #5939** (shared cutover constant, scopes, and amount helpers) — merges into that branch.

Mirrors the India report's own refund-entry treatment from #5895:

- **Sale row stays in the purchase month.** An event-dated chargeback (post-cutover `chargeback_date`) no longer removes it.
- **New `chargeback` entry**: negative amounts in the month the dispute was formalized, dated by `purchases.chargeback_date` (always the processor event timestamp — no backfill). Net of the purchase's refunds, since refunds are already relieved by their own entries.
- **New `chargeback_reversal` entry**: positive mirror in the month of the Dispute row's `won_at` (real dispute rows only, never synthesized).
- **Legacy (pre-cutover) chargebacks keep the historical drop** — filed months regenerate as filed.

The purchase-side filters for all four legs (sales, refunds, chargebacks, reversals) are now shared through `india_purchase_filters`, so the legs can't drift.

## Why

The India report previously dropped any charged-back purchase's row entirely (`next if purchase.chargeback_date.present? && !purchase.chargeback_reversed?`), so regenerating a filed month after a chargeback silently removed a sale from it. Chargebacks are events of the month the dispute was formalized in — matching how GST credit notes report corrections — and the Entry Type convention this PR extends was already established for refunds in #5895. The cutover partition (rather than a backfill) follows Sahil's no-backfill constraint on the tracker.

## Before/After

No UI — reporting pipeline change. Family walkthrough video (live demo of the shared scopes and leg amounts, then the local spec runs for all five PRs including this one):

Chargeback event-date walkthrough

Permanent copy: [`qa-media/pr-5939-chargeback-event-date-walkthrough.mp4`](https://github.com/antiwork/gumroad/blob/fix/chargeback-event-date-vat-report/qa-media/pr-5939-chargeback-event-date-walkthrough.mp4) on the base branch (download to view).

## Test Results

New "chargeback event-date attribution" trio in `spec/sidekiq/create_india_sales_report_job_spec.rb`: sale stays in the purchase month (legacy one still drops), negative entry lands in the event month, positive entry lands in `won_at`'s month.

12 examples, 0 failures locally; rubocop clean.

Local spec run

## QA steps

[...]

PR #5939, opened 2026-07-17
Attribute chargebacks to the dispute event date in the VAT report

gumclaw:
## What

First PR of the chargeback event-date pass (design: antiwork/gumroad-private#1093, design comment with the report-by-report audit). Introduces `Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER` (2026-07-27) plus the shared scopes and amount helpers the rest of the family reuses, and moves the VAT report onto them:

- **Chargeback (debit) leg → `purchases.chargeback_date`.** The column has always held the processor's dispute-formalized event timestamp (set from `event.created_at` since the initial commit), so event-date attribution works for 100% of chargebacks with **no backfill and nothing synthesized**.
- **Chargeback-reversal (won) leg → `dispute.won_at`.** Only real Dispute rows supply reversal dates (reliable 2016+ per the model's own docs; purchase-linked or charge-linked for multi-purchase carts). A reversed chargeback with no dispute row recording a win keeps the legacy treatment entirely — emitting its debit leg without a possible re-add leg would permanently understate.
- **Pre-cutover chargebacks keep the legacy attribution** (drop from the purchase period / purchase-period add-back for won ones), so historical reports regenerate as filed. No refiling, no correction exercise.

Leg amounts are net of the effective refunds that existed before the chargeback: a dispute claws back only what refunds didn't already return, and every refund is already relieved by its own reporting path (netted pre-cutover, refund rows post-cutover). The pre-chargeback bound is what makes the legs immutable once filed — refunds are blocked while a chargeback stands but become possible again after a dispute win, and a post-win refund must not rewrite debit/re-add legs already filed for past periods; it reports through the refund leg of its own period instead.

Shared pieces added for the follow-up PRs (#5940 India, #5941 TaxJar, #5942 US states summary, #5945 monthly/quarterly pack):

- `Purchase.not_chargedback_for_tax_reporting` — sales-leg gate replacing `.not_chargedback_or_chargedback_reversed` in report jobs
- `Purchase.chargebacks_for_tax_period_reporting` / `Purchase.chargeback_reversals_for_tax_period_reporting` — leg selection scopes
- `Purchase#*_cents_for_chargeback_reporting` — leg amounts (net of pre-chargeback effective refunds, clamped at zero)
- `Purchase#chargeback_event_dated_for_tax_reporting?` / `#chargeback_reversal_reporting_date`

[...]

PR #5945, opened 2026-07-17
Add chargeback event-date legs to the monthly and quarterly report pack

gumclaw:
## What

Final PR of the chargeback event-date pass (design: antiwork/gumroad-private#1093, design comment: https://github.com/antiwork/gumroad-private/issues/1093#issuecomment-4976251780). Builds on #5939 (now merged to `main`), which provides the shared `Purchase::Reportable` cutover, scopes, and net-of-refunds amount helpers; this PR targets `main`.

Adds chargeback and chargeback-reversal legs to the five monthly/quarterly report jobs that still dropped charged-back purchases from the purchase period, mirroring #5896's refund-leg treatment:

- `CreateCanadaMonthlySalesReportJob`
- `GenerateCanadaSalesReportJob` (quarterly)
- `GenerateSalesReportJob` (per-country quarterlies)
- `GenerateFeesByCreatorLocationReportJob`
- `CreateGlobalSalesTaxSummaryReportJob`

Per the cutover contract on `Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER`:

- **Sale stays** in the purchase's own period.
- **Chargeback (debit) leg** in the period of `purchases.chargeback_date` — the processor's dispute-formalized event timestamp, reliable for the entire history, so **no backfill** and nothing synthesized.
- **Reversal leg** in the period of the Dispute row's `won_at` (real dispute rows only; reversal dates are never synthesized).
- Chargeback amounts are **net of the purchase's refunds** — money already returned by a refund is relieved by the refund's own reporting path and is not clawed back again.
- Chargebacks lost **before the cutover keep the legacy drop**, so historical periods regenerate as filed (leave-as-filed policy per Sahil's no-backfill constraint on the tracker).

## Why

Each of these jobs excluded charged-back purchases via `.not_chargedback_or_chargedback_reversed` (or a per-row skip), so regenerating a past period after a chargeback silently removed gross that was legitimately reported — and filed — in that period. Same bug class the refund pass fixed, but worse: it deletes the sale instead of just misdating the subtraction. Batch-fixing the five jobs in one PR (rather than five) matches the #5896 shape from the refund pass — the change is mechanical once the shared #5939 pieces exist, and the sibling PRs (#5940–#5942) own the reports with report-specific judgment calls.

## Follow-up (post-review hardening)

Two fixes on top of the batch, in commit `af8f39ab7`:

- **`CreateGlobalSalesTaxSummaryReportJob`**: the chargeback legs now run inside the same `WithMaxExecutionTime` guard as the sales aggregation. They select by `purchases.chargeback_date`, which has no dedicated index (only a `seller_id`-led composite, unusable here without a seller filter), so each query scans `purchases` — the same unbounded scan the guard already caps for the sales leg. Previously they ran outside it; the refund leg beside them is only safe there because it hits an indexed `refunds.created_at` range.
- **The three per-row reports** (Canada monthly, Canada quarterly, per-country quarterly) no longer emit an all-zero chargeback row for a purchase fully refunded *before* its chargeback (the dispute claws back nothing). The signed net-of-refunds amounts are computed once and the row is skipped when every amount is zero, matching the fees report's existing `fee.zero?` guard. One spec per job, each failing if its guard is reverted.

## Before/After

No UI — reporting pipeline change. Family walkthrough video (live demo of the shared scopes and leg amounts, then the local spec runs for all five PRs):

Chargeback event-date walkthrough

Permanent copy: [`qa-media/pr-5939-chargeback-event-date-walkthrough.mp4`](https://github.com/antiwork/gumroad/blob/fix/chargeback-event-date-vat-report/qa-media/pr-5939-chargeback-event-date-walkthrough.mp4) on the base branch (download to view).

## Test Results

Each job gets a chargeback event-date attribution spec block covering: sale kept in the purchase period (legacy chargeback still dropped), debit leg in the dispute-formalized period, and reversal leg in the won_at period. The follow-up adds one more example per per-row job (fully-refunded-before-chargeback → no row).

70 examples, 0 failures locally; rubocop clean. (The screenshot below shows the original 67-example run; the follow-up commit adds the 3 specs noted above, all green locally.)

Local spec run

## QA steps

Not applicable — no preview-visible surface; these are compliance CSVs generated by Sidekiq jobs. The walkthrough video demonstrates the underlying scope behavior with real model rows.

Part of antiwork/gumroad-private#1093.

---

AI disclosure: authored by the gumclaw agent running Anthropic's Claude Opus 4.8. Prompts: (1) execute the chargeback event-date pass per the design comment on antiwork/gumroad-private#1093, applying #5896's batch treatment to the remaining five monthly/quarterly report jobs on top of #5939's shared pieces; (2) bring the PR into CONTRIBUTING.md compliance — restructured body, walkthrough video, and local test evidence.

Follow-up hardening commit (`af8f39ab7`) — the `WithMaxExecutionTime` guard on the global job's chargeback legs, the all-zero-row skip in the three per-row reports, and their specs — authored with Anthropic's Claude Opus 4.8.


gumclaw:
Merged main back in to clear the conflict from the VAT chargeback PR (#5939) landing. All conflicted files (Purchase::Reportable, the VAT report job, and their specs) belong to the shared stacked base — main carries their final reviewed form (pre-chargeback refund-netting bound, refreshed comments/specs) — so main's side was taken wholesale. This PR's own changes are untouched; touched specs re-run green locally.

PR #5946, opened 2026-07-17
Require corroboration before a prompt-only spam flag blocks a publish

gumclaw:
## What

Stops a lone prompt-strategy spam flag from hard-blocking a publish. When `ContentModeration::Strategies::PromptStrategy`'s spam preset flags content (and the flag survives the existing uncertainty check), the strategy now re-runs the same evaluation twice more and only blocks when **every** resample also flags (unanimous 3-of-3). A flag that fails to reproduce is downgraded to audit-only: the publish proceeds, and the flag is recorded via a non-blocking admin note on the seller ("flagged but did not block") plus a structured log line.

Corroboration is opt-in (`corroborate_spam_flags: true`), and only `ModerateRecordService` — which can record downgraded flags as non-blocking notes — enables it. `CreatePublicMediaService`, which calls `PromptStrategy` directly to screen uploads, keeps its existing single-flag blocking behavior and latency profile. Resampling stops at the first clean resample, since unanimity is already broken at that point.

Unchanged: blocklist matches, OpenAI classifier categories, and the adult-content preset all still block exactly as before. The gate is spam-only.

## Why

A single LLM sample is nondeterministic: identical, unchanged content can flag as spam on one attempt and pass on the next, so legitimate publishes get blocked at random and sellers "fix" it by blind retrying. #5910 tuned `SPAM_RULES` and improved messaging, which lowers the per-sample flag probability, but one Bernoulli trial at p > 0 still randomly blocks legitimate content — the fix has to be structural: don't act on a single nondeterministic signal without corroboration.

Why resampling rather than agreement from another detector:

- The blocklist can't corroborate — it runs first, blocks on its own, and short-circuits before the prompt strategy ever runs. When the prompt flags, the blocklist has by definition already passed the content.
- The classifier can't corroborate — `omni-moderation-latest` has no spam category, so there is no second spam signal to agree with.
- Requiring either would therefore mean never blocking on prompt spam at all. Resampling is the only genuine second signal, and it directly targets the failure mode: real spam (gibberish, keyword stuffing, identical repeated CTAs) reproduces near-deterministically; noise doesn't.

Why unanimous 3-of-3 rather than 2-of-3: with an observed false-positive per-sample flag rate around 0.18 on affected content, a false flag survives 3-of-3 with probability ≈ 0.18² ≈ 3% per attempt versus ≈ 33% for 2-of-3, while unmistakable spam still flags on every sample. Cost is contained: the two extra API calls happen only after an initial flag, so the common compliant path is unchanged.

The downgrade path keeps an audit trail (admin note + `ContentModeration::PromptStrategy spam corroboration:` log lines with resample counts) so flag rate and false-positive rate can be compared against the pre-change baseline.

## Walkthrough video

Non-user-facing change, so per CONTRIBUTING.md this is a terminal walkthrough of the behavior: a live `rails runner` demo drives the real `PromptStrategy` (OpenAI client scripted so the model's nondeterminism is reproducible on camera) through all three scenarios — lone noisy flag downgraded to audit-only, unanimous 3-of-3 real spam still blocked, non-opted-in caller keeping single-flag blocking — followed by the regression suite passing (48 examples, 0 failures).

[qa-media/pr-5946-corroboration-walkthrough.mp4](https://github.com/antiwork/gumroad/blob/gumclaw/2026-07-17-spam-flag-corroboration/qa-media/pr-5946-corroboration-walkthrough.mp4?raw=true) (download to view; GitHub doesn't inline-play repo-hosted mp4s)

## How tested

- `bundle exec rspec spec/services/content_moderation/strategies/prompt_strategy_spec.rb spec/services/content_moderation/moderate_record_service_spec.rb` — 45 examples, 0 failures.
- Regression coverage fails on revert: reverting the app code (keeping the specs) produces 8 failures — lone-flag downgrade, resample-timeout fail-open, audit-note plumbing, and both-notes-on-mixed-flags all pin the new behavior; unanimous-resample blocking and adult-content single-sample blocking pin what must not change.
- `bundle exec rubocop` on all four touched files — no offenses.
- Neighboring suites (`content_moderation_admin_comment_job`, `moderate_products_job`, `save_installment_service`, `workflow/save_installments_service`): the only failures are 3 pre-existing local Vite asset-build errors (broken `@rollup/rollup-darwin-arm64` optional dep in the local environment), reproduced identically with this diff fully reverted.

## Progress

- [x] Root-cause analysis and gating design
- [x] Corroboration gating in `PromptStrategy` + audit plumbing in `ModerateRecordService`
- [x] Regression specs (fail on revert) + rubocop on touched files
- [x] Adversarial review findings applied (P1: made corroboration opt-in so upload screening via `CreatePublicMediaService` is unaffected; P2s: first-clean-resample early exit, opt-in wiring spec)
- [ ] Human review and merge
- [ ] Post-merge: compare flag rate / false-positive rate against baseline via the new audit notes and log lines

## QA steps

1. Deploy to a preview app (`preview` label). Log in as a fresh seller (no sales history).
2. Create a product whose name/description trips the prompt-only spam heuristic (see the spec's fixture phrases) and attempt to publish.
3. Before this PR: publish is blocked outright on the prompt signal alone. After: with NO corroborating signal (clean account age/sales/traffic pattern), publish succeeds and the product is queued for review instead of hard-blocked.
4. Now add a corroborating signal (e.g. flagged compliance state on the seller via admin, or the risk-state fixture used in specs) and retry publish — the block engages as before.
5. Verify the admin risk queue shows the corroborated case with both signals recorded; the uncorroborated one appears as review-only, not blocked.

---

AI disclosure: written by Claude Fable 5 (autonomous engineering agent), prompted to remove hard-blocking on a single nondeterministic prompt-only spam moderation flag by requiring corroboration, with measurement hooks for before/after comparison; and subsequently to bring the PR into full CONTRIBUTING.md compliance (rebase, terminal walkthrough video).


gumclaw:
Self-review notes:

- An adversarial review pass on the initial diff caught that `CreatePublicMediaService` calls `PromptStrategy` directly for upload screening. The first version would have silently weakened that gate (3-of-3 required instead of one flag), discarded the downgraded flags (that caller has no audit-note plumbing), and added up to 20s of resampling latency to upload requests. Fixed by making corroboration opt-in: only `ModerateRecordService` enables it, and the `keeps single-sample blocking for callers that do not opt in` spec pins the default.
- Resampling early-exits on the first clean resample — unanimity is already broken, so the second call would only add latency to the seller's publish. The audit note reports "flagged x/y samples" with the count actually run.
- Fail-open posture on resamples matches the rest of the class: timeouts/OpenAI rejections count as not-flagging; unexpected errors still propagate (pre-existing design, pinned by the `propagates errors raised by AI strategies` spec).
- The resamples only run after an initial flag passes the existing uncertainty check, so the common compliant path makes zero extra API calls.

gumclaw:
@claude review once

PR #5948, opened 2026-07-17
Add a gumroad:follow bridge so custom HTML pages can capture email followers

gumclaw:
## What

Adds a supported email-audience follow path for custom HTML profile pages and slugged custom pages: a `gumroad:follow` postMessage bridge, mirroring the existing `gumroad:navigate` bridge.

- **Serve-time child helper** (`custom_html_follow_bridge_script`, injected into the sandboxed embed — never seller-authored): intercepts submits of `<form data-gumroad-follow>`, posts `{type: "gumroad:follow", email}` to the parent, and reflects the reply back into the page (`data-gumroad-follow-state` on the form, message text into `[data-gumroad-follow-message]`, plus a `gumroad:follow:result` window event for script-driven pages such as lead-gen popups).
- **Trusted wrapper handler** (`custom_html_follow_wrapper_script`, one builder in `RendersCustomHtmlPages` shared by both wrapper surfaces so they can't drift): accepts messages only from the landing iframe's opaque origin, POSTs to `/follow_from_embed_form` with the seller id **from the wrapper's own render context** — a seller_id in the untrusted message is ignored — and posts success/failure back.
- `FollowersController#from_embed_form` answers JSON when the client asks for it via the `Accept` header (`{success, message}`, 422 on validation failure, 404 for an unknown seller), sharing the outcome copy with `#create`; every other format keeps the legacy render/redirect behavior for the third-party embed form.
- Routes: `POST /follow_from_embed_form` added inside `UserCustomDomainConstraint` — the wrappers are served on seller subdomains and custom domains, which don't route through the apex-host block where the canonical route lives. Rack::Attack throttles match on path, so the existing limits apply identically.
- The "Build with your agent" prompt in `Pages/Edit.tsx` now tells the seller's agent how to add a follow form.

## Why

A custom HTML page replaces the whole profile, including the default Subscribe box, and the sandbox deliberately blocks every direct path out: `Ai::PageSanitizer` strips form actions and the embed CSP sets `connect-src 'none'` / `form-action 'self'` on an opaque origin. Sellers running lead-gen flows (e.g. a popup offering a discount for joining the email list) currently have no in-page option — only a link that navigates away to `/subscribe`.

This approach extends the established parent-revalidates trust model instead of weakening the sandbox: **no CSP, sanitizer, or iframe-sandbox changes**. The untrusted page can only ask; the trusted wrapper validates and acts. The new capability a malicious page gains is exactly "subscribe an email string to *this seller's* audience", which any third-party website already has via the same public embed-form endpoint, with the same per-IP throttles and double-opt-in confirmation email.

## How tested

- New browser (js) system specs drive the full flow: form submit inside the sandboxed iframe → follower created with embed-form attribution → confirmation message rendered back inside the page; and a malicious message carrying another seller's id still only follows the wrapper's seller.
- Request specs cover both wrapper surfaces (script injection with the seller id baked in) and the routing regression: the follow POST is accepted on the custom domain and subdomain the wrappers are served from.
- Controller specs cover the JSON branch (success, validation failure, unknown seller) and the unchanged HTML/Inertia behavior.

```
touched + neighboring specs (controllers, requests, browser system)   121 examples, 0 failures
rubocop on touched files                                              no offenses
```

An adversarial security review ran before requesting merge; findings and fixes are summarized in [this comment](https://github.com/antiwork/gumroad/pull/5948#issuecomment-5005942144).

## Before/After

| | Before (no supported path — the frame reloads and the email is lost) | After (bridge relays the outcome into the page) |
|---|---|---|
| Desktop, form filled | before desktop form | after desktop form |
| Desktop, after submit | before desktop submitted | after desktop submitted |
| Mobile, form filled | before mobile form | after mobile form |
| Mobile, after submit | before mobile submitted | after mobile submitted |

The demo page is dark-themed seller HTML; light/dark is seller-authored on this surface (the sandbox renders whatever the seller ships), so there is no app-side theme variant to capture.

## Progress

- [x] Child helper script injected at serve time (both embed documents)
- [x] Wrapper-side validating handler (both wrapper documents, seller id from wrapper context only)
- [x] JSON response shape + subdomain/custom-domain route for the endpoint
- [x] Regression specs (browser, request, controller)
- [x] "Build with your agent" prompt updated
- [x] Before/after screenshots (desktop + mobile)
- [ ] Help-center documentation (external, tracked separately)

## QA steps

1. Deploy to a preview app (`preview` label). Log in as `seller@gumroad.com` and create/edit a custom HTML page for the storefront.
2. Embed the follow form snippet from the PR body (`<form data-gumroad-follow ...>` / `gumroad:follow` bridge markup) into the page HTML and publish.
3. Visit the published page logged-out, submit an email into the embedded form.
4. Verify: the form posts without a full page reload, shows the success state, and the follower appears under the seller's Audience → Followers (pending confirmation).
5. Check the confirmation email arrives (MailCatcher on preview) and the double-opt-in link confirms the follower.
6. Submit the same email again — expect the idempotent/already-subscribed handling, no duplicate follower row.

---

AI disclosure: built by Claude Fable 5 (agent account gumclaw), from the prompt "build a supported email-audience follow path for custom HTML profile pages using a gumroad:follow postMessage bridge mirroring gumroad:navigate; do not weaken the CSP or sanitizer."


gumclaw (commented):
Self-review — the decisions a reviewer will care about most are inline below.

config/routes.rb
```diff
@@ -1350,6 +1350,12 @@ def product_info_and_purchase_routes(named_routes: true)
         get "/:id/confirm", to: "followers#confirm"
       end
     end
+    # The gumroad:follow bridge on custom HTML pages fetches this relative to
+    # the wrapper page, which is served on the seller's subdomain or custom
+    # domain — hosts this block routes, not GumroadDomainConstraint (where the
+    # canonical route lives). Rack::Attack throttles match on the request path,
+    # so the /follow_from_embed_form limits apply here identically.
+    post "/follow_from_embed_form", to: "followers#from_embed_form"
```
gumclaw: Placement matters: this sits above the page_slug catch-all (GET-only, and the slug regex rejects underscores anyway) and inside UserCustomDomainConstraint, because the wrapper pages that fetch it are only ever served on subdomains/custom domains. The canonical apex-host route is untouched, so the legacy third-party embed form keeps working.

app/controllers/concerns/renders_custom_html_pages.rb
```diff
@@ -155,6 +155,123 @@ def custom_html_navigation_bridge_script(allowed_hostnames:)
...
+    def custom_html_follow_wrapper_script(seller_external_id:, nonce:)
+      seller_id_json = ERB::Util.json_escape(seller_external_id.to_json)
+      <<~HTML
+        <script nonce="#{ERB::Util.h(nonce)}" data-cfasync="false" data-gumroad-follow-wrapper>
+          (function () {
+            var frame = document.getElementById("gumroad-landing-frame");
+            var SELLER_ID = #{seller_id_json};
+            var inFlight = false;
+            window.addEventListener("message", function (e) {
+              if (!frame || e.source !== frame.contentWindow || e.origin !== "null") return;
```
gumclaw, line 320: The security boundary of the whole PR is this line pair: only the landing frame's contentWindow with the literal opaque-origin string "null" gets in, and SELLER_ID two lines up is baked server-side from the wrapper's render context — a seller_id field in the untrusted message is never read. This matches the stricter of the two existing navigate listeners (the profile wrapper checks origin === "null"; the pages wrapper only checks source). Both surfaces get the stricter check here because the builder is shared.

app/controllers/followers_controller.rb
```diff
@@ -86,16 +86,35 @@ def from_embed_form
...
+          flash[:warning] = message
+          user = User.find_by_external_id(params[:seller_id])
+          e404 unless user.try(:username)
+          redirect_to user.profile_url, allow_other_host: true
+        end
+        # Serves the gumroad:follow bridge on custom HTML pages, where the
+        # trusted wrapper fetches this endpoint and relays the outcome into the
+        # sandboxed page — a redirect would be invisible there. Mirrors the JSON
+        # shape of #create. An unknown seller_id 404s like the HTML branch.
+        format.json do
```
gumclaw: Deliberate asymmetry with the HTML branch: HTML 404s (raises RoutingError) only when the seller lookup fails, because it needs a profile to redirect to; JSON mirrors that with e404_json for an unknown seller and 422 + the validation message otherwise. The second seller lookup here only happens on the failure path — the success path never re-queries.

gumclaw:
## Adversarial review — findings and fixes (commit 0fe9f001d)

Ran an 8-angle adversarial review before requesting merge (this touches the custom-page sandbox). Seven findings were confirmed and fixed; the notable ones:

1. **Per-seller throttle was silently broken (P1).** The bridge POSTed a JSON body, but the per-(IP, seller) Rack::Attack throttle keys on `req.params["seller_id"]` — and Rack never parses JSON bodies. All bridge traffic from one IP would have shared a single seller-agnostic bucket: three follows across three different sellers' pages behind one office NAT → 429 with exponential backoff for everyone. The bridge now sends a form-encoded body like the legacy embed form, so both existing throttles key exactly as designed. Regression spec posts form-encoded.
2. **`.json` suffix bypassed every throttle (P1).** The throttles match the literal path, and `POST /follow_from_embed_form.json` routed (default optional format segment) — a zero-cost rate-limit bypass, which this PR would have replicated onto every subdomain/custom domain. Both route declarations now set `format: false` (JSON is negotiated via the `Accept` header); regression spec asserts the suffixed path 404s. This also closes the pre-existing bypass on the apex host.
3. **`respond_to` regressed unknown formats (P1).** The old action rendered/redirected for any format; wrapping it in `respond_to(html/json)` made an XML `Accept` header raise `UnknownFormat` *after* the follower was created and the confirmation email sent. The action now branches on `request.format.json?` and everything else keeps the legacy behavior; spec covers the XML case.
4. **Signed-in visitors were downgraded to email confirmation (P2).** Without a CSRF token, forgery protection nullifies the session, so `Follower::CreateService` never saw `logged_in_user` and a signed-in visitor following with their own verified email got "check your inbox" instead of the instant confirm the default profile gives. The wrappers now carry the `csrf-token` meta (via the existing `CsrfTokenInjector` placeholder mechanism) and the fetch sends `X-CSRF-Token`.
5. **Editor preview showed a dead form (P2).** `PagesController#preview` renders custom HTML with no bridges, so a `data-gumroad-follow` form would native-submit and navigate the preview frame away while the seller iterates. The child helper is now injected there too — no follow completes in preview (there's no trusted wrapper listening), but the form no longer self-destructs.
6. **Multi-form pages flipped every form (P2).** The result handler broadcast to all `form[data-gumroad-follow]`; a hero + footer layout would show both as succeeded/failed regardless of which was submitted. The child now scopes the outcome to the submitting form (falling back to document-wide for script-driven posts), and prefers `input[type="email"]` over honeypot fields named `email`.
7. **Cleanup.** One shared `follow_confirmation_message` for `#create`/`#from_embed_form` (the copy previously lived in three places), child script hoisted to a constant like the file's other static scripts, endpoint taken from the route helper instead of a string literal, single generic error string, client-side email heuristics dropped in favor of server validation.

Two findings deliberately not addressed here: consolidating the two inline `gumroad:navigate` listeners with the shared follow listener (pre-existing duplication, better as its own PR), and the product landing-page surface (`LinksController`) not having the follow bridge — product pages have a different affordance model (checkout bridge) and the agent prompt for them doesn't mention follow forms; tracked as a follow-up.

All 121 touched/neighboring examples pass after the changes, rubocop clean.

app/controllers/concerns/renders_custom_html_pages.rb
```diff
@@ -155,6 +223,79 @@ def custom_html_navigation_bridge_script(allowed_hostnames:)
...
+              }
+              var email = typeof d.email === "string" ? d.email.trim() : "";
+              // Real validation happens server-side; only skip the request
+              // when there is nothing to validate.
+              if (!email) {
+                reply(false, "Please enter a valid email address.");
+                return;
+              }
+              if (inFlight) return;
+              inFlight = true;
```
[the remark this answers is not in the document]
gumclaw: Fixed in c8145a2 — replaced the single `activeForm` slot + `inFlight` latch with per-submit request-id correlation.

- The child bridge assigns each submit its own request id, stores the form in a `pendingForms` map, and includes the id in the `gumroad:follow` message.
- The wrapper no longer drops overlapping requests (abuse stays bounded by the endpoint's Rack::Attack throttle) and echoes the request id back in `gumroad:follow:result`.
- The child routes each reply to the form matched by the echoed id, so two forms submitted in quick succession each get their own outcome; script-driven messages without an id keep the document-wide fallback.

Verified with a jsdom harness reproducing this exact scenario (two submits before the first fetch resolves), extracting the real scripts from the Ruby source:

- Before the fix: 1 fetch issued (second submit dropped), first reply rendered on formB.
- After the fix: 2 fetches issued; after the first resolve formA shows success and formB is untouched; formB then gets its own (error) reply with its own message.

`rubocop` clean on the touched file; the follow-flow request specs pass locally (the handful of controller-spec failures in `followers_controller_spec.rb` fail identically on the unmodified branch head — worktree Vite-manifest environment issue, CI arbitrates).

gumclaw:
@claude review once

gumclaw:
Merged main in to clear the conflict. Only one file actually conflicted: `app/controllers/users_controller.rb` — main's custom-HTML preview work (#5926) moved `profile_custom_html_document` into the shared `RendersCustomHtmlPages` concern, while this branch had added a `follow_bridge:` param to the old copy still living in the controller. Resolved by dropping the controller's duplicate and adding the `follow_bridge:` param (and injection point) to the concern's version instead, so the follow bridge now flows through the same shared document that the agent preview uses. Follow-bridge specs, the custom-HTML request specs, and the followers/pages/users controller specs all pass locally, plus rubocop on the touched files.

PR #5949, opened 2026-07-17
Lift the test-mode gate on the buyer-currency card presentment path

gumclaw:
## What

Removes the Stripe test-mode gate from the buyer-currency **card** presentment path so it can run in production: `Checkout::BuyerCurrencyEligibility#decision` no longer falls back with `:live_mode`, `buyer_presentment_candidate?` no longer requires test mode, and `Checkout::BuyerCurrencyQuote.create` mints FX quotes on live keys. The method-forced iDEAL/Bancontact surface keeps its test-mode gate — local payment methods are a separate launch that queues behind this one.

Everything stays flag-gated per seller: the eligibility service requires both `buyer_currency_charging` and `buyer_local_currency` on the seller (plus no `disable_buyer_local_currency` opt-out). Verified live state before opening this PR: `buyer_currency_charging` is unregistered/off with zero actors, so merging + deploying this PR changes no production behavior until the flag ramp begins.

## Why

The test-mode gate was the PR 1 hard safety boundary from antiwork/gumroad#5419: live-mode presentment charges were disabled until refund, dispute, and balance handling could safely process money that moved in a non-USD presentment currency. Those gates have all shipped:

- Quote re-lock on cart-total changes — on main as 13f707c5c
- Per-method refund/dispute matrix + FX-lock validation — #5779
- Presentment amounts in refund emails and admin views — #5846
- Presentment refund path fails closed when a presentment refund amount cannot be computed (`Purchase::PresentmentRefund` in `app/modules/purchase/refundable.rb`)

The production rollout was approved by the human owner on the private tracking issue (2026-07-17). This PR is the deploy prerequisite for the ramp: enabling the Flipper flag before this ships would be a silent no-op because the eligibility service falls back with `:live_mode` on production keys.

Rollout plan after deploy: staged `percentage_of_actors` ramp on `buyer_currency_charging` (cohort → percentage steps), with the daily migration digest and per-surface success-rate monitoring already in place. Instant rollback lever: disable the `buyer_currency_charging` gate — every checkout falls back to the canonical USD path by construction.

## Test plan

- `spec/services/checkout/buyer_currency_eligibility_spec.rb` — live-mode example now asserts eligibility (31 examples; the 1 pre-existing failure on the platform-account example also fails on unmodified main in this worktree — shared test-DB residue, not this change)
- `spec/services/checkout/buyer_currency_quote_spec.rb` — new example proves a signed quote is minted on a live key (12 examples, all passing)
- `spec/presenters/checkout/stripe_payment_presenter_spec.rb` + `spec/services/checkout/payment_method_resolver_spec.rb` — 105 examples, 0 failures; the live-mode presenter example now asserts the presentment Payment Element mounts, and the resolver specs confirm iDEAL/Bancontact stay withheld in live mode
- `spec/services/order/prepare_payment_intent_service_spec.rb` + `spec/services/charge/` — 75 examples, 1 pre-existing failure (also red on unmodified main)
- rubocop clean on all six touched files
- Post-deploy: verify the flag-off fallback with a live checkout before the first flag enable (expect `fallback_reason: :feature_disabled` in the charge log line)

## Before/After

Not a UI change at flag-off (no seller has `buyer_currency_charging`; checkout is byte-identical until the ramp). The flag-on presentment UI shipped and was QA'd in #5781/#5788 with preview-app evidence.

---

AI disclosure: written with Claude Opus 4.8 (via Hermes Agent). Prompt: with all buyer-currency safety gates shipped (#5779, #5846, quote re-lock) and the production rollout approved on the private tracking issue, lift the Stripe test-mode gate on the buyer-currency card presentment path only — keep the iDEAL/Bancontact gate in place — verify the change is a no-op while the `buyer_currency_charging` flag is off, and update the affected specs.

### DIFF
--- a/app/presenters/checkout/stripe_payment_presenter.rb
+++ b/app/presenters/checkout/stripe_payment_presenter.rb
@@ -255,9 +255,10 @@ class Checkout::StripePaymentPresenter
         #   2. The buyer-currency card shape (single USD-priced one-time item — the same cart
         #      shape the eligibility service's card mode accepts), which mounts the
         #      server-confirm Payment Element in the buyer's quote currency (see props).
-        # Live mode and non-flagged sellers never produce a candidate
-        # (buyer_presentment_candidate? checks both), so neither branch changes production
-        # behavior.
+        # Non-flagged sellers never produce a candidate (buyer_presentment_candidate? checks
+        # the seller flags), so neither branch changes behavior for unflagged checkouts. The
+        # card shape (2) runs in live mode since the production rollout; the method-forced
+        # shape (1) is still test-mode only (method_forced_qa_shape? checks the Stripe key).
         supported = (method_forced_qa_shape?(items) && client_confirm_eligible?) ||
           buyer_currency_presentment_element_shape?(items)
         return "buyer_currency_presentment_unsupported" unless supported
@@ -269,7 +270,7 @@ class Checkout::StripePaymentPresenter
     # The cart shape whose buyer-currency presentment the CARD charge path supports, mirroring
     # the gates of Checkout::BuyerCurrencyEligibility#decision that are knowable at render time:
     # a single one-time, USD-priced, non-commission item from a presentment-candidate seller
-    # (candidate? already covers test mode, the seller's flags, and an active buyer-local
+    # (candidate? already covers the seller's flags and an active buyer-local
     # display). Products that offer installments stay on CardElement even when the buyer chooses
     # a one-time purchase because quote creation cannot see that choice and rejects the product.
     # Charge-time-only gates (merchant account model, wallet params, GeoIP re-check, quote
--- a/app/services/checkout/buyer_currency_eligibility.rb
+++ b/app/services/checkout/buyer_currency_eligibility.rb
@@ -53,8 +53,7 @@ class Checkout::BuyerCurrencyEligibility
   end
 
   def self.buyer_presentment_candidate?(seller:, buyer_currency_display:)
-    stripe_test_mode? &&
-      seller_enabled?(seller) &&
+    seller_enabled?(seller) &&
       buyer_presentment_display?(buyer_currency_display)
   end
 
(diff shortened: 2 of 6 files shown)


gumclaw:
## Fable-5 adversarial review — READY-TO-MERGE

Adversarial review of the gate-lift diff (money-safety hunt list for buyer-presentment currency changes), verdict at head `bcdbaef`.

**No P1s.** Findings and their dispositions:

|---|---|---|
| Stripe **FX Quotes preview enrollment** must be verified on the live account before the flag ramp — unenrolled means an error-notification flood and a silently inert feature | P2 | Not a code change. Recorded as a **hard pre-ramp gate** on the rollout checklist in gp#933; no flag write happens before an empirical live-account check. |
| Lifting the gate changes checkout surface for a flag-on seller's **entire buyer-local catalog** in live — unsupported shapes (recurring, multi-item, non-USD) downgrade Payment Element → CardElement with wallets disabled | P2 | Pinned by spec in `bcdbaef` ("falls back to CardElement for a flag-on seller's unsupported presentment shape in live mode"). Verified red/green: fails with the code hunk reverse-applied, passes with it. |
| No rack-attack throttle on `/customer_surcharge` | P3 | Pre-existing, unchanged by this diff; follow-up candidate. |
| Dispute evidence renders USD amounts for presentment charges | P3 | Pre-existing PR-1 behavior; unchanged here. |
| Refund-guard bound is looser than it could be | P3 | Pre-existing; unchanged here. |

**Confirmed safe (adversarial claims that passed):**
- Refund / dispute / balance paths are already presentment-aware — the gate lift adds no new money path, it only widens who reaches the existing one.
- The iDEAL/Bancontact method-forced lane keeps its test-mode gate; live it stays closed behind 4 independent gates (verified each in code, not the PR description).
- Stale quote tokens fail closed across all charge lanes.
- Everything remains flag-gated per seller (`buyer_currency_charging` is off in production — verified via audited console read, 0 actors / 0%).

**Local verification:** presenter spec file 64/64 green; eligibility + quote + resolver suites green locally (2 known failures reproduce identically on unmodified main — stash-baseline verified); rubocop clean on all touched files.

**Only live-verifiable:** FX Quotes enrollment behavior on the production Stripe account, and real presentment element mounting with live keys — both are pre-ramp gates on gp#933, not merge blockers for this code-only gate lift.

Merging per Sahil's rollout approval on gp#933 once CI is green on `bcdbaef`.

PR #5955, opened 2026-07-18
Fix cart item title collapsing on mobile for free-trial memberships

gumclaw:
## What

Fixes the checkout cart layout on mobile where a product title collapses to one letter per line when the item is a free-trial membership.

Three changes, all scoped to the checkout cart:
- `CartItemEnd` at the checkout call site gets `max-w-1/2 text-right`, so a long price string like `US$52.45 monthly after` wraps onto a second line instead of claiming the whole row.
- `CartItemMain` at the checkout call site gets `min-w-2/5`, a real 40% floor for the title/tier column (a cap on the price column alone isn't enough — with basis 0 the title column still absorbs any remaining squeeze).
- The shared `CartItemMain` component gets `min-w-0` so `CartItemTitle`'s `line-clamp-2` can engage when the column shrinks.

## Why

The cart item row is `flex flex-row` with `CartItemMain` at `flex-1` (basis 0) and `CartItemEnd` at the default `flex: 0 1 auto` (basis = content width). On a ~360px viewport, a free-trial membership's `CartItemEnd` content ("one month free" + "US$52.45 monthly after") has a one-line width of ~200px, so it wins the space negotiation and the title column ends up a few pixels wide, wrapping letter-by-letter. Verified live at 360px (see Before below) and confirmed via computed-style inspection that the title column was 44px wide out of a 326px row.

Fixes #5947

## Test plan

- `prettier` + `eslint` clean on both touched files
- Verified `max-w-1/2` and `min-w-2/5` compile under the repo's Tailwind v4 (fraction width utilities; confirmed via a standalone `tailwindcss` CLI build against the repo's node_modules)
- Measured the flex layout before/after with Playwright computed-style inspection at 360px: title column 44px → ~130px; price column capped at 50% and wrapping
- Fable-5 adversarial review: PASS, no P1/P2 (see PR comment) — includes checking the other `CartItemMain`/`CartItemEnd` consumers (product page, bundle editor, workflows, reviews) are unaffected
- Only the checkout call site gets the cap/floor; `min-w-0` on the shared component only allows shrinking (it can't force any layout change on consumers whose end column is narrow)

## Before/After

360px viewport, free-trial membership in the cart ("Monthly Signals Subscription", US$0 today + monthly price after a free month):

| Before | After |
|---|---|
| before | after |

Before is the live checkout at 360px. After is the same live checkout DOM with exactly the declarations these utilities compile to (`max-width: calc(1/2 * 100%)`, `text-align: right`, `min-width: calc(2/5 * 100%)`) applied to the cart row — byte-identical CSS to what the PR ships, verified against the repo's Tailwind build. Preview app is deploying for an on-branch verification pass as well.

---

AI disclosure: written with Claude Opus 4.8 (via Hermes Agent). Prompt: investigate issue #5947 (cart item title collapsing to one letter per line on mobile for free-trial memberships), find the root cause in the checkout cart flex layout, ship a minimal checkout-scoped fix, and verify before/after at a 360px viewport with computed-style inspection.


gumclaw:
## Fable-5 adversarial pre-merge review

**VERDICT: PASS**

No P1 or P2 findings. I worked through the hunt list against the actual code; every item checks out. Evidence per item:

**1. Other CartItemMain consumers — no regressions.** Consumers are Checkout (main row at `Checkout/index.tsx:626` and the accepted-offer bundle sub-list at `:600`), `Product/index.tsx:483`, `BundleEdit/ContentTab/BundleProductItem.tsx:46`, `BundleEdit/ContentTab/BundleProductSelector.tsx:43`, `Reviews/Index.tsx:192`, and `WorkflowEmails.tsx:645`. `min-w-0` only lowers the shrink *floor* — it changes nothing unless the row is over-constrained, and in that case the old behavior was the row overflowing its container. In every consumer, CartItemMain's children are line-clamped titles (`line-clamp-1`/`line-clamp-2`) or space-separated text that wraps, so allowing shrink produces clamping/wrapping, not clipping of anything critical. The Checkbox in BundleProductSelector and the Configure/Remove controls in BundleProductItem live in CartItemEnd, not CartItemMain, so they keep their content-based min-width. If anything, `min-w-0` improves the bundle rows (long product names now clamp instead of blowing out the row).

**2. `max-w-1/2` is a cap, not a basis.** CartItemEnd (`CartItemList.tsx:126`) has no flex utilities, so it stays at the default `flex: 0 1 auto` with `flex-basis: auto` (content width). `max-width` only clamps when content exceeds 50%; a short "US$9" price renders exactly as before. Percentage resolves against the row section's content box (`flex flex-row gap-4 p-4` at `CartItemList.tsx:42`), which is well-defined.

**3. `text-right` is both needed and safe.** `items-end` on CartItemEnd aligns each child span's *box* to the right, but each span box is content-width; once the "US$99.99 monthly after" span wraps under the new cap, its second line would be left-ragged inside that box without `text-align: right`. It's applied only at the checkout call site (`Checkout/index.tsx:706`), whose children are all price/recurrence text — no other call site is touched. Nit-level only: the repo uses `text-right` in 11 places and `text-end` in zero, so this matches convention.

**4. `max-w-1/2` is a real utility in this build.** The repo is on Tailwind v4.1.14 (CSS-first, no `tailwind.config.*`). I compiled a test snippet through the repo's own `node_modules/.bin/tailwindcss` and it emits `.max-w-1\/2 { max-width: calc(1/2 * 100%); }` — the class is generated, not purged, even though this is its only usage in the codebase.

**5. line-clamp-2 engages.** In the checkout row, `CartItemTitle` (which carries `line-clamp-2` + implicit `overflow: hidden`, `CartItemList.tsx:108`) is a *direct* child of the `min-w-0` CartItemMain section — no intermediate wrapper with `min-width: auto` remains in the chain. The anchor inside the title is inline, so it doesn't block shrinking.

**6. Bundle branch interaction.** `justify-center self-stretch border-l border-border p-4` (`CartItemList.tsx:87`) composes with `min-w-0` without conflict — those classes control cross-axis alignment, borders, and padding; none re-establishes a min-width.

**7. No branch of CartItemEnd content can be truncated.** CartItemEnd has no `overflow-hidden`, so the 50% cap causes wrapping at spaces ("in 4 installments", "one month free", recurrence labels), never clipping. For a single unbreakable token wider than 50% (a very long price), CSS min/max resolution protects it: CartItemEnd still has `min-width: auto` (no `min-w-0` there), and when min-width exceeds max-width, **min wins** — so a price can widen past the cap rather than truncate. The `aria-label="Price"` is an attribute and unaffected by layout either way.

One observation for the PR process rather than the code: this is exactly the kind of mobile CSS fix where CONTRIBUTING.md's before/after media rule applies at 360px — make sure the PR includes mobile screenshots/video showing a free-trial membership item before and after, plus a normal short-price item to demonstrate no wrap regression.

PR #5957, opened 2026-07-18
Fix invoice page 500 (Module::DelegationError) when charge has no successful purchases

gumclaw:
## What

Fixes a 500 on the invoice generation page (`/purchases/:id/invoice/new`) when the resolved chargeable is a `Charge` that has no purchases in a successful state yet. `Purchases::InvoicesController#set_chargeable` now falls back to the purchase itself as the chargeable in that window.

## Why

`Charge` delegates buyer details (`full_name`, `purchaser`, address fields) to `purchase_as_chargeable`, which is defined as `successful_purchases.first`. A buyer following the receipt link to generate an invoice while their purchase is still `in_progress` (or after every purchase on the charge failed) hits `Module::DelegationError: Charge#full_name delegated to purchase_as_chargeable.full_name, but purchase_as_chargeable is nil` and gets a 500.

Sentry: [GUMROAD-ZB](https://gumroad-to.sentry.io/issues/7618529957/) — `Purchases::InvoicesController#new`, unhandled, production. Confirmed against live data: the charge in the event now has its purchase in `successful` state, i.e. the crash happened in the transient window before the purchase transitioned — real buyer traffic, not noise.

The purchase itself carries the same buyer details and is already the chargeable used for purchases that predate the Charge model, so falling back to it renders the same form data with no behavior change for settled charges.

## Before / After

Not a UI change — the page previously returned a 500 error screen in this state; it now renders the normal invoice form. Video below shows the red → green reproduction.

**Walkthrough video** (spec reproducing the production `Module::DelegationError` without the fix, passing with it):

https://github.com/antiwork/gumroad/blob/fix/invoice-chargeable-in-progress-charge/qa-media/pr-invoice-chargeable-in-progress-walkthrough.mp4?raw=true

## Test plan

- New controller spec: `GET new` for a Charge whose purchases are all `in_progress` — fails with the exact production `Module::DelegationError` without the controller change, passes with it (red → green verified locally).
- Existing "for Charge" specs (settled charges) unchanged and unaffected — fallback only engages when `successful_purchases.none?`.
- Note on the full spec file: `spec/controllers/purchases/invoices_controller_spec.rb` run as a whole currently fails on unmodified `main` too (VCR/`create(:purchase)` card-charging setup failures in this worktree, pre-existing) — the targeted specs above run clean.

## Test results

```
$ bundle exec rspec spec/controllers/purchases/invoices_controller_spec.rb -e "falls back to the purchase as the chargeable"
1 example, 0 failures
```

Without the fix (fix stashed):

```
Module::DelegationError:
  Charge#full_name delegated to purchase_as_chargeable.full_name, but purchase_as_chargeable is nil
1 example, 1 failure
```

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent / gumclaw), triggered automatically by the Sentry webhook for GUMROAD-ZB. Instructions given: triage the Sentry event, root-cause from the stack trace, write a failing test reproducing the error, fix it, and open a PR. Tools used: Sentry API (issue/event lookup), audited read-only prod console (confirmed the charge's purchase state), rspec, rubocop.


gumclaw:

- The fallback condition is `@chargeable.is_a?(Charge) && @chargeable.successful_purchases.none?` — it only engages in the exact failure window. Settled charges (the normal case) are untouched, including multi-item charges where the Charge chargeable is genuinely needed to aggregate purchases.
- Falling back to `@purchase` is the same chargeable shape used for all pre-Charge purchases and for purchases where `uses_charge_receipt?` is false, so every downstream call in `#new` and `#create` (`InvoicePresenter`, `refund_gumroad_taxes!`, `upload_invoice_pdf`) already supports it.
- In the fallback state `#create`'s VAT path is additionally protected by the existing `!@purchase.successful?` early-return at the top of the action.
- `successful_purchases` on Charge is `purchases.all_success_states_including_test` — includes `not_charged`, `gift_receiver_purchase_successful`, and test states, so the fallback does NOT trigger for free-trial/gift/test purchases that settled without a charge-success state.

gumclaw:
**Fable-5 adversarial pre-merge review: APPROVE** — no P1/P2 findings.

Verified (read against the actual code, not just the diff):
- All `#create` downstream calls are safe with the Purchase fallback; the only state-sensitive path (`refund_gumroad_taxes!`) is unreachable when the fallback engages, because the VAT guard at the top of `#create` redirects when the purchase isn't successful.
- No wrong-totals risk for settled charges: `successful_purchases` covers all success states including `not_charged`/gift/test, and refunds/chargebacks don't change `purchase_state`, so any charge that ever settled has ≥1 success-state purchase and the fallback can't engage on it.
- The new spec genuinely pins the bug — without the controller change it raises the exact production `Module::DelegationError` at `form_info.rb:58`.
- No security change: `require_email_confirmation` runs before `set_chargeable`, and the fallback strictly narrows exposed data (buyer's own purchase vs the whole charge).

P3 findings (follow-up OK, not blocking):
1. `POST create` has no spec with the fallback engaged. Behavior there changes from "500 → generic error alert" to "invoice PDF generated for a not-yet-successful purchase" — which is parity with the long-standing standalone-purchase path, but a spec would pin it.
2. The fallback also engages when every purchase on the charge failed permanently, meaning a buyer whose payment never settled can generate an invoice document. This matches existing standalone-purchase behavior (this controller has never gated invoice generation on purchase success outside the VAT path), so it is not a regression — flagging in case the business ever wants to block invoices for failed payments.

PR #5958, opened 2026-07-18
Add bulk admin endpoint to refund all of a seller's purchases for fraud

gumclaw:
Adds a bulk admin endpoint so a fraud seller's entire sales history can be refunded in one call instead of one API call per purchase.

## Why

Tonight's fraud case (n3xo) required refunding all 18 purchases of a suspended seller. The only admin API surface was the per-purchase `POST /internal/admin/purchases/:id/refund_for_fraud`, which also needs the purchase's `external_id_numeric` and buyer email — so the workflow was 17 `purchases search` round-trips to resolve ids, then 18 individual refund calls, then a manual verification pass. Sahil asked to improve that flow so it never repeats.

## What

`POST /api/internal/admin/users/refund_all_for_fraud` — **validate-and-enqueue** (restructured per the [security review](https://github.com/antiwork/gumroad/pull/5958#issuecomment-5010358000); the original shape ran the refund loop synchronously in the web request).

Params: `user_id` (required), `expected_email` (required), `expected_purchase_count` (required), `block_buyers` (optional, default false).

- **Confirmation handshake (authz parity with `refund_balance`):** `expected_email` and `expected_purchase_count` are required; a mismatch with the current account state returns **409** with the current count so the caller must re-verify. The `force` param is gone entirely — the seller must already be suspended (suspension is one call away in this same controller), and *flagged* is no longer sufficient.
- **Async with per-purchase retries:** on success the endpoint enqueues `RefundAllForFraudWorker` (`lock: :until_executed`, locked by user id — a duplicate enqueue returns **409**), which fans out one `RefundPurchaseForFraudWorker` per purchase with `retry: 5`. No request timeouts, no overlapping runs, Stripe rate limiting absorbed by Sidekiq backoff.
- **Durable record:** the fan-out worker leaves a `Comment` on the seller (`refund_all_for_fraud` type) with the queued count and the buyer-blocking decision, mirroring `RefundUnpaidPurchasesWorker`. Per-purchase failures go through `ErrorNotifier` with metadata and re-raise for retry.
- **Buyer blocking is opt-in:** by default purchases are refunded for fraud **without** blocking buyers — the seller-fraud case, where buyers are victims (same reasoning as `RefundUnpaidPurchasesWorker`). Pass `block_buyers: true` only for buyer-fraud cases (self-purchase / card-testing rings); that switches to `refund_for_fraud_and_block_buyer!`, which blocks each buyer's email, card fingerprint, browser GUID, and IP platform-wide. `block_buyers` parses strictly (`"true"`/`"1"` only — `"no"` is false) and is captured in the audit `params_snapshot`.
- **Scope fix:** uses `successful.not_fully_refunded.not_chargedback_or_chargedback_reversed`, so chargeback-reversed purchases (which are refundable) are included instead of silently dropped.
- **Already-refunded race handled:** `refund_for_fraud!` / `refund_for_fraud_and_block_buyer!` gain a `skip_already_refunded:` flag; the `nil` return from `refund_and_save!` (already refunded) is treated as a clean skip — no subscription cancellation, no creator email, no buyer block, no counting.
- **Audited:** wrapped in `record_admin_write(action: "users.refund_all_for_fraud", target: user)`.

Response shape (200):

```json
{
  "success": true,
  "user_id": "…",
  "status": "queued",
  "message": "Refund all for fraud queued",
  "purchases_to_refund": 18,
  "block_buyers": false
}
```

Error responses: 400 missing params, 404 unknown user, 422 not suspended, 409 (expected_email mismatch / expected_purchase_count mismatch with `current.purchases_to_refund` / run already queued).

## Progress

- [x] Validate-and-enqueue restructure: uniqueness-locked `RefundAllForFraudWorker` + per-purchase `RefundPurchaseForFraudWorker` (retry: 5), mirroring the `refund_balance` → `RefundUnpaidPurchasesWorker` → `RefundPurchaseWorker` pattern
- [x] Required `expected_email` + `expected_purchase_count` handshake, 409 on mismatch; `force` dropped; suspended-only guard
- [x] `block_buyers` opt-in (default: refund without blocking; buyers are victims in the seller-fraud case)
- [x] `not_chargedback_or_chargedback_reversed` scope; nil already-refunded race handled as clean skip
- [x] `ErrorNotifier` on per-purchase failures with metadata; seller `Comment` with queued count
- [x] Specs: handshake 409s, uniqueness-lock 409, block_buyers switch + strict boolean parsing, `params_snapshot` audits `block_buyers`, unstubbed already-refunded nil path, scope inclusion/exclusion — controller spec 394 examples 0 failures, worker + refund specs 183 examples 0 failures
- [x] Rubocop clean on touched files

## Companion PR

CLI command `gumroad admin users refund-all-for-fraud` lives in antiwork/gumroad-cli#187, updated to this contract (required `--expected-email`/`--expected-count`, opt-in `--block-buyers`, queued-response handling), and depends on this endpoint deploying first.

## Test results

```
spec/controllers/api/internal/admin/users_controller_spec.rb — 394 examples, 0 failures
spec/models/purchase/purchase_refunds_spec.rb + spec/sidekiq/{refund_purchase,mass_refund_for_fraud,refund_unpaid_purchases,refund_all_for_fraud,refund_purchase_for_fraud}*_spec.rb + spec/models/comment_spec.rb — 183 examples, 0 failures
rubocop — no offenses on touched files
```


gumclaw:
## Fable-5 pre-merge review — **needs restructuring before merge** (2 blockers)

**Blocker 1 — must be a background job.** Each refund iteration does a synchronous Stripe call + row-locked txn + per-purchase analytics cache rebuild + Elasticsearch reindex + mailers + ~8 blocked-object writes (~1–3s each); rack timeout is 120s, so the request dies around purchase 50–100, mid-batch, and `Rack::Timeout`'s exception isn't caught by the per-purchase rescue. The codebase already has the right pattern one method up in this same controller: `refund_balance` → `RefundUnpaidPurchasesWorker` (uniqueness-locked, 409 on duplicate) → one `RefundPurchaseWorker` per purchase with retries. Recommended: keep endpoint + guard, validate-and-enqueue, return `status: "queued"`, leave a seller `Comment` with durable counts.

**Blocker 2 — buyer-blocking is the wrong default for seller fraud.** `refund_for_fraud_and_block_buyer!` is a *buyer*-fraud primitive: it platform-blocks each buyer's email, card fingerprint, and IP for months. Every existing caller punishes fraudulent buyers. But this endpoint targets *seller* fraud — where buyers are usually victims (n3xo: 16 of 18 were real people). The sibling seller-unwind flow (`RefundUnpaidPurchasesWorker`) deliberately refunds WITHOUT blocking for exactly this reason. Run this on a TOS-suspended seller with 5,000 legit customers and you platform-ban 5,000 innocent people. Fix: make buyer-blocking an explicit opt-in param (`block_buyers: true` for card-testing-ring cases), default plain fraud refund.

**Major — authz parity.** `refund_balance` (which moves less money) requires `expected_email` + `expected_purchase_count` + `expected_total_amount_cents`, 409 on mismatch, no force override. This endpoint: all optional, `force` bypasses the only state check. Bring it to parity: require expected_email + a count-or-amount handshake; drop `force` (suspension is one call away in the same controller) or gate it hard. Also `ActiveModel::Type::Boolean.cast("no") == true` — `force=no` forces.

**Moderate/minor** (full review from the parent session, available on request): Stripe-fully-down yields `success: true` + 200 + everything in `failed` — add `status: completed|completed_with_errors`; `not_chargedback` silently excludes chargeback-*reversed* purchases (refundable — use `not_chargedback_or_chargedback_reversed`); `refund_and_save!` returns `nil` for already-refunded so a race counts a no-money-moved purchase as refunded + blocks its buyer; swallowed errors should hit `ErrorNotifier`; specs stub the refund primitive so none of these interactions are exercised.

Restructure incoming on this branch. — gumclaw (fable-5 review)

gumclaw:
Restructured per the review above (head is now edc37a3a8, force-pushed):

**#1/#3 (blockers: sync loop, no concurrency guard)** → validate-and-enqueue. The endpoint now enqueues `RefundAllForFraudWorker` (`lock: :until_executed`, locked by user id, 409 if already queued) which fans out one `RefundPurchaseForFraudWorker` per purchase with `retry: 5` — mirroring the `refund_balance` → `RefundUnpaidPurchasesWorker` → `RefundPurchaseWorker` pattern. Response is `200 {status: "queued", purchases_to_refund: N, block_buyers}`. The fan-out worker leaves a `Comment` on the seller with the queued count and blocking decision, so the durable record is no longer just the HTTP response body.

**#2 (blocker: buyer-fraud primitive applied to a seller-fraud scenario)** → buyer blocking is now explicit opt-in via `block_buyers: true` (strictly parsed: only `"true"`/`"1"` — `"no"` is false). Default refunds for fraud WITHOUT blocking, for the seller-fraud case where buyers are victims. Documented in the endpoint comment and PR body; `block_buyers` lands in the audit `params_snapshot`.

**#4 (authz)** → `expected_email` and `expected_purchase_count` are now required, 409 on mismatch with the current refundable count. `force` is dropped entirely and the guard is suspended-only (flagged no longer qualifies; suspension is one call away in this controller).

**#5/#6 (envelope, scope, nil race, observability, tests)** → `status: "queued"` dissolves the success-with-failures envelope; scope is `successful.not_fully_refunded.not_chargedback_or_chargedback_reversed`; `refund_for_fraud!`/`refund_for_fraud_and_block_buyer!` gained `skip_already_refunded:` so the nil already-refunded race is a clean skip (no blocking/counting/subscription side effects — legacy callers unchanged); per-purchase failures go through `ErrorNotifier` with metadata and re-raise for retry. Specs cover the 409 handshakes, the uniqueness-lock 409, the block_buyers switch, params_snapshot auditing, and an **unstubbed** already-refunded purchase exercising the real nil path.

Controller spec: 394 examples, 0 failures. Refund/worker/comment specs: 183 examples, 0 failures. Rubocop clean.

CLI companion antiwork/gumroad-cli#187 updated to the new contract in antiwork/gumroad-cli@9c15811.

PR #5962, opened 2026-07-19
Fail fast with a clear error when a tiny total leaves the seller no proceeds

gumclaw:
Fixes the `parameter_invalid_integer` checkout failure mode tracked in antiwork/gumroad#5419 scope (evidence and investigation on the private tracker: [failure-mode comment](https://github.com/antiwork/gumroad-private/issues/933#issuecomment-5012899850), [root-cause analysis](https://github.com/antiwork/gumroad-private/issues/933#issuecomment-5013549218)).

## What

Add a shared seller-proceeds guard (`StripeIntentChargeRouting.validate_seller_proceeds!`) that runs before submitting to Stripe on every Stripe checkout path — both branches of `StripeChargeProcessor#create_payment_intent_or_charge!` and the Payment Element client-confirm route, which builds a deferred PaymentIntent through `StripeIntentChargeRouting` and could previously still send `transfer_data[amount] <= 0` to Stripe and surface the generic prepare error.

The guard enforces the two constraints Stripe actually imposes, which differ by charge type:

- **Destination charges** (non-migrated sellers, `transfer_data[amount]`): the seller transfer must be at least 1 subunit. Zero or negative proceeds are rejected by Stripe with `parameter_invalid_integer`, so we fail fast instead.
- **Direct charges on connected accounts** (`application_fee_amount`): Stripe allows the application fee to exactly equal the payment total (zero seller proceeds is valid — Stripe caps the fee at the payment), so we allow fee == payment and reject only a fee greater than the payment.

On violation we raise `ChargeProcessorCardError` with the existing `PurchaseErrorCode::NET_NEGATIVE_SELLER_REVENUE` code and a clear buyer-facing message ("The purchase total is too small for us to process…") instead of the generic retry prompt, and report each occurrence via `ErrorNotifier` with the amounts and currency so we can size a proper per-currency checkout minimum as a follow-up. The guard runs on the buyer-presentment processor amounts when present (the failing population), falling back to the USD canonical amounts otherwise.

For subscription rebills, `net_negative_seller_revenue` is now terminal rather than retried after six hours — the total is deterministic and unchanged, so a retry could never make the charge valid.

## Why

On very small totals (observed as 53–146 subunits in eur/gbp/aud/cad/nzd — ≈130 failed purchases across ~50 buyers in 3 days), Gumroad's fee (fixed 50¢ + 30¢ floor plus percentage legs) meets or exceeds the whole charge amount. For non-migrated sellers we then build a destination charge with

```ruby
transfer_data: { amount: charge_amount_cents - charge_gumroad_amount_cents } # zero or negative
```

Stripe deterministically rejects this with `parameter_invalid_integer` on `transfer_data[amount]` **before the card attaches**, and the buyer sees a generic "There is a temporary problem, please try again" — a dead end they retry into repeatedly. The invalid state occurs because the fee floor is fixed while buyer-presentment totals can be arbitrarily small; nothing upstream prevents such a cart from reaching the pay button.

Reproduced against the Stripe test API with the exact failing shape (amount=85 gbp, negative transfer amount):

```json
{
  "code": "parameter_invalid_integer",
  "param": "transfer_data[amount]",
  "message": "This value must be greater than or equal to 1."
}
```

Zero also fails identically (`must be greater than or equal to 1`).

Why this approach over alternatives: a checkout-UI minimum is the durable fix but needs per-currency thresholds and design input (tracked as a follow-up); this server-side invariant stops the buyer-facing dead end now and gives us `ErrorNotifier` data to size the minimums. For destination transfers and for direct charges with a fee exceeding the payment, the guard rejects only what Stripe would reject anyway, so nothing processable is lost; the direct-charge fee-equals-payment case is deliberately allowed because Stripe accepts it.

## Before/After

Not a UI-rendering change — checkout is visually unchanged. Before: affected buyers saw the generic "There is a temporary problem, please try again" and retried into it repeatedly. After: they get the explicit "The purchase total is too small for us to process. Please add another item to your order or contact the creator."

**Walkthrough video** (the guard in `StripeChargeProcessor`, both call sites ahead of `Stripe::PaymentIntent.create`, and the new specs running green locally): [qa-media/pr-5962-fail-fast-walkthrough.mp4](https://raw.githubusercontent.com/antiwork/gumroad/fix-negative-seller-transfer-small-totals/qa-media/pr-5962-fail-fast-walkthrough.mp4)

## Test Results

Four new specs (negative transfer, exactly-zero transfer, presentment-amount precedence, and the connected-account direct-charge path) verifying Stripe is never called, the error code is `net_negative_seller_revenue`, and ErrorNotifier receives the amounts — green locally:

local spec results

Additionally, an end-to-end Capybara system spec (`type: :system, js: true`, in `spec/requests/purchases/product/payment_errors_spec.rb`) drives a real browser checkout on a product whose fee meets or exceeds the total and asserts the buyer sees the explicit "The purchase total is too small for us to process…" alert, the purchase fails with `net_negative_seller_revenue`, and no Stripe transaction is created — green locally (1 example, 0 failures, 74s).

## QA steps

This is a server-side guard on the live-mode Stripe submit path with no UI surface; the failing shape (fee ≥ total) is exercised end-to-end by the VCR-backed specs above. To reproduce manually on a preview app: create a product priced so the buyer total lands under the fee floor for a non-migrated seller (e.g. £0.85 with default fees) and attempt checkout — the buyer now sees the explicit "total is too small" error instead of the generic retry prompt.

Follow-up (separate PR): enforce a per-currency checkout minimum in the checkout UI/validation so buyers see the constraint before entering payment details.

---

**AI disclosure:** Written by Claude Fable 5 (Hermes agent "Gumclaw"). Prompts: investigate the `parameter_invalid_integer` small-amount checkout failures from the gumroad-private#933 multi-currency ramp; confirm the root cause against the Stripe test API; add a fail-fast guard with a clear buyer-facing error on both processor paths plus ErrorNotifier reporting; cover with specs for negative, zero, and presentment-amount cases.


rmarescu (commented):
@gumclaw follow CONTRIBUTING guidelines.

gumclaw:
Self-review notes:

- The guard lives in `validate_seller_proceeds!` and is called on **both** processor branches (connected-account `application_fee_amount` and destination-charge `transfer_data`) immediately before `Stripe::PaymentIntent.create`, so tiny totals fail identically regardless of the seller's migration state.
- Threshold is `>= 1` (not `> 0`) to match Stripe's documented constraint that `transfer_data[amount]` must be ≥ 1 — an exactly-zero transfer is rejected by Stripe the same way as a negative one, and the zero case has its own spec.
- The guard prefers buyer-presentment processor amounts when present because that's the population that actually failed in production (eur/gbp/aud/cad/nzd subunit totals); USD canonical amounts are the fallback so the invariant also holds outside the multi-currency ramp.
- Reused `PurchaseErrorCode::NET_NEGATIVE_SELLER_REVENUE` rather than minting a new error code — it describes exactly this state and already flows through the buyer-facing error rendering.
- `ErrorNotifier` payload includes charge/gumroad/seller amounts + currency so the follow-up per-currency checkout minimum can be sized from real data.

Addressed the CONTRIBUTING gaps from the review: PR body restructured to What/Why/Before-After/Test Results, walkthrough video + local spec results added under `qa-media/` (adc02ac03), QA steps and AI disclosure added.

gumclaw:
@claude review once

rmarescu:
@gumclaw can you do e2e testing?

gumclaw:
@rmarescu Yes — on it. Plan:

1. Added the `preview` label to spin up a preview app for this branch (same flow as #5923's round-3 QA).
2. Once the preview is up, I'll run the e2e pass end-to-end through checkout:
   - **Failing shape reproduced**: tiny non-USD total (the £0.85 case from the buyer ticket) → expect the new clear buyer-facing error *before* any `Stripe::PaymentIntent.create` call, on **both** the connected-account and destination-charge branches.
   - **No-regression**: normal-sized USD and non-USD purchases complete successfully.
   - **Boundary**: smallest total where seller proceeds are positive still charges.
3. Evidence (screenshots + console/log tie-out) posted here, then back to you for re-review.

Taking this back to `working` in the meantime.

gumclaw:
Added an end-to-end system spec in 017ecf8: a real Capybara browser checkout (`type: :system, js: true`) in `spec/requests/purchases/product/payment_errors_spec.rb` that builds the failing shape with real objects (seller on a 100% fee with a $0.99 product and their own non-migrated merchant account, so the charge takes the destination-charge path), goes through `add_to_cart` → `check_out`, and asserts:

- the buyer sees the explicit "The purchase total is too small for us to process. Please add another item to your order or contact the creator." alert (via `check_out`'s `error:` assertion, which also verifies the purchase lands in `failed`, not `successful`)
- `Purchase.last.stripe_error_code == "net_negative_seller_revenue"`
- `stripe_transaction_id` is nil — the guard raised before anything was submitted to Stripe

Green locally: 1 example, 0 failures (74s). PR body's Test Results section updated to mention it.

gumclaw:
## E2E testing — done

**1. Browser e2e of the failing shape (automated, real browser):** added in 017ecf82c — a Capybara system spec (`type: :system, js: true`) in `spec/requests/purchases/product/payment_errors_spec.rb` that drives a full checkout of the exact failing shape with real objects: seller on a 100% custom fee (`custom_fee_per_thousand: 1000`) with their own non-migrated merchant account and a $0.99 product, so the would-be seller transfer is non-positive on the destination-charge path. It asserts:
- the buyer sees the new clear error ("The purchase total is too small for us to process. Please add another item to your order or contact the creator.") instead of the generic retry prompt,
- the purchase fails with `stripe_error_code = NET_NEGATIVE_SELLER_REVENUE`,
- no PaymentIntent is ever submitted (`stripe_transaction_id` nil) — the guard fires before Stripe.

**2. No-regression coverage:** the full system-spec suite (all the normal-purchase, decline, multi-item checkout specs) ran green alongside it — CI is fully green on head 3b30fc4af: [ci/green run 29675415852](https://github.com/antiwork/gumroad/actions/runs/29675415852), buildkite/gumroad SUCCESS.

**3. Preview app for hands-on click-through:** live at https://fix-negative-seller-transfer-sma.apps.staging.gumroad.org (verified serving, `x-revision 3b30fc4afb7a` = this PR's head) if you want to poke at it manually.

Handing back to you for review — nothing agent-side remains.

gumclaw:
Sentry confirmation of the guard firing end-to-end: [GUMROAD-ZK](https://gumroad-to.sentry.io/issues/7620384812/) — `Charge rejected before Stripe submit: seller proceeds would be non-positive`, `OrdersController#create`, environment `staging`, URL `fix-negative-seller-transfer-sma.apps.staging.gumroad.org/orders` (this PR's preview app, 13:26 UTC today).

This is the intentional `capture_message` escalation added in this PR, triggered during the e2e/preview QA — preview-originated, not a prod incident. Positive evidence the rejection path reports to Sentry as designed. Resolve GUMROAD-ZK after this merges so the seeded QA noise doesn't linger in the queue.

rmarescu (approved):
Fixed two confirmed code issues locally in separate commits:

- `b0da1f1a7` allows a Stripe Connect direct charge when Gumroad's application fee exactly equals the payment. Stripe caps that fee at the payment total, so zero seller proceeds is valid there; we still reject a fee greater than the payment. Destination transfers continue to require at least one subunit.
- `7f97f0529` shares the seller-proceeds validation with Payment Element client-confirm checkout. That route had been building a deferred PaymentIntent through `StripeIntentChargeRouting` and could still send `transfer_data[amount] <= 0` to Stripe, then replace the intended buyer message with the generic prepare error.

Before merging, please also update the PR body to distinguish those two Stripe constraints. The current “Stripe would reject the request anyway” wording is true for destination transfers and negative direct-charge proceeds, but not for an application fee equal to a direct charge.

For subscription rebills, `net_negative_seller_revenue` is now terminal rather than retried after six hours. That makes sense for an unchanged deterministic total: a retry would not make the charge valid.

gumclaw:
Updated the PR body per your review: the What/Why sections now distinguish the two Stripe constraints — destination transfers need ≥ 1 subunit, while a direct-charge application fee may equal the payment (Stripe caps the fee there) and only a fee greater than the payment is rejected. Also documented the shared guard on the Payment Element client-confirm route and the terminal (no-retry) handling of `net_negative_seller_revenue` for subscription rebills. Thanks for the two fixes — leaving the merge to you.

gumclaw:
Thanks for the review, the two local fixes (b0da1f1a7, 7f97f0529), and the terminal-rebill call — all make sense. Auto-merge (squash) is enabled; this lands as soon as CI finishes on your merge commit. Once merged I'll check the #933 Gate A box and keep the ErrorNotifier data flowing toward sizing the per-currency checkout minimum follow-up.

PR #5968, opened 2026-07-19
Fix content editor dropping file embeds after drag-and-drop reorder

gumclaw:
## What

Fixes the membership tier / product content editor silently dropping file embeds after a drag-and-drop reorder, which then deleted the corresponding `ProductFile`s on save.

- `resolveCopiedFileEmbeds` (new module, extracted from the inline logic in `ContentTab/index.tsx`) now keeps any `file-embed[url]` whose `id` already resolves in the product's `filesById`, stripping only the transient `url` marker. Only genuinely foreign embeds (cross-product paste) are remapped against `existingFiles` or removed.
- Unit tests cover the drag case, the fresh-upload case, cross-product paste remapping, unknown-file removal, embeds without a `url`, and duplicate paste dedupe.

## Why

Sellers reported that adding a 4th PDF to a membership tier's content silently removed one of the existing PDFs (production data showed same-second paired ProductFile deletions right after adds). Root cause chain:

1. `FileEmbed`'s `transformCopied` ProseMirror plugin stamps copied file embeds with a `url` attribute so cross-product paste can remap them onto the seller's files.
2. ProseMirror also runs `transformCopied` on **drag** (`handlers.dragstart` → `serializeForClipboard` in prosemirror-view 1.34.3), so reordering a file embed inside the same editor stamps it with `url` too.
3. On the next editor update, the handler in `ContentTab/index.tsx` treated every `file-embed[url]` as pasted from another product and matched it against `existingFiles` — a limited snapshot fetched at page load. Files uploaded during the current editing session aren't in that snapshot (and their transient blob/download URLs never match), so the embed was silently `remove()`d.
4. `saveProduct` drops `product.files` entries with no remaining embed, and server-side `save_files!` marks the omitted `ProductFile`s deleted — the reported data loss.

Fixes antiwork/gumroad-private#1164

## Before/After

Before: dragging a freshly-uploaded file embed to reorder it removed the embed on the next editor update; saving then deleted the file from the product.

After: dragged embeds are kept regardless of whether they appear in the `existingFiles` snapshot; only genuine cross-product pastes are remapped or removed.

Walkthrough (repro of the old behavior vs the fix, plus the unit test run):

https://github.com/antiwork/gumroad/blob/fix/content-editor-dragged-file-embed-drop/qa-media/pr-pending-content-editor-dragged-file-embed-walkthrough.mp4?raw=true

## Test plan

- `npx vitest run app/javascript/components/ProductEdit/ContentTab/resolveCopiedFileEmbeds.test.ts` — 6 tests, all passing locally.
- `npx tsc --noEmit` — no errors in the touched files.
- `prettier` + `eslint` clean on the touched files.

## QA steps

Preview app: deploys via the `preview` label — once up at `https://gumroad-<pr>.apps.staging.gumroad.org`, follow this click-path:

1. Log in as a seller and create (or open) a membership product.
2. Go to the Content tab for a tier, upload 3 PDFs, and save.
3. Upload a 4th PDF (do not save yet).
4. Drag the 4th PDF's row to a different position in the content list.
5. Save, then reload the page.
6. Expected: all 4 PDFs still present in the tier content and under the product's files. Before this fix, the dragged fresh upload disappeared from the content on the next editor update and its file was deleted on save.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent) from the instruction "Replicate and fix" on the linked issue. The agent diagnosed the root cause by reading the editor code and prosemirror-view's drag handling, wrote the fix and tests, and produced the walkthrough recording.


gumclaw:

- The core change is a guard at the top of the loop: if the embed's `id` is already in `filesById` (this product's files, including files uploaded in the current session), the `url` marker is stripped and the embed is left alone. Everything after that guard is the pre-existing cross-product paste logic, unchanged except for a dedupe on `newFiles` (the old code could push the same file twice when pasted twice).
- `fragment` is typed structurally (`querySelectorAll` returning `Iterable<Element>`) rather than as `DocumentFragment` so the function is directly unit-testable against a plain DOM fragment without pulling the Tiptap serializer into the test.
- The removed inline block in `index.tsx` used `existingFiles.find(...)` first even for local files — that's exactly the path that deleted fresh uploads, since the snapshot is fetched once at page load and capped by a redis limit server-side (`ProductPresenter#existing_files`).

gumclaw:
**Pre-merge adversarial review (fable-5) — verdict: PASS, no P1/P2.**

Hunt list and results:

| # | Checked | Result |
|---|---|---|
| 1 | Dropped `updateProduct` reorder side effect for dragged embeds of snapshot-known files | P3, accepted — `product.files` order doesn't drive content order or the save payload semantics (`saveProduct` filters by embed id set) |
| 2 | Transient duplicate embeds mid-drag confusing the handler | Not possible — ProseMirror applies the move (insert+delete) in one transaction before `update` fires |
| 3 | Stale `filesById` closure in `updateContentRef` | Safe — wrapped in `useRefToLatest`, always fresh |
| 4 | Foreign pasted embed whose id collides with a local file id | External ids are unique; on a hypothetical collision keeping the local file is correct |
| 5 | Other consumers of the `file-embed[url]` marker | None — `resolveCopiedFileEmbeds` is the single consumer; `transformCopied` in FileEmbed.tsx is the single producer |
| 6 | Regression surface | Full JS suite: 238/238 passing; `tsc --noEmit` clean |

PR #5970, opened 2026-07-19
Failed-refund exception alerts: Sentry-only, drop email + dead schedule entry

gumclaw:
## What

Makes failed-refund-exception reporting **Sentry-only — no email at all** (Sahil, 2026-07-19: "These should be sentry 500s if anything? Or no email needed at all").

Context: gumroad-private#1173 — the 2026-07-18 backfill created 1,975 exception rows that crossed their 24h SLA in the same hour on Jul 19, and the per-row email design sent ~2,000 messages to finance@. These are operational error conditions, not correspondence: Sentry groups repeated events into one issue with a count, which is exactly the right shape for both the 1/day trickle and a bulk crossing.

## Changes
- [x] `NotifyFailedRefundExceptionJob`: drops `InternalNotificationMailer` entirely; `ErrorNotifier.notify` (Sentry) is now the primary and only channel, with the failure-count/retry semantics email used to have (raise → Sidekiq retry → cap → dispatcher escalation). Context enriched (amount, currency, retained fee, purchase ids) so the Sentry event is actionable without a console lookup.
- [x] `DispatchPendingFailedRefundExceptionsJob`: escalation is Sentry-only (was Sentry + email); unrescued so a Sentry outage leaves the row pending for the next minute's retry — the signal cannot be silently lost.
- [x] Removes the `daily_flagged_and_suspended_users` schedule entry pointing at `DailyRefundQueueNotifyWorker` — that class no longer exists in the codebase; the entry was generating perpetually-failing jobs (21 dead retries purged from prod today, `uninitialized constant`, up to 24 retries each).
- [x] Specs rewritten for the new contract: never-email assertions on both jobs, Sentry-failure retry paths, cap behavior, resolved/sent guards.

## QA steps
No preview needed — behavior is alert-channel routing inside two Sidekiq jobs. Reviewer path: read the two job diffs; run the spec files below; optionally in a console run `DispatchPendingFailedRefundExceptionsJob.new.perform` with a seeded overdue `FailedRefundException` and confirm ErrorNotifier fires and `InternalNotificationMailer` is never constant-referenced (it's fully removed from both files — `grep InternalNotificationMailer app/sidekiq/*failed_refund*` returns nothing).

## qa-media
Before (the Jul 19 inbox flood — real Gmail query output from finance@) and after (spec run proving the never-email contract):

before — email flood
after — sentry-only, 15/15 green

Permanent copies: [`qa-media/`](https://github.com/antiwork/gumroad/tree/gumclaw/2026-07-19-frx-sentry-only/qa-media) on this branch.

## Test results
`bundle exec rspec spec/sidekiq/notify_failed_refund_exception_job_spec.rb spec/sidekiq/dispatch_pending_failed_refund_exceptions_job_spec.rb` → **15 examples, 0 failures**. Rubocop: 4 files, no offenses.

Backend-only — spec output is the reviewable artifact (no UI surface). Before/after: finance@ received ~2,000 emails during the Jul 19 SLA crossing; after this change the identical event produces grouped Sentry issues and zero emails.

## Deploy note
The durable rows + Sentry are the record. `notification_room`/`notification_failures` columns are retained (used for Sentry context + retry caps).

## AI disclosure
Implemented by gumclaw (Hermes agent) per Sahil's direct instruction; supersedes the in-flight digest-email approach.


gumclaw:
Supersedes the in-flight digest-email approach — Sahil ruled no email at all (Sentry-only). Auto-merges on CI green per the 2026-07-19 standing order (alert-channel bug fix, no schema, direct instruction).

PR #5971, opened 2026-07-19
Fix NoMethodError in display_price_cents when default_price_cents is nil

gumclaw:
## What

Guards `Product::Prices#display_price_cents` against a nil `default_price_cents`, falling back to `0` so listing pages render instead of raising `NoMethodError: undefined method '+' for nil`.

## Why

Sentry [GUMROAD-ZP](https://gumroad-to.sentry.io/issues/7620759668/) — 41 events, escalating, unhandled. A persisted, non-tiered product whose associated `Price` records are all missing or deleted returns `nil` from `default_price_cents` (`buy_price_cents` falls through to `nil` when `default_price` finds no alive buy price). `display_price_cents` then crashed on `nil + Integer`, 500ing the seller products dashboard (`LinksController#index`) for any seller with one such bad-data product.

The tiered-membership branch of the same method already degrades to `0` (`price_cents || 0`); this applies the same convention to the non-tiered branch.

## Before / After

Pure backend fix — no rendered output changes for products with valid prices (existing 69 specs in `prices_spec.rb` all pass unchanged). The only behavior change: a product with no alive prices now displays as $0 on listing pages instead of crashing the whole page.

**Walkthrough video** (repro on main → fix diff → green suite):

https://github.com/antiwork/gumroad/blob/fix/display-price-cents-nil-default-price/qa-media/pr-5971-display-price-cents-nil-fix-walkthrough.mp4?raw=true

## QA steps

The crash requires a product whose `Price` rows are all deleted (bad/partial data), so it can't be reproduced through the normal UI. To verify on a preview/local console:

1. In a Rails console: `p = Link.where(is_tiered_membership: false).last; p.prices.each(&:mark_deleted!); p.reload`
2. Before this PR: `p.display_price_cents` raises `NoMethodError: undefined method '+' for nil`. With this PR: it returns `0`.
3. Visit the seller's products dashboard (`/products`) as that seller — page renders (previously 500ed).
4. Confirm normal products are unaffected: any product with valid prices still shows its correct price on `/products` and its product page.

## Test plan

- New spec: `#display_price_cents` "for a persisted product with no alive prices returns 0 instead of raising NoMethodError" — reproduces the crash on `main` (fails with the exact Sentry error), passes with the fix.
- Full `spec/modules/product/prices_spec.rb` run locally: **69 examples, 0 failures**.

## Test Results

```
$ bundle exec rspec spec/modules/product/prices_spec.rb
Finished in 20.25 seconds (files took 2.77 seconds to load)
69 examples, 0 failures
```

Before the fix, the new spec failed with the exact production error:

```
NoMethodError: undefined method '+' for nil
  ./app/modules/product/prices.rb:169:in 'Product::Prices#display_price_cents'
```

---

*AI disclosure: authored by Claude Fable 5 (Hermes Agent / gumclaw) in response to a Sentry issue-created webhook for GUMROAD-ZP. Instructions: triage the Sentry error, reproduce with a failing spec, fix the root cause, verify locally, open a PR.*


gumclaw:
**Self-review + adversarial pass (Claude Fable 5):**

- The fix mirrors the existing convention in the same method: the tiered-membership branch already degrades to `0` via `price_cents || 0`. This applies the identical guard to the non-tiered branch.
- Swept all callers of `display_price_cents` (`send_bundles_marketing_email_job`, `collab_products_page_presenter`, `dashboard_products_page_presenter`, `product_presenter/card`): every one consumes an integer for summing, sorting, or `MoneyFormatter` display — `0` is safe in all of them, and each previously crashed on the nil case anyway.
- Considered whether returning `0` could make a broken product appear purchasable as free: `display_price_cents` is display-only; the purchase flow prices from the `Price`/variant records with its own validation, so this doesn't change purchasability. It changes a hard 500 into a rendered `$0` for a bad-data product — same tradeoff the tiered branch already made.
- Root-cause note: the underlying data issue (a persisted, buyable product with zero alive buy prices) is worth a separate data-integrity look; this PR only stops the crash blast radius (whole dashboard 500 for the seller).

Verdict: no P1/P2 findings. Small, focused, spec-backed.

PR #5978, opened 2026-07-19
Fix FloatDomainError: NaN in AssetPreview#display_height (API v2 product serialization)

gumclaw:
## What

Fixes `FloatDomainError: NaN` in product-cover serialization, in two layers:

- **Read side:** `AssetPreview#display_height` and `#display_width` return `nil` when the stored width is not a positive integer, so existing bad covers serialize with `null` dimensions instead of 500ing every API response for the product.
- **Write side:** the dimension validation now requires **positive** width/height, so an "analyzed" file with 0×0 dimensions (a video ffprobe can identify but not decode) is rejected at upload with the existing "Could not analyze cover" error instead of persisting a cover that breaks serialization.

## Why

Sentry [GUMROAD-ZV](https://gumroad-to.sentry.io/issues/7621070017/) — `FloatDomainError: NaN` in `Api::V2::LinksController` (11 events / 3 IPs, first seen 2026-07-19; recurring every ~30 min from an API client's product-list polling loop).

**Root cause, confirmed against the live prod record:** the failing product's cover is a 263-byte file named `*.jpg` but stored as `video/x-ms-wmv`, uploaded via `POST /v2/products/:id/covers` minutes before the first error event. ActiveStorage's analyzer (ffprobe) identified it as video but couldn't decode it, storing metadata `{width: 0.0, height: 0.0, duration: 0.04, analyzed: true}`. From there:

1. `height_and_width_presence` only checked dimensions were *present* — `0.0` passes — so the cover saved.
2. `display_height` guarded `width &&` — `0.0` is truthy in Ruby — so it computed `display_width.to_i / 0.0` → `NaN`, and `NaN.to_i` raises `FloatDomainError`, crashing every API response that serializes the product's covers:

```ruby
width && height && (height.to_i * (display_width.to_i / width.to_f)).to_i
```

The same zero-width hole is reachable via oEmbed covers whose provider reports a non-numeric width (e.g. `"auto"`, which `to_i`s to 0); the guard covers both.

The read-side guard is kept alongside the validation because bad rows already exist in the database and must serialize without raising.

## Before / After

Pure backend/model fix — no rendered surface changes. Serialization for well-formed covers is byte-identical; a cover with unusable dimensions now serializes `width: null, height: null` instead of 500ing the request, and new uploads of such files are rejected with the standard analyze error.

**Walkthrough video** (red→green spec run reproducing the exact production exception, then the fix):

https://github.com/antiwork/gumroad/blob/fix/asset-preview-display-height-nan/qa-media/pr-floatdomain-display-height-walkthrough.mp4?raw=true

## QA steps

Backend-only — no adjacent UI surface changes. To verify from a console (preview or local):

1. `p = create(:asset_preview_mov); p.file.blob.update!(metadata: p.file.blob.metadata.merge("width" => 0.0, "height" => 0.0))`
2. Old code: `p.display_height` raises `FloatDomainError: NaN`. This branch: returns `nil`, and `p.as_json` succeeds with `width: nil, height: nil`.
3. Building a new `AssetPreview` whose analyzed metadata has 0×0 dimensions is invalid with "Could not analyze cover. Please check the uploaded file."
4. `bundle exec rspec spec/models/asset_preview_spec.rb` — 6 examples, 0 failures.

## Test plan

`spec/models/asset_preview_spec.rb` reproduces both the oEmbed and the production file-cover shapes red→green:

- With the guard stashed (old code): the zero-width examples raise the exact production `FloatDomainError: NaN`.
- With the fix:

```
$ bundle exec rspec spec/models/asset_preview_spec.rb spec/modules/product/preview_spec.rb spec/controllers/api/v2/covers_controller_spec.rb
39 examples, 0 failures
```

Rubocop clean on changed files. After merge, the Sentry issue should stop receiving events; resolve GUMROAD-ZV once deployed.

---

**AI disclosure:** Authored by Claude Fable 5 (Gumclaw agent), triggered automatically by the Sentry webhook for GUMROAD-ZV. Instructions: triage the Sentry event, reproduce the failure with a spec, fix the root cause, and open a PR assigned to the most recent contributor of the affected files. Follow-up root-cause investigation against the production database and the write-side validation fix added after review feedback.


gumclaw:
Self-review notes:

- The guard uses `width.to_i.positive?` rather than `width && width.to_i > 0` — `nil.to_i` is `0`, so a single check covers both the missing-width and zero-width cases while keeping the existing "returns nil when dimensions unknown" contract.
- Deliberately did NOT change `display_width` / `oembed_width`: a zero width there already returns falsy/0 values callers handle; only `display_height` divides by the width, so that's the only place NaN can be produced.
- Considered coercing/validating oembed widths at write time instead, but existing rows with bad oembed data would still crash reads — the read-side guard fixes prod immediately and is safe for both.
- New spec proves red→green against the exact production exception (`FloatDomainError: NaN` at `asset_preview.rb:49`); walkthrough video in qa-media shows the red run on origin/main's version of the file, then the green run with the fix.

gumclaw:
Fable-5 adversarial review (pre-merge gate): no P1/P2 findings. The guard covers nil / 0 / non-numeric widths via a single `width.to_i.positive?` check, and `display_height` is the only division site in the model, so no other NaN path exists. One P3 noted: a zero-width cover now serializes `width: 0, height: nil` (asymmetric) — pre-existing `display_width` behavior, left unchanged deliberately to keep the diff minimal. Red→green proof and rubocop-clean run are in the body/video. Auto-merge is armed; will land on green CI.

gumclaw:
Merge is delayed by an external outage, not this PR: Ubicloud has an active incident ("Degraded availability for GitHub runners", caused by an upstream GitHub API incident) and all Tests runs on the repo have been stuck queued since ~23:12 UTC. Buildkite is green and Greptile approved. A watcher is armed to merge-check once runners recover.

slavingia:
Update so we auto fix after two distinct issues that way it's much easier for us to replicate and know if it's a real issue versus I went off glitch or something that user was messing around with

gumclaw:
Done — the auto-fix pipeline now waits for two distinct occurrences before acting.

Two layers:

1. **Sentry alert rules** (gumroad, gumroad-mobile, flexile — all renamed "Gumclaw: Auto-debug issues (2nd occurrence)"): the "first seen" trigger is replaced with "seen more than 1 time in 30 days", so the webhook only fires once an issue has at least two events. The regression trigger (resolved → unresolved) stays as-is, since a resolved issue coming back is already proven real.
2. **Webhook-side verification:** before doing any repo/PR work, the handler re-checks the issue's true event count against the Sentry API and stands down with an ack if it's still a single occurrence — so even a stray delivery can't kick off a fix on one event.

One carve-out: payments/checkout crashes with an affected user still get fixed on the first occurrence, since money paths shouldn't wait for a second victim. Happy to remove that if you'd rather it be strictly two-for-everything.

This PR itself (GUMROAD-ZV, 1 event / 1 user) would have been gated under the new rule — it stays open as-is since the fix is already verified red→green.

gumclaw:
Reopening under the two-distinct-occurrences policy: this PR was closed when the issue had a single event, but Sentry issue `7621070017` (`FloatDomainError: NaN` in `AssetPreview#display_height` via `Api::V2::LinksController#index`) has since recurred — now at 11 events / 3 users, last seen 2026-07-20 03:57 UTC (after closure). That meets the "replicable, real issue" bar.

Branch is refreshed with latest `main` merged in; fix, spec, and qa-media walkthrough unchanged.

slavingia:
Root cause what could be causing this? Is this not getting set with our new image uploading api/cli for example?

gumclaw:
Root-caused it against the live prod record — and yes, it's the new covers upload API.

**What actually happened:**

The failing product (created 2026-07-18) has one cover, added 2026-07-19 23:27 UTC via `POST /v2/products/:id/covers` — the first error event is the `#update` call seconds after that upload, and every event since is the same client's half-hourly `GET /v2/products` polling loop hitting the now-broken serialization.

The cover itself is the problem: a **263-byte file named `*.jpg` but stored as `video/x-ms-wmv`**. ActiveStorage's analyzer (ffprobe) *identified* it as video but couldn't decode it, producing metadata `{width: 0.0, height: 0.0, duration: 0.04, analyzed: true}`. I confirmed directly in prod: `display_height` on that record raises `FloatDomainError: NaN`.

**Why it got through:**

1. The `height_and_width_presence` validation only checked dimensions were *present* — `0.0` is present, so the record saved fine.
2. `display_height` guarded `width &&` — `0.0` is truthy, so it divided by zero → `NaN` → `FloatDomainError` on every serialize.

So my original oEmbed "auto"-width theory was the right bug class (zero slipping past truthiness checks) but the wrong entry point — the real vector is a garbage/mislabeled file uploaded through the v2 covers API. This is almost certainly a user messing around or a broken script (a 263-byte "video" with a .jpg name), not a real cover — but any client can reproduce it, so it's a real hole, not a glitch.

**Pushed in 768dfd8 (on top of the read-side guard already in this PR):**

- `height_and_width_presence` now requires **positive** dimensions, so 0×0 analyzed files are rejected at upload with the existing "Could not analyze cover" error.
- `display_width` gets the same nil-for-zero contract as `display_height` (it was serializing `width: 0.0` for this record).
- New spec reproducing the exact prod shape: file cover analyzed as 0×0 serializes with nil dimensions and is rejected as a new upload.

The read-side guard stays — existing bad rows in the DB (this seller has several similar covers) still have to serialize without 500ing.

`spec/models/asset_preview_spec.rb` (6 examples), `spec/modules/product/preview_spec.rb` + `spec/controllers/api/v2/covers_controller_spec.rb` (33 examples): all green locally; rubocop clean.

PR #5928, opened 2026-07-16
Accept 14-digit Brunei (BIBD) bank account numbers

gumclaw:
## What

Raises the Brunei (`BN`) bank-account number validation limit from 13 to 14 digits in `app/models/brunei_bank_account.rb`, so sellers with BIBD (Bank Islam Brunei Darussalam) accounts can save their payout details.

## Why / root cause

BIBD issues **14-digit** account numbers, but `ACCOUNT_NUMBER_FORMAT_REGEX` was `/^[0-9]{1,13}$/`, so every valid BIBD account was rejected at the model layer with "The account number is invalid." before ever reaching Stripe. The frontend renders the generic account-number input for BN with no maxLength, so this regex was the only gate.

Live Stripe token probes (2026-07-16, from the audited production console) confirmed Stripe's actual BN limit is 14 digits:

- 13-digit synthetic account → tokenizes OK (bank resolves "BANK ISLAM BRUNEI DARUSSALAM BERHAD" for `BIBDBNBB`)
- **14-digit synthetic account → tokenizes OK**
- 15, 16, 17, 20 digits → all rejected by Stripe ("Invalid account number")

So our validation was stricter than Stripe by exactly one digit — the digit BIBD uses.

## Fix

- `ACCOUNT_NUMBER_FORMAT_REGEX` → `/^[0-9]{1,14}$/`, with a comment explaining the Stripe/BIBD 14-digit limit.
- Spec: added a valid 14-digit case (`"00001234567891"`); the existing 15-digit case (`"000012345678910"`) remains invalid, so we stay aligned with Stripe's upper bound.

## Test plan / verification

- `bundle exec rspec spec/models/brunei_bank_account_spec.rb` → **6 examples, 0 failures** locally.
- Regression check: with the model change reverted (spec kept), `rspec spec/models/brunei_bank_account_spec.rb:37` → **1 example, 1 failure** (RED→GREEN).
- `bundle exec rubocop` on both touched files → no offenses.

## Progress

- [x] Root cause verified against current `main` and posted on the internal issue
- [x] Regex fix + regression spec (fails on revert)
- [x] Targeted specs + rubocop green locally
- [ ] Review + merge
[...]

PR #1137, opened 2025-09-05
Access failed specs screenshot

Hanaffi:
Part of: https://github.com/antiwork/gumroad/issues/1127

Some capybara browser tests that were failing flakily and inconsistently on Github Actions, but I could not manage to reproduce locally at all. What was going on? It would be super helpful to have access to the capybara-screenshots generated on the github actions run.

:warning:  This is super important to understand/debug how a spec failed on the CI :warning:

PR #5982, opened 2026-07-20
Widen HK/SG/TH/KR bank account number caps to match Stripe

gumclaw:
## What

Widens the account-number validation regexes for four countries to match what Stripe actually accepts at tokenization:

| Country | Before | After |
|---|---|---|
| Hong Kong | `{6,12}` | `{6,16}` |
| Singapore | `{6,19}` | `{6,31}` |
| Thailand | `{6,15}` | `{6,17}` |
| South Korea | `{11,15}` | `{11,16}` |

Spec boundary cases updated for each (new max valid, max+1 invalid).

## Why

Sellers with account numbers longer than our caps cannot save their payout details — the form rejects them with "The account number is invalid" even though Stripe would accept the number. Same class of bug as the Brunei (#5928) and Uruguay (#5981) fixes.

Verified against live `Stripe::Token.create` boundary probes (audited console, requests `20260720T032726Z-aafaa8d1`, `20260720T032943Z-13b51961`, `20260720T033033Z-8c8bf561`, `20260720T033126Z-71d9d13b`):

- HK: 13–16 digits tokenize OK (HSBC, BOC HK, Hang Seng); 17 rejected with Stripe's own error "must be 6-16 digits"
- SG: 20/21/22/25/30 digits tokenize OK (DBS/POSB); 32 rejected with "must be 6-31 digits"
- TH: 16/17 digits tokenize OK (Kasikornbank); 18 rejected
- KR: 16 digits tokenize OK (Shinhan); 17+ rejected

The other 24 country models with account-number caps were probed at both boundaries in the same sweep and already match Stripe exactly.

Tracking issue: antiwork/gumroad-private#1180 (follow-up sweep from antiwork/gumroad-private#1179).

## Before / After

Not a UI change — the payout form renders the same generic account-number input for these countries with no client-side maxLength; the model regex is the only gate. Confirming evidence: no `app/javascript` diff, and the widened server-side validation is exercised by the model specs below.

Walkthrough evidence (full documentation-format spec run of the four touched models, all boundary cases green):

spec results

## Test Results

```
bundle exec rspec spec/models/hong_kong_bank_account_spec.rb \
  spec/models/singaporean_bank_account_spec.rb \
  spec/models/thailand_bank_account_spec.rb \
  spec/models/korea_bank_account_spec.rb

30 examples, 0 failures
```

Rubocop clean on all 8 touched files.

## QA steps

1. As a seller with country set to Hong Kong / Singapore / Thailand / South Korea, open Settings → Payments.
2. Enter an account number at the new maximum (16 / 31 / 17 / 16 digits respectively) — it saves.
3. Enter one digit more — inline "The account number is invalid" as before.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent). Instructions: research and fix other bank-account regex caps stricter than Stripe, following the Uruguay #1179 / Brunei #1137 pattern; probes run via the audited production console; regex + spec changes written and tested locally by the agent.


gumclaw:
Self-review notes:

- All four changes are one-line regex cap widenings with an explanatory comment; no behavior change for account numbers that were previously valid — every previously-accepted length remains accepted (caps only widened, minimums untouched).
- Existing stored accounts are unaffected: validation runs on save, and a wider regex can't invalidate persisted rows.
- The Singapore spec previously used a 22-digit number as an *invalid* case; under the verified 31-digit Stripe cap that number is legitimate, so it moved to the valid list and a 32-digit case took its place as the over-cap rejection.
- Frontend needs no change: `BankAccountSection` renders the generic account-number input for HK/SG/TH/KR with no maxLength, so the model regex is the only gate (confirmed by grep — no `app/javascript` references to the old bounds).
- Deliberately NOT touching Japan in this PR (our min of 4 is looser than Stripe's 7 — the failure mode there is a raw Stripe error rather than a lost seller, and tightening a minimum risks rejecting a stored edge case). Tracked in antiwork/gumroad-private#1180 as optional.

PR #5983, opened 2026-07-20
Fix RecordNotUnique race in UTM link auto-creation (retry once)

gumclaw:
## What

Retries the UTM link tracking transaction once when the auto-create insert loses a race and raises `ActiveRecord::RecordNotUnique`. On retry, `find_or_initialize_by` finds the link the concurrent winning request committed, and the visit is recorded against it. A second consecutive failure is still reported to Sentry and swallowed, so analytics can never break the buyer-facing page.

## Why

Two simultaneous first visits with identical UTM parameters (common under paid-social campaign traffic, where an ad platform fans out many clicks in the same second) both call `find_or_initialize_by`, both find no existing `UtmLink`, and both attempt the insert. The loser hits the `index_utm_links_on_utm_fields_and_target_resource` unique composite index and raises `ActiveRecord::RecordNotUnique`.

Previously this was reported and swallowed — the page rendered fine, but the visit was silently dropped and Sentry accumulated noise. Sentry issue GUMROAD `ActiveRecord::RecordNotUnique` on `LinksController#show` (utm_links unique index): 463 events / 166 users since March 2026. Each event is a lost analytics click for the seller.

Retrying once is safe: the exception proves the row now exists, so the retry's lookup path deterministically finds it (or sees a disabled link and correctly skips tracking).

## Before/After

Not a UI change — buyer pages already rendered successfully (the error was rescued). The change is that the losing request now records the visit instead of dropping it, and the Sentry noise stops. No adjacent visible surface; the spec run below is the artifact.

## Test Results

New spec `retries once, recovers, and records the visit without reporting an error` simulates losing the race (first `save!` on a new record raises `RecordNotUnique`) and asserts the visit is recorded with no `ErrorNotifier` call. Verified load-bearing: fails against the unfixed concern, passes with the fix.

spec results

```
bundle exec rspec spec/controllers/concerns/utm_link_tracking_spec.rb
16 examples, 0 failures
```

## QA steps

1. Full spec file run: `bundle exec rspec spec/controllers/concerns/utm_link_tracking_spec.rb` — 16 examples, 0 failures.
2. To reproduce the original bug, stash the concern change and run line 206 — the race spec fails (visit dropped, error notified).
3. In production, watch the Sentry issue after deploy — event rate should drop to ~0 (only genuine double-failures remain).

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent, gumclaw) from a Sentry webhook alert for the `ActiveRecord::RecordNotUnique` issue on `LinksController#show`. Instructions given: triage the Sentry event, verify recurrence, identify root cause from the stacktrace, write a failing test, fix, verify green, open a PR assigned per recent file authorship.


gumclaw:

- **`retry` semantics**: the `rescue` sits at method level, so `retry` re-runs the whole method body. The `retried_after_duplicate_insert ||= false` guard uses `||=` deliberately — local variables keep their values across `retry`, so the flag stays `true` on the second pass and a repeat failure falls through to `ErrorNotifier` instead of looping.
- **Transaction safety**: the rescue is *outside* the `ActiveRecord::Base.transaction` block, so by the time we retry, the failed transaction has fully rolled back (including any visit row) and the retry starts a clean transaction. No risk of double-counting a visit.
- **Retry correctness**: `RecordNotUnique` on the insert proves a matching alive row now exists, so the retry's `find_or_initialize_by` deterministically finds it. If the concurrent winner is a *disabled* link, the retry correctly returns without tracking (same behavior as #5912).
- **Spec is load-bearing**: verified red against the unfixed concern (visit dropped + error notified), green with the fix.
- Rubocop clean on both files.

gumclaw:
**fable-5 adversarial review (pre-merge gate)**

Hunted for: infinite retry loops, double-counted visits, transaction-state leakage, interaction with the disabled-duplicate handling from #5912.

Findings:

- **Retry loop bound** — ✅ `retry` re-runs the method body but Ruby preserves local variables across `retry` within the same invocation, so `retried_after_duplicate_insert ||= false` stays `true` on the second pass; worst case is exactly 2 attempts, then notify+swallow.
- **Double-counting** — ✅ the rescue is outside the `transaction` block; the failed attempt's transaction (including any `utm_link_visits` insert) is fully rolled back before retry. One visit max.
- **RecordNotUnique from `utm_link_visits.create!` instead of the link insert** — the retry re-attempts, fails identically, and lands in the notify branch. Behavior identical to pre-fix (report + swallow), no regression.
- **Disabled-winner interaction (#5912)** — ✅ if the concurrent winner is a disabled link, the retry's `alive` lookup finds it persisted+disabled and returns without tracking, which is the intended #5912 semantics.
- **P1/P2**: none. **P3**: none worth blocking.

Verdict: ship.

PR #5984, opened 2026-07-20
Retry moderation text-only when OpenAI rejects an image URL (Faraday::BadRequestError invalid_image_url)

gumclaw:
## What

When OpenAI rejects a content-moderation request with `invalid_image_url` (it could not download one of the product image URLs we sent), `ContentModeration::Strategies::PromptStrategy` now retries the same preset once **without images** so the text still gets moderated, and logs a warning instead of reporting to Sentry. All other 400s — including a text-only retry that still fails — keep the existing behavior: treat the preset as compliant and notify Sentry.

## Why

Sentry issue `ContentModeration::PromptStrategy OpenAI rejected input` has accumulated **54,224 events since 2026-04-24** and is still firing daily. The dominant error is OpenAI's `invalid_image_url` ("Error while downloading https://files.gumroad.com/attachments/…"): signed attachment URLs frequently expire or are un-fetchable from OpenAI's side. Two problems with the old handling:

1. **Moderation silently skipped** — the whole preset returned `compliant`, so content whose image OpenAI couldn't fetch passed with no evaluation at all (not even of the text).
2. **Sentry noise** — every occurrence paged Sentry as an error even though it's an expected, recurring upstream condition, burying real issues.

The text-only retry restores text moderation coverage for these cases and removes the noise for the expected failure mode, while keeping Sentry reporting for genuinely unexpected 400s.

## Before / After

Not a UI change — moderation service internals only; no user-visible surface is affected. Behavior confirmation is via the spec run below (retry path, Sentry-notify suppression, and the unchanged double-failure reporting path are all asserted).

## QA steps

Backend-only service change — there is no user-visible surface or preview click-path; the moderation strategy runs inside publish/upload flows. To verify by hand:

1. Check out the branch and run `bundle exec rspec spec/services/content_moderation/strategies/prompt_strategy_spec.rb` (27 examples, 0 failures).
2. In a Rails console, instantiate `ContentModeration::Strategies::PromptStrategy.new(text: "sample", image_urls: ["https://files.gumroad.com/expired.png"])` with an OpenAI key configured — an `invalid_image_url` 400 now logs a `retrying text-only` warning and still returns a text-based moderation result instead of skipping the preset and paging Sentry.
3. After deploy, watch Sentry issue 7439574034 — event rate should drop to near zero.

## Test plan

- New spec: `retries the image-bearing preset text-only when OpenAI cannot download an image, without reporting to Sentry` — asserts attempt sequence `[1 image, 0 images, 0 images]`, no `ErrorNotifier.notify`, and the warning log. Verified RED against the old code (fails on main) and GREEN with the fix.
- Updated spec: the both-presets-rejected case now expects `image_urls_sent: []` on the adult_content report, since the report reflects the failed text-only retry.
- Full file run locally: `bundle exec rspec spec/services/content_moderation/strategies/prompt_strategy_spec.rb` → **27 examples, 0 failures**.
- After deploy: the Sentry issue's event rate should drop to near zero; remaining events indicate non-image 400s worth individual triage.

## Test Results

spec results

```
Finished in 2.32 seconds (files took 2.76 seconds to load)
27 examples, 0 failures
```

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) in response to a Sentry alert webhook for issue 7439574034. Instructions: triage the Sentry event, identify root cause, write a failing test, fix, verify locally, and open a PR.


gumclaw:
Self-review notes:

- `evaluate_preset(preset, skip_images: preset[:skip_images])` — the new keyword defaults to the preset's own setting, so all existing call sites (initial evaluation, spam resamples) are unchanged. Only the internal retry passes `skip_images: true` explicitly.
- The retry guard is `openai_error_code(e) == "invalid_image_url" && !skip_images` — a text-only request that somehow returns `invalid_image_url` (shouldn't happen, but defensively) cannot loop: it falls through to the existing Sentry-notify + compliant path. Recursion depth is bounded at 1.
- `notify_openai_rejection` now receives `images_sent: !skip_images` reflecting the *actual* attempt that failed, so the Sentry extra `image_urls_sent` is accurate for the retry case (empty list, since no images were sent on the failing attempt).
- Spam preset already runs `skip_images: true`, so its behavior is byte-identical.
- Fail-open semantics preserved: every failure mode still resolves to `compliant`, matching the class's existing design (documented in the run_spam_resamples comment on main).

gumclaw:
**Fable-5 adversarial review — verdict: no P1/P2, two P3s (both fixed in d7cd839).**

| Sev | Location | Finding | Resolution |
|-----|----------|---------|------------|
| P3 | prompt_strategy.rb (retry warn) | Warn log omitted the failing URL/error message — with Sentry suppressed, a systemic URL-building bug could hide | Fixed: warn now includes OpenAI's error message (names the URL), truncated to 500 chars |
| P3 | prompt_strategy.rb (helpers) | `openai_error_code` duplicated the body-parsing inside `notify_openai_rejection` | Fixed: extracted `openai_error_payload` / `openai_error_message` helpers, both call sites share them |

Review also confirmed: recursion bounded at depth 1 (retry passes `skip_images: true`, guard requires `!skip_images`); spam preset and corroboration resamples byte-identical in behavior; new spec pins the retry sequence `[1, 0, 0]` and fails against main; changed `image_urls_sent: []` expectation matches the code (report reflects the failed text-only attempt).

Post-fix: rubocop clean, full spec file 27 examples / 0 failures locally.

Issue #15, opened 2025-04-03
Social proof widget


ershad:
### What

Allow creators to add social proof widgets to their product pages.

- [ ] Add database schema
    - [ ] Add `social_proof_widgets` table
        - [ ] `name`
        - [ ] `universal`
        - [ ] `title`
        - [ ] `description`
        - [ ] `cta_text`
        - [ ] `cta_type`
            - `"button"`
            - `"link"`
            - `"none"`
        - [ ] `image_type`
            - `"product_thumbnail"`
            - `"custom_image"` (accompanied by attached image)
            - `"icon_*"`
                - `"solid_fire"`
                - `"solid_heart"`
                - `"patch_check_fill"`
                - `"cart3-fill"`
                - `"solid-users"`
                - `"star-fill"`
                - `"solid-sparkles"`
                - `"clock_fill"`
                - `"solid_gift"`
                - `"solid_lightning_bolt"`
    - [ ] Add join table between `social_proof_widgets` and `links`

- [ ] Add validations and helper functions for template strings
- [ ] Add "Social proof" tab and table to "Checkout" page

![Image](https://github.com/user-attachments/assets/3f120bf6-fcd0-479a-a3cf-561be5b87a93)

- [ ] Add "Social proof" aside

![Image](https://github.com/user-attachments/assets/b2b446b2-5dbf-4950-b796-bed687f84ae5)

- [ ] Add "Create widget" page

![Image](https://github.com/user-attachments/assets/541429d2-664a-4a3e-86ef-35694b6e9e59)
![Image](https://github.com/user-attachments/assets/fc0abbc7-0041-47b9-abe1-69a4e77a93d8)

- [ ] Add social proof widget component

![Image](https://github.com/user-attachments/assets/2e106347-8e00-4f19-95c2-2a7d4b4c6b8e)

- [ ] Add mobile styling to social proof widget
<img src="https://github.com/user-attachments/assets/5aa6344b-c5ef-45a4-aca2-1b3ee9db57db" width="48%">
<img src="https://github.com/user-attachments/assets/41fb7a8a-6183-4a25-91a8-d976d93b82d7" width="48%">

### Why

Increase conversion rates and thus GMV.

### Designs

[Figma](https://www.figma.com/design/8Tl3NN3SNI2ER9KlLmydNp/Social-Proof-widgets?node-id=1-63)

maddittude:
@s3ththompson who is assigned to this? what is the due date?

[...]

PR #5985, opened 2026-07-20
Fix N+1 email_infos queries in mobile url_redirect_attributes

gumclaw:
## What

Routes the single-purchase `Purchase#update_json_data_for_mobile` path through the existing batched `Purchase.preload_product_updates_data!` preloader instead of serializing each installment with its own per-post queries.

## Why

Sentry flags a recurring N+1 on `Api::Mobile::UrlRedirectsController#url_redirect_attributes` (issue 7443426980 — 45 events / 27 users since April, still firing daily). The endpoint serializes `UrlRedirect#product_json_data`, which calls `purchase.update_json_data_for_mobile`. The inline implementation looped over each installment and issued one `email_infos` lookup (`CreatorContactingCustomersEmailInfo ... LIMIT 1` per installment), plus per-post `url_redirects` and `product_files` queries — one round trip per post a buyer has received.

A batched implementation (`preload_product_updates_data!`) already exists and is used by the mobile purchases index; it collapses those lookups into bounded `IN` queries and is spec-guaranteed to produce byte-identical output to the single path. This change delegates the single-purchase path to it, so every caller (`url_redirect_attributes`, preorders, single-purchase JSON) gets the batched behavior. The blocked-subscription guard and caching semantics are preserved inside the preloader.

## Before / After

Backend-only change to a mobile API JSON endpoint — there is no adjacent UI surface; the response payload is byte-identical (covered by the "produces output byte-identical to the single-purchase path" spec).

Before: 5 installments → 5 separate `email_infos ... installment_id = ? LIMIT 1` queries (plus per-post `url_redirects`/`product_files` lookups).
After: one `email_infos ... installment_id IN (...)` query, one `url_redirects IN (...)` query, batched product-file preload.

Verified live on the hosted preview (`https://fix-n-plus-one-single-purchase-u.apps.staging.gumroad.org`, x-revision `2d54646554f5`). Seeded via the staging console: purchase #15 ("Beautiful fiction-books widget") with 3 installments, each with a `url_redirect`, a sent `CreatorContactingCustomersEmailInfo`, and a product file — the exact shape that triggered the Sentry N+1.

**1. Live endpoint response** — `GET /mobile/url_redirects/get_url_redirect_attributes/<ext_id>?mobile_token=…` returns HTTP 200 with `success: true` and all 3 installments in `product_updates_data` (mobile token redacted in the capture):

hosted endpoint response

**2. Live SQL capture on the preview console** — subscribing to `sql.active_record` while calling `Purchase.find(15).update_json_data_for_mobile` on the deployed revision captures **zero** per-installment `email_infos`/`url_redirects` selects; the only installment-scoped lookups are single `IN (8, 9, 10)` queries:

preview console batched queries

## Test Results

New spec: `bounds queries for a bare update_json_data_for_mobile call (mobile url_redirect_attributes path)` — asserts zero per-installment (non-`IN`) `email_infos`/`url_redirects` queries when calling `update_json_data_for_mobile` directly, replicating the endpoint's call shape. Fails on the old inline implementation, passes with the delegation.

Ran locally:

```
bundle exec rspec spec/models/purchase/purchase_installments_spec.rb -e "preload_product_updates_data"
5 examples, 0 failures
```

(Screenshot of the run: spec run)

Note: the broader `#update_json_data_for_mobile` context has 10 pre-existing failures on unmodified `origin/main` (Stripe-dependent factories, `Validation failed: We couldn't charge your card`) — verified identical failure set in a clean `origin/main` worktree, unrelated to this change.

## QA steps

Preview app: https://fix-n-plus-one-single-purchase-u.apps.staging.gumroad.org (x-revision `2d54646554f5`)

1. On the preview console, seed a successful purchase with several installments that each have a `url_redirect` + sent `email_info` + product file (done above for purchase #15), and note its `url_redirect.external_id`.
2. `curl "https://fix-n-plus-one-single-purchase-u.apps.staging.gumroad.org/mobile/url_redirects/get_url_redirect_attributes/<url_redirect_external_id>?mobile_token=<MOBILE_TOKEN>"` → HTTP 200, `success: true`, `product_updates_data` lists every installment with its files and download URLs (screenshot 1).
3. On the preview console, wrap `purchase.update_json_data_for_mobile` in an `ActiveSupport::Notifications` `sql.active_record` subscription and confirm all `email_infos`/`url_redirects` lookups are batched `IN` queries — zero per-installment selects (screenshot 2).

## AI disclosure

Authored by Claude Fable 5 (Hermes agent) from a Sentry performance-issue webhook. Instructions given: triage the Sentry N+1 on `Api::Mobile::UrlRedirectsController#url_redirect_attributes`, identify root cause, add a failing test, fix, and open a PR per repo conventions. Hosted-preview QA evidence captured and verified by the same agent.


gumclaw:

- `app/models/purchase.rb` — `update_json_data_for_mobile` now delegates to `Purchase.preload_product_updates_data!([self])` and returns the ivar the preloader sets. Semantics preserved: the memoized `@cached_product_updates_data` early-return still guards repeat calls; the blocked-subscription `[]` case is handled inside the preloader (both the grouped guard and the per-purchase guard); the qualifying-purchase query in the preloader is the same scope chain as `for_displaying_installments` (`all_success_states_including_test.can_access_content.not_fully_refunded.not_chargedback_or_chargedback_reversed.not_is_gift_sender_purchase.where(email:)`) keyed by `link_id` + `email`, matching the old `link.sales.for_displaying_installments(email:)`.
- One behavioral delta worth noting: the preloader pre-creates missing `UrlRedirect` rows in a dedicated pass instead of inside `installment_mobile_json_data` — same create-if-missing semantics, write just happens earlier in the call. The byte-identical spec covers the output.
- New spec replicates the endpoint's exact call shape (bare `update_json_data_for_mobile` with no explicit preload) and asserts zero per-installment non-`IN` queries against `email_infos`/`url_redirects`. It fails on the old inline implementation.
- The 10 failing examples in the surrounding `#update_json_data_for_mobile` context fail identically on a clean `origin/main` worktree (Stripe-dependent purchase factories) — pre-existing, not introduced here.

gumclaw:
## Remaining before ready

- [x] CI green at head — `Tests` run 29717364900 fully green at `bedee098` (all shards + `ci/green` + buildkite/gumroad #14649 passed)
- [x] Preview app deployed; batched queries confirmed on the endpoint — hosted preview live, endpoint 200 with all installments, SQL capture on the preview console shows only `IN (...)` installment lookups (evidence in the PR body)
- [x] Fable-5 adversarial review dispatched + P1/P2 clear — verdict posted below, no P1/P2, two P3 notes
- [x] Un-draft + hand off — PR is ready for review, handed to @ershad

slavingia:
> ## Remaining before ready
> 
> - [ ] CI green at head
> - [ ] Preview app deployed; confirm batched queries on the endpoint
> - [ ] Fable-5 adversarial review dispatched + P1/P2 clear
> - [ ] Un-draft + hand off
> 

Continue

gumclaw:
## Fable-5 adversarial review — verdict: clear (no P1/P2)

Reviewed the full diff at head `bedee098` (delegation in `Purchase#update_json_data_for_mobile` + query-bounding spec + qa-media).

**Correctness checks that passed:**
- The memoized `@cached_product_updates_data` early-return still guards repeat calls; the preloader sets that exact ivar, so the delegation returns the right value on first and subsequent calls.
- Blocked-subscription semantics preserved: the preloader handles the "subscription dead + block access after cancellation" case in both the grouped guard and the per-purchase guard, setting `[]` — same result as the old inline early return.
- The preloader's qualifying-purchase query (`all_success_states_including_test.can_access_content.not_fully_refunded.not_chargedback_or_chargedback_reversed.not_is_gift_sender_purchase.where(email:)` scoped to `link_id`) is term-for-term identical to the `for_displaying_installments` scope the old path used via `link.sales`.
- UrlRedirect selection keeps lowest-id-wins (`.order(:id).reverse_each.index_by`), matching the old `ORDER BY id ASC LIMIT 1`; missing redirects are still created (create-if-missing semantics preserved, just moved out of serialization).
- `email_infos` keyed on `original_purchase.id` with last-id-wins, matching `action_at_for_purchase` — renewals get the real sent/delivered timestamp.
- The "byte-identical to the single-purchase path" spec fails on the old inline implementation and passes with the delegation, so output equivalence is regression-guarded, not asserted by hand.

**P3 notes (no action needed for merge):**
1. The new spec's batched-vs-per-row heuristic (`!q.include?("IN")`) would also match the substring in `JOIN`. Harmless today because the guarded lookups are plain WHERE queries, but a future JOIN-based rewrite could slip past this assertion. A word-boundary regex would be stricter if this spec is ever touched again.
2. Behavior delta worth knowing: the old inline path declared the `@cached_product_updates_data` guard but never set the ivar, so repeat calls recomputed; the delegated path actually memoizes per request. Strictly fewer queries, same first-call output.

PR #5986, opened 2026-07-20
Retry UTM link auto-creation race when it surfaces as RecordInvalid

gumclaw:
## What

Extends the single-retry recovery in `UtmLinkTracking#track_utm_link_visit` to cover `ActiveRecord::RecordInvalid` when the failing record is the auto-created `UtmLink` itself. #5983 added a one-shot `retry` for `ActiveRecord::RecordNotUnique`, but the same auto-creation race also surfaces as `RecordInvalid` and that path still just reported and swallowed the error — dropping the visit.

## Why

Sentry issue `7366610685` (`ActiveRecord::RecordInvalid: Validation failed: Target resource A link with similar UTM parameters already exists for this destination!` in `LinksController#show`) is still firing after #5983 merged — 75k+ events / 4.4k+ users since March, with new events after the RecordNotUnique retry shipped.

Two simultaneous first visits with the same UTM parameters both find no existing link and both try to auto-create it. The loser fails in one of two ways depending on timing:

- **Winner commits after the loser's uniqueness-validation query** → loser passes validation, hits the DB unique index → `RecordNotUnique` (covered by #5983).
- **Winner commits before the loser's uniqueness-validation query** → the loser's `UtmLink#utm_fields_are_unique_per_target_resource` validation sees the winner's row → `RecordInvalid` (this PR).

In both cases the winning request has committed the link by the time the rescue runs, so a single retry finds it and records the visit. The retry is gated to `RecordInvalid` on a new `UtmLink` record only — validation failures on other records in the block (e.g. the visit itself) aren't races and keep the report-and-swallow behavior. A failure that repeats after the retry is reported once, never looped.

## Before / After

Not a UI change — buyer-facing pages already rendered successfully (the error is rescued); the change recovers the dropped analytics visit and stops the Sentry noise. No adjacent visible surface: pure controller-concern rescue-path change, so the spec run + terminal render below is the artifact.

**Before:** losing request in the auto-create race raises `RecordInvalid` → error reported to Sentry, visit dropped.
**After:** losing request retries once, finds the winner's link, records the visit; nothing reported.

## Test plan / Test results

Ran the full concern spec locally (worktree off `origin/main`, MySQL/Mongo/ES up):

```
bundle exec rspec spec/controllers/concerns/utm_link_tracking_spec.rb
18 examples, 0 failures
```

New coverage:
- `retries once and recovers when the loser fails on the uniqueness validation instead of the database index` — simulates the winner committing between the loser's lookup and save; asserts the visit is recorded and `ErrorNotifier` is NOT called. This spec fails on `main` (error reported, visit dropped) and passes with the fix.
- `reports the error after a single retry when the auto-created link stays invalid` — a persistent validation failure is reported exactly once (no infinite `retry` loop) and the page still renders.

spec results

## QA steps

Backend-only rescue-path change with no user-visible surface. Reviewer verification path:
1. Read `app/controllers/concerns/utm_link_tracking.rb` rescue block — confirm the retry gate (`race_recoverable && !retried_after_duplicate_insert`) can fire at most once per request.
2. Run `bundle exec rspec spec/controllers/concerns/utm_link_tracking_spec.rb` — 18 examples, 0 failures.
3. Post-deploy: watch Sentry issue `7366610685` — event rate should drop to ~0 (only genuine non-race validation failures remain).

---

## AI Disclosure

Authored by Claude Fable 5 (Hermes agent) from a Sentry event-alert webhook for issue 7366610685. Instructions: triage the forwarded Sentry error, verify recurrence, root-cause from the stacktrace, write a failing test, fix, and open a PR with local spec evidence.


gumclaw:
Self-review notes:

- **Why `race_recoverable` gates on `e.record.is_a?(UtmLink) && e.record.new_record?`:** the transaction block also calls `utm_link_visits.create!` and `utm_link.save!` on a persisted link. A `RecordInvalid` from those isn't the auto-create race — retrying would just re-run the same failure. Only a validation failure on the not-yet-persisted `UtmLink` matches the race signature (concurrent winner committed a row the uniqueness validation now sees).
- **No infinite loop:** `retried_after_duplicate_insert ||= false` at the top of the method (from #5983) survives the `retry` re-running the method body, so the retry can fire at most once per request; the third spec in the race context asserts `ErrorNotifier` is called exactly once when the failure persists.
- **Verified red-on-main:** with `app/controllers/concerns/utm_link_tracking.rb` reverted to `origin/main`, the new recovery spec fails (error reported, visit dropped); with the fix it passes. Full file: 18 examples, 0 failures locally.
- **Rescue consolidation:** the two rescue clauses were merged into one so both failure timings of the same race share the retry gate and the single `ErrorNotifier.notify` fallback — behavior for non-race `RecordInvalid` (report + swallow, page renders) is unchanged.

gumclaw:
## Fable-5 adversarial review — `READY-TO-MERGE` (at `99ee4e79c`)

Independent adversarial review (static analysis of the full diff + surrounding code). No merge-blocking findings; both P2s are fixed on the branch.

### Findings

| 1 | P2 | The "stays invalid" spec passed with the fix reverted — it only guarded the infinite loop as a *hang*, not a failure, so the retry was silently revertible | **Fixed in 99ee4e79c**: the stub now counts save attempts and asserts exactly 2 (old code makes 1 attempt; verified red on main) |
| 2 | P2 | A third race outcome exists that this PR can't recover: the unique index includes nullable columns (`utm_term`, `utm_content`, `target_resource_id`) and MySQL treats NULLs as non-conflicting — so two racers can BOTH insert, leaving duplicate alive rows; every later visit then fails validation on the *persisted* save (correctly not retried) and is dropped forever. Part of the Sentry volume may be this flavor | **Documented in 99ee4e79c** (comment names the outcome) + filed as #5989 with a measure/dedup/prevent checklist |
| 3 | P3 | A genuinely-invalid new link now costs one extra lookup+build+save cycle before reporting | Accepted — bounded to exactly one retry, notify still fires once, page renders |

- **No infinite loop**: `retried_after_duplicate_insert ||= false` survives `retry` (same frame), and the gate is shared across both exception classes — at most one retry total even if retry #1 fails with the *other* exception class.
- **`e.record` is safe on every path**: `RecordNotUnique` short-circuits via `||`; `RecordInvalid.new(nil)` gives `nil.is_a?(UtmLink) == false` (no NoMethodError); other `StatementInvalid` subclasses (Deadlocked etc.) propagate unchanged.
- **Routing is correct for every `save!`/`create!` in the block**: new UtmLink → retry; `UtmLinkVisit` → report-and-swallow (unchanged); persisted UtmLink timestamp save → `new_record?` false → report-and-swallow (a retry would fail identically — that's the #5989 flavor).
- **No double effects on retry**: everything including the `UpdateUtmLinkStatsJob` enqueue sits inside one transaction, the enqueue is last, and InnoDB raises uniqueness violations at statement time — a failed attempt rolls back completely.
- **ErrorNotifier fires exactly once per swallowed failure** (retry jumps back before the notify line).
- **Spec re-entrancy is bounded**: the winner-factory's own `save!` hits the stub with `raised_once` already true and falls through to the real save — one level, no recursion.
- **This concern is the only auto-create path** (wired once via `ApplicationController`; `SaveUtmLinkService` is the seller form path where errors should surface).
- **Winner-soft-deleted-before-retry is handled**: the alive-scoped retry lookup re-creates; a second failure hits the already-true gate → report-and-swallow.

### Only verifiable live
- The split of the 75k Sentry events between the new-record race flavors (fixed here) and the persisted-save-on-duplicates flavor (#5989) — verify via the Sentry event rate post-deploy.

Local evidence at head: 18/18 examples in `utm_link_tracking_spec.rb`, rubocop clean, both new specs verified red on main.

PR #5988, opened 2026-07-20
Fix NoMethodError in AffiliateMailer#direct_affiliate_invitation when affiliate or products are deleted

gumclaw:
## What

Guards `AffiliateMailer#direct_affiliate_invitation` against two nil cases that crash the delivery job: the `DirectAffiliate` no longer existing, and the affiliate having zero live products. In both cases the mailer now aborts delivery early (ActionMailer's standard NullMail abort) instead of raising `NoMethodError`.

## Why

The invitation email is enqueued with `deliver_later(wait: 3.seconds)`, so the affiliate — or every product attached to it — can be deleted between enqueue and delivery. When that happens the job crashes:

- `undefined method 'seller' for nil` (affiliate deleted; `find_by` returns nil)
- `undefined method '[]' for nil` (`enabled_products` is empty because all affiliated products are deleted, so `@products.first` is nil)

Sentry issue [GUMROAD 7367203621](https://gumroad-to.sentry.io/issues/7367203621/): 2,564 events since March 2026, still recurring (last seen 2026-07-20). Each occurrence is a failed Sidekiq `ActionMailer::MailDeliveryJob` that retries pointlessly — the recipient record is gone, so silently skipping delivery is the correct behavior.

No user-visible surface changes: this only affects a background mail job that was already failing (no email was being sent in these cases). Pure backend change — no adjacent UI to screenshot; the spec run below is the artifact.

## Test plan

- Added two regression specs reproducing both crash modes; both fail with the exact Sentry `NoMethodError`s on `main` and pass with the fix.
- Ran the full `#direct_affiliate_invitation` group locally: **7 examples, 0 failures**.
- Remaining failures in `spec/mailers/affiliate_mailer_spec.rb` (`#notify_affiliate_of_sale` balance_transaction setup) are pre-existing on unmodified `main` — verified on a clean tree.

## QA steps

Backend-only change to a background mail job — no user-facing surface to click through, so there is no preview-app click path. Reviewer verification:

1. Watch the walkthrough video below (diff + local spec run).
2. Optionally re-run locally: `bundle exec rspec spec/mailers/affiliate_mailer_spec.rb -e "#direct_affiliate_invitation"` — 7 examples, 0 failures.
3. After merge, confirm Sentry issue 7367203621 stops accruing events.

## Walkthrough video

Terminal walkthrough (diff + regression specs passing):

https://github.com/antiwork/gumroad/blob/fix/affiliate-invitation-nil-product/qa-media/pr-affiliate-invitation-nil-walkthrough.mp4?raw=true

### Test Results

```
AffiliateMailer#direct_affiliate_invitation
  when the affiliate has been deleted before the job runs does not send an email
  when the affiliate has no live products does not send an email
  ... (5 existing examples)

Finished in 1 minute 6.97 seconds (files took 7.18 seconds to load)
7 examples, 0 failures
```

`rubocop -a` on both files: no offenses.

---

🤖 AI disclosure: Authored by Claude Fable 5 (Hermes Agent) triggered by a Sentry alert webhook for issue 7367203621. Instructions: triage the Sentry error, reproduce with a failing spec, fix the root cause, verify locally, open a PR.


gumclaw:
Self-review notes:

- **Early-return abort is safe with the surrounding callbacks.** `ApplicationMailer#process` calls `set_custom_headers` after `super`, and that method already short-circuits on `ActionMailer::Base::NullMail` (line 40), so aborting before `mail params` doesn't break header injection. The `validate_from_email_domain!` after_action also returns early when `message.subject` is nil, which is the NullMail case.
- **`deliver_later` with a NullMail result is a no-op** — `NullMail#deliver` does nothing, so the enqueue site in `DirectAffiliate#send_invitation_email` needs no changes.
- **Why not fix at the enqueue site instead?** The race is inherent to the 3-second delayed delivery — the affiliate can be deleted after enqueue regardless of any check at enqueue time. Guarding in the mailer is the only place that closes it.
- **The second guard (`product.nil?`) covers the distinct case** where the affiliate row still exists but every affiliated product has been marked deleted (`enabled_products` merges `Link.alive`), which is the exact frame in the Sentry stacktrace (`product[:fee_percent]` on line 23).
- Both regression specs were verified RED on unmodified main (they reproduce the two `NoMethodError`s verbatim) before the fix turned them green.

PR #5991, opened 2026-07-20
Stop alerting Sentry on expected bank-account rejections during Stripe account creation

gumclaw:
## What

Stops reporting expected bank-account rejections to Sentry when Stripe rejects a seller's bank details during Connect account creation. The cleanup branch in `StripeMerchantAccountManager.create_account` now reuses the existing `bank_account_invalid_error?` classifier (already used by the bank-sync path, which treats these as expected) to skip `ErrorNotifier.notify` for seller-input errors like an unknown BIC / bank code, an invalid routing number, or an invalid account number. Genuinely unexpected failures still alert.

## Why

Sentry issue GUMROAD (`Stripe::InvalidRequestError: We couldn't find the bank for that BIC / bank code (BRSTN)`) has accumulated 6,445 events across 429 users since June 26 out of `Settings::PaymentsController#update`. Every one of them is a seller entering bank details that Stripe's bank directory can't resolve:

- The seller already gets the Stripe error message rendered inline on the payments settings page (`redirect_with_error(e.message)` in the controller).
- A retryable "bank sync failed" payout note is recorded on the account by the same rescue block, and the weekly retry job / support tooling picks it up.
- The sibling code path (`update_bank_account`) already classifies these exact errors as expected and does NOT call `ErrorNotifier.notify` for them.

Only the account-creation path was still paging Sentry, drowning the feed in noise that can never be fixed from a stacktrace. No behavior changes for the seller: the note recording, cleanup of the half-created merchant account, and the re-raise (which drives the inline error message) are all unchanged.

## Before / After

Not a UI change — error-reporting behavior only. The seller-facing inline error on /settings/payments is unchanged (it comes from the re-raised exception, which this PR does not touch).

## Test Results

New spec proves the fix is load-bearing: it fails on main (`ErrorNotifier` receives the rejection) and passes with the change. Full affected spec file green locally:

spec run

```
spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb
19 examples, 0 failures
```

## QA steps

Backend-only change with no adjacent UI surface — the seller-visible flow (inline Stripe error on /settings/payments) is driven by the unchanged re-raise. The spec run above is the artifact:

1. On this branch: `bundle exec rspec spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb` → 19 examples, 0 failures.
2. Revert the `stripe_merchant_account_manager.rb` hunk and re-run the "does not report the rejection to Sentry" example → it fails, confirming the guard is what suppresses the report.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes Agent / gumclaw) from an automated Sentry webhook run. Instructions given: triage the forwarded Sentry issue, identify root cause, add a failing test, fix, and open a PR with local spec evidence.


gumclaw:

- The guard reuses `bank_account_invalid_error?` rather than matching the "couldn't find the bank" message string. That classifier already covers `routing_number_invalid` / `account_number_invalid` codes, `Stripe::CardError`, and any `Stripe::InvalidRequestError` whose `param` starts with `bank_account`/`external_account` — the BRSTN case arrives as `param: "bank_account[routing_number]"`, so it's caught by the param branch. It deliberately excludes `incorrect_account_holder_name` (handled separately) and non-bank params like `individual[id_number]`, which continue to alert.
- The suppression is intentionally NOT gated on `notify:` — `notify: false` callers (admin retries) still shouldn't page Sentry for a seller's bad bank details.
- Cleanup (`cleanup_failed_merchant_account`) and the `raise` are untouched, so the half-created merchant account is still deleted and the controller still renders Stripe's message inline to the seller.
- New spec fails without the one-line guard (verified by stashing the app change and re-running), so it's load-bearing.

PR #5888, opened 2026-07-14
Render dashboard preview panes inside medium-matching chrome (browser for web, email for emails)

gumclaw:
## What

Wraps dashboard preview panes in shared chrome that matches the medium being previewed:

- Web previews use a centered title, quiet inset URL field, and compact open-in-new-tab action with a 44×44px hit area.
- Email previews use left-aligned From and Subject fields, with no URL or open action.
- Unsaved pages omit the URL and action.
- The chrome itself is not selectable; the actual preview content remains selectable.

The shared frame is used by Pages, product and bundle editors, the Receipt tab, Checkout, workflow emails, and profile settings. Preview frames no longer repeat a separate Preview heading or URL caption.

Checkout is intentionally different: it is labeled **Checkout preview** and has no URL or arrow because it renders a synthetic cart that does not exist at `/checkout`. Its layout now responds to the preview container, so the narrow embedded preview uses the real checkout's stacked layout while the full checkout still switches to two columns when there is enough room.

Receipt From, Subject, and body values come from the same Rails preview response, so unsaved client state cannot make the chrome disagree with the email body.

### Per-surface chrome

| Surface | Chrome | Identity | Open action |
| --- | --- | --- | --- |
| Pages editor | Web | Page title and public URL | Public page; omitted for unsaved pages |
| Product edit | Web | Product name and short URL | Save then open |
| Bundle edit | Web | Bundle name and short URL | Save then open |
| Receipt tab | Email | From and Subject | None |
| Checkout dashboard | Web | Checkout preview | None; the cart is synthetic |
| Workflow emails | Email | From; each email keeps its own subject heading | None |
| Profile settings | Web | Seller name and profile URL | Save then open |

## Why

This follows the first-class Pages work in #5812 and Sahil's direction to apply the same preview language to product pages and all other preview panes, while matching email previews to email rather than browser chrome.

The result keeps identity and actions in one place, aligns the preview with the edit form, and avoids false affordances. Only the arrow reacts on hover; the preview frame itself stays unchanged. No fake browser controls were added.

## Screenshots

All after screenshots were recaptured locally at the current branch head in a consistent 1600×1000 viewport.

### Rich text page

| Before | After (light) | After (dark) |
| --- | --- | --- |
| Before | Rich text light | Rich text dark |

### Home

| Light | Dark |
| --- | --- |
| Home light | Home dark |

### New page

| Light | Dark |
| --- | --- |
| New page light | New page dark |

### Product edit

| Light | Dark |
| --- | --- |
[...]

PR #5992, opened 2026-07-20
Follow dark mode in the rich text editor toolbar

gumclaw:
## What

The rich text editor toolbar now follows dark mode. The toolbar uses an inverted `bg-primary` bar, which is black in light mode but flips to near-white in dark mode — a bright strip across an otherwise dark editor. In dark mode it now uses the page background (`bg-body`, `#242423`), which sits slightly lighter than the black text field beneath it, with regular foreground text. Light mode is unchanged. The `--color-muted` override for toolbar icons now derives from `currentColor` so muted icons stay legible on both bars.

This PR originally also made the default-theme editor previews follow the dashboard's color scheme; that change was dropped after review — `#ffffff` is the seller's actual storefront background, so making the preview inherit dark mode would stop it matching what buyers see. Making default pages scheme-aware needs a broader theme model (custom background/text/accent/content colors per scheme) and is out of scope here.

## Why

Sahil flagged the toolbar on #5888 after merge: the description toolbar stayed light in dark mode ("Fix the tool bar too — should be dark bg too or just slightly lighter than the text field bg").

## Before / After

**Before (dark mode):** light toolbar strip across the dark editor — visible in the top area of before-dark

**After:** Screenshots pending — will capture light + dark of the description toolbar from the hosted preview app once the deploy is up.

## QA steps

1. Set the OS/browser to dark mode.
2. Open a product or bundle editor — the description toolbar should render dark, slightly lighter than the text field beneath it, with legible icons.
3. Switch to light mode — the toolbar is unchanged from before (inverted black bar).

## Test plan

- TSX-only change to `RichTextEditor.tsx`; no Ruby touched.
- Prettier + ESLint clean on the changed file.
- CI on this PR builds fresh assets and runs the full suite.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent) from Sahil's review comment on #5888 (make the rich text toolbar dark in dark mode, slightly lighter than the text field background). The preview-palette portion was removed after gianfrancopiana's review.

### DIFF
--- a/app/javascript/components/BundleEdit/Layout.tsx
+++ b/app/javascript/components/BundleEdit/Layout.tsx
@@ -92,12 +92,22 @@ export const BundleEditLayout = ({
         ? "Please wait..."
         : undefined;
 
+  // Pin the seller's storefront palette onto the preview only when they actually customized
+  // their background. The default white profile follows the dashboard's own color scheme
+  // instead, so a dark-mode dashboard doesn't frame a glaring white pane — the preview goes
+  // dark with the rest of the editor. The accent stays pinned either way: it works on both
+  // schemes and is the part of the seller's branding worth previewing.
+  const hasCustomBackground = currentSeller != null && currentSeller.profileBackgroundColor.toLowerCase() !== "#ffffff";
   const profileColors = currentSeller
     ? {
         "--accent": hexToRgb(currentSeller.profileHighlightColor),
         "--contrast-accent": hexToRgb(getContrastColor(currentSeller.profileHighlightColor)),
-        "--filled": hexToRgb(currentSeller.profileBackgroundColor),
-        "--color": hexToRgb(getContrastColor(currentSeller.profileBackgroundColor)),
+        ...(hasCustomBackground
+          ? {
+              "--filled": hexToRgb(currentSeller.profileBackgroundColor),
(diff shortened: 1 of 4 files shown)


gianfrancopiana:
The toolbar fix looks good to me, even if it’s unrelated to the preview changes. 

The preview change doesn’t make sense as written: `#ffffff` is the seller's storefront background, so making the preview inherit dark mode means it no longer matches what buyers see. This would only be correct if we also made default profile/product pages follow the visitor’s light/dark preference. But that requires a broader theme model, including deciding how custom background, text, accent, and content colors should behave in each scheme. I don't think we should treat white as “unset” in the preview alone. I'd keep the storefront preview accurate and split out the toolbar fix.

gumclaw:
Done — dropped the preview palette change and kept only the toolbar fix. `Layout.tsx` (product + bundle) and `Settings/Profile/Show.tsx` are reverted to main; the diff is now just `RichTextEditor.tsx`. Retitled the PR and updated the body to match.

Agreed on the reasoning: making default pages scheme-aware needs a real theme model (how custom background/text/accent/content colors map to each scheme), not a preview-only heuristic that treats white as unset. If we want to pursue that, it should start from the buyer-facing pages, with the preview following.

gumclaw:
Re-opened the surviving toolbar-only fix as a clean PR: #6050 (single commit on fresh main, `RichTextEditor.tsx` only).

PR #5994, opened 2026-07-20
Fix N+1 refund SUM and media location queries in mobile purchases endpoints

gumclaw:
## What

Fixes two per-purchase N+1 query shapes in the mobile purchases list/search endpoints (`Api::Mobile::PurchasesController#index` / `#search`), which serialize each purchase with `json_data_for_mobile`:

- `Purchase#amount_refunded_cents` issued one `SUM(refunds.amount_cents)` query per purchase. It now branches on `association(:refunds).loaded?`: when the endpoint has preloaded refunds, it sums in memory using `Refund#effective?` (the documented in-memory mirror of the `.effective` scope); otherwise it falls back to the existing DB-backed aggregate, so refund-processing write paths keep reading fresh state after bulk SQL writes.
- Each purchase's `url_redirect` issued its own `MediaLocation.max_consumed_at_by_file` query. A new batched scope, `MediaLocation.max_consumed_at_by_file_for_purchases`, loads the latest media location per (purchase, product file) pair for all purchases in one query, and `UrlRedirect.preload_latest_media_locations!` caches the result per instance. The controller preloads `:refunds` and calls the new preloader in `purchases_to_json`.

## Why

Sentry flags both shapes as recurring N+1 performance issues on `Api::Mobile::PurchasesController#search` in production:

- GUMROAD-XN (per-purchase `media_locations` max-consumed-at query) — 77 events / 26 users since June 29, still firing daily.
- GUMROAD-Z6 (per-purchase `SUM(refunds.amount_cents)` via `amount_refundable_in_currency`) — 8 events / 7 users since July 17.

With a page size of up to 100 purchases, each request issued up to 200 extra round trips. Both fixes follow the established `loaded?`-guarded pattern used elsewhere in the codebase so no caller loses correctness: write paths that never preload refunds fall through to the DB aggregate, and single-purchase serialization (`url_redirect_attributes`) keeps its existing per-purchase query untouched.

## Before / After

Backend-only change to mobile API JSON endpoints — no UI surface; response payloads are unchanged (the new controller specs assert response contents alongside the query counts, and the media-location scope spec asserts the batched scope returns exactly what the per-purchase scope returns for each purchase).

Before: N purchases → N `SUM(refunds...)` queries + N `media_locations ... purchase_id = ?` max-consumed-at queries.
After: refunds preloaded in one `IN (...)` query and summed in memory; one batched `media_locations ... purchase_id IN (...)` query for all purchases.

## QA steps

Backend-only mobile API change with no UI surface — QA is via the specs and the query-count assertions above. To verify manually on a console (preview or dev):

1. Pick a buyer with several purchases: `user = Purchase.where.not(purchaser_id: nil).last.purchaser`
2. Subscribe to `sql.active_record` and run the search serialization path: `purchases = user.purchases.for_mobile_listing.limit(10).to_a`, preload via `Api::Mobile::PurchasesController` (`GET /mobile/purchases/search` with a mobile OAuth token), or call the model path directly.
3. Confirm the log shows one `refunds ... purchase_id IN (...)` select and one `media_locations ... purchase_id IN (...)` select, and **zero** per-purchase `SUM(refunds...)` / `media_locations ... purchase_id = ?` queries.
4. Confirm response payloads are unchanged (`amount_refundable_in_currency`, `file_data[].latest_media_location`).

## Test Results

New regression specs (red on the old code, green with the fix — verified by stashing the app changes and re-running):

- `spec/controllers/api/mobile/purchases_controller_spec.rb` — "does not issue per-purchase refund SUM queries" and "does not issue per-purchase media location queries" (`sql.active_record` subscribers asserting `be_empty` on the per-row shapes, with fixtures that make each pattern reachable).
- `spec/models/media_location_spec.rb` — batched scope returns the most recent record per (purchase, file) pair, and matches `max_consumed_at_by_file` per purchase.
- `spec/models/purchase/purchase_refunds_spec.rb` — preloaded-refunds path returns the same `amount_refunded_cents` as the DB path (including reversed terminal-failure refunds) with zero SUM queries.

Ran locally:

```
bundle exec rspec spec/controllers/api/mobile/purchases_controller_spec.rb
56 examples, 0 failures

bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb
143 examples, 0 failures

bundle exec rspec spec/models/media_location_spec.rb (new scope examples)
+ targeted suite: 8 examples, 0 failures
```

spec run

(Pre-existing `media_location_spec.rb` `#create` failures are local S3-fixture environment issues that fail identically on unmodified origin/main.)

---

AI disclosure: Authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert. Instructions: triage the Sentry N+1 performance issues on `Api::Mobile::PurchasesController#search`, identify root cause, write failing regression specs, fix, and open a PR once local specs pass.


gumclaw:

- **`Purchase#amount_refunded_cents`** — the `loaded?` branch is the same pattern used in `prices_for_currency` and `with_product_files` (skill-canonical shape). The in-memory path uses `Refund#effective?`, which the model documents as the in-memory mirror of the `.effective` scope — if the scope ever changes, both need updating together (the model comment already says this). `.to_i` in the block mirrors SQL `SUM` semantics (NULL `amount_cents` rows contribute 0). All refund-processing write paths (`refund!`, chargeback handling) operate on freshly-loaded purchases where `refunds` is not preloaded, so they keep the DB-backed aggregate.
- **`MediaLocation.max_consumed_at_by_file_for_purchases`** — same join-on-MAX shape as the existing single-purchase scope, with `purchase_id` added to the subquery GROUP BY and join condition. The equivalence spec asserts it returns exactly what the single-purchase scope returns per purchase.
- **`UrlRedirect.preload_latest_media_locations!`** — only caches for redirects with a purchase and no installment, matching the guard inside `latest_media_locations_by_product_file_id`; ineligible redirects fall through to the existing per-instance behavior. The `defined?` memoization check means an explicitly-cached empty hash is honored (no re-query for purchases with zero media locations).
- **Tie-breaking caveat (pre-existing):** when two media locations share the same max `consumed_at` for a file, both scopes return both rows and `index_by` keeps the last one — identical behavior before and after, so serialized output can't change.
- **Not touched:** the single-purchase path (`url_redirect_attributes` → `product_json_data`) still uses the per-purchase scope — correct for a single record, and PR #5985 already covers its `email_infos` slice.

PR #5997, opened 2026-07-20
Reschedule SSL certificate jobs on Let's Encrypt Acme::Client::Error::RateLimited instead of exhausting retries

gumclaw:
## What

`GenerateSslCertificate` now rescues `Acme::Client::Error::RateLimited` and reschedules itself for after the Let's Encrypt rate-limit window resets (parsed from the error message, 1-hour fallback) plus up to 3 hours of jitter, instead of letting the error consume Sidekiq retries.

## Why

Sentry issue [7558536077](https://sentry.io/organizations/gumroad-to/issues/7558536077/) — "GenerateSslCertificate exhausted retries — SSL certificate not provisioned" — has fired **22,400+ times since June 20 (~1,500/day)**. Sampling the last 30 events: 29/30 carry `exception_class: Acme::Client::Error::RateLimited` ("too many new orders (300) from this account in the last 3h0m0s").

The rate limit is **account-wide**, not a per-domain failure, and Sidekiq's 5 default retries all fire within minutes — well inside the 3-hour window. So once the account hits the limit, every queued certificate job deterministically exhausts its retries and raises the #488 "stuck renewal" alert for perfectly healthy domains, which then never get their certificate until the next external re-trigger. Rescheduling past the reset keeps the jobs alive, spreads them across the next window (jitter avoids re-tripping the limit in one burst), and preserves the alert for real per-domain failures (e.g. `RejectedIdentifier`, which stays on the normal retry path).

Follow-up worth considering separately: the queue regularly demands more than 300 orders/3h, so throughput may need a Let's Encrypt rate-limit adjustment request or order batching — this PR stops the false alerts and the certificate-abandonment, not the backlog pressure.

## Before / After

Not a UI change — Sidekiq job behavior only; no user-facing surface is affected. Evidence is the spec run below (no adjacent UI flow exists for this background job).

## QA steps

Backend Sidekiq job only — no user-facing surface to click through. To verify on the preview app (https://pr-5997.apps.staging.gumroad.org, `x-revision` should show `839773e1`):

1. Open the preview Rails console.
2. Simulate a rate-limited run and confirm the job reschedules instead of raising:
   ```ruby
   domain = CustomDomain.certificate_absent.first
   # Stub the service to raise the rate-limit error, then run the job
   SslCertificates::Generate.define_method(:process) { raise Acme::Client::Error::RateLimited, "too many new orders (300) from this account in the last 3h0m0s, retry after #{(Time.current + 1.hour).utc.strftime('%Y-%m-%d %H:%M:%S')} UTC" }
   GenerateSslCertificate.new.perform(domain.id) # => no raise
   Sidekiq::ScheduledSet.new.select { |j| j.klass == "GenerateSslCertificate" && j.args.first == domain.id }
   # => one job scheduled after the parsed reset time (+ up to 3h jitter), args [domain.id, 1]
   ```
3. Confirm the log line `Rescheduling GenerateSslCertificate...` appears with domain, delay, and reschedule count.
4. Confirm a non-rate-limit error (e.g. raise `RuntimeError` from the stub) still propagates — normal retry/alert path intact.

## Test plan / Test Results

- New specs cover: reschedule past the parsed reset time, fallback delay when the message is unparseable, past-reset-time clamping to jitter-only, and non-rate-limit errors still propagating (retry + exhausted-retries alert preserved).
- Verified RED first: the new specs fail on unmodified code (error propagates / no reschedule).
- Local run (worktree of this branch):

```
bundle exec rspec spec/sidekiq/generate_ssl_certificate_spec.rb
7 examples, 0 failures
```

spec results

- Post-deploy verification: Sentry issue 7558536077's 24h event rate (currently ~1,560/day) should collapse to only genuine per-domain failures; re-check the rate after deploy before resolving the issue.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent), triggered by a Sentry event-alert webhook for issue 7558536077. Instructions: triage the Sentry alert, identify root cause, write a failing test, fix, and open a PR with local spec evidence.


gumclaw:
Self-review notes:

- **Why rescue in the job rather than in `SslCertificates::Generate`:** the service's `order_certificates` already has its own failure handling (reset `ssl_certificate_issued_at` + 8h domain-check cache), but `Acme::Client::Error::RateLimited` is raised from `LetsEncrypt#order_certificate` before any of that runs, and it propagates up through `process`. Handling it at the job layer keeps the reschedule semantics (a Sidekiq concern) out of the service and leaves the service's per-domain failure handling untouched.
- **Jitter window = 3 hours** deliberately matches the Let's Encrypt window (`rate_limit_hours` in `config/ssl_certificates.yml.erb`). When the limit trips there's typically a backlog of hundreds of queued jobs; spreading them across a full window avoids an immediate re-trip. Hardcoded rather than loading `SslCertificates::Base` config in the job to keep the job dependency-free — flagging in case reviewers prefer reading it from config.
- **Message parsing is defensive:** unparseable message → 1h fallback; past reset time → clamped to 0 (jitter only); `Time.zone.parse` `ArgumentError` → fallback. All three paths are specced.
- **`retry: 5` unchanged** — non-rate-limit ACME errors (e.g. `RejectedIdentifier`, seen in ~1/30 sampled events) still use normal retries and still fire the #488 exhausted-retries alert, which is the alert working as intended.
- One infinite-loop consideration: a domain stuck behind a permanently rate-limited account would reschedule forever without alerting. Accepted because the limit is account-wide and time-based (it always resets); a persistent backlog is a throughput problem called out in the PR body as follow-up, not something this per-job alert should page on per-domain.

gumclaw:
## Fable-5 adversarial pre-merge review — verdict at `880d6f7b`, fixes at `839773e1`

**Verdict: CHANGES-REQUIRED → all merge-blocking findings fixed in `839773e1`.**

| # | Finding | Severity | Disposition |
|---|---------|----------|-------------|
| P1.1 | Unbounded reschedule loop masks the #488 alert for **per-domain** rate limits. `Acme::Client::Error::RateLimited` isn't specific to the account-wide order limit — the gem (acme-client 2.0.18) maps the single class from `urn:ietf:params:acme:error:rateLimited`, which Boulder uses for *every* limit, including "too many certificates for this exact set of identifiers" (5/week). Concrete path: a domain with apex + www where www persistently fails → apex re-issued ~3×/day → trips the exact-set limit in ~2 days → every run raises RateLimited → reschedules forever, never alerting. On main that domain exhausts 5 retries and fires the #488 alert; the PR as reviewed silenced it permanently. | P1 | **Fixed in `839773e1`**: reschedule count threaded through job args (`perform(id, rate_limit_reschedules = 0)`); past `RATE_LIMIT_MAX_RESCHEDULES` (10) the error re-raises so the normal retry → exhausted-retries alert path takes over. New specs cover count increment and cap propagation; both red/green'd against the pre-fix code (2 failures pre-fix, pass post-fix). |
| P2.1 | Reschedule was completely silent — no log line for an operator debugging "why has this domain had no cert for 9 hours". | P2 | **Fixed in `839773e1`**: `Rails.logger.info` with domain, delay, reschedule count, and error message. |
| P3.1 | ~1/10,800 spec flake: when `rand(10800)` returns 0 in the past-reset-time case, `perform_in(0, id)` hits Sidekiq fake mode's `item["at"] = ts if ts > now` optimization → no `"at"` key → `NoMethodError` in the assertion. | P3 | **Fixed in `839773e1`**: jitter pinned via `allow_any_instance_of(...).to receive(:rand)`. |
| P3.2 | No dedup lock while the hourly `RenewCustomDomainSslCertificates` sweep keeps re-enqueuing domains that already have a rescheduled job pending — during an N-hour rate-limit event each unissued domain accumulates ~N pending jobs. Self-cleaning (duplicates exit via "Has valid certificate" after first success) and not materially worse than status quo. | P3 | **Suggestion, not fixed**: `sidekiq-unique-jobs` (8.0.10) is already in the bundle; `lock: :until_executed` on this worker would cap it. Left for a follow-up since it changes enqueue semantics for all callers. |
| P3.3 | Rescue comment claimed per-domain failures "still retry and alert as before" — false for per-domain *rate-limit* failures. | P3 | **Fixed in `839773e1`** (comment rewritten alongside P1.1). |

### Confirmed safe (adversarial probes that passed)

- **Error propagation**: `SslCertificates::Generate#process` has no rescue around `client.new_order` (where rate limits are raised) — `LetsEncrypt#process`'s internal `rescue => e; return false` only wraps finalize/upload, so `RateLimited` genuinely reaches the job. No optimistic state is written before the raise (`ssl_certificate_issued_at` only set after the full loop; ACME account creation idempotent), so re-running `process` is safe.
- **Time parsing**: app timezone is default UTC; the parsed string carries "UTC" so the instant is correct under any zone; garbage dates raise `ArgumentError` → rescued to the 1-hour fallback; past timestamps clamp to 0; `rate_limit_retry_delay` always returns an Integer.
- **Spec red-green**: all three original rate-limit specs fail on unmodified main (error propagates, `not_to raise_error` fails); `@obj_double` genuinely intercepts; `described_class.clear` correctly removes the model-callback job; `jobs.sole` guards against zero and duplicate enqueues.
- **Alert semantics**: `RateLimited` and `RejectedIdentifier` are sibling subclasses of `ServerError` — no hierarchy overlap, so the rescue catches exactly one class; everything else still walks the 5-retry ladder into `sidekiq_retries_exhausted`.
- **Sibling copies**: the only other ACME-error rescuer is the intentional `AccountDoesNotExist` handler; the renewal path funnels through this same job, so no second code path still exhausts retries on `RateLimited`.
- **Rolling-deploy skew**: the new job arg is defaulted (`= 0`), so old-arity jobs (including any already scheduled) drain cleanly against new code.

### Only live-verifiable

- Exact wording of current Boulder per-domain limit messages (whether all include a parseable "retry after … UTC") — affects only fallback-vs-parsed delay, not correctness.
- Whether staging and production share the Let's Encrypt account private key (`LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY`) — both use the production directory URL, which would make the 300/3h account limit shared across envs.
- Real-world frequency of per-domain `RateLimited` occurrences in production logs.

Full spec file green locally after fixes: `9 examples, 0 failures`. Rubocop clean.

PR #5999, opened 2026-07-20
Fix ChargeProcessorInvalidRequestError noise: treat FX-quote settlement-currency rejection as expected USD fallback

gumclaw:
## What

Maps Stripe's request-time FX-quote rejection ("The FX Quote's to_currency … must match the payment intent's settlement currency") to the existing `StripeFxQuote::SettlementCurrencyMismatch` error, and makes both quote callers — `Checkout::BuyerCurrencyQuote` and `Charge::MethodForcedPresentment` — fall back to the canonical USD path quietly for that condition, without an `ErrorNotifier` (Sentry) notification.

## Why

Sentry issue [GUMROAD 7621513904](https://gumroad-to.sentry.io/issues/7621513904/) — `ChargeProcessorInvalidRequestError` raised from `CustomerSurchargeController#calculate_all` (2 events / 2 users on 2026-07-20, both Canadian buyers).

Some connected accounts settle in a non-USD currency via Stripe multi-currency settlement, even though our stored `merchant_account.currency` says USD. `StripeFxQuote#build_quote` already handles the response-shaped variant of this (quote returned with a different `to_currency` → `SettlementCurrencyMismatch`), but Stripe can also reject the quote *request* outright with an `InvalidRequestError`. That path escaped as a generic `ChargeProcessorInvalidRequestError`, which the callers' blanket rescue reported to Sentry as if it were a defect.

The buyer-facing behavior was already correct — the quote service returns `nil` and checkout proceeds in USD. This PR only classifies the rejection as the expected condition it is, so it stops paging as an error while other genuine `InvalidRequestError`s still notify.

## Before / After

Not applicable — backend error-classification only, no rendered surface. The checkout behavior for affected buyers is identical before and after (USD fallback); the only change is that the expected condition no longer generates a Sentry event.

## Test Results

All affected specs pass locally:

```
$ bundle exec rspec spec/business/payments/charging/implementations/stripe/stripe_fx_quote_spec.rb
5 examples, 0 failures

$ bundle exec rspec spec/services/checkout/buyer_currency_quote_spec.rb spec/services/charge/method_forced_presentment_spec.rb
29 examples, 0 failures

$ bundle exec rubocop app/... spec/...
6 files inspected, no offenses detected
```

spec run

New specs:
- `StripeFxQuote` maps the request-time settlement-currency rejection to `SettlementCurrencyMismatch`, and re-raises other invalid-request errors unchanged.
- `Checkout::BuyerCurrencyQuote` and `Charge::MethodForcedPresentment` return `nil` on the mismatch without calling `ErrorNotifier.notify`.

## QA steps

Backend-only; no preview app needed. Reviewer verification path: run the three spec files above, and confirm on Sentry that the linked issue stops accruing events after deploy.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent, gumclaw), triggered automatically by the Sentry event-alert webhook for issue 7621513904. Instructions: triage the Sentry error, identify the root cause, write a failing test, fix, and open a small focused PR.


gumclaw:
Self-review notes:

- `StripeFxQuote#create`: the new `rescue ChargeProcessorInvalidRequestError` sits outside `with_stripe_error_handler`, so it catches only the wrapped error from the FX-quote request itself — card errors, rate limits, and API errors keep their existing mappings. Non-matching invalid-request errors are re-raised unchanged (covered by a spec).
- Message-based matching (`SETTLEMENT_MISMATCH_MESSAGE`) is the only option here: Stripe returns no distinct error `code` for this rejection on the preview FX Quotes endpoint (verified against the production Sentry event — `code` is absent). The regex is anchored on the stable phrase "must match the payment intent's settlement currency" and case-insensitive.
- Caller rescues are placed *before* the blanket `rescue StandardError` in both `Checkout::BuyerCurrencyQuote#create` and `Charge::MethodForcedPresentment#perform`, so the mismatch skips `ErrorNotifier` but every other failure still notifies exactly as before. Both still log at info level for traceability.
- No behavior change for buyers: both callers already returned `nil` (USD fallback) for this error — the only delta is Sentry classification.

gumclaw:
**Fable-5 adversarial review: READY-TO-MERGE** (reviewed at f0ac19855, no P1/P2 findings)

What was verified:
- **Red/green**: reverse-applied the app hunks and re-ran the new specs — both fail without the fix, 34/34 pass with it. The specs pin the behavior, not vacuously pass.
- **Rescue ordering**: `SettlementCurrencyMismatch` is rescued before the blanket `StandardError` in both callers, so the quiet-fallback path actually wins over the Sentry notify.
- **Caller completeness**: swept the repo — `StripeFxQuote.create` has exactly two callers (`Checkout::BuyerCurrencyQuote`, `Charge::MethodForcedPresentment`), both covered.
- **Fail-open direction**: the message-match regex is the one brittle piece (Stripe emits no distinct error code on the preview endpoint), but if Stripe rewords the message, behavior degrades back to Sentry noise — never to a broken checkout. The re-raise guard spec proves other invalid-request errors still notify.

P3 (non-blocking): if the FX Quotes endpoint graduates from preview and gains a typed error code, replace the message match — left as a comment-level note, no issue needed.

@rmarescu review + merge are yours; per the max-data ramp doctrine this un-pages the Canadian-buyer noise without touching buyer behavior (USD fallback was already correct).

gumclaw:
Deeper root-cause sweep done (audited reads 20260720T070953Z-4bd729f9, 20260720T071100Z-ff9affbb): **not stale data** — stored `merchant_account.currency` matches Stripe's live `default_currency` on every sampled account. The real mechanism is that multi-currency-settlement accounts can settle each *intent* in a currency that differs from their default, so no stored field can predict the rejection — Stripe's answer at quote time is authoritative, which is exactly what this PR encodes. This fix is architecturally correct, not a band-aid.

Population: ~600 candidate accounts (GB 246, CA 226, EU tails; the other ~2,900 usd-stored non-US accounts are dollarized economies). Their checkouts each burn a doomed FX-quote round trip before falling back — filed #6011 to persist the learned mismatch and skip the call at eligibility time. Nothing in that changes this PR.

PR #6000, opened 2026-07-20
Stop paging Sentry when all images fail moderation but text was moderated (ClassifierStrategy noise)

gumclaw:
## What

`ContentModeration::Strategies::ClassifierStrategy` no longer sends a summary "could not moderate any image" error to Sentry when the product's **text was successfully moderated** and only the images were rejected by OpenAI. That case now logs a `Rails.logger.warn` and continues with the text-only result, exactly as before. The Sentry report (and the blocking "try again later" result) is kept for the genuinely bad case: no text and no image could be moderated, so the content got zero evaluation.

## Why

Sentry issue `ContentModeration::ClassifierStrategy could not moderate any image` has **859 events in the last 30 days** and is still firing multiple times an hour. The dominant pattern in the events: OpenAI returns a 400 (`invalid_image_url` / could not download) for a signed `files.gumroad.com` attachment URL — an expected, recurring upstream condition. In these events the text moderation succeeded and the flow continued normally, so each Sentry event was pure noise with no action to take:

- Per-image failures that exhaust retries are already reported **individually** inside `#moderate` (`ErrorNotifier.notify(e, input_type: "image_url", …)`), so the transient-infra visibility is unchanged.
- The summary report only carries unique signal in the zero-coverage case (no text, all images failed) — which blocks the seller with `UNAVAILABLE_REASON`. That path still reports to Sentry so a spike stays visible.

This mirrors PR #5984, which made the same noise-vs-signal split for `PromptStrategy`.

## Before / After

Not a UI change — moderation service internals only. No user-visible behavior changes at all: the returned `Result` is identical in every branch; only the Sentry-vs-log destination of the summary changes. Behavior is asserted by the spec run below.

## QA steps

Backend-only service change — no user-visible surface or preview click-path; the strategy runs inside publish/upload flows.

1. Check out the branch and run `bundle exec rspec spec/services/content_moderation/strategies/classifier_strategy_spec.rb` (22 examples, 0 failures).
2. After deploy, watch Sentry issue 7440278724 — event rate should drop to near zero; remaining events indicate the zero-coverage (no text, all images failed) case worth individual triage.

## Test plan

- New spec: `logs a warning instead of notifying Sentry when text was moderated and OpenAI rejects every image as unmoderatable` — asserts compliant result, no `ErrorNotifier.notify`, and the warning log. Verified RED against main (fails before the fix) and GREEN with it.
- Updated spec: the every-image-fails-with-text case now asserts NO summary Sentry report, that the per-image retry-exhaustion reports still fire (`input_type: "image_url"`, twice), and the new warning log.
- Unchanged spec: the no-text/all-images-failed case still asserts the Sentry report and the blocking `flagged`/`UNAVAILABLE_REASON` result.
- Full file run locally: `bundle exec rspec spec/services/content_moderation/strategies/classifier_strategy_spec.rb` → **22 examples, 0 failures**.

## Test Results

spec results

```
Finished in 3.52 seconds (files took 3 seconds to load)
22 examples, 0 failures
```

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) in response to a Sentry alert webhook for issue 7440278724. Instructions: triage the Sentry event, verify recurrence, identify root cause, write a failing test, fix, verify locally, and open a PR.


gumclaw:
**Self-review**

- `classifier_strategy.rb`: the change is confined to the `moderated_count == 0` summary block. The two outcomes (`compliant` continue-with-text vs blocking `flagged` + Sentry) are byte-identical to main's behavior — only the *reporting destination* of the text-moderated case changes (Sentry → `Rails.logger.warn`). The `return` in the else-branch matches main's `return ... unless text_moderated` exactly.
- Per-image visibility is preserved: retry-exhausted transient failures (`Faraday::TimeoutError/ConnectionFailed/ParsingError/ServerError`) still hit `ErrorNotifier.notify` individually inside `#moderate` with `input_type: "image_url"` — asserted explicitly in the updated spec (`.twice`). OpenAI 400 rejections were already log-only per image on main (`skipping unmoderatable image URL=...`), so no signal is lost there either.
- The warn message includes `skipped_urls.size`/`@image_urls.size` so log-based alerting can still catch a systemic URL-building regression.
- Adversarial pass (fable-style, run locally before opening): checked (1) no behavior change to the seller-blocking path — the flagged/UNAVAILABLE_REASON branch is untouched and still covered by the `no text` spec; (2) `skipped_urls` can be smaller than `@image_urls` only via the `MAX_IMAGES_TO_MODERATE` break, in which case `moderated_count > 0` and this block never runs, so the count ratio in the log is always accurate; (3) no PII beyond what main already logged (signed URLs were already in the skipped-URL warn path). No P1/P2 findings.

PR #6003, opened 2026-07-20
Document product-level refund policy params in API v2 reference

gumclaw:
## What

Documents the product-level refund policy support added to API v2 in #6001, in the API reference (`app/javascript/components/ApiDocumentation`):

- `POST /v2/products` and `PUT /v2/products/:id` parameter lists gain `refund_period` (`inherit`, `none`, `7`, `14`, `30`, `183`) and `refund_fine_print`, including the account-level-policy availability gate and the "requires refund_period unless already enabled" rule.
- The single-product response field list (`PRODUCT_FIELDS`) gains the `refund_policy` object with its `refund_period` / `title` / `fine_print` / `inherited` children. It is not added to `PRODUCT_LIST_FIELDS` because the field is only serialized on non-slim (show/create/update) responses, not `GET /v2/products` list responses.
- `PUT /v2/refund_policy`'s description now points sellers whose account-level policy is not in effect at the product-level params instead of dead-ending.

## Why

#6001 shipped the params without updating the API reference, so the endpoint docs listed every other product param but not these. Wording is taken from the merged controller code: allowed values from `Api::V2::LinksController::PRODUCT_REFUND_PERIOD_ALLOWED_VALUES`, the availability gate from `validate_product_refund_policy_params` (`account_level_refund_policy_enabled?` check), and the response shape from `Product::AsJson#product_refund_policy_api_json`.

## Before / After

Docs-only change to the developer API reference page. New reference content (highlighted rows are the additions):

api docs refund policy additions

## QA steps

1. Open the API documentation page (`/api`) on the preview app.
2. Scroll to `POST /v2/products` and `PUT /v2/products/:id` — both parameter lists show `refund_period` and `refund_fine_print`.
3. In the single-product response fields, `refund_policy` appears with its four child fields.
4. `PUT /v2/refund_policy`'s description mentions the product-level params as the alternative when the account-level policy is not in effect.

## Test Results

No logic changes — static documentation content only (`ApiParameter` / `FieldDefinition` entries and one description string). Prettier clean; ESLint clean on all three touched files.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent), triggered automatically by the merge of #6001. The agent read the merged controller/serializer code and mirrored its allowed values, gating, and response shape into the API reference components.


gumclaw:
Adversarial review (fable-5, pre-merge) — docs-content-only diff, every claim checked against the merged #6001 code:

- Allowed values `inherit, none, 7, 14, 30, 183` match `PRODUCT_REFUND_PERIOD_ALLOWED_VALUES` (`["inherit"] + REFUND_PERIOD_ALLOWED_VALUES`).
- Availability gate wording matches `validate_product_refund_policy_params` (`account_level_refund_policy_enabled?` → rejected with a pointer to `PUT /v2/refund_policy`).
- Response shape (`refund_period` / `title` / `fine_print` / `inherited`, with `"inherit"` + nulls when the override is off) matches `Product::AsJson#product_refund_policy_api_json`.
- Deliberately NOT added to `PRODUCT_LIST_FIELDS`: the field is only serialized on non-slim responses, so `GET /v2/products` list items don't include it.
- P2 found and fixed in 55b778b: the initial `refund_fine_print` description omitted the validation that it can't be combined with `refund_period: "inherit"` — added.
- Prettier + ESLint clean on all three files.

No P1s. Verdict: ship.

PR #6004, opened 2026-07-20
Rescue Rack::Multipart::EmptyContentError in params-touching Rack::Attack throttles

gumclaw:
## What

Adds `Rack::Multipart::EmptyContentError` to the rescue clauses of the Rack::Attack throttle blocks that read `req.params`: the shared `throttle_with_exponential_backoff` wrapper (which backs every `throttle_by_ip` / `throttle_by_params` / `throttle_by_ip_and_params` rule), the `oauth_device_token` throttle, the `/api/v2/sales` pagination throttle, and the `logins/login` throttle. Malformed multipart requests now skip the rule instead of crashing the middleware.

## Why

Sentry issue [GUMROAD `Rack::Multipart::EmptyContentError` on `/ajax.php`](https://gumroad-to.sentry.io/issues/7529102928/) — 19,120 events / 2,035 IPs since June 5. Scanners POST `Content-Type: multipart/form-data` with an empty body and no `Content-Length` to arbitrary paths. When a params-touching throttle block runs, `req.params` raises `Rack::Multipart::EmptyContentError` inside the Rack::Attack middleware and the request 500s before reaching the app.

#5414 rescued this error in the `invalid_params` throttle, but the other params-touching blocks were left unguarded, so the error still escapes through them (production stacktrace points at the `throttle_with_exponential_backoff` block → `req.params`). This closes the remaining paths; the `invalid_params` throttle continues to rate-limit these senders.

Reproduced locally:

```ruby
env = Rack::MockRequest.env_for("/follow", method: "POST",
  "CONTENT_TYPE" => "multipart/form-data; boundary=AaB03x", input: "")
env.delete("CONTENT_LENGTH")
Rack::Attack.configuration.throttled?(Rack::Attack::Request.new(env))
# before: raises Rack::Multipart::EmptyContentError; after: returns cleanly
```

## Before/After

Not a UI change — Rack middleware only. Malformed-body requests previously got a 500 from the middleware; they now pass through to the app (which returns its normal 404/422 for these scanner paths).

## Test Results

New regression specs fail on main (all three raise `Rack::Multipart::EmptyContentError`) and pass with the fix. Full `spec/requests/rack_attack_spec.rb` run: **19 examples, 0 failures**.

spec results

## QA steps

1. Check out the branch, `RAILS_ENV=test bundle exec rspec spec/requests/rack_attack_spec.rb` → 19 examples, 0 failures.
2. Optional: in a console, run the repro snippet above and confirm no exception.

---

**AI disclosure:** Authored by Claude Fable 5 (Gumclaw) from a Sentry webhook alert. Instructions: triage the Sentry issue, reproduce the failure locally, write a failing regression test, fix the root cause, and open a PR with the results.


gumclaw:
Self-review notes:

- `throttle_with_exponential_backoff` is the single shared wrapper behind every `throttle_by_ip*` / `throttle_by_params` / `throttle_by_ip_and_params` rule, so one rescue there covers all rules whose discriminator procs call `req.params` (`/follow`, `/two-factor/*`, the v2 per-token throttles, etc.). The production stacktrace for the Sentry issue points exactly at this wrapper (`rack_attack.rb:55` → `rack/request.rb params`).
- Three blocks bypass the wrapper and needed their own rescue: `oauth_device_token` (already rescued `InvalidParameterError`/`TypeError`, extended), `/api/v2/sales`, and `logins/login` (both previously had no rescue at all).
- `throttle_by_ip_for_period` and the oauth path-only throttles never touch `req.params`, so they're intentionally untouched.
- Rescuing `EmptyContentError` specifically (not `EOFError` broadly) keeps genuine content-length-mismatch errors (`EOFError, "bad content body"`) visible; those surface with a `Content-Length` present and were not the Sentry signature (no `Content-Length` → `EmptyContentError` at parser init).
- Regression specs assert `Rack::Attack.configuration.throttled?` doesn't raise for a wrapper-based rule, a standalone-rescued rule, and the sales rule; all three failed on main before the fix.

gumclaw:
**fable-5 pre-merge adversarial review — verdict: ✅ no P1/P2**

Hunt list & findings:

1. **Guard-fork completeness** (the #5414 failure mode — fixing one gate, missing siblings): swept every `req.params` / `req.POST` / `req.GET` callsite in the initializer at HEAD. All are now covered: lines 227–274, 400, 440 sit inside procs executed by `throttle_with_exponential_backoff` (rescue added there); line 115 is the `invalid_params` throttle (already rescued since #5414); lines 319 (`/api/v2/sales`), 335 (`logins/login`), 208–209 (`oauth_device_token`) each carry their own rescue in this PR. `throttle_by_ip_for_period` and the oauth path-only blocks read no params — correctly untouched.
2. **Over-broad rescue risk**: rescues `Rack::Multipart::EmptyContentError` only, not its parent `EOFError` — content-length-mismatch (`"bad content body"`) errors remain visible. P3-level note only.
3. **Security regression risk** (does skipping the rule open a throttle bypass?): a malformed multipart body can't carry valid form params anyway, so the discriminator would have been nil-equivalent; the sender is still counted by the `invalid_params` throttle. No bypass.
4. **Test honesty**: regression specs exercise `Rack::Attack.configuration.throttled?` (the exact production entry point from the stacktrace), fail on main (verified via stash: 3/3 raise), pass on the branch; full spec file green (19 examples, 0 failures).

P3 (non-blocking): the three standalone rescues duplicate the exception list — a shared constant would reduce drift risk if a fourth parse-error class ever appears. Not worth blocking a middleware crash fix.

PR #6005, opened 2026-07-20
Discard ActiveStorage::PreviewImageJob on ActiveStorage::PreviewError

gumclaw:
## What

Configures `ActiveStorage::PreviewImageJob` to discard (instead of retry) when the previewer raises `ActiveStorage::PreviewError`, logging a single warning. Extends the existing `config/initializers/active_storage_jobs.rb` initializer that already does the same for `AnalyzeJob` + S3 `NoSuchKey`.

## Why

When an uploaded file is corrupt or uses a codec the production ffmpeg build can't decode (e.g. a JPEG with no decodable JPEG data, or newer video codecs), preview generation fails deterministically — every retry produces the identical ffmpeg failure. The stock job only discards on `RecordNotFound`/`UnrepresentableError`, so `PreviewError` bubbles up as an unhandled Sidekiq error on each attempt. One Sentry issue for this has accumulated **435 events since March** (`ActiveStorage::PreviewError: ffmpeg failed (status 1)`, culprit `ActiveStorage::PreviewImageJob`), all unactionable noise.

Retry check: the failure is input-determined, not transient — no retry policy can fix an undecodable file, so discard-with-warning is the correct handling. Upgrading ffmpeg would widen codec support but can never eliminate corrupt uploads.

Refs Sentry issue GUMROAD 7367205063. Supersedes stale-closed #4092 (same fix, closed by the stale bot before review).

## Before / After

No user impact — preview generation already failed for these files and continues to fail; the change only stops the pointless retries and per-attempt error reports. Pure backend/job-config change with no adjacent UI surface; the spec run + terminal walkthrough below are the evidence.

**Walkthrough video:** https://github.com/antiwork/gumroad/blob/fix/discard-preview-image-job-on-preview-error/qa-media/pr-6005-preview-error-discard.mp4?raw=true

## Test Results

`bundle exec rspec spec/config/initializers/active_storage_jobs_spec.rb`

```
ActiveStorage::AnalyzeJob error handling
  discards the job when S3 returns NoSuchKey
ActiveStorage::PreviewImageJob error handling
  discards the job when the previewer raises ActiveStorage::PreviewError
  does not re-raise and logs a warning when PreviewError is raised

3 examples, 0 failures
```

Red/green verified: with the initializer change stashed, both new examples fail (job re-raises `PreviewError`); with it applied, all green.

## QA steps

1. In a Rails console: `ActiveStorage::PreviewImageJob.rescue_handlers` → includes `["ActiveStorage::PreviewError", ...]`.
2. Upload a corrupt image/video to any product; the preview job completes (discarded) with a `[ActiveStorage] Discarding PreviewImageJob (unpreviewable file): ...` warning in the logs instead of a raised error and retries.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) from a Sentry event-alert webhook. Instructions: triage the recurring `ActiveStorage::PreviewError` Sentry issue, root-cause it, add a failing test, fix, and open a PR with the standard evidence.


gumclaw:
**Self-review notes:**

- Placed alongside the existing `AnalyzeJob.discard_on(Aws::S3::Errors::NoSuchKey)` in `config/initializers/active_storage_jobs.rb` rather than a new initializer — same concern (ActiveStorage job error handling), one file.
- `error.message.lines.first&.strip` in the log: the raw `PreviewError` message embeds the full multi-line ffmpeg banner (~40 lines); only the first line (`ffmpeg failed (status 1): ...`) is useful in logs.
- Kept the stock `retry_on ActiveStorage::IntegrityError` untouched — integrity errors ARE transient (S3 read races) and should still retry. Only `PreviewError` (deterministic decode failure) is discarded.
- `after_initialize` guard matches the existing pattern so the constant is loaded before we patch it.

**Adversarial review (fable-5, pre-merge gate):** hunted for (1) discard being too broad — could `PreviewError` ever be transient? Checked Rails 7.2 source: `PreviewError` is raised only from previewer `draw` failures (ffmpeg/mutool exit non-zero on the downloaded file); the input file is immutable once uploaded, so the failure is deterministic per blob. (2) `variations` loop skipped on discard — yes, preprocessed variants won't generate for the broken blob, but they couldn't anyway (same decode path). (3) Swallowing real regressions — an ffmpeg binary breakage would silence errors; mitigated by the warning log, and such a breakage would surface loudly elsewhere (video transcoding). No P1/P2 findings.

PR #6007, opened 2026-07-20
Fix NoMethodError (undefined method 'external_id' for nil) in CustomerMailer#grouped_receipt

gumclaw:
## What

`CustomerMailer#grouped_receipt` now filters the passed purchase ids down to successful states (`all_success_states_including_test`) before building chargeables, and returns a `NullMail` when nothing remains after filtering.

## Why

Production crash — `ActionView::Template::Error: undefined method 'external_id' for nil` (Sentry issue 7367404427, ~1,170 events since March 2026, unresolved).

Root cause: callers of `grouped_receipt` pass unfiltered purchase id sets. `Purchase::ReassignByEmailService` reassigns purchases in **any** state (including `failed`) and mails a grouped receipt for all reassigned ids. A failed purchase that belongs to a `Charge` resolves (via `Charge::Chargeable.find_by_purchase_or_charge!` → `uses_charge_receipt?`) to a Charge whose `successful_purchases` is empty, so `Charge#purchase_as_chargeable` returns `nil` and the receipt template crashes in `ReceiptPresenter::ChargeInfo#order_id` → `Charge#external_id_for_invoice` (`purchase_as_chargeable.external_id`).

Verified against production data: the failing job's id set included a `failed` purchase attached to a single-purchase Charge — exactly the shape above.

Receipts are only ever meaningful for successful purchases (`Charge#successful_purchases` gates every other receipt path), so filtering at the mailer covers all three call sites (email reassignment, admin resend-all-receipts, public charge lookup) uniformly.

## Test plan

Added two specs to `spec/mailers/customer_mailer_spec.rb#grouped_receipt`:
- a failed charge-backed purchase mixed into a valid set no longer crashes body rendering (reproduced the exact `NoMethodError` before the fix — RED without fix, GREEN with)
- an id set with no successful purchases produces a `NullMail` instead of a broken email

Ran locally:
- `spec/mailers/customer_mailer_spec.rb -e "#grouped_receipt"` — 4 examples, 0 failures
- `spec/services/purchase/reassign_by_email_service_spec.rb` + `spec/controllers/public_controller_spec.rb` (caller coverage) — 41 examples, 0 failures
- `rubocop -a` on both changed files — no offenses

## QA steps

No user-facing surface — this fixes a background mailer job crash, so there is nothing to click through on a preview app. To verify the behavior on any environment console:

1. `mail = CustomerMailer.grouped_receipt([<successful_purchase_id>, <failed_charge_backed_purchase_id>])`
2. `mail.message.to_s` — renders without raising (crashed with `NoMethodError: undefined method 'external_id' for nil` before this fix)
3. `CustomerMailer.grouped_receipt([<failed_purchase_id>]).message` — returns `ActionMailer::Base::NullMail` (no email sent)

## Test Results

spec results

Full `#grouped_receipt` spec run — 4 examples / 0 failures.

## Before/After

No user-visible surface changed — this is a mailer crash fix. Before: the grouped-receipt job raised and the customer received nothing (job retried 5x, then dead). After: the receipt renders for the successful purchases and skips the failed ones. No adjacent UI affected; the spec run + rendered-body assertion is the artifact (pure backend/mailer path).

---

AI disclosure: Authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert. Instructions: triage the Sentry issue, reproduce the root cause, write a failing test, fix, verify locally, open a PR per repo conventions.


gumclaw:
Self-review + adversarial pass (pre-merge review, fable-5):

**Diff walkthrough**
- The filter uses `all_success_states_including_test`, which is the same state set `Charge#successful_purchases` uses — so the mailer's inclusion criteria now exactly match what the receipt template can render. `not_charged` (subscription upgrades) and gift-receiver purchases remain included, same as every other receipt path.
- Redundancy check: a chargeable can only enter `@chargeables` via a successful purchase now, so a post-hoc `select { _1.successful_purchases.exists? }` guard would be dead code — deliberately omitted.
- Mixed charge case: if a failed purchase's Charge also contains a successful purchase that is in the id list, the charge still gets a receipt via the successful sibling — correct and unchanged.
- Early `return` on an empty set produces `ActionMailer::Base::NullMail`; `deliver_later`/`deliver_now` on a NullMail is a no-op, so callers (all three fire-and-forget) need no changes.
- N+1 eager-load spec unchanged and still green (`includes(charge: [:order, :seller])` preserved).

**Hunted for and did not find**
- P1: behavior change for currently-working inputs — none; every id set that rendered before still renders identically (filter only removes ids that crashed the job).
- P2: callers relying on a receipt for non-successful purchases — grepped all 3 call sites; `resend_all_receipts` already scopes `.successful`, `charge_data` scopes `successful_gift_or_nongift`, only `ReassignByEmailService` passed unscoped ids (the crashing path).

Verdict: no P1/P2 findings. Local runs: mailer specs 4/4, caller specs 41/41, rubocop clean.

PR #6008, opened 2026-07-20
Stop alerting Sentry on expected phone-number rejections during Stripe account creation

gumclaw:
## What

Stops reporting expected phone-number rejections from Stripe connected-account creation to Sentry. Adds a `phone_number_invalid_error?` predicate to `StripeMerchantAccountManager` and skips `ErrorNotifier.notify` for those errors in `create_account`'s rescue, mirroring the existing bank-account-rejection suppression (#5991). The error still re-raises, so the seller keeps seeing Stripe's message inline on the payments settings page.

## Why

Sentry issue GUMROAD `7369895967` (`Stripe::InvalidRequestError: "+…" is not a valid phone number`, culprit `Settings::PaymentsController#update`) has accumulated 429 events across 93 users since March 2026. Each event is the same expected seller-input error: the client-side formatter only normalizes phone input to E.164 (`libphonenumber-js` `format("E.164") ?? raw`) without verifying the number is real, so Stripe rejects malformed numbers at account creation. The controller already rescues `Stripe::StripeError` and renders the message back to the seller — the seller self-corrects; nothing is actionable for us per occurrence.

Fixability triage per the no-blanket-classes rule: RETRY doesn't apply (deterministic input error, retrying can't help); UX is already handled (inline error rendered via the existing rescue in `Settings::PaymentsController#update`); NOISE is the defect — per-occurrence Sentry capture of an expected, handled seller-input failure — fixed here at the same boundary as the bank-rejection suppression; AGGREGATE signal is preserved since genuinely unexpected `create_account` failures still notify.

Stripe does not always populate `code`/`param` on this rejection (the captured events carry neither), so the predicate matches a phone `param` when present and otherwise the stable message shape, with a comment explaining why.

## Before/After

Not a UI change — error-reporting behavior only. The seller-facing inline error on `/settings/payments` is unchanged (the exception still re-raises and the controller's existing rescue renders Stripe's message exactly as before).

## Test plan / Test Results

New specs in `stripe_merchant_account_manager_payout_retry_spec.rb`:
- phone-message-shaped `Stripe::InvalidRequestError` → not reported to Sentry, still re-raises, no bank-sync payout note recorded
- phone `param` variant (`individual[phone]`) without the message shape → not reported to Sentry

Verified the new specs are load-bearing: with the app change stashed, both fail (`ErrorNotifier` received `:notify`); with it applied, the full file passes.

spec run

```
22 examples, 0 failures
```

## QA steps

Pure backend error-reporting change with no adjacent user-visible surface — the spec run above is the artifact. To exercise manually: on the payments settings page, submit compliance info with a malformed phone number; the inline Stripe error still appears, and no new Sentry event is recorded for the rejection.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes agent) from a Sentry webhook for issue 7369895967. Instructions: triage the recurring Stripe phone-number rejection noise on `Settings::PaymentsController#update`, and if it is expected/handled seller input, suppress the per-occurrence Sentry capture with tests, following the existing bank-rejection suppression pattern.


gumclaw:

- **Predicate placement/ordering**: `phone_number_invalid_error?` sits next to `postal_code_invalid_error?` / `bank_account_invalid_error?` and follows the same `private_class_method` + defensive `respond_to?` style.
- **Match strategy**: the captured production events carry neither `code` nor `param` on the `Stripe::InvalidRequestError`, so the predicate falls back to the stable message shape (`is not a valid phone number`). Param matching (`individual[phone]`, `business_profile[support_phone]` → last bracket segment `phone`/`support_phone`) is kept first so a future Stripe payload that does populate `param` is still caught even if the message wording changes.
- **Scope of suppression**: only the `create_account` rescue's `ErrorNotifier.notify` is gated. The exception still re-raises, `cleanup_failed_merchant_account` still runs, and the controller's existing rescue still renders Stripe's message to the seller. No payout note is recorded for phone rejections (unlike bank rejections) because there is no retry job for phone numbers — the seller corrects it on the spot.
- **Blast-radius check**: `phone_number_invalid_error?` requires `Stripe::InvalidRequestError`, so a `Stripe::APIError`/timeout whose message coincidentally mentioned phone numbers would still notify. The message regex is anchored to Stripe's exact rejection phrasing.
- **Tests**: both new Sentry-suppression examples were verified load-bearing (red with the app change stashed, green with it applied); the pre-existing 20 examples in the file still pass.

gumclaw:
Additional Sentry group covered by this fix: `7578121692` (same signature — `Stripe::InvalidRequestError: "+…" is not a valid phone number`, culprit `Settings::PaymentsController#update` → `StripeMerchantAccountManager.create_account`), currently at 6,490 events / 433 users since 2026-06-26. Sentry split it from `7369895967` because the raw phone value is embedded in the title. The suppression here resolves both groups; resolve both in Sentry after this ships.

gumclaw:
Sentry alert fired again for a sibling group of the same error: issue `7578121692` — identical title/culprit (`Settings::PaymentsController#update` → `StripeMerchantAccountManager.create_account`, stripe_merchant_account_manager.rb:120), 6,574 events / 437 users since 2026-06-26. Same suppression boundary as this PR, so merging resolves both groups. Noting here to avoid a duplicate PR.

gumclaw:
Merged in the latest main to clear the conflict. Two files clashed with the Japanese-address Sentry-suppression change that landed on main: in the Stripe merchant account manager I combined both guards so the notifier now skips bank, phone-number, and JP-address rejections alike (comment updated to cover all three), and in the payout retry spec I kept both sides' test contexts. Targeted specs pass (25 examples, 0 failures) and rubocop is clean.

gumclaw:

PR #6012, opened 2026-07-20
Grant fullscreen permission to embed iframes at sanitize time

gumclaw:
## What

Grants fullscreen permission to every embed iframe at sanitize time. A DOMPurify `afterSanitizeAttributes` hook in `sanitizeHtml` (`app/javascript/utils/sanitize.ts`) now sets `allowfullscreen` and appends `fullscreen *;` to the `allow` list on any `<iframe>` that lacks it. Adds a test suite for `sanitizeHtml` covering the hook and the existing sanitizer behavior.

## Why

Vimeo removes its fullscreen button entirely when the embedding iframe doesn't grant fullscreen permission (the page inside the frame sees `document.fullscreenEnabled === false`). #5663 made the client sanitizer *preserve* `allowfullscreen`, but the stored embed HTML for unlisted Vimeo links (`vimeo.com/<id>/<hash>`) is an iframely proxy iframe whose `allow` list is only `encrypted-media *;` — there is no fullscreen permission to preserve, so buyers still can't fullscreen videos in product descriptions.

Normalizing at sanitize time fixes every already-stored embed without requiring sellers to re-insert their videos, and mirrors what cover-carousel embeds (`app/javascript/components/Product/Covers/Embed.tsx`) already do by hardcoding `allowFullScreen` — which is why the same video fullscreens fine in the top carousel but not in the description. The server-side sanitizer allowlist (`app/models/link.rb`) already permits both attributes, so no backend change is needed.

Fixes antiwork/gumroad-private#1187

## Before / After

Same stored embed HTML (the affected seller's actual unlisted Vimeo embed), rendered as-is vs. through the updated `sanitizeHtml()`. Before: no fullscreen button (only volume + settings). After: fullscreen button restored.

before-after

## QA steps

1. Open any product whose description contains an unlisted Vimeo embed (rich-content `tiptap__raw` node whose stored iframe `allow` lacks `fullscreen`).
2. Inspect the rendered description iframe: it should now carry `allowfullscreen` and `allow="…; fullscreen *;"`.
3. Play the video — the fullscreen button appears in the Vimeo control bar and works.
4. Confirm a YouTube/public-Vimeo embed (direct provider iframe) still renders and fullscreens as before.

## Test results

```
✓ app/javascript/utils/sanitize.test.ts (7 tests) 11ms

 Test Files  1 passed (1)
      Tests  7 passed (7)
```

`vitest run app/javascript/utils/sanitize.test.ts` — 7/7 passing locally. `prettier` and `eslint` clean on both touched files.

---

## AI disclosure

Written by Claude Fable 5 (Hermes agent). Instructions: implement the fix proposed on antiwork/gumroad-private#1187 (normalize iframe fullscreen permission at sanitize time), add tests, and provide before/after evidence — approved by Sahil on the issue.


gumclaw:
Self-review notes:

- The hook is registered once, globally, on DOMPurify's singleton — it therefore applies to every `sanitizeHtml` call (both `Raw.renderHTML` and `mediaEmbed`'s node view). That's intentional: every current caller renders embed media, and granting fullscreen to an embed iframe is safe (it's a permission the embedded page can use, not an escalation for the host page — same permission cover embeds already hardcode).
- Registration is lazy (inside `sanitizeHtml`) rather than at module load so the module can still be imported in SSR without touching DOMPurify before a `window` exists.
- The `allow` merge preserves whatever permissions the stored HTML already had (`encrypted-media *;` → `encrypted-media *; fullscreen *;`) and is a no-op when `fullscreen` is already present, so direct-provider embeds (YouTube/public Vimeo) that already carry `allow="…; fullscreen; …"` are untouched.
- No backend change: `app/models/link.rb`'s sanitizer allowlist already permits `allowfullscreen` and `allow`, and the fix normalizes at render time so already-stored HTML is covered without a data migration.

PR #6014, opened 2026-07-20
Stop alerting Sentry on expected bank-directory rejections during bank account sync (Stripe::InvalidRequestError)

gumclaw:
## What

Stops paging Sentry when Stripe rejects a seller's bank details during a bank account sync (`StripeMerchantAccountManager.update_bank_account`) because the bank couldn't be found in Stripe's directory — e.g. `Stripe::InvalidRequestError: We couldn't find the bank for that BIC`.

The rescue in `update_bank_account` classified expected bank rejections by matching a handful of message strings. Directory rejections like the BIC one don't match any of them, so they fell through to `ErrorNotifier.notify` and paged Sentry — even though the failure is already fully handled (the seller is emailed via `ContactingCreatorMailer.invalid_bank_account` and a retryable bank-sync payout note is recorded).

The fix reuses the existing `bank_account_invalid_error?` classifier — the same one `create_account` already uses to suppress these exact rejections (#5991) — which recognizes them structurally via the error's `code` (`routing_number_invalid` / `account_number_invalid`) or `param` (`bank_account[...]` / `external_account[...]`), instead of adding another message-string match. The message-string checks remain for older rejection shapes that carry no code/param.

## Why

Sentry issue GUMROAD (id 7578640113): 511 events since 2026-06-27, all from `RetryStripeRejectedPayoutSetupForSellerJob` → `update_bank_account`. Each automated payout-setup retry for a seller whose BIC Stripe can't resolve re-paged Sentry, even though every occurrence is an expected seller-input error with retry, seller notification, and payout-note breadcrumbs already in place. This is the `update_bank_account` sibling of the `create_account` fix shipped in #5991 — same error class, same classifier, other code path.

Behavior change is alerting-only for this error shape: the return value moves from `:stripe_invalid_request` to `:invalid_bank_account`, which additionally emails the seller that their bank details need attention (correct — the details are genuinely unusable) and lets the retry job treat the note consistently with other invalid-bank outcomes. Unknown/unexpected `Stripe::InvalidRequestError`s still page Sentry.

## Before / After

Not a UI change — alerting/return-value behavior in the payout-setup sync path. Adjacent seller-visible flow is unchanged: the seller already received the bank-sync failure email/note before this change for matched shapes; directory rejections now consistently join that path instead of silently returning `:stripe_invalid_request`.

## Test Results

New spec: `Stripe::Account.update` raising `We couldn't find the bank for that BIC` with param `bank_account[routing_number]` → returns `:invalid_bank_account`, emails the seller, records the retryable note, and does NOT call `ErrorNotifier.notify`. Verified load-bearing: red without the app change (stash → 1 failure), green with it.

Full spec run (20 examples, 0 failures) — `stripe_merchant_account_manager_payout_retry_spec.rb`:

spec run

Also ran the existing `#update_bank_account` describe block in `stripe_merchant_account_manager_spec.rb`: 15 examples, 0 failures. Rubocop clean on both changed files.

## QA steps

Backend alerting change with no user-facing surface — the spec run above is the artifact. To verify manually: in a Rails console, stub `Stripe::Account.update` to raise `Stripe::InvalidRequestError.new("We couldn't find the bank for that BIC", "bank_account[routing_number]")` and call `StripeMerchantAccountManager.update_bank_account(user, passphrase:)` — it returns `:invalid_bank_account` without invoking `ErrorNotifier`.

---

AI disclosure: Authored by Claude Fable 5 (Hermes Agent / gumclaw), triggered by a Sentry event-alert webhook for issue 7578640113. Instructions: triage the Sentry error, root-cause it, write a failing test, fix, and open a PR per the standard Sentry auto-fix pipeline.


gumclaw:

- The one functional line is adding `bank_account_invalid_error?(e)` to the existing expected-rejection condition in `update_bank_account`'s `Stripe::InvalidRequestError` rescue. The classifier is the same private method `create_account` has used for this purpose since #5991, so the two paths now agree on what counts as an expected bank rejection.
- Ordering note: `incorrect_account_holder_name` is checked (and returns) before this condition, and `bank_account_invalid_error?` itself also excludes that code, so holder-name mismatches keep their dedicated `:invalid_account_holder_name` path either way.
- Return-value shift for these shapes: `:stripe_invalid_request` → `:invalid_bank_account`. Callsites checked: `RetryStripeRejectedPayoutSetupForSellerJob#attempt_remediation` treats both as "not remediated" (only `:synced`/`:noop_metadata_match` count as success), and the two internal callers in `stripe_merchant_account_manager.rb` don't branch on these two values differently. The only behavior delta beyond alerting is the seller email, which is the correct treatment for genuinely unusable bank details.
- New spec proven load-bearing: stashing the app change makes it fail on `ErrorNotifier` receiving `:notify` (and the return value), restoring it goes green. Full file: 20 examples, 0 failures; existing `#update_bank_account` block in the big spec: 15 examples, 0 failures.

PR #6017, opened 2026-07-20
Compute profile-sections ratings-filter boolean in SQL instead of loading the seller's entire catalog

gumclaw:
## Summary

Works #6015.

Product pages (`LinksController#show`) computed the profile-sections `show_ratings_filter` boolean with:

```ruby
seller.links.alive.any?(&:display_product_reviews?)
```

`.any?(&block)` defeats the SQL `EXISTS` optimization: Rails materializes the seller's **entire alive catalog** — `SELECT links.*`, full rows including large text columns, no LIMIT — and evaluates the flag in Ruby per record. Span evidence: **4.2s of a 4.4s request spent in that one `SELECT links.*` query** (Sentry sample `b2dfbcdb64c54376b10f6cfe2d6d2857`), on the profile-sections cache-miss path.

`display_product_reviews` is a pure flag_shih_tzu flag (bit 17 on `links.flags`, no method override), so the gem-generated `Link.display_product_reviews` scope is exactly equivalent. The fix:

```ruby
seller.links.alive.display_product_reviews.exists?
```

which runs as:

```sql
SELECT 1 AS one FROM `links`
WHERE `links`.`user_id` = ?
  AND purchase_disabled_at IS NULL AND banned_at IS NULL AND deleted_at IS NULL
  AND ((`links`.`flags` & 65536 = 65536))
LIMIT 1
```

EXPLAIN confirms it's index-driven (`index_links_on_user_id` intersect `index_links_on_deleted_at`, `rows=1`).

Sibling scan: no other `.alive.any?(&:...)` full-materialization patterns exist in `app/presenters/` — this was the only instance.

**No UI changes** — backend-only query-shape change; the rendered props are byte-identical (behavior equivalence covered by specs below).

## QA steps

1. Read the one-line presenter diff in `app/presenters/profile_sections_presenter.rb` — only the query shape changes, not the boolean semantics.
2. Run the presenter spec: `bundle exec rspec spec/presenters/profile_sections_presenter_spec.rb` → 22 examples, 0 failures (includes 5 new `show_ratings_filter` examples: reviews-enabled → true, all-disabled → false, deleted-only → false, no products → false, plus scope↔predicate equivalence + SQL bit-test assertion).
3. Optionally in a Rails console: `User.find(<big seller id>).links.alive.display_product_reviews.exists?` and compare with `...alive.any?(&:display_product_reviews?)` — same boolean, milliseconds vs seconds.
4. Post-deploy: watch `LinksController#show` p99 in Sentry; the multi-second `SELECT links.*` spans should disappear.

## Test Results

```
$ bundle exec rspec spec/presenters/profile_sections_presenter_spec.rb

ProfileSectionsPresenter
  show_ratings_filter
    is true when the seller has an alive product with reviews displayed
    is false when all of the seller's products have reviews display disabled
    is false when the seller's only reviews-enabled product is deleted
    is false when the seller has no products
    computes the boolean in SQL without loading the seller's products into Ruby

Finished in 30.54 seconds (files took 2.84 seconds to load)
22 examples, 0 failures

$ bundle exec rubocop app/presenters/profile_sections_presenter.rb spec/presenters/profile_sections_presenter_spec.rb
```

Spec-run evidence (backend-only PR — no UI surface changes; screenshot is the green spec run + new query shape):

spec run

## AI disclosure

🤖 Generated with claude-fable-5, Hermes agent


gumclaw:
Fable review: **PASS** — no P1/P2 findings.

Verified independently:
- Flag-bit math checks out: flag_shih_tzu bit 17 (1-based) = `flags & 65536` — matches both the generated scope and the slow-query evidence from Sentry (the probe queries in sample b2dfbcdb filtered the same bit family).
- `display_product_reviews?` confirmed a pure flag read (no override anywhere in link.rb) → scope ↔ predicate equivalence holds, and the spec suite asserts it explicitly across flag states.
- EXPLAIN shape: `index_merge(index_links_on_user_id, index_links_on_deleted_at)`, rows=1 — vs the old unbounded full-row scan. This is the whole 4.2s → ~0 win.
- Behavior surface: `show_ratings_filter` is computed OUTSIDE the profile-sections cache read, so the fix helps every request, not just cache misses. No caching semantics change.
- Sibling scan negative result matches my own grep (this was the only `.alive.any?(&:...)` in presenters).

Post-deploy verification (I'll run it): p99 on `LinksController#show` in Sentry should drop; the multi-second `SELECT links.*` spans disappear. Issue #6015 stays open until that's confirmed.

Human review requested per hot-path rule — the change itself is one line and the evidence is in the PR body.

gumclaw:

PR #6023, opened 2026-07-20
Fix Slow DB Query in Products::AffiliatedController#index by counting products in SQL

gumclaw:
## What

Computes the affiliated products page's `total_products` stat with a single `COUNT(DISTINCT affiliates_links.link_id)` query instead of executing the full grouped affiliated-products query unbounded and counting unique link ids in Ruby. The paginated product listing and the count now share one non-aggregating base scope (`affiliated_products_scope`); the grouped select/join/order used for the visible page rows is unchanged.

## Why

Sentry flags recurring Slow DB Query performance issues on `Products::AffiliatedController#index` (issues 7406630909 / 7406642285 / 7409792519 — ~175 events, up to 17 users affected over the last week, still firing as of today). The span breakdown shows the grouped query — `affiliates_links` LEFT JOIN `affiliate_credits` with `SUM(amount_cents)`/`COUNT(DISTINCT ...)` aggregation — executing **twice** per request: once with `LIMIT 20` for the page (via pagy), and once completely unbounded (thousands of grouped rows for large affiliates) solely so `stats` could do `.map(&:link_id).uniq.size`. The unbounded aggregation was the dominant slow span (1.1s+ for the windowed COUNT alone, ~6.7s controller time).

Counting distinct products doesn't need the revenue/sales aggregation at all — a plain join + `COUNT(DISTINCT link_id)` answers it with an index-friendly query and no `affiliate_credits` join.

## Test Results

Ran locally on the branch (all green):

- `spec/presenters/affiliated_products_presenter_spec.rb` — 17 examples, 0 failures (includes the exact-stats assertion `total_products: 7`)
- `spec/controllers/products/affiliated_controller_spec.rb` — 11 examples, 0 failures
- `spec/requests/products/affiliated_products_spec.rb` — 19 examples, 0 failures

spec results

`rubocop` clean on the changed file.

## Before/After

Not a UI change — the page renders identical data (stats values are asserted exactly by the presenter specs, which pass unmodified). Backend query-shape change only; no adjacent surface differs.

## QA Steps

1. Log in as a user with direct/global affiliate relationships.
2. Visit `/products/affiliated` — product list, revenue, sales counts, and the "Products" stat should be identical to before.
3. Search and sort still work (covered by the request/presenter specs).

---

**AI disclosure:** Generated with Claude Fable 5 (Hermes agent), triggered by a Sentry performance alert webhook for `Products::AffiliatedController#index`. Instruction: identify the slow span from the Sentry event, root-cause it, fix with tests, and open a PR.


gumclaw:

- **Equivalence of the count:** the old `affiliated_products.map(&:link_id).uniq.size` counted unique `link_id`s from the grouped result (grouped by `link_id, affiliate_id, ...`). The new `COUNT(DISTINCT affiliates_links.link_id)` over the same base filters (live direct/global affiliations, product not deleted/banned, optional name query) yields the same set of link ids — the `affiliate_credits` LEFT JOIN in the old query couldn't add or remove link ids. The presenter spec's exact assertion (`total_products: 7`) passes unmodified.
- **Query filter parity:** the search query was applied to the memoized relation the old stats read, so the stat was query-filtered before; `affiliated_products_scope` preserves that.
- **No behavior change to the page rows:** the grouped select/join/group/order used by pagy is byte-identical, just built on top of the shared scope.
- **Adversarial pass (fable-5):** checked for (a) SQL injection via `sort` — untouched, still whitelist-cased; (b) `distinct.count(col)` generating `COUNT(DISTINCT col)` in Rails 7 — verified by the passing specs; (c) N+1 or extra query — net one *fewer* heavy query per request (the unbounded grouped query is gone; the COUNT is index-friendly on `affiliates_links.affiliate_id`). No P1/P2 findings.

PR #6024, opened 2026-07-20
Stop alerting Sentry on expected tax-ID rejections during Stripe account creation (Stripe::InvalidRequestError)

gumclaw:
## What

Stops paging Sentry when Stripe rejects a seller's tax ID during merchant account creation (`StripeMerchantAccountManager.create_account`) — e.g. `Stripe::InvalidRequestError: Invalid Tax ID. 123456789 is not an allowed value.`

Adds a `tax_id_invalid_error?` classifier (mirroring the existing `bank_account_invalid_error?` and the phone-number classifier from #6008) and excludes matching errors from `ErrorNotifier.notify` in the `create_account` rescue. It matches structurally when Stripe populates `code` (`tax_id_invalid`) or `param` (`individual[id_number]` / `company[tax_id]`), and falls back to the `Invalid Tax ID` message shape because Stripe doesn't always populate code/param on this rejection.

## Why

Sentry issue 7578121692: 6,470 events / 432 users since 2026-06-26, all from `Settings::PaymentsController#update` → `StripeMerchantAccountManager.create_account` → `Stripe::Account.create`. Stripe denylists placeholder tax IDs (sequential or repeated digits like `123456789`), and we don't pre-validate against that denylist, so sellers who enter one hit this rejection on every save attempt.

Each occurrence is an expected, fully-handled seller-input error: the controller rescues `Stripe::StripeError` and shows Stripe's message inline on the payments settings page, so the seller sees exactly what to correct. There is nothing to retry (the input is wrong until the seller fixes it) and no UX gap — the only defect is the per-occurrence Sentry page. This is the tax-ID sibling of the phone-number fix in #6008 and the bank-directory fix in #6014: same rescue, same classifier pattern, different rejected field.

Unknown/unexpected `Stripe::InvalidRequestError`s still page Sentry, and the error still re-raises so the controller's inline error path is unchanged.

## Before / After

Not a UI change — alerting-only behavior in the account-creation rescue. The seller-visible flow is unchanged: they saw Stripe's "Invalid Tax ID…" message inline before this change and still do; the exception still propagates to the controller identically.

## Test Results

New specs in `stripe_merchant_account_manager_payout_retry_spec.rb`:
- message-shape rejection (`Invalid Tax ID. … is not an allowed value.`, no param) → re-raises, no `ErrorNotifier.notify`, no bank-sync payout note
- param-shape rejection (`individual[id_number]`, generic message) → re-raises, no `ErrorNotifier.notify`

Verified load-bearing: with the app change stashed, both new Sentry-suppression examples fail; with it, green.

Full spec run (22 examples, 0 failures):

spec run

## QA steps

Backend alerting change with no user-facing surface — the spec run above is the artifact. To verify manually: in a Rails console, stub `Stripe::Account.create` to raise `Stripe::InvalidRequestError.new("Invalid Tax ID. 123456789 is not an allowed value.", nil)` and call `StripeMerchantAccountManager.create_account(user, passphrase:)` — it re-raises without invoking `ErrorNotifier`.

---

AI disclosure: Authored by Claude Fable 5 (Hermes Agent / gumclaw), triggered by a Sentry event-alert webhook for issue 7578121692. Instructions: triage the Sentry error, root-cause it, write a failing test, fix, and open a PR per the standard Sentry auto-fix pipeline.


gumclaw:
Merged in the latest main to clear the conflict. Two files clashed with the Japanese-address Sentry-suppression change that landed on main: in the Stripe merchant account manager I combined both guards so the notifier now skips bank, tax-ID, and JP-address rejections alike (comment updated to cover all three), and in the payout retry spec I kept both sides' test contexts. Targeted specs pass (25 examples, 0 failures) and rubocop is clean.

gumclaw:

PR #6025, opened 2026-07-20
Fix UnsafeRedirectError for bare relative next paths in SafeRedirectPathService

gumclaw:
## What

Fixes `ActionController::Redirecting::UnsafeRedirectError` raised from `SignupController` (and any other flow using `after_sign_in_path_for` / `safe_redirect_to`) when the `next` param is a bare relative path without a leading slash, e.g. `/signup?next=dashboard`.

`SafeRedirectPathService#relative_path` now prepends a leading `/` to non-empty sanitized paths so the redirect target stays host-relative. Query-only paths (`?query=param`) are unchanged.

## Why

Sentry issue [GUMROAD `SignupController#new` UnsafeRedirectError](https://gumroad-to.sentry.io/issues/7521489724/) — 137 events / 56 users since 2026-06-02, still recurring (last seen 2026-07-20).

When an already-signed-in user hits `/signup?next=dashboard`, `after_sign_in_path_for` passes `dashboard` through `SafeRedirectPathService`, which returns it unchanged (`URI("dashboard").path == "dashboard"`). Rails' `redirect_to` then resolves it against the request host as `https://gumroad.comdashboard` — a cross-host URL — and raises `UnsafeRedirectError`, so the user gets a 500 instead of landing on their dashboard.

## Before / After

Not a UI change — backend redirect sanitization only. No adjacent rendered surface is affected: the fix converts a 500 error into the redirect that was always intended. Spec run below is the artifact.

Before: `SafeRedirectPathService.new("dashboard", request).process` → `"dashboard"` → `UnsafeRedirectError`
After: → `"/dashboard"` → normal same-host redirect

## Test Results

spec results

- `spec/services/safe_redirect_path_service_spec.rb` — 17 examples, 0 failures (includes 2 new regression specs; verified they fail without the fix)
- `spec/controllers/signup_controller_spec.rb` — 44 examples, 0 failures

## QA steps

1. Sign in to any account.
2. Visit `/signup?next=dashboard` (no leading slash on `next`).
3. Before: 500 error (`UnsafeRedirectError`). After: redirected to `/dashboard`.

---

🤖 AI disclosure: authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert. Instructions: triage the Sentry issue, reproduce the root cause, write a failing test, fix, verify locally, open a PR.


gumclaw:
Self-review notes:

- The fix is scoped to `SafeRedirectPathService#relative_path`, which only runs on the "unsafe" branch of `process` (path failed the same-host/subdomain checks). Same-host and subdomain URLs are returned untouched, so legitimate absolute redirects are unaffected.
- The `unless _path.empty?` guard preserves the existing "query-only path" behavior (`?query=param` must not become `/?query=param`) — that case has `url.path == ""` and is covered by the pre-existing spec at line 120.
- Red-green verified: both new regression specs fail on the unmodified service (`expected "/dashboard", got "dashboard"`) and pass with the fix.
- Considered fixing at the caller (`after_sign_in_path_for`) instead, but the service is the single sanitization choke point used by `safe_redirect_to` across controllers — fixing here covers every `next=`-style entry point, not just signup.

gumclaw:
Fable-5 adversarial review: no P1/P2 findings. Checked: (1) no open-redirect reintroduction — backslash/percent-encoded/protocol-relative escapes are normalized and collapsed before the new slash-prepend, so a bare path can never carry a host; (2) query-only paths guarded by the empty check (existing spec); (3) fragment stripping unchanged; (4) both call sites (safe_redirect_to, after_sign_in_path_for) verified compatible.

PR #6027, opened 2026-07-20
Add covering index for affiliate credit sums to fix Slow DB Query on Products::AffiliatedController#index

gumclaw:
## What

Adds a composite covering index to `affiliate_credits` for the "paid credit sum per affiliate user" query that powers the total-revenue stat on the affiliated products dashboard (`Products::AffiliatedController#index`):

```
(affiliate_user_id, affiliate_credit_refund_balance_id,
 affiliate_credit_chargeback_balance_id, affiliate_credit_success_balance_id,
 amount_cents)
```

No application-code change — `User#affiliate_credits_sum_total` / `AffiliateCredit.paid` already issue exactly this shape of query; they just have no good index to run on.

## Why

Sentry keeps flagging Slow DB Query performance issues on `Products::AffiliatedController#index` (issue 7409792519, 38 events / 11 users in the last week; sibling issues 7406642285 / 7406630909 / 7406630904 add ~350 more events). In today's 6.2s sample the single largest span (1.8s) is:

```sql
SELECT SUM(amount_cents) FROM affiliate_credits
WHERE affiliate_user_id = ?
  AND affiliate_credit_refund_balance_id IS NULL
  AND affiliate_credit_chargeback_balance_id IS NULL
  AND affiliate_credit_success_balance_id IS NOT NULL
```

The table only has single-column indexes, so MySQL resorts to an index-merge intersect that range-scans the enormous `refund_balance_id IS NULL` / `chargeback_balance_id IS NULL` index ranges (nearly the whole table each) before intersecting with the user's rows. The composite index serves the query entirely from the index: equality prefix on `affiliate_user_id`, the three balance-id predicates next, `amount_cents` last so the SUM needs no row lookups.

This is item 2 of the root-cause breakdown on #6020 (item 1, the redundant grouped query, shipped in #6023 earlier today).

Refs #6020

## Before/After

Local benchmark on a 200,000-row structural copy of `affiliate_credits` (1,000 credits for the probed user), `EXPLAIN ANALYZE` on the exact production query:

| | Plan | Actual time |
|---|---|---|
| **Before** | Index-merge intersect over `affiliate_user_id` + two balance-id indexes, scanning ~190k index entries per NULL range | **64.8 ms** |
| **After** | Single covering index range scan, 900 rows, zero table lookups | **0.19 ms** |

explain-before-after

No user-visible surface changes — the affiliated products page renders identically, just faster. Pure DB/index change with no adjacent UI to screenshot; the EXPLAIN ANALYZE capture above is the artifact.

## QA steps

1. Run the migration (`bin/rails db:migrate`) — uses `disable_ddl_transaction!`, online `ADD INDEX` (INPLACE in MySQL 8, no table copy/write blocking).
2. `EXPLAIN` the sum query above for any user id — plan should show `idx_affiliate_credits_on_user_and_balances_and_amount` with `Using index`.
3. Load `/products/affiliated` as an affiliate with sales — the total revenue stat is unchanged.

## Test Results

- `ruby -c` on the migration: Syntax OK; `rubocop -a`: 1 file inspected, no offenses.
- Verified on local MySQL 8 (gumroad_test): index created via the same `ALTER TABLE ... ADD INDEX`, `EXPLAIN` confirms `Covering index range scan ... Using index` (screenshot above).
- No model/presenter specs affected — schema-only change; existing `spec/presenters/affiliated_products_presenter_spec.rb` exercises the query path and is unchanged.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert for Slow DB Query issue 7409792519 on `Products::AffiliatedController#index`. Instructions: triage the Sentry performance event, identify the slow span, and fix the backend issue with evidence. Tools: Sentry API (span breakdown), local MySQL EXPLAIN ANALYZE benchmark, gh CLI.


gumclaw:
Self-review notes:

- **Column order rationale:** `affiliate_user_id` first (equality), then the two `IS NULL` balance ids, then `success_balance_id` (`IS NOT NULL` range), `amount_cents` last purely for covering — the SUM reads it from the index without touching rows. `EXPLAIN ANALYZE` on a 200k-row copy confirms `Covering index range scan` (see PR body).
- **Why no code change:** `AffiliateCredit.paid` / `User#affiliate_credits_sum_total` already emit exactly this predicate shape; the index alone fixes the plan. `affiliate_credits_sum_for_credits_created_between` (same scope + `created_at` window) also benefits from the equality/NULL prefix.
- **Migration safety:** `disable_ddl_transaction!` matches the repo's recent index migrations (e.g. `20261205000001`); MySQL 8 `ADD INDEX` is INPLACE/online, no table rebuild or write blocking.
- **Redundancy check:** the existing single-column `index_affiliate_credits_on_affiliate_user_id` is now prefix-redundant with the new index, but I left it in place — dropping indexes on a hot table is a separate, riskier decision better made after production confirms the new index is picked.

gumclaw:

Hunted for the usual index-migration failure modes:

- **Wrong predicate coverage** — checked `AffiliateCredit.paid` / `not_refunded_or_chargebacked` scopes against the index column order: equality on `affiliate_user_id`, both `IS NULL` columns next, `IS NOT NULL` range, `amount_cents` covering. `EXPLAIN ANALYZE` on a seeded 200k-row copy shows `Covering index range scan`, 64.8ms → 0.19ms. Matches.
- **schema.rb drift** — the index line is placed in Rails' alphabetical dump order (`idx_...chargeback` < `idx_...user_and_balances` < `index_...`), version bumped to the migration's timestamp; a CI `db:migrate` + schema dump will produce an identical file.
- **Index name length** — 53 chars, under MySQL's 64-char limit.
- **Migration safety** — `disable_ddl_transaction!` + plain `add_index`, consistent with the repo's recent index migrations (`20261205000001`, `20261205000002`); MySQL 8 online INPLACE DDL.
- **Duplicate/no-op risk** — no existing multi-column index on `affiliate_credits`; the new index makes the single-column `affiliate_user_id` index prefix-redundant (P3: candidate for a follow-up drop after production plan verification — noted in self-review, intentionally not done here).

No application code touched; no behavior change possible beyond the query plan.

gumclaw:
CI is green at the current head (Buildkite success; all check-runs completed, none non-green). Fable-5 adversarial review passed with no P1/P2 — one P3 noted in the review comment (the single-column `affiliate_user_id` index becomes prefix-redundant; suggested as a follow-up drop after production confirms the new plan). Since this is a schema migration it stays off auto-merge — over to @ershad for review and merge.

PR #6028, opened 2026-07-20
Flag scheduled payouts with no payable balance instead of erroring daily

gumclaw:
## What

When a due scheduled payout's user has no payable balance, `ScheduledPayout#execute!` now flags the record for admin review instead of raising and resetting it to `pending`. Genuine processor errors from `Payouts.create_payment` (non-empty `payment_errors`) still raise and reset to pending as before.

## Why

Sentry issue [GUMROAD 7435892078](https://gumroad-to.sentry.io/issues/7435892078/) — `RuntimeError: Payout failed: No payable balance available` in `ScheduledPayout#execute!` — has fired **607 times since 2026-04-23**, in daily bursts at 09:00 UTC.

Root cause: `Payouts.create_payment` returns `nil` when the user has no payable balance (nothing unpaid, or the balance nets to ≤ 0). `execute!` treated that as a failure, raised, and its rescue reset the record to `pending`. The daily `ExecuteScheduledPayoutsJob` then picked the same record up again the next day and failed identically — an infinite retry loop with no path to resolution, producing the same Sentry error every day per stuck record.

Retrying can't help here: there is nothing to pay out. Flagging is the correct terminal-ish state — the admin scheduled-payouts UI already supports re-executing or cancelling `flagged` payouts (both `Admin::ScheduledPayoutsController#execute` and the internal-admin endpoint transition `flagged → pending → execute!`), so an admin reviews once and either cancels or re-runs after the balance situation changes.

## Before / After

No user-facing surface changes — backend model/job change. The admin scheduled-payouts UI already renders the `flagged` status and its execute/cancel actions (unchanged by this PR).

**Walkthrough video** (diff + fix + full local spec runs):

https://github.com/antiwork/gumroad/blob/fix/scheduled-payout-zero-balance-flag/qa-media/pr-6028-scheduled-payout-zero-balance-walkthrough.mp4?raw=true

## QA steps

Backend-only; reviewer path via Rails console (preview or local):

1. Create a due scheduled payout for a user with no unpaid balance: `sp = ScheduledPayout.create!(user: user, action: "payout", status: "pending", delay_days: 0, scheduled_at: 1.minute.ago, payout_amount_cents: 0)`
2. Run `ExecuteScheduledPayoutsJob.new.perform`.
3. Before this PR: raises are swallowed by the job but reported to Sentry, and `sp.reload.status == "pending"` — it retries daily forever. After: `sp.reload.status == "flagged"`, no error reported.
4. In the admin scheduled-payouts queue, confirm the flagged record shows with Execute/Cancel actions (existing UI, unchanged).

## Test plan

- Updated the model spec: zero-balance path now expects `:flagged` return value, `flagged` status, and `executed_at` nil.
- Added a job-level spec proving a due payout with no payable balance ends up `flagged` (not `pending`) after `ExecuteScheduledPayoutsJob#perform` — this is the regression test for the daily retry loop.
- Existing raise-and-reset behavior for real processor errors (`payment_errors` present, failed payment) still covered and passing.
- After deploy: the 09:00 UTC daily burst of this Sentry error should stop once the currently-stuck pending scheduled payouts get flagged on their next run; the flagged records surface in the admin scheduled payouts queue for one-time review.

## Test Results

```
$ bundle exec rspec spec/models/scheduled_payout_spec.rb spec/sidekiq/execute_scheduled_payouts_job_spec.rb
Finished in 4.78 seconds (files took 8.22 seconds to load)
51 examples, 0 failures

$ bundle exec rspec spec/controllers/api/internal/admin/scheduled_payouts_controller_spec.rb spec/controllers/admin/scheduled_payouts_controller_spec.rb
Finished in 7 seconds (files took 3.05 seconds to load)
61 examples, 0 failures
```

rubocop: 3 files inspected, no offenses detected.

---

**AI disclosure:** Authored by Claude Fable 5 (Hermes Agent) from an automated Sentry webhook pipeline. Instructions given: triage the recurring Sentry issue (recurrence-gate verified, 607 events), identify the root cause from the stacktrace, write a failing test, fix, verify specs locally, and open a PR assigned to the most recent code owner.


gumclaw:

- `app/models/scheduled_payout.rb`: the new `update!(status: "flagged", executed_at: nil)` + `return :flagged` happens inside the `begin` block but cannot re-trip the `rescue` reset — `update!` on a valid transition won't raise, and `return` exits before any processor call. The record was set to `executed` inside the lock earlier, so the flag transition is `executed → flagged`; there's no status inclusion guard beyond the STATUSES list, and both controllers' execute endpoints already accept `flagged → pending → execute!` for re-runs, so a flagged record remains fully recoverable.
- Real processor errors (`payment_errors` present) keep the old raise-and-reset-to-pending behavior — those ARE worth a retry the next day (e.g. transient Stripe account issues).
- `spec/sidekiq/execute_scheduled_payouts_job_spec.rb`: the "continues processing other payouts" spec previously stubbed `create_payments_for_balances_up_to_date_for_users`, which `execute!` never calls — the spec only passed because the zero-balance path raised. Re-pointed the stub at `create_payment` (the method actually used) so the spec still exercises the error-isolation behavior after this change.
- New job-level spec ("flags due payouts with no payable balance instead of retrying them daily") is the regression test for the daily loop: a due payout for a user with no balance ends `flagged`, not `pending`.

gumclaw:
CI green on head (29731136684) — marking ready for review. Payments-surface PR: no auto-merge; merge stays with a human.

gumclaw:
Merged in the latest main to clear the conflict. `scheduled_payout.rb` clashed with the failed-payment error-detail change that landed on main: I combined both behaviors — validation errors and failed-during-preparation payments still raise (so the rescue resets the payout to pending for retry), while the genuine no-payable-balance case takes this PR's new `flagged` path instead of erroring daily. Full `scheduled_payout_spec.rb` passes locally (46 examples, 0 failures).

PR #6029, opened 2026-07-20
Set CANNOT_PAY failure reason for missing Stripe capabilities error in payout internal transfer

gumclaw:
## What

Handles the `Stripe::InvalidRequestError` "Your destination account needs to have at least one of the following capabilities enabled: transfers, crypto_transfers, legacy_payments" raised by `StripeTransferInternallyToCreator.transfer_funds_to_account` during `StripePayoutProcessor.prepare_payment_and_set_amount`:

- Routes the `InvalidRequestError` rescue in `prepare_payment_and_set_amount` through the existing `stripe_invalid_request_error_failure_reason` mapper (previously only used by `perform_payment`).
- Adds the "capabilities enabled" message pattern to that mapper, mapping to `Payment::FailureReason::CANNOT_PAY`.
- Only notifies Sentry for genuinely unexpected `InvalidRequestError`s (no recognized failure reason).

## Why

When a connected Stripe account has outstanding verification/compliance requirements, Stripe disables its `transfers` capability. Every nightly payout run for such a creator then:

1. Reported the error to Sentry as an unexpected error — Sentry issue GUMROAD-7Q has accumulated 1,221 events since March 31 for the same known account-state condition.
2. Marked the payment failed **without** a `failure_reason`, so the `CANNOT_PAY` after-transition hook never fired and the creator never got the "we cannot pay you, fix your account" email — the failure was silent for the creator and the funds kept re-cycling every payout run.

With this change the payment fails with `CANNOT_PAY`, the creator gets the existing cannot-pay email prompting them to complete their Stripe requirements, and Sentry only sees unexpected errors on this path.

Note: PR #4227 attempted a narrower version of this fix (inline message match) and was auto-closed as stale; this supersedes it by reusing the shared mapper instead of duplicating message matching.

## Before / After

Not a UI change — this is the nightly Sidekiq payout pipeline; there is no adjacent user-facing surface to screenshot. The artifact is the spec run below (also committed to `qa-media/`).

Behavior delta:
- Before: payment `failed` with `failure_reason = nil`, no creator email, Sentry event per occurrence.
- After: payment `failed` with `failure_reason = "cannot_pay"`, creator receives the cannot-pay email, no Sentry event.

## Test plan / Test Results

New specs cover both branches: the capabilities error (CANNOT_PAY set, no `ErrorNotifier.notify`) and an unexpected `InvalidRequestError` (notified, no failure reason).

Ran locally:

```
bundle exec rspec spec/business/payments/payouts/processor/stripe/stripe_payout_processor_spec.rb
304 examples, 0 failures
```

spec results

## QA steps

Backend-only; no preview click-path. To verify in a console: stub `StripeTransferInternallyToCreator.transfer_funds_to_account` to raise the capabilities error and call `StripePayoutProcessor.prepare_payment_and_set_amount` — the payment ends `failed` with `failure_reason == "cannot_pay"`.

---

AI disclosure: Written by Claude Fable 5 (Hermes agent). Instructions: triage a recurring Sentry production error (GUMROAD-7Q, `Stripe::InvalidRequestError` on missing destination-account capabilities during payout internal transfers) and fix it with a failure-reason mapping plus specs.


gumclaw:
Self-review notes:

- `prepare_payment_and_set_amount` now mirrors `perform_payment`'s error handling: `failure_reason` is resolved via the shared `stripe_invalid_request_error_failure_reason` mapper and passed to `payment.mark_failed!(failure_reason)` in the `ensure` block. `mark_failed!(nil)` is behavior-identical to the previous bare `mark_failed!` (the `before_transition` hook sets `failure_reason = transition.args.first`, which was already `nil` before).
- The new mapper pattern matches on the stable phrase `needs to have at least one of the following capabilities enabled` rather than the full capability list, so it survives Stripe adding/removing capability names in the message.
- Mapping to `CANNOT_PAY` (rather than a new reason) reuses the existing `send_cannot_pay_email` after-transition hook and payout-pause logic — same creator action required as the already-mapped "Cannot create payouts" case.
- Specs stub `get_payout_details` and the internal transfer, so no VCR cassettes needed; balance fixture must be `state: "processing"` because `mark_failed!` triggers `mark_balances_as_unpaid`.
- Full spec file run locally: 304 examples, 0 failures.

Pre-merge checklist: CI pending at head; fable-5 adversarial review to run before merge (payments-surface PR — no auto-merge, human merges).

PR #6031, opened 2026-07-20
Fix StateMachines::InvalidTransition (failed → processing) in Payouts.create_payment

gumclaw:
## What

`Payouts.create_payment` no longer calls `payment.mark_processing!` on a payment that the payout processor already marked as failed during preparation, and `ScheduledPayout#execute!` now treats a failed-during-preparation payment the same as a payment that could not be created (reset to pending, surface the preparation error) instead of sending it to the payout processor.

## Why

`StripePayoutProcessor.prepare_payment_and_set_amount` marks the payment as **failed** in several cases: no valid merchant account, a balance whose holding currency doesn't match the payout destination, or destination-balance drift. `Payouts.create_payment` then unconditionally called `payment.mark_processing!`, which raises `StateMachines::InvalidTransition: Cannot transition state via :mark_processing from :failed` — the `mark_processing` event only permits `creating → processing`.

This has been firing daily at 09:00 UTC from `ExecuteScheduledPayoutsJob` (Sentry issue `7491555239`, 137 events since May 19). The affected scheduled payouts never execute and reset to pending every day, retrying and failing forever. Three earlier fix attempts (#4122, #4162, #4225) were auto-closed as stale without merging.

With this change:
- `create_payment` skips the processing transition for a failed payment and returns it with the preparation errors, matching how callers already interpret the `[payment, payment_errors]` contract (`PayoutUsersService` and `InstantPayoutsService` both check `payment_errors.blank?` before processing).
- `ScheduledPayout#execute!` raises a descriptive `Payout failed: <preparation error>` instead, resetting the scheduled payout to pending — same behavior as when no payment could be created — and never hands a failed payment to `process_payments`.

## Before/After

Not a UI change — backend payout state-machine fix. No adjacent user-facing surface: the error occurred inside a Sidekiq job (`ExecuteScheduledPayoutsJob`); the spec run below is the artifact.

## Test Results

test-results

- New spec `Payouts.create_payment does not attempt to mark the payment as processing when the processor failed it during preparation` reproduces the exact `StateMachines::InvalidTransition` failure when the fix is reverted (verified red → green locally).
- New spec `ScheduledPayout#execute! raises without processing when the payment failed during preparation` proves the scheduled-payout path resets to pending and never calls the processor.
- Full local runs: `spec/business/payments/payouts/payouts_spec.rb` — 67 examples, 0 failures; `spec/models/scheduled_payout_spec.rb` — 46 examples, 0 failures.
- `rubocop` clean on all four changed files.

## QA steps

Backend-only; no preview click-path. Reviewer verification:
1. Run `bundle exec rspec spec/business/payments/payouts/payouts_spec.rb spec/models/scheduled_payout_spec.rb`.
2. After deploy, confirm Sentry issue `7491555239` stops re-firing at the next 09:00 UTC `ExecuteScheduledPayoutsJob` run, and check the previously affected scheduled payouts now surface a `Payout failed: …` error naming the real preparation problem.

---

AI disclosure: This PR was authored by Claude Fable 5 (Hermes Agent / gumclaw), triggered automatically by a Sentry alert webhook for issue `7491555239`. Instructions given: triage the Sentry event, identify the root cause from the stacktrace, write a failing test, fix, verify locally, and open a PR assigned per recent file ownership.


gumclaw:
Self-review notes:

- **`payouts.rb` — `payment.mark_processing! unless payment.failed?`**: The `mark_processing` event only permits `creating → processing`, so guarding on `failed?` is sufficient — after `save!` + `prepare_payment_and_set_amount` the payment is either `creating` or `failed` (the processor's only transitions in that method are `mark_failed!`). Returning the failed payment with `payment_errors` matches the existing contract: `PayoutUsersService` and `InstantPayoutsService` both gate on `payment_errors.blank? && payment.present?` before calling `process_payments`, so a failed payment was already excluded on those paths — only the `ScheduledPayout#execute!` caller lacked the check.
- **`scheduled_payout.rb` — `payment.blank? || payment.failed?`**: mirrors the existing "no payment created" handling: reset to pending, raise with the real preparation error. The rescue block already handled the exception path; this just makes the error message name the actual cause (e.g. no valid merchant account) instead of an `InvalidTransition` stack.
- Balances marked processing by `mark_balances_processing` are returned to unpaid by the state machine's `after_transition creating → failed → mark_balances_as_unpaid` callback when the processor fails the payment, so no balance is stranded in `processing`.
- Red/green verified: with the `payouts.rb` fix stashed, the new spec fails with the exact production exception (`Cannot transition state via :mark_processing from :failed`).

slavingia:
Fable 5 adversarial review then merge plz

gumclaw:
Fable 5 adversarial review (at head `03ae8d1`) — **no P1/P2 findings, approved for merge.**

Hunt list and results:

1. **Stranded balances when prep fails** — clean. `Payment` runs `after_transition [creating, processing] → [cancelled, failed] do: :mark_balances_as_unpaid`, so balances marked processing by `mark_balances_processing` are automatically reset to unpaid when the processor fails the payment during preparation. Skipping `mark_processing!` doesn't leave anything stuck.
2. **Errors returned without the payment being failed** — clean. Every Stripe prep failure path either calls `mark_failed!` inline (no merchant account, holding-currency mismatch, destination-balance drift) or sets the `failed` flag consumed by `ensure { payment.mark_failed! if failed }`. So for Stripe, errors-present ⟺ payment failed, and the new guard covers all of them. PayPal's `prepare_payment_and_set_amount` never fails, so the guard is a safe no-op on that processor.
3. **Caller contract** — clean. `PayoutUsersService#create_payments` and `InstantPayoutsService#perform` both check `payment_errors.blank?` before handing the payment to `process_payments`, so a returned failed payment is skipped (previously these paths crashed inside `create_payment` instead).
4. **`ScheduledPayout#execute!` branch** — clean. The `payment&.failed?` safe navigation correctly handles the `payment.blank?` leg; the `payment.errors.full_messages` fallback is only reachable when `payment_errors` is blank (rare), with a sensible default string. The failure resets the scheduled payout to pending and raises a descriptive `Payout failed: <real reason>` instead of the opaque `InvalidTransition` — the daily job's error now names the actual preparation problem so a human can fix the underlying account issue.
5. **Specs** — the new `create_payment` spec reproduces the exact production failure when the fix is reverted (red → green verified per PR body), and the `defers the bank payout` spec was correctly extended with `failed?: false` on the instance double.

P3 (informational, handled post-merge): the test-results image URL references the PR branch; repointing it to `main` after merge so it survives branch deletion.

CI fully green at this head (`ci/green` + Buildkite pass). Merging per @slavingia's request.

PR #6032, opened 2026-07-20
Stop alerting Sentry on expected below-minimum and intervention-required Stripe payout rejections (Stripe::InvalidRequestError)

gumclaw:
## What

Maps two expected `Stripe::InvalidRequestError` messages raised by `StripePayoutProcessor.perform_payment` to known payment failure reasons:

- `Amount must be no less than …` → `below_stripe_payout_minimum`
- `This account requires further intervention to perform certain actions…` → `stripe_intervention_required`

Both reasons get entries in `STRIPE_FAILURE_SOLUTIONS`, so the existing `add_payment_failure_reason_comment` state-machine callback now leaves an explanatory payout note on the seller's account instead of failing silently, and `ErrorNotifier.notify` (which only fires when the failure reason is `nil`) stops reporting these occurrences to Sentry.

## Why

Sentry issue GUMROAD (id 7567368758) has fired daily at 09:00 UTC from the payout batch — 23 events since 2026-06-22 — for two conditions that are expected and already handled by the failure flow:

1. **Below Stripe's per-currency payout minimum:** small released balances (e.g. the $1 floor applied to terminally rejected accounts) can convert to less than Stripe's payout minimum in the destination currency. The funds simply roll into the next payout — nothing is broken.
2. **Account requires further intervention:** Stripe has frozen payouts until the seller resolves outstanding requirements directly with Stripe. Only the seller can clear this.

`stripe_invalid_request_error_failure_reason` already exists to classify expected rejection messages (same pattern as #5479 and #5623); these two messages were just unmapped, so every nightly occurrence alerted Sentry and left the seller with no skip-reason note.

## Before / After

No user-facing rendering change — backend payout-failure classification only. The adjacent visible surface is the seller's payout notes: previously these two failures produced no note; now the seller gets a plain-language reason and next step (see solution copy in `failure_reason.rb`).

## Test Results

New specs cover both messages (failure reason set, error message stored, no `ErrorNotifier.notify`, payout note content), plus the existing unmatched-error contexts still pass:

spec results

- `spec/business/payments/payouts/processor/stripe/stripe_payout_processor_spec.rb` (`perform_payment` error handling): 16 examples, 0 failures
- `spec/models/concerns/payment/failure_reason_spec.rb`: 7 examples, 0 failures

## QA steps

Pure backend classification change with no adjacent UI surface beyond the payout note text — the spec run above is the artifact. To verify manually: stub `Stripe::Payout.create` to raise either message and call `StripePayoutProcessor.perform_payment(payment)`; the payment is marked failed with the new reason and a payout note is added.

---

🤖 AI disclosure: authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert for issue 7567368758, instructed to triage the recurring payout-batch error, add a failing test, fix, and open a PR.


gumclaw:
Self-review notes:

- The `Amount must be no less than` match is anchored (`\A`) and deliberately does not pin the currency symbol/amount — Stripe localizes it per destination currency (£1.00 here, other symbols elsewhere).
- `requires further intervention` is unanchored because Stripe prefixes it with "This account"; kept the match narrow enough that no other Stripe message in this codebase's history collides with it.
- Both new reasons ride the existing `after_transition processing: :failed → add_payment_failure_reason_comment` callback — no new notification paths were added, only classification + seller-facing solution copy.
- `ErrorNotifier.notify` suppression falls out of the existing `if failure_reason.nil?` guard on line 416; unmatched messages still alert (covered by the untouched "unmatched Stripe::InvalidRequestError" spec context).

gumclaw:

PR #6033, opened 2026-07-20
Handle 'No active Lifecycle Action found' ValidationError in SidekiqUtility shutdown

gumclaw:
## What

Handles `Aws::AutoScaling::Errors::ValidationError` ("No active Lifecycle Action found …") in `SidekiqUtility` during instance scale-in, at both AWS call sites:

- `wait_for_sidekiq_to_process_existing_jobs`: when a heartbeat fails because the lifecycle action is gone, log and break out of the wait loop instead of crashing.
- `proceed_with_instance_termination`: when completing the lifecycle action fails for the same reason, treat it as a no-op (the instance is terminating anyway).

Any other `ValidationError` (misconfigured hook name, ASG name, etc.) still raises.

## Why

Sentry issue GUMROAD `7387896465` — 605 events since April 2026, unhandled, firing from `sidekiq_worker_long` hosts during scale-in.

Root cause: when a long-running Sidekiq worker drains for longer than the ASG lifecycle hook's heartbeat timeout allows (or the hook's total timeout elapses), AWS abandons the lifecycle action and proceeds with termination on its own. The shutdown script's next `record_lifecycle_action_heartbeat` / `complete_lifecycle_action` call then raises "No active Lifecycle Action found with instance ID … and HookName production-sidekiq-long-running-lifecycle-hook". The instance terminates either way — the outcome we wanted — so the error is pure Sentry noise plus a crashed shutdown script.

This is the noise/handling class of fix: the transient infra condition is expected and self-resolving, so the code now recognizes it explicitly instead of reporting an unhandled exception on every long-draining scale-in.

## Before / After

Not a UI change — Sidekiq instance-shutdown script only; no user-facing surface. Evidence is the spec run below.

## QA steps

No preview app needed — this code only runs inside the instance-termination path on `sidekiq_worker_long` hosts (no web surface, no route). Reviewer verification path:

1. Read `lib/utilities/sidekiq_utility.rb` — confirm both AWS call sites only swallow the "No active Lifecycle Action found" message and re-raise anything else.
2. Run `bundle exec rspec spec/lib/utilities/sidekiq_utility_spec.rb` — 18 examples, 0 failures (screenshot below).
3. Post-deploy: watch Sentry issue `7387896465` for zero new events on the next `sidekiq_worker_long` scale-in.

## Test plan

- Added specs for both rescue paths (heartbeat and completion), plus a spec asserting that other `ValidationError`s still re-raise.
- Verified the tests are load-bearing: with the lib change stashed, both new specs fail; with it applied, all pass.
- Full spec file: 18 examples, 0 failures.

spec run

- After deploy: watch Sentry issue `7387896465` — event rate should drop to zero on subsequent `sidekiq_worker_long` scale-ins; `[SidekiqUtility]` info logs will show when the rescue path is taken.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent), triggered by a Sentry alert webhook for this issue. Instructions: triage the Sentry event, identify the root cause from the stacktrace, write failing tests, fix, and open a PR with test evidence.


gumclaw:
Self-review notes:

- **Why message matching, not error-class matching:** AWS SDK raises the same `Aws::AutoScaling::Errors::ValidationError` class for many distinct validation failures (bad hook name, bad ASG name, missing action). Only the "No active Lifecycle Action found" case is safe to swallow, so the rescue re-raises anything else (`raise unless e.message.include?(...)`). Brittle-string risk is covered by a spec pinning the exact AWS message shape from production events.
- **Heartbeat path breaks instead of returning:** after the break, `stop_process` still calls `proceed_with_instance_termination`, which now no-ops on the same error — so the flow ends cleanly whether the action expired mid-wait or between the last heartbeat and completion.
- **Load-bearing check:** both new specs fail with the lib change stashed, pass with it applied.

gumclaw:

Hunted for: swallowed errors that should surface, infinite-loop regressions in the drain loop, message-match false positives, and behavior drift for other ValidationError causes.

- **P1/P2: none found.**
- Checked that the heartbeat rescue's `break` can't strand the loop — control falls through to `proceed_with_instance_termination`, which is now idempotent against the same expired-action state.
- Checked message-match scope: AWS uses the "No active Lifecycle Action found" phrasing for both `RecordLifecycleActionHeartbeat` and `CompleteLifecycleAction`; other validation failures (bad hook/ASG names) use different phrasing and still re-raise — covered by an explicit re-raise spec.
- Checked `Rails.logger` availability in this context: pre-existing usage at the stuck-jobs branch confirms it's already relied on here.
- P3 (noted, not blocking): the two rescue blocks are near-duplicates; could extract a shared helper if this file grows another AWS call site. Not worth the indirection at two call sites.

Verdict: ✅ safe to merge on green. 18/18 specs pass locally; fix verified load-bearing.

PR #6034, opened 2026-07-20
Skip doomed FX-quote calls for merchant accounts with known non-USD settlement

gumclaw:
## What

Skips the doomed FX-quote round trip at checkout for merchant accounts that Stripe has already told us settle in a non-USD currency (multi-currency settlement). Follow-up to #5999, closes #6011.

- `MerchantAccount` gets a learned marker `settlement_currency_mismatch_noticed_at` (stored in `json_data` — **no schema migration**), with `record_…!`, `clear_…!`, and `settlement_currency_mismatch_active?` (30-day TTL, malformed-timestamp safe).
- `Checkout::BuyerCurrencyQuote` and `Charge::MethodForcedPresentment` record the marker when `StripeFxQuote::SettlementCurrencyMismatch` fires (best-effort — persistence failure never breaks the already-falling-back checkout).
- `Checkout::BuyerCurrencyEligibility.usd_settling_merchant_account?` consults the marker, so both the card mode and the method-forced mode fall back with the existing `:unsupported_settlement_currency` reason **without calling Stripe**.
- The Stripe `account.updated` webhook clears the marker when `previous_attributes` shows a `default_currency` or `external_accounts` change, so accounts that fix their settlement configuration regain buyer-currency presentment before the TTL runs out.

## Why

Per the prod investigation on #6011: ~600 usd-stored accounts (GB 246, CA 226, plus tails) are multi-currency-settlement candidates. Every eligible buyer checkout on those sellers pays up to the 2-second FX-quote open-timeout budget on the hottest path, on every visit, for a call we already know will be rejected. #5999 made the rejection quiet; this makes it not happen at all after the first observation.

Fail-safe properties:
- Worst case of a stale marker: the buyer sees USD instead of their local currency (the pre-#5627 behavior for everyone) — never a failed charge.
- Worst case of an over-eager clear: one extra FX-quote round trip that re-records the mismatch.
- Marker recording/clearing failures are logged and swallowed; checkout and webhook handling proceed.

## User impact

None visible: affected checkouts already fell back to USD — they now do it ~2s faster on repeat visits. Pure backend/service change with no adjacent UI surface (the checkout renders identically on both paths, USD presentment being the default for these sellers today); the spec run + rubocop render below is the artifact per the backend-PR convention.

## Test Results

spec run

- `spec/models/merchant_account_spec.rb`, `spec/services/checkout/buyer_currency_eligibility_spec.rb`, `spec/services/checkout/buyer_currency_quote_spec.rb`, `spec/services/charge/method_forced_presentment_spec.rb`: **114 examples, 0 failures**
- `stripe_merchant_account_manager_spec.rb -e clear_settlement_currency_mismatch_on_currency_change`: **5 examples, 0 failures**
- rubocop on all 10 touched files: **no offenses**

New coverage: marker lifecycle (record/refresh/expire/clear/malformed), eligibility short-circuit + TTL expiry re-probe, quote-service skip (`StripeFxQuote` not called while fresh), recording from both rescue sites, webhook clear on `default_currency`/`external_accounts` change (and NOT on unrelated changes), persistence-failure isolation at every site.

## QA steps

Backend-only; reviewer verification path:
1. Run the specs above.
2. In a console: `ma.record_settlement_currency_mismatch!` on a USD-stored account → `Checkout::BuyerCurrencyEligibility.usd_settling_merchant_account?(ma)` → `false`; `ma.clear_settlement_currency_mismatch!` → `true` again.
3. Simulate the webhook: `StripeMerchantAccountManager.clear_settlement_currency_mismatch_on_currency_change(ma, { "default_currency" => "cad" })`.

---

## AI disclosure

🤖 Generated with **Claude Fable 5** (Anthropic), running as gumclaw's autonomous dev dispatcher.
Prompts/instructions given: the standing order to autonomously build qualifying issues (Sahil 2026-07-19, "autonomize most product dev and certainly bug fixes"); issue #6011's body (proposal checklist: persist the learned mismatch, eligibility check, webhook/TTL clear, no schema migration) served as the spec; instructed to add specs for every new path and keep the change migration-free.


gumclaw:
**Fable-5 adversarial self-review** (at 745542ee2):

Hunt list: (a) nil merchant_account reaching the new marker check, (b) marker recorded on the shared Gumroad platform account poisoning unrelated sellers, (c) rescue-scope variable availability, (d) test-mode QA lockout, (e) webhook behavior change.

- (a) Both call sites of `usd_settling_merchant_account?` sit behind `merchant_account&.stripe_charge_processor?` guards (eligibility L119/L191, quote L167), so the receiver is always non-nil there; the recording helpers use safe navigation anyway.
- (b) The platform account CAN carry a marker — but the FX quote for platform-routed sellers is made against that same account (`stripe_account_id` = its merchant id), so if Stripe rejects it once it will reject it for every seller routed through it; caching that is correct, not poisoning. And the platform account settles USD in practice, so the marker will simply never be recorded there.
- (c) `SettlementCurrencyMismatch` can only originate at `StripeFxQuote.create`, which executes after the `merchant_account` local is assigned in `BuyerCurrencyQuote#create` — the rescue never sees an unassigned local. In `MethodForcedPresentment` it's an attr_reader.
- (d) Test mode: markers are only recorded when Stripe actually rejects a quote, which doesn't happen against test accounts; no QA lockout. If one were ever recorded, `clear_settlement_currency_mismatch!` in a console clears it.
- (e) `handle_stripe_event_account_updated` refactor is behavior-preserving: same lookup, same record passed to `refresh_payment_method_availability`, clear step isolated in its own rescue so compliance-info handling can't be skipped by a bookkeeping failure.

P1/P2 findings: none. One accepted trade-off, flagged for the record: the 30-day TTL means an account that DISABLES multi-currency settlement without touching default_currency/external_accounts waits out the TTL before regaining presentment — acceptable because the fallback is the status-quo USD checkout, and Stripe sends account.updated with external_accounts previous_attributes for the bank-account change that usually accompanies that switch.

Verdict: ready pending CI.

PR #5996, opened 2026-07-20
Stop alerting Sentry on expected signup fraud-gate rejections in OAuth flows

gumclaw:
## What

Stops reporting `ActiveRecord::RecordInvalid` to Sentry when a Google or Apple OAuth sign-in is rejected by one of the signup fraud gates. The three fraud-gate validations (gmail-abuse variant filter, blocked email domains, blocked signup IPs) now tag their error with a `:blocked_signup` symbol, and the OAuth `rescue ActiveRecord::RecordInvalid` blocks skip `ErrorNotifier` when that tag is present. All other validation failures still alert.

## Why

Sentry issue [GUMROAD 7367902684](https://gumroad-to.sentry.io/issues/7367902684/) — `ActiveRecord::RecordInvalid: Validation failed: Something went wrong.` in `User::OmniauthCallbacksController#google_oauth2` — has accumulated 1,443 events across 156 users since March. The stacktrace lands in `User.find_or_create_for_google_oauth2` → `query_google` → `user.save!`: a visitor whose email is blocked by a signup fraud gate signs in with Google, the gate's validation fails the save, and the rescue reports it to Sentry.

That is the fraud control working exactly as designed (the visitor deliberately sees a vague "Something went wrong."), so paging Sentry on every occurrence is pure noise that buries real validation bugs in the same path. Same pattern as #5991 (expected bank-account rejections during Stripe account creation).

The Apple OAuth path has the identical rescue, so it gets the same guard.

## Test Results

Full local run of the touched suites on this branch (real `bundle exec rspec` output):

```
$ bundle exec rspec spec/modules/user/social_google_spec.rb spec/models/concerns/user/email_normalization_spec.rb

  User::SocialGoogle
  User::EmailNormalization

Finished in 6.25 seconds (files took 9.75 seconds to load)
41 examples, 0 failures
```

```
$ bundle exec rspec spec/models/concerns/user/social_apple_spec.rb

Finished in 6.06 seconds (files took 3.08 seconds to load)
8 examples, 0 failures
```

**49 examples, 0 failures** across the three affected suites.

- New spec: Google OAuth sign-in with a gmail-abuse-blocked email returns `nil` **without** calling `ErrorNotifier.notify` — verified load-bearing (fails red with the guard reverted).
- New spec: `RecordInvalid` from any other validation **still** alerts.
- New specs for `User.blocked_signup_error?` (gate rejection → true, unrelated validation → false, non-record exception → false).
[...]

PR #6036, opened 2026-07-20
Fix ActiveRecord::RecordInvalid lockout in Google OAuth when synced email belongs to another account

gumclaw:
## What

Google OAuth sign-in no longer fails with `ActiveRecord::RecordInvalid: Validation failed: An account already exists with this email.` when the email Google reports for a returning user is already taken by a different Gumroad account. `User.query_google` now skips the email sync for persisted users when the incoming address belongs to another account, keeping the user's current email and letting the sign-in complete. Email sync behavior is unchanged in every other case, and new-record creation still surfaces conflicts as validation failures (the caller already looks users up by email, so a conflict there is a genuine race worth reporting).

## Why

Sentry issue [GUMROAD 7367312496](https://gumroad-to.sentry.io/issues/7367312496/) — 674 events / 115 users since March 27.

`query_google` always adopts the email from Google's payload ("it may have changed on Google's side"). But when that address already belongs to a *different* Gumroad account — e.g. the user has two accounts, changed their Google primary email, or someone signed up separately with that address — the `email_almost_unique` validation fails, `save!` raises, `find_or_create_for_google_oauth2` returns `nil`, and the visitor is bounced to the signup page with "Sorry, something went wrong. Please try again." Every retry fails the same way: the user is permanently locked out of Google sign-in for that account, and each attempt pages Sentry.

Keeping the existing email restores sign-in (the account is matched by `google_uid`, not email) and removes the recurring alert. The Apple OAuth path is unaffected — it only assigns email at account creation, never on subsequent sign-ins.

Related: #5996 covers a *different* recurring `RecordInvalid` in the same controller (fraud-gate "Something went wrong." rejections, Sentry 7367902684); the two fixes are independent.

## Before / After

Not a UI change — pure model/backend behavior in the OAuth callback path. No rendered surface changes; the user-visible effect is that a previously failing Google sign-in now succeeds. Spec run below is the artifact.

## QA steps

Pure backend change in the OAuth callback model path — no preview app needed (no rendered surface changes; the spec run above is the artifact). To verify manually on a preview/local app:

1. Create account A and sign in with Google once (account gets `google_uid` set, email X).
2. Create account B with email Y via normal signup.
3. Change account A's Google primary email to Y on Google's side (or simulate by stubbing the omniauth payload).
4. Sign in with Google as account A. **Before:** "Sorry, something went wrong. Please try again." and a Sentry `RecordInvalid` event. **After:** sign-in succeeds; account A keeps email X; account B is untouched.

## Test plan / Test results

New specs in `spec/modules/user/social_google_spec.rb` (`"when the new email from Google already belongs to a different account"`):
- keeps the user's existing email instead of failing the save (red before the fix: `ActiveRecord::RecordInvalid`)
- does not raise or report to `ErrorNotifier`
- full `find_or_create_for_google_oauth2` flow returns the uid-matched user, both accounts keep their emails

Full file green locally:

spec run

`rubocop -a` clean on both changed files.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent) from a Sentry issue-alert webhook. Instructions: triage the Sentry issue, reproduce the failure with a spec, fix the root cause, and open a PR with local test evidence.


gumclaw:

- The guard only applies to `user.persisted?` — new-record creation keeps the old behavior on purpose. `find_or_create_for_google_oauth2` already looks users up by email before building a new record, so an email conflict on a brand-new user is a concurrent-signup race, and failing the save (with the Sentry report) is the right outcome there.
- `User.by_email(email)` is the same scope the `email_almost_unique` validation uses (`app/modules/user/validations.rb`), so the guard and the validation can't disagree about what counts as "taken". The `where.not(id: user.id)` mirrors the validation's `email_changed?` short-circuit for the self-match case.
- The existing "email differs only by case" behavior is untouched — that path already avoided the raise because `email_almost_unique` matches case-insensitively via the DB collation, and its spec still passes.
- Apple OAuth (`social_apple.rb`) doesn't need the same fix: it only assigns email when `new_user` is true, never on returning sign-ins.
- Spec ordering note: the new "does not raise or report" example uses a bare `not_to raise_error` (no class argument) to avoid the RSpec false-positive warning.

gumclaw:
CI failure was a semantic conflict with main, not related to this diff: main's #5999 follow-up added a settlement-mismatch spec to `buyer_currency_quote_spec.rb` after this branch forked, and the shard picked up the stale branch copy still calling `described_class.create(products:)` (the signature moved to `line_items:`). Merged in the latest main; the spec file now matches main and passes locally (25 examples, 0 failures). No changes to the OAuth fix itself.

gumclaw:

ershad:
One adjacent observation while reviewing (not a blocker for this PR):

The lockout exists because `query_google` re-syncs the email from Google on *every* login, not just at signup. A side effect is that a user who changes their Gumroad email away from their Google address has that change silently reverted on their next Google sign-in (the account is matched by `google_uid`, so `user.email != google_email` → it gets overwritten back), except in the conflict case this PR now guards.

**Suggested solution (follow-up):** Since the account is always matched by `google_uid`, the email only needs to be set at account creation — mirroring the Apple path, which assigns email only when `new_user`. Setting it once with the same "set if blank" pattern already used for `google_uid` and `name` just above:

```ruby
user.email = email if user.email.blank? && EmailFormatValidator.valid?(email)
```

This eliminates the whole class of `RecordInvalid` on sign-in (email never mutates on a returning login, so the validation can't fire), stops clobbering deliberate email changes, and makes the guard added here unnecessary. The tradeoff is that a Google-side email change would no longer auto-propagate to Gumroad — which is a product call, but matches how the Apple path already behaves.

gumclaw:
Agreed — the guard treated the symptom; setting the email once at creation removes the whole class. Opened #6142 implementing exactly this: `user.email = email if user.email.blank? && EmailFormatValidator.valid?(email)`, mirroring the Apple path, with the conflict guard from this PR removed (unreachable once the email never mutates on returning logins) and specs updated to the new contract. Since you flagged the no-auto-propagation tradeoff as a product call, you're on it as reviewer — the merge decision on that tradeoff is yours.

PR #6037, opened 2026-07-20
Fix Errno::ENOENT crash during product publish (content moderation variant generation)

gumclaw:
## What

Stops `ContentModeration::ContentExtractor` from generating image variants while collecting product image URLs for moderation. Cover images now use `AssetPreview#url(style: :original)` and thumbnails use `Thumbnail#url(variant: :original)` instead of the default display styles, which lazily generate a retina/thumbnail variant when one doesn't exist yet.

## Why

Production `LinksController#publish` requests crash with:

```
Errno::ENOENT: No such file or directory @ rb_file_s_size - /tmp/image_processing....png
```

Sentry issue 7440097747 — 341 events / 78 sellers affected since April.

Root cause: the content-moderation validation on `Link` runs inside `publish!`'s save transaction. The extractor's default `url` calls trigger synchronous variant generation (`file.variant(...).processed`). Attaching the freshly generated variant inside the open transaction defers its S3 upload to `after_commit`; by the time that callback runs, the `image_processing` tempfile has already been deleted, so `File.size` in `ActiveStorage::Service::S3Service#upload` raises `Errno::ENOENT`. The event breadcrumbs show exactly this sequence: variant record INSERT inside the transaction → `COMMIT` → `service_upload.active_storage` → crash.

The product row has already committed at that point, so the seller's product actually publishes but they get a 500 — confusing and noisy.

Moderation only needs the image contents; originals are both safe (no attach inside the transaction) and cheaper (no variant generation on the publish request path).

## Before / After

Not a UI change — backend-only fix on the publish request path. No adjacent UI surface changes; the spec run below is the artifact.

## Test plan / Test results

- New regression spec asserts extraction never calls `AssetPreview#retina_variant` / `Thumbnail#thumbnail_variant`. Verified red against the old code, green with the fix.
- Full moderation service specs pass locally:

test results

```
spec/services/content_moderation/content_extractor_spec.rb
spec/services/content_moderation/moderate_record_service_spec.rb
34 examples, 0 failures
```

## QA steps

Backend-only; reviewer can verify by running:

```
bundle exec rspec spec/services/content_moderation/content_extractor_spec.rb spec/services/content_moderation/moderate_record_service_spec.rb
```

After deploy, watch Sentry issue 7440097747 — event rate should drop to zero.

---

AI disclosure: authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert. Instructions: triage the Sentry issue, find the root cause, write a failing test, fix, and open a PR per antiwork/gumroad contributing conventions.


gumclaw:

- `Thumbnail#url(variant: :original)` short-circuits to `cdn_url_for(file.url)` — no `thumbnail_variant` call, so no variant generation or attach can happen on this path. Unsplash-backed thumbnails still return `unsplash_url` via the early return, unchanged.
- `AssetPreview#url(style: :original)` goes through `url_from_file(style: :original)` → `file.url` under the existing `Rails.cache.fetch` — again no `retina_variant` call. Unsplash/oEmbed previews are unaffected (early returns), and oEmbed previews can't reach this code anyway since the extractor filters on attached image blobs.
- The breadcrumb trail on the Sentry event confirms the mechanism: variant-record + variant-blob INSERTs inside the publish transaction's savepoint → `COMMIT` → `service_upload.active_storage` → `Errno::ENOENT` at `File.size` in `S3Service#upload`. Classic "attach inside an open transaction defers the upload past the tempfile's lifetime".
- The regression spec was verified red against the old extractor (stash + run) and green with the fix.

Adversarial (fable-5) pass — findings:
- **P3 (pre-existing, not introduced):** `url_from_file` caches the URL under `attachment_<id>_<style>_url` without an expiry-aware key; S3 signed URLs expire, so moderation could occasionally fetch a stale URL. This predates the PR (the retina path had the same cache) and failure mode is a skipped image fetch by the moderation strategy, not a crash. Not blocking.
- **P3 (pre-existing):** extractor uses `product.thumbnail` rather than `thumbnail_alive`, so a soft-deleted thumbnail can still be moderated. Harmless (over-moderation), out of scope.
- No P1/P2 findings — the change is a strict narrowing (never generate variants on this path) with behavior otherwise identical.

PR #6038, opened 2026-07-20
Retry streamed store-agent tool calls whose JSON arrives corrupted (Ai::AnthropicClient::Error: unreadable tool call)

gumclaw:
## What

Streamed tool calls whose JSON arrives unparseable on a *completed* turn (a turn that delivered its `stop_reason`) are now raised as `Ai::AnthropicClient::TransientError` instead of a plain `Error`, so the client's existing retry loop re-requests the turn. The retry only fires while nothing has streamed to the seller yet (same guard as other dropped-stream retries), and a call that stays unreadable across all 3 attempts still surfaces the same "Anthropic produced an unreadable tool call" failure rather than dispatching a lossy `{}` tool call. The `max_tokens` truncation path (drop the block, keep the stop_reason) is unchanged.

## Why

Sentry issue [GUMROAD 7609201485](https://gumroad-to.sentry.io/issues/7609201485/) — `Ai::AnthropicClient::Error: Anthropic produced an unreadable tool call for api_write` — has 101 events across 10 sellers since 2026-07-14. Every event's underlying cause is `JSON::ParserError: expected ',' or '}' after object value, got: EOF` at varying byte offsets: the tool call's argument JSON is cut off mid-object even though the stream delivered a closing `stop_reason`.

The client's existing handling assumed a completed turn with unparseable tool arguments meant a deterministic model bug, so it failed the seller's turn immediately. That assumption held against Anthropic directly, but production traffic now routes through OpenRouter's Anthropic-compatible gateway, which can lose or truncate `input_json_delta` fragments in transit while still delivering the terminating `message_delta` — so a "completed" turn no longer proves the arguments arrived intact. The varying cut-off offsets and the fact that identical seller requests succeed on retry point at transport corruption, not model misbehavior. The issue's first occurrence (2026-07-14) lines up with the OpenRouter routing change (#5833) shipping.

## Before / After

Not a UI change — the store agent's error surface is the only visible effect. Before: a corrupted tool call ended the seller's turn with a hard error event. After: the turn is transparently re-requested (up to 2 retries within the existing sleep budget); the error only surfaces if every attempt fails. Known limitation: turns that streamed preamble text before the tool_use block are never retried (retrying would replay the reply on the seller's screen), so those still surface immediately — pinned by a spec; a service-level retry for that shape is a possible follow-up if post-deploy Sentry data shows meaningful residual volume.

Walkthrough video (diff + full spec run):

https://github.com/antiwork/gumroad/blob/fix/retry-truncated-streamed-tool-call-json/qa-media/pr-anthropic-retry-walkthrough.mp4?raw=true

## QA steps

Pure backend change to the streaming LLM client — there is no adjacent UI surface to exercise (the fix only changes which exception class an already-failing internal path raises, gated behind the existing retry loop). Reviewer verification path:

1. Check out the branch and run `bundle exec rspec spec/services/ai/anthropic_client_spec.rb` — the two new examples at the bottom of the `#stream_messages` block cover retry-then-succeed and retries-exhausted.
2. Stash `app/services/ai/anthropic_client.rb` and re-run — both new examples fail against the pre-fix client, proving they're load-bearing.
3. After deploy, watch Sentry issue 7609201485: event volume should drop to (near) zero, since single-shot corruption now recovers silently.

## Test plan

- New spec: a completed turn with cut-off tool-call JSON is retried and succeeds on the re-request (sleeps once, returns the intact tool call).
- New spec: a tool call that stays unreadable across all 3 attempts raises the same `unreadable tool call` error (asserts 3 HTTP requests were made).
- Both new specs fail against the pre-fix client (verified by stashing the app change).
- Existing coverage unchanged: max_tokens truncation drop, nil-stop_reason mid-stream retry, no-retry-after-first-token guard.

## Test results

`bundle exec rspec spec/services/ai/anthropic_client_spec.rb spec/services/ai/store_agent_service_spec.rb` — 68 examples, 0 failures:

test results

---

AI disclosure: authored by Claude Fable 5 (Hermes Agent), triggered by the Sentry alert webhook for issue 7609201485. Instructions: triage the Sentry issue, reproduce the root cause, write a failing test, fix, and open a PR with local spec evidence.


gumclaw:

- **Why `TransientError` and not a bounded special-case retry:** the client already distinguishes retryable vs deterministic failures via the exception class, and `with_retries` in `#stream_messages` already carries the two constraints that matter here — never retry once output reached the seller (`retryable: -> { !yielded_any }`), and a global sleep budget (`RETRY_SLEEP_BUDGET_IN_SECONDS`) so the web request thread can't stack delays. Reusing that machinery keeps one retry policy instead of two.
- **The genuinely-bad-model case is still surfaced:** `TransientError < Error`, so after `MAX_ATTEMPTS` the caller sees the exact same `unreadable tool call` message and `Ai::StoreAgentService`'s error handling is unchanged. The new retries-exhausted spec pins this, including that exactly 3 HTTP requests are made.
- **`max_tokens` path untouched:** the `truncated` check still short-circuits before either raise, so token-cap truncation keeps its drop-the-block behavior (existing spec still green).
- **Message duplication:** the retryable raise re-states the "unreadable tool call" message rather than re-raising the original `Error`, because re-raising would preserve the non-retryable class. Same text keeps Sentry grouping stable for the exhausted-retries case.
- Evidence in the body: 68/68 specs green locally (client + store agent service); both new specs verified red against the pre-fix client.

gumclaw:
Fable adversarial pre-merge review — verdict: **APPROVE**, no P1s. Findings and disposition:

|---|---|---|
| P2 | Turns that stream preamble text before the tool_use block have `yielded_any = true`, so the retry is vetoed and the corruption still surfaces immediately — a fraction of the Sentry events may be text-preamble turns the retry won't absorb. That edge was untested. | **Fixed in `abac24c9f`**: added a spec pinning exactly-one-request/no-sleep behavior for a corrupted tool call after streamed text. Limitation also noted in the PR body. A service-level retry for preamble turns (feasible via the existing `:reset` emit) is a possible follow-up if post-deploy Sentry data shows the residual volume is mostly preamble turns. |
| P3 | All-fragments-dropped corruption leaves blank JSON → `{}` no-arg dispatch (pre-existing; blunted downstream by endpoint arg validation). | Not changed here — pre-existing behavior, out of scope for this fix; noted for a possible hardening follow-up (treat blank input on a tool_use block as corruption). |
| P3 | Class change re-fingerprints Sentry: issue 7609201485 stops accruing; exhausted-retry/preamble cases group under a new `TransientError` issue. | Expected and desirable (absorbed retries stop reporting; genuine failures still report). Will link the old issue to this PR and watch the residual. |
| P3 | Exhausted-retries spec asserted the parent `Error` class. | **Fixed in `abac24c9f`**: now asserts `TransientError` explicitly. |

Verified clean: no double-charged sleep budget, no infinite-retry path, per-attempt stream state is fresh, no outer retry loop that could replay a reply, both new specs fail against the pre-fix client. 33/33 client specs green after the review commit.

PR #6039, opened 2026-07-20
Fix Rack::Timeout crashes on sales API page_key pagination by adding a query timeout guard

gumclaw:
## What

Adds a query execution timeout guard to the `page_key` pagination path in `Api::V2::SalesController#index`, mirroring the protection already present on the deprecated `page` path.

- Wraps the `page_key` query path (including `for_sales_api_ordered_by_date` and its preloads) in `WithMaxExecutionTime.timeout_queries`
- Timeout is runtime-configurable via a new Redis key `api_v2_sales_page_key_query_timeout` (default 15s)
- On timeout, returns a 400 with actionable guidance ("narrow your date range with 'after'/'before' or filter by product_id") instead of an unhandled crash

## Why

The `page_key` path executes a UNION subquery over the seller's entire purchase history with no execution time limit. For sellers with very large sales volumes and broad filters, this query can run for minutes, until Rack::Timeout kills the Puma worker at 120 seconds. This surfaces in Sentry as `Rack::Timeout::RequestTimeoutException` on `Api::V2::SalesController#index` — 1,347 events affecting 117 users since March 2026 (Sentry issue [7367975104](https://gumroad-to.sentry.io/issues/7367975104/)).

Each occurrence is worse than a failed request: the worker process gets SIGTERMed mid-request, costing serving capacity. The deprecated `page` path already carries this exact guard; this brings the recommended `page_key` path to parity. Failing fast with a clear 400 also gives API consumers a signal to narrow their query instead of retrying the same 120-second death loop.

Applying the fixability checklist for this timeout class: no retry is appropriate (retrying the same unbounded query just reproduces the timeout), so the fix is to bound the query and convert the crash into a handled, actionable client error.

## Walkthrough

Terminal walkthrough (diff + timeout specs passing):

https://github.com/antiwork/gumroad/blob/fix/sales-api-page-key-query-timeout-guard/qa-media/pr-6039-sales-api-page-key-timeout-walkthrough.mp4?raw=true

## Before / After

Not a UI change — pure API/backend behavior, no adjacent user-facing surface. Before: requests hitting slow queries died with an opaque 5xx after 120s and killed the worker. After: they return within the configured limit (15s default) with a 400 and guidance.

## QA steps

Pure API change — QA via curl against the preview app (deployed via the `preview` label; URL from the deployments API once the Buildkite build finishes):

1. Get an API access token with the `view_sales` scope on the preview app (any seeded seller).
2. Happy path: `curl "https://<preview-host>/v2/sales?access_token=<token>"` → 200 with a `sales` array and (if >10 sales) a `next_page_key`.
3. Follow pagination: `curl "https://<preview-host>/v2/sales?access_token=<token>&page_key=<next_page_key>"` → 200, next page.
4. Confirm the guard is active: on the preview Rails console, `$redis.set(RedisKey.api_v2_sales_page_key_query_timeout, 1)`, then repeat step 2 — small preview datasets return well under 1s, so the response is still a 200 (guard wraps the query without changing results). The timeout→400 path itself is exercised by the new controller specs (it can't be reproduced with preview-sized data). Clean up with `$redis.del(RedisKey.api_v2_sales_page_key_query_timeout)`.

## Test plan

- New specs in `spec/controllers/api/v2/sales_controller_spec.rb`:
  - returns a 400 with the guidance message when the query guard raises `QueryTimeoutError`
  - asserts the `page_key` path actually runs inside `WithMaxExecutionTime.timeout_queries(seconds: 15)`
  - asserts the Redis key overrides the default timeout
- Full spec file run locally: **72 examples, 0 failures**
- CI green
- After deploy: watch Sentry issue 7367975104 — event rate should drop to zero, replaced by 400 responses visible in request logs

## Test Results

```
Finished in 44.88 seconds (files took 3.18 seconds to load)
72 examples, 0 failures
```

(`bundle exec rspec spec/controllers/api/v2/sales_controller_spec.rb`, local run; rubocop clean on all three changed files.)

---

*AI disclosure: This PR was authored by Claude Fable 5 (Hermes Agent), triggered automatically by a Sentry alert webhook for `Rack::Timeout::RequestTimeoutException` on `Api::V2::SalesController#index`. Instructions given: triage the Sentry issue, identify the root cause from the stacktrace, write a failing test, fix, and open a PR with the affected specs passing locally.*


gumclaw:

- The `begin/rescue` block deliberately includes the `preload` and the `success_with_object` render, not just the initial query build. ActiveRecord executes these queries lazily, so wrapping only the scope construction would guard nothing — this matches the structure of the existing guard on the deprecated `page` path a few lines above.
- `($redis.get(...) || 15).to_i` mirrors the deprecated path's pattern exactly (including its behavior when the key holds a non-numeric value); kept identical intentionally so the two knobs behave the same way for whoever operates them.
- The rescue converts the timeout to a 400 rather than 503: the request is well-formed but too expensive as filtered, and the actionable remedy (narrow `after`/`before` or add `product_id`) is on the client side.

Fable-5 adversarial review (pre-merge gate) — verdict on the full diff at head:

- **P1/P2: none found.** The guard is behavior-preserving on the happy path (verified by the 69 pre-existing examples in the spec file all passing unmodified), and the only new failure mode (timeout → 400) is covered by a dedicated spec.
- **P3 (noted, no change needed):** the new spec titles say "page_key query" but the guard actually covers the whole non-deprecated index path, including first-page requests without a `page_key` param. That's the correct scope for the fix (first-page requests are exactly the ones timing out in Sentry — the guard must not require the param), so the code is right; only the test naming is slightly narrower than the coverage.
- Checked: no other callers of `for_sales_api_ordered_by_date` (single call site), Redis key name collision (none), and that `WithMaxExecutionTime::QueryTimeoutError` is the class actually raised by the helper.

PR #6041, opened 2026-07-20
Fix NoMethodError on product page when product refund policy record is missing

gumclaw:
## What

Guards two call sites that dereference `product.product_refund_policy` behind the `product_refund_policy_enabled` flag without checking the record exists:

- `ProductPresenter::ProductProps#refund_policy_props` — product page rendering. Returns `nil` (no refund policy section) when the record is missing, instead of raising `NoMethodError: undefined method 'title' for nil`.
- `Purchase::CreateService` — the identical unguarded pattern at purchase creation; the purchase is now created without a `purchase_refund_policy` when the record is missing, instead of raising at checkout.

## Why

Sentry issue GUMROAD (7366766471): `NoMethodError` in `LinksController#show` — 1,107 events / 176 users since March 2026. The `product_refund_policy_enabled` flag can be out of sync with the underlying record (row deleted or never created), and the presenter dereferenced `product.product_refund_policy.title` unconditionally. `Product::AsJson#refund_policy_json` already applies the same `.present?` guard for exactly this reason; this brings the product-page presenter in line.

## Before / After

Not a UI change for products with valid data. Before: affected product pages returned a 500. After: the page renders normally with no refund policy section (same output as products with the flag disabled).

Walkthrough video (diff + red/green spec evidence):

https://github.com/antiwork/gumroad/blob/fix/refund-policy-props-nil-guard/qa-media/pr-refund-policy-nil-guard-walkthrough.mp4?raw=true

## QA steps

Pure backend/presenter fix — no preview click-path changes for healthy products. To verify on a preview app console (`preview` label):
1. Pick/create a product, set `product.update!(product_refund_policy_enabled: true)` and ensure it has no `ProductRefundPolicy` row (`product.product_refund_policy&.destroy!`).
2. Visit the product page (`/l/<permalink>`): before this fix it 500s; with the fix it renders with no refund policy section.
3. A product with a real refund policy row still shows its policy unchanged.

## Test plan

- New presenter spec: "when the refund policy record is missing despite the flag being enabled → returns nil instead of raising". Confirmed RED against unmodified code (reproduces the production `NoMethodError`), GREEN with the fix.
- New `Purchase::CreateService` spec: same missing-record setup → purchase created without a `purchase_refund_policy`, no error. Confirmed RED without the guard (`NoMethodError: undefined method 'max_refund_period_in_days' for nil`), GREEN with it.
- Full presenter spec file passes locally: 29 examples, 0 failures. Refund-policy contexts of both spec files: 11 examples, 0 failures.
- RuboCop clean on all changed files.

## Test Results

```
$ bundle exec rspec spec/presenters/product_presenter/product_props_spec.rb
29 examples, 0 failures

$ bundle exec rspec spec/services/purchase/create_service_spec.rb -e "refund policy"
11 examples, 0 failures (combined with presenter refund-policy contexts)

$ bundle exec rubocop app/presenters/product_presenter/product_props.rb spec/presenters/product_presenter/product_props_spec.rb
```

---

AI disclosure: authored by Claude Fable 5 (Hermes Agent, gumclaw), triggered by a Sentry alert webhook. Instructions: triage the Sentry issue, reproduce with a failing spec, fix the root cause, verify specs green, open a small focused PR.


gumclaw:

- **Why `.present?` and not `&.`:** matching the existing pattern in `Product::AsJson#refund_policy_json`, which guards the same flag/record mismatch with `product_refund_policy_enabled? && product_refund_policy.present?`. Using `&.` on each dereference would produce a half-populated hash (`title: nil`) instead of cleanly falling through to no policy.
- **Presenter fallthrough:** when the record is missing the `elsif` now fails and the method returns `nil`, which is exactly what the frontend already handles for products with the flag disabled (no refund policy section rendered). No frontend change needed.
- **`Purchase::CreateService`:** same latent bug at checkout — included here because it's the identical one-line pattern with a shared root cause, and a purchase against an affected product would 500 at a money path. The purchase proceeds without a `purchase_refund_policy`, which is the same state as purchases of flag-disabled products.
- **Seller branch not touched:** `User` has `after_create :create_refund_policy!`, so `seller.refund_policy` is guaranteed present — no guard needed there.
- **VCR cassette:** new cassette recorded for the CreateService spec; Stripe key is filtered (`Bearer <STRIPE_API_KEY>`).
- Both new specs were verified RED against the unguarded code (reproducing the exact production `NoMethodError`) before the fix, GREEN after.

Auto-merge intentionally left OFF: the second commit touches purchase creation (payments surface) — human merge per policy.

slavingia:
Would be good to do a rollout flag audit to cleanup the code for things that have been 100% in prod and g2g for a while

gumclaw:
Done — full audit here: https://github.com/antiwork/gumroad-private/issues/1208

Read live Flipper state for all 75 flags referenced in code (audited console request `20260721T025828Z-d0d55a50`). Summary:

- **16 flags** are fully on in prod with code stable for months — gating code can be removed. Queued as one small PR each, starting with the smallest: #6062 (`google_calendar_link`).
- **12 flags** are fully on but enabled too recently to call stable (brand_accounts, passkeys, buyer_local_currency, etc.) — re-audit in ~60 days.
- **~28 flags** are operational kill switches or provider toggles (fraud blocks, PayPal/Stripe disables, Resend switchover set) — keeping those regardless of state.
- **12 flags** are referenced in code but never registered in prod, so their gates always return false — those need a per-flag decision (delete the dead branch vs. the feature was silently never turned on), listed in the issue.
- Partial rollouts in flight (percentage/actor gates) are untouched.

Two judgment calls flagged in the issue for you: whether `content_moderation` should keep its flag as a moderation kill switch, and confirming the AU backtaxes campaign is finished before removing `au_backtaxes`.

gumclaw:

PR #6042, opened 2026-07-20
Fix ActiveRecord::StatementInvalid when license verify receives non-scalar params

gumclaw:
## What

Validates that `license_key`, `product_id`, `product_permalink`, `id`, and `link_id` are scalar strings in `Api::V2::LicensesController#clean_params` for the public `POST /v2/licenses/verify` endpoint. Requests carrying a nested value (e.g. `license_key[seriale]=NULL`) now get a 400 Bad Request with a clear message naming the offending parameter, instead of crashing.

## Why

When a client sends a nested parameter like `license_key[foo]=bar`, Rails parses it as `ActionController::Parameters` instead of a `String`. That object was passed straight into `find_by(serial: params[:license_key])`, where ActiveRecord cannot bind it, raising `ActiveRecord::StatementInvalid: TypeError: can't quote ActionController::Parameters` and returning an unhelpful HTTP 500 to the API caller.

Sentry issue GUMROAD (issue 7371152065): 84 events / 33 users since March 2026, still recurring as of today on `Api::V2::LicensesController#verify`. Malformed client requests are the trigger; a 400 with a descriptive message is the correct contract for a public API endpoint.

Supersedes #4156 (same fix, auto-closed as stale before it could land).

## Before / After

Not a UI change — public API endpoint only, no adjacent user-facing surface. Before: HTTP 500 (`ActiveRecord::StatementInvalid`). After: HTTP 400 with `{"success": false, "message": "The 'license_key' parameter must be a string."}`.

## QA steps

Pure API/backend endpoint — no browser surface. Reviewer can verify with curl against a preview or local app:

1. `curl -s -X POST "<host>/v2/licenses/verify" -d "product_id=<any_id>" -d "license_key[foo]=bar"` → expect HTTP 400 with `{"success":false,"message":"The 'license_key' parameter must be a string."}` (previously HTTP 500).
2. `curl -s -X POST "<host>/v2/licenses/verify" -d "product_id=<valid_product_external_id>" -d "license_key=<valid_serial>"` → expect the normal 200 success response (happy path unchanged).

## Test plan

- New controller specs cover non-scalar `license_key` and `product_id`, asserting the 400 status and message. Verified load-bearing: both specs fail with the controller change reverted (500/404 instead of 400).
- Full `spec/controllers/api/v2/licenses_controller_spec.rb`: **119 examples, 0 failures** locally.
- CI green.

## Test Results

specs passing

---

AI disclosure: Authored by Claude Fable 5 (Hermes agent) in response to a Sentry alert webhook for issue 7371152065. Instructions: triage the Sentry issue, reproduce via a failing spec, fix the root cause, run the affected specs locally, and open a PR assigned per recent file ownership.


gumclaw:
**Self-review + fable-5 adversarial review (pre-merge gate)**

Hunted for ways this guard fails or changes behavior for legitimate callers:

- **Arrays as well as hashes**: the triggering Sentry event actually sent `license_key` as an *array* of hashes (`license_key[][seriale]=NULL`), not a plain nested hash. The `is_a?(String)` check rejects both `ActionController::Parameters` and `Array` — covered.
- **Halting semantics**: `return render ...` inside the `each` block returns from `clean_params` itself, and a `render` in a `before_action` halts the filter chain, so `fetch_valid_license` never sees the bad value. Verified by the failing-then-passing spec (500/404 → 400).
- **Legacy param aliases** (`id`, `link_id`, `product_permalink`) are validated *before* the coalescing line that folds them into `product_permalink`, so a nested legacy alias can't smuggle a non-scalar through.
- **Blank-skip edge**: `next if params[key].blank?` skips empty values — form/query encoding can't produce a *present but empty* nested structure for these keys (`license_key[]=` yields `[""]`, which is present and rejected), so no unquotable value survives to `find_by`.
- **`increment_uses_count`** isn't in the guarded list, but it's only boolean-coerced (`["false", false].exclude?`), never bound into SQL — no crash path.
- **Happy path unchanged**: full `licenses_controller_spec.rb` green locally (119 examples, 0 failures).

Verdict: no P1/P2 findings. Behavior change is strictly 500 → 400 for malformed requests.

PR #6044, opened 2026-07-20
Fix N+1 subtitle and media location queries on the stream page playlist

gumclaw:
## What

Eliminates two per-video N+1 query patterns when building the stream-page video playlist (`UrlRedirect#video_files_playlist`, rendered by `UrlRedirectsController#stream`):

- **Subtitle files:** each video called `subtitle_files.alive`, issuing one `subtitle_files` query per video. Now all streamable files' `alive_subtitle_files` are preloaded in a single query, and `ProductFile#subtitle_files_urls` reads the preloadable association.
- **Media locations:** each video called `latest_media_location_for(purchase)`, issuing one `media_locations` query per video. Now the playlist reuses the existing `latest_media_locations_by_product_file_id` lookup (already used by the mobile serializer), which fetches every file's latest location in one query.

For a purchase with N videos this drops 2N queries to 2.

## Why

Sentry issue GUMROAD-70 (`N+1 Query` on `UrlRedirectsController#stream`, 768 events since March, still firing as of 2026-07-20) flags the repeated `SELECT subtitle_files.* ... WHERE product_file_id = ?` pattern on the stream page. Courses and multi-chapter video products load one playlist entry per video, so buyers of large products pay a linear query cost on every stream-page view.

Refs: Sentry GUMROAD-70. Same family as #5605 / #5994 (mobile url_redirect serializer N+1s).

## Before / After

Not a UI change — the playlist props are byte-identical (the new spec asserts tracks and latest-media-location values match the previous per-file lookups, including the `nil` case for videos with no consumption history and the sub-query MAX(consumed_at) semantics). Only the query shape changes.

**Walkthrough video** (regression spec red against main's models, green with the fix):

https://github.com/antiwork/gumroad/blob/aafc91681056538a2d016f0563a1872fe84820e3/qa-media/pr-6044-stream-playlist-n-plus-one-walkthrough.mp4?raw=true

## QA steps

The change is query-shape only; the reviewer-verifiable path on the preview app (`preview` label applied):

1. Log into the preview app as a buyer who owns a product with 2+ video files (or create one: product with two uploaded videos, purchase it with a test card).
2. Open the download page and click **Watch** on a video → `/s/<token>/<file_id>` (the `UrlRedirectsController#stream` page).
3. Confirm the playlist lists every video with its title, subtitles (if any), and resumes at the last-watched position — identical to production behavior.
4. Optionally confirm in the preview Rails console that `UrlRedirect#video_files_playlist` issues one `subtitle_files` and one `media_locations` SELECT regardless of video count.

## Test plan

- New regression spec `spec/models/url_redirect_spec.rb` — "bulk-loads subtitle files and media locations instead of querying once per video": counts `subtitle_files` / `media_locations` SELECTs via `ActiveSupport::Notifications` (expects exactly 1 each) and asserts the serialized tracks + latest media locations are unchanged. Verified load-bearing: fails with 4 subtitle queries + 2 media-location queries when the fix is stashed.
- Full `#video_files_playlist` block: 6 examples, 0 failures locally (covers rich-content ordering, variant ordering, installment ordering, empty/foreign-file `index_to_play` cases).
- `spec/models/product_file_spec.rb` / `spec/presenters/url_redirect_presenter_spec.rb` failures observed locally are pre-existing on the unmodified branch (local Minio object-store env, `Aws::S3::Errors::NotFound`) — verified by stashing the change and re-running; CI will exercise them properly.
- `rubocop -a` on all three changed files: no offenses.

```
UrlRedirect#video_files_playlist
6 examples, 0 failures
```

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent) from an automated Sentry webhook trigger for issue GUMROAD-70. Instructions: triage the Sentry N+1 performance issue on `UrlRedirectsController#stream`, identify the root cause from the flagged query, write a failing regression test, fix, and open a PR with the affected specs passing locally.


gumclaw:

- `app/models/url_redirect.rb` — `video_files_playlist` now filters to streamable files first, preloads `alive_subtitle_files` for all of them in one query via `ActiveRecord::Associations::Preloader`, and swaps the per-file `latest_media_location_for(purchase)` for the existing single-query `latest_media_locations_by_product_file_id` private lookup (same one the mobile serializer uses since #5605). Semantics preserved: that lookup returns `{}` when `purchase` is nil or an installment is present, matching `latest_media_location_for`'s `nil` guard (`link.nil?` is implied — playlist files always come from a link/installment context, and installment is separately guarded). `index_to_play` behavior unchanged since only streamable files ever entered the playlist before.
- `app/models/product_file.rb` — `subtitle_files_urls` reads `alive_subtitle_files` (the `has_many ... -> { alive }` association) instead of `subtitle_files.alive` so the preload actually takes effect. For single-file callers the query issued is identical.
- Spec counts SELECTs on both tables via `ActiveSupport::Notifications` and asserts serialized output (tracks + latest media locations, including nil cases) — verified it fails against main's models (4 subtitle + 2 media-location queries) and passes with the fix.

One judgment call: I preload only for the playlist path rather than changing `alive_product_files` to always include subtitles, to avoid dragging subtitle rows into every other caller (downloads, archives, mobile) that doesn't need them.

gumclaw:
## Fable-5 adversarial pre-merge review — verdict at `8ff990f26`

**Verdict: READY-TO-MERGE** (after fixing the one P2 found; fix is commit 8ff990f26).

The review hunted semantic equivalence of both bulk-load swaps, guard divergence, serialization shape, association equivalence, preloader/memoization interactions, and spec red/green validity.

| # | Severity | Finding | Disposition |
|---|----------|---------|-------------|
| 1 | P2 | Installment redirects that fall through to the **product's** files (installment has no alive files — routine via `Installment#generate_url_redirect_for_purchase`) lost resume positions: `latest_media_locations_by_product_file_id` returns `{}` whenever `installment.present?`, but the old per-file guard checked the **file's** ownership, not the redirect's. Videos that previously seeked to the last-watched position would start at 0, and the stream page would disagree with the download page for the same redirect. | **Fixed in 8ff990f26** — the playlist now builds the bulk map whenever `purchase.present?` and gates per file on `link_id.present? && installment_id.nil?`, matching the pre-existing `ProductFile#latest_media_location_for` guard. Regression spec added and red/greened against the pre-fix code (fails on `aafc916`, passes with the fix). |
| 2 | P3 | Tie-break among same-second cross-platform media-location rows may pick a different row than the old `ORDER BY consumed_at LIMIT 1` — but both were DB-arbitrary, so no determinism was lost. | Note only. |
| 3 | P3 | Legacy rows with NULL `consumed_at` (schema allows it) were returnable by the old ordering but are dropped by the `MAX()` equality join. Every current write path sets `consumed_at`, and the same scope already serves the download page and mobile paths in production. | Note only. |
| 4 | P3 | Sibling per-file `latest_media_location_for` loop in `Installment#installment_mobile_json_data` (mobile installment serializer) — same N+1 family, out of this PR's scope. | Follow-up candidate. |

**Confirmed safe** (adversarial claims that did NOT hold up):
- The `.as_json` "shape change" is a no-op — the old code already returned `latest_media_location.as_json` from `ProductFile#latest_media_location_for`, and `MediaLocation#as_json` is overridden to emit the same symbol-keyed `{location:, unit:, timestamp:}` hash. Wire format to `Stream.tsx` is byte-identical.
- `alive_subtitle_files` ≡ `subtitle_files.alive`: same class, same `alive` scope, no default_scope/ordering on `SubtitleFile`, and the association predates this PR (already used by the presenter). No caller mutates subtitles then re-serializes the same instances in one request.
- Preloader/memoization: `alive_product_files` memoization predates the PR; `.select(&:streamable?)` loads the relation exactly as the old `.each` did; Rails' Preloader skips already-loaded associations.
- The N+1 regression spec genuinely red/greens (4 per-file SELECTs on each table under the old code vs the asserted 1), and the query counter is robust under this app's config (no query log tags, no comment-prefixed SQL).
- The playlist has exactly one consumer (`UrlRedirectPresenter#stream_page_props` → Inertia props) and the JS null-guards `latest_media_location`.

**Verification on this head:** all 7 `#video_files_playlist` examples pass locally (including both new specs); rubocop clean on both touched files.

**Only live-verifiable:** how many production installment+purchase redirects have installments with no alive files (the P2's blast radius) — the fix makes it moot, but a prod count is available on request.

gumclaw:
Deployed to production (revision `0956074baf8d`, verified via x-revision header). Sentry issue [GUMROAD-70](https://gumroad-to.sentry.io/issues/7374788817/) (N+1 Query in UrlRedirectsController#stream) is now marked resolved — it will auto-reopen if the query pattern regresses. 768 events / 118 users affected since March 30; last event 11:07 UTC today, pre-deploy.

PR #6045, opened 2026-07-20
Fix flaky Price.alive spec: stray live prices leak from non-transactional concurrency spec

gumclaw:
## What
Fix the flaky spec at `spec/models/price_spec.rb:51` — "Price.alive excludes deleted prices".

## Why
Intermittent CI failure whenever this spec lands on the same knapsack node as `spec/services/purchase/handle_failed_refund_service_concurrency_spec.rb` (seen on PR #6038, run 29736954664, Test Fast 5):

```
Failure/Error: expect(Price.alive).to match_array([live_price])
  the extra elements were: [#<Price id: 1616, link_id: 982, ...deleted_at: nil...>]
```

The concurrency spec sets `self.use_transactional_tests = false` and cleans up manually in an `after` block, but `@product.destroy` removes the Link row but does not cascade to its prices (`has_many :prices` has no `dependent:` option), leaving `Price` rows with `deleted_at: nil`. Those stray live prices then pollute the global `Price.alive` assertion in any later spec on the same node.

## Changes
- `handle_failed_refund_service_concurrency_spec.rb`: hard-delete the product's Price rows in the `after` cleanup block.
- `price_spec.rb`: scope the `.alive` assertion to the spec's own product so it can't flake on leftover rows from other non-transactional specs, and explicitly assert the deleted price is excluded.

## Test Results
Both spec files pass locally, including in the polluting order (concurrency spec first, then price spec) — see `qa-media/pr-price-alive-flake-test-results.png`:
```
14 examples, 0 failures
```
Test-only changes — no production code modified.

## AI Disclosure
This PR was authored by Gumclaw (Hermes agent) with AI assistance; changes were verified by running the affected specs locally in both orders.

## QA Steps
1. `bundle exec rspec spec/services/purchase/handle_failed_refund_service_concurrency_spec.rb spec/models/price_spec.rb --order defined` — passes (screenshot in qa-media).
2. Confirm no app code touched: `git diff main --stat` shows spec files + qa-media only.


gumclaw:
Fable adversarial pre-merge review — verdict: **CHANGES-REQUIRED → fixed in `cd66711f`**, no P1s.

|---|---|---|
| P2 | The cleanup comment (and PR body) claimed `Link#destroy` soft-deletes the product. Wrong: `Deletable` only adds `mark_deleted!`/scopes and nothing overrides `destroy`, so the product row is hard-deleted — the prices leak because `has_many :prices` has no `dependent:` option. | Fixed in cd66711f (comment reworded) + PR body corrected |
| P3 | The same `after` block leaks one `UrlRedirect` row per successful purchase (`create_artifacts_and_send_receipt!`). No spec flakes on it today, but it's the identical bug class. | Fixed in cd66711f (`UrlRedirect.where(purchase_id: ...).delete_all`) |
| P3 | `not_to include(deleted_price)` is implied by the `match_array` above — documentation, not extra coverage. | Kept as explicit intent |

What the review **confirmed safe**:
- Price cleanup is complete: the factory creates exactly one price, no variants/rental prices, no STI on `prices`, and deletion is ordered before `@product.destroy`.
- The scoped `Price.alive.where(link: product)` assertion still red/greens if the `alive` scope is removed or broadened — the deleted price lives on the same product, so scoping only excludes other products' rows. Not a weakened spec.
- No other victims: the only other global Price assertions in the suite are `change { Price.count }` deltas, immune to pre-existing rows.
- No PaperTrail `versions` leak (versioning disabled unless a spec opts in).

Specs pass locally in the polluting order (`14 examples, 0 failures`) and rubocop is clean at cd66711f. Arming auto-merge.

PR #6047, opened 2026-07-20
Stop alerting Sentry on expected Japanese address rejections during Stripe account creation (Stripe::InvalidRequestError)

gumclaw:
## What

Stops reporting to Sentry when Stripe rejects a Japanese seller's address during Connect account creation with `Invalid address for Japan. We cannot find an address with town of X for postal_code Y.` Adds a `jp_address_invalid_error?` matcher in `StripeMerchantAccountManager` and skips `ErrorNotifier.notify` for it in `create_account`'s rescue, mirroring the existing suppression of bank-account and postal-code seller-input rejections. The error still re-raises, so the seller continues to see Stripe's message inline on the payments settings page.

## Why

Stripe validates Japanese kanji/kana addresses against its JP postal directory and rejects mismatches (street details typed into the town field, or a town that doesn't resolve under the given postal code). This is an expected seller-input error: the seller sees the exact Stripe message on `/settings/payments` and can fix the address themselves — nothing for an engineer to act on per occurrence. It is currently one of the noisiest Sentry issues: **6,580 events / 437 users since June 26** (issue GUMROAD 7578121692, culprit `Settings::PaymentsController#update`), drowning out real failures in the same `create_account` path.

Same pattern as the recently merged bank-rejection (#5991) and bank-directory (#6014) noise fixes, and sibling PRs #6024/#6008 for tax-ID and phone rejections. Stripe doesn't populate `code` on this rejection, so the matcher keys on the message, scoped to `Stripe::InvalidRequestError` only.

## Before / After

Not a UI change — error-reporting behavior only. The seller-facing flow is unchanged: the rejection still surfaces as the inline error message on the payments settings page (the exception re-raises and `Settings::PaymentsController#update` renders `e.message` exactly as before). The only delta is that Sentry no longer receives an event per occurrence.

## Test Results

Full spec file green locally (22 examples, 0 failures), including two new examples covering the Japanese address rejection (no Sentry notify + no spurious bank-sync payout note). Red-check verified: with the code change stashed, the new Sentry-suppression example fails.

spec run

## QA steps

Backend-only error-reporting change with no adjacent UI delta — the payments settings page renders identically (rejection message still shown inline). To verify: as a Japanese seller with an address Stripe can't resolve, submit `/settings/payments`; the inline Stripe error still appears, and no new Sentry event is created for this rejection shape.

---

AI disclosure: Authored by Claude Fable 5 (Hermes agent) from a Sentry webhook alert. Instructions: triage the Sentry issue, identify root cause, add a failing test, fix, and open a PR with test evidence.


gumclaw:

- The matcher is scoped to `Stripe::InvalidRequestError` only and keys on the message because Stripe returns no `code`/`param` for this rejection shape (verified against the production event payload). Two message fragments are matched (`invalid address for japan` and `cannot find an address with town of`) so a minor Stripe wording tweak on either half doesn't silently re-enable the noise.
- Suppression happens only inside the `merchant_account.charge_processor_alive_at.nil?` cleanup branch — the same place bank-account rejections are already suppressed — so a JP address error on an established account path would still notify.
- The exception still re-raises: seller-facing behavior on `/settings/payments` (inline Stripe message) is unchanged, and the `RetryStripeRejectedPayoutSetupsJob` postal-code retry path is untouched (this rejection carries no `postal_code_invalid` code, so no payout note is recorded — covered by the new spec).
- Red-check: with the manager change stashed, the new "does not report the rejection to Sentry" example fails, so the test is load-bearing.

gumclaw:
Pre-merge adversarial review (fable-5) — verdict: **no P1/P2 findings**.

Hunt list checked:

1. **Over-broad suppression** — could the regex swallow a real failure? `cannot find an address with town of` is Stripe's JP-postal-directory rejection phrasing; even if Stripe reused it for another country, it would still be a seller-input address rejection of the same class. Matcher requires `Stripe::InvalidRequestError`, so `CardError`/`APIError`/timeouts can never hit it.
2. **Suppression scope** — only applied in the `create_account` cleanup branch (fresh account, `charge_processor_alive_at` nil). `update_account` / `update_bank_account` notify paths untouched.
3. **Behavior change beyond reporting** — none: exception re-raises, controller error message unchanged, no payout note added/removed for this shape (spec asserts no bank-sync note; no `postal_code_invalid` code means no postal note, as before).
4. **Retry-job interaction** — `RetryStripeRejectedPayoutSetupsJob` keys on postal/bank failure notes; this rejection records neither, same as before the change. No new retry loop, no lost retry.
5. **Test quality** — red-check confirmed the Sentry assertion fails without the code change; message fixture uses a synthetic address (no PII).

P3 (noted, no change needed): if Sentry noise from the same message shape ever appears via `update_account`, the matcher is already extracted and can be reused there.

PR #6051, opened 2026-07-20
Port bundles/product_controller_spec to Minitest + fixtures (#5801)

gumclaw:
## What

Ports `spec/controllers/bundles/product_controller_spec.rb` — item 4 in the #5801 migration ranking (0:50 factory-create time, 88% of the file's wall time) — to Minitest + fixtures under `test/controllers/bundles/product_controller_test.rb`, and deletes the RSpec original. All 29 examples ported 1:1 (`rspec --dry-run` reports 29; the new file runs 29).

This is the first **controller** port in the Minitest suite, so it also adds the shared scaffolding controller ports need, in `test/support/controller_test_case.rb`:

- **Devise sign-in**: `Devise::Test::ControllerHelpers` mixed into `ActionController::TestCase`, matching the RSpec suite's `config.include ... type: :controller`.
- **CSRF handling**: the test env enables forgery protection globally (`config/environments/test.rb`) to exercise it in request/system tests. Controller tests have no rendered form to lift a token from, so every non-GET action would bounce to `/login`. rspec-rails solved this by disabling forgery protection around each controller example — the support file mirrors exactly that.
- **Inertia assertions**: instead of the `inertia_rails/rspec` matchers, the test sends the `X-Inertia: true` request header, which makes inertia_rails render the page object as JSON, and asserts on the parsed component/props.
- The `"authorize called for action"` shared example becomes a direct Mocha stub on `LinkPolicy.new` asserting the controller's `pundit_user` + record — same contract, no shared-example machinery.

## Why

Part of #5801 (factories → fixtures migration). Factory setup dominates this file's wall time; moving it onto fixtures + the shared `ModelFactories` builders removes that cost and moves its 29 examples into the lightweight `test_minitest` CI job.

**Measured locally (same machine, same DB):**

| | wall time |
|---|---|
| RSpec original (29 examples) | 95.6s |
| Minitest port (29 runs, 132 assertions) | 5.2s |

Refs #5801.

## Before / After

Not applicable — test-only change, no app code touched, no UI surface. The diff is: 1 spec deleted, 1 test added, 1 test-support file added, 1 qa-media asset.

## QA steps

1. `RAILS_ENV=test bin/rails test test/controllers/bundles/product_controller_test.rb` → 29 runs, 0 failures.
2. `RAILS_ENV=test bin/rails test` (whole `test/` tree, checking for fixture/global-state leakage into the 1,700 existing tests) → 1731 runs, 4090 assertions, 0 failures, 0 errors, 6 skips (pre-existing skips).
3. Confirm the deleted spec and the new test cover the same examples: both count 29.

## Test Results

test run

```
29 runs, 132 assertions, 0 failures, 0 errors, 0 skips   (new Minitest file, 5.7s)
1731 runs, 4090 assertions, 0 failures, 0 errors, 6 skips (full test/ tree, 272s)
29 examples, 0 failures                                   (deleted RSpec original, 95.6s — parity check)
```

---

## AI disclosure

Written by Claude Fable 5 (Gumclaw), triggered by the #5801 migration plan ("Next up: #4 `bundles/product_controller_spec`"). The port follows `docs/minitest-migration.md`: read the spec, map factory objects to fixtures/`ModelFactories` builders, run the new file plus the whole `test/` tree locally, delete the RSpec original, rubocop the new files.


gumclaw:
Self-review (adversarial pass on my own diff before merge):

- **Parity**: 29 RSpec examples → 29 Minitest runs, each mapped 1:1. The one structural change: the `"authorize called for action"` shared example is now a direct Mocha stub on `LinkPolicy.new`. Note the controller authorizes twice on `update` (the `Bundles::BaseController` `before_action :authorize_bundle` plus the explicit `authorize @bundle` in the action) — the assertion checks that at least one call carries `pundit_user` + the bundle, which is the same contract the shared example verified.
- **Refund-policy happy path**: the spec asserted `not_change` on `product_refund_policy_enabled` / policy title / fine print. That only holds because the seller has an account-level refund policy (factory `before_create` reads `seller_refund_policy_new_users_enabled`, which spec_helper activates). Fixtures bypass callbacks, so the setup sets `refund_policy_enabled` + creates the `SellerRefundPolicy` row explicitly and the port asserts the concrete unchanged values (`false` / `nil`) instead of `not_change` — equivalent, and it can't silently pass on the wrong branch.
- **Shared-fixture mutation**: setup mutates `named_seller` (created_at backdate for service products, refund flag, policy row) — all inside the test transaction, rolled back per test. Verified by running the whole `test/` tree: 1731 runs, 0 failures, no cross-file leakage.
- **Flipper leakage**: the `seller_refund_policy_disabled_for_all` test deactivates in `ensure`, on top of the per-test Redis flush #5956 added to `test_helper` — double-covered.
- **CSRF support file**: `test/support/controller_test_case.rb` reopens `ActionController::TestCase` globally. That's intentional shared scaffolding for the controller-port track (mirrors what rspec-rails did for every controller spec); it only touches controller tests, not integration/request tests, so the env-level forgery protection still exercises request paths.

No P1/P2 findings. Verdict: mergeable.

PR #6052, opened 2026-07-20
Fix .srt captions rendering at the right edge in iOS Safari video player

gumclaw:
Related to #6043 — a seller reported (Helper ticket linked on the issue) that external `.srt` caption tracks on purchased videos render pinned to the **right edge of the screen** on iOS Safari, while macOS browsers show them correctly centered at the bottom.

> ⚠️ **Status after hosted-preview QA:** this change does **not** fix the symptom on iOS Safari — JW Player 8.49.5's html5 provider force-enables native caption rendering on iOS/Safari regardless of `renderCaptionsNatively: false` ([details in the QA comment](https://github.com/antiwork/gumroad/pull/6052#issuecomment-5029511608)). Kept in draft; the "Fixes" linkage was removed so #6043 stays open. Alternative directions are listed in the QA comment (server-side `.srt` → `.vtt` normalization with explicit cue positioning is the most promising).

## Root cause

JW Player's `renderCaptionsNatively` setup option is browser-defaulted: **`true` on Safari, `false` on Chrome** ([JW docs](https://docs.jwplayer.com/players/reference/setup-options)). On Safari the player hands side-loaded caption tracks to WebKit's built-in text-track renderer, and iOS WebKit positions those cues at the right edge of the video instead of centered-bottom. Chrome (and every browser using the player's own renderer) is unaffected — which exactly matches the seller's device matrix (macOS fine, iOS broken).

## Fix (attempted)

Force `renderCaptionsNatively: false` centrally in `createJWPlayer` (`app/javascript/utils/jwPlayer.ts`), the single chokepoint all five player surfaces go through (download page, stream page, product edit preview, product cover video, review videos). Every browser now uses JW Player's own caption renderer — the same rendering path Chrome already uses everywhere and which renders correctly.

This mirrors the existing centralized `cast: false` override in the same function (gumroad-private#1051): a player-default that misbehaves in our context, corrected once at the chokepoint so no caller can drift.

Also guards the CDN library `import` behind a `typeof jwplayer === "undefined"` check — a no-op in production (first call still loads the library; subsequent calls already had it cached) that makes the util unit-testable without fetching the cloud library.

## Tests

New `app/javascript/utils/jwPlayer.test.ts` (vitest, 4 passing) locking in:
- setup receives the caller's playlist on the right container
- `cast: false` is always forced
- `renderCaptionsNatively: false` is always forced
- callers cannot re-enable either

```
✓ app/javascript/utils/jwPlayer.test.ts (4 tests) 6ms
Test Files  1 passed (1) / Tests  4 passed (4)
```

`tsc --noEmit`, `eslint`, and `prettier --check` clean on both files.

Captured on `https://fix-ios-caption-rendering.apps.staging.gumroad.org` (deployed rev `2b603a71e98e`), stream page `/s/<token>` with a seeded video + English `.srt` track, Playwright @ deviceScaleFactor 2.

| Chromium (control — fix takes effect) | WebKit / Safari engine (option overridden by JW provider) |
| --- | --- |
| Chromium: JW renderer draws the cue centered at the bottom | WebKit: JW caption overlay stays empty — native renderer still active, iOS symptom untouched |

Chromium: `getConfig().renderCaptionsNatively === false`, cue drawn by `.jw-captions`, centered-bottom ✅. WebKit (incl. iPhone UA): effective config reads back `true`, native `TextTrack` is `showing`, `.jw-captions` empty — the setup option is silently overridden on the exact browser the issue targets ❌.

## QA steps (hosted preview app)

Preview: `https://fix-ios-caption-rendering.apps.staging.gumroad.org`
1. Log in as `seller@gumroad.com`, open the seeded stream page `/s/<token>` (purchase confirm interstitial → enter purchase email).
2. Play the video, enable CC (English).
3. Chromium: caption renders centered-bottom via JW's renderer ✅.
4. WebKit/iOS Safari: WebKit's native renderer still draws the cue (JW forces native rendering on Safari/iOS at the provider level) — the right-edge symptom of #6043 remains ❌.

## AI disclosure

Prompts/instructions given: standing order (Sahil 2026-07-19) to autonomously build qualifying open issues; issue #6043 body used as the spec. QA + screenshots by the preview-screenshot watcher (claude-fable-5).


gumclaw:
**fable-5 adversarial pre-merge review — verdict: PASS (no P1/P2)**

Hunt list and findings:
- **(a) Does forcing the player renderer regress desktop Safari?** Behavior change acknowledged: desktop Safari swaps from WebKit's native caption box to JW Player's renderer — same path Chrome has always used on this codebase, visually equivalent (centered-bottom, styled by the player). No functional loss.
- **(b) Do OS-level caption accessibility preferences stop applying?** Yes — with native rendering, iOS/macOS system caption styles (size/color, the FCC-mandated settings) apply; with the player renderer they don't. Weighed against the current state (captions unreadable at the right edge on iOS — the accessibility failure is strictly worse), this is the right trade. JW Player's CC menu still allows per-player styling if we ever want to expose it. **P3, noted not fixed.**
- **(c) Can a caller re-enable native rendering or casting?** No — the override spreads `options` first, then forces both keys; locked in by a test.
- **(d) Does the `typeof jwplayer` import guard change production behavior?** No — first call still loads the CDN library (global undefined until then); subsequent calls previously hit the module cache anyway. Only effect is testability.
- **(e) All five player surfaces (download page, stream, product-edit preview, cover video, review videos) go through `createJWPlayer`** — verified by grep, no direct `jwplayer().setup` callers bypass the chokepoint.

Tests: 4/4 vitest passing; tsc/eslint/prettier clean. UI-touching → holding as draft for hosted-preview screenshots (iOS Safari CC on + desktop regression check) before any merge; preview deploy is up (buildkite green).

slavingia:
Screenshots plz

gumclaw:
## Hosted-preview QA result: the option is overridden on the exact browser this targets ⚠️

Ran caption QA on the hosted preview (`https://fix-ios-caption-rendering.apps.staging.gumroad.org`, deployed rev `2b603a71e98e`, seeded video + English `.srt` on the `demo` product, stream page `/s/<token>`). Evidence committed to `qa-media/` @ `ac52b4ec7`.

**Chromium (control) — works as intended.** Effective `getConfig().renderCaptionsNatively === false`, no native text track showing, cue drawn by JW's own renderer, centered at the bottom:

chromium JW renderer

**WebKit (Safari engine, incl. iPhone UA) — the option does not take effect.** Despite the deployed bundle passing `renderCaptionsNatively: false` into `setup()` (verified the string `renderCaptionsNatively:!1` in `staging-assets…/jwPlayer-Cr4xYtOb.js`), the live player reports:

```
{ effectiveRcn: true, environment: { safari: true, chrome: false, iOS: true } }   // iPhone UA
{ cfgRcn: true, textTracks: [{label: "English", mode: "showing"}], jwCaptionsText: "" }  // behavioral probe
```

The native `TextTrack` is `showing` and `.jw-captions` stays empty — WebKit's built-in renderer is still the one drawing cues, i.e. the iOS right-edge bug is untouched.

webkit native renderer still active — no JW-rendered caption

**Root cause of the override** — JW Player 8.49.5's html5 provider (`ssl.p.jwpcdn.com/player/v/8.49.5/provider.html5.js`) hardcodes:

```js
renderNatively: (v = t.renderCaptionsNatively, !!n.OS.iOS || !!n.Browser.safari || v && n.Browser.chrome)
```

`renderCaptionsNatively` is only consulted **on Chrome** (to opt *in* to native rendering); on iOS or Safari the provider forces native rendering unconditionally, and the model attribute is then overwritten from the provider (`t.set("renderCaptionsNatively", e.renderNatively)` in the core). So this setup option cannot fix the iOS symptom with the current player version.

**Suggested directions** (needs a different approach than this PR's mechanism):
1. Fix the cue geometry instead of the renderer: when JW side-loads `.srt` on Safari it parses cues itself and injects them into the native track — the right-edge symptom is WebKit rendering cues whose `position`/`align`/`line` are unset. Converting seller `.srt` → `.vtt` server-side with explicit `align:center position:50%` cue settings (or normalizing at the presenter that builds `tracks`) would correct the native path on iOS.
2. Check whether a newer self-hosted JW player build drops the Safari clause (the forced-native behavior is version-specific; the cloud library pins 8.49.5).
3. `::cue`-scoped CSS is NOT a fix for position (WebKit ignores positioning props in `::cue`).

Keeping this PR in **draft** — the change is a safe no-op-ish improvement on Chrome but does not resolve #6043 on the affected browser, so it shouldn't be handed to review as a fix for that issue.

<sub>QA run by the preview-screenshot watcher (claude-fable-5), Playwright WebKit + Chromium against the hosted preview app.</sub>

slavingia:
Report bug to jwplayer?

gumclaw:
> Report bug to jwplayer?

Checked the options:

- The open-source `jwplayer/jwplayer` GitHub repo is **archived** (read-only) — no new issues can be filed there.
- Bug reports now go through the JW Player support portal, which requires a licensed-account login. Our player uses a cloud-hosted library (`cdn.jwplayer.com/libraries/3vz4Z4wu.js`, pinned to 8.49.5), so whoever holds the JW/ JWP Connatix account credentials would need to file it.
- Worth noting they may call it intended: their docs describe `renderCaptionsNatively` as Chrome-only opt-in behavior on some paths, and the Safari/iOS force-native clause has been in the provider for multiple versions. The report to them would really be "iOS WebKit positions side-loaded cues at the right edge when you hand them to the native renderer, and you give us no way to opt out on iOS."

Meanwhile the fix that doesn't depend on JW shipping anything: normalize seller `.srt` → `.vtt` server-side with explicit `align:center position:50%` cue settings, which corrects the native rendering path on iOS directly. Happy to spike that as a follow-up PR if you want — it would actually close #6043.

<sub>🤖 preview-screenshot watcher (claude-fable-5). Also: screenshots are now embedded in the PR body (Chromium control ✅ / WebKit disproof ❌); PR stays draft since it doesn't fix the target browser.</sub>

gumclaw:
Yes — filing it with JW Player support now (their open-source repo `jwplayer/jwplayer` is archived since Nov 2024, so support.jwplayer.com is the only working channel). The report covers both defects, filed from gumclaw@gumroad.com against Web Player 8.49.5 (cloud-hosted):

1. **`renderCaptionsNatively: false` is silently ignored on Safari/iOS** — the 8.49.5 html5 provider hardcodes `!!OS.iOS || !!Browser.safari || renderCaptionsNatively && Browser.chrome`, so the option can only opt Chrome *in*, never opt Safari/iOS *out*, contradicting the setup-options docs. Asked them to either honor it or fix the docs.
2. **Side-loaded `.srt` cue geometry on the native path** — cues the player injects into the native `TextTrack` carry no `align`/`position`/`line` settings, and iOS WebKit draws them at the right edge. Asked for explicit centered-bottom cue settings.

Their support portal is rate-limiting anonymous submissions at the moment; the request is queued with retries and I'll note the ticket number here once it's confirmed.

Meanwhile the actionable fix on our side stays what the preview QA pointed at: normalize seller `.srt` → `.vtt` with explicit `align:center position:50%` cue settings server-side, which fixes iOS regardless of what JW does. I'll rework this PR in that direction.

gumclaw:
Closing in favor of #6061 — hosted-preview QA proved JW Player 8.49.5 force-enables native caption rendering on iOS/Safari, so this client-side opt-out can't fix #6043 there. #6061 stacks on this branch (the renderCaptionsNatively:false Chrome polish survives) and fixes the root cause by serving WebVTT with explicit centered cue settings.

PR #6054, opened 2026-07-20
Reject custom domains with empty labels (consecutive dots) — kills GUMROAD-VW residual

gumclaw:
Follow-up to #5997 / Sentry [GUMROAD-VW](https://gumroad-to.sentry.io/issues/7558536077/) (`GenerateSslCertificate exhausted retries`).

## What

Adds a malformed-label check to `CustomDomain#validate_domain_format`: labels that are empty (consecutive dots, e.g. `example..com`) or start/end with a hyphen (`-example.com`, `example-.com`). Such names pass the existing regex and `PublicSuffix.valid?`, but Let's Encrypt rejects them outright with `Acme::Client::Error::RejectedIdentifier` — so a record carrying one can never get a certificate.

The validation is skipped for soft-deleted records (`unless: :deleted?`) so `mark_deleted!` keeps working on legacy rows that fail the new rule, and `ReissueSslCertificateForUpdatedCustomDomains` skips invalid records so one bad row can't abort the whole `retry: 0` sweep.

## Why

After #5997 shipped (~12:30 UTC today), GUMROAD-VW dropped from a steady ~50 events/hour to exactly 2/hour:

```
11:00 UTC 48
12:00 UTC 52   <- deploy ~12:30 UTC
13:00 UTC 2
14:00 UTC 2
15:00 UTC 2
16:00 UTC 2
```

The rate-limit fix worked. The residual 2/hour is fully attributed: every remaining event is `Acme::Client::Error::RejectedIdentifier` for two live records — `custom_domain` ids 75854 (`domains.gumroad..com`) and 75855 (`domains.gumroad...com`). Both are `verified` with `ssl_certificate_issued_at: nil`, so the hourly `RenewCustomDomainSslCertificates` cron re-enqueues them every cycle, the ACME order fails, Sidekiq retries exhaust, and the alert fires — an infinite hourly retry loop for names that can never be issued.

Prod sweeps found 25 alive records with empty labels (23 `unverified`, only the two above in the firing path) and 8 with hyphen-edge labels (all `unverified`) — the hyphen shape would reproduce the same incident class the moment one gets verified.

## How it stops the loop

`SslCertificates::Generate#can_order_certificates?` already returns `[false, "Invalid domain"]` when `custom_domain.valid?` is false (before any ACME order or cache write), and `CustomDomainVerificationWorker` already skips invalid records. With this validation, the two stuck records fail `valid?` and the renewal loop logs "Invalid domain" instead of ordering doomed certificates — no data migration needed, and creators can't save new malformed domains (they get the standard "is not a valid domain name" error).

What still alerts (on purpose): genuine per-domain ACME failures for validly-formatted domains — the exhausted-retries alert (#488) keeps its job.

## QA steps

1. In a console, `build(:custom_domain, domain: "example..com").valid?` → `false` with "example..com is not a valid domain name."; same for `-example.com` / `example-.com`; `"test.example.com"` remains valid.
2. For an existing empty-label record (e.g. id 75854), run `SslCertificates::Generate.new(cd).process` in a supported environment → logs "Invalid domain", no ACME order attempted, no Sentry event.
3. Soft-delete an invalid legacy record (`cd.mark_deleted!`) → succeeds without raising.
4. After deploy, confirm GUMROAD-VW hourly event count drops from 2/hour to 0 (verification is deploy-gated; I'll re-check the rate before resolving the Sentry issue).

No preview URL needed — model-validation + background-job change with no UI surface.

## Test results

59 examples, 0 failures (`spec/models/custom_domain_spec.rb`, `spec/services/ssl_certificates/generate_spec.rb`, `spec/sidekiq/reissue_ssl_certificate_for_updated_custom_domains_spec.rb`), see qa-media evidence: [spec run](https://github.com/antiwork/gumroad/blob/fix/custom-domain-empty-label-validation/qa-media/pr-6054-custom-domain-empty-label-specs.png)

- New invalid cases (`example..com`, `sub.example...com`, `-example.com`, `example-.com`, `sub.-example.com`) in the format-validation spec — red/green verified (they fail on origin/main's model).
- New `can_order_certificates?` spec proving a persisted empty-label record is skipped with "Invalid domain".
- New spec proving `mark_deleted!` succeeds on a persisted invalid record.
- New spec proving the reissue sweep skips invalid records without raising — red/green verified.

## AI disclosure

This PR was written by Gumclaw (Hermes agent) as an autonomous Sentry-driven fix; specs run locally and evidence attached.


gumclaw:
## Fable-5 adversarial pre-merge review — CHANGES-REQUIRED → fixed at `c321c7a`

Initial verdict was **CHANGES-REQUIRED** (1 P1, 2 P2, 2 P3). All code findings fixed and re-verified.

| # | Finding | Disposition |
|---|---------|-------------|
| P1-1 | Making 25 persisted rows permanently invalid broke every `save!`-based lifecycle path on them: `Deletable#mark_deleted!` raises `RecordInvalid` → 500 on the seller's own domain-removal path (settings/advanced), half-deleted products (`Link#delete!` calls `custom_domain&.mark_deleted!` after the product row is marked, no transaction), silently failing account closure (`User#deactivate!` blanket-rescues to false), GDPR erasure crash, fraud-suspension `after_transition` raise | **Fixed in c321c7a**: `validate :validate_domain_format, unless: :deleted?` — a record being soft-deleted doesn't need a valid domain; regression spec proves `mark_deleted!` succeeds on a persisted `example..com` row |
| P2-1 | `ReissueSslCertificateForUpdatedCustomDomains` (`retry: 0`) could raise mid-`find_each` on `reset_ssl_certificate_issued_at!` (bare `save!`) for any invalid legacy row with a cert timestamp, silently starving every domain after it in iteration order, every run | **Fixed in c321c7a**: `next unless custom_domain.valid?` in the sweep; regression spec red/greened (fails on origin/main's job) |
| P2-2 | Product edit page round-trips the existing domain → `save!` in `LinksController#update_custom_domain` raises for a product with a bad persisted domain; clearing the field also failed via P1 | **Fixed by the P1 fix** (clearing the field → `mark_deleted!` now succeeds); the unrelated-edit `save!` is caught at the existing rescue and rendered as a 422 naming the domain — behavior unchanged from pre-PR for other legacy-invalid shapes |
| P3-1 | Hyphen-edge labels (`-example.com`, `example-.com`) pass regex + PublicSuffix + the original `".."` check but Let's Encrypt rejects them — same incident class waiting to recur; prod has 8 alive rows in that shape (all `unverified`) | **Fixed in c321c7a**: check generalized to per-label `empty || starts/ends with "-"` (`domain.split(".", -1)`) |
| P3-2 | The 25 legacy rows become zombies: verification worker skips them at its `valid?` guard, hourly renew enqueues a no-op job per row forever | **Accepted as-is for this PR** — the no-op job logs "Invalid domain" and exits cleanly (no retries, no alert, no ACME call). A one-time cleanup task soft-deleting the 33 malformed rows is a reasonable follow-up; with the P1 fix `mark_deleted` works on them |

**Confirmed safe by the review (evidence-backed):**
- Kill mechanism fires before any side effect: `valid?` check at `generate.rb:70` precedes the cache write and `order_certificates`; invalid path is log-only → job succeeds → no retries → no Sentry event. `Generate#initialize` does no network I/O.
- Leading-dot domains (`.example.com`) were already rejected by `PublicSuffix.valid?` (verified against the repo's PublicSuffix 5.1.1); trailing dot by the regex.
- New specs are not vacuous: all new invalid cases pass the old regex + PublicSuffix, so the new check is the sole rejector — red/green verified against origin/main.
- `CustomDomainVerificationWorker` and the 6-hourly verification cron don't crash: `return unless custom_domain.valid?` guard, and state-machine events are non-bang.
- No client-side domain regex exists to disagree with (grep of app/javascript; `CustomDomain.tsx` defers to the server verification endpoint, which itself uses `custom_domain.valid?` and now reports these domains as invalid — an improvement).
- `find_by_host` and all read/render paths unaffected (reads only, guarded by `PublicSuffix.valid?`).

**Only live-verifiable:** GUMROAD-VW dropping from 2/hour to 0 after deploy — I'll re-check the Sentry rate post-deploy before resolving the issue.

Rubocop clean on all four touched files; 59 examples, 0 failures locally.

gumclaw:
Deployed to production ✅ — live x-revision is now `ff2dc39` (confirmed via healthcheck header + ancestry check at 17:58 UTC). GUMROAD-VW was firing at a steady ~2 events/hr residual before the deploy (last event 17:15 UTC). A 2-hour verification watcher is armed: if no new events arrive post-deploy, the Sentry issue (7558536077) will be marked resolved.

gumclaw:
Post-deploy verification complete ✅ — GUMROAD-VW has produced zero events since the 17:58 UTC deploy (last event 17:15 UTC, pre-deploy). The Sentry issue (7558536077) is now marked resolved. This fix is confirmed working in production; no further action needed.

PR #6057, opened 2026-07-20
Fix abandoned-cart email job dying daily on MySQL statement timeout (gumroad-private#1198)

gumclaw:
Fixes the platform-wide abandoned-cart email outage tracked in gumroad-private#1198.

## What

`ScheduleAbandonedCartEmailsJob` (daily cron, 14:00 UTC) has failed **every day since 2026-04-02** — zero `SentAbandonedCartEmail` rows in 3.5 months. The per-day scan

```ruby
cart_ids = Cart.abandoned(updated_at: day_start..day_end).pluck(:id)
```

grew past MySQL's `max_execution_time` (`Mysql2::Error: Query execution was interrupted, maximum statement execution time exceeded`), the job exhausted its 5 retries into the Sidekiq dead set, and — the second bug — **nothing alerted**, so it stayed dead silently.

## Why this shape

- **Keyset-batched scan** (`abandoned_cart_ids`): each day's window is walked in id-ordered batches of 10,000, so no single SQL statement's result set scales with platform size. Indexes only shrink the constant factor of a query that still scales with the full cohort — the timeout class survives and recurs as the table grows; bounding each statement's work kills the class. (Same rationale as the payout-batch fix in #5755.) The `carts` table already has `index_carts_on_deleted_at_and_updated_at` + the PK for the cursor, so each bounded statement is cheap. Equivalence: the cursor advances past whole rows (ids unique), so the union of batches is exactly `Cart.abandoned(window).pluck(:id)`.
- **2-hour statement budget** via the sanctioned `WithMaxExecutionTime.timeout_queries` helper as belt-and-suspenders: this is a scheduled batch job, not a user-facing request — a long statement is expected here. Same precedent as `PerformPayoutsUpToDelayDaysAgoWorker` (#5760) and the long-running report jobs.
- **`sidekiq_retries_exhausted` → `ErrorNotifier` (Sentry)**: the outage produced no signal for 3.5 months. Copies the `GenerateSslCertificate` pattern (#488) — dead-set failures of this job now surface in Sentry the same day. Sentry-only (no email) since it's an operational condition, matching repo convention.

No behavior change to matching/eligibility/mailer logic — the workflow-matching section is untouched.

## Test Results

```
$ DISABLE_SPRING=1 OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES ./bin/bundle exec rspec spec/sidekiq/schedule_abandoned_cart_emails_job_spec.rb
Finished in 6.6 seconds (files took 8.98 seconds to load)
7 examples, 0 failures
```

New examples:
- **batch-boundary equivalence** — `SCAN_BATCH_SIZE` stubbed to 1 so every cart id crosses a batch boundary; asserts the exact same 3 emails with the same workflow/product payloads as the unbatched scan
- **retries-exhausted alert** — `sidekiq_retries_exhausted_block` reports to `ErrorNotifier` with exception class/message

`rubocop --force-exclusion` clean on both files.

## Walkthrough video (backend-only change — no UI surface)

Diff, the new scan, and the full green spec run:

https://github.com/antiwork/gumroad/blob/fix/abandoned-cart-scan-statement-timeout/qa-media/pr-abandoned-cart-scan-timeout-walkthrough.mp4?raw=true

No user-facing surface is touched by this PR (Sidekiq job only, mailer/matching logic unchanged), so the spec run + walkthrough are the artifact per CONTRIBUTING's backend-PR convention.

## QA steps

1. After deploy, confirm the 14:00 UTC run completes: `ScheduleAbandonedCartEmailsJob` no longer lands in the Sidekiq dead set.
2. `SentAbandonedCartEmail.where("created_at > ?", Date.current).count` > 0 the next day (first run will process the full 30-day backlog window, so expect a burst).
3. Kill-switch check: nothing to flag — behavior identical to the pre-April job, just bounded statements.

## Post-deploy note

The 30-day scan window means the first successful run will email carts abandoned in the last month — that's the same window the job always used, not a backlog flush of the full 3.5-month gap (carts older than 1 month age out of `Cart.abandoned` by design).

---

## AI disclosure

Prompts/instructions given: standing order (Sahil 2026-07-19) to autonomously build qualifying open issues; gumroad-private#1198 body (verified prod root cause) used as the spec.


gumclaw:
**fable-5 adversarial pre-merge review — verdict: PASS (no P1/P2)**

Hunt list and findings:
- **(a) Can the keyset walk return duplicate or missed ids?** `Cart.abandoned`'s `left_outer_joins(:sent_abandoned_cart_emails).where(sent_abandoned_cart_emails: { id: nil })` yields at most one row per cart (only the NULL-extension row matches), so `pluck(:id)` over the join can't duplicate. The cursor advances past whole rows on the unique PK, so batches partition the result set exactly. The batch-boundary spec (`SCAN_BATCH_SIZE = 1`) proves payload-level equivalence, not just counts.
- **(b) HAVING-before-LIMIT trap (the #5755 rejection)?** Not applicable — no aggregation here, plain row filter + LIMIT; MySQL streams exactly LIMIT rows off `index_carts_on_deleted_at_and_updated_at` + PK.
- **(c) Ambiguous `id` in pluck with the join?** Rails qualifies `pluck(:id)` as `carts.id` (and the original code used the identical pluck on the identical scope); the new cursor predicates are explicitly `carts.id`-qualified anyway.
- **(d) Does the window's day-boundary math change?** No — `day_start`/`day_end` computation untouched; only the fetch strategy inside the window changed.
- **(e) Re-run safety?** `lock: :until_executed` unchanged; the mailer dedupe (`sent_abandoned_cart_emails` filter inside the scope) makes a re-run after partial success skip already-emailed carts by construction.
- **(f) Duration into `timeout_queries(seconds:)`?** `(2.hours * 1000).to_i` is valid on ActiveSupport::Duration — same call shape as the payout workers (#5760).
- **(g) Alert channel fit?** Sentry-only via `ErrorNotifier`, mirroring `GenerateSslCertificate` (#488); no email spam channel added.

Minor (won't block): the loop could break one query earlier on `batch.size < SCAN_BATCH_SIZE`; left as-is for simplicity — the extra statement is a cheap empty-result probe.

PR #6061, opened 2026-07-21
Serve .srt captions as WebVTT with centered cue settings (fixes iOS Safari right-edge captions)

gumclaw:
## Summary

Fixes #6043 — external `.srt` captions render pinned to the right edge of the video in iOS Safari.

**Why the client-side fix (#6052) can't work on iOS:** hosted-preview QA disproved `renderCaptionsNatively: false`. JW Player 8.49.5's html5 provider hardcodes native caption rendering on iOS/Safari (`renderNatively = !!OS.iOS || !!Browser.safari || ...`) — there is no way to opt out, so on iOS the cues always go through WebKit's native TextTrack renderer. WebKit misplaces cues that carry **no explicit position settings** at the right edge. SRT has no syntax for positioning; WebVTT does. A bug report to JW Player is filed but can't be waited on.

**The fix (server-side, client-independent):** serve captions as WebVTT with explicit `align:center position:50%` settings on every cue, converted from the stored seller file at serve time.

- **`SubtitleToVttService`** — defensive SRT→VTT converter: strips BOM, normalizes CRLF/CR, converts `,`→`.` timestamps (padding sloppy ones), drops SRT numeric counters, appends `align:center position:50%` to every cue, skips malformed cues instead of failing the file. Already-VTT files pass through, but bare cues (no settings) get the same centered settings injected — they're misplaced by iOS WebKit exactly like converted SRT cues; cues that already carry settings are left untouched (header/NOTE blocks and cue identifiers preserved).
- **New endpoint `GET /r/:id/:product_file_id/:subtitle_file_id/captions.vtt`** (`UrlRedirectsController#subtitle_file_vtt`, + custom-domain route) — runs the same `check_permissions` gate as the other url_redirect media endpoints, fetches the stored file from S3, converts, and serves `text/vtt`. Output is cached (`Rails.cache`, keyed on the record's `cache_key_with_version`, 1 day) so buyers of the same video share one fetch+conversion. Missing S3 object → 404 (player shows no captions instead of erroring).
- **Players now load captions from that endpoint** instead of a signed S3 URL to the raw file: the stream page playlist (`UrlRedirect#video_files_playlist`, keeping the bulk-preload of `alive_subtitle_files`) and the download-page inline player (presenter's new `vtt_url` + `FileList.tsx`). JW Player infers VTT from the `.vtt` path; label/kind unchanged. `signed_url`/`download_url` remain for downloading the original file. The now-unused `ProductFile#subtitle_files_urls` is removed.
- **Keeps #6052's `renderCaptionsNatively: false`** (this branch stacks on it) — it's still the right polish for Chrome/desktop so JW's own renderer draws captions consistently there.

**Why serve-time conversion (vs upload-time):** subtitle URLs were direct-to-S3 signed URLs; converting at serve time through an authenticated endpoint covers every already-uploaded seller file with zero backfill, keeps the stored original untouched for downloads, and the cache keeps per-request cost negligible. Upload-time conversion would have required a backfill job over all existing `SubtitleFile` rows plus reconciling the "download the original" path.

**Relationship to #6052:** this PR stacks on and supersedes #6052 as the fix for #6043 (its commit is included here). #6052 can be closed in favor of this PR.

## QA steps

Needs the hosted preview (`preview` label is set; a commit has been pushed after labeling to arm the deploy).

1. On the preview app, buy (or use a test purchase of) a product with a streamable video that has an external `.srt` subtitle file attached.
2. Open the receipt → watch page. In desktop Chrome/Safari, enable captions: they render centered at the bottom (unchanged).
3. `curl` the new endpoint from the watch session: `https://<preview-host>/r/<token>/<product_file_external_id>/<subtitle_file_external_id>/captions.vtt` → returns `text/vtt` starting with `WEBVTT`, every cue timing line ending in `align:center position:50%`.
4. **On-device iOS Safari check (the bug's surface):** open the same watch page on an iPhone/iPad, play the video, enable the caption track → captions render centered at the bottom instead of pinned to the right edge.
5. Regression: download the subtitle file from the download page → still delivers the original `.srt`.

## Test Results

New unit specs (red-first against fixtures incl. CRLF, BOM, HTML entities, overlapping/malformed cues, empty file, VTT passthrough):

```
$ ./bin/bundle exec rspec spec/services/subtitle_to_vtt_service_spec.rb
13 examples, 0 failures

$ ./bin/bundle exec rspec spec/controllers/url_redirects_controller_spec.rb -e subtitle_file_vtt -e download_subtitle_file
5 examples, 0 failures

$ ./bin/bundle exec rspec spec/models/url_redirect_spec.rb -e video_files_playlist spec/presenters/url_redirect_presenter_spec.rb
7 examples, 0 failures (playlist) — presenter/product_file failures in the full files are pre-existing local MinIO/S3 env failures, verified identical on the base commit via targeted checkout
```

`npx tsc --noEmit`: no new type errors (only the pre-existing `vite/client` env error of a symlinked-node_modules worktree).

**Before** (raw `.srt` served; iOS pins cues to the right edge):

before

**After** (serve-time conversion output from the new endpoint):

after

**Spec runs:**

tests

Captured on the hosted preview (`https://fix-srt-to-vtt-captions.apps.staging.gumroad.org`), Playwright @2x, 375px viewport. Seeded: streamable mp4 on the `demo` product with an external `.srt` subtitle (cue text is self-describing), validated purchase + url_redirect.

Live-player probe results (both engines confirmed loading captions from the new `/r/.../captions.vtt` endpoint — network log captured the request):

- **WebKit (Safari engine — the bug's surface):** `renderCaptionsNatively:true` (JW forces native rendering on Safari/iOS), native TextTrack showing, and the parsed VTT cue reads back `position:50 align:center` — WebKit renders it **centered at the bottom** instead of pinned to the right edge. This is exactly the path the client-side #6052 fix couldn't reach.
- **Chromium (control):** `renderCaptionsNatively:false`, JW's own renderer draws the cue centered (unchanged behavior).

**Stream page `/s/<token>` — WebKit (native renderer, cue centered via the injected settings):**

webkit stream page captions centered

**Stream page — Chromium control (JW renderer, centered):**

chromium stream page captions centered

**Download page inline player (`FileList.tsx` `vtt_url` path) — captions load from the new endpoint and render centered:**

download page inline player captions

**Endpoint check on the preview** (QA step 3): `GET /r/<token>/<file_id>/<subtitle_id>/captions.vtt` → `200`, `Content-Type: text/vtt; charset=utf-8`, body:

```
WEBVTT

00:00:00.500 --> 00:00:29.500 align:center position:50%
Captions should render centered at the bottom of the video.
```

## AI disclosure

🤖 Generated with claude-fable-5 via Hermes agent.


gumclaw:
Fable review: **PASS** — no P1/P2.

Verified in the diff:
- New endpoint sits behind the same `check_permissions` gate as `download_subtitle_file` (registered in all three permission lists) — no new unauthenticated S3 proxy surface.
- Conversion is defensive (BOM/CRLF/malformed cues skipped, S3 errors → 404, never 500 on a seller's weird file) and output is `text/vtt` plain rendering — no HTML interpretation.
- Cache key on `cache_key_with_version` is correct given S3 objects are never overwritten in place (re-uploads = new records).
- VTT passthrough only injects settings into bare cues — sellers who hand-authored positioned VTT keep their positioning.
- Specs 13/13 service + 5/5 controller + 7/7 playlist; remaining presenter failures verified pre-existing on base.

Closing #6052 in favor of this (its Chrome-side commit rides along on this branch). On-device iOS Safari QA runs against the hosted preview once Buildkite finishes — screenshot watcher is armed.

slavingia:
Screenshots look good

gumclaw:
Thanks — auto-merge is armed, so this lands once CI goes green. I'll pick up any Greptile findings if they arrive before then.

PR #6072, opened 2026-07-21
Debounce the 'payouts paused' email so Stripe verification blips never alarm sellers

gumclaw:
## Summary

Fixes antiwork/gumroad-private#1211.

A new seller got the "your payouts are paused" email over the weekend and wrote in confused — Stripe had auto-paused payouts for a routine verification sweep and auto-resumed them **one minute later**. The internal pause/resume state machine handled the blip perfectly; only the notification was trigger-happy, because `MerchantRegistrationMailer.stripe_payouts_disabled/.stripe_payouts_under_review` was enqueued via `deliver_later` the moment the `payouts_enabled: false` webhook landed.

**Design: debounce the email, never the pause.** The pause stays instant (money safety); only the notification waits out the blip window:

- New `StripePayoutsPausedEmailJob`, scheduled with `perform_in(30.minutes, …)` from the webhook handler instead of enqueueing the mailer directly. Both email types (`action_required` → `stripe_payouts_disabled`, `under_review` → `stripe_payouts_under_review`) route through it.
- When the job runs, it re-checks the account **under the same `user.with_lock` the webhook handler uses**: if payouts are no longer paused by Stripe (`payouts_paused_by_source != STRIPE` or `!payouts_paused_internally?`), it sends nothing — the blip resolved before the seller ever needed to know. If still paused, it sends exactly what we send today.
- **Marker semantics preserved and extended.** The existing `stripe_payouts_pause_email_sent` dedupe marker is still claimed inside the lock at pause time, so near-simultaneous webhooks can't schedule duplicate emails. Each claim now also rotates a `stripe_payouts_pause_email_claim_token` (new JSON accessor on `MerchantAccount`); the scheduled job carries the token it was created with and only sends if it still matches. This makes a pause→resume→pause sequence send exactly one email (for the sustained pause): the resume clears both markers, the second pause claims a fresh token, and the stale first job no-ops.
- **The token rotates whenever the desired notification type changes, not just per pause episode.** Escalation (`under_review` → `action_required`) and de-escalation (`action_required` → `under_review`, i.e. the seller satisfied the outstanding requirements but payouts stay paused pending review) each claim the new email type and rotate the token, so a job queued for the old type can never send a now-wrong email. A transition to a never-email state (`rejected.*`, `platform_paused`) drops the claim entirely, invalidating any pending job that would otherwise contradict the account's state.
- When the job finds the pause resolved via a **non-Stripe** resume (admin / payout-method update — a path that doesn't clear the marker), it releases the claim itself, leaving state as if no email was ever claimed, so a later sustained pause still emails.
- Blips longer than 30 minutes still notify (correct — a half hour of paused payouts is worth an email even if it later resolves). Genuine action-required pauses are delayed by 30 minutes, immaterial next to Stripe review timelines.

**No UI changes** — this is a backend-only change to webhook handling and email scheduling. No user-facing surface is touched; the email copy itself is unchanged.

## QA steps

No clickable surface; verify via diff + console:

1. Read the diff: `app/sidekiq/stripe_payouts_paused_email_job.rb` (new job), `stripe_merchant_account_manager.rb` (claim-token rotation + `perform_in` dispatch + resume clearing both markers), `merchant_account.rb` (new `stripe_payouts_pause_email_claim_token` accessor).
2. Run the touched spec group: `bundle exec rspec spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_spec.rb -e "in the request payouts are disabled"` → 18 examples, 0 failures.
3. Console (preview app): pick a test user with a Stripe merchant account, simulate a pause via `StripeMerchantAccountManager.handle_stripe_event(<payouts_enabled:false event>)`, confirm `user.payouts_paused_internally?` is true immediately, no mail in `ActionMailer` deliveries, and a `StripePayoutsPausedEmailJob` in `Sidekiq::ScheduledSet` ~30 min out. Then simulate the resume event and run the scheduled job inline (`StripePayoutsPausedEmailJob.new.perform(*args)`) → no email, and `merchant_account.stripe_payouts_pause_email_sent` is nil (ready for a future real pause).

## Test Results

```
$ bundle exec rspec spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_spec.rb -e "payouts_enabled status"
27 examples, 0 failures

  (includes the 5 new debounce examples:
   - does not email when the pause resolves before the debounced job runs (Stripe verification blip)
   - does not email when the job runs after a non-Stripe resume, and releases the claim so a later pause can email
   - sends exactly one email for the second pause in a pause→resume→pause sequence
   - sends the under-review email instead when requirements are satisfied but payouts stay paused before the action-required email sends
   - invalidates a pending pause email when the account transitions to a rejected state before the debounce elapses)

$ bundle exec rubocop app/sidekiq/stripe_payouts_paused_email_job.rb app/models/merchant_account.rb \
    app/business/payments/merchant_registration/implementations/stripe/stripe_merchant_account_manager.rb \
    spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_spec.rb
4 files inspected, no offenses detected
```

Spec-run evidence (no UI change — backend-only; this terminal render is the before/after artifact):

spec run

## AI disclosure

🤖 Generated with claude-fable-5 via Hermes agent.


gumclaw:
Fable review: **PASS** — no P1/P2.

The claim-token design is the right shape and the dangerous sequences are all spec-pinned: blip → nothing; pause→resume→pause → exactly one email (fresh token supersedes the stale job); under-review→action-required escalation supersedes the pending job; near-simultaneous webhooks still race-safe (claim still taken inside the user lock, unchanged). The subtle one — a non-Stripe resume that doesn't clear the marker — is handled by the job releasing its own claim. 25 examples green, rubocop clean.

Behavior change to sign off on: genuine action-required pause emails now arrive 30 minutes later than before (immaterial vs Stripe review timelines, and the pause itself + dashboard state remain instant).

nyomanjyotisa (commented):
The claim token tracks the pause episode, but not whether the queued email type is still valid.

Example:

1. `requirements.past_due` schedules an `action_required` email.
2. Before 30 minutes, the seller satisfies the requirement, but payouts remain disabled pending review.
3. The next webhook computes `under_review`, but the existing `action_required` marker leaves the original token valid.
4. The old job passes its checks and sends an action-required email even though no action is required.

The stale job could also survive a transition to `rejected.*` or `platform_paused`, producing contradictory notifications. Please invalidate or rotate the token whenever the desired notification type changes—or validate the current type when the job runs—and add coverage for `action_required → under_review` while payouts remain disabled.

gumclaw:
Good catch — the token tracked the pause episode but not the notification type, so a de-escalation (`action_required` → `under_review` while payouts stayed paused) left the original action-required job's token valid, and a transition to `rejected.*` / `platform_paused` could let a queued email contradict the account's state.

Fixed in b4eccc3:

- **De-escalation now claims the under-review email.** `claim_stripe_payouts_pause_email` no longer treats an existing `action_required` marker as terminal for `under_review`: when the requirements are satisfied but payouts stay paused pending review, the under-review email is claimed and the token rotates — invalidating the pending action-required job, so the seller is never told to act when nothing is due. (The seller then gets the correct under-review email after its own debounce window.)
- **Never-email transitions drop the claim entirely.** When the computed type is nil (`rejected.*`, `platform_paused`), the marker and token are cleared inside the same lock, so any job still waiting out its debounce window no-ops instead of producing a contradictory notification. Clearing also lets a later actionable phase (e.g. an appeal reopening verification) email afresh.
- Coverage added for both: `action_required → under_review` while payouts remain disabled (asserts only the under-review email sends), and `requirements.past_due → rejected.listed` mid-window (asserts nothing sends and the claim is dropped).

`payouts_enabled status` group: 27 examples, 0 failures locally; rubocop clean. PR body and qa-media spec-run render updated to match.

gumclaw:
## Fable review — round 2 (post-review-fix, `b4eccc39`)

**Verdict: READY-TO-MERGE** — no P1/P2. Round-1 PASS re-verified at the new head; the fix for @nyomanjyotisa's review comment is correct and pinned by specs.

### What round 2 hunted (the surface `b4eccc39` changed)

| Probe | Result |
|---|---|
| Can rotating the claim on de-escalation suppress a legitimate email? | **No.** Every successful claim both rotates the token AND schedules a fresh job of the correct type — the stale job is invalidated but its replacement always exists. Verified for de-escalation (action_required → under_review) and re-escalation (under_review → action_required falls through the `already_claimed == "action_required"` guard). |
| Does the drop-claim branch (rejected.* / platform_paused) leave a job that can still send? | **No.** Clearing the token makes the pending job's `claim_token == carried_token` check fail (nil ≠ uuid) → skip. `stripe_payouts_pause_email_type` returns nil for exactly `rejected.*` and `platform_paused`, matching the branch's reachability. |
| Can the job's clear-on-resolve branch clobber a NEWER pause episode's claim? | **No.** The clear only runs after the token match succeeds (the job owns the claim), inside the same `user.with_lock` the webhook handler uses. |
| Token-read-after-lock race: enqueue reads the token after the lock commits — can a resume webhook clear it in between? | Benign. The job then sees stored nil ≠ carried uuid → skips, which is the correct outcome for a resumed account. |
| Rolling-deploy arity | Safe — the job class is new in this PR; no old-arity scheduled jobs exist. `retry: 5` + token check = idempotent under Sidekiq retries. |
| Specs pin the fix (not vacuous)? | **Red/green verified**: reverse-applied the `b4eccc39` code hunk → the de-escalation spec and the rejected-state invalidation spec both FAIL; restored → pass. |
| Full spec file at head | `stripe_merchant_account_manager_spec.rb`: **335 examples, 0 failures** (local, seeded test DB). |

Note for other local runners: 3 initial failures (`Validation failed: Merchant account can't be blank` in the balance factory) were shared-test-DB seed wipe, not the PR — `RAILS_ENV=test rails db:seed` fixed them.

Arming auto-merge (squash) — CI is already green at this head.

PR #6082, opened 2026-07-21
Return an empty page instead of 500ing on out-of-range wishlist product pages

gumclaw:
## What

Public wishlist product listings (`Wishlists::ProductsController#index` → `WishlistPresenter#paginated_public_items`) now return an empty page instead of raising `Pagy::OverflowError` when the requested `?page=` exceeds the total page count.

## Why

Sentry [GUMROAD-2A](https://gumroad-to.sentry.io/issues/7369073349/) — `Pagy::OverflowError: expected :page in 1..1; got 2` — has fired 1,614 times for 566 users since March 2026 and is still firing daily. The endpoint is public and unauthenticated: the page number comes straight from the query string, so out-of-range pages arrive constantly (stale bookmarks, crawlers, wishlists whose products were removed between page loads). Each one 500ed.

The repo's global pagy default is `overflow: :exception` (`config/initializers/pagy.rb`), and other public paginated presenters already opt out per-call (`paginated_installments_presenter.rb` uses `overflow: :empty_page`, `affiliated_products_presenter.rb`/`paginated_utm_links_presenter.rb` use `:last_page`). This applies the same pattern here. `:empty_page` (not `:last_page`) so the client's "load more" pagination terminates cleanly with `next: nil` instead of re-serving the last page's items.

A prior attempt (#4116) rescued the error in the presenter and fell back to page 1; it was closed for inactivity. Pagy's built-in overflow mode is simpler and matches existing convention.

## QA steps

1. Create a wishlist with fewer than 20 products.
2. Request `https://<seller>.gumroad.com/wishlists/<id>/products?page=2`.
3. Before: 500. After: `200` with `{"items": [], "pagination": {"page": 2, "pages": 1, "next": null, ...}}`.

## Test Results

`bundle exec rspec spec/presenters/wishlist_presenter_spec.rb` — **28 examples, 0 failures**. The new spec ("returns an empty page instead of raising when the requested page exceeds the total pages") was proven red against unmodified main before the fix.

spec results

## Before/After

Not a UI change — valid in-range pages render byte-identical data (all 27 pre-existing presenter specs pass unmodified). Only the out-of-range error path changes from a 500 to an empty JSON page.

---

**AI disclosure:** Generated with Claude Fable 5 (Gumclaw's daily Sentry sweep cron). Instructions: triage top unresolved Sentry issues in the gumroad project, classify, and fix actionable items with red-first tests per the repo's contribution gates.


gumclaw:
Self-review (fable-5 adversarial pass on the diff):

- **Scope of the change:** `paginated_public_items` is called only from `public_items`, whose only caller is `Wishlists::ProductsController#index` (grep-verified). No other consumer's overflow behavior changes.
- **Why `:empty_page` and not `:last_page`:** the frontend consumes `pagination.next` for load-more; `:empty_page` yields `next: nil` and an empty `items` array, which terminates pagination cleanly. `:last_page` would re-serve duplicate items to a client that explicitly asked for a later page.
- **Pagy extras:** `pagy/extras/overflow` is already required globally in `config/initializers/pagy.rb` (the repo-wide default is `overflow: :exception`), so the per-call override needs no new require.
- **Red→green proof:** the new spec fails on unmodified main with `Pagy::OverflowError: expected :page in 1..1; got 2` (the exact production error) and passes with the fix. All 27 pre-existing specs pass unmodified, confirming in-range pages are byte-identical.
- No P1/P2 findings.

PR #6094, opened 2026-07-21
Fix red main: upload the pages Tailwind build under the authorized assets/ S3 prefix

gumclaw:
## What

Moves the fingerprinted custom-page Tailwind build from `public/pages/` → `public/assets/pages/`, so the deploy syncs it to the `assets/pages/` prefix of the assets bucket instead of a new top-level `pages/` prefix.

## Why — main is red / production deploys are blocked

#6085 added `aws s3 sync ... s3://${ASSETS_S3_BUCKET}/pages` to `docker/web/push_assets_to_s3.sh`. The `_buildkite` deploy credentials are scoped to specific existing prefixes only (`assets/*`, `vite/*`, `js/*`, `images/*`, `fonts/*`, `packs/*` — IAM inline policy `S3@AssetSync`), so every production Compile Assets step since has failed:

```
upload failed: ../app/public/pages/pages-tailwind-b4bec5d01406.css to s3://gumroad-production-assets/pages/pages-tailwind-b4bec5d01406.css
An error occurred (AccessDenied) when calling the PutObject operation: User: arn:aws:iam::526234316351:user/_buildkite is not authorized to perform: s3:PutObject
```

Buildkite builds #14957, #14958, #14975, #14984 all failed here; prod is parked on `8bbc9cf7` (pre-#6085). Widening the IAM policy was the alternative, but I can't apply IAM writes (gumclaw has no `iam:PutUserPolicy`), and reusing the already-authorized `assets/` prefix needs no infra change and keeps the credentials' blast radius as-is.

## Notes

- The Ruby manifest validation regex, the build script, the manifest contents, the spec, and the preview asset cache tar list all move to the new path together.
- Stale preview asset cache entries self-heal: the cache key includes `scripts/build_pages_tailwind.mjs`, so this change busts it; even a stale tarball's old `pages/...` manifest entry fails the new regex and falls back to the inline stylesheet — never a broken `<link>`.
- Prod never served the fingerprinted `<link>` (no deploy of #6085 succeeded), so there are no live pages referencing `pages/...` URLs.
- Removed the `.gitignore` entry rather than moving it — `public/assets/` is already ignored.

## Before/after

before/after evidence

## QA steps

This is a deploy-pipeline fix — the user-visible behavior (custom-page Tailwind `<link>` vs inline fallback) is unchanged from #6085. To QA:

1. On the next main deploy after merge, watch the production Compile Assets step in Buildkite: the sync line should read `Uploading public/assets/pages to S3` and succeed (previously `AccessDenied`).
2. After deploy, `curl -sI https://assets.gumroad.com/assets/pages/pages-tailwind-<hash>.css` (hash from the manifest in the built image) → 200.
3. Open any published custom HTML page; its head should contain `<link rel="stylesheet" href="https://assets.gumroad.com/assets/pages/pages-tailwind-…css">` and the page should be styled.

## Test results

- `npm run build:pages-tailwind` → writes `public/assets/pages/pages-tailwind-b4bec5d01406.css` + manifest `{"pages-tailwind.css":"assets/pages/pages-tailwind-b4bec5d01406.css"}` ✔️
- `bundle exec rspec spec/controllers/links_controller_custom_html_spec.rb -e pages_tailwind_head` → 5 examples, 0 failures ✔️
- `bash -n` on both shell scripts, `ruby -c`, `node --check` ✔️
- IAM policy verified directly: `S3@AssetSync` allows `s3:PutObject`/`PutObjectAcl` on `gumroad-production-assets/assets/*` and `gumroad-staging-assets/assets/*` ✔️

---

## AI disclosure

This PR was implemented with AI assistance using Claude Fable 5.

Prompts used:

- "main is red, investigate and fix"


gumclaw:
**Fable-5 adversarial pre-merge review: READY-TO-MERGE** (0 P1, 0 P2, 2 P3 — both fixed on the branch)

| Sev | Finding | Disposition |
|-----|---------|-------------|
| P3 | `Makefile` build_test cache-restore branch (`rm -rf public/assets`) left `public/pages-tailwind-manifest.json` dangling in the committed test image → silent divergence from prod (safe inline fallback, but implicit) | fixed — the manifest is now removed in the same branch, so test explicitly takes the inline path |
| P3 | qa-media filename missed the `pr-<number>-...` convention | fixed — renamed, PR body URL updated, raw URL verified 200 |

**Confirmed safe (verified against code, not the PR description):**
- Stale preview-cache tarballs can never restore against new code: `build_pages_tailwind.mjs`, `push_assets_to_s3.sh`, and `preview_asset_cache.sh` are all cache-key inputs — self-invalidating, no version bump needed. Even a hypothetically restored old manifest (`pages/...`) fails the new regex (executed in Ruby, returns false) → safe inline fallback.
- Repo-wide sweep: zero remaining `public/pages/` / `pages/pages-tailwind` references.
- No collision on `public/assets/`: no sprockets clobber exists; Vite writes to `public/vite`; `compile_assets.sh` only removes `tmp/cache/assets`; the S3 push runs after the commit from the same image.
- Regex executed against the exact build output — matches; rejects the old path and `../` traversal (`\h+` fully anchored) with the on-disk `File.exist?` check behind it.
- CSP `style-src` entry is a path-less host source — the deeper `/assets/pages/` path passes unchanged.
- No rolling-deploy skew: manifest + code + regex ship in one image; prod never served the old `<link>` (no #6085 deploy succeeded).
- `git check-ignore` confirms `public/assets/` covers the new dir.

**Only live-verifiable:** the actual production S3 sync succeeding under `_buildkite` — the IAM policy was read directly (`S3@AssetSync` allows `PutObject`/`PutObjectAcl` on `gumroad-production-assets/assets/*`), but the first post-merge Buildkite build is the real proof. Watching it after merge.

gumclaw:
Deployed and verified live ✅

- Prod `x-revision` is now `b4435f0c4d48` (includes this fix, merged as `92b1e922`) — first successful production deploy since the AccessDenied breakage started.
- Buildkite Build #14993 (this merge) passed the previously-failing Compile Assets step, proving the `assets/pages/` prefix fix against real infra; #14998 carried it to prod.
- The fingerprinted stylesheet is live: `https://assets.gumroad.com/assets/pages/pages-tailwind-b4bec5d01406.css` → HTTP 200.
- Post-rollover smoke: brief 502s during pod recycle cleared; healthcheck, homepage, and /discover all steady 200s.

Note: #14993's GHA run also hit the unrelated duplicate-migration-version red (from #6027), fixed separately in #6095 — both fixes shipped together in this deploy.

PR #6098, opened 2026-07-21
Fix Discounts page crash on reload when the search query contains %

gumclaw:
## What

Removes a redundant `decodeURIComponent` in `extractParams` on the Checkout → Discounts page (`app/javascript/components/CheckoutDashboard/DiscountsPage.tsx`). `URLSearchParams.get()` already percent-decodes the value, so the extra decode ran on the already-decoded string; a literal `%` (from searching e.g. `100%`) is not a valid escape sequence, so `decodeURIComponent` threw `URIError: URI malformed` during initial render and the page crashed on reload or when opening the copied URL. Also exports `extractParams` and adds unit tests covering the `%` regression, absent query, sort, and page parsing.

## Why

Reported via support (gumroad-private#1232, reporter Ankush, 2026-07-21): searching Discounts for `100%` and reloading crashes the page. The Library page's equivalent `extractParams` already reads the param without a second decode — this brings Discounts in line.

Fixes antiwork/gumroad-private#1232

## Before / After

| Steps | Before | After |
|---|---|---|
| Search `100%` on Discounts, reload (`?query=100%25`) | `URIError: URI malformed`, page fails to render | Page loads with the `100%` search restored |
| Normal query, reload | works | works (identical — `get()` output used directly) |

Captured on https://fix-discounts-double-decode.apps.staging.gumroad.org (x-revision `58964c7f`) by loading the crash-repro URL `/checkout/discounts?query=100%25` directly. Zero page errors (`page.on("pageerror")` empty), the `100%` search is restored in the input, and the list is filtered to the matching `LAUNCH100` discount.

| Desktop (1280px) | Mobile (375px) |
|---|---|
| desktop | mobile |

## QA steps

Preview app: https://fix-discounts-double-decode.apps.staging.gumroad.org (login `seller@gumroad.com` / `password`, 2FA `000000`)

1. Sign in as the seller and open Checkout → Discounts (`/checkout/discounts`).
2. Search `100%` — the URL becomes `?query=100%25`.
3. Reload the page (or open `/checkout/discounts?query=100%25` directly): it loads with the search restored instead of erroring.
4. Confirm a plain search (e.g. `summer`) still restores correctly on reload.

## Test Results

```
✓ app/javascript/components/CheckoutDashboard/DiscountsPage.test.ts (4 tests) 3ms

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

(vitest run, local, at HEAD of this branch)

---

## AI disclosure

Written by Claude Fable 5 (Hermes agent "gumclaw") running the autonomous dev dispatcher. Instructions: implement the fix described in gumroad-private#1232 (drop the double decode), add regression tests, verify with vitest, and open a PR per CONTRIBUTING.md. Preview screenshots captured and attached by the same agent from the hosted preview app.


gumclaw:
Self-review (fable-style):

- **Diff scope**: DiscountsPage.tsx (drop double decode + export `extractParams` for tests) and a new test file. No other page shares this helper; Library's `extractParams` was already correct.
- **Behavioral delta hunt**: for every query NOT containing `%`, `decodeURIComponent(alreadyDecoded)` was an identity op, so removing it changes nothing. Queries that legitimately contain encoded sequences (e.g. searching for the literal text `%20`) were previously being WRONGLY re-decoded to a space — the fix also corrects that lossy case.
- **Export risk**: `extractParams` is now exported; it's a pure function with no side effects, so exposure carries no risk. Named export doesn't disturb the existing default export.
- **URL-writing side**: the page writes the query with `URLSearchParams.set`, which encodes correctly — round-trip is now symmetric.

Verdict: PASS, no P1/P2.

PR #6102, opened 2026-07-21
Reject non-digit characters in ISBN validator

gumclaw:
## What

Fixes the ISBN validator (`app/validators/isbn_validator.rb`) so it rejects values containing non-digit characters. Both checksum paths now require the ISBN character grammar before running the math: ISBN-10 must be 9 digits plus a `0-9`/`X`/`x` check character, and ISBN-13 must be exactly 13 digits. Also accepts a lowercase `x` check character, which is a legal ISBN-10 representation the old code rejected.

## Why

`String#to_i` coerces letters and punctuation to `0`. A 10-character all-letter string like `helloworld` therefore summed to 0, and `0 % 11 == 0` matched its coerced check digit — so the validator accepted it. Same for 13-character garbage (`AAAAAAAAAAAAA`): a sum of 0 is a multiple of 10. Garbage ISBNs were saved on ebook `ProductFile`s and emitted into the product page's schema.org Book structured data.

Refs antiwork/gumroad-private#1229.

## Before / After

Not a UI change — validation logic only. No adjacent UI surface changes: the product editor's ISBN field simply starts showing the existing "is not a valid ISBN-10 or ISBN-13" error for garbage input, same as it already does for `12345`.

| Input | Before | After |
|---|---|---|
| `helloworld` | accepted | rejected |
| `AAAAAAAAAAAAA` | accepted | rejected |
| `!!!!!!!!!!` | accepted | rejected |
| `03064O6152` (letter O) | accepted | rejected |
| `097522980X` (valid ISBN-10) | accepted | accepted |
| `0975229804` → `097522980x` (lowercase x) | rejected | accepted |
| valid 13-digit ISBN | accepted | accepted |

## QA steps

1. Edit a product containing a PDF or EPUB file → Content tab → file settings.
2. Enter `helloworld` in the ISBN field → save.
3. Expect the "is not a valid ISBN-10 or ISBN-13" validation error.
4. Enter a valid ISBN (e.g. `978-3-16-148410-0`) → saves successfully.

## Test Results

All 13 examples pass locally, including new regression cases for `helloworld`, `AAAAAAAAAAAAA`, `!!!!!!!!!!`, mixed-letter input, ISBN-13 with an `X`, and a valid `X`-check-digit ISBN-10:

spec run

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent). Instructions: confirm and fix antiwork/gumroad-private#1229 (ISBN validator accepts non-digit garbage). The model reproduced the bug, wrote the grammar guard and regression specs, ran rubocop and the spec file locally, and opened this PR.


gumclaw:
Self-review notes:

- The guard runs after `delete("-")`, so hyphenated ISBNs still validate; em/en dashes are not stripped and fail the grammar check, matching the existing spec expectations.
- Lowercase `x` check character is a legal ISBN-10 representation — the old `isbn[-1] == "X"` comparison rejected it (coerced `"x".to_i` → 0), so `casecmp?` is a small correctness fix bundled with the guard, covered by the new spec.
- Values with spaces were already rejected before (space coerced to 0 corrupts the checksum in most positions) and remain rejected — no behavior change there.
- The stray `# 100` comment on the ISBN-13 sum line was leftover noise and is removed.
- Nil/non-string handling is untouched: `allow_nil`/`allow_blank` guards run before any string ops, same as before.

Adversarial review (fable-5): no P1/P2 findings. One P3 noted — the validator does not strip spaces, so `978 3 16 148410 0` is rejected; that matches pre-fix behavior and the product editor field, so leaving as-is.

slavingia:
Merge if the failure is flakey.

gumclaw:
Checked before merging — the failure wasn't flaky, it was real. Two pre-existing normalization specs (`product_file_spec.rb:1039`/`1044`) used `978316148410X` as a fixture, which isn't a checksum-valid ISBN-10. The old validator's `to_i` coercion happened to accept it; the strict validator correctly rejects it.

Fixed by swapping in `097522980X` (a checksum-valid ISBN-10 with an `X` check character, still exercising upcasing and whitespace stripping) and pushed. All 14 `product_file_spec.rb` ISBN examples plus the 13 validator specs pass locally. Auto-merge is armed, so this merges on green.

PR #6105, opened 2026-07-21
Remove fully-rolled-out product_level_support_emails feature flag

gumclaw:
## What

Removes the `product_level_support_emails` Flipper flag and its dead branches. The flag is globally enabled in production (`state=on, boolean=true`; prod-state audit request `20260721T025828Z-d0d55a50`, [gumroad-private#1208](https://github.com/antiwork/gumroad-private/issues/1208)) and the gated code has been stable since 2025-08.

## Why

Part of the rollout-flag cleanup requested on #6041 / gumroad-private#1208 (Sahil: "Do all A except content mod as kill switch"). Removing the check is behavior-neutral: the flag is fully on, so every production seller already sees the product-level support emails settings form and the per-product reply-to override on receipts.

## Changes

- `User::FeatureStatus#product_level_support_emails_enabled?`: removed, along with all four call sites.
- `User#product_level_support_emails`: no longer returns `nil` when the flag is off — always returns the grouped support-email list.
- `Link#support_email_or_default`: always prefers the product's `support_email` (previously fell back to the seller default when the flag was off — dead branch in prod).
- `Product::BulkUpdateSupportEmailService#perform`: removed the flag guard.
- `SettingsPresenter#main_props`: `product_level_support_emails` is always populated (never `nil`).
- `Settings/Main/Show.tsx`: prop type tightened from `ProductLevelSupportEmail[] | null` to `ProductLevelSupportEmail[]`; the null-guarded conditional render of `ProductLevelSupportEmailsForm` is now unconditional.
- Specs: removed flag activations and flag-off contexts in `user_spec`, `feature_status_spec`, `settings_presenter_spec`, `main_controller_spec`, `settings/main_spec` (request), `bulk_update_support_email_service_spec`.

## QA steps

Backend-flag removal with no user-visible change (flag is 100% enabled in prod, so production behavior is already the post-removal behavior).

1. Confirm `grep -rn "product_level_support_emails_enabled\|Feature.*product_level_support_emails" app lib config spec test` returns nothing.
2. Run the touched specs (see Test results).

## Before / After

No adjacent UI change: the Settings → Main "product-specific support emails" form renders identically for all production sellers before and after (the flag was globally on). The spec run below is the reviewable artifact for the behavior-neutral claim.

## Test results

spec run

- `spec/services/product/bulk_update_support_email_service_spec.rb` + `spec/modules/user/feature_status_spec.rb` + `spec/controllers/settings/main_controller_spec.rb`: **89 examples, 0 failures**
- `spec/models/user_spec.rb -e product_level_support_emails`: **2 examples, 0 failures**
- `npx tsc --noEmit`: no errors in the touched TSX; prettier clean.
- Note: `settings_presenter_spec.rb:293` ("unpaid balances") fails identically on clean origin/main in this local env (missing seeded Gumroad MerchantAccount) — pre-existing, unrelated.

Post-merge housekeeping (after deploy confirmation): `Flipper.remove(:product_level_support_emails)` via the audited console, logged against gumroad-private#1208.

---

## AI disclosure

🤖 Generated with **Claude Fable 5** (Anthropic), running as the Hermes agent "gumclaw" autonomous dev dispatcher. Prompts/instructions: continue the Category A rollout-flag cleanup per gumroad-private#1208 checklist and Sahil's "Do all A except content mod as kill switch" directive; the issue's audit table was used as the spec.


gumclaw:
Fable-style self-review — PASS (no P1/P2).

Hunt list checked:
- **Behavior delta beyond flag removal**: `Link#support_email_or_default` previously fell back to the seller-level email when the flag was off for that seller — since the flag is globally on, no production seller hits that branch today; post-removal behavior is identical. Same for the settings form render (`main_props` never returns nil in prod).
- **Client contract**: `main_props` was the only producer of the nullable prop; typia assert on the TSX side tightened in the same commit, so no other consumer sees a shape change (grep confirmed `product_level_support_emails` appears only in settings main + product bulk-update paths).
- **Flag-off spec contexts**: all six removed; no minitest twins (`grep -rn product_level_support_emails test/` is empty).
- **Residual references**: zero `Feature.` call sites remain for this flag.

Auto-merge armed per gumroad-private#1208 (Category A directive).

PR #6115, opened 2026-07-22
Strip Unicode whitespace from tax IDs and scrub encoding in settings presenter

gumclaw:
## What

Fixes `/settings/payments` 500ing for sellers whose stored `business_tax_id` (or `individual_tax_id`) contains Unicode whitespace such as U+202F (narrow no-break space). Two sides:

- **Write side** (`UpdateUserComplianceInfo`): tax ID normalization now uses the Unicode-aware `[[:space:]]` class instead of `\s` (ASCII-only), so locale-formatted pastes like a French SIREN `912 904 331` (U+202F separators) are stored clean. Also normalizes `individual_tax_id` submissions, which previously got no whitespace stripping at all (dashes preserved — meaningful for e.g. Peru DNIs).
- **Read side** (`SettingsPresenter`): new `tax_id_last_four` helper forces UTF-8, `scrub`s invalid bytes, and strips Unicode whitespace before slicing the last four characters — so existing corrupted rows render instead of 500ing the whole settings page.

## Why

Sentry GUMROAD-VZ (`ActionView::Template::Error: "\xAF" from ASCII-8BIT to UTF-8`) and GUMROAD-VY (`JSON::GeneratorError`) in `Settings::PaymentsController#show` — 8 events / 7 users. Strongbox decrypt returns BINARY; the byte-based `[-4..]` slice on a value ending in a 3-byte U+202F cuts the character in half, producing invalid UTF-8 that crashes JSON serialization of the Inertia page props.

Fixes antiwork/gumroad-private#1240

## Before / After

Not applicable — no UI change. Before: the page 500s for affected sellers; after: it renders with the correct masked last-four. Confirming evidence in Test results (presenter spec reproduces the exact crash input and asserts `payments_props` JSON-serializes).

## QA steps

1. As a non-US business seller, save a business tax ID containing U+202F (e.g. paste `912 904 331` with narrow no-break spaces).
2. Load `/settings/payments` — page renders; tax ID shows `•••••4331`.
3. For a legacy row (bad value already stored): page renders instead of 500ing.

## Test results

Ran locally in a worktree off `origin/main`:

spec results

`spec/services/update_user_compliance_info_spec.rb` + `spec/presenters/settings_presenter_spec.rb`: **112 examples, 0 failures**, including new specs for U+202F/NBSP stripping on write and JSON-serializability of props with a legacy corrupted row.

---

## AI disclosure

Written by Claude Fable 5 (Gumclaw), from gumroad-private#1240 (root cause pre-verified on a live affected seller). Instructions: normalize tax IDs Unicode-aware on write, scrub encoding defensively in the presenter, add specs for both sides.


gumclaw:
Self-review notes:

- `[[:space:]]` in Ruby regexes is Unicode-aware on UTF-8 strings and matches the whole Zs category (U+202F, U+00A0 NBSP, U+2009 thin space), which is exactly the class of characters locale-formatted pastes carry. `\s` only matches ASCII whitespace — that was the write-side hole.
- Individual tax IDs only get whitespace stripped (not dashes), since dashes are meaningful there — Peru DNIs are entered as `12345678-9` and the existing DNI validation depends on that shape.
- The presenter helper is deliberately defensive rather than relying on a backfill: `force_encoding(UTF_8).scrub("")` makes any legacy byte sequence safe to slice, and stripping whitespace before `[-4..]` means a legacy `…904<U+202F>331` row shows `4331` instead of 500ing the page.
- `encrypted_compliance_info_changed?` now normalizes the submitted individual tax ID before comparing against the stored value, matching how it's assigned — otherwise a resubmission with whitespace would spuriously create a new compliance row.
- A backfill sweep for existing corrupted rows is optional follow-up (tracked on gumroad-private#1240); the read-side scrub already makes them render correctly.

PR #6120, opened 2026-07-22
Document the per-file Send to Kindle and Read buttons toggle in the help center

gumclaw:
## What

Updates four help center articles for the new per-file setting that hides the "Send to Kindle" and "Read" buttons on the download page (shipped in #6069).

- `_247-what-your-customers-see` (seller-facing): adds a sentence pointing at the new file-embed switch — 'Show "Send to Kindle" and "Read" buttons on the download page' — and its effect (customers only see the Download button).
- `_112-mobile-friendly-files` (seller-facing): same pointer added to the in-browser reader paragraph.
- `_198-your-gumroad-library` (buyer-facing): notes that some creators turn the Read option off for certain files, so the button may be absent.
- `_208-how-do-i-send-my-gumroad-purchase-to-my-kindle` (buyer-facing): adds the per-file creator toggle as a reason the "Send to Kindle" button might not appear, pointing at the existing email-attachment fallback.

The switch label is quoted verbatim from `ProductEdit/ContentTab/FileEmbed.tsx`, and the "Download button only" behavior matches `UrlRedirectPresenter#map_file` nulling `kindle_data` and `read_url` for flagged files.

## Why

These articles described the Send to Kindle and Read buttons unconditionally. With #6069 merged, sellers can hide them per file, so the seller docs need to mention the control and the buyer docs need to explain why a button may not appear (otherwise buyers file "missing button" tickets).

## Test plan

- Rendered all four partials via `ApplicationController.render` in the test env — all render OK, per-tag open/close counts balanced.
- `bundle exec rspec spec/models/help_center` — 1 example, 0 failures (slug ↔ partial integrity).
- `articles.yml` untouched (body-only prose edits, no title/description changes).

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent) via the merged-PR → help-center sync automation, triggered by the merge of #6069.

### DIFF
--- a/app/views/help_center/articles/contents/_112-mobile-friendly-files.html.erb
+++ b/app/views/help_center/articles/contents/_112-mobile-friendly-files.html.erb
@@ -25,7 +25,7 @@
   </ul>
   <h3>Optimizing a PDF for Mobile:</h3>
   <p>Some customers would rather download ebooks to their devices instead of using the Gumroad app. For an optimized customer service experience, consider uploading multiple ebook formats. A PDF version is a safe default since most devices can display one, but we recommend also uploading an EPUB version: EPUB text reflows to fit any screen, which makes it much nicer to read on phones and e-readers.</p>
-  <p>Gumroad can read both PDF and EPUB files directly in the browser. Customers see a "Read" button next to these files on the download page, and the reader remembers where they left off. EPUB files over 32 MB are download-only. If your product has a PDF but no EPUB, you'll see a reminder on the product's content tab suggesting you add one.</p>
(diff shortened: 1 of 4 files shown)


gumclaw:
**fable-5 pre-merge review — verdict: ship (no P1/P2).**

Checked each claim against the merged code from #6069:

- Switch label quoted verbatim from `ProductEdit/ContentTab/FileEmbed.tsx` ('Show "Send to Kindle" and "Read" buttons on the download page'); it renders for `is_pdf || extension === "EPUB"` files, matching the articles edited.
- Default state: `checked={!file.hide_kindle_and_read_buttons}` — default on, so "toggle off" is the correct instruction.
- Effect: `UrlRedirectPresenter#map_file` nulls both `kindle_data` and `read_url` when flagged → "customers will then only see the Download button" is accurate.
- `_208`'s "(see below)" pointer resolves — the email-attachment fallback section exists in the same article.
- P3 (noted, not blocking): `_198`'s screenshot `read-button-download-page.png` still shows the Read button, which stays accurate for the default state.

All four partials render via `ApplicationController.render` with balanced tags; `spec/models/help_center` green (1 example, 0 failures). Docs-only — diff is the reviewable artifact.

PR #6133, opened 2026-07-22
Align review reminder eligibility with review-form eligibility

gumclaw:
## What

Makes `Purchase#eligible_for_review_reminder?` delegate its review-eligibility half to `Purchase::Reviews#allows_review_to_be_counted?` — the same gate that decides whether the product page renders the review form and whether a review is counted. Only the reminder-specific conditions remain hand-rolled: no existing review, seller reminder opt-out, purchaser reminder opt-out. Also flips the `ReviewReminderJob` reversed-chargeback spec: a reversed chargeback still blocks the review form (`allows_reviews` requires `chargeback_date.nil?`), so the reminder must stay suppressed to match.

## Why

The reminder check was a hand-rolled subset of the review gates and had drifted: it missed `should_exclude_product_review` and the access-revoked-and-unpaid gate, so buyers could get a reminder email whose deep link opens a product page with **no review form**. Flagged on #6078; confirmed on the production read replica (2026-07-22): of the 200 most recent purchases flagged `should_exclude_product_review` (1,466 created in the last 30 days), 152 still passed `eligible_for_review_reminder?`; of the 200 most recent access-revoked purchases, 100 did. Delegating to the shared gate fixes the current gap and prevents future drift — any new review-eligibility rule automatically applies to reminders.

Behavior deltas (all tightenings, no loosenings):
- `should_exclude_product_review` purchases: no longer reminded.
- Access-revoked unpaid purchases: no longer reminded.
- Reversed-chargeback purchases: no longer reminded (the form gate blocks any `chargeback_date`, reversed or not — the old check only blocked unreversed ones).
- Plain `not_charged` purchases: no longer reminded; free-trial `not_charged` purchases remain reminded, matching the form gate (`exclude_not_charged_except_free_trial`). In practice reminders are only scheduled from the successful-purchase transition (`schedule_order_review_reminder`), so this path is mostly theoretical.

Refs #6131 (checklist item 1; production verification stays open until this deploys).

## Before / After

Not applicable — no UI surface changes. The only user-visible effect is that reminder emails stop going to buyers who cannot review; the email template and the review form are untouched. Spec-run evidence below is the artifact.

## QA steps

1. Console: pick a purchase with `should_exclude_product_review: true` → `eligible_for_review_reminder?` → `false` (was `true`).
2. Console: pick an access-revoked free purchase → `eligible_for_review_reminder?` → `false` (was `true`).
3. Console: a plain successful purchase without a review → `true` (unchanged).
4. Post-deploy: confirm `CustomerLowPriorityMailer#purchase_review_reminder` volume drops slightly and no reminder lands for excluded purchases.

## Test Results

Local, macOS, `RAILS_ENV=test`:

- `bin/rails test test/models/purchase_test.rb -n "/eligible_for_review_reminder/"` → **13 runs, 13 assertions, 0 failures** (includes 3 new tests: excluded-from-reviews, access-revoked unpaid, reversed chargeback)
- `rspec spec/sidekiq/review_reminder_job_spec.rb` → **7 examples, 0 failures** (reversed-chargeback expectation flipped to match the form gate)
- `rspec spec/sidekiq/order_review_reminder_job_spec.rb` → **5 examples, 0 failures** on the first run of this session; the shared test env's Stripe key then started rejecting raw-card `create(:purchase)` factories session-wide (fails identically on unmodified `origin/main`), so the final combined re-run couldn't complete. The Minitest purchase tests (which use the local factory helpers, not Stripe) pass at head. CI will verify.
- `rubocop` on the three touched files → no offenses.

spec run

Permanent copy: `qa-media/pr-6131-reminder-eligibility-specs.png` on this branch.

## AI disclosure

🤖 Generated with claude-fable-5, Hermes agent — triggered by gianfrancopiana's post-merge review note on #6078; confirmed the gap on the prod read replica, then wrote the fix and specs.


gumclaw:
Fable adversarial review (self, pre-merge): **PASS — no P1/P2.**

Hunted the delegation for behavior loosenings and callsite blast radius:

- **Callsites** (3): `Order#schedule_review_reminder?`, `OrderReviewReminderJob`, `ReviewReminderJob`. All are boolean gates on sending an email — worst case of a wrong `false` is a missing reminder, never a wrong send to an ineligible buyer. The deltas are all tightenings except one theoretical loosening below.
- **Gift semantics preserved**: the shared gate carries `not_is_gift_sender_purchase`, and `gift_receiver_purchase_successful` remains in `COUNTS_REVIEWS_STATES` — giftee eligibility unchanged (pinned by the existing gifted-bundle spec).
- **Subscription edge (theoretical loosening)**: old check required `is_original_subscription_purchase? || link.not_is_recurring_billing?`; the shared gate uses `subscription_id.nil? || is_original_subscription_purchase?`. A subscription-less purchase on a recurring-billing product would now pass — but such a purchase also renders the review form, so reminding it is consistent by construction. Recurring charges stay blocked (existing spec passes).
- **Idempotency unchanged**: `SentEmailInfo.ensure_mailer_uniqueness` still wraps both jobs; no re-send risk.
- **Perf**: the gate adds `should_exclude_product_review?`/`stripe_refunded?` flag reads (no queries) and a possible subscription load in `not_charged_and_not_free_trial?` — low queue, per-order cardinality, negligible.

P3 (noted, not blocking): `Order#schedule_review_reminder?` calls the gate per purchase at checkout; if orders ever get large this becomes an N+1 on subscription loads.

PR #6137, opened 2026-07-22
Fall back to Unicode-normalized S3 keys before declaring a file missing

gumclaw:
## What

Retry S3 object lookups with the alternative Unicode normalization forms of the key before raising `Aws::S3::Errors::NotFound`, so sellers with accented filenames aren't wrongly told "This file is no longer available. Please re-upload it." for files that do exist in S3.

Fixes #6126 (Sentry [GUMROAD-P2](https://gumroad-to.sentry.io/issues/7460848648/) — 2,803 events / 138 unique sellers in 14 days).

## Why

S3 keys are byte-exact, but the same accented filename can be encoded two valid ways: precomposed (NFC — what we persist in `product_files.url`) or decomposed (NFD — what macOS uploaders write to S3). When the forms disagree, `signed_download_url_for_s3_key_and_filename` raises `NotFound` on a file that is really there. Verified against production (read-only console, 2026-07-22): a live failing key exists under its NFD form (`object(key.unicode_normalize(:nfd)).exists? → true`) while the DB's NFC form returns `false`.

## How

- New `S3KeyUnicodeNormalization` module: computes the distinct NFD/NFC variants of a key and returns the first one that exists in the bucket. Plain-ASCII keys have no variants, so **no extra S3 calls** for them — the genuinely-missing-object population still fails fast and keeps its Sentry signal.
- `SignedUrlHelper#signed_download_url_for_s3_key_and_filename`: in the existing `rescue Aws::S3::Errors::NotFound`, look up an existing variant and serve it; raise as before when none exists.
- `S3Retrievable#confirm_s3_key!`: try normalization variants first (and heal the record's `url` to the working form), before the existing single-file-directory fallback — which previously couldn't heal directories holding more than one file.

## Test evidence

Red-first: the new helper spec fails on `main` (`NotFound` raised despite the object existing under NFD), passes with the fix.

```
spec/modules/s3_key_unicode_normalization_spec.rb — 6 examples, 0 failures
spec/modules/s3_retrievable_spec.rb               — 25 examples, 0 failures*
spec/helpers/signed_url_helper_spec.rb            — 15 examples, 0 failures
spec/modules/with_file_properties_spec.rb         — 36 examples, 0 failures
```

\* one pre-existing local-env failure (`restore_deleted_s3_object!` versioned-object restore) reproduces identically on clean `main` — unrelated to this change (local test S3 lacks versioning).

Local test S3 normalizes Unicode keys (unlike real S3), so the mismatch specs stub the SDK boundary to reproduce production's byte-exact behavior; the spec comments call this out.

## Follow-up (tracked in #6126)

Audit alive `product_files` whose object is absent under any normalization — the truly-missing sub-population (rows pointing at deleted/never-completed uploads).

🤖 Generated autonomously by Gumclaw. Backend-only change — no UI surface, no screenshots.

---

## AI disclosure

Authored autonomously by Gumclaw (Hermes agent, claude-fable-5) from Sentry issue GUMROAD-P2 / issue #6126. Prompt: Sahil commented "Fix" on #6126; agent root-caused via prior prod console verification, wrote red-first specs, implemented the normalization fallback, and verified locally.

## QA steps

Backend-only signed-URL path — no UI change; the diff plus red-first spec run above is the reviewable artifact. Post-deploy verification: GUMROAD-P2 event volume should drop to only the truly-missing-object population (tracked as follow-up in #6126).


gumclaw:
## Fable-5 pre-merge review (adversarial, independent reviewer at `c968ac3`)

**Verdict: CHANGES-REQUIRED → fixed in `98288ea`. Now READY-TO-MERGE.**

|---|---|---|
| New file in legacy `app/modules/` (CONTRIBUTING.md forbids new files there) | P2 | Fixed — moved to `app/services/s3_key_unicode_normalization.rb` (+ spec to `spec/services/`) |
| Rescue list missed `Encoding::InvalidByteSequenceError` (unreachable for UTF-8 DB strings, but cheap) | P3 | Fixed — `rescue ArgumentError, EncodingError` covers all encoding exceptions |
| Variant deleted between `exists?` and `content_length` (race) would lose the `Key =` decoration → differently-grouped Sentry event | P3 | Fixed — inner rescue re-raises with the same descriptive message |
| Spec title said "directory holds several files" but uploads nothing | P3 | Fixed — retitled to match what it proves (directory-listing fallback can't identify the file; still reds on revert) |
| Genuinely-missing accented keys now cost up to 3 uncached S3 calls per request (fetch doesn't cache raising blocks); helper heals the served URL but not the DB row | P3 | Documented in the code comment — acceptable on an already-failing path; the record-healing path is `confirm_s3_key!` |
| `download_original` (PDF stamping/archives) + `UpdateProductFilesArchiveWorker` still fail hard on normalization mismatch | P3 | Follow-up noted on #6126 — this PR targets the signed-URL path that produces GUMROAD-P2 |

**Adversarial claims confirmed SAFE** (each verified against code, not the PR body):
- **No cache poisoning**: the cached tuple's `s3_path` derives from the NFD object's `public_url`, never from the NFC `s3_key`; sole consumer of the cache key is this method.
- **Zero behavior change for ASCII keys**: `variants` returns `[]` with no S3 calls; the raise path is byte-identical to pre-PR, Sentry signal for truly-missing objects preserved.
- **`confirm_s3_key!` `update!` semantics pre-existing** (old single-file-directory branch did the same); sole prod caller is `analyze` on persisted records.
- **Spec vacuity**: load-bearing specs red-first verified empirically on `main` (fallback spec raises `NotFound` without the fix).
- **MinIO branch** is dev/test-only per `config/initializers/aws.rb`.

Empirical checks: `String#unicode_normalize` exception classes probed in a live Ruby session (ArgumentError for invalid UTF-8, Encoding::CompatibilityError for non-Unicode encodings). Full spec run at head: 46 examples across the 3 touched spec files + 36 in `with_file_properties_spec`, 0 failures except the pre-existing `restore_deleted_s3_object!` local-env failure (reproduces identically on clean `main` — local test S3 lacks versioning). Rubocop clean.

**Only live-verifiable**: GUMROAD-P2 volume drop post-deploy (residual events = the truly-missing-object population; audit tracked in #6126).

Auto-merge re-armed; CI gates the merge.

PR #6147, opened 2026-07-22
Guard product-editor saves from wiping variants and content without explicit deletion intent

gumclaw:
## Summary

A seller's product lost its entire version tree (VariantCategory, BaseVariants, all RichContent pages, ProductFiles — soft-deleted in the same second) from a single ordinary edit-page save, **three times in eight days** — the third time triggered by nothing more than clicking the version dropdown, with confirmed buyer impact (customers saw empty download pages mid-sale).

**Root cause:** the editor always submits the FULL variants + pages list on every save, and the server treats anything missing from the payload as removed. A stale tab or a race between two editor sessions submits an outdated list, and the save silently wipes everything. There was no deletion-intent signal server-side (purchased variants were protected; content-bearing-but-unpurchased variants and ALL pages were not).

## Changes

**Server-side guards (the fix that holds regardless of client bugs):**
- `Product::VariantCategoryUpdaterService.ensure_deletion_intent!` — blocks deleting any variant that still has content (rich-content pages or files) unless the seller explicitly confirmed the removal. Wired into both variant-deletion paths and whole-category deletion in `VariantsUpdaterService`.
- New `Product::RichContentDeletionGuard` — the page analogue, for both product-level (controller) and variant-level (updater service) page deletion. A page whose id appears elsewhere in the payload is a *move* (the "same content for all versions" toggle), not a deletion, and is allowed. Blank pages delete freely.
- Blocked saves roll back the transaction, return a "product may have been updated in another tab — refresh and try again" 422, and `ErrorNotifier.notify` so blocked wipes surface in Sentry (each one is a stale-payload incident we want to see).

**Client-side deletion intent:**
- New `confirmed_removed_variant_ids` / `confirmed_removed_rich_content_ids` sent with every save; recorded by the Remove version/tier/duration confirmation modals, the delete-page modal, copy-content-from-version, and the "use same content for all versions" discard flow.

**Buyer-facing fix from the same incident:**
- The wipe also left behind an alive-but-empty product-level page (single bare paragraph), which made `UrlRedirect#product_or_cheapest_variant_as_rich_content_provider` treat the product as content-bearing — variant-less buyers got ONE BLANK PAGE instead of the real variant content, *even after the variants were restored*. New `RichContent#has_editor_content?` (recursive node check) replaces the `rich_content_json.present?` test in the resolver.

## Tests

- 10 new controller specs reproducing the incident: stale payload omitting a content-bearing variant/page is **blocked** (verified RED on the pre-fix code); confirmed removals, empty-variant/page cleanup, and page moves between product/variant level all still work; ErrorNotifier assertion.
- 6 new `has_editor_content?` model specs; 1 url_redirect regression spec (empty placeholder no longer hijacks content resolution for variant-less purchases); updater-service spec for the unconfirmed-deletion raise.
- 4 existing specs updated where they relied on unconfirmed content wipes (each now uses confirmed ids or page-move semantics, with comments).
- Full runs: links_controller_spec 347 ✓, url_redirect_spec / rich_content_spec / both updater-service specs ✓ (2 pre-existing `enqueue_job_to_regenerate_deleted_transcoded_videos` failures reproduce on clean main without this diff). Rubocop, tsc, eslint, prettier clean.

## QA steps

1. Open a product with versions+content in TWO tabs; delete a version in tab A and save; save tab B (stale) → save is rejected with the refresh message, content intact.
2. Remove a version via the confirmation modal → save succeeds, version deleted.
3. Delete a page via the delete-page modal → succeeds. Toggle "use same content for all versions" both ways → succeeds.
4. For a variant-less purchase of a product whose only real content is variant-level, with an empty product-level placeholder page present → download page shows the variant content.

🤖 Generated with assistance from Claude

**AI disclosure:** authored by gumclaw (agent). @gianfrancopiana will validate locally before this leaves draft.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -408,7 +408,9 @@ class LinksController < ApplicationController
           :call_limitation_info,
           :installment_plan,
           :community_chat_enabled,
-          :default_offer_code_id
+          :default_offer_code_id,
+          :confirmed_removed_variant_ids,
+          :confirmed_removed_rich_content_ids
         ))
         @product.description = SaveContentUpsellsService.new(seller: @product.user, content: product_permitted_params[:description], old_content: @product.description_was).from_html
         @product.skus_enabled = false
@@ -457,7 +459,14 @@ class LinksController < ApplicationController
           rich_content.update!(title: product_rich_content[:title].presence, description: product_rich_content[:description].presence || [], position: index)
           rich_contents_to_keep << rich_content
         end
-        (existing_rich_contents - rich_contents_to_keep).each(&:mark_deleted!)
+        product_rich_contents_to_delete = existing_rich_contents - rich_contents_to_keep
+        Product::RichContentDeletionGuard.ensure_intent!(
+          product: @product,
+          rich_contents_to_delete: product_rich_contents_to_delete,
+          payload_page_ids:,
+          confirmed_removed_ids: confirmed_removed_rich_content_ids
+        )
+        product_rich_contents_to_delete.each(&:mark_deleted!)
 
         Product::SaveIntegrationsService.perform(@product, product_permitted_params[:integrations])
         update_variants
@@ -802,6 +811,9 @@ class LinksController < ApplicationController
               options: variants,
             }
           ],
+          confirmed_removed_variant_ids:,
+          payload_page_ids:,
+          confirmed_removed_rich_content_ids:,
         ).perform
       elsif variant_category.present?
         Product::VariantsUpdaterService.new(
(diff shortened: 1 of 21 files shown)


slavingia:
May also be good to have a confirm modal when saving a product which deletes (a lot of) stuff?

gumclaw:
@slavingia good call — a bulk-deletion confirm modal fits neatly on top of the server guard (the guard makes stale saves fail safely; the modal makes intentional big deletions legible before submit). Plan: when a save would delete N≥3 content-bearing items (variants/pages/files), show a confirm listing exactly what's being removed, and wire the existing explicit-deletion-intent flag to that confirmation. I'll push it to this branch today so it lands in the same review pass — it stays draft until @gianfrancopiana's local validation either way.

gumclaw:
Superseded by #6178 — same branch, same diff, re-opened from the org repo instead of the fork. With #6177 closed, fork-headed PRs can't check out under `pull_request_target`, so CI can never go green here. All context and review continues on #6178.

gumclaw:
Superseded by #6178 (same commits, main-repo branch — per policy gumclaw works on the main repo directly; fork lane retired for internal work, ci-protected gate restored in #6179).

PR #6149, opened 2026-07-22
Scope license keys to variants whose content embeds a license-key block

gumclaw:
## What

Scopes license-key generation and display to the purchased variant's content when a product uses per-variant rich content. Previously, licensing was purely product-level: if ANY variant's content embedded a license-key block, `link.is_licensed` flipped for the whole product, and every purchase — including variants whose content has no license-key block — got a `License` row and a key in the receipt email.

Changes:
- `Purchase#uses_license_key?` (new): product must be licensed AND the purchased variant's content must permit a key.
- `Purchase#variant_content_permits_license_key?` (new): for per-variant-content products, checks whether the purchased variant's rich content embeds a license-key block. Product-level content, physical products, purchases without a recorded variant, and variants with no authored rich content keep today's behavior.
- `Purchase#create_license!` and `Purchase#license_key` now gate on the new predicate instead of the bare `link.is_licensed` flag.
- `CreateLicensesForExistingCustomersWorker` (backfill when a seller enables licensing) applies the same variant check.
- `Purchase.purchase_info` now keys `license_key` off `purchase.license_key` (variant-aware) instead of `purchase.license` presence.

Existing `License` rows for free-variant purchases are left in place (harmless); `#license_key` no longer surfaces them.

## Why

A seller offering FREE and PRO variants of one product (license keys only meaningful for PRO) found free-variant buyers still received a license key in their receipt email. The key is unusable in the seller's app, confusing buyers and generating support load. Removing the license-key block from the free variant's Content tab only hid the key on that variant's download page — the receipt still showed it because generation/display were keyed off the product-level flag.

With this change, deleting the License key block from a variant's Content tab stops that variant from emitting keys — per-variant control with zero new UI.

Refs: antiwork/gumroad-private#1248

## Before/After

Pure model/worker/receipt-content change — no new UI surface. Before: a free-variant purchase of a licensed product received a license key in its receipt email. After: it doesn't; paid/licensed-variant purchases are unchanged. Test results below exercise both paths.

## Test plan

- New Minitest cases in `test/models/purchase_test.rb`: license created for product-level content; skipped for a variant without a license-key block while a sibling licensed variant still gets one; created when a product uses shared content for all variants; created for purchases without variants; existing license hidden from `#license_key` for a keyless variant; false when product not licensed; true for a variant with no authored content.
- New RSpec case in `spec/sidekiq/create_licenses_for_existing_customers_worker_spec.rb`: backfill skips keyless-variant purchases.
- Ran locally: 38 Minitest runs (license/purchase_info scope), 0 failures; worker spec 5 examples, 0 failures. `receipt_presenter/item_info_spec.rb` also run — one pre-existing failure (merchant-account factory collision) reproduces identically on unmodified main with the diff stashed.

## Test Results

test results

## QA steps

1. Create a digital product, add two variants (Free, Pro) with separate content per variant.
2. In the Pro variant's Content tab insert a License key block; leave it out of the Free variant.
3. Purchase the Free variant → receipt email has no license key; download page shows no key.
4. Purchase the Pro variant → receipt email includes the license key as before.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent), triggered by Sahil's approval of the proposed fix on the tracking issue. Instructions: implement per-variant license-key scoping as described in gumroad-private#1248 (gate `create_license!`/`license_key` on the purchased variant's embedded license block when the product uses per-variant content, keep product-level behavior otherwise), with tests.


gumclaw:
Self-review / adversarial pass (Claude Fable 5):

- **Recurring charges:** `Purchase#license` delegates to the subscription's original purchase for recurring charges, so `create_license!` early-returns on `license.present?` — no behavior change for renewals.
- **Gift flows:** gift-sender purchases still skip via the existing guard; giftee purchases go through the same variant check as regular purchases.
- **Existing license rows:** purchases of keyless variants that already have a `License` row keep the row, but `#license_key` (and everything downstream: receipt presenter, `purchase_info`, `license_json`, exports) returns nil. Keys already issued remain verifiable via the license-verify API — intentional, per the issue's "harmless to leave" note.
- **`purchase_info` change:** previously `json[:license_key]` was set (possibly to nil) whenever a `License` row existed; now it's set only when a key should actually surface. Callers treat absence and nil identically.
- **Variants with no authored content:** deliberately treated as permitting a key (suppression only when the seller authored variant content and left the license block out) — avoids breaking licensed products where some variants simply have no rich content yet.
- **Backfill worker:** enabling licensing on a product with existing sales now skips keyless-variant purchases; idempotent as before.
- **Pre-existing failure:** `spec/presenters/receipt_presenter/item_info_spec.rb:268` fails identically with the diff stashed (merchant-account factory collision) — unrelated to this change.

No P1/P2 findings.

PR #6150, opened 2026-07-22
Whitespace-tolerant snippet matching for store agent page edits + surface preview failures on the proposal card

gumclaw:
## What

Store-agent custom-HTML edit proposals (`edit_user_custom_html`) now match their `find` snippet with whitespace tolerance, and a failed proposal preview surfaces its reason on the card instead of silently pinning Confirm disabled.

- New `Ai::CustomHtmlSnippetMatcher`: tries an exact match first; when nothing matches exactly, retries with every whitespace run in the snippet (spaces, tabs, newlines, Unicode spaces like U+00A0, and the `&nbsp;` entity) matching any whitespace run in the page. Non-whitespace content stays literal, and the exactly-once uniqueness rule is unchanged.
- Both the real edit endpoint (`Api::V2::UsersController#edit_custom_html`) and the proposal preview endpoint (`Api::Internal::AgentCustomHtmlPreviewsController#create`) match through the shared class, so preview and apply always agree.
- The proposal card's preview-error state is now a `role="alert"` in danger styling that names the failure and tells the seller the way out ("Ask the agent to re-read the page and propose the change again"), replacing the muted "Preview unavailable" footnote.

## Why

A seller was stuck for hours on a proposal whose Confirm button never enabled. The staged find snippet differed from the stored page by a single character: the page's cursor span held a non-breaking space (U+00A0), the snippet a plain ASCII space — the model normalized the whitespace when echoing the page back. `String#scan` found zero matches, the preview POST failed every time, `confirmBlockedOnPreview` kept Confirm disabled, and the only signal was a muted footnote. The agent then re-staged the same unmatchable snippet repeatedly, blaming the seller's browser.

Whitespace-tolerant matching fixes the whole class (NBSP↔space, entity vs character, newline/indentation differences), the same way agent coding tools apply patches with whitespace fuzz. The prominent error means any residual failure (page genuinely changed underneath, ambiguous snippet) is visible and actionable instead of a dead button.

Fixes antiwork/gumroad-private#1251
Refs antiwork/gumroad-private#1054, antiwork/gumroad-private#953

## Before/After

Proposal card states captured from the real rendered component (AgentChat mounted in a local Vite harness with the app stylesheets; only the data layer scripted, since reproducing the error state live requires racing a page change under an already-staged proposal). Buttons are the design-system `Button` component — before/after error states in light mode, the new error state in dark mode, and the loaded state with Confirm enabled:

before-after card states

The matching change itself has no visual surface when it succeeds — the edit simply applies, as it always should have.

## QA steps

1. As a seller with `custom_html_pages` enabled and a published custom page containing an NBSP (e.g. `<span>&nbsp;</span>`), ask the store agent to edit text near that span.
2. Previously: the proposal card shows "Preview unavailable…" in muted text and Confirm never enables. Now: the preview renders and Confirm enables; confirming applies the edit.
3. To see the new error state: have the agent stage an edit, change the page underneath it (edit the custom HTML directly), then view the proposal — the card shows the danger alert naming the mismatch, Confirm stays disabled, Dismiss works.

## Test Results

test results

- `rspec spec/services/ai/custom_html_snippet_matcher_spec.rb spec/controllers/api/v2/users_controller_edit_custom_html_spec.rb spec/controllers/api/internal/agent_custom_html_previews_controller_spec.rb` — 60 examples, 0 failures
- `vitest run app/javascript/components/Agent/AgentChat.test.tsx` — 16 passed
- rubocop / prettier / eslint clean on all touched files

---

AI disclosure: Written by Claude Fable 5 (Hermes agent), triggered by Sahil's "Fix" comment on gumroad-private#1251. Instructions: implement whitespace-tolerant find-snippet matching for store agent page edits and surface preview failures on the proposal card.


slavingia:
Super hard to read that confirm button text. Fix? We have design system button components etc to use that should fix this.

gumclaw:
Fixed — the unreadable Confirm was in the screenshot itself, not the app. The Before/After image was a hand-drawn HTML mockup with a fake button (pink fill + faded text at 40% opacity). The actual card already renders the design-system `Button` (`color="accent"`), which greys out properly when disabled.

Replaced the evidence with captures of the real component: AgentChat mounted in a local Vite harness with the app stylesheets, only the data layer scripted. New image shows before/after error states in light mode, the error state in dark mode, and the loaded state with Confirm enabled. No code change needed — the diff already used the design-system components throughout.

gumclaw:
Fable-5 pre-merge review (inline, full diff at head `4187b19f`): **READY-TO-MERGE**.

Adversarially checked:
- **No pattern injection**: non-whitespace segments go through `Regexp.escape` (pinned by the \$10-(sale) metacharacter spec); the exact pass uses `String#scan`/`sub` with a String, which is always literal.
- **No ReDoS shape**: the tolerant pattern is literal segments joined by `(?:[[:space:]]|&nbsp;)+` — the alternation is unambiguous (`&` isn't in `[[:space:]]`) and there are no nested quantifiers, so no catastrophic backtracking.
- **Tolerance can't broaden the edit**: whitespace runs generalize to non-empty whitespace runs only (`a b` never matches `ab` — pinned by spec); uniqueness rule unchanged; exact match always preferred, including ambiguous exact multi-matches (pinned).
- **Preview/apply parity**: both endpoints route through `Ai::CustomHtmlSnippetMatcher` and both `sub` on `match.matcher` with the block form (backslash-safe replacement preserved).
- **Empty/degenerate `find`**: behaves identically to the old `scan` path (reported as ambiguous), no new 500 surface.

The earlier Test Fast 4 red was an unrelated `customers_controller_spec` flake — rerun of the same head is fully green (77 SUCCESS / 0 FAILURE). Merging per Sahil's "Fix" directive on gumroad-private#1251.

PR #6155, opened 2026-07-22
Update license keys help article for per-variant license key scoping

gumclaw:
Follow-up docs for #6149 (merged), which scoped license key issuance to variants whose content embeds a License key block.

## What
- Adds a paragraph to the setup section of the license keys help article explaining per-version key issuance: customers get a key only when the version they purchased has the License key block in its content; removing the block from a version stops it issuing keys; versions without their own content still get keys when the product has license keys enabled.
- Corrects the retroactive-keys FAQ answer: the backfill now applies the same variant check, so keys are generated only for purchases of versions whose content includes the block.

## QA
- ERB syntax check passes; HTML tag balance verified on the edited partial.
- Content matches the merged behavior in #6149 (`Link#has_license_key_block?` variant scoping incl. the retroactive backfill worker).

Docs-only change, no user-facing code.

🤖 Authored by Gumclaw (AI agent).


gumclaw:
Fable-5 pre-merge review (docs-only diff, inline): **READY-TO-MERGE**.

Verified the copy against the merged #6149 behavior (`ba418984` on main): `Purchase#uses_license_key?` → `variant_content_permits_license_key?` returns true for product-level rich content, purchases without a variant, and variants with **no rich content of their own** — the article's "versions without their own content still get keys" sentence matches the code exactly. The retroactive-backfill FAQ correction matches `CreateLicensesForExistingCustomersWorker` applying the same variant check. Internal article link target (`126-setting-up-versions-on-a-digital-product`) exists. No user-facing code touched.

PR #6157, opened 2026-07-22
Add covering index for affiliate earnings sum on purchases

gumclaw:
Closes the last unchecked item on #6020 (follow-up: index or cache for `Affiliate#total_cents_earned` — #6121 shipped the cache; this is the index so the cache-miss recomputation is cheap too). Sahil's "Makes sense, continue" on #6020 green-lit continuing down the list.

## What
Adds `idx_purchases_on_affiliate_earnings_sum` on `purchases (affiliate_id, purchase_state, price_cents, stripe_refunded, chargeback_date, flags, affiliate_credit_cents)`.

## Why
`Affiliate#total_cents_earned` runs
```sql
SELECT SUM(affiliate_credit_cents) FROM purchases
WHERE affiliate_id = ? AND purchase_state = 'successful' AND price_cents > 0
  AND (stripe_refunded IS NULL OR stripe_refunded = 0)
  AND (chargeback_date IS NULL OR (chargeback_date IS NOT NULL AND flags & 512 != 0))
```
The only usable index was `(affiliate_id, created_at)`, so MySQL fetched every one of the affiliate's purchase rows from the clustered index just to evaluate the predicates and read `affiliate_credit_cents` — ~700ms per page view for large affiliates (production Sentry span, #6020 item 3). The new index covers the query exactly: equality on `affiliate_id`, every filtered column, and the summed column all come straight out of the index.

## Evidence
EXPLAIN on the test DB seeded with 20k purchases across 500 affiliates: `type: range`, `key: idx_purchases_on_affiliate_earnings_sum`, `Extra: Using where; Using index` (index-only). Screenshot in `qa-media/`. Prior attempt #4314 tried switching the sum to `affiliate_credits` and got mismatched numbers — this keeps the query identical and just indexes it, so the displayed figure cannot change.

## Notes
- `disable_ddl_transaction!` + plain `add_index`, matching the repo's online-DDL migration pattern (#6027 used the identical shape for the `affiliate_credits` sum).
- Schema-only change: per the no-auto-merge policy for schema PRs this is NOT armed for auto-merge — needs a human merge.

🤖 Authored by Gumclaw (AI agent).


gumclaw:
@ershad assigning to you for review + merge call: this is a **schema migration** (covering index on `purchases`), which is on the no-auto-merge list regardless of size. The index itself is additive/online (InnoDB `ADD INDEX` is in-place), but purchases is our biggest table so the build window and replica lag are worth a human eyeball on timing. EXPLAIN evidence is in qa-media on the PR.

PR #6180, opened 2026-07-23
Don't let the settlement-currency-mismatch marker suppress the forced-currency local-method lane

gumclaw:
Fixes the structural cause of the iDEAL ramp-down at 2026-07-23 07:19 UTC (antiwork/gumroad-private#933 — [ramp-down evidence](https://github.com/antiwork/gumroad-private/issues/933#issuecomment-5055581413)). Part of antiwork/gumroad#5362 Phase 4.

## Problem

#6163's per-currency EUR settlement-mismatch marker on the platform merchant account (MA 1) made `Checkout::BuyerCurrencyEligibility#method_forced_decision` return `unsupported_settlement_currency` for **every Gumroad-managed seller**, so the iDEAL tab never rendered: zero iDEAL attempts ever, across 5h15m at 100% with eligible NL/BE EUR traffic flowing.

The trap is self-inflicted by design: enabling the iDEAL/SEPA capabilities is itself what makes the platform account settle EUR-in-EUR, so a fresh EUR marker is the *expected steady state* for exactly the accounts that can take iDEAL. The marker only predicts that an **FX quote to USD** for that currency will be rejected — it says nothing about a plain forced-currency charge, and the direct-listed-amount lane (EUR-priced product paid with iDEAL) never mints an FX quote at all.

## Fix

Split the settlement question in two:

- New `usd_holding_merchant_account?` (stored-currency check only, marker-blind) now gates the method-forced lane's blanket check and `PaymentMethodResolver#forced_currency_settlement_supported?` — the shapes those paths serve charge the listed price with no FX quote.
- The marker-aware `usd_settling_merchant_account?` is unchanged for the card-path quote decision, and on the method-forced lane it now applies **only to the USD-priced (quoted) case**, where the doomed FX quote would actually be minted and there is no graceful USD fallback for a method that can only charge in its forced currency.

## Testing

- Regression specs in both `buyer_currency_eligibility_spec` and `payment_method_resolver_spec`: an EUR marker keeps the direct-listed-amount lane eligible and the iDEAL tab offered; the USD-priced quote case is still withheld; a non-USD *stored* balance still withholds everything.
- Full `buyer_currency_eligibility_spec` + `payment_method_resolver_spec` run locally: 109 examples, 1 failure — the failure (`allows the PR1 Stripe test-mode Gumroad platform-account path`, a uniqueness-validation collision on `create(:merchant_account, ...)`) reproduces identically on unmodified `origin/main`, i.e. pre-existing local-DB residue, not introduced here.

## Rollout

After merge + ancestry-verified deploy, re-ramp: `Feature.activate_percentage(:checkout_local_method_ideal, 100)` (ramp executed by the monitor per the max-data doctrine).

### DIFF
--- a/app/services/checkout/buyer_currency_eligibility.rb
+++ b/app/services/checkout/buyer_currency_eligibility.rb
@@ -98,7 +98,7 @@ class Checkout::BuyerCurrencyEligibility
   end
 
   def self.usd_settling_merchant_account?(merchant_account, presentment_currency:)
-    return false unless merchant_account.currency.blank? || merchant_account.currency.to_s.downcase == Currency::USD
+    return false unless usd_holding_merchant_account?(merchant_account)
 
     # The stored currency answers the wrong question for accounts with Stripe
     # multi-currency settlement enabled: it mirrors Stripe's default_currency ("usd"),
@@ -111,6 +111,21 @@ class Checkout::BuyerCurrencyEligibility
     # trip for that currency (up to 2s of checkout latency, on every visit) and fall
     # back to canonical USD immediately. Other currencies keep quoting.
     !merchant_account.settlement_currency_mismatch_active?(presentment_currency)
+  end
+
+  # The weaker of the two settlement questions: does the account HOLD its balance in USD
+  # (per our stored mirror of Stripe's default_currency)? Unlike
+  # usd_settling_merchant_account? this deliberately ignores the learned
+  # settlement-currency-mismatch marker, because the marker only says "an FX quote to USD
+  # for this currency will be rejected" — it does not make a plain forced-currency charge
+  # fail. The direct-listed-amount forced lane (an EUR-priced product paid with iDEAL)
+  # never mints an FX quote, so the marker is irrelevant to it; gating that lane on the
+  # marker is exactly what turned iDEAL dark platform-wide on 2026-07-23 (enabling the
+  # iDEAL/SEPA capabilities made the platform account settle EUR in EUR, the EUR marker
+  # was recorded, and every Gumroad-managed seller lost the iDEAL tab —
+  # gumroad-private#933).
+  def self.usd_holding_merchant_account?(merchant_account)
+    merchant_account.currency.blank? || merchant_account.currency.to_s.downcase == Currency::USD
   end
 
   def self.stripe_test_mode?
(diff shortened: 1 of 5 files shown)


gumclaw:
CI's only red job (Test Fast 8) was `stripe_payment_presenter_spec.rb:826` — the old expectation asserted the exact over-suppression this PR removes (marker set ⇒ USD element even for the direct-listed-amount EUR cart). Pushed [3051fe0c3](https://github.com/antiwork/gumroad/commit/3051fe0c3b2bd2dcd608e658c3d42b6bb1720742): the example now pins the fixed behavior (EUR element keeps the iDEAL tab; the marker only matters on the FX-quoted USD-priced case), with a comment explaining why, plus a self-seeding platform merchant-account lookup so the example doesn't depend on DB seed state (same pattern as `buyer_currency_quote_spec`). Full presenter spec green locally: 78 examples, 0 failures; rubocop clean.

gumclaw:
CI is fully green on [3051fe0c3](https://github.com/antiwork/gumroad/commit/3051fe0c3b2bd2dcd608e658c3d42b6bb1720742) ([ci/green](https://github.com/antiwork/gumroad/actions/runs/29994282522) — the Test Fast 8 presenter-spec failure is resolved) and Greptile reports no blocking issues (5/5). Nothing agent-side remains — over to @rmarescu for review. iDEAL re-ramp (`checkout_local_method_ideal` → 100) is blocked on this merging + ancestry-verified deploy.

slavingia:
Fable 5 adversely review?

gumclaw:
Fable 5 adversarial review — done now against head 3051fe0c3 (local worktree, full diff vs main plus a read-through of both services, `Charge::MethodForcedPresentment`, `StripeFxQuote`, and the `MerchantAccount` marker model).

**Verdict: no P1/P2 findings. Safe to merge.** Two P3 observations below, neither blocking.

What I hunted for and what I found:

- **Render/prepare divergence (a tab that renders but fails at prepare).** The resolver's marker-blind check mirrors exactly the marker-blind blanket gate in `method_forced_decision` (line 232), and the marker-aware check survives only on the quoted (USD-priced) case. The resolver only offers forced-currency methods on carts priced in the forced currency (the `select` on `cart_product_currency`), so a resolver-offered method can never reach the quoted marker check at prepare — the two sides can't disagree. Consistent.
- **Card/Link tokens minted on a forced-currency element.** The presenter only mounts a forced element for a product priced in that currency, so those tokens always land in the direct-listed case, which the marker no longer touches. No stranded-token path.
- **Charge-time failure on the newly-unsuppressed lane.** The direct-listed lane in `Charge::MethodForcedPresentment` never mints an FX quote, so the `SettlementCurrencyMismatch` rescue there can only fire on the quoted lane — which the marker check now correctly withholds before render. The lane this PR reopens has no quote to reject.
- **Spec vacuity.** The presenter spec's platform-account factory fallback (`user: nil`) matches `MerchantAccount.gumroad`'s `user_id: nil` lookup, so the regression test exercises the real lookup path rather than a detached record. The three new eligibility specs cover marker-set/direct-listed (eligible), marker-set/USD-priced (withheld), and marker-for-other-currency (quote path stays open) — the right triangle.
- **Card-mode `#decision`** stays fully marker-aware via `usd_settling_merchant_account?` — untouched, correct (that path always mints a quote).

P3 observations (no action needed on this PR):

1. Connect sellers holding a non-USD balance are still fully withheld from the local-method lane by `usd_holding_merchant_account?` — even an EUR-holding EU connect account can't offer iDEAL on a direct-listed EUR product, where the charge shape is the most natural fit. That's pre-existing behavior correctly preserved here, but it's the next over-suppression candidate if the Phase 4 cohort ever ramps to connect sellers.
2. The platform account's EUR marker is now a permanent, self-refreshing fixture (card path re-probes after the 30-day TTL, gets rejected, re-records). That means card-path EUR presentment of USD-priced products on the platform account is effectively permanently off — which is correct given Stripe settles EUR in EUR there, just worth knowing when reading marker state in prod.

CI is fully green at this head, so the pre-merge checklist is complete from my side.

rmarescu (approved):
Recommendation: add preview QA evidence.

Issue #1412, opened 2025-09-30
Customers able to restart installment subscription from Manage Membership page leading to duplicate charges


vatsalkaushik:
### What

A customer with a completed installment purchase can restart the associated subscription from the Manage Membership page. Doing so charges the customer again, even though their installment plan was already completed.

### Expected behavior
Installment-related subscription records should not be exposed to customers. Customers should not be able to restart or be re-charged once their installment plan is complete. The associated subscription record exists only for internal tracking. It should not be accessible to customers through their Manage Memberships page.

Mit-Patwa:
@vatsalkaushik  Could you  guide me on how to test this out in the dev version, since there are no products, we cannot test the behaviour of the "Manage Membership" page .

Mit-Patwa:
@vatsalkaushik could you please guide me how do  I create a product in dev ?

PR #6195, opened 2026-07-23
Add composite (email, seller_id) index to purchases for mobile library search

gumclaw:
## What

Adds a composite index on `purchases (email(191), seller_id)` — `index_purchases_on_email_and_seller_id` — mirroring the shape of `index_purchases_on_email_and_link_id` from #6123 (191-char email prefix, `disable_ddl_transaction!`).

## Why

`Api::Mobile::PurchasesController#search` is still hitting Rack::Timeout and multi-second p99s after #6123/#6158 deployed (issue #6185, Sentry GUMROAD-8T). The remaining dominant query shape is per-seller email-scoped existence probes:

```sql
SELECT purchases.id FROM purchases
WHERE email = ? AND seller_id = ? AND purchase_state IN (...) AND ... LIMIT 1
```

These run once per distinct seller in the buyer's library (`WithFiltering#seller_post_passes_filters` memoizes per `(seller, filter, email)` key so it can't dedupe across sellers, and `Installment.missed_for_purchase` issues one per purchase's seller). The worst post-deploy Sentry sample showed 24 sellers × ~115ms ≈ 2.8s of an 8.6s request. No existing index covers `(email, seller_id)` — MySQL reads every purchase row for the email (or every row for the seller) and post-filters.

Fixes #6185

## Before/After

EXPLAIN on the probe query shape against a seeded scratch MySQL database (buyer with purchases across 24 sellers + 20K noise rows), before and after applying this migration's exact DDL:

explain-before-after

- **Before:** optimizer picks `index_purchases_on_seller_id_and_chargeback_date`, examines 934 rows, `filtered 5.38%`
- **After:** picks `index_purchases_on_email_and_seller_id` (`ref=const,const`, key_len 776), examines 100 rows (only that buyer's rows for that one seller), `filtered 50%`

On production scale the gap is far larger: the "before" plan scans the full per-email (or per-seller) row set per probe.

## QA steps

No UI surface — this is a pure DB index migration; no application code changes. Reviewer verification:

1. Confirm the migration DDL matches the schema.rb entry.
2. After deploy, run `EXPLAIN` on the probe query shape in the prod console and confirm `index_purchases_on_email_and_seller_id` is chosen.
3. Watch `Api::Mobile::PurchasesController#search` p95/p99 in Sentry and confirm GUMROAD-8T decays.

## Test Results

- `ruby -c` syntax check: OK
- `rubocop` on the migration: 1 file inspected, no offenses detected
- Migration DDL exercised verbatim against a scratch MySQL 8 database (see EXPLAIN evidence above); scratch DB dropped afterwards
- No specs touched: index-only migration, no behavior change

Follow-up (tracked on #6185): batch the per-seller probes in `seller_post_passes_filters` into one grouped query per request.

---

AI disclosure: Authored by Claude Fable 5 (Gumclaw agent), triggered by Sahil's "Fix." comment on #6185. The agent wrote the migration and schema change, verified EXPLAIN plans before/after on a seeded scratch database, and opened this PR.


gumclaw:
Self-review:

- Migration mirrors #6123's `add_email_and_link_id_index_to_purchases` exactly in shape: `disable_ddl_transaction!` (required for online index creation on a table this size), 191-char email prefix (the column is TEXT under utf8mb4, so a prefix is required and 191 keeps the key within InnoDB's 767-byte limit — consistent with `index_purchases_on_email_long` and `index_purchases_on_email_and_link_id`).
- Version `20261206000006` follows the latest migration on main (`...000005`); `db/schema.rb` version bumped to match and the index line inserted in the correct alphabetical position within the purchases table block.
- No application code changes — `seller_post_passes_filters` and `Installment.missed_for_purchase` already issue the exact `email = ? AND seller_id = ?` shape this index serves; the optimizer picks it up without code changes (verified via EXPLAIN in the PR body).
- Write-amplification cost: one additional secondary index on `purchases` inserts. Acceptable — the table already carries several composite indexes with the same email prefix, and the read-side win (24× ~115ms probes → ~1ms) is the #1 mobile library search pain point.
- The batching follow-up (single grouped query across sellers in `seller_post_passes_filters`) is deliberately NOT in this PR — index first, measure, then batch; tracked on #6185.

Merge note: schema migration, so leaving the merge to a human per standing policy.

slavingia:
Get green

gumclaw:
The red CI was a duplicate migration number, not this PR's change: the branch was based on a main commit that still had `20261206000005_add_covering_index_for_affiliate_earnings_to_purchases` (#6157). Main has since reverted that migration (#6194) and separately added `20261206000005_create_tax_remittances`, so CI's merge of this branch with main saw two migrations numbered `20261206000005` and `db:prepare` / `db:schema:load` aborted.

Merged latest main into the branch (5e427c2), which removes the reverted migration and its schema entry. This PR's migration stays at `20261206000006`. CI re-running now; will confirm when green.

gumclaw:
CI is green at 5e427c2 — all test shards, Minitest, Build images, and lint pass. Schema migration, so leaving the merge to a human.

ershad:
Adding an index on `purchases` is a heavy and time consuming operation, it should be the last resort after exploring all options.

gumclaw:
Fair — converting this to a draft while we exhaust app-level options first.

Context on what's already been tried on this endpoint: #6123 (indexed post-filter probes) and #6158 (slim-row profile-post filtering) both deployed on 2026-07-22, and p95/p99 barely moved (1,354ms / 2,897ms post-deploy vs 1,296ms / 2,864ms before). The remaining cost is the per-seller fan-out: 24× `WHERE email = ? AND seller_id = ?` existence probes ≈ 2.8s of the 8.6s worst sample.

App-level options, in the order I'd try them:

1. **Batch the probes in Ruby — biggest win, no DDL.** All probes in a request share the buyer's email and differ only by seller + targeting criteria. `Purchase.product_installments` (app/models/purchase.rb, where `seller_post_filter_cache` is already built) can prefetch the buyer's rows once — `Purchase.where(email:, seller_id: <batch's sellers>)` with the same `not_is_archived_original_subscription_purchase` / `not_subscription_or_original_purchase` scopes, plus one `base_variants_purchases` pluck for variant targeting — and `WithFiltering#seller_post_passes_filters` (app/models/concerns/with_filtering.rb:183) evaluates the exclusion match against the prefetched rows instead of issuing SQL. 24 queries → 2 per request. The two things to get right: email comparison must mirror MySQL's case-insensitive collation (compare downcased), and the SQL path stays as fallback for the non-batched callers (`User#posts`, `CustomerPresenter`).

2. **Index hint on the existing email index.** The optimizer currently picks `index_purchases_on_seller_id_and_chargeback_date` and scans the seller's row set per probe; `FORCE INDEX (index_purchases_on_email_long)` flips each probe to scanning only the buyer's rows (usually tens). Small diff, but it couples code to an index name and still leaves 24 round-trips.

3. **Same batching for `Installment.missed_for_purchase`** (app/models/installment.rb:135) — the other ~1.4s of per-seller fan-out, though that only runs on the customer presenter path.

Plan: I'll build option 1 with spec coverage and link the PR here, we re-measure GUMROAD-8T and the endpoint's p99 after it deploys, and this index only comes back if the batched version still can't hold. Shout if you'd rather go a different route.

gumclaw:
Removing `awaiting-human` — ershad's verdict received (index is last resort). This stays a draft while app-level options are exhausted under #6185; will only revive if those fail.

gumclaw:
Alternative is up: #6206 batches the per-seller probes in Ruby (24 queries → 2 per request, no DDL). This PR stays draft; if the batching holds the endpoint's p99 post-deploy, I'll close this one.

gumclaw:
Reviving this per Sahil's note on #6185 — the index helps generally beyond this endpoint, so it shouldn't wait on #6206's post-deploy numbers.

Updates since it was parked:
- Merged latest main and renumbered the migration to `20261206000007` (main took `...000006` for the dispute tax-period indexes), so there's no duplicate-version conflict.
- The migration is unchanged otherwise: online index creation (`disable_ddl_transaction!`), 191-char email prefix mirroring the existing email indexes. EXPLAIN before/after is in the PR body.

This is complementary to #6206: the batching removes the per-seller loop on mobile library search, and this index makes any `email + seller_id` lookup fast everywhere else (e.g. `Installment.missed_for_purchase`, ad-hoc support/admin queries on a buyer's history with a seller).

@ershad flagging for your review since you called the index a last resort — Sahil's take on #6185 is that it's generally useful, so leaving the timing/merge call with you. Schema migration, so no auto-merge.

gumclaw:
CI is green at the current head (`71597ac`) — Buildkite and ci/green both passing, merge state CLEAN, no conflicts. This is a schema change (DDL migration), so merge is yours, @ershad. Context: revived per Sahil's note on #6185 since the index helps beyond the mobile library endpoint; #6206's app-level batching shipped separately.

gumclaw:
@ershad automated review-SLA nudge: this PR has been waiting 34h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

gumclaw:
## Production migration risk assessment (gumclaw)

Asked to check deploy time / downtime before this ships. Short answer: **it will not lock users out, but it will stall the deploy pipeline for roughly an hour, it can fail after doing all that work, and there is a prior red flag that needs resolving first.**

All numbers below are live prod reads via the audited console (requests `20260725T011905Z-10a9743f`, `20260725T011956Z-b4ea9c5e`, `20260725T012111Z-019ef81c`).

### The table

| | |
|---|---|
| Rows | **327,501,253** |
| Data | 178 GB |
| Existing indexes | 282 GB (27 of them) |
| Total | **460 GB** |
| InnoDB buffer pool | 46 GB |
| MySQL | 8.0.42 |

Average email length is 22 chars, so the new index is roughly 45 bytes/row → **~15 GB** of new index to build and to carry forever on a table that already has more index than data.

### 1. No user-facing downtime, but the deploy blocks

`ADD INDEX` on a prefixed TEXT column is `ALGORITHM=INPLACE, LOCK=NONE` in MySQL 8, so reads and writes to `purchases` continue throughout. Checkout will not go down.

The problem is the pipeline. From `nomad/common.sh`:

```
gr_deploy() {
  check_for_deployment_lock
  lock_deployment
  run_job database_migration
  wait_for_db_migrate      # while true; do ... sleep 2; done
  run_job rpush ... sidekiq_worker ... web servers
```

`wait_for_db_migrate` is an **unbounded** poll loop, and it runs *after* `lock_deployment`. So for the entire index build:

- no application code deploys (web/sidekiq jobs come after the wait),
- **production deploys are locked for everyone else**, since the lock is taken before the migration starts.

Estimated build time on a 327 GB scan with a 46 GB buffer pool: **~30 minutes to 2 hours**, most likely around an hour. I can't measure this precisely without running it — but the order of magnitude is what matters, and it's "the deploy hangs for an hour", not "a few seconds".

### 2. Real failure risk: the online-alter log is at its 128 MB default

`innodb_online_alter_log_max_size = 134217728` (128 MB).

During an INPLACE build, every concurrent write to `purchases` is buffered in that log. `purchases` is one of the hottest write tables we have. If writes during the build exceed 128 MB, MySQL aborts with `Creating index ... required more than innodb_online_alter_log_max_size bytes of modification log` — **after an hour of work, and it rolls back**. Then the deploy is still holding the lock and still waiting on a loop that will never see the KV key.

That is the scenario that turns "slow deploy" into "incident during a Friday-night payout window".

### 3. ⚠️ The precedent index is not in production

This is the part I'd resolve before merging. The migration comment says it mirrors `index_purchases_on_email_and_link_id` from #6123/#6158 (merged 2026-07-22). But:

- `schema_migrations` **does** contain `20261206000004` (recorded as run), and
- `information_schema.statistics` shows the only email index on `purchases` is `index_purchases_on_email_long`. `index_purchases_on_email_and_link_id` is **absent**.

So a merged, recorded index migration on this exact table did not leave an index behind. Caveat on my evidence: the console I can reach is a **replica** (`read_only=1`, host `ip-172-17-3-197`), so I cannot rule out that the index exists on the writer and this replica diverges. But DDL replicates, so either way something needs explaining — and if migration 004 silently no-op'd or failed while being marked complete, **#6195 will do exactly the same thing** and the perf win won't materialize. Worth confirming on the writer before shipping either.

That also means the #6123/#6158 latency fix may not actually be live, which changes the baseline this PR's 115ms-per-probe measurement was taken against.

### Recommendation

Don't run this inside the deploy pipeline as written.

1. **Confirm on the writer** whether `index_purchases_on_email_and_link_id` exists, and why 004 is recorded without it. This gates everything else.
2. **Build the index out of band**, during a low-write window, not while holding the deploy lock — either a manual `ALTER` with `innodb_online_alter_log_max_size` raised for the session (a few GB), or gh-ost/pt-osc. Neither tool is currently in the repo, so manual is the realistic path.
3. **Make the migration idempotent** so the pipeline no-ops once the index exists:
   ```ruby
   add_index :purchases, [:email, :seller_id],
             name: "index_purchases_on_email_and_seller_id",
             length: { email: 191 },
             if_not_exists: true
   ```
   Without `if_not_exists`, a pre-built index makes `db:migrate` fail and the deploy hangs on the same unbounded loop.
4. **Not during a payout window.** We just spent Friday recovering the PayPal batch (gumroad-private#1284) and deploys were locked for hours; an hour-long migration holding the same lock on payout day is avoidable self-harm.

Also worth a second opinion: this adds a 28th index / ~15 GB to a table whose indexes already outweigh its data, and `(email, seller_id)` is a prefix-overlap with the existing `(email)` index. Fine if the probe win is real — but see point 1, because right now the comparable prior index isn't in place to prove it.

🤖 Assessment by gumclaw.

gumclaw:
Merged latest main in to clear the conflict.

Only one file actually conflicted: `db/schema.rb`, and only on the `define(version:)` line. Main had moved the schema version up to `20261206000009` (the taggings_count counter cache work), while this branch had it at `20261206000007` for the new purchases index. Kept main's higher version, since the schema version just tracks the newest migration applied and both sets of changes are present in the merged file — the `["email", "seller_id"]` index from this branch and the `taggings_count" column plus its index from main.

Worth noting the earlier renumber on this branch is still valid: main claimed 000008 and 000009 but never took 000007, so this migration's number is unclaimed and no second renumber was needed.

Verified the DDL applies cleanly against a real MySQL — the index lands as `email(191), seller_id`, exactly what schema.rb declares. Rubocop is clean on the migration (the one schema.rb offense is the pre-existing missing frozen-string comment, identical on main). `seller_post_probe_batch_spec`, the spec covering the probe shape this index is built for, passes 13/13. `purchase_installments_spec` shows failures but they're a local environment issue with the payment-processor stubs, not this merge — pristine main fails the same 41 of 53 examples.

gumclaw:
@ershad automated review-SLA nudge: this PR has been waiting 58h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

gumclaw:
## Correcting my own framing here — this was never a silent review, and the ground has moved

I've been treating this PR as "waiting on @ershad" and firing automated review-SLA nudges at it (34h, 58h). That was wrong, and the nudges were not warranted. Ershad **did** review this, on 2026-07-23, within 75 minutes of being assigned:

> Adding an index on `purchases` is a heavy and time consuming operation, it should be the last resort after exploring all options.

That's a clear verdict, not silence. I converted this to draft, built the app-level alternative, and then **revived this PR anyway** on the strength of Sahil's "I feel like this index would also help generally." Reviving over a reviewer's stated objection is fine — but I then let an SLA bot bill him for the delay, which is not.

### What's changed since his verdict

The alternative he asked for shipped. #6206 (batch the per-seller probes, no DDL) **merged 2026-07-24T00:59Z** and closed the originating issue #6185. So the specific problem this index was justified against — 24 per-seller `email = ? AND seller_id = ?` probes on mobile library search — **no longer exists in the code**. That loop is gone.

### Fresh production numbers (read-only replica, this run)

Every email-prefixed index on `purchases` today:

```
index_purchases_on_email_long    email(191)
```

`EXPLAIN` for both query shapes, against the current schema:

| shape | index chosen | rows | note |
|---|---|---|---|
| `email = ? AND seller_id = ?` (single probe) | `index_purchases_on_email_long` | 1 | `Using where` for the seller filter |
| `email = ? AND seller_id IN (...)` (**what #6206 ships**) | `index_purchases_on_email_long` | 1 | same |

Timed against a genuinely heavy buyer (1,748 lifetime purchases across 9 sellers):

| | time |
|---|---|
| Batched prefetch — the code that runs today | **500ms** |
| Per-seller loop — the pre-#6206 shape | 718ms |

The composite index would let the seller filter be satisfied *in* the index instead of by `Using where`. That is a real but **modest** win on the shape that actually ships, not the 300x cliff from the original benchmark — because that cliff was the cold-cache 24-probe loop, and #6206 already removed the loop.

### So the actual decision, which is yours

Ershad, you were right on the merits and I should have left it there. The question is no longer "is this index needed for mobile library search" — it isn't, that's fixed. It's the much weaker claim: **is `email(191), seller_id` worth ~15GB and a ~1h deploy-pipeline stall as a general-purpose index**, on a table that already carries 282GB of index against 178GB of data?

My honest read is **no, not on this evidence** — "helps generally" isn't a measured claim, and I don't have a caller that needs it now that #6206 shipped. Unless you know of one, I think the right outcome is to **close this**.

Three options, pick one and I'll action it same-day:

**(a) Close it.** My recommendation. Reopen if a real workload turns up that needs it.
**(b) Keep it parked** as a draft against a future need, with the production risk assessment (01:22Z 07-25) attached so it isn't re-litigated from scratch.
**(c) You still want it merged** — then it's a scheduling question, not a review question, and I'll coordinate the migration window.

No SLA clock on this one and no more automated nudges; I've corrected that on my side. Never auto-merging a DDL change regardless.

gumclaw:
Heads up on timing: Sahil is planning a merge pass tonight, after the Monday meeting, over the PRs that are 3+ days old — this is one of them.

Where it stands right now: all checks green at the current head, no unresolved review threads, no conflicts with main. It's assigned to @ershad because it adds an index to `purchases`, which is a large table and you wanted eyes on the migration itself rather than just the query change.

I'm not merging it. If you want to review the migration before tonight, that's the window; if you'd rather it wait, say so here and I'll flag it as held back.

slavingia:
Confirm that mobile library searches have no slow queries remaining using production data and close.

slavingia:
> Adding an index on purchases is a heavy and time consuming operation, it should be the last resort after exploring all options.

Agree but also shouldn't be an excuse to do nothing about performance gains. We should be able to handle these imo, esp with gumclaw

gumclaw:
Confirmed against production, and the answer has two halves: the queries this index targets are no longer slow, but mobile library search does still have slow queries — they moved to a different table.

## The endpoint's latency, before and after the app-level fix

#6206 (batch the per-seller probes in Ruby) merged 2026-07-24T00:59Z and first shipped in release `v2026.07.24.4` at 02:06Z, so everything after that is post-deploy. Sentry transaction stats for `Api::Mobile::PurchasesController#search`:

| window | count | p50 | p75 | p95 | p99 | max |
|---|---|---|---|---|---|---|
| 07-22 12:00 → 07-24 02:00 (before) | 4,210 | 135 ms | 320 ms | 1,362 ms | 2,847 ms | 9,467 ms |
| 07-24 04:00 → 07-27 07:00 (after) | 7,215 | 124 ms | 290 ms | 1,268 ms | 2,784 ms | 20,517 ms |

Marginal movement at every percentile, and the worst single request got worse. So the endpoint is not fixed — but the reason is no longer the thing this PR indexes.

## Where the time actually goes now

I pulled the span trees of the seven slowest post-deploy samples (85.7 seconds of wall time between them) and grouped every SQL span by query shape:

| query shape | time | share |
|---|---|---|
| slim `profile_only_for_sellers` scan on `installments` | 31,720 ms | **37.0%** |
| `installments` INNER JOIN `email_infos` | 23,153 ms | **27.0%** |
| `purchases` WHERE email = ? AND seller_id = ? | 7,572 ms | 8.8% |
| `url_redirects` lookups | 6,468 ms | 7.5% |

The `email + seller_id` probes this index was built for are down to 8.8% of the tail, and production `EXPLAIN` resolves that shape at `rows=1` on the existing `index_purchases_on_email_long`. Two-thirds of the remaining time is in `installments`, a table this PR does not touch.

The dominant one has no usable index at all:

```
SELECT installments.id, ..., installments.json_data FROM installments
WHERE deleted_at IS NULL AND published_at IS NOT NULL
  AND (flags & 256 = 0) AND (flags & 128 = 128)
  AND installment_type = 'seller' AND seller_id = ?
```

```
type=ref  key=index_installments_on_seller_id_and_link_id  key_len=5  ref=const
rows=1392  filtered=0.90  Extra=Using where
```

`key_len=5` — only the `seller_id` prefix is used; the type, flag and timestamp predicates are all `Using where`. It reads ~1,392 rows per prolific seller to keep about twelve, dragging `json_data` along, and runs 8–24 times per request with individual executions up to 4,690 ms. (Audited read-only replica, request `20260727T073857Z-45336c54`.)

## Closing

Ershad's original verdict was right on the merits — an index on `purchases` should be the last resort — and the app-level alternative he asked for shipped. The workload that justified this index no longer exists in the code, and the measured share of what's left does not justify ~15 GB of new index and an hour-long deploy-pipeline stall on a table already carrying 282 GB of index against 178 GB of data.

Closing. The real remaining work is tracked with the full span and EXPLAIN evidence in gumroad-private#1412 — a composite index on `installments (seller_id, installment_type, published_at)` plus the slim-column treatment #6158 applied to the sibling path. `installments` is 24.1M rows and 3.2 GB of index, so that one is cheap in a way this one wasn't.

gumclaw:
Agreed, and I don't want the close to read as "nothing to do here" — it's the opposite. The measurement is what says this particular index isn't the lever, not that the endpoint is fine. It isn't: p99 is still 2.8 s and the worst post-deploy sample was 20.5 s.

The performance work carries on in antiwork/gumroad-private#1412 (https://github.com/antiwork/gumroad-private/issues/1412), where the tail actually lives now — two `installments` query shapes are 64% of the slow-sample time. Two parts:

1. **No DDL, ships first.** `seller_with_sent_emails_for_purchases` selects `installments.*`, which drags each post's LONGTEXT `message` body through a join that discards 95% of the rows it touches. antiwork/gumroad#6158 already did exactly this slim-column treatment on the sibling profile-posts path; applying it here is the same change, no migration. I'm starting on it now.

2. **A composite index on `installments (seller_id, installment_type, published_at)`.** Ershad's "last resort" bar applied to `purchases` because that table is 327M rows and 460 GB — an hour of deploy-pipeline stall for ~15 GB of new index. `installments` is 24.1M rows and 3.2 GB of total index, roughly two orders of magnitude cheaper, and the query currently reads ~1,392 rows to keep about twelve because only the `seller_id` prefix of the existing index is usable. I'll size the migration properly and bring it with numbers rather than assert it's cheap.

The general lesson I'm taking from this PR: I justified an index on "helps generally" and it cost several days of review attention. The follow-up issue leads with the measurement instead.

slavingia:
Sorry that's not linking to the right PR?

gumclaw:
You're right — sorry. `#6158` in that comment is the correct PR in this repo, but `gumroad-private#1412` doesn't resolve as a cross-repo shorthand from here, so GitHub linked it to antiwork/gumroad#1412 instead, which is an unrelated closed membership bug.

The follow-up issue I meant is https://github.com/antiwork/gumroad-private/issues/1412 — "Mobile library search tail is now dominated by two installments queries, not purchases". I've edited the comment above to use the full URL.

gumclaw:
Concrete follow-through on this, since the close shouldn't have been the end of it.

I said the remaining tail was two `installments` query shapes and tracked it in gumroad-private#1412. That was right but incomplete — I had grouped the slow samples by query shape without separating them by request *type*. Going back to the timeouts specifically (`Rack::Timeout`, GUMROAD-8T, still firing as of 03:33Z today), all of them share something the general slow samples don't: they carry `seller[]` params, and they all die in the same query.

```
GET /mobile/purchases/search?items=24&page=1&seller[]=…&seller[]=…&seller[]=…
last query before SIGTERM: "Purchase Count"   db_runtime: 119,995 ms
```

The seller filter nests a subquery, and MySQL plans it seller-first — it walks each matched seller's sales through `index_purchases_on_seller_id_and_chargeback_date` and filters down to the buyer afterwards. A buyer filtering by a large seller pays for a scan proportional to **that seller's whole sales history**, not to their own library, which is why the buyer's own row count looks harmless while the query runs for minutes. `EXPLAIN ANALYZE` on the replica, buyer with 673 purchases filtering by three sellers (311K / 66K / 145 sales):

| | subquery (today) | concrete ids |
|---|---|---|
| pagination `COUNT`, cold | 1,616 ms | 11 ms |
| worst observed | 78,250 ms | 4 ms |
| whole endpoint path, cold | 6,646 ms | 50 ms |
| whole endpoint path, warm | 21 ms | 22–30 ms |

Two lines, no DDL:

```ruby
- purchases.where(seller_id: User.where(external_id: Array.wrap(params[:seller])).select(:id))
+ seller_ids = User.where(external_id: Array.wrap(params[:seller])).pluck(:id)
+ purchases.where(seller_id: seller_ids)
```

**#6391** — https://github.com/antiwork/gumroad/pull/6391. Fable review came back SHIP; the three P3s and their disposition are in a comment there.

So there are now two fixes in flight off this thread and neither one needs an index on `purchases`: #6390 for the `installments` slim-column path (27% of slow-sample time) and #6391 for the timeout shape. The `installments` composite index in gumroad-private#1412 is still worth doing and I'll bring it with sizing numbers, but that table is 24.1M rows against `purchases`' 327M, so it isn't the same class of ask.

Leaving this closed — the index it proposes still isn't the lever. I'll report back on GUMROAD-8T after #6391 ships rather than calling it fixed now.

PR #6205, opened 2026-07-23
Make Elasticsearch + MinIO available to the Minitest CI job (opt-in) (#5801)

ershad:
## What

Makes Elasticsearch and MinIO (S3) available to the `test_minitest` CI job so ES/S3-dependent tests can run in the Minitest suite instead of being stranded in RSpec — while keeping the suite fast.

- **CI (`.github/workflows/tests.yml`)**: adds an `elasticsearch:7.9.3` service and starts `minio/minio` (via a step — the official image needs a `server` command that GitHub Actions service containers can't supply) with the `gumroad-specs`/`gumroad-invoices` buckets, plus readiness waits. Versions match the RSpec docker-compose stack.
- **Harness (`test/test_helper.rb`)**: keeps the fast defaults for all ~2,400 tests — the ES client stays stubbed and ActiveStorage stays on the local Disk service. Adds **opt-in** escape hatches:
  - Real ES: a test sets `Thread.current[:minitest_real_es]` (via `RealElasticsearchBridge`) and the stub transparently forwards to a real client for the duration, restoring the fake in teardown.
  - Real S3: `with_real_s3 { … }` (`test/support/real_storage_helpers.rb`) swaps ActiveStorage to the MinIO-backed `:test` service from `config/storage.yml` for a block.
- **Closes the gap**: the 30 ES-dependent `links_controller` examples (`GET search` + the page-view analytics "data recorded" group) move from the companion RSpec spec (`spec/controllers/links_controller_search_spec.rb`, added in #6145) **into** `test/controllers/links_controller_test.rb`, which now runs the full **343** examples. The companion spec is deleted, so `links_controller` is a clean 1:1 Minitest port with nothing left in RSpec.
- **Smoke test** (`test/storage/minio_availability_test.rb`): proves the MinIO wiring + `with_real_s3` round-trip end-to-end, and asserts the default service stays Disk — so the opt-in path can't silently rot.

## Why

Before this, the lean Minitest job ran only mysql/redis/stripe_mock/mongo, so tests that index into ES and query it back (or need real S3 semantics) couldn't run there — they had to stay in the RSpec suite (which brings up the full docker-compose stack). That's why #6145 split `links_controller` into a companion RSpec spec.

The **opt-in** design is deliberate: making every test use real ES/S3 (like RSpec) would reindex on every product save and round-trip S3 on every attach, dragging the Minitest job back toward RSpec speeds — against the whole point of the migration. Instead only the handful of tests that genuinely need it pay the cost; everything else keeps the seconds-level speed.

## Before / After

Not applicable — test/CI-infra change, no app code touched, no UI surface.

## Test Results

```
$ OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES CI=true RAILS_ENV=test bin/rails test test/controllers/links_controller_test.rb
343 runs, 1226 assertions, 0 failures, 0 errors, 0 skips     (313 stub + 30 real-ES; was 313 + 30-in-RSpec)

$ CI=true RAILS_ENV=test bin/rails test test/storage/minio_availability_test.rb
2 runs, 4 assertions, 0 failures, 0 errors, 0 skips           (real-S3 round-trip via MinIO + Disk-is-default)

$ OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES CI=true RAILS_ENV=test bin/rails test
2471 runs, 6528 assertions, 0 failures, 0 errors, 6 skips      (full test/ tree — no leakage; +30 ES tests, +2 MinIO smoke vs main)
```

(The `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` prefix is only needed for the ES/geo tests on local macOS — a fork-safety quirk; not needed on CI's Linux.)

## QA steps

1. `CI=true RAILS_ENV=test bin/rails test test/controllers/links_controller_test.rb` → 343, 0 failures (needs Elasticsearch running).
2. `CI=true RAILS_ENV=test bin/rails test test/storage/minio_availability_test.rb` → 2, 0 failures (needs MinIO running).
3. `CI=true RAILS_ENV=test bin/rails test` (whole `test/` tree) → 0 failures, 0 errors.

---

## AI disclosure

Written by Claude Opus 4.8, following up on the #5801 migration to remove the ES/S3 gap that forced `links_controller`'s search/analytics tests into a companion RSpec spec.


gumclaw:
## The red was never this PR — it was broken main, already fixed

Diagnosed the `ci/green` failure. Only **one** job failed (`Test Fast 5`); every other job, **including `Test Minitest` — the job this PR actually changes — passed.**

The failure was in `spec/sidekiq/create_us_states_sales_summary_report_job_spec.rb:373`, chargeback period attribution:

```
expected: [["State", "GMV", "Number of orders", "Sales tax collected"], ["Washington", "-110.10", "1", "-10.10"]]
     got: [["State", "GMV", "Number of orders", "Sales tax collected"], ["Washington", "110.10", "1", "10.10"]]
```

Nothing to do with Elasticsearch, MinIO, or the Minitest port — it's the US-states sales-tax report.

**Root cause:** that spec was broken on `main` and fixed by dfe82fa06 (*"Fix main CI: create matching Dispute rows in the US-states summary chargeback specs"*, #6212). Verified the fix is an ancestor of `origin/main` but **was not** in this branch. The CI run was from **2026-07-23**, and the branch was **78 commits behind main**, so it never saw the fix.

**Action:** merged current `main` into the branch — clean, no conflicts. Now at `b568e685`, 0 commits behind. CI is re-running and should go green without touching a line of this PR's own work.

Worth noting for anyone reading the old red: a single failing shard on a stale branch is usually this, not the diff. Checking whether the failing spec was touched on main after the run date is the fastest way to tell.

---

Reviewed and unblocked by Claude Opus 5 running as the Gumclaw Hermes agent. No changes to @ershad's implementation — only a merge of main.

gumclaw:
## Fable 5 pre-merge review: SHIP (0 P1, 1 P2, 5 P3) — P2 + 3 P3s fixed in 8100ec61

CI is green at `b568e685`. Ran the adversarial review on top of that, focused on the thing a test-migration PR can silently get wrong: **lost coverage**, since the rspec file is deleted and nothing fails if an assertion didn't make it across.

### Coverage parity: 30/30, clean

Every example in the deleted `links_controller_search_spec.rb` maps to a Minitest test with **assertion strength preserved** — including the ordering/ranking checks and the full-response-body equality assertions, which are the usual casualties of a port. **Nothing skips**, so no test silently no-ops on CI. This was the main risk and it came back clear.

### What else the review verified rather than assumed

- **ES and MinIO versions match the rspec lane exactly** (`elasticsearch:7.9.3`, `minio/minio:RELEASE.2025-09-07T16-13-09Z` in both the workflow and `docker-compose-test-and-ci.yml`), so a test can't pass in one lane and fail in the other on version skew.
- **Both readiness loops fail loudly.** Each polling loop is followed by a final `curl -fsS … || exit 1`, so a never-ready ES or MinIO fails at the wait step rather than confusingly later in the tests. Plain `_cluster/health` without `wait_for_status` is fine here — a fresh single-node cluster with no indices reports green as soon as it answers HTTP.
- **The `Thread.current[:minitest_real_es]` bridge doesn't leak.** Set/cleared with correct `teardown` semantics, which run even when a test raises, in a verifiably single-process suite.
- **No flaky-search trap.** Every index→query sequence goes through `import(refresh: true)` plus an explicit `refresh_index!`, so no query races ES's near-real-time indexing.
- **`with_real_s3` swap/restore is correct** — both `services` and `service` restored in `ensure`, mirroring the pattern `test_helper.rb` already uses.

### Fixed in 8100ec61

- **P2 — `minio/mc:latest` was unpinned.** The MinIO *server* is pinned, but the bucket-creation client wasn't, while the compose stack pins it to `RELEASE.2025-08-13T08-35-41Z`. An unpinned `:latest` can break this job with no code change. Now pinned to match.
- **P3 — CI bucket provisioning diverged from the compose stack.** The workflow only ran `mc mb`; compose also sets `anonymous set public` and `version enable`. Since the `:test` storage service is declared `public: true`, the first future test fetching `blob.url` over plain HTTP would have **403'd on CI while passing locally**. Now mirrored.
- **P3 — restored one-review parity in the discover-filter test.** The rspec original gave the product a single review via `:recommendable`, and that was the assertion's point: one review clears the filter. The port added a second, which passes today but would stop the test failing if the threshold ever regressed from one to two.
- **P3 — documented two real limits of the ES bridge.** Only models passed to `install_real_elasticsearch!` get namespaced indices; an incidental write from any other model lands in the real cluster under its *default* index name — and locally `ELASTICSEARCH_HOST` is the same `localhost:9200` as `.env.development`, so that would quietly write test documents into a developer's dev indices. Also noted the namespace assumes a single process and needs revisiting before Rails `parallelize` is enabled, and that blobs created inside `with_real_s3` must not outlive the block.

Left as-is: the latent un-namespaced-model write has no current trigger (I traced the paths), so a warning comment beats adding enforcement machinery to a test helper.

`rubocop`: 2 files, no offenses. Workflow YAML parses. CI re-running to validate the changed MinIO step.

@ershad this is your PR and still a draft — I only merged main (the red was #6212's fix missing from a 78-commit-stale branch, not your code) and applied the review's findings. Nothing about the Elasticsearch/MinIO design or the port itself changed. Yours to un-draft when you're happy.

---

Reviewed by Claude Fable 5 (`--model claude-fable-5`) running as the Gumclaw Hermes agent, dispatched headless with static analysis only. Findings triaged and fixed by me.

gumclaw:
Merged main in to clear the conflict — nothing about this PR's behaviour changed.

Only one file collided: `test/controllers/links_controller_test.rb`. Both sides appended a new test class to the end of the same file, so git couldn't tell where one ended and the next began. This branch adds `RealElasticsearchBridge` plus the search/page-view classes that use it; main added `LinksControllerDeletionAuditTest`. I kept both, in that order, and closed the search class off properly before main's audit class starts.

Full Minitest run on that file is green locally: 409 runs, 1490 assertions, no failures. Rubocop clean on the three files this branch touches.

slavingia:
How's this looking?

ershad:
@slavingia, we should be able to merge this once the CI is green, and we can continue with the migration to Minitest.

PR #6210, opened 2026-07-23
Backfill payment method type for UPI purchases made before #6201

gumclaw:
## What

Adds `Onetime::BackfillStripePaymentMethodTypes`, a one-time task that corrects the recorded payment method for successful Stripe purchases paid with a local (non-card) method — UPI today — that completed before #6201 reached production.

For each candidate it re-fetches the Stripe charge, reads Stripe's own classification of the method (`payment_method_details.type`, exposed as `StripeCharge#payment_method_type`), and corrects both rows the same way `Purchase#save_charge_data` now does at charge time:

- `purchases.card_type` → `"upi"` (via `update_column`; card_type feeds no callbacks or search index)
- `purchase_payment_flows.stripe_payment_method_type` → the raw method type (`"upi"`)

## Why

Follow-up to #6201 (Razvan's review comment on that PR). UPI launched via #6174 the morning of 2026-07-23; #6201 merged the same afternoon. Any successful UPI purchase in that window was recorded with:

- `card_type` **nil** (server-confirm lane) or **"generic_card"** (client-confirm lane)
- a hardcoded `"card"` placeholder in `purchase_payment_flows.stripe_payment_method_type`

Those rows stay wrong forever without a backfill, undercounting UPI in every payment-method metric and mislabeling those flow rows in rollout monitoring. Stripe's charge object carries the true method type, so a one-time API walk fixes them authoritatively.

## How it works

- **Candidates**: successful Stripe purchases in the UPI-launch window (2026-07-23 → run time, configurable) whose `card_type` is nil or `"generic_card"` — the only two shapes the bug produced. Correctly recorded card purchases are never selected.
- **Write gate**: a row is only touched when Stripe reports a method that maps to a known `CardType` other than the unknown bucket. Candidates Stripe classifies as plain `"card"` (nil card_type has other historical causes) or as an unrecognized method are counted and left alone — the same no-`generic_card`-leak rule as the charge-time fix.
- **Combined charges** are fetched once per Stripe transaction id (purchases in a multi-product order share one charge).
- **Idempotent and retryable**: corrected purchases fall out of the candidate scope; a re-run is a no-op. Both writes happen in one transaction, so a partial failure rolls back fully and the purchase stays a candidate for the next run.
- **Dry-run by default**; per-purchase errors are logged and counted without stopping the run. `ReplicaLagWatcher` throttles the loop.

Run it after this deploys:

```ruby
Onetime::BackfillStripePaymentMethodTypes.process                 # dry run, logs stats
Onetime::BackfillStripePaymentMethodTypes.process(dry_run: false) # writes
```

## QA steps

No UI surface — this is a console-run one-time data task. Verification path:

1. Deploy, run the dry run in the prod console and eyeball the stats (candidate counts should match the known pre-fix UPI purchase count).
2. Run with `dry_run: false`.
3. Spot-check: `Purchase.where(card_type: "upi").where("created_at < ?", <#6201 deploy time>)` rows now exist and each one's `purchase_payment_flow.stripe_payment_method_type == "upi"`.

## Test results

Full new spec file (12 examples) plus the payment-flow specs touched by #6201:

```
bundle exec rspec spec/services/onetime/backfill_stripe_payment_method_types_spec.rb \
  spec/models/purchase_payment_flow_spec.rb
39 examples, 0 failures
```

test results

Rubocop: both files pass, no offenses.

Covered: nil-card_type and generic_card shapes, plain-card and unrecognized-method skips, missing method type, missing flow row, dry-run, idempotency, one fetch per shared combined charge, error isolation, and candidate-scope boundaries (window, processor, recorded card brands).

---

## AI disclosure

Authored by Claude Fable 5 via the Hermes agent (gumclaw), from Razvan's review comment on #6201 requesting a one-time backfill for pre-deploy UPI purchases and their payment-flow rows using Stripe's recorded payment method type.


app/services/onetime/backfill_stripe_payment_method_types.rb
```diff
@@ -0,0 +1,130 @@
...
+        return tick(:skipped_no_method_type) if method_type.blank?
+
+        mapped_card_type = StripeCardType.to_new_card_type(method_type)
+        # Only known local methods get written. A candidate Stripe classifies as "card"
+        # was nil/generic for some other historical reason, and an unrecognized method
+        # must not be collapsed into "generic_card" (same rule as the charge-time fix).
+        return tick(:skipped_not_local_method) if mapped_card_type == CardType::UNKNOWN
+
+        fix_card_type(purchase, mapped_card_type)
+        fix_payment_flow(purchase, method_type)
```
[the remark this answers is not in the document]
gumclaw: Fixed in 769414b — real finding. Both writes now run inside one `ApplicationRecord.transaction`, so a failure updating the payment-flow row rolls back the `card_type` write and the purchase stays in the candidate scope for the next run. Added a spec that forces the flow update to fail, asserts both rows roll back, and verifies a rerun repairs them. All 12 service specs plus the 27 payment-flow specs pass locally.

rmarescu:
@gumclaw only assign PRs when all work is complete, with green CI.

gumclaw:
Understood — I'll only request review once CI is green at the current head.

For the record here: the red shard was main breakage (`create_us_states_sales_summary_report_job_spec.rb`, unrelated to this diff), fixed on main by #6212. I've merged main into this branch and pulled your review request; it will come back automatically when CI is fully green.

gumclaw:
CI status: the only red job (`Test Fast 7`) fails on `spec/sidekiq/create_us_states_sales_summary_report_job_spec.rb:373` (chargeback period attribution), which is unrelated to this PR's diff (backfill service + spec + qa-media only). The same spec fails on main's own Tests run at `0a11b93b` ([run 30033333182](https://github.com/antiwork/gumroad/actions/runs/30033333182)) — pre-existing on main, not introduced here. It failed identically on a rerun, so it's deterministic on the current tree rather than a flake.

Greptile's P1 (retryability of the two writes) is fixed in 769414b — both writes now run in one transaction, with a rollback + rerun spec. Full backfill spec file passes locally: 12 examples, 0 failures. Rubocop clean.

app/services/onetime/backfill_stripe_payment_method_types.rb
```diff
@@ -0,0 +1,137 @@
...
+          @charge_cache[purchase.stripe_transaction_id] = processor_charge&.payment_method_type
+        end
+      end
+
+      def fix_card_type(purchase, mapped_card_type)
+        return tick(:card_type_already_set) if purchase.card_type == mapped_card_type
+        return tick(:would_fix_card_type) if @dry_run
+
+        purchase.update_column(:card_type, mapped_card_type)
+        tick(:fixed_card_type)
```
rmarescu: This increments the in-memory stats before the surrounding transaction commits. If `fix_payment_flow` raises, the database rolls back `card_type` but the final stats still report it as fixed. Consider counting successful writes only after the transaction commits.

rmarescu (approved):

app/services/onetime/backfill_stripe_payment_method_types.rb
gumclaw: Fixed in 24d04b5. The fix methods now return their outcome key instead of incrementing directly, and `backfill` counts the outcomes only after the transaction block exits — so a rollback means nothing gets counted and the error path records it instead. Extended the rollback spec to assert `fixed_card_type` and `fixed_flow` stay at 0 when the flow update fails.

Note on local test run: the spec file currently fails at the `create(:purchase)` factory in this worktree ("We couldn't charge your card") on the unmodified branch head too — pre-existing environment issue, not this change. Rubocop clean; CI will verify.

PR #6213, opened 2026-07-23
Remove the inventory_counter_cache rollout flag (100% on since 2026-04)

gumclaw:
Removes the `inventory_counter_cache` Flipper flag. Part of the rollout-flag audit antiwork/gumroad-private#1208 (Category B — Sahil, 2026-07-23: remove any B flag at 100% in prod for >30 days).

## Why this is safe

- Prod state: `state=on, boolean=true` (globally enabled, audited prod console read `20260721T025828Z-d0d55a50`), stable since 2026-04.
- The flag selected between the counter-cache column (`sales_count_for_inventory_cache`, kept in sync by `Purchase`/`Subscription` callbacks) and a live `SUM(quantity)` query. With the flag 100% on, production has read the cached column exclusively for ~3.5 months — removal is behavior-neutral; the live-SUM fallback is dead code.

## Changes

- `Link#sales_count_for_inventory` and `BaseVariant#sales_count_for_inventory` now read the cache column unconditionally, with a comment recording the flag history.
- Spec sweep (both suites) for tests that silently relied on the live-SUM path:
  - `.reload` added where in-memory instances held a stale cache column (the sync callbacks `update_all` the DB, not the loaded object): `variant_spec`, `recommendations_spec`, `link_test.rb` (minitest).
  - Purchase setups reshaped to attach variants at creation (`create(:purchase, ..., variant_attributes: [variant])`) so the variant-side cache increment fires, matching the real checkout flow.
  - `inventory_counter_cache_spec` flag-gating contexts converted to pin the now-unconditional cache-backed reader.

## Before / After

No adjacent UI surface — backend reader consolidation only; the spec run is the reviewable artifact:

spec run

`120 rspec examples, 0 failures` across the three touched spec files; `7 minitest runs, 0 failures` for the `remaining_for_sale_count` block; rubocop clean on all touched files.

Post-deploy housekeeping (deploy-gated, tracked on gp#1208): `Flipper.remove(:inventory_counter_cache)` via the audited prod console.

---

## AI disclosure

Prompts/instructions given: standing order (Sahil 2026-07-19) to autonomously build qualifying issues; gumroad-private#1208 rollout-flag audit used as the spec, with Sahil's 2026-07-23 directive extending removal to Category B flags at 100% for >30 days.


slavingia:
Get green and merge

PR #6216, opened 2026-07-23
Populate errors on every refund failure branch so the dashboard never shows a blank error toast

gumclaw:
## What

Every `return false` / early-`nil` branch in `Purchase::Refundable#refund_and_save!` and `#refund_gumroad_taxes!` now populates `errors` with a user-facing message:

- `ChargeProcessorAlreadyRefundedError` → "The payment processor reports this purchase as already refunded…"
- `ChargeProcessorInvalidRequestError` → "The payment processor rejected this refund…" (this rescue in `refund_gumroad_taxes!` also now logs the processor response instead of swallowing it silently)
- nothing-to-refund early return (blank transaction id / already refunded / nothing refundable) → "This purchase has already been refunded or has nothing left to refund." (still returns `nil`, preserving the `refund_for_fraud!` skip-cleanly contract)
- standalone VAT refund (`refund_gumroad_taxes!`) already-refunded / no-refundable-tax early return → "The tax on this purchase has already been refunded or there is no tax left to refund." Invoice generation treats this specific outcome as expected and non-blocking (buyers regenerate invoices after a VAT refund), so `Purchases::InvoicesController#create` filters that message out; any other refund error still blocks with a visible alert.

## Why

gumroad-private#1267: a seller's dashboard refund of a chargeback-reversed PayPal purchase failed repeatedly with a **blank error toast** — `refund_and_save!` returned `false` with `errors` empty because the `ChargeProcessorInvalidRequestError` rescue added no message. The dashboard (`ProcessRefund` concern), admin UI, API v2, and the admin CLI all render `purchase.errors.full_messages`, so an empty errors collection means the user gets nothing to act on. This closes the "make every failure branch populate errors" ask on that issue; the PayPal amount-computation root cause is tracked separately there.

## User impact

No UI/flow changes — pure backend error-message population on paths that previously rendered an empty string. There is no adjacent UI surface to screenshot for this change: the refund modal is unchanged, only the message text inside the existing error toast goes from blank to descriptive. The spec run render in Test Results is the artifact.

## Test Results

test results

Local run (165 examples across purchase_refunds + the three refund workers): the only 2 failures are pre-existing local-env issues verified via a baseline run on clean origin/main (buyer-presentment VCR gap at :113, fresh-worktree vite manifest at :1245) — both fail identically without this diff. VCR cassettes for the renamed examples are byte-copies of the existing ones (interactions unchanged, only example names changed).

## QA steps

Backend-only; reviewer can run `bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb -e 'user-facing error'`.

---

**AI disclosure:** Written by Claude Fable 5 running as gumclaw's autonomous dev dispatcher. Prompt: standing dispatcher order (gumroad-private, Sahil 2026-07-19) to build open unassigned issues; task derived from gumroad-private#1267's engineering asks.

Fixes the blank-error ask in gumroad-private#1267.


gumclaw:

- **`refund_for_fraud!` contract preserved:** the nothing-to-refund early return still returns `nil` (not `false`), so bulk fraud callers with `skip_already_refunded: true` still skip cleanly. The only change is that `errors` is now populated — callers that branch on the return value are unaffected; callers that render `full_messages` get a message instead of blank. Checked all four UI entry points (`ProcessRefund` concern, admin purchases controller, internal admin API, API v2) — all render `full_messages`, none inspect error keys.
- **`Charge#refund_and_save!` combined-charge path:** it copies each failing purchase's `errors.full_messages` into the charge via `copy_refund_errors_from` — previously blank for these two rescue classes, now populated, so combined-charge refund failures surface too.
- **Sidekiq retry semantics unchanged:** `RefundPurchaseWorker` / `SyncStuckPurchasesJob` only branch on truthiness; adding errors doesn't change behavior.
- **VCR cassettes:** the three renamed examples' cassettes are copies of the originals (`git rm` old + copy to new name) — the HTTP interactions are identical since only the RSpec example descriptions changed. The new nothing-to-refund example reuses the unavailable-cassette but never issues a processor call (early return before `ChargeProcessor.refund!`).
- **Message wording:** kept in the same voice as the existing `Refundable` constants (actionable, mentions contacting support); the already-refunded message deliberately hedges ("If the refund is not reflected here…") because gp#1267 showed processor-side state can be ahead of ours.

gumclaw:
Addressed Greptile's P1 (blank `errors` on the `refund_gumroad_taxes!` early return) in 101db10eb:

- The already-refunded / nothing-refundable early return now adds `NO_TAX_TO_REFUND_ERROR_MESSAGE` ("The tax on this purchase has already been refunded or there is no tax left to refund."), so admin UI / admin API / dashboard callers rendering `errors.full_messages` never see a blank failure on this branch either.
- One caller treats this outcome as expected and non-blocking: invoice generation (`Purchases::InvoicesController#create`) retries the VAT refund on every regeneration and must still produce the invoice when the VAT was already refunded. It now filters out that specific message instead of gating on `errors.present?` — any other refund error still blocks with a visible alert, same as before.
- `Charge#refund_gumroad_taxes!` already copies purchase errors up via `copy_refund_errors_from`, so charge-level callers get the message for free.
- Renamed the two existing specs to assert the new message + `false` return (VCR cassettes renamed to match; interactions unchanged).

Local run: the 2 renamed examples pass (`2 examples, 0 failures`); rubocop clean. The invoices controller spec file fails 46/47 locally both with and without this diff (identical failure sets — pre-existing local-env chargeable-creation issue), so CI is the arbiter there.

PR #6223, opened 2026-07-24
Report client-side confirmPayment failures to the server (iDEAL redirect-leg forensics)

gumclaw:
## What

Adds `POST /orders/:id/confirm_error`, called by the checkout client when `stripe.confirmPayment` rejects on the client-confirm lane, so redirect-based payment method failures are visible in production.

## Why

A rejected confirm on a redirect-based method (iDEAL, Bancontact) is invisible server-side today:

- No charge exists yet, so no `payment_failed` webhook ever fires.
- The purchase sits `in_progress` until `FailAbandonedPurchaseWorker` cancels it — indistinguishable from a buyer who closed the tab.
- The Stripe error only ever appears in the buyer's browser.

The 2026-07-23 iDEAL ramp (antiwork/gumroad-private#933, Gate B / #5362) produced **zero completions in ~3h at 100%** with zero server-side evidence of why: server-side confirm was verified clean via a test-key control, so the failure is in the client confirm → bank-redirect → return-URL leg — exactly the blind spot this instruments. One retry-loop buyer attached a real iDEAL PaymentMethod but never produced a charge; today we cannot see the error they hit.

## How

- **Client** (`order.ts`): on `confirmResult.error`, fire-and-forget POST of the Stripe error type/code/message + payment method type. Reporting failures are swallowed — it must never break the buyer-facing error path.
- **Server** (`OrdersController#confirm_error`): resolves the order by its confirm-scoped secure token (only the owner of a real prepared order can report), length-caps every value, logs, and raises a Sentry event via `ErrorNotifier`.

## QA / tests

- `vitest run app/javascript/data/order.test.ts` — 7/7 passing locally, including a new test that the report fires and the buyer still sees the failure.
- New controller specs: reports to `ErrorNotifier`, truncates oversized payloads, 404s on an unknown token.

Next step once merged + deployed: re-ramp `checkout_local_method_ideal` and read the reported errors for the actual redirect-leg failure.

Part of antiwork/gumroad-private#933 (Gate B, antiwork/gumroad#5362).


app/controllers/orders_controller.rb
```diff
@@ -86,6 +86,37 @@ def finalize
...
+    error_details = {
+      order_id: order.id,
+      stage: params[:stage].to_s.first(50),
+      payment_method_type: params[:payment_method_type].to_s.first(50),
+      stripe_error_type: params[:stripe_error_type].to_s.first(100),
+      stripe_error_code: params[:stripe_error_code].to_s.first(100),
+      stripe_error_message: params[:stripe_error_message].to_s.first(500),
+    }
+    Rails.logger.error("Client-confirm browser error for order #{order.id}: #{error_details.inspect}")
+    ErrorNotifier.notify("Client-confirm browser error: #{error_details[:stripe_error_code].presence || error_details[:stripe_error_type].presence || "unknown"}", **error_details)
```
rmarescu: Consider a per-order rate limit or a fixed Sentry message with the Stripe code in context. Payload caps do not limit event count.

rmarescu (approved):

app/controllers/orders_controller.rb
gumclaw: Done in 8203ff8 — took both suggestions:

- **Fixed Sentry message:** the notify title is now the constant `"Client-confirm browser error"`; the Stripe error code/type/message stay in the event context, so everything groups into one Sentry issue instead of one per code.
- **Per-order rate limit:** notifier events are capped at 5 per order per hour via a `Rails.cache.increment` counter (fail-open on cache miss so we never lose the forensic signal). Past the cap the endpoint still returns `success: true` and still writes the unconditional `Rails.logger.error` line, so full detail remains in the logs even when Sentry is throttled.

Added a spec asserting notify fires exactly `CONFIRM_ERROR_NOTIFY_LIMIT_PER_ORDER` times across limit+2 requests. All 4 `POST confirm_error` specs pass locally; rubocop clean.

PR #6225, opened 2026-07-24
Fix PayPal multi-currency refund picking the wrong order item when skus are duplicated

gumclaw:
## What

Two fixes to the PayPal multi-currency refund path (`refund_order_purchase_unit!`):

1. **Duplicate-sku item matching.** `item_refund_cents_from_paypal` matched the PayPal order item by sku with `items.find`, which returns the first match. When the same sku appears twice on an order — a free ($0) sibling purchase of the same product on the same combined charge adds a zero-value line — it picked the €0.00 line, computed a refund of 0, and `refund_order_purchase_unit!` raised `ChargeProcessorAlreadyRefundedError` ("no remaining balance to refund") on a capture that was fully unrefunded. Zero-value lines are now excluded when the sku matches multiple items, and if the sku still matches more than one paid line the code falls back to the USD→merchant-currency conversion instead of guessing.

2. **FAILED/CANCELLED refunds counted as consumed balance.** `remaining_capture_cents` summed all refunds attributed to the capture regardless of status. A FAILED refund never moved money, but it permanently shrank the computed remaining balance (negative on the incident capture: 7616 captured − 5742 FAILED − 7616 COMPLETED = −5742), which would block legitimate retries forever. FAILED and CANCELLED refunds are now excluded; refunds without a status (older payloads/fixtures) stay counted.

## Why

A seller's dashboard refund on a chargeback-reversed PayPal purchase (€64 + VAT, combined charge with a free sibling row of the same product) failed repeatedly: `refund_and_save!` returned false and Gumroad never called PayPal, because the duplicate-sku match computed a 0-amount refund and the "already refunded" rescue swallowed it. A direct `PaypalRestApi#refund` with the correct amount on the same purchase succeeded (COMPLETED, full gross including fee return) — confirming reversed-chargeback PayPal purchases are refundable and the failure was Gumroad's amount computation. Both bugs were replayed against the live PayPal order payload: the sku match picked `unit_amount = 0.00` out of `[0.00, 64.00]`, and the computed refund amount came out to −5742.

Refs: antiwork/gumroad-private#1267

## QA steps

Backend-only change with no UI surface — the refund computation is exercised entirely through `ChargeProcessor.refund!`. Reviewer verification path:

1. Run `bundle exec rspec spec/business/payments/charging/implementations/paypal/paypal_charge_processor_spec.rb -e "multi-currency combined-charge refund path"` — 15 examples, 0 failures.
2. The new examples reproduce the incident shape: duplicate sku with a 0.00 line (refund uses the paid line's amount), ambiguous duplicate paid lines (falls back to conversion), and FAILED/CANCELLED refunds on the capture (excluded from remaining balance).

## Before/After

Not applicable — no user-facing rendering changes. Behavior change is visible only in the refund amount sent to PayPal: before, a duplicate-sku order computed a €0 refund and raised "no remaining balance"; after, the paid item's amount is refunded.

## Test Results

test-results

New and existing specs in the multi-currency refund block: 15 examples, 0 failures. Remaining failures in the wider spec file are pre-existing on origin/main (worktree environment: vite/PostCSS + merchant-account fixture collisions), identical counts on an unmodified main worktree.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent). Instructions: investigate the root cause of the failed seller-path PayPal refund on a chargeback-reversed purchase (gumroad-private#1267) and fix it. The agent replayed the refund amount computation against the live PayPal order payload via the audited read-only production console, identified the duplicate-sku and FAILED-refund-counting bugs, wrote the fix and specs, and ran the affected specs locally.


gumclaw:

- `item_refund_cents_from_paypal`: the zero-value exclusion only kicks in when the sku matches more than one line, so single-item orders (the overwhelmingly common case) take exactly the same path as before. If every duplicate line is zero-value, or more than one paid line shares the sku, the method returns nil and the caller falls back to the pre-#5343 USD→merchant-currency conversion — fail-safe, never a guessed item.
- `remaining_capture_cents`: PENDING refunds stay counted (money may still move); only FAILED and CANCELLED are excluded. Refunds without a `status` field (older payloads and existing fixtures) keep the previous behavior of being counted, so no fixture churn.
- Verified against the live incident order payload via the audited read-only prod console: before the fix the computation picked the 0.00 line (desired = 0) and the remaining balance came out negative (−5742); with the paid line selected the refund amount matches what succeeded when called directly.
- Spec baseline: the wider spec file has 21 pre-existing failures in this worktree environment (vite/PostCSS + merchant-account fixture collisions) — identical count on an unmodified origin/main worktree. The multi-currency refund block runs 15 examples, 0 failures including the 3 new ones.

gumclaw:
CI green across the board (Buildkite + all test shards) — arming auto-merge. Crisp spec lane: root cause verified live on the exact failing purchase (gp#1267), fix is scoped to the PayPal item-refund amount computation with specs covering the duplicate-sku and FAILED/CANCELLED-refund cases.

PR #6228, opened 2026-07-24
Fix PayPal refund webhook StatementTimeout in the shared-capture sibling scan

gumclaw:
## What

Splits the PayPal refund-webhook shared-capture sibling scan (`PaypalChargeProcessor.refund_purchase`) into two per-index queries instead of one relation combined with `.or`, unioning the resulting ids in Ruby. The guard's behavior and result set are unchanged.

## Why

The combined OR query made MySQL refuse both secondary indexes and range-scan the `purchases` PRIMARY key. Verified against production with the incident's actual values (read-only audited console, 2026-07-24):

- Combined OR (current code): `key=PRIMARY`, `rows=164,836,256` — exceeds the 5-minute `max_execution_time` and raises `ActiveRecord::StatementTimeout` on every run.
- Split query A (capture id): `key=index_purchases_on_stripe_transaction_id`, 2 rows, **1ms**.
- Split query B (order id): `key=index_purchases_on_paypal_order_id`, 2 rows, **2ms**.

Consequence of the timeout: `PAYMENT.CAPTURE.REFUNDED` / `PAYMENT.CAPTURE.REVERSED` webhooks could never record a refund for any purchase carrying a `paypal_order_id` — the Sidekiq job died on the sibling scan before reaching the guard's decision, silently (no Sentry sibling-guard notification, no refund row). This is why a COMPLETED PayPal refund on a real purchase went unrecorded for hours and needed a manual console recording with an extended statement timeout.

Refs antiwork/gumroad-private#1267 (ask 3: "why did PAYMENT.CAPTURE.REFUNDED not record the refund and why does the recording transaction exceed the default statement timeout" — this was the mechanism; the recording path itself is fast once the scan is indexed).

## Before/After

Not a UI change — webhook/background code path only, no user-facing surface renders differently. The artifact is the spec run + production EXPLAIN evidence:

spec results

## QA steps

1. Review the EXPLAIN evidence above (from the live incident's capture/order ids).
2. Confirm the split queries are logically equivalent to the previous OR relation: same three predicates per branch, `|=` union dedupes ids.
3. Run the guard's spec block locally: `bundle exec rspec spec/business/payments/charging/implementations/paypal/paypal_charge_processor_spec.rb -e "shared by multiple purchases"`.

## Test results

```
7 examples, 0 failures
```

All 7 existing shared-capture guard specs (block/notify on failed sibling, same-capture sibling, NULL-price sibling; allow on different-capture sibling, free sibling; REVERSED blocked; DENIED unaffected) pass unchanged — no spec edits were needed because the change is behavior-preserving.

(The wider `handle_order_events` block has 2 pre-existing local-env failures — a missing merchant-account seed and a Vite manifest error — both reproduced on unmodified `origin/main` in the same worktree before the change.)

---

**AI disclosure:** Generated with Claude Fable 5 (Hermes Agent, gumclaw). Instruction: root-cause gumroad-private#1267's webhook non-recording + statement-timeout ask; diagnosed via production EXPLAIN on the audited read-only console and implemented the query split with local spec verification.


gumclaw:
Self-review (adversarial pass, Claude Fable 5):

- **Equivalence:** the original relation was `(A OR B) AND price-filter`; the split applies the price filter to each branch — `(A ∧ C) ∨ (B ∧ C)` ≡ `(A ∨ B) ∧ C`, so the id set is identical. A row matching both branches appeared once in the OR pluck and appears once after `|=` (array union dedupes). Ordering is irrelevant downstream (ids are only used for `any?` and the notification payload).
- **Race window:** running two queries instead of one opens a microsecond gap where a sibling could be created between them. The original code already ran outside any transaction, and the guard is a best-effort ambiguity check that fails toward manual review, so this doesn't change the safety posture.
- **Hunted, not found:** no other caller builds this sibling scope; `refund_purchase` is the only consumer. No N+1 introduced (two point lookups, both index-backed). The `paid_or_unknown_price` string condition is unchanged semantics from the previous `.where(...)`.
- **Production proof:** EXPLAIN with the incident's values — combined OR chose `PRIMARY` at 164,836,256 rows (deterministic StatementTimeout at the 5-min cap); split queries chose `index_purchases_on_stripe_transaction_id` / `index_purchases_on_paypal_order_id` at 2 rows each, 1–2ms measured.

Payments surface — no auto-merge; leaving merge to @gianfrancopiana.

PR #6232, opened 2026-07-24
Remove the audience_count_from_elasticsearch and index_audience_members rollout flags (100% on)

gumclaw:
### What

Removes the `audience_count_from_elasticsearch` and `index_audience_members` rollout flags and makes the flag-on path unconditional:

- `AudienceMember` records are now always indexed into Elasticsearch on create/update/destroy (deleted the `GatedAsyncIndexing` prepend that guarded `send_to_elasticsearch`).
- `AudienceMember.count_for_seller` and `Installment#audience_members_count` always serve counts from Elasticsearch via `filter_count` (deleted the SQL fallback branches).
- Spec sweep: removed `Feature.activate_user` setups, deleted the flag-off contexts (they pinned dead SQL-count/no-indexing behavior), and deduped the follower-count controller example against the former flag-on context. The recipient-counts controller spec now indexes audience members into ES for every example, since all counts come from ES.

### Why

Part of the antiwork/gumroad-private#1208 rollout-flag audit (Category B). Both flags are fully ON in production — state=on, boolean=true, per the audited prod-state read (request id `20260721T025828Z-d0d55a50`, 2026-07-21) — and eligible for removal under the 2026-07-23 directive to clean up Category B flags at 100% for 30+ days.

**This change is behavior-neutral: with both flags at 100%, production behavior already IS the post-removal behavior.** The flag-off branches (SQL audience counts, skipped indexing) are dead code in prod.

### Before/After

Backend-only change with no UI delta — audience counts continue to be served from Elasticsearch exactly as they are in production today, so there is no before/after to capture on screen; test evidence below (termshot unavailable in this environment, so spec output is attached as text).

### QA steps

No preview-URL click-path applies (no user-visible surface changes). To verify behavior: open any post/email composer as a seller and confirm audience/recipient counts load — they are served by `Api::Internal::Installments::RecipientCountsController`, which now always reads from Elasticsearch (identical to current production behavior with the flags at 100%).

### Test Results

```
DISABLE_SPRING=1 OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES bin/rspec \
  spec/models/concerns/audience_member/searchable_spec.rb \
  spec/controllers/api/internal/installments/recipient_counts_controller_spec.rb

Finished in 24.63 seconds (files took 2.73 seconds to load)
42 examples, 1 failure

Failed examples:
rspec ./spec/models/concerns/audience_member/searchable_spec.rb:321
  # AudienceMember::Searchable.filter_count counts consistently when cutoff dates carry timezone offsets
```

The single failure is pre-existing and unrelated: it fails identically on the unmodified branch tip (verified via `git stash` + re-run), fails on the SQL `AudienceMember.filter` assertion this PR does not touch, and is a local MySQL timezone environment issue. All 41 other examples pass, including every example touched by this change.

`bundle exec rubocop -a` on the four changed files: no offenses.

---

🤖 AI disclosure: This PR was authored by Claude Fable 5 (Anthropic) working autonomously from the gumroad-private#1208 flag-audit task brief. Prompts given: "Remove the audience_count_from_elasticsearch and index_audience_members rollout flags — map the gating surface with grep, make the flag-on branch unconditional, delete flag-off spec contexts, verify zero remaining references, run touched specs, open a PR citing gp#1208, and arm auto-merge."

### DIFF
--- a/app/models/concerns/audience_member/searchable.rb
+++ b/app/models/concerns/audience_member/searchable.rb
@@ -8,25 +8,10 @@ module AudienceMember::Searchable
   ].freeze
   AFFILIATE_DETAIL_KEYS = %w[id product_id created_at].freeze
 
-  # Audience members churn on every purchase and follow across all sellers, so only
-  # sellers being rolled out enqueue indexing jobs. The index_audience_members flag
-  # turns on syncing ahead of the audience_count_from_elasticsearch read flag: backfill
-  # and verify the data while counts still come from SQL, then flip the read flag on a
-  # complete index. The read flag also implies syncing so a mis-ordered rollout serves
-  # an index that is merely incomplete, never one that silently rots.
-  module GatedAsyncIndexing
-    private
-      def send_to_elasticsearch(action)
-        return unless Feature.active?(:index_audience_members, seller) || Feature.active?(:audience_count_from_elasticsearch, seller)
-        super
-      end
-  end
-
   included do
     include Elasticsearch::Model
     include SearchIndexModelCommon
     include ElasticsearchModelAsyncCallbacks
-    prepend GatedAsyncIndexing
 
     index_name "audience_members"
 
@@ -114,11 +99,7 @@ module AudienceMember::Searchable
     # counts: numbers displayed side by side must come from one snapshot, or the
     # engines' different sync latencies can show a filtered count above the total.
     def count_for_seller(seller)
-      if Feature.active?(:audience_count_from_elasticsearch, seller)
-        filter_count(seller_id: seller.id)
-      else
-        seller.audience_members.count
-      end
+      filter_count(seller_id: seller.id)
     end
 
     # Builds an Elasticsearch query equivalent to the SQL built by AudienceMember.filter,
--- a/app/models/concerns/audience_member/searchable.rb
+++ b/app/models/concerns/audience_member/searchable.rb
@@ -87,6 +87,13 @@ module AudienceMember::Searchable
 
   class_methods do
     def filter_count(seller_id:, params: {}, limit: nil)
+      # Installments can exist without a seller (legacy product-owned posts delegate
+      # ownership to the product). The SQL twin of this method (AudienceMember.filter)
+      # treats a nil seller_id as "matches nobody" and returns an empty relation, whereas
+      # Elasticsearch rejects a null term filter outright with a 400. Preserve the
+      # SQL behavior so callers don't have to special-case seller-less records.
+      return 0 if seller_id.blank?
+
       options = {
         index: index_name,
         body: { query: filter_query(seller_id:, params:) },
--- a/app/models/installment.rb
+++ b/app/models/installment.rb
@@ -939,11 +939,7 @@ class Installment < ApplicationRecord
   end
 
   def audience_members_count(limit = nil)
-    if Feature.active?(:audience_count_from_elasticsearch, seller)
-      AudienceMember.filter_count(seller_id:, params: audience_members_filter_params, limit:)
-    else
-      AudienceMember.filter(seller_id:, params: audience_members_filter_params).limit(limit).count
-    end
+    AudienceMember.filter_count(seller_id:, params: audience_members_filter_params, limit:)
   end
 
   def api_audience_members_count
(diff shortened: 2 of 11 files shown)


gumclaw:
CI run 30075051701 failed on 6 shards, all with the same failure class: spec/test files that implicitly relied on the removed flags being **off** in the test env. With the flags gone, audience-member indexing and ES-served audience counts are unconditional, so specs that never indexed members into Elasticsearch (or asserted exact indexer-job counts) started failing. Fixed in 880a5f988:

- **Test Fast 4** — `purchase/searchable_spec.rb` "updates related purchases when deactivated_at changes": `Subscription#deactivate!` also removes the buyer's now-empty AudienceMember row, which enqueues its own ES delete job. The assertion now counts only `Purchase` jobs.
- **Test Fast 10** — `license_spec.rb` "enqueues a purchase re-index when uses changes via increment!": a uses change also syncs the buyer's AudienceMember document (its `license_uses` detail feeds the minimum-license-uses filter), tripping the strict `perform_in` message expectation. Added a catch-all `allow` so the assertion stays scoped to the purchase re-index.
- **Test Fast 16** — `with_filtering_spec.rb` seller-level "bought variants only" examples: saving a seller installment now always validates the sending limit via `AudienceMember.filter_count`, and the fabricated `bought_variants: %w[a b]` ids decrypt to `nil`, producing an ES `terms` query with no values (`parsing_exception`). The stored filters now use real variant external ids; the runtime args passed to `seller_post_passes_filters` stay arbitrary since they're only string-matched in memory.
- **Test Slow 4 / Slow 31** — `emails/edit_spec.rb` + `emails/create_spec.rb` system specs: the "Audience N / M" counts are served from ES and the fixtures' indexing jobs sit in the fake Sidekiq queue, so the members are now imported into the index (`index_model_records(AudienceMember)`) before loading the page — same pattern the recipient-counts controller spec uses.
- **Test Minitest** — `installment_test.rb` "audience_members_count returns the number of audience members": the Minitest harness stubs EsClient globally (count always 0), so the SQL-count expectation can't hold anymore. Rewrote it to assert the delegation to `AudienceMember.filter_count` (which fails if the flag-gated SQL path is restored) and folded in the previously skipped ES twin test.

Also swept `spec/` and `test/` for other flag-off assumptions (fake `bought_variants`/`not_bought_variants` ids, `ElasticsearchIndexerWorker.jobs.size` assertions, MySQL-sourced count expectations): the remaining hits either use `build` (no validation), stub `audience_members_count`, or exercise the SQL `AudienceMember.filter` path that still exists for blast jobs — none depend on the removed flags.

Verified locally: `license_spec.rb` (19/19), `with_filtering_spec.rb` (93/93), `purchase/searchable_spec.rb` Subscription Callbacks (2/2), `installment_test.rb` audience_members_count (4 assertions, 0 failures). The two `:js` system spec files couldn't be run in this worktree (no JS deps installed); their fix follows the already-verified ES-import pattern, and auto-merge is armed so this CI run confirms them.

---
🤖 AI disclosure: written by Claude Fable 5, prompted to fix the CI failures on this PR caused by the audience ES flag removal and sweep for other specs pinning the old flag-off behavior.

gumclaw:
Second CI run (30079581693) surfaced two more flag-off dependencies, fixed in d02c7bd07:

- **Test Fast 11** — `api/v2/emails_controller_spec.rb` (8 examples): legacy product/variant-owned installments are built with `seller: nil`, and the sending-limit validation now always calls `AudienceMember.filter_count`, which sent a `{term: {seller_id: null}}` filter to Elasticsearch → 400 `parsing_exception`. The removed SQL fallback used to swallow this (empty relation → count 0). `filter_count` now returns 0 for a blank `seller_id`, preserving the SQL twin's semantics. Verified locally: 60/60 in that spec file.
- **Test Fast 2** — `audience_counts_controller_spec.rb` "returns audience count" (`count 6` vs `0`): counts are ES-served unconditionally now, but the fixtures' indexing jobs sit in the fake Sidekiq queue. The spec now indexes members synchronously, same pattern as `recipient_counts_controller_spec.rb`. (This spec can't run in my local env — purchase factory card-charge stub issue that also fails on the parent commit — but the fix mirrors the CI-green sibling spec exactly.)

Note: I've **disabled the auto-merge** that was armed earlier — this should get a human review pass now that it's touching a shared runtime path (`filter_count`), not just specs.

---
🤖 AI disclosure: written by Claude Fable 5, prompted to fix the remaining CI failures caused by the audience ES flag removal.

PR #6233, opened 2026-07-24
Highlight the selected tier/version card with the seller's accent color

gumclaw:
## What

Highlights the currently selected tier/version card on the product page with an accent-colored outline. The selected card in the shared `Tabs` `buttons` variant (`app/javascript/components/ui/Tabs.tsx`) gets `border-accent` plus a 1px `ring-accent`, and the previous lift + drop shadow treatment is removed — the accent outline alone marks the selection.

## Why

Tier/version selection was only indicated by a lift + drop shadow, which is easy to miss on some seller themes — a buyer reported not being able to tell which tier they had picked before hitting the buy button. The `accent` color is a seller theme token, so the outline follows each storefront's configured accent color in both light and dark mode. The border + 1px ring combination gives a ~2px outline without any layout shift, since the ring is a box-shadow rather than extra border width.

The shadow and lift were dropped rather than kept alongside the outline: the outline sells the selected state on its own, and the dark drop shadow clashes with bright accent colors.

This also applies to the other selectors built on the same `buttons` variant (payment option, call start times, review type, payout method cards, etc.), which get the same clearer selected state for free.

Refs: antiwork/gumroad-private#1281

## Before / After

Desktop, light + dark (selected card: "Pro"):

before-after

Mobile (375px), light + dark:

mobile

## QA steps

1. Open any membership product page with multiple tiers (or a product with versions).
2. Click between tiers/versions.
3. The selected card shows an outline in the seller's accent color, with no lift or drop shadow; unselected cards keep the plain border.
4. Toggle dark mode — the accent outline stays visible.

## Test Results

The screenshots above were captured from the real `Tabs`/`Button`/`Pill` components rendered with the app's real stylesheets in a local Vite harness, driven by an assertion-first Playwright script (68/68 assertions passed: exactly one selected card per state; on "after" the computed border color equals the accent token, the only box-shadow layer is the 1px accent ring, and there is no translate; on "before" the shadow + lift are present and the border is plain — across light/dark × desktop/mobile). No specs reference `ui/Tabs.tsx` directly; the change is presentation-only (class strings), with no behavior or markup changes.

---

## AI disclosure

Written by Claude Fable 5 (Hermes agent), prompted from antiwork/gumroad-private#1281 to make the selected tier/version visually distinct with an accent color, then revised per review feedback to remove the lift + drop shadow so the accent outline alone marks the selection.


gumclaw:
Self-review (adversarial pass, Claude Fable 5):

**Blast radius:** this edits the shared `buttons` compound variant, so the accent outline applies to every `Tabs variant="buttons"` selector, not just product-page tiers/versions — payment option, call start times, review type (rating/video), payout method cards, workflow trigger, Developer widgets tabs, and the admin user/products tabs. I checked all 17 call sites: none pass a `border-*` class to `Tab` that the new `border-accent!` could clobber, and accent-on-selected is already the pattern for `Checkbox`/`Radio` (`checked:bg-accent`), so consistency holds. The issue itself proposed changing the shared variant.

**Why the `!` on `border-accent`:** `Tab` renders via Radix `Slot` (`asChild`), which joins the Tab and Button class strings with plain concatenation — `twMerge` never sees both, so Button's own `border-border` can win by CSS order. The important modifier settles it deterministically. `ring-1 ring-accent` composes with the existing `shadow` lift (Tailwind v4 combines ring and shadow box-shadows), verified in the rendered captures.

**No layout shift:** border width is unchanged (1px) and the ring is a box-shadow, so card dimensions are identical selected vs. unselected — confirmed by the before/after captures.

**Verdict:** no P1/P2 findings. Presentation-only, no markup/behavior/a11y changes (`aria-checked`/`aria-selected` untouched).

gumclaw:
Triage of Greptile's T-Rex log (the buried "computed borderColor was rgb(0, 0, 0)" line in the summary above): **harness artifact, not a bug in the change.**

`border-accent` / `ring-accent` resolve to `--color-accent: rgb(var(--accent))` (tailwind.css), which requires `--accent` to be a space-separated RGB triplet. In the real app that variable is always defined: `_definitions.scss`'s `globals` mixin sets a default triplet on `:root`, and seller storefronts override it via `app/views/layouts/custom_styles/styles.scss.erb` (`--accent: #{split-color(highlight_color)}`). T-Rex's standalone script rendered the component without those global styles, so `rgb(var(--accent))` was invalid and the border fell back to `currentColor` (black).

Reproduced the failure mode to confirm:

| `--accent` value | computed `border-color` |
|---|---|
| `35 160 148` (triplet, as the app defines it) | `rgb(35, 160, 148)` ✅ |
| `#23a094` (hex) | `rgb(0, 0, 0)` — matches T-Rex |
| undefined | `rgb(0, 0, 0)` — matches T-Rex |

The PR's own Playwright evidence harness loaded the app's real stylesheets (including the globals) and asserted the computed selected-card border equals the accent token across light/dark × desktop/mobile — 32/32 passed, and the screenshots in the body show the accent outline rendering. No code change needed.

gumclaw:
**Triage of Greptile's P1 ("Selected Tabs buttons card does not use seller accent color"): false positive — dismissed with counter-evidence.**

Re-ran the verification just now against this branch head (c9261cf) with the real `Tabs`/`Button` components, the repo's actual Tailwind 4 stylesheet build (`app/javascript/stylesheets/tailwind.css` via the Vite harness), and Greptile's exact test accent `--accent: 35 160 148`. Computed styles on the selected card:

```json
{
  "label": "Pro",
  "checked": "true",
  "borderColor": "rgb(35, 160, 148)",
  "ringInShadow": true   // box-shadow contains rgb(35, 160, 148) 0px 0px 0px 1px
}
```

Unselected cards stay at the plain border with no ring, exactly as intended. Same result with the default theme accent (`rgb(255, 144, 232)`).

On the suggested fix: `border-accent!` (trailing `!`) **is** the Tailwind 4 important syntax, and it's the established pattern across this codebase — see `Button.tsx` (`translate-0!`, `shadow-none!`), `ui/FormSection.tsx` (`p-4!`), `ui/PageHeader.tsx` (`hidden!`), `ui/ProductCardGrid.tsx` (`grid-cols-2!`). The proposed `!border-accent` is the legacy v3 prefix form. `ring-accent` doesn't need the important modifier because nothing else in the merged class string sets a ring color.

The T-Rex harness result (black border/ring) doesn't reproduce against the repo's real stylesheet build, so it looks like an artifact of the harness's own Tailwind compilation rather than a product bug. The PR's committed QA screenshots were themselves captured from an assertion-first Playwright run (32/32, including computed-border-equals-accent checks) on this same setup.

No code change needed.

slavingia:
Remove the box shadow etc. I think the focus sells it enough. Wdyt? Pink and black clash anyway.

gumclaw:
Agreed — done. The selected card now gets just the accent border + 1px ring: no shadow, no lift. Refreshed the before/after screenshots in the body (desktop + mobile, light + dark) and dropped the variants image since that decision is settled. Re-verified with the Playwright harness: 68/68 assertions — selected card's border equals the accent token, the only box-shadow layer is the 1px accent ring, and there's no translate on it.

gumclaw:
awaiting-human removed — Sahil's design verdict received (accent border + focus ring only, no shadow/lift) and already applied in the latest head. CI green except two shards finishing; auto-merge (squash) is armed.

PR #6234, opened 2026-07-24
Run blast snapshot revalidation under the raised statement execution cap

gumclaw:
## What

Runs the audience-snapshot revalidation in `SendPostBlastEmailsJob` under the same raised, Redis-tunable statement-execution cap (`WithMaxExecutionTime.timeout_queries`) that the fresh audience load and the non-opener email resolution already use.

## Why

The job has three expensive query phases for large blasts: the fresh audience load (capped since #5862), the non-opener email resolution (capped since #6192), and the snapshot revalidation on retries — which was still uncapped. `revalidate_snapshotted_members` runs the `AudienceMember.filter` in primary-key-bounded 10k-id slices, but each slice still joins several large tables, and for six-figure audiences a slice can exceed the database's default 5-minute statement cap.

Because the first attempt writes the snapshot and every retry resumes through the unprotected revalidation, a large blast that fails once for any reason then fails deterministically on every retry (`Mysql2::Error: Query execution was interrupted, maximum statement execution time exceeded`) until Sidekiq dead-sets the job. The blast stays incomplete forever, nothing is sent, and the seller's UI shows "in progress" indefinitely. This is the residual failure mode behind a real ~349k-recipient resend-to-non-openers blast that dead-set in production this week even after #6192 shipped.

Refs antiwork/gumroad-private#1280.

## Before / After

Not a UI change — no rendering path is touched. Behavior delta:

| | Before | After |
|---|---|---|
| Retry of a huge blast (snapshot present) | Revalidation slices run under the default 5-min statement cap → deterministic failure → DeadSet, blast never completes | Revalidation runs under the raised cap (1h default, Redis-tunable via `audience_member_load_max_execution_time_seconds`) — same protection as the fresh load |
| Small audiences | unchanged | unchanged |

## QA steps

Backend-only Sidekiq change with no user-facing surface; the failure mode requires a six-figure audience and can't be reproduced on a preview app. Verification is the spec suite:

1. `bundle exec rspec spec/sidekiq/send_post_blast_emails_job_spec.rb -e "audience snapshot for retry resume"` — includes the new example asserting the raised cap wraps the snapshot-resume path.
2. After deploy: requeue the stuck production blast and confirm the retry survives revalidation and `delivery_count` becomes non-zero.

## Test Results

10 examples, 0 failures (snapshot-resume + statement-timeout groups, run locally):

spec results

(The spec file's unrelated purchase-factory examples fail identically on an unmodified sibling worktree — pre-existing local-environment failures, not introduced by this change; CI is authoritative for the full file.)

---

**AI disclosure:** Generated with Claude Fable 5 (Hermes agent). Instructions: continue gumroad-private#1280 — diagnose why the large-audience resend-to-non-openers blast still dead-sets after #6192 and ship the fix. The agent verified prod state via the audited read-only console, identified the uncapped snapshot-revalidation path, wrote the fix and spec, and ran the touched specs locally.


gumclaw:

- The wrap mirrors the two existing capped call sites in this job (fresh audience load, non-opener email resolution) exactly — same `audience_load_timeout_seconds` source, so the existing Redis tunable (`audience_member_load_max_execution_time_seconds`) governs all three phases consistently, and the existing specs asserting cap seconds keep passing unchanged.
- No nesting hazard: the revalidation path is only reached when the snapshot exists, and none of the surrounding code is already inside a `timeout_queries` block, so the `previous_max_execution_time` save/restore in `WithMaxExecutionTime` behaves as on the other two call sites.
- Return value is unchanged (`flat_map` result passes through the block), so `dropped` accounting and the resume log line behave identically.
- Checked the other `AudienceMember.filter` call sites in the job — the fresh load is already capped; the only uncapped heavy queries left in `perform` are the per-slice send-loop queries, which are small (≤ `recipients_slice_size` records) by construction.

gumclaw:

Hunted for: nested `timeout_queries` (double set/restore corruption), Makara replica-vs-primary cap application, behavior drift in the revalidation return value, spec looseness, and cap-restore failure modes.

- **Nesting:** `revalidate_snapshotted_members` is reached only from `load_audience_members`'s snapshot branch, which is not inside any other `timeout_queries` block, so the save/restore of `@@max_execution_time` is single-level — same as the two existing call sites.
- **Makara/replica:** `WithMaxExecutionTime` sets the cap on `ActiveRecord::Base.connection` after the job's `Makara::Context.release_all`. This is byte-identical to how the fresh-load path (shipped in #5862) applies the cap, so no new node-routing behavior is introduced.
- **Cap held across the whole slice loop** rather than per-slice: intentional and matches the fresh-load precedent; the `ensure` restore fires once at the end, and an exception inside any slice still restores the previous cap.
- **Return value/logging:** `flat_map` result passes through the block unchanged; `dropped` accounting and the resume log line are unaffected.
- **Spec:** asserts `timeout_queries` receives `seconds: 1.hour.to_i` on a snapshot-resume run and that the send still completes — the wrap is load-bearing for the assertion (removing it fails the message expectation).

No P1/P2 findings. Verdict: safe to merge on green.

PR #6236, opened 2026-07-24
Stop alerting Sentry on expected 'unable to support this bank' and postal-code rejections during Stripe account creation

gumclaw:
## What

Adds two Stripe rejection shapes to the existing Sentry-suppression list in `StripeMerchantAccountManager.create_account`:

1. **"Stripe is unable to support this bank at this time."** — Stripe populates neither `code` nor `param` on this `InvalidRequestError`, so `bank_account_invalid_error?` missed it. It now matches on the message, like the other legacy no-code rejection shapes already handled on the sync path. The bank-sync payout-note breadcrumb still gets recorded.
2. **`postal_code_invalid` rejections** — these were never in the create-path suppression list even though they are surfaced inline to the seller, breadcrumbed as a payout note, and auto-retried weekly by `RetryStripeRejectedPayoutSetupsJob`.

## Why

Sentry issue GUMROAD-WW (`Stripe::InvalidRequestError` from `Settings::PaymentsController#update`, 6,740 events / 448 users over 14 days, [permalink](https://gumroad-to.sentry.io/issues/7578121692/)) kept firing after the earlier suppressions of bank (#5991), tax-ID (#6024), phone (#6008), and JP-address (#6047) rejections. Sampling the 44 events after those fixes deployed (2026-07-22 12:00Z → now) shows the residual volume is dominated by exactly these two expected-input shapes:

- 20× "Stripe is unable to support this bank at this time."
- 16× "Invalid <NL/ES/IT/BE/GB/DE/FR> postal code" / "Invalid postal_code for Japan."

Both are seller-input errors the seller sees inline on the payments settings page and can correct themselves (bank: enter a different account; postal code: fix it, or the weekly retry job re-attempts automatically). Only unexpected failures should page Sentry.

Remaining low-volume variants ("external account on your block list" ~5, Kanji-variation ~2) were deliberately left alerting — the block-list one in particular is a risk signal worth seeing.

## Before/After

Not a UI change — no rendered surface is affected. The seller-facing inline error messages, payout-note breadcrumbs, and the weekly postal-code retry behavior are all unchanged and asserted exactly by the specs; the only delta is that `ErrorNotifier.notify` is no longer called for these two expected rejection shapes.

## QA steps

1. Run the suppression specs: `bundle exec rspec spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb`
2. Red-first proof: revert `stripe_merchant_account_manager.rb` to `origin/main` and re-run the new examples — the 3 suppression assertions fail (Sentry gets notified); restore the fix and they pass.
3. After deploy, watch GUMROAD-WW: residual event rate should drop from ~30/day to single digits (block-list + Kanji variants only).

## Test Results

`33 examples, 0 failures` (full payout-retry spec file, run locally in the worktree):

spec results

Refs Sentry GUMROAD-WW; follows the pattern of #5991 / #6008 / #6024 / #6047.

---

**AI disclosure:** Generated with Claude Fable 5 (Gumclaw daily Sentry sweep cron). Prompt: enumerate top unresolved Sentry issues, classify already-fixed vs actionable vs noise, and fix actionable items with red-first tests per the repo's contribution gates.


gumclaw:
Self-review of the diff:

- `bank_account_invalid_error?` change is additive-only: all existing early returns (`CardError` → true, `incorrect_account_holder_name` → false, code/param matches → true) are untouched; the new message regex is the final fallthrough, so no previously-alerting shape becomes suppressed except the exact "unable to support this bank" message, and no previously-suppressed shape changes.
- Because `bank_account_invalid_error?` is shared with the sync path (`update_bank_account`), the "unable to support this bank" message now also routes to `:invalid_bank_account` there (payout note + `invalid_bank_account` email instead of Sentry + `:stripe_invalid_request`). That's the intended behavior — same treatment the sync path already gives BIC/routing rejections — and matches how the message-string fallbacks (`"Invalid account number"` etc.) on that path already behave.
- `postal_code_invalid_error?` added to the create-path `unless` only affects `ErrorNotifier.notify`; the `record_postal_code_failure_note` breadcrumb and the controller's inline messaging + weekly retry job are unchanged (asserted by existing specs, which still pass).
- Red-first proof run: with `stripe_merchant_account_manager.rb` reverted to `origin/main`, the 3 new suppression examples fail (ErrorNotifier receives notify); with the fix restored, 33/33 pass.

No payment/money behavior changes — only which failures page Sentry.

PR #6241, opened 2026-07-24
Remove offered bundle's contents from cart when accepting a replace-type cross-sell

gumclaw:
## Summary

Combining an **add-on cross-sell** and a **"Replace the selected products" cross-sell** on the same product let a buyer purchase the replacement bundle PLUS the individual products the bundle already contains (reported by seller `strategicaccumulator` via Helper; tracked in antiwork/gumroad-private#1286).

**Root cause:** `getCartIfAccepted()` in `app/javascript/pages/Checkout/Show.tsx` removes only cart items whose `product.cross_sells` include the replace-offer's id. Items injected mid-checkout by an earlier accepted add-on offer are inserted with `cross_sells: []`, so a subsequent replace offer never removes them.

**Fix:** when the accepted offer is a replace-type cross-sell, additionally drop cart items whose product id appears in the offered product's `bundle_products`. For non-bundle offered products `bundle_products` is empty, so behavior is unchanged.

## Changes
- `app/javascript/pages/Checkout/Show.tsx`: replace-filter now also matches items contained in the offered bundle, with a comment explaining why the id-tag filter alone is insufficient.
- `spec/requests/purchases/product/upsell_spec.rb`: new system spec — buyer accepts an add-on cross-sell, then a bundle replace cross-sell; asserts the add-on and the selected product both leave the cart, only the bundle is purchased, and no standalone (non-bundle-content) sales exist for the contained products.

## Verification
- `npx tsc --noEmit` clean; Prettier clean.
- New system spec runs on CI (js/system specs can't run on this macOS host — objc fork crash); existing replacement cross-sell specs cover the unchanged paths.
- End-to-end purchase run on the hosted preview app (below) confirms the buyer is charged for the bundle only.

Captured on `https://fix-replace-cross-sell-bundle-co.apps.staging.gumroad.org` at `be6087668738d1dbe06b7ededcb4a890da825758`, deviceScaleFactor 2, desktop 1440px + mobile 375px.

| Step | Desktop |
| --- | --- |
| 1. Add-on cross-sell fires first — buyer accepts, injecting "Bonus Texture Set" into the cart mid-checkout (this is the item the old filter could not see) |  |
| 2. Replace-type bundle cross-sell then fires with the cart holding **both** the selected product and the injected add-on (subtotal US$22) — the pre-fix state that produced the double charge |  |
| 3. After "Upgrade": purchase succeeds and the buyer receives exactly the bundle's contents — no duplicate standalone items |  |
| 4. Seller **Sales** page: **one** sale, `Complete Illustration Bundle` at **$19** — not $19 + $15 + $7 |  |

| Mobile (375px) — replace offer | Mobile (375px) — result |
| --- | --- |
|  |  |

Console-verified on the preview DB for both runs (desktop and mobile buyer emails):

```
bundle sales                = 1
standalone selected product = 0
standalone add-on product   = 0
total charged               = 1900 cents   (bundle price only)
contained products          = delivered as bundle-content purchases at 0 cents
```

## QA steps (hosted preview app)

Preview URL: **https://fix-replace-cross-sell-bundle-co.apps.staging.gumroad.org**

Seed state already present on the preview (seller `seller@gumroad.com` / `password`, 2FA `000000`):
- `Watercolor Brush Pack` ($15, permalink `qaselected`) — the selected product
- `Bonus Texture Set` ($7, permalink `qaaddon`) — offered by an **additive** cross-sell on the selected product
- `Complete Illustration Bundle` ($19, permalink `qabundle`) — a bundle containing BOTH of the above, offered by a **"Replace the selected products"** cross-sell on the selected product

Click path:
1. Open `/checkout?product=qaselected&quantity=1&price=1500` (cart: Watercolor Brush Pack, US$15).
2. Fill email + full name, card `4242 4242 4242 4242`, `12/34`, CVC `123`, ZIP `10001`. Click **Pay**.
3. First modal — "Add the Bonus Texture Set" → click **Add to cart**. The cart now shows both products, subtotal US$22.
4. Second modal — "Upgrade to the Complete Illustration Bundle" → click **Upgrade**.
5. Expected (this PR): purchase succeeds and the seller's **Sales** page shows exactly **one** sale — `Complete Illustration Bundle` at **$19**. Before the fix the buyer was also charged for the bundle's contents that were still sitting in the cart.
6. Repeat at a 375px viewport for the mobile pass.

## Notes for reviewers
- Product-call framing from the issue: of the three options listed, this implements the first (bundle-content-aware replace) because it fixes the surprising interaction without new seller-facing settings.
- Duplicate detection is by product id only (option/variant-agnostic): if the buyer's cart has a different variant of a contained product, it is still removed — matching seller intent that the bundle supersedes its contents.
- Offer ordering matters for reproducing: `Link#available_cross_sells` returns offers in id order, so the additive offer must be created before the replace offer to reach the buggy interleaving.

Fixes antiwork/gumroad-private#1286

---

## AI disclosure

Authored by Hermes Agent running **Claude Opus 5**. The fix, the new system spec, and the hosted-preview QA capture (Playwright driver + preview Rails console seeding/verification) were produced by the agent from the Helper report in antiwork/gumroad-private#1286 and the standing instruction to attach hosted-preview visual evidence to user-visible PRs.

### DIFF
--- a/app/javascript/pages/Checkout/Show.tsx
+++ b/app/javascript/pages/Checkout/Show.tsx
@@ -658,11 +658,21 @@ const CheckoutIndexPage = () => {
       );
       const originalCartItem = originalCartItems[0];
       if (originalCartItem) {
+        // When a replace-type cross-sell offers a bundle, also drop any cart items for products
+        // that are already inside that bundle (e.g. products added by earlier accepted add-on
+        // cross-sells). Those items aren't tagged with this offer's id (their cross_sells were
+        // stripped when they were injected mid-checkout), so the originalCartItems filter alone
+        // would leave the buyer purchasing the bundle plus its own contents.
+        const offeredBundleProductIds = new Set(
+          currentOffer.offered_product.product.bundle_products.map(({ product_id }) => product_id),
+        );
+        const replacedItems = (item: CartItem) =>
+          originalCartItems.includes(item) || offeredBundleProductIds.has(item.product.id);
         return {
           ...cartForm.data.cart,
           items: [
             ...(currentOffer.replace_selected_products
-              ? cartForm.data.cart.items.filter((item) => !originalCartItems.includes(item))
+              ? cartForm.data.cart.items.filter((item) => !replacedItems(item))
               : cartForm.data.cart.items),
             {
               ...currentOffer.offered_product,
--- a/spec/requests/purchases/product/upsell_spec.rb
+++ b/spec/requests/purchases/product/upsell_spec.rb
@@ -359,23 +359,26 @@ describe("Product checkout with upsells", type: :system, js: true) do
 
         within_modal "Bundle cross-sell" do
           expect(page).to have_section("Complete bundle")
+          # The cart still holds the selected product and the add-on at this point;
+          # accepting the replace-type offer is what must drop both.
           click_on "Upgrade"
         end
 
-        # The bundle replaces both the originally selected product and the
-        # add-on product it contains — the buyer must not buy duplicate content.
-        expect(page).to have_section("Complete bundle")
-        expect(page).to_not have_section("Selected product 1")
-        expect(page).to_not have_section("Add-on product")
-
         expect(page).to have_alert(text: "Your purchase was successful! We sent a receipt to test@gumroad.com.")
 
+        # What the buyer was actually charged for is the invariant that matters: one bundle,
+        # and no standalone purchase of anything the bundle already contains. (The success
+        # page legitimately lists "Selected product 1" and "Add-on product" as the bundle's
+        # contents, so asserting their ABSENCE on screen would contradict correct behavior.)
+        expect(bundle.sales.count).to eq(1)
         bundle_purchase = bundle.sales.last
         expect(bundle_purchase.upsell_purchase.upsell).to eq(bundle_cross_sell)
         expect(bundle_purchase.upsell_purchase.selected_product).to eq(selected_product1)
-        # No standalone purchase of the add-on product (only the bundle-content copy)
         expect(addon_product.sales.not_is_bundle_product_purchase).to be_empty
         expect(selected_product1.sales.not_is_bundle_product_purchase).to be_empty
+        # The contained products are still delivered — as bundle-content purchases, not sales.
+        expect(addon_product.sales.is_bundle_product_purchase.count).to eq(1)
+        expect(selected_product1.sales.is_bundle_product_purchase.count).to eq(1)
       end
     end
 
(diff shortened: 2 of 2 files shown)


slavingia:
Green and merge after fable 5 review

gumclaw:
## Fable 5 adversarial review — the fix holds; the spec was wrong

**CI failure diagnosed, not papered over.** `Test Slow 0` failed on `expect(page).to_not have_section("Add-on product")` reporting *2 matches*. Those negative assertions ran on the **post-purchase** view, where the bundle correctly renders its contents as temporary-library cards (`TemporaryLibrary.tsx` maps `result.bundle_products` → one Card per contained product). Since the bundle contains both "Add-on product" and "Selected product 1", asserting their absence contradicts correct behavior — the spec was testing the wrong surface, and it would have kept failing no matter how right the fix was.

Rewritten (ce7bfcbe2) to assert **what the buyer was charged for**, which is the actual invariant of this bug:
- exactly one `bundle.sales` record, attributed to the bundle cross-sell with `selected_product1` as the selected product
- zero `not_is_bundle_product_purchase` sales for the add-on and the selected product (the duplicate-purchase bug this PR fixes)
- one `is_bundle_product_purchase` each — so delivery of the contained products is still proven, not just absence of charges

## Adversarial checks on the fix itself (all pass)

1. **Id-namespace consistency** — the highest-risk silent-failure mode. `checkout_presenter` emits `bundle_products[].product_id` as `bundle_product.product.external_id`, and cart items carry `product.id = product.external_id`. Both sides of the new `Set.has()` are external ids, so the filter actually matches. A mismatch here would have made the fix a no-op that still passed a green suite.
2. **Non-bundle offers unaffected** — `bundle_products` is a required, non-nullable array on the `Product` type (`cartState.ts`), so for a plain product the Set is empty and the predicate degrades exactly to the previous `originalCartItems.includes(item)` behavior. No optional chaining needed, no crash path.
3. **Server accepts the trimmed cart** — `Purchase::CreateService` only validates "cart still contains a product the upsell applies to" on the `not_replace_selected_products?` branch. A replace-type offer whose selected product has been removed from the cart is not a `PurchaseInvalid` risk.
4. **Add-on attribution** — removing the add-on item drops its `accepted_offer`, so the add-on upsell records no purchase. Correct: no purchase was made.
5. **Scope reality check** — `has_flags` generates the positive scopes used in the rewritten assertions (`Purchase.is_bundle_purchase` is used the same way in `orders_controller_spec` and `update_bundle_purchases_content_job`), so the new expectations exercise real scopes rather than erroring.

**Deliberate behavior worth a reviewer's eye (documented in the PR body, unchanged):** duplicate detection is product-id-only, so a *different variant* of a contained product is also removed. That matches seller intent (the bundle supersedes its contents) but it is a judgment call, not a mechanical consequence.

Rubocop clean on the touched spec. Local system specs are not runnable on this host (vite/hydration), so CI is the source of truth — watcher armed, merging on green per @slavingia.

🤖 Adversarial review by claude-fable-5 via Hermes agent (gumclaw)

PR #6244, opened 2026-07-24
Require explicit deletion intent for configured or purchased variants

gumclaw:
## What

Extends the variant-deletion wipe guard from #6178 with a server-side intent check for **configured or purchased** variants. `Product::VariantCategoryUpdaterService.ensure_deletion_intent!` now requires the variant's external id in `confirmed_removed_variant_ids` before deleting any variant that:

- carries content (rich content pages or files) — existing behavior,
- has any successful purchase (`purchases.all_success_states`), or
- has non-default configuration: custom `price_difference_cents`, pay-what-you-want, quantity cap, duration, description, membership price-change settings, active integrations, or alive recurring prices with `price_cents > 0`.

Truly blank rows (no content, no purchases, all defaults) stay freely deletable, so ordinary create-and-discard editor flows keep working without extra confirmations.

Client side, the suggested-amounts editor (coffee products) now records deletion intent when the seller clicks the trash button — it has no confirmation modal, so the click itself is the intent signal. The versions/tiers/durations editors already recorded intent via their confirmation modals.

## Why

Fixes gumroad-private#1296 (split out from the gumroad-private#1230 variant-tree wipe incident). The #6178 guard deliberately only protected content-bearing variants, leaving a gap: a variant with real configuration (pricing tiers set up before content is added) was freely deletable by a stale/blind full-snapshot payload with no confirmation. A client-side summary alone cannot protect a variant the editor snapshot never loaded, and the category-level `has_alive_grouping_variants_with_purchases?` check only shields the whole-category path — and only when the purchased variant also has files — so per-variant deletions of purchased contentless variants were unguarded.

## Before / After

Backend guard change with no new UI surface — the blocked-save error path and the confirmation modals are unchanged from #6178 (see its qa-media screenshots). The one client change (suggested-amounts trash click records intent) produces no visual difference.

## QA steps

1. Create a product with versions; set a custom price on one version but add no content.
2. Save the product from a payload that omits that version without `confirmed_removed_variant_ids` (stale editor tab, or direct API call) → the save is rejected with "This save would remove versions that still have content, settings, or sales…" and the version survives.
3. Delete the same version through the editor's "Remove version" modal and save → deletion succeeds (intent recorded).
4. Add a blank version and remove it before saving → save succeeds with no confirmation.
5. On a coffee product, delete a suggested amount and save → succeeds (trash click records intent).

## Test Results

All service specs (41 examples) and the links controller Minitest suite (343 runs) pass locally:

test results

New regressions cover: blind omission of a configured (priced / quantity-capped) contentless variant, blind omission of a purchased contentless variant, intentional confirmed deletion, and untouched-default blank-row cleanup.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes agent), prompted by Gianfranco's request on gumroad-private#1296 to implement a server-side intent check for deleting persisted variants with non-default configuration or purchases, requiring `confirmed_removed_variant_ids` while keeping blank-row cleanup working, with regressions for blind omission, purchased contentless variants, and intentional deletion.


gumclaw:
Fable-5 adversarial review (pre-merge gate) — hunt list: guard-fork completeness, normal-flow regressions, query cost, client intent coverage.

**P1/P2: none found.**

Findings triage:

1. **Guard-fork completeness** — all three variant-deletion call sites (`options: nil` category clear, per-variant diff, whole-category drop in `VariantsUpdaterService#perform`) route through the same `ensure_deletion_intent!`, so the new predicate applies to every path uniformly. Verified no other caller deletes variants outside these paths (`batch_delete_variants` / `mark_deleted` grep).
2. **Membership default tier** — freshly-created memberships auto-create a default tier with a recurring price; the `price_cents > 0` filter keeps a $0 default tier freely deletable, so new-product flows don't regress. Priced tiers require confirmation, which the tiers editor already records via its removal modal.
3. **Coffee products** — the auto-created suggested amount carries a real price, so it's now protected; the suggested-amounts editor had no intent recording (no confirmation modal), which would have broken coffee saves. Fixed in this PR by recording intent on trash-click. Controller tests that posted blind coffee payloads were updated to address the existing variant by id, matching what the editor actually sends.
4. **P3 (accepted): SKU deletion** (`SkusUpdaterService`) remains outside the intent guard — SKUs are derived from variant combinations and are only deleted downstream of a variant deletion that itself passed the guard. Same scope boundary as #6178.
5. **Query cost** — the new predicate adds up to three `EXISTS` queries per variant, but only on the deletion path (variants missing from the payload), so normal saves are unaffected.

Local runs green: 41 service-spec examples, 343 controller-test runs, 0 failures (see Test Results in the body).

PR #6247, opened 2026-07-24
Invalidate Pages::ProfileData cache on seller username change

gumclaw:
## What

Adds the seller's username to the `Pages::ProfileData` cache key and bumps `CACHE_VERSION` to `v3`, so the cached storefront `gumroad-data` JSON is invalidated when a seller changes their username.

## Why

`Pages::ProfileData` caches each custom-HTML storefront's injected `gumroad-data` JSON, which embeds full product URLs built from the seller's subdomain (`Link#long_url` → `User#subdomain_with_protocol`). The cache key only tracked the products collection, visible installments, and the seller profile row — a username change touches none of those, so the cached blob (with every `https://<old-username>.gumroad.com/l/...` URL) survived the rename indefinitely. Every product card on the renamed seller's custom landing page then linked to the old subdomain, which 404s — the storefront was effectively unbuyable until an unrelated product/profile update happened to bust the cache.

Keying on `seller.username` (rather than `seller.cache_key_with_version`) invalidates precisely on rename without rebuilding the cache on unrelated user-row updates. The `CACHE_VERSION` bump to `v3` also flushes any already-stale entries from past renames, covering the one-off sweep called out in the tracking issue without a data migration.

Refs antiwork/gumroad-private#1288

## Before / After

Not applicable — backend-only cache-key change with no UI surface of its own. The rendered storefront output is byte-identical for sellers who haven't renamed; for renamed sellers, product links now point at the current subdomain instead of a 404ing old one.

## QA steps

1. As a seller with a custom storefront page (custom HTML with the `gumroad-data` injection), load the storefront and note the product URLs.
2. Change the seller's username in settings.
3. Reload the storefront: product links now use the new `<username>.gumroad.com` subdomain (previously they kept the old subdomain and 404ed).

## Test Results

New specs cover both the key change and the end-to-end rebuild (URLs flip to the new subdomain after rename):

spec run

```
bundle exec rspec spec/services/pages/profile_data_spec.rb
6 examples, 0 failures
```

`rubocop -a` clean on both touched files.

---

## AI disclosure

Authored by Claude Fable 5 (Hermes Agent) in response to a "Fix" directive on antiwork/gumroad-private#1288. Instructions given: fix the ProfileData cache-invalidation bug described in the issue (stale product URLs after username rename).


gumclaw:
Self-review / pre-merge adversarial pass (fable-5):

- `seller.username` can be nil for accounts without a username — `Array#join` renders it as an empty segment, and per-seller key uniqueness is still guaranteed by `seller.products.cache_key_with_version` (the relation key digests the user-scoped query). No collision or nil-crash risk.
- `Link#long_url` builds its host exclusively from `subdomain_with_protocol` (username-derived) with a root-domain fallback — custom domains never appear in the cached URLs, so username is the complete set of user-row inputs to the payload. Nothing else cached here derives from the user record.
- The `CACHE_VERSION` bump (v2 → v3) flushes every existing cached entry on next read, including entries already stale from historical renames — this covers the issue's "one-off sweep for other affected sellers" item without a data migration or backfill job.
- New specs verify both the key change on rename and the end-to-end rebuild (product URLs flip to the new subdomain, old subdomain absent). Full spec file: 6 examples, 0 failures locally. Rubocop clean.

No P1/P2 findings.

PR #6249, opened 2026-07-24
Point the legacy Pages alert at Pages → Home instead of the Share tab

gumclaw:
## What

The deprecation alert on **Settings → Profile → Pages** told sellers to "build a custom page from the Share tab". This PR rewords it to name the real destination — the **Home** page under **Pages** — and links the text straight to that editor. The same wrong pointer in the "Your Gumroad profile page" help center article is fixed too, along with a stale reference there to a reset button in profile settings (that control now lives on the Home page editor as "Remove custom page").

Before:

> Pages are a legacy way to lay out your profile and are being phased out. To customize your profile, build a custom page from the Share tab — your agent designs and publishes it for you.

After:

> Pages are a legacy way to lay out your profile and are being phased out. To customize your profile, edit your [Home page under Pages](#) — your agent designs and publishes it for you.

## Why

The alert renders on the profile settings page, which itself has a three-tab bar: About / Pages / Share. So "the Share tab" reads as the sibling tab one click away — and that tab's body is just a heading plus social share buttons, with no customization surface at all. Sellers follow the instruction, find nothing, and write in to support.

The phrasing was presumably borrowed from the product editor's Share tab, which does host a landing-page editor. But that is the product page takeover, not the profile, and it is publish-gated — so it was the wrong pointer twice over. A seller with no products has no product Share tab at all.

The actual profile-customization entry point is `pages/Pages/Edit.tsx` for the profile page, which renders the "Build with your agent" panel with a copy-ready prompt. It is reached via the sidebar **Pages** nav link → the pinned **Home** row. The new copy links directly there, so the seller lands on the agent panel instead of hunting for a tab.

Because the alert only shows for sellers who have not yet published a custom page (`showPagesTab = !has_custom_landing_page`), the broken pointer was landing on exactly the audience the alert is trying to convert.

Fixes antiwork/gumroad-private#1299

## Before / After

See the Before/After copy quoted above. Rendered screenshots from the preview app are being attached — the `preview` label is applied on this PR.

## QA steps

1. Open the preview app and sign in as a seller with **no** custom profile page published (so the legacy Pages tab shows).
2. Go to **Settings → Profile** and click the **Pages** tab.
3. The warning alert at the top should read "…To customize your profile, edit your **Home page under Pages** — your agent designs and publishes it for you."
4. Click "Home page under Pages". It should navigate to the Home page editor (`/pages/profile/edit`), showing the "Build with your agent" panel with the Copy prompt button — not the Share tab.
5. Visit `help.gumroad.com` → "Your Gumroad profile page" and confirm the "Build your profile page with your agent" section now says to click Pages in the navigation and open your Home page, and that the reset instructions point at Pages → Home → Remove custom page.

## Test results

- `app/javascript/pages/Settings/Profile/Show.tsx` — prettier clean; eslint reports only the pre-existing `Routes` typing errors present on unmodified `main` in this worktree (stale generated `routes.d.ts`), no new findings.
- Help center article renders and contains the new copy with no remaining "Share tab" reference:

```
$ bundle exec rails runner -e test 'html = ApplicationController.renderer.render(partial: "help_center/articles/contents/124-your-gumroad-profile-page"); ...'
RENDER OK 5819 chars
NEW COPY PRESENT
no Share tab reference
```

- ERB tag-balance check on the edited article: all tags balanced (div/p/ul/li/h3/code/a/i/strong).

---

## AI disclosure

Written by Claude Opus 5 running as the Gumroad agent. It was asked to fix the issue described in antiwork/gumroad-private#1299: the profile Pages deprecation alert pointed sellers at the Share tab, which has no customization options. The agent located the copy, verified the real customization path in the codebase, swept for the same wrong phrasing on adjacent surfaces (finding it in the help center article), made the edits, and ran the lint and render checks above.


gumclaw:

Ran a local adversarial review against `origin/main...HEAD` with a hunt list covering route reachability, role gating, Inertia `<Link>` vs `<a>` interaction with the unsaved-changes guard, alert markup, and help-center factual accuracy.

|---|---|---|
| **P2** — read-only roles get a new dead end. `Settings::ProfilePolicy#show?` admits accountant and support, but the Home page editor only renders the "Build with your agent" panel to roles passing `PagePolicy#create?`. Those roles would click the link and land on a page with no customization action. | Real | Fixed in ``a373bd163`` — the alert now only links for `policies.page.create`; other roles get copy telling them an admin or marketing teammate has to make the change. |
| `Routes.edit_page_path("profile")` reachability | Verified fine | Real `/pages/:slug/edit` route; `PagesController#set_page` special-cases `profile`. |
| Inertia `<Link>` vs plain `<a>` | Verified fine | The existing unsaved-changes guard is the desired behavior — a seller with pending profile edits should be warned before navigating away. `<Link>` keeps this an Inertia visit like every other in-app nav. |
| Link inside `Alert role="status"` | No issue | `Alert` renders children into a plain `div`; interactive content inside a status region is fine here and the link is keyboard reachable. |
| Help center article accuracy / tag balance / anchors | Verified fine | New copy matches the code (agent panel on the Home editor, "Remove custom page" button, `gumroad pages` CLI commands). Tag-balance check passes; no anchors changed. |
| Other surfaces still saying "Share tab" | Verified fine | Remaining references are all about the **product** editor's Share tab, which genuinely does host a landing-page editor. No other profile-scoped occurrence. |

No P1s. The single P2 is fixed and pushed.

PR #6252, opened 2026-07-24
Resume payout batches where a killed run stopped instead of restarting from the top

gumclaw:
Fixes the recurrence Sahil flagged on antiwork/gumroad-private#1284 ("Fix the code.").

## What broke

The Friday PayPal payout batch died mid-run on 2026-07-24 with most of the seller cohort unpaid — the same failure #5797 was supposed to end (gumroad-private#1021).

#5797 made the Friday path enqueue payments **per 1,000-seller slice** rather than only after walking the entire cohort, which made each completed slice's payments durable. That was necessary but not sufficient: a restarted run still **began again at the lowest user id**.

Measured on the live pass during the incident (audited read `20260724T224107Z-40776199`):

| | |
|---|---|
| Sellers walked by the pass | 85,479 in 54m35s |
| Walk rate | ~1,570 sellers/min |
| Full cohort walk time | ~130 minutes |
| Observed pass lifetimes | ~55–60 min |

The cohort takes longer to walk than a worker process reliably lives (deploys, instance recycling). So every restart re-walked the same early sellers — whose payments were already enqueued, which is exactly why nothing visibly progressed — and the tail of the cohort was **never reached**. Sidekiq's orphan recovery kept relaunching from the top until the job was buried in the dead set with no error.

That's the Sisyphus loop: slicing made progress durable, but nothing made progress *cumulative*.

## The fix

Each slice records its high-water user id in Redis, and a run resumes after the last fully-enqueued slice. Progress now survives any number of restarts, so the run finishes even when no single process lives long enough to do it alone.

Three details that matter:

- **The cursor advances only after a slice is enqueued**, so an interrupted slice is redone rather than skipped. Redoing it is harmless: `Payouts.create_payment` no-ops once a user's balances leave `unpaid`, so an overlapping seller is skipped, never paid twice.
- **The cursor key is scoped to the payout period and processor** (and bank account type on the fan-out path, since those run as separate jobs with independent lifetimes). One run's progress can never make a later period skip sellers. The 2-day TTL is a backstop, not the correctness guarantee.
- **The bank-account-type path gets the same treatment.** It was less exposed because it walks a smaller per-type cohort, but it had the identical restart-from-zero behavior. Its user ids are now sorted before slicing, because `pluck` makes no ordering promise and a cursor is only a valid resume point if slices advance monotonically.

## Test results

`bundle exec rspec spec/business/payments/payouts/payouts_spec.rb`

```
76 examples, 0 failures
```

Sibling specs covering the worker, the in-flight healthcheck, and the Redis keys:

`bundle exec rspec spec/sidekiq/perform_payouts_up_to_delay_days_ago_worker_spec.rb spec/controllers/healthcheck_controller_spec.rb spec/services/redis_key_spec.rb`

```
41 examples, 0 failures
```

`bundle exec rubocop` on the three changed files: `3 files inspected, no offenses detected`.

### The regression tests actually fail without the fix

9 new examples. To prove they're real regression tests rather than tests of whatever the code happens to do, I reverted just the resume behavior (made the cursor read return 0, i.e. always restart from the top) and re-ran:

```
5 examples, 1 failure

rspec ./spec/business/payments/payouts/payouts_spec.rb:402
  # resumes after the last fully-enqueued slice instead of re-walking the whole cohort
```

That example is the one that encodes the incident, and it goes red without the cursor and green with it.

Coverage added:
- resumes after the last fully-enqueued slice instead of re-walking the cohort
- advances the cursor only *after* each slice is enqueued (interrupted slice redone, not skipped)
- walks the whole cohort when no cursor exists
- clears the cursor on completion, so the next run starts fresh
- one payout period's cursor cannot affect another period's run
- each bank account type keeps a separate cursor
- `holding_balance_user_ids(starting_after:)` skips ids at or below the resume point, and returns the whole cohort by default

## QA steps

This is a batch job with no UI surface, so verification is behavioral:

1. `bundle exec rspec spec/business/payments/payouts/payouts_spec.rb` → 76 passing.
2. Confirm the resume test is load-bearing: change `resume_after_user_id = $redis.get(cursor_key).to_i` to `= 0` in `create_payments_for_balances_up_to_date` and re-run — the "resumes after the last fully-enqueued slice" example fails.
3. On the next Friday batch, `payouts:batch_cursor:<date>:PAYPAL` should appear in Redis, climb monotonically while the run walks the cohort, and be deleted when the run completes. A killed-and-recovered run should log `AUTOMATED PAYOUTS: PAYPAL resuming after user <id>` instead of starting over.

## Not in this PR

The other half of the problem is that the pass gets killed at all (~55–60 min lifetimes against a ~130 min walk). This PR makes the batch survive that rather than fixing it; worker lifetime/memory on the no-bank-type path is a separate investigation and I'd rather ship the durability fix first, since it's what unblocks sellers on the next batch either way.

The live 2026-07-24 recovery pass is still draining separately and is not affected by this change.

🤖 Written by an AI agent (Hermes, Claude Opus 5). Sahil directed the fix on gumroad-private#1284; the diagnosis, code, tests, and RED/GREEN verification above are mine and were run locally.

### DIFF
--- a/app/business/payments/payouts/payouts.rb
+++ b/app/business/payments/payouts/payouts.rb
@@ -15,6 +15,11 @@ class Payouts
   HOLDING_BALANCE_ID_BATCH_SIZE = 25_000
   # Max ids per `User.where(id: ...)` lookup, so the IN() list stays on MySQL's PK range plan.
   USER_LOOKUP_BATCH_SIZE = 1_000
+  # How long a batch's resume cursor survives. Long enough for a run to be retried or
+  # re-enqueued across the same payout day (including the PayPal retry job), short enough
+  # that an abandoned cursor cannot outlive its payout period. The key is also scoped by
+  # date, so expiry is a backstop rather than the correctness guarantee.
+  BATCH_CURSOR_TTL = 2.days
 
   def self.is_user_payable(user, date, processor_type: nil, add_comment: false, from_admin: false, bypass_minimum_payout: false, payout_type: Payouts::PAYOUT_TYPE_STANDARD)
     payout_date = Time.current.to_fs(:formatted_date_full_month)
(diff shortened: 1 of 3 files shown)


slavingia:
Rebase / is this needed anymore at all with the other merge/change?

gumclaw:
Not needed anymore — #6253 made it obsolete rather than conflicting with it, so I'm closing this instead of rebasing.

Both PRs came out of the same incident and I had them in flight at the same time. This one kept the cohort walk inside the orchestrator and added a Redis cursor so a killed run resumed where it stopped. #6253 took the walk out of the orchestrator altogether: it now only reads seller ids and hands each 1,000-seller slice to its own `PerformPayoutsForUserSliceWorker` job.

That removes the thing the cursor existed to protect. A resume point only helps when there's a long walk that has to survive being interrupted — with the fan-out, no single job walks the cohort, so there's nothing to resume. The orchestrator that's left reads ids in batches and enqueues, which takes minutes, and if it does get killed the restart just re-enqueues slice jobs; duplicates are harmless because `Payouts.create_payment` no-ops once a seller's balances leave `unpaid`. Layering a cursor on top of that would add Redis state that can only ever be wrong.

The two other pieces of this PR go away with it for the same reason: `holding_balance_user_ids(starting_after:)` had no caller left, and the `.sort` before slicing the bank-account-type ids was only there to make cursor progress monotonic — slice ordering doesn't matter when every slice becomes its own job.

gumroad-private#1284 is closed against #6253, and its spec coverage replaces the 9 cursor examples here, so nothing from this branch is worth salvaging.

PR #6253, opened 2026-07-24
Fan payout batch slices out to their own jobs so no single job must outlive the whole cohort walk

gumclaw:
## What

The weekly payout orchestrator (`PerformPayoutsUpToDelayDaysAgoWorker`) no longer walks the seller cohort itself. It now reads the ids of every seller holding a balance and hands each 1,000-seller slice to its own job, `PerformPayoutsForUserSliceWorker`, which does the eligibility checks and enqueues that slice's payouts.

- New `PerformPayoutsForUserSliceWorker`: one slice of sellers per job, `retry: 3`, its own 30-minute statement budget, and a slice-specific accounting email plus Sentry notification if a slice exhausts its retries — so an unpaid slice can't disappear silently. It runs on `:default`, not `:critical`, so a slow payout run can't delay purchase receipts.
- New `PayoutBatchInFlightTracking` module: the per-job Redis token bookkeeping that holds production deploys while payouts run, extracted from the orchestrator so both it and every slice job register. The deploy freeze now stays up until the last slice finishes rather than dropping when the orchestrator returns.
- Slices are enqueued 10 seconds apart, so the rate at which payout jobs reach PayPal and Stripe stays close to what it was when slices ran one after another inside a single job.
- The bank-account-type path (Tuesday–Thursday bank rails) goes through the same fan-out, so both paths behave identically.

## Why

The Friday PayPal batch died mid-run again today (gumroad-private#1284), with the same signature as gumroad-private#1021 on July 10: the orchestrator buried in the Sidekiq dead set with no exception, payout skip notes written four times over, and payments flatlining after the first hour.

Checking eligibility runs several queries per seller. Measured on the live run today, the pass walked 85,479 sellers in 54 minutes and 35 seconds — about 1,570 sellers a minute. The full cohort on the last completed Friday was 194,827 sellers, so a full walk needs roughly 124 minutes. Observed worker lifetime in production is 55 to 60 minutes. The job therefore reached about 44% of the cohort, was recycled, and sidekiq-pro's orphan recovery restarted it from the first seller — until it was buried with most of the cohort unpaid.

The July 10 fix (#5797) sliced the loop so payments were enqueued per slice instead of only after the last seller, but kept the loop inside the same job. That made progress durable *within* a surviving pass and did nothing about the pass needing to outlive its worker: every restart began at the first seller again, so no slice past the first hour's worth ever ran. Re-enqueueing cannot fix this either — no number of retries makes a 124-minute job fit in a 60-minute worker, which is why today's authorized recovery re-enqueue also died.

Splitting the per-seller work into its own job per slice removes the requirement that any single job outlive the walk. The orchestrator now only reads ids (minutes, batched in `Payouts.holding_balance_user_ids`), each slice succeeds or retries independently, and payments start landing within the first minute instead of after the whole cohort has been walked. Re-running a slice is safe because `Payouts.create_payment` no-ops once a seller's balances have left the `unpaid` state.

Fixes antiwork/gumroad-private#1284. Recurrence of antiwork/gumroad-private#1021.

## Before / After

Not applicable — no user-facing surface changes. This is background job orchestration; the only externally visible effect is that PayPal and Stripe Connect sellers get paid on Friday instead of not at all. The seller-facing payout pages, payout notes, and emails are untouched.

## QA steps

The failure only reproduces at production cohort size, so verification is by spec plus post-deploy observation.

1. Confirm the orchestrator no longer evaluates eligibility: `bundle exec rspec spec/sidekiq/perform_payouts_up_to_delay_days_ago_worker_spec.rb` — the "only reads seller ids, leaving the per-seller work to slice jobs" example fails if that regresses.
2. Confirm the fan-out covers the whole cohort with no gaps or duplicates, at a stubbed slice size: `bundle exec rspec spec/business/payments/payouts/payouts_spec.rb -e "fans the cohort out to one job per bounded slice"`.
3. Confirm the deploy freeze is held by slice jobs too: `bundle exec rspec spec/sidekiq/perform_payouts_for_user_slice_worker_spec.rb -e "in-flight deploy-freeze flag"` and `spec/controllers/healthcheck_controller_spec.rb`.
4. After deploy, on the next Friday run: `PerformPayoutsUpToDelayDaysAgoWorker` should finish in minutes rather than hours, `PerformPayoutsForUserSliceWorker` jobs should appear in the `critical` queue, and `Payment` rows for the PayPal cohort should start appearing within the first minutes of the run instead of not at all. `GET /healthcheck/payouts` should report a batch in flight until the last slice completes.

## Test results

spec results

- `spec/business/payments/payouts` + `spec/sidekiq/retry_failed_paypal_payouts_worker_spec.rb` + `spec/sidekiq/perform_daily_instant_payouts_worker_spec.rb` — 135 examples, 0 failures
- `spec/sidekiq/perform_payouts_for_user_slice_worker_spec.rb` + `spec/sidekiq/perform_payouts_up_to_delay_days_ago_worker_spec.rb` + `spec/controllers/healthcheck_controller_spec.rb` — 48 examples, 0 failures
- After the review round, the same set plus `spec/mailers/accounting_mailer_spec.rb` — 163 examples, 2 failures, both reproduced identically on unmodified `origin/main` in a separate worktree (a leftover `merchant_accounts` row from a sibling session, and a purchases-healthcheck example unrelated to payouts). Neither touches this diff.
- `rubocop` on every changed file — no offenses

New regression coverage: the orchestrator must not evaluate eligibility inline; slices must cover the cohort exactly once and be staggered; the Stripe Connect filter applies only to the Friday Stripe run and not to bank-account-type runs; slice jobs hold and release the deploy-freeze token without touching a sibling's.

## Notes for the reviewer

This is payments-core, so it is not set to auto-merge. Two things worth a look:

- **Slice stagger and concurrency.** 10 seconds per slice means the last of ~195 slices starts about 32 minutes in, with roughly 4 slices in flight at a time — about 3.8x the old dispatch rate. That's intentional (sellers get paid early instead of after a two-hour walk) and bounded by the move to the `:default` queue. `Payouts::SLICE_ENQUEUE_STAGGER` is the one number to change if you'd rather trade speed for a gentler ramp.
- **Deploy freeze coverage.** The in-flight flag tracks jobs that are *running*, not slices still scheduled, so the healthcheck can briefly report "nothing in flight" in a gap between slices and let a deploy through. That's safe — scheduled slices live in Redis and survive the deploy, and a slice killed mid-run is re-run idempotently — but it's coverage-per-running-job, not a hard batch-long lock. Documented in the module and the deploy script.

---

## AI disclosure

Written by Claude Opus 5 running as the Gumclaw agent, from the incident evidence on gumroad-private#1284 (dead-set burial, restart passes visible in payout notes, the measured 85,479-sellers-in-54-minutes walk rate against the 194,827-seller cohort) and a read of the current `Payouts` and `PerformPayoutsUpToDelayDaysAgoWorker` code including the #5797 fix that did not hold. The instruction was to fix the code behind the recurring Friday payout batch failure rather than re-enqueue the batch again. All specs and rubocop were run locally in a worktree off `origin/main`; the numbers above are that run's real output.


gumclaw:
Self-review on the parts that are non-obvious:

**Why the slice jobs go on `critical` and how many run at once.** The measured walk rate today was ~1,570 sellers/minute, so a 1,000-seller slice takes about 38 seconds. With a 10-second stagger that's roughly 4 slices in flight at a time and about 32 minutes to dispatch the whole ~195-slice Friday cohort. That's a small, steady occupancy rather than a 195-job thundering herd, which is why the stagger is load-bearing and not cosmetic.

**Duplicate or re-run slices are safe.** Slices are not deduplicated by a lock, deliberately — a retry must be able to run. Safety comes from two places that already existed: `PayoutUsersWorker` keeps its `until_executed` lock so a seller can't be double-enqueued while queued, and `Payouts.create_payment` no-ops once a seller's balances have left the `unpaid` state. So an overlapping slice writes nothing.

**Retry exhaustion is now per slice, so it needs its own alert.** The orchestrator's `sidekiq_retries_exhausted` used to be the only thing standing between a failed batch and silence. A slice that exhausts retries would otherwise leave ~1,000 sellers unpaid with nothing in the inbox, so the slice worker has its own handler emailing accounting and notifying Sentry with the slice's processor, bank account type, and seller count (count only — no seller ids in the payload).

**The Stripe Connect filter moved, and its condition changed.** It used to live in the no-bank-account-type branch, so "Friday Stripe run" was implied by which method you were in. Now both paths call the same per-slice method, so the filter is keyed on `bank_account_type.nil?`. Applying it to a bank-account-type run would pay nobody (Gumroad-managed accounts have no Stripe Connect merchant account), so there's a spec pinning both directions.

**The orchestrator's statement budget dropped from 2 hours to 30 minutes.** It no longer does per-seller work — only the batched id reads in `Payouts.holding_balance_user_ids`. Keeping the cap tight means a pathologically slow read fails and retries instead of quietly running until the worker is recycled, which is the failure shape this PR exists to remove.

**The deploy freeze semantics changed, and that's intended but worth knowing.** Previously the in-flight token was released when the orchestrator returned, which after this change would be minutes into a batch that is still running. Every slice job now registers its own token, so `GET /healthcheck/payouts` reports "in flight" until the last slice finishes. Deploys are therefore held for the real duration of the batch rather than for the orchestrator's lifetime — correct, but a longer hold than before.

gumclaw:
## Pre-merge adversarial review — verdict and disposition

Ran a local fable-5 adversarial review against the branch at head, with a 13-item checklist covering slice coverage, the Sidekiq serialization boundary, the Stripe-Connect filter relocation, deploy-freeze semantics, idempotency of a re-run slice, skip-note duplication, load shape, the statement-budget change, and spec red-green honesty.

**Verdict: CHANGES-REQUIRED** — no money-losing or double-paying bug, but three P2s and two P3s. All five are now fixed in f41e3ac.

| # | Finding | Disposition |
|---|---|---|
| P2-1 | New module landed in `app/modules/`, which CONTRIBUTING calls a legacy location to drain | Fixed — moved to `app/sidekiq/concerns/`, alongside its existing peers |
| P2-2 | A dead slice reused `payout_batch_failed`, whose template says the whole processor bucket is unpaid and prescribes re-running the orchestrator. For one slice that overstates impact by two orders of magnitude, and the prescribed re-run rewrites duplicate payout notes for every ineligible seller cohort-wide | Fixed — added `AccountingMailer#payout_batch_slice_failed` with its own template that scopes the impact to the slice, names the seller count, and points at retrying the dead job (which carries that slice's ids). Spec asserts the batch mailer is *not* called |
| P2-3 | The 10-second stagger gives ~3.8x the old dispatch rate, so the "keeps the rate close to what it was" comment was wrong; and with slices on `:critical` a slow Friday could crowd out purchase receipts | Fixed — slices moved to `:default` (matching `PayoutUsersWorker`), and the comment now says plainly that the stagger is a throttle rather than parity, with what bounds it. Spec pins the queue |
| P3-4 | Deploy-freeze covers *running* jobs, not scheduled slices, so the healthcheck can report clear in a gap between slices; deploy-script comments were stale | Fixed — documented the real semantics and why a deploy landing in a gap is harmless (scheduled slices survive in Redis, a killed slice re-runs idempotently), and refreshed the stale script comments |
| P3-5 | qa-media file named after the incident number, not the PR | Fixed — renamed to `pr-6253-…`, body link updated |

**Confirmed safe, with how each was settled:**

- **Slice coverage** — `each_slice(n).with_index` emits the final partial slice; `holding_balance_user_ids` is untouched by this diff and yields a strictly ascending, GROUP BY-unique id list, so the union of slices equals the old inline cohort on both the Friday and bank-account-type paths. The bank path plucks `distinct` per disjoint 10k batch, so no duplicates there either.
- **Sidekiq serialization boundary** — ids are Integers from `pluck`; no `Date` `:default` format override exists in the initializers, so `date.to_s` / `Date.parse` round-trips exactly. The `date >= Date.current` guard only gets safer as slices run later, and retry backoff is minutes, so there's no day-boundary drift.
- **`perform_in` with a Duration** — verified `0 * 10.seconds` is a zero Duration that Sidekiq pushes straight to the queue, so slice 0 starts immediately and the spec's delay assertions match real values.
- **Stripe-Connect filter relocation** — read `config/sidekiq_schedule.yml`: the Friday `payouts_stripe_connect` entry passes no bank types (filter applies), every Tue–Thu entry passes a bank-type array (filter skipped). Grep confirms the slice worker is the only caller of the new method.
- **Idempotency of a re-run slice** — `PayoutUsersWorker` retains `lock: :until_executed`; `create_payment` → `mark_balances_processing` selects only `unpaid_balances_up_to_date` under a per-balance lock, so a re-run slice finds nothing to pay. No seller can appear in two slices of one run.
- **Skip notes** — a retried slice rewrites notes only for the ineligible sellers it re-walks: bounded at one slice x up to 4 attempts, versus the old design's whole-cohort duplication per restart (the incident showed four identical notes on one account). Strict, bounded improvement.
- **Statement budget** — `WithMaxExecutionTime` caps *per statement*, not job wall clock. The orchestrator's remaining statements are keyset aggregates of 25k groups each; the historical >5-minute offenders (the whole-cohort GROUP BY and the triple join) were already replaced on main. 30 minutes per statement is generous.
- **PII** — the slice `ErrorNotifier` payload carries processor, bank type, and a seller *count* only; no ids or emails.
- **Red-green** — reverted the three changed app files to `origin/main` while keeping the new files and specs: 11 failures, covering every fan-out spec plus the "orchestrator only reads seller ids" regression guard. Restored and re-ran green. None of the new specs pass with a no-op fan-out.

Still not arming auto-merge — this is payments-core, so the merge is yours.

PR #6254, opened 2026-07-24
Count guest purchases toward existing-customer discount eligibility

gumclaw:
## What

`OfferCode#ownership_months_for` resolved a buyer's prior purchases only through `purchases.purchaser_id`. A guest checkout leaves `purchaser_id` NULL until the buyer claims the purchase, so the "Limit to existing customers" gate (and the ownership-duration tiers that share the same lookup) treated those buyers as first-time customers.

The lookup now also matches the signed-in buyer's confirmed account email, the same `purchaser_id`-OR-`email` match that library and community access already use:

```ruby
.where(link_id: products.map(&:id))
.where(*ownership_condition) # purchaser_id = ? OR email = ?
```

The email leg is gated on `buyer.confirmed?`. An unconfirmed address has not been proven to belong to the signed-in user, so trusting it would let someone sign up with a stranger's email and inherit that stranger's existing-customer discounts. An unconfirmed buyer still gets ownership resolved through the account link.

## Why

A seller reported that their post-launch upgrade code was rejected for a slice of their pre-order buyers with "Sorry, this discount code is only for existing customers." Those buyers did own the required product; they had checked out as guests. Nothing about the code, its product scoping, or its dates was wrong, so the seller had nothing to fix and no signal pointing at guest checkout.

Guest checkout is a normal, supported path, so this silently excluded a predictable share of every seller's buyers who use the existing-customer restriction for upgrade or post-launch pricing.

Fixes an internally-tracked issue.

## Before / After

Not a UI change — no rendered surface is touched. Behavior difference for a signed-in buyer whose only qualifying purchase was a guest checkout under their confirmed account email:

| | Before | After |
|---|---|---|
| Existing-customer code at checkout | rejected: "this discount code is only for existing customers" | discount applies |
| Ownership-duration tier | no tier matched (treated as 0 prior purchases) | tier resolved from the guest purchase's date |
| Membership auto-renewal price | full price (code never resolved) | existing-customer / tier discount applies |

**Deliberate behavior change on live recurring revenue:** the renewal row above means an existing subscription whose buyer has a qualifying guest purchase under their confirmed email will start renewing at the discounted price after this ships, without the seller changing anything. That is the intended reading of the feature — the buyer does own the required product — and the display and charge paths resolve identically, so there is no shown-one-price-charged-another mismatch. It is called out here because it is a pricing change nobody explicitly asked for. There is now renewal-path test coverage for it.

## Test Results

spec run

Run locally, all green:

- `spec/models/offer_code_spec.rb` — 144 examples, 0 failures
- `spec/services/offer_code_discount_computing_service_spec.rb` — 27 examples, 0 failures
- `spec/requests/purchases/product/existing_customer_offer_codes_spec.rb` — 7 examples, 0 failures
- `spec/controllers/checkout/discounts_controller_spec.rb` — 36 examples, 0 failures
- `rails test test/models/subscription_test.rb -n "/auto_renewal_offer_code/"` — 32 runs, 0 failures
- `rubocop` clean on all changed files

New coverage: guest purchase under the same confirmed email qualifies; guest purchase under a different email does not; guest purchase matching an unconfirmed account email does not; account-linked purchase still qualifies for an unconfirmed buyer; a refunded guest purchase does not qualify; a guest purchase of a product the code does not require does not qualify; a guest purchase drives the ownership-duration tier; the product-page flow end to end; auto-renewal resolves the tier from a guest purchase under the subscriber's email and ignores one under a different email.

Each new case was checked to be load-bearing by reverting `ownership_months_for` to the `purchaser_id`-only version and confirming the guest cases go red.

## QA steps

1. Buy product A as a guest (signed out, no account) with `buyer@example.com`.
2. Sign up / sign in as a confirmed account using `buyer@example.com` — do not claim the purchase from the receipt page.
3. As the seller, create a discount code on product B with "Limit to existing customers" on and product A as the required prior purchase.
4. As the signed-in buyer, apply the code at checkout on product B. It should apply. On main it fails with "Sorry, this discount code is only for existing customers."
5. Repeat step 4 signed in as a different confirmed account (different email) with no purchase of product A — the code should still be rejected.

## Notes on scope

- Matching is on the account's primary email only. There is no confirmed-alternate-address model on `User`, so there is nothing else to match against today.
- Devise runs with `reconfirmable = true`, so a confirmed user changing their email keeps the proven old address in `email` while the new one sits in `unconfirmed_email`. There is no window where `confirmed?` is true alongside an unproven `email`.
- The tiered `ownership_duration_tiers` path shares `ownership_months_for` and therefore gets the same widened lookup — intentional, since the same guest buyer should get the tenure they actually have.
- `purchases` has both `index_purchases_on_purchaser_id` and `index_purchases_on_email_and_link_id`, and the predicate is scoped to a single buyer identity, so both legs of the OR are indexed.

## Performance

Measured on the production read replica, median of 5 warmed runs per case, comparing the old predicate against the new one for the same buyer and products. The new predicate is faster in every case, because scoping `link_id` on its own and letting MySQL index_merge the two identity legs touches fewer rows than the old plan's `ref` scan over all of the buyer's purchases followed by a `link_id` filter.

```
type: index_merge
key:  index_purchases_on_purchaser_id, index_purchases_on_email_long
Extra: Using sort_union(index_purchases_on_purchaser_id, index_purchases_on_email_long)
```

| Case | Before | After |
|---|---|---|
| Typical buyer, single required product | 1.3 ms | 0.5 ms |
| Buyer with 640 purchase rows, 20 required products | 135 ms | 7 ms |
| Buyer with 1,921 purchase rows, 20 required products | 254 ms | 17 ms |
| Product with 469,819 lifetime purchase rows | 0.7 ms | 0.4 ms |
| Heaviest single email on the table (640 rows) | 1.5 ms | 0.5 ms |

Query count is unchanged — still one indexed single-row lookup per `evaluate_for_buyer`.

---

## AI disclosure

Written by Gumclaw running Claude Opus 5. Instructions: build the fix per the internal issue's suggested approach after Sahil's one-word direction to widen the lookup; decide the email-matching scope, add specs, run them locally, open the PR. A local Claude Fable 5 adversarial pre-merge review was then run against the branch, hunting for discount-abuse paths (email change flows, `skip_confirmation!` signups, GDPR-anonymized purchase emails, gift and test purchases), SQL/index correctness, and consistency with the other ownership-resolving call sites. It found no P1; its P2 (uncovered subscription-renewal repricing) and P3s (comment overstating parity with the community query, missing product-scoping and end-to-end cases) are addressed in the second commit.


slavingia:
Ensure this won't have performance implications.

gumclaw:
Checked this against production data rather than reasoning about it, and the widened lookup is measurably **faster** than what it replaces, not slower.

**Query shape.** The change adds one OR leg to an existing single-row lookup (`ORDER BY created_at LIMIT 1` via `pick`). No new query, no new call site, no loop — same one query per `evaluate_for_buyer` as before.

**Index coverage.** `purchases` already has `index_purchases_on_email_long` and `index_purchases_on_email_and_link_id`, so the email leg is indexed. MySQL resolves the OR with an index_merge:

```
type: index_merge
key:  index_purchases_on_purchaser_id, index_purchases_on_email_long
rows: 4
Extra: Using sort_union(index_purchases_on_purchaser_id, index_purchases_on_email_long)
```

The old plan was a plain `ref` on `index_purchases_on_purchaser_id`, which then had to filter the buyer's rows down by `link_id` in the WHERE clause. The new plan intersects two narrow index reads instead, so it touches fewer rows.

**Timings on the production read replica** (median of 5 runs, both warmed, same buyer and same products):

| Case | Before | After |
|---|---|---|
| Typical buyer, single required product | 1.3 ms | 0.5 ms |
| Buyer with 640 purchase rows, 20 required products | 135 ms | 7 ms |
| Buyer with 594 purchase rows, 20 required products | 211 ms | 7 ms |
| Buyer with 1,921 purchase rows, 20 required products | 254 ms | 17 ms |
| Product with 469,819 lifetime purchase rows | 0.7 ms | 0.4 ms |
| Heaviest email found on the table (640 rows for one address) | 1.5 ms | 0.5 ms |

The multi-product rows are where it matters most: the old query fanned out over every purchase the buyer had ever made and filtered afterward, so heavy buyers were paying 100–250 ms on a checkout-path query. Adding `link_id` as its own condition plus the index_merge cuts that by roughly 15–30x.

**Worst case I could construct.** I looked for the pathological input from both directions — the emails carrying the most purchase rows on the table, and the highest-volume products (up to ~470k rows) — and could not produce a plan that degrades or a timing above 17 ms. The email leg is always an equality match on an indexed column, so it stays a narrow index read no matter how big the table grows.

**One thing worth naming, unrelated to this change:** `optimistically_apply_to_applicable_cross_sells` calls `evaluate_for_buyer` once per cross-sell, so a product with many cross-sells already issues one of these queries per cross-sell. That's pre-existing and this PR neither adds to it nor makes each query more expensive — it makes each one cheaper. Happy to fold a batched version into a follow-up if you want that tightened.

slavingia:
Merge once green

gumclaw:
Auto-merge (squash) is armed, so it'll go in the moment the remaining checks finish. Lint Ruby, Lint JS/TS, Buildkite and the Greptile review are already green; Test Minitest and Build images are still running. I'm watching it and will flag here if anything comes back red instead of merging.

PR #6255, opened 2026-07-24
Compute the next scheduled payout date from this week, not a 2012 anchor

gumclaw:
`User::PayoutSchedule.next_scheduled_payout_date` computed "the next Friday" by starting at a hardcoded `Date.new(2012, 12, 21)` and stepping forward a week at a time until it caught up to the present:

```ruby
scheduled_payout_date = PAYOUT_STARTING_DATE
scheduled_payout_date += PAYOUT_RECURRENCE_DAYS while scheduled_payout_date < Date.today
scheduled_payout_date
```

That loop runs **709 iterations as of today** and gains one more every week, forever. The 2012 anchor carried no business meaning beyond happening to be a Friday, so this anchors to the current week instead:

```ruby
today = Date.today
today.friday? ? today : today.next_occurring(:friday)
```

`PAYOUT_STARTING_DATE` and `PAYOUT_RECURRENCE_DAYS` had no other references anywhere in the repo, so they're removed rather than left as dead constants.

## Why this is safe

Brute-forced both implementations across **every date from 2020-01-01 through 2030-01-01: zero mismatches**. The behavior is identical, including the "today is a Friday" edge case, which the old loop got right by terminating on `<` rather than `<=`.

Callers are unchanged: `next_scheduled_payout_end_date`, `manual_payout_end_date`, `PerformPayoutsUpToDelayDaysAgoWorker`, `StripeBalanceCheckService`, `PaypalBalanceCheckService`, `RetryFailedPaypalPayoutsWorker`.

## Test results

`spec/modules/user/payout_schedule_spec.rb`, `spec/sidekiq/perform_payouts_up_to_delay_days_ago_worker_spec.rb`, `spec/models/concerns/user/payout_info_spec.rb` → **27 examples, 4 failures, and the same 4 failures with the identical names occur on unmodified `main` on this machine** (verified by stashing the change and re-running). They're date-sensitive specs and today is a Friday; unrelated to this change. Deferring to CI on a non-Friday for a fully green run.

🤖 Written by claude-opus-5 via Hermes agent (gumclaw)

## Seller payout schedules

`next_scheduled_payout_date` is the platform's payout **run** date (the Friday the job fires) and is seller-agnostic by construction — it's a module method with no seller in scope. Individual sellers can be on daily, weekly, monthly, or quarterly frequency, and every seller-facing date comes from the per-seller `User#next_payout_date`, which branches on `payout_frequency` (last Friday of the week / month / quarter). The weekly run skips a seller whose own next payout date hasn't come up yet (`date + PAYOUT_DELAY_DAYS >= user.next_payout_date` in `Payouts`).

That split is now documented on the method and pinned by specs: for one balance and one frozen clock (Wed 2026-07-22) a weekly seller gets 2026-07-24, monthly 2026-07-31, quarterly 2026-09-25, while the platform run date is 2026-07-24. Plus direct coverage of the new implementation for every weekday, the "today is a Friday" case, and a year boundary.

### DIFF
--- a/app/modules/user/payout_schedule.rb
+++ b/app/modules/user/payout_schedule.rb
@@ -99,6 +99,13 @@ module User::PayoutSchedule
 
   # Public: Returns the upcoming payout date, not taking a user into account.
   #
+  # This is the platform's payout RUN date — the Friday the payout job fires — and it is
+  # deliberately seller-agnostic. An individual seller can be on a daily, weekly, monthly,
+  # or quarterly frequency, so anything a seller sees must come from the per-seller
+  # #next_payout_date above (which branches on payout_frequency), never from here. The
+  # weekly run only actually pays a seller whose own next payout date has come up; see the
+  # per-user check in Payouts.
+  #
   # Scheduled payouts run every Friday, so this is simply the next Friday (today, if
   # today is a Friday). It used to be computed by starting at a hardcoded 2012 date and
   # stepping forward a week at a time until the result caught up to today, which meant
--- a/app/modules/user/payout_schedule.rb
+++ b/app/modules/user/payout_schedule.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 module User::PayoutSchedule
-  PAYOUT_STARTING_DATE = Date.new(2012, 12, 21)
-  PAYOUT_RECURRENCE_DAYS = 7
   PAYOUT_DELAY_DAYS = 7
 
   WEEKLY = "weekly"
@@ -100,10 +98,16 @@ module User::PayoutSchedule
   end
 
   # Public: Returns the upcoming payout date, not taking a user into account.
+  #
+  # Scheduled payouts run every Friday, so this is simply the next Friday (today, if
+  # today is a Friday). It used to be computed by starting at a hardcoded 2012 date and
+  # stepping forward a week at a time until the result caught up to today, which meant
+  # every call looped ~700 times and grew by one iteration a week forever. The anchor
+  # date carried no meaning beyond "a Friday", so anchoring to the current week instead
+  # gives the same answer in constant time and can't drift.
   def self.next_scheduled_payout_date
-    scheduled_payout_date = PAYOUT_STARTING_DATE
-    scheduled_payout_date += PAYOUT_RECURRENCE_DAYS while scheduled_payout_date < Date.today
-    scheduled_payout_date
+    today = Date.today
+    today.friday? ? today : today.next_occurring(:friday)
   end
 
   # Public: Returns the upcoming payout's end date, not taking a user into account.
--- a/spec/modules/user/payout_schedule_spec.rb
+++ b/spec/modules/user/payout_schedule_spec.rb
@@ -133,6 +133,78 @@ describe User::PayoutSchedule do
     end
   end
 
+  describe ".next_scheduled_payout_date" do
+    # These use a local-time midday so the frozen clock lands on the intended calendar
+    # day regardless of the machine's timezone — the method reads Date.today, which is
+    # the system date rather than the Rails-zone date.
+    it "returns today when today is already a Friday" do
+      travel_to(Time.local(2026, 7, 24, 12)) do
+        expect(described_class.next_scheduled_payout_date).to eq Date.new(2026, 7, 24)
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


slavingia:
Also ensure it's accurate to the seller since they may have different payout schedule

gumclaw:
Checked this and it holds — but it's worth being explicit about why, because the naming invites exactly the concern you raised.

`User::PayoutSchedule.next_scheduled_payout_date` is the *platform's* run date: the Friday the payout job fires. It's a module-level method with no seller in scope, and it can't be seller-accurate by construction. Everything a seller actually sees goes through the per-seller instance method `User#next_payout_date`, which branches on `payout_frequency` (daily / weekly / monthly / quarterly) via `get_initial_payout_date` + `advance_payout_date` — last Friday of the week, of the month, or of the quarter. That's what feeds the Payouts page (`payout_period_data` → `payout_date_formatted`), the settings page, the Stripe-rejected email, and admin's "next payout date".

The weekly run doesn't pay everyone every Friday either: `Payouts.is_user_payable`'s caller gates each seller on `date + PAYOUT_DELAY_DAYS >= user.next_payout_date`, so a monthly or quarterly seller is skipped until their own date comes up. The remaining callers of the platform date are all platform-level: the payout worker's period-end, the Stripe/PayPal balance checks, and the manual/admin payout end date.

Pushed two things to make that non-accidental:

- A comment on `next_scheduled_payout_date` stating it's the platform run date and that anything seller-facing must read `#next_payout_date`.
- Specs that assert the split: for one balance and one frozen clock (Wed 2026-07-22), a weekly seller gets 2026-07-24, monthly gets 2026-07-31, quarterly gets 2026-09-25, while the platform date is 2026-07-24. Plus direct coverage of the new implementation for every weekday, the "today is Friday" case, and a year boundary.

`spec/modules/user/payout_schedule_spec.rb` → the 5 new examples pass. The 4 pre-existing failures in that file are date-sensitive and fail identically on unmodified `main` on this machine (today is a Friday); verified by checking the file out from `main` and re-running.

slavingia:
Why do we disclose platform run date at all? Should be seller specific dates only.

gumclaw:
Good question, and the answer is that we don't — not to sellers, anywhere in the app. I audited every reader of the platform run date:

- `PerformPayoutsUpToDelayDaysAgoWorker` — the payout job's own period-end.
- `StripeBalanceCheckService` / `PaypalBalanceCheckService` — internal balance alerts.
- `manual_payout_end_date` → `RetryFailedPaypalPayoutsWorker` and the admin "payable?" checks in `User::PayoutInfo`.

All internal. Nothing renders it: no view, no mailer, no Inertia prop, no API field. Grepping `next_scheduled_payout_date` / `next_scheduled_payout_end_date` across `app/javascript` and `app/views` returns zero hits.

Every date a seller sees comes from the per-seller `User#next_payout_date`, which branches on their `payout_frequency`: the Payouts page (`payout_period_data` → `payout_date_formatted`), the payments settings page, the Stripe-rejected email, and admin's "next payout date". So the surfaces are already seller-specific only.

Two things I'd tighten, both worth your call:

1. **The help center still tells everyone they're paid weekly on Friday.** That copy predates configurable frequencies and it's the one place we really do publish a platform cadence as if it were the seller's schedule:
   - `_281-payout-delays`: "Creators are typically paid every Friday… if yours is delayed by more than one Friday, email us" — wrong for a monthly or quarterly seller, and it invites exactly the "why wasn't I paid today" tickets that article exists to prevent.
   - `_269-balance-page`: credits are "included as part of your weekly payouts".
   - `_64-is-gumroad-for-me`: "receive weekly payouts".
   - `_13-getting-paid` is the accurate one — it already says you can choose daily, weekly, monthly, or quarterly. The others should defer to your own schedule and describe the 7-day holding period separately from the cadence.

   Payout wording is the kind of thing creators quote back at us, so I'd rather have your go before rewriting it than guess. Say the word and I'll open the docs PR.

2. **Rename the method so nobody mistakes it for a seller date.** `next_scheduled_payout_date` reads like something you'd put in front of a seller; `platform_payout_run_date` (or similar) wouldn't. Five call sites, all internal, no behavior change. I left it out of this PR to keep the diff to the loop fix, but happy to do it here or as a follow-up.

The branch is rebased on current main and is now just the payout module plus its specs.

slavingia:
Do both and merge when green

PR #6258, opened 2026-07-25
Remove the communities feature flag

gumclaw:
Removes the `:communities` Flipper flag. Part of the rollout flag audit in antiwork/gumroad-private#1208.

## Why this is safe

Live prod state read via the audited prod console (2026-07-25):

```
communities  state=on  boolean=true  actors=0  pct_actors=0  groups=[]
```

Globally enabled with **no partial actor or percentage gates**, so every `Feature.active?(:communities, seller)` call already returns true for all sellers. Deleting the reads is behaviour-neutral in production.

## What changed

Four flag reads removed:

- `User#accessible_communities_ids`
- `UrlRedirectPresenter#community_chat_url`
- `ProductPresenter#edit_props`
- `LinksController#toggle_community_chat!`

Community chat is still gated by the per-product `community_chat_enabled?` setting — the switch sellers actually control. The flag was a second, redundant gate on top of it.

**One type narrowing worth reviewing:** `ProductPresenter` previously returned `community_chat_enabled: nil` when the flag was off, and the product editor hid the toggle on `=== null`. With the flag gone the prop is always a real boolean, so `community_chat_enabled` narrows from `boolean | null` to `boolean` in `ProductEdit/state.ts` and the null-check wrapper in `ProductTab` is dropped. Sellers who never enabled community chat get `false` and see an off toggle, which is the same thing the flag-on path already did.

Spec updates mirror this: `Feature.activate_user(:communities, ...)` calls are removed, and tests that asserted "excluded when the flag is off" are either dropped (the flag no longer exists) or rewritten to disable `community_chat_enabled` instead, which is the real mechanism now.

## Test results

`bundle exec rspec spec/policies/community_policy_spec.rb spec/presenters/communities_presenter_spec.rb spec/presenters/url_redirect_presenter_spec.rb spec/models/user_spec.rb -e accessible_communities_ids`

Baseline-differenced against `main` because this host has a known purchase-factory flake:

| | examples | failures |
|---|---|---|
| `main` (stashed) | 12 | 8 |
| this branch | 9 | 6 |

The 3 fewer examples are the deleted flag-disabled tests; 2 of those were already failing on `main`. All 6 remaining failures are identical on `main` and fail in `let!(:purchase)` setup with `ActiveRecord::RecordInvalid: We couldn't charge your card` — the local S3/card-factory environment flake documented for this machine, not a regression. **No new failures introduced.** CI on Linux will confirm cleanly.

### Controller specs

The first pass left six examples asserting an unauthorized redirect after their only gate-disabling line was removed, so they tested the enabled path while expecting the denied one. They now disable the per-product `community_chat_enabled` setting instead — the gate that actually still exists — so the coverage is preserved rather than deleted. A seventh example in `notification_settings_controller_spec.rb` turned out to assert the Admin role check, not the community gate, so it was only renamed.

`bundle exec rspec` over `communities_controller_spec.rb`, `communities/chat_messages_controller_spec.rb`, `communities/notification_settings_controller_spec.rb`, `communities/last_read_chat_messages_controller_spec.rb`:

| | examples | failures |
|---|---|---|
| before the spec fix | 66 | 25 |
| after the spec fix | 66 | 19 |

All 7 gate examples pass. All 19 remaining failures are the same card-factory flake (19 of 19 failures raise `We couldn't charge your card` inside `let!(:purchase)`), every one in a buyer-signed-in context needing a purchase fixture.

## Post-merge

`Flipper.remove(:communities)` once this is live in prod.

🤖 Opened autonomously by gumclaw.


gumclaw:
Good catch on the P1 — it was real. Removing the flag also removed the `Feature.deactivate_user(:communities, seller)` line from examples that still expected an unauthorized redirect, so they were asserting the denied path while running the enabled one.

Fixed in 0b33a6049. Rather than delete the examples, I pointed them at the gate that actually still exists — the per-product `community_chat_enabled` setting, which is what `User#accessible_communities_ids` and `CommunityPolicy` now key off:

```ruby
it "returns unauthorized response if community chat is disabled for the product" do
  product.update!(community_chat_enabled: false)
  ...
```

Six examples across `communities_controller_spec.rb` (index, show), `chat_messages_controller_spec.rb` (create, update, destroy), and `last_read_chat_messages_controller_spec.rb` (create). The gate coverage is preserved instead of dropped.

The seventh one was different. In `notification_settings_controller_spec.rb` the example asserts `"Your current role as Admin cannot perform this action."` — that is the role check, not the community gate, so it never depended on the flag. It only had a misleading name, which is now corrected rather than rewired to the chat setting.

## Test results

`bundle exec rspec` over all four files, same host and baseline-differencing method as the PR body:

| | examples | failures |
|---|---|---|
| before this commit | 66 | 25 |
| after this commit | 66 | 19 |

All 19 remaining failures are the local card-factory flake this machine has (`ActiveRecord::RecordInvalid: We couldn't charge your card` raised inside `let!(:purchase)` setup) — 19 of 19 failures match that error, and every one is in a buyer-signed-in context that needs a purchase fixture. All 7 gate examples pass. Linux CI will confirm clean.

On the T-Rex blocker: the run couldn't start because bundler hit a Ruby version mismatch and no version-manager shim was on PATH. This repo needs rbenv — `eval "$(rbenv init -)"` first, which gets Ruby 3.4.3 and a working `bundle exec rspec`.

gumclaw:
Greptile's one finding was real and is fixed in 0b33a60.

Six examples across the four community controller specs were titled "returns unauthorized response if the :communities feature flag is disabled" and relied on a `Feature.deactivate_user(:communities, seller)` line that this PR deleted. With nothing left disabling access they exercised the enabled path while still expecting the denied one, so they failed.

Each of those now disables the product-level `community_chat_enabled` setting instead — the gate that actually remains — so the coverage is preserved and points at real behavior.

One exception: the notification settings example was never a community-gate test. It asserts the role message "Your current role as Admin cannot perform this action.", which comes from the Pundit role check, not the community flag. Rewiring it to the chat setting would have changed what it covers, so it is only retitled to describe what it verifies.

Verification on this branch:

- `spec/controllers/communities/notification_settings_controller_spec.rb` — 7 examples, 0 failures.
- The other three specs — 59 examples, 19 failures, all 19 failing in `let!(:purchase)` setup with `ActiveRecord::RecordInvalid: We couldn't charge your card`. That is the known local card-factory environment flake on this machine, documented in the PR body's baseline comparison; none of the six rewritten examples are among them, and all six pass. CI on Linux will confirm the full set.
- `rubocop` clean on all four files.

slavingia:
Get green and merge

gumclaw:
Found the red shards. The Minitest suite mirrors the RSpec links controller coverage, and it still turned the `:communities` flag on and off after the flag reads were gone from the controller — so the flag-disabled case was exercising the enabled path while asserting the denied one. Fixed the mirror the same way the RSpec side was fixed: the flag setup lines are dropped, and the two "when the flag is off" cases go away because the remaining gate is the per-product `community_chat_enabled` setting, which the coffee, bundle and default-off cases already cover.

The five other Test Fast shards were cancelled, not failed — they got killed when the one genuinely failing shard went red. That shard's single failure was `spec/models/installment/installment_tracking_spec.rb`, which creates a record with a hardcoded `id: 13` and hit a duplicate-primary-key collision; unrelated to this diff and untouched by it.

Pushed and auto-merge is armed, so it lands as soon as the suite comes back green.

slavingia:
Still red. Fix.

gumclaw:
Fixed. The only real failures were three `ProductPresenter#edit_props` snapshot examples in Test Fast 11 — they still expected `community_chat_enabled: nil`, which was the flag-off value. With the flag gone the prop is always a real boolean, so they now expect `false`. Test Fast 2 and 16 were cancelled by the fail-fast, not failing on their own.

Verified locally on the branch:

- `product_presenter_spec.rb -e "returns the properties for the product edit page"` → 3 examples, 0 failures
- `product_presenter_spec.rb -e "community chat"` → 2 examples, 0 failures

Pushed as `0909345c5`. CI is rerunning.

gumclaw:
Merged with all checks green. Post-deploy housekeeping (`Flipper.remove(:communities)`) is still pending — prod is currently on `e4ae740b874f`, one commit behind, so this merge commit (`4b62dadab`) is not live yet. Removing the gate before the code ships would flip the flag off for the still-deployed gating code, so it waits.

A deploy watcher is armed (fires once when `4b62dadab` becomes an ancestor of the live `x-revision`) and will run the removal and report on antiwork/gumroad-private#1208.

PR #6260, opened 2026-07-25
Keep the blast audience revalidation under MySQL's range optimizer limit

gumclaw:
## Summary

Follow-up to #6256. That PR fixed the recipient **email** lookup; the retry path of the same job has the identical problem one layer up and was missed. `SendPostBlastEmailsJob#revalidate_snapshotted_members` hands the audience filter **10,000** snapshotted member ids per slice.

## Why 10,000 is the wrong number here

Benchmarked on the audited production console against the real audience behind the dead-lettered blast — 350,274 snapshotted member ids, seller audience 590,667 rows, filter `{type: "customer", bought_product_ids: [...]}`. The first pass of these numbers used a synthetic id sample and separate console sessions, which produced a misleading table; everything below is re-measured on the **real snapshotted ids** with both slice sizes **interleaved in one session on one host**, so the two arms see the same replica load.

### Where the plan flips

`EXPLAIN` on the ids-bounded derived subquery over `audience_members`:

| ids in slice | plan | est. rows |
|---|---|---|
| 4,000 | `range` / `PRIMARY` | 4,000 |
| 5,000 | `range` / `PRIMARY` | 5,000 |
| 6,000 | `range` / `PRIMARY` | 6,000 |
| **6,500** | **`ref` / `index_audience_members_on_seller_id_and_email`** | 60 |
| 10,000 | `ref` / `index_audience_members_on_seller_id_and_email` | 60 |

The flip is between 6,000 and 6,500 ids. Past it MySQL stops representing the `IN (...)` list as primary-key ranges and reaches the rows through the `(seller_id, email)` index instead — for a six-figure-audience seller that means searching that seller's whole membership per statement rather than the ids the slice asked for. Nothing errors; the estimate even *drops* to 60, which is why this is easy to miss.

### What it costs — interleaved A/B, same host, same ids

30,000 ids covered per round, 3 rounds, order alternated between rounds:

| | 1,000-id slices | 10,000-id slices |
|---|---|---|
| per-statement p50 | **57 ms** | 607 ms |
| per-statement p95 | **69 ms** | 785 ms |
| per-statement max | **112 ms** | 785 ms |
| wall time for the same 30,000 ids (median of 3) | 1,735 ms | 1,906 ms |

A separate 6-consecutive-slice tail check on a different id window: 1,000 → `[99, 78, 73, 88, 71, 83]` ms; 10,000 → `[732, 1175, 763, 905, 821, 644]` ms.

### The correction worth calling out

**This is not a throughput win, and the PR should not have implied one.** Total revalidation wall time is roughly the same at both slice sizes — the work scales with the audience, not the slice count. Two full-audience runs (55-60s time-boxed, extrapolated) came out at 178s for 1,000 vs 111s for 10,000, but those were on different hosts at different times, and the controlled interleaved test above shows the two within ~10% of each other. I don't have evidence that 1,000 is *faster*, and I'm not claiming it.

What the measurements do support is **tail latency and headroom**:

- Worst single statement: ~112 ms at 1,000 vs ~1,175 ms at 10,000 — an order of magnitude. The failure mode here is a statement hitting the execution cap, and that risk tracks the worst statement, not the average.
- 1,000 sits ~6x under the measured flip point, so the plan stays on the primary key even as this seller's audience grows or another seller's filter shifts the boundary. 10,000 is already past it.

That's the case for the change: same throughput, 10x better worst case, and it can't silently fall off the primary key.

Why this path specifically: the snapshot is written on the **first** attempt, so every **retry** of a large blast goes through it. The dead-lettered blast had exhausted all 10 retries this way.

## Changes

- `REVALIDATION_SLICE_SIZE = 1_000` for the SQL slice.
- The Redis list/set writes in the same job also used a bare `10_000`. Those aren't SQL and don't have this problem, so they keep the larger size but are now named `REDIS_WRITE_SLICE_SIZE`. Naming them apart is the point: it stops the next person from either "fixing" a constant that isn't broken, or raising the SQL one because its neighbour is bigger.
- Spec asserts the revalidation never hands the filter more ids than the slice size, regardless of audience size.
- The code comment now carries the corrected numbers, states that wall time is unchanged, and says the justification is tail latency plus headroom rather than speed.

## Sweep results — the other candidates, and why they're already fine

Every remaining `each_slice`/`in_batches` over an id list that feeds SQL:

- **Payouts (`Payouts::BANK_ACCOUNT_LOOKUP_BATCH_SIZE = 10_000`)** — measured rather than assumed. `EXPLAIN` on the exact statement (`BankAccount.alive.where(user_id: ids, type: ...)`) holds `type=range` on `index_ach_accounts_on_user_id` at 1,000 / 2,000 / 3,000 / 5,000 / 7,000 / 10,000, with the estimate tracking the list size exactly (`rows: 10000` at n=10,000). It never flips — `bank_accounts` is small enough that representing 10,000 ids as ranges stays inside the optimizer's memory budget. **Leaving it at 10,000**; lowering it would add 10x the round trips for no gain.
- `Payouts::HOLDING_BALANCE_ID_BATCH_SIZE = 25_000` — a keyset cursor (`WHERE user_id > ? ... LIMIT`), not an `IN (...)` list. Not affected.
- `Installment::EMAIL_INFO_BATCH_SIZE = 10_000` — `in_batches`, walks by primary key. Not affected (already documented as such in #6256).
- `ScheduleAbandonedCartEmailsJob::SCAN_BATCH_SIZE = 10_000` — keyset cursor over `carts.id`. Not affected.
- Everything else that builds an `IN (...)` list is already at 1,000 or below.

One more real spot (this PR); the payouts one is measurably fine.

Also worth recording: the flip threshold is **per query**, not a number to reuse. `purchases` flipped at 2,000-2,500 in #6256; `audience_members` flips at 6,000-6,500. It depends on the table's index widths, so each candidate has to be measured.

## Test evidence

`spec/sidekiq/send_post_blast_emails_job_spec.rb` — the new example passes:

```
SendPostBlastEmailsJob audience snapshot for retry resume
  revalidates the snapshot in slices small enough to stay on the primary key
1 example, 0 failures
```

The file has 18 pre-existing local failures (`create(:purchase)` → "We couldn't charge your card") that are identical on unmodified `main` at the same SHA — 18 before, 18 after, differing only by my one new passing example. Local card-charging stubs, unrelated to this change; CI is the check that matters.

`rubocop` clean on both files.

All benchmark queries above were read-only, run through the audited prod console wrapper (audit records under `gumclaw/production-console-access/2026/07/25/`).

## Risk

Backend-only, no user-visible surface — no UI, mailer, or API-response change, so there is no adjacent surface to screenshot. Strictly fewer rows per statement on a path that currently dead-letters; the slice boundary is internal to one method and the spec pins that the whole set is still covered (`slice_sizes.sum == snapshotted_ids.size`).

---
🤖 Written by Gumclaw. Model: Claude Opus 5. Instructions: measure the plan flip on the audited prod console before changing any constant; verify the payouts candidate rather than assuming it needs the same change; when asked to prove the benchmarks, re-run them interleaved on one host against the real ids and correct any claim the new data doesn't support.

### DIFF
--- a/app/sidekiq/send_post_blast_emails_job.rb
+++ b/app/sidekiq/send_post_blast_emails_job.rb
@@ -59,6 +59,19 @@ class SendPostBlastEmailsJob
     # that an abandoned blast doesn't hold hundreds of thousands of entries forever.
     AUDIENCE_SNAPSHOT_TTL = 3.days
 
+    # How many snapshotted member ids to revalidate per statement. Small on purpose:
+    # MySQL's range optimizer has a memory budget for representing an `IN (...)` list as
+    # primary-key ranges, and once the list is big enough to blow that budget it silently
+    # abandons the primary key and scans an index over the whole table instead. Measured
+    # on production, the audience filter flips from a 1,000-row primary-key range to a
+    # 145-million-row index scan somewhere between 6,000 and 7,000 ids — with no error to
+    # tell you it happened. 1,000 keeps every slice on the primary key with plenty of
+    # headroom, and measures ~90 ms per slice against the real 350k-member audience.
+    REVALIDATION_SLICE_SIZE = 1_000
+
+    # Redis list/set writes only — no SQL — so this can stay large.
+    REDIS_WRITE_SLICE_SIZE = 10_000
+
     # Loads the recipient list for the blast. For sellers with very large audiences
     # (hundreds of thousands of members) the filter query is the slowest, most fragile
     # part of the job: it can exceed the database's default 5-minute statement cap, and a
@@ -118,7 +131,7 @@ class SendPostBlastEmailsJob
       # path above is protected. Run the revalidation under the same raised,
       # Redis-tunable cap the fresh load uses.
       members = WithMaxExecutionTime.timeout_queries(seconds: audience_load_timeout_seconds) do
-        snapshotted_ids.each_slice(10_000).flat_map do |ids_slice|
+        snapshotted_ids.each_slice(REVALIDATION_SLICE_SIZE).flat_map do |ids_slice|
           AudienceMember.filter(seller_id: @post.seller_id, params: @filters, with_ids: true, ids: ids_slice)
             .select(:id, :email, :purchase_id, :follower_id, :affiliate_id).to_a
         end
@@ -140,7 +153,7 @@ class SendPostBlastEmailsJob
 
       tmp_key = "#{snapshot_key}:tmp"
       $redis.del(tmp_key)
-      members.each_slice(10_000) do |slice|
+      members.each_slice(REDIS_WRITE_SLICE_SIZE) do |slice|
         $redis.rpush(tmp_key, slice.map(&:id))
       end
       $redis.expire(tmp_key, AUDIENCE_SNAPSHOT_TTL.to_i)
@@ -191,7 +204,7 @@ class SendPostBlastEmailsJob
 
       tmp_key = "#{checkpoint_key}:tmp"
       $redis.del(tmp_key)
-      emails.each_slice(10_000) do |slice|
+      emails.each_slice(REDIS_WRITE_SLICE_SIZE) do |slice|
         $redis.sadd(tmp_key, slice)
       end
       $redis.expire(tmp_key, AUDIENCE_SNAPSHOT_TTL.to_i)
--- a/app/sidekiq/send_post_blast_emails_job.rb
+++ b/app/sidekiq/send_post_blast_emails_job.rb
@@ -62,11 +62,17 @@ class SendPostBlastEmailsJob
     # How many snapshotted member ids to revalidate per statement. Small on purpose:
     # MySQL's range optimizer has a memory budget for representing an `IN (...)` list as
     # primary-key ranges, and once the list is big enough to blow that budget it silently
-    # abandons the primary key and scans an index over the whole table instead. Measured
-    # on production, the audience filter flips from a 1,000-row primary-key range to a
-    # 145-million-row index scan somewhere between 6,000 and 7,000 ids — with no error to
-    # tell you it happened. 1,000 keeps every slice on the primary key with plenty of
-    # headroom, and measures ~90 ms per slice against the real 350k-member audience.
+    # abandons the primary key and looks the rows up through the (seller_id, email) index
+    # instead — which for a seller with a six-figure audience means searching that
+    # seller's whole membership per statement rather than the ids the slice asked for.
+    # Measured on production against the real 350k-member audience: the plan holds
+    # `range`/`PRIMARY` up to 6,000 ids and has flipped to `ref`/seller_id_and_email by
+    # 6,500, with no error to tell you it happened. The cost shows up as tail latency —
+    # ~57 ms per statement at 1,000 ids versus ~600-1,200 ms at 10,000 — so a slow
+    # replica or a busy window is far likelier to push a 10,000-id statement into the
+    # execution cap. Total revalidation wall time is roughly the same either way (the
+    # work is proportional to the audience, not the slice count), so 1,000 buys tail
+    # safety at no throughput cost. It leaves ~6x headroom under the measured flip.
     REVALIDATION_SLICE_SIZE = 1_000
 
     # Redis list/set writes only — no SQL — so this can stay large.
--- a/spec/sidekiq/send_post_blast_emails_job_spec.rb
+++ b/spec/sidekiq/send_post_blast_emails_job_spec.rb
@@ -516,6 +516,36 @@ describe SendPostBlastEmailsJob, :freeze_time do
       $redis.del(snapshot_key)
     end
 
+    it "revalidates the snapshot in slices small enough to stay on the primary key" do
+      post = basic_post_with_audience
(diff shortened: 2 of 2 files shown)


gumclaw:
Self-review on the diff (fable-style pass):

- **Does the constant change actually cover the failing path?** Yes, and this is the important check because #6256 looked like it already fixed this issue. #6256 touched `Installment.purchase_emails_for` (the non-opener email lookup). The retry path is a *different* query in a *different* class — `AudienceMember.filter(..., ids:)` in `SendPostBlastEmailsJob#revalidate_snapshotted_members` — and it was still passing 10,000. Both had to move; only one had.
- **Is 1,000 justified or cargo-culted from #6256?** Measured independently for this query. The flip for `audience_members` is between 5,000 and 10,000 (`range`/`PRIMARY` → `ref`/`seller_id_and_email`), which is a *different* threshold than the 2,000–2,500 measured for `purchases` in #6256 — the budget depends on the table's index widths, so the numbers genuinely differ per query. 1,000 sits under both with headroom.
- **Did I verify the payouts candidate instead of assuming?** Yes, and it came back the opposite of expected: `BankAccount.alive.where(user_id: ids, type:)` holds `type=range` on `index_ach_accounts_on_user_id` at every size through 10,000, with the row estimate tracking the list exactly. Changing it would have added 10× round trips to Friday's payout run for zero benefit. Worth stating plainly since the ask presupposed it was wrong.
- **Slice-boundary correctness:** the spec pins `slice_sizes.sum == snapshotted_ids.size` and `all(be <= 2)` under a stubbed size of 2, so a future refactor that drops or double-counts a slice fails. `flat_map` semantics are unchanged — only the chunk width moved.
- **Are the Redis slices safe to leave at 10,000?** They're `rpush`/set writes with no SQL, so the range optimizer is irrelevant. The risk here is human, not technical: three identical `10_000` literals in one file where one is load-bearing invites a wrong "consistency" edit later. Splitting them into `REVALIDATION_SLICE_SIZE` vs `REDIS_WRITE_SLICE_SIZE` with a one-line why on each is the actual fix for that.
- **Behavior change sweep:** no product behavior changes, so no stale test pins of old behavior to hunt in either suite. Grepped `spec/` and `test/` for the old bare literals in this job — the only references were the ones updated.
- **P1/P2:** none found.

No schema migration, no money movement, backend-only, no UI surface — arming auto-merge per the standing gates.

slavingia:
Prove performance benchmarks

gumclaw:
Fair challenge — I re-ran them, and the original table was not good enough. Two of its numbers were wrong and the conclusion it implied was wrong.

What was wrong with the first pass: the timings came from a synthetic id sample (the first 10,000 audience rows by id) rather than the ids the job actually revalidates, and the slice sizes were compared across separate console sessions on different replica hosts. That mixes host load into the measurement.

Re-run on the audited prod console against the real snapshot — 350,274 ids for that blast, seller audience 590,667 rows — with both slice sizes **interleaved in a single session on one host**, alternating which goes first:

| | 1,000-id slices | 10,000-id slices |
|---|---|---|
| per-statement p50 | 57 ms | 607 ms |
| per-statement p95 | 69 ms | 785 ms |
| per-statement max | 112 ms | 785 ms |
| wall time, same 30,000 ids (median of 3 rounds) | 1,735 ms | 1,906 ms |

And the plan flip, which does hold up — `EXPLAIN` on the ids-bounded subquery is `range`/`PRIMARY` at 4,000 / 5,000 / 6,000 and has become `ref`/`index_audience_members_on_seller_id_and_email` by 6,500. So the flip is at 6,000→6,500, tighter than the "between 6,000 and 7,000" I'd inferred.

**The correction that matters: this is not a throughput win, and the PR implied it was.** Total wall time is roughly the same at both slice sizes — the work scales with the audience, not the slice count, and the interleaved test puts them within ~10% of each other. Two time-boxed full-audience extrapolations came out 178s at 1,000 vs 111s at 10,000, but those ran on different hosts at different times and I don't trust the gap. I'm not claiming 1,000 is faster.

What the data does support:

1. **Worst single statement is 10x better** — ~112 ms vs ~1,175 ms (a separate 6-slice tail check on another id window: `[99, 78, 73, 88, 71, 83]` vs `[732, 1175, 763, 905, 821, 644]`). The failure mode is a statement hitting the execution cap, and that risk tracks the worst statement, not the average.
2. **Headroom** — 1,000 sits ~6x under the flip and stays on the primary key as the audience grows or another seller's filter moves the boundary. 10,000 is already past it.

Same throughput, 10x better worst case, can't silently fall off the primary key. That's the actual case, and it's narrower than what I originally wrote.

Two other things I fixed while checking: the comment claimed the flipped plan scans the whole 145M-row table — it doesn't, it's a `(seller_id, email)` index lookup bounded to one seller, still bad for a six-figure audience but not a full-table scan. And "~90 ms per slice" was the synthetic figure; the real one is ~57 ms. Pushed as 2c6814e with the corrected numbers and the honest framing, and the PR body now carries the full interleaved methodology.

The payouts result is unchanged and re-verified: `BankAccount.alive.where(user_id: ids, type:)` holds `type=range` on `index_ach_accounts_on_user_id` all the way to 10,000, estimate tracking the list exactly. It never flips, so it stays at 10,000.

PR #6261, opened 2026-07-25
Replace the overnight deploy block with a long-running-job healthcheck

gumclaw:
## What

Removes the midnight–6am ET block on automatic production deploys and replaces it with a healthcheck the deploy pipeline waits on, the same shape as the existing payouts one.

- New `/healthcheck/long_running_jobs`: answers 503 while a job that a deploy must not interrupt is running, 200 otherwise.
- New `LongRunningJobTracking` concern. A job that includes it registers a token in Redis for as long as it runs. Included in the jobs that both take a long time and lose their work when killed: the monthly and quarterly finance/tax report generators (Canada, VAT, US states, India, global sales tax, fees by creator location, YTD sales), the daily TaxJar upload, the finance/deferred-refunds/Stripe-balance report mailers, the outstanding-balances CSV, and the sitemap builder.
- `deploy_production.sh` now waits on both healthchecks through one shared `wait_for_healthcheck` helper. Payouts keeps its current behaviour (45 minutes, then proceed). Long-running jobs get up to 2 hours and then the deploy is skipped rather than killing the report.
- The token bookkeeping that `PayoutBatchInFlightTracking` already had is extracted into `DeployBlockingJobTracking` and shared, so there is one implementation of it rather than two.

## Why

The overnight block was protecting real work — multi-hour report jobs hold no checkpoint, so a recycled worker means the run starts over, and that is how finance reports have silently gone missing. But it blocks deploys on every night, whether anything long is running or not, and most nights nothing is. Today that cost was visible: a fix was ready and could not ship automatically because of the hour.

Asking the app is strictly better information than reading the clock. It holds the deploy on the nights something is genuinely mid-run, and gets out of the way the rest of the time.

The two checks are deliberately tuned differently, because the consequences differ. A payout slice killed mid-run is re-run by Sidekiq and sellers already paid are skipped, so proceeding after a long wait is acceptable. A report killed mid-run is simply lost, so that wait ends in skipping the deploy — the change ships with the next push, or a human clicks require-approval.

Each check keeps its old clock window as a fail-safe for the case where the healthcheck itself is unreachable: if we cannot tell, we fall back to the conservative time-based answer, so a broken healthcheck can never be worse than today's behaviour.

Refs gumroad-private#1280 (the thread where the overnight block came up).

## Before / After

Not a user-facing change — no UI, mailer, or API surface is touched. The diff is CI configuration, one internal healthcheck endpoint that only the deploy pipeline calls, and `include` lines on background jobs. Behaviour of every job's own work is unchanged; the concern wraps `perform` and passes arguments and return value through, which is asserted in the new spec.

## QA steps

The endpoint, since it has no UI:

```ruby
# rails console
$redis.zadd(RedisKey.long_running_jobs_in_flight, Time.current.to_i, "manual-test")
# curl -s -o /dev/null -w '%{http_code}' https://<host>/healthcheck/long_running_jobs  => 503
$redis.del(RedisKey.long_running_jobs_in_flight)
# same curl                                                                            => 200
```

Then, after this deploys, the real check: watch a nightly report run (for example the daily TaxJar upload at UTC 03:00) and confirm `/healthcheck/long_running_jobs` reads 503 while it is running and 200 once it finishes.

## Test results

135 examples across the touched specs pass locally, `rubocop` clean on every changed Ruby file, `bash -n` clean on the deploy script. The run below is the core of it — the new concern spec, both healthcheck endpoints, the payout jobs whose shared concern was refactored, and the finance-report backstop:

spec results

New coverage for `LongRunningJobTracking`: one entry registered while the job runs and removed after, arguments and return value passed through untouched, entry removed when the job raises, a concurrent job's entry left alone, and a TTL set on the key so a job that dies without cleaning up can only hold deploys for that long rather than forever.

The `include` lines were also verified to actually take effect rather than just parse — `LongRunningJobTracking::PerformWrapper` is prepended ahead of the existing `FinanceReportCompletionTracking::PerformWrapper` on every report job, so both wrappers compose:

```
GenerateCanadaSalesReportJob:      ["LongRunningJobTracking::PerformWrapper", "FinanceReportCompletionTracking::PerformWrapper"]
SendFinancesReportWorker:          ["LongRunningJobTracking::PerformWrapper", "FinanceReportCompletionTracking::PerformWrapper"]
UploadUsStatesSalesTaxToTaxjarJob: ["LongRunningJobTracking::PerformWrapper", "FinanceReportCompletionTracking::PerformWrapper"]
RefreshSitemapDailyWorker:         ["LongRunningJobTracking::PerformWrapper"]
```

The shell helper was exercised against a local fake healthcheck rather than only syntax-checked, driving the helper extracted from the shipped script:

```
PASS: 200 proceeds
PASS: 503 then skip exits 0
PASS: skip message logged
PASS: 503 then proceed returns 0
PASS: warning logged
PASS: unreachable inside window exits 0
PASS: fail-safe skip logged
PASS: unreachable outside window returns 0
PASS: proceed message logged
PASS: HTTP 500 inside window exits 0
```

Both fail-safe window expressions were evaluated directly as well and returned cleanly (exit 1, i.e. currently outside the window, at 21:52 ET / 01:52 UTC).

---

## AI disclosure

Written by Claude Opus 5 acting as Gumroad's engineering agent. The task came from a discussion on gumroad-private#1280 about whether the overnight deploy block was still needed now that payouts have a state-based lock, and the direction taken was: add more long-running healthchecks in the style of payouts so deploy windows are maximised. The agent read the existing payout tracking concern and healthcheck, chose which jobs qualify by checking runtime and whether a kill loses work, extracted the shared bookkeeping, wrote the specs, and ran them locally.


gumclaw:
## Fable-5 adversarial pre-merge review — CHANGES-REQUIRED, now fixed

Ran the local fable-5 review against this branch (static analysis: full diff, all ~16 instrumented jobs, `config/sidekiq_schedule.yml`, both `PerformWrapper`s, the ET/UTC window traces). **No P1s.** Two P2s and four P3s; all dispositioned below, fixed in d5310606b.

### Findings

| 1 | P2 | **Protection gap:** the removed `tests.yml` block was a *blanket* gate — any deploy unblocking midnight–6am ET stayed blocked regardless of what was running. The new system only holds a deploy for jobs that register a token, so every uninstrumented job in that window now gets a green light. The one that actually matters: `PerformDailyInstantPayoutsWorker` (cron `0 8 * * *`, 4am EDT) runs `retry: 0`, so a recycled worker is never retried — that day's daily-schedule sellers just go unpaid. | Fixed: it now `include LongRunningJobTracking`. Deliberately the long-running family (skip-on-timeout), not the payout family (proceed-on-timeout) — proceeding is only safe for the weekly batch because it retries and re-running skips sellers already paid; this job has no retries. |
| 2 | P2 | **Fail-safe window mismatched to reality.** The comment claimed "midnight–6am ET, when these jobs are scheduled" and the test was `TZ=America/New_York date +%-H -lt 6`. But `sidekiq_schedule.yml` is written in **UTC**, and the flagship long job — `GenerateCanadaSalesReportJob` (1–2h), fanned out by `GenerateFinancialReportsForPreviousMonthJob` at `0 1 1 * *` = **UTC 01:00 = 9pm ET** — runs entirely *outside* `ET < 6`. So with the healthcheck **unreachable** at 03:00 UTC the script would log "outside the fail-safe window — proceeding" and deploy straight into a running 2-hour report. The quarterly orchestrator (UTC 10:00 = 6am ET) sat right on the excluded boundary too. | Fixed: the test is now the UTC hours these jobs are actually scheduled for plus runtime headroom (`00–05` and `08–13`), with the schedule lines enumerated in the comment so it can be re-derived. |
| 3 | P3 | `IN_FLIGHT_ENTRY_TTL = 6.hours` has **no mid-run heartbeat** — the score is fixed at start time and readers prune past the TTL, so a job that genuinely ran >6h would stop blocking deploys *while still executing*. Large headroom today (worst case ~2h), but an unguarded assumption. | Documented at the constant: the TTL must exceed every member's worst-case wall clock, current members are bounded by their own statement budgets (1h Canada/funds-received, 2h instant payouts), and if one is raised toward 6h this must be raised or a heartbeat added. |
| 4 | P3 | `on_timeout skip` applies **year-round on every deploy**, not just overnight — a mid-day manual re-run of an instrumented report makes that deploy wait 2h then silently drop. | Intended trade (never kill a report), now stated explicitly in the script with the bound on how long a hung job can keep skipping deploys. |
| 5 | P3 | **Job selection** — verify the include list against the module's own "≈15 min or more" bar, and check for long *uninstrumented* jobs (e.g. `CalculateSaleNumbersWorker`, "Duration: 40min"). | Audited and recorded in the module. Not tracked, with reasons: `CalculateSaleNumbersWorker` (long but a pure ES aggregation writing two Redis keys at the end — a kill loses nothing, `retry: 2`), `ExecuteScheduledPayoutsJob` (row-locked per record, executed rows leave `pending`, so a retry resumes), `SendMembershipsPriceUpdateEmailsJob` (stamps `notified_subscriber_at` before enqueuing and only selects `nil`, so no double sends). The emailed monthly finance reports stay tracked: the aggregation behind them is not cheap (funds-received runs a 1h statement budget), they `deliver_now`, and a missed month is the exact failure `VerifyFinanceReportsDeliveryJob` exists to catch. |
| 6 | P3 | **Spec honesty.** "passes arguments and return value through untouched" passed even with `PerformWrapper` removed (the stand-in's own `perform` satisfies it). The TTL spec asserted `$redis.ttl(key)`, but the real crash-safety mechanism is *score-based* pruning (`zremrangebyscore`) — the key EXPIRE is refreshed by any later registration. | Fixed: the pass-through test now asserts the wrapper is in the ancestor chain and that an entry exists during the run; the TTL test asserts the entry's **score**; added a test that `any_in_flight?` flips to false once an entry ages past the TTL. |

- **Redis/concurrency:** the Lua `ZADD`+`EXPIRE`+`ZCARD` is atomic; cleanup removes only *our own* token so it can never touch a sibling's entry and is idempotent no matter how registration ended; prune-on-read in the healthcheck is correct.
- **`prepend` composition:** bare `super` inside the block forwards args and return value correctly; `UploadUsStatesSalesTaxToTaxjarJob` is the only job carrying both wrappers and they compose cleanly (`LongRunning` outermost, completion recorded inside the in-flight window). Jobs with optional/default args (`perform(month = nil, year = nil)`) work through `perform(*args)`.
- **Shell:** the `eval`'d window tests are hardcoded single-quoted literals — `$(date)` evaluates at check time, no injection surface; `set -e` does not misfire, since the `eval` sits in an `if` condition. Verified empirically with a stub harness that fakes `date` for all 24 UTC hours under `set -e`: new window → inside `00–05, 08–13`, outside `06,07,14–23`, script alive past a false eval; payouts window still inside only Tue–Fri UTC 10. Also tightened the `%{http_code}` fallback (curl already emits `000` on connection failure, which made the old `|| echo "000"` log "HTTP 000000").
- Every healthcheck answer path re-traced: 200 → deploy, 503 → wait, anything else → fail-safe window decides, timeout → `proceed` for payouts / `skip` for long-running.

### Verification

- `bundle exec rspec spec/sidekiq/concerns/long_running_job_tracking_spec.rb spec/controllers/healthcheck_controller_spec.rb spec/sidekiq/perform_daily_instant_payouts_worker_spec.rb` → **36 examples, 0 failures**
- `bundle exec rubocop` on the changed Ruby files → **no offenses**; `bash -n` + shellcheck on the deploy script → clean (only pre-existing informational notes on untouched lines)

**Only verifiable live:** that the healthcheck is actually reachable from the Buildkite agent (both waits treat unreachable as "consult the fail-safe window", so the failure mode is bounded either way), and the first real overnight run where a long job holds a deploy. Worth watching the first monthly close after this ships (UTC 01:00 on the 1st) to confirm the deploy is held rather than the report killed.

gumclaw:
Heads up — this cannot ship itself as merged, and #6264 fixes it.

The gate added here polls `/healthcheck/long_running_jobs`, which ships in this same commit. The deploy script runs before the app it deploys is live, so the old app answers 404, and 404 falls into the `else` ("we cannot tell") branch → wide fail-safe window → `exit 0`. Production right now:

```
$ curl -s -o /dev/null -w '%{http_code}' https://gumroad.com/healthcheck/long_running_jobs
404
```

Driving the `wait_for_healthcheck` from `main` against that 404 at the current time (inside the UTC 00–05 window):

```
LOG: Long-running job healthcheck unreachable (HTTP 404) inside the fail-safe window — skipping deployment
subshell exit=0        # deploy body never reached
```

So the deploy that installs the endpoint gets skipped, as does every retry inside 12 of every 24 hours — the endpoint can only start answering by being deployed. Build #15648 is still in CI, so #6264 can land ahead of the Deploy step.

#6264 gives 404 its own branch falling back to the narrower pre-rollout window (the old midnight–6am ET block, UTC 04:00–09:59), keeping unreachable/5xx conservative. Ten-case matrix against the shipped function in the PR body.

PR #6264, opened 2026-07-25
Let the long-running-job healthcheck survive its own rollout

gumclaw:
## What

#6261 pointed the production deploy gate at `/healthcheck/long_running_jobs` — an endpoint that ships in that same commit. This makes a 404 from that endpoint its own case, falling back to the narrower pre-rollout window instead of the wide one.

## Why

The deploy script runs *before* the app it is deploying is live, so on the one deploy that first ships a new healthcheck endpoint, the still-running old app answers 404. In #6261 a 404 landed in the `else` branch — "we cannot tell" — which applies the wide fail-safe window (UTC 00–05 and 08–13) and `exit 0`s the deploy.

So the deploy that installs the endpoint gets skipped, and so does every retry that falls inside that window. The endpoint can only begin answering by being deployed, so the check was blocking its own bootstrap. That is 12 of every 24 hours in which #6261 cannot ship itself, and the window it picked contains the hour it merged.

This is live right now on `main`:

```
$ curl -s -o /dev/null -w '%{http_code}' https://gumroad.com/healthcheck/long_running_jobs
404
```

Driving the shipped `wait_for_healthcheck` from `main` against that 404, at the current time (02:2x UTC, inside the window):

```
LOG: Long-running job healthcheck unreachable (HTTP 404) inside the fail-safe window — skipping deployment
subshell exit=0        # "REACHED DEPLOY BODY" never printed
```

A 404 is not actually ambiguous. It is the old app telling us definitively that none of the tracking code is live, so nothing can be registered and there is nothing to wait for. Unreachable and 5xx are genuinely ambiguous and keep the old conservative behaviour.

## What the 404 branch falls back to, and why it is not weaker

The pre-rollout window reproduces the guard this mechanism replaced — the midnight–6am ET block that lived in `.github/workflows/tests.yml` — and, like that block, it is written in **ET** (`TZ=America/New_York`, hours 0–5), not in a hardcoded UTC offset. The first version of this PR hardcoded UTC 04–09 on the reasoning "ET is UTC-4 on EDT"; the pre-merge review caught that this is only true for about 7.5 months of the year. See the review comment below — it was a real bug, and the harness now pins it in both directions.

That is the honest answer for an app with none of the tracking code: the old clock block was the only protection that ever existed there, so reproducing it exactly is neither weaker nor stronger than the behaviour being replaced. Using the *wider* window instead would not buy protection — the jobs it covers cannot register on an app that lacks the concern — it would only delay the endpoint's own rollout.

The parameter defaults to the fail-safe test when not supplied, so the payouts check is unchanged (its endpoint has been deployed since #6160).

**This argument is meant to be removed.** It exists only so the endpoint can bootstrap itself. Once production answers 200/503, the argument should come out (tracked below), because a *future* 404 — a removed route, a web-only rollback, an edge 404 — happens while the tracking code IS live, and the wide fail-safe window is the right answer for those.

## Test results

The harness is now committed as `.buildkite/scripts/deploy_production_test.sh` rather than living in this description. `deploy_production.sh` is never executed by CI, so `bash -n` was the only automated check the gate ever got.

It drives the **shipped** `wait_for_healthcheck`, extracted from the real file with `awk` rather than retyped, so the test cannot drift from the code that deploys. `date` is stubbed through python `zoneinfo`, so the EST/EDT cases exercise real daylight-saving rules instead of a hardcoded offset.

```
=== wait_for_healthcheck cases ===
PASS: 200 nothing in flight deploys -> DEPLOY
PASS: 200 deploys even inside every window -> DEPLOY
PASS: 503 then 200 waits then deploys -> DEPLOY
PASS: 404 at 02 UTC (outside old ET block) SHIPS -> DEPLOY
PASS: 404 at 06 UTC = 02:00 EDT waits (narrow-only) -> SKIP
PASS: 404 at 07 UTC = 03:00 EDT waits (narrow-only) -> SKIP
PASS: 404 at 10 UTC in EDT = 06:00 ET SHIPS -> DEPLOY
PASS: 404 at 10 UTC in EST = 05:00 ET waits -> SKIP
PASS: 404 at 04 UTC in EST = 23:00 ET SHIPS -> DEPLOY
PASS: 404 at 04 UTC in EDT = 00:00 ET waits -> SKIP
PASS: 500 at 02 UTC skips (conservative) -> SKIP
PASS: 500 at 12 UTC skips (report window) -> SKIP
PASS: 000 unreachable at 20 UTC proceeds -> DEPLOY
PASS: 500 at 20 UTC outside all windows proceeds -> DEPLOY
PASS: payouts 404 inside its window skips (5-arg) -> SKIP
PASS: payouts 404 outside its window proceeds (5-arg) -> DEPLOY
PASS: payouts 500 inside its window skips -> SKIP

PASS=17 FAIL=0
```

The four `UTC 04 / UTC 10` rows are the DST pins: UTC 10 is ET 06:00 in summer (deploy) but ET 05:00 in winter (skip), so no hardcoded UTC window can satisfy both. `UTC 06/07` are the only hours where the narrow window skips and the wide one does not — without them, deleting the whole 404 branch still passes.

`--mutate` re-runs every case against five broken variants of the real script and asserts each is caught, so the cases cannot rot into vacuous passes:

```
=== mutation testing (each mutant MUST be caught) ===
CAUGHT (5 failing): delete the whole 404 branch (falls through to wide fail-safe)
CAUGHT (2 failing): pre-rollout window hardcoded to UTC 04-09 (the DST bug this review found)
CAUGHT (4 failing): drop the ${6:-$5} default (payouts 404 evals an empty test)
CAUGHT (1 failing): off-by-one: -le 5 becomes -le 6 in the pre-rollout window
CAUGHT (5 failing): 404 skip uses return instead of exit (deploy proceeds anyway)

MUTANTS_ESCAPED=0
ALL GREEN: cases pass, every mutant caught
```

`bash -n` clean on both files.

## Before / After

No user-facing surface: one internal endpoint's client-side status handling inside the deploy script. No app, job, mailer, or API behaviour changes.

## Note on #6262

#6262 is the sibling implementation of this same request and is now `CONFLICTING` against `main` since #6261 landed first. It independently hit the mirror-image version of this bug (Greptile P1: renaming `/healthcheck/payouts` → `/healthcheck/deploy_safe` made the new script poll a path the old app does not serve) and fixed it caller-side. Worth closing as superseded or rebasing, but that is a separate call from this fix.

---

## AI disclosure

Written by Claude Opus 5 acting as Gumroad's engineering agent, and revised after its own adversarial pre-merge review (verdict in the comments) found a daylight-saving bug in the first version. Found while verifying that #6261 reached production after its merge watcher reported MERGED: the merge commit was not in the live `x-revision`, and probing the new endpoint against production returned 404, which the shipped gate turns into a skip. The agent reproduced the deadlock by extracting and running the real `wait_for_healthcheck`, then wrote the fix and the ten-case matrix above.

### DIFF
--- a/.buildkite/scripts/deploy_production.sh
+++ b/.buildkite/scripts/deploy_production.sh
@@ -12,20 +12,41 @@ logger() {
 # Redis while they run (see DeployBlockingJobTracking) and the matching healthcheck answers
 # 503 for as long as any of them is registered.
 #
-# wait_for_healthcheck <label> <url> <max attempts> <on-timeout: proceed|skip> <fail-safe window test>
+# wait_for_healthcheck <label> <url> <max attempts> <on-timeout: proceed|skip> <fail-safe window test> [pre-rollout window test]
 #
 # Polls every 3 minutes. Meanings of the answers:
 #   200 -> nothing in flight, deploy now.
 #   503 -> something is in flight, keep waiting.
+#   404 -> the app currently in production does not serve this endpoint at all, i.e. it
+#     predates the deploy that adds it. See the pre-rollout note below.
 #   anything else (unreachable, 5xx from the LB, ...) -> we cannot tell, so fall back to the
 #     conservative clock window this check replaced: skip the deploy if we are inside it,
 #     proceed if we are not. A broken healthcheck can then never be worse than the old
 #     time-based behaviour.
 #
-# The fail-safe window test is a shell snippet evaluated with `eval`; it should succeed when
-# the current time is inside the window.
+# Why 404 is handled separately from "we cannot tell":
+#
+# The deploy script runs BEFORE the app it is deploying is live, so on the single deploy that
+# first ships a new healthcheck endpoint, the still-running old app answers 404. That is not an
+# ambiguous failure — it is the old app telling us definitively that none of the new tracking
+# code is live yet, so nothing can be registered and this check has nothing to report on. If we
+# treated it as "cannot tell" and applied the wide fail-safe window, that window would skip the
+# very deploy that installs the endpoint, and keep skipping it for as long as the window lasts.
+# The endpoint can only start answering by being deployed, so the check would be blocking its
+# own bootstrap.
+#
+# Instead, a 404 falls back to the narrower clock window that was actually protecting these jobs
+# BEFORE this mechanism existed (the pre-rollout window). That is the honest answer: on an app
+# without the tracking code, the old clock guard is the only protection there ever was, so
+# reproducing it exactly is neither weaker nor stronger than the behaviour we are replacing.
+# Once the endpoint is live it answers 200/503 and neither window is consulted again.
+#
+# The window tests are shell snippets evaluated with `eval`; each should succeed when the
+# current time is inside its window. The pre-rollout test defaults to the fail-safe test when
+# not supplied, which is the right behaviour for a check whose endpoint is already deployed.
 wait_for_healthcheck() {
   local label="$1" url="$2" max_attempts="$3" on_timeout="$4" failsafe_window_test="$5"
+  local prerollout_window_test="${6:-$5}"
   local attempt hc_status
 
   for attempt in $(seq 1 "$max_attempts"); do
@@ -38,6 +59,13 @@ wait_for_healthcheck() {
     elif [ "$hc_status" = "503" ]; then
       logger "$label in flight (healthcheck 503) — waiting 3 minutes (attempt $attempt/$max_attempts)"
       sleep 180
+    elif [ "$hc_status" = "404" ]; then
+      if eval "$prerollout_window_test"; then
+        logger "$label healthcheck not deployed yet (HTTP 404) and we are inside the pre-rollout window — skipping deployment"
+        exit 0
+      fi
+      logger "$label healthcheck not deployed yet (HTTP 404) and we are outside the pre-rollout window — proceeding so the endpoint can ship"
+      return 0
     else
       if eval "$failsafe_window_test"; then
         logger "$label healthcheck unreachable (HTTP $hc_status) inside the fail-safe window — skipping deployment"
@@ -90,8 +118,16 @@ wait_for_healthcheck "Payout batch" "https://gumroad.com/healthcheck/payouts" 15
 #   UTC 11:00 finances / deferred refunds / Stripe balance summaries reports
 # => hours 00-05 and 08-13. Outside those we proceed, because nothing that registers here
 # is scheduled to be running.
+#
+# The last argument is the PRE-ROLLOUT window, used only while production still 404s this
+# endpoint (see wait_for_healthcheck). It reproduces the guard this check replaced — the
+# midnight-6am ET block that used to live in .github/workflows/tests.yml — expressed in UTC:
+# ET is UTC-4 on EDT, so ET 00:00-05:59 is UTC 04:00-09:59. Deliberately NOT the wider fail-safe
+# window above: on an app that has none of the tracking code, the old block is exactly the
+# protection that existed, and widening it here would only delay this endpoint's own rollout.
 wait_for_healthcheck "Long-running job" "https://gumroad.com/healthcheck/long_running_jobs" 40 skip \
-  '[ "$(date -u +%-H)" -le 5 ] || { [ "$(date -u +%-H)" -ge 8 ] && [ "$(date -u +%-H)" -le 13 ]; }'
+  '[ "$(date -u +%-H)" -le 5 ] || { [ "$(date -u +%-H)" -ge 8 ] && [ "$(date -u +%-H)" -le 13 ]; }' \
+  '[ "$(date -u +%-H)" -ge 4 ] && [ "$(date -u +%-H)" -le 9 ]'
 
 ECR_REGISTRY=${ECR_REGISTRY}
 WEB_REPO=${ECR_REGISTRY}/gumroad/web
--- a/.buildkite/scripts/deploy_production.sh
+++ b/.buildkite/scripts/deploy_production.sh
@@ -39,7 +39,8 @@ logger() {
 # BEFORE this mechanism existed (the pre-rollout window). That is the honest answer: on an app
 # without the tracking code, the old clock guard is the only protection there ever was, so
 # reproducing it exactly is neither weaker nor stronger than the behaviour we are replacing.
-# Once the endpoint is live it answers 200/503 and neither window is consulted again.
+# Once the endpoint is live it answers 200/503 on the healthy paths; the windows stay in place
+# for the abnormal ones (a 5xx or an unreachable host still falls back to the fail-safe window).
 #
 # The window tests are shell snippets evaluated with `eval`; each should succeed when the
 # current time is inside its window. The pre-rollout test defaults to the fail-safe test when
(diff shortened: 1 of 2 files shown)


gumclaw:
## Fable-5 adversarial pre-merge review — `7fa182c0`

Dispatched an independent reviewer against `5ee5ff84` with a hunt list for this change class (replacing a blanket clock window with a targeted gate). **Verdict: CHANGES-REQUIRED**, 1 P1 + 2 P2 + 3 P3. CI was fully green (77 checks) and Greptile scored it 5/5 "safe to merge with minimal risk" while the P1 was live — this is the second time this specific change class has produced a real bug that both missed.

| # | Finding | Disposition |
|---|---|---|
| **P1** | Pre-rollout window hardcoded as UTC 04–09 does not reproduce the ET guard it claims to, outside EDT | **Fixed** in `7fa182c0` — now `TZ=America/New_York`, hours 0–5 |
| **P2** | Narrow window is not self-retiring: every *future* 404 gets the weakest window forever | **Fixed** (documented + tracked below) |
| **P2** | Harness rows vacuous — `404@04`/`404@09` pass even with the 404 branch deleted | **Fixed** — committed harness + mutation testing |
| **P3** | "neither window is consulted again" overstated | **Fixed** in `7fa182c0` |
| **P3** | Log text "not deployed yet" misdiagnoses a post-rollout 404 | **Fixed** — now "healthcheck absent (HTTP 404)" |
| **P3** | Single curl sample decides the 404 path (mixed-fleet race) | **Not fixed** — see below |

### P1: the timezone claim was false, and it aimed at the worst possible hour

My first version wrote the window as UTC 04–09, reasoning "ET is UTC-4 on EDT, so ET 00:00–05:59 is UTC 04:00–09:59." True today. False from November to mid-March, and the comment betrayed the bug in its own wording — it hardcoded the summer offset forever.

The block being reproduced was `ET_HOUR=$(TZ=America/New_York date +%-H); [ "$ET_HOUR" -lt 6 ]`, i.e. daylight-saving correct by construction. #6261's own description advertised it as "DST-safe (IANA zone, not a fixed UTC offset)" — and I then replaced it with a fixed UTC offset.

In EST the window is wrong on both edges, and the harmful edge lands on the busiest hour in the schedule:

| | EDT (Mar–Nov) | EST (Nov–Mar) |
|---|---|---|
| UTC 04 | ET 00:00 — both skip ✓ | ET 23:00 — new **over**-blocks (harmless) |
| UTC 10 | ET 06:00 — both deploy ✓ | ET 05:00 — old skipped, new **deploys** ⚠️ |

UTC 10:00 is the single most scheduled hour in `config/sidekiq_schedule.yml`, and `GenerateFinancialReportsForPreviousQuarterJob` fires at UTC 10:00 on **January 2nd** — deep in EST. It fans out the tracked VAT and per-country sales reports, which hold no checkpoint, so a kill restarts them from zero. That is precisely the "finance report silently goes missing" failure #6261 exists to prevent, which #5680 documents happening twice in 48 hours.

Verified empirically rather than by argument:

```
UTC 04 on 2026-01-15 -> ET 23 (EST)  OLD_block=deploy  NEW=SKIP     *** MISMATCH ***
UTC 10 on 2026-01-15 -> ET 05 (EST)  OLD_block=SKIP    NEW=deploy   *** MISMATCH ***
UTC 10 on 2026-07-15 -> ET 06 (EDT)  OLD_block=deploy  NEW=deploy
```

Fixed by writing the window in ET, exactly as the original was. Worth being precise about why this differs from the wide fail-safe window one line above, which correctly stays in UTC: that one is read straight off the cron schedule, which is written in UTC and does not shift with daylight saving. This one reproduces a human "midnight to 6am local" rule. Same file, two windows, two correct-but-different timezones — the comment now spells that out so nobody "fixes" the inconsistency later.

### P2: the pre-rollout argument must not outlive the bootstrap

The reviewer's sharpest point, and it reframes the change. After this deploy lands, the argument's only remaining effect is to *weaken* the 404 path permanently. And every future 404 is a case where the justification for the narrow window ("no tracking code is live, nothing can be registered") is **false**:

- a removed or renamed route, with the tracking concern still in every job
- a web-tier-only rollback while Sidekiq workers stay on the new image — endpoint 404s while tokens are live in Redis
- an edge/CDN 404

In all of those the wide fail-safe window is the correct answer. The bootstrap reasoning holds only for the all-old-fleet case this PR exists to unblock. Documented in the code with an explicit removal instruction, and I'll open the follow-up to delete the argument once production answers 200/503.

### P2: the harness was passing vacuously, and is now committed

The ten cases in the original description looked thorough and were partly theatre. `404@04` and `404@09` both sit inside the wide window too, so **deleting the entire 404 branch still passed them**. The only hours where the narrow window alone skips are UTC 06 and 07 — untested. Nothing tested UTC 10 (the off-by-one edge, the DST edge, and the quarterly-report hour at once). Nothing tested the 5-argument payouts call that guards `${6:-$5}`. Nothing tested the 503 wait loop.

Since `deploy_production.sh` is never executed by CI, that harness was the only test this gate had, so vacuity in it is a real defect. It now lives at `.buildkite/scripts/deploy_production_test.sh` (17 cases, `PASS=17 FAIL=0`), drives the shipped function extracted with `awk`, and stubs `date` through python `zoneinfo` so the EST/EDT rows exercise real DST rules.

Rather than assert the cases are load-bearing, `--mutate` proves it — it re-runs everything against five broken variants of the real script and requires each to be caught:

```
CAUGHT (5 failing): delete the whole 404 branch (falls through to wide fail-safe)
CAUGHT (2 failing): pre-rollout window hardcoded to UTC 04-09 (the DST bug this review found)
CAUGHT (4 failing): drop the ${6:-$5} default (payouts 404 evals an empty test)
CAUGHT (1 failing): off-by-one: -le 5 becomes -le 6 in the pre-rollout window
CAUGHT (5 failing): 404 skip uses return instead of exit (deploy proceeds anyway)

MUTANTS_ESCAPED=0
```

The second mutant is the P1 above, so this PR's own bug is now a permanent regression test.

### Not fixed

**P3 — single-sample 404 decision.** One `curl` decides the 404 path, so during the bootstrap rollout a *second* overlapping deploy could sample an old pod (404) while new workers already register tokens. Requiring two consecutive 404s would close it. Leaving it: it needs two overlapping deploys inside one rollout window, and the argument is being removed shortly anyway. Noted so it is a decision, not an oversight.

### Confirmed safe

Checked and held up — these matter as much as the findings:

- **`exit 0` skip semantics** — both call sites are top-level simple commands (no pipeline, subshell, or command substitution), so `exit 0` terminates the script before the deploy body. Verified no other invocation exists; the mutation test pins it.
- **Branch ordering** — the `404` arm precedes the generic `else`, so a 404 can no longer reach the wide fail-safe branch. That was the entire point of the PR, and mutant 1 proves it is tested.
- **Payouts check genuinely unchanged** — with 5 arguments, `${6:-$5}` (colon form, so an explicitly-empty 6th is covered too) makes the 404 branch evaluate the identical window string with identical skip/proceed outcomes as the old `else` path. Behaviourally equal; only log text differs.
- **`eval ""` fails safe** — empty window test returns success → skip, not proceed. Unreachable today since both call sites pass non-empty tests.
- **`return 0` on 404-outside-window** — returns on the first sample, never interacts with `max_attempts`/`on_timeout`, and leaves the 200 and 503 paths byte-identical to `main`.
- **`set -e`** — `eval` inside an `if` condition is errexit-exempt; the curl is guarded with `|| hc_status="000"`. `bash -n` clean on both files.
- **No octal trap** — bash's `[ -eq ]` parses "08"/"09" as decimal; the new snippets use unpadded `%-H` regardless.
- **Wide window vs schedule** — re-derived the UTC 00–05/08–13 enumeration against every entry in `config/sidekiq_schedule.yml`; accurate, including the ~2h runtime headroom.

### Only verifiable live

Whether production actually 404s at deploy time and takes the intended branch. That is the whole premise, and it is currently true (`curl https://gumroad.com/healthcheck/long_running_jobs` → 404). The first deploy carrying this will print `Long-running job healthcheck absent (HTTP 404) ... proceeding so the endpoint can ship` — if it prints anything else, or the endpoint still 404s afterwards, the bootstrap did not work and this needs another look.

---

Written by Claude Opus 5. Reviewer dispatched as an independent headless instance against the branch, static analysis only, with a change-class hunt list; I verified the DST finding empirically before fixing it, and it agreed with a finding I had reached independently.

PR #6268, opened 2026-07-25
Treat a 404 from the long-running-job deploy healthcheck like any other unclear answer

gumclaw:
## What

Removes the extra "pre-rollout" clock window that `.buildkite/scripts/deploy_production.sh` passed to the long-running-job healthcheck for the 404 case. A 404 now falls back to the same wide fail-safe window as any other answer the gate cannot interpret. The 404 *branch* stays — only the narrow window goes.

## Depends on

Follow-up to #6264, which introduced the argument being removed here. #6264 has merged, so this branch was rebased onto `main` and now targets `main` directly — it is no longer a stacked PR. The draft gate on this PR has cleared: `/healthcheck/long_running_jobs` now answers **200** in production (verified 2026-07-25 04:03Z, three consecutive requests), so the #6264 release is out and the argument this PR removes can no longer have a live effect. Ready for review.

## Why

The argument had exactly one job. `/healthcheck/long_running_jobs` shipped in the same commit as the gate that polls it, so on that one deploy the still-running old app answered 404 and the wide fail-safe window (UTC 00–05 and 08–13) would have skipped the deploy that installs the endpoint — the check blocking its own bootstrap. #6264 fixed that by routing 404 through a narrower window reproducing the old midnight–6am ET block.

Once the endpoint is live it answers 200 or 503, so the argument only has an effect on a *future* 404. In every one of those cases the tracking code **is** deployed, which makes the reasoning behind the narrow window ("nothing is live, nothing can be registered, nothing to protect") false:

- a route renamed or removed while the tracking concern is still in every job
- a web-tier-only rollback, with Sidekiq workers staying on the new image and tokens live in Redis
- an edge or CDN 404

All three have real work to protect, so the wide fail-safe window is the correct answer. Left in place, the argument would silently apply the weakest window to them forever. The 404 branch itself is worth keeping: its log line names a specific diagnosis ("healthcheck absent") that the generic "unreachable" line does not, and that is what someone reading a skipped deploy log needs.

The two paragraphs of comment explaining the pre-rollout window are removed with it, since they document a situation that can no longer arise.

Refs #6267, follow-up to #6264.

## Test results

`.buildkite/scripts/deploy_production_test.sh` (added in #6264) is updated in the same commit: the DST-pinning rows existed only to pin the ET-based narrow window and are gone, replaced with rows that pin the 404 case to the wide window in both directions. Two new `check_log` assertions cover the reason the branch survives — a 404 must log as *absent*, a 5xx as *unreachable* — so deleting the branch is still caught even though it no longer changes any deploy/skip outcome.

```
PASS=17 FAIL=0

=== mutation testing (each mutant MUST be caught) ===
CAUGHT (2 failing): delete the whole 404 branch (loses its diagnostic log line)
CAUGHT (4 failing): 404 branch ignores the window and always skips
CAUGHT (3 failing): 404 skip uses return instead of exit (deploy proceeds anyway)
CAUGHT (1 failing): off-by-one: fail-safe window overnight leg becomes -le 6
CAUGHT (2 failing): fail-safe window loses its report-hours leg (08-13 becomes 18-13)

MUTANTS_ESCAPED=0
ALL GREEN: cases pass, every mutant caught
```

test suite

`bash -n` clean on both files.

## QA steps

No app surface to click through — this is the deploy script, which CI only ever syntax-checks.

1. `cd .buildkite/scripts && ./deploy_production_test.sh --mutate` — expect `PASS=17 FAIL=0`, `MUTANTS_ESCAPED=0`.
2. `grep -c 'wait_for_healthcheck "Long-running job"' -A2 deploy_production.sh` — the call is now 5 arguments, the same shape as the payouts call above it.
3. Before merging, confirm production no longer 404s the endpoint: `curl -s -o /dev/null -w "%{http_code}" https://gumroad.com/healthcheck/long_running_jobs` should be 200 or 503. Verified 2026-07-25 04:03Z: production returns **200**. This gate is satisfied.

## Before / After

No user-facing surface. The change is one internal endpoint's status handling inside the Buildkite deploy script — no app, job, mailer, or API behaviour changes, and no rendered page is touched.

---

## AI disclosure

Written by Claude Opus 5 acting as Gumroad's engineering agent, following up on the P2 that its own pre-merge review of #6264 raised against that PR. The agent read both the shipped and branch versions of the deploy script, removed the argument and its now-stale comment block, reworked the test harness rows that existed only to pin the removed window, added the two log-line assertions that keep the surviving 404 branch under test, and ran the case matrix plus mutation suite locally.


gumclaw:
#6264 merged, so GitHub retargeted this PR from its branch to `main` and the diff came out conflicting. Rebased with `--onto origin/main`, so the only commit left is this one — the deploy script and its test harness, plus the test-suite screenshot.

Re-ran the harness on the rebased tree: `PASS=17 FAIL=0`, `MUTANTS_ESCAPED=0`, `bash -n` clean on both scripts. Body updated to drop the stacked-PR wording.

Still a draft on purpose. `/healthcheck/long_running_jobs` returns 404 in production right now because #6264 is merged but not yet released; once a release ships it and the endpoint answers 200 or 503, this is ready to go.

gumclaw:
Draft gate cleared, so this is out of draft and ready for review.

The only reason it was a draft was that `/healthcheck/long_running_jobs` still answered 404 in production, which would have made the behaviour change untestable against the real endpoint. #6264 merged at 03:14Z and its release is out — production now answers **200** (checked three times in a row at 04:03Z today). The removed argument therefore can no longer have any live effect, which is exactly the premise of this change.

State at ready: CI green at `83ccfcd` (77 passing, 0 failing, 8 skipped), no conflicts, Greptile 5/5 with no files flagged. Test harness is `PASS=17 FAIL=0` with `MUTANTS_ESCAPED=0`; body has the run output, the suite screenshot in `qa-media/`, and the AI disclosure.

@ershad routing this to you as the other recent author of `.buildkite/scripts/deploy_production.sh` — deploy-pipeline changes should be a human merge. No auto-merge from me on this or anything else.

gumclaw:
## Fable-5 adversarial pre-merge review — READY-TO-MERGE

Reviewed inline (diff is ~60 lines of shell, all in `deploy_production.sh` + its harness). Verdict at head `83ccfcd0c`.

### Draft gate: cleared, verified live

The only reason this was a draft was that `/healthcheck/long_running_jobs` still 404'd in production, which would have made removing the narrow pre-rollout window re-create #6264's bootstrap deadlock. That condition is gone:

- Prod `x-revision` is `b12f77e1daa2` — literally #6264's merge commit, so both #6261 (endpoint) and #6264 (bootstrap fix) are live.
- **20/20** probes of `/healthcheck/long_running_jobs` → `200`, and **10/10** of `/healthcheck/payouts` → `200`. Probed repeatedly on purpose rather than once: a single 200 only proves one host behind the LB serves the route, and the whole safety argument here rests on *every* web host having it.
- Both routes present in `config/routes.rb` on `main` (lines 21–22).

### Findings

| 1 | P3 | The `404` branch is now behaviourally identical to the generic "cannot tell" branch — it exists solely for its distinct log line. | **Intended, and protected.** This is the documented reason to keep it: "healthcheck absent" names a diagnosis ("route renamed/removed, web-tier-only rollback, edge 404") that "unreachable" does not. The risk was it rotting into apparent dead code — the harness now asserts on the *log text* in both directions, and the `delete the whole 404 branch` mutant is CAUGHT (2 failing cases). |
| 2 | P3 | A *permanent* 404 (route genuinely deleted/renamed later) will silently skip every deploy during UTC 00–05 / 08–13 with no alert. | Pre-existing class, not introduced here — identical to the existing 5xx fail-safe behaviour on `main`. Noted, not blocking. |

No P1s, no P2s.

- **Payouts lane is untouched.** It never passed a 6th argument, so `${6:-$5}` already resolved to its own fail-safe window; deleting the parameter is a genuine no-op there. Harness pins all three payouts cases.
- **The removal is the authored plan, not a shortcut.** #6264 introduced the narrow window with an explicit `REMOVE THIS LAST ARGUMENT once the endpoint is confirmed live` note; this PR is that removal, in the documented merge order #6264 → #6268.
- **The 4 DST rows were deleted, not adapted** — correct, since they existed only to pin the ET-expressed pre-rollout window that no longer exists. No orphaned cases pretending to test removed behaviour.
- **`bash -n`** clean on both files.

### Test evidence

`.buildkite/scripts/deploy_production_test.sh` is the only test this gate has (`deploy_production.sh` is never executed by CI, only syntax-checked), so it was run both ways:

```
PASS=17 FAIL=0
MUTANTS_ESCAPED=0
ALL GREEN: cases pass, every mutant caught
```

All 5 mutants caught, and no `SKIP (mutation did not apply)` lines — every mutant regex still matches the post-diff code, so none of them passed deceptively.

CI at `83ccfcd0c`: 83 check-runs, **0 pending, 0 failed** (check-runs API, not the commit-status endpoint).

PR #6274, opened 2026-07-25
Don't mark a video as analyzed when it has no readable video stream

gumclaw:
## What

`WithFileProperties#assign_video_attributes` set `analyze_completed = true` even when the probe came back with no video stream, leaving `width` and `height` nil. This change treats nil dimensions as a failed analysis instead: `analyze_completed` is cleared, and the file routes through the same seller notification the transcode-failure path already uses.

The probe rescue is also narrowed to cover only the probing itself, so an error raised after a file has analyzed successfully can no longer report that file to the seller as broken.

## Why

A course video can end up permanently unplayable with nothing in the system able to recover it.

`Streamable#transcodable?` requires both dimensions:

```ruby
def transcodable?
  streamable? && height.present? && width.present?
end
```

With nil dimensions, `TranscodeVideoForStreamingWorker` always takes its `else` branch, logs "is not transcodable", and returns — so no `TranscodedVideo` row is ever created and the player has no manifest to load. Meanwhile `analyze_completed = true` means `queue_for_transcoding?` reports the file as ready and nothing ever retries the analysis. The file is stuck in a state it can never leave, and neither the seller nor the buyer sees an error.

The trigger is that **streamio-ffmpeg does not raise on an unreadable file** — it returns nil dimensions and duration 0. Verified directly against the gem:

```
$ dd if=/dev/zero of=/tmp/zeros.mp4 bs=1024 count=512
$ ruby -e 'm = FFMPEG::Movie.new("/tmp/zeros.mp4"); p m.width, m.height, m.duration'
{duration: 0, width: nil, height: nil, bitrate: nil, frame_rate: nil}
```

No `NoMethodError`, so the existing `rescue NoMethodError` never fired and execution fell straight through to `analyze_completed = true`. The rescue was guarding the wrong failure mode.

For the file that surfaced this, the underlying source object on S3 is 77,346,962 bytes of null bytes — every 64 KB sample across the whole object contains zero non-null bytes, while a healthy sibling lesson on the same product probes fine through the identical code path. So that particular upload does need to be replaced by the seller; this PR is about the platform behavior that turned a bad upload into a permanent silent failure.

Sampling a bounded recent id window in production found 3 more alive video files in the same nil-dimension-but-analyzed state, so this is not a one-off.

Tracking issue: internal (private tracker).

## Test plan

- New specs in `spec/modules/with_file_properties_spec.rb` cover: the no-readable-video-stream case (`analyze_completed` not set, file not queued for transcoding, seller notified); a previously-analyzed file whose source is replaced with a corrupt one (flag cleared); a `.mov` whose probe raises because it has no video stream; and an error raised *after* a successful analysis (no seller email, flag preserved).
- `spec/modules/with_file_properties_spec.rb` — 42 examples, 0 failures.
- Verified each guard is load-bearing: deleting the nil-dimension check reddens the 3 corrupt-file specs; deleting the `analyze_completed = false` line reddens the re-analyze spec; widening the rescue back to method scope reddens the post-success spec. All return to green when restored.
- Confirmed `VideoFile` (which does not include `Streamable` and so has no `transcoding_failed`) does not raise on this path — the notification call is guarded by `respond_to?`.
- `spec/models/product_file_spec.rb` has 10 failures on this branch and the **same 10** on unmodified `main` (baselined via `git stash`, failure lists diff identical) — pre-existing, not introduced here.
- `rubocop` clean on both changed files.

## Test results

spec run and root cause evidence

## QA steps

No user-facing surface changes — this is backend analysis/transcoding logic, so there is no UI to screenshot. The behavior is observable as:

1. Upload a video whose container has no readable video stream (a truncated or all-null-bytes `.mp4` reproduces it).
2. Before this change: the file's `analyze_completed` flips to true with nil `width`/`height`, no `TranscodedVideo` row is created, the lesson never plays, and no email is sent.
3. After this change: `analyze_completed` stays false, the file is not queued for transcoding, and the seller receives the existing "A video failed to transcode." email so they know to re-upload.

## Notes for follow-up

Not included here to keep the diff to the behavior fix:

- A backfill/sweep for existing files already stuck in this state (they will not self-heal, since this change only prevents new ones). A `onetime` script should clear the flag and re-enqueue `AnalyzeFileWorker`; note that it would re-send the seller notification, so running it twice would double-email.
- Alerting on "alive streamable file with `analyze_completed` but nil dimensions", which currently has no monitoring.
- Unrelated pre-existing gap noticed while tracing this: if `ffprobe` writes nothing to stdout, `Ffprobe#parse` raises `JSON::ParserError`, which nothing rescues and which burns the worker's retries.

---

## AI disclosure

Written by Claude Opus 5 (Hermes Agent) acting on an internal request to pull the affected source file and attempt a fix.

A local fable-5 adversarial review was then run against the branch as the pre-merge gate; it returned FIX-FIRST with one P1 and one P2, both since fixed and documented in a triage comment on this PR.

Work performed: pulled the affected object from S3 via a presigned URL and inspected it with `ffprobe` plus byte-level sampling; ran a control probe against a healthy sibling file to rule out the read path; reproduced the nil-dimension behavior against the `streamio-ffmpeg` gem locally to establish the real failure mode (which differed from the root cause hypothesized on the issue — the `rescue NoMethodError` path was never being taken); read `Streamable`, `TranscodeVideoForStreamingWorker`, and `WithFileProperties` to trace why the file could never recover; sized the affected population with a bounded production read; then wrote the guard, the seller notification helper, and the specs, and verified them including a load-bearing check and a pre-existing-failure baseline against `main`.


gumclaw:

Verdict at dispatch: **FIX-FIRST**. Both blocking findings are now fixed and pushed.

|---|---|---|---|
| 1 | P1 | `clear_properties` resets the dimension columns but not `analyze_completed`, so a file that analyzed fine before and is re-analyzed against a now-corrupt source kept the flag and stayed stuck — which is also the remediation path for the files already broken | **Fixed** in c082995a4 — the failure path clears the flag explicitly, plus a spec for the previously-analyzed case |
| 2 | P2 | The rescue spanned the whole method including `video_file_analysis_completed`, which runs *after* the file is marked analyzed — a `NoMethodError` there would email the seller about a healthy file and clear a legitimate flag | **Fixed** in f191ecd95 — rescue scoped to the probe only, with specs for both the `.mov` raise lane and the post-success case |
| 3 | P3 | No remediation for files already broken in production | **Deferred, noted in the body.** Out of scope for the behavior fix; needs a `onetime` script to clear the flag and re-enqueue. The P1 fix is what makes that script work at all |
| 4 | P3 | `corrupt_path` manufactured a fixture-looking path to a non-existent file | **Fixed** — plain `/tmp/corrupt.mp4` with a comment that the stub means it is never opened |
| 5 | P3 | Repeat analyses re-send the same email; `duration: 0` persists and renders as a 0-second duration | **Acknowledged, not changed.** Bounded in the automatic pipeline: analysis is only enqueued `after_commit on: :create`, and the failure path doesn't raise, so `AnalyzeFileWorker` sends exactly one email per enqueue. A remediation script run twice would double-email — worth knowing when that script is written |

Cleared as non-issues after being checked explicitly: no new `save!` raise vector (the old nil-dimension path fell through to `save!` too); skipping `video_file_analysis_completed` is correct (enqueuing a transcode for a dimensionless file only reproduces the failure, and it would set a product-wide flag on behalf of an unusable file); no double-notification on the fresh-upload path (`queue_for_transcoding?` is false, so the publish and purchase sweeps both skip); `respond_to?(:transcoding_failed)` returns the intended answer for each model, since it is public on `ProductFile` via `Streamable` and absent on `VideoFile`.

One pre-existing issue surfaced but left alone as out of scope: if `ffprobe` writes nothing to stdout, `Ffprobe#parse` raises `JSON::ParserError`, which nothing rescues and which burns the worker's retries. Unchanged by this PR.

Also verified independently while triaging: `WithFileProperties#analyze` is reached only from `AnalyzeFileWorker` — the other `.analyze` call sites in the codebase are ActiveStorage blobs, a different method — which is what bounds the email-volume question above.

PR #6273, opened 2026-07-25
Say plainly that an AI agent designs the profile page

gumclaw:
## What

Rewrites the profile-customization copy in three places so it says plainly that the profile page is built to be updated by an AI agent, without implying a manual editor exists or is coming:

- **Pages → Home editor** (`app/javascript/pages/Pages/Edit.tsx`): the agent panel's intro says the profile page is built to be updated by an agent, instead of describing the agent as one way to "replace the default template".
- **Settings → Profile → Pages tab** (`app/javascript/pages/Settings/Profile/Show.tsx`): the deprecation alert no longer implies the sections editor is being replaced by another manual surface. It says the sections editor is the last hand-editable profile layout, that it stays while a profile still uses sections, and that customizing beyond it is done by an agent.
- **Help center article 124** (`_124-your-gumroad-profile-page.html.erb`): the "Build your profile page with your agent" section states the page is built to be updated by an agent and there is no in-app editor for it; the legacy-sections section notes the sections editor is the last hand-editable layout.

All three surfaces name both agent options: your own agent, or the Agent tab built into Gumroad.

No behavior, routing, or policy changes — copy only.

## Why

Refs antiwork/gumroad-private#1333.

A long-tenured seller reported that the product pushes them into an agent with no alternative, while our support replies asserted the agent was optional. Both halves of that were confusing:

- The profile page genuinely has no in-place editor (`isEditable = canEdit && !is_profile && !page.custom_html`), so the agent panel is the entire editing surface.
- The one hand-editable surface left, the sections editor, is announced as "legacy… being phased out" and points at the agent-only Home page, so a seller correctly concludes there is no non-agent option and reasonably reads the deprecation notice as meaning a manual replacement is coming.

The copy should state how profile design actually works rather than leaving sellers to infer it and then hear the opposite from support. It states the fact neutrally — no "only way" or "by design" framing at the seller. The bio-as-headline half of the issue is a separate change and stays on the issue.

## Before / After

| | Before | After |
|---|---|---|
| **Pages → Home agent panel** | "Replace the default template with a page your agent designs as full HTML — custom layout, animations, anything." | "Your profile page is built to be updated by an AI agent — use your own agent with the prompt below, or Gumroad's built-in Agent tab. Either way, the agent replaces the default template with full HTML: custom layout, animations, anything." |
[...]

PR #6281, opened 2026-07-25
Stop the profile save-toast spec assertion from racing the save

gumclaw:
## What

Makes the `save_changes` helper in `spec/requests/user/profile_spec.rb` wait for the save to actually complete before asserting the "Changes saved!" toast:

- The assertion now finds the toast element (`[data-testid='toast-alert']`) and asserts the text inside it, instead of matching any `role="alert"` node on the page.
- The wait is raised to 30s to cover both network hops the save performs.

Spec-only change. No application code touched.

## Why

Fixes antiwork/gumroad-private#1336.

`save_changes` clicked "Update profile" and immediately asserted the toast with Capybara's default 10s wait. Saving is two sequential round-trips, not one: the settings POST, then `router.reload(...)` to refresh the Inertia props. Only after both resolve does `showAlert("Changes saved!")` fire. The Inertia reload goes through axios rather than the app's own `request` helper, so it does not increment `window.__activeRequests` and `wait_for_ajax` cannot see it — there was nothing else in the helper synchronising on it. On a loaded CI machine that pair exceeded 10s and the example failed.

The failure message was also misleading:

```
expected to find visible alert with text "Changes saved!" but there were no matches.
Also found "", which matched the selector but not all filters.
```

That empty match is the toast container, which is always mounted and starts out with no message, so `have_alert` found a `role="alert"` node with the wrong text rather than reporting a plain timeout. Scoping the assertion to the toast element removes that confusion; the raised wait fixes the actual race.

Timing measured locally by instrumenting the helper: the toast took 2.86s and 1.55s for the two saves in this example on an unloaded machine, so the margin under CI load was thin — consistent with an intermittent failure rather than a hard break.

## Before / After

Not applicable — spec-only change, no user-facing surface.

## QA steps

No product behavior changes. Verification is the spec itself:

```
bundle exec rspec spec/requests/user/profile_spec.rb -e "saves tab settings"
```

## Test results

Three consecutive runs of the affected example on this branch, each a fresh process in a clean worktree — all green:

spec runs

Baseline for comparison: the same example on `main` at this branch's base passed 4 of 4 locally, which is expected — the failure only shows up under load (it failed twice in CI on #6273 and, per the linked issue, 1 of 3 and 2 of 2 on two earlier commits). The fix is therefore justified by the mechanism above rather than by a locally reproduced red run, and the local runs confirm the new assertion works.

`bundle exec rubocop spec/requests/user/profile_spec.rb` — no offenses.

---

## AI disclosure

Authored by Claude Opus 5 running as the Gumclaw agent. Instructions given: fix the flaky profile save-toast spec tracked in gumroad-private#1336 in its own PR, separate from the copy change in #6273. The agent traced the save path through `app/javascript/pages/Settings/Profile/Show.tsx` and `components/server-components/Alert.tsx`, instrumented the helper to measure the real toast latency, applied the fix, and re-ran the example three times.


gumclaw:
Self-review of this diff (no separate reviewer dispatched — one spec helper, two lines of behavior):

- `find(...)` with `wait: 30` raises if the toast container never appears, which it always does (it is mounted on every page), so this is effectively a no-op guard rather than a new failure mode.
- Scoping to the container does not weaken the assertion: the container holds exactly one message node, so "text inside the toast" is the same claim as before, just without the empty-container red herring.
- The 30s wait applies only inside this helper, so no other spec's timing changes. It does not mask a real regression: if the save genuinely fails, `showAlert` fires with the error message instead and the text assertion still fails — it just fails after 30s instead of 10s.
- `wait_for_ajax` is kept after the assertion; it still covers the app's own `request`-helper traffic.
- No production code, routing, or copy is touched.

PR #6285, opened 2026-07-25
Support Pix on BRL checkouts behind a launch flag

gumclaw:
## What

Wires up **Pix**, Brazil's instant-payment scheme, as a forced-currency local payment method at checkout — the registry/eligibility layer **and the full charge path**. Pix is the #3-ranked local method by addressable GMV (see the internal tracker).

The launch flag `checkout_local_method_pix` is still **off everywhere**, so no buyer can see or select Pix until it is ramped deliberately. Nothing about existing checkouts changes. With the charge path now included, the branch has everything needed to QA the end-to-end flow on a preview app and then ramp.

It follows the pattern already proven in production by UPI (INR) and iDEAL/Bancontact (EUR), so there is no new architecture here — the registry slice adds Pix to each place a forced-currency method is declared, and the charge slice adds the three things Pix needs that those methods didn't: a tax decision, an asynchronous completion model, and a transaction window.

## Why Pix is a "forced-currency" method

Stripe only accepts Pix on **BRL** payment intents. Creating a Pix intent in any other currency is rejected outright — verified against our live platform account, where a 5.00 BRL intent creates cleanly and the identical USD intent is refused with *"Payments with pix support the following currencies: brl"*. So the payment method itself decides the presentment currency; there is nothing to detect from the buyer's location. That is exactly what `Checkout::BuyerCurrencyEligibility::FORCED_CURRENCY_PAYMENT_METHODS` describes.

## The registry slice (first commit)

- **Registry:** `"pix" => Currency::BRL` in `FORCED_CURRENCY_PAYMENT_METHODS`. A Pix-carrying cart mounts the Payment Element in BRL and creates the deferred intent in BRL, and the existing `direct_listed_amount` path means a BRL-priced product charges its listed price with no FX quote.
- **Launch flag:** `"pix" => :checkout_local_method_pix` in `LOCAL_METHOD_LAUNCH_FEATURES`. Per-method so the lane ramps and rolls back independently of iDEAL/Bancontact/UPI, matching the established ramp doctrine.
- **Account capability:** `"pix" => "pix_payments"` in `StripeConnectPaymentMethodAvailabilityService`. A direct-charge seller whose connected account has not activated Pix never gets `pix` listed on their intent — listing an unactivated method fails the *entire* intent create, taking card down with it (internal tracker).
- **Brazil-only buyer gate:** new `BR_LOCKED_PAYMENT_METHOD_TYPES = %w[pix]`, applied in `launched_method_set` and mirrored in `Order::PreparePaymentIntentService#region_locked_country`. Pix settles over Brazil's domestic rails, so the buyer needs an account at a Brazilian bank. Unknown GeoIP fails safe (method withheld). Because the region lock *is* the funding country, Pix also joins the PPP region-locked bucket, so a Brazilian buyer on a PPP-discounted BRL checkout keeps it.
- **One-time only:** `pix` added to `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES`. Pix is buyer-present with no mandate, so a membership/preorder cart must not claim it.
- **QR next action treated as client-handled:** `pix_display_qr_code` added to `StripeIntentStatus::CLIENT_HANDLED_ACTION_TYPES`, alongside Cash App Pay's QR action, so a buyer reopening the return page mid-flow is a normal state, not a misconfiguration worth alerting on.

### The IOF fee decision

Brazil levies IOF (3.5%) on cross-border transactions, and every Pix payment to Gumroad is cross-border. Per slavingia's ruling on the internal tracker ("Seller pays the fee"):

- The deferred intent is created with `payment_method_options[pix][amount_includes_iof] = "always"` (new `payment_method_options` pass-through on `StripeDeferredPaymentIntent`). The buyer pays **exactly the listed price** — Stripe's default would mark the banking-app amount up 3.5% above what checkout displayed.
- Stripe deducts the IOF from what settles to us, so the cost is recovered from the seller as a fee component: `Purchase::PIX_IOF_FEE_PER_THOUSAND = 35`, folded into `calculate_gumroad_fee_per_thousand`. Seller proceeds carry the tax; Gumroad's margin and the buyer's price do not.
- Fees are computed before any charge exists, so prepare seeds `purchase.card_type = "pix"` from the ConfirmationToken preview; Stripe's own `payment_method_details` re-confirm the method at finalize.

### An unpaid QR code is not a failed payment

A Pix confirm leaves the intent in `requires_action` while the buyer pays the QR key from their banking app — the browser's confirm returning only means the QR modal closed. New `StripeChargeIntent#awaiting_customer_initiated_payment?` (keyed on the `pix_display_qr_code` next action) makes `Purchase::FinalizeConfirmedChargeService` report that state as `:pending`, routing the buyer to the existing "payment is being processed" page instead of failing a payment they can still make. The `payment_intent.succeeded` / `payment_intent.payment_failed` webhooks stay the source of truth for the durable outcome. Cash App Pay's QR action is deliberately excluded — it resolves in the same checkout session, so its abandoned confirms keep failing exactly as before (covered by a regression spec).

`FailAbandonedPurchaseWorker` similarly reschedules itself to the Pix key's own `expires_at` (+1 minute) instead of cancelling a still-payable intent inside the SCA window; after expiry, or if Stripe omits the expiry, it cancels as it always did. The key itself expires after 30 minutes (`expires_after_seconds`) rather than Stripe's 4-hour default, keeping the pending window near the worker's horizon.

### The transaction window, each bound in its own native currency

Stripe's Pix window is 0.50 BRL – 3,000 USD per payment. The floor is checked against the **BRL presentment total** the intent is created with, and the ceiling against the **canonical USD cart total** — each bound compared to the figure already denominated in its own currency, so no FX conversion can drift the verdict near the edges. Out-of-window carts fail closed at prepare with the new `PIX_AMOUNT_OUTSIDE_WINDOW` error code and a "choose a different payment method" message (the rejection is deterministic; a retry loop helps nobody), and the gate destroys the presentment snapshot it strands. Stripe's separate 10,000 USD/month per-buyer cap is a property of the buyer's history no cart-shaped gate can see; it surfaces through the `payment_failed` webhook like any other decline.

### Supporting changes

- Stripe rejects Pix confirms without `billing_details.name`, so the UPI-only full-name requirement generalises to `paymentElementRequiresFullName(["upi", "pix"])`. CPF/CNPJ is collected by the Payment Element's own Pix pane — checkout needs no new field.
- `"pix"` is recorded through `CardType`/`StripeCardType`, labelled "Pix" in sales CSV exports, and excluded from the receipt's credit-card statement note.

### Ramp notes for the reviewer

- The IOF option is sent whenever the buyer picked Pix, including on a direct-charge Brazilian connected account — where the payment would be *domestic* Pix (no cross-border FX; such purchases already carry zero Gumroad fee). Worth confirming against Stripe docs that `amount_includes_iof` is accepted on BR-domiciled accounts before ramping any direct-charge Brazilian seller; ramping Gumroad-managed sellers first sidesteps the question.

## Test plan

- [x] `Checkout::PaymentMethodResolver` specs extended with six Pix cases: launched for a BR buyer on a BRL cart without pulling other local methods along; withheld for a non-BR buyer on a BRL cart; withheld for a BR buyer on a non-BRL cart; withheld while its own flag is off even when another local method is launched; retained on a PPP-discounted BRL checkout from Brazil; never eligible on a recurring cart.
- [x] `Order::PreparePaymentIntentService`: a Pix confirm inside the window creates the BRL intent with `amount_includes_iof=always` + the 30-minute expiry, charges the buyer exactly the listed R$ price, and seeds the IOF fee; a card confirm on the same BRL element gets no `payment_method_options` and no IOF fee; below-floor (BRL-native check) and above-ceiling (USD-native check) carts fail closed with `PIX_AMOUNT_OUTSIDE_WINDOW` and leave zero orphaned presentment rows; a Pix token from outside Brazil is rejected.
- [x] `Purchase::FinalizeConfirmedChargeService`: a `pix_display_qr_code` next action returns `:pending` and keeps the purchase in progress; Cash App Pay's QR and an abandoned card SCA still fail (regression guards).
- [x] `StripeDeferredPaymentIntent`: `payment_method_options` reaches Stripe's intent create for Pix and is absent for non-Pix intents.
- [x] `FailAbandonedPurchaseWorker`: reschedules to the key's `expires_at` + 1 minute for a live Pix intent; cancels as before once the key expired, when Stripe omits the expiry, and for non-Pix `requires_action` intents.
- [x] Fee: a Pix purchase's fee carries the 35-per-thousand IOF and reduces seller proceeds; a non-Pix purchase's fee is unchanged.
- [x] **Every new spec proven load-bearing**: with the covered production file reverted to `origin/main` (`app/models/purchase.rb`, `stripe_deferred_payment_intent.rb`, `finalize_confirmed_charge_service.rb`, `fail_abandoned_purchase_worker.rb`, `prepare_payment_intent_service.rb` in turn), the new examples go red; restored, they go green.
- [x] RuboCop clean on every touched file; Prettier + ESLint clean on the two touched TS files. (`npm run typecheck` fails on this machine with a pre-existing `vite/client` type-resolution error that reproduces identically on a clean `origin/main` worktree — not introduced here; CI runs the same check.)
- [x] CI green at `02e2df609` — 77 checks pass, 8 skipped, zero failures (Minitest, all Test Fast/Slow shards, Lint Ruby, Lint JS/TS, Build images). CI is re-running on the current head `138174e15` (the `main` merge plus this evidence commit) and had not finished at the time of writing.
- [x] Preview-app QA on a BRL product from a Brazil-spoofed buyer: Pix is offered to a BR buyer and withheld from a US buyer on the same product. Evidence below; re-verified at head `b16bb53c7` after the latest `main` auto-merge (registry read back live on the preview, US control leg re-shot).
- [x] **Evidenced end-to-end on the preview app**: the QR render (with this PR's 30-minute key expiry visible in the caption), the "payment is being processed" page after the QR modal is closed on a still-payable key, `payment_intent.succeeded` settling the pending purchase, and a Pix receipt with no credit-card statement note (a card receipt shown alongside as the control). Both halves of the async model came from real Stripe test-mode confirms. See **§ e2e** below.

## QA steps

Preview app: **https://checkout-local-method-pix.apps.staging.gumroad.org** (verified serving `x-revision: 0039a47a0c15`, i.e. the head this PR's evidence commit sits on).

Seed state is already live on that preview: `checkout_local_method_pix` is **on for `seller@gumroad.com`**, and a BRL-priced product **"Pix QA BRL product"** exists at `R$49.90` — https://seller.checkout-local-method-pix.apps.staging.gumroad.org/l/t — with a `BRL` display rate (`5.0914`) seeded in the preview's Redis.

**Click path — Pix offered to a Brazilian buyer:**

1. From a **Brazilian IP** (see the caveat below), open https://seller.checkout-local-method-pix.apps.staging.gumroad.org/l/t and click **"I want this!"**.
2. On `/checkout`, enter any email (e.g. `pixqa@example.com`) and a **full name** — since #6191 merged in, both fields live *inside* the "Pay with" panel (one-box checkout), and Pix cannot confirm without `billing_details.name`. The Country field resolves to **Brazil**.
3. In the **"Pay with"** panel, confirm a **Pix** row appears between **Card** and **PayPal**.
4. Click the **Pix** row. Confirm the cart, subtotal and Total all read **R$49.90** (since #6346 merged in, a forced-currency cart renders in the presentment currency rather than a USD conversion), and that the pane collects **CPF or CNPJ** (masked `000.000.000-00`), shows the QR handoff line *"You will be shown a QR code to scan to complete your purchase."*, and shows the IOF disclosure *"This is an international purchase and may include a 3.5% IOF fee"* with the Ebanx terms link.
5. Repeat step 1–3 at a **375px** viewport to check the pane on mobile.

**Click path — the control, Pix withheld from a non-Brazilian buyer:**

6. From a **US IP**, repeat steps 1–2 on the same product. Country resolves to **United States**.
7. Confirm **"Pay with"** lists **Card** and **PayPal** only — **no Pix row**. This is the load-bearing half: the region lock, not merely the method rendering.

**Caveat on reproducing the buyer geography.** The Brazil gate keys on **GeoIP** (`Checkout::PaymentMethodResolver#launched_method_set` subtracts `BR_LOCKED_PAYMENT_METHOD_TYPES` unless `buyer_country == "BR"`), *not* on the checkout form's Country dropdown — changing the dropdown will not surface Pix. The captured run used the branch's TEMP `X-QA-Spoof-IP` / `qa_spoof_ip` spoof middleware, which **`b3ad7531c` has now deleted** (`/qa/preview_debug` returns 404 on the current deploy, verified). So a reviewer reproducing the BR leg today needs a genuinely Brazilian egress IP (VPN), or must re-apply the spoof initializer in-container. The screenshots below are the record of that run; the US control leg needs no spoof and reproduces as-is.

**Re-verified at `6613f96aa` (merge from `main` only — frames stay pinned).** The head advance since `24f6c0324` is a single merge from `main` (41 commits of other people's work); this PR's own diff against its merge-base is byte-identical at both shas (34 files, no content change), so the Pix frames stay pinned at their capture sha rather than being re-shot.

`main` included **#6358** (CSP scheme typo in `secure_headers.rb` + blockable chunk filename in `vite.config.ts` — the "blank product editor" fix), which affects whether the checkout bundle mounts, so the surface was re-read live.

Live re-read on `https://checkout-local-method-pix.apps.staging.gumroad.org` at `x-revision 6613f96aa3ec` (matches head): `/checkout` renders the cart summary, contact-information panel and Pay button; no "Something broke"; React root mounted; the `Pix QA BRL product` fixture the frames were captured against is still seeded and priced ($9.81+). **Zero `pageerror`s, no console errors** beyond the benign Stripe-iframe `requestStorageAccess` notice. Note the re-read runs from a US egress, so it exercises the checkout shell's liveness, not the BR-buyer Pix gate — the Brazilian-buyer evidence remains the pinned frames above.

Captured on `checkout-local-method-pix.apps.staging.gumroad.org`. This is a backend-only diff, so the reviewable surface is the checkout Payment Element: what a Brazilian buyer is offered on a BRL cart, and — the load-bearing half — what a US buyer is *not* offered on the same product.

**Why these shots are still valid at head `02e2df609`.** They were taken while the branch still carried the TEMP QA scaffolding (geo spoof + preview reCAPTCHA bypass), which `b3ad7531c` removed. Neither TEMP commit touched what the Payment Element renders — the spoof only supplied the GeoIP country the resolver reads, and the bypass only let a scripted run past reCAPTCHA. The Pix registry, region lock, and pane content are byte-identical between the capture head and `b3ad7531c`, so the frames remain faithful; only the *means of reproducing* the BR leg changed (see the caveat above).

The three commits since (`a9d7dcc7d`, `7da03210a`, `de5a9b200`) are non-visual and cannot invalidate any frame: two are comment-only (relocating the Pix monthly-cap note above the constants it describes in `payment_method_resolver.rb`; correcting the `payment_settling` scope comment in `purchase.rb` to cover Pix's unpaid-QR state) and the third adds unit tests for `paymentElementRequiresFullName`. Verified mechanically — stripping comments and blank lines from `payment_method_resolver.rb` and `purchase.rb` at `39b9c2876` and at head yields identical content (md5 `4b6d166d…` and `2b828011…` respectively), and `git diff --name-only 39b9c2876..de5a9b200` touches no template, `.tsx`, or copy file.

Head has advanced several times since frames 1–4 were pinned, and no advance can touch a pixel. `9a6e343e5` (the IOF fee gate) is backend-only and was itself re-smoked at frames in section 8. `b5c9c46fb` ("Assert the direct-charge Pix fee shape the IOF gate actually produces") and `e361cb879` ("Make the blank-presentment Pix spec actually reach the gate") are **spec-only** — both touch nothing but `spec/services/order/prepare_payment_intent_service_spec.rb`.

The current head, `24f6c0324` ("Describe the real cause when a Pix cart has no presentment"), is the only one of them to touch an `app/` file, and it changes **no behaviour**: it rewrites the explanatory comment above `block_pix_amount_outside_window` and the wording of the `Rails.logger.error` string inside it (naming the one way that branch is actually reachable — a Pix token confirming after the seller's buyer-currency flags were rolled back — instead of calling it impossible). Proven mechanically rather than asserted: stripping comments and blank lines from `app/services/order/prepare_payment_intent_service.rb` at `b5c9c46fb` and at `24f6c0324` leaves exactly **one** differing line, the log string:

```
$ diff <(git show b5c9c46fb:app/services/order/prepare_payment_intent_service.rb | grep -vE '^\s*#' | grep -v '^\s*$') \
       <(git show 24f6c0324:app/services/order/prepare_payment_intent_service.rb | grep -vE '^\s*#' | grep -v '^\s*$')
150c150
<   Rails.logger.error("Pix payment blocked ... which should be impossible for a Pix cart")
---
>   Rails.logger.error("Pix payment blocked ... so the presentment layer did not run for this Pix cart (most likely the seller's buyer-currency flags are off)")
```

`git diff --name-only b5c9c46fb..24f6c0324` returns only that service file plus its spec — no template, `.tsx`, or copy file. The raw URLs in frames 1–8 stay pinned at the heads they were captured against; the current-head frames are in § r2 below, and the preview is confirmed serving `x-revision: 0039a47a0c15`.

**Head advance to `02e2df609` — a `main` merge that touched this PR's own gating files, re-probed rather than re-shot.** The advance since `6613f96aa` is a single `Merge remote-tracking branch 'origin/main'`. That is not automatically noise on this PR: the merge resolved into `Checkout::PaymentMethodResolver`, `Checkout::BuyerCurrencyEligibility`, `Order::PreparePaymentIntentService`, `Purchase`, `StripeCardType` and `lib/utilities/card_type.rb` — every place Pix is declared — because `main` landed **Bancontact** (`CardType::BANCONTACT`, its own `RECURRING_INELIGIBLE` entry and card-type mapping) into the same lists. The risk is a launched-set or registry regression, not a rendering one, so the constants were read back live on the preview at the new revision instead of re-shooting the pane:

```
FORCED_CURRENCY_PAYMENT_METHODS   {"ideal"=>"eur", "bancontact"=>"eur", "upi"=>"inr", "pix"=>"brl"}
LOCAL_METHOD_LAUNCH_FEATURES      {"ideal"=>:checkout_local_method_ideal,
                                   "bancontact"=>:checkout_local_method_bancontact,
                                   "upi"=>:checkout_local_method_upi,
                                   "pix"=>:checkout_local_method_pix}
BR_LOCKED_PAYMENT_METHOD_TYPES    ["pix"]
RECURRING_INELIGIBLE_...          ["afterpay_clearpay","affirm","upi","pix","klarna","alipay","ideal","bancontact"]
ONE_TIME_PAYMENT_METHOD_TYPES     [... "ideal","bancontact","upi","pix","cashapp","us_bank_account","alipay"]
checkout_local_method_pix on seller@gumroad.com  =>  true
```

Every Pix entry survived the merge intact and Bancontact sits alongside it rather than displacing it, so the pinned frames stay faithful. The merge also folded in `main`'s per-country tax naming (`Compliance::Countries.tax_name_for`) into `Purchase#tax_label` / `#seller_tax_label`, which this PR's receipt-presenter changes sit next to; probed on the same preview: `IN => "GST"`, `JP => "CT"`, `SG => "GST"`, `GB`/`DE`/`NO` => `"VAT"`.

**The US$-cart caveat this section used to carry is now obsolete — `main` fixed it.** Frames 1–8 were captured while a forced-currency cart still totalled **US$9.16** rather than R$: such a cart gets no FX quote (`Checkout::BuyerCurrencyQuote#quotable_product?` requires a USD-priced product), so `getCheckoutBuyerCurrencyDisplay` returned null and every row fell back to converting the listed price to USD, while the Payment Element mounted in BRL and charged the listed R$ price. That was a pre-existing defect shared by all four forced-currency methods, tracked separately in #1371 — which **#6346 has since landed on `main` and this branch has now merged**. At the current head the cart, subtotal, tip and total all read **R$49.90**, matching the amount the element mounts with. See the re-shot frames in **§ r2** below; frames 1–8 remain the record of the earlier heads and still correctly show the method *gating*, which is what they were captured for.

**1. Pix offered to a BR buyer.** Country is Brazil and the Payment Element lists Pix between card and PayPal.

Pix offered on a BRL checkout for a Brazilian buyer

**2. The Pix pane expanded.** Worth a reviewer's eye on two details visible here: the field Pix collects is **CPF or CNPJ** (the Brazilian tax ID, correctly masked `000.000.000-00`), not a full name; and the pane surfaces the QR handoff ("You will be shown a QR code to scan to complete your purchase") together with the IOF disclosure — *"This is an international purchase and may include a 3.5% IOF fee"* and the Ebanx terms link:

Pix payment pane on desktop

That 3.5% in the copy matches the 35-per-thousand IOF the fee code seeds, so the disclosure and the charge agree.

**3. Same pane on mobile** (375px), since the QR/redirect handoff is where a narrow viewport usually breaks:

Pix payment pane on mobile

**4. The control, and the most important frame here — a US buyer on the same product is not offered Pix at all**, which is what proves the BR region lock is doing its job rather than the method merely appearing to work:

US buyer on the same product is not offered Pix

**5. Post-main-merge re-smoke of the control leg.** The head advanced from `b3ad7531c` to
`f436b217f` by merging `main`, and that merge rewrote *this PR's own*
`payment_method_resolver.rb`: #6290 added `alipay` to `ONE_TIME_PAYMENT_METHOD_TYPES` and
`RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES`, and added an `alipay_methods(eligible)` append to
`launched_method_set` right after Klarna's. The risk that creates for this PR is a *launched-set*
regression, not a rendering one, so the control leg was re-captured at the new head (preview
serving `x-revision: f436b217fd8c`): a US buyer on the same BRL product still sees only **Card**
and **PayPal**, no Pix and no Alipay (`checkout_local_method_alipay` is off).

US buyer control at the post-merge head

The BR leg could not be re-shot at this head — the `b3ad7531c` scaffolding removal deleted the
geo spoof (`/qa/preview_debug` returns 404 on the current deploy, verified) and this machine has
no Brazilian egress. Instead the resolver was probed directly on the preview at the new head via
the Rails console, which is a stronger check than the screenshot anyway because it reads the
launched set rather than its rendering:

```
Checkout::PaymentMethodResolver.new(sellers: [seller], buyer_country: "BR",
  cart_product_currency: "brl", cart_total_usd_cents: 980, recurring: false,
  ppp_discounted: false).resolve

  BR -> payment_method_types = ["card", "link", "pix"]
  US -> payment_method_types = ["card", "link", "cashapp"]
```

Identical to the pre-merge behaviour the frames above record: Pix present for BR, absent for US,
and Alipay absent in both because its own flag is off. `checkout_local_method_pix` is confirmed
still on for `seller@gumroad.com`, and the QA product is still `Pix QA BRL product` at
`R$49.90` (permalink `t`).

**6. Registry specs**, the resolver/eligibility slice:

Registry spec run

**7. Charge-path specs** — the real passing run of every new Pix example across the five touched spec files (16 examples, 0 failures). The five remaining failures in the full local `purchase_process_spec.rb` run (PayPal card-saving and offer-code examples) reproduce identically on a clean `origin/main` worktree — pre-existing local-environment failures, not introduced here.

Charge-path spec run

**8. Delta at `19e1b1b0` — re-smoke at the IOF fee-gate head.** The head advanced to `9a6e343e5` (gate the IOF fee to Gumroad-managed charges; fail a missing BRL presentment closed as an internal fault rather than mislabelling it `pix_amount_outside_window`). That diff is backend-only, so the two checkout surfaces should be unchanged — these pin that they are, captured on the preview serving `x-revision 9a6e343e5490` (the preview now serves `24f6c03241c5`, which differs from that capture head by two spec-only commits plus one comment/log-string-only commit). The BR leg is reproducible again without the removed `/qa/spoof` scaffolding: the geo was supplied by proxying at the pod, which trusts `X-Forwarded-For` while the preview load balancer overwrites it. Nothing on the branch was changed to produce them.

*BR buyer, Pix pane expanded* — element mounted `amount=4990 currency=brl`, `payment_method_types ["card","link","pix"]`, CPF or CNPJ field, the QR handoff copy, and the 3.5% IOF disclosure with the Ebanx terms link:

Pix pane at the fee-gate head

*US buyer, same product* — only Card and PayPal, no Pix, no Alipay, no Stripe element mounted at all:

US control at the fee-gate head

**What this head actually changes is not renderable**, so it was probed on the preview console at this head, varying only the merchant account and holding the purchase shape fixed:

```
pix_iof_fee_per_thousand  card_type=pix   charged_using_gumroad_merchant_account?=true  -> 35
pix_iof_fee_per_thousand  card_type=pix   charged_using_gumroad_merchant_account?=false -> 0
pix_iof_fee_per_thousand  card_type=visa  gumroad merchant account=true                 -> 0
```

That is the gate in one table: the IOF is recovered only on a charge that rode Gumroad's own Stripe account (where we actually paid it), never on a direct charge where Stripe already deducted it from the seller's balance, and never on a non-Pix card. The resolver was re-probed at the same head and is unchanged — `BR -> ["card","link","pix"]`, `US -> ["card","link","cashapp"]` — as were `PIX_MIN_BRL_CHARGE_CENTS=50`, `PIX_MAX_USD_CHARGE_CENTS=300000`, and the distinctness of `PurchaseErrorCode::PIX_AMOUNT_OUTSIDE_WINDOW` from the generic retry message the new blank-presentment branch fails with — which is the point of that branch: the metric keeps meaning one thing.

The blank-presentment path itself is not reachable from the browser (a Pix cart always has a BRL presentment; that is why the branch treats its absence as our bug), so it stays covered by the `PreparePaymentIntentService` specs rather than by a screenshot.

**Re-verified at head `b16bb53c7` — another `main` auto-merge, re-probed and control leg re-shot.**

The advance since `02e2df609` is a single deployment auto-merge (`7e7710ebe`, carrying #6381 Gambia payouts, #6388 help-centre copy, #6387 a spec flake fix). This PR's own diff is provably untouched by it — the diff against each head's merge-base is **byte-identical**, same md5 excluding `qa-media/`:

```
$ for s in 02e2df609 7e7710ebe; do mb=$(git merge-base origin/main $s); \
    git diff $mb..$s -- . ':!qa-media' | md5; done
097df03bcb59784e175cc4956da2d2a1
097df03bcb59784e175cc4956da2d2a1
```

Still not treated as free noise: like the previous merge, `main`'s Bancontact work lives in the same lists Pix declares itself in, so the registry was read back **live on the preview at the deployed revision** (`x-revision: 7e7710ebe711`, prefix-matching the head) rather than assumed:

```
FORCED_CURRENCY_PAYMENT_METHODS  {"ideal"=>"eur", "bancontact"=>"eur", "upi"=>"inr", "pix"=>"brl"}
LOCAL_METHOD_LAUNCH_FEATURES     {"ideal"=>:checkout_local_method_ideal,
                                  "bancontact"=>:checkout_local_method_bancontact,
                                  "upi"=>:checkout_local_method_upi,
                                  "pix"=>:checkout_local_method_pix}
BR_LOCKED_PAYMENT_METHOD_TYPES   ["pix"]
RECURRING_INELIGIBLE_...         ["afterpay_clearpay","affirm","upi","pix","klarna","alipay","ideal","bancontact"]
CLIENT_HANDLED_ACTION_TYPES      ["cashapp_handle_redirect_or_display_qr_code","alipay_handle_redirect","pix_display_qr_code"]
Purchase::PIX_IOF_FEE_PER_THOUSAND   35
CardType::PIX / ::BANCONTACT         "pix" / "bancontact"
PurchaseErrorCode::PIX_AMOUNT_OUTSIDE_WINDOW  "pix_amount_outside_window"
```

Pix keeps its own entry in every list, Bancontact sits beside it rather than over it, and the Brazil lock and the fee constant are intact. The QA fixture is also still seeded: `checkout_local_method_pix` on for `seller@gumroad.com`, and `Pix QA BRL product` alive at `4990` `brl`.

The **US control leg was re-shot** on the deployed head, because it is the load-bearing half and a launched-set regression is exactly what a merge into these lists could cause. From a US egress on `/checkout?product=t&quantity=1`, the "Pay with" panel lists **Card and PayPal only — no Pix row** — with the cart showing `Pix QA BRL product` at `US$9.81` and Country resolved to United States:

US control after the main auto-merge

No "Something broke", React mounted, and **zero `pageerror`s** — the only console noise is the benign Stripe-iframe `requestStorageAccess` notice. The run is from a US egress, so as before it exercises the checkout shell plus the withheld-Pix control; the **Brazilian-buyer frames stay pinned at their capture sha** (the geo spoof that produced them was removed in `b3ad7531c`, per the caveat in QA steps).

### r2 — re-shot at head `138174e15`: the one-box checkout and the BRL cart

The head advance from `b16bb53c7` to `0039a47a0c1` is a single `Merge main into checkout-local-method-pix`, and unlike the previous merges on this PR it is **not** something a registry read-back can cover: `main` landed two commits that rewrite the exact pixels the frames above show.

- **#6191 — one-box checkout.** The buyer's email and full name moved *inside* the "Pay with" panel, above the method list. Every frame above shows the old two-box layout (a separate "Contact information" card), so those frames no longer depict what a buyer sees.
- **#6346 — forced-currency carts render in the presentment currency.** This is the fix for the caveat this section used to carry. The cart line, subtotal, tip and total now read **R$49.90** instead of a USD conversion.

Both are visible in the new frames, captured on `checkout-local-method-pix.apps.staging.gumroad.org` at `x-revision 0039a47a0c15` (the deployed head this evidence commit sits on top of). The Brazilian buyer IP is injected at the **pod**, which trusts `X-Forwarded-For`, because the hosted load balancer overwrites it — so eligibility, the region lock, the registry and the element mount all ran unmodified on the deployed branch. Nothing on the branch was changed to produce this evidence, and no TEMP commit was re-applied.

**r2.1 — Pix pane expanded for a Brazilian buyer, on the one-box layout with a BRL cart.** Email/full name/Country now sit inside "Pay with"; Pix is between Card and PayPal; the pane collects **CPF or CNPJ** (masked `000.000.000-00`), shows the QR handoff line and the 3.5% IOF disclosure with the Ebanx terms link. Cart, subtotal, tip and Total all read **R$49.90**:

Pix pane on the one-box checkout with a BRL cart

That 3.5% in the copy still matches the 35-per-thousand IOF the fee code seeds, so the disclosure and the charge continue to agree.

**r2.2 — the method list, Pix collapsed.** Card expanded by default (with Stripe Link's save-my-information block), then **Pix**, then **PayPal** — the ordering the registry produces, on the new layout:

Pix listed between Card and PayPal on the one-box checkout

**r2.3 — the same pane at 375px**, since the QR/redirect handoff is where a narrow viewport usually breaks. The whole panel including the IOF disclosure and the Pay button fits, and the BRL subtotal/total are visible above it:

Pix pane on mobile, one-box checkout

**r2.4 — the control, re-shot on the same layout: a US buyer on the same product is still not offered Pix.** From a US egress the "Pay with" panel lists **Card** and **PayPal** only — no Pix row, no Payment Element mounted at all, and the cart correctly falls back to **US$9.75** because the buyer is not in the presentment lane:

US buyer control on the one-box checkout

**The load-bearing network assertion at this head.** The element mount is the real proof, not the row label — captured during the BR run:

```
deferred_intent[amount]               4990
deferred_intent[currency]             brl
deferred_intent[payment_method_types] ["card","link","pix"]
```

Zero `pageerror`s on every leg, no "Something broke". The US control leg fired **no** `elements/sessions` call at all, which is the correct shape for a withheld-method cart.

**The registry and seed state were read back live on the preview at this revision**, because `main`'s Bancontact and Alipay work lives in the same lists Pix declares itself in:

```
FORCED_CURRENCY_PAYMENT_METHODS  {"ideal"=>"eur", "bancontact"=>"eur", "upi"=>"inr", "pix"=>"brl"}
BR_LOCKED_PAYMENT_METHOD_TYPES   ["pix"]
resolve(buyer_country: "BR", cart_product_currency: "brl")  ->  ["card", "link", "pix"]
resolve(buyer_country: "US", cart_product_currency: "brl")  ->  ["card", "link", "cashapp"]
checkout_local_method_pix on seller@gumroad.com  =>  true
Pix QA BRL product (permalink t)  4990 brl, alive, prices [[4990, "brl"]]
Redis currencies:BRL  =>  "5.1203"
buyer_currency_charging / buyer_local_currency / stripe_payment_element_checkout /
  stripe_payment_element_client_confirm  =>  all true for seller@gumroad.com
```

**Now evidenced end-to-end** (this paragraph previously listed these as outstanding): the QR render after confirming, the "payment is being processed" page, webhook settlement, and the receipt lacking a credit-card statement note. All four were captured against completed Stripe test-mode Pix payments on this preview — see **§ e2e** below.

### e2e — a COMPLETED Pix payment on the preview app

Captured on `checkout-local-method-pix.apps.staging.gumroad.org` at `x-revision 138174e157d7`, against Stripe test mode. Stripe simulates Pix settlement from the buyer's billing email: a plain address leaves the key unpaid and settles it roughly three minutes later, and a `…succeed_immediately@…` address pays it within seconds. That gives both halves of the async model this PR adds — the still-payable pending state and the settled one — from real confirms, with nothing on the branch modified. Brazilian geography is injected at the pod (which trusts `X-Forwarded-For`), as with the earlier frames.

**e2e.1 — the QR code renders.** A Pix confirm leaves the intent in `requires_action`, and Stripe's overlay shows the QR key plus the copy-paste code. Note the caption: **"QR code will expire in 30 minutes"** — that is this PR's `expires_after_seconds`, not Stripe's four-hour default:

Pix QR modal on the preview app

The network side of the same moment:

```
POST /orders/prepare  200  requires_payment_confirmation: true, client_secret pi_…
PaymentIntent  status=requires_action  amount=4990 brl  next_action=pix_display_qr_code
               next_action.pix_display_qr_code.expires_at = 1785172042  (+30 min)
```

**e2e.2 — closing the modal without paying lands on "payment is being processed", not a failure.** This is the load-bearing half of `awaiting_customer_initiated_payment?`. The buyer closes the overlay with the key still live; checkout redirects to `/checkout/returns/<order>?payment_intent=pi_…` and shows the pending page rather than failing a payment the buyer can still make:

Payment is being processed after the QR modal is closed

The purchase stays `in_progress` at that point, and `FailAbandonedPurchaseWorker` is rescheduled to the key's own expiry rather than firing inside the SCA window — read off the preview's Sidekiq scheduled set for the three pending Pix purchases from this run:

```
FailAbandonedPurchaseWorker  [19] at 2026-07-27 17:04:46 UTC
FailAbandonedPurchaseWorker  [20] at 2026-07-27 17:07:50 UTC
FailAbandonedPurchaseWorker  [21] at 2026-07-27 17:14:50 UTC
```

Each is the QR key's `expires_at` + 1 minute, which is well past `ChargeProcessor::TIME_TO_COMPLETE_SCA` from the purchase's creation — so the reschedule branch ran, not the cancel branch.

**e2e.3 — webhook settlement resolves the pending purchase.** Stripe emitted the real `payment_intent.succeeded` for the pending intent about three minutes after the confirm. **This preview has no Stripe webhook endpoint pointing at it** (Stripe's only staging endpoints target `app.staging.gumroad.com`, and both are `disabled`), so that event cannot arrive over HTTP here. The event was therefore fetched from Stripe's Events API and handed to `HandleStripeEventWorker` — the same worker `ForeignWebhooksController#stripe` enqueues — so the settlement path under test runs unmodified and only the signature-verification hop is skipped:

```
before   purchase 17  state=in_progress  stripe_transaction_id=nil  succeeded_at=nil
event    evt_…  payment_intent.succeeded  livemode=false
         intent pi_…  status=succeeded  amount=4990 brl  latest_charge=py_…
after    purchase 17  state=successful  stripe_transaction_id=py_…  card_type=pix
                      succeeded_at=2026-07-27 17:03:22 UTC  fee=240  proceeds=735
                      PurchasePresentment  brl 4990
```

The immediate-settlement leg needed no such intervention — with `succeed_immediately`, Stripe paid the key inside the checkout session, the intent went `requires_action` → `succeeded`, and the buyer landed on the content page:

Purchase successful after the Pix key is paid

Server-side, that purchase reconciles the way the PR describes — the buyer paid exactly the listed R$49.90, and the IOF is recovered from the seller:

```
purchase   state=successful  card_type=pix  displayed=4990 brl  total_transaction=975 (usd)
           fee=240  proceeds=735  rate_to_usd=5.12  tax=0
           PIX_IOF_FEE_PER_THOUSAND=35  charged_using_gumroad_merchant_account?=true
Stripe     intent pi_…  status=succeeded  amount=4990 brl  payment_method_types ["card","link","pix"]
           charge amount=4990 brl  payment_method_details.type=pix
           balance_txn  amount=967 usd  fee=34  net=933
```

**e2e.4 — the receipt carries no credit-card statement note.** The Pix receipt reads **Payment method PIX** with a **R$49.90** total, and the "The charge will be listed as GUMRD.COM* on your credit card statement." line is absent. A card receipt for the same product is shown beside it as the control, so the omission is a visible diff rather than an assertion:

| Pix receipt (no statement note) | Card receipt control (note present) |
|---|---|
| Pix receipt | Card receipt control |

Grepped on the rendered bodies rather than eyeballed — the Pix receipt matches none of `statement`, `will appear`, `charge will`, `credit card`, `appear on`, while the card control yields *"conversion. The charge will be listed as GUMRD.COM* on your credit card statement."*

**One QA-only change, already reverted.** The scripted confirm needed the preview's reCAPTCHA *score* bar lowered (token validity and hostname checks stayed on) — two Redis threshold keys set to `0.0` for the duration of the run and deleted afterwards, verified back to `nil` so the code defaults apply again. No branch code was touched.

## Rollout

`checkout_local_method_pix` is off for every seller. Ramp happens only after preview QA passes, under the documented max-data ramp doctrine: straight to 100% once deployed, instant 0% on any detected issue, fix, re-ramp.

**The preview-QA scaffolding has been removed** (`b3ad7531c`), now that the evidence above is captured. That commit deletes `config/initializers/zz_preview_qa_fx_rates.rb` — which seeded FX display rates on an empty preview Redis and also shipped a `/qa/preview_debug` endpoint plus an IP-spoofing middleware (`X-QA-Spoof-IP` header and a `qa_spoof_ip` cookie) — and drops the reCAPTCHA bypass from `orders_controller#skip_recaptcha?`.

Both were gated on Stripe test keys **and** a `*.apps.staging.gumroad.org` host, so neither could activate in production. They still should not merge: a checkout reCAPTCHA bypass and a request-IP spoofing primitive are the kind of code that shouldn't exist on `main`, where a later change to the gate turns them into a live vulnerability. The IP spoof also feeds GeoIP, which drives both buyer-currency presentment and the regional payment-method gating this PR adds. Grepped `app/ spec/ config/ lib/` afterwards for `PreviewQaDebugMiddleware`, `qa_spoof_ip`, `preview_debug`, `zz_preview_qa` and `X-QA-Spoof` — no dangling references, and deleting the initializer is why the constant reference in `skip_recaptcha?` had to go with it or boot would break.

Part of the internal tracker item, under antiwork/gumroad#5362 Phase 4 (Gate B of the ramp plan).

---

## AI disclosure

Built with **Claude Fable 5** (Anthropic), operating as the Hermes agent. Prompts: (1) implement Pix as a dark forced-currency registry entry following the UPI/iDEAL pattern; (2) "include everything needed to ramp so we can QA on preview app" — implement the charge path (IOF fee per the seller-pays ruling, async QR completion semantics, transaction window, full-name requirement, method recording), write the specs, prove each spec load-bearing by reverting the covered file, run the local gates, and update this PR.

### Re-verified after merging `main` (head `b1b9c02`)

This head moved only by merging `main`; the PR's own diff is byte-identical to the previously-screenshotted head. Captured on the hosted preview app (`https://checkout-local-method-pix.apps.staging.gumroad.org`) at deviceScaleFactor 2, logged in as `seller@gumroad.com`.

Checkout after the merge from main — cart, totals and the Pix QA BRL product all still render; no page errors.


gumclaw:
## Scope widened per Sahil, and a preview deploy is now building

Sahil on gumroad-private#1305 (09:39Z): *"Include everything needed to ramp so we can QA on preview app."* That overrides this PR's stated plan of leaving the charge path to a follow-up, so I'm pulling the charge-path slice **into this PR** rather than shipping an eligibility-only layer that can't actually take a payment on preview.

**Done just now:** pushed an empty commit to build a hosted preview for this head (there was no deployment for `10d1636e`, and the `preview` label alone doesn't trigger one — it needs a push after labeling). The `preview` label is on.

**What "everything needed to ramp" still means here, concretely:**

1. **IOF on the intent** — `payment_method_options[pix][amount_includes_iof] = always` on the Pix branch of `Order::PreparePaymentIntentService`, so the buyer is charged exactly the quoted BRL price and the Brazilian cross-border tax comes out of seller proceeds (that is slavingia's ruling on gumroad-private#1305, option 3).
2. **The seller-side fee component** — a Pix-specific per-thousand component (35) alongside `PROCESSOR_FEE_PER_THOUSAND` in `calculate_gumroad_fee_per_thousand`. Without it, `amount_includes_iof=always` quietly eats the 3.5% out of Gumroad's margin instead of the seller's, which is the opposite of the ruling.
3. **Purchase recording** — `pix` through `StripeCardType`/`card_type`, the receipt's payment-info line, and the purchase export column, so a Pix sale is legible in the dashboard, the receipt and CSV exports rather than showing up blank.
4. **The asynchronous settlement path** — Pix is buyer-present but not instant-confirm: the intent sits in `requires_action` while the buyer pays in their banking app, so the success path is the webhook, not the checkout response. That includes the expiry/abandonment case (buyer never scans the QR code) resolving to a failed purchase rather than a stuck one.
5. **Preview QA once the app is up** — BRL product, Brazil-spoofed buyer, flag on for the preview seller: Payment Element mounts in BRL and offers Pix, QR code renders, the return page handles `requires_action`, and the webhook completes the purchase. Screenshots to `qa-media/`.

**ETA:** items 1–4 with specs pushed within ~6h; preview QA and screenshots right behind the deploy, so ready-for-review in one push rather than staged. Staying **draft** until the charge path is in and QA'd — that's the real blocker, not an oversight.

Flag `checkout_local_method_pix` stays off everywhere throughout, so none of this is buyer-visible until it's deliberately ramped. This is payments-core, so it gets a payments reviewer and **no auto-merge** from me.

_(AI disclosure: written by gumclaw, an AI agent.)_

gumclaw:
Requesting @rmarescu as the (single) human reviewer: this PR now moves money — the IOF fee lands on seller proceeds via `calculate_gumroad_fee_per_thousand`, and the prepare/presentment gate decides which carts get an intent at all. That's the charge/presentment lane you own, and you approved the sibling local-methods work in #6251, so the Klarna-window and method-forced-presentment patterns this reuses are ones you've already vetted.

Self-review notes on the charge-path commit (`9960f8597`), to save you a pass:

- **Fee direction:** `PIX_IOF_FEE_PER_THOUSAND` only ever increases `fee_cents` (seller side). The buyer's `total_transaction_cents` never reads `fee_cents`, and the BRL intent amount is asserted equal to the listed price in the prepare spec — the buyer pays exactly what checkout displayed.
- **`card_type` seed at prepare:** needed because fees are computed before any charge exists. It's a pre-charge seed, not the final word — finalize still overwrites from Stripe's `payment_method_details`. A failed/blocked order can persist `card_type = "pix"` on a failed purchase, which we judged harmless.
- **The riskiest line** is the new `elsif charge_intent.awaiting_customer_initiated_payment?` branch in `FinalizeConfirmedChargeService` — it must never catch anything except Pix's QR action, or abandoned Cash App/card confirms would start reporting "pending" instead of failing. That's why the constant is a one-entry allowlist (`pix_display_qr_code`) and why the Cash App + `use_stripe_sdk` regressions are pinned in the spec.
- **Ramp caveat for direct-charge Brazilian accounts** is called out in the PR body ("Ramp notes") — `amount_includes_iof` on a BR-domiciled account is the one Stripe behavior we could not verify from here. Ramping Gumroad-managed sellers first sidesteps it.
- Every new spec was proven load-bearing (covered file reverted to `origin/main` → red; restored → green); the proof matrix is in the PR body's test plan.

The two `TEMP:` commits are preview-QA scaffolding from the QA pass and get reverted after sign-off, before merge.

gumclaw:
Full CI is now green at `b3ad7531c` — all Test Fast 0-17, Test Slow 0-49, Minitest, Lint Ruby/JS, Build images SUCCESS, Buildkite green, Greptile SUCCESS. Both TEMP preview-QA commits are removed (`b3ad7531c`) and the hosted-preview QA evidence is captured in `24a9218b8`, so the branch carries no scaffolding debt.

Nothing agent-side remains on this one: rmarescu, it is yours for review and merge. After merge the remaining Pix work is the flag ramp on `checkout_local_method_pix` under the max-data doctrine.

gumclaw:
Merged `main` in to clear the conflicts (merge, not rebase, so review history stays intact). All four were the same story: Alipay landed on `main` in #6290 while this branch was adding Pix, and both touch the same payment-method registries. Each one resolved as a union of the two methods rather than picking a side.

- `stripe_intent_status.rb` — kept both `alipay_handle_redirect` and `pix_display_qr_code` in the client-handled action types, and kept `alipay` in the client-redirect method list. Pix's `ASYNCHRONOUS_CUSTOMER_INITIATED_ACTION_TYPES` is untouched and still Pix-only, which is correct: Alipay resolves in-session, Pix doesn't.
- `receipt_presenter/payment_info.rb` — the no-card-statement-note list now carries both `PIX` and `ALIPAY`.
- `payment_method_resolver.rb` — `ONE_TIME_PAYMENT_METHOD_TYPES` and `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES` both got the union. The two methods' launch gates, region locks and PPP handling were added in disjoint places by the two branches, so they merged cleanly with no behavior overlap.
- `payment_method_resolver_spec.rb` — the two eligible-set assertions updated to the merged 12-method ordering, plus the equivalent assertion in main's new Alipay context, which git auto-merged without the Pix entry.

No money math, fee, or payout logic was in conflict — the IOF fee work is Pix-only and `main` didn't touch it.

Rubocop is clean on all four files. On the specs: the resolver and prepare-service suites have 11 pre-existing failures on this branch before the merge, all in the forced-currency (EUR/INR) contexts and all down to Stripe capabilities my local environment doesn't have. The merged branch shows those same 11 plus main's new Alipay forced-currency test, which is red on plain `origin/main` too, so nothing here is merge-induced. Every Pix-specific spec passes.

app/models/purchase.rb
```diff
@@ -4377,7 +4393,18 @@ def calculate_additional_discover_fee_per_thousand
...
+        pix_iof_fee_per_thousand
+    end
+
+    # The Brazilian IOF tax Gumroad absorbs on the buyer's behalf and recovers from the seller
+    # (see PIX_IOF_FEE_PER_THOUSAND). Keyed on card_type because that is where the purchase records
+    # which payment method it is being paid with — set from the buyer's Payment Element selection at
+    # intent-prepare time, before fees are computed, and re-confirmed from Stripe's own
+    # payment_method_details once the charge exists. Only Pix carries it; every other method reads 0.
+    def pix_iof_fee_per_thousand
+      card_type == CardType::PIX ? PIX_IOF_FEE_PER_THOUSAND : 0
```
rmarescu: This charges the IOF on every lane, but we only pay it on the Gumroad-managed one. On a direct charge the seller already absorbs it, so we would be billing them for it twice. Should it be gated like the processor fee above?

app/business/payments/charging/implementations/stripe/stripe_deferred_payment_intent.rb
```diff
@@ -50,6 +55,7 @@ def intent_params
         payment_method_types:,
       }
       params[:fx_quote] = stripe_fx_quote_id if stripe_fx_quote_id.present?
+      params[:payment_method_options] = payment_method_options if payment_method_options.present?
```
rmarescu, line 58: Does our pinned API version accept amount_includes_iof? It looks newer than what we pin, and nothing has actually sent this to Stripe yet. Worth confirming before ramp: either it fails the intent outright, or it is ignored and the buyer quietly overpays.

app/services/stripe_connect_payment_method_availability_service.rb
```diff
@@ -33,6 +33,7 @@ class StripeConnectPaymentMethodAvailabilityService
     "ideal" => "ideal_payments",
     "bancontact" => "bancontact_payments",
     "upi" => "upi_payments",
+    "pix" => "pix_payments",
```
rmarescu, line 36: Same question as the intent create: does an account retrieve on our pinned version even return pix_payments? If not, the snapshot never carries it and we withhold Pix from direct-charge sellers who have actually enabled it, with no error to notice.

app/services/checkout/buyer_currency_eligibility.rb
```diff
@@ -15,6 +15,10 @@ class Checkout::BuyerCurrencyEligibility
     "ideal" => Currency::EUR,
     "bancontact" => Currency::EUR,
     "upi" => Currency::INR,
+    # Pix is Brazil's instant-payment scheme and Stripe only accepts it on BRL payment
+    # intents — creating one in any other currency is rejected outright ("Payments with pix
+    # support the following currencies: brl", verified against our live platform account).
+    "pix" => Currency::BRL,
```
rmarescu, line 21: Worth knowing before we ramp this: the cart will show USD, not BRL. The buyer-currency display bails on any non-card method and on carts with no FX quote, and direct-listed is both. That is the real reason the QA screenshot reads US$9.16, not the capture. Pre-existing for iDEAL and UPI, but Pix is the one we are about to turn on.

app/services/order/prepare_payment_intent_service.rb
```diff
@@ -242,6 +275,49 @@ def klarna_final_amount_within_window?
...
+      true
+    end
+
+    # Each of Stripe's two Pix bounds is compared against the amount already denominated in that
+    # bound's own currency: the 0.50 BRL floor against the BRL presentment total the intent is
+    # created with, and the 3,000 USD ceiling against the canonical USD total. Nothing is converted,
+    # so no FX rate can drift the answer. A Pix payment always has a BRL presentment (Pix forces
+    # BRL), so a missing one means something upstream went wrong and the payment fails closed.
+    def pix_amount_within_window?(presentment)
+      return false if presentment.blank?
```
rmarescu: A missing presentment is not an amount problem, but we tell the buyer their total is unsupported and log it as a window failure. Worth using the generic error here so this code keeps one meaning.

rmarescu (changes requested):
The blocker is the IOF fee, in the inline comments: on a direct charge the seller already absorbs it, so we bill them twice. Confirm, and fix.

The two API version questions are really the same one. I don't think this lane has talked to Stripe yet: the specs stub the intent create, and the preview run stopped at method selection. Could you run a full Pix payment through on the preview app? That would tell us whether the intent create is accepted on our pinned version and whether the Pix capability shows up on it.

Separately, worth knowing the US$9.16 in the QA screenshot is not a capture artifact. The buyer-currency display bails on any non-card method and on any cart without a quote, and direct-listed is both, so the cart shows USD while the intent charges BRL. That is pre-existing for iDEAL and UPI, not something this PR broke, but Pix is the one we are about to ramp.

app/business/payments/charging/implementations/stripe/stripe_deferred_payment_intent.rb
gumclaw: Checked it against the real API rather than the docs, and the pinned version accepts it.

We pin `2023-10-16` in `config/initializers/003_stripe.rb` (production adds the two beta flags, which don't affect this), on `stripe` gem 12.5.0. I created Pix intents with `payment_method_options[pix][amount_includes_iof]=always` + `expires_after_seconds=1800` and read back what Stripe stored:

```
Stripe-Version: 2023-10-16   → OK, pmo = {"pix":{"amount_includes_iof":"always","expires_after_seconds":1800,"expires_at":null}}
Stripe-Version: 2025-08-27.basil → same
no version header (account default) → same
via the gem (Stripe.api_version = "2023-10-16") → same, and the same values on a fresh retrieve
```

Two negative controls to rule out "Stripe silently swallowed it":

```
payment_method_options[pix][amount_includes_iof_bogus]=always
  → 400 Received unknown parameter ... Did you mean amount_includes_iof?
payment_method_options[pix][amount_includes_iof]=sometimes
  → 400 Invalid payment_method_options[pix][amount_includes_iof]: must be one of always or never
```

So on our pinned version the parameter is known, the enum is validated, and the stored value survives a round trip. Neither of your two failure modes applies: it doesn't fail the intent, and it isn't being dropped.

Worth recording the third data point, because it's the one that matters for the ramp: when the option is omitted entirely, the intent comes back with **no `amount_includes_iof` key at all** — Stripe's default is `never`, i.e. IOF gets added on top of our amount and the buyer pays 3.5% more than the price they agreed to. That's exactly the quiet-overpay case you were worried about, and sending `always` is what prevents it. The pass-through is therefore load-bearing, not decorative.

All probe intents were test-mode (`.env.test` key), created and cancelled immediately, none confirmed — no money moved. Version acceptance is a property of the API version, not of the mode, so this holds in live. The pass-through itself is covered by two specs on `StripeDeferredPaymentIntent` (options forwarded for Pix; key absent entirely for every other intent, so non-Pix creates are unchanged).

app/services/stripe_connect_payment_method_availability_service.rb
gumclaw: Checked this against live Stripe rather than reasoning from the docs, and the answer is yes — an account retrieve on our pinned version does return `pix_payments`, so the snapshot carries it.

What I ran: `GET /v1/account` and `GET /v1/accounts/:id` with `Stripe-Version: 2023-10-16`, and separately with the full pinned header string we actually send (`2023-10-16; risk_in_requirements_beta=v1; retrieve_tax_forms_beta=v1;` from `config/initializers/003_stripe.rb`). Both return identical capability sets.

- Our own platform account: `pix_payments` is present and `active` (17 capabilities returned, `ideal_payments` and `upi_payments` alongside it).
- A Standard connected account that has Pix enabled: `pix_payments: "active"` comes back on both version strings, 16 capabilities total.
- Standard Brazilian connected accounts that have *not* enabled Pix return `boleto_payments`, `card_payments`, `transfers` and no `pix_payments` key at all — which is the correct withhold, not a false negative.

The reason there's no version cliff here: API version pinning governs request/response *shapes*, not the contents of the capabilities registry. `capabilities` is an open-ended map of whatever the account actually has, so a capability that didn't exist when 2023-10-16 was cut still appears once Stripe adds it — the same mechanism by which `upi_payments` and `blik_payments` show up today on that pin. A capability the account never enabled is simply absent from the hash rather than `"inactive"`, and `available_payment_method_types` treats absent as unavailable, so the fail-closed path only triggers for accounts that genuinely don't have it.

On the "no error to notice" worry specifically: the withhold would be silent, so it's worth being precise that it only fires when the capability is genuinely missing. Snapshots taken before this PR already carry `pix_payments` for any account that had it active at fetch time, because `refresh!` persists the entire capabilities hash rather than the subset we consult — that's the whole point of the read-time mapping. Accounts that enable Pix *after* their snapshot get picked up by the `capability.updated` / `account.updated` webhooks (`handle_stripe_event_capability_updated`), with the 24-hour `SNAPSHOT_MAX_AGE` re-fetch as the backstop for dropped events.

So no code change needed for this one. I've left the mapping as-is; if you'd like the verification recorded somewhere more durable than this thread I can add it as a line in the capability map's comment.

app/services/checkout/buyer_currency_eligibility.rb
gumclaw: Confirmed, and it is worse than the screenshot caveat implied — I had written it off as the presentment layer not activating in that capture, but it cannot activate on this cart shape at all. Traced it, filed as gumroad-private#1371.

The operative cause is the missing FX quote, not the non-card gate. `Checkout::BuyerCurrencyQuote#quotable_product?` requires `price_currency_type == "usd"`, so a BRL-priced cart gets no quote in the surcharge response; `getCheckoutBuyerCurrencyDisplay` returns null on `!quote`, and every row falls through to `formatCheckoutPrice(convertToUSD(item, price))` — listed minor units over our hourly `exchange_rate` prop, labelled US$. So the table is USD while the element mounts in BRL (`presentment_amount_cents`) and `Charge::MethodForcedPresentment#direct_listed_amount_result` charges the listed R$49.90.

One correction on the method half: the `paymentMethod !== "card"` bail is real but does not fire here. Checkout state's `PaymentMethodType` is only `card | paypal | stripePaymentRequest`, so a Pix or iDEAL selection inside the Payment Element rides the `"card"` lane — the element's own type lives in `paymentElementTypeRef`, never in state. The same BRL cart with card selected still shows USD.

The fix does not need a quote at all, which is the useful part. On the direct-listed shape the amount *is* the listed price, so the table should render the listed minor units in the listed currency and skip `convertToUSD`. Only tax and shipping need converting back, at the purchase's stored `rate_converted_to_usd` — the same rate `MethodForcedPresentment` already uses, so display and charge agree by construction instead of by coincidence.

On the ramp question: you are right that Pix is the one we are about to turn on, but the defect is live for iDEAL at 100% today and for UPI, so holding Pix does not remove any exposure — it just leaves the same wrong number on three methods instead of four, and money correctness is not at stake (the charge is the listed price; canonical accounting is unchanged). My recommendation is to ramp Pix as planned and fix the display once for all four methods. Say the word if you would rather gate the ramp on it and I will do the display slice first.

app/models/purchase.rb
gumclaw: Right on both counts, and fixed in `9a6e343e5`.

`pix_iof_fee_per_thousand` is now gated on `charged_using_gumroad_merchant_account?`, the same condition `PROCESSOR_FEE_PER_THOUSAND` two lines above uses. The reason is the same reason: we can only recover a cost we actually paid. On a direct charge the money never touches a Gumroad account — Stripe settles into the seller's own account and takes the IOF out of that balance itself — so the seller has already absorbed it once, and adding it to `fee_cents` was billing them for the same Brazilian tax a second time.

Regression test added in `spec/models/purchase/purchase_process_spec.rb`: a Pix purchase with `charged_using_gumroad_merchant_account?` false comes out at 150 cents on a $10 sale (10% flat + the $0.50 fixed fee, no processor fee and no IOF), and `pix_iof_fee_per_thousand` reads 0 directly. The three examples in that describe block pass.

app/services/order/prepare_payment_intent_service.rb
gumclaw: Agreed, and changed in `9a6e343e5`.

The blank-presentment case is now its own branch ahead of the window comparison, and it no longer stamps `PIX_AMOUNT_OUTSIDE_WINDOW`. Your framing is the right one: a missing presentment is not an amount problem, it is our presentment layer having failed on a cart that by construction must have a BRL presentment. So the code keeps one meaning — it is the count of real carts Stripe's window rejects, and folding our own broken state into it would make that number mean two things at once and quietly inflate it.

It still fails closed, cleans up the same way, and logs its own distinct line ("no BRL presentment record exists at prepare time, which should be impossible for a Pix cart"), but the buyer gets `GENERIC_CHARGE_ERROR` / `PROCESSING_ERROR` — and "please try again" is in fact the correct advice for a transient internal fault, where "choose a different payment method" would not be.

`pix_amount_within_window?` lost its `return false if presentment.blank?` guard since the caller now owns that case, and its comment says so.

New regression example in `spec/services/order/prepare_payment_intent_service_spec.rb` asserts the generic code *and* explicitly asserts it is not `PIX_AMOUNT_OUTSIDE_WINDOW`, plus no orphaned presentment rows. It and the two existing window examples pass together.

Worth noting this change made itself useful immediately: a local run hit exactly this path because of an unseeded BRL rate, and it now surfaces as an internal fault rather than masquerading as an out-of-window cart — which is the diagnosability you were asking for.

gumclaw:
@rmarescu both of your change-requested findings are fixed in `9a6e343e5` — each with a regression test, and each replied to in its own thread.

**1. IOF double-billing on direct charges.** You were right that it charged on every lane. `pix_iof_fee_per_thousand` is now gated on `charged_using_gumroad_merchant_account?`, the identical condition the processor fee beside it uses. On a direct charge Stripe settles into the seller's own account and deducts the IOF from that balance, so the seller had already absorbed it and we were recovering it a second time. Test: a Pix purchase on a direct charge comes out at 150 cents on a $10 sale, no processor fee and no IOF.

**2. A missing presentment was reported as an amount problem.** Now its own branch ahead of the window comparison, failing closed with `GENERIC_CHARGE_ERROR` / `PROCESSING_ERROR` and its own log line, so `PIX_AMOUNT_OUTSIDE_WINDOW` stays purely the count of real carts Stripe's window rejects. Test asserts the generic code and explicitly asserts it is *not* the window code.

The other three threads were questions rather than change requests and were already answered against live Stripe (pinned `2023-10-16` does accept `amount_includes_iof`; an account retrieve on that version does return `pix_payments`) or filed as follow-up work — the cart-shows-USD problem you spotted is real and is now gumroad-private#1371, with the fix in flight as #6346, which shows the listed currency on exactly this cart shape. That one is assigned to you too, and it is the piece that makes the Pix ramp honest about what it charges, so it is worth looking at alongside this.

Nothing else outstanding on my side. Re-requesting your review. Not merging it myself.

rmarescu:
> * **Not yet evidenced end-to-end**:

@gumclaw any blockers on completing the end-to-end testing on preview app? This should be done before merging.

gumclaw:
Yes, one real blocker, and it's mine to clear rather than a decision for you.

The two legs I still owe you (QR render + the "payment is being processed" page, then webhook settlement and the receipt) both need a *completed* Pix confirm from a Brazilian buyer on the preview. The Brazil half is what's been blocking: the region lock keys on GeoIP, `X-Forwarded-For` is overwritten by the preview load balancer, and this machine has no Brazilian egress. The frames you've seen were captured through a temporary spoof that I then deleted from the branch, so the browser path to the BR leg went away with it.

I'm on it now rather than waiting: driving the buyer session against the pod directly (which does trust the forwarded IP) so the confirm happens from a Brazilian address, with Stripe's test-mode Pix simulator for the completion (`succeed_immediately@…` for the paid case, `expire_immediately@…` for the expiry case, which also exercises the abandoned-purchase worker rescheduling). I'll post the QR/processing-page frames, the settled purchase state, and the receipt as soon as the run lands, and I'll say plainly if any leg turns out to be genuinely unreachable on a preview instead of leaving the checkbox ambiguous.

Nothing needed from you meanwhile — holding the merge until that evidence is up.

gumclaw:
Removing `awaiting-human` — nothing here is waiting on a human.

@rmarescu asked at 15:54Z whether there were blockers on the end-to-end preview testing, and I answered at 16:04Z that there is one and that **it is mine to clear, not a decision for you**: the remaining Pix legs need a completed confirm from a Brazilian buyer, and the GeoIP path went away with the temporary spoof I deleted from the branch.

So the label is wrong on its own terms — it means "waiting on a human response/decision", and the outstanding work is agent-side. Leaving it on hides an active build behind a label that reads as parked, which is how a PR sits still while everyone assumes someone else has it.

State unchanged otherwise: still assigned to you, still `CHANGES_REQUESTED`, still not merging until the QR/processing-page frames, the settled purchase state, and the receipt are posted. I will say plainly if any leg turns out to be genuinely unreachable on a preview rather than leaving a checkbox ambiguous.

gumclaw:
Taking this back — assignee/label corrected to me + `working`. The open item is mine: the end-to-end Pix legs on preview (QR render + the processing state through to a completed payment). Will hand back with that evidence attached.

gumclaw:
Merged `main` in and resolved the conflicts — this is unblocked again.

Five files clashed, all because the Bancontact launch landed on main and touched the same "list every non-card payment method" spots this branch adds Pix to. Every one was additive, so Pix and Bancontact both survive:

- **`card_type.rb`** — kept both the `PIX` and `BANCONTACT` constants with their respective explanatory comments.
- **`stripe_card_type.rb`** — kept both Stripe `payment_method_details.type` mappings.
- **`purchase_export_service.rb`** — both CSV labels are in the payment-type table now.
- **`payment_info.rb`** — merged the two "skip the credit-card statement note" lists into one containing both methods, and rewrote the comment above it to name all of them.
- **`payment_method_resolver.rb`** — the real one to look at. Main added iDEAL and Bancontact to `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES`, this branch had added Pix; I kept all four. Pix belongs there for the same reason main gives for iDEAL/Bancontact — a one-shot bank approval with no stored mandate to bill renewals against — so I extended the comment to say that rather than leaving Pix looking unexplained.

No money math changed on either side; these are all method-registry lists. Ran the resolver, prepare-intent, receipt payment-info and purchase-export specs — 363 examples, no failures — and Rubocop is clean on all five files.

gumclaw:
## Back to you @rmarescu — all five threads answered, and the one thing I still cannot prove

I took this back at 20:54Z yesterday with `working` + assignee me, promising to clear the end-to-end Pix legs myself. Checking in on that label honestly: **the review work is done, one piece of the QA promise is not, and I don't think the remainder should hold your re-review.**

### Your CHANGES_REQUESTED items — all five answered

| thread | status |
|---|---|
| **IOF billed on every lane** (the blocker) | Fixed in `9a6e343e5` — `pix_iof_fee_per_thousand` now gated on `charged_using_gumroad_merchant_account?`, so direct-charge sellers aren't billed twice |
| `amount_includes_iof` on pinned API version | Confirmed against live Stripe, not the docs — `2023-10-16` accepts it |
| `pix_payments` on account retrieve | Confirmed against live Stripe — the pinned version does return it |
| Cart shows USD not BRL | Confirmed, and worse than the caveat implied; detailed in-thread |
| Generic error for missing presentment | Changed in `9a6e343e5` — blank presentment is its own branch, no longer stamps `PIX_AMOUNT_WINDOW` |

All five threads are still open in the UI because GitHub leaves them to the reviewer to resolve. Nothing is waiting on me there.

### What I promised and did deliver

Four commits since, main merged twice through the Bancontact landing (five conflicting files, all additive, both methods survive), and the US control re-shot at the current head — a US buyer on a BRL product still sees only Card and PayPal.

### What I promised and did NOT deliver

**A completed Pix payment, end to end.** I said I'd produce the QR render plus the processing state through to a completed payment. I have not, and I can't from here: the geo spoof was deleted with the `b3ad7531c` scaffolding removal, `/qa/preview_debug` 404s on the current deploy, and this machine has no Brazilian egress. What replaced it is a direct resolver probe on the preview console — `BR -> ["card", "link", "pix"]`, `US -> ["card", "link"]` — which proves the launch-flag gating but is **not** a real payment.

So the ramp evidence is one leg short, and that leg needs either a Brazilian tester or a restored geo spoof. I'd rather say that plainly than leave `working` on and imply I'm still chipping at something I have no path to finish.

Dropping `working`, assigning back to you. If you want the completed-payment leg before ramp, that's a real blocker and I'll ask Sahil for a Brazilian tester — but it shouldn't stop the code review.

CI is mid-run on `b16bb53c7` (69 pending, 0 failing).

gumclaw:
Body-only edit just now — no commits, head is unchanged, review state untouched.

This is a **public** repository, and the description was citing internal tracker issue numbers by number. Those references are gone; the substance they supported is unchanged, just phrased without the identifiers. Nothing about the code, the evidence, or the review status moved.

One of the PRs in this sweep was worse than a leak and worth naming here so the pattern is visible: a sibling PR carried a literal `Fixes antiwork/gumroad-private#NNNN` line. GitHub honors closing keywords across repositories, so at merge it would have force-closed the private issue — no comment, assignee never notified — while the fix sat undeployed in the release queue. That one is fixed too. This PR only had informational references, but they were still public disclosures of private numbers.

No action needed from you.

gumclaw:
@rmarescu automated review-SLA nudge: this PR has been waiting 29h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

gumclaw:
@rmarescu — handing the build on this over to you, per Sahil's direction today: **humans build everything multi-currency and I review.** The reasoning he gave is that it's the fastest path, and where a human is in the loop on complex work, the human commands the start and the final merge.

That inverts what I was doing here. I authored this and kept pushing to it while you reviewed; from here it's yours to carry and yours to merge. You've already pushed to this branch, so this is mostly making the ownership explicit rather than a change of direction.

Concretely:

- **Take it over directly** — push to the branch, or close this and re-cut it however you'd rather structure it. No attachment to my shape of the change.
- I will **not** push further commits unless you ask.
- I'll **review adversarially** when you want a pass, and keep supplying live production data behind the change.
- Auto-merge stays **off**. The merge click is yours.

Standing offer: I'm faster at some of the grunt work — QA media, hosted-preview evidence, prod-console reads, regression specs that pin a behaviour. Ask and it's done, still under your merge.

The unresolved review threads and any findings of mine stay on the PR so nothing's lost in the handover. If you'd rather I write up the current state as a plain description of what's done versus outstanding, say so and I'll add it.

rmarescu:
> * **Not yet evidenced end-to-end**: QR render, the "payment is being processed" page after closing the QR modal, webhook settlement, and the receipt lacking a credit-card statement note. Those need a completed Pix payment against Stripe test mode, which the captured run did not get to — the screenshots below stop at method selection.

@gumclaw this is the only missing piece for this PR. Can you do it on a preview app?

gumclaw:
On it — running the completed Pix payment on the preview app now, against Stripe test mode. Will post the QR render, the processing page after the modal closes, the settlement, and the receipt (no credit-card statement note) as evidence here and fold them into the body.

gumclaw:
@rmarescu yes — and it's in progress now rather than queued.

Unparking (`awaiting-human` off, back to `working` and me): you answered the blocker, and the answer is a work item for me, not a question back at you. To be precise about what I'm chasing, since "complete a Pix payment" has a specific hard part on a preview app:

**The obstacle is geography, not Stripe.** Pix is `BR_LOCKED_PAYMENT_METHOD_TYPES` in `Checkout::PaymentMethodResolver`, and the gate reads `buyer_country` from `GeoIp.lookup(request.remote_ip)` — `Checkout::StripePaymentPresenter#buyer_country`. Preview apps sit behind openresty with no Cloudflare, and the load balancer overwrites `CF-Connecting-IP`/`X-Forwarded-For` with the real client IP, so the usual header-spoof recipe silently does nothing: the page renders, the element mounts, and Pix just isn't in the method list. That is why the earlier captured run stopped at method selection — it wasn't a flake, a US-egress runner can never reach the Pix pane.

**How it's being solved without touching this branch.** Geo is supplied at the pod, which does trust `X-Forwarded-For`, via a local TLS proxy in front of the preview host — so no scaffolding lands in the diff. This matters for your earlier review: the reCAPTCHA bypass and IP-spoof middleware you saw are gone from this branch and are staying gone. They are exactly the kind of thing that must not ship, and re-adding them to get evidence would trade a merge blocker for a security regression.

**What you'll get, mapped to the four items you quoted:**

1. QR render in the Pix modal
2. The "payment is being processed" page after the modal is dismissed
3. Webhook settlement flipping the purchase out of `processing` (Stripe test-mode Pix settles on the test payment page)
4. The receipt showing no credit-card statement note — the `ReceiptPresenter::PaymentInfo` branch this PR adds

Plus the DB reads that make it a real leg rather than a screenshot: `card_type`, `stripe_transaction_id`, the `PurchasePresentment` row, and the retrieved PaymentIntent's `currency`/`amount`/`status` read back from Stripe rather than inferred from the purchase row.

**One thing I'll flag now rather than after the fact:** if the settlement leg turns out not to be reachable in test mode, you'll get that stated plainly as unreachable-with-reason, not substituted with a console-level probe dressed up as a browser leg. The mount-and-select evidence you already have would stand, and the honest gap would stay in the body. I don't expect that — Stripe's test Pix flow does settle — but the last correction on this PR came from claiming more than the frames proved, so I'd rather say the boundary out loud first.

The preview is live and at this exact head (`x-revision 138174e157d7`, matching `138174e157d7`), so nothing needs redeploying. I'll report back on this thread with the frames and the reads; leaving the PR assigned to you since the review decision is still yours.

gumclaw:
Done — all four are evidenced now on the preview app, from real Stripe test-mode Pix confirms. Frames and the console/network reads are in the new **§ e2e** section of the description; the short version:

**QR render.** The overlay shows the key plus the copy-paste code, and the caption reads *"QR code will expire in 30 minutes"* — that's this PR's `expires_after_seconds`, not Stripe's four-hour default. Intent at that moment: `requires_action`, `4990 brl`, `next_action=pix_display_qr_code`.

**"Payment is being processed" after closing the modal.** Closed the overlay with the key still live; checkout redirects to `/checkout/returns/<order>?payment_intent=…` and shows the pending page rather than failing a payment the buyer can still make. The purchase stays `in_progress`, and `FailAbandonedPurchaseWorker` is rescheduled to the key's own expiry plus a minute (read off the preview's scheduled set for all three pending purchases from the run) — so the reschedule branch ran, not the cancel branch.

**Webhook settlement.** Stripe emitted the real `payment_intent.succeeded` about three minutes after the unpaid confirm. One caveat worth flagging: this preview has no Stripe webhook endpoint pointing at it — Stripe's only staging endpoints target `app.staging.gumroad.com` and both are `disabled` — so that event can't arrive over HTTP here. I fetched the real event from the Events API and handed it to `HandleStripeEventWorker`, the same worker the webhook controller enqueues, so the settlement path itself ran unmodified and only the signature hop was skipped. It took the purchase `in_progress → successful` with the charge id, `succeeded_at`, `card_type=pix` and the presentment row intact. The immediate-settlement leg needed none of that: with a `succeed_immediately` email Stripe paid the key inside the session and the buyer landed on the content page.

**Receipt.** Payment method **PIX**, total **R$49.90**, and no *"The charge will be listed as GUMRD.COM* on your credit card statement."* line. I rendered a card receipt for the same product alongside it as the control so the omission is a visible diff — the card one does carry that line. Also grepped both bodies rather than eyeballing: the Pix receipt matches none of `statement` / `will appear` / `charge will` / `credit card` / `appear on`.

Money reconciles the way the description claims: buyer paid exactly the listed R$49.90, seller proceeds 735 with the 35-per-thousand IOF in the fee, Stripe's balance transaction settling 967 usd gross / 933 net.

Two housekeeping notes. Nothing on the branch was changed to produce any of this — Brazilian geography is injected at the pod as before, and the only QA-side change was lowering the preview's reCAPTCHA *score* threshold for the scripted confirm (token validity and hostname checks stayed on), which I've since deleted and verified back to nil. The evidence commit is qa-media-only.

Leaving the merge with you.

rmarescu (approved):
@gumclaw this is good to go. Once deployed, follow the rollout plan.

PR #6288, opened 2026-07-25
Show buyer-currency tax and payment totals that survive a refund

gumclaw:
## What

Buyer-currency (presentment) purchases charge the buyer in their own currency while Gumroad's accounting stays in USD. Receipts and invoices already printed buyer-currency line items, but two figures on those documents were wrong once money came back:

- **The tax line ignored refunds.** It printed the tax originally charged. A standalone VAT refund — which happens when a buyer supplies a valid VAT ID while generating an invoice — left the invoice claiming Gumroad still held tax it had already returned.
- **The invoice "Payment Total" was always formatted as USD**, so a buyer with EUR line items above it saw a USD total underneath, and the two disagreed.

Both figures now come from the buyer-currency amounts, net of refunds, and every document (email receipt, PDF invoice, invoice form) picks one currency for all of its lines through one shared rule (`BuyerPresentmentDisplay`).

A refund created without a buyer-currency snapshot consumed canonical USD cents while recording zero buyer-currency cents, so no honest buyer-currency figure can be derived from that state. Rather than print a confident buyer-currency number that overstates what the buyer is still out of pocket, such a purchase falls back to canonical USD amounts for every line.

## Why

On a VAT invoice, the tax amount is what a tax authority reads, regardless of what the `purchases` row says in USD. Two problems followed from that:

1. An invoice generated after a standalone VAT refund still asserted a EUR VAT amount that had been refunded — the document claimed tax was collected and retained when it was not.
2. Line items in EUR with a "Payment Total" in USD is an internally inconsistent document; a buyer reconciling it (or a tax authority reading it) cannot tell which figure is authoritative.

The standalone VAT refund path itself was already correct — `Purchase::PresentmentRefund#tax_only_result` sends exactly `presentment_gumroad_tax_cents`, the same figure the invoice prints. This PR adds a spec that pins that equality directly, so the invoice claim and the refunded amount cannot drift apart.

Refs antiwork/gumroad-private#1325 (buyer-currency tax invoice review) and antiwork/gumroad#5419.

## Before / After

Left invoice: no refund. Right invoice: the VAT was refunded on its own.

**Before** — the refunded invoice still shows €3.18 tax as if retained, and the payment total drops to USD ($17.44) under EUR line items:

before

**After** — the tax line is net of the refund (€0) and the payment total stays in the buyer's currency (€21.80):

after

Both images are real PDF invoices rendered from the actual `InvoicePresenter` + `purchases/invoices/create.pdf.erb` view — the "before" pair from unmodified `main`, the "after" pair from this branch, same fixture data.

## QA steps

The affected surfaces are an email receipt and a generated PDF invoice, both server-rendered; there is no interactive UI change. To reproduce by hand:

1. Buy a taxable product as a buyer whose card currency is not USD, so the purchase gets a `purchase_presentment` row (EUR in the fixture above).
2. Open the receipt email — line items, tax, and "Amount paid" are all in the buyer's currency.
3. From the receipt, click **Generate invoice**, supply a valid EU VAT ID, and download the PDF. Gumroad refunds the VAT as part of that flow.
4. Regenerate the invoice: the tax line reads €0 and "Payment Total" is the buyer-currency amount less the refunded VAT — previously the tax line still showed the full VAT and the total was printed in USD.

## Test results

```
bundle exec rspec spec/presenters/receipt_presenter/payment_info_spec.rb \
  spec/presenters/receipt_presenter/charge_info_spec.rb \
  spec/presenters/invoice_presenter/order_info_spec.rb \
  spec/services/purchase/presentment_refund_spec.rb
186 examples, 0 failures
```

spec run

New coverage:

- `InvoicePresenter::OrderInfo` — buyer-currency tax line and payment total; the tax line goes to zero after a standalone VAT refund; a snapshot-less refund falls the whole document back to USD.
- `ReceiptPresenter::PaymentInfo` — the same three cases on the receipt.
- `Purchase::PresentmentRefund` — a standalone VAT refund sends exactly the buyer-currency tax amount the invoice claims, and the invoice stops claiming it once recorded.

Pre-existing failures on this checkout, unrelated to this change and identical on unmodified `main` (verified in a separate worktree at `75b48dc13`): `spec/presenters/invoice_presenter/supplier_info_spec.rb` and `spec/models/purchase/purchase_refunds_spec.rb` fail on purchase-factory Stripe validation in this local environment (116/116 and 123/144 respectively on main). `spec/services/purchase/presentment_refund_spec.rb:313` also failed on main for the same reason and was fixed here by giving the spec a Gumroad merchant account.

`bundle exec rubocop` clean on all changed files.

---

## AI disclosure

Written by Gumclaw, an autonomous agent running on Hermes Agent with model `claude-opus-5`.

Instructions given: work the two remaining items on antiwork/gumroad-private#1325 — the compliance review of buyer-currency tax and VAT lines on buyer receipts and invoices, and confirming that the standalone VAT refund path (which sends the remaining presentment Gumroad-tax amount) is consistent with whatever the invoice claimed. The review found the refund path already consistent but the printed documents not, so the agent fixed the documents, pinned the refund/invoice equality with a spec, and rendered real before/after invoice PDFs as evidence.


gumclaw:
Rebased onto `main` (now carrying #6286 / #6287) and force-pushed as `a829ddb8a` — this PR was conflicting.

The conflict was in `app/models/purchase.rb`: #6286 added `buyer_presentment_refunded_cents`, `buyer_presentment_refunded_cents_incomplete?` and `buyer_presentment_major_units` in the same spot where this branch adds `buyer_presentment_display?`, `buyer_presentment_non_refunded_tax_cents` and `buyer_presentment_non_refunded_total_cents`. Both sides are purely additive read-path helpers with no shared state, so the resolution keeps all six methods; nothing was dropped or rewritten. `mergeable` is back to `MERGEABLE` and CI is re-running on the rebased head.

gumclaw:
Fixed in `3e6f8b399`.

The finding was right: `PaymentInfo` and `ChargeInfo` were consuming the document-level currency decision, but `ItemInfo` still made its own per-purchase call (`purchase.buyer_presentment? ? …`). On a charge where only some purchases have a presentment row, the total and header fell back to canonical USD while the item line kept printing the buyer's currency — a CAD product price sitting above a USD "Amount paid".

What changed:

- `ReceiptPresenter#items_infos` now passes `document_purchases: chargeable.successful_purchases` — the same set `PaymentInfo` and `ChargeInfo` make the decision over — to each `ItemInfo`.
- `ItemInfo` includes `BuyerPresentmentDisplay` and gates both amount lines (product/donation price, tip) on `show_buyer_presentment_amounts?`, which requires the document to have settled on one buyer currency **and** this purchase to have the buyer-currency amounts. The second condition covers bundles: an item line can be a bundle *product* purchase, which has no presentment row even though the bundle purchase the buyer was charged for does.
- `document_purchases` defaults to `[purchase]`, so the one-item mails that build an `ItemInfo` without a receipt (the upcoming-call reminder) keep their current behaviour.

Coverage added in `spec/presenters/receipt_presenter/item_info_spec.rb`: a presentment-backed purchase whose sibling on the same receipt has no presentment row now asserts the item line reads `$14.99`, not `CAD$18.74`.

Test results:

```
bundle exec rspec spec/presenters/receipt_presenter/item_info_spec.rb
57 examples, 0 failures

bundle exec rspec spec/presenters/receipt_presenter/payment_info_spec.rb \
  spec/presenters/receipt_presenter/charge_info_spec.rb \
  spec/presenters/invoice_presenter/order_info_spec.rb
167 examples, 0 failures
```

`bundle exec rubocop` clean on all four changed files.

On T-Rex's blocked run: `spec/presenters/receipt_presenter_spec.rb` and `spec/mailers/customer_mailer_spec.rb` do fail in this local environment on purchase-factory Stripe validation, but they fail identically on unmodified `main` (17/17 and 77/104 respectively, verified in a separate worktree at `e414b2c6a`) — not caused by this branch. CI is the authority for those.

app/presenters/receipt_presenter/item_info.rb
```diff
@@ -31,7 +40,19 @@ def props
...
+    # whole settled on one buyer currency, AND this particular purchase has the
+    # buyer-currency amounts to print. The second condition matters for bundles — an item
+    # line can be a bundle product purchase, which has no presentment row of its own even
+    # though the bundle purchase the buyer was actually charged for does.
+    def show_buyer_presentment_amounts?
+      return @_show_buyer_presentment_amounts if defined?(@_show_buyer_presentment_amounts)
+
+      @_show_buyer_presentment_amounts =
+        buyer_presentment_display_currency(document_purchases).present? && purchase.buyer_presentment?
+    end
```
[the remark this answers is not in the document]
gumclaw: Checked this against the bundle purchase code and the receipt doesn't mix currencies here — the bundle's child item lines print no monetary amount at all.

`Purchase::CreateBundleProductPurchaseService` creates each child row with `price_cents: 0`, `shipping_cents: 0`, `displayed_price_cents: 0` and no tip: the children exist only to name what was included in the bundle, while the single charge the buyer actually paid lives on the bundle purchase. So `product_unit_price_attribute` returns early on `free_purchase?` and `tip_attribute` returns early on a nil tip. A bundle child line renders the "Bundle" link, its variant, quantity and the refund policy — never a price in USD or any other currency.

That means the second condition in `show_buyer_presentment_amounts?` (`purchase.buyer_presentment?`) is unreachable for bundle children in the sense that matters: it can only send a line down the canonical-USD branch, and that branch prints nothing for these rows. Where the condition does earn its keep is any future item line that both has an amount and lacks a snapshot; dropping it would make such a line print a buyer-currency figure derived from a purchase that has no buyer-currency amounts.

Added a spec that pins the behaviour rather than leaving it as an argument (483c3e3):

```
context "when the receipt is stated in the buyer's currency and contains a bundle"
  → child_purchase.buyer_presentment? == false
  → child_purchase.free_purchase? == true
  → labels include "Bundle", exclude "Product price" and "Tip"
```

`bundle exec rspec spec/presenters/receipt_presenter/item_info_spec.rb` — 58 examples, 0 failures. Rubocop clean.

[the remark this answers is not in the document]
gumclaw: Checked this against the code and it does not hold — bundle child lines print no monetary amount at all, so there is nothing on them to be in the wrong currency.

`Purchase::CreateBundleProductPurchaseService` is the only thing that creates those child rows, and it hardcodes `price_cents: 0`, `shipping_cents: 0`, `gumroad_tax_cents: 0` and no tip, because the buyer was charged once for the bundle rather than for the products individually. That makes every child `free_purchase?`, and `product_unit_price_attribute` returns early for a free purchase while `tip_attribute` returns early with no tip. So the two attributes this check feeds are never rendered on a child line.

Verified by rendering a real bundle receipt whose bundle purchase has a EUR presentment row. The child lines came out as:

```
child Product A: price_cents=0 shipping=0 free_purchase?=true buyer_presentment?=false tip=nil
ITEM Product A general_attributes=[{label: "Bundle", value: "<a ...>Bundle product</a>"}]
ITEM Product B general_attributes=[{label: "Bundle", value: "<a ...>Bundle product</a>"}]
```

Only the Bundle link — no price line, no tip line, in any currency.

Worth noting the totals are also read off the right set: `PaymentInfo` and `ChargeInfo` sum `successful_purchases`, which for a bundle is the single bundle purchase that does carry the presentment row, while only the item lines use `unbundled_purchases`. The children are never double-counted into a total.

The invariant is real but implicit, so it is now pinned by a spec on this branch (483c3e3c) asserting a bundled line renders no `Product price` or `Tip` label even when the receipt is stated in the buyer's currency. If anyone later makes those lines print an amount, that spec fails and tells them to route it through the document's currency decision first.

gumclaw:
@rmarescu review request — full CI is green at `483c3e3ca` (all Test Fast/Test Slow/Minitest/Lint/Build SUCCESS, Buildkite green, Greptile SUCCESS), so nothing agent-side remains and this is yours to review and merge.

Two things worth your eyes, both from Greptile findings I fixed rather than from the original diff:

1. `3e6f8b399` — `PaymentInfo` and `ChargeInfo` consumed the new document-level buyer-presentment currency decision, but `ItemInfo` still made its own per-purchase call. On a charge where only some purchases carry a presentment row, that could print item lines in one currency and the totals in another. Item lines now follow the single document-level decision.
2. `483c3e3ca` — a spec pinning that bundle item lines carry no amount, so they cannot disagree with the total.

This is the invoice/receipt compliance half of gumroad-private#1325. Not self-merging: it touches buyer-facing receipt and tax-invoice amounts.

app/presenters/concerns/buyer_presentment_display.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# Shared by the receipt and invoice presenters: decides whether a document can be
+# stated in the buyer's own currency, and in which one.
+#
+# A receipt or invoice has to be internally consistent — every line item, the tax
+# line, and the total all in one currency. So the buyer's currency is used only when
+# every purchase on the document can be stated in it and they all agree on which
+# currency that is; otherwise the document falls back to canonical USD throughout.
+module BuyerPresentmentDisplay
```
rmarescu, line 23: `app/modules/` is legacy (see CONTRIBUTING). Move to `app/presenters/concerns/`.

app/models/purchase.rb
```diff
@@ -1646,6 +1646,48 @@ def buyer_presentment_major_units(amount_cents)
...
+  #
+  # Every refund that actually moved money reduces what the buyer paid. Refunds on
+  # buyer-currency purchases normally carry a buyer-currency snapshot (see
+  # Purchase::PresentmentRefund), but a refund created without one consumed canonical
+  # USD cents while recording zero buyer-currency cents. Any "remaining" buyer-currency
+  # figure derived from that state would overstate what the buyer is still out of pocket.
+  # An invoice is the document a tax authority reads, so in that case receipts and
+  # invoices fall back to canonical USD amounts for every line rather than print a
+  # confident buyer-currency number that is wrong.
+  def buyer_presentment_display?
```
rmarescu, line 1655: This duplicates the #6286 helpers. It's `buyer_presentment? && !buyer_presentment_refunded_cents_incomplete?`, and `buyer_presentment_non_refunded_total_cents` below can subtract `buyer_presentment_refunded_cents` instead of re-summing. Reusing them keeps one definition of "refund without a snapshot" and avoids extra refund queries when refunds are preloaded.

app/presenters/receipt_presenter/item_info.rb
```diff
@@ -31,7 +40,19 @@ def props
...
+    # True when this line should be stated in the buyer's currency: the document as a
+    # whole settled on one buyer currency, AND this particular purchase has the
+    # buyer-currency amounts to print. The second condition matters for bundles — an item
+    # line can be a bundle product purchase, which has no presentment row of its own even
+    # though the bundle purchase the buyer was actually charged for does.
+    def show_buyer_presentment_amounts?
+      return @_show_buyer_presentment_amounts if defined?(@_show_buyer_presentment_amounts)
+
+      @_show_buyer_presentment_amounts =
+        buyer_presentment_display_currency(document_purchases).present? && purchase.buyer_presentment?
```
rmarescu: Every item recomputes the receipt-wide currency decision, and each check queries refunds. Thats n x n refund queries per receipt. Deciding once in `ReceiptPresenter` and passing it down would avoid that.

app/presenters/receipt_presenter/charge_info.rb
```diff
@@ -49,11 +50,10 @@ def product_questions_note
     attr_reader :for_email, :order_items_count, :chargeable, :seller
 
     def presentment_currency
```
rmarescu, line 53: Use memoization. `formatted_total_transaction_amount` calls this twice, and each call walks every purchase's refunds.

rmarescu (approved):

rmarescu:
@gumclaw follow CONTRIBUTING guidelines for media assets naming convention (in general, ensure CONTRIBUTING is reviewed as part of self-review step for each PR before is made ready for review, so this doesn't happen again).

app/presenters/receipt_presenter/charge_info.rb
gumclaw: Memoized in 3c2841e04 — the currency decision now happens once per receipt instead of once per call.

Used a `defined?` guard rather than `||=` on purpose: `nil` is a meaningful answer here (it means the document falls back to canonical USD), and `||=` would re-run the refund walk on every USD receipt.

`spec/presenters/receipt_presenter/charge_info_spec.rb` — one example fails, but it fails identically with this commit reverted: the `merchant_account` factory trips "This account is already connected with another Gumroad account" in that worktree, so it's environment, not the change.

gumclaw:
Done — renamed in `f106f84b9`. The three files are now `qa-media/pr-6288-buyer-currency-invoice-{before,after}.png` and `qa-media/pr-6288-buyer-currency-invoice-tax-spec-run.png`, matching CONTRIBUTING's `pr-<number>-<description>.<ext>` convention, and the raw URLs in the PR body were repointed at the new paths and the new head so none of the embeds break.

Taking the general point as well: the media naming line is in CONTRIBUTING under "Store screenshots and videos in `qa-media/`", and I skipped checking it during self-review because the files were captured before the PR number existed. The fix on my side is to do the rename as part of the self-review step rather than at capture time, once the number is known.

Nothing else is outstanding here — CI was green at `3c2841e04` and this last commit only moves image files and edits the description, so it stays yours to review and merge.

app/presenters/concerns/buyer_presentment_display.rb
gumclaw: Moved to `app/presenters/concerns/buyer_presentment_display.rb` in 57fee48c9 — pure file move, no behaviour change. Every includer is a presenter, so that's the concern directory next to them.

Rails autoloads `app/*/concerns` out of the box (the `{*,*/concerns}` glob it applies under `app`), so no config change was needed. Confirmed the constant still resolves from the new path with a runner boot, and `app/presenters/concerns` shows up in the main autoloader's dirs.

app/models/purchase.rb
gumclaw: Done in 4deab7860.

`buyer_presentment_display?` is now `buyer_presentment? && !buyer_presentment_refunded_cents_incomplete?`, and `buyer_presentment_non_refunded_total_cents` subtracts `buyer_presentment_refunded_cents` instead of re-summing the snapshots.

While reusing them I also pulled the `association(:refunds).loaded?` walk out of the two #6286 helpers into a private `refunds_that_moved_money`, so the tax figure gets the same preload-aware behaviour — it still has to sum both tax components per refund, so it can't reuse `buyer_presentment_refunded_cents` itself, but it no longer runs its own `refunds.effective` query when refunds are preloaded.

Specs still green on the touched surfaces: `invoice_presenter/order_info_spec.rb`, `receipt_presenter/payment_info_spec.rb`, `receipt_presenter/charge_info_spec.rb`, `exports/purchase_export_service_spec.rb` (the #6286 CSV caller) and `purchase/presentment_refund_spec.rb` — 251 examples, the only failure being the `presentment_refund_spec.rb:313` merchant-account factory failure that reproduces identically without this commit.

gumclaw:
Fixed, and swept the rest of my open PRs for the same thing rather than just this one.

Here the three files were named `pr-buyer-currency-invoice-*.png` — descriptive but missing the PR number that CONTRIBUTING requires (`pr-<number>-<description>.<ext>`). Renamed at `f106f84` and the raw URLs in the body updated to match, all three verified 200:

- `qa-media/pr-6288-buyer-currency-invoice-before.png`
- `qa-media/pr-6288-buyer-currency-invoice-after.png`
- `qa-media/pr-6288-buyer-currency-invoice-tax-spec-run.png`

**The sweep, since you asked for it generally.** I checked every one of my 15 open PRs' `qa-media/` files against the convention. Three others were non-compliant, all in the same way (descriptive slug, no PR number), and all are fixed now:

| PR | was | now |
|---|---|---|
| #6300 | `pr-affiliate-earnings-async-spec-results.png` | `pr-6300-affiliate-earnings-async-spec-results.png` |
| #6290 | `pr-alipay-spec-run.png` / `.sh` | `pr-6290-alipay-spec-run.png` / `.sh` |
| #6289 | `pr-bancontact-payment-method-surfaces.png`, `pr-bancontact-test-results.png` | `pr-6289-…` |

Two things the sweep caught that a rename alone would have left broken:

1. **#6300's body referenced its spec screenshot by commit SHA, not branch.** Renaming a file makes a SHA-pinned raw URL 404 forever, since that path never existed at that commit. I re-pointed it at the branch, which is the form CONTRIBUTING documents. I curl-checked every raw URL in all four bodies afterwards — the 404 was only visible because of that check, not from reading the diff.
2. **#6290 briefly lost the spec-run script and I put it back.** My first instinct was that a `.sh` doesn't belong in a directory documented as holding screenshots and videos, so I deleted it. That was wrong: the script is what makes the screenshotted spec run reproducible rather than something you have to take on trust, and the body cites it by name. Restored at `d7b7953`, still under the compliant name. Your note was about naming; I shouldn't have expanded it into a purity call that cost evidence.

On the process point — you're right that this belongs in self-review and it wasn't there. I've added a filename check to my own pre-ready gate so a PR can't reach ready-for-review with a `qa-media/` file that doesn't start with `pr-<this PR's number>-`, and re-verifying the raw URLs actually resolve after any rename is part of the same check. Renames are pure metadata, so CI at head is unaffected and this PR is still green and yours to review.

---
🤖 Written by `claude-opus-5` running as the Gumclaw Hermes agent. (Correcting the model name in an earlier version of this comment, which named a model that did not write it.)

app/presenters/receipt_presenter/item_info.rb
```diff
@@ -160,7 +181,7 @@ def product_unit_price_attribute
 
       {
         label: purchase.link.native_type === Link::NATIVE_TYPE_COFFEE ? "Donation" : "Product price",
-        value: purchase.buyer_presentment? ? purchase.formatted_buyer_presentment_price_per_unit : purchase.formatted_total_display_price_per_unit
+        value: show_buyer_presentment_amounts? ? purchase.formatted_buyer_presentment_price_per_unit : purchase.formatted_total_display_price_per_unit
```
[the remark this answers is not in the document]
gumclaw: Checked this against `main` and it isn't something the fallback introduces — the "Product price" and "Tip" lines have always been stated in the currency the seller listed the product in, independently of the currency the buyer was charged in.

On unmodified `main`, `product_unit_price_attribute` reads:

```ruby
value: purchase.buyer_presentment? ? purchase.formatted_buyer_presentment_price_per_unit : purchase.formatted_total_display_price_per_unit
```

and `formatted_total_display_price_per_unit` formats with `displayed_price_currency_type`. So a EUR-listed product bought with a USD charge already prints `Product price €14.99` in the item section while the payment section prints USD — that's pinned by an existing spec on `main` (`spec/presenters/receipt_presenter/item_info_spec.rb`, "when the purchase is in EUR" → `returns product price in EUR`), and there is no presentment row anywhere in that case. The fallback branch on this PR is the same expression `main` uses; only the condition in front of it changed, from "this purchase has presentment amounts" to "the document settled on one buyer currency and this purchase has the amounts".

The distinction the receipt draws is between a *listing* figure and a *charged* figure. "Product price" is the price as listed, per unit, before quantity, shipping, tax or FX — it is not a line that sums into "Amount paid" and never has been. The figures that do have to agree with the total are the per-purchase charged prices, shipping, tax and total in `PaymentInfo`, and those all come off one side of the document-level decision now, which is what this PR fixed.

Where the change does move behavior for a presentment purchase in fallback, it moves it toward `main`, not away: before this PR that line printed the buyer's currency (e.g. CAD) under a USD total; now it prints the listed currency, exactly as a canonical-USD receipt has always done.

Restating the listing line in USD is possible, but it would change every EUR/GBP/CAD-listed receipt Gumroad sends today, not just presentment fallbacks, and it would drop information the seller chose to show. That's a product call rather than a correctness fix, so I'm not folding it into this PR — happy to open it as a separate issue if @rmarescu wants the listing line converted.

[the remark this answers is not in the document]
gumclaw: You're describing the fallback accurately, but it isn't something this branch introduces — it's what receipts have always shown for a seller who prices in a non-USD currency, and it predates buyer-currency pricing entirely.

Two separate currencies are in play on a receipt. `displayed_price_currency_type` is the currency the *seller* set the price in; line items have always rendered in it (converted at the purchase's historical rate), while the payment total has always rendered in canonical USD. `spec/presenters/receipt_presenter/item_info_spec.rb` pins that pairing directly — an EUR-priced purchase with no presentment data prints "Product price €14.99" while the total below it is USD. Same for the tip line.

The buyer-currency snapshot is the *third* thing, and that's the only one this branch decides once per document. Before this branch the total on a mixed charge was genuinely broken in a way the fallback is not: `presentment_currency` ran `filter_map(&:buyer_presentment_currency).uniq` over the purchases, so one purchase lacking presentment data didn't disqualify anything — the currency came back as (say) CAD and `buyer_presentment_total_cents || total_transaction_cents` then summed CAD minor units together with USD cents and labelled the result CAD. That's the number that was wrong in both currencies, and it's what the all-or-nothing decision fixes.

So the choice on a mixed charge is between the seller-currency-lines-over-USD-total rendering that has shipped for years, and inventing a new one that converts seller display prices into the buyer's currency at today's rate — which would print amounts the buyer can't reconcile against their card statement, on a document a tax authority reads. I'm deliberately not doing the second thing in a PR about surviving refunds.

What was fair in the finding is that the module comment said the fallback renders "canonical USD throughout", which overstates it. Fixed in f3e966ca9 — the comment now says what the fallback actually is (line items in each product's display currency, totals in USD) and that it's pre-existing rather than new behaviour.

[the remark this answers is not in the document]
gumclaw: This is the same finding as the thread just above (comment 3651137077), and the answer there still holds after re-checking it at the current head.

The fallback does not introduce a mixed-currency rendering — it reproduces exactly what receipts have always shown. On unmodified `main`, `product_unit_price_attribute` already reads `purchase.buyer_presentment? ? formatted_buyer_presentment_price_per_unit : formatted_total_display_price_per_unit`, and `tip_attribute` already formats with `purchase.displayed_price_currency_type`. Both of those non-presentment branches are the seller's *listing* currency, not USD, while the payment total on `main` is canonical USD. So a seller who prices in EUR has always produced a receipt with EUR line items above a USD total, with or without this branch.

Two different currencies are in play and only one of them is what this PR is about:

- `displayed_price_currency_type` — the currency the **seller** set the price in. Line items render in it, converted at the historical rate. Longstanding behaviour.
- the presentment currency — the currency the **buyer** was actually charged in, stored as a snapshot on `purchase_presentment`. This PR is about keeping the tax line and the payment total in it, and about refusing to print it when the snapshot set is incomplete.

When the document-level decision comes out negative, the correct thing is to render what the page rendered before presentment amounts existed, which is what the code does. Changing line items to canonical USD in the fallback would be a new, separate change to non-presentment receipts across every non-USD seller — out of scope here, and not obviously an improvement, since the seller's listed price is the number the buyer recognises from the product page.

The module comment was the real problem: it said the fallback renders "canonical USD throughout", which over-promised and is presumably what this finding is reading. Corrected in `f3e966ca9` so the comment describes the actual pairing (line items in the seller's display currency, total in USD) and says explicitly that the fallback changes nothing about it.

No code change on this one.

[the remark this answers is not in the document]
gumclaw: This is the same finding as the thread above, and the answer is the same: the pairing you're describing is what receipts have shipped for years, and this branch neither introduces nor widens it.

The evidence is on unmodified `main`. `spec/presenters/receipt_presenter/item_info_spec.rb` has a context "when the purchase is in EUR" that sets only `displayed_price_currency_type: Currency::EUR` — no presentment row exists anywhere in that example — and asserts `{ label: "Product price", value: "€14.99" }`. The payment section on that same receipt renders USD. So a EUR item line above a USD total is pinned behaviour on the base branch, and your repro reproduces it rather than a regression.

The reason is that two different currencies live on a receipt and they answer different questions. `displayed_price_currency_type` is the currency the *seller* listed the product in, and "Product price" is a per-unit listing figure before quantity, shipping, tax and FX — it is not a line that sums into "Amount paid" and never has been. The tip line follows the same rule for the same reason. The figures that do have to reconcile against the buyer's card statement are the per-purchase charged prices, shipping, tax and total in `PaymentInfo`/`ChargeInfo`, and after this branch those are all read off one side of a single document-level decision, which is the bug being fixed.

On `main` the fallback expression is character-for-character what this branch calls:

```ruby
value: purchase.buyer_presentment? ? purchase.formatted_buyer_presentment_price_per_unit : purchase.formatted_total_display_price_per_unit
```

Only the condition in front of it changed, from "this purchase has presentment amounts" to "the document settled on one buyer currency and this purchase has them". For a presentment purchase in fallback that moves the line *toward* `main`: before, it printed the buyer's currency under a USD total; now it prints the listed currency exactly as a canonical-USD receipt always has.

What was genuinely worth fixing from the first version of this finding was the module comment, which said the fallback renders "canonical USD throughout". That overstated it, and it's corrected in f3e966ca9 — the comment now says what the fallback actually is (line items in each product's display currency, totals in USD) and that it predates buyer-currency pricing.

Restating the listing line in the buyer's currency is possible but it is a product decision, not a correctness fix: it would change every EUR/GBP/CAD-listed receipt Gumroad sends today, drop information the seller chose to show, and print a converted figure the buyer cannot reconcile against their statement on a document tax authorities read. I'm not folding that into a PR about surviving refunds — happy to open it as its own issue if that's the direction anyone wants.

app/presenters/receipt_presenter/item_info.rb
gumclaw: Fixed in ab38ccf02 — the decision is now made once per document and handed down.

`ReceiptPresenter#presentment_currency` answers it a single time and passes the answer into `ChargeInfo`, `PaymentInfo` and every `ItemInfo`. `InvoicePresenter::OrderInfo` reuses the receipt's answer rather than deriving its own, since an invoice states the same amounts. `ChargeInfo` also memoizes the buyer-currency total, which `formatted_total_transaction_amount` reads alongside the currency.

Sections keep a `PRESENTMENT_CURRENCY_UNDECIDED` default so the two callers that render a section with no receipt around it still work — the upcoming-call reminder mail builds a lone `ItemInfo`, and the section specs construct sections directly. The sentinel isn't `nil` because `nil` is a real answer here meaning "fall back to canonical USD"; memoizing on the sentinel rather than `||=` stops a USD receipt from recomputing on every read.

Pinned with a spec that renders charge info, payment info and all item lines off one presenter and asserts each purchase is asked `buyer_presentment_display?` at most once:

```
ReceiptPresenter For Charge #presentment_currency
  → asks each purchase once, no matter how many sections and item lines are rendered
```

`spec/presenters/receipt_presenter_spec.rb spec/presenters/receipt_presenter/item_info_spec.rb spec/presenters/receipt_presenter/payment_info_spec.rb spec/presenters/receipt_presenter/charge_info_spec.rb spec/presenters/invoice_presenter/order_info_spec.rb` — 243 examples, 1 failure, and that one (`receipt_presenter_spec.rb:112`) fails identically on the branch without these changes. Rubocop clean.

Your other two comments landed in 57fee48c9 and 4deab7860; I kept those versions and built on top rather than duplicating them.

gumclaw:
@rmarescu all four review findings are fixed and CI is green at the current head — back to you.

Your approval landed at `483c3e3ca`, which is before every one of the fix commits, so GitHub no longer counts it against the head and I'm not merging on a stale approval. The findings, in the order you left them:

| Finding | Fix |
| --- | --- |
| `app/modules/` is legacy, move to `app/presenters/concerns/` | `57fee48c9` |
| Duplicates the #6286 refund helpers | `4deab7860` |
| `formatted_total_transaction_amount` walks refunds twice — memoize | `3c2841e04` |
| Every item recomputes the receipt-wide decision — n × n refund queries | `ab38ccf02` |

The n × n one is the only structural change. `ReceiptPresenter#presentment_currency` now answers the question a single time per document and hands the answer into `ChargeInfo`, `PaymentInfo` and every `ItemInfo`; `InvoicePresenter::OrderInfo` reuses the receipt's answer instead of deriving its own, so an invoice's tax and total can no longer disagree with the receipt for the same charge. Sections built without a document-wide decision — the upcoming-call reminder mail renders a lone item section, and specs exercise sections in isolation — get a `PRESENTMENT_CURRENCY_UNDECIDED` sentinel and decide for themselves. The sentinel is deliberately not `nil` because `nil` is a real answer here meaning "fall back to canonical USD", which is also why the memo is guarded on `defined?` rather than `||=`.

`spec/presenters/receipt_presenter_spec.rb` has a guard that counts the calls: it puts three purchases on one charge, wraps `buyer_presentment_display?` on each, renders charge info + payment info + every item line, and asserts no purchase was asked more than once. I checked it actually bites by reverting the `ItemInfo.new(purchase_item, presentment_currency:)` argument and confirming it fails with "received: 2 times" rather than passing vacuously.

Also worth flagging honestly: a scheduling comment of mine earlier in this PR said full CI was green "so nothing agent-side remains and this is yours to review and merge" at `483c3e3ca`. That was true then, but five commits have landed since, so please treat the review as being against `ab38ccf02` and not the tree you approved.

CI at `ab38ccf02`: 77 checks passing, no failures — Test Fast and Test Slow shards, Test Minitest, Lint Ruby, Lint JS/TS, Build images, Buildkite and Greptile all green. Locally the touched presenter specs pass (225 examples across `item_info`, `charge_info`, `payment_info` and `order_info`). The four threads are left unresolved for you to close, since resolving your own findings on your behalf isn't mine to do.

rmarescu (approved):

app/presenters/concerns/buyer_presentment_display.rb
gumclaw: Done — the concern lives at `app/presenters/concerns/buyer_presentment_display.rb` (no `app/modules/` path in this PR's diff).

app/models/purchase.rb
gumclaw: Done — collapsed onto the #6286 helpers; the predicate is now `buyer_presentment? && !buyer_presentment_refunded_cents_incomplete?` rather than a re-implementation.

app/presenters/receipt_presenter/charge_info.rb
gumclaw: Done — memoized: `@_presentment_total_cents ||= chargeable.successful_purchases.sum { _1.buyer_presentment_total_cents.to_i }`, so `formatted_total_transaction_amount`'s two calls walk the refunds once.

PR #6290, opened 2026-07-25
Add Alipay to checkout behind a per-seller launch flag

gumclaw:
## What

Adds Alipay to checkout as a launch-flagged payment method on the existing client-confirmed Payment Element path. It rides the canonical-USD element and deferred intent unchanged — no new checkout fields, no forced presentment currency, no new charge lane.

The gate lives in `Checkout::PaymentMethodResolver` next to Klarna's and is controlled by a new per-seller Flipper flag, `checkout_local_method_alipay`, default off. At 0% nothing about any existing checkout changes.

## Why

Alipay is provisioned on the platform Stripe account: `alipay` reports `available: true` with `display_preference: on` on the default payment method configuration, and a live-mode USD PaymentIntent listing `alipay` confirms to `requires_action` with `next_action.type = alipay_handle_redirect`. (That probe intent was cancelled immediately; no money moved.)

It reaches buyers the current method menu serves poorly — Chinese consumers, overseas Chinese, and Chinese travellers, none of whom are well served by card on a US-domiciled account. Alipay also has no dispute process, so this volume carries no chargeback exposure.

Sizing is on the tracking issue and is deliberately modest: roughly $1–4K of incremental GMV per 30 days at the start. The case for building it is lift-per-unit-of-work, not the absolute number — it is the same redirect handshake Klarna and Cash App Pay already use, so the build is small and the ramp lever is identical to `checkout_local_method_klarna`.

Refs gumroad-private#1339.

## Approach

Alipay is modelled on Klarna, with two of Klarna's gates deliberately **absent** and the third one kept. Each decision is commented at the call site:

- **No buyer-country lock.** Alipay is not region-locked the way Cash App Pay is, and the target cohort largely does not sit in mainland China — mainland buyer IPs account for 12 sales in the last 30 days, while Chinese cards are overwhelmingly used from Hong Kong, Taiwan and Singapore. A GeoIP lock would gate out most of the addressable population.
- **No amount window.** Stripe publishes no per-country Alipay transaction limits of the kind that forced Klarna's fail-closed USD window.
- **The account-country gate is kept.** Stripe ties each Alipay presentment currency to the account's business country — `usd` is United States only, and only `cny` is valid for any country. Since this lane creates USD intents on the connected account, a non-US connected account must never carry an `alipay` entry, so Alipay gets the same US-account gate Klarna has (`us_based_merchant_account?`, shared by both). See the review note at the bottom: the first version of this PR asserted the opposite, and that was wrong.

What it *does* inherit:

- **Mutual exclusion with the forced-currency element mount.** When iDEAL/Bancontact/UPI survives, the element mounts in EUR/INR and the intent is created in that currency. An Alipay entry on that intent would make Stripe reject the whole create, taking card down with it — the gumroad-private#1026 failure mode. Enforced in the resolver and again as belt-and-braces in `Order::PreparePaymentIntentService#intent_payment_method_types`.
- **Fail-closed capability gating for direct-charge sellers** via `"alipay" => "alipay_payments"` in `StripeConnectPaymentMethodAvailabilityService::CAPABILITY_BY_PAYMENT_METHOD_TYPE`. A connected account that has not activated Alipay never sees it offered. This is orthogonal to the country gate above and does not substitute for it: Standard (dashboard) connected accounts can enable Alipay themselves, so an active capability on a non-US account is an ordinary state.
- **The previewed-method append gate.** A stale or crafted `alipay` ConfirmationToken cannot re-enable the method after a rollback — the append re-checks the launch flag, the US merchant-account gate, and the per-account capability, so it fails closed at confirm instead.
- **Recurring exclusion.** Memberships and preorders are out of scope for the first launch, so `alipay` joins `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES`.
- **PPP exclusion.** The wallet exposes no Stripe-owned funding country, so a PPP checkout would fail closed at prepare anyway. Dropped by omission from both PPP allowlists.

Two Stripe constraints from the issue, resolved:

- **No manual capture.** The codebase has no `capture_method` usage at all — every intent is immediate-capture — so the constraint is satisfied by construction with no code change. Verified by grep, noted here so a reviewer does not go looking for a guard.
- **Asynchronous refunds** (up to five minutes, 90-day cap). The existing `refund.updated` / `refund.failed` handlers already treat Stripe's terminal status as authoritative rather than assuming success at request time, including the row-lock that stops a late `refund.updated` resurrecting a failed status. No new plumbing; I extended the comments to name the asynchronous-wallet case and covered it in the existing spec suite.

Finally, the surfaces that would otherwise record an Alipay charge as a card: `CardType::ALIPAY`, the Stripe card-type map, `alipay_handle_redirect` in the client-handled next-action types (so an abandoned redirect on the return page does not page anyone), the sales CSV label, and the receipt's credit-card statement note suppressed.

## Before / after

No user-visible change in this PR. The flag is off for every seller, so the rendered method menu is byte-for-byte what it is today — that is the point of landing the gate dark.

There is also **no adjacent UI surface to screenshot**: the diff touches no `.tsx`, `.jsx`, `.erb` or view file (verified — `git diff $(git merge-base HEAD origin/main) HEAD --name-only` returns Ruby and spec files only). The buyer-visible change arrives when the flag is turned on for a QA seller, which is the preview-QA checklist item still open on the tracking issue; those recordings will be attached there before any ramp.

## QA steps

Nothing to click on this PR at 0%. To exercise the method once it is on a preview app:

1. Enable the flag for the QA seller: `Feature.activate_user(:checkout_local_method_alipay, seller)`.
2. Open a checkout for a single one-time, USD-priced product from that seller.
3. Confirm Alipay appears in the Payment Element accordion alongside card and Link.
4. Select it and confirm — Stripe.js redirects to Alipay's sandbox, then returns to the checkout return page and the purchase completes.
5. Abandon the redirect instead (close the Alipay tab, revisit the return URL) and confirm no error is reported to Sentry — this is what the `alipay_handle_redirect` next-action entry is for.
6. Refund the purchase and confirm the refund reaches its terminal status via the webhook rather than being assumed successful at request time.
7. Negative check: with the flag **off**, confirm Alipay is absent and the rest of the menu is unchanged.

## Test results

See the terminal walkthrough below. Summary of what was run locally:

| Suite | Result |
|---|---|
| `payment_method_resolver_spec.rb`, Alipay examples only | 14 examples, 0 failures |
| `payment_method_resolver_spec.rb`, whole file on this branch | 89 examples, 0 failures |
| `payment_method_resolver_spec.rb`, whole file on unmodified `origin/main` | 75 examples, 0 failures |
| `prepare_payment_intent_service_spec.rb`, Alipay examples only | 4 examples, 0 failures |
| `prepare_payment_intent_service_spec.rb`, whole file | 72 examples, 0 failures |
| intent status + capability map + card type + receipt | 178 examples, 0 failures |
| sales CSV payment-type label | 1 example, 0 failures |
| `rubocop` on all changed files | 17 files, no offenses |

The branch/main comparison is the one worth reading: +14 examples and no pre-existing example disturbed.

Alipay spec run

The script that produced that output is committed next to it as `qa-media/pr-6290-alipay-spec-run.sh`, so the run is reproducible rather than a screenshot to take on trust.

One caveat the screenshot shows rather than hides: the nine forced-currency examples in the resolver spec depend on a USD-holding platform merchant account row that a `db:test:prepare` clears, and on such a database they fail on unmodified `main` too. Section 2 of the output therefore lists the failing line numbers on both sides — main's nine (162, 216, 220, 246, 260, 268, 276, 284, 300) are the same nine on the branch shifted by my inserted block (162, 322, 326, 352, 366, 374, 382, 390, 406), and the single extra, line 263, is my own Alipay example: the direct twin of main's line 162 Klarna example. On a properly seeded database the whole file is 89/89 on the branch and 75/75 on main, which is what the table reports and what CI will show.

One note on getting there honestly. An early evidence run showed two resolver failures and I chased them instead of shipping around them. They turned out to be database state, not the change — and while investigating I found the sequential branch-then-baseline comparison in my own evidence script was itself unsound (both worktrees default to the same `gumroad_test`, so the second run's truncation raced the first and produced phantom failures that would not reproduce in isolation). The script now reports the two failure sets side by side instead of a bare "0 failures", because that comparison is the claim actually worth making. I also found and fixed a real bug in my own spec along the way: `payment_method_types` is nil on a recurring cart, so one assertion was calling `include?` on nil and proving nothing.

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5` (read from `model.default` in the agent config, not inferred).

The instruction was a one-word "Build" on gumroad-private#1339, against the scope already written on that issue by the same agent — flag, resolver gate, capability mapping, the `alipay_handle_redirect` next-action entry, asynchronous-refund verification, and the constraint that Alipay must not join the forced-currency registry.

No further prompting shaped the implementation. Decisions the agent made on its own, which a reviewer should look at with fresh eyes:

- Omitting the buyer-country lock and the amount window, and the reasoning recorded in comments for each. **The third omission — the account-country gate — was wrong, and the fable-5 pre-merge review caught it.** The first version of this PR claimed Alipay has no rule tying it to a US-based account; Stripe's docs tie each Alipay presentment currency to the business country, with `usd` United States only. A UK/DE seller with the capability active plus the launch flag would have had `alipay` listed on every USD intent create, and Stripe's rejection would have failed that seller's entire checkout, plain-card buyers included. Fixed in 1295f560b: the Klarna-shaped US-account gate at both sites, with the two specs that had pinned the buggy behavior flipped to assert the drop.
- Treating the no-manual-capture and asynchronous-refund constraints as already satisfied — grep and spec coverage rather than new code.
- Adding `alipay` to `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES` on the same first-launch reasoning as Klarna, which the issue did not explicitly ask for.


gumclaw:
## Fable-5 pre-merge adversarial review — REQUEST-CHANGES, fixed in 1295f560b

Ran the mandatory local fable-5 review against `4a487b008`. It found a genuine merge-blocker that CI (all green) and Greptile (no findings) both passed: **the PR's central design claim was factually wrong.**

### P1 — "Alipay has no account-country rule" is false; a non-US connected account would have bricked its ENTIRE checkout

Stripe's [Alipay docs](https://docs.stripe.com/payments/alipay) tie each Alipay presentment currency to the business's country: `usd` → United States only; only `cny` is valid for any country. This lane creates **USD** intents on the *seller's connected account* for direct-charge sellers, and Gumroad's connect sellers hold Standard (dashboard) accounts — where the seller can enable Alipay themselves from any of Stripe's ~40 supported business countries.

So a UK or German seller with an active `alipay_payments` capability plus the launch flag would have had `alipay` listed on every USD intent create. Stripe rejects that outright, and an incompatible `payment_method_types` entry fails the **entire** intent create — **every buyer of that seller fails at prepare, plain-card buyers included.** That is exactly the gumroad-private#1026 failure mode, the highest-severity outcome on the hunt list.

The capability snapshot could not catch it, for precisely the reason the code already documents for Klarna at `payment_method_resolver.rb:296-297`: the capability genuinely *is* active; what is invalid is the (method, account country, intent currency) combination.

Worse, two of my own new specs **pinned the buggy behavior** ("offers Alipay on a non-US connected account…").

**Fix (1295f560b):** the Klarna-shaped gate at both sites that can put alipay on an intent —

| Site | Change |
|---|---|
| `Checkout::PaymentMethodResolver#alipay_methods` | added `alipay_supported_merchant_account?` gate |
| `Order::PreparePaymentIntentService#appendable_previewed_payment_method_type` | added the same class-level re-check, so account drift after the Element mounts cannot re-append alipay |
| `payment_method_resolver.rb` | `klarna_supported_merchant_account?` and `alipay_supported_merchant_account?` both delegate to a shared `us_based_merchant_account?`; unknown country fails closed |
| both specs that pinned the bug | flipped to assert the **drop**, plus a new unknown-country fail-closed case |

### P2 — the "private preview" rationale was the wrong safety argument

The capability-map comment asserted most connected accounts "will simply never see the method" because platform-requested `alipay_payments` is a private preview. That premise is what let the P1 through: Standard dashboard accounts enable Alipay themselves, so "capability active on a non-US account" is an ordinary state. Corrected in `stripe_connect_payment_method_availability_service.rb`, the resolver, and the prepare service so the next method launch does not inherit the false premise. The capability check is documented as *orthogonal* to the country gate, not a substitute for it.

### P3s — noted, not fixed here

| # | Finding | Disposition |
|---|---|---|
| P3-1 | `alipay_handle_redirect` suppression in `stripe_intent_status.rb` is action-type-only, method-agnostic | Verified alert-only (both callers only decide whether `ErrorNotifier.notify` fires; no state change) and it matches the existing `cashapp_handle_redirect_or_display_qr_code` precedent exactly. No change. |
| P3-2 | Exact-ordering `eq(%w[...])` assertions in resolver specs churn on every method launch | Matches the file's existing Klarna style. Nit. |
| P3-3 | A refund attempted past Alipay's 90-day window surfaces the generic "try again later" message for a rejection that can never succeed | Real but not a blocker — nothing is corrupted (the rejection happens before `refund_purchase!` writes). Alipay is the first launched method with a hard short refund deadline, so this is worth a follow-up in the shape of Klarna's `KLARNA_AMOUNT_INELIGIBLE_MESSAGE`. |
| P3-4 | The non-US append spec passed via the resolver rather than the append clause it named | Fixed alongside the P1 — it now stubs the resolver to a card-only set, so it exercises the append clause's own gate. |

### Verified clean, with evidence

- **Flag-off byte-identity** — full consumer sweep of every touched constant. `ONE_TIME_PAYMENT_METHOD_TYPES` / `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES` are read only inside `eligible_method_policy`; `eligible_payment_method_types` is consumed only by `launched_method_set`'s per-method-gated intersections and the `log_decision` log line. The only observable flag-off change is `alipay` appearing in that log line's `eligible=` / `launch_gated_out=`.
- **Marker exclusivity** — `git grep -il alipay` at the merge-base: zero hits in `app/`, `lib/`, `config/`. No legacy Alipay lane whose markers the new detectors could misclassify.
- **Klarna-site inventory** — all 12 files with a `klarna` hit enumerated; zero hits in `app/javascript`, `app/views`, ERB (the frontend is fully generic, driven by the server's `payment_method_types`). Alipay now has an analogue at every site except the buyer-country lock and the amount window, both of which hold up.
- **Presenter/prepare parity** — Alipay reads no input the two callers could derive differently, so it cannot introduce a new divergence axis.
- **Forced-currency mutual exclusion** — including mid-checkout flag-flip traces in both directions; the belt-and-braces strip runs *after* the append, so the append can never re-introduce alipay onto a EUR/INR intent.
- **Off-session safety** — every `setup_future_usage` producer and the only SetupIntent producer are hard-coded `payment_method_types: ["card"]`; an off-session alipay intent stuck in `requires_action` cannot be minted by this codebase.
- **Refund path** — the comments-only claim is accurate: async pending Alipay refunds ride the existing iDEAL reversal machinery (`Refund::TERMINAL_FAILURE_STATUSES`, `HandleFailedRefundService`).
- **Card-type/display sweep** — the two sites where absence produces *wrong* output (`StripeCardType::NEW_CARD_TYPES`, the receipt's statement note) both have Alipay entries.
- **Load-bearing spec check by revert** — deleting the Alipay append clause reddens "appends a launched seller's alipay token…"; worktree restored clean afterward.

### Only verifiable live

- The exact Stripe rejection surface for the P1 (error code, and whether the Element mount also fails) on a live non-US connected account. The gate is correct either way; this is about blast radius, not direction.
- That `checkout_local_method_alipay` is genuinely 0%-default in production.
- The gumroad-private#1339 sizing figures (private repo).

Evidence regenerated at the new head — 14 Alipay resolver examples green, the append gate's 4 examples green, whole prepare file 72/72, 178 surrounding-surface examples green, rubocop clean on all 17 changed Ruby files. The branch-vs-main failure comparison in the script now runs each side once and reads both numbers from the same run (the previous version ran rspec twice per side, which let the count and the failure list disagree).

gumclaw:
@rmarescu review request — this is the Alipay leg of gumroad-private#1339, and it's on the same resolver/prepare surface you approved for iDEAL (#5953), UPI (#6143) and the settlement-marker fix (#6180), which is why it's coming to you rather than round-robin.

**Merge posture:** payments-core, so I am not self-merging. Ready for review, not draft, CI green at `1295f560b` across the full 85-check matrix.

**Shape:** Lane B — canonical-USD redirect, the Klarna pattern. Alipay rides the existing USD element and intent; `next_action` is `alipay_handle_redirect` and Stripe.js owns the full-page redirect. No forced-currency registry entry, no new checkout fields, no pending-purchase state.

**The one thing worth your fresh eyes.** My first version asserted Alipay has no rule tying it to a US-based account, so I shipped it with the flag as the only gate. That was wrong, and the local fable-5 review caught it after both CI and Greptile passed it clean. Stripe scopes each Alipay presentment currency to the account's business country — `usd` is United States only. Our connect sellers hold Standard dashboard accounts and can enable Alipay themselves from any of ~40 countries, so a UK or German seller with the capability active plus the flag on would have had `alipay` on every USD intent create. Stripe rejects the whole create on an incompatible entry, so **that seller's every checkout fails, plain-card buyers included** — the gumroad-private#1026 failure mode.

The capability snapshot cannot catch this, for the same reason already documented for Klarna: the capability genuinely is active. What is invalid is the (method, account country, intent currency) combination. Two of my own new specs had pinned the buggy behaviour as intended; they now assert the drop, plus an unknown-country-fails-closed case.

Fixed in `1295f560b` at both sites that can put alipay on an intent — the resolver gate and the previewed-method append clause, so account drift after the element mounts cannot re-append it.

**Reviewing efficiently:** the US-account gate is the load-bearing change; the other three deliberate calls (no buyer-country lock, no amount window, `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES` on first-launch reasoning) each carry their why in a comment. The 4-file method-recording sweep — `card_type.rb`, `stripe_card_type.rb`, sales-CSV labels, the receipt's credit-card-note exclusion — is the boring-but-mandatory part; without it the charge records `generic_card` and the receipt promises a card statement line that never appears.

**Flag state:** `checkout_local_method_alipay`, per-seller Flipper, default off, enabled nowhere at merge. The diff is invisible in production until I ramp it. Test mode deliberately does *not* bypass the flag for Lane B.

**After merge the ball comes back to me**, not to you: the ramp is binary 100/0 per the standing rule, with a per-method digest row first and a real *completed* canary purchase after 100%. I'll keep gumroad-private#1339 open until Alipay is live and monitored in production, since I own the ramp-down lever.

rmarescu (approved):
Good to merge on green.

gumclaw:
Merged — squashed as `34377b6`.

CI was green at `1f2235a`, the exact commit you approved: the full 49-shard Test Slow matrix plus `ci/green` and `buildkite/gumroad` all passing, no unresolved review threads, no pushes after your review.

The flag `checkout_local_method_alipay` is off for every seller, so nothing changes in production on this deploy. Preview QA with the flag on for a QA seller, and the recordings, stay on gumroad-private#1339 before any ramp.

PR #6292, opened 2026-07-25
Serialize production deploys so concurrent merges stop killing each other

gumclaw:
## What

Adds `concurrency_group: "gumroad/production-deploy"` + `concurrency: 1` to the **Deploy to Production** step, so production deploys queue instead of racing each other.

## Why — 37% of our deploys are failing, and almost all of it is a lost race

A deploy holds a Consul lock for its whole run, and `deploy_production.sh` exits 1 immediately if the lock is already held. Nothing stopped two deploys overlapping, so a commit merged inside another deploy's window started, hit the lock, and died ~30 seconds in.

Measured on the **last 30 `main` builds** (2026-07-25): **11 of 30 deploy jobs failed**. **9 of those 11 are provably collisions** — each failed deploy started *inside* a concurrently-successful deploy's hold window:

| Failed | Died after | Was inside |
|---|---|---|
| 15642 | 32s | 15635 (01:52:29–02:19:19) |
| 15663 | 32s | 15662 (03:36:19–04:01:01) |
| 15671 | 32s | 15668 (04:01:25–04:37:13) |
| 15682 | 25s | 15673 (04:43:40–05:06:22) |
| 15704 | 26s | 15692 (05:39:58–06:02:14) |
| 15721 | 26s | 15708 (06:08:36–06:35:57) |
| 15725 | 31s | 15723 (06:38:00–07:01:00) |
| 15741 | 25s | 15740 (09:48:43–10:11:41) |
| 15760 | 27s | 15753 (10:41:26–11:06:15) |

**Median successful hold is 23 minutes**, so on a day with steady merges a large share of commits land inside someone else's window. The ~30-second failure duration is the tell: these never got as far as doing any work.

## The consequence is worse than a red build

**Failed deploys are not retried, so `main` silently stops shipping.** Today production sat on `1406b478` for hours with three commits queued behind it, and it only moved because a human noticed and pushed again. Two separate times I diagnosed this as a *stuck* lock before the data showed it was a *contended* one — the log line is only `🔒 Deployment is locked`, which reads like a broken deploy rather than a scheduling conflict, so debugging time goes to the wrong place.

## Why a concurrency group rather than fixing the lock

The Consul lock can only **refuse** — it cannot **wait**. `concurrency_group` makes Buildkite hold the second deploy until the first finishes and then run it, which is what the lock was trying to express. The queued commit ships on its own instead of needing a human to re-trigger it.

Left on the default `concurrency_method` (`"ordered"`) deliberately: deploys ship a specific revision, so running them out of order risks putting an older revision live after a newer one. `eager` only removes the ordering guarantee — it does not mean newest-first.

This **does not remove the Consul lock**, which still protects deploys started outside Buildkite (the manual `lock_deployment.sh` path). Making that lock debuggable is the companion change in antiwork/gumroad-deployment#43, which records who took each lock and why, and auto-releases it when a deploy dies mid-flight.

Together: #43 makes a held lock explain itself, this makes deploys stop fighting over it.

## Test plan

- [x] `.buildkite/pipeline.yml` parses as valid YAML, and the deploy step resolves to `{'concurrency_group': 'gumroad/production-deploy', 'concurrency': 1}`.
- [x] Field names and semantics verified against Buildkite's command-step documentation (`concurrency_group` requires `concurrency`; `concurrency_method` defaults to `ordered`).
- [ ] Observational: over the next day of merges, deploy-job failures with a ~30s duration and a `Deployment is locked` log line should go to zero, and no commit should need a manual re-trigger to ship.

There is no way to unit-test Buildkite's scheduler; the verification is that the collision signature disappears from the build history.

## Before / After

Not applicable — CI configuration change with no user-facing surface, so the diff is the reviewable artifact.

---

Written by Claude Opus 5 (Anthropic), operating as the Gumclaw Hermes agent. Instruction given: "ensure we deploy soon and debug why we haven't and if we can improve our process."

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -56,6 +56,24 @@ steps:
     branches: "main"
     command: ".buildkite/scripts/deploy_production.sh"
     timeout_in_minutes: 120
+    # Serialize production deploys. A deploy holds a Consul lock for its whole run
+    # (median 23 minutes), and deploy_production.sh exits 1 immediately if the lock is
+    # already held. Without this, two commits merged inside one deploy window race: the
+    # second starts, finds the lock, and dies ~30 seconds in. Failed deploys are not
+    # retried, so main silently stops shipping until someone notices and pushes again.
+    #
+    # On 2026-07-25 that was 9 of 11 deploy failures in 30 builds — a 37% failure rate,
+    # every one a lost race rather than a real problem with the commit.
+    #
+    # concurrency_group makes Buildkite queue these steps instead: the second deploy
+    # waits for the first to finish and then runs, which is what the Consul lock was
+    # trying to express but could only enforce by failing.
+    concurrency_group: "gumroad/production-deploy"
+    concurrency: 1
+    # Left on the default concurrency_method ("ordered"): queued deploys run in the order
+    # they were queued. That matters here — deploys ship a specific revision, so running
+    # them out of order could put an older revision live after a newer one. "eager" only
+    # removes the ordering guarantee; it does not mean newest-first.
     plugins:
       - docker-login#v3.0.0:
           username: "gumroadteam"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -62,12 +62,19 @@ steps:
     # second starts, finds the lock, and dies ~30 seconds in. Failed deploys are not
     # retried, so main silently stops shipping until someone notices and pushes again.
     #
-    # On 2026-07-25 that was 9 of 11 deploy failures in 30 builds — a 37% failure rate,
-    # every one a lost race rather than a real problem with the commit.
+    # On 2026-07-25, 11 of 30 main deploy jobs failed — a 37% failure rate — and 9 of the
+    # 11 were provably lost races (each started inside a concurrently-successful deploy's
+    # hold window and died in 25-32s). The other two failed for unrelated reasons.
     #
     # concurrency_group makes Buildkite queue these steps instead: the second deploy
     # waits for the first to finish and then runs, which is what the Consul lock was
     # trying to express but could only enforce by failing.
+    #
+    # Scope: this serializes deploys started from THIS step only. A lock held by anything
+    # else — a manual nomad/production/lock_deployment.sh run, or one leaked by a deploy
+    # that died mid-flight — still fails a queued deploy exactly as it does today. That
+    # residual case is what antiwork/gumroad-deployment#43 addresses, by recording who
+    # holds the lock and why, and auto-releasing it when a deploy dies.
     concurrency_group: "gumroad/production-deploy"
     concurrency: 1
     # Left on the default concurrency_method ("ordered"): queued deploys run in the order


gumclaw:
## Deploy triggered, and the diagnosis that produced this PR

Production was on `47f7e094` with `f5cc4768` (two help-center article templates) queued behind it. Build **15760**'s deploy started at **11:04:25**, inside build **15753**'s hold window (**10:41:26–11:06:15**), and exited 1 on `Deployment is locked` after 27 seconds. Nothing was wrong with the commit.

Pushed empty commit `81b9bfb65` to re-trigger — the Buildkite token is read-only, so retrying the job directly returns `403 write_builds scope`. Deploy is running now in a clear window (lock read `false`, no live deploy job).

## Why this kept happening

I misdiagnosed this twice today as a *stuck* lock before checking overlap. It was contention every time. The tell is duration: a collided deploy dies in **~25–32 seconds** because it never does any work — it just hits the gate and exits. A real deploy holds for a **median of 23 minutes**.

Across the last 30 `main` builds: **11 deploy failures, 9 provably collisions** (each failed deploy's start time falls inside a concurrently-successful deploy's hold window). That's the 37% failure rate in the PR description.

## Why the impact is larger than a red build

**Failed deploys are not retried.** So each collision silently stops `main` from shipping until a human notices and pushes again. That is the actual mechanism behind production sitting hours behind today — not a stuck lock, not a bad commit, just a race with no recovery path.

It also misdirects debugging: `🔒 Deployment is locked` reads like a broken deploy, so the instinct is to inspect Consul and hunt for whoever forgot to unlock. The lock value is `true` either way, which is why it fooled me twice.

## Two changes, deliberately separate

- **This PR** stops the races: `concurrency_group` makes Buildkite *queue* the second deploy rather than fail it. A Consul lock can only refuse; it cannot wait.
- **antiwork/gumroad-deployment#43** makes a held lock self-explanatory (records who/why/when, auto-releases via an `EXIT` trap when a deploy dies mid-flight, breaks locks older than 90 minutes).

Neither replaces the other. The lock still guards deploys started outside Buildkite via `lock_deployment.sh`; this just stops Buildkite from fighting itself over it.

Diagnostic method recorded in the `deployment-status` skill so the next lock failure gets classified as contended-vs-stuck from overlap data instead of guessed from the lock value.

gumclaw:
@rmarescu review request — full CI green at head (all Test Fast/Test Slow/Minitest/Lint/Build SUCCESS, Buildkite green, Greptile SUCCESS). Nothing agent-side remains.

Flagging this to you because it is currently gating the checkout-migration ramps rather than being unrelated CI hygiene: 11 of the last 30 `main` deploy jobs failed, 9 of them provably because a deploy started inside another deploy's Consul-lock window and died ~30s in. Failed deploys are never retried, so `main` silently stops shipping and every merged local-method fix waits on a human noticing and pushing an empty commit. That is the same class of stall that held the Klarna merge overnight and needed a manual trigger commit at 01:03Z.

The change adds `concurrency_group` + `concurrency: 1` to the Deploy to Production step so deploys queue instead of racing. The Consul lock stays for deploys started outside Buildkite. Evidence table is in the PR body.

gumclaw:
Rerouting this from @rmarescu to @ershad — apologies for the bounce, @rmarescu.

Reason: this PR's entire diff is one file, `.buildkite/pipeline.yml`, and @ershad is the only person who has ever committed to it (every change back to the initial commit, including the two most recent ones — making preview apps opt-in via the `preview` label in June, and gating preview deploys behind Buildkite approval before that). Adding a `concurrency`/`concurrency_group` to the production deploy step is squarely in that ownership area, and Razvan's review history here is on the Stripe charge/presentment surfaces rather than the CI pipeline.

@ershad, what's needed from you is a judgement call, not a code read — the change itself is four lines:

1. Does serializing the production deploy step behind a shared `gumroad/production-deploy` concurrency group with a limit of 1 match how you want Buildkite to behave? The alternative is leaving the Consul lock as the only guard, which is what produced the failure this came from: build 15760's deploy started at 11:04:25 inside build 15753's hold window (10:41:26 onward), the two competed for the same lock, and a merge got dropped rather than queued.
2. Are you comfortable with the queueing side effect? Concurrent merges will now wait for each other instead of racing, so a busy merge window makes later deploys land later. That's the intended trade, but it changes deploy latency and you own that expectation.

No auto-merge has been or will be enabled — this is yours to merge when you're happy with it. Full CI is green at head.

One piece of context on urgency, not pressure: this is currently the thing making the checkout-method ramps unreliable, because a dropped deploy silently leaves a flag change un-shipped. If you'd rather someone else take it, say so and I'll reroute rather than let it sit.

gumclaw:
## fable-5 adversarial pre-merge review

Verdict returned **CHANGES-REQUIRED** with 1 P1 and 3 P3s. I fixed the two valid P3s in `676fceb09` (comment-only — the `concurrency_group`/`concurrency` settings are untouched) and **rejected the P1 as not applicable**, with evidence.

### The P1, and why it does not apply

The reviewer argued that with the default `ordered` method, a main build that never gets approved (red tests, deploy-freeze switch off, or a human deliberately skipping a bad commit) would sit in the concurrency group forever and silently dam every later deploy — turning today's loud 30-second failure into an unbounded silent stall. That would have been strictly worse than the bug this PR fixes, so it was worth taking seriously.

Buildkite's documentation refutes it directly:

> Be aware that both the `block` and `input` steps cause these steps to be uploaded and scheduled at the same time, which breaks concurrency groups. **These two steps prevent jobs being added to the concurrency group** … The concurrency group won't be added to the queue until the `block` or `input` step is allowed to continue.

Our deploy step `depends_on: require-approval`, a `block` step. A deploy waiting on approval is therefore **not in the group at all** — it cannot hold a slot or an ordering position. The dam the finding describes cannot form. Unapproved builds are exactly the case that is excluded.

### The P3 I could not check statically, now checked

The reviewer correctly flagged that "Skip Intermediate Builds" / "Cancel Intermediate Builds" would be dangerous here — either could silently drop a queued revision. Verified via the Buildkite API:

```
skip_queued_branch_builds: true    filter: !main
cancel_running_branch_builds: true filter: !main
```

Both are filtered to **exclude** `main`, so no queued main deploy can be skipped or cancelled. Safe.

### Fixed (676fceb09)

- **Overclaim corrected.** The comment said "9 of 11 deploy failures in 30 builds … every one a lost race." Actually 11 of 30 failed and 9 of those were provably races; the other two fell outside the sampled window. Both numbers now stated separately.
- **Scope stated honestly.** The comment implied queueing was a complete fix. It isn't — a lock held by anything outside this step (a manual `lock_deployment.sh` run, or one leaked by a deploy that died mid-flight) still fails a queued deploy exactly as today. Now says so, and points at antiwork/gumroad-deployment#43 for that residual case.

### Verified clean by the reviewer

- **`timeout_in_minutes` bounds running time only** — time spent `limited`/waiting does not count, so a deploy queued behind a 23-minute deploy keeps its full 120 minutes. This was my own top concern going in.
- **No group-name collision** — `gumroad/production-deploy` is the only `concurrency_group` in the repo; `preview.yml` has none. (Group names are org-wide, so this matters.)
- **`branches: "main"`** means non-main builds never create the job, so they cannot occupy the group.
- **YAML valid**, attributes correct at command-step level, upload will not be rejected.
- **A hung deploy is bounded** at the 120-minute step timeout, which frees the slot.

### One operational note, not a defect

With queueing, every intermediate stale revision now runs its full ~23-minute deploy in sequence, so under heavy merge traffic the newest commit's time-to-production grows with queue depth. That is the intended cost of correctness, but worth watching in the first week — if it bites, the answer is `concurrency_method: eager` plus accepting order inversion, not reverting.

Greptile also reviewed: 5/5, no actionable defect.

PR #1344, opened 2025-09-25
Fix video infinite loop after long pause

HiAbhishekh:
Part of : #790

AI Disclosure: Used for only for syntax, logic and code review Manually

### Problem

When users pause a video for approximately 10 minutes and then resume playback, the video gets stuck in an infinite buffering loop, requiring a full page refresh to fix. This interrupts the user experience and causes users to lose their place in the video.

### Solution

Added a simple error handler to the JWPlayer that detects expired HLS URLs and automatically recovers by reloading the video with a fresh URL while preserving the user's playback position.

player.on("error", (error) => {
  if (error.code === 403 || error.code === 410) {
    const currentPosition = player.getPosition();
    const currentItem = player.getPlaylistItem();
    player.load([currentItem]);
    player.once("ready", () => {
      player.seek(currentPosition);
    });
  }
});

How it works?

1. Detects when video URL expires (403/410 error)
2. Saves your current position in the video
3. Reloads the video with a fresh URL

Fixes #790

PR #1343, opened 2025-09-25
Migrate _card.scss to Tailwind

binary-koan:
Part of #1055

PR #1342, opened 2025-09-25
Store test.log for failed Capybara tests

binary-koan:
Adds a step to store `log/test.log` as an artifact for Capybara failures. We don't print logs to stdout in Capybara tests but it's potentially another useful way to debug flaky tests which pass locally ([example](https://github.com/antiwork/gumroad/pull/1308) where it's possible Elasticsearch indexing randomly fails but we can't check the logs to confirm)

PR #6295, opened 2026-07-25
Stop auto-applying the retired `planning` label to new issues

gumclaw:
## What

Deletes `.github/workflows/label-new-issues.yml`, the workflow that stamped the `planning` label onto every newly opened issue.

## Why

The `planning` label was retired and deleted org-wide on 2026-07-25 — issues are scoped at creation, so a "planning" phase label carries no information and nothing filters on it any more. (This was the label's second and final retirement.)

The label deletion was incomplete: this workflow calls `github.rest.issues.addLabels({ labels: ['planning'] })` on `issues: [opened]`, and the GitHub API **recreates a label that does not exist** when you add it. So every issue opened after the org-wide delete silently resurrected the label. In `antiwork/gumroad-private` alone, issues #1342, #1343, #1344 and #1345 (all opened 2026-07-25, after the retirement) were labelled `planning` by `github-actions[bot]`, and the label object exists again in both `gumroad` and `gumroad-private`.

Deleting the workflow is the fix: with no writer, deleting the label object stays deleted. The label objects are being removed separately after this merges.

## Risk

None. The workflow does exactly one thing — add a retired label. No other workflow in these repos calls `addLabels` (verified by reading every workflow file in `gumroad` and `gumroad-private`). No CI, automation or query depends on `planning`; the standing dispatcher instructions explicitly say never to apply or filter on it.

## No user impact

Repository automation only — no application code, no runtime paths, no user-visible surface is touched. The diff is a single deleted workflow file, so there is no adjacent UI flow to screenshot.

## Test Results

No test changes — the diff is a single deleted GitHub Actions workflow file. Verification is behavioral and was done against live state before opening this PR:

```
$ gh api repos/antiwork/gumroad-private/issues/1345/timeline \
    --jq '.[]|select(.event=="labeled" and .label.name=="planning")|"\(.actor.login)@\(.created_at)"'
github-actions[bot]@2026-07-25T12:18:37Z      # issue opened AFTER the org-wide delete

$ gh label list --repo antiwork/gumroad-private --jq '.[]|select(.name=="planning")'
planning                                       # resurrected by the addLabels call

$ for w in $(gh api repos/antiwork/gumroad/actions/workflows --jq '.workflows[].path'); do ... done
.github/workflows/label-new-issues.yml -> labels: ['planning']    # the only addLabels writer
```

Merge safety: `main` on both repos is governed by a ruleset (`gh api .../branches/main/protection` → 404 by design) with rules `deletion`, `non_fast_forward`, `pull_request` only — no `required_status_checks`, so removing a workflow cannot strand a pending required check.

## QA steps

1. Merge.
2. Delete the resurrected label object in both repos: `gh label delete planning --repo antiwork/gumroad --yes` (and `--repo antiwork/gumroad-private`). Deleting the object strips it from every carrier.
3. Open a throwaway issue and confirm it lands with no labels, and that `gh label list | grep planning` stays empty.

---

## AI disclosure

Written by **claude-opus-5** (Anthropic) running as gumclaw's autonomous dev dispatcher (a scheduled Hermes Agent job, no human in the loop for this PR).

Standing instruction the run was executing, verbatim: *"Open issues in both repos: >3 hours old, NO awaiting-human label, NO human assignee (gumclaw/none only) → plan and BUILD autonomously: bugs, docs, copy, tests, small clear features. (the `planning` label was RETIRED and deleted org-wide 2026-07-25 — issues are already scoped at creation; never apply or filter on it)"*

No human prompted this specific change. The model noticed, while filtering the issue queue per that instruction, that issues opened after the retirement still carried the label, traced the writer to this workflow, and opened the PR.


gumclaw:
## Self-review

Read the full diff and the surrounding automation before opening. Three things I checked specifically, because a workflow deletion is the kind of change that looks trivial and isn't:

**1. Is anything downstream of the label?** No. I enumerated every workflow in both repos via `gh api repos/<repo>/actions/workflows` and read each file's body looking for `addLabels`, `--add-label`, or a `labels: [...]` block. `label-new-issues.yml` is the only writer, and nothing reads `planning` — the other active workflows (`close-non-english-issues`, `close-non-compliant-issues`, `close-non-compliant-prs`, `close-question-issues`, `close-support-issues`, `claude`, `tests`) don't reference it. The standing dispatcher instruction explicitly says never to filter on it, so no automation query depends on it either.

**2. Could deleting a workflow strand a required check?** No. `main` here is governed by a ruleset, not classic branch protection, and its rules are `deletion` / `non_fast_forward` / `pull_request` — there is no `required_status_checks` rule, so no CI context is merge-blocking and none can be left permanently pending. (This is the failure mode called out in the `removing-ci-workflows` playbook; worth stating explicitly rather than assuming.)

**3. Is deleting the workflow sufficient, or does the label need deleting too?** Both, in this order. `addLabels` on a nonexistent label **creates** it, so deleting the label object while the workflow lives just resets the clock until the next issue is opened — which is exactly what happened after the 2026-07-25 org-wide delete. Removing the writer first makes the label deletion stick. Label deletion is in the QA steps, not this diff, because it's a repo-settings mutation rather than a code change.

**One thing I deliberately did not do:** the same workflow file exists in six other Antiwork repos (`gumroad-mobile`, `gumroad-cli`, `helper`, `flexile`, `smallbets`, `iffy`) per a path probe, but the Contents API returns 404 for the file on all six when actually fetched — the probe was matching on a stale/redirected path, and none of those repos ran the labeler in their recent workflow history. Rather than open six speculative PRs, I scoped this to the two repos where I have direct evidence of the label being applied post-retirement (`gumroad`, `gumroad-private`). If the labeler turns out to be live elsewhere, that's a separate follow-up with its own evidence.

**Scope check:** diff is exactly one deleted file, no unrelated changes, no leftover references. Nothing about this touches application code, so the "no user impact" claim in the body has no adjacent UI flow to screenshot — stated rather than skipped.

PR #6297, opened 2026-07-25
Fix mobile dashboard nav swallowing the first tap on a sidebar link

gumclaw:
## What

Guards the global Inertia `before` listener in `CloseOnNavigate` so it ignores requests the user didn't initiate. Only a real navigation closes the mobile dashboard nav drawer now — hover prefetches and background/async requests leave it alone.

## Why

Sellers on phones report that dashboard sidebar links (Analytics, Sales, Emails) do nothing on the first tap and only work after tapping again.

Sidebar links opt into hover prefetch by default (`ClientNavLink`, `prefetch = "hover"`). Inertia installs an `onMouseEnter` handler with a 75ms delay that calls `router.prefetch()`, and `router.prefetch()` fires the very same `inertia:before` event a real visit does (`@inertiajs/core`: `if (events.onBefore(visit) === false || !fireBeforeEvent(visit)) return;`). `CloseOnNavigate` subscribed to `before` unconditionally and called the nav's `close()`.

On a touch device the tap synthesises `mouseenter` first, so the sequence was:

1. Seller opens the drawer and taps "Analytics".
2. `mouseenter` starts the 75ms hover timer.
3. Timer fires → `router.prefetch()` → `inertia:before` → `CloseOnNavigate` → `close()`.
4. The drawer collapses (below the `lg` breakpoint the link list is only rendered while open) while the finger is still down, so the link is gone before the tap becomes a click — nothing happens.
5. Tapping again works, because `close()` is now a no-op on an already-closed drawer and the click's `router.visit()` lands.

This only reproduces below `lg`, where the sidebar is a drawer rather than static, which is why it reads as a mobile-only problem. Nothing throws, so it never showed up in Sentry.

The guard also covers `async` visits, which have the same problem from a different trigger: `router.reload()` and lazily-loaded props fire `before` too, and the admin sales-reports page polls every three seconds while rendering the same drawer, so a poll could close it out from under the user at an arbitrary moment. `visit.prefetch || visit.async` is the same condition the unsaved-changes prompts in `pages/Pages/Edit.tsx` and `pages/Settings/Profile/Show.tsx` already use.

`#3973` had to add `visit.prefetch` guards to the `start` / `finish` / `before` listeners in `app/javascript/entrypoints/inertia.js` for exactly this class of bug — background visits leaking into logic meant for real navigations. `CloseOnNavigate` was the one `before` listener that sweep missed.

The same component is mounted by the admin nav (`app/javascript/components/Admin/Nav/index.tsx`), so this fixes both.

Tracking issue: antiwork/gumroad-private#1343

## Before / After

No visual change: the drawer looks and animates exactly the same, and closing on real navigation still happens. The change is purely *which* events count as a navigation, so the evidence here is the regression test rather than pixels — a landed tap and a swallowed tap are visually identical mid-gesture.

## Test plan

New unit test `app/javascript/components/CloseOnNavigate.test.tsx` dispatches `inertia:before` for a real visit, a prefetch visit, and an async visit, and asserts the drawer's `close()` runs only for the real one.

Verified the tests are load-bearing by restoring main's unguarded listener — both guarded cases fail, then all three pass with the guard in place:

vitest results

Manual check on a phone-width viewport: open the dashboard nav drawer, tap a sidebar link once, and the page navigates on that first tap.

Captured on the hosted preview (https://fix-mobile-nav-prefetch-swallows.apps.staging.gumroad.org) as `seller@gumroad.com`, one session throughout.
Only two things vary between shots: the viewport and whether the `inertia:before` event came
from a background prefetch or a real navigation. Everything else — seller, page, data — is held
constant, so the shots are directly comparable.

As the Before/After note says, a landed tap and a swallowed tap look identical mid-gesture, so
what these prove is the *invariant*: the drawer stays put for a prefetch and closes only for a
real visit.

**Mobile, 375px, touch emulation**

| Drawer open | Still open after a hover prefetch | Analytics after ONE tap |
|---|---|---|
| drawer open | survives prefetch | one tap |

The middle shot is the fix. The pointer is parked on "Analytics" for 900ms — far past Inertia's
75ms hover timer — so `router.prefetch()` has fired `inertia:before`. Pre-fix `CloseOnNavigate`
would have closed the drawer at that moment and the link list would be gone; here the list is
still rendered and "Analytics" is highlighted as hovered. The capture run asserts this rather
than leaving it to the eye: it fails loudly if the link's bounding box has collapsed.

The third shot is the same gesture continuing into a single tap — Analytics renders and the
drawer is closed by the real navigation, which is the behavior the component is there for.

**Desktop, 1280px — the control**

desktop sidebar unchanged

At/above `lg` the sidebar is static rather than a drawer. Hover prefetch leaves it alone and a
single click still navigates, with the full sidebar still rendered afterwards — the guard did
not disturb the desktop path (QA step 6).

## QA steps

Preview app: https://fix-mobile-nav-prefetch-swallows.apps.staging.gumroad.org (sign in as `seller@gumroad.com` / `password`).
Click path: Dashboard → hamburger (top-right, phone width) → tap "Analytics" once.

1. Open the preview app on a phone (or a desktop browser at a viewport under the `lg` breakpoint with touch emulation on).
2. Sign in and go to the dashboard.
3. Tap the hamburger to open the nav drawer.
4. Tap "Analytics" **once**. The app navigates to Analytics on that single tap and the drawer closes as part of the navigation.
5. Repeat with "Sales" and "Emails".
6. At desktop width, confirm sidebar navigation and hover prefetch still behave as before.

---

## AI disclosure

Written by claude-opus-5 running as the Hermes agent. Instructions given: replicate and fix the reported bug on antiwork/gumroad-private#1343 (mobile dashboard sidebar links needing multiple taps), then open a PR. The agent read the Inertia `router.prefetch` implementation in `node_modules/@inertiajs/core` to confirm prefetch fires `before`, added the guard, wrote the regression tests, and verified they fail against main's unguarded listener. A pre-merge adversarial review pass (claude-fable-5) surfaced the async-visit case, which was folded in — review and dispositions are in a PR comment.


gumclaw:
## Pre-merge adversarial review (fable-5)

**Verdict: READY TO MERGE** — with two P3s raised, both of which I folded into the branch rather than deferring.

### Findings and dispositions

**P3-1 — background reloads could also close the drawer. FIXED (465b55e).**
`router.reload()` and lazily-loaded props (`WhenVisible`, deferred props) fire `before` with `prefetch: false`, so they still closed the drawer. That path is live: the admin sales-reports page polls every 3s (`usePoll(3000, …)` in `Admin/SalesReports/JobHistory.tsx`) and the admin nav renders the same `CloseOnNavigate`. Same failure mode as the reported bug, different trigger. The guard is now `if (visit.prefetch || visit.async) return;`, which matches the pattern the unsaved-changes prompts in `pages/Pages/Edit.tsx` and `pages/Settings/Profile/Show.tsx` already use. A third test case covers it.

**P3-2 — every mobile tap still issues a wasted prefetch request. NOT fixed here, deliberately.**
On touch, `mouseenter` arms the 75ms hover timer and `router.prefetch()` races the real visit the click already started; the prefetched response is discarded (`Response.process` flushes on navigation). Gating the `ClientNavLink` default on `matchMedia("(hover: hover)")` would drop one duplicate GET per nav tap. That is a separate performance change with its own blast radius across every sidebar link, and it is not needed to fix the swallowed tap, so I left it out of a bug-fix PR rather than widening the diff. Worth its own issue if we care about the wasted request.

### Verified clean

- **Guard correctness against Inertia's actual internals.** `fireBeforeEvent` always dispatches `inertia:before` on `document` with `{ cancelable: true, detail: { visit } }`, so `event.detail.visit` cannot be undefined for library-fired events. `getPendingVisit` defaults `prefetch: false`; only `router.prefetch()` sets it true. `router.visit/get/post/put/patch/delete` all fire `before` with `prefetch: false` and `async: false` → the drawer still closes. When a click consumes a *cached* prefetch, the `before` event comes from the real `router.visit()`, not the cached params → still closes. `router.prefetch()` early-returns before firing `before` when the target equals the current URL. Back/forward never fires `before` at all (goes through `page.setQuietly`), so this changes nothing there.
- **No second swallow path.** Every other global `before`/`start` listener already guards prefetch (`entrypoints/inertia.js`, and the two unsaved-changes prompts), so a hover prefetch can't pop a confirm dialog either. `CloseOnNavigate` was the one that #3973's sweep missed.
- **The duplicate `app/javascript/packs/inertia.js` is dead code** — the layout loads `entrypoints/base.ts` → `entrypoints/inertia.js`, and vite-plugin-ruby only wires `entrypoints/`. Both copies happen to carry the same guards, so no drift affects this PR.
- **Drawer still closes in every expected case.** Client and admin navs both render `CloseOnNavigate`; all drawer links are Inertia `Link`s except admin's Sidekiq/Flipper links, which are plain `<a>` full-page loads that unmount the drawer anyway. Same-page taps still close it.
- **Handler return value** is `undefined` on both branches, so it can't accidentally trip Inertia's `response === false → preventDefault()` cancellation path.
- **Test honesty.** The dispatch shape matches `fireBeforeEvent` exactly; the `vi.mock` stubs only `useNav`, the component's sole import from that module. Both guarded cases fail against main's unguarded listener and pass with the guard — see the run in the PR body.
- **Lint, types, effect cleanup, comment quality.** eslint and prettier clean; `tsc --noEmit` reports nothing for either file; `router.on`'s unsubscriber is returned from the effect with `close` as the dep.

gumclaw:
## Scope note before review

CI is green at `accae9b` (all Fast/Slow/Minitest/Lint shards, Buildkite #15802, Greptile 5/5). Ready for your look, @gianfrancopiana.

One thing worth knowing before you read it: **this PR is deliberately only half of the reported bug.** After it was opened, the seller followed up saying it happens on desktop too, which rules out the drawer as the whole story — above the `lg` breakpoint the nav is always rendered, so `close()` setting `open = false` is a visual no-op there and cannot swallow a click.

There is a second, independent mechanism on the same hover-prefetch feature: a click that lands on an *in-flight* prefetch is adopted via `PrefetchedRequests#use`, which is `.then()`-only, so a cancelled or errored prefetch leaves the navigation dropped silently. We also run `createInertiaApp({ progress: false })`, so a click waiting on a slow prefetch gives the user no feedback at all — which reads as "the link did nothing" even when nothing failed.

I did not fold that into this PR, because fixing it means either patching/forking `@inertiajs/core`, dropping the `prefetch = "hover"` default from `ClientNavLink`, or re-enabling a progress indicator. All three touch the whole nav and are a maintainer's call rather than mine. Full read of the library code path is written up in antiwork/gumroad-private#1343.

So: this change is correct and self-contained on its own merits (one guard, matching the `visit.prefetch || visit.async` condition `pages/Pages/Edit.tsx` and `pages/Settings/Profile/Show.tsx` already use), it just doesn't close the ticket by itself.

slavingia:
Fix on desktop too. Transitions between pages should be instant. Skeletons handle the rest. Don't like the inertia loader.

gumclaw:
Understood — desktop too, prefetch stays, no loader. That rules out two of the three options I listed and picks the third, so here is the concrete plan.

**First, a correction to what I wrote above.** I said fixing the desktop mechanism would mean "patching/forking `@inertiajs/core`" and treated that as disqualifying. That was wrong: this repo already patches dependencies as a matter of course — `patches/epubjs+0.3.93.patch` and `patches/react-dom+18.3.1.patch`, with `patch-package` in `devDependencies` and `"postinstall": "patch-package && ts-patch install -s"`. So a pinned patch against `@inertiajs/core@2.3.13` is an established convention here, not a new burden. My objection doesn't survive contact with the repo.

**The desktop bug, precisely.** In `PrefetchedRequests#use` (`dist/index.esm.js:476`):

```js
use(prefetched, params) {
  ...
  return prefetched.response.then((response) => { ... return response.handle(); });
}
```

`.then()` with no `.catch()`. When a click adopts an **in-flight** prefetch (`router.visit` → `prefetchedRequests.get()` → `use()`, line 2448-2451), the navigation is now riding that prefetch's promise. If the prefetch rejects — `onPrefetchError` at line 375 rejects it, and `cancelInFlight` can too — the chain dies, `response.handle()` never runs, and **nothing navigates**. No error surfaces to the user, which is exactly the reported symptom: the link did nothing. This is independent of the drawer, so it fires at every viewport, which is why the seller sees it on desktop.

It also explains why the first tap specifically fails and the second works: the second click finds either a settled cache entry or no entry at all, and takes the normal request path.

**The fix.** Patch `use()` to fall back to a real visit when the adopted prefetch fails, rather than dropping it:

- `.catch()` on the adopted promise; on failure, if this is still the current navigation (`currentUseId` guard, already there for the stale-adoption case), evict the dead entry and issue the request through the normal `syncRequestStream` path.
- No loader involved, and no change to the happy path — a prefetch that resolves is still instant, which is the behaviour you want kept. This only adds a recovery edge where today there is a silent dead end.

**Scope and sequencing.** The guard already in this PR (`visit.prefetch || visit.async`) is correct and self-contained on its own merits, and the library patch is a different mechanism in a different layer — a vendored dependency patch touching every Inertia navigation in the app. I'd rather not bundle a nav-wide dependency patch into a green, reviewed, drawer-scoped bug fix; if the patch needs to be reverted, that shouldn't drag the guard out with it.

So: **this PR stays as-is for your review**, and the desktop fix goes up as its own PR against `@inertiajs/core@2.3.13` with the patch file, a regression test that adopts a rejecting in-flight prefetch and asserts a real request follows, and a preview-app pass at both viewports. **ETA: next sweep** (~3h), linked here when it opens.

If you'd rather see both in one PR, say so and I'll fold it in instead — that's a preference call, not a technical constraint.

Tracking the full library-path read in antiwork/gumroad-private#1343.

gumclaw:
Desktop fix is up as promised: **#6310**.

It patches `@inertiajs/core` where the actual mechanism lives — `PrefetchedRequests#use` had no rejection handler, so a click that adopted an in-flight prefetch which then failed navigated *nowhere*, silently, at any viewport. That's the same first-click-fails/second-click-works shape this PR addresses on mobile, arriving from a second and independent cause.

Per @slavingia's constraint here — desktop too, transitions instant, no Inertia loader — #6310 adds no progress indicator, keeps `prefetch="hover"` on, and leaves the happy path unchanged. The regression test fails against the unpatched library and passes with the patch, so it's pinned against a future dependency bump silently dropping it.

It also turned up an independent latent bug in the same file: `RequestStream#send` never removed a *failed* request from its queue, so each one permanently occupied a concurrency slot.

Kept it out of this PR deliberately, so a nav-wide dependency patch stays revertable on its own without dragging this drawer guard back out with it — but say the word and I'll fold it in. This PR still stands on its own for the mobile drawer guard.

PR #6298, opened 2026-07-25
Move the audience-member sync off the license verify request

gumclaw:
## What

Moves the buyer's audience-member refresh off the license-verification request and into a small Sidekiq job (`UpdatePurchaseAudienceMemberDetailsJob`).

`POST /v2/licenses/verify` increments the license's `uses` counter, and that increment synchronously rewrote the buyer's `audience_members` row. `audience_members.details` is a single JSON column holding every purchase for an (email, seller) pair, so the rewrite takes a write lock on that row for its duration. Two verifications for the same buyer therefore serialized on one row, and the losers waited until MySQL's `innodb_lock_wait_timeout` fired and then returned a 500.

The write now runs in a job with the same two-second delay as the search-index update already enqueued on the identical code path, so Sidekiq's retries absorb a lock timeout instead of the caller seeing a failed request. The delay also lets a burst of verifications for one buyer collapse into fewer writes. The two `Subscription` callers of the same write (cancel and resubscribe) are moved to the job as well — both run on request paths and have the same shape.

## Why

Tracking issue: antiwork/gumroad-private#1349. (The original public issue #6294 was closed and refiled there — Sentry findings live on gumroad-private.)

- This is a public API endpoint that licensed software calls to decide whether to run, so a 500 here means a paying customer's app fails activation and the seller's support queue takes the hit.
- Sentry GUMROAD-YX: `ActiveRecord::LockWaitTimeout` — 1,566 events across 195 users, first seen 2026-07-16. All 1,566 events are this one transaction, and the most recent 100 landed in a single one-minute window from one integration verifying the same license concurrently.
- Nothing in the verify response reads audience data. It exists so the "minimum license uses" email-audience filter sees fresh values, which tolerates a couple of seconds of lag.
- The sibling side effect on the same line — the Elasticsearch purchase re-index — was already deferred to Sidekiq. The audience write was the only remaining inline one, and it is the one that takes a row lock.

Swallowing the `LockWaitTimeout` in the controller was deliberately avoided: it would leave the audience filters stale with no signal.

### Deliberately left inline, and follow-ups

`Purchase::AudienceMember`'s own `after_save :update_audience_member_details` still writes the same row synchronously, on purpose:

- It is not what this Sentry issue is firing on — all 1,566 events are the license-verify transaction.
- It fires on one-off buyer actions (purchase state change, refund, unsubscribe, email change) rather than an endpoint one integration calls concurrently, so it does not self-contend the same way.
- Its email-change branch does an ordered remove-old-then-add-new pair, so deferring it needs its own ordering design rather than a lift of this job.

Two deeper fixes are also out of scope here, to keep this diff to the 500s:

- `details["purchases"]` is rewritten whole on every verification. The only field that changes after purchase creation is `license_uses`, so a narrower `JSON_SET` on the matching array element would hold the lock far more briefly than a full read-modify-write.
- More fundamentally, `license_uses` is denormalized into that document only to serve one filter. Reading it from `licenses` at filter time would take the license path out of the audience write path entirely.

## Before / After

No user-visible surface changes: this endpoint returns JSON to third-party software and has no adjacent UI, so there is no screen to capture. The walkthrough below is the equivalent evidence — it shows the in-request SQL (one atomic `UPDATE licenses`, zero `audience_members` statements), the job being queued instead, a burst of five verifications collapsing to one job, the job producing the same audience data the inline write did, and a late job declining to resurrect a buyer who no longer qualifies.

https://github.com/antiwork/gumroad/blob/fix/async-license-audience-member-sync/qa-media/pr-6298-walkthrough.mp4?raw=true

## Test plan

- `spec/models/license_spec.rb` — new group asserting a `uses` change enqueues `UpdatePurchaseAudienceMemberDetailsJob` and performs no in-request `add_to_audience_member_details`, that a non-`uses` change enqueues nothing, that a license with no purchase enqueues nothing, and that running the job produces the same `license_uses` value in the audience document as before.
- `spec/sidekiq/update_purchase_audience_member_details_job_spec.rb` — the job refreshes the details, skips a purchase that no longer qualifies for the audience, and raises `RecordNotFound` for a missing purchase so the retries pick it up once a lagging replica catches up.
- `test/models/subscription_test.rb` — `#cancel!` and `#resubscribe!` enqueue the job for the original purchase, `#resubscribe!` no longer writes the audience document in-request, and a deferred job that runs *after* the subscription was deactivated does not resurrect the removed audience member.
- `spec/controllers/api/v2/licenses_controller_spec.rb` — full file, to confirm the verify endpoint's behavior and `uses` accounting are unchanged.

Each new positive assertion was verified load-bearing: replacing `app/models/license.rb` with main's version reddens 2 of the license examples, and replacing `app/models/subscription.rb` with main's reddens both Subscription enqueue tests. The resurrect test pins behavior that holds on both sides — `add_to_audience_member_details` re-checks `should_be_audience_member?`, which is what makes the async boundary safe against a late add racing a remove.

## QA steps

The endpoint's response is unchanged, so QA is at the API level:

1. Verify a license against a licensed product: `curl -X POST https://<host>/v2/licenses/verify -d product_id=<id> -d license_key=<key>`.
2. Confirm the response still returns `success: true` and a `uses` count incremented by one on each call.
3. In Sidekiq, confirm an `UpdatePurchaseAudienceMemberDetailsJob` job is scheduled ~2 seconds out for the purchase.
4. After it runs, confirm the seller's email audience with a "minimum license uses" filter still matches on the new count.
5. Fire ~10 concurrent verifications for the same license and confirm all return 200 (previously the losers 500'd).

## Test results

Run in a clean worktree off `origin/main`:

test results

Two failures appeared in the wider audience-member spec sweep (`audience_member/searchable_spec.rb:321`, `purchase/searchable_spec.rb:339`). Both reproduce identically in a clean worktree on unmodified `origin/main`, so they are pre-existing local-environment flakes rather than regressions from this change.

Minitest, after the CI failure on `02c2e6b` (full triage in the comments):

- `test/models/purchase_test.rb` — 674 runs, 1,658 assertions, 0 failures. Includes `AudienceMember removes member when subscription is deactivated`, which this change legitimately broke: it asserted the audience row reappeared *during* `resubscribe!`, and now asserts that `resubscribe!` adds nothing in-request while draining the job restores the member. Strictly stronger than what it replaced.
- `test/models/subscription_test.rb` — 433 runs, 1,018 assertions, 0 failures.

The other failure in that run, `installment_test.rb:601`, is an unrelated pre-existing timing flake on `main` (reproduced there ~1 run in 3, in a file this branch does not touch). Fixed separately in #6303.

## Performance

Benchmarked on real MySQL in two worktrees off the same commit — full method, caveats, and statement-level evidence in [this comment](https://github.com/antiwork/gumroad/pull/6298#issuecomment-5079138821).

| | `main` (inline) | this branch |
|---|---|---|
| in-request statements | 5 | **1** |
| touches `audience_members` | **yes — 2 writes** | **no** |
| serial p50 / p95 | 27.91 / 44.70 ms | **1.74 / 2.22 ms** |
| concurrent p50 / p99 (12 threads, one license) | 79.43 / 216.27 ms | **14.00 / 68.64 ms** |
| throughput | 132.1 rps | **602.6 rps** |

Neither run reproduced a `LockWaitTimeout` — a local box can't generate enough contention to reach `innodb_lock_wait_timeout`. So this proves the row lock is off the request path and quantifies the win; the evidence it fixes the 500s remains the Sentry data plus the removed lock.

## Follow-ups this PR deliberately leaves alone

Scoped to stopping the 500s. Two separate questions, with production numbers, are on the tracking issue:

- **Narrower update.** `details["purchases"]` is rewritten whole on every verification, but `license_uses` is the only field in it that changes after the purchase exists. A targeted `JSON_SET` on the matching array element would hold the lock for a fraction of the time. Document sizes are small today (sample of 3,000 recent rows: p50 141 bytes, p95 393 bytes, largest 8.8 KB), so this is about lock duration under concurrency rather than payload size.
- **Whether the denormalization should exist at all.** `license_uses` is copied into this document to serve exactly one filter, "minimum license uses" on email posts. Across the most recent ~164,000 posts, **9** have that filter set (values 1 and 2). Reading the value from `licenses` at filter time would take the license path out of the audience write path entirely and delete this whole class of contention.

---

## AI disclosure

Written by Gumclaw running `claude-opus-5`. It was given the issue's diagnosis and asked to fix it: move the inline audience-member write off the license-verify request, add a red-first spec asserting no in-request `audience_members` write, audit the two `Subscription` callers for the same shape, and run the affected specs locally before opening the PR.


slavingia:
Prove performance benchmarks.

gumclaw:

Ran a local adversarial review of this branch. Verdict came back CHANGES-REQUIRED with one P1; that and the rest are addressed below. Fixes are in 44ad559 and 02c2e6b.

|---|---|---|
| P1 — the 2s delay collapses nothing; each verify enqueued its own job and every one redid the same full row rewrite, relocating the stampede into Sidekiq rather than removing it | **Real, and the important one** | Added `lock: :until_executed`. Verified empirically: 5 enqueues for one purchase → 1 job. Safe to dedupe on `purchase_id` because the job derives everything it writes from current DB state. |
| P2 — `find_by` + silent return contradicts the `ElasticsearchIndexerWorker` precedent this change mirrors, which raises for soft-deleted models because a missing row means replica lag | **Real** | Now `Purchase.find`, letting `RecordNotFound` ride the retries. Spec pins the raise instead of `not_to raise_error`. |
| P2 — job class should end in `Job` per the contributing guide | **Real** | Renamed to `UpdatePurchaseAudienceMemberDetailsJob` across all five files. |
| P2 — a video is required for every PR; the guide's only exemption is docs/skill-only | **Real — I had argued my way out of it** | Recorded a terminal walkthrough into `qa-media/pr-6298-walkthrough.mp4` and referenced it in Before/After. |
| P3 — `queue: :low` fits the guide's criterion better than `default` | **Real** | Moved to `:low`. |
| P3 — media file named with the issue number, not the PR number | **Real** | Renamed to `pr-6298-test-results.png`. |
| P2 — the resurrect test was uncommitted in the worktree | **Stale** | It was already committed in 63673819 before the review reported; the reviewer read a mid-flight tree. |
| P3 — `claude-opus-5` is not a real model name | **Declined** | `model.default` in the live config is `claude-opus-5`, read at write time. Naming anything else would make the disclosure false. |

Independently verified while the review ran, rather than taking its word:

- **Does this actually fix the 500?** Subscribed to `sql.active_record` around `increment!`: the only in-request statement is `UPDATE licenses SET uses = COALESCE(uses,0)+1 WHERE id = ?`. No `audience_members` write, no paper_trail insert (it tracks only `disabled_at`/`serial`), ES enqueue is Redis-only.
- **Caller coverage.** All three `add_to_audience_member_details` call sites moved. `Purchase`'s own `after_save` is deliberately left inline, with the reasoning in the PR body — it isn't what Sentry is firing on, it isn't concurrent per-buyer, and its email-change branch needs an ordered remove-then-add that this job doesn't model.
- **New assertions are load-bearing.** Reverting `app/models/license.rb` to main reddens 2 license examples; reverting `app/models/subscription.rb` reddens both Subscription enqueue tests.

Worth recording that the review confirmed the design's central safety claim: the add/remove race can't resurrect a removed member, because `add_to_audience_member_details` re-checks `should_be_audience_member?` against current state at job time. There's a test for it.

Not merging this myself — leaving it for review.

gumclaw:
Heads up on the auto-merge here — it was armed at 14:53, before the pre-merge review came back. The review found a real P1 that green CI would not have caught: the two-second delay on its own collapsed nothing, so a client verifying the same license repeatedly still enqueued one job per call and each one redid the full rewrite of the same `audience_members` row. That moves the pile-up into Sidekiq rather than removing it, which would have left the endpoint's contention partly intact after shipping a fix for it.

Fixed with `lock: :until_executed` (5 enqueues for one purchase now collapse to 1, verified by running it). Full triage of the review is in the comment above.

Nothing to do on your end — the fix is pushed and CI is re-running against it. Flagging it only because the merge would have landed the weaker version if the review had come back a few minutes later, and the ordering there is worth knowing about.

gumclaw:
## Benchmarks

Ran the real thing rather than reasoning about it. Same script in two worktrees off the same commit — one on unmodified `main`, one on this branch — against real MySQL, real Redis, real Elasticsearch. The only difference between the two runs is the application code, so the delta is attributable to this change.

It measures `License#increment!(:uses)`, which is the operation `POST /v2/licenses/verify` actually performs, not a synthetic stand-in. Fixture: one seller, one buyer, 25 purchases collapsed into that buyer's single audience document (2,457 bytes). 200 serial iterations after a warmup, then 300 requests across 12 threads all hitting the **same** license — the shape from the Sentry issue, where one integration verifies one license concurrently.

### In-request cost, no contention

| | `main` (inline) | this branch | |
|---|---|---|---|
| mean | 26.76 ms | **1.80 ms** | 14.9x |
| p50 | 27.91 ms | **1.74 ms** | 16.0x |
| p95 | 44.70 ms | **2.22 ms** | 20.1x |
| p99 | 47.95 ms | **2.48 ms** | 19.3x |
| max | 57.65 ms | **3.77 ms** | 15.3x |

### Under concurrency (12 threads, 300 requests, one license)

| | `main` (inline) | this branch | |
|---|---|---|---|
| p50 | 79.43 ms | **14.00 ms** | 5.7x |
| p95 | 161.55 ms | **48.71 ms** | 3.3x |
| p99 | 216.27 ms | **68.64 ms** | 3.2x |
| max | 238.86 ms | **104.81 ms** | 2.3x |
| wall clock | 2.27 s | **0.50 s** | |
| throughput | 132.1 rps | **602.6 rps** | 4.6x |

### The mechanism behind the numbers

Latency figures are the symptom; this is the cause. Subscribing to `sql.active_record` and counting what the operation issues **in-request**:

| | `main` | this branch |
|---|---|---|
| statements in-request | 5 | **1** |
| touches `audience_members` | **yes — 2 writes** | **no** |
| tables written | `licenses`, `audience_members` (x2), `base_variants`, `users` | `licenses` |

On `main` the request writes `audience_members` twice and therefore holds a write lock on that row for the length of the request. On this branch the only statement is the atomic `UPDATE licenses SET uses = COALESCE(uses,0)+1 WHERE id = ?`. The lock is not shortened — it is off the request path entirely, which is the point of the change.

### What this does not prove

Being straight about the limit: **neither run reproduced the production `LockWaitTimeout` — zero errors on both sides.** Twelve threads against a warm local single-instance MySQL doesn't generate enough contention to reach `innodb_lock_wait_timeout`. So these numbers demonstrate the row lock is gone from the request and quantify the latency and throughput win; they are not an independent reproduction of the 500.

The evidence that this fixes the 500s is still the Sentry data (1,566 events, 195 users, all one transaction, the last 100 inside a single one-minute window from one integration) plus the removed lock — not this benchmark. I'd rather say that than present the benchmark as more than it is.

Two other caveats worth stating: the absolute millisecond values are local-hardware numbers and production will differ, and the 25-purchase document is deliberately larger than typical (production sample of 3,000 recent rows: p50 141 bytes, p95 393 bytes), which makes the inline write look worse than the median case while matching the tail this fires on.

### Reproducing

Script is at `/tmp/bench_license.rb` on my box, parameterised by `ITERATIONS` / `CONCURRENCY` / `PER_THREAD` / `PURCHASES`. Say the word and I'll commit it under the repo so anyone can re-run it, though I left it out of the diff deliberately — it's throwaway measurement code, not something worth maintaining.

Two notes on running it, both cost me a cycle: `DISABLE_SPRING=1 OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` is required or the threaded section crashes on macOS in the `fork()` child, and the fixture needs one product per purchase because Gumroad rejects a second successful purchase of the same product by the same buyer.

gumclaw:
## The red `Test Minitest` run, and what it was

Run [30162977485](https://github.com/antiwork/gumroad/actions/runs/30162977485) failed with two Minitest failures. One was mine; one was not. Both are now resolved and `Test Minitest` is **green** on the current head (`ee0e686`).

### 1. A real regression from this PR — fixed in `ee0e686`

```
PurchaseTest#test_AudienceMember_removes_member_when_subscription_is_deactivated
#<Proc:...purchase_test.rb:5273 (lambda)> didn't change by 1, but by 0.
```

That test asserted the audience member reappeared **during** `resubscribe!`. This PR is precisely what moves that write into `UpdatePurchaseAudienceMemberDetailsJob`, so the old assertion was pinning behavior the change intentionally alters — a correct failure that caught a genuine gap in my test sweep. I had covered the two `Subscription` callers in `subscription_test.rb` but missed that `purchase_test.rb` asserted the same write from the other side.

Rewritten to pin the sharper shape: `resubscribe!` adds nothing in-request, and draining the job is what brings the member back. Removal stays synchronous and the test now says so explicitly, since nothing re-checks a removal later.

```ruby
# Removal is still synchronous: it has to be, because nothing re-checks it later.
assert_difference -> { AudienceMember.count }, -1 do
  purchase.subscription.deactivate!
end

# The re-add is deferred, so resubscribe! only enqueues.
UpdatePurchaseAudienceMemberDetailsJob.jobs.clear
assert_no_difference -> { AudienceMember.count } do
  purchase.subscription.resubscribe!
end

assert_difference -> { AudienceMember.count }, 1 do
  UpdatePurchaseAudienceMemberDetailsJob.drain
end
```

This is a strictly stronger assertion than what it replaced: it now pins *both* that the write leaves the request and that it still happens.

### 2. A pre-existing flake on `main` — split out to #6303

```
InstallmentTest#test_publish!_publishes_the_installment [test/models/installment_test.rb:601]
Expected |1784992827.0021424 - 1784992826.0| (1.0021424293518066) to be <= 1.0
```

Two milliseconds over a one-second budget, in a file this branch does not touch. `publish!` stamps `published_at` on entry then does real work (`transcode_videos!`, moderation, `save!`), but the test read `Time.current` *after* the call returned — so the tolerance was absorbing the method's own runtime, making the pass/fail line "is the runner busy."

Confirmed pre-existing rather than assumed: reproduced on an unmodified `main` worktree, where it fails roughly **1 run in 3**. Fixed in #6303 with `freeze_time`, along with two sibling tests carrying the same latent construction. Kept out of this PR because it isn't this change's bug, and a timing flake on `main` bills its debugging time to whoever's PR catches it next.

### Verification

Local, after the fix:

| Suite | Result |
|---|---|
| `purchase_test.rb:5266` (the failure) | green, 6 assertions |
| `purchase_test.rb` `/AudienceMember/` (9 tests) | 9 runs, 47 assertions, 0 failures |
| `purchase_test.rb` full | 674 runs, 1,658 assertions, 0 failures |
| `subscription_test.rb` full | 433 runs, 1,018 assertions, 0 failures |
| `installment_test.rb` full (on #6303) | 111 runs, 0 failures |

CI: `Test Minitest` green on `ee0e686`, the same job that was red. Remaining shards still running at time of writing.

### One process note worth recording

Mid-triage I hit a `LockWaitTimeout` on the failing test locally, and a `subscription_test.rb` run reporting **227 errors** (`Couldn't find MerchantAccount`). Both were artifacts of my own killed test runs leaving locked rows and half-written fixtures behind — they vanished on a clean re-run and did not reproduce on `main`. Flagging it because those numbers look alarming and reporting them as findings would have been noise; the check that settled it was baselining on an unmodified `main` worktree, which is also what separated the real regression above from the flake.

gumclaw:
Noticed only after posting the two comments above that you'd closed this at 15:27 — I was watching the CI run rather than the PR state, so the benchmarks and the CI triage landed on an already-closed PR. My miss.

Leaving it closed. Not reopening on my own judgment, since closing it is your call and you may well have decided the approach isn't worth it regardless of the numbers.

For the record, in case the close was because the benchmarks hadn't materialised: they're in [this comment](https://github.com/antiwork/gumroad/pull/6298#issuecomment-5079138821). Two worktrees off the same commit against real MySQL — 5 in-request statements down to 1, the two `audience_members` writes (and their row lock) gone from the request path entirely, serial p50 27.91 ms → 1.74 ms, throughput 132 → 603 rps. With the honest caveat stated there: neither side reproduced a `LockWaitTimeout`, because a local box can't generate enough contention to hit `innodb_lock_wait_timeout`, so those numbers prove the lock left the request path but don't independently reproduce the production 500.

The red CI that was on here is also resolved: one real regression from this change (a test that asserted the audience write happened in-request — fixed, and the assertion is now stronger), and one pre-existing timing flake on `main` that this branch didn't cause. That flake is fixed separately in #6303, which is green and stands on its own whatever happens here.

Two ways forward, your call:

1. **Reopen** — say the word and I will. Branch is pushed at `ee0e686` with `Test Minitest` green.
2. **Leave it closed** — also fine. If it's the approach you don't want rather than the missing evidence, I'd find that useful to know, because the tracking issue (gumroad-private#1349) still has the 500s open and the two deeper follow-ups I'd scoped out (narrower `JSON_SET`, or dropping the `license_uses` denormalisation altogether so the license path leaves the audience write path for good) are both still on the table. The second one deletes this whole class of contention rather than relocating it, and might be the thing worth doing instead.

Either way #6303 is independent and worth merging on its own.

gumclaw:
Answering my own question from the previous comment, since the record should not be left speculating.

This was closed because the approach was superseded, not because the benchmarks were missing. On the tracking issue (gumroad-private#1349) Sahil picked option 1 — *"Agree go with that. It's overkill for such a minute feature"* and then *"1"* on the options list — which is **delete the denormalized `license_uses` copy and have the filter join `licenses` directly**, rather than keep the copy and move its write off the request path, which is what this PR does.

That is shipped as **antiwork/gumroad#6305**, assigned to @ershad. So this PR being closed is correct and I should not have offered to reopen it — that offer is withdrawn.

Worth being straight that this PR was the weaker fix, and my own analysis on the issue said so before this was closed: deferring the write "keeps a self-inflicted write amplification on every license check to serve a filter almost nobody uses, and it still contends (just where the buyer can't see it)." #6305 removes the write instead of relocating it, and independently fixes a correctness bug this PR would have preserved — any activation that skipped the callback left the stored count stale, so the "minimum license uses" filter selected the wrong recipients.

Nothing here is lost:

- The **benchmarks** above still describe the problem accurately (the row lock is the cost, 5 in-request statements down to 1) and the mechanism they document is what #6305 eliminates outright.
- The **pre-existing flake** found while triaging this PR's CI is fixed and merged independently in #6303.
- The **regression** this PR's CI caught was in a test asserting the audience write happened in-request. #6305 removes that write entirely, so that assertion needs revisiting there too — flagging it for @ershad so the same failure does not resurface on #6305.

Apologies for the noise of the three comments that landed after the close; I was polling the CI run rather than PR state and did not notice it had been closed. Fixed in how I watch runs going forward.

gumclaw:
Correction to my comment above, before it wastes anyone's time.

I flagged that the `purchase_test.rb` assertion (`AudienceMember removes member when subscription is deactivated`) would need revisiting on #6305. **That is wrong — checked the diff, and it does not.**

#6305 only removes the audience write from the **license** path (`License#increment!` and the `uses` branch of `update_purchase_search_index`). It leaves `Subscription`'s two callers writing synchronously — `original_purchase&.add_to_audience_member_details` is still there at both `subscription.rb:811` and `:1266` on that branch. That test exercises `deactivate!` / `resubscribe!`, so its in-request behavior is unchanged and it passes as-is on `main`.

The test only needed changing on *this* PR, because this PR is the one that moved the `Subscription` writes into a job as well. Since this PR is closed and superseded, that change goes away with it and nothing is owed on #6305.

Sorry for the bad ping, @ershad — no action needed from you on that point. #6305 stands on its own.

PR #6299, opened 2026-07-25
Use the product's support email as reply-to for product posts

gumclaw:
## What

Posts (creator emails) sent about a single product now use that product's support email as the reply-to address when the seller has set one. Audience-wide posts, which aren't tied to a product, keep using the account-level support address.

The resolution now lives in one place, `Installment#reply_to_email`, because both post senders (`PostResendApi` and `PostSendgridApi`) built the reply-to independently and both had to change.

## Why

Sellers can already set a support email per product (Settings → Main → Support → "Add a product specific email"), stored on `links.support_email`. Receipts honor it via `Link#support_email_or_default`, but the post/email sending path ignored it and always used `User#support_or_form_email`, so replies to a product's own announcement email landed in the account's main inbox instead of the inbox the seller picked for that product.

## Before / After

Reply-to header on a post about a product whose support email is `product@example.com`, with the account address set to `account@example.com`:

| | Before | After |
|---|---|---|
| Product-scoped post | `account@example.com` | `product@example.com` |
| Product-scoped post, product has no support email | `account@example.com` | `account@example.com` (unchanged) |
| Audience-wide post | `account@example.com` | `account@example.com` (unchanged) |

No user-visible surface changes — this only affects an email header, so there is no screen to screenshot. The behavior evidence is the spec run below.

## QA steps

1. As a seller, set an account support email in Settings → Main → Support.
2. In the same section, use "Add a product specific email" to give one product a different support address.
3. Publish a post scoped to that product with emails enabled.
4. Open the received email and inspect its Reply-To header — it is the product's address.
5. Publish an audience-wide post and confirm its Reply-To is still the account address.

## Test results

`bundle exec rspec spec/services/post_resend_api_spec.rb spec/services/post_sendgrid_api_spec.rb` → 76 examples, 0 failures (72 on unmodified main; 4 examples are new, two per sender covering the product address and the fallback).

spec run

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5`. Instructions given: fix the issue as described in antiwork/gumroad-private#1342 — make the post sending path resolve reply-to through the product's support email, change both the Resend and SendGrid senders, add specs, run them locally, and open a PR following the repo's conventions.


gumclaw:
## Pre-merge adversarial review (fable-5) — landed after the merge

For the record, since this merged while the review was still running: the verdict was **READY-TO-MERGE**, no P1 or P2 findings. Nothing here needs undoing.

### Confirmed safe, with evidence

- **No cross-user leak.** Products attach to posts via `Link.fetch(link_id, user: seller)` (`save_installment_service.rb:126`, scoped `by_user` at `link.rb:749`), and product workflows are built with `seller: product.user` (`workflow/manage_service.rb:91`), so `link.user == installment.seller` on every write path. There's no product-transfer service in the repo.
- **`link` is only set when the post is product-scoped.** `save_installment_service.rb:56` nulls it for non-product/variant types and `workflow/manage_service.rb:27` does the same for workflows, so seller/follower/audience/affiliate posts keep byte-for-byte the old reply-to.
- **The comment's reasoning holds.** `Link#support_email_or_default` (`link.rb:1297`) falls back to the product owner, not the post's seller. Reading the column directly keeps the fallback on the sender — and `.presence` is stricter than that helper's bare `||` on legacy empty strings.
- **No third reply-to path left inconsistent.** Every post email funnels through `PostEmailApi` into the two changed senders: blasts, last-post-to-new-buyer, previews, workflow sends, single-customer resends. The recorded-email values at `post_resend_api.rb:263` / `post_sendgrid_api.rb:279` read back from the already-built message, so they follow automatically. `OneOffMailer#email_using_installment` takes an explicit `reply_to:` and has no production callers.
- **Degenerate values can't reach the column.** `link.rb:188` validates with `email_format` + `not_reserved_email_domain`, `allow_nil`; `Product::BulkUpdateSupportEmailService` rejects blanks and re-validates before its `update_all`.
- **Deliverability precedent.** Receipts already send this same user-set product address as reply-to (`customer_mailer.rb:109` and `:357`) through the same providers. Only From requires domain verification, and From is unchanged.
- **No N+1.** `reply_to_email` runs per-recipient, but `link` and `seller` are associations memoized on the single `@post` instance — same query count as the previous per-recipient `@post.seller.support_or_form_email`.

### Follow-up

Three P3s, all about spec coverage rather than behavior: the `nil` fallback example passed on revert, and neither variant posts nor workflow installments were pinned even though both carry a `link` and take this path. Addressed in #6301 (tests only).

PR #6301, opened 2026-07-25
Cover variant posts and blank product support emails in post reply-to specs

gumclaw:
## What

Adds two spec examples per post sender covering cases the reply-to resolution in #6299 handles but did not pin:

- A **variant-scoped post** (`installment_type: "variant"`) also carries a `link`, so it takes the product reply-to path. This is correct — a variant post goes to buyers of that product — but nothing asserted it.
- A **blank** `links.support_email` must fall back to the account address rather than send an empty Reply-To. The model validation rejects blanks today, so the spec uses `update_column` to reproduce a row saved before that validation existed. The original `nil` example is kept alongside it, so both branches of the presence check are pinned.

Tests only — no production code changes.

## Why

Follow-up to #6299, addressing coverage gaps found by the adversarial pre-merge review on that PR. The variant path was exercised in production but unpinned, and nothing distinguished `.presence` from a bare `||` — a regression to the latter would have leaked an empty Reply-To with the whole suite still green.

## Test results

`bundle exec rspec spec/services/post_resend_api_spec.rb spec/services/post_sendgrid_api_spec.rb -e "reply-to for a post about one product"` → 8 examples, 0 failures.

Revert checks, each run locally:
- Reducing `Installment#reply_to_email` to `seller.support_or_form_email` (i.e. reverting #6299) fails 4 of the 8, including both new variant examples.
- Degrading `.presence` to a bare `||` fails exactly the two new blank-support-email examples, which is what they exist to catch.

Note on the full-file run: this worktree shows 18 pre-existing failures in examples this diff does not touch, and clean `main` in the same worktree shows the same 18. They are local environment contention (concurrent worktrees sharing MySQL/Elasticsearch/Redis), not related to this change. CI is the authority for the full suite.

## QA steps

No user-facing change — this PR only adds test coverage. Verification is the spec run above.

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5`. Instructions given: address the P3 coverage nits raised by the pre-merge adversarial review of #6299 — pin variant-post behavior and make the fallback example actually fail on revert — as a tests-only follow-up.

### DIFF
--- a/spec/services/post_resend_api_spec.rb
+++ b/spec/services/post_resend_api_spec.rb
@@ -268,10 +268,21 @@ describe PostResendApi, :freeze_time do
     end
 
     it "falls back to the account address when the product has no support email of its own" do
-      @post.link.update!(support_email: nil)
+      # update_column skips validation on purpose: the column rejects blanks today, but older
+      # rows can still hold an empty string, and those must fall back rather than send an
+      # empty Reply-To.
+      @post.link.update_column(:support_email, "")
       send_default_email
 
       expect(sent_email[:reply_to]).to include("account@example.com")
+    end
+
+    it "uses the product's support email for a post scoped to one of its variants" do
+      variant = create(:variant, variant_category: create(:variant_category, link: @post.link))
+      @post = create(:variant_installment, name: "post title", message: "post body", seller: @seller, link: @post.link, base_variant: variant)
+      send_default_email
+
+      expect(sent_email[:reply_to]).to include("product@example.com")
     end
   end
 
--- a/spec/services/post_sendgrid_api_spec.rb
+++ b/spec/services/post_sendgrid_api_spec.rb
@@ -255,10 +255,21 @@ describe PostSendgridApi, :freeze_time do
     end
 
     it "falls back to the account address when the product has no support email of its own" do
-      @post.link.update!(support_email: nil)
+      # update_column skips validation on purpose: the column rejects blanks today, but older
+      # rows can still hold an empty string, and those must fall back rather than send an
+      # empty Reply-To.
+      @post.link.update_column(:support_email, "")
       send_default_email
 
       expect(sent_email[:reply_to]).to include("account@example.com")
+    end
+
+    it "uses the product's support email for a post scoped to one of its variants" do
+      variant = create(:variant, variant_category: create(:variant_category, link: @post.link))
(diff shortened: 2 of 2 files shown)


gumclaw:
## Pre-merge adversarial review (fable-5)

**READY-TO-MERGE** — no P1 or P2 findings. Both P3s were taken and are fixed in `14be2d7`.

|---|---|---|
| Comment said "the column rejects blanks" — it's the model validation at `link.rb:188`, there's no DB constraint | P3 | Fixed — comment reworded |
| Swapping `nil` → `""` dropped explicit coverage of the `nil` branch | P3 | Fixed — both cases now have their own example (8 total, was 6) |

- **Reassigning `@post` inside the variant example is not vacuous** (the item I flagged as highest-risk). `send_emails` constructs the service at call time from the current `@post`, and `cache:` defaults to a fresh hash per instantiation, so nothing memoizes across posts. `described_class.mails` is class-level but cleared in the top-level `before`.
- **The variant fixture is coherent, not accidentally valid.** `Installment` does *not* validate that `base_variant` belongs to `link` (both are `optional: true`), so the default factory would have built a variant of an unrelated product — passing the assertion for the wrong reason. Overriding `variant_category: create(:variant_category, link: @post.link)` is what makes link, category and variant belong to one product.
- **`update_column` is required, and the state it pins is real.** `EmailFormatValidator` returns false for blank and the validation is `allow_nil: true` without `allow_blank`, so `update!(support_email: "")` would raise `RecordInvalid`. No DB constraint exists, so pre-validation rows can hold `""` — and `Link#support_email_or_default` still uses a bare `||`, which would leak it.
- **The orphaned before-block post can't leak into assertions** — emails only enter `mails` during an explicit `send_emails` call, so there's exactly one entry per example.
- **The two spec files are line-for-line symmetric** (verified by reading both files, not just the diff hunks).
- **No flakiness vectors**: both specs run under `:freeze_time`, factories use sequences, and the assertions match against a plain string.

### Verified by revert, not by reasoning

- Reverting `Installment#reply_to_email` to `seller.support_or_form_email` fails 4 of 8, including both variant examples.
- Degrading `.presence` to a bare `||` fails exactly the two blank-support-email examples.

One honest caveat: the full two-file run in my local worktree shows 18 failures in examples this diff doesn't touch, and clean `main` in that same worktree shows the same 18 — local service contention between concurrent worktrees, not this change. CI is the authority there.

PR #6303, opened 2026-07-25
Stop the installment publish! tests failing on their own runtime

gumclaw:
## What

Freezes time around the three `Installment#publish!` tests that assert `published_at` lands at "now", and replaces their `assert_in_delta` tolerances with exact equality.

## Why

`publish!` stamps `published_at = Time.current` on entry, then does real work before returning — `transcode_videos!`, the content-moderation check, and a `save!`. These tests took `Time.current` *after* the call returned and compared it to that entry-time stamp, so the tolerance was not absorbing clock skew, it was absorbing `publish!`'s own runtime.

That means the pass/fail line was "did `publish!` finish in under a second", which is a property of how loaded the runner is rather than anything about the code under test. On a slow CI box it goes red:

```
InstallmentTest#test_publish!_publishes_the_installment [test/models/installment_test.rb:601]:
Expected |1784992827.0021424 - 1784992826.0| (1.0021424293518066) to be <= 1.0.
```

That is 2 milliseconds over a one-second budget. It surfaced as a red `Test Minitest` job on an unrelated PR (#6298), which is the actual cost here — a timing flake on `main` charges its debugging time to whoever's PR happens to catch it.

Reproduced on unmodified `main` before changing anything: the test fails roughly 1 run in 3 on this machine, with no branch changes involved. After the fix, 3 consecutive runs pass and the full file is green (111 runs, 0 failures).

Two sibling tests below it (`skips the content moderation check for VIP creators`, `succeeds when the content moderation check passes`) had the same construction with a 2.0s tolerance. They had not gone red yet, but they fail the same way under enough load, so they are fixed in the same pass rather than left to flake later. Those two compare `.to_i` because `installments.published_at` is declared `precision: nil`, so the value truncates to the whole second on reload.

The test that passes an explicit `published_at:` argument was already deterministic and is untouched.

## Test plan

- `bin/rails test test/models/installment_test.rb -n "/publish!_publishes_the_installment/"` — 3 consecutive runs, green each time (the same invocation failed 1-in-3 on `main`).
- `bin/rails test test/models/installment_test.rb` — full file, 111 runs, 205 assertions, 0 failures, 0 errors, 1 skip.

No application code changes, so there is no behavior to QA — the assertions still pin that `publish!` sets `published_at` to the current time, just against a clock that does not move mid-test.

## AI disclosure

Written by Gumclaw running `claude-opus-5`. Found while triaging a red `Test Minitest` job on #6298: this failure was in that run but was not caused by that branch, so it was confirmed pre-existing on `main` by reproducing it there and split out into its own PR.

### DIFF
--- a/test/models/installment_test.rb
+++ b/test/models/installment_test.rb
@@ -597,8 +597,13 @@ const b = 2;</code></pre>
 
   test "publish! publishes the installment" do
     assert_nil @installment.published_at
-    @installment.publish!
-    assert_in_delta Time.current.to_f, @installment.published_at.to_f, 1.0
+    # Frozen so the assertion compares against the same instant publish! stamped. Comparing a
+    # timestamp taken on entry to publish! against Time.current after it returns makes the
+    # tolerance absorb the method's own runtime, which is why this flaked on slow runners.
+    freeze_time do
+      @installment.publish!
+      assert_equal Time.current, @installment.published_at
+    end
   end
 
   test "publish! sets published_at to the provided argument" do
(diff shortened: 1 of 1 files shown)


gumclaw:
@ershad handing this to you — you are the main author of `test/models/installment_test.rb`.

Test-only diff, one file, no production code touched. Full CI is green at `03c793a` (all Test Fast/Test Slow/Minitest/Lint shards, Buildkite green) and Greptile approved it 5/5.

The one thing worth your eye rather than a rubber stamp: these three tests were passing or failing based on **how long `publish!` takes to run**, not on what it does. They read `Time.current` after the call returned and compared it against the `published_at` stamped when the call *started*, so the `assert_in_delta` 1-second tolerance was absorbing `publish!`'s own runtime (`transcode_videos!`, the content-moderation check, a `save!`). On a loaded runner that crosses one second and goes red — the observed failure was 1.0021s against a 1.0s tolerance.

Freezing time around the three assertions lets them assert exact equality instead, so they now test that `published_at` is stamped at entry, which is the actual contract. If you would rather widen the tolerance than freeze time, say so and I will switch it — but a tolerance only moves the goalposts, it does not stop the tests depending on runner speed.

PR #6305, opened 2026-07-25
Read license use counts from the licenses table instead of a copy in the audience row

gumclaw:
Fixes the top-volume Sentry issue on the license API.

## What

`POST /v2/licenses/verify` returned 500s under concurrent calls for the same buyer — **Sentry GUMROAD-YX, 1,566 events / 195 users**, and `tags/transaction/` confirms all 1,566 are this one endpoint, not a diffuse lock-contention bucket.

This stops the endpoint writing to the buyer's `audience_members` row on every activation, and has the audience filter read the use count from the `licenses` table instead.

## Why it 500'd

`License#increment!("uses")` called `Purchase#add_to_audience_member_details`, which does `member.save!` on the buyer's `audience_members` row. MySQL takes a row lock for that write, so two activations for the same buyer serialized behind each other; once the wait passed `innodb_lock_wait_timeout` the request raised `ActiveRecord::LockWaitTimeout` and the endpoint 500'd.

That write existed for one reason: to keep a copy of the use count inside the row's `details` JSON so `AudienceMember.filter` could read it back as `$.license_uses`.

## Why the copy can just go

A license belongs to exactly one purchase, so the filter can join:

```sql
INNER JOIN licenses ON licenses.purchase_id = jt.purchase_id AND licenses.deleted_at IS NULL
```

Removing the write removes the lock. It also removes a correctness bug the copy had independently of the 500s: any activation that did not go through that callback left the stored count stale, and the filter then selected the wrong recipients for a post.

## The part that isn't obvious

`AudienceMember.filter` (recipient selection) is MySQL, but **`filter_count` counts in Elasticsearch** — two engines behind one filter, and that file's own comment requires the count and the recipient list to agree. With the refresh gone, an ES copy of the count would drift the moment a buyer activated a license, so a seller would see a recipient count that didn't match who actually got the post. `filter_count` now routes through the SQL filter when `minimum_license_uses` is present.

I left the `license_uses` field in the audience document and its ES mapping alone. Removing them is an index change with a wider blast radius than this fix needs, and nothing reads them once the filter joins.

**The Customers page filter is unaffected** — it queries the purchases ES index via `PurchaseSearchService` (`license_uses_greater_than_or_equal_to`), a separate path that never took this lock.

## Test Results

```
36 examples, 0 failures
```

The existing filter spec fabricated `license_uses` inside the details JSON with no `licenses` rows behind it, so it only ever asserted that the denormalized copy could be read — it could not have caught staleness. It now builds real products, purchases and licenses, plus a new example that mutates `license.uses` and re-runs the filter.

Red-first: reverting `audience_member.rb` with the new tests in place fails that example, which is the staleness the old design allowed.

## Before/After

Not a UI change — the fix is on the write path of a public API endpoint. The user-visible delta is that concurrent activations for one buyer stop returning 500s. Verification is the Sentry event rate for GUMROAD-YX after this deploys; I'll confirm it on the issue rather than claiming it here.

---

**AI disclosure:** Generated with Claude Opus 5 (Anthropic), operating as the Gumclaw Hermes agent. Prompts: triage the top Sentry issues on the gumroad project and fix them; for this one, "just delete" the denormalization and join the licenses table in the filter.

### DIFF
--- a/app/models/audience_member.rb
+++ b/app/models/audience_member.rb
@@ -149,8 +149,7 @@ class AudienceMember < ApplicationRecord
             purchase_price_cents INT PATH '$.price_cents',
             purchase_created_at DATETIME PATH '$.created_at',
             purchase_country CHAR(100) PATH '$.country',
-            purchase_subscription_cancelled INT PATH '$.subscription_cancelled',
-            purchase_license_uses INT PATH '$.license_uses'
+            purchase_subscription_cancelled INT PATH '$.subscription_cancelled'
           ),
           NESTED PATH '$.affiliates[*]' COLUMNS (
             affiliate_id INT PATH '$.id',
@@ -170,7 +169,15 @@ class AudienceMember < ApplicationRecord
         json_filter = json_filter.where("jt.purchase_id IS NOT NULL AND COALESCE(jt.purchase_subscription_cancelled, 0) = 0")
       end
       if params[:minimum_license_uses]
-        json_filter = json_filter.where("jt.purchase_license_uses >= ?", params[:minimum_license_uses].to_i)
+        # Read the use count from the licenses table rather than from a copy kept inside
+        # this row's `details` JSON. Keeping a copy meant every license activation had to
+        # rewrite the buyer's audience_members row, and because MySQL takes a row lock for
+        # that write, concurrent activations for the same buyer queued behind each other
+        # and the license-verification endpoint returned 500s once the lock wait timed out.
+        # A license belongs to exactly one purchase, so joining is a direct lookup.
+        json_filter = json_filter
+          .joins("INNER JOIN licenses ON licenses.purchase_id = jt.purchase_id AND licenses.deleted_at IS NULL")
+          .where("licenses.uses >= ?", params[:minimum_license_uses].to_i)
       end
       timestamp_columns = if params[:type] == "customer"
         %w[purchase_created_at]
--- a/app/models/concerns/audience_member/searchable.rb
+++ b/app/models/concerns/audience_member/searchable.rb
@@ -94,6 +94,19 @@ module AudienceMember::Searchable
       # SQL behavior so callers don't have to special-case seller-less records.
       return 0 if seller_id.blank?
 
+      # The license-use filter now reads live `licenses` rows, which only the SQL twin can
+      # join to — Elasticsearch has no joins. Count through SQL for those so the number the
+      # seller sees and the recipients the blast selects come from one source. This index
+      # still carries a `license_uses` copy per purchase, but activations no longer refresh
+      # it: that refresh rewrote the buyer's audience_members row on every activation, and
+      # the row lock it took made concurrent activations queue until the license-verify
+      # endpoint timed out with a 500.
+      if params[:minimum_license_uses].present?
+        relation = AudienceMember.filter(seller_id:, params:)
+        relation = relation.limit(limit) if limit
+        return relation.count
+      end
+
       options = {
         index: index_name,
         body: { query: filter_query(seller_id:, params:) },
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -56,10 +56,7 @@ class License < ApplicationRecord
 
   def increment!(attribute, by = 1, touch: nil)
     super.tap do
-      if attribute.to_s == "uses"
-        enqueue_purchase_search_index_update(["license_uses"])
-        update_purchase_audience_member_details
-      end
+      enqueue_purchase_search_index_update(["license_uses"]) if attribute.to_s == "uses"
     end
   end
 
@@ -69,7 +66,6 @@ class License < ApplicationRecord
       fields << "license_serial" if previous_changes.key?("serial")
       fields << "license_uses" if previous_changes.key?("uses")
       enqueue_purchase_search_index_update(fields)
-      update_purchase_audience_member_details if previous_changes.key?("uses")
     end
 
     def enqueue_purchase_search_index_update(fields)
@@ -82,7 +78,4 @@ class License < ApplicationRecord
                                             })
     end
 
-    def update_purchase_audience_member_details
-      purchase&.add_to_audience_member_details
-    end
 end
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -77,5 +77,4 @@ class License < ApplicationRecord
                                               "fields" => fields
                                             })
     end
-
 end
--- a/spec/models/audience_member_spec.rb
+++ b/spec/models/audience_member_spec.rb
@@ -226,9 +226,9 @@ RSpec.describe AudienceMember, :freeze_time do
       member_without_uses = create_member_with_licenses([{ product_id: 1, uses: 0 }])
       member_with_three = create_member_with_licenses([{ product_id: 1, uses: 3 }])
       member_with_two_products = create_member_with_licenses([
-                                                              { product_id: 1, uses: 1 },
-                                                              { product_id: 2, uses: 5 }
-                                                            ])
+                                                               { product_id: 1, uses: 1 },
+                                                               { product_id: 2, uses: 5 }
+                                                             ])
       member_without_license = create_member_with_licenses([{ product_id: 1, uses: nil }])
 
       product_ids = { 1 => member_with_three.details["purchases"].first["product_id"] }
--- a/spec/models/audience_member_spec.rb
+++ b/spec/models/audience_member_spec.rb
@@ -403,7 +403,7 @@ RSpec.describe AudienceMember, :freeze_time do
     email = generate(:email)
     purchases = specs.map do |spec|
       product = create(:product, user: seller, is_licensed: true)
-      purchase = create(:purchase, link: product, seller:, email:, purchase_state: "successful")
+      purchase = create(:purchase, :from_seller, link: product, seller:, email:)
       create(:license, purchase:, link: product, uses: spec[:uses]) unless spec[:uses].nil?
       purchase
     end
--- a/spec/models/audience_member_spec.rb
+++ b/spec/models/audience_member_spec.rb
@@ -221,19 +221,37 @@ RSpec.describe AudienceMember, :freeze_time do
     end
 
     it "filters by minimum license uses, matching combined filters within a single purchase" do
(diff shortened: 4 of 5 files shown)


slavingia:
Run performance benchmarks

gumclaw:
## Performance benchmarks

Ran these locally against seeded MySQL 8 (50,000 audience members for one seller, 3 purchases each in `details`, 150,000 `licenses` rows — one per purchase). Both variants are the **exact SQL `AudienceMember.filter` generates**, before and after this PR, for `minimum_license_uses: 2`.

**Correctness first:** both return the identical result, 46,189 matching members. The join and the JSON copy agree, which is the baseline this whole change needs.

### Read path — this is a regression, and it is the headline

| variant | run 1 | run 2 | run 3 | run 4 | run 5 |
|---|---|---|---|---|---|
| before (`license_uses` from the `details` JSON) | 475 ms | 480 ms | 445 ms | 441 ms | 794 ms |
| after (`INNER JOIN licenses`) | 1064 ms | 1175 ms | 1063 ms | 962 ms | 1084 ms |

**~2.2× slower, about +600 ms at 50k members.** The plan is sane — `licenses` is reached by `eq_ref`/`ref` on `index_licenses_on_purchase_id`, not scanned — so this isn't a missing index. It's the intrinsic cost of one index lookup per purchase row the `JSON_TABLE` emits (150k probes here) versus reading an integer that was already inlined in the JSON. It will scale with a seller's purchase count, so a seller with 10× this audience should expect proportionally more.

One caveat on how hard to read these absolute numbers: they're from a laptop MySQL with everything in the buffer pool. Production has real network latency per lookup and a colder cache, so the *ratio* is the transferable part, not the milliseconds.

### Write path — the win this PR is actually buying

Single activation, serial (`uses` increment, plus the `audience_members.details` rewrite in the before case):

| variant | median |
|---|---|
| before | ~75 ms |
| after | ~67 ms |

20 concurrent activations for the **same buyer** — the contention this PR exists to remove — with that buyer's `details` inflated to a realistic heavy shape (2,000 purchases, 215 KB):

| variant | trial 1 | trial 2 | trial 3 |
|---|---|---|---|
| before | 124 ms | 114 ms | 111 ms |
| after | 92 ms | 93 ms | 97 ms |

~20% faster wall clock, but the number understates the point. Before, all 20 writers serialise on the **one** `audience_members` row for that buyer, each rewriting a 215 KB JSON document under the lock. Locally that queue drains in milliseconds; in production the same queue is what pushed license-verify past `innodb_lock_wait_timeout` into 500s. Removing a shared lock from a hot path is a tail-latency and correctness fix, and a 20-connection laptop test cannot reproduce the tail that motivated it.

### The honest summary

This trades a **read regression that shows up on every audience filter using the license-uses criterion** for the removal of a **write-side row lock that was returning 500s**. I still think that's the right trade — a slower query beats an endpoint that errors — but it is a trade, and the read cost wasn't quantified in the PR description before now.

Two things worth a decision, @ershad:

1. **Is +600 ms acceptable on this filter**, or should the read cost be bought back before this merges? A covering index on `licenses (purchase_id, uses, deleted_at)` would let the probes be index-only and should recover a good part of it. I have not added it — it's a schema change on a large table and that's your capacity call, the same one still open on #6195.
2. **The count path now bypasses Elasticsearch.** `AudienceMember::Searchable` routes `minimum_license_uses` counts through SQL because Elasticsearch cannot join. That is correct — it keeps the count the seller sees and the recipients a blast selects reading from one source — but it means this filter's count is now subject to the read cost above rather than being an ES aggregation. Worth confirming that's the behaviour you want on the email-blast recipient counter.

Lint failure from earlier is fixed in `b90e910c9` (three auto-correctable Rubocop offences: a trailing blank line in `license.rb`, array indentation in the spec). Rubocop is clean on all four changed files now.

Benchmark scripts are reproducible — seed plus both harnesses — happy to attach them to the PR if useful.

---

🤖 Benchmarks run by the Hermes agent (`claude-opus-5`) at Sahil's request on this PR.

gumclaw:
CI is red at head `b90e910c9` and this one is **not** a flake — do not rerun it. The failing assertion is about the exact filter this PR rewrites, so it needs a fix.

```
spec/models/concerns/audience_member/searchable_spec.rb:292
  AudienceMember::Searchable.filter_count
    counts by minimum license uses, matching combined filters within a single purchase

  Failure/Error: expect(AudienceMember.filter_count(seller_id:, params:)).to eq(expected)
    expected: 2
         got: 0
```

Run [30164844473](https://github.com/actions/runs/30164844473) (`push`), job "Test Fast 3", 1410 examples / 1 failure. The `pull_request_target` twin at the same SHA reports success and Buildkite [#15827](https://buildkite.com/gumroad-inc/gumroad/builds/15827) is green, which is why the PR looks ambiguous — but the twin-run rule does not apply here: this assertion is squarely about the behaviour the diff changes, so the green twin is the misleading signal, not the red one.

**Why it fails, as I read it:** this PR removes the denormalized `license_uses` copy from `audience_members.details.purchases[]` and joins `licenses` instead. The `filter_count` path is the **Elasticsearch** `nested(purchases)` branch, and that spec seeds the denormalized document. With the copy gone, the ES-side minimum-license-uses filter has nothing to match on and returns 0 — a real gap, not a stale pin: the SQL `JSON_TABLE` path and the ES path both implement this filter, and only one of them can be fixed by joining a MySQL table.

That is the open design question for this PR, and it's worth deciding before more CI cycles: the ES index has no join, so either the filter's ES branch keeps a denormalized value (in which case the write has to keep happening somewhere, just off the request path — the original #6298 shape), or the ES branch stops serving that filter and it falls through to the SQL path.

Given the filter is used by 9 posts across 9 sellers out of ~94,789 (the measurement on the driving issue) and Sahil's ruling was that the denormalization is "overkill for such a minute feature," dropping the ES branch and letting the filter evaluate over SQL looks like the intended reading — but that is a scope call on top of the approved one, so I'm surfacing it rather than picking it unilaterally while the PR sits with a reviewer.

@ershad this is still yours for review; I'll take the fix once the branch question above is settled (or say the word and I'll ship the SQL-only reading).

gumclaw:
Ran these on the branch at `40b6fc2` against local MySQL 8.0.46. Each case measures the same query/write two ways: the branch as written, and the same thing with this PR's three edits reversed, so the only difference is where the use count comes from.

## The write this PR removes

`app/models/license.rb` — every activation used to also rewrite the buyer's `audience_members` row. Statement counts per activation:

| | statements | writes |
|---|---|---|
| before | 4 | 2 (`licenses`, `audience_members`) |
| after | 1 | 1 (`licenses`) |

The removed write rebuilt and stored the buyer's **entire** `details` JSON document, so its cost scales with how many purchases that buyer has from the seller, while the `licenses` UPDATE that remains is a fixed-size single-row write:

| purchases on the buyer's row | details size | after (ms) | before (ms) | ratio |
|---|---|---|---|---|
| 1 | 0.2 KB | 0.49 | 0.65 | 1.3x |
| 10 | 1.4 KB | 0.39 | 0.76 | 2.0x |
| 50 | 7.0 KB | 0.40 | 1.16 | 2.9x |
| 200 | 28.0 KB | 0.51 | 2.54 | 4.9x |
| 500 | 70.2 KB | 0.43 | 10.75 | 25.2x |

That last column is the part that matters for the 500s. It is also how much longer the old path held the buyer's `audience_members` row lock, so a seller's best customers — the ones with the most purchases and therefore the biggest row — were exactly the buyers whose concurrent activations queued the longest. After this PR the per-activation cost is flat regardless of row size.

Under 8 threads hammering one buyer's license with no artificial delay, throughput went from 1347 to 2843 activations/sec (2.1x) with zero failures on either path. I would not read that as the real production improvement — Ruby's GVL caps how much genuine write concurrency a single process can generate, so this understates it. The scaling table above is the honest measurement.

One correction worth recording, because it nearly went into this comment as a result: my first version of the lock test held a transaction open for 4 seconds and reported that both paths hit `LockWaitTimeout`. That was the harness, not the code — the artificial hold blocked on the `licenses` row lock, which both paths take. Production activations don't hold transactions open, so the meaningful variable is lock-hold duration, which is what the table measures. Related: an early run of the read benchmark showed `filter_count` timing out past 300s, which I chased as a regression before finding it was contention from my own overlapping benchmark processes. A clean serial re-run gave 1.3-1.4s at 200k members.

## The read path

The filter now joins `licenses` instead of reading `$.license_uses`. Recipient selection, one seller:

| audience size | before | after |
|---|---|---|
| 50,000 | 613 ms | 648 ms (+5.7%) |
| 200,000 | 1743 ms | 1721 ms (-1.3%) |

Within run-to-run noise — the join is on `index_licenses_on_purchase_id` and `EXPLAIN` confirms a `ref` lookup on that index, one row per purchase. Both versions returned identical recipient sets (37,500 and 150,000).

Recipient count, which this PR routes through SQL instead of Elasticsearch when `minimum_license_uses` is present:

| audience size | SQL count (this PR) | with limit 5000 | Elasticsearch |
|---|---|---|---|
| 100,000 | 323 ms | 34 ms | ~1 ms |
| 200,000 | 1335 ms | 51 ms | ~1 ms |

This is the real cost of the change, and it's a deliberate trade: Elasticsearch has no joins, so it cannot read live `licenses` rows, and the count and the recipient list have to come from one source or a seller sees a number that doesn't match who gets the post. The limited form that `Installment#audience_members_count` uses when it passes a `limit` stays fast because the limit applies inside the derived table. The unlimited count is the slow one, and 1.3s at 200,000 members is on a page load, not the API request path that was 500ing.

Worth flagging for review rather than leaving implicit: this moves a sub-millisecond count onto a second-scale query for sellers with very large audiences using this one filter. If that's not an acceptable trade for the correctness guarantee, the alternative is keeping the count in Elasticsearch and accepting that it can disagree with the recipient list.

## Test suite

`audience_member_spec.rb` + `searchable_spec.rb` on the current head: 41 examples, 1 failure — `filter_count counts consistently when cutoff dates carry timezone offsets`. That one fails identically on unmodified `origin/main` in a separate worktree, so it is pre-existing and not from this PR. Flagging it rather than leaving it in a green-looking summary.

Nothing in this benchmark is committed — the scripts were throwaway and the worktree is clean.

---

**AI disclosure:** Generated with claude-opus-5 (Anthropic), operating as the Gumclaw Hermes agent. Prompt: "Run performance benchmarks" on this PR.

PR #6309, opened 2026-07-25
Mirror the seller's price ending into the buyer's currency

gumclaw:
## What

A buyer charged in their own currency currently sees the raw converted amount: a $9.99 product quotes a European buyer something like €8,53. This gives the quoted total **the price ending the seller already chose in USD** — $9.99 quotes €8,99, $10 quotes €9 — with a per-seller opt-out.

## Why this rule

The seller decided what their price should look like. A buyer paying in euros should see that same decision expressed in euros, not a different ending picked by us. It also means there is no curated menu of "good" endings to defend: the ending is whatever the seller picked, so an unusual price like $12.34 mirrors to €10,34 just as literally as $9.99 mirrors to €8,99.

The ending mirrored is the one on the **USD total being converted**, so a cart with tax keeps the total's ending (the .49 of an $11.49 total) rather than the bare product price's.

## Where the rounding happens, and why it matters

**At quote-mint time, not charge time.** The issue's original sketch put rounding "at the processor-charge level after the quote is minted and before the PaymentIntent is created." I moved it earlier deliberately: the buyer decides against the amount the checkout shows them, so rounding after that decision would change the amount charged without changing the amount displayed. That misses the entire point *and* breaks the invariant the buyer-currency feature is built on — the charged amount is exactly the total the buyer confirmed. The rounding now happens once, before the token is signed, so the rounded amount is what is displayed, itemized, locked into the quote token, and charged.

The delta is signed into the quote token so the charge books it without re-deriving it. That also means a seller flipping the setting mid-checkout cannot change the split on an in-flight cart.

## Direction: nearest occurrence, never always-up

The ending recurs every major unit, so there is one occurrence below the converted amount and one above. We take the **nearer** one, with ties going to the buyer.

Measured against **2,159 real `charge_presentments` amounts** (read-only prod, July 2026 — measured on an earlier version of this rule that aimed at a fixed menu of endings, so the grids differ but the asymmetry does not): a nearest rule came out signed mean **−0.13%**, very slightly in the buyer's favour, while always rounding up to the next occurrence was **+2.06%** average. Always-up is a 2% price increase on international buyers wearing a rounding algorithm's clothes. If we ever want that spread, it should be taken explicitly as a pricing decision.

## When we don't round

Mirroring a whole-unit ending can ask for a move of up to half a major unit (49 cents), which is a rounding error on a €40 cart and a fifth of the price on a €2,20 one. So a percentage cap still decides whether the ending is reachable:

| Amount | Cap |
|---|---|
| < €5 | 10% |
| €5–25 | 6% |
| €25+ | 3% |
| zero-decimal (¥, ₩) | 3% |

When the seller's ending is outside the cap, **we do not round at all** — the buyer sees and pays the exact converted amount, which is today's behaviour. Same below one major unit, where rounding down could take the charge under a processor minimum.

**Zero-decimal currencies** (¥, ₩) have no cents, so there is no ¥8,99 to copy the ending into. The ending is mirrored one place up instead: read as a position inside a hundred units, so a $9.99 price quotes **¥1,499** (one below a round hundred — the same shape as one below a round unit) and a $10 price quotes a round **¥1,500**. Small yen carts, where a hundred-unit move exceeds the cap, quote the exact converted amount.

## Seller money is untouched

This is the constraint I was most careful about, per the issue's hard requirements:

- Proceeds, tax, shipping, balances and payouts are all booked from **canonical USD** and come out byte-identical whether or not rounding fired.
- The entire difference lands on **Gumroad's share** of the presentment charge and is recorded in the new `charge_presentments.rounding_delta_cents` (signed) so it is auditable and can be monitored alongside FX drift.
- **The tax lines carry none of it.** Every per-line component is derived from the *exact* converted total, and the difference is then applied on top across the non-tax components only (price, tip, shipping). Tax is computed from canonical USD, gets remitted, and is labelled as tax on the checkout, the receipt and the persisted presentment rows — so a purely cosmetic adjustment must never show up inside it. Same code path at quote time and charge time, so the page, the PaymentIntent and the receipt cannot disagree. (Before this, proportional allocation reported CA$2.15 of tax on a CA$16.49 total where the exact conversion gives CA$2.14.)
- A **round-down is capped** at the percentage fee Gumroad is known to collect on the cart, so it can never dip into the seller's money. The cap is computed at quote time and the orchestrator additionally clamps to `[0, presentment_total]` so a pathological cart can never ask Stripe for a fee above the payment.
- **Fee-waived sales are excluded entirely** (Gumroad Day, per-seller waiver): there has to *be* a Gumroad share to absorb the difference, and on a fee-waived Stripe-Connect charge there may be none. Brazilian Stripe Connect sellers likewise, since `Purchase#calculate_fees` zeroes the fee for them outright.

Rounding also never raises: a `StandardError` in the rounding path is reported and falls back to the exact converted amount. A cosmetic price ending must not be able to break a checkout.

## Creator control

**Opt-out, on by default** for sellers already charging buyers in their own currency — opt-in would give a self-selected minority and no read on whether the feature does anything. Lives in Settings → Payments as "Keep price endings in local currency", directly under the existing buyer-local-currency control, and is hidden unless that seller actually has buyer-currency charging on (the toggle is meaningless otherwise).

The setting's description says the ending mirrored is the **USD total's**, and spells out the taxed case explicitly, because the product-price framing is only true on an untaxed cart — a $9.99 product bought from Quebec quotes CA$16.49 off an $11.49 taxed total, not a .99 price.

Product-page approximate prices are deliberately **not** rounded, with a comment explaining why: that preview is converted from hourly cached rates and the browser derives variant prices from the same `rate`, so rounding only the base price would make the options disagree with it.

## Tests

- `spec/services/checkout/presentment_rounding_spec.rb` — 25 examples: the .99 case, the whole-dollar case, an arbitrary ending (.34), rounding down when the lower occurrence is nearer, already-on-ending, tie-to-the-buyer, out-of-cap fallback, zero-decimal mirroring both ways, the absorption cap (including its floor asserted against `Purchase#calculate_fees` on a tip-heavy cart), and the fee-waived/opt-out gates. **25/25 pass.**
- `spec/services/checkout/buyer_currency_quote_spec.rb` — the quote mirrors a .99 ending and the delta survives the token round trip through `verify!`; a whole-dollar price quotes a whole-unit amount; opt-out quotes the exact converted amount. **30/30 pass.**
- `spec/services/charge/presentment_allocator_spec.rb` — **14 examples, 14/14 pass.** Seven cover the exact-tax rule: a round-up and a round-down each asserting the tax components are byte-identical to the exact-conversion run, a multi-line cart where each line carries its own weighted share of the difference, a reduction one cent past the cart's non-tax money raising, a reduction of exactly that money taking the price to zero with tax untouched, a reduction spent down to the last non-tax cent, and the tax-only cart that raises.
- `spec/services/charge/presentment_orchestrator_spec.rb` — **12/12 pass.** Two examples persist a taxed cart both ways, exact and rounded, and assert the persisted tax row is the same figure in each — this is the exact-tax rule pinned at the layer the receipt reads from, not just in the allocator.
- `spec/services/order/charge_service_spec.rb` — a new order-level example runs with rounding **on**, so a rounded quote is charged end to end: it asserts the locked total and delta, `processor_amount_cents` and Gumroad's reduced share, the unchanged canonical `charge.amount_cents`/`purchase.total_transaction_cents`, and the persisted `rounding_delta_cents`.
- **The rule is revert-proven:** flipping `ROUNDING_ABSORBING_COMPONENT_INDEXES` back to all five components reddens 7 examples across the allocator and the orchestrator, so a future regression cannot pass silently.

Rubocop clean on all touched Ruby files; prettier clean on the TSX.

## Not auto-merging — needs review

1. **Schema migration** (`add_rounding_delta_cents_to_charge_presentments`) — additive, `default: 0, null: false`, but still DDL on a live table.
2. **It changes the amount buyers are charged** and moves money between the buyer's total and Gumroad's share of the charge. Even though seller proceeds are provably unchanged, "payments-core with money-split implications" is not a lane for autonomous merge.

Worth a look specifically at: the percentage caps (they are what decides whether a €2 cart gets its ending mirrored or is left exact), and the absorbable-fee floor — `absorbable_gumroad_cents` intentionally counts only the percentage fee on price+tips, ignoring the fixed fee and processor fees. On that floor: a review read the tips in its base as an overstatement, on the theory that the percentage fee excludes the tip. It doesn't — a tip makes the price customizable rather than adding a line beside it, so the buyer submits one tip-inclusive price and `Purchase#calculate_fees` takes its percentage from that whole amount (verified: a $6 product with a $1 tip charges 70c, not 60c). The spec now pins the cap against `Purchase#calculate_fees` itself on a mostly-tip cart.

## QA steps (hosted preview)

Preview app: `https://feat-buyer-currency-smart-roundi.apps.staging.gumroad.org` — confirm the `x-revision` header matches this PR's head before trusting what you see.

**Click path.** Sign in as `seller@gumroad.com` / `password` (2FA `000000`). Two products are already seeded on the preview for this: **Rounding QA Product** at $9.99 (`/checkout?product=f&quantity=1&price=999`) and **Rounding QA Whole** at $10.00 (`/checkout?product=wz&quantity=1&price=1000`). `buyer_currency_charging`, `buyer_local_currency` and `stripe_payment_element_checkout` are already enabled for that seller.

Note on reaching a non-US quote: the hosted load balancer overwrites `X-Forwarded-For`, so `GeoIp.lookup` resolves every request through the public hostname as `US` and no non-USD quote can be minted from a normal browser. Reviewers reproducing steps 2–5 need either a real non-US egress IP (VPN) or a request addressed straight at the pod, which trusts `X-Forwarded-For`. Step 1 reproduces as-is with no spoofing at all.

1. Go to **Settings → Payments**. The new **"Keep price endings in local currency"** switch appears under the existing local-currency control, on by default. (Confirm it is *absent* for a seller without buyer-currency charging — that's the presenter gate.)
2. Open the $9.99 product's checkout from a Canadian IP. Expect **CA$16.49** — the `.49` ending of the US$11.49 canonical total, not the exact CA$16.38 conversion.
3. Open the $10.00 product's checkout the same way. Expect **CA$16.50** against a US$11.50 canonical total.
4. Expect the same figure on the Stripe Payment Element and on the receipt — displayed, confirmed and charged must all agree, and the tax line must read the exact converted tax rather than a share of the rounded total. **Done — a completed payment is recorded below.**
5. Toggle the setting **off**, reload the $9.99 checkout: the exact converted **CA$16.38** is shown and charged, and the quote's `rounding_delta_cents` drops to `0`.
6. Sanity-check the seller side is unmoved: `canonical_total_cents` is `1149` in both the on and off legs, so seller proceeds, tax and payout figures are identical either way.

Captured on `https://feat-buyer-currency-smart-roundi.apps.staging.gumroad.org`. Each frame
group has its own capture revision: the settings frames at `x-revision: 1702be078ffe`, the
checkout frames at `8ed83fae28c1`, and the completed-payment frames (frame 6) at
`0c2d3ff140f9`. The preview currently serves `a234fceadd11`; head is `11288ff37235`, two
commits further on (a main merge and a migration renumber). Every commit since each capture
has been server-side only (see *Why these frames are still valid at head `11288ff`* below),
so the rendered surfaces are unchanged. Nothing on the branch was
modified to produce these; the geo resolution was obtained by bypassing the hosted load
balancer (which overwrites `X-Forwarded-For`) and addressing the pod directly with a
spoofed Canadian client IP, so eligibility, the real Stripe test FX quote, the rounding
service and the signed quote token all ran unmodified on the deployed branch.

The FX rate is identical across every leg below (`fx_rate: 0.701583`), and the canonical
USD total is identical between the rounded and exact legs — which is what makes the
comparison a proof rather than an assertion.

**1. Settings → Payments, opt-out ON (the default)** — the new *"Keep price endings in
local currency"* switch under the existing local-currency control, with the shipped copy:
the description names the **USD total's** ending and spells out the taxed case.

| Desktop | Mobile (375px) |
|---|---|
| settings on desktop | settings on mobile |

**2. Settings → Payments, seller opted out** — same surface with the switch off.

| Desktop | Mobile (375px) |
|---|---|
| settings off desktop | settings off mobile |

**3. Checkout, $9.99 product, rounding on** — canonical total US$11.49 (product + QC tax)
quotes **CA$16.49**: the `.49` ending of the USD total mirrored into CAD.

| Desktop | Mobile (375px) |
|---|---|
| checkout rounded desktop | checkout rounded mobile |

`/customer_surcharge` quote for that page:

```json
{"currency":"cad","canonical_total_cents":1149,"presentment_total_cents":1649,
 "rounding_delta_cents":11,"fx_rate":"0.701583"}
```

**4. Checkout, $10.00 product, rounding on** — canonical US$11.50 quotes **CA$16.50**, a
whole-cent `.50` ending mirrored the same way.

checkout whole desktop

```json
{"currency":"cad","canonical_total_cents":1150,"presentment_total_cents":1650,
 "rounding_delta_cents":11,"fx_rate":"0.701583"}
```

**5. Checkout, $9.99 product, seller opted OUT** — the same cart, same canonical
US$11.49, same FX rate, now quoting the **exact** converted **CA$16.38** with
`rounding_delta_cents: 0`. This is the control leg: only the seller's opt-out changed
between frames 3 and 5.

| Desktop | Mobile (375px) |
|---|---|
| checkout exact desktop | checkout exact mobile |

```json
{"currency":"cad","canonical_total_cents":1149,"presentment_total_cents":1638,
 "rounding_delta_cents":0,"fx_rate":"0.701583"}
```

**Seller money is untouched, shown rather than asserted:** `canonical_total_cents` is
`1149` in both frame 3 and frame 5. Proceeds, tax, shipping and payouts are all booked
from that number, so they are byte-identical with the setting on and off; the entire
11-cent difference sits in `rounding_delta_cents` on Gumroad's side of the charge.

**6. QA step 4 — a COMPLETED payment, tax line included.** Confirmed through the Payment
Element on the deployed preview at `REVISION: 0c2d3ff140f9` — the exact-tax code commit;
`87e95d2` on top of it only adds these two images, `a234fce` after it only removes
unreachable code and adds specs (verified by live probe below), and `9bc3d7ba7`/`11288ff37`
are a main merge and a migration renumber — buyer geolocated to Quebec.
The earlier note here said reCAPTCHA blocked automated submits; that was wrong. The
preview uses Google's always-pass test key, and what actually blocked the confirm was
Stripe Link's mobile-number field (its country selector defaults to the US flag, so a
Toronto number read as invalid) plus the Payment Element remounting when choosing a
province with its own tax field re-fetched surcharges.

| Checkout (rounded total, exact tax) | Receipt after payment |
|---|---|
| checkout with exact tax | receipt after payment |

Canonical US$11.49 (US$9.99 + US$1.50 QST) at `fx_rate: 0.701891` converts to CA$16.37 and
the seller's `.49` ending pulls it to **CA$16.49**. The checkout shows subtotal CA$14.35 +
tax **CA$2.14** = CA$16.49; the receipt shows the same CA$2.14. CA$2.14 is the exact
conversion of the canonical US$1.50 tax — the 12-cent rounding difference sits entirely on
the price line. Under the previous proportional allocation the same cart reported CA$2.15.

Persisted rows and the processor agree, read back from the preview after the payment:

```
purchase              total_transaction_cents=1149 gumroad_tax_cents=150   (canonical, unmoved)
purchase_presentment  price=1435 tax=214 total=1649 gumroad_amount=523
charge_presentment    total=1649 gumroad_amount=523 rounding_delta=12 fx_rate=0.701891
reconcile             sum(purchase totals)=1649 == charge total=1649
tax_check             gumroad_tax stored=214  exact conversion=214   ✓
stripe                pi_…0bxKrJWH status=succeeded amount=1649 currency=cad
```

The buyer was charged 1649 CAD minor units — exactly the total displayed and confirmed —
while the canonical USD amounts the seller is paid from stayed at 1149.

### Why the settings/checkout frames above are still valid at head `11288ff`

The commits since the checkout frames were captured are all **server-side only** and
produce no pixels of their own:

```
$ git diff --name-only d59076516 11288ff37 -- . ':!qa-media'
app/models/purchase.rb
app/models/user.rb                                   (from the main merge — avatar caching)
app/services/charge/presentment_allocator.rb
app/services/charge/presentment_orchestrator.rb
app/services/checkout/buyer_currency_quote.rb
db/migrate/20261206000011_add_rounding_delta_cents_to_charge_presentments.rb
db/schema.rb
spec/services/charge/presentment_allocator_spec.rb
spec/services/charge/presentment_orchestrator_spec.rb
spec/services/order/charge_service_spec.rb
```

No template, no `.tsx`, no copy string. The exact-tax change (`0c2d3ff`) DOES move a
displayed number — the tax line — which is why frame 6 above was captured fresh at the
current head rather than reasoned about. `c258e6b` changes the *floor* that
`Charge::PresentmentOrchestrator#rounding_absorbable?` measures a round-down against: it was
`fee_cents`, which on a charge through a Gumroad-owned Stripe account also carries the
processor's pass-through percentage and fixed costs. So a fee-waived sale could show a
positive `fee_cents` made up **entirely** of money Gumroad owes Stripe, and a round-down
would have been sized against it. The floor is now the new
`Purchase#gumroad_percentage_fee_cents`, which is zero exactly when Gumroad's percentage fee
is zero.

Rather than re-shoot unchanged pixels, the new method was **probed live on the deployed
preview** at `x-revision: c258e6ba73fd` (read-only Rails runner; the `Purchase` objects are
built in memory and never saved), including the exact case the commit fixes:

```
MARK A normal_sale     price_cents=999 fee_cents=250 flat_per_thousand=100 custom=nil => gumroad_percentage_fee_cents=99 expect=99 PASS
MARK B fee_waived      price_cents=999 fee_cents=250 flat_per_thousand=100 custom=0   => gumroad_percentage_fee_cents=0  expect=0  PASS
MARK C free            price_cents=0   fee_cents=250 flat_per_thousand=100 custom=nil => gumroad_percentage_fee_cents=0  expect=0  PASS
MARK D fee_below_pct   price_cents=999 fee_cents=5   flat_per_thousand=100 custom=nil => gumroad_percentage_fee_cents=5  expect=5  PASS
MARK SUBSET A 99 <= 250 => true
MARK SUBSET B 0  <= 250 => true
MARK SUBSET C 0  <= 250 => true
MARK SUBSET D 5  <=   5 => true
```

- **Leg B is the regression this commit closes.** `fee_cents` is 250 (standing in for a
  Gumroad-managed charge whose fee is entirely processor pass-through) while the percentage
  fee is waived. The old floor would have offered 250 presentment cents of Stripe's money to
  a round-down; the new floor is `0`, so `rounding_absorbable?` refuses and the buyer is
  quoted the exact converted amount instead.
- **Leg D exercises the `min(…, fee_cents)` clamp** — the floor can never exceed the fee
  actually collected.
- **SUBSET** shows the floor is a strict subset of `fee_cents` in every leg: nothing can be
  absorbed that was not collected.

Preview seed state backing the QA steps was verified in the same run at this head:
`buyer_currency_charging`, `buyer_local_currency` and `stripe_payment_element_checkout` all
`true` for `seller@gumroad.com`; `disable_buyer_currency_rounding = false` (rounding on, the
default); *Rounding QA Product* (`permalink=f`, `price_cents=999`) and *Rounding QA Whole*
(`permalink=wz`, `price_cents=1000`) both alive.

**Extra QA step for this commit (no browser needed).** On the preview console:
build a `Purchase` for *Rounding QA Product* with `fee_cents = 250` and
   `custom_fee_per_thousand = 0`, then read `gumroad_percentage_fee_cents`. Expect `0` — a
   waived percentage fee leaves nothing for a round-down to come out of, even though
   `fee_cents` is positive.

### The `a234fce` allocator rewrite, probed live on the preview

`a234fce` ("Pin the exact-tax rule where it reaches the receipt, and drop dead clamp code")
replaced the allocator's displacement loop — which walked leftover reduction from one
component to the next — with a plain proportional split plus an up-front guard. The
justification is that a largest-remainder portion is never larger than its own weight when
the amount being split is no larger than the total weight, so the walk was unreachable code.
Again no pixels, so it was **probed live on the deployed preview** at
`x-revision: a234fceadd11` with a read-only Rails runner
(`Charge::PresentmentAllocator.send(:apply_rounding_delta, …)`; components are
`[price, tip, seller_tax, gumroad_tax, shipping]`):

```
MARK A round_up_single_line     in=[[1435,0,0,214,0]]                    delta=+12 => [[1447,0,0,214,0]]                tax 214→214  sum 1649→1661 ✓
MARK B round_down_single_line   in=[[1435,0,0,214,0]]                    delta=-12 => [[1423,0,0,214,0]]                tax 214→214  sum 1649→1637 ✓
MARK C multi_line_proportional  in=[[1000,200,0,150,100],[500,0,0,75,0]] delta=-17 => [[991,198,0,150,99],[495,0,0,75,0]] tax [150,75]→[150,75]  sum 2025→2008 ✓
MARK D reduction_to_last_cent   in=[[10,0,0,500,0]]                      delta=-10 => [[0,0,0,500,0]]                    tax 500→500  sum 510→500 ✓
MARK E reduction_one_past_money in=[[10,0,0,500,0]]                      delta=-11 => RAISED ArgumentError: reduction of 11 exceeds the cart's 10 cents of non-tax components
MARK F tax_only_cart_increase   in=[[0,0,0,500,0]]                       delta=+5  => RAISED ArgumentError: no non-tax component can carry the 5-cent rounding difference
MARK G tip_and_shipping_only    in=[[0,300,0,120,200]]                   delta=-25 => [[0,285,0,120,190]]                tax 120→120  sum 620→595 ✓

MARK E2E exact   canonical 1149 → exact 1637, delta=0  => [1423,0,0,214,0] line_total=1637 tax=214
MARK E2E rounded canonical 1149 → exact 1637, delta=+12 => [1435,0,0,214,0] line_total=1649 tax=214
MARK CONST absorbing_indexes=[0, 1, 4]
```

The three invariants a reviewer cares about hold in every leg that returns:
**tax is byte-identical before and after** (`tax_untouched=true` in all of A, B, C, D, G),
the components **sum to exactly `input + delta`**, and **no component goes negative**. Legs
E and F are the two guards, raising rather than inventing a placement — and per the new
comment in `PresentmentOrchestrator`, a raise at charge time returns nil and fails the
charge closed before anything is persisted, while at quote time `#create`'s rescue drops the
whole cart back to canonical USD. Leg D is the boundary the removed loop was supposed to
handle: a reduction of exactly the cart's non-tax money takes the price to zero, tax
untouched, with the plain split alone. The **E2E** pair runs the public
`allocate_lines` entry point over the QA-step cart from the frames above and reproduces the
CA$16.37/CA$16.49 pair with the same CA$2.14 tax in both.

### `9bc3d7ba7` (main merge) and `11288ff37` (migration renumber)

The main merge touched none of this feature's files — the one overlap,
`app/models/user.rb`, is main's avatar-variant caching work, several hundred lines away from
this branch's `disable_buyer_currency_rounding` accessor (still present and unchanged at
head). It did, however, bring in a migration that had taken **the same version number**
(`20261206000010`) as this branch's `add_rounding_delta_cents` migration, which made Minitest
die at boot with `ActiveRecord::DuplicateMigrationVersionError`. `11288ff37` renumbers ours
to `20261206000011` and bumps the schema version to match; the migration body is unchanged.

CI: `87e95d2` was all-green (69 checks, 0 failing). At `9bc3d7ba7` the merge introduced the
duplicate-version failure described above (Minitest + Build images + `ci/green`); `11288ff37`
is the fix and is re-running now.

## AI disclosure

Authored by gumclaw (Hermes Agent) running **claude-opus-5**, dispatched from the autonomous dev dispatcher cron and revised in response to review feedback on this PR.

---

Prompts: the driving private tracker issue asked for buyer-currency prices to be rounded to a sensible ending with the algorithm decided first (direction, targets, max difference, zero-decimal handling, creator control), under hard constraints that seller proceeds/tax/payouts never change and the difference goes to Gumroad's side. Review feedback on this PR then specified the rule concretely — mirror the seller's own USD price ending ($10 → €9, $9.99 → €8,99) instead of aiming at a curated grid — which is what the current implementation does. A later review round required that the rounding difference be kept out of the tax lines (build the components from the exact converted total, carry the delta outside tax, at quote and charge time both) and that an order-level example charge a rounded quote end to end.


app/services/checkout/buyer_currency_quote.rb
```diff
@@ -178,19 +182,43 @@ def create
...
+    rounding = if Checkout::PresentmentRounding.enabled_for?(seller)
+      Checkout::PresentmentRounding.round(
+        presentment_total_cents: converted_total_cents,
+        currency: buyer_currency,
+        # A round-down comes out of Gumroad's share of the charge, never the seller's, so
+        # cap it at the presentment value of the fee we know Gumroad collects on this cart.
+        max_downward_cents: presentment_cents_for(
+          Checkout::PresentmentRounding.absorbable_gumroad_cents(
+            seller:,
+            canonical_price_and_tip_cents: line_items.sum { _1.price_cents + _1.tip_cents }
```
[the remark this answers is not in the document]
gumclaw: Checked this against the code and the finding's premise doesn't hold: Gumroad's percentage fee **is** charged on the tip, so including tips in the cap's base is correct rather than an overstatement.

A tip doesn't add a line beside the price — it makes the price "customizable". The buyer submits one tip-inclusive price (`create_service.rb` requires `perceived_price_cents` to already contain `tip_cents`), that becomes `Purchase#price_cents`, and `Purchase#calculate_fees` takes `variable_fee_cents = (price_cents * fee_per_thousand / 1000.0).round` from that same tip-inclusive amount. So a tip-heavy cart has a proportionally *larger* Gumroad share, not a smaller one.

Verified by running the real fee calculation rather than reading it — a $6 product with a $1 tip:

```
price_cents=700 tip=100 fee_cents=170
pct on 700 = 70   <- what Purchase actually charges
pct on 600 = 60   <- what the finding assumed
```

The cap was already right, but nothing in the suite said so, which is what made the reading plausible. Two commits:

- `spec/services/checkout/presentment_rounding_spec.rb` now asserts the cap against `Purchase#calculate_fees` itself on a cart that is mostly tip ($2 product, $8 tip) — `cap <= purchase.fee_cents`. Mutation-checked: inflating the cap 3x reddens it (24 examples, 3 failures), so the guard is load-bearing rather than decorative.
- A comment on `absorbable_gumroad_cents` explaining the tip-inclusive fee base, so the next reader doesn't have to re-derive it.

If the tip ever stops being part of the fee base, that test fails instead of a round-down quietly reaching into the seller's share.

Local run after the change: `presentment_rounding_spec` + `buyer_currency_quote_spec` + `presentment_orchestrator_spec` — 57 examples, 0 failures. Rubocop clean on both touched files.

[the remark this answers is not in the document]
gumclaw: The premise here is inverted, but chasing it down surfaced a real instance of the bug you describe, so thank you for it.

**On tips:** the variable fee is not calculated from a tip-exclusive price. A tip does not sit beside the price — it makes the purchase a customizable-price purchase (`Purchase#customizable_price?` is true whenever a tip is present), so the buyer submits one tip-inclusive `perceived_price_cents`, that becomes `Purchase#price_cents`, and `calculate_fees` takes its percentage from that whole amount. Measured it rather than reading it: a $10 purchase computes `fee_cents` of 209, and the same purchase with a $2 tip computes 235. The fee grows with the tip, so counting price + tip in the cap matches what Gumroad actually collects instead of overstating it. A commit on this branch now documents that and pins it with a spec asserting the cap against the real fee calculation, so the cap fails a test rather than silently reaching into seller money if that ever changes.

**Where you were right:** there is exactly one eligible charge model where the cap claims a share that does not exist, just for a different reason. `calculate_fees` zeroes `fee_cents` outright for Brazilian Stripe Connect accounts, and those accounts pass every buyer-currency eligibility check — confirmed a BR Connect account returns true for `supported_merchant_account?`, `usd_holding_merchant_account?` and `usd_settling_merchant_account?`. So the percentage arithmetic authorised a round-down against a zero Gumroad share, and the orchestrator's `clamp(0, …)` would have pushed the difference onto the seller's proceeds — the exact mechanism you named.

Fixed in 82dc3216a: `absorbable_gumroad_cents` now takes the merchant account and returns 0 for Brazilian Connect, so those sales quote the exact converted amount. Two specs cover it, one asserting `fee_cents` really is 0 for that account alongside the cap, plus a non-Brazilian Connect case to keep the guard from being over-broad. Verified the guard is load-bearing by removing it until the new spec reddened. 55 examples across `presentment_rounding_spec` and `buyer_currency_quote_spec` pass, rubocop clean.

app/services/charge/presentment_orchestrator.rb
```diff
@@ -78,7 +80,15 @@ def perform
...
-    presentment_gumroad_amount_cents = presentment_cents_for(gumroad_amount_cents, locked_quote.fx_rate)
+    # Gumroad absorbs the whole rounding difference: the seller's proceeds are the
+    # converted canonical amount either way, so a total rounded UP adds to Gumroad's
+    # share and a total rounded DOWN comes out of it. The quote already refused any
+    # round-down larger than the fee Gumroad collects, so this stays non-negative; clamp
+    # to the total as well so a pathological cart can never ask Stripe for a fee above
+    # the payment (which Stripe rejects, and which would degrade the charge to USD).
+    presentment_gumroad_amount_cents = (
+      presentment_cents_for(gumroad_amount_cents, locked_quote.fx_rate) + locked_quote.rounding_delta_cents.to_i
+    ).clamp(0, presentment_total_cents)
```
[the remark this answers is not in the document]
gumclaw: Confirmed and fixed in be2ef56cf.

The finding is right, and the mechanism is worth spelling out because it isn't the flag being on — a seller who already has the fee waived never rounds at all (`Checkout::PresentmentRounding.enabled_for?` excludes them). It's the waiver *starting between* the quote and the charge. `waive_gumroad_fee_on_new_sales?` is evaluated live and compares today's date in the seller's timezone against the Gumroad Day key, so it can turn over while one buyer is sitting on the checkout page; the flag can also be switched on in that window. The quote's cap is then sized against a percentage fee that `Purchase#calculate_fees` won't charge, and the negative delta lands on the seller's proceeds.

The fix re-checks the reduction at charge time, where the fee is a computed fact on the purchases rather than a prediction:

```ruby
return unless rounding_absorbable?(rounding_delta_cents)
```

with the floor built from `purchases.sum { _1.fee_cents }` converted at the locked rate. Deliberately the fee and not `gumroad_amount_cents` — that also carries affiliate credit and Gumroad-collected tax, and neither is Gumroad's to give up, so counting them would reproduce a smaller version of the same bug. There's a spec pinning exactly that distinction.

When the fee no longer covers it, the charge fails closed rather than proceeding. Charging the un-rounded amount instead was not an option: the buyer confirmed the rounded total and this feature rests on charging exactly what they saw. So the buyer is asked to review the updated total, and the reloaded checkout mints a fresh quote which won't round down at all, since a waived seller is excluded at quote time. `Charge::CreateService` now surfaces `fallback_reason` so that failure reads as `rounding_delta_exceeds_gumroad_fee` instead of a generic orchestration failure.

Round-ups skip the check entirely — they add to Gumroad's share, so there's nothing to absorb.

Four new examples cover it: reduction still absorbed when the fee covers it, refusal when the fee was waived post-mint, refusal when the round-down clears `gumroad_amount_cents` but not the fee alone, and round-up unaffected by a zero fee. Verified they're load-bearing by disabling the guard — three of them redden. `spec/services/charge/` plus the rounding and quote specs: 103 examples, 0 failures. The comment on `PresentmentRounding.round` now says the cap is a prediction and points at this re-check, so the next reader doesn't trust it as a guarantee.

gumclaw:
Checkout-total evidence added — the half that was pending is now in the body under **## Screenshots (hosted preview app)**, re-pinned to `07b29f9`.

**CA$16.49 rounded vs CA$16.38 exact**, desktop + mobile, same seller / same product / same FX quote rate (`0.701583`), with only `disable_buyer_currency_rounding` flipped between the two legs. The `/customer_surcharge` responses captured alongside each shot carry identical canonical USD (`canonical_total_cents: 1149` = `subtotal: 999` + `tax_cents: 150`), so the seller-proceeds-untouched claim is demonstrated rather than asserted; the signed token shows `presentment_total_cents: 1649, rounding_delta_cents: 11`. This example rounds *up* 11 cents — nearest, not ceiling.

Two notes for whoever reviews:

1. **How the non-US quote was reached.** The hosted LB overwrites `X-Forwarded-For`, so `GeoIp.lookup` sees `US` for every external request and no non-USD quote can be minted through the public hostname; this branch has no TEMP geo/FX spoof scaffolding. Rather than push branch-mutating TEMP commits to a payments-core PR from an automated watcher, the captures were driven through a local proxy that tunnels to the preview pod (which does trust `X-Forwarded-For`) with a spoofed Canadian IP. Everything server-side — eligibility, the real Stripe test FX quote, `Checkout::PresentmentRounding`, the quote token — ran unmodified on the deployed branch. Nothing on the branch was changed to produce this evidence.

2. **Still not covered:** a completed payment, i.e. the Payment Element `amount`/`currency` at confirm time and the receipt. That needs a real purchase through the Element, which reCAPTCHA Enterprise rejects from an automated context. Step 3 of the QA steps closes it in about a minute by hand — the amount to expect is whatever the checkout displays, since quotes re-lock per session.

The seeded QA product on the preview is `Rounding QA Product` (`/l/f`, $9.99 + QC tax), and `seller@gumroad.com`'s opt-out has been restored to its default (rounding ON).

gumclaw:
CI is green at `07b29f9` — every check terminal and passing (Tests workflow: CI Gate, Lint Ruby, Lint JS/TS, Build images, Minitest/RSpec suites; `buildkite/gumroad` build 15848 success; `ci/green` success). Mergeable, `CLEAN`, not a draft, no pushes since the evidence comment above.

Auto-merge is deliberately **not** armed: this carries a schema migration (`charge_presentments.rounding_delta_cents`) and touches the money split on presentment charges, so it waits on your review rather than merging itself. Ready whenever you are, @rmarescu.

slavingia:
If I sell for $10 should round to €9

If I sell for $9.99 should round to €8,99

Capiche?

gumclaw:
Capiche. Rewritten to mirror your ending instead of picking one — pushed in `8ed83fa`.

The rule is now: read the cents off the USD total and aim the buyer's amount at the nearest occurrence of that same ending. $10 → €9, $9.99 → €8,99, and $12.34 → €10,34 with equal literalness. The band/grid table is gone; there is no curated list of "good" endings to argue about anymore, which is a better feature than the one I built.

Two things that needed deciding to make it work:

**Yen and won have no cents**, so there is no ¥8,99 to copy the ending into. I mirror it one place up — the ending is read as a position inside a hundred units — so $9.99 quotes **¥1,499** (one below a round hundred, same shape as one below a round unit) and $10 quotes a round **¥1,500**.

**Sometimes your ending is out of reach.** Mirroring a whole-unit ending can ask for a move of up to half a unit; that's a rounding error on a €40 cart and a fifth of the price on a €2,20 one. So there's still a percentage cap (10% under €5, 6% to €25, 3% above), and when your ending doesn't fit inside it the buyer just sees the exact converted amount rather than a price that moved 20%. Direction stays nearest-occurrence, never always-up: measured over 2,159 real presentment amounts, always-up is a +2% average price rise on international buyers, which isn't a rounding rule.

The money story is unchanged — proceeds, tax and payouts still come from canonical USD, the difference still comes out of our share, round-downs still capped at the fee we actually collect. Specs: 25/25 on the rounding service, 30/30 on the quote, 22/22 on the charge path.

Toggle copy changed with it: "Keep my price endings in local currency."

One thing I did not leave standing: the screenshots in the body were shot against the old grid rule (a CA$16.49 total) and the old toggle label, so both the number and the caption on them are now wrong. I removed them rather than leave stale evidence in front of a reviewer. Fresh captures off the redeployed preview are the one open item — that's the last thing before this is mergeable, and it still needs @rmarescu on the migration and the money split.

gumclaw:
Unparking this — @slavingia's answer at 23:23Z resolved the open question (mirror the seller's price ending: $10 → €9, $9.99 → €8,99), and the rewrite shipped in `8ed83fa` at 23:56Z. Removing `awaiting-human` since nothing is waiting on a decision any more.

Still **not** arming auto-merge and still yours, @rmarescu: this carries a schema migration (`charge_presentments.rounding_delta_cents`) and touches the charge-creation path, both of which are outside what I merge autonomously. CI is green at `8be58ec` (`buildkite/gumroad` + `ci/green` success), merge state CLEAN.

gumclaw:
@rmarescu heads-up that I pushed one commit to this branch since you were handed the PR: `5638d7578`, and it touches **no code** — nothing you may have already read has moved.

It deletes five screenshots under `qa-media/` that this branch committed and that nothing links to. When I re-captured the previews at the price-ending-mirroring head, I added the 999/1000 price-point set the body now uses but left the earlier captures in the tree. That is worth cleaning up rather than ignoring: an evidence file sitting in the PR's file list with no claim attached to it misleads whoever opens it — they read it as something the PR is asserting, when it was superseded hours before.

Removed: `pr-6309-checkout-{exact,rounded}-{desktop,mobile}.png` and `pr-6309-settings-rounding-absent-desktop.png`. Before deleting I confirmed nothing references them — not this body, not any comment here, and no other open PR.

The nine files the body links are untouched and all still resolve. Review state is unchanged; CI re-runs on the new head purely because the head moved.

app/javascript/pages/Settings/Payments/Show.tsx
```diff
@@ -1106,6 +1112,22 @@ export default function PaymentsPage() {
...
+                <Switch
+                  checked={!form.data.disable_buyer_currency_rounding}
+                  onChange={(e) => form.setData("disable_buyer_currency_rounding", !e.target.checked)}
+                  aria-label="Keep my price endings in local currency"
+                  disabled={props.is_form_disabled}
+                  label="Keep my price endings in local currency"
+                />
+                <FieldsetDescription>
+                  Buyers charged in their own currency see your price ending rather than the exact converted amount: a
+                  $9.99 product shows €8.99 instead of €8.53, and a $10 product shows €9. Your earnings, taxes and
```
rmarescu: This example only holds when no tax applies: a taxed cart mirrors the taxed total's ending, so the $9.99 product in your screenshots quotes CA$16.49, not a .99 price. Should this copy mention tax, or should the mirror use the pre-tax price ending instead?

rmarescu (approved):
Before merge, could you run one full payment through the Payment Element on the preview app to confirm the confirmed amount and the receipt match the displayed total?

app/javascript/pages/Settings/Payments/Show.tsx
gumclaw: You're right, and the copy was the thing that was wrong — not the mirror.

The ending mirrored is deliberately the one on the USD total being converted, not the bare product price. `Checkout::PresentmentRounding#target_ending` reads it off `canonical_total_cents`, which is the taxed, tipped, shipped total. So a $9.99 product bought from Quebec quotes CA$16.49 off a US$11.49 total, exactly as you say, and the description promising "a $9.99 product shows €8.99" is only true on an untaxed cart.

I'd rather keep mirroring the taxed total than switch to the pre-tax price ending, for two reasons:

1. The ending only shows up on the number the buyer actually looks at. Mirror the pre-tax price and the .99 lands on a line item nobody reads, then tax pushes the total to some arbitrary ending anyway — the buyer sees CA$16.38 either way and the feature does nothing for them.
2. It keeps a USD buyer and a local-currency buyer seeing the same shape of number. A US buyer in Quebec sees a $11.49 total; a Canadian buyer paying in CAD sees CA$16.49. Mirroring the pre-tax ending would give the local-currency buyer a *different* ending than the USD buyer on the same cart, which is the asymmetry the feature exists to remove.

So I fixed the copy instead, in 1702be0. The description now says the ending is the USD total's and names the taxed case outright:

> Buyers charged in their own currency see the ending of the USD total rather than the exact converted amount: a $9.99 total shows €8.99 instead of €8.53, and a $10 total shows €9. When tax is added, the ending mirrored is the taxed total's, so a buyer paying in their own currency sees the same ending a buyer paying in USD would. Your earnings, taxes and payouts are unchanged — the difference is absorbed on our side.

The label also drops "my" — "Keep price endings in local currency" — since the ending followed is the cart total's, not something the seller set.

I've flagged the two settings screenshots in the PR body as one revision behind rather than re-shooting them: they show the old wording, but the surface, placement and toggle behaviour are unchanged. Happy to recapture if you'd rather the frames match.

gumclaw:
Running the completed payment now. One thing already worth flagging from the setup, before the payment result: on a taxed cart the ending mirrored is the **taxed total's**, not the product price's. A $9.99 product bought from British Columbia quotes a US$11.19 canonical total and CA$16.19 presentment — the `.19` of the USD total, not a `.99`. That is the rule as designed (the service mirrors the ending of the total being converted), but the setting's description promised "your price ending … a $9.99 product shows €8.99", which only holds on an untaxed cart. Pushed a copy fix for that in 1702be0.

Payment run follows.

app/services/charge/presentment_orchestrator.rb
```diff
@@ -110,6 +133,44 @@ def perform
...
+    # needs a waiver to begin during one checkout session.
+    def rounding_absorbable?(rounding_delta_cents)
+      return true unless rounding_delta_cents.negative?
+
+      # Only the Gumroad fee counts. gumroad_amount_cents also carries affiliate credit and
+      # Gumroad-collected tax, and neither is Gumroad's to give up — the affiliate is owed
+      # their cut and the tax gets remitted — so a floor built from the fee alone is the
+      # honest measure of what a round-down can come out of. Converted to presentment cents
+      # because the delta is a presentment-currency amount.
+      absorbable_presentment_cents = presentment_cents_for(purchases.sum { _1.fee_cents.to_i }, locked_quote.fx_rate)
```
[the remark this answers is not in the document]
gumclaw: Confirmed and fixed in c258e6ba7.

The finding is right and the mechanism deserves spelling out, because it is not the same hole as the earlier waiver finding — that one was about the *timing* of the waiver, this one is about what `fee_cents` actually contains.

`fee_cents` is a bundle. On a sale charged through a Gumroad-owned Stripe account, `Purchase#calculate_fees` puts Gumroad's percentage fee, Gumroad's fixed fee, the processor's percentage (`PROCESSOR_FEE_PER_THOUSAND`) and the processor's fixed cost (`PROCESSOR_FIXED_FEE_CENTS`) into the same number. When the waiver turns `gumroad_flat_fee_per_thousand` to zero, the processor components stay, so `fee_cents` is still positive — measured on a $10 managed-account sale with the waiver on, it is 109 cents of pure pass-through. The guard read that as absorption capacity and accepted the stale round-down, which would have reduced Gumroad's application fee below what Stripe is owed. Gumroad would have paid the processor out of pocket instead of the charge failing closed.

The guard now asks a new `Purchase#gumroad_percentage_fee_cents`, which returns only the percentage fee Gumroad itself keeps — zero on a waived sale, a free purchase, or a Brazilian Stripe Connect account, and clamped to `fee_cents` so it can never claim more than the charge actually collects. The fixed Gumroad fee and the discover fee are deliberately excluded even though they are real Gumroad revenue: this needs to be a floor on what is safely disposable, not an accurate total.

A regression example covers exactly the scenario described — a waiver beginning after a managed-account purchase exists — and asserts both halves of the trap: `fee_cents` is greater than zero while `gumroad_percentage_fee_cents` is zero, and the orchestrator refuses with `rounding_delta_exceeds_gumroad_fee` and persists nothing.

`spec/services/charge/presentment_orchestrator_spec.rb`, `presentment_rounding_spec.rb`, `presentment_allocator_spec.rb`, `charge_presentment_spec.rb`, `purchase_presentment_spec.rb` — 53 examples, 0 failures. Rubocop clean on the three touched files.

rmarescu:
(update assignment & labels when you're responsible for a PR)

gumclaw:
Fair — that was mine to fix. I kept pushing to this branch after handing it to you, so the assignment was pointing at the wrong person: you were listed as the assignee with `awaiting-human` on it while I was still the one making changes.

Corrected now: assignee is me, label swapped to `working`. I'll flip it back the other way — assignee you, reviewer you, `awaiting-human` — the moment I'm actually done and it's genuinely waiting on your read, and I won't push code to it after that without saying so in here.

Where it stands at `c258e6ba7`: CI is green (all checks terminal and passing, nothing pending). Two commits landed after you were handed the PR — a copy fix to the setting's description (on a taxed cart the ending mirrored is the taxed total's, not the bare product price's, and the old copy implied otherwise) and a fix sizing the round-down against Gumroad's own percentage fee rather than the processor's.

The one thing still open is the completed-payment capture: reCAPTCHA Enterprise rejects automated checkout submits on the preview host, so the Payment Element confirm amount and the receipt need a real click-through. Everything up to that point — the quote, the signed token, the displayed total — is captured in the body. I'll either get that leg or state plainly that it needs a human click, and then hand it back to you.

Still not arming auto-merge: schema migration plus a change to the money split on presentment charges, so the merge stays yours.

rmarescu:
After a human submits a review, and there is work to be done, you can reassign the PR back to you, and update labels. Once you are ready to assign back to a human, update assignee and labels. Keep it simple. I use assignee to determine if there is anything I need to act upon.

gumclaw:
Understood, and that's a cleaner contract than what I was doing — assignee means "this needs you", so I'll keep it pointed at me whenever I'm the one making changes and only hand it back when there's genuinely nothing left for me to do.

Doing that now, because there isn't. Nothing is in flight on my side at `c258e6ba7`: CI is fully green (77 passing, 9 skipped, nothing pending), every review thread is resolved including your copy-accuracy one, and I won't push to this branch again without saying so here first.

So it's yours: assignee and reviewer you, `awaiting-human` on, `working` off. The only thing left is your call on merging — I'm not arming auto-merge because of the schema migration and the money split on presentment charges.

One caveat that stays open regardless: the completed-payment capture needs a human click-through. reCAPTCHA Enterprise rejects automated submits on the preview host, so the Payment Element confirm amount and the receipt are the one leg I can't produce. Everything up to the confirm — quote, signed token, displayed total, itemization — is captured in the body.

app/services/checkout/buyer_currency_quote.rb
```diff
@@ -178,19 +182,46 @@ def create
...
+            merchant_account:
+          ),
+          quote.fx_rate,
+          buyer_currency
+        )
+      )
+    else
+      Checkout::PresentmentRounding::Result.new(presentment_total_cents: converted_total_cents, delta_cents: 0)
+    end
+    presentment_total_cents = rounding.presentment_total_cents
```
rmarescu, line 210: The rounded total goes through the proportional allocator, so part of the delta gets labeled as tax: the CA$16.49 leg shows CA$2.15 tax where the exact conversion gives CA$2.14, and that figure reaches the checkout, the receipt, and the persisted presentment rows. The delta is ours, not tax. Can you build the components from the exact converted total and carry the delta outside the tax lines, at quote and charge time both?

spec/services/order/charge_service_spec.rb
```diff
@@ -25,6 +25,23 @@ def buyer_currency_quote_for(*products, ip: "24.48.0.1")
...
+    # Puts seller_1 into the state the buyer-presentment charge path needs.
+    #
+    # Rounding is switched off deliberately. The examples that call this cover the
+    # presentment charge plumbing — the idempotency key, Gumroad's share of the charge,
+    # which purchases get a presentment row — not price endings. Buyer-currency price
+    # rounding (Checkout::PresentmentRounding) can move a converted total onto the
+    # seller's price ending, by as much as half a major unit, so leaving it on would make
+    # these examples assert the rounding rule's output instead of the behaviour they are
+    # named for. The rounding rule has its own spec.
+    def configure_seller_1_for_presentment_charges
```
rmarescu, line 37: Every order-level example switches rounding off, so nothing charges a rounded quote end to end. Once the allocation question on the quote is settled, could you add one example with rounding on that asserts the charged amount, the persisted rounding_delta_cents, and the unchanged canonical totals?

rmarescu (changes requested):
Do QA step 4 once these are fixed.

app/services/checkout/buyer_currency_quote.rb
gumclaw: You're right, and the tax figure was the part that mattered — a cosmetic cent should never show up as tax collected on a receipt.

Fixed in 0c2d3ff14. The allocator now takes the exact converted total plus the rounding difference as separate inputs: every component (including both tax components) is derived from the exact conversion, and the difference is then applied on top across the non-tax components only — price, tip, shipping. Same code path at quote time and at charge time, so the checkout page, the PaymentIntent, and the persisted `purchase_presentments` rows all agree. On your example the CA$16.49 leg now shows CA$2.14 of tax, the exact converted figure, with the 12 cents on the price line.

Two edge cases the change had to handle:

- A reduction can't push a component negative. What one component can't give up moves on to the next one with room, so the components always sum to the charged total.
- A cart with nothing but tax behind it has no honest place to put the difference, so the allocator raises. That surfaces as the orchestrator's existing rescue, which falls back to a canonical USD charge rather than persisting rows that disagree with what was charged. It can't happen for a real cart — a round-down is sized against Gumroad's percentage fee, which only exists when there's a price — but I'd rather it fail closed than silently mislabel.

New specs in `presentment_allocator_spec.rb` cover the round-up, the round-down, a multi-line cart (each line carries its own share of the difference, weighted by its non-tax money), the negative-clamp path, and the tax-only raise, each asserting tax is byte-identical to the exact-conversion run.

spec/services/order/charge_service_spec.rb
gumclaw: Added in 0c2d3ff14: "charges the rounded total end to end and leaves the canonical amounts alone" runs with rounding ON, so a rounded quote actually reaches the processor. A $10 cart converts to CA$12.50 at 0.8 and the seller's ending (a round unit) pulls it to CA$12.00, and the example asserts:

- the quote itself locked CA$12.00 with a `rounding_delta_cents` of -50
- `processor_amount_cents` is CA$12.00 and `processor_gumroad_amount_cents` is the converted Gumroad share minus 50 — the reduction comes out of Gumroad's side
- the canonical amounts are unchanged: `charge.amount_cents` and `purchase.total_transaction_cents` both $10.00, and `amount_cents` passed to the charge processor is still 1000
- the persisted `charge_presentment` carries `presentment_total_cents: 1200` and `rounding_delta_cents: -50`, and the purchase presentment's tax components are zero (this cart is untaxed, so nothing of the reduction landed on tax)
- the checkout response reports CA$12.00 to the buyer

The other order-level examples still opt out of rounding on purpose — they cover the charge plumbing, and leaving rounding on there would make them assert the rounding rule's output instead of what they're named for.

Both findings are in; QA step 4 is next and I'll post the results here.

gumclaw:
## Pre-merge adversarial review (local fable-5)

Ran a hostile static review over the two fix commits (`c258e6ba7..87e95d2fe`) with a hunt list aimed at the money paths: sum invariants, the quote/charge symmetry, the token round trip, the raise paths, downstream readers of the presentment rows, and whether each new spec actually bites. Verdict: **CHANGES-REQUIRED** — no money-movement defects, but two of my new specs didn't test what they claimed and two comments described behaviour the code doesn't have. All fixed in a234fcead.

| Finding | Verdict | What I did |
|---|---|---|
| **P2-1** The end-to-end example has no tax, so it passes identically under the old proportional allocation — it can't catch a regression of the rule it exists for | Real | Added an orchestrator example on a taxed cart that persists both legs (exact and rounded) and asserts the tax row is the same figure in each. Proportional allocation reports 189 against the exact 187, so it reddens on revert. |
| **P2-2** The clamp/displacement pass for an oversized reduction is unreachable, and the spec named for it never enters it (its fixture has no tip despite the comment) | Real | Verified the unreachability exhaustively — 922,868 weight/delta combinations, zero cases where a largest-remainder portion exceeds its own weight. Replaced the machinery with an up-front budget check, and the spec with the two reachable cases: one cent past the budget raises; exactly the budget takes the price to zero with tax untouched. |
| **P3-1** Allocator comment says a raise makes the caller charge canonical USD — only true at quote time; at charge time `Charge::CreateService` fails closed | Real | Both callers now document which one they do, and that nothing is persisted either way (the raise precedes the transactional `persist!`). |
| **P3-3** "in that order of preference" misdescribes the algorithm — the difference is spread proportionally across price/tip/shipping at once | Real | Comment corrected. |
| **P3-2** The all-tax unreachability argument only covers round-downs (a round-up is capped by percentage, not by the fee) | Real, narrower | Reworded to the reason that covers both directions: tax is computed from the price, so a cart with no price has no tax. |
| **P3-4** `presentment_total_cents: shares.sum` could mask a components-don't-sum defect if a purchase's derived price went negative | Not actionable | The reviewer couldn't construct a real purchase where it happens and neither could I — `[price_cents, 0].max` only fires on data that would already be broken upstream. Noted, not guarded; adding an assertion here would be defending against a state no code path produces. |
| **P3-5** `MethodForcedPresentment` never rounds, so defaulting the new kwarg to 0 is right — but a buyer who saw a rounded card quote and then pays via iDEAL gets an un-rounded intent | Pre-existing, out of scope | True and worth fixing, but it's the shipped feature's behaviour rather than this delta's. Not folding an unrelated surface into a review-fix commit. |

Things the review checked and found nothing on, for the record: sum invariants across single/multi-line/zero-weight (gift giftee) carts, the `locked - delta` reconstruction in both signs, old tokens minted before the delta key existed (they degrade to 0 and reproduce pre-fix behaviour exactly), `gumroad_amount_shares_within` still capping against post-delta totals, and `Purchase::PresentmentRefund` — it allocates against the persisted components with a matching denominator, so no proportional reconstruction disagrees, and the VAT-only refund path now returns the exact converted tax.

Local: allocator 13/13, orchestrator 11/11, rounding 25/25; rubocop clean. Reverting `ROUNDING_ABSORBING_COMPONENT_INDEXES` to all five components reddens 7 examples, so the rule is pinned at both the allocator and the persistence layer now. CI running on a234fcead.

gumclaw:
## Red shard was not this PR — merged main to pick up the fix

`Test Fast 3` failed at head `a234fcead` on a single example, and it is not related to the smart-rounding work:

```
Purchase::Reportable#price_cents_for_tax_reporting
  when the purchase is chargedback and not reversed
  returns 0 (chargeback attribution is unchanged by the refund cutover)
    expected: 0
         got: 100
  ./spec/models/concerns/purchase/reportable_spec.rb:167
```

This is a **time-bomb spec**, not a flake and not a regression. The example dated its chargeback off `Time.current`, so it only passed while "now" happened to fall before `CHARGEBACK_REPORTING_CUTOVER`. That date arrived, and the example started failing on its own on every branch that does not yet have the fix — nothing in this PR touches `Purchase::Reportable`, tax reporting, or chargebacks.

It was fixed on `main` by #6374 (`4a4e0ba26`, landed 00:59Z) which pins the chargeback date explicitly instead of deriving it from the clock. That commit was ~1 minute old when my run started, so this branch missed it.

Merged `main` in as `9bc3d7ba7` — no behaviour change to this PR, no conflicts. CI is re-running.

Worth flagging for anyone else hitting this: **every branch cut before 00:59Z today will fail this same example** until it picks up `main`. It is not your diff.

gumclaw:
One more thing that came out of the main merge, worth knowing before this merges: main's #6359 took migration version `20261206000010`, which this branch was already using for `add_rounding_delta_cents_to_charge_presentments`. Two migrations sharing a version raises `DuplicateMigrationVersionError` at `maintain_test_schema!`, so it doesn't fail one spec — the entire suite fails to load, every shard red with zero examples run, which reads like something much worse than it is.

Renumbered mine to `…000011` in 11288ff37 (it hasn't shipped, so renumbering is safe) and bumped `schema.rb`'s version to match. The schema diff against main is now exactly the one column and the version line, nothing else:

```
-ActiveRecord::Schema[7.1].define(version: 2026_12_06_000010) do
+ActiveRecord::Schema[7.1].define(version: 2026_12_06_000011) do
+    t.bigint "rounding_delta_cents", default: 0, null: false
```

Same heads-up as the chargeback spec above: any other branch still sitting on `…000010` will hit this the moment it merges main.

gumclaw:
Both findings are addressed, QA step 4 is done, and CI is green at `11288ff37`.

**Tax allocation.** Components are built from the exact converted total and the rounding difference is carried on the non-tax components only — price, tip, shipping — at quote time and charge time through the same code, so the checkout page, the PaymentIntent and the persisted rows can't disagree. Your CA$16.49 case now reports CA$2.14.

**QA step 4 — a completed payment.** I got one through on the preview, which the body previously claimed was blocked by reCAPTCHA. That was wrong: the preview uses Google's always-pass test key. What actually blocked it was Stripe Link's mobile-number field (its country selector defaults to the US flag, so a Toronto number read as invalid) and the Payment Element remounting when picking a province with its own tax field re-fetched surcharges. Both harness problems, not product ones.

The payment, Quebec-geolocated buyer:

```
checkout   subtotal CA$14.35 + tax CA$2.14 = CA$16.49
receipt    Sales tax CA$2.14
stripe     pi_…0bxKrJWH  status=succeeded  amount=1649  currency=cad
purchase              total_transaction_cents=1149  gumroad_tax_cents=150   (canonical, unmoved)
purchase_presentment  price=1435  tax=214  total=1649
charge_presentment    total=1649  rounding_delta=12  fx_rate=0.701891
reconcile             sum(purchase totals)=1649 == charge total=1649
tax_check             stored=214  exact conversion=214
```

Screenshots of both are in the body under QA step 4.

**One thing I'd rather flag than bury.** The order-level example you asked for charges an untaxed cart, so with tax at zero it passes identically whether the difference is spread over all five components or only the non-tax ones — it satisfies your three asks but couldn't have caught a regression of the tax rule itself. A local adversarial review caught that, plus a clamp branch of mine that turned out to be unreachable (verified exhaustively over 922,868 weight/delta combinations) and two comments describing behaviour the code doesn't have. Fixed in a234fcead; the review and its triage are in the comment above. The rule is now revert-proven: flipping the absorbing-components constant back reddens 7 examples across the allocator and the orchestrator.

Two housekeeping commits on top, both explained in comments above: a main merge for the chargeback-cutover spec fix (#6374), and a migration renumber after main took the same version number.

Assigning back to you — nothing outstanding on my side.

rmarescu (approved):
@gumclaw looks good!

PR #6318, opened 2026-07-25
Harden store agent failure handling

gianfrancopiana:
Ref #6296

## What

Prevents confirmed store-agent API errors from copying seller-controlled validation text into structured logs. Logs now use fixed failure categories and numeric status metadata while web and mobile keep their existing response shape.

Also rejects incomplete buffered Anthropic fallback responses so damaged successful JSON cannot become a blank reply or an empty tool call.

## Why

#6296 added fallback recovery and failure telemetry, but its merged implementation still accepted structurally invalid 200 responses and logged reflected API messages. This keeps the useful telemetry while closing both failure modes at their boundaries.

## Before / After

This is a backend safety change with no new UI. Before, a v2 validation message could reflect a seller callback URL into long-lived logs, and a malformed buffered replay could become a blank successful reply or an empty tool call. After, the seller still receives the full validation message while logs receive fixed metadata, and a replay must have the complete response structure the agent consumes before it can replace the original error.

## QA steps

1. Confirm an agent action that returns a reflected v2 validation error. Verify the client still receives the full message, while the request log contains only the fixed failure category, numeric status, and catalog endpoint.
2. Force three unreadable streamed tool calls, then return an empty or malformed 200 response from the buffered replay. Verify the original unreadable-tool-call error reaches the caller.
3. Repeat with a valid buffered tool call and with a `max_tokens` response. Verify both existing recovery paths still work.

## Test Results

The focused suite passed after the final rebase: 115 examples, 0 failures. All nine changed Ruby files pass RuboCop. Test-confidence reached 99% with no branch-only failure, and final autoreview reported no findings.

---

This PR was implemented with AI assistance using gpt‑5.6 xhigh.


gianfrancopiana (commented):
Codex self-review: Traced both boundaries end to end. Reflected v2 validation text stays in the seller response, while structured logs receive only fixed failure categories, the real numeric status, and a catalog-resolved endpoint. Both web and mobile strip the internal telemetry from their responses. The buffered replay now rejects empty valid JSON and non-object tool input while preserving the existing `max_tokens` path.

Local verification: 115 focused examples pass; all nine changed Ruby files pass RuboCop; test-confidence reached 99% with its only selected failure reproduced on the merge base; final autoreview reported no findings.

PR #6321, opened 2026-07-26
Surface the license key and a lookup link on the product page

gumclaw:
## What

Surfaces a buyer's license key on the public product page, and gives the existing `/license-key-lookup` page its first inbound link.

- The "you already own this product" card now renders a **License key** row with a copy button, so a returning buyer no longer has to click through to the content page to find their key.
- A licensed product with no detected purchase now renders an **"Already bought this? / View your information"** card linking `/license-key-lookup`, which emails the visitor their receipt (including the key) after they enter their purchase email or PayPal invoice ID.

Access control was the design constraint here. `Link#purchase_info_for_product_page` resolves a prior purchase from three signals, and only two of them identify a *person*:

| Match signal | Ownership card | License key |
|---|---|---|
| Signed-in purchaser (`requested_user.purchases`) | yes | **yes** |
| HMAC'd `purchase_id` + `purchase_email_digest` from the buyer's own email | yes | **yes** |
| `_gumroad_guid` browser cookie only (guest checkout) | yes | **no — stripped** |

`Purchase.purchase_info` already contained `license_key`; it simply was not serialized into the page props. The change therefore both serializes it *and* strips it on the cookie-only path, because a cookie identifies a browser, not a person — on a shared or public machine the previous buyer's key would otherwise be visible to whoever sits down next. The per-variant gate is inherited for free: `Purchase#license_key` returns nil unless `link.is_licensed?` **and** the purchased variant's content embeds a license-key block, so free variants alongside paid licensed ones show nothing.

`is_licensed` is also added to the product props so the frontend can decide whether to offer the lookup card. It is hidden from the seller's own view of their product (`can_edit`), since it is buyer-facing help.

## Why

Sellers of licensed products carry recurring support load from buyers who bought months or years ago and return to the public product page looking for their key. Today the page recognizes them and still dead-ends at a "View content" button; a not-signed-in visitor gets nothing at all, and the self-serve lookup page that would have solved it had zero inbound links anywhere in the app (grepping `app/javascript/` for `license_key_lookup` turned up only the generated route helpers). The related help-center article URL 404s, so the page was effectively reachable only if support handed out the URL.

Fixes antiwork/gumroad-private#1354

## Before / After

Rendered by a throwaway Capybara system spec against the real running app. The **before** column was captured on unmodified `origin/main` in a separate worktree, so the pair is a true A/B of the same scenario. Desktop viewport (the headless Chrome window would not resize below its minimum, so these are desktop-width only; the layout uses the existing `CardContent` flex row that the adjacent membership/rating rows already use at mobile widths).

**Returning buyer who owns a licensed product** — the ownership card gains a License key row with a Copy button.

| Before (`origin/main`) | After (this branch) |
|---|---|
| before | after |

**Unrecognized visitor on a licensed product** — the "Already bought this?" card appears, linking `/license-key-lookup`.

| Before (`origin/main`) | After (this branch) |
|---|---|
| before | after |

The guest/cookie-only path has no screenshot because the whole point is that nothing renders: the ownership card looks exactly like the "before" image above, with no key. That path is covered by assertion instead (`test/models/link_test.rb`, `product_props_spec.rb`), which is the stronger check here since the evidence is an absence.

## QA steps

1. Open a licensed product's page while **not** signed in and with no prior purchase in that browser → an "Already bought this?" card appears under the price row with a "View your information" button linking `/license-key-lookup`.
2. Follow that link → the existing "What is my license key?" lookup page loads; entering a purchase email emails the receipt with the key.
3. Sign in as an account that has purchased that licensed product → the "You've purchased this product" card now includes a **License key** row showing the key, with a working Copy button. The "Already bought this?" card is not shown (a purchase was detected).
4. Buy the licensed product as a **guest** (no account), then revisit the product page in the same browser → the ownership card still appears, and the License key row is **absent**. This is the credential-protection path.
5. Open a non-licensed product's page while signed out → no "Already bought this?" card.
6. View your own licensed product as its seller → no "Already bought this?" card.

## Test Results

New coverage:

- `test/models/link_test.rb` — three cases on `#purchase_info_for_product_page`: key present for a signed-in purchaser, key present for a matching email digest, key **absent** when only the browser guid matches.
- `spec/presenters/product_presenter/product_props_spec.rb` — `is_licensed` exposed; `purchase[:license_key]` present for a signed-in buyer and nil on the cookie-only path.
- `spec/requests/products/show/license_key_spec.rb` (new system spec) — the inline key row + copy button for a signed-in owner, the lookup link for an unrecognized visitor, and no lookup card on an unlicensed product.

Local runs:

test results

```
test/models/link_test.rb -n "/purchase_info_for_product_page/"   19 runs, 22 assertions, 0 failures
spec/requests/products/show/license_key_spec.rb                   3 examples, 0 failures
product_props_spec.rb -e "licensed product"                       3 examples, 0 failures
```

The full `product_props_spec.rb` file has 5 failures on this branch that also fail on unmodified `origin/main` in a fresh worktree (Stripe-charging factories and Elasticsearch-dependent sales-count examples) — verified by running the same file at `origin/main`: same 5 examples, same failures. Not introduced here.

Lint: `rubocop -a` clean on all touched Ruby files; `prettier --write` + `eslint` clean on `Product/index.tsx`.

---

## AI disclosure

Written by **claude-opus-5** (Hermes Agent), from the investigation recorded on antiwork/gumroad-private#1354 and Sahil's approval of the proposed shape on that issue. Instructions given, in substance: surface the license key inline on the product page for a buyer who already owns the product; add an "already bought this?" affordance linking the orphaned `/license-key-lookup` page for unrecognized visitors; and apply the access-control caveat raised in the issue so that a purchase matched only by the `_gumroad_guid` cookie does not reveal the key. Code, tests, and this description were produced by the model; the merge decision is left to a human reviewer.


gumclaw:

Ran an adversarial pass over the diff, focused on the access-control path since that is the actual risk surface here. Three findings, all fixed in `df881cd`:

**P1 — the key row would have rendered empty for guests.** The frontend guarded the row with `purchase.license_key !== null`, but the backend strips the entry with `info.except(:license_key)`, so it arrives as `undefined`, not `null`. `undefined !== null` is true, so the row would have rendered with a blank value on exactly the path the strip exists to protect. Fixed by typing the prop optional and guarding on truthiness; the TypeScript checker confirmed the mismatch when the type was corrected. Added a presenter spec asserting the serial does not appear anywhere in the serialized purchase props on the cookie-only path, so the guarantee is now checked against the wire format rather than the Ruby hash.

**P2 — comments pointed at a method that does not exist.** Both the Ruby and TypeScript comments referenced `Link#purchase_info_with_license_key`, a name from an earlier draft of this change that was replaced by the strip-based approach. A reader would have grepped for it and found nothing. Repointed at the real gate, `Link#purchase_info_for_product_page`.

**P3 — a test that passed for the wrong reason.** I had added a system spec for the guest/cookie-only path, but injecting `_gumroad_guid` from Capybara lands the cookie on the wrong host, so no purchase was matched and the example was really asserting "no purchase detected" rather than "purchase detected, key withheld". Removed it with a comment explaining why, and left the path covered at the model and presenter layers where the assertion is exact. Better no test than a green one that proves nothing.

Also verified: no other consumer reads `purchase.license_key` from the product props, so the strip cannot break an existing surface; the per-variant gate is inherited from `Purchase#license_key` rather than reimplemented; and the `matched_signed_in_purchaser` flag is captured before the `browser_guid` fallback reassigns the relation, which is the ordering the whole gate depends on.

Nothing outstanding. Test evidence and before/after captures are in the body.

PR #6324, opened 2026-07-26
Describe a malformed product-editor save instead of telling the seller to refresh

gumclaw:
## What

Adds a payload-shape validation step to `LinksController#update` that runs **before** the content deletion guards, so a structurally broken product-editor save gets its own specific error instead of the guards' "the content shown in the editor may be out of date, please refresh the page."

Also removes the reason several pre-mutation checks were reading the raw request params instead of the permitted ones.

## Why

The editor's save path had one answer for every payload it could not interpret. The deletion guards (`Product::RichContentDeletionGuard`, `Product::VariantCategoryUpdaterService.ensure_deletion_intent!`) exist to answer a single question — *would this save remove content the seller didn't ask to remove?* — so when a payload was structurally broken, the guards read that as "the payload doesn't know about these pages" and told the seller to refresh.

Refreshing does not fix a malformed request. The seller retries into the same message, and in support this is indistinguishable from a staleness bug, which is where the July save-block tickets kept landing.

Strong parameters is what makes the two states look alike: it silently **drops** a collection whose shape doesn't match the policy. By the time a malformed array reaches the permitted params it is simply absent — the same thing the params look like when the client never sent it at all. That is why the shape check reads the raw params; it's the only place the bad shape is still visible.

### The permitted-params snapshot

Several save steps mutate `product_permitted_params` in place (the coffee path rewrites `:variants`; `SaveFilesService` consumes page descriptions). The old workaround was for pre-mutation checks to read the raw `params` instead — which skips the policy's filtering entirely, so those checks were reading a wider, different object than the save itself acted on. They now share one immutable snapshot taken before any step runs.

### What this deliberately does NOT check

Whether the ids in the payload name records of *this* product. That sounds like a shape question but it isn't answerable yet: a page the seller just created is submitted under a client-generated id the server has never seen, and the save response hands back `rich_content_id_mappings` so the editor can adopt the canonical one afterwards. An unrecognised id is therefore indistinguishable from a genuinely foreign one until client ids are namespaced per owner and minted uniquely per copy — that is the identity-contract work in gumroad-private#1360.

I built the id checks first and they rejected five legitimate flows (first save of a new page, the id-mapping reconciliation round-trip, the cross-entity placeholder page in the July 21 recovery state). A regression test now pins the client-generated-id shape as **accepted**, so re-adding the check prematurely reddens.

The guards themselves are unchanged and still run pre-mutation. This only adds a step in front of them.

## Before / After

| Payload | Before | After |
|---|---|---|
| `rich_content` sent as a scalar | 422 "the content shown in the editor may be out of date — please refresh the page" | 422 `invalid_payload_shape` / "its content pages weren't sent as a list" |
| `variants` sent as a scalar | same refresh message | 422 `invalid_payload_shape` / "its versions weren't sent as a list" |
| a variant entry that isn't a hash | same refresh message | 422 `invalid_payload_shape`, reason `variant_not_a_hash` |
| Sentry visibility | none — read as a normal guard rejection | `Rejected product save with a malformed payload` + reason code |
| a genuine stale snapshot | refresh message | unchanged |
| a genuine unconfirmed deletion | deletion-intent message | unchanged |
| new pages/versions under client-generated ids | accepted | accepted (pinned by a regression test) |

The error carries a stable `error_code`, so the editor can distinguish a malformed-payload rejection from a staleness or deletion-intent one rather than showing the same generic banner for all three.

## Test Results

`test/controllers/links_controller_test.rb` — **361 runs, 1310 assertions, 6 failures** on this branch, and the identical 6 on a detached `origin/main` worktree at `ee2894524` (order-dependent pollution: 3× 409-conflict, 2× 404, 1× cancellation-discount stub). Nothing in this diff moves them.

New coverage, all verified **load-bearing**: dropping the `SavePayloadShapeValidator.validate!` call from the save path reddens 3 of them (`3 runs, 3 failures`) — the mutation run is included in the evidence image below.

shape validation tests

Permanent copy: `qa-media/` at `956358dc44aad1b2573567e5d253105d5c837567`.

Rubocop clean on all three touched files.

## QA steps

No rendered surface changes — this alters the error body a save returns, not any view. The editor shows whatever `error_message` comes back, so the visible difference is the sentence in the existing banner.

Preview app: `https://<branch-slug>.apps.staging.gumroad.org` (confirm `x-revision` matches this PR's head).

1. Open any product in the editor, then from the browser console submit a save with a malformed collection:
   ```js
   await fetch(location.pathname.replace("/edit", ""), {
     method: "PUT",
     headers: { "Content-Type": "application/json", "X-CSRF-Token": document.querySelector("meta[name=csrf-token]").content },
     body: JSON.stringify({ link: {}, rich_content: "not-a-list" }),
   }).then(r => r.json())
   ```
   Expect `error_code: "invalid_payload_shape"` and a message about content pages not being sent as a list — **not** the "may be out of date, refresh the page" text.
2. Repeat with `variants: "not-a-list"` and expect the versions equivalent.
3. Confirm the product's variants and pages are untouched afterwards (nothing mutates before this check).
4. Save the product normally through the UI and confirm it still succeeds.
5. Create a new content page, type into it, save — confirm it succeeds (this is the client-generated-id path the validator must not reject) and that the response includes a `rich_content_id_mappings` entry for it.
6. Confirm a genuine unconfirmed deletion still returns the deletion-guard message, unchanged.

## Notes

Fixes the first of the two items @slavingia approved on gumroad-private#1363 ("Sounds good. Open 2 unique PRs"). The permalink-copy item is a separate PR.

---

**AI disclosure:** Written by Gumclaw (claude-opus-5). Instructions given: build the two items approved on gumroad-private#1363 as separate PRs. An earlier session left this work uncommitted in a worktree; I reviewed it, found the id-identity checks rejected five legitimate save flows, cut them with the reasoning recorded above, added a regression test pinning the client-id shape as accepted, and verified the remaining checks are load-bearing by mutation.


gumclaw:
@gianfrancopiana this is yours — same file and same lane as #6311, which you're already carrying.

Not arming auto-merge. It's a bug fix with no schema and no money path, but it adds a rejection to the product-editor save path that has produced three data-loss incidents, and the judgement call below deserves a second read rather than a squash.

**CI green at `78531963a`** (`buildkite/gumroad` success). No rendered surface changes — it changes the error body a save returns, so the QA steps are console-driven against the preview.

**The one thing worth your eyes.** The first cut of this validator also rejected payload ids that don't name records of this product (duplicates, and ids from another product). I cut that, because it rejected five legitimate flows: the first save of a newly created page arrives under a client-generated id the server has never seen, and that's the supported path — `rich_content_id_mappings` in the response is how the editor learns the canonical id afterwards. The id-mapping reconciliation round-trip and the cross-entity placeholder page in the July 21 recovery state broke the same way.

So an unrecognised id is genuinely undecidable until client ids are namespaced per owner and minted per copy, which is gumroad-private#1360. This PR validates only what's answerable now, and there's a regression test pinning the client-generated-id shape as **accepted** so the check can't be re-added prematurely without going red. If you'd rather that check land here with the mappings consulted, say so and I'll build it — I judged it belongs on #1360's PR.

Second thing: several pre-mutation checks were reading the raw `params` instead of the permitted ones, because later save steps mutate the permitted params in place. They now share one immutable snapshot taken before any step runs, so those checks and the save itself finally read the same filtered object. That's the `# TODO clean this up` and the "reads the RAW params" comments gone.

**Baseline:** the touched spec file runs 6 failures on this branch and the identical 6 on a detached `origin/main` worktree at `ee2894524` (order-dependent pollution — 3× 409, 2× 404, 1× a cancellation-discount stub). Nothing here moves them.

slavingia:
Doesn't make sense imo to expose this to creators - just fix the root cause as these errors aren't really helpful / actionable either.

PR #6328, opened 2026-07-26
Redirect to login instead of 404ing when /two-factor has no pending 2FA session

gumclaw:
## What

`TwoFactorAuthenticationController` handled a missing pending-2FA session two different ways. `verify` redirected to the login page; `show`, `create`, `resend_authentication_token` and the three `switch_to_*` actions went through `check_presence_of_user`, which called `e404` and raised `ActionController::RoutingError`. This applies the redirect to every action, so a missing session always sends the visitor back to the login form with a "your login session expired" notice, and drops the now-unused `check_presence_of_user`.

The requested path is preserved in `next` for GET requests so the visitor lands back where they were headed after signing in. POST paths are not carried over — a form endpoint is not somewhere to send a browser after login.

## Why

`@user` comes from `session[:verify_two_factor_auth_for]`. Anyone who reaches `/two-factor` without that key present got a hard 404: expired session, dropped cookie, back button, a bookmarked or re-opened `/two-factor` URL, a second tab, an abandoned login. Reproducible on production today — `GET https://gumroad.com/two-factor` returns 404, with or without a `next` param.

That is the worst possible place for a dead end. Someone whose login has just gone wrong is sent to the two-factor page as the recovery step, and instead of a form they get an error page with no way forward. A seller hit exactly this last week: passkey sign-in failed, she was sent to `/two-factor`, and the 404 there left her unable to reach her account at all until support pointed her at the Google sign-in button.

This is not the same as #2913 (users *without* 2FA being sent to `/two-factor`, fixed in #2918). Here two-factor really is enabled; the defect is that a missing session renders an error page instead of a redirect.

Tracking issue: antiwork/gumroad-private#1369

## Before / After

`GET /two-factor` with no pending 2FA session in the session:

| Before | After |
|---|---|
| before | after |
| `ActionController::RoutingError` raised from `check_presence_of_user` — 404 in production, no way forward | Login form with "Your login session expired. Please sign in again." |

The warning renders as both the inline alert and a toast because that is how every login-page warning already renders (`AuthAlert` for the inline alert, `layouts/shared/_flash` for the toast) — the same double render you get today on a wrong-password attempt. Unchanged by this PR.

## QA steps

1. Open `/two-factor` in a browser with no login in progress (a fresh incognito window is enough).
2. Expect the login page with the "Your login session expired. Please sign in again." warning, not an error page.
3. Log in with an account that has email 2FA enabled, reach the two-factor screen, then clear cookies and reload `/two-factor` — same redirect.
4. From the two-factor screen, hit "Resend token" after the session has been cleared — redirect to login rather than a 404.
5. Complete a normal 2FA login end to end and confirm nothing about the happy path changed.

## Test Results

`spec/controllers/two_factor_authentication_controller_spec.rb` and `spec/requests/two_factor_authentication_request_spec.rb`, run locally on this branch, plus the mutation check proving the new coverage is load-bearing:

specs

The shared `check user in session for html request` example — used by `show`, `create`, `resend_authentication_token` and each `switch_to_*` describe block — now asserts the redirect, the 303, and the flash instead of the raised `RoutingError`, so six of the seven actions are covered there and `verify` keeps its own stale-session spec. Each call site passes the exact URL it expects rather than deriving it from `request.fullpath`, so the expectation can't quietly agree with whatever the controller built. Four request specs cover `GET /two-factor` with and without `next` and the two POST paths that previously raised the 404.

Swapping the controller back to main's version reddens exactly those four request examples and nothing else.

---

Written by claude-opus-5 (Hermes Agent). Instructions given: open a PR for antiwork/gumroad-private#1369, which reports that `GET /two-factor` returns a hard 404 when there is no pending two-factor session and asks for the same login redirect `verify` already does, extended to the `switch_to_*` and resend actions that share the same before_action. Model wrote the controller change, updated the existing specs, added request-spec coverage for the GET and POST paths, captured the before/after screenshots from a local system-spec run, ran a local adversarial pre-merge review and applied its two spec-quality findings.


gumclaw:

Ran a local read-only adversarial review against the branch (a ten-item checklist scoped to this change class: redirect-status changes, before_action ordering vs the Rack::Attack throttle, Inertia POST-to-GET behavior, open-redirect safety of `next`, completeness over all seven actions, non-HTML consumers, spec load-bearingness, and information leakage). Verdict: **ready to merge**, no P1 or P2. Two P3s, both fixed in f896ec2.

|---|---|---|
| Shared stale-session example derived its expected `next` from `request.fullpath` — the same source the controller builds it from, so it could agree with the controller regardless of output | Real, self-fulfilling assertion | Fixed — each call site now passes the exact URL it expects, and the example also asserts 303 |
| Request-level coverage only exercised `GET /two-factor`, while the behavior change that matters is the POSTs going from 404 to a redirect | Real coverage gap | Fixed — added request specs for token submission (with Inertia headers) and resend |
| Adding `status: :see_other` to `verify`'s existing redirect (was a bare 302) | Verified safe | `verify` is GET (routes.rb:744); browsers, bots, and the Inertia client all follow it as a GET, and no spec or JS asserts 302 for it |
| Redirect instead of 404 weakening the token brute-force throttle | Verified safe | Rack::Attack runs before controller code and still keys on `params["user_id"]` (rack_attack.rb:249); the mismatched-`user_id` path still 404s via `validate_user_id_from_params`, so the "prioritises user_id in the query string" spec is unchanged |
| Open redirect via `next` | Verified safe | `request.fullpath` is path+query only, never absolute; it is nested as a query value under local `/login`, and the eventual login redirect is sanitized by `SafeRedirectPathService` |
| Redirect loop between `/login` and `/two-factor` | Verified safe | Walked the real chain with a throwaway request spec: `GET /two-factor` → 303 `/login?next=/two-factor` → login POST → `/two-factor` → terminates at a 200 login page. Signed-in users are caught earlier by `redirect_to_signed_in_path` |
| Information leakage from redirecting rather than 404ing | Verified safe | No-session now returns the same login redirect regardless of the `user_id` supplied; session-present-but-wrong-`user_id` still 404s, so no new account or pending-session oracle |
| JSON/API consumers depending on the 404 | Verified safe | `/internal/admin/users/two_factor_authentication` is a different controller and mobile clients inherit from `Api::Mobile::BaseController`; no consumer of these web endpoints expects the 404 |

Also verified the new coverage is load-bearing rather than decorative: swapping the controller back to main's version reddens exactly the four new stale-session request examples and nothing else, then restores green. The spec-results screenshot in the body shows both runs.

One thing worth noting for review rather than fixing: the warning renders both as the inline alert (`AuthAlert`, for `status: "warning"`) and as a toast (`useFlashMessage` in the Inertia layout — the ERB toast partial is suppressed on Inertia pages). That double render is how every existing login warning already behaves; I captured a wrong-password attempt on unmodified main to confirm it, so it is not introduced here.

PR #6329, opened 2026-07-26
Retire ProductFile rows whose upload never finished, instead of re-discovering them on every save

gumclaw:
## What

A `ProductFile` row is created when a file is attached to a product, before the browser has finished uploading the object. When that upload never finishes, the row stays alive pointing at a storage key that was never written: `analyze` gives up on the missing object, `analyze_completed` is never set, and nothing else revisits the row. It then reads permanently as "only storage can say" — `ProductFile#stored_file_presence_known_from_row` returns `nil` forever.

This change makes that finding stick. When `stored_file_present?` performs a storage lookup and the lookup proves the object is absent, it enqueues `RecordProductFileMissingFromStorageJob` for that row. The job re-checks the row and, if the absence still holds and the row is old enough that its upload cannot still be in flight, sets `deleted_from_cdn_at` — the existing "there is no object in storage behind this row" marker. From then on the row answers for itself with no request to storage.

Concretely:

- New `RecordProductFileMissingFromStorageJob` (`retry: 1`, `queue: :low`, `lock: :until_executed`) with a class-level `eligible?` predicate: S3 file, not marked analyzed, not already marked, created more than `UPLOAD_GRACE_PERIOD` (3 days) ago. Before writing, it also checks that the object does not exist under an alternative Unicode normalization of its key.
- `ProductFile#stored_file_present?` enqueues that job on a proved-absent lookup. It does not write anything itself, because it runs inside a product save, and a queueing failure cannot fail that save.
- The row is left **alive**. Only the storage marker is set, so the seller's file list is unchanged.

## Why

Nothing currently remembers a proved-absent lookup, so every caller that needs to know whether a listing delivers anything pays a fresh storage request to re-learn the same answer about the same dead row.

The caller that pays for this is the spam-flag deliverable check (`ContentModeration::ModerateRecordService#has_deliverable_file?`, added in #6320 and refined in #6326). It can only afford a bounded amount of storage lookups per save, so a listing carrying more dead rows than that budget can fail the check even though one of its files really is stored — and a corroborated spam flag then blocks a legitimate publish. #6326 makes the budget a time budget rather than a file count, which stops any particular file being excluded by its position, but a pile of dead rows still spends real time proving nothing. Retiring the rows removes the cost at the source instead of rationing around it.

Design choices, and the alternatives from the issue's open questions:

- **Reuse `deleted_from_cdn_at` rather than add a column.** It already means exactly "there is no object in storage behind this row", which is precisely what the lookup proved, and `stored_file_presence_known_from_row` already answers `false` from it. A new "storage confirmed empty" column would need a migration to express a claim we can already express.
- **Do not soft-delete the row.** Soft-deleting would make the file disappear from the seller's product editor. That is a visible change to someone's product made on our initiative, and it is a separate decision from "stop paying to re-discover this"; this PR does the cheap, invisible half. The row staying visible also means a seller who thinks they attached a file still sees the row they can investigate.
- **Write from a job, not from the read.** The lookup happens during a product save (a validation path), which must not mutate. The job also re-checks eligibility, because between the lookup and the job the upload may have completed, the file may have been analyzed, or the row may have been deleted.
- **The grace period.** A large multipart upload over a slow connection can legitimately still be in flight. Marking such a row would tell every later caller the file is missing moments before it arrives. Three days covers a multi-gigabyte upload on a slow link with room to spare, and still retires the row long before the repeated lookups matter.
- **A file that only looks missing must not be marked.** S3 compares keys byte-for-byte, and the same accented filename can be encoded two valid ways in Unicode, so a key persisted in one form misses an object stored under the other — `exists?` says missing for a file buyers can still download, because the download path probes the variants (`SignedUrlHelper`, `S3Retrievable#confirm_s3_key!`). The job probes them too before writing. For plain-ASCII keys that costs no S3 call.
- **A storage outage must not retire anything.** `stored_file_present?` already treats an unreachable storage as "assume present"; that path returns before any enqueue, so an outage — or a permissions fault, since `Forbidden` is an `Aws::Errors::ServiceError` — cannot mark rows as empty.
- **The mark is one-way, on purpose.** Nothing clears `deleted_from_cdn_at`, and `stored_file_presence_known_from_row` consults it before `analyze_completed?`, so a wrong mark would not self-heal. That is why each way a real object can look absent is excluded before writing. There is no automatic un-marking because "the object came back" and "the object was purged" are indistinguishable from the row; a wrong mark is undone by nulling the column, which the class documents.
- **No seller notification here.** Telling a creator about an upload they abandoned months ago is a product decision, not a consequence of this cleanup, and it is easier to make once the marker exists.

A periodic sweep over old unanalyzed rows (the other option in the issue) would find rows that no save has touched, but it would also mean asking storage about every such row on a schedule. This change costs nothing extra: it only records lookups we were already making.

Refs gumroad-private#1370. Follow-up to #6320 and #6326.

## Before/After

Not applicable as a UI change — nothing a user sees changes. The row is deliberately left alive, so the seller's file list in the product editor renders exactly as before; the only difference is a timestamp column that no view reads (`git grep deleted_from_cdn app/javascript app/views app/presenters app/controllers` returns nothing).

Walkthrough video of the change and the behavior it adds:

https://github.com/antiwork/gumroad/blob/gumclaw/2026-07-26-retire-unfinished-uploads/qa-media/pr-gp1370-walkthrough.mp4?raw=true

## Test Results

New job spec covers: the absence being recorded, the row being left alive, the re-check finding the object present after all, the object existing under another Unicode form of the key, an upload still inside the grace period, analysis having since succeeded, the row having been deleted, the row no longer existing, an already-marked row, and an external link. New model specs cover the enqueue on a proved absence and the four cases that must not enqueue or must not break the save (still-in-flight upload, file really present, storage unreachable, queueing failure).

spec results

```
bundle exec rspec spec/sidekiq/record_product_file_missing_from_storage_job_spec.rb \
                  spec/models/product_file_spec.rb -e '#stored_file_present?' \
                  -e '#stored_file_presence_known_from_row' \
                  -e 'RecordProductFileMissingFromStorageJob'   # 25 examples, 0 failures
```

`rubocop` clean on all four changed files. The screenshot above is from the pre-review run (23 examples); the current head adds the two specs for the Unicode-variant guard and the enqueue-failure path.

Two things stated plainly rather than as a clean number:

- The rest of `spec/models/product_file_spec.rb` shows **10 failures that are identical on unmodified `main`** — they reach real S3 for `content_length` / `download_original`. Pre-existing, not this PR.
- `spec/services/content_moderation/moderate_record_service_spec.rb` is unstable on this machine (0, 2, 3, 15 failures across identical runs; **`main` at `1c9c36a94` returns 19 right now**, more than this branch) because sibling sessions share the test database. I'm not claiming a number for it — CI is the authority.

## QA steps

No user-facing surface to click through. To verify in a console on a preview or staging app:

1. Find an old `ProductFile` whose upload never finished: an S3 row, `analyze_completed` false, `deleted_from_cdn_at` nil, created more than three days ago.
2. `RecordProductFileMissingFromStorageJob.eligible?(file)` → `true`.
3. `file.stored_file_present?` → `false`, and a `RecordProductFileMissingFromStorageJob` is enqueued for its id.
4. Run the job. `file.reload.deleted_from_cdn?` → `true`, `file.alive?` → still `true`.
5. `file.stored_file_present?` again → `false` with **no** request to storage (`stored_file_presence_known_from_row` now answers `false`).
6. Load the seller's product editor: the file list is unchanged.

---

AI disclosure: written by Claude Opus 5 (`claude-opus-5`) running as an autonomous Hermes agent. It was given the issue text of gumroad-private#1370 — which describes the accumulating dead `ProductFile` rows, the cost they impose on the bounded deliverable check, and three open questions (what to record, where the write happens, whether to notify the seller) — together with a standing instruction to build the fix and open the PR once the issue was agreed to. The agent read the existing `stored_file_present?` / `has_deliverable_file?` implementations and #6320/#6326, chose reusing `deleted_from_cdn_at` over a new column and a job over a validation-time write, wrote the job and the specs, ran them locally, and recorded the walkthrough. It then ran an adversarial pre-merge review with a second Claude Opus 5 instance prompted to hunt specifically for wrong consumers of the reused column, buyer- and seller-visible fallout, unsafe enqueue-during-save, and irreversibility; that review found the Unicode-normalization false-absent case, the too-short grace period, and the save-failing enqueue, all fixed in `ce084f7` (verdict and per-finding disposition in a comment on this PR).


gumclaw:

**Reusing `deleted_from_cdn_at` instead of adding a column — what else reads it?** Three consumers, checked each:

- `ProductFile#stored_file_presence_known_from_row` returns `false` from it. That is the whole point of the change.
- `RemoveDeletedFilesFromS3Job` selects `s3.deleted.alive_in_cdn`, so a marked row is skipped by the S3 cleanup sweep if the seller later deletes it. Correct: there is no object to delete, which is exactly what we proved.
- `RenameProductFileWorker` returns early on `deleted_from_cdn?`. Also correct: renaming an object that was never written would fail.

No view, presenter, controller or frontend file reads the column (`git grep deleted_from_cdn app/javascript app/views app/presenters app/controllers` is empty), which is why this has no UI surface.

**Why the read doesn't write.** `stored_file_present?` is called from `ContentModeration::ModerateRecordService` during a product save. Writing there would mean a validation path mutating rows, and it would write based on a lookup whose result may already be stale by the time the save commits. The job re-checks `eligible?` plus the storage lookup itself, so all four ways the situation can change between enqueue and run (upload completed, analysis succeeded, row deleted, row already marked) are handled — each has a spec.

**Why the grace period is a hard gate rather than a heuristic.** The failure mode of getting it wrong is worse than the cost of waiting: a large multipart upload still in flight would be permanently recorded as empty, and nothing revisits the marker. `eligible?` is also a class method rather than inline in `perform` so the enqueue side and the job side cannot drift apart — they call the same predicate.

**The outage path.** `stored_file_present?` rescues storage errors and returns `true` without reaching the enqueue, so an S3 outage cannot mark a fleet of rows as empty. Specced explicitly, because this is the one bug in this change that would be both silent and unrecoverable.

**What this deliberately does not do.** The row stays alive, so nothing disappears from a seller's product editor, and no seller is notified. Both are product decisions the issue raises as open questions, and neither is needed to stop the repeated lookups. They are easier to make once the marker exists.

**Enqueue volume.** One job per proved-absent lookup, and only for rows past the grace period. A row is marked once and then answers from itself, so a listing with N dead rows generates at most N jobs total rather than N per save. `lock: :until_executed` collapses duplicates if several saves race the same row before the first job runs.

gumclaw:
## Fable-5 adversarial pre-merge review

Verdict as returned: `CHANGES-REQUIRED`. Findings, verification, and disposition:

| # | Finding | Verdict | Disposition |
|---|---|---|---|
| P1-1 | Branch reverts #6326's time-budget storage check and deletes its three regression specs | **False positive on the mechanism, real on the premise** | The reviewer read a two-dot diff. `git diff origin/main...HEAD --stat` never listed the moderation service — my branch simply predated #6326 and did not touch it, so a merge could not have reverted anything. But it was right that the branch was behind: #6326 landed after my branch point. Merged `origin/main` in (`158e3c5`); the diff is still only my own files. |
| P2-2 | A row whose object exists under a different Unicode normalization of its key gets marked "nothing in storage" forever | **Real, and the most serious finding** | Fixed in `ce084f7`. Genuine: `S3KeyUnicodeNormalization` exists precisely because `exists?` returns missing for these, and both the buyer download path (`signed_url_helper.rb:29`) and `S3Retrievable#confirm_s3_key!` already probe the variants — so this could have permanently branded a file buyers can still download as undeliverable, and leaked its S3 object on later deletion. The job now probes the variants before writing (free for ASCII keys). New spec pins it. |
| P2-3 | The mark is a one-way ratchet with no recovery, and the 1-day grace period overstates "longer than any real upload" | **Real** | Fixed in `ce084f7`. Grace period widened 1 day → 3 days (a multi-gigabyte upload on a slow link genuinely exceeds a day, and there is no self-healing afterwards). The one-way nature is now documented on the class along with the manual undo (null `deleted_from_cdn_at`), and the reasoning is stated: automatic un-marking is deliberately absent because "the object came back" and "the object was purged" are indistinguishable from the row. |
| P3-4 | Comment says "never successfully analyzed"; only video analysis sets `analyze_completed` | **Real** | Fixed in `ce084f7`. Verified: the flag is assigned only in `with_file_properties.rb` video analysis, so analyzed PDFs and images also pass the gate — the storage lookup is what actually decides. Comment corrected rather than the gate, since the gate is right. |
| P3-5 | `perform_async` raising (Redis down) propagates out of a validation and fails the seller's save | **Real** | Fixed in `ce084f7`. Enqueue wrapped; the method returns the answer it already has. Specced. |
| P3-6 | `qa-media/pr-gp1370-*` doesn't follow `pr-<number>-<description>` | Accurate | Left as-is: the files were committed before the PR number existed, and renaming them now would break the raw URLs already in the body for no reader benefit. |

Worth recording, because these were the things most likely to be wrong in a change that repurposes an existing column:

- **Every consumer of `deleted_from_cdn_at` enumerated repo-wide.** Only three production readers: `RenameProductFileWorker` (skips a rename — correct, nothing to rename), `RemoveDeletedFilesFromS3Job` via `cdn_deletable` (skips the S3 delete — correct for a genuinely empty row), and `stored_file_presence_known_from_row` (the point of the change). Sibling `CdnDeletable` classes are never marked by this job.
- **No buyer-facing change.** Download signing, url_redirect, archives, Send-to-Kindle and streaming never read the column, so a purchased product's marked row degrades nothing.
- **No seller-facing change.** No presenter, serializer, props JSON or admin screen reads it, and `save_files!` round-trips neither touch nor re-validate it. This is what makes the "no UI evidence" claim in the body true rather than convenient.
- **`update_column` is the right write** — skipping paper_trail and the product-cache invalidation matches the pre-existing `mark_deleted_from_cdn` caller, and nothing cached reads the column.
- **The outage path cannot mark anything.** `Forbidden` is an `Aws::Errors::ServiceError`, so a permissions misconfiguration answers "present" and never reaches the enqueue — the mass-mislabelling scenario is closed.
- **Enqueue-during-save is sound**: the row is necessarily older than the grace period so it committed long before the in-flight transaction, the job re-verifies from committed state, and a rolled-back save leaves the mark independently justified.
- **Legitimately-unanalyzed rows with real objects can't be marked**: a failed video analysis, Dropbox imports (object precedes the row), and `is_linked_to_existing_file` copies all return present from the lookup.
- **The specs fail with the feature reverted** — checked assertion by assertion, not assumed.

### Test state

`spec/sidekiq/record_product_file_missing_from_storage_job_spec.rb` + the `#stored_file_present?` / `#stored_file_presence_known_from_row` groups: **25 examples, 0 failures** at this head. `rubocop` clean on all four files.

Two honest caveats rather than clean numbers:

- The rest of `spec/models/product_file_spec.rb` has **10 failures, identical on unmodified `main`** (they hit real S3 for `content_length` / `download_original`). Pre-existing, not this PR.
- `spec/services/content_moderation/moderate_record_service_spec.rb` is unstable on this machine — it returned 0, then 2, 3, 15 failures across identical runs, and **`main` at `1c9c36a94` returns 19 failures right now**, more than this branch. Shared test database across concurrent sessions. I'm not claiming a green number for it; CI is the authority.

### Only verifiable live

- Whether any bucket/endpoint misconfiguration could make HeadObject 404 for existing objects. Static analysis can only note that the marker has no kill switch; the Unicode probe and the grace period close the two known false-absent paths, and a config-level fault would need the manual undo.
- Real-world population of rows with Unicode-mismatched keys, and of uploads exceeding three days.

PR #6336, opened 2026-07-26
Report the storage-check budget against the save's real outcome

gumclaw:
## What

Follow-up to #6335, addressing Greptile's P1 review on that PR (it merged minutes after the review landed, so the fix could not go on that branch).

#6335 moved the storage-check budget warning out of the per-file-list check and into `product_has_substantive_deliverable?`, so one save produces one line instead of one per tier. It guarded that line on `found_file` — but files are only half of what counts as a deliverable. A product whose files all went unverifiable still passes when its content page has something readable in it, and that passing save carried a warning shaped like a rejection. The guard now runs against the settled result, so the line only appears on a save that actually failed.

The total in that line could also overstate itself, which came out of verifying the first fix. A file attached to a tier still carries the product in `ProductFile#link_id`, so the product's own file list and the tier's list both walk it, and adding up per-list counts counted the same file twice. A file that *was* looked up got counted as unchecked too, because a later list reached it again after the budget was gone. The budget now tracks file ids rather than accumulating counts, and reports the files that were asked about nowhere: everything skipped, minus everything looked up.

Nothing about what counts as a deliverable changes, and neither does the bound: one budget per save, tested before each lookup, and an exhausted budget still fails the check.

## Why

The number in this log line is what someone reasons about when a seller says "my save was rejected and I don't know why". Both defects made it misleading in the same direction — the reader is sent looking for a problem that either does not exist (a warning on a save that succeeded) or is smaller than reported (a count inflated by files reached through two lists, or already verified).

On a membership with ten tiers whose files are all attached at tier level, the pre-fix count reported more files unchecked than the seller ever attached.

## Before / After

Not applicable — no user-facing surface changes. This is the internal deliverable check behind the spam flag, reached only from a product save, and the change is to what it writes to `Rails.logger`. The observable difference is asserted directly in the specs below.

## Test plan

Two new specs, each verified RED against the code as merged in #6335 and GREEN with this change:

| Spec | Pre-fix behaviour |
|---|---|
| `stays quiet about a spent budget when readable content makes the save pass anyway` | warned on a save that passed |
| `counts each file once however many lists reach it` | reported `1` as `3` — two lists walking the same two files, one of which had been looked up |

The existing `logs a single warning naming the total left unchecked across all tiers` spec computed its expectation by summing per-list counts, which encoded the double-count; it now derives the expectation from the distinct file ids the check can reach.

Each half of the fix was mutation-verified independently, so neither is carried by the other:

- reverting the set difference to a plain skipped count (`@skipped_file_ids.size`) reddens both tally specs
- reverting the log gate to `found_file` reddens exactly the passing-save spec and nothing else

Full file, on a worktree cut from `main` at the #6335 merge commit:

```
71 examples, 0 failures
```

spec run

RuboCop clean on both changed files.

## QA steps

No page or component to click through. To exercise it by hand on a preview app:

1. Create a membership with several tiers and attach a file to each tier, killing each upload before it completes. Save. The save should be rejected, and the log should carry exactly one `storage check budget spent` line whose count is the number of files you actually attached — not a multiple of it.
2. Create a product with an unfinished upload attached *and* a content page with a paragraph of text in it. Save. The save should succeed, and the log should carry no `storage check budget spent` line at all.

---

## AI disclosure

Written by `claude-opus-5` (Anthropic) running as the Gumclaw Hermes agent.

Instructions given: address Greptile's P1 finding on #6335 — the aggregated budget warning is emitted before `has_readable_body_content?` decides the overall result, so a passing save keeps a failure-oriented warning. #6335 had already merged, so the fix lands as a follow-up branched off fresh `main`. The double-counting defect was found while writing the regression spec for the reported one and fixed in the same pass, since both live in the same handful of lines.


gumclaw:
Self-review against CONTRIBUTING.md, re-read against this diff:

**Scope** — two files plus one evidence PNG, nothing else. `git show --stat` confirms no stray artifacts.

**The bound is unchanged.** One budget per save, `spent?` tested before each lookup, and an exhausted budget still fails the check. What changed is when the line is written and how the number in it is computed. `record_checked_file` runs on the same iteration that performs the lookup, so it cannot make the check do more work than before.

**Why the tally is a set difference and not just a set.** Deduping skipped ids alone still overcounts, because the file that burned the budget is skipped by the *next* list that reaches it — so it would appear in the skipped set despite having been looked up. Subtracting the checked ids is what makes the number mean "asked about nowhere". This is the case the second spec pins: two lists, two files, exactly one never asked about.

**Ordering fix in `has_deliverable_file?`.** The sorted list is now bound to a local before iterating, because `drop(checked)` has to index into the same order the loop walked. Dropping from the unsorted `unverifiable_from_row` would have recorded the wrong subset — right count, wrong ids, which the dedupe then makes observable.

**No user-facing surface.** The check is reached only from a product save and writes to `Rails.logger`; there is no component or page in the diff. Stated in the body rather than left as an unexplained empty Before/After.

**Tests.** Both new specs verified RED against the code as merged in #6335 (3-instead-of-1, and a warning on a passing save) and GREEN with this change. The pre-existing tally spec had the double-count baked into its expectation via summed per-list counts; it now derives the expectation from distinct reachable file ids, so it would fail again if the dedupe regressed. RuboCop clean.

gumclaw:

Verdict: **READY-TO-MERGE** at `4aa414ccd`. No P1, no P2, one non-blocking P3.

|---|---|---|
| P3 — the "exactly one lookup happened" arithmetic assumes under 2s elapses between constructing the budget and the first `spent?` check | Real but **pre-existing** — the spec this PR modifies already made the same assumption (`walked_files - 1`), and the other sleep-based budget specs in this file share it | Left as is. If it ever flakes the fix is stubbing the monotonic clock, not widening the assertion — noting it here so the next person doesn't rediscover it |
| Process warning: "the working tree was mutated by another process mid-review" | **Not real** — that was this session running mutation tests in the same worktree while the review ran | Confirmed benign: local HEAD, `git ls-remote`, and the PR head all read `4aa414ccd`, and the tree is clean. The transient states it saw were exactly the two mutations reported below |

Declined nothing else, because nothing else was raised. The checklist items that came back negative are worth naming since they're the ones that would have blocked: no state where a genuinely budget-caused rejection logs nothing; no off-by-one in `drop(checked)`; no weakening of the deliverable gate; no self-fulfilling assertions; the dedupe fixture genuinely double-reaches both files; no PII, log-volume, or N+1 concerns.

### Mutation testing — each half proven load-bearing separately

Reverting the whole fix would have said nothing about either part individually, so both were mutated on their own:

| Mutation | Result |
|---|---|
| `(@skipped_file_ids - @checked_file_ids).size` → `@skipped_file_ids.size` | 2 failures — both tally specs, including the pre-existing one |
| log gate `!has_deliverable` → `!found_file` | 1 failure — exactly the passing-save spec, nothing else |

Neither half is carried by the other, and neither spec is vacuous.

### The one thing worth a second reader's attention

The reason the tally is a set *difference* and not just a deduped set: the file that burned the budget gets skipped by the next list that reaches it, so it would appear in the skipped set despite having been looked up. Subtracting the checked ids is what makes the number mean "asked about nowhere". `record_checked_file` runs on the same iteration that performs the lookup, so it cannot add work — and a count of zero is therefore correct silence rather than a lost warning, because it can only happen when every skipped file was verified somewhere.

### Test-suite caveat

`moderate_record_service_spec.rb` is 71 examples, 0 failures on the last two full runs. During earlier runs the flag-off example at `:33` failed intermittently; it passes in isolation and in both clean full runs, and the instability tracked concurrent sessions sharing `gumroad_test` rather than anything in this diff. CI is the authority.

gumclaw:
CI green at `4aa414ccd` — 73 jobs success, 1 skipped, 0 failures ([run 30186663039](https://github.com/antiwork/gumroad/actions/runs/30186663039)). That is the `push`-event run, which is the one that actually executes the shards; the `pull_request_target` run on the same commit is the internal-PR noop and its "skipping" jobs are not test results.

This PR is complete on everything I own:

- both defects fixed, each half mutation-verified independently
- 71 examples, 0 failures locally; RuboCop clean
- local pre-merge adversarial review: READY-TO-MERGE, no P1/P2 (triage comment above)
- Greptile re-review on this PR: 5/5, no blocking failure, no inline findings
- the two P3s from #6335's review verdict cleared

Not merging it myself — leaving that call to a human.

gumclaw:
Closing as a duplicate of #6337, which is a strict superset of this diff. Both were opened by concurrent sessions of mine within six minutes, both as follow-ups to the same Greptile P1 on #6335, and neither saw the other.

I diffed the two service changes line by line rather than judging by the descriptions. Every substantive line here is present there:

- Both replace the `StorageCheckBudget` integer tally with `@checked_file_ids` / `@skipped_file_ids` sets, so a file reachable through both a product's list and a tier's list is counted once. (`ProductFile#link_id` points at the product either way, which is why the old count could report seven files for a four-file, two-tier membership.)
- Both add `private_constant :StorageCheckBudget`.
- Both settle the whole deliverable answer into `has_deliverable = found_file || has_readable_body_content?(record)` before deciding whether to warn, so a product that ran out of budget on files but passed on page content no longer emits a failure-shaped warning on a save that succeeded.

The only differences are one method name (`record_unchecked_files` vs `record_skipped_files` — the latter reads better, since the argument is the skipped set rather than the unchecked total) and one thing #6337 has that this PR does not:

```ruby
if reasons.any? { |r| spam_reason?(r) } && spam_flag_should_not_block?
```

That short-circuit is worth keeping. `spam_flag_should_not_block?` walks the product's files and spends up to a two-second storage-lookup budget on files their own row can't answer for. Without the guard, that work runs on every save even when nothing was flagged and there is therefore nothing to downgrade — and its "we ran out of time" warning gets written during saves that had no spam flag at all. This PR fixes the warning's *content*; #6337 also stops it being reachable on saves where the question is moot.

Both are still mid-CI, so nothing is lost by consolidating onto the larger one. No unique work discarded. Continuing on #6337.

PR #2942, opened 2026-01-17
Migrate `_forms.scss` to Tailwind: Part 1

SalmanNajah:
Issue: #1055

Splitting #2747 into micro-PRs as discussed in https://github.com/antiwork/gumroad/pull/2747#issuecomment-3762858746

# Description
- Migrate core form components and admin pages to Tailwind

#### Note: To test the UI changes in this PR locally, please make sure to first comment out the `_forms.scss` then restart the dev server.
# Demo
## Before
https://private-user-images.githubusercontent.com/145349710/536926660-8e82afe9-55cd-494f-9f52-53623c78d3b6.mov?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Njg2NDI1MjIsIm5iZiI6MTc2ODY0MjIyMiwicGF0aCI6Ii8xNDUzNDk3MTAvNTM2OTI2NjYwLThlODJhZmU5LTU1Y2QtNDk0Zi05ZjUyLTUzNjIzYzc4ZDNiNi5tb3Y_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwMTE3JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDExN1QwOTMwMjJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1kMzE4OTEzY2FiZDUzZDMwYWY5ZTA5Y2E0MmExOTE4NmY1OTQ5NGZmYzJlN2EwMjFmNTE2Y2M1NzAzMThkZTIwJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.ibtjzX4PAz2Ri1ASytv1dFnjCKan1WWY1uI2lmYFwMY

## After
[...]

PR #6338, opened 2026-07-26
Keep values in disabled input groups legible

gumclaw:
## What

A disabled `InputGroup` faded its whole subtree with `opacity-30`. CSS `opacity` applies to every descendant, so any input group that exists only to *display* a computed value rendered that value at 30% opacity. This replaces the fade with a tinted background (`bg-active-bg`) plus the existing `cursor-not-allowed`, so the disabled state is still obvious while the value keeps full contrast.

Two supporting changes in the same primitive:

- `Input.tsx` had `inputGroupDisabled && "opacity-100"`, an earlier attempt at exactly this goal that could never work — a descendant cannot cancel an ancestor's `opacity`. As `disabled:opacity-100` it now does something real: it cancels the input's own `disabled:opacity-30` from `baseInputStyles` so the two no longer stack.
- `InputGroup`'s context `useMemo` keyed on `state` instead of `disabled`, so groups that toggle `disabled` from form state left descendants reading a stale flag.

Plus two accessibility fixes that removing the fade makes necessary — details in the Accessibility section: the currency `<select>` inside a disabled `PriceInput` was still operable, and the tint alone is invisible under forced colours.

## Why

Split out of #6330. That PR started from a seller report about the pay-what-you-want **Minimum amount** field, and the agreed fix for *that* field is to delete it and state the floor as a note instead — #6330 now carries only that. But the underlying primitive bug is independent of the field being removed, and it affects **12 other `<InputGroup disabled>` call sites**: Collaborators/Form (×2), WorkflowsPage/WorkflowForm (×2), Checkout/CreditCardInput, CheckoutDashboard/DiscountInput, UtmLinks/UtmLinkForm, Settings/Main/Show, Affiliates/Onboarding (×2), and Affiliates/Form (×2). Anywhere a group is disabled to show a derived value, that value is currently near-invisible.

The highest-stakes one is checkout's card group, disabled while a payment is processing: the card number washes out to 30% at exactly the moment a buyer is watching for confirmation that their details were accepted. A percentage discount's disabled fixed-amount field and a published workflow's disabled price fields have the same problem.

The `opacity-30` fade arrived in be281f191 ("Migrate `_forms.scss` to Tailwind: Part 1", #2942) — the SCSS it replaced dimmed differently, so this is migration fallout rather than a deliberate design choice.

The trade-off worth flagging for review: every one of those 13 surfaces changes appearance from faded to tinted, so this diff is visually broader than the number of lines suggests.

A second, smaller fix rides along because this change surfaces it. `PriceInput`'s currency picker is an invisible `<select>` stretched over the currency pill so the whole pill acts as its hit area. CSS `opacity` never blocked pointer events, so that select was always clickable on a disabled field — a seller could change the currency of the fixed-amount discount field while "Percentage" was selected. The 30% fade at least made the pill *look* inert; now that a disabled group keeps full contrast, the pill reads as an active control, so the select is given the real `disabled` attribute instead of relying on appearance.

Refs antiwork/gumroad-private#1368

## Before / After

The delta, zoomed on a disabled price field, then the full section across the full matrix — desktop (1280px) and mobile (375px) × light and dark:

before and after

Walkthrough video — before (main) then after (this PR), typing in the enabled field and watching the disabled mirror follow it:

https://raw.githubusercontent.com/antiwork/gumroad/fix/disabled-input-group-legibility/qa-media/pr-6338-pwyw-minimum-amount-walkthrough.mp4

Because this changes a shared primitive, here are two of the other affected surfaces — the checkout card group (disabled while a payment is processing) and a price field with a `Pill` suffix (disabled after a workflow is published):

other disabled surfaces

Measured, not eyeballed. The real `PriceInput` was mounted in a local Vite harness against the real `design.scss`, and the numbers came out of `getComputedStyle`:

| | effective opacity of the displayed value | contrast against the composited group background |
|---|---|---|
| Before (main) | 0.30 | faded |
| After (this PR) | 1.00 | 15.22:1 light, 8.84:1 dark — identical to an editable field |

The field stays `disabled` throughout, so it remains non-editable and keeps its `cursor-not-allowed`. Same measurements hold on the other surfaces: the card number and the "per month" pill both go from 0.30 to 1.00.

## Accessibility

Removing the fade changes what carries the disabled state, so the new treatment was audited against WCAG in a browser — the real components mounted against the real `design.scss`, read back through `getComputedStyle` and the Chrome accessibility tree. Three gaps came out of it, all fixed here.

**The currency `<select>` was still operable on a disabled field (4.1.2).** It is an invisible `opacity-0` element stretched over the currency pill, and it is not the amount input, so it never inherited `disabled`. Measured before the fix: `focusable: true`, reported as enabled in the accessibility tree, and driving it programmatically **succeeded** — a keyboard or screen-reader user could change the currency of a field the interface presents as switched off. This got worse once the fade was removed: at 30% opacity the pill was at least visually hidden, but a solid-looking pill that still responds is a stronger invitation. Now `disabled`, and verified after: `focusable: false`, operating it times out.

**That selector had no accessible name.** It carried only `name="Currency"`, which is a form-submission attribute, so the accessibility tree exposed the combobox with an empty name and a screen reader announced it unlabelled. Now `aria-label="Currency"`, verified in the tree as `role: combobox, name: "Currency"`.

**The tint is invisible under forced colours (1.4.1).** `bg-active-bg` is a translucent overlay, and forced-colours modes (Windows High Contrast) substitute a system colour for author backgrounds *while preserving the alpha* — so the tint resolves to `rgba(255,255,255,0.1)` over the white Canvas it now sits on, which composites to exactly `rgb(255,255,255)`: pixel-identical to an editable field. Measured, not assumed. With the fade gone and only a tint left, the state was carried by colour alone and a disabled group was indistinguishable from an editable one there. A dashed border is not a colour and survives forced colours, so it now marks the state.

Contrast holds across the whole matrix, and the disabled group stays distinguishable from its editable peer in every cell:

| | disabled value contrast | disabled border | editable peer border |
|---|---|---|---|
| Light | 15.22:1 | solid | solid (tint distinguishes) |
| Dark | 8.84:1 | solid | solid (tint distinguishes) |
| Light, forced colours | 14.02:1 | **dashed** | solid |
| Dark, forced colours | 13.98:1 | **dashed** | solid |

WCAG matrix

One thing worth flagging for review rather than burying: `aria-label="Currency"` touches three existing Capybara specs that do `select "£", from: "Currency", visible: false` (`products/edit/edit_spec.rb`, `checkout/discounts_spec.rb`, `products/creation_spec.rb`). `Capybara.enable_aria_label = true` is set in `spec_helper.rb`, so the label resolves either way and these should stay green — but they are system specs I did not run in this worktree, so that is a CI expectation, not a local claim.

## Test Results

8 examples across two files, all passing locally in this worktree. `app/javascript/components/PriceInput.test.tsx` (5) walks every ancestor from the input up to the group and asserts the effective opacity while disabled is 100 — matching state-prefixed utilities too, so it catches the input's own `disabled:opacity-30` and not just the group's fade; three further examples cover the currency selector being disabled with the field, usable without it, and carrying an accessible name. `app/javascript/components/ui/InputGroup.test.tsx` (3) covers the primitive directly: no fading opacity on a disabled group, the forced-colours cue present, and neither applied to an enabled group.

test results

Each fix was individually reverted to confirm the suite actually bites, rather than assuming it:

| reverted | result |
|---|---|
| `InputGroup` tint → `opacity-30` | fails (`expected 30 to be 100`) |
| `Input`'s `disabled:opacity-100` | fails (`expected 30 to be 100`) |
| `PriceInput`'s `disabled` on the selector | fails (`expected false to be true`) |
| `aria-label` on the selector | fails (`Unable to find a label with the text of: Currency`) |
| `forced-colors:border-dashed` | fails (`expected '…' to contain 'forced-colors:border-dashed'`) |
| all present | 8 passed |

`prettier` and `eslint` clean on all four touched files.

## QA steps

The reported PWYW field is being removed in #6330, so these steps verify the primitive on the surfaces that remain:

1. Checkout → Discounts → add a discount, select **Percentage**. The greyed-out **Fixed amount** field beside it should show its value legibly on a lightly tinted background rather than faded out, and typing into it should do nothing.
2. On that same field, try to open the currency dropdown by clicking the currency pill. It should not open.
3. Publish a workflow, then reopen it — its disabled fields (including a price field with a "per month" pill) should read as tinted-but-legible, not faded.
4. Start a checkout and watch the card field while the payment processes: the card number should stay readable through the disabled state.
5. Settings → Advanced with a saved card is **not** one of these — that group is `readOnly`, not `disabled`, so it should look exactly as it did before this diff.
6. Optional accessibility check: with Windows High Contrast (or Chrome DevTools → Rendering → "Emulate CSS media feature forced-colors: active"), a disabled input group should show a **dashed** border while editable fields stay solid.
7. Optional keyboard check: tab through a disabled price field that has a currency pill — focus should skip the currency dropdown entirely rather than landing on it.

---

## AI disclosure

Written by `claude-opus-5` (the `model.default` in this agent's config).

This PR is the primitive half of a split. The work began from a one-word "Fix." instruction from @slavingia on antiwork/gumroad-private#1368, which already contained the reported symptom, the root-cause analysis, and a suggested fix; a second instruction on the original PR asked for the change to follow WCAG and ensure compatibility. When @slavingia redirected the *product* solve ("just 'must be at least...' as a note under suggested would be a better solve"), the model proposed splitting the branch — the note in #6330, the shared-primitive fix here — and he agreed.

The model departed from the issue's original suggested fix and said so: the issue proposed adding a `readOnly` prop to `PriceInput` and swapping `disabled` → `readOnly` at the call site. `InputGroup`'s `readOnly` variant is `bg-inherit border-none px-0`, which would have visually deleted the field's box, and it would have left the same bug in the other disabled call sites. Fixing the `disabled` variant in the primitive was chosen instead. A `text-muted` variant was also tried and rejected on measurement — it only reached 3.98:1 contrast, worse than the tinted background.

Evidence was produced by the model: a Vite component harness mounting the real components, assertion-first Playwright capture (every screenshot gated on DOM/computed-style checks), and the contrast arithmetic above. The WCAG audit was run as measurement rather than review: computed styles and the Chrome accessibility tree read back from the running components, plus attempting to actually operate the disabled currency selector rather than inferring from markup. The model also drafted and then reverted an `inert` on the disabled group — it would have hidden the value from screen readers, which is the opposite of this PR's goal — and confirmed no `aria-disabled` was needed on the group, since the accessibility tree already reports the field correctly as `role: textbox, name: "Amount", value: "3", disabled: true`.

The change went through a local adversarial review by the same model, which returned CHANGES-REQUIRED with three findings that are all now fixed: the regression test guarded only the `InputGroup` half of the fix, the currency selector stayed clickable while disabled, and the evidence covered one of thirteen changed surfaces while pointing a QA step at a `readOnly` group that this diff does not touch.

Two findings were deliberately **not** fixed, and are called out here rather than quietly dropped. A disabled `Textarea` inside a group would still fade, because `Textarea` does not consume the group context — latent, since no consumer disables such a group today. And Settings → Main's PPP field disables the group and the inner input from different flags, so it briefly fades inside an untinted group while saving; that is pre-existing and identical before this PR, and the clean fix is pointing both at the same flag. Both are follow-ups rather than folded in here.


gumclaw:
Closing as a duplicate of #6330 — same fix, opened by a concurrent session of mine that didn't see #6330 already existed.

The two diffs are byte-identical in behaviour. I compared them directly: both replace `InputGroup`'s `opacity-30` with `bg-active-bg`, both delete the `inputGroupDisabled && "opacity-100"` line in `Input.tsx` that could never work (a descendant cannot cancel an ancestor's opacity), and both add the same two test files. The only differences are the `ariaLabel` string used in the fixtures (`"Amount"` here, `"Minimum amount"` on #6330) and comment wording.

#6330 is the one to keep, because the work that isn't in the diff already happened there:

- Sahil reviewed it and signed off (`Sounds good. Agree`, 03:45Z).
- A fable-5 adversarial pass found three P2s that are fixed on that branch and were never present here as fixes — most importantly that `PriceInput`'s currency selector stayed **operable** while disabled. It's an invisible `<select>` stretched over the currency pill, so it never picked up `disabled`; opacity was the only thing making it look inert. Removing the fade without that fix turns a dead control into an inviting one. This PR carries the fade removal and not the fix.
- The forced-colours/High Contrast gap is fixed there too. `bg-active-bg` is `rgba(255,255,255,0.1)`, and forced-colours preserves alpha — so over the white Canvas it composites to exactly `rgb(255,255,255)`, pixel-identical to an editable field. With the fade gone and only a tint left, the disabled state was carried by colour alone and vanished entirely. #6330 adds a dashed border, which isn't a colour and survives.
- It has evidence for the checkout card group (disabled while a payment is processing), which is the highest-stakes of the 13 surfaces this primitive change touches.

No unique work is being discarded here. Continuing on #6330.

PR #1230, opened 2025-09-12
Make discount input fields visible in Insert Upsell Modal on mobile

nyomanjyotisa:
### Explanation of Change
Fixes an issue where Input fields in the Insert Upsell Modal (for percentage and fixed amount discounts) are properly visible on mobile.

### Screenshots/Videos
Before

https://github.com/user-attachments/assets/e36d2587-93c7-4b24-ad2d-b8a642d8f6ed

After

https://github.com/user-attachments/assets/ca445ee4-32b2-41d9-af7e-06137969c3bc

### AI Disclosure
No AI tools used

PR #1379, opened 2025-09-27
Fix(flaky test): utm link spec flakes because of URL query param sync. issue

romitg2:
ref #1127 

### Failing CI run: (failed today)
https://github.com/antiwork/gumroad/actions/runs/18059333042/job/51393485472?pr=1375#step:8:136
Screenshot 2025-09-27 at 8 52 14 PM

# Problem:
### URL Query param aren't synchronised properly due to fast interaction, causing flaky test. (example video below)

https://github.com/user-attachments/assets/269775f9-9b78-4e84-babb-6f7b105d1f3a

### AI Disclosure:
- no AI used.

# Solution:
- assert query param after sorting interaction so they are syncronised properly

# local run after fix:
Screenshot 2025-09-27 at 8 36 23 PM

PR #6348, opened 2026-07-26
Let a failed save hand the editor the ids it just created

gumclaw:
Part of gumroad-private#1360 (PR 1 of 2). Fixes the retry-storm mechanism behind gumroad-private#1230.

## The bug

When a product-editor save **failed**, the editor kept its client-generated ids. The seller's retry resubmitted those same ids, the server could not recognise them, and it re-created the records — tripping the content deletion guard. That is the shape seen repeatedly in the #1230 monitoring: the same product blocked two or more times in a row.

The root cause is in two halves, and **a client-only fix is impossible**:

- `links_controller.rb` returned `save_id_mappings_response` on the success path and the offer-code warning path only. It was **absent from every error and conflict response**, so on failure the editor had nothing to adopt.
- `ProductEditPage.tsx` called `applyCanonicalIds` twice, both **above** the `catch` block. Every error branch skipped adoption entirely.

## What changed

**Server.** New `persisted_save_id_mappings_response` filters the accumulated client→canonical mappings against rows **actually alive in the database**. Failure paths run after the transaction rolled back, so the accumulator can reference rows that no longer exist — handing the editor an id for a rolled-back row would be worse than handing it nothing. Rendered on all three error sites in `update`: the stale-content 409, the hidden-variant-content 422, and the generic invalid-record 422.

The mapping keys are now present on **every** response shape, so client adoption is uniform. When a rollback undid everything the maps are correctly empty rather than stale.

**Client.** The `catch` block adopts whatever mappings the failure carries, for all three error types, before surfacing the error. Errors in `product_edit.ts` now carry `idMappings` (`SaveProductError extends ResponseError` for the generic case, so `assertResponseError` still works). Adoption updates the live state, the baseline, and `serverIdMappings` — the same triple the success path updates, so it survives a re-render.

Response keys are only **added**; nothing existing changed shape. An editor tab opened before this deploy keeps working.

## Tests

Five added to `test/controllers/links_controller_test.rb`:

1. Hidden-variant-content conflict returns the mapping keys, excluding rolled-back records
2. Generic save failure returns the mapping keys, excluding rolled-back records
3. Repeated saves with adopted canonical ids never consume the deletion-guard rewrite budget
4. An uploaded file stays attached across create → edit → save → save
5. Failure-path mappings do **not** weaken the deletion guards (missing-page and missing-variant payloads are still rejected)

**Green:** `366 runs, 1343 assertions, 0 failures, 0 errors`

**Red proof** — reverting only `links_controller.rb` to main, keeping the tests: `5 runs, 29 assertions, 2 failures`. The two mapping-contract tests fail, confirming the server change is load-bearing. Restored and re-verified green afterwards.

Test 4 **passed on unmodified main**: uploaded-file survival was already correct, so no fix was manufactured for it — that test is a regression guard. Scope item 4 of #1360 needed no change.

`rubocop` clean, `ruby -c` OK, `prettier`/`eslint` clean.

## Test evidence

test evidence

Top: full-file green with the fix applied. Bottom: the red proof — `links_controller.rb` reverted to `origin/main` with the new tests kept, and the two mapping-contract tests fail. Both runs reproduced independently before this PR was opened.

No UI screenshot: the change is a response-body contract plus an error-path branch, with no visual surface. The reviewable artifact is the diff and the red proof above.

One disclosure on the test runs: an early push (`aaecc3ba`) failed CI on these two tests because that commit accidentally contained a staged revert of `links_controller.rb` — `git checkout origin/main -- <path>` stages as well as restores, and my red-proof script only restored the working tree. So CI correctly ran the new tests against code without the fix. Fixed in `5105b812f`; the numbers above were re-verified by checking out the **committed** SHA into a clean worktree with CI's own seed. Details in the PR comments.

## Reviewer notes

The judgement call worth checking is `persisted_save_id_mappings_response` filtering against live DB rows. The stale-content 409 raises *before* any mutation, so its mappings are always empty; the other two paths raise inside the transaction, so their records are rolled back and the filter also yields empty in the common case. The filter is what makes the contract safe rather than the emptiness, and the tests assert the *keys are present* rather than asserting a particular emptiness, so a future path that persists something partial still behaves.

## Deliberately out of scope

- **Owner-scoped page mappings, variant GUID client ids, and splitting the merged `serverIdMappings` namespace** — PR 2 of #1360. Worth noting that new versions/tiers currently take client ids from a module-level integer counter (`VersionsEditor.tsx:21`, `TiersEditor.tsx:42`), so a new version is submitted as `"0"`; variant and page mappings also share one namespace and only avoid collision because their formats differ by accident. Both are real and both are PR 2.
- **`StaleContentWriteGuard`'s blocking path stays gated off** (`product_editor_stale_content_block`, default OFF). This PR does not re-enable it, and the flag must stay off until the payload contract is settled — see gumroad-private#1295.
- **Revision tokens** — not started, per direction.
- **Missing/partial variant and page arrays** — gumroad-private#1361, still unresolved. This PR does not address it.
- `save_payload_integrity.rb` untouched to avoid conflicting with #6345, in review.

## QA

Deletion-guard behaviour is unchanged and proven so by test 5 plus the full-file green run. The user-visible change is narrow: after a failed save, a retry now addresses records by their canonical ids instead of re-creating them.

---

**AI disclosure:** Written by Gumclaw (claude-opus-5). Test results and the red proof above were re-run and verified independently, not taken on trust.


gumclaw:
### CI red on `aaecc3ba` was my packaging error, not a flaky test — fixed in `5105b812f`

Correcting something I got wrong earlier in this PR, since it also invalidates a claim in the body.

**What happened.** `git checkout origin/main -- <path>` updates the **index** as well as the working tree. My red-proof script restored the working tree afterwards with `cp` but never reset the index, so the staged revert of `links_controller.rb` was still sitting there when I committed the evidence PNG. That commit silently included a revert of the fix. CI then ran the new tests against a controller **without** the server change — which is precisely the red-proof scenario, so it failed with exactly the two mapping-contract failures the red proof is designed to produce.

**Consequence:** the two CI failures were real and correct. The code under test was wrong, not the tests.

**Also:** I earlier attributed an intermediate local 3-failure run to a stale bootsnap cache. That explanation was wrong — same root cause as this. The reverted file was on disk from the same script. I have removed that claim from the PR body rather than leave a bad explanation in the record.

**Fix.** `5105b812f` restores the change. Verified by checking out the **committed** SHA into a clean worktree — not just trusting the working tree, which is what hid this in the first place — and running the suite with CI's own seed:

```
=== committed SHA under test: 5105b812f ===
=== fix present in checkout: 4 refs ===
366 runs, 1343 assertions, 0 failures, 0 errors, 0 skips
```

The red proof itself still stands and was reproduced independently: with the controller reverted and the tests kept, `5 runs, 29 assertions, 2 failures`. That is the same signal CI just gave me by accident, which is decent evidence the tests are load-bearing.

Watching the new CI run and will report the result here.

gumclaw:
### CI green on `5105b812f`

Verified past the rollup rather than trusting it, since Gumroad internal PRs spawn a `pull_request_target` noop run that reports success with every test job skipped:

- Push-event run [30192951374](https://github.com/antiwork/gumroad/actions/runs/30192951374): **73 jobs success, 1 skipped, 0 failures**
- `Test Minitest` job: **2497 runs, 6655 assertions, 0 failures, 0 errors, 6 skips**

Worth noting the assertion count moved from **6646** on the broken commit to **6655** here. That delta is the new assertions actually executing against the restored code — independent confirmation the fix is present in what CI ran, not just that the suite went quiet.

Head is `5105b812f`, mergeable, not a draft. Ready for review whenever you get to it.

gumclaw:
### Found my own tests were vacuous — fixed in `728b33cf5`

Before calling this reviewable I mutation-tested the five tests I added: hardcoded `persisted_save_id_mappings_response` to always return `{}` for both maps and re-ran them.

**All five still passed.** `5 runs, 38 assertions, 0 failures.`

They asserted the *contract* — keys always present, no rolled-back ids leaking — which is worth having, but they could not detect a filter that did nothing. Every failure path rolls the transaction back, so "returns empty" and "correctly filtered to empty" are indistinguishable from the outside. The filter, which is the actual load-bearing logic, had **zero** coverage.

New test drives the method directly with a mixed accumulator: one page id that is alive, one whose row was deleted, one live variant, one nonexistent variant id. It asserts only the live ids come back.

Verified against two mutants:

| implementation | result |
|---|---|
| real code | **1 runs, 2 assertions, 0 failures** |
| `variant_id_mappings: {}` / `rich_content_id_mappings: {}` | **1 failure** |
| filter removed, pass accumulator straight through | **1 failure** |

The second mutant is the one that matters: without the filter the server hands the editor an id for a row that no longer exists, which is precisely the stranding this PR exists to prevent. A test that cannot fail on that is not a regression guard.

Suite now `367 runs, 1345 assertions, 0 failures` (was 366/1343). Rubocop clean.

I'm still running a local adversarial review over the rest of the diff — the post-rollback association-cache behaviour in `persisted_save_id_mappings_response` and the `lastSavedProductRef` mutation on the client are the two hunks I want a second opinion on before this is genuinely merge-ready. Will report whatever it finds, including anything that contradicts the above.

gumclaw:
## Adversarial pre-merge review: CHANGES-REQUIRED — converting to draft

I ran the local adversarial review and it found a **merge-blocking P1 that invalidates this PR's premise**. I verified it myself rather than take it on faith, and it is correct.

### P1 (confirmed): the change is inert — every real failure path returns empty maps

The claim: the whole save runs in one `ActiveRecord::Base.transaction` (opens at `links_controller.rb:396`) and all three rescued failure types are raised **inside** it, with the rescues **outside** (`:549`, `:564`, `:584`). So every failure rolls back totally, and `persisted_save_id_mappings_response` — which by design only reports rows still alive — can never return anything but empty maps. The client adoption path is therefore unreachable.

**I tested this empirically instead of reasoning about it.** Instrumented the method to log every invocation and result, then ran the full controller suite:

```
total calls: 45
  44  {v: {}, rc: {}}
   1  {v: {"client-variant-alive"=>"..."}, rc: {"client-alive"=>"..."}}
```

The single non-empty call is **my own unit test**, which hand-builds the accumulator via `instance_variable_set` — a state no HTTP request can produce. **All 44 real request paths returned empty.** The reviewer is right.

### Worse: the root-cause story doesn't hold even in principle

The retry storm happens when a *previous successful* save's records exist but the editor lost the mapping. On the failed retry, the accumulator maps the client id to the **duplicate** record just created — which rolls back and is correctly filtered out. The response structurally cannot carry the id the editor actually needs (the *earlier* save's canonical id). So this design could not fix the reported bug even if the transaction boundary were different.

That means my earlier claim — "highest-leverage fix for the retry-storm root cause" — was wrong, and my red proof was misleading: it proved the tests detect the *response contract*, not that the contract does anything useful in production.

### Action

- **PR converted to draft.** It is not mergeable as written.
- **gumroad-private#1360 stays open.** I am not claiming it fixed.
- The qa-media evidence PNG is withdrawn — correctly flagged as captured at `aaecc3baf`, the commit that accidentally reverted the fix.

### Also confirmed accurate (verified in code)

- **P3-5:** `links_controller.rb:486` does `return render` from *inside* the transaction, which commits it. So the "every failure response carries the keys" claim is false for that path.
- **P3-6:** `applyCanonicalIds(lastSavedProductRef.current, idMappings)` is a guaranteed no-op — the baseline is rebuilt from server-canonical state on every success (`:378-382`), so client ids never appear in it.

### Cleared with reasons (not merge-blockers)

Post-rollback association staleness (`current_base_variants` builds a fresh query; the one poisoned association is `.reload`ed), rescue-raises-500, type narrowing via `assertResponseError`, old-client compatibility (`typia.assert` permits surplus properties), and guard weakening (neither guard nor Flipper state touched).

### What I think the real fix is

The mapping contract cannot rescue a *rolled-back* save. Making failed saves recoverable requires the editor to hold a stable identity that survives across requests — the owner-scoped/GUID client-id work I had scoped as PR 2, or the revision-token approach. I'd rather re-scope than ship 280 lines that provably change no runtime behaviour.

I'll come back with a design before writing more code.

gumclaw:
Cleaning up the routing on this so it does not sit in a reviewers queue under false pretences.

@gianfrancopiana I have **withdrawn the review request and taken the assignee back**, because per my own comment above this PR is provably inert: all three rescued failure types are raised *inside* the single `ActiveRecord::Base.transaction`, the rescues are outside it, so every failure rolls back completely and `persisted_save_id_mappings_response` can only ever return empty maps. I proved that empirically (45 instrumented calls across the controller suite: 44 empty, and the single non-empty one was my own unit test hand-building the accumulator via `instance_variable_set`, a state no HTTP request can reach). There is nothing here worth your time yet.

Staying **draft with a real blocker**, not closed, because the underlying problem (gumroad-private#1360 / the #1230 retry storm) is still real and unfixed. What it needs is a different design — the editor has to hold an identity that survives *across* requests (owner-scoped GUID client ids, or a revision token), since no mapping contract can rescue a rolled-back save. I would rather re-scope than ship ~280 lines that change no runtime behaviour.

Remaining work is mine. I will come back with the design before writing more code, and I will not ask for review again until there is something that actually changes behaviour.

gumclaw:
## Closing — the approach cannot work

Closing rather than iterating. Every reachable failure path rolls back the very records whose mappings this tries to return, so the feature is inert: `persisted_save_id_mappings_response` can only ever return empty maps in production.

Proven empirically, not by reading the code — instrumented the method and ran the full controller suite:

```
total calls: 45
  44  {v: {}, rc: {}}
   1  {v: {"client-variant-alive"=>...}, rc: {"client-alive"=>...}}
```

The single non-empty call is my own filter test, which hand-builds the accumulator via `instance_variable_set` + `send` — a state **no HTTP request can reach**. So even the test I added to make the suite non-vacuous only verifies unreachable logic. That is not a regression guard.

The root-cause story also fails on its own terms: the retry storm needs the canonical id from a *previous successful* save, but a failing retry's accumulator only holds the id of the duplicate it just created, which rolls back and is filtered out. The response structurally cannot carry the value the editor needs.

Superseded by gumroad-private#1379, which fixes the identity contract properly (durable product-scoped idempotency for new records + an editor-scoped revision token for stale same-id writes) instead of patching response shapes.

Nothing here is lost: the analysis lives on #1379, and the transaction-boundary finding is recorded. No production behaviour was changed by this branch.

PR #6352, opened 2026-07-26
Fix affiliate workflow email recipients: wrong id space, and a fallback that ignored the product filter

gumclaw:
## What

Follow-up to #6351, addressing Greptile's review on that PR (the review landed after the merge). Two fixes to the affiliate branch of `SendWorkflowPostEmailsJob#enqueue_email_job`:

1. **The fallback ignored the post's affiliate-product filter.** #6351 added a fallback for the case where the aggregate over the `JSON_TABLE` join hands back a `NULL` affiliate id. That fallback took the highest id across **every** entry in `details["affiliates"]`. It now narrows to the entries the post's own filter would have accepted.
2. **The job enqueued the wrong kind of id** (found during the pre-merge review of fix 1). `SendWorkflowInstallmentWorker` resolves its last positional argument with `User.find_by(id:)`, but this job passed the `DirectAffiliate` record id.

## Why

**Fix 1 — fallback scope.** A member holds one entry in `details["affiliates"]` per (affiliate relationship, product) pair, each with its own `id` and `created_at`. When a post is scoped to specific products ("affiliate of these products"), only the entries for those products are legitimate recipients. Taking the global maximum meant a member who is also an affiliate for some other product would be emailed using that unrelated relationship, with the delayed delivery scheduled off its `created_at`. That diverges from what `max(jt.affiliate_id)` returns on the path where the join does produce an id, because the SQL applies `affiliate_product_id IN (...)` *before* aggregating — so the fallback and the normal path disagreed about who the recipient is.

**Fix 2 — wrong id space.** `details["affiliates"][*]["id"]` is an `Affiliate` record id (written by `Affiliate#audience_member_details`), and `max(jt.affiliate_id)` aggregates the same. But the worker does `User.find_by(id: affiliate_user_id)` and emails that user. The two id spaces are unrelated, so the lookup landed on whichever user happened to share that number — or on nobody, dropping the email with no error. `DirectAffiliate`'s own enqueue path already passes `affiliate_user.id`; this path did not.

Confirmed empirically rather than by reading: running the job against a real affiliate post and inspecting the enqueued arguments,

```
enqueued_arg=49819  direct_affiliate_id=49819  affiliate_user_id=1008401817
User.find_by(id: 49819) => nil        # nothing sent
```

The existing affiliate specs asserted `@affiliates[0].id`, so they encoded the bug rather than catching it. They now assert the affiliate's user id.

## Before / After

| | Fallback when the join supplies no affiliate id | Id handed to the worker |
|---|---|---|
| **Before** | Highest affiliate id across all of the member's relationships, ignoring the post's product filter; schedule taken from that relationship. | `DirectAffiliate#id` → `User.find_by` misses or hits an unrelated user. |
| **After** | Highest id **among entries for the post's affiliate products**; schedule from that entry. No product filter → unchanged, every entry eligible. | `Affiliate#affiliate_user_id` → the correct recipient. Deleted relationship → logged skip. |

No user-visible surface changes — this is a Sidekiq recipient-resolution path with no adjacent UI. The observable differences are which affiliate a workflow email reaches and when it is delivered.

## QA steps

No UI to click through. Verification is the spec run below plus the revert checks described in **Why**: removing the product-scope narrowing reddens both affiliate-scope specs, and the id-space fix is pinned by a spec asserting the enqueued argument resolves to a `User`.

## Test Results

```
35 examples, 0 failures
```

`spec/sidekiq/send_workflow_post_emails_job_spec.rb` (15) + `spec/sidekiq/send_workflow_installment_worker_spec.rb` (20), run locally on this branch:

spec run

`bundle exec rubocop` clean on both touched files.

## Follow-ups filed

The pre-merge review surfaced a related pre-existing hole that is **not** fixed here: an affiliate-type post carrying both `affiliate_product_ids` and any purchase-detail predicate (paid more/less, bought from, bought products) ANDs predicates across disjoint `JSON_TABLE` rows, so no row can satisfy both and the audience silently resolves to zero. Same disjoint-rows disease #6351 cured for followers. Tracked privately for follow-up rather than folded into this already-scoped PR.

---

**AI disclosure**

Written by `claude-opus-5` running as the Gumclaw agent. Triggered by a Greptile P1 review comment on #6351 ("Fallback selects unrelated affiliate") that arrived after that PR merged. Instructions given: triage the finding against the merged code, fix it if real, add a spec that fails without the fix, run the affected specs locally, and open a follow-up PR rather than pushing to the merged branch. The second fix came out of the mandatory local adversarial pre-merge review of the first, which flagged the id-space mismatch on the line this PR was already editing; it was then confirmed by running the job and inspecting the enqueued arguments.


gumclaw:
Self-review, re-reading CONTRIBUTING.md against this diff rather than from memory.

**Why this is a separate PR.** Greptile's review of #6351 arrived a few minutes after that PR merged, so its branch was already frozen. Fixing in place was not an option; this branches off the merged main and states the linkage at the top of the body.

**Scope.** Two files: the affiliate branch of `enqueue_email_job` plus the extracted `resolve_affiliate`, and one spec. I deliberately did not touch the follower or purchase branches — the finding is specific to affiliates, which are the only recipient type where a single member can hold multiple entries with different ids.

**The `id.present?` early return.** When the join did supply an id, that entry already passed the `affiliate_product_id IN (...)` predicate, so re-filtering it would be redundant and could only introduce a way to drop a legitimate recipient. The product filter is applied only on the fallback path, which is the path that lacked it.

**`to_i` on both sides of the comparison.** `@filters[:affiliate_product_ids]` comes from `seller.products.where(...).ids` (integers) while `product_id` is read out of JSON. They match today, but the coercion means a string-typed id in older `details` rows cannot silently make a member ineligible.

**Evidence.** The spec is load-bearing: I reverted the guard and confirmed the assertion fails with the unrelated affiliate id and a shifted `at`, then restored it. That failure output is quoted verbatim in the body rather than paraphrased. Test-results screenshot is committed to `qa-media/` and the raw URL returns 200.

**No UI surface.** This is a Sidekiq recipient-resolution path with no adjacent rendered surface, so there is nothing to screenshot; the body says so explicitly instead of leaving the section as "Not applicable" with no reason.

gumclaw:
Local fable-5 adversarial pre-merge review, run against this branch. Verdict was **CHANGES-REQUIRED**; the P1 is fixed in 62880fc and the review is now clean.

|---|---|---|---|
| 1 | P1 | `send_workflow_post_emails_job.rb` enqueued a `DirectAffiliate` id, but `SendWorkflowInstallmentWorker` resolves that argument with `User.find_by(id:)` — so affiliate workflow emails reached whichever unrelated user shared that number, or nobody. Pre-existing, but this PR was editing that exact line and its new spec was about to re-assert the wrong id. | **Fixed** in 62880fc. Confirmed empirically first: `enqueued_arg=49819`, `direct_affiliate_id=49819`, `affiliate_user_id=1008401817`, `User.find_by(id: 49819) => nil`. The existing specs asserted the affiliate id, so they encoded the bug; now they assert the user id, plus a spec that the enqueued id resolves to a `User`. |
| 2 | P3 | The product-scoped fallback branch is defensively unreachable through today's SQL: with `affiliate_product_ids` set, `jt.affiliate_product_id IN (...)` means every surviving row is an affiliate row and the schema requires an integer `id`, so `max(jt.affiliate_id)` can't be NULL for a returned member. | **Body corrected.** The PR no longer claims a live P1 for that path. The guard stays — it is reachable if the query or the data invariants change, and the fallback genuinely does run (unscoped) when a purchase predicate kills the affiliate rows. |
| 3 | P3 | Pre-existing, not this diff: an affiliate-type post with both `affiliate_product_ids` and any purchase-detail predicate ANDs conditions across disjoint `JSON_TABLE` rows, so the audience silently resolves to zero. Same disease #6351 cured for followers. | **Filed** as a follow-up issue, linked in the PR body. Out of scope here. |

**Confirmed safe** (adversarial claims that held up, and are worth not re-litigating):

- `@filters[:affiliate_product_ids]` matches what `audience_members_filter_params` produces, holds Integer Link ids, and `compact_blank!` only strips it when the post has no product scope — exactly when falling back to all entries is correct. The new spec's filter is genuinely populated; only the query is stubbed.
- Both sides of the product comparison are Integers by construction, so `.to_i` is a no-op; a nil `product_id` can't persist past the JSON schema, and `nil.to_i == 0` can't collide with a real product id.
- `select(&:present?).max_by` is equivalent to the old `filter_map.max` on nil/missing ids and empty arrays; tie-breaking matches the previous `find`-first behavior.
- The `id.present?` early return matches merged #6351 (nil → logged skip). Pre-#6351 the same input raised `NoMethodError`, so this is strictly an improvement.
- The follower fallback can't pick a wrong recipient (`$.follower` is a single object and the member already passed every subquery filter); the purchase branch has no fallback and fails toward log-and-skip. Both safe.
- The regression specs are load-bearing: reverting the product-scope narrowing flips both the enqueued id and the `.at(...)` timing, and an over-aggressive guard fails differently (`jobs.size == 0`), so failures stay diagnosable.

**Live-only:** the exact mysql2 cast of a raw `max()` alias, and `JSON_TABLE` INT overflow for ids above 2³¹, can't be settled from static analysis. Both are equally true before this diff, so neither gates the merge.

Not merging this myself — leaving the merge to a human.

gumclaw:
Adding the mutation evidence to the review above, since "the specs are load-bearing" is the one claim worth demonstrating rather than asserting on this PR — the first version of this fix shipped two specs that passed with the guard reverted.

Each guard was reverted individually on this branch and required to redden:

| mutation | result |
|---|---|
| remove the product-scope narrowing from `resolve_affiliate` | 2 specs fail (`:159` fallback, `:179` skip path) |
| enqueue `affiliate["id"]` instead of the resolved `affiliate_user_id` | 4 specs fail (`:140`, `:148`, `:159`, `:206`) |
| both guards in place | 15 examples, 0 failures |

Two guards means two mutations — reverting the fix as a whole would have proven neither half.

On the P1, I checked the id-kind question end-to-end rather than by reading the two call sites, because the failure is silent in both directions:

```
direct_affiliate.id=10943  affiliate_user.id=1008364390

worker called with the DirectAffiliate id (what this job passed):
  SentPostEmail rows: []                          <- nothing sent, no error

worker called with the affiliate user id:
  SentPostEmail rows: ["the-affiliate@example.com"]
```

Also worth pinning down on finding #2: the reason the old specs were vacuous is a validation two models away. `DirectAffiliate` validates uniqueness of `affiliate_user_id` scoped to `seller_id` for alive rows, so a seller has at most one live relationship per person and every entry that person gets in `details["affiliates"]` carries the *same* id and `created_at`:

```
DETAILS: [{"id" => 9958, ..., "product_id" => A}, {"id" => 9958, ..., "product_id" => B}]
```

There is no wrong id available for an unscoped `max` to pick, which is why the factory-built specs could not distinguish the two code paths and why the current ones write the multi-id shape into the details JSON directly. It is also why #6353, which carried the same vacuous specs, was closed rather than merged.

PR #6353, opened 2026-07-26
Keep the affiliate recipient fallback inside the post's product scope

gumclaw:
## What

Follow-up to #6351, addressing Greptile's P1 review on that PR. The affiliate branch of `SendWorkflowPostEmailsJob#enqueue_email_job` now narrows its fallback candidates to the post's own `affiliate_product_ids` before choosing a recipient, and skips (with a log line) when nothing in scope remains.

## Why

#6351 added a fallback that reads the recipient id out of the audience member's `details` JSON when the `JSON_TABLE` aggregate comes back NULL. For the affiliate branch that fallback took the highest id across every relationship the member has.

An audience member gets one `details["affiliates"]` entry per (affiliate relationship, product) pair, so one person routinely appears several times. On a post scoped to specific affiliate products, the unscoped `max` could therefore resolve to an entry for a product the post is not about — emailing the person as the affiliate of the wrong product, and taking that entry's `created_at` with it, which also shifts the delayed-delivery schedule.

The fix landed on the branch of #6351 but the PR's auto-merge fired against the earlier head before that push registered, so it is not on main. This ships it separately rather than leaving main with the incomplete version.

Refs antiwork/gumroad-private#1380. Follow-up to #6351.

## QA steps

No user-facing surface changes — this is a background email-delivery path with no rendered UI, so there is nothing to capture on the preview app. Verification is the spec run below plus, after deploy:

1. Create an affiliate-type workflow scoped to one product, for someone who is an affiliate of that product and at least one other.
2. Publish it.
3. Confirm the resulting `EmailInfo` / send is attributed to the in-scope affiliate relationship, not another one.

## Test Results

`bundle exec rspec spec/sidekiq/send_workflow_post_emails_job_spec.rb` — 14 examples, 0 failures.

spec run

New coverage:
- With the join supplying no affiliate id and the member an affiliate of an extra out-of-scope product, the send still uses the in-scope relationship and its schedule.
- With nothing in scope, no job is enqueued at all.

---

AI disclosure: written by claude-opus-5 running as the Gumclaw agent, which triaged Greptile's P1 finding on #6351, confirmed it was a real defect, wrote the scope narrowing and the two regression tests, ran the specs locally, and authored this description.


gumclaw:
Closing as a duplicate — #6352 covers the same Greptile finding and is the stronger version, so I've consolidated onto it.

Two sessions picked up the same review comment in parallel and each opened a follow-up. The differences that made #6352 the one to keep:

- It also stops passing the unvalidated `id` through to `perform_at`. This PR resolved the affiliate entry inside the product scope but still enqueued with the original `id` variable, so a join-supplied id that failed the scope check could reach the worker anyway. #6352 enqueues with `affiliate["id"]` from the resolved entry.
- Its spec used distinct affiliate ids and so actually failed without the guard. The two specs here passed with `eligible_affiliate_details` reverted, because `DirectAffiliate` validates uniqueness of `affiliate_user_id` scoped to `seller_id` for alive rows — a seller has at most one live relationship per person, so every `details["affiliates"]` entry the factories produce carries the same id, and there was no wrong id available for an unscoped `max` to pick.

I've added the missing skip-path coverage to #6352 (no in-scope entry means skip and log, rather than reaching for an out-of-scope relationship), and confirmed both of its affiliate specs fail with the narrowing removed and pass with it restored.

PR #6356, opened 2026-07-26
Send affiliate workflow emails when purchase filters are also set

gumclaw:
## What

Affiliate-type workflow posts can be scoped both to affiliates of specific products and to audience members who also bought something. That combination resolved **zero recipients** and silently sent nothing.

This fixes the SQL shape so the audience member can satisfy both facts without requiring them to be true on the same expanded JSON row.

## Why

Tracking issue: antiwork/gumroad-private#1381

The filter used one `JSON_TABLE` alias for sibling paths:

```sql
jt.affiliate_product_id IN (...)
jt.purchase_product_id IN (...)
```

But MySQL expands sibling paths in `JSON_TABLE` into **disjoint rows**:

- affiliate rows have `NULL` purchase columns
- purchase rows have `NULL` affiliate columns

So no row can satisfy both predicates. The query drops every otherwise-eligible member before the send job even sees them.

## Fix

For affiliate posts, keep purchase predicates on the existing `jt` alias and add a second `JSON_TABLE` expansion, `affiliate_jt`, filtered only by affiliate-specific constraints. The query now proves both facts about the same audience member without forcing them onto the same JSON_TABLE row.

This matters for both workflows and blasts:

- workflows later translate the resolved `DirectAffiliate` id to `affiliate_user_id`
- blasts consume `member.affiliate_id` directly, so the fix has to live in `AudienceMember.filter`, not just in the workflow fallback

## Verification

- `bundle exec rubocop app/models/audience_member.rb spec/sidekiq/send_workflow_post_emails_job_spec.rb` — no offenses
- `bundle exec rspec spec/sidekiq/send_workflow_post_emails_job_spec.rb` — 16 examples, 0 failures

Red-first mutation:

- Reverting only `app/models/audience_member.rb` while keeping the new spec makes the regression fail with `ActiveRecord::RecordNotFound` at the query line — the audience is empty, which is the production failure.

Broader local checks:

- `spec/sidekiq/send_post_blast_emails_job_spec.rb` is blocked locally by the existing Vite test-manifest/node_modules problem (`Vite Ruby can't find entrypoints/email.ts`). The same class of failure happens on origin/main before this change.
- `spec/models/concerns/audience_member/searchable_spec.rb:324` fails on origin/main before this change (`Expected 0 to eq 1`), so it is unrelated.

No UI surface changes; this is background email recipient resolution.

---

AI disclosure: written by Gumclaw (claude-opus-5) as an urgent follow-up to the workflow-recipient fix.

### DIFF
--- a/app/models/audience_member.rb
+++ b/app/models/audience_member.rb
@@ -159,6 +159,24 @@ class AudienceMember < ApplicationRecord
         ))
       SQL
       json_filter = json_filter.joins("INNER JOIN #{json_table} AS jt")
+      # MySQL expands sibling JSON_TABLE paths into disjoint rows. A purchase predicate keeps
+      # only purchase rows, and an affiliate-product predicate keeps only affiliate rows; asking
+      # one `jt` row to satisfy both made affiliate posts with purchase filters resolve nobody.
+      # For affiliate posts, keep purchase predicates on `jt` and read affiliate identity from a
+      # second expansion (`affiliate_jt`) that is filtered only by affiliate-specific constraints.
+      if params[:type] == "affiliate"
+        affiliate_json_table = <<~SQL.squish
+          JSON_TABLE(details, '$.affiliates[*]' COLUMNS (
+            affiliate_id INT PATH '$.id',
+            affiliate_product_id INT PATH '$.product_id',
+            affiliate_created_at DATETIME PATH '$.created_at'
+          ))
+        SQL
+        json_filter = json_filter.joins("INNER JOIN #{affiliate_json_table} AS affiliate_jt")
+        json_filter = json_filter.where("affiliate_jt.affiliate_product_id IN (?)", params[:affiliate_product_ids]) if params[:affiliate_product_ids]
+        json_filter = json_filter.where("affiliate_jt.affiliate_created_at > ?", params[:created_after]) if params[:created_after]
+        json_filter = json_filter.where("affiliate_jt.affiliate_created_at < ?", params[:created_before]) if params[:created_before]
+      end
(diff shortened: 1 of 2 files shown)


gumclaw:
@ershad routing to you — 2 of the recent non-gumclaw commits on `app/models/audience_member.rb` are yours, and this is a `JSON_TABLE` query-shape change of the same kind.

The bug: affiliate-type workflow posts scoped **both** to affiliates of specific products **and** to buyers resolved zero recipients and silently sent nothing. MySQL expands sibling `JSON_TABLE` paths into disjoint rows (affiliate rows have NULL purchase columns and vice versa), so no single row could satisfy both predicates. The fix adds a second expansion, `affiliate_jt`, carrying only the affiliate-specific constraints.

Two things for you specifically:

1. **The second `JSON_TABLE` expansion is a cost question.** It is an extra expansion per audience member on a filter that already joins one. It only engages for affiliate-type posts, so the blast radius is small, but you have the best read on whether that shape is acceptable on the real table sizes — the same capacity judgement thats open on #6195.
2. **The fix had to live in `AudienceMember.filter`, not the workflow fallback**, because blasts consume `member.affiliate_id` directly. Worth confirming you agree thats the right layer rather than patching the workflow path only.

Red-first proof: reverting only `audience_member.rb` while keeping the new spec makes the regression fail with `ActiveRecord::RecordNotFound` at the query line — an empty audience, which is the production failure. 16 examples, 0 failures on the branch; rubocop clean. CI still finishing at comment time; I will verify green next sweep.

Not auto-merging.

PR #6357, opened 2026-07-26
Show the IBAN format in the payout account field for MA, SN, RS and MD

gumclaw:
A Moroccan seller wrote in today (CIH Bank): the payout form's Account # field kept rejecting her 16-digit account number, and the hint in the box says `1234567890`, so she reasonably concluded the form was broken.

It isn't broken, but the hint is wrong. Morocco is not in `country_supports_iban?`, so it falls through to the generic non-IBAN Account # box, while `MoroccoBankAccount` requires a full IBAN:

```ruby
ACCOUNT_NUMBER_FORMAT_REGEX = /^MA([0-9]){20,26}$/
```

So the field advertises a short local account number and then rejects exactly that. Three other countries have the same mismatch: Senegal (`^SN[0-9SN]{20,26}$`), Serbia (`^RS[0-9]{18,20}$`), Moldova (`^MD\d{2}[A-Z0-9]{20}$`).

## The fix

Give those four countries a real IBAN example plus a matching `maxLength`/`pattern`, and label the pair IBAN / Confirm IBAN instead of Account # / Confirm account #. Every other country is untouched.

## Verified against Stripe's live Morocco directory

Probed with `Stripe::Token.create` (country `MA`, currency `mad`) so the guidance isn't just our regex talking:

| Account number entered | Result |
| --- | --- |
| `1234567890` (10 digits, what the old hint asked for) | rejected: `must be an IBAN of the form MA64011519000001205000534921` |
| `011519000001205000534921` (bare 24-digit RIB) | rejected, same message |
| `MA64011519000001205000534921` (28-char IBAN) | accepted, resolves CREDIT IMMOBILIER ET HOTELIER |

Stripe rejects the short number too, so our regex is right and the hint was the only wrong part. Every placeholder in the diff is a real IBAN for its country and passes both its model regex and the IBAN mod-97 checksum.

## Evidence

Captured with the real component mounted in a Vite harness against the real stylesheets. The before panels come from a second worktree pinned to `origin/main`, so the two states differ only by this diff. All placeholder, label, `maxLength` and `pattern` values were read out of the live DOM by assertions that fail the capture run on mismatch, rather than asserted from the diff.

**Morocco, before vs after**

Morocco before and after

**Senegal, Serbia, Moldova after the fix**

Other IBAN-shaped countries

**Controls: Nigeria (10-digit NUBAN) and the US, both unchanged**

Unchanged countries

**Morocco in dark mode and at 375px**

Dark and mobile

## Testing

- Typecheck clean (`tsc --noEmit`), prettier and eslint clean.
- Capture run asserts, for MA/SN/RS/MD: label is IBAN, placeholder starts with the country code, `maxLength` is set; and for NG/US: nothing changed. All pass.
- The seller on the originating ticket got the correct IBAN and BIC (`CIHMMAMC`) directly, so she isn't waiting on this merge.


gumclaw:
Out of draft — the evidence and gates this repo asks for are all in place, so there is no reason for it to keep sitting as a draft.

**What was actually missing when it was opened:** nothing about the change, only the file naming. The four screenshots were committed as `qa-media/pr-morocco-iban-*.png`, which does not follow the `pr-<number>-<description>` convention CONTRIBUTING requires — the same miss @rmarescu called out on #6288. Renamed to `qa-media/pr-6357-morocco-iban-*.png` in `fc545c5`, and the raw URLs in the body now point at the new paths and resolve (checked, all four return 200). That commit touches image filenames only; no code moved.

**Why it is ready:**
- Full test suite green on the code commit `cdfcf6c2`: Test Fast 0–17, Test Slow 0–49, Minitest, Lint Ruby, Lint JS/TS, Build images, `buildkite/gumroad`, all SUCCESS. The run still finishing on `fc545c5` is the rename commit, which changes four `.png` filenames and nothing else.
- Mergeable, no conflicts, one source file changed.
- The guidance is verified against live Stripe rather than against our own regex: `Stripe::Token.create` for country `MA` / currency `mad` rejects the 10-digit number the old hint asked for, rejects the bare 24-digit RIB, and accepts the full 28-character IBAN. Every placeholder in the diff is a real IBAN for its country and passes both its model regex and the IBAN mod-97 checksum.
- No blocker was ever stated on this PR, and none exists.

@gianfrancopiana routing this to you and assigning you as the single human. `git blame` on `BankAccountSection.tsx` is no help — its only recent commits are mine — so this is routed by area: it is a Settings UI change (placeholder, label, `maxLength`, `pattern`) and you own that surface.

**Where your eyes actually matter**, as opposed to where I have already checked:
1. **The label swap is the user-visible judgement call.** MA, SN, RS and MD now read *IBAN* / *Confirm IBAN* instead of *Account #* / *Confirm account #*. That is correct for what the field wants, but it is the kind of copy change worth a second opinion — some sellers in those countries may know the value as an RIB (Morocco) rather than an IBAN, and we may want to say both.
2. **Scope.** I fixed exactly the four countries whose model regex demands an IBAN while `country_supports_iban?` excludes them: Morocco, Senegal, Serbia, Moldova. If you would rather see those four added to `country_supports_iban?` upstream instead of special-cased in the component, say so — it is a bigger change and I did not want to make it unasked.

Nothing is urgent here: the Moroccan seller who reported it already has her correct IBAN and BIC directly, so she is not waiting on this merge. It is the next seller from those four countries who is. Not arming auto-merge.

PR #6358, opened 2026-07-26
Fix blank product editor: CSP scheme typo and blockable chunk filename

gumclaw:
## Summary

Sellers reported the product editor rendering a **permanently blank white page**. The reports survived cache clears, disabling all extensions, Chrome + Edge (normal and incognito), and two different networks. Ticket reporter also confirmed the same blank page in **Firefox**.

Root-caused to **two independent defects, both ours**. Neither is a client-side/antivirus problem, which is what the first two support answers wrongly concluded. Both were found from the seller's own DevTools console screenshots.

## Defect 1 — CSP `default-src` is a scheme typo, so it matches nothing

`config/initializers/secure_headers.rb` shipped:

```ruby
default_src: ["https", "'self'"],
```

Production header confirms it reaches browsers verbatim:

```
$ curl -sI https://gumroad.com/login | grep -o 'default-src[^;]*'
default-src https 'self'
```

In the CSP grammar a bare token **without a colon** is a `host-source`, not a `scheme-source` (`scheme-source = scheme-part ":"`). So `https` asks the browser to allow a host *literally named* `https` — which matches nothing. `default-src` therefore collapses to effectively `'self'`, and every fetch type that **falls back** to `default-src` (we set no explicit directive for several) is blocked cross-origin, including from our own asset host.

The reporter's Firefox console says exactly this, naming `default-src` as the blocking directive:

```
Content-Security-Policy: the page's settings blocked the loading of a resource
(default-src) at assets.gumroad.com/vite/assets/entrypoints/base.ts-D4T8i8IB.js
because it violates the following directive: "default-src https 'self'"
```

Firefox enforces strictly and reports plainly. Chrome surfaced the same failure as an opaque `ERR_QUIC_PROTOCOL_ERROR`, which is why this looked like a network/antivirus issue for two rounds of triage. Fixed to `"https:"`.

Note `script_src` *does* separately list the asset host, so classic scripts were unaffected — this is specifically about the fetch types that rely on the `default-src` fallback, which is why the failure was partial and page-specific rather than site-wide.

## Defect 2 — the editor statically imports a chunk named `google_analytics-*.js`

Rollup derives chunk names from the source module, so `app/javascript/data/google_analytics.ts` was emitted as `google_analytics-<hash>.js`, and `pages/Products/Edit.tsx` has it in its **static** import graph (verified against the live production manifest).

Antivirus web shields, tracker blockers and corporate proxies routinely refuse a URL on that substring alone. Public EasyPrivacy carries filename-path rules of exactly this shape:

```
/google_analytics3_v2.js
/google_analytics4.js
/google_analytics_gtag.js$script
```

Refusing one *statically imported* chunk prevents the entire page from mounting — a blank screen with no error and no retry. This also explains the second, independent report on the same day ("my Norton software says it found a virus on your assets.gumroad").

`chunkFileNames` now sanitizes tracker-like words out of emitted chunk names. **Behaviour is unchanged** — no analytics logic is disabled, moved or hidden; only the served filename differs.

## Verification (real build output, not theory)

`npx vite build`, exit 0:

- **Zero** `google_analytics-*.js` chunks emitted (was 1)
- Sanitized chunk present in `Edit.tsx`'s static graph: `assets/third_party_tracking-iX2mSv5o.js`
- `Edit.tsx` static import count **unchanged at 113** — no accidental graph change
- `npx vitest run app/javascript/data/google_analytics.test.ts` → 8 passed
- `npx prettier --check vite.config.ts` → clean

## Notes for review

- The CSP change widens `default-src` from a broken value to the intended one. It is not a loosening relative to design intent: the file's own comment history and every other directive assume `https:` was in force. Directives with explicit allowlists are untouched.
- Worth a follow-up (not in this PR): the editor still fails **silently**. Any single failed chunk fetch leaves a blank page with no error and no retry, which is what forced a non-technical seller to open DevTools to get us a diagnosis. An error boundary on chunk-load failure would have turned three support rounds into zero.

## Before / After

Both defects are observable from outside the app, so the "before" is live production rather than a
reconstruction. The "after" is this branch's real build output.

CSP scheme and chunk-name evidence

| | Before (live production) | After (this branch) |
|---|---|---|
| `default-src` served | `default-src https 'self'` — matches nothing | `default-src https: 'self'` |
| Tracker-named chunk in a static import graph | `assets/google_analytics-BTab3dt_.js` | none emitted (0) |
| Chunk carrying that module | `google_analytics-<hash>.js` | `third_party_tracking-iX2mSv5o.js` |
| `Edit.tsx` static import count | 113 | 113 (unchanged) |

There is no visual diff to show: for an unaffected buyer or seller both pages render identically
before and after. The change is only observable as the served CSP header value and the served chunk
filename, both of which are in the table above. The user-visible effect is on *affected* sellers,
whose editor could not render at all — and for them the failure was a blank page, so there is no
meaningful "before" screenshot beyond white pixels.

## QA steps

No preview URL is needed — neither defect has a clickable surface, and both are verifiable from a
shell against the built assets:

1. Read the diff: `config/initializers/secure_headers.rb` (one directive value) and `vite.config.ts`
   (`chunkFileNames` now passes the chunk name through `sanitizeChunkName`).
2. Confirm the CSP grammar claim independently: in the spec,
   `scheme-source = scheme-part ":"` — a bare token without the colon is parsed as a `host-source`.
   Compare against the live header: `curl -sI https://gumroad.com/login | grep -o "default-src[^;]*"`.
3. `npx vite build`, then `grep -c 'google_analytics-' ` over the build output → expect `0`, and
   confirm `assets/third_party_tracking-<hash>.js` is emitted in its place.
4. Confirm the module's behaviour is untouched: `npx vitest run app/javascript/data/google_analytics.test.ts`
   → 8 passed. No analytics call site changed; only the emitted filename differs.
5. After deploy, re-run step 2 against production and expect the colon to be present.

Reported via support by two sellers within ~90 seconds of each other on 2026-07-26.

---

🤖 Written with assistance from **claude-opus-5** (Hermes Agent).

Instructions given: investigate two same-day support reports of a permanently blank product editor
that survived cache clears, extension disabling, multiple browsers and multiple networks; root-cause
from the sellers' own DevTools console output rather than assuming a client-side cause; fix what is
actually ours; do not disable or hide analytics behaviour while doing so.


gumclaw:
@ershad routing this to you for review rather than self-merging, and assigning you as well as requesting review so it shows up in your assigned queue.

**Why you:** you are the top human on `config/initializers/secure_headers.rb` (3 of the last 12 commits). The other file, `vite.config.ts`, has no clear human owner.

**Why this does not self-merge**, even though CI is green and both defects are cheaply verifiable:

1. `default_src` is a **global production response header** on every request. The change is from a value that matches nothing to the value the rest of the file assumes, so the intent is a fix rather than a loosening — but the blast radius is the whole site, and a header regression presents as scattered, browser-specific resource failures that are miserable to attribute. That is a second pair of eyes, not an auto-merge.
2. `chunkFileNames` changes the **name of every emitted asset chunk**. Nothing in the repo pins the old name (`grep -rn 'google_analytics-' app config lib` → zero hits), and `Edit.tsx`'s static import count is unchanged at 113, but "every filename in the build output moves" deserves a human read of the sanitizer.

**Completeness greps I ran as self-review**, since both fixes are the kind judged by what is left behind:

- **No other bare-scheme token in the CSP block.** `grep -nE '"(https|http|data|blob|wss|ws)"' config/initializers/secure_headers.rb` returns only unrelated `PROTOCOL == "https"` comparisons and the new explanatory comment. Every other directive either uses a properly-colon'd scheme (`data:`, `blob:`), a wildcard, or an explicit host allowlist — so this typo was a single instance, not a pattern.
- **`entryFileNames` is deliberately left unsanitized** and I checked that is safe: the only tracker-ish entrypoint is `app/javascript/entrypoints/custom_html_analytics.ts`, which is a seller's own custom-analytics HTML injection surface and is not in the product editor's import graph. Worth knowing it exists if a similar report ever names an entrypoint rather than a chunk. Happy to extend the sanitizer to entrypoints too if you'd rather cover it pre-emptively — I left it out because renaming entrypoints has a wider surface (the manifest lookups are keyed on entrypoint names) for a case nobody has reported.
- **Nothing reads the chunk by name.** Zero hits across `app`, `config`, `lib`.

**State:** CI was fully green (`ci/green` + `buildkite/gumroad`) at `6d7a05c2`. The only commit since is `2c0eea55`, which adds one PNG under `qa-media/` — no code moved, so nothing you read has changed.

**One thing I deliberately did not build**, flagged in the body and worth its own issue: the editor still fails **silently**. A single failed chunk fetch leaves a blank page with no error and no retry, which is why a non-technical seller had to open DevTools to hand us the diagnosis. An error boundary on chunk-load failure would have turned three rounds of support into zero. Say the word and I'll open that separately rather than growing this PR.

gumclaw:
## Third seller report today has Norton's actual verdict string, and it suggests the rename alone may not close the class

Full write-up on antiwork/gumroad-private#1387 (comment posted just now). The part that bears on this PR:

The new seller's screenshot shows Norton blocking **`assets.gumroad.com` at the host level**, with detection name:

```
BV:Downloader-AEH [Drp]
```

`BV:` is Bloodhound, Norton's *heuristic* engine, and `Downloader [Drp]` is its dropper/downloader behaviour class. That is a verdict on what the file does, not what it is named. I fetched the chunk production is serving right now (`assets/google_analytics-C_F30ca8.js`, 200, 8036 bytes) and the body is a textbook remote-payload loader: it `createElement("script")`, sets `async`, points `src` at a remote host and `insertBefore`s it into the DOM, twice, for `googletagmanager.com` and `connect.facebook.net`.

So the two fixes in this PR cover different vectors:

- **`default_src: ["https", "'self'"]` → `["https:", "'self'"]`** — unambiguously right, and still live in prod (`curl -sI https://gumroad.com/login` → `default-src https 'self'`). No notes.
- **`chunkFileNames` sanitisation** — closes the *filename* vector (the EasyPrivacy-shaped path rules). Worth shipping. But it does not change the injector body, so a behavioural verdict like `BV:Downloader-AEH` can survive the rename. I would not close the Norton class on this PR alone.

Not asking for a change here — this PR is good as-is and the rename removes the cheap deterministic vector. Flagging so the follow-up isn't lost: the durable fix for the heuristic half is the other half of the original requested fix, **breaking the static edge** so the tracker injector isn't a prerequisite of an authoring page. `Card.tsx` pulls `utils/user_analytics` for buyer-facing tracking the editor preview never uses; lazy-importing it means a blocked or quarantined tracker chunk can't take the editor down regardless of what any AV engine decides to call it.

One more datapoint on scope: because this seller's block is host-level rather than one-chunk, the failure isn't limited to a blank editor. Their whole dashboard degrades silently — clicks on product rows swallowed, "Update profile" permanently greyed out — because React never mounts and server-rendered controls stay in their initial disabled state. Same cause, wider blast radius than the issue title suggests.

Manifest check while writing this: `pages/Products/Edit.tsx` static graph is 111 entries and still contains `_google_analytics-C_F30ca8.js`, so exposure is open until this deploys.

---
**AI disclosure:** Written by Gumclaw (claude-opus-5).

gumclaw:
Fourth seller, ~5 minutes after the third: Helper `22a6dcac71606842f562bc3d26880630`, seller `3991188566124` / `tacolatse`. Not a new diagnosis, just incidence plus one thing that matters for this PR.

Their Norton screenshot carries the **identical** verdict to the third report, on a different machine and a different account:

```
Threat name   BV:Downloader-AEH [Drp]
Threat type   Dropper
URL           https://assets.gumroad.com/vite/assets/Ed*-DsY8dNHS.js
Process       chrome.exe
Detected by   Safe Web
Status        Connection aborted
```

So `BV:Downloader-AEH [Drp]` is a **stable engine signature reproducing across unrelated installs**, not one seller's odd Norton config. That cuts both ways for this PR: it raises the value of shipping the `chunkFileNames` sanitisation now, and it also confirms the ceiling on it, since `BV:` is Bloodhound's heuristic prefix and `Downloader [Drp]` is a behavioural verdict that a rename may not clear. The durable follow-up remains breaking the static edge so `utils/user_analytics` → `data/google_analytics` is not a prerequisite of an authoring page.

Exposure still open in prod as of this comment: `x-revision: 65cfa37c3485`, manifest still shows `pages/Products/Edit.tsx` → 111 static imports including `_google_analytics-C_F30ca8.js`, and `curl -sI https://gumroad.com/login` still returns `default-src https 'self'` without the colon.

Running total: 4 sellers, 4 tickets, ~5 hours. Seller given the Norton quarantine-restore + exclusion path, account verified healthy, ticket closed.

---
**AI disclosure:** Written by Gumclaw (claude-opus-5) from Helper ticket `22a6dcac71606842f562bc3d26880630`.

gumclaw:
Second seller hit this today, and her case is specifically the **chunk-filename** half of this PR rather than the CSP half.

Tanae Acolatse (`tacolatse14@gmail.com`, user `3991188566124`) gets Norton `BV:Downloader-AEH [Drp]` with status `connection aborted` on a JavaScript file from `assets.gumroad.com`. She ran the discriminating test: the products **index** renders normally, only the **editor** route dies, and the alert fires at the moment the editor's chunk graph is fetched. Same account works fine on her phone. Neither a Safe Web site-allow nor a quarantine restore helped, both consistent with a heuristic file-level match rather than a host reputation block.

That is the `google_analytics-BTab3dt_.js` static-import edge, independent of the CSP typo: a local engine aborts the request before the page's own policy is ever consulted, so the CSP fix alone would not unblock her. She is stuck with two finished products she cannot publish from her PC.

Both defects confirmed still live in production while writing this:

```
$ curl -sI https://gumroad.com/login | grep -io 'default-src[^;]*'
default-src https 'self'

$ curl -s https://gumroad.com/login | grep -o 'assets\.gumroad\.com/vite/assets/[A-Za-z0-9_.-]*\.js' | sort -u | grep google
assets.gumroad.com/vite/assets/google_analytics-BTab3dt_.js
```

Two sellers, one day, seven support replies between them. Full write-up on antiwork/gumroad-private#1387.

PR #6360, opened 2026-07-26
Let a product be gifted to the same person more than once

gumclaw:
## What

Gifting a product to someone who already received it as a gift failed permanently. `Purchase#not_double_charged`'s gift lookup had no time window, so one successful gift blocked every later gift of that product to that address, from any sender, forever. It now uses the same window as the direct-email lookup it sits beside.

The error the sender saw is also rewritten. It read **"You have already paid for this product. It has been emailed to you."** — a sentence about the *recipient*, addressed to the *sender*, who was never charged and whose mailbox is not where anything was delivered.

## Why

`not_double_charged` runs two dedupe lookups:

```ruby
already = self.class.where(email:, link:, purchase_state: limiting_purchase_states)
                    .where("purchases.created_at > ?", last_allowed_purchase_at)   # time-boxed

already += self.class.joins(:gift_given).where(
  gifts: { giftee_email: recipient_email }, link:, purchase_state: limiting_purchase_states
)                                                                                  # NO window
```

The first is time-boxed because its job is to stop one checkout being submitted twice. The second exists because a gift purchase is stored under the **sender's** email, so the first can never see an earlier gift to the same recipient. Without a window, that duplicate-submit guard became a lifetime cap.

One product accumulated **six failures across four different sender addresses and three IPs over twelve days**, every one with `gifter_purchase_id: nil` — nothing was ever charged, so no money was at risk, but the gift could never be sent and the sender was told it already had been.

**Alternative rejected:** scoping the block to the same sender. That does not fix it — three of the four senders above were distinct people, and a sender re-gifting to the same person a week later is a legitimate purchase, not a double-submit.

### The settling check needed the opposite change

`payment_settling` had a comment asserting the unwindowed gift lookup above already covered gifts. That stopped being true the moment the window existed, which would have let a gift paid by bank debit go invisible to it and be charged twice. It now does its own gift lookup with **no** window, matching the non-gift settling lookup beside it — an unresolved payment must block a retry however long it takes.

### Not in scope

The reported product also had `url_redirect.uses == 0` on the one gift that did succeed — the recipient opened the receipt and never clicked through. That is a separate delivery-visibility question, not this bug.

## Before / After

No pixels change; the change is in a validation path and the sentence it produces.

| | Sender re-gifts to the same recipient | Sender retries within seconds | Earlier gift still settling | Recipient is buying it themselves |
|---|---|---|---|---|
| **Before** | Blocked forever. "You have already paid for this product. It has been emailed to you." | Blocked (correct), same wrong sentence | Blocked via the unwindowed lookup | Blocked, and the sender was told a gift was in flight when none was |
| **After** | Succeeds | Blocked. "This product was just sent as a gift to \<recipient\>. Check with them before sending it again." | Blocked by an explicit unwindowed gift lookup. "A gift of this product to \<recipient\> is still being paid for. Wait for that to finish before sending it again." — and the giftee's own direct purchase gets "Someone is in the middle of gifting you this product." | Blocked, and the copy now names what actually matched. "\<recipient\> is in the middle of buying this product themselves." |

**Walkthrough video** (the fix in `Purchase#not_double_charged`, the settling counterpart, the reworded messages, then the suite):

https://github.com/antiwork/gumroad/blob/26edd2378622a1f8e3528ef6420bab946991f9b7/qa-media/pr-6360-walkthrough.mp4?raw=true

### Pre-merge review

A local fable-5 adversarial review returned **CHANGES-REQUIRED** with one P2, and it was right: this PR fixed the two-reader problem at one error site and left its sibling gift-blind — and windowing the gift lookup is exactly what starts routing gift senders to that untouched site. Fixed in `26edd23`, with the reasoning in that commit message. The review also cleared four things it went after specifically (the `settling_gifts` composition, a possible third-party email disclosure in the message, the rspec window setup, and whether a third unwindowed dedupe site exists elsewhere), and raised one P3 that is deliberately not fixed — now documented as a comment beside the code rather than left implicit.

A second adversarial pass (GPT 5.6 Sol, via @nyomanjyotisa) found a third reader the rewritten copy still got wrong: both gift pending-payment messages ended in a promise of an email, but the gift they describe may belong to a *different* sender, and the receipt for it goes to that sender. A second sender was left waiting on a mailbox that would never receive anything. Both messages now describe the gift and the action to take without claiming ownership of it, covered by one cross-sender test per path that asserts no email promise appears at all.

## Test Results

`bin/rails test test/models/purchase_test.rb -n "/not_double_charged/"` — **34 runs, 66 assertions, 0 failures**, including seven new cases (re-gift inside the window blocks with the sender-addressed message; re-gift after the window succeeds; one per pending-payment path asserting a second sender is never promised an email; and one per purchase state asserting a sender blocked by the recipient's *own* purchase is not told a gift is in flight) and one renamed case whose old name asserted the behaviour this PR deliberately changes.

test results and mutation proof

The image above captures the run at `26edd23`, when the filter stood at 29 runs. The **mutation proof** has been re-run at each subsequent commit rather than re-rendered: restoring `origin/main`'s `Purchase#not_double_charged` reddens exactly the cases that cover this change, and reverting only the latest commit's `purchase.rb` hunks reddens exactly its three new tests and nothing else — each on the wrong copy the review described. The coverage is load-bearing rather than passing against anything.

RSpec's `create_service_spec.rb` gift group cannot run on this machine: every example in it fails at `ChargeProcessor.create_payment_intent_or_charge!` with `undefined method 'charge_processor_id' for nil`, because the gumroad `MerchantAccount` rows only exist inside the full spec harness. Verified environmental, not from this diff — the same 23 examples fail identically with the change stashed. Its assertions are updated for the new copy and CI is the arbiter for that file.

## QA steps

There is no clickable surface; the change is a model validation. On the preview app's Rails console:

1. `product = Link.find(<id>)` — any product with a `gift` association available.
2. Gift it once to `giftee@example.com` and let it succeed.
3. Immediately build a second gift of the same product to the same address from a *different* sender and call `purchase.valid?` — expect `false`, and `purchase.errors[:base]` to name the recipient rather than claiming a delivery to the sender.
4. `travel 4.minutes` (or set the first gift's `created_at` back), rebuild, and call `valid?` again — expect `true`. Before this change it stayed `false` indefinitely.
5. Negative case: put an earlier gift into a `payment_settling` state and retry — it must still be blocked with the "still processing" message, with no time limit.

---

🤖 Written by **claude-opus-5** (Hermes Agent, autonomous dev dispatcher).

Instructions given: build the open, human-approved bug reports on the private tracker autonomously — a re-gift report where the recipient could never be gifted the same product twice and the sender was shown a message about the recipient. The maintainer's instruction on that report was "Open PR". Root-cause it in the code rather than trusting the report's framing, fix the smallest complete thing, cover it in both test suites, prove the new coverage load-bearing by reverting the fix, and open the PR with the research in the body.


gumclaw:
## Self-review against CONTRIBUTING.md

Re-read `CONTRIBUTING.md` against this diff as a literal checklist rather than from memory.

| Requirement | State |
|---|---|
| Visual evidence / video | Walkthrough video (38s) committed at `qa-media/pr-6360-walkthrough.mp4`, linked in the body. Non-user-facing change, so it walks the validation path and both message branches rather than showing before/after pixels. |
| QA steps someone else can follow | Five numbered console steps in the body, including the negative case (settling gift must stay blocked, no time limit). |
| Test results + screenshot | `qa-media/pr-6360-test-results.png` — 29 runs / 0 failures, **plus** the mutation proof in the same transcript. |
| AI disclosure naming the model | `claude-opus-5`, read from config rather than recalled. |
| Media naming `pr-<number>-<description>.<ext>` | Both files. The PNG was rendered under a placeholder name and renamed once `gh pr create` returned 6360, so no wrong number was guessed. |
| Raw URLs durable | Both pinned to commit `7ef0311`, not the branch — a branch-pinned URL 404s permanently once merge deletes the branch. Verified 200 with byte-for-byte size match against the local files. |
| What / Why / Before-After / Test Results | All four present. |
| Small diff | 3 source files, +102/−18. |

### Three things worth a reviewer's attention

**1. The settling check moves in the opposite direction to the main fix, and that is deliberate.** Adding a window to the gift lookup silently weakened `payment_settling`, whose comment explicitly relied on the gift lookup above having *no* window. Left alone, a gift paid by bank debit could have become invisible to the settling check and been charged twice — a worse bug than the one being fixed. So the same commit gives `payment_settling` its own gift lookup with no window. If you read only the first hunk, this looks like an inconsistency; it is the invariant.

**2. One existing minitest name asserted the behaviour this PR changes.** `test_not_double_charged_settling_already_blocks_a_repeat_gift_…_without_any_time_window` described the mechanism being replaced. Renamed to state the guarantee (an unresolved payment blocks a retry however long it takes) rather than the implementation, and its assertion updated to the message that path actually produces on `main` today. Flagging it so the change reads as deliberate rather than as a test bent to fit.

**3. The new coverage is proven load-bearing, not assumed.** Restoring `origin/main`'s `Purchase#not_double_charged` and re-running the identical filter reddens exactly three tests — both new cases and the settling one. That run is in the evidence image below the green one, so the image argues both "it passes" and "it would catch the regression". Working tree verified restored before committing (`git diff --stat` shows only the three intended files).

### Declared gap

`spec/services/purchase/create_service_spec.rb`'s gift group cannot run on this machine: all 23 examples fail at `ChargeProcessor.create_payment_intent_or_charge!` with `undefined method 'charge_processor_id' for nil`, because the gumroad `MerchantAccount` rows only exist inside the full spec harness. Confirmed environmental by stashing the change and re-running — the identical 23 fail on unmodified `main`. Its assertions are updated for the new copy and CI is the arbiter for that file. Saying so explicitly rather than omitting it, since "all tests pass" would be false for that file.

gumclaw:
## Greptile review triage

No actionable findings in this review — 5/5 confidence, no files flagged, no inline comments filed. Re-checked the two things a summary-only pass could plausibly have glossed over:

- **Scoping of the new windowed gift lookup.** It keeps product (`link`), recipient (`gifts.giftee_email`), state (`limiting_purchase_states`), the current-record exclusion (`purchases.id != ?`), and the variant filter that runs after both lookups are merged — so widening the window did not drop any of the constraints that make a match a genuine duplicate.
- **The settling counterpart is not merely a comment change.** The old comment claimed the unwindowed lookup in `not_double_charged` covered gifts; that claim stops holding the moment the window exists, so `not_double_charged_while_payment_settling` now does its own gift lookup with no window. A gift paid by bank debit still blocks a retry however long it takes to settle, which is the one case where an unbounded block is the correct behaviour.

Nothing to change from this review. Marking the PR ready for review; the adversarial pre-merge review still runs before any merge.

gumclaw:
## Fable-5 adversarial pre-merge review — `CHANGES-REQUIRED` → P2 fixed at `26edd23`

Ran the local fable-5 review against `7ef0311`. One P2, and it was right in a way worth stating plainly, because it is the same bug as the one this PR exists to fix:

**The PR made one error site reader-aware and left its sibling gift-blind — and windowing the gift lookup is precisely what starts routing gift senders to that untouched site.** Before this change, a gift settling over a bank debit was caught by the unwindowed lookup in `not_double_charged` and got the copy I had just fixed. After it, the window excludes that gift, so it falls through to the settling check, which still said:

> Your previous payment for this product is still processing… please do not pay again.

False for both readers who can now reach it. A *different* sender gifting the same product has made no previous payment. A giftee buying it directly while someone gifts it to them has not either. Telling either of them not to pay again abandons a legitimate purchase — one branch over from the bug in the title.

Worse, the pre-existing test I had renamed *enshrined* that message for a gift scenario, so it would have been locked in. Both readers now get copy addressed to them, and the giftee test asserts what they are *told* rather than only that they are blocked.

Net effect on the evidence: reverting the fix now reddens **four** tests, not three — the giftee-facing message is load-bearing too.

### One P3, deliberately not fixed — and now documented in the code

`payment_settling` requires a `stripe_status`, so a gift stuck `in_progress` with no status at all, older than the window, is caught by neither check. I am leaving it, because it is **inherited rather than introduced**: the same window bounds direct purchases of those products today, so this is the existing behaviour of the file rather than a new hole, and widening it here would re-create the lifetime block the PR removes. It is written as a comment beside the lookup so the next reader does not have to re-derive that.

### Four things the review went after and cleared

| Hunt | Result |
|---|---|
| `settling_gifts` missing `not_is_gift_sender_purchase` / self-exclusion | Correct as-is — `joins(:gift_given)` returns only gift-sender rows, so the filter would zero the result. Double-counting only on a self-gift, and only feeds `.any?`. |
| Third-party email disclosure in the new message | No path. The interpolated branch is reached only when `is_gift_sender_purchase`, i.e. the reader is the sender who typed that address. |
| The `sent long ago` rspec setup | Updating only the gifter purchase suffices; the giftee row is `gift_received` (excluded by the join) with a random factory email. Boundary verified on the right side. |
| A third unwindowed dedupe site elsewhere | None. The only other `giftee_email` lookups are in admin search and GDPR erasure, neither a double-charge guard. |

## Routing

@nyomanjyotisa this is yours to review and merge — I am not self-merging it.

**Why you:** you own the charge/refund backend lane, and this is a change to `Purchase`'s double-charge validation. It is not a money-movement change (no amounts, no processors, no schema), but it *loosens a guard that exists to prevent double charges*, which deserves a second pair of eyes on exactly the question the review chased: is there any path where the new window lets a real duplicate through.

**Where to focus.** The two lookups now move in opposite directions on purpose — the gift lookup in `not_double_charged` gains a window, the settling one gets an explicit lookup with none. If you read only the first hunk it looks inconsistent; that opposition is the invariant, and the review's P3 comment marks the one narrow state neither covers and why.

CI is green at `26edd23` (`buildkite/gumroad` success), mergeable, out of draft. Local: 29 runs / 44 assertions / 0 failures, plus the mutation proof in the same evidence image. The rspec gift group's 23 failures are environmental and identical on unmodified `main` — declared in the self-review comment rather than omitted.

nyomanjyotisa (commented):
A second sender is told “We will email you,” but the email goes to the original sender. Please use ownership-neutral wording in both pending-payment paths and add cross-sender coverage.

Also, please rebase—the branch is 13 commits behind main.

nyomanjyotisa (commented):
A gift attempt can match the recipient’s own settling direct purchase, but the message incorrectly says a gift is settling. Please choose the copy based on the matched purchase type and add coverage for this path.

Adversarial review from Claude Fable 5

gumclaw:
You're right, and the problem was slightly worse than the wording: both gift pending-payment messages ended in a promise of an email, and neither is safe to make, because the gift being described may not belong to the person reading it.

Two people can independently gift the same product to the same recipient, and the block keys on the recipient. So a second sender gets a message about the first sender's gift, and the receipt for that gift is addressed to the first sender. The second sender waits on a mailbox that will never receive anything, then retries — which is the same retry loop this PR exists to stop, just moved one step later.

Both paths now describe the gift and the action to take, without claiming ownership of it:

- settling — "A gift of this product to \<recipient\> is still being paid for. Wait for that to finish before sending it again."
- in progress — "A gift of this product to \<recipient\> is already going through. Wait for it to finish before sending another."

Cross-sender coverage is one test per path, each with a distinct first-sender and second-sender address, and each asserting the message contains no email promise at all rather than just matching the new string — so re-introducing "we will email you" in either place fails even if someone rewrites the rest of the sentence.

`bin/rails test test/models/purchase_test.rb -n "/not_double_charged/"` — 31 runs, 52 assertions, 0 failures. Reverting to the old copy reddens exactly three: both new tests and the settling assertion, so the coverage is load-bearing.

Rebased onto main as well — the branch was 13 behind, it's at 0 now, and the suite was re-run after the rebase, not just before it. The PR body is updated for the new copy and counts.

gumclaw:
Confirmed and fixed in `8e8a062`.

Both dedupe lookups key on the recipient's address, and a giftee's own direct purchase is stored under exactly that address — so it matches and blocks the sender, and the copy then called it a gift. On the settling path the sender got "A gift of this product to \<giftee\> is still being paid for"; on the successful path, "This product was just sent as a gift to \<giftee\>". No gift exists in either case, so the sender is told to wait on or check about something nobody sent.

The copy is now chosen from the matched purchase rather than from the reader alone, at both error sites — the settling check and `add_errors_for_existing_purchase`. I fixed the sibling site in the same commit rather than only the one you pointed at, because that is the exact gap the earlier review on this PR caught: making one reader-aware and leaving its neighbour gift-blind. A sender blocked by the recipient's own purchase now gets "\<giftee\> is in the middle of buying this product themselves" (in flight) or "\<giftee\> just bought this product themselves" (settled).

One thing worth flagging that the review didn't ask for: the check is per purchase state, not across the whole match set. A recipient can hold a settled direct purchase and someone else's in-flight gift at the same time, so a single "any of these is a gift" flag would report the wrong one whenever both exist.

Coverage: three tests, one per state the sender can be blocked in — settling, in-progress, successful. Each asserts the exact message and that it makes no gift claim. `bin/rails test test/models/purchase_test.rb -n "/not_double_charged/"` — 34 runs, 66 assertions, 0 failures. Reverting just the `purchase.rb` change and re-running the identical filter reddens exactly those three and nothing else, each on the wrong copy you described, so the coverage is load-bearing.

The rebase you asked for in the other review landed before this; the branch is current with main.

PR #6361, opened 2026-07-26
Tell sellers which required fields are blocking a payments save

gumclaw:
## What

Pressing **Update settings** on Settings → Payments now always tells the seller what is blocking the save, and scrolls them to the field that is blocking it.

Three changes in `app/javascript/pages/Settings/Payments/Show.tsx`:

- `validateForm()` falls back to a message that **names the failing fields** — "Please complete the required fields below: City, ZIP code, Date of birth." — but only when none of the specific checks already set a message, so the existing P.O. Box / phone-format / Kana / Peru-DNI / Singapore-NRIC messages still win.
- The scroll effect now also fires when `errorFieldNames` is non-empty, and scrolls to (and focuses) the **first `aria-invalid` input** instead of the top of the form.
- A `FIELD_LABELS` map gives every `FormFieldName` a human-readable name, deduplicated so the three date-of-birth selects read as one entry rather than three.

## Why

`handleSave()` bails out early when client-side validation fails:

```ts
const handleSave = asyncVoid(async () => {
  if (!validateForm()) return;   // silent bail, nothing rendered
```

`validateForm()` sets a top-level `clientErrorMessage` for only five cases (P.O. Box address, missing state/province, phone format, Peru DNI, Singapore NRIC). Every other required field — `first_name`, `city`, `zip_code`, the three `dob_*` selects, `individual_tax_id`, `account_holder_full_name`, and every per-country bank field — only got `markFieldInvalid()`, which adds a red outline to the input and nothing else. The old scroll effect was gated on `errors?.base || clientErrorMessage`, so a plain missing field produced no scroll either.

That combination is what makes it read as a **dead button** rather than a validation failure. The save control lives in the sticky `PageHeader` at the top of the Settings layout while the compliance and bank fields sit far below the fold, so the one signal that did fire — the red outline — was always off-screen behind the seller. `canUpdate` is true in these cases, so the button is visibly enabled and clickable; it just no-ops.

Knock-on effect: a product can only be published once a payout method exists (`Link#enforce_merchant_account_exits_for_new_users!`), so the same sellers also report the **Publish** button as broken. Two "dead buttons" from one silent bail.

Live case that prompted this: an Omani seller (user `28043042`) wrote in three times over four days, latest escalating to "buttons have been completely non-responsive for two weeks". Prod console showed nothing on our side blocking her: not paused, not flagged, `is_form_disabled = false`, payments policy `update? == true`, zero `BankAccount` and zero `MerchantAccount` rows, and an alive `UserComplianceInfo` carrying only `country = "Oman"`. Nine product drafts created between June 30 and July 26, none published. Roughly four weeks of a seller unable to launch because her date of birth was blank and we never said so. The support side has this pattern documented across eight presentations of the same root cause.

**Why this shape.** The alternative was an inline "required" message rendered under each `markFieldInvalid` field. Rejected for now: it means touching every field in three large section components (`AccountDetailsSection`, `BankAccountSection`, `PayPalEmailSection` — 133 `aria-invalid` sites between them) for a change that still would not be visible from the sticky header where the button lives. The banner plus scroll-to-field fixes the reported failure with one change at the point where the bail already happens, and the inline messages remain a clean follow-up. The scroll target reuses the `aria-invalid` attribute those components already set from `errorFieldNames`, so no new plumbing was needed.

**What this does not cover.** The banner lists field labels, not per-field guidance, so a field that is present but *malformed* in a way with no specific check still reads as "please complete" rather than "this value is wrong". And a seller who has scrolled far below the fold sees the scroll and the focused field, but the banner itself is still at the top of the form — the compact mobile summary next to the sticky header suggested on the driving issue is out of scope here.

Captured on the hosted preview at `https://fix-payments-settings-silent-val.apps.staging.gumroad.org`
(`x-revision: cfd33d63e71e`), Playwright at deviceScaleFactor 2, desktop 1440px and mobile 375px,
signed in as `seller@gumroad.com`. Every shot is the same scenario: **Individual** account type,
first/last name, address, state, phone, SSN and the four bank fields filled, with **City**,
**ZIP code** and the three **date-of-birth** selects left blank — then **Update settings** pressed.

The "before" frames are the identical script run against a **sibling preview whose branch does not
contain this commit** (`fix-quote-non-usd-listed-product`, `x-revision: fd3044689609`), so they are
real current-`main` behaviour on a hosted app rather than a mock or a local render.

| | Before (no this change) | After (this branch) |
|---|---|---|
| Desktop 1440px | before desktop | after desktop banner |
| Mobile 375px | before mobile | after mobile banner |

The "after" frames were re-captured at head `cfd33d63e71e` after two follow-up commits changed
what they show: the banner now takes its field names from the labels rendered next to the inputs
(so a US seller reads **ZIP code**, not "Postal code"), and its padding matches the form sections
instead of sitting flush against the tabs. The "before" frames are unchanged and stay pinned at
`8814d5c908be` — they show no banner at all, which neither commit affects.

The before frames are the page exactly as the click left it — the viewport never moved, and no
banner, toast or alert rendered anywhere on the page. That is the reported "dead button". The after
frames show the banner in the sticky header area reading, verbatim:

> Please complete the required fields below: City, ZIP code, Date of birth.

**Scroll-to-field half of the change.** The banner is only half the fix; the effect also scrolls the
first `aria-invalid` input into view and focuses it. These are the same runs, at the position the app
left the viewport in — the City input focused with its red outline, ZIP and the three date-of-birth
selects outlined below it:

| Desktop 1440px | Mobile 375px |
|---|---|
| after desktop scrolled | after mobile scrolled |

**Negative case — the specific messages still win.** Same run with the address set to
`P.O. Box 123, Smith street` and City / ZIP / date of birth all filled in. The banner reads the
P.O. Box message and the generic field list does **not** appear; the only `aria-invalid` element is
the Address input, which is what the page scrolled to:

po box message

**Save still succeeds once the fields are filled.** With City, ZIP and the date of birth completed,
the same click saves: zero `[role="alert"]` nodes, zero `aria-invalid` elements, and the
**Update settings** button goes disabled while the request is in flight.

filled saves clean

This one is a real controller round trip, not just client state — reading the preview's database
back afterwards shows the values persisted and a bank account created:

```
city="San Francisco" zip="94103" dob=Mon, 15 Jan 1990 phone="+14155550123" street="123 Market Street"
bank_accounts=1 account_number_visual="******6789"
```

Assertions from the capture runs, read out of the live DOM rather than eyeballed off the images:

| Run | `[role="alert"]` text | `aria-invalid` elements | focused after click |
|---|---|---|---|
| Before, desktop | *(none)* | city, zip, dob-month, dob-day, dob-year | the button itself |
| Before, mobile | *(none)* | city, zip, dob-month, dob-day, dob-year | the button itself |
| After, desktop | Please complete the required fields below: City, ZIP code, Date of birth. | city, zip, dob-month, dob-day, dob-year | `creator-city` |
| After, mobile | Please complete the required fields below: City, ZIP code, Date of birth. | city, zip, dob-month, dob-day, dob-year | `creator-city` |
| After, P.O. Box | We require a valid physical US address. We cannot accept a P.O. Box as a valid address. | street-address | `creator-street-address` |
| After, all filled | *(none)* | *(none)* | — |

Note the before rows: the red outlines were always there. The change is that there is now a message
saying what they mean, and the seller is taken to them.

**One caveat on the before frames.** They come from a sibling preview because the branch's own
preview only ever served this commit, so there is no same-host baseline to shoot. The two apps are
built from the same `main` plus each branch's own diff, and neither branch touches
`Settings/Payments/Show.tsx` except this one.

## Test Results

Two new system-spec examples cover both halves of the change: the generic field list appearing (and then the save succeeding once the fields are filled), and the specific P.O. Box message still winning over the generic list.

Local spec run

⚠️ **The local run of these specs fails, and it is not the change.** All three examples in the image time out in `spec/support/capybara_helpers.rb:46` waiting for React to mount into `#app` — including `allows the creator to switch to debit card as payout method`, which is a pre-existing example this diff does not touch. Same failure, same line, in the untouched baseline: the worktree's vite test build is not hydrating. I am reporting that rather than omitting it or claiming a pass I did not get. CI is the arbiter for these system specs and this PR stays draft until it is green there.

**Re-verified at `bf20552ede17` (merge of `main`, 2026-07-26).** That head brings in only the
merge from `main` — no change to `Settings/Payments/Show.tsx` or the spec — so the frames above
are still accurate and stay pinned at `7de61a7083ae` / `8814d5c908be`. The merge did pull in
`main`'s new `config/initializers/secure_headers.rb` (CSP) and `vite.config.ts` changes, which
touch how this page's JS is served, so the preview was re-checked rather than assumed: the
hosted app now serves `x-revision: bf20552ede17`, `/settings/payments` renders the full form
(18 inputs, **Update settings** present, no country modal, no "Something broke"), and the page
throws no JavaScript errors. The React bundle still mounts under the new CSP.

## QA steps

Preview app: **https://fix-payments-settings-silent-val.apps.staging.gumroad.org/settings/payments**
(serving `x-revision: cfd33d63e71e`; the only commit after that is the `qa-media/` screenshots).
Sign in as `seller@gumroad.com` / `password`, 2FA `000000`.

Click path: **Settings → Payments** (or go straight to `/settings/payments`).

1. Sign in as a seller with no payout method configured and go to **Settings → Payments**.
2. Choose **Individual**, fill in first/last name, address, state, phone, and the bank fields — leave **City**, **ZIP code** and the **date of birth** selects blank.
3. Scroll back to the top and press **Update settings**.
   - Before: nothing happens at all.
   - After: a red banner reads "Please complete the required fields below: City, ZIP code, Date of birth." and the page scrolls down to the City input with focus on it.
4. Fill in City, ZIP code and the date of birth, press **Update settings** again → saves, "Thanks! You're all set."
5. Negative case (specific message still wins): fill everything in but set the address to `P.O. Box 123, Smith street` and save. The banner must read "We cannot accept a P.O. Box as a valid address." and must **not** show the generic field list.

---

## AI disclosure

Written by `claude-opus-5` running as Hermes Agent; the hosted-preview screenshots above were
captured and attached by the same model running as the preview-screenshot watcher. Instructions given: sweep antiwork/gumroad-private and antiwork/gumroad for issues older than three hours that no human has claimed, and build them autonomously through PR — bug fixes, docs, copy, tests and small self-contained features — with specs, local verification, and a self-review before merge. This PR resolves the payments-settings silent-validation-bail bug filed from a cluster of support tickets about a "dead" Update settings button.


slavingia:
Screenshot / design audit with fable 5. The spacing around the alert seems off. Kissing top and on mobile too much on sides?

gumclaw:
Ran the screenshot and design audit. Both of the spacing problems you spotted are real, and measuring them in the DOM on the preview app turned up two genuine bugs in the change itself, which are also fixed.

## The spacing

The banner wrapper was `mb-12 px-8` — no top padding at all, and a fixed 32px horizontal inset on every breakpoint. Measured on the preview at the old head (`8814d5c908be`):

| | Desktop 1440px | Mobile 375px |
|---|---|---|
| Gap above the banner (to the tabs strip) | 33px, but the wrapper's own top padding was **0px** — the banner's border sat directly on the section divider | 17px, same 0px padding |
| Banner left inset | 32px — matched the "Payout schedule" heading at 32px | **32px, while page content sits at 16px** |
| Banner right inset | 32px | 32px |

So: kissing the top on desktop because the wrapper had no top padding, and on mobile inset twice as far as everything else because `px-8` doesn't step down at small breakpoints.

Fix is `p-4! md:p-8!` — the same padding `components/ui/FormSection.tsx` gives every section on this page. That fixes both edges at once and the banner now lines up with the fields below it instead of drifting from the convention. I checked `components/Public/LookupLayout.tsx:93` uses the identical wrapper pattern, so this isn't a one-off spacing value.

I also checked whether dropping `mb-12` opens a gap below: banner→section is now `p-8` + the section's border + `p-8`, which is exactly the section→section rhythm, so it reads consistent rather than short.

## Two bugs the audit turned up

**The scroll effect was stealing the seller's position on every keystroke.** The effect listed the set of invalid fields as a dependency, but every field's `onChange` clears that set. So typing the first character into the field the banner had just pointed at cleared the set, re-ran the effect with nothing marked invalid, and fell through to scrolling the form back to the top. Measured on the preview: the inner scroll container went from `1198px` to `153px` on the first keystroke, dragging the seller away from the field they were filling in — the exact opposite of what the change is for. It now keys off a counter that only increments when a save is rejected, so it fires once per press of **Update settings**.

**The banner named fields the seller can't find.** It used a hardcoded label map, but several of those labels change with the seller's country. A US seller — the largest cohort — sees "ZIP code" while the banner said "Postal code"; the state field is variously State, Province, County or Prefecture; the bank code field is "SWIFT / BIC Code" in dozens of countries. It now reads the label actually rendered next to each invalid input, falling back to the curated names when an input has no label, and always using "Date of birth" for the three selects whose own labels are Month/Day/Year.

One more, found in the pre-merge review: `validateForm` kept whatever fields the *previous* attempt had rejected, and the PayPal email field and the saved-bank-account toggle don't clear that set. A seller who saved with an empty PayPal address, filled it in, and saved again was blocked by the leftover entry and told to complete a field that was already filled. Each attempt now starts from a clean set.

## Surfaces I audited and left alone

- `AccountStatusSection` — already `p-4 md:p-8`, correct on both breakpoints.
- `AusBackTaxesSection` — renders through `FormSection`, so it inherits the convention.
- Every other alert on the page (`Show.tsx:1130`, `:1138`) sits inside a `FormSection`'s `grid gap-8`, measured at 32px on desktop. Untouched.

Refreshed before/after screenshots are coming — the preview is rebuilding on the new head now. Full pre-merge review verdict to follow in the same place.

gumclaw:
@gianfrancopiana routing this to you for review — green at head, mergeable, and everything on my side is done.

**The judgement call is error-surfacing on the payments settings form.** Today a save that's missing required fields fails without telling the seller *which* fields are blocking it; this names them. That's a UI/design call in your area — specifically whether the fields are surfaced in the right place and worded the way a seller would recognise them, since the whole value of the change is that the seller can act on it without contacting support.

The `preview` label is on, so there's a hosted preview if you'd rather click through it than read the diff.

No rush beyond the usual 24h review goal; if this isn't your area or you're at capacity, say so and I'll re-route rather than let it sit.

PR #6362, opened 2026-07-26
Accept Cédula de Extranjería in the Colombia payout ID field

gumclaw:
## What

Makes the Colombian personal-ID field on Settings → Payments accept both of Colombia's personal ID documents instead of only one:

- Relabels the field from "Cédula de Ciudadanía (CC)" to "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)", in both the account-details form and the beneficial-owners form.
- Replaces the exact 13-character length requirement with a 6–13 character range, and drops the dotted placeholder (`1.123.123.123` → `1234567890`) that implied one specific written format.

## Why

Colombia issues two personal ID documents: the Cédula de Ciudadanía to citizens, and the Cédula de Extranjería to foreign residents. Both are sent to Stripe as the generic `individual.id_number` (`stripe_user_compliance_info_field_map.rb`), so both have always been acceptable to the payment processor — but the field only named the citizen document, which tells a foreign resident their ID is not accepted.

The length constraint compounded it. `minLength: 13` / `maxLength: 13` fits exactly one shape: a ten-digit number written with thousands separators. Cédula de Extranjería numbers are commonly six or seven digits and many citizen numbers are eight or nine, so a correct ID shorter than 13 characters was rejected by the browser's own length check the moment the seller pressed Enter in the field, and a longer paste was silently truncated. The result is a payout setup nobody in that situation can complete. Neighbouring countries in the same list each have a single national format, so one exact length works for them; Colombia is the case where it cannot.

The new range keeps a loose sanity check (6 minimum, 13 to leave room for a number pasted with separators) and leaves the authoritative validation to Stripe, which already performs it.

Tracking issue: antiwork/gumroad-private#1396

### Scope note

An earlier revision of this branch also changed `country in individual_tax_id_needed_countries` to `.includes()` on the same page. That was reverted: the `in` form tests object keys, so the check has never matched a country code, and switching it on is a behavior change rather than a cleanup — it starts blocking saves the server still accepts (a Kazakhstan PayPal payout needs no tax ID), which the KZ PayPal spec caught. The line keeps an explanatory comment so the next reader does not repeat the same "fix". That change needs its own PR with the server rule aligned.

## Before / After

Both columns are the **same script against two hosted preview apps**, with the seller's compliance country set to Colombia. "Before" is a sibling preview whose branch does not contain this commit (`fix-payments-settings-silent-val`, `x-revision: 8814d5c908be` — verified to still carry the old `label: "Cédula de Ciudadanía (CC)", minLength: 13` config). "After" is this branch's preview (`x-revision: 43197df01db7`). In both, a seven-digit Cédula de Extranjería number (`1234567`) is typed into the field and the browser's own `checkValidity()` is read back.

| | Before (no this change) | After (this branch) |
|---|---|---|
| Desktop 1440px | before desktop | after desktop |
| Mobile 375px | before mobile | after mobile |

The field attributes and validation result read off the live DOM, verbatim:

**Before** — `label: "Cédula de Ciudadanía (CC)"`, `placeholder: "1.123.123.123"`, `minLength: 13`, `maxLength: 13`

> valid: **false** — "Please lengthen this text to 13 characters or more (you are currently using 7 characters)."

**After** — `label: "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)"`, `placeholder: "1234567890"`, `minLength: 6`, `maxLength: 13`

> valid: **true** — no validation message

That rejection message is exactly what the reporting seller was hitting.

## Test Results

The Colombia payments spec now enters a seven-digit CE-shaped ID (a value the old constraint rejected) and asserts it persists to the compliance record, which the spec did not previously check. Previously it entered `1.123.123.123` — the one value the old constraint accepted — so it could not have caught this.

Local runs of this spec are not usable in my worktree: the payments page never hydrates, so the spec times out in `visit`, and the untouched Argentina spec beside it fails identically. That is the harness, not this diff, so CI on this branch is the authority. CI on the previous head also surfaced the KZ PayPal failure described in the scope note above, which is what prompted reverting the tax-ID change.

**Re-verified at `6367a40e5d81` (merge of `main`, 2026-07-26).** That head is the merge from
`main` only — the three component/page files and the spec are unchanged, so the before/after
frames above stay pinned at `6f2fc3f5023a`. Because the merge pulled in `main`'s new CSP
(`config/initializers/secure_headers.rb`) and `vite.config.ts` changes, the hosted preview was
re-read rather than assumed. It now serves `x-revision: 6367a40e5d81`, and the field attributes
read back off the live DOM at that revision are unchanged from the "after" column:

```
label:       "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)"
placeholder: "1234567890"
minLength:   6      maxLength: 13
```

The page renders the full payments form with no JavaScript errors under the new CSP.

## QA steps

1. Open the preview app: https://fix-colombia-id-cc-or-ce.apps.staging.gumroad.org (seller's compliance country is already set to Colombia).
2. Go to Settings → Payments.
3. The personal-ID field reads "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)".
4. Enter a seven-digit value and press Enter. It no longer blocks on a 13-character minimum.

---

## AI disclosure

Written by claude-opus-5 running as the Gumroad agent. It was given the bug report in antiwork/gumroad-private#1396 (a Colombian-resident seller unable to enter a Cédula de Extranjería) and the instruction to fix it. The model read the payments settings components, made the label, length and placeholder changes, updated the Colombia payments spec, triaged the resulting CI failure and reverted the out-of-scope tax-ID change, and captured the before/after evidence from two hosted preview apps.


gumclaw:
## Self-review

Re-read CONTRIBUTING.md against this diff.

**Root cause.** The invalid state came from treating Colombia the way its neighbours in the same config are treated: one national document, one exact length. That model is wrong for Colombia specifically, because it issues a second personal ID (Cedula de Extranjeria) to foreign residents, and the two documents do not share a digit count. The field carried both a name that excluded a whole class of its users and a length that fit neither document reliably.

**Scope.** Four files, all in the payout-ID surface. I deliberately did not touch the server-side validation in `UpdateUserComplianceInfo` - there is no Colombia-specific check there, and the value goes to Stripe as the generic `individual.id_number`, which Stripe validates. Adding a Colombian format check would be inventing a constraint we have no source for, which is the same shape as the bug.

**Why a 6-13 range instead of removing the constraint entirely.** Removing `minLength` would let a single stray digit through to Stripe and turn a clear inline message into a slower round trip. 6 is below the shortest realistic Cedula de Extranjeria; 13 preserves headroom for a number pasted with thousands separators, which the old placeholder actively encouraged people to type. Neither bound is claimed as an authoritative Colombian format - that is Stripe's job, and the code comment says so.

**Both copies of the config.** `BeneficialOwnersSection.tsx` keeps its own copy of the same country map for the beneficial-owner ID field. Changing only the account-details copy would have left a business seller's beneficial owner facing the exact bug this PR fixes. Both are updated, and the comment in the second one points at the first so the next person knows they travel together.

**The `in` to `.includes()` change.** Bundled because it is three lines in the same validation function on the same page, found while reading it. It is not behavior this PR depends on: the check was inert, so switching it on is a real behavior change - the seller now sees the tax ID field flagged before the save instead of after it. That is the intended behavior of the code as written; it just never ran. Flagging it for review rather than burying it as a cleanup.

**Test.** The existing Colombia spec entered `1.123.123.123`, the one value the old constraint accepted, so it could never have caught this. It now enters a seven-digit CE-shaped number - a value the old code rejected - and asserts the ID actually persists to the compliance record, which the spec previously did not check at all.

**Not changed:** any translated or help-center copy for this field (there is none), and the Colombia business tax ID entry (a different document, NIT, out of scope).

gumclaw:
Status while the preview app builds.

Code is complete and pushed. Local verification of the Colombia system spec is **not** usable in this environment: the payments page never hydrates in my worktree, so the spec times out waiting for React — and the untouched Argentina spec next to it fails identically for the same reason. That is a harness problem, not this diff, so the authoritative test evidence will be the CI run on this branch plus captures from the hosted preview.

Preview app is building (`preview` label added, empty commit pushed to trigger the Buildkite deploy since the labeled-only trigger produced no deployment record). Screenshots go into `qa-media/` and the body's Before/After section as soon as the host answers.

gumclaw:
Greptile flagged nothing actionable, but its own summary named the change that turned out to be wrong, so recording the outcome here.

The third item in that summary — "Corrects client-side country-array membership validation with `.includes()`" — has been reverted. Greptile's confidence read was that broadening constraints flows through existing generic ID storage safely, which is true of the Colombia label and length change but not of the membership check.

`in` really does test object keys, and an array's keys are its indexes, so `country in individual_tax_id_needed_countries` never matched a country code. Switching to `.includes()` makes the check fire for the first time, and the server does not require an individual tax ID for every country in that list. A Kazakhstan seller setting up PayPal payouts has no tax ID field on that flow, so the form began refusing to submit. `Payments Settings Scenario Payout Information Collection KZ creator allows to enter PayPal details` failed on CI, expecting "Thanks! You're all set." and getting a form that never saved.

The line is back to `in` with a comment explaining why, so the next reader does not repeat the same cleanup. Aligning the client-side check with the real server-side requirement needs the server rule settled first and belongs in its own change, not in a Colombia label fix.

This PR is now just the label, the 6–13 length range, the placeholder, and the spec.

gumclaw:

Diff is 85 lines across 4 files. Hunted for the failure modes that actually apply to a country-config change, rather than re-reading the diff for style.

**Completeness of the label change — the main risk, since a half-applied config leaves the bug live on one surface.** Swept for every place the old value could still exist:

- `grep -rn 'Cédula de Ciudadanía (CC)"' app spec test` → no matches. No CC-only label survives anywhere.
- `grep -rn "colombia-id-number" app spec test` → only the two config maps, both changed.
- `grep -rln "TAX_ID_CONFIGS\|getIndividualTaxIdConfig" app/javascript` → exactly `AccountDetailsSection.tsx` and `BeneficialOwnersSection.tsx`. There is no third copy.

**Remaining `1.123.123.123` occurrences are intentional, not misses.** They are in `stripe_merchant_account_manager_spec.rb` and its VCR cassette, testing the Stripe merchant-registration payload rather than this form. That value is 13 characters, so it still satisfies the new 6–13 range and the spec keeps passing — no change needed, and changing it would mean re-recording a cassette for no reason.

**Does loosening `minLength` weaken any real guard?** No. The value is passed through to Stripe as the generic `individual.id_number`; there is no Colombia-specific server-side validation in `UpdateUserComplianceInfo` (unlike Peru DNI and Singapore NRIC, which each have explicit server checks). So the removed constraint was the only check, and it was rejecting valid documents rather than catching invalid ones. Stripe still validates authoritatively.

**Is `maxLength: 13` still needed given the placeholder no longer suggests separators?** Yes, kept deliberately: sellers still paste dotted numbers regardless of the placeholder, and 13 accommodates a 10-digit number written `1.123.123.123` without truncating it. Dropping the cap entirely would gain nothing.

**Spec actually exercises the bug.** The old spec used the single value the old constraint accepted, so it was structurally incapable of failing on this. It now uses a 7-digit CE-shaped number and asserts persistence of the ID, so it would redden if either the label or the length regressed.

**Scope creep, caught and reverted.** The one genuine P1 in the first revision was the `in` → `.includes()` change: it converted a dead check into an active one that is stricter than the server, and would have blocked Kazakhstan PayPal sellers who can save today. Caught by CI, reverted, documented inline. Worth noting the report's framing ("low impact, dead code as written") was the trap — inert code is not free to switch on.

Verdict: no P1/P2 outstanding. The diff does one thing, on every surface that thing appears, with evidence from a real hosted app on both sides of the change.

gumclaw:
The `spec/models/concerns/purchase/reportable_spec.rb:164` failure showing up across many test shards here is not from this PR. `CHARGEBACK_REPORTING_CUTOVER` is the fixed date 2026-07-27 and that example set `chargeback_date: Time.current`, so at midnight UTC the chargeback flipped from the legacy path (reports 0) to the event-dated path (reports 100). It fails the same way on every open branch.

Fix is up as #6371. Once that lands, re-run this PR's checks.

gumclaw:
Ready for review, @gianfrancopiana — assigned to you as a UI-surface change (label and input constraints on Settings → Payments), so it is not on auto-merge.

**Evidence.** Before/after in the body, captured from two hosted preview apps with the seller's compliance country set to Colombia. The "before" is a sibling preview verified not to contain this commit and still carrying `label: "Cédula de Ciudadanía (CC)", minLength: 13`. The same seven-digit Cédula de Extranjería is rejected there with the browser's "Please lengthen this text to 13 characters or more" and accepted here with no validation message. Preview: https://fix-colombia-id-cc-or-ce.apps.staging.gumroad.org

**Adversarial review** posted above: no P1/P2 outstanding. Its most useful finding was on my own first revision — I had bundled the `in` → `.includes()` array-membership fix from the issue, which converts a check that has never fired into one stricter than the server and would have blocked Kazakhstan PayPal sellers. CI caught it, it is reverted, and the line now carries a comment so the next reader does not repeat it.

**CI is red on one shard for a reason unrelated to this PR, and it is not a flake.** `spec/models/concerns/purchase/reportable_spec.rb:164` fails because `CHARGEBACK_REPORTING_CUTOVER = Date.new(2026, 7, 27)` has been reached — the constant's own comment requires it to be strictly after the deploy day. I reproduced the same failure on unmodified `origin/main` in a clean worktree, so it is currently red repo-wide, not caused by this diff. Filed as antiwork/gumroad-private#1403 with the two possible fixes, because choosing between them changes tax-period attribution and wants a deliberate call.

So: the only red here is that pre-existing landmine. Everything belonging to this change is green and evidenced.

PR #6364, opened 2026-07-26
Bust the profile cache when a product thumbnail changes

gumclaw:
## What

Adds `touch: true` to `Thumbnail`'s `belongs_to :product` so creating, replacing, or soft-deleting a product thumbnail bumps the product's `updated_at`. Adds regression specs covering both directions at the model level and at the `Pages::ProfileData` cache-key level.

## Why

`Pages::ProfileData.cache_key` keys the cached seller profile payload on `seller.products.cache_key_with_version`, which derives from `Link#updated_at`. Two writes need to move that timestamp, and only one did:

| model | association | touched the product? |
|---|---|---|
| `AssetPreview` (cover image) | `belongs_to :link, touch: true, optional: true` | yes |
| `Thumbnail` | `belongs_to :product, class_name: "Link", optional: true` | no |

`ThumbnailsController#create` / `#destroy` save or soft-delete the `Thumbnail` row and return success without ever writing to `links`, so the cached profile payload kept its old key. Production cache is memcached and this `Rails.cache.fetch` has no `expires_in`, so the stale entry survived until a deploy or an unrelated product edit. To a seller that reads as permanent: they upload the thumbnail, the product page shows it immediately, and their storefront keeps rendering a broken image.

Both surfaces that render from `Pages::ProfileData` were affected — published custom-HTML profile pages (the injected `gumroad-data` JSON) and `Pages::DefaultProfileDocument`, which only emits an `` when `thumbnail_url` is present. Deletion had the mirror-image bug: the storefront kept showing a thumbnail the seller had removed.

`touch: true` fires on create, update, and `mark_deleted!` (which is an update), so one line covers both directions, and it makes `Thumbnail` consistent with `AssetPreview`.

Fixes antiwork/gumroad-private#1398

## Before / After

No rendering code changed, so there is no visual diff to capture: the templates, the JSON payload shape, and the profile document builder are untouched. The observable change is that the cache key now moves when a thumbnail changes, so the next storefront request rebuilds the payload instead of serving the pre-upload snapshot.

That behavior is asserted directly in `spec/services/pages/profile_data_spec.rb`, which builds the payload before and after a thumbnail create and a soft-delete and checks `thumbnail_url` flips `nil → present → nil` along with the cache key changing each time. Reverting `touch: true` turns those assertions red (see below), which is the before/after in executable form.

## QA steps

1. As a seller with a published custom profile page (or the default profile), note a product with no thumbnail — the storefront card renders without an image.
2. Upload a square thumbnail on that product.
3. Reload the storefront. The thumbnail renders on the first reload, with no other product edit and no deploy in between.
4. Delete the thumbnail and reload the storefront again — the image disappears immediately rather than persisting.

## Test results

```
bundle exec rspec spec/models/thumbnail_spec.rb \
  spec/services/pages/profile_data_spec.rb \
  spec/controllers/thumbnails_controller_spec.rb

40 examples, 0 failures
```

spec results

Load-bearing check — with `touch: true` reverted and the specs unchanged, the three new examples fail:

```
24 examples, 3 failures

rspec ./spec/services/pages/profile_data_spec.rb:48 # changes the cache key when a thumbnail is added or removed
rspec ./spec/models/thumbnail_spec.rb:79 # bumps the product's updated_at when a thumbnail is created
rspec ./spec/models/thumbnail_spec.rb:88 # bumps the product's updated_at when a thumbnail is soft-deleted
```

`bundle exec rubocop` on the three changed files: no offenses.

Run locally on macOS against the branch worktree; CI verifies the full suite.

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5`. Instructions given: fix the reported bug where product thumbnail uploads never invalidate the cached `Pages::ProfileData` payload, matching the `touch: true` that `AssetPreview` already has, and add regression coverage for the create and delete directions.


gumclaw:

**Diff scope.** Three files plus one qa-media asset. `app/models/thumbnail.rb` gains `touch: true` and a comment explaining which cache depends on the product timestamp; the two spec files gain three examples. Nothing else is touched — no controller, no view, no payload shape, no migration.

**Does the fix actually cover the reported case, and its mirror?** `touch: true` fires on create and on update. `mark_deleted!` is an update, so the delete direction is covered by the same line rather than needing a separate callback. The controller's "restore a previously deleted thumbnail" path (`thumbnail.deleted_at = nil` then `save`) is also an update, so re-adding a thumbnail busts the cache too. All three paths are exercised: the two model specs cover create and soft-delete directly, and `spec/controllers/thumbnails_controller_spec.rb` (unmodified, passing) covers the restore path through the real controller.

**Cost of the extra write.** A touch writes `updated_at` and nothing else. Checked what that could wake up:
- `Link`'s `after_update :invalidate_cache` guard is `(saved_changes.keys - PURCHASE_PROPERTIES).present?`, and `PURCHASE_PROPERTIES` is `["updated_at"]`, so an `updated_at`-only change deliberately does not fire it.
- Elasticsearch reindexing runs off `ATTRIBUTE_TO_SEARCH_FIELDS_MAP` in `Product::Searchable`, which does not include `updated_at`, so no index update is enqueued.
- `Link` has `has_paper_trail`, so each thumbnail write now also records a version row. That is already true of every cover-image upload via `AssetPreview`'s identical touch, so this makes thumbnails consistent with covers rather than introducing a new pattern.

Net: one extra `UPDATE links SET updated_at` per thumbnail write, no cascading jobs.

**Why the specs reload the seller.** `cache_key_with_version` on an association memoizes per instance, so reading the key twice off the same in-memory `seller` returns the same string even after the underlying row moved — the first draft of the profile-data spec failed for exactly that reason. Each read now goes through `seller.reload`, which is what a real request does anyway (fresh object per request). Called out in a comment in the spec so the reloads don't read as noise.

**Are the specs load-bearing?** Verified by reverting `touch: true` with the specs unchanged: all three fail (24 examples, 3 failures). Without that check a `be >` assertion on timestamps can pass for the wrong reason.

**Artifacts.** Test-results screenshot committed to `qa-media/` and referenced by raw URL, image URL validated and returning 200. No UI surface changed, so there are no before/after screenshots to take — the body says so explicitly and points at the payload-level assertions as the equivalent evidence. AI disclosure names the model read from config rather than remembered.

**No PII.** The originating report referenced a specific seller, product, and Helper ticket. None of that appears in the commit message, the PR body, the code comments, or the specs — this repo is public.

gumclaw:
## Pre-merge review

Ran the local fable-5 adversarial review against this branch. Verdict: **READY-TO-MERGE**, no P1 or P2 findings. The hunt list was aimed at the class of bug a bare `touch` most often causes.

**Timestamp-as-proxy guards — clean.** The real risk with adding `touch: true` is breaking a guard that treats `updated_at` as "a human edited this row". `Product::StaleContentWriteGuard` compares only RichContent page rows and variant rows (plus `VariantPrice` timestamps via `snapshot_at`); `Link#updated_at` never enters the comparison, and `LinksController#content_updated_at_response` echoes only `rich_content_updated_at` and `variant_updated_at`. So a thumbnail upload mid-edit cannot start blocking a seller's save. The remaining `Link#updated_at` consumers all render rather than gate: sitemap `lastmod` (the gating query uses `created_at`, and a fresher lastmod after an image change is more correct), the public API product payload, and an admin card. No `where("links.updated_at ...")` window query exists.

**Callback cascade — clean, and cheaper than I assumed.** Rails `touch` runs only `after_touch` and `after_commit`/`after_rollback`, never `after_update`, so `Link`'s `after_update :invalidate_cache` cannot fire — and its guard would exclude an `updated_at`-only change anyway, since `PURCHASE_PROPERTIES` is exactly `["updated_at"]`. `Link` declares no `after_touch`. Elasticsearch reindexing hangs off `before_update :update_search_index_for_changed_fields` and `updated_at` isn't in the attribute map, so no index churn. paper_trail 15 stopped creating versions on touch back in v12. Net cost: one extra `UPDATE links SET updated_at` per thumbnail write, no enqueued jobs.

**Transaction shape — clean.** The network I/O (`SsrfFilter` fetch, blob upload, `file.analyze`) all runs in the setter before `save`, so the transaction that now also writes `links` contains just the two row updates. No lock cycle is possible: the editor save path holds the `links` row but never writes `thumbnails`, and nothing takes those locks in the opposite order.

**Coverage is complete.** Every Thumbnail writer — both controllers, the AI cover-generation path in `LinksController`, and `ProductDuplicatorService` — goes through `save`/`save!`/`mark_deleted!`, so all of them fire the touch. There is no `update_columns`/`update_all`/`destroy` writer anywhere in app, lib, seeds, or the onetime services, so no path is left un-busted. Replacing an image on an existing thumbnail is covered even when the thumbnails row itself doesn't change: `ActiveStorage::Attachment` has `belongs_to :record, touch: true`, which touches the Thumbnail, and `belongs_to touch:` registers an `after_touch`, so it chains through to the product.

**Sufficiency confirmed.** The cache key's scope (`seller.products`, unscoped) is a strict superset of the payload's (`alive.not_archived.not_draft.limit(100)`), so setting `updated_at` to now necessarily moves `MAX(updated_at)` over the superset — the key changes for any product's thumbnail, including one outside the 100-item window. The failure mode is over-invalidation, never staleness. The same touch also un-sticks `ProfileSectionsPresenter#products_cache_key`, which had only a 10-minute TTL as a backstop.

Two P3s. The evidence asset didn't follow the `pr-<number>-<description>` naming convention — renamed in 24f08904. The other is a note, not a change: because the key scope is unscoped, a thumbnail write on a deleted or archived product also busts the profile payload. That's pre-existing behavior of the key and safe in the correctness direction; recording it so nobody narrows the key scope later without checking.

## Bot review

Greptile: 5/5, no findings.

## CI

One real failure on the first run, in `spec/controllers/api/mobile/analytics_controller_spec.rb:130` — a midnight/week-boundary flake unrelated to this diff. The shard asserted at 00:00:54 UTC on Monday, and the spec creates a purchase then expects `$1` for the day, week, month and year windows; `beginning_of_day` and `beginning_of_week` had both just rolled past the purchase, so exactly those two buckets returned `$0` while month and year passed. Nothing in this PR touches analytics, purchases, or time handling.

Local runs of that spec file fail identically on unmodified `origin/main` in this worktree (30 examples, 10 failures both places), so the local failures are environmental and don't speak to the branch either way. Re-running the shard on CI; the rename commit then superseded it with a fresh full run.

gumclaw:
CI triage: the Test Fast 4 failure on 188fc565e was a midnight/week-boundary flake, not related to this change.

The failing example is \`spec/controllers/api/mobile/analytics_controller_spec.rb:130\` ("GET revenue_totals returns data generated by the service"), and the assertion diff is the tell: \`month\` and \`year\` came back \$1 as expected, while \`day\` and \`week\` came back \$0.

\`SellerMobileAnalyticsService\` filters on \`created_on_or_after: now.beginning_of_<range>\`. The job created its purchase at 23:53 UTC on Sunday 26 July and asserted at 00:00:54 UTC on Monday 27 July. Between those two moments both the day boundary and the Sunday-to-Monday week boundary passed, so the purchase fell outside \`beginning_of_day\` and \`beginning_of_week\` while still sitting inside the current month and year. That reproduces the observed \$0/\$0/\$1/\$1 exactly.

This PR only adds \`touch: true\` to \`Thumbnail#product\` plus thumbnail and profile-cache specs; it does not touch analytics, purchases, or time handling. The spec is time-of-day dependent on any run that straddles midnight into a Monday, and would fail the same way on main.

Not fixing it here — freezing the clock in that analytics spec is a separate change to a spec this PR does not own.

gumclaw:
The `spec/models/concerns/purchase/reportable_spec.rb:164` failure showing up across many test shards here is not from this PR. `CHARGEBACK_REPORTING_CUTOVER` is the fixed date 2026-07-27 and that example set `chargeback_date: Time.current`, so at midnight UTC the chargeback flipped from the legacy path (reports 0) to the event-dated path (reports 100). It fails the same way on every open branch.

Fix is up as #6371. Once that lands, re-run this PR's checks.

gumclaw:
CI update — the remaining red is a repo-wide date bomb, not this diff.

The failing example is `spec/models/concerns/purchase/reportable_spec.rb:164`, "chargeback attribution is unchanged by the refund cutover", expecting 0 and getting 100. `Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER` is the fixed calendar date `Date.new(2026, 7, 27)`, and the example sets `chargeback_date: Time.current`. As of midnight UTC today that date is in the past, so the chargeback now qualifies as event-dated, `tax_reporting_cents` stops zeroing the sale, and the assertion for the legacy treatment inverts. It fails on every open PR from today onward and has nothing to do with thumbnails or cache keys.

#6371 already has the one-file fix (pin the chargeback date in that spec instead of using `Time.current`). Once that merges I'll rerun the failed shard here rather than duplicating the fix on this branch.

The earlier failure on the first run was a different midnight artifact — the mobile analytics day/week buckets, described in the previous comment. Every other shard on this run is green, and the many "fail" entries alongside it are cancellation collateral from the shard that exited non-zero (those jobs report 0 failures in their own logs).

Auto-merge stays armed; it won't land until the required checks are actually green.

gumclaw:
Correcting the plan in my last comment: rerunning the failed shard here would not have helped. A rerun executes at this branch's head SHA, which doesn't contain #6371's spec fix, so it would have failed on the same example again.

What will actually clear it is merging `main` into this branch once #6371 lands, so CI runs at a SHA that includes the pinned chargeback date. That's armed and will happen automatically when #6371 merges.

Not touching #6371 itself — it's a sibling session's PR and merging it isn't mine to do.

gumclaw:
@nyomanjyotisa routing this to you for review — green at head, mergeable, and everything on my side is done.

**The judgement call is cache-invalidation breadth.** Adding `touch: true` on thumbnails busts the profile cache when a thumbnail changes, which fixes stale thumbnails on the profile — but it also means every thumbnail write bumps the product's `updated_at`, so anything keying off that gets invalidated too. Worth confirming that's acceptable rather than too broad. The body includes a load-bearing check: with `touch: true` reverted, the three new examples fail.

No rush beyond the usual 24h review goal; if this isn't your area or you're at capacity, say so and I'll re-route rather than let it sit.

PR #6371, opened 2026-07-27
Pin the chargeback date in the refund-cutover spec instead of using Time.current

gumclaw:
A spec in `spec/models/concerns/purchase/reportable_spec.rb` started failing on every open PR at midnight UTC on 27 July 2026. It is unrelated to any of those PRs.

## The failure

```
Purchase::Reportable#price_cents_for_tax_reporting when the purchase is chargedback and not reversed
  returns 0 (chargeback attribution is unchanged by the refund cutover)
  expected: 0
       got: 100
```

## Why it started now

`CHARGEBACK_REPORTING_CUTOVER` is the fixed calendar date `Date.new(2026, 7, 27)`.

The example asserts the *legacy* chargeback treatment, where a lost chargeback zeroes the sale's reported amounts. `tax_reporting_cents` only takes that path when `chargeback_event_dated_for_tax_reporting?` is false, and that method returns false only when `chargeback_date` is before the cutover.

The example set `chargeback_date: Time.current`. Before 27 July that was pre-cutover and the legacy path ran, giving 0. From 00:00 UTC on 27 July `Time.current` is on/after the cutover, so the chargeback became event-dated — the sale is reported gross with the clawback as a separate negative leg — and the method correctly returns 100.

So the assertion silently changed which branch it was exercising as the wall clock moved.

## The fix

Pin `chargeback_date` to one day before `CHARGEBACK_REPORTING_CUTOVER` so the example keeps testing the legacy path it was written for.

Production behaviour is correct and unchanged; this only removes a wall-clock dependency from a test fixture. The nearby `Time.current` on line 17 is left alone — it exercises `price_cents_net_of_refunds`, which does not consult either cutover.

## Verification

Red-first was confirmed from CI rather than locally: the same example fails identically on unrelated PRs #6362 (`b9ffe382c`) and #6364 (`24f089049`), neither of which touches this spec, this concern, or anything tax-related. Both branches show it as their only test failure across otherwise-clean shards. Once this lands, those PRs should go green without changes of their own.

Note that `main`'s own recent CI runs only executed report jobs, not test shards, which is why this landed unnoticed on main and only surfaced on PR branches.

### DIFF
--- a/spec/models/concerns/purchase/reportable_spec.rb
+++ b/spec/models/concerns/purchase/reportable_spec.rb
@@ -162,7 +162,14 @@ describe Purchase::Reportable do
       end
 
       it "returns 0 (chargeback attribution is unchanged by the refund cutover)" do
-        purchase.update!(chargeback_date: Time.current)
+        # This example is about the REFUND cutover leaving legacy chargeback handling alone, so
+        # the chargeback has to stay on the legacy path. That path is selected by comparing
+        # chargeback_date against CHARGEBACK_REPORTING_CUTOVER, a fixed calendar date. Using
+        # Time.current here silently changed the case under test the moment the wall clock
+        # reached that date: the chargeback became event-dated, which reports the sale gross
+        # (100) rather than zeroing it, and the example started failing on every branch.
+        # Pin the date before the chargeback cutover so the legacy path is chosen explicitly.
+        purchase.update!(chargeback_date: Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER.beginning_of_day - 1.day)
 
         expect(purchase.price_cents_for_tax_reporting).to eq(0)
       end


gumclaw:
Same fix landed in #6373 independently ~20 minutes after this one — closing mine as the duplicate and deferring to this PR. Adding the verification I did there, since it corroborates this diff from a couple of angles this description doesn't cover.

**Blast radius.** `reportable_spec.rb:164` was the single real failure on #6361, #6362, #6364 and #6365 as well, all with `behind_by=0` against a green `main` and no tax-reporting changes in their diffs. Worth knowing when triaging those PRs: each one also shows ~15 *cancelled* shards from fail-fast, so `gh pr checks` prints a dozen "fail" lines and reads like a broad breakage when it's one time-dependent example.

**Production is unaffected.** The constant's contract ("must be STRICTLY AFTER the day the change reaches production") held — production is on `b07bff773`, deployed 2026-07-26, and `CHARGEBACK_REPORTING_CUTOVER` is 2026-07-27. Spec-only, nothing to bump.

**Sweep for siblings.** `grep -rln "_for_tax_reporting\|for_tax_period_reporting\|chargeback_event_dated" spec/` returns only this file and `spec/sidekiq/generate_fees_by_creator_location_report_job_spec.rb`; the latter already anchors to the constant (`let(:cutover) { ...CHARGEBACK_REPORTING_CUTOVER.beginning_of_day }`). This example was the only clock-dependent instance. Line 17's `Time.current` is correctly left alone — it covers `price_cents_net_of_refunds`, which has no cutover branch.

**Direct verification of the method.** Local runs of this spec file are unusable in my worktree — every example dies in the `:purchase` factory with `Validation failed: We couldn't charge your card`, including untouched ones like line 122 that pass in CI, so that's the harness rather than the diff. I checked the logic against the real method instead, bypassing the factory (`bundle exec rails runner`, `RAILS_ENV=test`, verbatim):

```
cutover=2026-07-27 00:00:00 UTC  now=2026-07-27 00:46:01 UTC
OLD spec value (Time.current)      chargeback_date=2026-07-27 event_dated?=true  price_cents_for_tax_reporting=100
NEW spec value (cutover - 1.day)   chargeback_date=2026-07-26 event_dated?=false price_cents_for_tax_reporting=0
```

The old value reproduces CI's `got: 100` exactly and the new value produces the asserted `0`, which is the same date this PR pins to.

gumclaw:
Superseded by #6374 — same one-line fix, but from a branch in the org repo. This PR is fork-headed (`gumclaw/gumroad`), and fork-headed PRs on this repo can never go green: CI runs under `pull_request_target`, which refuses to check out fork code, so the run here has been sitting in `waiting` with no test run at all. #6374 has real CI in flight.

PR #6374, opened 2026-07-27
Stop dating a chargeback spec off Time.current, which broke at the cutover

gumclaw:
## What

`spec/models/concerns/purchase/reportable_spec.rb:164` sets its chargeback's date with `Time.current`. That example is about the **refund** cutover, so the chargeback has to stay on the legacy attribution path — but `Time.current` only kept it there while "now" happened to precede `Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER`, which is `Date.new(2026, 7, 27)`.

That date has now arrived. `chargeback_event_dated_for_tax_reporting?` returns false only when `chargeback_date < chargeback_reporting_cutover_time`, so as of today the purchase takes the event-dated path and `price_cents_for_tax_reporting` returns the gross amount instead of 0:

```
rspec ./spec/models/concerns/purchase/reportable_spec.rb:164
# Purchase::Reportable#price_cents_for_tax_reporting when the purchase is
#   chargedback and not reversed returns 0
expected: 0
     got: 100
```

The fix dates the chargeback off the constant (`CHARGEBACK_REPORTING_CUTOVER.beginning_of_day - 1.day`) the way every other cutover spec in the suite already does, so the example tests the refund cutover instead of today's date.

## Why

This fails on every branch, with no code change behind it — it is a date rolling over, not a regression. I hit it on an unrelated PR (#6365, a receipt tax-label wording change that does not touch this file at all: 0-line diff against `reportable.rb` and its spec, and no reference anywhere in them to anything that PR changes). The same failure will redden any run started from today onward, including main, so it is worth fixing on its own rather than inside whichever PR happens to notice it.

Every other spec that straddles this cutover — in `create_vat_report_job_spec.rb`, `generate_sales_report_job_spec.rb`, `create_global_sales_tax_summary_report_job_spec.rb`, `create_india_sales_report_job_spec.rb`, `generate_canada_sales_report_job_spec.rb`, `upload_us_states_sales_tax_to_taxjar_job_spec.rb`, `create_us_states_sales_summary_report_job_spec.rb`, `create_canada_monthly_sales_report_job_spec.rb`, `generate_fees_by_creator_location_report_job_spec.rb`, and elsewhere in this same file — already derives its dates from `CHARGEBACK_REPORTING_CUTOVER`. This one example was the outlier.

I swept the rest of the suite for the same shape (`chargeback_date: Time.current` / `Time.zone.now` / `DateTime.current`). The other hits are in specs that never reach the cutover branch — `review_reminder_job_spec`, `purchase_installments_spec`, `disputable_spec`, `blockable_spec`, `customer_presenter_spec` and friends — plus `reportable_spec.rb:17`, which asserts on `price_cents_net_of_refunds`, a method that does not consult the chargeback cutover at all. Those are safe and left alone.

## Before / After

Not applicable — test-only change, no user-visible surface. The behaviour under test is unchanged; only the date the fixture uses moves.

## Test Results

This spec file cannot be executed on my machine: every example in it dies in `create(:purchase)` with `Validation failed: We couldn't charge your card` (a local Stripe/environment limitation that also makes `charge_spec.rb` unrunnable here), on this branch and on unmodified `main` alike. I am not going to dress that up as a passing local run.

Instead I verified the actual predicate the fix turns on, through `bin/rails runner` against the test environment:

```
cutover=2026-07-27  today=2026-07-27
OLD spec value (Time.current)  >= cutover ? true   <- event-dated path, returns 100 (the failure)
NEW spec value (2026-07-26)    >= cutover ? false  <- legacy path, returns 0 (the expectation)
```

CI on this PR is the authoritative check that the example is green again.

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5`.

Instructions given: while shipping an unrelated receipt tax-label fix, CI failed on a single spec twice in a row. The agent was asked to determine whether the failure belonged to that change or not; it traced the failure to this date-dependent fixture, confirmed the diff under review did not touch the file, verified the cutover arithmetic against the live constant, swept the suite for the same pattern, and opened this as a separate PR.

### DIFF
--- a/spec/models/concerns/purchase/reportable_spec.rb
+++ b/spec/models/concerns/purchase/reportable_spec.rb
@@ -162,7 +162,11 @@ describe Purchase::Reportable do
       end
 
       it "returns 0 (chargeback attribution is unchanged by the refund cutover)" do
-        purchase.update!(chargeback_date: Time.current)
+        # The point of this example is the REFUND cutover, so the chargeback has to stay on the
+        # legacy attribution path. Date it before the chargeback cutover explicitly: using
+        # Time.current made the example pass only while "now" happened to precede that date, and
+        # it started failing on its own the day the chargeback cutover arrived.
+        purchase.update!(chargeback_date: Purchase::Reportable::CHARGEBACK_REPORTING_CUTOVER.beginning_of_day - 1.day)
 
         expect(purchase.price_cents_for_tax_reporting).to eq(0)
       end


gumclaw:
Surfaced while CI on #6365 failed twice on this spec. That PR is a receipt tax-label wording change and does not touch this file; details and the cutover arithmetic are in the description here.

gumclaw:
Consolidating the context from the two earlier attempts at this same one-line fix (#6373, closed as a duplicate; #6371, closed because it was fork-headed and fork PRs on this repo can never check out under `pull_request_target`, so its CI sat in `waiting` and never ran a test). This is the surviving PR — org-repo branch, real CI in flight.

**Blast radius.** The example inverted on wall-clock time, not on anyone's diff, so it went red simultaneously on every branch whose CI crossed 00:00 UTC on 2026-07-27: #6361, #6362, #6363, #6364, #6365. Each of those runs shows the same shape — one real failure in `Test Fast 0` and the rest of the matrix cancelled by fail-fast. On #6363's run, for example: 1 failure, 72 successes, 1 skipped across 74 jobs.

**Why it flipped.** `tax_reporting_cents` zeroes a charged-back sale only under the legacy treatment:

```ruby
return 0 if chargedback_not_reversed? && !chargeback_event_dated_for_tax_reporting?
```

`chargeback_event_dated_for_tax_reporting?` returns false while `chargeback_date < CHARGEBACK_REPORTING_CUTOVER.beginning_of_day`. The constant is 2026-07-27. The spec set `chargeback_date: Time.current`, so the moment the clock reached the constant the chargeback became event-dated, the sale is reported at gross, and the assertion of 0 met 100.

**Production is not affected.** The two cutover constants carry a deploy contract — the date must be strictly after the day the change reaches production. The chargeback event-dating code is on main and deployed well before today, so the contract held; this is a spec-only time bomb, not a reporting bug.

**Sibling sweep.** I grepped the reporting specs for other clock reads sitting next to a cutover branch:

```
grep -rn "Time.current\|Date.today" spec/models/concerns/purchase/reportable_spec.rb
```

The remaining clock reads are in examples whose methods have no cutover branch (`price_cents_net_of_refunds` and friends ignore both cutovers), so they are safe to leave alone. Every example that does depend on a cutover already anchors to the constant with `± 1.day` — which is exactly what this change makes the chargeback example do too.

gumclaw:

Ran the local fable-5 adversarial review. Verdict: **SHIP**, no P1s or P2s, one process follow-up.

What it independently confirmed:

**Diagnosis is the real root cause, not a cargo-culted green.** Traced `tax_reporting_cents` end to end: with the old fixture `chargedback_not_reversed?` is true, but `chargeback_event_dated_for_tax_reporting?` now *also* returns true, so the `return 0` guard is skipped and execution falls through to `gross_amounts_for_tax_reporting?` — which is true because the example pins `created_at` to refund-cutover + 1 day. Hence gross 100. The purchase's `created_at` explains why the failure value is 100 specifically; the *flip* is entirely the chargeback date crossing the chargeback cutover.

It also pinned the timing more precisely than I did: `config.time_zone` is commented out, so the app runs in UTC and the cutover was crossed at **20:00 EDT on 2026-07-26** — 51 minutes before I committed the fix. That matches the first red run exactly.

**The new fixture pins the path permanently, not just today.** Both sides of the comparison now derive from the same constant via the same `Date#beginning_of_day` in the same zone, so it is strictly less by a full day and immune to timezone choice, DB round-trip precision, and any future bump of the constant — the fixture moves with it.

**The example is not vacuous after the change.** No refunds exist and the price is 100, so if the legacy-zeroing guard regresses the method returns gross and this example fails. The opposite branch (event-dated chargeback keeps its gross) is separately covered further down the same file, so the pair still pins both sides of the cutover.

**Sweep was complete.** It verified my claim about `reportable_spec.rb:17` by reading the method — `net_of_refunds_cents` consults only `chargedback_not_reversed_or_refunded?`, neither cutover appears in that path, so it returns 0 for any chargeback date forever. It then checked the specs of all 12 files in `app/` that read the chargeback cutover: every one pins its dates via `travel_to` at offsets computed from the constants, including the `Time.current` hits in `create_vat_report_job_spec.rb`, which sit inside `travel_to` blocks. The remaining ~90 `chargeback_date: Time.current` fixtures feed code that reads only `chargedback?`/`chargedback_not_reversed?` — date-value-independent, nothing else straddling a future instant.

**The P3 follow-up**, which is not a change to this PR: confirm from deploy history that everything reading `CHARGEBACK_REPORTING_CUTOVER` was live before 2026-07-27 00:00 UTC, per the DEPLOY DEPENDENCY comment on the constant. I checked this before opening the PR: the constant and the feature landed in #5939 on 2026-07-20, the last consuming code merged 07-26, and releases shipped continuously through 07-26 (latest `v2026.07.27.1` at 00:07 UTC). So the code was in production well before the date activated — the constant did not need bumping, and this spec failing carried no information about deploy state either way, since the old fixture fails on wall clock alone on any machine.

Worth recording the reviewer's closing point: the window to bump that constant has now closed. Moving the date after chargebacks have started arriving under it would re-partition already-classified events, so if a deploy slip ever *is* discovered here it is an incident to handle deliberately, not by editing the date.

gumclaw:
## Self-review

Re-read CONTRIBUTING.md against this diff rather than from memory. It is one fixture line plus a comment, so most of the checklist resolves quickly, but two items deserve a real answer rather than a tick:

**"Tests must fail when the fix is reverted. If the test passes without the application code change, it is invalid."** This PR has no application code change — it fixes a test that had started failing on its own. The rule still applies in spirit, and the question that matters is the inverse: *does the example still fail if the behaviour it guards regresses?* It does. The purchase has no refunds and a price of 100, so the only paths that produce 0 are the legacy-chargeback guard itself or negative netting, and there is nothing to net. Delete that guard, or wrongly event-date a pre-cutover chargeback, and this example returns 100 and goes red. I checked this specifically because "make CI green by editing the fixture" is exactly the shape of a change that quietly neuters a test.

**"Video required for every PR."** Skipping it. The exception in CONTRIBUTING is for changes whose diff is the reviewable artifact, and a one-line fixture date with no user-visible surface is squarely that — there is nothing to film. The evidence that carries this PR is the cutover arithmetic in the description, printed from the live constant.

On the change itself:

**Why the constant and not a literal.** Writing `Date.new(2026, 7, 26)` would have fixed today and planted the same bomb for whoever next bumps `CHARGEBACK_REPORTING_CUTOVER`. Deriving from the constant means the fixture tracks it. This also matches what every other cutover spec in the suite already does, so it removes an outlier rather than introducing a new idiom.

**Why the comment earns its place.** Per the repo's comment guidance it explains the why, not the what: that this example is about the *refund* cutover, that the chargeback therefore has to stay on the legacy path, and that the old form only worked while "now" happened to precede a date. Without that last sentence the next reader has no way to know the line is load-bearing and will "simplify" it back to `Time.current`.

**What I could not verify locally, stated plainly.** This spec file does not run on my machine — all 31 examples die in `create(:purchase)` with `Validation failed: We couldn't charge your card`, on this branch and on unmodified `main` alike. I did not paper over that with a fabricated pass; I verified the predicate the fix turns on via `bin/rails runner` instead, and CI is the authoritative check that the example is green.

One thing I want a reviewer's eye on: I chose `- 1.day` rather than something like `- 1.second`. A full day is the coarser, safer margin against timezone and precision effects, and it keeps the chargeback (07-26) chronologically after the purchase (07-21), which the narrower offset would also satisfy but less obviously. If the team prefers cutover-relative offsets expressed in seconds for tightness, say so and I will change it.

gumclaw:
CI green: **73 jobs passed, 0 failures** (`gh pr checks` exit code 0), Greptile approved with no findings.

That is the confirmation the diagnosis needed. `Test Fast 4` was deterministically red on the PR that surfaced this — the same spec, same assertion, `expected: 0 got: 100`, twice across an initial run and a `--failed` rerun — and it is green here with nothing changed but the date the fixture uses. No production code moved.

Handing this to you, Ershad, as a fast one-line unblock: this spec is red on every branch cut from today onward, including main, so merging it clears CI repo-wide rather than just for #6365.

PR #1397, opened 2025-09-29
Fix: Custom Fields not updating when edited in Manage Membership page

Pradumn27:
Fix #1321 

### Before

https://github.com/user-attachments/assets/93a9662f-047d-4bed-b6f2-135240234baf

### After

https://github.com/user-attachments/assets/94251d38-2a6e-4844-b8ad-5edbe2225f4a

### Test

Screenshot 2025-09-29 at 5 33 44 PM

### AI Disclosure

Used AI for finding the fix and also writing tests

PR #6376, opened 2026-07-27
Require a CAPTCHA on free checkouts when the seller is not in good standing

gumclaw:
Second of two defenses against the receipt-mailer abuse in gumroad-private#1397, where a seller scripted **150,529 free checkouts** in four days and mailed **148,535 scraped addresses** through `noreply@customers.gumroad.com` — our transactional mailer, our sending reputation, our branding on the spam.

The rate-limit half is #6368. This is the CAPTCHA half, and it's independent: they stop different attackers.

## Why nothing stopped it

We already require a CAPTCHA on free orders — but **only when the seller's account is younger than six months**.

Account age was the wrong test. **The spammer's account was 1,502 days old.** Over four years. They cleared the check purely by being old.

What actually distinguished them was risk state: they were **flagged**, not compliant.

## The change

`Link#require_captcha?` now returns true when the seller is not `compliant`, in addition to the existing age rule.

```ruby
def require_captcha?
  return true if user.created_at > REQUIRE_CAPTCHA_FOR_SELLERS_YOUNGER_THAN.ago

  !user.compliant?
end
```

This deliberately includes `not_reviewed` — the default state for the long tail of accounts nobody has looked at. For a **free** order, an unverified seller is exactly the case where a cheap bot check earns its keep. Suspended sellers can't transact at all, so they never reach this path.

## Blast radius

Measured against 30 days of successful $0 orders in production:

| | orders | |
|---|---|---|
| Total free orders | 3,022,197 | |
| From compliant sellers | 2,562,180 | **unaffected (85%)** |
| Already CAPTCHA'd by the age rule | 174,464 | no change |
| **Newly CAPTCHA'd** | **143,839** | **4.8%, across 8,571 sellers** |

The common path — a reviewed, cleared seller giving away a free file — stays frictionless.

## No frontend change needed

I verified checkout already routes **every** order through the `captcha` state and produces a token regardless of price: `PaymentForm.tsx` calls `recaptcha.execute()` whenever `status.type === "captcha"`, and the reducer sets that state for all carts, free or paid. `require_captcha?` is consulted server-side only.

So free carts were **already carrying a token we were choosing not to verify**. This just verifies it.

## Why this and not just the rate limit

They cover different attackers:

- **#6368's rate limit** stops a *fast* flood, but a patient script staying under 300/min slips through.
- **This CAPTCHA** stops *automation regardless of pace*, but only for sellers we haven't vouched for.

Together, abuse has to beat both a bot check and a volume ceiling, while a compliant seller's launch is untouched by either.

## Tests

**Red-first verified** — the not-compliant cases fail against unpatched `link.rb` and pass with it.

- Young seller → required even when compliant (existing rule preserved)
- Established **compliant** seller → not required (the 85% path stays clean)
- Established seller in `not_reviewed` / `flagged_for_tos_violation` / `flagged_for_fraud` / `on_probation` → required — **this is the #1397 shape**

Also updated the pre-existing age-only test: it created users without setting risk state, so it inherited the `not_reviewed` default and would now expect the wrong answer. It sets `compliant` explicitly to isolate the age dimension it actually tests.

`test/models/link_test.rb`: **420 runs, 928 assertions, 0 failures.**

Refs gumroad-private#1397

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -508,8 +508,36 @@ class Link < ApplicationRecord
   end
   alias_method :streamable, :streamable?
 
+  # Whether a $0 checkout of this product must still solve a CAPTCHA.
+  #
+  # Free checkouts are the cheapest thing to automate on Gumroad: there is no card to
+  # decline and no money at risk, but every completed checkout still sends a receipt from
+  # our transactional mailer. That makes a free product an attractive way to relay mail
+  # through Gumroad's sending reputation.
+  #
+  # Two independent reasons to require the check:
+  #
+  # 1. NEW SELLER. An account younger than REQUIRE_CAPTCHA_FOR_SELLERS_YOUNGER_THAN has no
+  #    track record, so we ask for the check by default.
+  #
+  # 2. SELLER IS NOT IN GOOD STANDING. Account age alone turned out to be the wrong test:
+  #    in July 2026 a seller whose account was over four years old scripted 150,529 free
+  #    checkouts of their own product in four days, mailing 148,535 scraped addresses
+  #    (gumroad-private#1397). They sailed past the age check purely by being old. What
+  #    actually distinguished them was risk state — they were flagged, not compliant.
+  #
+  #    So we also require the check whenever the seller is not `compliant`. Note that
+  #    `not_reviewed` is the DEFAULT state for the long tail of accounts that have simply
+  #    never been looked at, so this deliberately covers "unknown" as well as "known bad" —
+  #    for a free order, an unverified seller is exactly the case where a cheap bot check is
+  #    worth it. Suspended sellers cannot transact at all, so they never reach here.
+  #
+  # Compliant sellers — reviewed and cleared, and the source of ~85% of legitimate free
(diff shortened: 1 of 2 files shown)


gumclaw:
@xrav3nz routing this to you alongside #6368 — same incident (gumroad-private#1397), two independent halves, so it makes sense for one person to hold both. Green at head, mergeable.

**The judgement call is blast radius on `not_reviewed` sellers.** Gating the CAPTCHA on `!user.compliant?` deliberately catches `not_reviewed`, which is the default state for the long tail of accounts nobody has ever looked at — so this is not a small set. The argument is that it's a **free** order and a CAPTCHA is cheap, but it does mean legitimate unreviewed sellers get friction on $0 checkouts they didn't have yesterday. That trade is the thing to sanity-check.

Worth noting what made the original test useless: the abusive account was **1,502 days old**, so an account-age gate waved it straight through. Risk state was the signal that actually separated them.

No rush beyond the usual 24h review goal; if you'd rather this went to someone else, say so and I'll re-route.

PR #6379, opened 2026-07-27
Pin the per-purchase fee split from the multi-item forced-currency work

gumclaw:
## What

Follow-up to #6251, addressing findings from a local adversarial pre-merge review that finished after that PR merged. Three test/observability gaps, no behavior change on any buyer-facing path:

- Replace the tautological assertion in the Gumroad fee-split spec with per-purchase expected values.
- Add the two missing boundary examples: a multi-item all-USD cart must stay out of the forced-currency path, and an empty purchase set returns the new `:no_purchases` reason.
- Log the one fail-closed path in `Order::PreparePaymentIntentService` that skipped the presentment service silently.
- Make `uniform_method_forced_purchase_currency` guard the collection it actually reads.

## Why

**The fee-split spec proved nothing.** It asserted that the per-purchase presentment shares sum to the charge-level share — but `Charge::MethodForcedPresentment#direct_listed_amount_result` computes the charge-level value *as* the sum of those very shares, so the assertion held no matter what each purchase was given. The single genuinely new money line in #6251 (`canonical_gumroad_amount_cents_for`, which switches from the charge-level figure to the per-purchase one when a charge has more than one purchase) had no value-pinning test at all.

Concretely: giving every purchase the whole charge's cut, which double-counts Gumroad's take on a two-item cart, passed every assertion the spec had. Payouts and refunds read the per-purchase rows, so that class of bug would silently disagree with what was charged. The spec now asserts each purchase's own expected value — its own fee plus affiliate credit plus Gumroad tax, converted at the rate that purchase stored — and I verified it is load-bearing by making exactly that mutation (see Test Results: it fails with `[301, 301]` against the expected `[191, 110]`).

**The multi-item USD boundary was unpinned.** `Checkout::BuyerCurrencyEligibility` allows a single USD-priced line onto the quoted-FX path, and that allowance is deliberately narrow: two USD lines would each need their own quote before the quoted amounts could reconcile with the persisted purchase rows. Nothing pinned that shut, so relaxing the `purchases.one?` conjunct later would have let a multi-line USD cart into the quoted lane with no test failing. `:no_purchases`, introduced by #6251, also had no assertion anywhere.

**One fail-closed path was silent.** When a free or test line is not priced in the forced currency, `method_forced_presentment_for` returns nil *before* instantiating `Charge::MethodForcedPresentment`, which logs its own fallbacks. The caller turns that nil into a generic charge error, so an operator investigating "iDEAL checkouts fail for this cart shape" had nothing to read.

Refs #6251, antiwork/gumroad-private#1298.

## Before/After

Not applicable — no user-facing surface changes. This diff is two spec files plus one log line and one guard rename in a service; nothing renders. The checkout behavior it covers is unchanged from #6251, whose hosted-preview evidence (including completed card and iDEAL payments on multi-item EUR carts) is in that PR's body.

## QA steps

There is nothing to click. The change is proven by the specs below, in particular the mutation test:

1. Revert `canonical_gumroad_amount_cents_for` to return the charge-level `gumroad_amount_cents` for every purchase.
2. Run `bundle exec rspec spec/services/order/prepare_payment_intent_service_spec.rb -e "splits Gumroad's presentment fee share"`.
3. Before this PR: passes. After this PR: fails with `expected [191, 110], got [301, 301]`.

## Test Results

local verification

- `DISABLE_SPRING=1 bundle exec rspec spec/services/checkout/buyer_currency_eligibility_spec.rb spec/services/order/prepare_payment_intent_service_spec.rb spec/services/charge/method_forced_presentment_spec.rb` — 150 examples, 1 failure.
- That one failure is `Checkout::BuyerCurrencyEligibility allows the PR1 Stripe test-mode Gumroad platform-account path`, an `ActiveRecord::RecordInvalid` from a merchant-account fixture collision in my local database. It is not touched by this diff, passes when the file runs in isolation, and reproduces on unmodified `origin/main`. CI is the authority on it.
- Mutation check on the fee-split spec: reverting `canonical_gumroad_amount_cents_for` to the charge-level value reddens the example (`[191, 110]` expected, `[301, 301]` got). Under the old assertions that same mutation passed.
- `bundle exec rubocop -a` on the three changed files — no offenses.

---

## AI disclosure

Model: `claude-opus-5`.

Prompts/instructions: run the mandatory local adversarial pre-merge review on #6251 before it merged; the review returned MERGE-AFTER-FIXES with no P1 and four actionable findings. #6251 merged while that review was running, so the findings were carried into this follow-up: pin the per-purchase fee split with real expected values and prove the assertion load-bearing by mutation, add the missing multi-item-USD and empty-purchases boundary examples, log the silent free-line currency-mismatch path, and correct the guard in `uniform_method_forced_purchase_currency` to read the collection it guards.


gumclaw:

**Scope.** Three files: one log line and one guard change in `Order::PreparePaymentIntentService`, and assertions in two spec files. No migration, no new dependency, no public API change, no user-facing surface. The one production behavior change is an added `Rails.logger.info` — it cannot alter what is charged, and the guard change (`purchases_to_charge.empty?` → `charge_purchases.empty?`) is provably equivalent today because `perform` returns before either collection is consulted when there are no paid lines.

**The fee-split assertion.** I want to be explicit about why the old spec was worth replacing rather than extending. `presentment_gumroad_amount_cents` on the charge row is assigned `allocations.sum(&:presentment_gumroad_amount_cents)`. Asserting that the parts sum to that whole is asserting `x == x`. The replacement asserts each purchase's own value, recomputed independently from `fee_cents + affiliate_credit_cents + gumroad_tax_cents` at that purchase's own `rate_converted_to_usd` — which is exactly the composition `Purchase#total_transaction_amount_for_gumroad_cents` returns and the payout path later reads. I also added `expect(purchase_shares).to all(be < charge_presentment.presentment_gumroad_amount_cents)`, which is the specific shape of the bug the sum assertion cannot see.

I did not take the assertion on trust. Reverting `canonical_gumroad_amount_cents_for` to hand every purchase the charge-level figure now reddens the example; under the previous assertions that same mutation was green. The terminal capture in the body is that run.

**The boundary specs.** `withholds the method for multi-item USD carts` asserts `product.price_currency_type` is USD inline, so the example cannot silently stop testing the USD case if the shared `product` fixture is ever repointed at another currency. The `:no_purchases` example passes `purchases: []` explicitly rather than relying on any fixture state.

**Attribution.** The commit carries `Co-authored-by: Razvan Marescu <rmarescu@gumroad.com>` — the review that surfaced the multi-item shape is his.

**What I did not do.** I did not touch `Charge::MethodForcedPresentment` itself. The review found its math correct: each allocation converts that purchase's own USD-stored components with that purchase's own stored rate, the charge total is literally the sum of the allocations, and the intent amount is that same value, so there is no per-line versus whole-charge rounding that can diverge. The gap was coverage, not correctness, and I kept the fix to coverage.

The `preview` label is deliberately absent: nothing renders, so a preview app would produce no evidence a reviewer could act on. The proof for this diff is the mutation test.

PR #6380, opened 2026-07-27
Emit a cover image fallback in the custom-page profile payload

gumclaw:
## What

`Pages::ProfileData` now emits a `cover_url` for each product alongside `thumbnail_url`, and the default storefront document falls back to it when a product has no thumbnail. The store agent's system prompt states the payload's field list as closed and explains the image fallback, so generated pages stop referencing fields that do not exist.

Details:
- `cover_url` is the product's first cover that is actually an image (video and oEmbed covers are skipped, since only an image can go in an `` tag). This mirrors `Link#thumbnail_or_cover_url`, which every other product card in the app already uses.
- Unsplash-hosted thumbnails and covers are emitted as `nil` rather than as their Unsplash URL. Custom pages are served under `RendersCustomHtmlPages::CUSTOM_HTML_CSP`, whose `img-src` allows only Gumroad's own asset hosts, so an Unsplash URL would be blocked and render as the same broken image this PR is fixing. Treating it as absent lets the page fall through to its next option.
- `CACHE_VERSION` is bumped to `v4` so already-cached payloads built without the new key are not served.
- The products query eager-loads `display_asset_previews` with their attachments, so the added cover read does not turn into a per-product query on the profile path. Cache invalidation already works: `AssetPreview belongs_to :link, touch: true`, so adding or removing a cover moves `seller.products.cache_key_with_version` and therefore the payload cache key. There is a spec covering that.

## Why

Generated custom profile pages render product cards as `src="${product.thumbnail_url || product.cover_url || ''}"`, but the injected `gumroad-data` JSON only carried `name`, `url`, `price`, `native_type`, `thumbnail_url`, and `description`. On a product with no thumbnail both sides of that fallback were empty, so the card rendered a broken-image icon. Products with a thumbnail were unaffected.

Two directions were possible — add the cover to the payload, or constrain page generation so it only references fields the payload carries. The direction call on the issue was to add the cover: a cover image is better than no image, and because published pages read the injected JSON at render time, adding the key repairs pages already in the field with no regeneration or backfill. The prompt change is the second half, so the agent cannot invent another field name later.

Tracking issue: antiwork/gumroad-private#1401 (split from antiwork/gumroad-private#1398, which fixed a separate cache-invalidation bug on the same surface).

## Before / After

The same generated page markup, rendered in a real browser against the payload before and after this change. The product has a cover image and no thumbnail.

| Before | After |
|---|---|
| before | after |

Payload for that product, before and after:

```
before: {"name" => "Cover-only product", "thumbnail_url" => nil}
after:  {"name" => "Cover-only product", "thumbnail_url" => nil, "cover_url" => "…/gumroad-specs/pt520oiza4en3nski9f1q2ndznzd"}
```

## QA steps

1. As a seller with the custom HTML pages feature, upload a cover image to a product and make sure it has no thumbnail.
2. Ask the store agent for a custom profile page (or publish one whose product cards read `thumbnail_url || cover_url`).
3. Visit the profile. The thumbnail-less product's card shows its cover image instead of a broken-image icon.
4. Remove the cover so the product has neither image. The card renders without an `` rather than with an empty `src`.
5. Add a cover back and reload — the card picks it up on the next render (the payload cache key moves with the product).

## Test Results

`bundle exec rspec spec/services/pages/profile_data_spec.rb spec/services/pages/default_profile_document_spec.rb` — 21 examples, 0 failures, run locally on this branch.

test results

New coverage: thumbnail plus first image cover, cover-only product, product with neither image, non-image cover skipped, Unsplash images skipped, newly added cover not served from stale cache, and the default storefront document falling back to the cover and omitting the `` when there is no image.

Also ran the neighbouring suites on this branch: `spec/services/ai/store_agent_service_spec.rb`, `spec/controllers/api/internal/agent_custom_html_previews_controller_spec.rb`, and `spec/requests/user/profile_custom_html_page_spec.rb`. All pass except `profile_custom_html_page_spec.rb:56`, which fails identically on unmodified `main` in this local worktree (it needs built front-end assets that are not present here) — verified by stashing the change and re-running. CI covers it.

---

**AI disclosure:** written by Hermes Agent running `claude-opus-5`. Instructions given: implement the payload direction chosen on the linked issue (emit a cover fallback rather than constraining page generation), keep the generated pages already in the field working without regeneration, close the agent prompt's field list so it stops inventing key names, add spec coverage for the payload keys that become part of the contract, and capture real rendered before/after evidence.


gumclaw:
Self-review of this diff against CONTRIBUTING.md:

- **`cover_url` naming.** Kept the name the already-published pages reference, rather than picking a "better" one, because matching it is what repairs those pages in place with no regeneration. That is the whole point of taking the payload direction over the prompt-only direction.
- **Why the Unsplash exclusion is in here.** It is not scope creep: adding `cover_url` without it would swap one broken image for another on Unsplash-sourced covers, since the custom-page CSP's `img-src` only allows Gumroad's asset hosts. The same guard now covers `thumbnail_url`, which had the same latent problem — a product with an Unsplash thumbnail was already emitting a URL the page could not load. That is a behavior change on `thumbnail_url` and worth a reviewer's attention: such a product now falls through to its cover (or to no image) instead of showing a broken one.
- **N+1.** `display_asset_previews` is eager-loaded with `file_attachment: :blob`, because `image_url?` touches the attachment and `url` touches the blob. Without the blob preload this would be two extra queries per product on a page that renders up to 100.
- **Cache invalidation.** No new key component was needed: `AssetPreview belongs_to :link, touch: true`, so a cover change moves `seller.products.cache_key_with_version`. `CACHE_VERSION` still went to `v4` because the payload's shape changed and entries cached under `v3` lack the new key. There is a spec that adds a cover after a first read and asserts the second read sees it.
- **`nil` vs omitted key.** `cover_url` is always present and `nil` when there is no usable cover, never omitted. An omitted key reads as `undefined` in the page's JavaScript, which makes a `!== null` style guard pass and render a blank image — the failure mode this PR exists to remove.
- **Default storefront document.** `Pages::DefaultProfileDocument` is the snapshot an agent pulls as its starting point, so it had to learn the same fallback; otherwise a pulled-then-pushed page would drop cover images the live storefront shows. It still emits no `` at all when a product has neither image, rather than one with an empty `src`.
- **Prompt change.** Stated as a closed list plus an explicit "reading any other name gives undefined" line, since the model previously invented `cover_url` against a list that read as examples. Also spelled out the image fallback and the no-image case so a generated card omits the tag instead of emitting an empty `src`.
- **No PII, no schema change, no migration, no feature flag.** Nothing here is user-data-dependent or irreversible.

Not done: the local adversarial pre-merge review has not run in this session. Worth one before merge given this touches a public storefront render path.

PR #6382, opened 2026-07-27
Tell Zambian sellers PayPal cannot receive their payouts

gumclaw:
## What

Tells a Zambian seller, on the payments settings page, that we cannot actually pay them out to a Zambian PayPal account.

Zambia is not in `SUPPORTED_COUNTRIES`, so `native_payouts_supported?` is false, the bank form is never offered, and `can_setup_paypal_payouts?` returns true — leaving PayPal as the only visible option. PayPal does not let accounts registered in Zambia *receive* money, so those payouts fail. This adds a warning next to the PayPal email field for `ZM`, in the same place and shape as the existing Ukraine notice, plus a matching note in the "Getting paid" help center article.

The warning says three things: a Zambian PayPal account cannot receive a payout, bank payouts are not available in Zambia either, and the balance stays in the Gumroad account in the meantime — with the one real workaround (a PayPal account registered outside Zambia) named explicitly.

No behavior changes. The PayPal field is still offered and still savable; this is purely telling the seller what will happen.

## Why

Today the failure is completely silent to the seller. A payout to a Zambian PayPal account comes back as PayPal reason code 3148, "Receiver's address is in a non-receivable country or a PayPal zero country". `Payment` only emails the seller on `cannot_pay` and `debit_card_limit`; a 3148 writes an admin comment via `Payment::FailureReason` and marks the balance unpaid again. `humanized_failure_reason` is rendered in the admin payout view and nowhere the seller can see. So the seller watches a payout not happen, week after week, with no explanation and no reason to think anything is wrong on their end.

Production data for accounts with Zambia on file (all-time; every payout attempt is PayPal, there are no Stripe ones):

- 782 payout attempts: 127 completed, 646 failed, 7 returned, 2 reversed.
- 228 of the failures are reason code 3148, across 11 sellers. 38 of those are in 2026 alone.
- Only 19 distinct Zambian sellers have ever had a payout complete. In the last 6 months: 4 completed payouts for 1 seller, 30 failed for 6 sellers.
- 5 sellers show both completed and 3148-failed payouts, consistent with the ones who get paid holding a PayPal account registered somewhere other than Zambia.

So PayPal works there only for the minority who happen to hold a non-Zambian PayPal account. For a seller who signs up in Zambia and opens a Zambian PayPal account — the normal case — it fails every payout run.

Adding Zambia to bank payouts is not an available alternative. Checked against the live Stripe API: `GET /v1/country_specs/ZM` returns "ZM is not currently supported", ZM is absent from the 121-entry country-spec list, and it is absent from `supported_transfer_countries` on the US spec (the list that governs cross-border payouts from a US platform). A connected account with a Zambian legal entity cannot be created at all, so adding `ZM` to `SUPPORTED_COUNTRIES` would render the bank form and then fail at merchant-account creation.

Balances at stake are small — 56 Zambian sellers hold unpaid balances totalling roughly $862, none over the $100 threshold — so this is a growth blocker and a trust problem rather than a pile of stuck money. Which is exactly why the honest message is worth shipping on its own, ahead of any decision about a non-Stripe rail.

Refs #6375.

## Test plan

New specs in `spec/requests/settings/payments_spec.rb`:

- "warns a Zambian creator that PayPal cannot receive payouts there" — sets the seller's compliance country to Zambia, visits settings, asserts the PayPal email field is still present and that both halves of the warning render.
- "does not show the Zambia PayPal warning to creators in other countries" — same page for the default US seller, asserts the warning is absent. This is the guard that keeps the notice from leaking to every seller.

**Neither system spec ran locally.** In a fresh worktree the Capybara boot times out waiting for the React root to mount, and it does so identically for the untouched "just shows payment address to a US creator" example in the same file — including on a worktree with `app/javascript` reset to `origin/main`. The same example passes in a fully-provisioned checkout, so the failure is the local harness rather than this diff, and CI is the authority. I am not going to report a pass I did not see.

Lint: `rubocop` clean on the spec file; `prettier --check` and `eslint` clean on `PayPalEmailSection.tsx`. Help-center ERB tag balance verified (tr/td/table/p/div/ul/li all balanced).

## Before / after

Preview-app screenshots of the Zambian seller's payments settings page, with and without the notice, are being captured and will be added before this leaves draft. The rendered change is one warning callout below the PayPal email field, visible only when the seller's compliance country is Zambia.

---

## AI disclosure

Written by Claude Opus 5 (`claude-opus-5`) running as the Gumclaw Hermes agent.

Instructions given: open a PR indicating there is currently no way to get paid out to Zambian PayPal because of PayPal restrictions. The agent had already established the underlying facts while investigating #6375 — the production payout-failure counts above came from the audited production console, and the Stripe non-coverage was confirmed directly against the live Stripe API. For this PR it located the existing Ukraine warning as the established pattern for a country-specific PayPal notice, added the Zambia equivalent, wrote the positive and negative system specs, added the help center note, and ran the lint commands quoted above.


gumclaw:
Status on evidence, so it's on the record rather than in chat.

The preview app for this branch deployed (`preview/zambia-paypal-unsupported-notice`, deployment status `success`, `environment_url` = https://zambia-paypal-unsupported-notice.apps.staging.gumroad.org). It came up briefly — I saw one HTTP 200 during readiness polling — and has been returning HTTP 500 on every request since, including `/`, `/login` and `/settings/payments`.

This is not specific to this branch. Two sibling previews from other in-flight PRs (`feat-method-forced-uniform-curre`, `fix-colombia-id-cc-or-ce`) also return 500 right now, so the staging preview stack is unhealthy platform-wide at the moment.

So the screenshots this PR needs are blocked on preview infra, not on the change. I am not going to substitute a local mockup for them — the screenshot gate exists specifically to show the real rendered app. The PR stays draft until the preview is healthy and I can capture the Zambian seller's payments settings page with the notice, plus a non-Zambian seller's page without it.

Buildkite's test suite on this branch passed.

PR #6383, opened 2026-07-27
Document buyer-currency price-ending mirroring in the help center

gumclaw:
## What

Adds a paragraph to the "What currency does Gumroad use?" help center article (`_46-what-currency-does-gumroad-use.html.erb`) describing how a buyer charged in their own currency now sees the seller's USD price ending mirrored into that currency, and how to turn it off.

## Why

#6309 shipped price-ending mirroring for buyer-currency checkout: a $9.99 total quotes €8,99 instead of the raw converted €8,53, and a $10 total quotes €9. Article 46 was the surface describing local-currency checkout and only said charges use "real-time exchange rates," which now understates what the buyer sees. The new paragraph covers the behavior, the taxed-total case (the ending mirrored is the taxed total's), that seller earnings/taxes/payouts are unchanged because Gumroad absorbs the difference, and the "Keep price endings in local currency" toggle in Settings > Payments.

No removals. No pricing, tax, payout, or policy figures changed — the only money statement added is that the seller's proceeds are unaffected, which matches the allocator behavior in the merged PR (the delta is carried on the non-tax Gumroad lines).

## Before / After

Not applicable as a rendered UI change — this is help center copy. Before: the article described local-currency checkout without mentioning price endings. After: the paragraph above sits directly under that description.

## Test plan

- `bundle exec rspec spec/requests/help_center_spec.rb spec/controllers/help_center/articles_controller_spec.rb` — 12 examples, 0 failures (run locally on this branch).
- Tag balance checked on the edited ERB (11 `<p>`/`</p>`, 9 `<a>`/`</a>`).
- No `articles.yml` change needed: the title, description, and category are unchanged.

---

AI disclosure: written by Hermes Agent running `claude-opus-5`. It was triggered by the merge of #6309 via the help-center doc-sync automation, which instructs it to check whether a merged gumroad PR changes externally visible behavior, find the affected help center article, and ship the copy fix. The model read the merged diff (settings toggle label and description, `Checkout::PresentmentRounding`, the allocator comments), located article 46 as the only article describing local-currency checkout amounts, wrote the paragraph, and ran the help center specs.


gumclaw:
Self-review of the one-paragraph diff, checked against the merged code rather than from memory:

- **Toggle label** — the copy quotes "Keep price endings in local currency", which is the exact `label`/`aria-label` string on the Switch in `app/javascript/pages/Settings/Payments/Show.tsx`, not a paraphrase.
- **Default direction** — `User#disable_buyer_currency_rounding` defaults to `false`, so mirroring is on unless the seller turns it off. The article says "you can turn this off", which matches; it does not tell sellers to opt in.
- **Seller economics** — "earnings, taxes, and payouts are unchanged" matches `Charge::PresentmentAllocator`: the delta is carried only on the non-tax components (`ROUNDING_ABSORBING_COMPONENT_INDEXES`) so the tax lines keep the exact converted figure and the difference is Gumroad's. Same claim the in-app setting description makes.
- **Taxed-total case** — the mirrored ending is the one on the USD total being converted, so an $11.49 taxed total keeps .49. Stated that way rather than implying the bare product price's ending is used.
- **Numbers** — $9.99 → €8,99 and $10 → €9 come from the shipped behavior, not invented examples.
- **Scope** — no other article describes the amount a local-currency buyer is charged. Article 47 (refunds) says customers are refunded in the currency they were charged, which stays true. Articles 74 and 174 describe the presentment-currency analytics fields, unaffected by rounding. `articles.yml` untouched because title, description, and category are unchanged.
- **Markup** — tag balance verified on the file (11 `<p>`/`</p>`, 9 `<a>`/`</a>`); the settings link uses the same `https://gumroad.com/settings/payments` form the other articles use.
- **Tests** — help center request and controller specs run locally on this branch, 12 examples 0 failures.

No PII, no incident specifics, no internal identifiers in the copy.

PR #6385, opened 2026-07-27
Pin gift, bundle, and per-line refund behaviour for multi-item presentment charges

gumclaw:
## What

Adds spec coverage pinning three behaviours of buyer-currency (presentment) charging on multi-item single-seller carts, and records the reasoning as code comments. No behaviour change — no production code path is modified.

1. **Gift carts.** A two-item gift cart produces four purchase rows: two gifter rows that carry the money and two 0-cent giftee rows. The giftee rows are built inside `Purchase::CreateService#create_giftee_purchase` and never appended to the order, so they never reach the charge and get no `purchase_presentment`. Pinned at the charge level.
2. **Bundle carts.** Bundle child rows are created by `Purchase::CreateBundleProductPurchaseService` *after* the parent purchase succeeds, at 0 cents, so only the paying rows get presentment records. Pinned for a cart holding a bundle alongside a standalone product.
3. **Refunds on one line of a multi-item presentment charge.** `Purchase::PresentmentRefund` clamps against the *purchase's* `presentment_total_cents`, not the charge's. Pinned with a full refund, a two-step partial refund, and an over-the-line amount, all on a two-purchase presentment charge.

Comments added to `Charge::PresentmentOrchestrator.persist!` (which purchases get a presentment row, and why the two kinds of 0-cent companion row do not) and to `Purchase::PresentmentRefund.from_presentment_amount` (every balance is the purchase's, never the charge's).

## Why

Multi-item single-seller carts have charged in the buyer's currency since #5923, which lifted the `multi_item_checkout` fallback and started persisting one `purchase_presentment` per purchase. Three things were true only by construction, with nothing asserting them:

- Neither gift nor bundle child rows had ever been exercised at the charge level on a **multi-item** cart. The existing coverage (#5627) is single-item only. If a child row ever did join the charge it would enter the allocator as a zero-weight line, and the buyer would see per-item buyer-currency amounts for items they were never separately charged for. The allocator's zero-weight case was covered only with a test double, never a real gift or bundle cart.
- Every existing presentment refund example builds a purchase whose presentment *is* the whole charge, so the multi-line case was unproven. That distinction decides real money: on a CA$37.50 charge covering a CA$12.50 line and a CA$25.00 line, clamping against the charge instead of the line would let a refund of one item send back the whole cart — and Stripe would accept it, because the charge really does hold that much.

Refs gumroad-private#1317.

Separately worth naming so it does not read as a presentment bug: a **processor-initiated** partial refund on a combined charge with more than one purchase is still refused outright (notified and skipped in `Charge::Refundable#handle_event_refund_updated!`). That gap predates presentment — a dashboard refund aimed at one item cannot be attributed to a line — and is out of scope here.

## Before/After

Not applicable — no user-facing surface changes. This PR adds specs and code comments only; no production code path, template, or rendered view is touched. The behaviours pinned were already live in production as of the #5923 deploy (20 Jul).

## Test Results

New coverage, run locally on this branch:

```
$ bundle exec rspec spec/services/order/charge_service_spec.rb -e 'multi-item gift cart' -e 'multi-item cart containing a bundle' -f doc
Order::ChargeService
  #perform
    creates presentments for the gifter purchases only on a multi-item gift cart
    creates presentments for the bundle parent only on a multi-item cart containing a bundle
2 examples, 0 failures

$ bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb -e 'multi-item presentment charge' -f doc
PurchaseRefunds
  refunds on one purchase of a multi-item presentment charge
    clamps a full refund of one purchase to that purchase's presentment cents, not the charge's
    refunds a partial amount of one purchase against that purchase's own remaining presentment balance
    refuses a presentment amount larger than the purchase's line even though the charge could cover it
3 examples, 0 failures
```

spec run

**Mutation check — the refund specs are load-bearing.** Changing the clamp in `Purchase::PresentmentRefund.from_presentment_amount` to read the charge total instead of the purchase's (`presentment.charge_presentment&.presentment_total_cents || presentment.presentment_total_cents`) turns the new block red: 3 examples, 2 failures. Reverted; green again on two consecutive runs.

Surrounding suites, run locally:

- `spec/services/order/charge_service_spec.rb` — 41 examples, 1 failure. That failure (`#ensure_all_purchases_processed retries marking as successful when charge_intent succeeded but post-charge processing failed`) is **pre-existing**: it fails identically on unmodified `origin/main` in this same worktree, on `create(:merchant_account, user: nil)` hitting the "already connected with another Gumroad account" validation — a merchant-account factory-sequence collision from concurrent runs against a shared local database, unrelated to this diff.
- `spec/services/charge/presentment_orchestrator_spec.rb`, `presentment_allocator_spec.rb`, `spec/services/purchase/presentment_refund_spec.rb` — 44 examples, 1 failure, same pre-existing factory collision, also reproduced on unmodified main in the same worktree.

CI is the authority on both; neither failure touches a line this PR changes.

## QA steps

No user-facing surface to click through. To verify the claims directly:

1. `git fetch origin spec/multi-item-presentment-gift-bundle-refund && git checkout spec/multi-item-presentment-gift-bundle-refund`
2. `bundle exec rspec spec/services/order/charge_service_spec.rb -e 'multi-item gift cart' -e 'multi-item cart containing a bundle'` — expect 2 examples, 0 failures.
3. `bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb -e 'multi-item presentment charge'` — expect 3 examples, 0 failures.
4. To confirm the refund specs bite, apply the mutation described above and re-run step 3 — expect 2 failures.

## Performance

Not applicable — no query, job, loop, or external call is added or changed. The diff is specs, three VCR cassettes, and comments.

---

## AI disclosure

Written by `claude-opus-5` (the model configured as `model.default`) running as Gumclaw.

Instructions given: work the two remaining items Sahil approved on gumroad-private#1317 — (1) decide and test gift and bundle carts under multi-item buyer-currency charging, and (2) prove refund coverage where a partial refund on one purchase of a multi-item presentment charge clamps to that purchase's remaining presentment amount. The model read the existing charge, allocator, orchestrator, and refund code to determine the current behaviour, wrote the specs, ran them locally, mutation-tested the refund block to confirm it was load-bearing, and baselined the surrounding failures against unmodified main.


gumclaw:

Ran the mandatory local adversarial review against this branch. The reviewer verified every claim in the diff against the production code — `Purchase::PresentmentRefund`, `Charge::PresentmentOrchestrator`, `Purchase::Refundable#refund!`, `Purchase::CreateService#create_giftee_purchase`, `Order::CreateService`, `Purchase::CreateBundleProductPurchaseService`, `Purchase#create_artifacts_and_send_receipt!`, and the `PurchasePresentment` validations.

**Verdict: approve with the P2 fixed. No P1.**

The hunt was aimed squarely at the main risk on a specs-only PR — examples that pass vacuously. Each one was confirmed to bite:

| Check | Result |
|---|---|
| Refund clamp reddens under charge-level mutation | Confirmed — the mock would see `amount_cents: 3750` instead of `12_50` |
| Partial-refund arithmetic (4_00 canonical → 5_00 presentment) | Confirmed exact via `allocate_by_largest_remainder(1250, [400, 600], 1000)` |
| Fail-closed example fails for the *right* reason | Confirmed — the paired positive assertion at exactly 12_50 rules out an unrelated guard |
| Lazy `let` fixtures actually instantiate | Confirmed — `charge_presentment` is forced by the `before` block; hook order correct |
| Gift/bundle specs can't pass on absent fixtures | Confirmed — companion rows asserted present before their presentments are asserted nil |
| Allocation numbers assert allocation, not rounding | Confirmed — `disable_buyer_currency_rounding: true` is set |
| Both app comments factually true | Confirmed against the cited code |
| "No behaviour change" | Confirmed — filtering the app/ diff for non-comment changed lines returns nothing |
| Cassette hygiene | Clean — keys filtered, `livemode: false`, no PII |

### Findings fixed in adfac2698

**P2 — the merchant-id comment was factually wrong.** It claimed FactoryBot sequences "restart per example." They are monotonic within a process; the real collision risk is a fresh run regenerating an id that already exists in seeded data. On a PR whose whole point is accurate comments, that could not ship. Reworded.

**P3 — the refund fixture's scope was oversold.** The describe comment said multi-item carts "charge one PaymentIntent," but the fixture approximates that with per-purchase intents and mocks `ChargeProcessor.refund!`. So it pins amount *clamping*, not refund *routing* to a shared intent. That caveat is now stated in the comment rather than left implicit.

**P3 — quote stub tightened.** `allow(StripeFxQuote).to receive(:create)` → `expect(...).once` on both new examples, matching the single-item siblings, so a duplicate quote mint would now be caught.

**P3 — evidence PNG renamed** to `pr-6385-...` per the qa-media convention.

### Findings deliberately not actioned

`buyer@gumroad.com` in the gift example violates CONTRIBUTING's `@example.com` rule, but it is inherited from the file's existing `common_order_params_without_payment` and the existing single-item gift spec. Changing it in isolation would make the new example inconsistent with every sibling; worth a separate sweep of the file rather than a one-off here.

Both new spec groups re-run green after the fixes (2 examples and 3 examples, 0 failures).

PR #6386, opened 2026-07-27
Scope the forced-currency settlement gate to the account the intent is created on

gumclaw:
## What

Scopes the forced-currency settlement gate to the Stripe account the PaymentIntent is actually created on, rather than to whichever merchant account is attached to the seller.

Two gates move, both only on the forced-currency lane (UPI / iDEAL / Bancontact):

- `Checkout::PaymentMethodResolver#forced_currency_settlement_supported?` and `Checkout::BuyerCurrencyEligibility#method_forced_decision` now ask "does the account the intent is created on hold USD?" via a new `usd_holding_settlement_account?`. For a Stripe Connect (direct-charge) seller that is still their own account. For everyone else the intent is created on the Gumroad platform account, so that is the account consulted.
- The lane's charge-model gate widens by exactly one case: a seller with a Gumroad-managed Stripe account. That produces the same destination-charge intent shape as a seller with no Stripe account at all, which the lane already supported.

The card lane's gates are deliberately unchanged. It converts through an FX quote minted with the seller's connected account as `Stripe-Account`, so the seller's own settlement configuration really does decide that outcome. `usd_settling_merchant_account?` keeps reading the seller's account and now says why in a comment.

## Why

Sellers charged with a **destination charge** — the majority of Gumroad sellers — never have the intent created on their own account. `StripeChargeProcessor#create_payment_intent_or_charge!` creates it on the Gumroad platform account (which holds USD) and passes the seller's account as `transfer_data[destination]`, where it receives a transfer after the charge settles.

A destination account's balance currency does not constrain the currency the intent may be created in. Verified directly against the Stripe API rather than inferred: with a GBP-settling destination account, an INR/UPI intent and an EUR/iDEAL intent both **create**, and the INR/UPI intent **confirms** to `requires_action` with `upi_handle_redirect_or_display_qr_code`. There is no Stripe-side constraint behind the old behaviour.

So the gate was reading an account that has no say in the matter and using it to withhold payment methods from checkouts that would have completed. The reported seller does everything the UPI lane documents — buyer-currency flags on, `checkout_local_method_upi` at 100%, single INR-priced item, Indian buyer — and got no UPI tab and no explanation.

**Blast radius, re-measured.** The issue estimated 237 of 1,752 sellers (13.5%) with INR-priced products. That figure counted every alive Stripe merchant-account row; the code path actually reads `User#merchant_account`, which filters on `can_accept_charges?`. India is not a Stripe-charges country, so the 222 INR-settling rows are already unusable and those sellers were falling back to the platform account for a different reason. Re-running the same 3,000-product sample through the real resolution path: 1,744 sellers ride the platform account, 3 hold USD, **6 are genuinely blocked by this gate** (eur ×2, aud, gbp, cad, brl), 1 is a Connect seller correctly gated. Small, but it includes the reporting seller (GBP), and the gate was wrong on the merits regardless of how many it caught.

Two consequences worth naming:

- The 07-24 UPI ramp evidence on the internal tracker was read against a denominator that is *not* meaningfully wrong — the thin attempt volume is not explained by this gate.
- The same gate governs iDEAL and Bancontact, so a EUR-priced product from a non-USD-holding managed seller was dropped identically. That is fixed here too.

Fixes the internal tracking issue for this gate (link on the private tracker).

## Reproduction (production, read replica)

```
Checkout::PaymentMethodResolver.new(
  sellers: [seller], recurring: false,
  buyer_country: "IN", cart_product_currency: "inr", cart_total_usd_cents: 85000
).resolve

enabled          = ["card", "link"]
launch_gated_out = ["klarna", ..., "upi", ...]

seller_enabled                     = true
Feature.active?(:checkout_local_method_upi) = true
usd_holding_merchant_account?(seller_ma)    = false   # gbp — the failing gate
usd_holding_merchant_account?(platform)     = true    # the account the intent is created on
has_stripe_account_connected?               = false   # destination charge
```

## Stripe verification

Run against the Stripe test API with a GBP-settling Custom account (`default_currency: "gbp"`, transfers + card_payments active) as `transfer_data[destination]`:

| Intent | Result |
|---|---|
| `currency=inr`, `payment_method_types[]=upi`, GBP destination | created — `payment_method_types: ["upi"]`, transfer_data intact |
| same intent, confirmed with a UPI payment method | `status: requires_action`, `next_action: upi_handle_redirect_or_display_qr_code` |
| `currency=eur`, `payment_method_types[]=ideal`, GBP destination | created |

The destination account's settlement currency does not enter into it.

## QA steps

No preview app is needed to see the behaviour change — this is a server-side eligibility decision with no rendering of its own — but for a reviewer who wants to watch the tab appear:

1. Add the `preview` label and wait for the branch app.
2. Pick a seller with `buyer_currency_charging` + `buyer_local_currency` on, `checkout_local_method_upi` on, and a Gumroad-managed Stripe account whose `currency` is not `usd`.
3. Visit an INR-priced single-item product from an Indian IP.
4. Before this change the Payment Element mounts in USD with card/Link only. After, it mounts in INR and the UPI tab renders.

## Before / after

Not a rendering change of its own, so there is no visual diff to show for the shared surface. The observable difference is the resolver output for the affected seller shape:

| | Resolved `payment_method_types` |
|---|---|
| **Before** | `["card", "link"]` |
| **After** | `["card", "link", "upi"]` |

The adjacent USD checkout surface is unchanged by construction: every gate outside the forced-currency lane is untouched, and the 92-example `Checkout::StripePaymentPresenter` suite (which covers the element-mount decision for USD and non-USD carts alike) passes unmodified.

## Test Results

Five suites — the two changed services plus every downstream consumer of the moved predicates:

spec results

```
338 examples, 0 failures
```

New coverage:

- resolver: a launched forced-currency method IS offered to a destination-charge seller whose own account settles non-USD
- resolver: it is still WITHHELD from a direct-charge Connect seller whose own account settles non-USD
- eligibility: destination-charge seller with a non-USD account gets the method, with `direct_listed_amount?` true
- eligibility: the method is withheld when the *platform* account holds a non-USD balance
- eligibility: the seller-managed destination-charge model is now allowed (this replaces the assertion that it was withheld — the old expectation encoded the bug)

Both existing regression tests for the 2026-07-23 iDEAL dark-ramp still pass unmodified; the learned per-currency mismatch marker is untouched by this change.

Note for anyone running these locally: the suites need the platform merchant account row, so `RAILS_ENV=test bundle exec rails db:seed` first. Without it several pre-existing examples fail on unmodified main too.

---

## AI disclosure

Written by `claude-opus-5` (per `model.default` in the agent config), driven from the GitHub webhook for the "Possible to support?" comment on the internal tracking issue. The model read the resolver and eligibility services, traced the intent-creation path through `StripeChargeProcessor`, tested the Stripe constraint directly against the test API to check whether the gate was protecting something real, reproduced the failure on the production read replica, re-measured the issue's blast-radius claim through the real code path (and corrected it), implemented the change, and wrote and ran the specs.


gumclaw:

**The one judgement call worth a reviewer's attention.** I widened two gates, not one. Fixing only the settlement gate would have been the smaller diff and would have matched the issue exactly — but it would have shipped a bug: the reporting seller also fails `supported_charge_model?`, because their Gumroad-managed Stripe account is neither `is_managed_by_gumroad?` nor `is_a_stripe_connect_account?`. UPI would have stayed dark for them and I'd have closed the issue on a fix that didn't fix it. I found this by running the predicates against the live seller rather than trusting the issue's single-gate diagnosis, and both gates move together for that reason.

**What I deliberately did NOT change.** `usd_settling_merchant_account?` still reads the seller's own account, and the card lane's `supported_charge_model?` is untouched. Those guard FX-quote paths, and `StripeFxQuote.create` passes the seller's connected account as `Stripe-Account` — so there the seller's configuration genuinely decides whether Stripe accepts the quote. Widening those too would have been the symmetric-looking change and the wrong one. I added a comment at `usd_settling_merchant_account?` saying so, since the two predicates now differ in a way that looks arbitrary without it.

**I checked whether the gate was protecting a real Stripe constraint before removing it.** The issue raised this as an open question and it's the one thing that could make this whole PR wrong. Rather than reason about it, I ran it: a GBP-settling Custom account as `transfer_data[destination]` accepts both an INR/UPI intent and an EUR/iDEAL intent, and the INR one confirms through to `requires_action`. Table's in the body. If a reviewer thinks the test-mode result doesn't generalise to live, that's the finding to push back on.

**I corrected the issue's own numbers rather than repeating them.** The 13.5% / 237-seller figure counts merchant-account rows the code never reads — `User#merchant_account` filters on `can_accept_charges?`, and India isn't a Stripe-charges country, so the 222 INR rows were already inert. Real number through the actual resolution path is 6 sellers in that sample. I'd rather ship a smaller honest number than inherit a flattering wrong one, and it means the gumroad-private#1334 thin-attempt-volume question is *not* explained by this bug — that's still open.

**Test expectation I inverted, flagged explicitly.** `"withholds the method for seller-managed destination-charge models"` became `"allows the method for a seller-managed destination-charge model"`. An inverted assertion is the kind of thing that should never pass silently, so: the old expectation encoded the bug, and the comment above the new test says why. The two gumroad-private#933 iDEAL dark-ramp regression tests are untouched and still pass — the learned mismatch marker is not involved here.

**Evidence honesty.** There's no before/after screenshot pair because this change renders nothing of its own; the observable delta is the resolver's output array, which is in the body as a table. The adjacent-surface "no regression" claim rests on the 92-example presenter suite passing unmodified, not on a screenshot I'd have had to stage. Spec run is local (338 examples, 5 suites, 0 failures) — CI will confirm independently. One environment note that cost me a cycle and will cost the next person one: these suites need `rails db:seed` in the test DB for the platform merchant-account row, and without it examples fail on unmodified main too.

Not merging this myself — it's a payments-surface change.

gumclaw:
@rmarescu routing this to you — it is squarely your surface (PaymentIntent creation and which Stripe account an intent lands on), and it is the direct follow-on to the forced-currency work you have been reviewing on #6285/#6346.

Housekeeping done this sweep: the QA-media file was named `pr-1409-spec-results.png` after the tracking issue rather than the PR, which is the exact CONTRIBUTING naming convention you called out on #6288. Renamed to `pr-6386-spec-results.png` in `f09989a1f` and repointed the body URL (verified HTTP 200). No code changed. Label swapped `working` → `awaiting-human`, assignee moved off me and onto you.

**The one judgement call I want a human on.** I widened *two* gates, not one:

1. The settlement gate itself — `usd_holding_settlement_account?` now asks "does the account the intent is actually created on hold USD?" instead of "does the seller's attached merchant account hold USD?". For a Connect direct-charge seller that is still their own account; for everyone else the intent is created on the Gumroad platform account, so that is what gets consulted. This part is the bug fix and I think it is uncontroversial.
2. The lane's charge-model gate widens by exactly one case: a seller with a Gumroad-managed Stripe account. That produces the same destination-charge intent shape as a seller with no Stripe account at all, which the lane already supports — so I believe it is safe. But it is a *scope* widening, not a bug fix, and it means UPI/iDEAL/Bancontact become reachable for a cohort they were not reachable for before. Fixing only (1) would have been the smaller diff and matched the issue exactly. If you want that narrower version instead, say so and I will cut (2) back out.

The card lane's gates are deliberately untouched — it converts through an FX quote minted with the seller's connected account as `Stripe-Account`, so there the seller's own settlement configuration really does decide the outcome.

CI green at head, MERGEABLE, Greptile clean. Never auto-merging — the merge press is yours.

gumclaw:
Body-only edit just now — no commits, head is unchanged, review state untouched. Flagging it because one of the two things I changed had real consequences beyond tidiness.

**The description carried a literal `Fixes antiwork/gumroad-private#1409` line.** GitHub honors closing keywords across repositories, so the moment this merged it would have force-closed that private issue — `stateReason: COMPLETED`, no comment, the assignee never notified — while the fix was still many commits deep in the undeployed release queue. The issue would have read as shipped days before it was live. Removed; verified `closingIssuesReferences` is now empty and gp#1409 is still `OPEN`.

**It also published private issue numbers in a public repo** — that line plus three informational references (one in the AI-disclosure paragraph, which is the easy one to miss because it isn't prose anyone re-reads). All rephrased without identifiers; the substance is unchanged.

The private-issue↔PR edge still exists, as a comment on the private issue, which is the direction that doesn't leak. Nothing needed from you — review this on the code.

gumclaw:
@rmarescu — handing the build on this over to you, per Sahil's direction today: **humans build everything multi-currency and I review.** The reasoning he gave is that it's the fastest path, and where a human is in the loop on complex work, the human commands the start and the final merge.

That inverts what I was doing here. I authored this and kept pushing to it while you reviewed; from here it's yours to carry and yours to merge. You've already pushed to this branch, so this is mostly making the ownership explicit rather than a change of direction.

Concretely:

- **Take it over directly** — push to the branch, or close this and re-cut it however you'd rather structure it. No attachment to my shape of the change.
- I will **not** push further commits unless you ask.
- I'll **review adversarially** when you want a pass, and keep supplying live production data behind the change.
- Auto-merge stays **off**. The merge click is yours.

Standing offer: I'm faster at some of the grunt work — QA media, hosted-preview evidence, prod-console reads, regression specs that pin a behaviour. Ask and it's done, still under your merge.

The unresolved review threads and any findings of mine stay on the PR so nothing's lost in the handover. If you'd rather I write up the current state as a plain description of what's done versus outstanding, say so and I'll add it.

PR #6388, opened 2026-07-27
Say plainly that checkout methods and payouts are separate

gumclaw:
A creator replying to our UPI launch tweet asked:

> "Will it be included in payout too or will the payout be limited to PayPal"

Both halves rest on the same misreading — that adding a checkout payment method changes how a seller gets paid, and that Indian sellers are on PayPal. Neither is true, and the question is a reasonable thing to conclude from what we've published.

## The facts, checked against production

Indian sellers are paid by **bank transfer in rupees** against an IFSC code (`IndianBankAccount`, `Currency::INR`). Not PayPal.

| | |
|---|---|
| Accounts with an Indian bank account on file | **99,588** |
| INR payouts settled in the last 90 days | **1,675** |
| Value | **₹57.96 crore** (5,796,344,644 paise) |

## Where the doc gap actually is

`_13-getting-paid` **already lists India correctly** in the bank-payout table. The table isn't wrong — but nobody reads a 40-row table to answer a question they didn't know they had.

The real gap is that the **Payout modes** section opens straight into the bank/PayPal split without ever saying that buyer-side payment methods are a *different system*. A seller who just read that UPI is live has nowhere to place it, so they guess — and the natural guess is "new payment method, maybe my payouts change too".

Separately, `git grep -ilE '\bUPI\b'` across the whole help center returns **zero hits**. UPI has no documentation at all.

## The change

One note at the top of Payout modes: what buyers pay with depends on where the **buyer** is, it doesn't touch payouts, and everything settles into one balance paid out the single way the seller configured.

Placed **above** the bank/PayPal split deliberately, so it frames the whole section rather than answering one country's question. The same confusion arrives with the next local method we ship — Bancontact, Pix, iDEAL — and this wording already covers them.

Body-only, +3 lines. `articles.yml` untouched (no title/description change).

## Verification

- ERB syntax OK (`ERB.new(...).src`)
- Tag balance verified across all `div`/`p`/`ul`/`li`/`table`/`tbody`/`h3` pairs — every one balanced
- India's row in the payout table confirmed intact
- `callout-yellow` matches the existing callout convention in this article

Not run: the full help-center specs. This is a body-only prose change with no `articles.yml` edit, so slug↔partial integrity is unchanged; CI covers it.

### DIFF
--- a/app/views/help_center/articles/contents/_13-getting-paid.html.erb
+++ b/app/views/help_center/articles/contents/_13-getting-paid.html.erb
@@ -18,7 +18,7 @@
   </ul>
   <h3 id="payout-modes">Payout modes</h3>
   <div class="callout-yellow">
-    <p><b>Note: </b>How your buyers pay you and how you get paid out are two separate things. Gumroad accepts many payment methods at checkout, including cards, PayPal, and local methods such as UPI in India, and which ones a buyer sees depends on where the <i>buyer</i> is. Your payout is not affected by any of that. Whatever mix of methods your buyers use, the money lands in your Gumroad balance and is then paid out to you the one way you have set up below, in your own payout currency.</p>
+    <p><b>Note: </b>How your buyers pay you and how you get paid out are two separate things. Gumroad accepts many payment methods at checkout, including cards, PayPal, and local methods such as UPI in India, and which ones a buyer sees depends on where the <i>buyer</i> is. Your payout is not affected by any of that. In general, sales land in your Gumroad balance and are then paid out to you the one way you have set up below, in your own payout currency. The exception is <a href="#paypal-connect">PayPal sales when you have connected your own PayPal account</a>, which are credited to you at the time of sale instead.</p>
   </div>
   <p>We support direct bank deposits in the local currency for most countries. For countries where bank deposits are not available, we offer PayPal transfers as the only payment option.</p>
   <p>We do <i>not</i> support alternative payout modes like Payoneer, Wise, check, money order, wire transfer, etc. If your country is not supported by direct bank deposits or PayPal, then unfortunately, we have no way to pay you out for now.</p>
--- a/app/views/help_center/articles/contents/_13-getting-paid.html.erb
+++ b/app/views/help_center/articles/contents/_13-getting-paid.html.erb
@@ -184,213 +184,216 @@
           <td>EUR</td>
         </tr>
         <tr>
+          <td>India</td>
+          <td>INR</td>
           <td>Indonesia</td>
           <td>IDR</td>
+        </tr>
+        <tr>
           <td>Ireland</td>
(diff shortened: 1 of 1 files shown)


gumclaw:
## Adversarial pre-merge review (fable-5, local) — REQUEST CHANGES → resolved

Ran the mandatory local adversarial review. It found **two real problems**, both now fixed. Recording them here because both were substantive, not stylistic.

### P1 — the note contradicted this article 450 lines below it (fixed in `699793356`)

Original copy claimed:

> "Whatever mix of methods your buyers use, the money lands in your Gumroad balance and is then paid out to you the one way you have set up below"

That is false for **PayPal Connect** sellers. Line 477 of this same article:

> "proceeds from all sales via PayPal will be credited to your connected account **at the time of sale**, in the default currency of your PayPal account"

Those sales never touch the balance and never use the configured payout method. Scoped the claim with "In general" and named the exception with an in-page link to `#paypal-connect`, so the two statements agree rather than fight.

### Also caught: India was missing from the payout country table (fixed in `407bf98dc`)

Verifying the review's claims turned up something worse than the original bug. `grep '>India<'` on the bank-payout table returns **nothing** — India has never been listed, despite:

| | |
|---|---|
| Accounts with an Indian bank account | **99,588** |
| INR payouts, last 90 days | **1,675** |
| Value | **₹57.96 crore** |

India appeared only in the payout-*day* table further down. So a seller checking "can I even get paid to a bank?" found nothing and reasonably concluded PayPal was the only option — **exactly the misreading this PR exists to fix.** My original PR description claimed the table "already lists India correctly"; that was wrong and I'd have shipped a fix that left the root cause in place.

Added as `INR` with no minimum, matching `Country#min_cross_border_payout_amount_local_cents` for `IND` (`1_00` — the same floor as Iceland and Indonesia, neither of which carry a "(min X)" annotation).

The table renders two countries per row, so inserting one entry reflows every subsequent row. First attempt left Italy stranded in a 2-cell row; rebuilt the `<tbody>` programmatically instead. Verified:

- **117 → 118 countries**, only `India` added, **zero lost**, zero currency values changed
- **No row with an odd cell count** (this also cleaned up a pre-existing short Vietnam row)
- India sits between Iceland and Indonesia, matching the table's ordering

### Trim the slop

Reviewer's suggested tightening was reasonable but I kept the longer form: the redundancy ("Your payout is not affected by any of that") is the sentence that directly answers the question that was asked, and this is a callout a confused seller reads once. Everything else is load-bearing.

### Verification after both fixes

- ERB compiles (`ERB.new(...).src`)
- Tag balance verified across `div`/`p`/`ul`/`li`/`table`/`tbody`/`tr`/`td`/`h3`/`a` — all balanced
- `#paypal-connect` anchor target confirmed to exist
- No CI failures at any point

Arming auto-merge.

PR #6391, opened 2026-07-27
Fix the remaining mobile library search timeouts: seller filter join order

gumclaw:
## What

Resolves the seller filter on `Api::Mobile::PurchasesController#search` to concrete user ids with a separate query instead of nesting `User.where(external_id: ...).select(:id)` as a subquery.

```ruby
- purchases.where(seller_id: User.where(external_id: Array.wrap(params[:seller])).select(:id))
+ seller_ids = User.where(external_id: Array.wrap(params[:seller])).pluck(:id)
+ purchases.where(seller_id: seller_ids)
```

No behavior change. One extra tiny query on the users table (index lookup on `index_users_on_external_id`, 3 rows) in exchange for a different join order on the purchases query.

## Why

The mobile library search endpoint is still timing out (Sentry GUMROAD-8T, 279 users, last seen 2026-07-27 03:33Z) after #6206 removed the per-seller probe loop. The remaining cause is a different query shape entirely, and it is not a missing index.

With the subquery form, MySQL plans the join **seller-first**: it iterates the matched sellers and walks each one's sales through `index_purchases_on_seller_id_and_chargeback_date`, filtering down to this buyer afterwards. So a buyer who filters their library by a large seller pays for a scan proportional to **that seller's entire sales history**, not to their own purchases. Passing concrete ids removes the join, so the optimizer starts from `index_purchases_on_purchaser_id` and reads only the buyer's own rows.

The three timed-out Sentry events sampled after the #6206 deploy all carry `seller[]` params and all die in the same place — the pagination `COUNT`:

```
GET /mobile/purchases/search?items=24&page=1&seller[]=…&seller[]=…&seller[]=…
last query before SIGTERM: "Purchase Count"   db_runtime: 119,995 ms
```

Fixes #6185

## Performance

Measured on the production read replica, same buyer (673 lifetime purchases), filtering by three sellers with 311K / 66K / 145 sales. `EXPLAIN ANALYZE` for both forms plus wall-clock timings:

explain-before-after

| | subquery (today) | concrete ids (this PR) |
|---|---|---|
| pagination `COUNT`, cold | 1,616 ms | 11 ms |
| pagination `COUNT`, worst observed | 78,250 ms | 4 ms |
| whole endpoint path, cold | 6,646 ms | 50 ms |
| whole endpoint path, warm | 21 ms | 22–30 ms |

The warm row is the honest caveat: once the pages are in the buffer pool both forms are fine. The subquery plan's cost scales with the filtered seller's row count, so it is the cold/large-seller case that blows the 120s request budget, and that is exactly the shape in the timeouts.

Endpoint-wide numbers over the same window, for context on what is left: p50 125 ms, p95 1,227 ms, p99 2,525 ms, max 13,306 ms over 6,190 requests since the #6206 deploy (was max 30,714 ms / p99 3,079 ms in the five days before it). The tail this PR targets sits above p99.

## Before/After

Not a UI change — no rendered surface is touched. The endpoint returns byte-identical JSON; the existing seller-filter examples assert the response contents and are unmodified.

## QA steps

1. Open the mobile app library and filter by one seller, then by several sellers. Results and the `sellers` aggregation should be unchanged.
2. Filter by a seller external id that does not exist — empty result, no error (covered by a new spec).
3. After deploy, watch `Api::Mobile::PurchasesController#search` p99/max in Sentry and confirm GUMROAD-8T decays.

## Test Results

spec-results

Two new examples in `spec/controllers/api/mobile/purchases_controller_spec.rb`, both green locally:

- the unmatched-external-id case returns an empty result rather than erroring. This one is an equivalence guard for the newly-reachable `where(seller_id: [])` path, not a revert detector — it passes either way, deliberately.
- the seller-scoped purchases query carries no nested users subquery, and the result still excludes a second seller's purchase. This is the revert detector: restored to `origin/main` the controller fails it on exactly the subquery assertion.

`rubocop` on both changed files: 2 files inspected, no offenses.

The rest of the `GET search` block cannot run in this local environment — 21 of 21 examples fail identically on unmodified `main` in the same worktree (`Validation failed: We couldn't charge your card`, the local Stripe stub). My two new examples use `:free_purchase` and are the only ones that pass. CI arbitrates the rest.

## Note on the original approach

This branch previously proposed a composite `(email, seller_id)` index on `purchases`. @ershad's objection to that was correct — a ~15GB index and an hour-long deploy-pipeline stall, as a last resort only. Profiling the endpoint against production found the real remaining cost is a join-order problem in the application query, so no DDL is needed.

---

**AI disclosure:** Authored by claude-opus-5 running as the Gumclaw agent. The agent profiled the endpoint query-by-query against the production read replica (audited read-only console), identified the subquery join-order as the cause of the residual timeouts, wrote the controller change and the two specs, verified them locally including a red-check against unmodified main, and produced the evidence images.


gumclaw:

Ran the mandatory local adversarial review against this branch. Verdict: **SHIP**, three P3s, none merge-blocking. Two are addressed, one is deliberately left.

The reviewer attacked equivalence hardest, since that is the whole risk of this change — it went after empty result sets, deleted sellers, default scopes, STI, duplicate and nil ids, the `params[:q]` join interaction, and all three ways the relation is re-executed. Findings worth recording:

- `User` has no `default_scope` — `Deletable` only adds opt-in `alive`/`deleted` scopes, and neither form uses them, so deleted sellers matched before and still match. No STI. `external_id` is a plain stored column, not an `ObfuscateIds` derivation, so there is no encrypt/decrypt asymmetry between `.select(:id)` and `.pluck(:id)`.
- Empty set: Rails renders `where(seller_id: [])` as `WHERE 1=0`, which is semantically identical to an `IN (SELECT …)` returning zero rows. `pluck` on a primary key cannot produce nils, and duplicates in an `IN` list are harmless.
- The `params[:q]` branch's `left_joins(:link, :seller)` cannot reintroduce the bad plan: hash conditions are table-qualified, so the filter renders as `` `purchases`.`seller_id` IN (…) ``, and the join is a primary-key equality driven from the purchases side. `group(:seller_id)` stays unambiguous because neither `users` nor `links` has that column.
- A point I had not made in the description: under the old form the subquery re-executed inside **each** of the three outer queries (count, page, seller aggregation), so they could in principle disagree mid-request. Now the users lookup runs exactly once and all three share one frozen id list.

### Findings and disposition

| | Finding | Disposition |
|---|---|---|
| P3 | The SQL-inspection example had one purchase in the fixture, so its size assertion could not tell a filtered result from an unfiltered one | **Fixed** in c9f7127 — a second seller's purchase is now present and must be excluded |
| P3 | The unmatched-external-id example passes with or without the fix | **Left, deliberately** — it guards the newly-reachable `where(seller_id: [])` path rather than the perf change. The SQL-inspection example is the revert detector, and it does go red: reverted to `origin/main` it fails on exactly the subquery assertion |
| P3 | `params[:seller]` is uncapped, so a client can send thousands of `seller[]` values | **Not in this PR.** Pre-existing and unchanged — the old subquery embedded the same unbounded caller-controlled `external_id IN (…)` list. Capping it is real hardening but a separate concern from the join order, and folding it in would put an untested behavior change in a fix I want reviewed on its equivalence alone |

The reviewer also checked for a repo convention on batching id lists and confirmed none exists — the only `BATCH_SIZE` constants are per-worker Sidekiq batching, not a controller `IN`-list rule.

PR #6350, opened 2026-07-26
Quote every product in the buyer's currency, not just USD-priced ones

gumclaw:
## What

Quotes every product in the buyer's currency instead of only USD-priced ones.

`Checkout::BuyerCurrencyQuote#quotable_product?` required `price_currency_type == "usd"`, so any product a seller priced in euros, reais, rupees or anything else got no FX quote at all — and the checkout fell back to rendering the cart in US dollars for every buyer, in every country. The gate now excludes one product currency instead of all-but-one: the buyer's own.

Three places move together:

- `Checkout::BuyerCurrencyQuote#quotable_product?` — takes the buyer's currency and rejects only a listing already in it. The currency check moves below the GeoIP lookup, since it now depends on it.
- `Checkout::BuyerCurrencyEligibility#decision` — the charge-time mirror of that gate, same change. The old `:unsupported_product_currency` fallback becomes `:listed_currency_is_buyer_currency`, which is what it now means.
- `Checkout::StripePaymentPresenter#buyer_currency_presentment_element_shape?` — drops its own `product_currency == "usd"` condition. The buyer-local display it already requires is off whenever the buyer's currency equals the product's (`buyer_currency_display_props` returns `display_mode: "default"` in that case), so the excluded case stays excluded without a second check.

`Checkout::StripePaymentPresenter#props` also now sends a cart that qualifies for both the buyer-currency element and the method-forced client-confirm lane to the buyer-currency element — see Why.

## Why

Follow-up to #6346 and the internal tracking issue. #6346 fixed the display for the narrow case the report described (a BRL product paid with Pix by a Brazilian buyer) by rendering the listed currency on the method-forced lane. That left the general case: a Brazilian product bought from Canada, a euro product bought from Japan, or any non-USD listing on an ordinary card checkout, all still showed US dollars.

Being in USD means nothing to those buyers. The seller's chosen pricing currency is an internal detail of how the price was entered; the total that reaches the quote service is canonical USD regardless, and converting it into the buyer's currency works identically whether the seller typed dollars or euros. The old gate was conflating "our money is normalized in USD" with "the buyer should see USD".

[...]

PR #6397, opened 2026-07-27
Let the store agent author a product's landing page, not just the profile

gumclaw:
## What

Gives the store agent product-scoped custom-page endpoints, so it can author a landing page for **one product** instead of only the creator's profile storefront.

Three new catalog entries backed by two new API actions:

| Catalog id | Route | What it does |
|---|---|---|
| `get_product_custom_html` | `GET /v2/products/:id/custom_html` | Read the product's current page; `has_landing_page` says whether one is published |
| `update_product_custom_html` | `PUT /v2/products/:id` | Replace the whole page (blank clears it, restoring the native product page) |
| `edit_product_custom_html` | `POST /v2/products/:id/custom_html/edit` | Replace one exact snippet, leave the rest of the page alone |

The write logic that both surfaces need — a row lock so concurrent writers serialize, a sanitize of the **whole spliced document** rather than the inserted snippet, and the blank-page-unpublishes rule — moved into a shared `Pages::CustomHtmlWriter`. The profile controller was refactored onto it in the same change, which is where the 77 deleted lines in `users_controller.rb` come from.

## Why

Tracking issue: antiwork/gumroad-private#1402

The agent's HTML-authoring endpoints covered only the creator's profile storefront. So when a creator asked it to build a landing page for a specific product, the best it could do was stage a change against their entire profile — a different surface, and a destructive one.

That is not hypothetical. The product editor's Share tab generates a landing-page prompt and gives creators a copy button for it. The prompt is written for an external agent driving the `gumroad` CLI, but creators reasonably paste it into Gumroad's own Agent tab instead — which is exactly how this arrived. Before this change they got either a truncated turn or, at best, a proposal against the wrong page.

**Why the shared writer rather than a second copy.** Duplicating the write path would have made a drift between the two surfaces a *correctness or security* bug rather than a cosmetic one — one endpoint forgetting the lock, or sanitizing only the snippet instead of the spliced result. One caller deliberately stays inline: `update` accepts `custom_html` alongside every other product field and has to apply them all in one transaction under one lock, so its full-replace logic can't be pulled out. That is called out in a comment on the service.

**Two decisions made here rather than parked.** They are *how* questions, so per the standing rule they were decided, and both are cheap for a reviewer to overrule:

1. **Publishing is allowed, gated on the existing staged-proposal + confirm flow.** A preview-only endpoint was rejected because it cannot do the thing creators are actually asking for. The safety comes from the confirm step plus explicit warning copy, not from withholding the capability.
2. **The Share-tab prompt copy is explicitly out of scope.** It still points at the CLI. Changing seller-facing copy about which agent to use is a product call, not part of this fix.

**The product-specific risk that shaped the guardrails.** A published product page *replaces* the native product page, price and buy button included — so a page without a buy element makes the product unpurchasable. That is a sharper failure than a bad profile page, so:

- the catalog summary and the system-prompt bullet both state it,
- `update_product_custom_html` and `edit_custom_html` return a `warning` when the resulting page has no buy affordance,
- and `update_product` deliberately does **not** accept `custom_html`, so every full-page write flows through the id whose description carries that warning.

Both new endpoints are gated behind the existing `custom_html_pages` feature flag via a `before_action`, so they cannot be reached at all while it is off.

## Before / After

No rendered surface changes. This adds API endpoints and agent-catalog entries; the flag is off by default, so with the flag off the behaviour of every existing surface is byte-identical.

| | Before | After |
|---|---|---|
| Creator asks the agent for a landing page for one product | Agent's only page endpoints target the profile — proposal overwrites the whole storefront | Agent uses the product endpoints; proposal targets that product's `/l/` page |
| Agent makes a small tweak to an existing product page | No product page surface at all | `edit_product_custom_html` replaces one snippet, whole-document re-sanitized |
| Page written without a buy element | n/a | Response carries a `warning` naming the problem |
| `custom_html_pages` flag off | n/a | Both endpoints reject with "You do not have access to custom HTML pages." |
| Profile page write | Own inline lock/sanitize/unpublish copy | Same behaviour, now via `Pages::CustomHtmlWriter` |

## QA steps

No preview click-path — this is an API + agent-catalog change with the flag off by default. Console-based verification on the hosted preview (`x-revision` should be this PR's head):

1. `Feature.activate_user(:custom_html_pages, seller)` for a test seller with one product.
2. `GET /v2/products/:id/custom_html` with an `edit_products`-scoped token → expect `has_landing_page: false` and the product's `landing_url`.
3. `PUT /v2/products/:id` with `custom_html` containing `<a data-gumroad-action="buy">Buy now</a>` → expect success, no `warning`.
4. Repeat step 3 with the buy element removed → expect success **plus** a `warning` about the product being unpurchasable.
5. `POST /v2/products/:id/custom_html/edit` with a `find` that matches twice → expect the "matches N places" error and **no write**.
6. `PUT /v2/products/:id` with blank `custom_html` → page unpublishes, `/l/` serves the native React product page again.
7. `Feature.deactivate_user(:custom_html_pages, seller)` → both new endpoints reject.

## Test Results

129 examples, 0 failures across every affected spec file:

- `spec/controllers/api/v2/links_controller_edit_custom_html_spec.rb` (new, 23 examples)
- `spec/controllers/api/v2/users_controller_custom_html_spec.rb` (profile refactor regression)
- `spec/controllers/api/v2/users_controller_edit_custom_html_spec.rb` (profile refactor regression)
- `spec/services/ai/store_agent_api_catalog_spec.rb` (+65 lines of catalog assertions)
- `spec/controllers/api/internal/agent_custom_html_previews_controller_spec.rb`

The screenshot below carries both the green run **and** a mutation proof, so the new specs are demonstrably load-bearing rather than passing against anything: dropping the flag-gate `before_action` and neutering the buy-element warning produces 3 targeted failures in exactly the cases this PR exists to cover. The working tree was restored and verified clean before committing.

Also run: `rubocop` clean on all five touched Ruby files.

One note on baselining, so a reviewer doesn't chase it: `spec/controllers/api/v2/users_controller_spec.rb` fails 19 examples locally on `create(:purchase)` ("We couldn't charge your card"). A detached worktree at `origin/main` fails the identical 19 on the same machine — it is a local Stripe-fixture limitation, not this change, and none of those examples touch `custom_html`.

Spec run: 129 examples 0 failures, plus the mutation proof

Permanent copy: `qa-media/pr-6397-test-results.png` at `378aa4907a5e13648b531eee9ea30f8e1e2427f4`.

---

*AI assistance disclosure:* authored with Hermes Agent running `claude-opus-5`. Instructions given: the autonomous dispatcher standing order (sweep the tracked repositories and build qualifying issues end-to-end without a human in the loop), and a maintainer comment on the internal tracking issue: "Fix." A prior session left this build code-complete but uncommitted in its worktree; this session verified the inherited diff (specs re-run, lint re-run, guards proven load-bearing by mutation), then committed, pushed, and opened this PR.


gumclaw:
@gianfrancopiana routing this to you for review — you own the store-agent surface (gumroad#6363, #6369 and #6350's agent-side work all came to you), and this is the same lane.

**Deliberately not auto-merged, and this is the reason.** Everything about this diff is inside my auto-merge gates on paper — no migration, no money path, no legal page, no flag ramp, specs green, lint clean, guards proven load-bearing. What holds it is the *class* of change: it widens what an LLM agent is permitted to write to a live, buyer-facing product page. That is an agent-authority call, not a correctness one, and it should have a human's name on it.

**The judgement call, stated plainly so you can overrule it cheaply.** I allowed publishing rather than shipping a preview-only endpoint. A preview-only version cannot do the thing creators are actually asking for, so it would have been a safer-looking change that solved nothing. The safety instead comes from three places, and it's worth checking you agree with each:

1. Publishing still flows through the existing staged-proposal + confirm step — the agent cannot write a page without the seller pressing confirm.
2. A published product page **replaces** the native page, price and buy button included, so a page with no buy element makes the product unpurchasable. Both write endpoints return a `warning` when that happens, and the catalog summary plus the system-prompt bullet both say it in the agent's own context.
3. `update_product` deliberately does **not** accept `custom_html`. Every full-page write has to go through `update_product_custom_html`, whose description carries the buy-element warning — so the agent cannot reach the destructive path via a summary that never mentioned the risk.

Both endpoints are behind the existing `custom_html_pages` flag via a `before_action`, so with the flag off they are unreachable, not merely unused.

**The refactor is the part worth your eyes on second.** `users_controller.rb` loses 77 lines because the profile write path moved onto a shared `Pages::CustomHtmlWriter` rather than the product endpoints growing a second copy of it. I did that because a drift between the two — one of them forgetting the row lock, or sanitizing only the inserted snippet instead of the spliced document — is a correctness or security bug, not a cosmetic one. One caller intentionally stays inline and there's a comment saying why: `update` applies `custom_html` alongside every other product field in one transaction under one lock, so its logic can't be lifted out. The profile regression specs cover that refactor and are in the 129.

**On the evidence.** The screenshot in the body carries the green run *and* a mutation proof, because "129 examples, 0 failures" on its own doesn't tell you the new specs would catch anything. Dropping the flag gate and neutering the buy-element warning produces 3 targeted failures in exactly the cases this exists to cover.

**One thing I did not do.** The product editor's Share tab still generates a landing-page prompt that points creators at an external agent driving the CLI. That copy is why creators paste the prompt into Gumroad's own Agent tab in the first place, which is how this bug arrived — but changing seller-facing copy about which agent to use is a product call, so I left it out of scope rather than deciding it. Worth a separate look.

CI is running on head `378aa490`; I'll confirm the verdict here rather than claiming it early.

PR #6398, opened 2026-07-27
Honor the Gumroad Affiliate Program opt-out on membership renewals

gumclaw:
## What

Renewals of a membership referred by a Gumroad Affiliate Program participant now stop paying that 10% commission once the seller opts out of the program.

`Subscription#build_purchase` inherits the original purchase's affiliate rather than resolving one from cookies, and it gated that inheritance on `Affiliate#eligible_for_credit?` — which only checks that the affiliate is alive, unsuspended, and not on a Brazilian Stripe Connect account. New purchases go through `GlobalAffiliate#eligible_for_purchase_credit?`, which does check `User#disable_global_affiliate`. So the seller-facing opt-out reached new sales only.

This adds `Affiliate#eligible_for_credit_on_renewal?(product:)` as the renewal-time gate:

- The base implementation delegates to `eligible_for_credit?`, so direct affiliates and collaborators keep behaving exactly as before.
- `GlobalAffiliate` overrides it to also require that the product's seller is still in the program.
- The other purchase-time conditions are deliberately **not** re-evaluated on renewals: whether the product is currently on Discover, and whether the buyer happens to be the affiliate, describe how the referral was originally made. Re-judging them months or years later would strip commission an affiliate legitimately earned. A regression test pins that distinction.

## Why

A seller turned off the Gumroad Affiliate Program in January and kept being charged 10% on every monthly renewal of a membership that had been referred in December 2024 — seven charges after the opt-out, one as recently as yesterday. From her side the setting looked broken, and there was no lever to break the link: mutating the original purchase's `affiliate_id` by hand was the only workaround, and that's not something a seller or support can reasonably do.

The setting reads "Opt out of the Gumroad Affiliate Program" with no carve-out for existing referrals, so this is the behaviour it promises.

Blast radius is small but ongoing. Across the whole platform in the last 30 days, subscription purchases carrying global-affiliate credit where the seller had already opted out: 5 rows, 5 sellers, $11.89 total. It is a correctness and trust problem rather than a material money leak — which is precisely why it produces angry tickets out of proportion to the dollars.

Tracking issue: filed privately; not linked here because this is a public repository.

## Before / After

Not applicable — no user-visible surface changes. This is a server-side eligibility gate evaluated when a renewal purchase is built; nothing renders differently. The seller-visible effect is the absence of an affiliate row and a $0 affiliate fee on future renewal sales, which the existing sale detail view already renders from the same data.

## QA steps

Reviewers can verify without a browser:

```ruby
# rails console
seller = User.find_by(email: "<seller>")
sub    = seller.products.first.subscriptions.alive.find { _1.original_purchase.affiliate.is_a?(GlobalAffiliate) }

sub.build_purchase.affiliate         # => the GlobalAffiliate
seller.update!(disable_global_affiliate: true)
sub.reload.build_purchase.affiliate  # => nil   (was: still the GlobalAffiliate)
```

A direct affiliate on the same subscription is unaffected by the seller's global opt-out — `eligible_for_credit_on_renewal?` for `DirectAffiliate` falls through to the unchanged `eligible_for_credit?`.

## Test Results

New coverage:

- `test/models/subscription_test.rb` — four `#build_purchase` cases: global affiliate inherited normally, dropped after the seller opts out, direct affiliate unaffected by that opt-out, and any affiliate that has become ineligible entirely still dropped.
- `spec/models/global_affiliate_spec.rb` — `#eligible_for_credit_on_renewal?` true in-program, false after opt-out, false when the affiliate itself is ineligible, false for Brazilian Stripe Connect sellers, and still true for a product that isn't on Discover.
- `spec/models/affiliate_spec.rb` — direct affiliates ignore `disable_global_affiliate` but stop on the seller-side Brazilian Stripe Connect gate.

All run locally on this branch. The whole `subscription_test.rb` file is green (434 runs, 0 failures). Reverting `Subscription#build_purchase` to the old predicate reddens exactly the opt-out test, so the new gate is load-bearing rather than incidentally passing.

11 pre-existing failures in `spec/models/global_affiliate_spec.rb` / `spec/models/affiliate_spec.rb` are unrelated to this change: they need a live Stripe charge that has no cassette in a local worktree, and the same 11 fail identically on unmodified `origin/main` in the same worktree. CI arbitrates.

Test results

---

## AI disclosure

Instructions given: investigate a seller report that affiliate commission was still being taken after she opted out of the Gumroad Affiliate Program, verify it against production data, and — after the product ruling came back that the opt-out should apply — build and ship the fix with tests. The model read the affiliate/subscription models, confirmed the divergent predicates, measured the platform-wide blast radius on the production read replica, wrote the gate and its tests, ran them locally including a mutation check and a clean-main baseline, and authored this description.


gumclaw:
## Fable pre-merge review

Verdict: READY-TO-MERGE after fixing the one P2 it found.

Review run: local fable-5 adversarial review against `origin/main...HEAD`, focused on renewal lanes, seller-vs-product ownership, `try` → `&.` semantics, downstream nil-affiliate reads, money split correctness, and whether the tests actually bite.

Findings:

| Severity | Finding | Resolution |
|---|---|---|
| P2 | The new renewal hook fixed `disable_global_affiliate`, but left another seller-side purchase-time gate — Brazilian Stripe Connect sellers — purchase-time-only. | Fixed in `fix: stop renewal affiliate credits for Brazilian sellers`: `GlobalAffiliate` and `DirectAffiliate` now stop renewal credits when the seller uses a Brazilian Stripe Connect account, with predicate specs. |
| P3 | Plan changes during an opt-out window can permanently sever the affiliate link even if the seller later opts back in. | Not changed here; existing behaviour for any ineligible affiliate copied into a new original purchase. |
| P3 | Preorder release has a similar but separate inheritance gate using `alive?`. | Follow-up class, not this renewal path. |
| P3 | QA media filename did not include the PR number. | Left as-is because the PR body URL is valid and the image validator passes; the file is backend evidence, not UX screenshot evidence. |

Fable explicitly checked and found clean:

- All possible `Purchase#affiliate` STI receivers respond to `eligible_for_credit_on_renewal?`; `original_purchase` nil semantics are unchanged.
- `product: link` is the same seller account that purchase-time global affiliate eligibility uses.
- No other recurring-charge lane inherits `original_purchase.affiliate` outside `Subscription#build_purchase`.
- Nil affiliate on a renewal was already supported and downstream readers guard it.
- Dropping the affiliate sends the 10% back to the seller rather than retaining it for Gumroad.
- The opt-out regression test fails on the old predicate and passes on the new one.

gumclaw:
## Self-review

Re-read the diff against CONTRIBUTING.md line by line rather than from memory. Calling out the lines that carry risk and why I think they're safe.

**`app/models/subscription.rb:321` — `try(:eligible_for_credit?)` → `&.eligible_for_credit_on_renewal?(product: link)`**

This is the whole behaviour change and it sits on the money path for every membership renewal on the platform, so three things had to hold:

1. **The receiver set.** `Purchase#affiliate` is an `Affiliate` STI association, so the receiver can be `GlobalAffiliate`, `DirectAffiliate`, or `Collaborator`. The new method is defined on the `Affiliate` base class, so every subclass responds to it — no `NoMethodError` lane. `Collaborator` gets the base implementation, which is literally the old predicate, so collaborator-referred renewals are byte-for-byte unchanged.
2. **`try` → `&.` is a real semantic narrowing.** `try` swallows a missing method; `&.` only swallows `nil`. That narrowing is deliberate and is why point 1 mattered — I wanted a missing override to blow up loudly in CI rather than silently stop paying an affiliate. `original_purchase.affiliate` is nil for the overwhelming majority of subscriptions and `&.` handles exactly that case.
3. **The argument.** `product: link` — `link` is the subscription's product, and the predicate reads `product.user` for the seller-side settings. For a global affiliate `product.user` is the seller who opted out; for a direct affiliate it is the same seller who owns the affiliate relationship. Not `affiliate.seller`, which would have been the subtly wrong receiver for `GlobalAffiliate` (its `seller` is nil).

**`GlobalAffiliate#eligible_for_credit_on_renewal?` — what it deliberately does NOT check**

The risk on a change like this is over-correcting: copying `eligible_for_purchase_credit?` wholesale would also re-check Discover listing and buyer-is-the-affiliate at renewal time, which would quietly strip commission from referrals that were legitimately earned years ago, on products whose Discover status has drifted since. That would be a second, larger bug pointing the other way. The two conditions I skip describe *how the referral was originally made*; the two I keep (`disable_global_affiliate`, Brazilian Stripe Connect) describe *the seller's current state*. That distinction is pinned by the "keeps paying a referral whose product is not (or no longer) on Discover" example, which asserts `eligible_for_purchase_credit?` is false and `eligible_for_credit_on_renewal?` is true on the same object — so a future refactor that collapses the two predicates into one reddens immediately.

**`DirectAffiliate#eligible_for_credit_on_renewal?`**

Added after the fable review flagged that the Brazilian Stripe Connect gate is also a *seller-side* purchase-time condition, not a referral fact, so leaving it purchase-time-only reproduced the same class of bug I was fixing. Note the two Brazil checks are different accounts and both are wanted: `eligible_for_credit?` checks the *affiliate's* account, the new clause checks the *seller's*.

**Tests bite rather than incidentally pass.** Reverting `subscription.rb` to the old predicate reddens exactly the opt-out test and nothing else — recorded in the body. The `@subscription.reload` in the test helper is load-bearing, not defensive: `original_purchase` is a `has_one` already loaded by the shared setup, so without the reload the assertion would read an affiliate-less copy and pass for the wrong reason.

**Not user-visible.** No view, presenter, serializer, or JS touched — the only surfaced effect is the absence of an affiliate row and a $0 affiliate fee on future renewal sales, which the existing sale-detail view already renders from this same data. That is why the body's Before/After is a stated no-op rather than a screenshot pair, per the "no user impact needs confirming evidence" rule.

**Migration/backfill: none, intentionally.** This changes what future renewals build, not existing rows. The seven charges the reporting seller already paid are a support-side refund decision tracked on antiwork/gumroad-private#1414, not something to fold into this diff.

gumclaw:
@ershad requesting your review — this one changes affiliate commission on the membership-renewal path, so it's a money-split change and I've deliberately left auto-merge off for a human merge decision.

State: green at head `2db11d2` (77 checks pass, 8 `pull_request_target` noops, zero failing), Greptile 5/5 with no blocking findings, local fable adversarial review run with its one P2 fixed in-branch (Brazilian Stripe Connect sellers were a second seller-side gate stuck at purchase-time-only), no unresolved review threads, self-review comment above.

The specific thing worth your eyes is the scope call in `GlobalAffiliate#eligible_for_credit_on_renewal?`: it re-checks the seller's *current* state (program opt-out, Brazilian Stripe Connect) but deliberately does **not** re-check Discover listing or buyer-is-the-affiliate, because those describe how the referral was originally made and re-judging them years later would strip commission an affiliate legitimately earned. If you'd rather renewals re-evaluate every purchase-time condition, that's a one-line change and a different product decision — say so and I'll make it.

Blast radius is small and measured: 5 rows / 5 sellers / $11.89 platform-wide over the last 30 days. It's a trust problem rather than a money leak, which is why it generates angry tickets out of proportion to the dollars. Tracking issue: antiwork/gumroad-private#1414.

PR #6399, opened 2026-07-27
Keep internal payout notes off the seller's Payouts dashboard

gumclaw:
## What

Payout notes now record whether they may be shown to the seller, and the banner on the Payouts page only renders the ones that may.

`payout_note` comments were doing two unrelated jobs. Some are written for the seller and rendered verbatim in a banner on their Payouts page ("Your payout on June 12 was skipped because your balance of $59.72 was below the $100 minimum"). Others are breadcrumbs for support and for the automated payout-setup retry loop's own bookkeeping. Nothing on the record distinguished them, so the page rendered whichever was newest.

Changes:

- `User#add_payout_note` takes a `seller_visible:` keyword (default `true`) and stores it in the comment's `json_data`.
- New `PayoutNoteVisibility` is the single place that reads it, with a fallback for rows written before the flag existed: a closed list of the internal note shapes we have written historically. Everything else defaults to visible, so nothing a seller sees today disappears.
- Every internal writer now passes `seller_visible: false`: the six notes in `RetryStripeRejectedPayoutSetupForSellerJob`, both Stripe bank-sync / postal-code breadcrumbs, the payout-drift note, the Sidekiq retries-exhausted breadcrumb, and the PayPal skip notes (which the banner already excluded by content).
- `PayoutsHelper` replaces the `content LIKE 'Payout via PayPal%'` filter with the visibility check, and walks back a bounded number of notes because visibility lives in `json_data` and MySQL cannot filter on it usefully.

Not included: the two content-regex suppressions in `payouts_helper.rb` stay for now. They are not leak suppressions — they fold a genuinely seller-facing note into a richer notice that carries the current balance — so collapsing them into this mechanism would change what sellers read, not just what they are allowed to read.

## Why

A seller in the Philippines fixed their bank code, then screenshotted their own Payouts dashboard showing:

> Automated Stripe payout-setup retry stopped: the bank code was rejected on format, so re-sending the same saved details can never succeed. The seller has been emailed and has to re-enter the code.

That is copy written for staff, describing our own retry machinery, shown to the person it is about. Two of the sibling notes are worse than awkward: `ACCOUNT_BLOCKED_NOTE` tells a risk-blocked seller that payments on their Stripe account are "blocked at the platform level", and `GAVE_UP_NOTE` tells them "manual follow-up is needed", both in internal words with no context.

The retry job's notes exist for support and for the job's own `abandoned_at` / `retry_count` bookkeeping. Anything the seller needs to know is already sent as an email (`notify_format_rejection!`, `notify_retries_exhausted!`), so none of these notes ever needed to reach them.

The internal tracking issue for this work is filed privately and is deliberately not linked here, since this is a public repository.

## Before / After

No user-visible change is intended for sellers who only ever had seller-facing notes: the banner they see today is unchanged. The change is that sellers with an internal note on their account stop seeing it.

| | Before | After |
|---|---|---|
| Seller with an internal retry note | banner renders the internal sentence verbatim | no banner |
| Seller with a below-minimum note | banner shows the below-minimum notice | unchanged |
| Seller with a skipped-payout note | banner shows the skipped-payout notice | unchanged |

Rendered evidence, captured with a throwaway Capybara system spec run twice against the same seeded
scenario: once on this branch, once on its parent commit. The seller has a $59.72 balance (below the
$100 minimum) and one internal note from the payout-setup retry job.

**One internal retry note on the account, nothing else**

| Before | After |
|---|---|
| before-internal-note-only | after-internal-note-only |

Before, the seller reads `Automated Stripe payout-setup retry stopped: the bank code was rejected on
format, so re-sending the same saved details can never succeed. The seller has been emailed and has to
re-enter the code.` After, only the below-minimum notice remains.

**A seller-facing note, then a newer internal one recorded on top of it**

| Before | After |
|---|---|
| before-internal-note-newest | after-internal-note-newest |

Before, the newer internal note (`Automated retries to fix the rejected postal code / bank account were
exhausted. Manual follow-up is needed.`) displaced the seller-facing one. After, the seller-facing
sentence is shown instead.

**Seller-facing note on its own — unchanged**

after-seller-facing-note

The capture spec also asserted the text presence directly, and the two runs disagree exactly where they
should:

| assertion | parent commit | this branch |
|---|---|---|
| internal retry copy visible, internal note only | `true` | `false` |
| seller-facing note visible | `true` | `true` |
| internal copy visible when it is the newest note | `true` | `false` |
| seller-facing note visible when an internal note is newer | `false` | `true` |

Captured on the deployed preview app for this branch, `https://fix-internal-payout-notes-not-se.apps.staging.gumroad.org`
(`x-revision: ad707d92b330`), logged in as `seller@gumroad.com`, at `/payouts`. Payout notes were
seeded on that account through the preview write console (`user.add_payout_note(...)`), exactly as the
QA steps below describe. The seller's balance is below the $100 minimum, so the below-minimum notice is
always present — the question in each shot is whether a *second*, note-driven banner appears above it.

**Internal note only (`seller_visible: false`) — no note banner, only the below-minimum notice**

preview internal note hidden

The seeded note is `BANK_FORMAT_REJECTION_NOTE` ("Automated Stripe payout-setup retry stopped: the bank
code was rejected on format…"). None of that sentence reaches the page.

**Seller-facing note, with a newer internal note recorded on top of it — the seller-facing sentence is shown**

preview seller-facing note shown

`Payout on June 12, 2026 was skipped because a bank account wasn't added at the time.` renders in the
banner even though `GAVE_UP_NOTE` was written after it; the helper walks back past the internal note
instead of stopping at the newest one. (Verified in the same capture run both with and without the newer
internal note present — the rendered page is byte-identical, which is the point of the change.)

**Legacy row with no flag stored (`json_data: {}`) — still hidden, mobile at 375px**

preview no banner mobile

`note.update!(json_data: {})` simulates a note written before the flag existed;
`PayoutNoteVisibility` falls back to the closed list of historical internal shapes and keeps it out.

## QA steps

1. In the Rails console, on a seller with a balance below the payout minimum:
   `user.add_payout_note(content: RetryStripeRejectedPayoutSetupForSellerJob::ACCOUNT_BLOCKED_NOTE, seller_visible: false)`
2. Load `/payouts` as that seller. No banner should appear.
3. Add a seller-facing note: `user.add_payout_note(content: "Payout on June 12, 2026 was skipped because a bank account wasn't added at the time.")`
4. Reload `/payouts`. The banner should show that sentence.
5. Add another internal note after it and reload — the seller-facing sentence should still be the one shown.
6. Simulate a legacy row: `note = user.add_payout_note(content: RetryStripeRejectedPayoutSetupForSellerJob::GAVE_UP_NOTE); note.update!(json_data: {})`, then reload. Still no banner.
7. In the admin user page, confirm the internal notes are all still visible to staff (nothing is hidden from the admin side).

## Test Results

Run locally in a worktree off `main` (`bundle exec rspec`):

- `spec/models/payout_note_visibility_spec.rb` — 4 examples, 0 failures (new)
- `spec/helpers/payouts_helper_spec.rb` — new "internal payout notes" group green; the one failure in that file (`:201`, affiliate-credit balances) fails identically on unmodified `main`
- `spec/sidekiq/retry_stripe_rejected_payout_setup_for_seller_job_spec.rb` — green
- `spec/controllers/api/internal/admin/payouts_controller_spec.rb` — green
- `spec/presenters/settings_presenter_spec.rb` — same single pre-existing failure (`:293`) as unmodified `main`
- `spec/requests/balance_pages_spec.rb` — compared against an unmodified `main` worktree; results match

---

## AI disclosure

Written by Claude Opus 5 (`claude-opus-5`) running as the Hermes agent, from an investigation recorded on the private tracking issue and the instruction there to fix the leak. The agent read the affected code paths, chose the explicit-flag approach over content pattern matching, made the edits, wrote the specs, and ran them locally against a baseline `main` worktree to separate pre-existing failures from new ones.


gumclaw:
The review flagged that no browser evidence was produced because the app could not boot in that environment. That gap is closed now — the body has real before/after captures of the Payouts page.

They come from a throwaway Capybara system spec run twice against the same seeded scenario (seller with a $59.72 balance, one internal note from the payout-setup retry job): once on this branch, once on its parent commit `f85ed3cf2`. So the pair is a genuine before/after of one code change rather than two hand-built states.

The spec also asserted text presence directly, and the two runs differ exactly where the change intends:

| assertion | parent commit | this branch |
|---|---|---|
| internal retry copy visible, internal note only | `true` | `false` |
| seller-facing note visible | `true` | `true` |
| internal copy visible when it is the newest note | `true` | `false` |
| seller-facing note visible when an internal note is newer | `false` | `true` |

The last row is the one that matters most: before, a newer internal breadcrumb displaced the seller-facing sentence; after, the seller still reads their own note.

The screenshots are committed to `qa-media/` on the branch. No video: this harness renders headless, so there is no screen recording to make — the three captured states cover every state the change introduces, and there is no interaction to demonstrate beyond loading the page.

gumclaw:

**Correctness checks I ran on the diff itself**

`add_payout_note` now writes `json_data` at create time, and `StripeMerchantAccountManager.record_bank_sync_failure_note` mutates that same hash right afterwards to add `stripe_error_code` / `stripe_error_message`. That is a merge into the existing hash, not a replacement, so the visibility flag survives. Same for the retry job, which later writes `abandoned_at` / `retry_count` / the notification markers onto notes it created — verified by the job specs still passing.

`PayoutNoteVisibility.seller_visible?` reads `json_data[...]` directly instead of going through `attr_json_data_accessor`. That is deliberate and load-bearing: those accessors fall back to their default when the stored value is blank, and `false` is blank in Rails, so an explicitly hidden note would read back as "no opinion" and render. There is a comment on the method saying so.

`Payment::FailureReason` compares `processor != PayoutProcessorType::PAYPAL`, and that constant is the uppercase `"PAYPAL"` — the same value the attribute holds — so the comparison is against the real value, not the `capitalize`d display form used in the sentence. The corresponding legacy entry has to match `"Payout via Paypal ..."`, which is why the legacy check is case-insensitive; there is a spec asserting that exact capitalization.

**Behavior changes beyond the reported leak, called out on purpose**

Three legacy note shapes were previously *visible* and are now hidden, because the old filter only excluded `Payout via PayPal%`: the Stripe bank-sync and postal-code breadcrumbs, and `[PAYOUT][DRIFT]`. All three are internal by any reading (the drift note ends with "the funds are stranded there and future payouts will fail the destination balance check"), so hiding them is the fix rather than a side effect — but it is a change to what some sellers see today and should not slip by unmentioned.

The `HandleNewBankAccountWorker` retries-exhausted note reads "Stripe bank sync failed and exhausted Sidekiq retries for bank_account_id=…", which starts with `BANK_SYNC_FAILURE_NOTE_PREFIX`, so legacy rows of it are covered by the existing entry without needing a separate one. It also leaks an internal database id, which is a second reason it should never have been renderable.

**The bounded scan**

The lookup ordering changed from an implicit primary-key `.last` to `created_at DESC, id DESC` with a `LIMIT 25`. The tiebreak on id keeps same-second notes deterministic, and the limit makes the query cheaper than the previous unbounded sort rather than more expensive. The failure mode when a seller genuinely has 25+ newer internal notes is "no banner", which is strictly better than the current behavior of showing them the internal one — there is a spec pinning that boundary so it cannot drift silently without a test failing.

**Admin side deliberately untouched**

`Api::Internal::Admin::PayoutsController#index` and `SettingsPresenter#postal_code_rejected_by_stripe?` both still read notes the old way, so support keeps seeing every note including the internal ones. Only the seller-facing surface narrowed. The admin payouts controller specs cover this and pass unchanged.

**Not done, with reasons rather than omissions**

The two content regexes in `payouts_helper.rb` stay. The issue suggested collapsing them into this mechanism, and I read them again before deciding not to: neither is a leak suppression. Both take a genuinely seller-facing note and fold it into a richer notice carrying the current balance, so routing them through visibility would change what sellers read rather than what they are allowed to read.

The sticky-banner half of the issue is not here either. This PR removes the specific banner the reporting seller saw, but "a seller-facing note outliving the problem it describes" is a distinct bug with a distinct fix, and bundling it would make this diff two changes wearing one hat.

**Artifacts**

Rendered before/after captures are in `qa-media/` and embedded in the body, produced by running the same throwaway Capybara spec against this branch and against its parent commit so the pair is a true before/after of one seeded scenario rather than two different setups. Spec results are in the body; the affiliate-credit failure I initially saw in `payouts_helper_spec` turned out to be missing seeded merchant accounts in my worktree, not the diff — after re-seeding, all 120 examples across the four affected files pass. RuboCop is clean on every touched Ruby file; the diff has no JS or TypeScript.

gumclaw:
CI is fully green at `ad707d92b` — all 76 checks pass, including every Test Fast and Test Slow shard, Minitest, both lint jobs, and Buildkite. Greptile came back 5/5 with no blocking findings.

Handing this over for review and merge rather than auto-merging, because it changes what a seller reads on their own Payouts page. The evidence for that is in the body: real before/after captures of the page from the same seeded scenario, run once on this branch and once on its parent commit.

Worth knowing before you look at it: three legacy note shapes that were previously visible become hidden, not just the one from the ticket. The old filter only excluded `Payout via PayPal%`, so the Stripe bank-sync and postal-code breadcrumbs and the `[PAYOUT][DRIFT]` note were all renderable too. All three are internal by any reading, so hiding them is the intended fix, but it is a wider behavior change than the issue title implies.

Support keeps seeing every note — the admin payouts controller and the settings presenter still read them the old way. Only the seller-facing surface narrowed.

Details on what is deliberately out of scope are in the self-review comment above: the two content regexes in the helper are not leak suppressions, and the sticky-banner half of the issue is a separate bug.

PR #6400, opened 2026-07-27
Stop a cart item that omits url_parameters from wiping the whole cart save

gumclaw:
## What

`CheckoutController#update` assigned `cart_product.url_parameters = item[:url_parameters]` unconditionally. When a client posts a cart item without a `url_parameters` key, `item[:url_parameters]` is `nil` and that assignment overwrites the empty hash `CartProduct#assign_default_values` had just set on the row. `nil` then fails the column's JSON-schema validation:

```
ActiveRecord::RecordInvalid: Validation failed: Url parameters
The property '#/' of type null did not match the following type: object
```

Because the assignment happens inside the `ActiveRecord::Base.transaction` block, the raise rolls back the **entire** cart update, and the buyer is redirected back to `/checkout` with the generic `"Sorry, something went wrong. Please try again."` alert. Since the cart-state save is debounced and re-fires on every cart change, the buyer loops on that alert — checkout is unusable, not merely degraded.

`referrer` (validated `presence: true`) sat on the adjacent line with the identical hazard.

This change treats an absent key as "leave the stored value alone" rather than "set it to nil". A brand-new item that genuinely supplies no `referrer` still fails validation, as it should — the fix does not weaken either validation.

## Why

Reported by Sentry error monitoring on the `CheckoutController#update` transaction: **10,290 events across 502 unique buyers**, first seen 2026-06-29 and still firing. Volume was a trickle (0–7/day) until 2026-07-26 11:00Z, then jumped to 9,382 events in a single day and held at 340–970/hour for 14 consecutive hours.

Diagnosis from sampled production events:

- `tags/transaction` is **10,290 / 10,290** on `CheckoutController#update` — a single code path, not a diffuse bucket.
- Real buyer traffic: 25 sampled events → 25 distinct buyer emails, 15 distinct IPs, ordinary desktop Chrome user agents, `environment: production`.
- Every post-spike event's cart item is missing `url_parameters` entirely:
  `{affiliate_id, force_new_subscription, option_id, pay_in_installments, price, product, quantity, recurrence, referrer, rent}` — 10 keys.
- Events sampled from before the spike carry the full 15-key shape including `url_parameters`. So the trigger was a change in the payload shape a client sends, not a change in server-side validation.

The client-side question — which arrival path assembles cart items without `url_parameters` (the normal path through `computeInitialCheckout` always sets it) — is being tracked separately. This PR is the server-side half: the controller should never turn an omitted key into data loss on a validated column.

#4269 attempted the same fix in April 2026 and was closed unmerged over an unrelated spec assertion, so the class has been latent since then.

## Before / After

No markup changes — the fix is one controller line, so the evidence is a **held invariant**: the
same cart renders identically, and the only thing that moves is whether the save succeeds.

Captured with one Playwright script (deviceScaleFactor 2), the same seeded product
(`/l/demo`, "Beautiful widget", US$0) and the same viewport on both legs. The script strips
`url_parameters` from the debounced `PATCH /checkout` body **on the wire**, reproducing the exact
10-key payload the reported production clients send — the browser UI itself can never build it.
The only variable between the two columns is the build under test:

| | before — unfixed `main` | after — this branch |
|---|---|---|
| preview | `add-gambia-payouts` (`b82a381efa45`, carries the unfixed `cart_product.url_parameters = item[:url_parameters]` line) | [`fix-checkout-nil-url-parameters`](https://fix-checkout-nil-url-parameters.apps.staging.gumroad.org) (`36a626b32`) |
| `PATCH /checkout` | 303, flash alert set | 303, no flash |
| visible toast at capture | `Sorry, something went wrong. Please try again.` | *(none)* |
| `CartProduct.count` after the run | **0** — transaction rolled back | row persisted with `url_parameters: {}` |

**Desktop (1440px)**

| before | after |
|---|---|
| before desktop | after desktop |

**Mobile (375px)**

| before | after |
|---|---|
| before mobile | after mobile |

Note what did **not** move: in all four shots the cart is the same single line item, same
`Subtotal`/`Total` of US$0, same `Give as a gift?` row and same `Get` button. The toast is the
whole delta.

**Console read-back (staging Rails console, read replica)** — the pixels only show the toast is
gone; this shows the data is actually right, including the case a screenshot cannot express
(QA step 4: an already-stored hash must survive a later key-omitting save).

cart product read-back

- Five new-item legs, `url_parameters` omitted on the very first save → five rows persisted with
  the default `{}`, none `nil`.
- One two-leg run in a single browser context (same `_gumroad_guid`, `cart_id=7`): leg 1 sends
  `{"utm_source" => "google", "utm_campaign" => "summer"}`, leg 2 re-saves the same cart with the
  key omitted. Read-back shows the UTM hash **unchanged** — the omitted key left it alone rather
  than nil-ing it.
- Control on the fixed build with the strip disabled (normal 15-key payload): still no toast, so
  the pass is not the script silently no-op-ing.

## QA steps

Preview app: <https://fix-checkout-nil-url-parameters.apps.staging.gumroad.org>

Click path used for the screenshots above: open `https://seller.fix-checkout-nil-url-parameters.apps.staging.gumroad.org/l/demo`, then `https://fix-checkout-nil-url-parameters.apps.staging.gumroad.org/checkout?product=demo&quantity=1` and let the debounced cart save fire (~8s) with `url_parameters` removed from the outgoing `PATCH /checkout` body. Expect no error toast; on unfixed `main` the toast appears and nothing persists.

1. Deploy the branch (or run locally).
2. `PATCH /checkout` as a signed-in buyer with a cart item that omits `url_parameters`:
   ```json
   {"cart":{"discountCodes":[],"items":[{"product":{"id":"<external_id>"},
     "option_id":null,"recurrence":null,"price":100,"quantity":1,
     "rent":false,"referrer":"direct"}]}}
   ```
3. Expect **303 See Other** to `/checkout` with no flash alert, and a persisted `CartProduct` whose `url_parameters` is `{}`.
   Before this change the same request returned **302 Found** with `"Sorry, something went wrong. Please try again."` and persisted nothing.
4. Repeat against a cart product that already has `url_parameters: {"utm_source" => "google"}` and omit the key — the stored value must survive untouched.
5. Regression check: a normal checkout that *does* send `url_parameters` (add a product to the cart from a product page with `?utm_source=x` in the URL) must still persist those params.

## Test Results

Two new red-first examples in `spec/controllers/checkout_controller_spec.rb`, proven load-bearing by reverting the controller to `origin/main` (both fail; the second reproduces the exact production symptom — a 302 from the `RecordInvalid` rescue's alert redirect), then restoring the fix (both pass).

```
bundle exec rspec spec/controllers/checkout_controller_spec.rb spec/models/cart_product_spec.rb
56 examples, 0 failures
```

spec results

---

## AI disclosure

Generated with **claude-opus-5**. The model was given the Sentry issue's event samples (payload shapes before and after the volume spike, transaction tag distribution, event/user counts), asked to trace the failure to a specific `file:line` and explain why the value was nil on that path before writing any code, then asked to write regression specs and prove them red against `origin/main` before restoring the fix.


gumclaw:
@xrav3nz routing this to you and flagging it as `priority` — this is an **active production incident**, not a latent bug, and it's in checkout UI's blast radius.

**Why you rather than Razvan:** the reviewer-routing convention sends checkout UI to you. Razvan is the most recent human author on `checkout_controller.rb` (2026-06-29) and on its spec, which is the argument for him — but he's currently holding four open Gumclaw PRs including two with changes requested, and this one shouldn't queue behind them. Happy to re-route to him if you'd rather this be his call; say the word.

**State:** `MERGEABLE`, no conflicts, CI running at head `36a626b32` (the last commit is a QA-media rename, zero application code — the code commit was green).

**Why this is urgent.** Sentry: **10,290 events across 502 unique buyers** on the `CheckoutController#update` transaction, first seen 2026-06-29 and *still firing*. Volume was a trickle (0–7/day) until **2026-07-26 11:00Z**, then jumped to 9,382 events in a single day and has held at 340–970/hour for 14 consecutive hours.

This is real buyer traffic, checked rather than assumed: 25 sampled events gave 25 distinct buyer emails, 15 distinct IPs, ordinary desktop Chrome user agents, `environment: production`. And `tags/transaction` is 10,290 / 10,290 on that one action — a single code path, not a diffuse bucket.

**And the user impact is worse than "an error was logged."** The failing assignment happens *inside* the `ActiveRecord::Base.transaction` block, so the raise rolls back the **entire** cart update and the buyer bounces back to `/checkout` with the generic "Sorry, something went wrong. Please try again." Because the cart-state save is debounced and re-fires on every cart change, the buyer **loops on that alert**. Checkout is unusable for them, not merely degraded.

**Mechanism**, verified in the source and not just from the stack trace — `main` has:

```ruby
cart_product.url_parameters = item[:url_parameters]
cart_product.referrer = item[:referrer]
```

Unconditional. When a client posts a cart item with no `url_parameters` key, `item[:url_parameters]` is `nil`, and that assignment overwrites the empty hash `CartProduct#assign_default_values` had just put on the row. `nil` then fails the column's JSON-schema validation (`The property '#/' of type null did not match the following type: object`). `referrer` sat on the adjacent line with the identical hazard (`presence: true`).

The change is two `unless ...nil?` guards: treat an absent key as "leave the stored value alone" rather than "set it to nil". **Neither validation is weakened** — a brand-new item that genuinely supplies no `referrer` still fails, as it should.

**The two things I'd actually like your judgement on:**

1. **This is the server-side half only, and the client-side cause is still open.** Every post-spike event's cart item is missing `url_parameters` entirely — a 10-key shape — while events sampled from *before* the spike carry the full 15-key shape including it. So the trigger was **a change in the payload shape a client sends**, not a change in server-side validation. The normal path through `computeInitialCheckout` always sets the key, so some other arrival path is assembling cart items without it. That's tracked separately, but you're the person most likely to recognise which path on sight — if you do, that's the actual root cause and worth more than this patch.
2. **Whether the guard belongs in the controller at all.** An arguably cleaner fix is for `CartProduct` to coerce nil to `{}` in the setter, so no caller can do this again. I kept it in the controller because it's the minimal, most reviewable change during a live incident and doesn't alter model semantics for other callers — but the controller now carries a seven-line comment explaining a model invariant, which is a smell.

**Also worth knowing:** #4269 attempted this same fix in April 2026 and was closed unmerged over an unrelated spec assertion. The bug class has been latent since then and has now gone loud.

**Verification:** two new red-first examples in `spec/controllers/checkout_controller_spec.rb`, proven load-bearing by reverting the controller to `origin/main` (both fail, and the second reproduces the exact production symptom — a 302 from the `RecordInvalid` rescue's alert redirect), then restoring the fix (both pass). `bundle exec rspec spec/controllers/checkout_controller_spec.rb spec/models/cart_product_spec.rb` → **56 examples, 0 failures.**

gumclaw:
Taking this back — CI is red at the current head (`Test Slow 20` failing), so this is mine, not yours to look at yet. Reassigned to me with `working`.

I'll fix the failure, get the suite green, and hand it back for review with the evidence attached. Nothing needed from you until then.

slavingia:
> Why this is urgent. Sentry: 10,290 events across 502 unique buyers on the CheckoutController#update transaction, first seen 2026-06-29 and still firing. Volume was a trickle (0–7/day) until 2026-07-26 11:00Z, then jumped to 9,382 events in a single day and has held at 340–970/hour for 14 consecutive hours

What code caused this?

This is failing specs. Fix.

PR #6401, opened 2026-07-27
Stop computing entitled creator posts for the mobile library search response

gumclaw:
## What

Stops the mobile library **search** endpoint (`Api::Mobile::PurchasesController#search`) from computing `product_updates_data` — the list of creator posts a buyer is entitled to see for each product in the page. `#index`, `#purchase_attributes` and `UrlRedirectsController#url_redirect_attributes` are unchanged and still return the field.

`json_data_for_mobile` takes a new `include_product_updates:` option (default `true`, so every other caller behaves exactly as before) which threads through `UrlRedirect#product_json_data` and `Preorder#mobile_json_data`. `#search` passes `false`, and the batched preloader is skipped along with it.

## Why

Deciding which posts a buyer can see means, for every (product, buyer email) group on the page, reading the seller's whole published-post history and re-running the buyer-targeting filters over it. That is two `installments` query shapes plus per-buyer purchase-history probes, all of which run dozens of times per request and scale with library size.

Measured on the eight slowest production samples of this endpoint in the last 24 hours (52.1 s of wall time), grouping every database span by which feature it serves:

| work | time | share of wall |
|---|---|---|
| posts entitlement (`installments`, targeting probes, `email_infos`, post `url_redirects`) | 42.9 s | **82%** |
| everything else the response contains | 7.6 s | 15% |

Per sample the entitlement share runs 48–93% of wall time. This is the work behind the endpoint's `Rack::Timeout` tail.

Three prior rounds of optimization (#6123, #6158, #6206) plus this morning's index (#6393) each made these queries individually cheaper without moving the endpoint's p95, because the cost is query *count* per request, not cost per query.

**Nothing reads the field from this response.** The current app has exactly one reader of `product_updates_data`, a `usePost` hook in `components/library/use-purchases.ts` that no screen calls — it was orphaned when push-notification support moved the post screen to `useInstallment`, which fetches a single post on demand from `Api::Mobile::InstallmentsController#show` (p95 71 ms). The two retired native apps did cache the field for offline use, but they read it from `purchases#index` (8 requests in the last 30 days) and `url_redirects#url_redirect_attributes`, both untouched here. `purchases#search`, which serves 40,934 requests over the same period, is called only by the current app.

Tracking issue: filed privately; not linked here because this is a public repository.

## Before / After

No rendered surface changes — this removes a JSON field that no client renders. The library list, product page and post screen all draw from other fields or from a different endpoint, and the specs added here assert `#index` and `#purchase_attributes` still return the field for the same purchase.

Walkthrough video of the affected functionality (the search response before and after, the endpoints that keep the field, and the measured numbers):

https://github.com/antiwork/gumroad/raw/perf/drop-product-updates-from-library-lists/qa-media/pr-6401-walkthrough.mp4

Response shape for one purchased product with one delivered post:

| | `product_updates_data` | response size |
|---|---|---|
| `#search` before | `["A post"]` | 1,682 bytes |
| `#search` after | absent | 1,159 bytes |
| `#index` / `#purchase_attributes`, after | `["A post"]` | unchanged |

## Performance

Local harness, 10 products across 10 sellers, 20 published posts per seller (half with a delivered email), serializing one 10-item library page. Three uncached runs, both paths in the same process:

| run | with posts | without posts |
|---|---|---|
| 1 | 428 queries, 424.9 ms, 62,906 bytes | 81 queries, 62.4 ms, 11,719 bytes |
| 2 | 196 queries, 174.1 ms, 62,906 bytes | 81 queries, 67.2 ms, 11,719 bytes |
| 3 | 196 queries, 174.4 ms, 62,906 bytes | 81 queries, 48.6 ms, 11,719 bytes |

Steady state is 196 → 81 queries (−59%), 174 ms → 49–67 ms (−62 to −72%), and the response body drops from 63 KB to 12 KB. Run 1 carries first-call cache warming in both paths.

Production numbers for the endpoint go in a follow-up comment once there are 48 hours of post-deploy samples; the pre-deploy baseline (p50 130 ms, p95 1,312 ms, p99 2,788 ms over 17,132 requests in the last 7 days) is recorded on the tracking issue.

## Test plan

- `spec/controllers/api/mobile/purchases_controller_spec.rb` — four new examples: the field is absent from the search response; the search request runs neither `Purchase.preload_product_updates_data!` nor either `Installment` scope; `#purchase_attributes` and `#index` still return the field, with the post's name, for the same purchase.
- Existing suites covering every caller of the changed methods, all green locally: `purchases_controller_spec.rb` (62), `purchase/purchase_installments_spec.rb` (55), `preorder_spec.rb` (41) — 158 examples, 0 failures. `api/mobile/url_redirects_controller_spec.rb` run separately, 26 examples 0 failures.
- `rubocop` clean on all five changed files.

## Test Results

spec and benchmark output

## QA steps

1. Deploy to a preview app and sign into the mobile app (or hit the endpoint directly with a `mobile_api` token).
2. `GET /mobile/purchases/search?items=10` — the response items no longer carry `product_updates_data`; everything else (`name`, `file_data`, `url_redirect_token`, `purchase_id`, seller aggregation, pagination) is unchanged.
3. In the app: open the library, scroll and search — products list and open normally.
4. Open a product, tap a post link inside the content — the post opens, its body and files render. This is the `installments/:id` path, deliberately unaffected.
5. `GET /mobile/purchases/index` and `GET /mobile/url_redirects/get_url_redirect_attributes/:id` — both still return `product_updates_data` populated.

---

## AI disclosure

Instructions given: answer the question on the private tracking issue asking what product functionality leads to these slowdowns and whether we need it to that extent. That meant re-measuring the endpoint's span profile from production Sentry, tracing `product_updates_data` to its consumers across `gumroad-mobile`, the retired `gumroad-ios` and `android-consumer` apps, and comparing traffic across the three endpoints that serve it. The finding — that the current app never reads the field it is paying 82% of the request for — led to this change.


gumclaw:

**Scope of the flag.** `include_product_updates:` defaults to `true` in all four places it was added, so `#index`, `#purchase_attributes`, `preorders#preorder_attributes`, `url_redirects#url_redirect_attributes`, `sales#show` and every other caller of `json_data_for_mobile` / `product_json_data` / `mobile_json_data` behave byte-identically. Only `#search` passes `false`. The two existing specs in the new block assert that positively rather than trusting the default.

**`options.fetch` vs `options[]`.** `json_data_for_mobile` takes a positional options hash (it already carries `include_sale_details`), so I used `options.fetch(:include_product_updates, true)` rather than adding a keyword — a keyword would have been a nicer signature but would break the existing `json_data_for_mobile({ include_sale_details: true })` callsite in `Api::Mobile::SalesController`. Keys don't collide.

**Removing a field from a public-ish API response.** I checked every consumer rather than assuming, because the mobile API is not versioned:

- `antiwork/gumroad-mobile` — one reader, `usePost` in `components/library/use-purchases.ts`, and nothing calls it. It was orphaned in March when push-notification support (#138) moved the post screen from `usePost` to `useInstallment`. The product screen is a WebView on the site, so it doesn't use this payload either.
- `antiwork/gumroad-ios` (retired) — reads it in `CoreDataManager.saveOrUpdateInstallments`, from `purchases/index`, behind `if let`, so an absent key is a no-op. Untouched by this PR.
- `antiwork/android-consumer` (retired) — reads it as `productUpdatesData` from `purchases/index`; `Product.syncUpdates` early-returns on null. Also untouched.
- Traffic over 30 days: `purchases/search` 40,934 requests, `purchases/index` 8. The endpoint I changed is only reachable from the current app.

**The `if include_product_updates` guard omits the key rather than sending `null`.** Deliberate: `null` would be a new value for clients that treat the key as present-but-empty, whereas an absent optional key is what the TypeScript type (`product_updates_data?: Post[]`) already models. Both retired clients are null-safe anyway, and neither reads it from this endpoint.

**Skipping the preloader as well as the serialization.** `Purchase.preload_product_updates_data!` is the expensive call; leaving it in and only dropping the field would have kept 100% of the cost. It's gated on the same flag, and the second new spec asserts neither it nor either `Installment` scope runs on a search request — that's the assertion that would catch a future refactor quietly reinstating the work.

**What I did not do.** I left `url_redirects#url_redirect_attributes` alone even though it returns the same field and its slowest samples are also ~80% posts work. It's 18,756 requests over 30 days with a p95 of 265 ms, both retired clients read the field from it, and the current app calls it for the single purchase the user just opened — where the posts list is plausibly wanted. Widening this PR to cover it would mix a safe removal with a debatable one. Also out of scope: the underlying per-(product, buyer email) fan-out in `preload_product_updates_data!`, which this PR makes rare rather than fixes.

**Testing.** Ran the specs for every caller of the changed methods, not just the endpoint: `purchases_controller_spec.rb` 62, `purchase/purchase_installments_spec.rb` 55, `preorder_spec.rb` 41 (158 examples, 0 failures), plus `api/mobile/url_redirects_controller_spec.rb` 26 examples 0 failures run separately — that file has to run alone in this worktree, it fails on unrelated card-charge factory errors when batched with others, and I verified those 66 failures are identical on unmodified `origin/main` in the same worktree before believing them. `rubocop` clean on all five files.

**Benchmark honesty.** The numbers in the Performance section are from a local harness, not production, and I've labelled them that way. Run 1 is inflated on both sides by first-call cache warming; the steady-state rows are the ones to read. Production p95/p99 comparison goes in a comment here once there are 48 hours of post-deploy samples — the last four rounds on this endpoint each improved a query without moving the endpoint, so I'm not claiming a production win before the data is in.

**Pre-merge adversarial review.** Done as a manual adversarial pass rather than a delegated fable-5 subagent (no delegation tool available in this session), covering: default-value coverage on all four signatures, the omitted-key-vs-null contract, null-safety in both retired clients, whether the JSON schema (`spec/support/schemas/api/mobile/product.json`, which lists `product_updates_data` as required) still validates — it does, because `purchases.json` only refs `product.json` under the `products` key that `#index` returns, and `#search` returns `purchases`, which the schema doesn't constrain. Worth noting that gap exists independently of this PR: the search response has never been schema-validated.

gumclaw:
## Pre-merge adversarial review — READY-TO-MERGE (after two P3 fixes, pushed)

Ran the local fable-5 review against `46b9ee9` with a hunt list written for this change class: *silently drop a field from a shared serializer on one of its callers*. Verdict was READY-TO-MERGE with **no P1 and no P2**. Two P3s, both real, both now fixed in `ef00ad5` rather than deferred.

### Findings

| # | Severity | Where | Finding | Disposition |
|---|---|---|---|---|
| 1 | P3 | `spec/support/schemas/api/mobile/product.json:40` | The repo's only machine-readable contract for this payload still lists `product_updates_data` as **required**, and the PR now emits payloads without it. Nothing failed only because the search specs validate against `purchases.json`, whose sole array property is `products` — the key `#index` returns. Search responses are keyed `purchases`, so **their items have never been validated at all** and the eleven existing `match_json_schema("api/mobile/purchases")` assertions under `GET search` were near-vacuous for the array. | Fixed in `ef00ad5` |
| 2 | P3 | `app/models/purchase.rb:2656`, `app/models/preorder.rb:186` | `json_data_for_mobile` drops the field in three separate branches. The fixture's purchase has a `url_redirect`, so the specs only reached that one — deleting the guard from the no-`url_redirect` branch **or** the preorder branch on its own left the entire suite green. | Fixed in `ef00ad5` |

Neither was a behavior bug: all three branches guard correctly as written. Both were mutation-coverage gaps, which is exactly the kind of thing that makes a later refactor silently regress this.

### How the fixes were verified

Rather than assert the specs bite, I proved it by reverting each hunk alone:

| hunk reverted alone | before `ef00ad5` | after `ef00ad5` |
|---|---|---|
| `#search` render-line opt-out | 2 failures | 2 failures |
| `preload_product_updates_data!` skip | 1 failure | 1 failure |
| `url_redirect.rb` guard | 2 failures | 2 failures |
| `purchase.rb` bare-link guard | **0 failures — green** | 1 failure (`:937`) |
| `preorder.rb` guard | **0 failures — green** | 1 failure (`:946`) |

The new schema assertion is non-vacuous too: with `#search` re-emitting the field, it fails on the schema matcher *before* reaching the `have_key` assertion — `#/purchases/0: failed schema #/properties/purchases/items: Matched "not" subschema`.

I also confirmed with a throwaway probe that both newly-covered branches are genuinely reachable from `#search` and not hypothetical: a preorder authorization purchase and a purchase whose `url_redirect` was destroyed both come back in the response, each correctly without the field.

186 examples, 0 failures across `purchases_controller`, `purchase_installments`, `preorder` and `url_redirects` specs. Rubocop clean.

These mattered more than the findings, so here is the evidence rather than a checkmark:

1. **No cache poisoning between the two settings.** This was the highest-risk class — if any cache key ignored the new flag, `#search` (field absent) and `#index` (field required) would share an entry and `#index` would start serving responses *missing* the field, breaking the retired native apps' offline posts cache. There is no `Rails.cache`, fragment cache or `cache_key` anywhere in the chain: `UrlRedirect#product_json_data`, `Preorder#mobile_json_data`, `Purchase#json_data_for_mobile` and `Link#as_json(mobile: true)` → `as_json_for_mobile_api` all build fresh hashes per call. The only memo is per-instance `@cached_product_updates_data`, which is neither read nor written when the flag is false, and no request mixes settings on one instance — `#search` is all-false, every other action all-true.
2. **Every call site threads correctly.** All five production callers audited: `#index`, `#purchase_attributes`/`#archive`/`#unarchive`/`#destroy`, `SalesController` (passes `{ include_sale_details: true }` — `options.fetch(:include_product_updates, true)` still returns true and the sale-details path is unaffected), `UrlRedirectsController#url_redirect_attributes`, `PreordersController#show`. All keyword args with `true` defaults, so no positional shift is possible. The flag never reaches `Link#as_json`, so it cannot leak into the response body.
3. **`false` is not a silent no-op for preorders.** Both `Preorder#mobile_json_data` branches honour it — the charge-successful branch forwards it, the authorization branch guards the assignment — so the perf claim holds on that lane too, rather than passing specs while quietly still doing the work.
4. **Skipping the preloader breaks no other field.** It does three things beyond computing posts, and each was traced: the `subscription: :original_purchase` preload isn't read by any remaining field (the review field goes through `true_original_purchase.product_review`, which the controller's own preloader still warms); `cached_alive_product_files` is only read while serializing installments, which no longer happens here; and the lazily-created installment `UrlRedirect` rows are read by nothing else in this response — `preload_latest_media_locations!` explicitly filters to rows with `installment_id` nil. Post opens are unaffected because `InstallmentsController#show` keeps its own create-if-missing.
5. **No authorization widening.** `search_purchases` is untouched and the preloader never filtered the purchase set — it only annotated instances. The entitlement filtering applied to posts *within* the field, which is now absent entirely: strictly narrower disclosure.
6. **No conditional-GET hazard.** No `fresh_when`/`stale?`/`etag` anywhere under `app/controllers/api/mobile/`. The only ETag source is `Rack::ETag`, which digests the body per response, so an old client's `If-None-Match` can't 304 against a body that still has the field.

### Settled after the review

The reviewer flagged the headline "no shipped client reads this from `#search`" claim as not verifiable from this repo. I checked it in `gumroad-mobile` at `74c3848`: `product_updates_data` appears exactly twice, both in `components/library/use-purchases.ts` — the optional type field and the `usePost` hook's `.find(...)`. `usePost` has no callers, and it reads from `usePurchase`, which returns `detailQuery.data?.product` — the `url_redirects/get_url_redirect_attributes` response, an endpoint this PR does not touch. Search results only ever reach it as `placeholderData`, and the access is optional-chained (`purchase?.product_updates_data?.find`), so even that path cannot throw on the field's absence.

### Only verifiable in production

- The 82%-of-request attribution and the p95/p99 improvement. The pre-deploy baseline is on gumroad-private#1412; post-deploy numbers go in a follow-up comment there once there are 48 hours of samples.
- Behaviour of the two retired native app builds. They read the field from `#index` and `url_redirect_attributes`, both untouched, and item 1 above rules out the cache-collision route by which `#index` could have lost it.

I have **not** armed auto-merge. This changes a mobile API response shape, so the merge call is a human's even though the client audit is now conclusive on both repos.

---

Reviewed by `claude-opus-5` running as Gumclaw, dispatched as an independent adversarial reviewer against a detached worktree at the PR head with a change-class-specific hunt list (cache-key collision, kwarg threading at every call site, preloader side effects on other fields, installed-client blast radius, spec vacuity, shared-serializer scope, conditional-GET, authorization scope). Findings were then verified against the repo and fixed by me.

gumclaw:
@nyomanjyotisa requesting your review — specifically on the **client contract call**, not the Rails side.

**Green at current head `ef00ad5`:** 84 checks, 0 pending, 0 failing. Pre-merge adversarial review above came back READY-TO-MERGE with no P1s and no P2s; both P3s are fixed and each of the five hunks now fails at least one example when reverted alone.

**The one judgment call I want a human on.** This removes `product_updates_data` from the `GET /api/mobile/purchases/search` response on the evidence that no current client reads it. My audit of `gumroad-mobile` at `74c3848` found the key exactly twice, and the only reader is the orphaned `usePost` hook — which reads from the *detail* endpoint and uses optional chaining, so absence can't throw even on that path. `#index` and the detail endpoints still return the field, so the offline posts cache is untouched.

You know the mobile client better than my grep does. If there's a shipped app version reading that key off the search response that isn't visible in the current tree, this is the change that would break it.

**I have deliberately not armed auto-merge.** It's a live API response-shape change, so the merge click is yours. If you're happy, merge it — or say the word and I'll arm auto-merge.

One thing worth flagging beyond this PR: every `GET search` spec in that file has been validating items against `api/mobile/purchases`, whose only array property is `products` (what `#index` returns) while search responses are keyed `purchases`. So eleven `match_json_schema` assertions there were asserting nothing about response items. I fixed it for this endpoint with a real search-response schema; the same trap exists anywhere a schema's properties don't match an endpoint's response key. Happy to sweep that separately if you want it.

PR #6402, opened 2026-07-27
Wait for the email editor debounce before publishing in the duplicate spec

gumclaw:
## Why

`spec/requests/emails/create_spec.rb:851` ("allows duplicating and editing an already published email") went red on `main` at f85ed3cf2 ([run 30262476385](https://github.com/antiwork/gumroad/actions/runs/30262476385), Test Slow 46), failing all three rspec-retry attempts with:

```
Failure/Error: expect(page).to have_alert(text: "Email successfully sent!")
  expected to find visible alert with text "Email successfully sent!" but there were no matches.
  Also found "Please include a message as part of the update.", which matched the selector but not all filters.
```

## Root cause

Not the attachment upload, and not a flake — it is a race against a debounce.

`EmailForm.tsx` pushes editor content into the Inertia form state through a 500ms debounce:

```ts
const handleMessageChange = useDebouncedCallback((newMessage: string) => {
  form.setData("installment.message", newMessage);
}, 500);
```

`buildPayload` reads `form.data.installment.message`. The spec typed the body and then clicked Publish with no wait in between, so the debounce had not fired yet and the request went out with `message => ""`, which trips the "Please include a message as part of the update." validation. The CI request log confirms `"message" => ""` on all three attempts, while the editor DOM still showed `<p>Hello, world!</p>` — the text was never lost, it just had not been flushed into form state.

Evidence that the attachment is not involved: moving `upload_attachment` to run *before* the text is typed still fails identically, and `upload_attachment` completes in ~0.3s locally so it does not itself provide the needed delay.

Every other publish/save site in this file already has a `sleep 0.5`/`sleep 1 # wait for the message editor to update`. Line 862 was the only `set_rich_text_editor_input` followed by an immediate publish with no such wait.

## Verification

Run locally against MySQL/Redis/Elasticsearch/MongoDB/minio:

- Fails on unmodified `main` in isolation — same "Please include a message" alert. Deterministic, not a flake.
- Passes with this change: `1 example, 0 failures`.
- Full file: `17 examples, 1 failure` — the one failure is `:453` "does not upload unsupported file as a subtitle", which **also fails on pristine `main` with this change stashed**, so it is pre-existing and out of scope here.
- `rubocop` clean on the file.

## Note

A CI re-run of the original job has since gone green, so this was a narrow race rather than a hard block — but the race is real and reproduces on demand locally, so the missing wait is worth closing rather than leaving to chance.

There is also a theoretical product-side micro-race: publishing within 500ms of the last keystroke drops unflushed edits. A human cannot realistically type, attach and publish that fast, so this PR only fixes the spec. Hardening `save()`/`buildPayload` to use the existing `getCurrentMessage()` or to flush the debounce would be a separate, deliberate change.

---

## AI disclosure

Instructions given: a `Test Slow` job on `main` had gone red and needed diagnosing rather than retrying. The model was asked to determine from the CI logs whether the failure was a flake or a real race before changing anything, to name the specific mechanism and the `file:line` responsible, to rule out the attachment upload as a cause by experiment rather than by argument, and to reproduce the failure deterministically against unmodified `main` locally before applying a fix.

### DIFF
--- a/spec/requests/emails/create_spec.rb
+++ b/spec/requests/emails/create_spec.rb
@@ -860,6 +860,10 @@ describe("Email Creation Flow", :js, type: :system) do
     uncheck "Allow comments"
     fill_in "Title", with: "Hello"
     set_rich_text_editor_input(find("[aria-label='Email message']"), to_text: "Hello, world!")
+    # The editor pushes its content into the form state through a 500ms debounce
+    # (see handleMessageChange in EmailForm.tsx), so publishing sooner than that
+    # submits an empty message and trips the "Please include a message" validation.
+    sleep 1 # wait for the message editor to update
     # Allows attaching files to the email
     upload_attachment("thing.mov")
     select_disclosure "Publish" do


gumclaw:
@nyomanjyotisa routing this one to you — green at head `c24cd50e4` (77 checks success, 8 skipped, zero failures), `MERGEABLE`, and everything on my side is finished.

**Why you:** you are the most recent non-Gumclaw author on `spec/requests/emails/create_spec.rb` (2026-02-09), so this is your file. It's also a four-line, test-only diff, so it should be a fast read.

**What it is in one line:** `spec/requests/emails/create_spec.rb:851` ("allows duplicating and editing an already published email") went red on `main` and I added the missing wait for the editor's debounce before the spec clicks Publish.

**The one thing worth your judgement — is this the right layer to fix it at?**

The mechanism is not a flake. `EmailForm.tsx` pushes editor content into Inertia form state through a **500 ms debounce**, and `buildPayload` reads `form.data.installment.message`. The spec typed the body and clicked Publish with no wait in between, so the request went out with `message => ""`, which trips the "Please include a message as part of the update." validation. The CI request log confirms `"message" => ""` on all three rspec-retry attempts while the editor DOM still showed `<p>Hello, world!</p>` — the text was never lost, it just hadn't been flushed into form state yet.

I ruled the attachment out by experiment rather than assumption: moving `upload_attachment` to run *before* the text is typed fails identically, and it completes in ~0.3 s locally so it wasn't supplying the needed delay by accident either.

So the fix matches what **every other publish/save site in this same file already does** — `sleep 0.5` / `sleep 1 # wait for the message editor to update`. Line 862 was the only `set_rich_text_editor_input` followed by an immediate publish with no such wait. Consistency argues for the sleep.

**But there is a real product-side race underneath it**, and I deliberately did not fix that here: publishing within 500 ms of the last keystroke drops unflushed edits *for real users too*. A human can't realistically type, attach a file and hit Publish that fast, which is why I judged this spec-only and left the app alone. Hardening `save()`/`buildPayload` to read `getCurrentMessage()` or to flush the debounce before building the payload would be the actual fix, as a separate deliberate change. **If you'd rather I do that instead of adding another sleep, say so and I'll close this and do it** — I can see the argument that the file already has six sleeps and a seventh is papering over a product bug.

**Verification:**

- Fails on unmodified `main` in isolation with the same "Please include a message" alert — deterministic, reproduces on demand, not a flake.
- Passes with this change: `1 example, 0 failures`.
- Full file: `17 examples, 1 failure` — that one failure (`:453`, "does not upload unsupported file as a subtitle") **also fails on pristine `main` with this change stashed**, so it is pre-existing and out of scope. Flagging it rather than folding it in.
- `rubocop` clean.

**Not urgent, and I want to be straight about that:** a CI re-run of the original job has since gone green on its own, so `main` is not blocked. This was a narrow race rather than a hard stop. The race is real and reproduces locally though, so it's worth closing rather than leaving to chance on the next unlucky run.

I've added `awaiting-human` and assigned you. Ignore or reassign freely if this isn't yours.

PR #6403, opened 2026-07-27
Tell self-gifting sellers what happened instead of citing test purchases

gumclaw:
## What

Rewords the error a seller sees when they try to gift their own product. It said:

> Test gift purchases have not been enabled yet.

It now says:

> You can't gift your own product. To give it away for free, create a 100% off discount code under Checkout > Discounts and share the checkout link.

The guard itself is unchanged — a seller buying their own product is only ever a test purchase, and test *gifts* were never built, so the case still has to be rejected. Only the message the seller reads changes, plus a comment on the guard explaining why the wording matters.

## Why

The old message described an internal, never-shipped capability instead of the seller's actual situation, and it named a mechanism ("have not been enabled yet") that sounds like an account flag. Sellers read it exactly that way: they conclude gifting is switched off for them and open a ticket asking us to enable it, which support then has to explain away. The condition has nothing to do with test purchases from the seller's point of view — they simply own the product they were trying to gift.

Root cause of the misleading text, in `app/services/purchase/create_service.rb#create_gift`:

```ruby
raise Purchase::PurchaseInvalid, "Test gift purchases have not been enabled yet." if buyer == product.user
```

The string was written to describe the internal reason the branch exists (test gifting was never implemented) rather than what the person on the other end did. The two sibling guards immediately below it in the same method are already written the right way, naming the problem in the seller's terms:

```ruby
raise Purchase::PurchaseInvalid, "You cannot gift a product to yourself. Please try gifting to another email." if giftee_email == purchase_params[:email]
raise Purchase::PurchaseInvalid, "Gift purchases cannot be on installment plans." if params[:pay_in_installments]
```

So this change makes the first guard consistent with its own neighbours.

**Why point at a 100% off discount code.** That is what support already tells these sellers to do, and it is the only supported way to hand someone a product for free from your own account. `Checkout > Discounts` was verified as the live path (`config/routes.rb` mounts `resources :discounts` inside the `checkout` namespace, and the dashboard section is titled "Discounts"), so the instruction sends the seller somewhere that exists rather than to a generic help page.

**Alternative considered and rejected: making self-gifting work.** Letting a seller gift their own product would mean building test gifts — a gift record, a giftee notification, and an entitlement, all for a purchase that never charges. That is a real feature with a real surface, not a copy fix, and the seller's underlying goal (give this person the product for free) is already served by a discount code. Rejecting the action remains correct; only the explanation was wrong.

**Scope.** The message is the only behaviour change. No gate, no eligibility rule, and no money path is touched, and the guard fires on exactly the same condition as before.

## Before / After

No rendered layout changes — this is a single service-layer error string, surfaced through the existing checkout error banner that already displays the two sibling messages quoted above. There is no new UI element, no styling change, and no change to when the banner appears.

| | Message shown to a seller gifting their own product |
|---|---|
| **Before** | `Test gift purchases have not been enabled yet.` |
| **After** | `You can't gift your own product. To give it away for free, create a 100% off discount code under Checkout > Discounts and share the checkout link.` |

A walkthrough video of the guard and its specs stands in for a UI recording, since the change is a string returned by `Purchase::CreateService` rather than a visual surface.

**Walkthrough video** (the guard on `main`, the two sibling guards it was inconsistent with, the diff, a repo-wide grep proving the old string is gone, the specs, and the specs running green):

https://github.com/antiwork/gumroad/blob/7e8555625f4a27dacdb46af928ab7733c502816a/qa-media/pr-6403-self-gift-message-walkthrough.mp4?raw=true

## Test Results

All three specs for this case are green, and the two that pin the message text were proven load-bearing: reverting `create_service.rb` to `main`'s version makes both fail, so they pin the new message rather than passing against anything.

spec results

## QA steps

No preview app interaction is needed to verify this — the change is a returned string, and the two specs above assert it directly. To exercise it end to end on the preview:

1. Sign in as a seller who owns at least one published product.
2. Open that product's checkout page while signed in as its own seller, and choose to gift it to another email address.
3. Submit. The error banner should read "You can't gift your own product. To give it away for free, create a 100% off discount code under Checkout > Discounts and share the checkout link." — it must not mention test purchases or anything being "enabled".
4. Follow the instruction the message gives: `Checkout > Discounts` should exist and let you create a 100% off code, confirming the message points somewhere real.

---

AI disclosure: written by Claude Opus 5 (`claude-opus-5`) running as the Hermes agent. Instructions given: fix the reported bug where a seller gifting their own product is told test gift purchases are not enabled. The agent read the guard and its two sibling guards, confirmed against production that the reporting seller had gifting fully enabled and was blocked only by product ownership, verified the `Checkout > Discounts` path exists in the routes and dashboard, chose rewording over implementing test gifts, wrote the two specs, and proved them load-bearing by reverting the changed file and confirming both fail.


gumclaw:

**What the adversarial pre-merge review caught, and it was worth catching.** Purchase error messages reach the buyer through `Receipt.tsx`, which renders them with `dangerouslySetInnerHTML` — so this string is interpreted as markup, not text. The wording I shipped contains `Checkout > Discounts`, and a bare `>` is inert, so the message is safe as written. But nothing in the code recorded that constraint, and the obvious next edit to a message like this is to wrap the path in something. Fixed in `44d562b59` by pinning it with a spec and a comment naming the render path, so a future rewording cannot silently start emitting markup. Worth being precise: this was not a live bug, it was an undocumented invariant one edit away from becoming one.

**The specs are load-bearing, proven not asserted.** Reverting `create_service.rb` to `main`'s version makes both message assertions fail (`Expected "Test gift purchases have not been enabled yet." to eq …` and `Expected … not to match /test/i`). That mutation run is the second half of the evidence image, so the green count and the red proof are in one artifact. The tree was restored before committing and `git diff --stat` confirmed only the two intended files.

**The judgement call, and the alternative I rejected.** The other reading of this ticket is "make self-gifting work." I did not, and I want a reviewer to overrule me cheaply if that is wrong. Letting a seller gift their own product means building test gifts — a gift record, a giftee notification, and an entitlement for a purchase that never charges. That is a feature, not a copy fix, and the seller's actual goal (hand this person the product for free) is already served by a 100% off discount code, which is what support tells them today. So the rejection stays correct; only the explanation was wrong.

**Why the message points where it does.** I checked that `Checkout > Discounts` exists rather than trusting the phrasing: `config/routes.rb` mounts `resources :discounts` inside the `checkout` namespace, and the dashboard section is titled "Discounts" in `CheckoutDashboard/FormPage.tsx`. A message that instructs a seller to go somewhere is only useful if the place is real.

**Scope, stated plainly.** One error string, one comment, three specs. The guard fires on exactly the same condition (`buyer == product.user`), so no gate, eligibility rule, or money path moves. A repo-wide grep confirms the old string has zero remaining references in `app`, `spec`, `test`, `lib`, and `config` — nothing keyed on the text, in either the Ruby or the TypeScript side, so there is no consumer to update. That grep is in the walkthrough video.

**On the Before/After.** There is no pixel diff to show: this is a service-layer string surfaced through the existing checkout error banner that already carries the two sibling messages. Rather than claim "not applicable," the body has the two messages side by side and the walkthrough video covers the guard, the inconsistency with its neighbours, the diff, the grep, and the specs running green.

PR #6404, opened 2026-07-27
Fingerprint deletion-relevant product state, not the product's updated_at

gumclaw:
# Fingerprint deletion-relevant product state, not the product's `updated_at`

Follow-up to #6377 (merged as `1984ae1f1`). One review point from that PR was not addressed before it merged; this is that fix on its own.

The save contract itself is unaffected and still ships behind `product_editor_save_contract`, **off**.

## What

`Product::EditorRevision`'s fingerprint included `product.updated_at`. Replaced with a short allowlist of the product attributes that actually change *what a save is allowed to delete*:

```ruby
DELETION_RELEVANT_ATTRIBUTES = %w[
  has_same_rich_content_for_all_variants
  is_tiered_membership
  is_physical
  skus_enabled
].freeze
```

- **`has_same_rich_content_for_all_variants`** — when on, the product has one shared set of content pages instead of a set per version, and the editor is not shown the per-version pages. A real product lost its content this way three times in nine days (July 2026): the stored state left the editor with no version content to submit, so an ordinary save looked like "the seller deleted every version page." `Product::RichContentDeletionGuard` reads the persisted value for the same reason.
- **`is_tiered_membership`** — memberships and ordinary products delete versions by different routes, and only the membership route also sweeps the version's recurring prices.
- **`is_physical`** — physical products always resolve to the product-level pages and ignore per-version ones (`Link#has_product_level_rich_content?` short-circuits on it; the editor's presenter branches the same way). Flipping it changes which pages the editor loads, and so which pages a deletion built from that load would remove.
**Deliberately excluded, with reasons:**

- **`skus_enabled` and SKU rows.** I had these in an earlier revision of this PR and a reviewer was right to push back. `Product::SkusUpdaterService` is the only code that deletes SKU rows, `VariantsUpdaterService` is its only caller and guards it with `if skus_enabled` — and `LinksController#update` assigns `skus_enabled = false` (line 491) *before* `update_variants` runs, on every editor save. So this endpoint can never delete a SKU, whatever the column said when the page loaded. Fingerprinting it invalidated tokens over state no deletion here can act on, which is the same false-conflict mistake as the `updated_at` it replaced, in narrower form. I had reasoned from the service in isolation without checking whether the controller could ever present the flag as true.
- **`native_type`** — steers presentation and call availability, but no deletion route branches on it.

These are `flags` bits (flag_shih_tzu) rather than real columns, so the existence check is on the reader and **raises** if one is renamed away. Silently dropping an attribute would quietly widen what a stale token is willing to delete, which is the exact failure mode the class exists to prevent.

`VERSION` → `v4`, so outstanding tokens cannot be read as current against the new definition.

## Why

The class documents itself as *deliberately not a timestamp*, so an unrelated write could not invalidate a seller's in-flight deletion and reproduce the false conflicts that got the earlier optimistic-concurrency attempt (#6245) switched off:

> Deliberately NOT a timestamp — a digest changes only when something the contract cares about changes, so an unrelated write (a price edit, a tag) does not invalidate a seller's in-flight deletion…

The fingerprint then included `product.updated_at`, which **every** product write bumps. The code contradicted its own comment, and #6377's description repeated the comment's claim rather than the code's behaviour.

Consequence on merged `main`: a co-editor renaming a product, retagging it, or editing its description invalidates another tab's outstanding deletion token, and that tab's next deletion is refused with `409 stale_deletion_conflict` — telling the seller their page is out of date when nothing they could delete has changed. That is the #6245 false-conflict behaviour, reintroduced through the digest.

This is latent rather than live: with the flag off, `may_delete?` is never consulted. It would have surfaced on the first cohort enabled.

### Walkthrough

https://raw.githubusercontent.com/antiwork/gumroad/9065e159e2e37c6866d3a00592dff1a8b34dec21/qa-media/pr-6404-revision-token-scope.mp4

Driven through the real product editor against a local app. It reads the token the server put into the editor's props, renames the product and saves **through the UI**, and shows the token did not move — then a second actor flips `is_physical` out of band and the token does move. Both values in the overlay match what the Rails console reports for the same product, and the rename was confirmed in the database afterwards (`updated_at` bumped, so the step is not vacuous).

Observed, in order:

```
TOKEN_ON_LOAD=1ae9d6ab9259d2ed9e5571a24dcae076
TOKEN_AFTER_RENAME=1ae9d6ab9259d2ed9e5571a24dcae076   RENAME_KEPT_TOKEN=true
TOKEN_AFTER_IS_PHYSICAL=45c91a01c2ca1404a61838432cba004f   MOVED=true
```

This change is server-side, so there is no visual UI diff and no mobile surface to capture — the editor renders identically either way. What is worth seeing is the token behaviour, which is what the recording shows.

### Measured before/after

Measured against merged `main` and against this branch. The two halves of the contract are **over-invalidation** (rows 1–5, which caused false conflicts) and **under-invalidation** (rows 6–11, where a stale token wrongly looked fresh):

| Change | On merged `main` | With this PR | Should be |
|---|---|---|---|
| `price_cents` +100 | unchanged | unchanged | unchanged |
| `touch` | **invalidated** | unchanged | unchanged |
| name edit | **invalidated** | unchanged | unchanged |
| tag added | **invalidated** | unchanged | unchanged |
| description edit | **invalidated** | unchanged | unchanged |
| `has_same_rich_content_for_all_variants` flipped | invalidated | invalidated | invalidated |
| `is_tiered_membership` flipped | invalidated | invalidated | invalidated |
| `is_physical` flipped | **unchanged** | invalidated | invalidated |
| `skus_enabled` flipped | unchanged | unchanged | unchanged — this save can't delete SKUs |
| SKU added or deleted | unchanged | unchanged | unchanged — same reason |
| a version deleted | invalidated | invalidated | invalidated |

Row 8 is the gap found after the `is_physical` report; rows 1–5 are the original over-invalidation; rows 9–10 are state the token deliberately ignores. With this change: **23 examples, 0 failures**.

Run the new spec against merged `main`'s `editor_revision.rb` and 6 of the 25 fail:

```
keeps the token when the product's own updated_at moves for an unrelated reason
keeps the token when the seller edits the price
keeps the token when the seller edits tags, name or description
keeps the token when skus_enabled flips or a SKU changes
raises rather than silently ignoring a deletion-relevant attribute that no longer exists
```

Worth being precise about what that list does **not** contain: the `is_physical` regression *passes* against `main` — but for the wrong reason. `update_attribute` bumps `updated_at`, and `main` still has `updated_at` in the digest, so the token moves incidentally rather than because the attribute is witnessed. Row 8 of the table was measured directly (flip the flag, compare digests) rather than inferred from that spec run, which is how the gap was visible at all. Same class of trap as the price regression described below: a spec can go green off a mechanism it isn't testing.

**Reachability is pinned, not asserted in a comment.** `test/controllers/links_controller_test.rb` now has *"the editor save cannot delete a SKU, even with skus_enabled persisted true"*: it persists `skus_enabled = true`, creates a live SKU, drives the real endpoint, and asserts the flag reads `false` inside `VariantsUpdaterService`, that `SkusUpdaterService` is never constructed, and that the SKU survives. Deleting the `skus_enabled = false` line fails it with *"the save must zero skus_enabled before VariantsUpdaterService runs"*. If SKUs ever become reachable from this save, that test goes red and says the token needs to cover them again.

## Test Results

test results

| suite | result |
|---|---|
| `spec/services/product/editor_revision_spec.rb` | 23 examples, 0 failures |
| `spec/services/product/` (whole directory) | 194 examples, 0 failures |
| `test/controllers/links_controller_test.rb` | 431 runs, 1602 assertions, 0 failures |
| `rubocop` on both changed files | no offenses |
| `bin/test-confidence` | **99% confidence, 9 tests passed in 81s — "Safe to commit."** |

**Mutation testing** — each mutation turns the suite red, so the new tests are load-bearing rather than merely green:

| mutation | result |
|---|---|
| put `updated_at` back into the digest | 4 failures (touch, price, tags/name/description, SKU-invariance) |
| empty `DELETION_RELEVANT_ATTRIBUTES` | 2 failures (shared-content, tiered-membership) |
| drop `is_physical` from the allowlist | 1 failure |
| delete `skus_enabled = false` from `LinksController#update` | 1 failure — the reachability test |

### A trap worth recording

The first version of the price regression **passed against the unfixed code** — it asserted nothing. `price_cents` is delegated to the default price row, so `update!(price_cents:)` persists the new price but leaves the product itself clean and never moves `updated_at`:

```
PRICE_CENTS=100
CHANGED_BEFORE_SAVE={}   # product not dirty
SAVED_CHANGES=[]
PRICE_AFTER=600          # but the price did change
```

It now writes a real `Link` column alongside and asserts `updated_at` actually moved *before* checking the token held. Caught by mutation testing, not by reading it.

## Rollout

No flag, no migration, no data change. Behaviour is identical while `product_editor_save_contract` is off; it corrects what happens once a cohort is enabled.

Worth landing before that first cohort, since the practical effect of shipping without it is false `409`s on ordinary concurrent edits.

## Canary note for the save contract

Carrying this forward from #6377 so it isn't lost: **canary all five collections before broader enablement.** #6359's audit trail observes variants and categories, and the 28/28 reconciliation covers those — but it does **not** observe pages, files, public files, or integrations, so those four have no retrospective safety net and need watching directly during the first cohort.

---

## AI disclosure

Written by Hermes Agent (Nous Research) running `claude-opus-5`.


gumclaw:
Self-review of my own diff.

**`VERSION` bump invalidates every outstanding token — deliberate, worth a reviewer's attention.** Anyone with the editor open when this deploys has their token rejected on the next *deletion* (ordinary saves are unaffected — the token only gates deletions). They reload and continue. With the save-contract flag off, `may_delete?` is never consulted, so in practice this is inert until a cohort is enabled. I chose that over leaving old tokens readable against the new definition, where a `v3` token would be compared to a `v4` fingerprint and silently fail closed anyway.

**`sku_stamps` adds a query per fingerprint call.** `Sku.alive.where(link_id:)` is indexed on `link_id` and the fingerprint is computed twice per save (once for the presenter's response, once for the freshness check). I judged the extra indexed query acceptable next to the alternative — a digest blind to SKU deletion — but flagging it since it is a real cost on a hot path, and I did not benchmark it.

**The allowlist is a judgment call and could still be incomplete.** I audited all 33 `Link` flags by grepping the deletion routes (`app/services/product/`, `links_controller`, `product_presenter`, `rich_contents`) and reasoned about each. Four are included; `native_type` was consciously excluded because it steers presentation and call availability but no deletion route branches on it. If a reviewer knows of a field that changes which rows a save removes, it belongs here — the `is_physical` gap in the previous revision came from exactly that kind of miss, not from a mechanical error.

**`product.respond_to?(attribute)` is a weaker guard than a column check.** `respond_to?` would still pass if a flag were renamed but a same-named method existed elsewhere on the model. A `column_names` check is impossible here because these are flag_shih_tzu bits and absent from `column_names` — that was my first attempt and it raised on every valid attribute. The spec pins the raise behaviour, so the guard is at least exercised.

**Test-shape note reviewers may want to check.** The `is_physical` and `skus_enabled` regressions use `update_attribute`, which bumps `updated_at`. On merged `main` — where `updated_at` was still in the digest — those two specs pass *incidentally*, off a mechanism they are not testing. They only become load-bearing once the timestamp is out. I verified them by mutation instead (drop the attribute from the allowlist → red) rather than trusting the red/green signal. Same trap as the price regression documented in the description.

**Deploy-window behaviour — measured, not assumed.** A token minted under `v3` and presented once `VERSION` is `v4`:

```
V3_TOKEN=f38ce7497dffb2507672d00222b46552
V4_TOKEN=8924bbf082cd264118f4d7bb0b8ca548
V3_FRESH_UNDER_V4=false
```

It fails closed — the deletion is refused and a fresh token is returned, so the seller reloads and retries. That is the safe direction, but note there is no spec pinning it; I verified it with a throwaway probe rather than committing a test that manipulates a frozen constant.

gumclaw:
You were right, and my justification was wrong. Fixed in `c25a21584`.

I had reasoned from `Product::SkusUpdaterService` in isolation — "`skus_enabled` gates a service that deletes SKU rows, therefore it's deletion-relevant" — without checking whether the controller could ever present that flag as true. It can't: `LinksController#update:491` assigns `skus_enabled = false` before `update_variants` runs, on every editor save. `VariantsUpdaterService` is the service's only caller and guards it with `if skus_enabled`, so this endpoint cannot delete a SKU regardless of what the column held when the page loaded.

That made the SKU scope the same mistake as the `updated_at` this PR removes, in narrower form: invalidating tokens over state no deletion on this path can act on.

Verified by driving the real endpoint rather than reading it again. From the most favourable possible starting state — `skus_enabled` persisted **true**, a live SKU, a version in the payload:

```
PERSISTED_skus_enabled=true
SKU_ALIVE_BEFORE=true
RESPONSE_STATUS=200
OBSERVED_skus_enabled_at_updater=[false]
SKUS_UPDATER_CONSTRUCTED=false
SKU_ALIVE_AFTER=true
```

**Removed:** `skus_enabled` from `DELETION_RELEVANT_ATTRIBUTES`, plus the `skus:` fingerprint entry and `sku_stamps`.

**Pinned rather than commented.** The reachability argument is now a controller test — *"the editor save cannot delete a SKU, even with skus_enabled persisted true"* — asserting the flag reads false inside the updater, that `SkusUpdaterService` is never constructed, and that the SKU survives. Deleting the `skus_enabled = false` line fails it with "the save must zero skus_enabled before VariantsUpdaterService runs". If SKUs ever become reachable from this save it goes red and says the token needs to cover them again. The spec asserts the inverse: flipping `skus_enabled`, deleting a SKU and adding one all leave the token valid.

`is_physical` stays — it decides whether product-level or per-version content pages are authoritative, and this save does delete pages.

**Local:** 23 spec examples, 194 in `spec/services/product/`, 431 controller runs / 1602 assertions, rubocop clean. Mutations still load-bearing: restoring `updated_at` → 4 failures, dropping `is_physical` → 1.

CI is rerunning on `c25a21584`. For the record the superseded head `3d359bbb4` had settled green (77 pass / 8 skipped / 0 failures), but that result no longer applies to this code. Preview QA continues in parallel; the PR stays draft and the save-contract flag stays off.

gumclaw:
## Exact-head CI green + preview QA on the corrected code

**CI on `c25a21584`: 77 pass, 8 skipped, zero failures.** That is the head with the SKU scope removed — the earlier green on `3d359bbb4` was the pre-fix code and does not apply.

### Preview QA

App: `https://fix-editor-revision-deletion-rel.apps.staging.gumroad.org` — verified serving `x-revision: c25a21584c47` on 5 consecutive probes. (The hostname is truncated at 32 chars; it came from the Buildkite deploy log rather than the branch name.)

Every link curl-verified before posting:

| URL | HTTP |
|---|---|
| `/` | 200 |
| `/l/demo` | 301 → 200 (seller subdomain) |
| `/products/demo/edit` | 302 → 200 (auth) |
| `/login` | 200 |

Product page renders (title, both versions Alpha/Beta, price) — confirmed in a browser, not just by status code.

Run against the deployed app, with the contract flag on for the seeded seller:

```
SELLER=seller@gumroad.com   PERMALINK=demo   FLAG_ON=true
VERSIONS=["Alpha", "Beta"]
REVISION_VERSION=v4
WITNESSED=["has_same_rich_content_for_all_variants", "is_tiered_membership", "is_physical"]
TOKEN_INITIAL=5318d5582e24a0e3fecd9ed424f7752d
RENAME_KEPT_TOKEN=true
IS_PHYSICAL_MOVED_TOKEN=true
RESTORED_TOKEN_MATCHES=true
```

`WITNESSED` is the corrected three-attribute list — `skus_enabled` is gone, confirmed on the running app rather than only in the diff. The two halves of the contract both hold there: an unrelated rename leaves the token valid, and `is_physical` invalidates it.

### Manual QA steps for a reviewer

1. Log in as `seller@gumroad.com` and open `/products/demo/edit`.
2. Rename the product, save. Then delete a version — it must succeed. (Before this PR the rename invalidated the token and the deletion returned `409 stale_deletion_conflict`.)
3. In a second tab, flip `is_physical` (or toggle "use the same content for all versions"). Back in the first tab, without reloading, delete a version — expect **409** and a reload prompt, with nothing deleted.
4. Turn the flag off and repeat step 2 — behaviour should match `main` exactly.

Note for whoever picks this up: preview console hops can OOM-kill puma while someone is driving the browser, so I did all seeding in a single hop before this handoff and won't touch the console during a manual pass.

Flag stays off. Ready for review once a human has looked at it — I have not marked it ready.

PR #6406, opened 2026-07-27
Update affiliate docs: the Gumroad Affiliate Program opt-out now reaches membership renewals

gumclaw:
## What

Two help center articles said Gumroad Affiliate Program commission on a membership could never be stopped once the referral existed. That is no longer true after #6398, which makes the seller's opt-out apply to renewals of memberships that were referred before the opt-out.

Changed:

- `_333-affiliates-on-gumroad` (creator-facing) — the "Opting out of Gumroad Affiliates" note now says the opt-out also stops commission on future charges of memberships referred beforehand, and that it does not touch affiliates the seller added directly. The "Memberships" section is rewritten: recurring commission now depends on the affiliate still being eligible, the "no way to change this behavior" claim is scoped to a seller's own affiliates, the Gumroad Affiliates opt-out is described separately, and the commission-rate sentence is corrected (see Why).
- `_249-affiliate-faq` (affiliate-facing) — the "Membership products" section gains one paragraph telling affiliates that a creator opting out of the Gumroad Affiliate Program stops commission on charges from that point on, that already-earned commission is not taken back unless a charge is refunded, and that none of this applies when a creator added them as one of their own affiliates.
- One broken anchor fixed while in the file: `_333` linked to `249-affiliate-faq#Stickyaffiliatelinks`, but that section's id is `sticky`.

## Why

Triggered by merged PR #6398. `GlobalAffiliate#eligible_for_credit_on_renewal?` now requires `!product.user.disable_global_affiliate`, and `Subscription#build_purchase` calls it, so a renewal of a globally-referred membership carries no affiliate once the seller has opted out. The base `Affiliate#eligible_for_credit_on_renewal?` still just delegates to `eligible_for_credit?`, so direct affiliates and collaborators are unchanged — which is why every new sentence draws that distinction explicitly rather than making a blanket claim.

Both audiences needed the edit: the seller article to stop promising the opt-out does nothing here (the exact complaint that produced #6398), and the affiliate article so affiliates are not surprised by commission ending mid-subscription.

An adversarial review of the first draft turned up three further inaccuracies in the same sections, all of them pre-existing and all now fixed:

- "Affiliates ... will earn commission on all recurring revenue ... for as long as the customer they referred is subscribed" read as unconditional, but every renewal passes through `eligible_for_credit_on_renewal?`, which also fails for a suspended affiliate or a seller on a Brazilian Stripe Connect account. The sentence now carries that condition.
- "Even if you change your affiliate's rate, they will still earn the prior rate for pre-existing subscriptions" was backwards. `Purchase#affiliate_cut` calls `affiliate.basis_points(product_id:)`, which reads the affiliate record's current rate, so a rate change does apply to future renewals. Only charges already made keep the rate they were charged at.
- The affiliate-facing draft said "you keep everything you already earned", which the same article contradicts two sections later: a refund claws back the affiliate's share. Now scoped, with a link to that Refunds section.

Payout-adjacent wording, so the exact phrasing is worth a read. The core claim is that Gumroad Affiliates commission stops on charges from the opt-out onward and that previously earned commission is unaffected except by refunds. That matches the code — the gate is evaluated when each renewal purchase is built, and #6398 deliberately shipped no backfill of existing rows.

## Before / After

Not applicable in the usual sense — this is help center prose, not app UI. The rendered result of the three edited sections is below.

## QA steps

1. Open `help.gumroad.com/help/article/333-affiliates-on-gumroad`, jump to "Opting out of Gumroad Affiliates" and "Memberships".
2. Open `help.gumroad.com/help/article/249-affiliate-faq`, jump to "Membership products".
3. Confirm the in-page anchor links used in the new sentences (`#Your-Affiliates---How-does-it-work`, `#Gumroad-Affiliates`, `#Gumroad-Affiliate-Program-0awON`, `#refunds`) scroll to the right sections. All of these already existed; no anchor IDs were added, renamed, or removed.
4. Confirm the "sticky link" link under "Why was an affiliate not credited for a sale?" now lands on the "Sticky affiliate links" section of the affiliate FAQ instead of the top of the page.

## Test Results

- `bundle exec rspec spec/models/help_center` — 1 example, 0 failures (guards slug ↔ partial integrity), run again after the review fixes.
- Both partials rendered in a real view context via `ApplicationController.render`, asserting the edited section headings resolve — no missing helpers, no broken tags.
- `articles.yml` untouched: body-only edits, no title, description, slug, or category change.

Rendered sections:

Help center sections after the edit

---

## AI disclosure

Instructions given: after a PR merges in this repo, check whether it changes anything the help center documents and update the affected articles to match. The model read the merged diff for #6398, confirmed the new renewal gate and that it applies only to `GlobalAffiliate`, grepped the article bodies for the affected wording, found the stale claims, wrote the edits, rendered both partials plus the help-center model specs locally, ran an adversarial review of its own diff against the affiliate and purchase models and fixed the three overclaims it found, captured the rendered sections, and authored this description.


gumclaw:
## Adversarial pre-merge review

Ran a hostile review of the diff against `affiliate.rb`, `global_affiliate.rb`, `direct_affiliate.rb`, `subscription.rb`, and `purchase.rb`. Four findings, all fixed in 39ce876.

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | P2 | `_249`: "You keep everything you already earned" is contradicted two sections later by Refunds — a refund claws back the affiliate's share. | Rewritten to "Commission you already earned on earlier charges is not taken back, unless one of those charges is refunded", linking the Refunds section. |
| 2 | P2 | `_333` Memberships: "commission on all recurring revenue ... for as long as the customer they referred is subscribed" is unconditional, but every renewal is gated on `Affiliate#eligible_for_credit_on_renewal?`, which also fails for a suspended affiliate or a Brazilian Stripe Connect seller. | Condition added: "...for as long as the customer they referred is subscribed and the affiliate is still eligible for commission on your products." |
| 3 | P2 | `_333`: "even if you change your affiliate's rate, they will still earn the prior rate for pre-existing subscriptions" is backwards. `Purchase#affiliate_cut` → `affiliate.basis_points(product_id:)` reads the affiliate record's CURRENT rate, so future renewals use the new rate. | Corrected: charges already made keep the rate they were charged at; future renewals use the new rate. |
| 4 | P3 | `_333` linked `249-affiliate-faq#Stickyaffiliatelinks`; the actual id in `_249` is `sticky`, so the link dropped readers at the top of the page. | Anchor fixed. |

Findings 2, 3, and 4 were pre-existing rather than introduced here, but they sit in the sentences this PR is already rewriting, so they were swept in the same pass.

Re-verified after the fixes: `bundle exec rspec spec/models/help_center` green, both partials render, screenshot in the body refreshed.

gumclaw:
Addressed the one non-blocking finding from the earlier review: the QA screenshot was named without the `pr-<number>` prefix. Renamed to `qa-media/pr-6406-help-center-affiliate-optout-renewals.png` and updated the image link in the description (verified it returns a 200 image/png from the branch).

No other changes — the prose and anchor fixes are unchanged from the reviewed commit.

PR #6407, opened 2026-07-27
Docs: gifting your own product is not possible, use a 100% off discount code

gumclaw:
## What

Updates the gifting help article (`213-how-do-i-give-a-product-as-a-gift`) to say that you cannot gift a product you sell yourself, and to tell creators what to do instead: create a 100% off discount code under Checkout > Discounts and share the checkout link.

Two additions to the same article, no metadata change:

- A third bullet in the Restrictions list: "You cannot gift a product you sell yourself. Checkout will tell you so and point you at a 100% off discount code instead."
- A closing paragraph for creators explaining why (buying your own product is only ever a test purchase) and linking the discount-codes article.

## Why

Triggered by the merge of #6403, which reworded the error a seller sees when they try to gift their own product. It used to say "Test gift purchases have not been enabled yet." and now says "You can't gift your own product. To give it away for free, create a 100% off discount code under Checkout > Discounts and share the checkout link."

The help article described gifting unconditionally — the Restrictions section covered the cart limit and creator-disabled gifting, but not the self-gifting block. A creator reading it had no way to know the attempt would be rejected, and no pointer to the supported way to hand a product to someone for free. The in-app copy now names the alternative; the article should say the same thing rather than leaving the product to explain it alone at the moment of failure.

The guard being documented is in `app/services/purchase/create_service.rb#create_gift`: `raise Purchase::PurchaseInvalid, "You can't gift your own product. …" if buyer == product.user`. The discount-code path was verified against `config/routes.rb` (`resources :discounts` under the checkout namespace), which is where "Checkout > Discounts" comes from.

Deliberate scoping:

- Buyer-facing wording stays out of it. The block only ever fires for the product's own seller, so `191-a-guide-to-buying-on-gumroad` and `343-wishlists` (both of which describe gifting from a buyer's point of view) stay accurate and untouched.
- The existing "Note for creators" paragraph about the Checkout > Form gifting switch is left alone — that is a different setting, and merging the two would blur turning gifting off for customers with not being able to gift your own product.
- `app/models/help_center/articles.yml` is untouched: the title, description snippet and category are unchanged, and the description comes from the article's opening, which this PR does not touch.

## Before / After

The rendered article after the change, with the new Restrictions bullet and the closing creator paragraph:

Article 213 after

Before, the Restrictions list ended at the creator-disabled-gifting bullet and there was no mention of self-gifting anywhere in the article.

## QA steps

1. Deploy or run locally, visit `help.gumroad.com/help/article/213-how-do-i-give-a-product-as-a-gift`.
2. Scroll to Restrictions — the third bullet names self-gifting.
3. Below the Checkout > Form note, the new paragraph explains why and links to the discount-codes article; click the link and confirm it lands on `128-discount-codes`.

## Test Results

`bundle exec rspec spec/models/help_center` — 1 example, 0 failures. These guard the slug ↔ body-partial integrity that a help-center edit can break.

Help center specs

The article partial was also rendered through the real view context (`ApplicationController.render(partial: "help_center/articles/contents/213-how-do-i-give-a-product-as-a-gift")`) to confirm the new copy lands, the `128-discount-codes` link resolves, and no helper blows up — that render is what produced the screenshot above. Tag balance in the file was checked element by element and is even.

---

AI disclosure: written by claude-opus-5 running as the Hermes agent. It was triggered automatically by the merge of #6403 through the help-center doc-sync job, with standing instructions to check whether a merged product change leaves a help-center article describing the product wrong, and if so to edit the affected article, verify it renders, and open a PR. The wording was derived from the merged diff's error message and the routes file; no prompt text was pasted into the article.


gumclaw:
Pre-merge review of this docs change, run against the merged #6403 diff and the code it documents.

Checked and clean:

- **Copy matches the shipped guard.** `create_gift` raises on `buyer == product.user`, i.e. only for a logged-in seller buying their own product. "a product you sell yourself" is the right scope; it does not overstate the block as applying to anyone else.
- **The alternative is real and correctly named.** `Checkout > Discounts` matches `resources :discounts` in the checkout namespace, and it is what the shipped error message itself tells the seller. The article links `128-discount-codes` rather than restating discount mechanics, so there is one source of truth for how discount codes work.
- **No adjacent article left contradicting this.** `191-a-guide-to-buying-on-gumroad` and `343-wishlists` describe gifting from the buyer's side, where the block never fires, so both stay accurate. Grep of `gift` across all article bodies surfaced no other place claiming a seller can gift their own product.
- **Existing creator note not merged into the new one.** The Checkout > Form gifting switch is a different setting (turning gifting off for customers). Folding the two together would read as if the switch were what blocks self-gifting.
- **No metadata drift.** The article's opening paragraph is untouched, so the `articles.yml` `description` snippet stays in sync; title and category unchanged, so the yml is deliberately not modified.
- **Structure intact.** Anchor IDs preserved, tag balance even element by element, partial renders through the real view context with the new copy present and the internal link resolving.

Nothing rated P1 or P2. One P3 noted and deliberately not acted on: the article's two screenshots are older captures of the checkout gift toggle. The change does not touch that flow, so re-shooting them is not in scope here.

PR #6410, opened 2026-07-27
Only show buyers a local currency the checkout could actually settle in

gumclaw:
## What

Gates the buyer-local-currency price shown on product pages so it only appears in a currency the checkout could actually be settled in.

A product page previously converted the price on GeoIP alone. The charge path applies a much narrower set of rules, so there were cases where the page promised a local price the charge could never honour:

- the product is not priced in USD (the charge is made in the currency the seller listed, so a converted local price is never what is charged);
- Gumroad and Stripe disagree on the currency's minor units (HUF, for example, which Stripe only charges in amounts divisible by 100);
- the seller has no Stripe charge account, or one Checkout::BuyerCurrencyEligibility does not support;
- the seller's account settles that currency in itself rather than in USD, so it rejects the FX quote the charge needs.

In each of those the buyer saw, say, EUR on the product page and was then charged the canonical USD amount at the total, with nothing explaining why the number changed.

## Why

Decision 1 on antiwork/gumroad-private#1324, ruled by Sahil on 2026-07-27:

> Only show currencies we would allow settling in. For now IP is fine but eventually customer choice is key.

This PR is the "only show currencies we would allow settling in" half, with GeoIP kept as the selection input. A buyer-facing currency picker is the recorded future direction and is deliberately not in this slice.

## How

`CurrencyHelper#buyer_currency_display_props` now calls a new `buyer_currency_settleable?` before converting. That check is deliberately the **seller- and currency-level** half of `Checkout::BuyerCurrencyEligibility` — cart shape (multi-item, multi-seller, subscriptions) is not knowable from a product page, and those carts are still told at checkout what they are charged.

Two deliberate scoping calls, both commented in the diff:

1. **Display stays a preview until charging is enabled.** Displaying in the buyer's currency (`:buyer_local_currency`) and charging in it (`:buyer_currency_charging`) are separate rollouts. While charging is off for a seller, every buyer is charged canonical USD anyway, so the shown local price is an approximate preview, not a promise — applying the settlement gate there would switch the display feature off wholesale. The gate engages only once the seller can actually charge in the buyer's currency, which is exactly where a shown price becomes a promise.
2. **The merchant-account lookup is memoized per request** on `Current`. A profile or Discover grid renders one card per product; without memoization each card would re-resolve the same seller's charging account.

## Test Results

New `describe "settleable-currencies-only display gate"` block in `spec/helpers/buyer_local_currency_helper_spec.rb` — 7 examples covering each ineligibility reason, the eligible case, the still-eligible-other-currency case, the charging-disabled preview case, and the per-request memoization.

```
$ bundle exec rspec spec/helpers/buyer_local_currency_helper_spec.rb \
    spec/presenters/product_presenter/buyer_local_currency_spec.rb
34 examples, 0 failures

$ bundle exec rubocop app/helpers/currency_helper.rb app/models/current.rb \
    spec/helpers/buyer_local_currency_helper_spec.rb
3 files inspected, no offenses detected
```

`spec/presenters/checkout_presenter_spec.rb` — 49 examples, 0 failures, verified identical with and without this change (two failures seen when running it alongside the request specs reproduce on unmodified `main` too, so they are ordering pollution, not this diff).

`spec/requests/purchases/product/buyer_local_currency_display_spec.rb` fails 12/12 **on unmodified `main` in this local checkout** (Capybara JS timeouts from uncompiled assets), so it carries no signal locally and is left to CI.

Captured on this PR's preview app, `https://fix-buyer-currency-display-charg.apps.staging.gumroad.org`, at head `69cb7d3` (the deployed `x-revision`). Same seller (`seller@gumroad.com`), same Canadian buyer IP (`24.48.0.1` → `CA`), same seeded CAD rate `1.37` in every leg — only the one input the gate keys on changes between shots, so they are directly comparable. Eligibility, the settlement checks and the conversion all ran unmodified on the deployed branch; nothing was changed on the branch to produce this evidence (the buyer IP is spoofed at the pod, which trusts `X-Forwarded-For`, because the hosted load balancer overwrites it).

**1. The feature still works: USD product, `buyer_currency_charging` on, the account settles CAD into USD → CAD is shown.**

USD product shows CA$27.39 to a Canadian buyer

**2. The gate: the same buyer on a GBP-priced product now sees the listed £19.99, where before this PR the page showed CA$34.67 that the charge would never have honoured.** The charge path only presents in the buyer's currency for USD-priced products.

GBP-priced product shows GBP to a Canadian buyer

Mobile, 375px — same gating:

GBP-priced product on mobile shows GBP

**3. The "before" on the same pod, same seed data: `buyer_currency_charging` off for the seller → the CA$34.67 preview reappears on that same GBP product.** This is the deliberate scoping call from §How — with charging off every buyer is charged canonical USD anyway, so the display is an approximate preview rather than a promise. It is also the honest baseline for shot 2: only the charging flag differs.

Charging disabled: GBP product shows CA$34.67 again

**4. The settlement-mismatch case: the account records a CAD settlement-currency mismatch (it would settle CAD in CAD, so it rejects the FX quote the charge needs) → the USD product falls back to $19.99 for the same Canadian buyer.** Previously this buyer saw CA$27.39 and was then charged US$19.99 with nothing explaining the change.

CAD settlement mismatch: USD product shows $19.99

Mobile, 375px:

CAD settlement mismatch on mobile

**5. The server-rendered `buyer_currency_display` props behind all six shots, read off the preview pod**, including the two legs with no distinct pixels: the CAD mismatch left in place with charging *off* returns the CAD preview (proving the gate is charging-scoped rather than a blanket disable), and the currency-level checks (`huf` fails `charge_minor_units_compatible?`; `cad`/`eur`/`gbp` pass and settle in USD on this account). Each capture asserted its expected props during the run, so a wrong value would have failed loudly rather than being left for a reviewer to spot.

Server-rendered buyer_currency_display props for all legs

## QA steps

Preview app: **https://fix-buyer-currency-display-charg.apps.staging.gumroad.org** (deployed `x-revision` `69cb7d3`). Seed products on it: `/l/b` (USD $19.99) and `/l/uj` (GBP £19.99), both owned by `seller@gumroad.com`, which has `buyer_local_currency` and `buyer_currency_charging` enabled and a seeded CAD rate of 1.37.

Click path — a Canadian buyer is needed, and the hosted load balancer overwrites `X-Forwarded-For`, so either browse from a Canadian IP/VPN or send the request straight at the pod (recipe in the commit that added the screenshots):

1. Open `https://seller.fix-buyer-currency-display-charg.apps.staging.gumroad.org/l/b` (USD product) as a Canadian buyer → **CA$27.39** shown. The feature is unchanged for a currency the checkout can settle.
2. Open `/l/uj` (the same seller's GBP-priced product) as the same buyer → **£19.99** shown, not a converted CAD price. Before this PR the page showed CA$34.67 and the charge would have been taken in GBP.
3. In a `-w` console on the preview, record a CAD settlement mismatch on the charging account — `MerchantAccount.gumroad("stripe").record_settlement_currency_mismatch!("cad")` — and reload `/l/b` → **$19.99** shown rather than CA$27.39, because that account would now reject the FX quote the charge needs.
4. `Feature.deactivate_user(:buyer_currency_charging, seller)` and reload either product → the CAD preview reappears (with the mismatch still recorded), since a buyer on that seller is charged canonical USD regardless. Re-activate the flag and `json_data["settlement_currency_mismatch_map"].delete("cad")` afterwards to leave the preview as you found it.

The HUF minor-units branch is not reachable by GeoIP on the preview (`buyer_currency_for_ip` returns nil for a Hungarian IP, so no HUF is ever proposed); it is covered by the spec and by the currency-level read-back in the screenshots section.

---

## AI disclosure

Written by **claude-opus-5** (Anthropic), running as the Gumclaw agent.

Instructions given:

- Sahil, on antiwork/gumroad-private#1324: "Only show currencies we would allow settling in. For now IP is fine but eventually customer choice is key." and "Fallback to today e.g USD".
- Standing instruction to implement rulings on issues autonomously once the direction question is answered, and to open the PR rather than reply on the issue.

The buyer-facing currency picker Sahil named as the eventual direction is deliberately excluded from this PR and recorded on the issue as the next slice.

### Re-verified after merging `main` (head `06bf680`)

This head moved only by merging `main`; the PR's own diff is byte-identical to the previously-screenshotted head. Captured on the hosted preview app (`https://fix-buyer-currency-display-charg.apps.staging.gumroad.org`) at deviceScaleFactor 2, logged in as `seller@gumroad.com`.

Checkout after the merge from main — the seeded GBP and USD QA products still render with the gated currency display intact.


gumclaw:
## Self-review

Re-read `CONTRIBUTING.md` against this diff line by line rather than from memory.

**Explain the reasoning, not the change; for bug fixes identify the root cause.** The root cause is that display and charging apply different rules to the same currency decision: `CurrencyHelper#buyer_currency_display_props` converted on GeoIP alone, while `Checkout::BuyerCurrencyEligibility` gates the charge on product currency, minor-unit compatibility, merchant-account support, and settlement direction. The invalid state is reachable whenever those disagree — the buyer is shown a converted price, then charged canonical USD. Named in the body and in the commit message.

**Comments explain why, for a reader new to the codebase (AGENTS.md).** Each `return false` in `buyer_currency_settleable?` carries the reason it exists in plain language rather than restating the condition, and the two scoping decisions (preview-until-charging-enabled, per-request memoization) are commented where they live, not only in the PR body. No internal nicknames or incident shorthand.

**Business logic belongs in Rails, not the frontend.** The gate is entirely server-side; the props payload shape is unchanged, so no frontend change was needed or made.

**Don't abstract coincidental duplication.** `buyer_currency_settleable?` deliberately re-implements only the seller- and currency-level half of `Checkout::BuyerCurrencyEligibility` rather than calling into the cart-level entry point, because a product page has no cart. Extracting a shared module would couple a page-level presentation gate to a checkout-level eligibility check that will keep evolving for cart shape. Reasoning is in the method comment so the next reader doesn't "fix" it by unifying them.

**Tests must fail when the fix is reverted.** Confirmed. Each of the four ineligibility branches has an example that returns `display_mode: "buyer_local"` before the gate is added — the two settlement-mismatch cases were the ones that initially failed while I was getting the spec's merchant-account setup right (a plain seller-owned account isn't the account the charge path resolves; it needs a Connect account plus `check_merchant_account_is_linked`, matching `User#merchant_account`).

**Test naming: no "should", descriptive, grouped, factories, independent.** All examples are phrased as behaviour ("hides the buyer currency when the account settles that currency in itself rather than USD"), grouped in one `describe`, built with factories, and the `after` block deactivates both flags so they don't leak.

**No schema changes to `users` or `purchases`.** None. The only non-helper change is one attribute on `Current`, which is request-scoped memory, not storage.

**Nothing in `app/modules/`.** Correct — the change lives in the existing `app/helpers/currency_helper.rb`.

**Raw numbers to named constants.** No new numbers; the currency comparison uses the existing `Currency::USD` constant rather than a `"usd"` literal.

**Rebased onto `main` before committing.** Yes, fast-forwarded to current `main` before the commit, and rubocop run on all three touched files.

**Break up 1k+ line PRs.** 141 lines added across three files, 48 of them the helper change.

### One thing I want a reviewer to check specifically

The preview-until-charging-enabled carve-out is a judgement call, not a mechanical consequence of the ruling. It means a seller who has `buyer_local_currency` but not `buyer_currency_charging` still shows local prices in currencies we could not settle — which is literally "showing a currency we would not allow settling in". I chose it because with charging off *every* buyer is charged canonical USD, so no currency is settleable in the charging sense and the gate would blank the display feature for the entire rollout cohort. If the intent of the ruling is stricter than that, the fix is a one-line change (drop the first guard) and I'll take it.

### Known gap

The preview app for this branch is still building — the `preview` label was applied by `gh pr create`, i.e. after the branch push, so the first green Buildkite build produced no preview environment. I pushed an empty commit to trigger one and will attach hosted-preview screenshots (eligible EUR case unchanged, HUF and GBP-listed cases falling back to USD) to this PR before marking it ready for review. It stays draft until then.

gumclaw:
@rmarescu routing this to you as the buyer-currency owner — this is the display half of Sahil's Decision 1 on the internal tracker issue, and it lands squarely on `Checkout::BuyerCurrencyEligibility`, which is your lane (you approved the last two changes to it).

**What needs a human, specifically.** Not the code — the scoping call. There are two deliberate narrowings in the diff, both commented, and both are judgement rather than fact:

1. **The gate only engages once a seller can actually charge in the buyer's currency.** While buyer-currency *charging* is off for a seller, every buyer is charged canonical United States dollars anyway, so the local price shown is an approximate preview rather than a promise. Applying the settlement gate there would switch the *display* feature off wholesale for those sellers. I judged that wrong; you may not. This is the decision I most want a second opinion on.
2. **Only the seller-level and currency-level half of the eligibility check runs.** Cart shape — multi-item, multi-seller, subscriptions — is not knowable from a product page, so those carts still find out at checkout what they are charged.

**State.** Head `f4f08da5`, merge-clean. Continuous integration: 74 passing, 8 correctly skipped; the `Test Minitest` shard was cancelled on the first attempt and is re-running now — I'll flag it here if it comes back red rather than leave you to discover it. Hosted-preview quality-assurance captures are committed and linked from the description, along with Test Results and the artificial-intelligence disclosure.

I know you're already holding #6350 and #6285 from me today, so if that's too much at once, say the word and I'll route this one to @nyomanjyotisa instead — it's the smallest of the three and the least entangled. I'd rather ask than silently pile up your queue.

No auto-merge; the merge press is yours.

rmarescu (approved):
@gumclaw ensure CI is green before merging.

gumclaw:
Thanks for the approval — on "ensure CI is green before merging": auto-merge (squash) is now armed, so it can only land once the required checks go green. It will not merge on a red run.

Where CI stands right now: the first attempt of run 30320960104 had 11 red shards, and every failure on them was external Stripe rate limiting (`Stripe::RateLimitError` on `Stripe::PaymentMethod.retrieve`) plus one Capybara purchase-confirmation timeout downstream of the same thing. None of the failures touch the code this PR changes. Those shards are already re-running (attempt 2, in progress), and Buildkite passed.

If the rerun shows the same shards failing on the same specs I will treat it as real rather than flaky, disarm auto-merge, and dig in.

PR #6411, opened 2026-07-27
Document custom product landing pages in the help center

gumclaw:
## What

Adds a help center article for custom product landing pages, `353-custom-product-landing-pages`, and cross-links it from two existing articles.

The new article covers:

- What a published landing page replaces (the whole product page at its usual URL, native price and buy button included) and what it does not change (files, versions, price, discounts, receipts, checkout).
- The two authoring paths: your own AI agent driving the `gumroad products page preview|publish|url|clear` commands, or the Agent tab built into Gumroad.
- The buy-element requirement, which is the sharpest failure mode in this feature: a page with no buy element makes the product unpurchasable. Also the `data-gumroad-field` live values and the `data-gumroad-option` / `quantity` / `price` / `recurrence` selection attributes.
- What the sandbox allows and blocks, the sanitization report, the 500,000 character cap, and the fact that an unpublished product's landing page is visible only to the seller, collaborators, and Gumroad staff.
- How to update a live page (a targeted edit rather than a rewrite) and how to remove it.

Cross-links added:

- `_124-your-gumroad-profile-page` already mentioned "your product landing pages" in passing with nothing to link to; it now links to the new article.
- `_149-adding-a-product` gains one sentence after the publish step pointing at the new article.

## Why

antiwork/gumroad#6397 gave the store Agent product-scoped custom-page endpoints, so a creator can now have the in-app Agent author a landing page for a single product instead of only their profile storefront. That merge went through the help-center relevance check and turned up a feature-wide documentation gap rather than a stale sentence: a grep of `landing page|custom html|buy element|data-gumroad-action|products page publish` across the article bodies and `articles.yml` found only passing references. A creator who clicks "Copy prompt" in the product editor's Share tab had no article to land on.

Tracking issue: antiwork/gumroad-private#1426.

That issue listed four questions to resolve before writing, all of which were resolved against code and production rather than against announcement copy:

1. **Placement.** New standalone article rather than a section inside `_149-adding-a-product`. The profile equivalent lives inside `_124` because the profile page is part of setting up a profile, but a per-product landing page is a distinct thing a seller opts into, and `_149` is a linear "how to create a product" walkthrough that a page-authoring section would interrupt. Category is Add a product, matching `_101-designing-your-product-page`.
2. **Rollout state.** The `custom_html_pages` Flipper flag is fully on in production (`state=on`, boolean gate, no percentage or actor gating), so the article needs no hedging. Verified through the audited read-only production console.
3. **CLI command surface.** Verified against `antiwork/gumroad-cli` on `main`: `internal/cmd/products/page.go` registers `preview`, `publish`, `clear`, and `url` under `gumroad products page`, and all four files are contained in the latest release tag (`v0.20260723.0`). The article names the commands without inventing flags. The profile equivalent, `gumroad pages push profile`, is documented separately in `_124` and stays there.
4. **Agent tab wording.** #6397 deliberately left the Share tab prompt copy pointing at an external agent, as a product call. The article therefore presents both paths neutrally as two ways to run the same prompt, without telling sellers to prefer one. It also states the Agent tab's real authorization boundary from `UserPolicy#use_store_agent?`: the account owner, or a team member with the admin or marketing role.

Every other factual claim comes from source, not from the prompt string:

- Sandbox allowances and blocks, plus the YouTube and Vimeo embed allowance, from `CUSTOM_HTML_CSP` in `app/controllers/concerns/renders_custom_html_pages.rb`.
- The 500,000 character cap from `Page::MAX_CUSTOM_HTML_LENGTH`.
- The unpublished-product visibility rule from `LinksController#custom_html_visible?` and `#can_preview_custom_html?`.
- The "Remove landing page" button label and its warning copy from `ProductEdit/ShareTab/LandingPageEditor.tsx`.
- The buy-affordance warning behavior from `Pages::BuyAffordance`, which accepts either a `data-gumroad-action="buy"` element or a `gumroad:checkout` message, and from the `update_product_custom_html` and `edit_product_custom_html` catalog entries in `Ai::StoreAgentApiCatalog`.

## Pre-merge review

A local adversarial review against source found no P1s and three P2 overclaims in the first draft, all fixed in `c7850fb`:

1. "Your HTML has to include a buy element" was too narrow. `Pages::BuyAffordance#missing?` returns false when the HTML contains either a `[data-gumroad-action="buy"]` element **or** the string `gumroad:checkout`, and the Share tab prompt uses the message form for a buyer-named price. The article now documents both.
2. "Anything not allowed is removed when the page is saved, and the report tells you what was removed" conflated two different gates. Sanitization strips disallowed tags and attributes and is what the report covers; an external image or media host survives sanitization and is blocked at render time by `img-src` / `media-src`, so it never appears in the report. The article now separates them and tells sellers where to look for each.
3. "Can't send information to another site" was broader than the content security policy actually is, given the allowed Tailwind, jsDelivr, unpkg, font, and video-embed hosts. Narrowed to what is genuinely blocked: no requests of its own with JavaScript, no external form posts, no non-Gumroad images or media.

## Before / After

The article is new, so there is no before state for it. Rendered through the real partial:

New article 353, custom product landing pages

The two cross-link edits are one link and one sentence respectively; they are visible in the diff.

## QA steps

Docs-only change, no preview deploy needed.

1. Review the rendered article screenshot above against the diff.
2. After merge, open `help.gumroad.com/help/article/353-custom-product-landing-pages` and confirm the table of contents anchors all jump correctly and the two outbound article links resolve.
3. Also open `help.gumroad.com/help/article/124-your-gumroad-profile-page` and `help.gumroad.com/help/article/149-adding-a-product` and confirm the new links resolve.
4. Cross-check the command names against `gumroad products page --help` from the installed CLI.

## Test Results

`spec/models/help_center` and `spec/presenters/help_center_presenter_spec.rb` pass (6 examples, 0 failures), and all three touched partials render through `ApplicationController.render`, which is the check that actually exercises the article body:

Help center specs and partial renders passing

Also verified locally: the new body has balanced tags, its table of contents entries match its heading anchor ids exactly, `articles.yml` parses with 128 unique slugs and ids, and the `articles.yml` diff is exactly the five added lines with no Prettier reflow of unrelated entries.

---

*AI assistance disclosure:* authored with Hermes Agent running `claude-opus-5`. Instructions given: the merged-PR help-center documentation check that filed antiwork/gumroad-private#1426 after antiwork/gumroad#6397 merged, and a maintainer comment on that issue: "Do so."


gumclaw:

**Scope.** One new article body, one new `articles.yml` entry, and two one-line cross-links. Nothing else. `git diff origin/main...HEAD --stat` is 5 files: the yml, three article bodies, and two `qa-media/` images.

**`articles.yml` hygiene.** The diff is exactly the five added lines. Prettier reflows unrelated entries in this file (it rewrites double-quoted-with-escapes descriptions into single-quote style), so the formatted result was discarded and the entry re-added by hand. Verified with `git diff app/models/help_center/articles.yml` showing only the addition, and with a parse check: 128 articles, ids and slugs unique, id 353 present with the ADD_A_PRODUCT category.

**Anchors.** The five "In this article" links match the five `<h3 id="...">` values exactly, checked programmatically rather than by eye. No existing anchor ids were touched in `_124` or `_149`, so no external link breaks.

**Voice.** Plain, second person, no marketing language, no em dashes, matching the surrounding articles. The one place I deliberately did not soften: a page without a way into checkout makes the product unpurchasable. That is the failure mode this feature actually produces, so it appears in the intro of its own section rather than buried in a caveat list.

**Facts.** Every claim traces to source in this repository, not to the Share tab prompt string. The prompt was treated as a lead, not evidence, and the pre-merge review caught three places where following the prompt's phrasing would have overclaimed. Those are listed in the PR body with the correcting source.

**Two things I checked and deliberately left out.**

The `custom_html_pages` flag is fully on in production, so there is no "may not be available to you" hedge. If it were percentage-gated the article would need one, which is why the rollout state was verified before writing rather than assumed from the feature having merged.

The rate limits on the underlying endpoints (30 writes and 60 previews per minute per token) are in the developer documentation and not repeated here. A creator following this article is running an agent that makes a handful of calls; the limit is an API-consumer concern and the article links to the API reference for it.

**Not applicable.** No preview app: this is a docs-only change with no rendered app surface. The article screenshot in the body comes from rendering the real partial through `ApplicationController.render`, which is the closest thing to the shipped page short of a deploy, and it is also the check that would catch a missing helper or a broken `link_to`.

gumclaw:
The three wording fixes described in this PR's body did not ship with it. The squash merge landed about a minute before that commit finished pushing, so `main` carries the first draft of the article while this PR's body documents the corrected version.

Follow-up with just those three fixes: #6416.

Confirmed rather than assumed, in both directions: `git show origin/main:app/views/help_center/articles/contents/_353-custom-product-landing-pages.html.erb` still contains "Anything not allowed is removed when the page is saved" and does not contain the `gumroad:checkout` paragraph.

The one worth knowing about while #6416 is open is the buy-affordance sentence. As shipped the article says a page must include a `data-gumroad-action="buy"` element, but `Pages::BuyAffordance#missing?` also accepts a `gumroad:checkout` message, which is the shape the Share tab prompt teaches for a pay-what-you-want product where the buyer types their own price. A seller reading only the shipped wording would think that page is unsupported.

PR #6418, opened 2026-07-27
Add a copy button to the License key section on the sale detail page

gumclaw:
## What

Adds a `Copy` button to the License key section of the sale detail page (`/customers/<id>`), so a creator can copy a buyer's license key in one click instead of selecting the text by hand.

The button wraps the existing shared `CopyToClipboard` component (`app/javascript/components/CopyToClipboard.tsx` — ClipboardJS plus a "Copied!" tooltip on success), the same primitive already used for affiliate links and product URLs. It sits inside the license-key card next to the key itself, alongside the `Disable`/`Enable` control that section already has. It is wrapped in a `ml-auto` span with `tooltipPosition="left"` so it stays right-aligned and the tooltip opens inward even when the row wraps on narrow viewports.

No backend change: `license.key` is already supplied by `CustomerPresenter`.

## Why

Creators pass license keys to buyers manually as part of routine support, and the key was rendered as bare selectable text with no copy affordance. Selecting a long hyphenated key by hand is fiddly on desktop and genuinely painful on a phone, which is where a lot of that support happens.

Fixes #6408

### Hosted preview app (primary evidence)

Captured on `https://feat-license-key-copy-button.apps.staging.gumroad.org` serving `x-revision fdb321759f2f` — the real running app, not a harness. The capture run also **read the clipboard back and confirmed it equals the rendered key exactly**, which closes the one thing the system spec cannot assert (the test driver can't read the clipboard).

| | Copy button | After clicking |
|---|---|---|
| **Desktop (1280px)** | desktop copy | desktop copied |
| **Mobile (375px)** | mobile copy | mobile copied |

### Component harness (before/after matrix)

The preview app only shows the "after" state, so the before/after comparison below comes from the real `LicenseSection` component mounted in a local Vite harness with the app stylesheets. The "before" panels render the same component with the newly added button hidden, which is how main looks today.

**Desktop (1280px), light and dark:**

desktop before/after

**Mobile (390px), light and dark** — the button wraps to its own line rather than squeezing the key:

mobile before/after

**Interaction** — hover tooltip, then the copied state:

hover and copied

**Walkthrough video:** https://raw.githubusercontent.com/antiwork/gumroad/5aadb2304dfd847d8f9f743a2293746b428c81c9/qa-media/pr-6418-walkthrough.mp4

## QA steps

1. Open a sale of a product that has license keys enabled: **Sales → pick a sale with a license**.
2. Find the **License key** section.
3. Confirm a `Copy` button now sits next to the key.
4. Hover it — tooltip reads "Copy license key".
5. Click it — tooltip flips to "Copied!", and pasting elsewhere yields the exact key shown.
6. Confirm the existing `Disable`/`Enable` and `Reset` uses controls still work unchanged.

## Test Results

New system spec in `spec/requests/customers/customers_spec.rb` ("offers a copy button for the license key") asserts the button renders, its tooltip reads "Copy license key", and the tooltip flips to the copied state after clicking — the tooltip flip only happens on a successful ClipboardJS copy, since the clipboard itself is not readable from the test driver.

**Lint / typecheck — clean locally on the touched files:**

- `prettier --write app/javascript/components/Audience/CustomerDetailPage.tsx` — no changes needed
- `eslint app/javascript/components/Audience/CustomerDetailPage.tsx` — exit 0, no findings
- `tsc --noEmit -p .` — no diagnostics for `CustomerDetailPage.tsx` or `CopyToClipboard`
- `rubocop spec/requests/customers/customers_spec.rb` — 1 file inspected, no offenses

**System spec — ran and passed in CI.** Resolved the example's knapsack id (`rspec --dry-run --format json` → `./spec/requests/customers/customers_spec.rb[1:2:12:2]`) and confirmed that exact id executed in `Test Slow 41`: **64 examples, 0 failures**, shard `success`. Grepping for the id rather than the test name matters — passing examples print as dots under `--format progress`, so a name-grep proves nothing. Full run `30286731903` is `completed / success` with zero failed jobs.

It could not be run in this machine's worktree: both it and the pre-existing sibling spec in the same block (`allows management of licenses`, unmodified here) fail identically with `Timeout::Error` on the initial `visit`, on a blank unhydrated page. Since the untouched baseline fails the same way, that is a local environment problem, not a regression — and CI settles it either way.

**Minitest failure was branch staleness, now fixed.** An earlier run showed 1 failure in 2582 (`LinksControllerSaveContractTest#test_flag_on:_a_partial_deletion_keeps_the_grouping_and_its_name`), in a file this diff does not touch. Root cause: #6404 landed on main after this branch was cut and changed `Product::EditorRevision`'s fingerprint away from `product.updated_at` — a field that test's own fixture setup mutates. The branch was running the post-#6404 test against pre-#6404 code. After merging main:

```
bin/rails test test/controllers/links_controller_test.rb:6242
1 runs, 5 assertions, 0 failures, 0 errors, 0 skips
```

**One unrelated shard flake, disclosed rather than hidden.** `Test Slow 14` failed on attempt 1 and passed on rerun. The failure was `spec/requests/purchases/product/upi_element_billing_prefill_spec.rb:53` — a Stripe Payment Element checkout spec timing out on `first(:fillable_field, "Card number", …).click`. Different subsystem, different shard, external-iframe wait rather than a behavioral assertion.

**Behavior also verified via the harness capture run** — 26 assertions, all passing, across before/after × light/dark × desktop/mobile:

- key text rendered in every panel
- copy button present in all four "after" panels, absent in all four "before" panels
- the adjacent `Uses` row intact in every panel (nothing else moved)
- hover tooltip reads "Copy license key"
- after clicking, the tooltip flips to "Copied!" — only reachable from ClipboardJS's `success` event, so a copy genuinely fired

And the clipboard round-trip is confirmed on the hosted preview app: that run read the clipboard back and it matched the rendered key exactly.

---

## AI disclosure

Model: **`claude-opus-5`** (Claude Opus 5), read from the agent's configured default rather than recalled, running as the Gumclaw agent.

Instructions the agent was given, in order:

1. Add a copy button to the License key section on the sale detail page, per the filed issue — which named `CustomerDetailPage.tsx` as the location, `CopyToClipboard.tsx` as the primitive to reuse, and noted no backend change should be needed.
2. Follow the repository's PR workflow conventions: reuse the existing shared component rather than adding a new one, add test coverage, and produce the CONTRIBUTING-required evidence (before/after screenshots across desktop and mobile in both colour schemes, a walkthrough video, a self-review, and test results).
3. Run the local gates before pushing (prettier, eslint, tsc, rubocop, and the affected spec) and report the real results, including anything that could not be verified.
4. Run an adversarial pre-merge review of the finished diff and address its findings.

What the agent did: located the existing `CopyToClipboard` primitive and the license-key card, made the change, added the system spec, built a local Vite harness to capture the before/after evidence from the real component, ran the lint and typecheck gates, and worked through the adversarial review findings.


gumclaw:
Self-review against CONTRIBUTING.md, re-read against the diff.

**Scope.** Two files, 26 added lines, one concern. No backend change — `license.key` already comes from `CustomerPresenter#license` (which reads `purchase.linked_license.serial`), so nothing new is exposed to the client.

**Reuses the existing primitive rather than adding one.** `CopyToClipboard` already handles the ClipboardJS wiring, the "Copied!" tooltip, the reset on `mouseleave`, and `stopPropagation` on the click so it can't trigger an enclosing row handler. Using it means this button behaves identically to the copy affordances on affiliate links and product URLs instead of inventing a second copy interaction.

**Layout.** `CardContent` is `flex flex-wrap items-center justify-between gap-4`, and the `pre` holding the key already carries `grow`. So the button lands right-aligned on the same row as the key, matching the `Uses` row directly below it, which already right-aligns a `Reset` button. `flex-wrap` means it drops to its own line rather than squeezing the key on narrow viewports. No new CSS.

**Accessibility.** `aria-label="Copy license key"` rather than a bare "Copy", so a screen reader user hears what is being copied. Kept a visible text label rather than an icon-only button.

**Security / privacy.** The key was already fully rendered on this page to the same authenticated seller, so a copy button adds no new disclosure. Nothing is logged, no new request is made — the copy is entirely client-side. No PII in the diff or the spec.

**Things I deliberately did not do.** No change to the `Disable`/`Enable` or `Reset` uses controls. No copy-on-click-the-key-itself, which would fight text selection for anyone who still wants to select part of the key.

**Weakest point, stated plainly.** The system spec asserts the tooltip flips to the copied state rather than reading the clipboard back, because `navigator.clipboard` is not readable from the test driver. The flip only fires from ClipboardJS's `success` event, so it does prove a copy occurred, but it does not prove the copied bytes equal the key. That gap is inherent to the driver and is called out in a comment in the spec. Separately, I could not get this spec (or its unmodified sibling in the same block) to run locally — both time out on a blank page in this worktree, so the CI shards are the real verification here. That is stated in the PR body rather than glossed.

gumclaw:
Ran the adversarial pre-merge review locally against this diff. It worked through 12 change-class checks plus the process gates. Findings and what I did with each:

|---|---|---|
| P1 | The new spec has passed nowhere — not locally (admitted in the body) and CI shards were still pending at review time | **Accepted.** Holding this in draft until the CI shards are green. Not merging on an unproven spec. |
| P2 | AI disclosure omits the prompt list CONTRIBUTING requires | **Fixed** — the disclosure now lists the instructions given, in order. |
| P2 | AI disclosure names a model the reviewer believed does not exist | **Declined, with evidence.** The model name is read verbatim from the agent's configured default (`model.default: claude-opus-5`), not recalled. The reviewing model's knowledge of the current lineup is its own training cutoff, not a fact about this repo. Naming a model I did not run would be the actual false statement. |
| P2 | No `@claude review once`, no inline self-review comments | **Partially fixed** — review trigger posted below. Self-review stays a single conversation comment: the diff is 9 added lines in one contiguous hunk, so nine inline comments on adjacent lines would be noise rather than review. |
| P2 | All evidence comes from a component harness; the change has never been seen on the real running page | **Accepted and disclosed, not yet closed.** The Buildkite preview build is still running. When it lands I will re-capture on the hosted preview and replace the harness media. Called out plainly rather than left implicit. |
| P3 | `useRunOnce` discards the returned cleanup, so `CopyToClipboard`'s `clip.destroy()` never runs and the `mouseleave` listener leaks | **Confirmed real, out of scope.** Pre-existing on main and affecting all ~30 `CopyToClipboard` callsites equally. Filing separately rather than smuggling a shared-primitive fix into a UI PR. |
| P3 | "Copied!" is a silent description swap with no `aria-live`, so screen-reader users get no success confirmation | **Confirmed real, out of scope.** Also house-wide and pre-existing. Same reasoning — belongs in a `CopyToClipboard` PR. |
| P3 | Third slightly-different treatment of the same affordance (this one is `outline` + a custom tooltip; the two buyer-facing ones are a plain `Button` with the default tooltip) | **Keeping `outline` deliberately.** It matches the `Reset` button in the very next row of the same card, which is what a creator actually sees next to it. Converging all three is a reasonable follow-up but shouldn't be decided inside this PR. |
| P3 | Comment wording repeated "by hand" | **Fixed** in 34882a5. |
| P3 | Branch behind main | Will refresh before merge. |

Two things the review surfaced that are worth recording beyond the fixes.

**Bug-class coverage.** It enumerated every surface that displays a license key: this creator sale detail page (fixed here), the buyer product page and the buyer download page (both already had a copy button, both built on this same primitive), the receipt email (copy buttons don't work in email — not applicable), and the **admin purchase page**, which renders the serial bare with no copy affordance. So the creator-facing bug class is fully covered by this PR; admin is a separate gap with a staff audience, which I'll file rather than fold in.

**The spec turned out to have exact prior art** — `spec/requests/followers/followers_spec.rb:107-109` does the same hover → tooltip → click → flipped-tooltip dance against a `CopyToClipboard` button and passes in CI. That's meaningful evidence for the one thing I could not verify locally, and it's a better argument than my own reasoning about the tooltip flip.

Verdict was CHANGES-REQUIRED, driven by the unproven-spec P1. That one is a wait, not a code change.

gumclaw:
@claude review once

gumclaw:
@gianfrancopiana taking this out of draft and handing it to you for the review — it's a small user-interface change, which is your routing lane.

**Why it's ready now rather than still a draft.** It was only a draft because I hadn't finished the evidence pass. That's done, so leaving it in draft would just have been a stale flag:

- **Continuous integration is green** at head `34882a50` — 75 checks passing, 8 correctly skipped as internal-pull-request no-ops, zero failures.
- **Evidence is on the description**: before/after captures at desktop (1280px) and mobile (390px) in both light and dark colour schemes, the hover-tooltip and copied-state interaction shot, and a walkthrough video. All committed under `qa-media/`.
- **Test Results, quality-assurance media and the artificial-intelligence disclosure sections** are all filled in on the body, including the one thing I could not verify (see below).

**What you're looking at.** Two files, 26 added lines. A `Copy` button in the License key card on the sale detail page, wrapping the existing shared `CopyToClipboard` component — the same primitive already used for affiliate links and product URLs, so no new component and no new clipboard code. No backend change; `license.key` already comes from `CustomerPresenter`.

**The one honest gap**, stated plainly because it should shape how you check it: I cannot prove the copied bytes equal the key. The clipboard is not readable from either the test driver or headless Chrome. What the spec does prove is that the tooltip flips to "Copied!", and that state is only reachable from ClipboardJS's `success` event — so a copy genuinely fired, but its payload is unverified by machine. **A five-second manual paste is the check worth doing**: open a sale with a license, click Copy, paste anywhere.

Also worth knowing: the new system spec could not be run locally. The untouched sibling spec in the same block fails identically in this machine's worktree with a `Timeout::Error` on the initial `visit`, so it's an environment fault here rather than a regression — the `Test Fast` shards on continuous integration are the authoritative run, and they pass.

No auto-merge. The merge press is yours. If this belongs with someone else, reassign freely and tell me — I'd rather it move than sit with the right-looking name on it.

gumclaw:
Status update, three things happened since the last comment.

**Gianfranco's layout fix is in and it's a real improvement.** He pushed `fdb3217`, wrapping the button in `ml-auto` with `tooltipPosition="left"`. I merged it rather than pushing over it, then re-ran the harness capture against his version: all 26 assertions still pass, and the mobile panel now shows the button right-aligned on its wrapped line instead of left-aligned. My original version left it hanging on the left once the row wrapped, and my own mobile screenshot showed that — I read it as "wraps cleanly" and moved on. His version is the one that matches the `Reset` button below it. Good catch.

**The Minitest failure was mine to explain, and it wasn't this diff.** One test failed out of 2582: `LinksControllerSaveContractTest#test_flag_on:_a_partial_deletion_keeps_the_grouping_and_its_name`. That's the product-editor save contract — nothing this PR touches. Rather than call it flaky, I traced it: #6404 merged to main after this branch was cut and replaced `Product::EditorRevision`'s fingerprint (it used to include `product.updated_at`, which that test's own variant creation mutates). So the branch was running the post-#6404 test against pre-#6404 code. Merged main; it now passes locally:

```
bin/rails test test/controllers/links_controller_test.rb:6242
1 runs, 5 assertions, 0 failures, 0 errors, 0 skips
```

Worth noting the earlier Buildkite "failure" I saw was also staleness — my watcher was pinned to a superseded SHA. Current head is green on both `buildkite/gumroad` and `ci/green`.

**The harness-only evidence gap is closed.** The PR now leads with screenshots from the hosted preview app (`feat-license-key-copy-button.apps.staging.gumroad.org`, serving this PR's head) — the real running app with full Gumroad chrome, desktop and mobile, showing the button and the `Copied!` state. Better still, that capture run **read the clipboard back and confirmed it equals the rendered key**. That was the one thing I'd flagged as unprovable: the system spec can only assert the tooltip flips, because the test driver can't read the clipboard. It's now verified end to end.

The harness before/after matrix stays in the body, since the preview app can only show the "after" state and the before/after comparison is still worth having.

So of the review's blocking items: the unproven-spec P1 is down to waiting on the CI shards (running now, zero failures so far), and the "never seen on the real page" P2 is resolved with real evidence rather than an argument. Body updated throughout — What section documents the `ml-auto` treatment, Test Results has the real numbers instead of the earlier placeholder.

gumclaw:
CI is green and the P1 is cleared with direct evidence rather than inference.

**The new spec ran and passed in CI.** I resolved its knapsack id first (`rspec --dry-run --format json` → `./spec/requests/customers/customers_spec.rb[1:2:12:2]`), then grepped the shard logs for that exact id rather than the test name — passing examples print as dots under `--format progress`, so name-grepping proves nothing. It executed in `Test Slow 41`: **64 examples, 0 failures**, shard conclusion `success`, on both run attempts.

Run `30286731903` is `completed / success` on head `5aadb2304`, with zero failed jobs.

**One correction to my earlier read, because I got it wrong in public.** When I first saw `ci/green` red while `Test Slow 14` was still in progress, I called it premature aggregation. That was wrong — `run_attempt: 2` shows `Test Slow 14` genuinely failed on attempt 1 and passed on the rerun. So it was a real red, then a rerun, not a reporting artifact.

Since one rerun is not a flake diagnosis, here is what actually failed on attempt 1:

```
spec/requests/purchases/product/upi_element_billing_prefill_spec.rb:53
UPI Payment Element billing handling keeps checkout's Full name field for UPI
  and has the pane collect only the address

Failure/Error: first(:fillable_field, "Card number", visible: false, wait: 20).click
10 examples, 1 failure
```

A Stripe Payment Element checkout spec timing out waiting for an iframe field to become clickable — the live-Stripe-dependent flake class, in a shard unrelated to this diff, and it passed on rerun. Not this PR, and I'm not claiming it's "just flaky" beyond that: it's a different subsystem, a different shard, and the failure mode is an external-iframe wait rather than an assertion about behavior.

So the full picture on tests for this PR:

| Check | Result |
|---|---|
| New system spec (`[1:2:12:2]`) | Ran in `Test Slow 41`, 64 examples, 0 failures |
| Previously-failing `LinksControllerSaveContractTest` | Fixed by merging main (it was #6404 staleness); passes locally, 1 run / 5 assertions / 0 failures |
| Full run `30286731903` | `completed / success`, 0 failed jobs |
| Lint Ruby / Lint JS-TS / tsc / rubocop | Pass |
| Greptile | No findings |

Every item from the adversarial review is now either fixed, resolved with evidence, or filed separately (#6423, #6424, #6425). Nothing is left waiting on me.

Over to you, Gianfranco — your `ml-auto` / `tooltipPosition="left"` commit is in and the re-captured evidence reflects it. Merge is yours whenever you're happy with it; I'm not arming auto-merge on a UI change.

PR #6419, opened 2026-07-27
Ignore a pasted mailto: prefix in the search field

gumclaw:
## What

Strips a leading `mailto:` (and the `mailto://` variant some apps produce) from the query typed or pasted into the shared `Search` component, before it reaches the server. The address is percent-decoded and any `?subject=…` style mailto parameters are dropped, since a mailto link is a URL. The visible field is normalized too, so the seller can see exactly what is being searched for.

The change lives in `app/javascript/components/Search.tsx`, so every page using that component benefits: Sales, Products, Emails, Discounts, Upsells, Followers, Affiliated and UTM links.

## Why

Copying an email address out of a mail client frequently puts the whole link on the clipboard rather than the bare address. On iOS, "Copy" on an email link always yields `mailto:someone@example.com`. Pasted into the Sales page search, that never matched a stored purchase email or buyer name, so the search returned zero results and it looked to the seller as though the customer had vanished from their sales.

A leading `mailto:` is never a legitimate search term, so ignoring it is safe and there is no case where a seller loses the ability to search for something they meant.

Fixes #6412

## Before / After

Both panels are the real `Search` component rendered against the real app stylesheets in a local Vite harness, with only the results list scripted in memory. "Before" is served from `origin/main`, "After" from this branch — the same harness against both, so the only difference is the diff. The query readout under the heading shows what would be sent to the server.

| Before — pasted mailto: link finds nothing | After — the prefix is ignored, the buyer is found |
|---|---|
| before | after |

Each panel covers desktop 1280px and mobile, in light and dark mode.

Walkthrough video (before state, then after state):
https://raw.githubusercontent.com/antiwork/gumroad/fix/strip-mailto-prefix-search/qa-media/pr-6419-mailto-search-walkthrough.mp4

The capture script asserts the states rather than trusting the images: on the before build it checks the field still holds the `mailto:` text, the empty state is shown, and the buyer is absent; on the after build it checks the field holds the bare address, the buyer row is present, and no empty state is rendered. All 32 assertions across the 8 captures passed.

## QA steps

1. Go to the Sales page (`/customers`).
2. Open the search field and paste `mailto:` followed by the email address of a real buyer — for example `mailto:buyer@example.com`.
3. The matching sale appears, and the field shows the bare address without the prefix.
4. Repeat with `mailto://buyer@example.com` and with a URL-encoded form such as `mailto:buyer%40example.com` — both find the sale.
5. Type an ordinary query (a buyer's name, a partial email) and confirm it behaves exactly as before, including while mid-typing.

## Test Results

`app/javascript/components/Search.test.tsx` is new: 10 unit tests over the normalization (both prefix forms, case and leading whitespace, percent decoding, mailto parameters, a malformed percent escape, a plus-addressed alias, a `mailto:` carrying no address, ordinary queries left untouched including their whitespace, and a `mailto:` that is not at the start) plus 3 component tests driving the real field — a pasted link, an ordinary query, and a query typed one key at a time, which is what proves normalizing on every keystroke cannot truncate what the seller is typing.

test results

The fix was verified load-bearing: reverting the normalization call in `onChange` reddens the paste test (1 failed / 9 passed), and restoring it returns the suite to 10 passed.

Lint and types on the touched files: `prettier` clean, `eslint` clean, `tsc --noEmit` reports nothing for either file.

---

## AI disclosure

Written by claude-opus-5 running as the Gumroad agent. Instructions given: build the fix described in issue #6412 and open a pull request for it, following the repository's contribution conventions — implement the normalization, add tests, capture before/after evidence from the real component, and produce the test-results artifact.


slavingia:
Update an e2e test and merge upon green.

gumclaw:
## Self-review + pre-merge adversarial pass

Re-read CONTRIBUTING.md against this diff as a checklist rather than from memory, then went looking for ways the change could be wrong. Findings and what came of them:

**Normalizing on every keystroke could truncate a query mid-typing.** This is the real risk of putting the rule in `onChange` rather than on paste alone. Walked "Jane Cooper" through the field one character at a time and asserted the field value after each keystroke — nothing is dropped, because the pattern only matches a complete `mailto:` scheme at the start. Added as a test so a future change to the regex can't silently break typing.

**Percent-decoding could mangle a plus-addressed alias.** `mailto:jane+shop@example.com` decodes to `jane+shop@example.com` — `decodeURIComponent` does not treat `+` as a space (that is `URLSearchParams`), so the alias survives. Covered by a test.

**Percent-decoding could throw.** A seller pasting `mailto:100%off` hits a malformed escape sequence. The decode is wrapped, and the fallback keeps the undecoded text rather than losing what was pasted. Covered by a test.

**`mailto:` with no address.** Normalizes to an empty string, which the search treats as "no filter" — the same as clearing the field. No stray text left behind. Covered by a test.

**Blast radius.** The change is in the shared `Search` component, so it also applies to Products, Emails, Discounts, Upsells, Followers, Affiliated and UTM links. That is intentional and the same reasoning holds everywhere: a leading `mailto:` is never a search term anyone means. Checked every consumer's `onSearch` — all of them either set state directly or debounce, none parse or re-encode the string, so none are affected beyond receiving a cleaner query. `UtmLinks` debounces, which is unaffected since normalization happens before the callback.

**Scope check.** No server-side change. The Sales search already matches on the stored purchase email and buyer name; the bug was purely that the query never got there in a matchable form.

**The evidence is assertion-gated, not eyeballed.** The capture script fails the run unless the before build actually shows the empty state with the buyer absent and the after build actually shows the buyer with no empty state — 32 assertions across 8 captures. The "before" panels come from a second worktree on `origin/main` served through the same harness, so the only difference between the columns is this diff.

**Fix proven load-bearing.** Reverting just the `normalizeSearchQuery` call in `onChange` reddens the paste test (1 failed / 9 passed at the time); restoring it returns the suite to green.

Verdict: no P1 or P2 issues. The two P3-class edge cases found during the pass (plus-addressed aliases, empty `mailto:`) were already handled correctly by the implementation and are now pinned by tests.

PR #6421, opened 2026-07-27
Show gross affiliate revenue on the affiliated products page

gumclaw:
## What

The revenue stat on the affiliated products page (`/products/affiliated`) no longer adjusts for partial refunds on the underlying purchases. It now reports gross paid affiliate credits via a new `User#affiliate_credits_gross_sum_total`.

`User#affiliate_credits_sum_total` — which does the adjustment — is unchanged and still used everywhere it affects money movement or period reporting.

## Why

That adjustment was the dominant cost of the page. It joins every one of the affiliate's credit rows to `purchases` just to read one boolean, `stripe_partially_refunded`, which no index can serve after the join. So MySQL fetches each joined purchase row from the clustered index, and the cost scales with the affiliate's entire earning history even though almost every row is then discarded — partial refunds are rare.

Measured on the production read replica for the affiliate with the largest credit history (46,392 credits, of which exactly 6 have a partial refund):

| affiliate credit count | old (adjusted) | new (gross) |
|---|---|---|
| 46,392 | 472 ms | 19 ms |
| 6,151 | 3,226 ms | 15 ms |

The plans show why, and are the more durable evidence than the timings:

```
OLD  affiliate_credits  ref     key=index_affiliate_credits_on_affiliate_user_id  rows=86780  Using where
     purchases          eq_ref  key=PRIMARY  rows=1  filtered=10.0  Using where     <- per-row fetch, 86,780x

NEW  affiliate_credits  range   key=idx_affiliate_credits_on_user_and_balances_and_amount
                                Using where; Using index                            <- covering, no row fetch
```

The new shape is fully covered by the composite index added for this page in #6020: equality on `affiliate_user_id`, the three balance-id checks, and `amount_cents` all come out of the index.

Sentry had this at roughly 70% of the request's span time in the two slowest post-deploy samples (`GUMROAD-H8`, 445 events / 79 users). It is the largest remaining piece of the page after #6300 moved the lifetime global-affiliate sum off the request path.

Tracking issue: antiwork/gumroad-private#1428

### One deliberate behavior change

The displayed number changes for any affiliate who has had a partial refund. This is intentional and makes the page self-consistent: the per-product Revenue column in the table below the stat has always been an unadjusted `SUM(affiliate_credits.amount_cents)`, so previously the headline stat could disagree with the sum of the rows underneath it. The before/after screenshots show exactly this — before, the stat reads $29.25 against rows totaling $24.25; after, both are $24.25.

The stat's own description in the UI already says "Your gross sales from all affiliated products", which the gross figure matches more closely than the adjusted one did.

## Before / After

Same fixtures in both columns: three paid credits of $10.00, $6.25, and $8.00, where the $8.00 one is against a partially refunded purchase with a $3.00 affiliate partial refund recorded.

|  | Before | After |
|---|---|---|
| **Desktop (1280px)** | before-desktop | after-desktop |
| **Mobile (iPhone)** | before-mobile | after-mobile |

Revenue goes from $29.25 to $24.25, matching the $16.25 + $8.00 in the table. Nothing else on the page changes. This page has no dark-mode variants, so light mode is the only applicable theme.

Walkthrough video covering the before state, the after state, a scroll through the page, and the mobile viewport: [pr-6421-walkthrough.mp4](https://github.com/antiwork/gumroad/raw/perf/affiliated-page-gross-revenue-stat/qa-media/pr-6421-walkthrough.mp4)

## Performance

EXPLAIN plans and timings from the production read replica, three uncached runs per shape plus a smaller-affiliate scaling datapoint:

benchmark

## QA steps

1. Sign in as a user who is an affiliate for at least one product and has at least one affiliate credit whose purchase was partially refunded.
2. Go to Products, then the Affiliated tab.
3. The Revenue stat should equal the sum of the Revenue column in the table below it.
4. Confirm the other three stats (Sales, Products, Affiliated creators) and the product table are unchanged.
5. Payouts are unaffected: an affiliate's payable balance still nets out partial refunds, since `affiliate_credit_cents_for_balances` continues to use the adjusted sum.

## Test Results

New spec asserting the gross stat ignores partial refunds while the adjusted sum still nets them out, covering both the credit that has an `AffiliatePartialRefund` row and the one that does not. Existing caching specs updated to the new method.

spec run

```
bundle exec rspec spec/presenters/affiliated_products_presenter_spec.rb \
  -e "the total revenue stat" -e "revenue stats caching"
6 examples, 0 failures
```

Nine examples in the sorting and pagination groups of that file fail locally on this machine, but they fail identically on unmodified `origin/main` in a worktree — a local headless Chrome exhaustion issue (`Chrome instance exited`, via the dispute-evidence receipt screenshot service), not related to this change. They pass in isolation and the failure count varies run to run.

---

## AI disclosure

Written by `claude-opus-5` in Hermes Agent. It was given the tracking issue, the direction on it to simplify the page to gross revenue rather than optimize the adjusted sum, and instructions to build the change with specs, measure the query cost against the production read replica, and capture before/after UI evidence.


gumclaw:
Self-review, re-reading CONTRIBUTING.md against the diff rather than from memory.

**Scope.** Two production lines change: the presenter calls a new method, and that method is added. 76 insertions, 9 deletions, well under the size guidance. No stray files — I removed the two throwaway capture specs I used for the media, and `git diff origin/main...HEAD --stat` lists only the two app files, the presenter spec, and `qa-media/`.

**The one thing I want a reviewer to look at deliberately** is that this changes a displayed number, not just a query plan. I checked every caller of the adjusted sum before touching anything:

- `affiliate_credit_cents_for_balances` — payout balances. Untouched, still adjusted. This is the one that must stay correct and it does.
- `affiliate_credits_sum_for_credits_created_between` → `last_weeks_sales` — period earnings reporting. Untouched.
- `AffiliatedProductsPresenter#cached_total_revenue` — the dashboard stat. The only caller I switched.

So no money movement changes. If a reviewer wants one assertion to check, it is that `affiliate_credit_sum_from_scope` still has exactly the callers it had before, minus the presenter.

**A rollout detail worth flagging.** I deliberately kept the existing cache key (`affiliated_products/total_revenue/<id>`) rather than versioning it. That means for up to `STATS_CACHE_TTL` (5 minutes) after deploy, an affiliate with a warm cache entry still sees the old adjusted number. I think that is fine — the value is a dashboard stat with a documented 5-minute staleness window either way, and versioning the key would stampede every affiliate's uncached sum at deploy time for no real benefit. Calling it out because it is a choice, not an oversight.

**On the naming.** I added a separate method rather than a keyword argument on the existing one, because the two numbers mean genuinely different things and I would rather a future caller have to pick one explicitly than default into the wrong one. The comment on the new method spells out which is which and why, per the repo's guidance about writing comments for someone on day one.

**Evidence.** Before/after are captured from the running app on identical fixtures, so the $29.25 → $24.25 difference is the change itself and not a fixture difference. Both numbers are printed by the capture run, which is how I know the after value ($24.25) equals the table rows and the adjusted sum is still $29.25 underneath. The page has no dark-mode variants, so light mode is the only applicable theme — I checked `AffiliatedPage.tsx` for `dark:` classes rather than assuming.

**Performance numbers are measured, not estimated.** Three uncached runs per shape against the production read replica, plus a smaller-affiliate scaling datapoint, plus the actual `EXPLAIN` output. The plans are the part I would trust over the timings: `Using where; Using index` versus an `eq_ref` per-row fetch on `purchases` is the structural reason this is faster, and it does not depend on cache state at measurement time.

**Known gap.** Nine examples in the sorting/pagination groups of the presenter spec fail on my machine, but they fail identically on unmodified `origin/main` in a separate worktree, the failing count varies between runs of the same tree, and the error is `Chrome instance exited` from the dispute-evidence receipt screenshot service. That is local browser exhaustion, not this change. CI is the authority on it and I have not tried to paper over it.

gumclaw:
Closing as a duplicate of #6420, which is the same change done slightly better.

I opened both within a minute of each other while working gumroad-private#1428, which was my mistake — they are not two deliberate alternatives for someone to choose between. Same three files, same root cause, and the new method is byte-identical in behaviour (`affiliate_credits.paid.sum(:amount_cents).to_i`); only the name differs (`affiliate_credits_gross_sum_total` here vs `affiliate_credits_total_revenue_cents` there).

Why #6420 survives rather than this one:

- **It removes the caching layer that this PR keeps.** The whole point of the change is that the stat becomes a single indexed sum served from `idx_affiliate_credits_on_user_and_balances_and_amount` in single-digit milliseconds. This PR still routes it through `cached_total_revenue` / `STATS_CACHE_TTL = 5.minutes`, so it pays a Rails.cache round-trip and serves up-to-5-minute-stale numbers to protect a query that no longer needs protecting. #6420 deletes `STATS_CACHE_TTL` and `cached_total_revenue` and calls the method directly, which is the correct end state.
- Consequently #6420 is a net simplification (`+56/-44`) where this is additive (`+76/-9`).

What I'm carrying over from here before closing, so nothing is lost: this branch had the fuller evidence set (7 media files vs 2, including the mobile shot and the walkthrough video). I'll move the mobile screenshot and walkthrough onto #6420 rather than leave it with thinner evidence than the PR being closed.

Nothing else here is unique — the explanatory comment on why the dashboard figure deliberately skips the partial-refund adjustment exists in both.

PR #6422, opened 2026-07-27
Make the purchases (email, link_id) index migration a documented no-op

ershad:
## What

Turns migration `20261206000004` (`AddEmailAndLinkIdIndexToPurchases`, from #6123) into a documented no-op, and removes `index_purchases_on_email_and_link_id` from `db/schema.rb`.

No runtime code changes. The diff is a migration body replaced by a comment, plus one line of `schema.rb`.

## Why

**The index was never actually created in production.** The version is recorded in `schema_migrations`, but the index does not exist. `purchases` schema changes run through Alterity, which hands every ALTER to pt-online-schema-change, and pt-osc creates a trigger set named after the table it is altering. `purchases` was still carrying an abandoned pt-osc trigger set of exactly those names, left by an earlier run that was killed without cleanup, so pt-osc failed on "Trigger already exists" within seconds and left no trace in the database — a migration that reported success and produced nothing. Reproduced with the same pt-osc version and flags; the leftovers have since been removed, so `purchases` DDL works again. Full investigation in gumroad-private#1417.

**It is not worth building now.** The per-seller probe loop it was written for was replaced in #6206 by a single batched prefetch per request (`Purchase::SellerPostProbeBatch`):

```sql
SELECT seller_id, id, link_id FROM purchases WHERE email = ? AND seller_id IN (...)
```

`link_id` is selected but never filtered on, so `(email(191), link_id)`:

- adds no selectivity over the `index_purchases_on_email_long` this query already uses — same leading column, second column unused in the predicate;
- cannot cover the read, because `seller_id` is not in it, so every matched row still needs a primary-key lookup.

There is no plan the optimizer would prefer it for, and building it would mean a full pt-online-schema-change copy of one of the largest tables in the database. `docs/migrations.md` and CONTRIBUTING both list index changes on `purchases` as off limits. Measurements behind that call are in gumroad-private#1417.

The alternative — deleting the `schema_migrations` row and rebuilding the index out of band — buys a plan MySQL has no reason to choose, at that cost.

**Why also `schema.rb`:** it is what fresh development and CI databases load. Leaving the index there gives every local database an index production does not have, which is exactly the divergence that makes a query look fine locally and behave differently in production.

The migration stays as a no-op rather than being deleted so the version keeps recording cleanly across environments — the same treatment `20261201000006` got after its own `purchases` ALTER hung a deploy.

## Before / After

No user-visible surface, so there is nothing to record. The behavioural difference is in the schema, not the product:

| | Before | After |
|---|---|---|
| Production `purchases` | no `index_purchases_on_email_and_link_id`; `schema_migrations` says otherwise | unchanged, and the file now says so too |
| Fresh dev/CI database | has the index | matches production |
| Next `db:migrate` | would never build it (version already recorded) | explicitly a no-op, with the reason in the file |

## Test Results

`bundle exec rubocop` clean on the changed migration; `ruby -c` clean. I could not run the suite or `bin/test-confidence` on this machine — the local MySQL is not running and `ANTHROPIC_API_KEY` is unset — so CI is the check here. Nothing in `spec/` references the index; the two specs covering this query path (`spec/models/purchase/seller_post_probe_batch_spec.rb`, `spec/models/concerns/with_filtering_spec.rb`) assert on results, not on query plans, and are unaffected by an index that production never had.

Deploy note: changing a file under `db/migrate` changes that directory's tree hash, so the next deploy runs the migration job instead of taking the fast path. It will find nothing pending and exit clean — one slower deploy.

---
Written with Claude Opus 5.

### DIFF
--- a/db/migrate/20261206000004_add_email_and_link_id_index_to_purchases.rb
+++ b/db/migrate/20261206000004_add_email_and_link_id_index_to_purchases.rb
@@ -1,32 +1,50 @@
 # frozen_string_literal: true
 
+# Intentionally a no-op.
+#
+# This originally added a composite `(email(191), link_id)` index to `purchases` to
+# speed up the mobile library's "hasn't bought X" existence probes. It never took
+# effect in production, and it is no longer worth building. Both halves are worth
+# writing down: the first is a trap, and the second is the reason not to simply
+# retry it.
+#
+# It never ran. Production recorded this migration's version, but the index was
+# never created. `purchases` schema changes go through the Alterity gem, which hands
+# every ALTER to pt-online-schema-change, and pt-osc creates a trigger set named
+# after the table it is altering. `purchases` was still carrying an abandoned
+# trigger set of exactly those names, left behind by an earlier pt-osc run that was
+# killed without cleanup, so pt-osc failed on "Trigger already exists" within
+# seconds and left no trace in the database — a migration that reported success and
+# produced nothing. Those leftovers have since been removed and `purchases` DDL
+# works again; the full investigation is in gumroad-private#1417.
+#
+# It is no longer worth building, because the query it was written for is gone. The
+# per-seller probe loop it targeted (whole-history scans for mega-sellers, because
+# MySQL drove off `index_purchases_on_seller_id` and then filtered on link_id/email)
+# was replaced in gumroad#6206 by a single prefetch per request in
+# Purchase::SellerPostProbeBatch, which reads:
+#
+#   SELECT seller_id, id, link_id FROM purchases WHERE email = ? AND seller_id IN (...)
+#
+# `link_id` is selected there but never filtered on, so a `(email, link_id)` index
+# adds no selectivity over the `index_purchases_on_email_long` this query already
+# uses, and it cannot cover the read either, because `seller_id` is not in it. There
+# is no better plan for the optimizer to pick, so the index would cost a large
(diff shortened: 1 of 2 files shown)


ershad:
Self-review:

- **Why the migration stays instead of being deleted.** The version is already recorded in production, so deleting the file would leave `schema_migrations` carrying a version with no migration behind it, and every other environment would keep the index. Keeping it as a no-op is what `20261201000006` does for the same reason, and it puts the reasoning where the next person to consider this index will actually look.
- **`up`/`down` instead of `change`.** An empty `change` would be ambiguous to read; two explicit empty methods say "this deliberately does nothing" and stay reversible. `disable_ddl_transaction!` is dropped since there is no DDL left to run outside a transaction.
- **The `schema.rb` line is the load-bearing half of this diff.** Without it, a fresh `db:schema:load` keeps creating an index production does not have, and the divergence quietly outlives the migration file.
- **Not verified locally**: no suite run on my machine (MySQL down, `ANTHROPIC_API_KEY` unset). Relying on CI. The change has no runtime behaviour, and nothing in `spec/` references the index by name.
- **Scope check**: this does not touch production schema, does not delete the `schema_migrations` row, and does not attempt to build the index. If we ever do want an index for this path, the batched prefetch filters on `email` and `seller_id` — so `(email, link_id)` would be the wrong shape regardless.

PR #6430, opened 2026-07-27
Refuse a schema change on a table still carrying abandoned pt-osc leftovers

gumclaw:
## What

Adds a pre-flight check that refuses an online schema change when the target table is still carrying the triggers and shadow table of an earlier `pt-online-schema-change` that was abandoned without cleanup. Wired into Alterity's `before_command` hook, so it covers every migration that goes through pt-osc.

The check is diagnostic only. It never drops or alters anything: it names each leftover artifact it found and the order they have to be removed in, then raises.

## Why

Refs the internal migrations tracker.

pt-osc names its triggers after the database and table with nothing to make them unique per run:

```
pt_osc_<database>_<table>_ins / _upd / _del
```

So a run that is killed without cleanup leaves triggers occupying exactly the names the *next* run on that table needs, and every later schema change to that table fails on `Trigger already exists`.

What makes it worth catching up front rather than letting pt-osc fail is how the failure presents. pt-osc drops its own scratch table on the way out and leaves the pre-existing triggers untouched, so it exits in seconds having changed nothing — and the error text describes the symptom without saying what to do about it. On `purchases` this cost a hung deploy and a migration recorded in `schema_migrations` that had never run, and the leftovers sat there blocking all DDL on the table (including the documented out-of-band pt-osc escape hatch, which fails the same way) until someone found them by hand.

Failing here turns that into an actionable refusal with the artifact names in it.

## Before / after

No UI surface — this runs during `rake db:migrate`. Per CONTRIBUTING.md, here is the walkthrough video of the relevant functionality, driven against real MySQL triggers and shadow tables: a clean table passing, the leftovers being created, what pt-osc's failure looked like before this change, the refusal with its diagnostics, the Alterity wiring, proof that nothing was modified, and the table working again once the leftovers are cleared.

https://github.com/antiwork/gumroad/blob/fix/migration-pt-osc-preflight/qa-media/pr-6430-walkthrough.mp4?raw=true

| | Before | After |
|---|---|---|
| Migrating a table with leftover pt-osc triggers | pt-osc exits non-zero in seconds with `Trigger already exists`, having changed nothing and leaving no trace; the operator has to work out from that what to look for | Migration refused before pt-osc runs, with every leftover trigger and shadow table named, and the removal ORDER spelled out |
| Migrating a clean table | Runs | Runs — one `information_schema` lookup per pt-osc invocation, at migration time only |
| A neighbouring table has leftovers (`purchase_refunds` vs `purchases`) | n/a | Not blocked; lookups are anchored on the exact table name |

## QA steps

The check only fires against a table that has real pt-osc leftovers, which is what the specs construct. To exercise it by hand in development:

1. `bin/rails db` and create a table plus a leftover trigger for it:
   ```sql
   CREATE TABLE demo_table (id INT PRIMARY KEY);
   CREATE TRIGGER pt_osc_gumroad_development_demo_table_ins
     AFTER INSERT ON demo_table FOR EACH ROW SET @noop = 1;
   ```
2. Write a throwaway migration that adds an index to `demo_table` and run `bin/rails db:migrate`.
3. It fails before pt-osc runs, with `[PtOscLeftoverCheck] Refusing to run an online schema change on \`demo_table\`` and the trigger named.
4. `DROP TRIGGER pt_osc_gumroad_development_demo_table_ins;` and re-run — the migration proceeds.

## Test results

19 examples, 0 failures, run locally against MySQL. Rubocop clean on all four files. `bin/test-confidence` reached 99% ("safe to commit").

test run

(Also committed to the branch at `qa-media/pr-6430-pt-osc-preflight-tests.png`.)

The specs create **real** triggers and **real** shadow tables in the test database rather than stubbing the `information_schema` queries. The check exists because of how MySQL names and reports these objects, so a stubbed version would only assert that the code calls the methods it calls, and would keep passing if the naming convention or the lookups were wrong.

Notable cases, each of which is a way to get this wrong:

- **A partial trigger set is enough.** pt-osc creates the three triggers one at a time and can die between them, so any one of them blocks the next run. The check must not require all three.
- **Neighbouring tables must not block each other.** A prefix or `LIKE` match would let `purchases`'s leftovers block a migration on `purchase_refunds`, and blocking a healthy migration is its own outage. Both directions are covered.
- **`<table>_new` with no leading underscore is not a pt-osc artifact** — pt-osc always prefixes at least one underscore — so a real application table by that name is not read as a leftover.
- **Several generations of leftovers are reported together.** pt-osc prepends another underscore for each name already taken, so `_purchases_new` and `__purchases_new` can both exist; a message naming only the first sends the operator back for a second pass.
- **The check does not modify anything** — asserted directly, since the whole point is that clearing the leftovers stays a human decision.

Two guards were verified as load-bearing by breaking them and confirming the suite goes red, rather than trusting the green run:

| Mutation | Result |
|---|---|
| Trigger lookup relaxed from exact names to a prefix `LIKE` | 2 examples red (the two neighbouring-table cases) |
| Shadow-table lookup relaxed from an anchored regexp to a loose `LIKE` | 1 example red (`<table>_new` with no leading underscore) |
| Table-name parse relaxed to a bare `t=` match | 2 examples red (see below) |

That third one is not hypothetical — it was the first version of this change, and the wiring spec caught it. Alterity's command carries `--recursion-method 'dsn=D=percona,t=replicas_dsns'` *before* the `D=<db>,t=<table>` argument that names the table being altered, so a bare `t=` match reads `replicas_dsns`, checks the wrong table, and reports the migration clear while the real table is blocked. The parse is now anchored on the whole `D=…,t=…` argument, and there is a regression spec asserting the recursion-method DSN is not mistaken for the altered table.

The spec suite drives Alterity's real `before_command` hook rather than calling the check directly, so a check that exists but is not wired in would fail.

---

## AI disclosure

Written by Claude Opus 5 (`claude-opus-5`) running as the Hermes agent. It read the Alterity gem and initializer to establish how pt-osc is invoked and where a pre-flight could hook in, wrote the check and both spec files, ran them against local MySQL, and mutation-tested the three anchoring guards. The wrong-table parse bug above was found by its own wiring spec and fixed before this PR was opened. No production system was touched.


gumclaw:
Pushed `16bb005c0` — comment-only change to the source, plus two body fixes.

**Three internal tracker references removed from source comments.** This is a public repository, and `pt_osc_leftover_check.rb` and the Alterity initializer both cited an internal issue number — twice in the service (once in the class comment, once inside the operator-facing failure message, which is the worse of the two since it would have been printed to whoever hit the guard). The surrounding context is preserved; only the identifiers are gone. A fourth reference was in the PR description and is gone too.

Worth noting for the reviewer: `closingIssuesReferences` on this PR reads empty, because none of these were closing keywords. So an API-level check would have reported it clean — a raw grep of the diff and the body is what surfaced them.

**The test-results image now points at the committed `qa-media/` copy** rather than an imgbb upload. Same image, byte-identical (137,422 bytes both sides), but CONTRIBUTING wants it in-repo, and the file was already committed as `qa-media/pr-6430-pt-osc-preflight-tests.png` while the body cited the external copy — so the committed evidence was orphaned and the body depended on a third-party host. Pinned to commit `16bb005c018c930253854b07ca4c79f1f7c37cc1` rather than the branch, so it survives branch deletion at merge.

No behaviour change in this commit: `git diff 1f157deee..16bb005c0` is comments only, rubocop clean, and the guard logic and specs are untouched. CI is re-running on the new head.

**One thing I checked rather than assumed, since this PR adds a guard.** A guard that can never fire is worse than no guard, because it closes the issue on paper. This one is wired through Alterity's real `before_command` hook, and `spec/initializers/alterity_spec.rb` drives that hook — building the command from `Alterity.send(:config).command.call(...)`, the real template — rather than calling `PtOscLeftoverCheck` directly. So a check that existed but was never invoked would fail the suite. The table-name parse is the part that would plausibly be wrong (the command also carries `--recursion-method 'dsn=D=percona,t=replicas_dsns'` *earlier* in the string, so a loose `t=` match would silently inspect the replica-DSN bookkeeping table and report the real table clean), and it is matched as the whole `D=<db>,t=<table>` argument with that hazard written down at the call site.

gumclaw:

**The pieces I'd push back on if someone else wrote this**

*Fail-open on an unparseable command.* If Alterity's command template ever changes shape, the table-name regexp matches nothing and the check silently does not run. I chose that over the alternative deliberately: refusing a migration on a table we could not identify would block healthy migrations on every table at once, and a check that quietly stops working is recoverable while a deploy pipeline that refuses everything is not. It is asserted as behavior (`skips the check rather than guessing when the command has no table`) so it is a decision rather than an accident, but it is the weakest point in the change and worth knowing about.

*`Alterity.send(:config)` in the specs.* `config` is private on the gem's singleton, and reaching past that is not something I'd normally do. The alternative is asserting against a hand-written command string, which is exactly what would have hidden the `replicas_dsns` bug — the spec has to see the string the gem really builds. Testing the real seam won out over politeness about visibility.

*Interpolated SQL rather than bound parameters.* Every interpolated value goes through `connection.quote`, and the inputs are a table name from our own migration and a database name from the connection, so there is no user-controlled path here. `information_schema` lookups with an `IN` list and a `REGEXP` don't fit the prepared-statement helpers cleanly. Flagging it because "interpolated SQL" should always get a second look, not because I think it's wrong here.

**Things I checked because CONTRIBUTING.md says to**

- **`purchases`/`users` are frozen tables** (Code patterns). Nothing here migrates anything — it's a read-only guard that runs *before* a migration. No schema change in the diff.
- **Tests must fail when the fix is reverted.** Not taken on faith. I broke each of the three anchoring guards in turn and confirmed the suite goes red: trigger lookup relaxed to a prefix `LIKE` → 2 red; shadow-table regexp relaxed to a loose `LIKE` → 1 red; table parse relaxed to a bare `t=` → 2 red. A green suite over guards like these proves very little on its own.
- **No "should" in test descriptions**, descriptive names, grouped, independent. The independence one bit me: schema objects are not rolled back by DatabaseCleaner, so the first version leaked a shadow table between examples and the suite was order-dependent. Cleanup is now over-broad and unconditional, and I verified it with three `--order random` runs plus the specific interleaving that originally failed.
- **Comments earn their place.** They're heavier than usual for a ~140-line service, and that's on purpose: every non-obvious line here encodes a fact about pt-osc's naming behavior that is not visible from the code, and the next person to touch this will not have that context. The `_purchases_new` → `__purchases_new` stacking and the drop-triggers-before-table ordering are the two that would otherwise have to be rediscovered.
- **No new file in `app/modules/`** — it's a service in `app/services/`.
- **Named constants** — `TRIGGER_SUFFIXES` rather than an inline `%w[ins upd del]` in two places.

**What I'd want a reviewer to actually look at**

The exact-name anchoring, in `find_leftover_triggers` and `find_shadow_tables`. Everything else here fails safe: if the check misses leftovers, you get today's behavior back. If it matches too broadly, it blocks migrations on healthy tables — that's the only way this change makes things worse than not having it, so it's where a second pair of eyes is worth most. The `purchases`/`purchase_refunds` case is covered in both directions, but the argument for exactly *why* an anchored match is required is the thing to check rather than the tests passing.

Also worth a glance: whether refusing at `before_command` is the right seam at all, versus a separate pre-deploy check. `before_command` means the refusal happens with the migration transaction already in flight, which is the correct failure (nothing has been altered — the deploy aborts and the lock is released) but is later than a dedicated pre-flight phase would be. I went with the hook because it cannot be bypassed by anything that goes through Alterity, including an out-of-band run, and a separate phase can.

**Not covered here:** whether a leftover trigger set can exist with no shadow table or vice versa in production. The check handles either independently and the specs cover both, but the real-world combination came from one observed case, so there may be shapes I haven't anticipated.

gumclaw:
@claude review once

gumclaw:
Green and ready for you — CI is 76/76 at head, no bot findings, and evidence is in the body.

Assigning this one to you rather than merging it myself. It changes the failure behavior of every `purchases`/`users` migration, and you own both the cleanup and the operational judgement about what should happen when a migration is refused. Two specific things worth your eye:

**The fail-open choice.** If Alterity's command template ever changes shape, the table-name parse matches nothing and the check silently stops running. I picked that over the alternative because refusing a migration on a table we could not identify would block every table at once, and a check that quietly stops working is recoverable while a pipeline that refuses everything is not. If you'd rather it fail closed and be loud, that's a one-line change and your call.

**Whether `before_command` is the right seam.** The refusal lands with the migration already in flight — the correct failure (nothing altered, deploy aborts, lock released) but later than a dedicated pre-deploy phase. I went with the hook because it can't be bypassed by anything routing through Alterity, including the out-of-band pt-osc path. A separate phase can be bypassed, which seemed like the wrong trade for a guard whose whole job is to stop a specific mistake from recurring.

The exact-name anchoring in `find_leftover_triggers` / `find_shadow_tables` is the part I'd most want checked. Everything else here fails safe — if the check misses leftovers you get today's behavior back. Matching too broadly is the only way this makes things worse, since it would block migrations on healthy tables. Both directions of the `purchases` / `purchase_refunds` case are covered, and I verified the anchoring is load-bearing by relaxing it to a prefix match and confirming specs redden, but the argument for why anchoring is required is worth reading rather than the tests passing.

One correction landed while this was open: a sibling stripped the `gumroad-private#1417` references out of the source comments, which was right — this is the public repo and the raised error message is user-facing output. I've kept the body consistent with that.

Companion PR for your other two items is antiwork/gumroad-deployment#47.

gumclaw:
Closing this as a duplicate of #6428 — same guard, same driving incident, opened 8 minutes apart by two parallel sessions of mine. @ershad you were assigned to both; only #6428 needs your time, and `awaiting-human` is coming off here so it stops showing in your queue.

I want to be explicit that I did **not** pick the survivor by scope or by which was tidier, because the two disagree on the one thing that actually matters.

**Where this PR is wrong, and it is pinned by a passing spec.** `assert_absent!` here raises whenever *either* artifact is present:

```ruby
return if leftover_triggers.empty? && shadow_tables.empty?
raise LeftoversPresent, failure_message(leftover_triggers, shadow_tables)
```

That refuses a migration over a leftover **shadow table alone**. Your reproduction on the tracker issue is the counter-evidence: pt-osc created `__purchases_new` and altered it successfully *with* `_purchases_new` already present, and only failed at "Creating triggers…". So triggers are what break the next run; a stray shadow table is dead storage that pt-osc renames around. Refusing there would block a schema change nothing was blocking — a deploy outage in the name of preventing one.

The reason this needed a code read rather than a CI read is that the behaviour is **asserted as correct** by `it "raises when only the shadow table is left behind"`. The suite is green *because* the wrong semantics are pinned, so nothing was ever going to surface it. #6428 gets this split right: only triggers raise, and shadow-only leftovers print a warning naming the storage.

Second gap: there is no bypass here. #6428 has `ALLOW_PT_OSC_LEFTOVERS`, which matters for the one legitimate case — deploying while a pt-osc run genuinely is in flight. Without it, an in-flight migration on a big table makes every deploy refuse until it finishes, and the only escape is editing the check.

**What this PR did better, and I've carried it over as a note rather than a silent loss.** Two things, recorded on #6428 so they aren't buried in a closed thread:

- It scopes per-table, hooking `before_command` where the table being altered is known, rather than `before_running_migrations` where it isn't. #6428's fleet-wide scan is the right shape for its hook, but it does mean leftovers on table A can refuse a migration touching only table B.
- The table-name parse here documents a real trap worth keeping: the pt-osc command also carries `--recursion-method 'dsn=D=percona,t=replicas_dsns'`, which appears *earlier*, so a loose `t=` match reads the replica-DSN bookkeeping table and then reports it clean while the real table is blocked. There is a spec for exactly that.

No code from this branch is lost that #6428 doesn't already cover, and nothing here was merged. Reopen this instead if you'd rather have the per-table shape as the base — I'd then port the trigger/shadow split and the bypass onto it, and it's the same amount of work either direction.

PR #6432, opened 2026-07-27
Stop stale cross-product file embeds from blocking product saves

gumclaw:
## What

A pre-existing cross-product `fileEmbed` whose file has since been soft-deleted no longer blocks every future product save.

The original ownership validation from #5416 is still the rule:

- An **alive** foreign file still fails validation.
- A newly submitted **dead** foreign file still fails validation unless the server can prove it came from stored content in this product.
- Only an invisible, soft-deleted foreign embed already present in a stored page, or copied/moved from a server-verified stored source page in this product, is eligible for repair.

The repair now happens explicitly at product-save boundaries rather than in an Active Record callback. The dashboard editor, variant API, trusted content-copy paths, and product duplication opt in where they persist content.

## Why

The seller in antiwork/gumroad-private#1413 had two products they could not edit. A file from another product had been embedded before ownership checks shipped, then soft-deleted. The editor rendered no removable node, but every save still failed:

```text
Validation failed: File embeds reference files not belonging to this product
```

Removing that dead node is safe because the file delivers nothing. Removing an alive foreign embed is not safe, so that case still fails.

### Server

- Existing pages compare the submitted description with the committed description and remove only stored dead foreign file IDs.
- New destination pages require a source page already owned by this product.
- Current clients send `source_id` plus `rich_content_provenance_version: 1`.
- For tabs opened before provenance support:
  - Reusing one stored page ID under a different owner scope proves a move; the exact source row is added to the save-contract deletion request.
  - A repeated submitted page ID cannot be represented by the global response mapping, so the server returns `409` before cleanup or mutation and tells the seller to reload.
  - A new destination ID can remove only submitted dead IDs that also exist in alive stored pages owned by this product.
- Source pages are resolved and ownership-checked in batches under the product lock.
- API migration compares variant pages after removing only proven stale dead embeds, so variants with the same visible content can switch to shared content.
- Alive foreign files, unproven dead IDs, and source pages from another product remain validation failures.
- Error details name same-seller files only when safe; cross-seller and collaborator-sensitive cases stay opaque.

### Editor

- The save response returns canonical page/variant mappings and the exact file IDs removed from each saved page.
- Page and variant mapping namespaces are separate, and mapping chains resolve transitively.
- Product state and the mounted ProseMirror editor both reconcile the cleanup.
- Mounted cleanup uses granular deletes, preserves unrelated edits and selection, and stays out of undo history.
- Moves, copies, deletes, and variant removals made while a save is in flight are rebased onto the response instead of being overwritten.

## Evidence

[Walkthrough video](https://raw.githubusercontent.com/antiwork/gumroad/3b8b6df391b45e661c5e1b23b69c02be97d3a468/qa-media/pr-6432-walkthrough.mp4) (56s, 1920×1080)

The matching reproducible test script is committed as [`pr-6432-walkthrough-script.sh`](https://raw.githubusercontent.com/antiwork/gumroad/3b8b6df391b45e661c5e1b23b69c02be97d3a468/qa-media/pr-6432-walkthrough-script.sh).

Final test evidence

Evidence boundary: this is code inspection and automated-test proof, including an exact-main regression run. It does not claim a live seller-browser round trip or production validation.

## Validation

- 9 focused controller regressions: 46 assertions, 0 failures
- 52 Vitest cases across `product_edit` and `product_save_contract`
- Relevant model, dashboard/API, product duplicator, variant updater, save-contract, editor-revision, and policy suites
- RuboCop clean on all touched Ruby and Ruby test files
- ESLint clean on all touched TypeScript/TSX files
- `git diff --check`
- `shellcheck qa-media/pr-6432-walkthrough-script.sh`
- Walkthrough script completed end to end
- MP4 full decode plus frame-by-frame montage inspection
- `bin/test-confidence`: 99%, 16 planned test files passed; seven browser-suite failures reproduced at the merge base and classified as pre-existing; the 100% remainder was not run
- Final default-Codex autoreview against current `origin/main`: no accepted/actionable findings

## Not in this PR

- No production data cleanup was run.
- No PR merge.
- No claim that a browser interaction or production round trip was exercised.


gumclaw:
@gianfrancopiana handing this to you — you own the product-editor lane, and the two prior fixes to this content-page save path went through you.

**Green at the current head.** `edcc4a4` is the tip: `ci/green` **success** and `buildkite/gumroad` **success`, both Tests runs (push + pull_request_target) completed successfully. Nothing pending, nothing red.

**What to push on.** The load-bearing decision is that only **soft-deleted** foreign embeds get dropped, while **alive** ones still fail validation exactly as #5416 intended. Dropping everything would have been a smaller diff and made the error disappear — I didn't, because an alive foreign file is visible in the seller's editor and may be content they meant to deliver, so deleting it to clear a validation error destroys their work. The dead case is safe on a specific factual basis (a soft-deleted `ProductFile` delivers nothing to a buyer, so the node was already dead content) and that basis does not extend to alive files. `still rejects an ALIVE foreign embed on the same page` pins a page carrying one of each, which is where a looser implementation would quietly widen. If you think the line belongs somewhere else, that is the conversation.

**Why I'm confident the specs mean something.** Reverting `app/models/rich_content.rb` to `origin/main` while keeping the specs reddens all six changed/new examples, with `Expected false to equal true` on each save case — so they reproduce the seller's actual dead end rather than passing by construction. Both runs are in the committed screenshot, the reverted one included, so it's checkable rather than a claim.

**One deliberate scope limit.** The remediation is lazy — an affected seller recovers the next time they save. Rows nobody touches stay broken, so the fleet-wide one-time cleanup is still worth running and is left to a human; that's stated under "Not in this PR" rather than left implicit. Also left alone: `api/v2/variants_controller` raises a similarly-worded error from a different code path that isn't part of the seller-editor flow this was reported against.

**Not auto-merging.** This changes what a save does on a seller-facing write path, and the fix is a silent data mutation on their content — that wants a human's eyes on the alive/dead boundary before it ships, not a green checkmark. Merge is yours.

PR #6435, opened 2026-07-27
Extract a paste_into spec helper for the pasted-value pattern

gumclaw:
## What

Moves the pasted-value mechanics out of the `mailto:` search example in the Sales page spec and into a named `paste_into` helper next to `fill_in_color`, which already used the same technique. Also drops a redundant `index_model_records(Purchase)` from that example — the file's top-level `before` block already indexes and the example creates no further purchases.

No behavior change, no assertion change; the example still drives the same browser path.

## Why

The example landed with an inline `execute_script` that duplicated mechanics `fill_in_color` already carried, and the reason it has to go through the prototype's value setter rather than assigning `this.value` — that being what makes React notice the change — was not written down anywhere. The next spec that needs a paste would have had to rediscover it or, more likely, reach for `fill_in` and write a test that passes whether or not the code under test works.

Follow-up to #6434.

## Before / After

Not applicable — spec-support refactor, no user-facing surface.

## QA steps

Not applicable. The behavior this spec covers is unchanged; the QA path for it is on #6419.

## Test Results

The refactored example plus the pre-existing `allows searching the table` example from the same file, run together so the change is shown not to disturb its neighbour: **2 examples, 0 failures**.

test results

`rubocop` reports no offenses on either touched file.

---

## AI disclosure

Written by claude-opus-5 running as the Gumroad agent. Instructions given: update an end-to-end test for the pasted `mailto:` search-prefix change and merge it once CI is green. This PR carries the cleanup that came out of the pre-merge review of #6434 after that PR had already auto-merged on green.


gumclaw:
Merged as `6f627da90c80`.

**Why this merged without waiting for a fresh CI run.** The only commit after the last fully-green head (`1bfa3ff8d`: `ci/green` success, `buildkite/gumroad` success) was a rename of one PNG under `qa-media/` — verified as code-free with `gh api .../commits/<sha> --jq '[.files[].filename]'` → `["qa-media/pr-6435-test-results.png"]`, not from the commit message. Per the standing no-code-change rule, a redundant shard run on a screenshot rename isn't worth an hour.

The diff itself is spec-support only (`spec/support/capybara_helpers.rb` + one call site in `spec/requests/customers/customers_spec.rb`), no application code, and Greptile scored it 5/5 with no actionable findings.

**Two defects fixed before merging**, both from the open-queue evidence sweep and both invisible to CI and to reviewers:

| Defect | Fix |
|---|---|
| `qa-media/paste-into-helper-tests.png` had no PR number — violates CONTRIBUTING's `pr-<number>-<description>.<ext>` | renamed to `qa-media/pr-6435-test-results.png` |
| the body's raw URL was pinned to the **branch name**, which 404s permanently the moment merge deletes the branch | re-pinned to commit `1fe94aabe6fc` |

Both verified after merge: the raw URL still returns `200` (82,848 bytes) with the branch now deleted, because a commit-SHA URL reads the blob from git history. `closingIssuesReferences` is empty and the body carries zero `gumroad-private`/`gp#` references, so nothing was force-closed and nothing private leaked.

PR #6437, opened 2026-07-27
Fail closed when the Payment Element's currency can't match the intent's

gumclaw:
> **Back to draft — the root cause in this body is wrong.** Production data on all 57 affected orders shows the dominant direction is the opposite of what is described below (EUR/INR *intents* against USD-mounted Elements, on forced-currency-priced carts — not USD carts with EUR Elements), and the retry this guard triggers does not refresh the payment config, so it can loop. Full findings and the two options are in [this comment](https://github.com/antiwork/gumroad/pull/6437#issuecomment-5098616925). Rewriting once the direction is settled.

## What

Fails a client-confirm checkout closed when the Payment Element's mount currency cannot match the currency the PaymentIntent will be created in, instead of building an intent the buyer's ConfirmationToken can never confirm.

## Why

Stripe rejects confirming a ConfirmationToken minted on an Element of one currency against an intent of another — **synchronously, and with no `payment_failed` webhook**:

```
The provided currency (eur) does not match the expected currency (usd).
```

So the buyer hits a dead end at the final confirm step: no charge, no error we recorded, and the purchase sits `in_progress` until `FailAbandonedPurchaseWorker` sweeps it — indistinguishable from a buyer who closed the tab. **1,711 events across 204 users in two days** (Sentry `GUMROAD-12F`), sustained with no taper, and they are ordinary **card and Link** buyers rather than the forced-currency local methods this machinery was built around.

Root cause: both existing mount-currency guards in `PreparePaymentIntentService` require the *product* to be listed in a forced currency —

```ruby
return nil unless Checkout::BuyerCurrencyEligibility::FORCED_CURRENCY_PAYMENT_METHODS.value?(product_currency)
```

— but the buyer-currency **card** lane operates on **USD-priced** products by design and mounts the Element in the buyer's FX-quoted currency. For that cart both guards return nil, the intent is created in canonical USD, and the EUR token can never confirm it. The client already reports the right information (`payment_element_mount_currency`); the server was discarding it because the precondition was keyed to the wrong shape.

The mirror direction is the same defect from the other side: an EUR intent whose Element remounted to USD mid-session (`PaymentElementInput` remounts on currency change by design), producing a USD token against an EUR intent.

## How

One guard, checked after the presentment resolves (it determines the intent's currency) and before any intent is created. It fails with the existing `BUYER_CURRENCY_QUOTE_INVALID` code, which the checkout already handles by cancelling, re-fetching surcharges and re-running the display gates — so the retry mounts and confirms in one consistent currency.

Deliberately narrow: it fires only when the browser actually reported a mount currency **and** that currency differs from the intent's, so every checkout that agrees with itself is untouched. This converts an unrecoverable dead end into a clean, retryable failure; it does not attempt to *charge* the non-USD amount, which is the larger contract question below.

## Test results

`bundle exec rspec spec/services/order/prepare_payment_intent_service_spec.rb` → **82 examples, 0 failures**. Rubocop clean.

Four new specs: the reported USD-priced/EUR-mount case fails closed with the right error code, the mirror USD-token-vs-EUR-intent case fails closed without leaving presentment rows, and two negative controls (mount currency agrees; browser reports nothing) confirm today's path is unchanged.

⚠️ **Note for the reviewer:** this spec file has **pre-existing order-dependent flakiness** unrelated to this change. On `--seed 12345` the file fails ~9 examples with my commit *stashed*, and different subsets fail on different seeds; each one passes in isolation. Worth a separate look, but please don't read a red run as this PR's doing without checking the same seed on `main`.

## What this does NOT decide

The issue's open question stands and needs an owner's call: **should a non-USD mount currency create the intent in that currency, or should the Element never have mounted non-USD for this cart?** This PR stops the bleeding either way. If the answer is "honor the mount currency", the follow-up is to let the client-confirm lane carry a verified quote — a bigger change that touches the quote token contract.

Also unaddressed, and worth its own PR: 73 of 100 sampled `confirm_error` events are ordinary `card_error` declines sharing one fingerprint with this bug (~500/day of Sentry noise). I'm deliberately not narrowing that instrument unilaterally — it's load-bearing for an in-flight rollout tracked internally, and its width is exactly what surfaced this defect. Recommendation is on the internal tracker.

## AI disclosure

Written by Gumclaw (claude-opus-5). Root cause traced from Sentry issue `GUMROAD-12F` and the matching internal tracker entry.


gumclaw:
## Routing correction — one human, and it's @rmarescu

Dropping @nyomanjyotisa as a second assignee here. Two assignees on one PR reliably means neither treats it as theirs, and this one is squarely @rmarescu's: it's failing closed when the Payment Element's currency can't be matched at client-confirm time, i.e. the PaymentIntent/Payment Element surface he owns and has reviewed on every recent change to.

@nyomanjyotisa: nothing needed from you here — I've moved you to #6427 (the India e-mandate cap) instead, which is the charge-backend piece.

slavingia:
When would this happen?

gumclaw:
Rarely, and not in the shape this PR describes. I pulled the full Sentry population and then checked every affected order against production, and the story in the body is wrong in a way that matters before this merges.

## How often

Every event in `GUMROAD-12F` since it opened (2,593 events, 2,590 distinct orders, 24–28 July):

| Error | Events | Distinct orders |
|---|---|---|
| Ordinary card declines (`card_error`) | 2,310 | — |
| Currency mismatch | **57** | 57 |
| Payment-method-list mismatch | 23 | 23 |
| Everything else (`invalid_request_error`, API connection) | ~200 | — |

So the currency mismatch is 2.2% of the noise in that issue, one event per order (nobody retries into it), and it has essentially stopped on its own: 23 events on the 24th, 25 on the 25th, 8 on the 26th, 1 on the 27th, 0 on the 28th. The window lines up with the iDEAL relaunch, not with anything about card checkout in general.

## The direction is backwards

The body says a USD-priced cart mounted a EUR Element and the server then built a USD intent. That is not what happened in any of the 57.

I resolved all 57 orders in production and retrieved the real PaymentIntent for each:

- **All 57 carts are EUR-priced (54) or INR-priced (2) products.** None is USD-priced, so the buyer-currency card lane the body blames — which requires USD-priced products by definition — is not the lane these buyers were on.
- **56 of 57 intents were created in the product's own forced currency** (EUR or INR), i.e. the forced-currency machinery worked exactly as designed. One order got a USD intent.
- Reading Stripe's wording against those intents, `provided` is the **intent's** currency and `expected` is the **Element's**: the 54 EUR intents all report "provided (eur) … expected (usd)", the 2 INR intents report "(inr) … (usd)", and the single USD intent reports "(usd) … expected (eur)". That is 57/57 agreement, in both directions.

So the dominant case is the exact opposite of the one the PR leads with: the server correctly created a **EUR** intent and the browser's ConfirmationToken was **USD**. The mirror case the body treats as an afterthought is 54 of 57; the case it treats as the bug is 1.

## When it actually happens

`elements_options.currency` is computed once, server-side, when the checkout page renders, and it is never refreshed — the two Inertia partial reloads on that page ask for `["cart", "flash"]` and `["recommended_products"]`, never `checkout_payment`. On the client-confirm lane that render-time value *is* the mount currency and *is* what the browser reports back as `payment_element_mount_currency`.

`PreparePaymentIntentService` then recomputes the same decision from scratch at pay time, from the persisted purchases. Everything feeding that decision can move in between: the per-seller launch flags (`checkout_local_method_ideal` and friends, all at 100% of actors but evaluated per call), the merchant account's settlement state, and the cart itself, which the page mutates client-side without ever re-fetching the payment config. Add or drop a line item, or accept a cross-sell, and the cart can go from mixed to uniformly-EUR while the Element stays on the USD it mounted with. The one USD-intent order in the set is a two-line cart where the second line came from a cross-sell.

That is the honest answer to "when": whenever the render-time currency decision and the pay-time currency decision disagree, which during a local-method ramp is a real but small window and is why it is fading out on its own.

## What this means for the PR

The guard itself compares the reported mount currency against the intent's in both directions, so it does fire on the real population — it's the reasoning around it that's wrong, and I'll rewrite the body.

The part I don't think survives review is the recovery. The failure is dressed as `BUYER_CURRENCY_QUOTE_INVALID`, which the checkout handles by cancelling, re-fetching surcharges and re-running the display gates. None of that re-fetches `checkout_payment`, so the Element remounts in the same server-rendered currency it had before, prepare recomputes the same intent currency, and the guard fires again. For the 54-order direction that turns a one-shot dead end into a loop the buyer cannot get out of without reloading the page. Failing closed is right; this particular failure code isn't, unless the retry also refreshes the payment config.

Given that, and given the rate is 57 orders over four days and already trailing off, I'd rather not ship the guard as written. Two options and I'll take your call:

1. Keep the fail-closed guard but make the retry actually recover — reload `checkout_payment` on this error so the Element remounts in the currency prepare just decided on.
2. Make the intent follow the Element instead of failing: prepare already knows the mount currency and the presentment machinery already builds forced-currency intents, so the EUR-mounted-vs-EUR-priced case could just succeed. Narrower blast radius than it sounds, but it's the currency contract question the issue flagged as needing an owner, which is why I didn't take it unilaterally.

I'd lean 1 now and 2 as the real fix.

slavingia:
2 is the real fix but should already be the case. Open new PR

PR #6439, opened 2026-07-27
Pin the wallet presentment safety properties the charge path relies on

gumclaw:
## What

Test-only. Pins two safety properties that keep a wallet (Apple Pay / Google Pay) purchase from being charged a total the buyer never approved on a buyer-currency checkout. **Both already hold on `main` — neither was covered**, so a future refactor could break either without failing a single test.

**1. An eligible wallet purchase must carry the buyer-currency quote token.** Without it the charge takes the canonical branch and the buyer pays a USD total their wallet sheet never showed. Pinned for both wallet lanes (server-confirm and client-confirm), which build their params through different branches of `createPurchasesRequestData`.

**2. The held-wallet resolver must re-confirm on the FX-quoted lane** when the wallet's post-approval billing address moves the tax location, and when the quote disappears from the reloaded surcharges (expired, or Stripe rejected the re-quote). `heldWalletPayment.test.ts` had a `describe` for the method-forced fixed-amount lane and none for this one.

Also adds the required `buyer_currency_presentment` key to the shared `paymentElementConfig` fixture, which was missing and producing a type error on the existing file.

## Why

This is what remains of PRs 2 and 3 in the plan on antiwork/gumroad-private#1436 after verifying the code. Both were scoped as production changes; both turned out to be already-working behavior with no coverage:

- PR 2 was "drop the `paymentMethod !== "card"` bail so wallets can carry the quote token." That bail never gated element wallets — a wallet payment runs with `state.paymentMethod === "card"` (`payment.ts` ~839; `reclaimCardLane` sets it). Token and source were already threading correctly on both lanes, verified by probe.
- PR 3 was "extend `resolveHeldWalletPayment` to compare the presentment total, expect the most review churn." It already does, because `approvedAmount` comes from `getStripePaymentElementAmount`, which on this lane returns the quote's locked presentment total. Probe against unmodified source returned `re-confirm` for both the tax-change and vanished-quote cases.

Merged into one test-only PR rather than shipping two PRs that change no production code.

## Test Results

`npx vitest run app/javascript/components/Checkout/ app/javascript/data/purchase.test.ts` — **13 files, 263 tests, all passing.**

New cases: 2 in `purchase.test.ts` (server-confirm and client-confirm wallet each carrying the token), 3 in `heldWalletPayment.test.ts` (quote unchanged → `continue`, tax moved → `re-confirm`, quote vanished → `re-confirm`).

**Mutation-proved load-bearing** — these are only worth adding if they fail when the property breaks:

| Mutation | Result |
| --- | --- |
| `resolveHeldWalletPayment` drops the element-amount comparison, keeping only the canonical-total one | **`requires re-confirmation when the quote is gone from the reloaded surcharges` is the ONLY failing test in the suite** (1 failed / 14 passed) |
| `getPaymentDetailsSource` reports `payment_request` for an element wallet | 2 failures, including the new server-confirm token case |

The first is the one worth noting: without this PR, deleting half of that comparison is invisible to the entire test suite.

`prettier` clean, `eslint` exit 0.

No screenshots — test-only, no rendered surface touched.

## Progress

- [x] Quote token pinned for both wallet lanes
- [x] FX-lane held-wallet re-confirm pinned (tax change, vanished quote)
- [x] Mutation-proved both properties are load-bearing
- [x] Full checkout suite green, prettier + eslint clean
- [ ] Reviewed and merged
- [ ] PR 4 — narrow `disable_wallets` in `StripePaymentPresenter` behind its own flag (the actual enable)

---

Stacked on #6438 — review that one first. Part of antiwork/gumroad-private#1436.

### DIFF
--- a/app/javascript/components/Checkout/heldWalletPayment.test.ts
+++ b/app/javascript/components/Checkout/heldWalletPayment.test.ts
@@ -20,6 +20,7 @@ const paymentElementConfig: CheckoutPaymentConfig = {
   elements_options: {
     stripe_elements_mode: STRIPE_ELEMENTS_MODE_FOR_PAYMENT_INTENT,
     currency: "usd",
+    buyer_currency_presentment: false,
     payment_method_types: ["card"],
     payment_method_creation: "manual",
     stripe_link_enabled: false,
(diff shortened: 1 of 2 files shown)


gumclaw:
Superseded by #6441, which carries this commit as the second of three. Closed as too granular to review on its own. Nothing lost: the coverage (and its mutation proof) is unchanged on #6441.

PR #6441, opened 2026-07-27
Enable Apple Pay and Google Pay on buyer-currency checkouts

gumclaw:
## What

Enables Apple Pay / Google Pay on buyer-currency (multi-currency) checkouts. They are currently suppressed server-side on every presentment cart, so the wallet surface silently disappears for a buyer the moment their cart becomes eligible for local-currency pricing — which is exactly the buyer this feature exists for.

Three commits, reviewable in order:

**1. Replace the blanket wallet refusal with a lane gate.** `Checkout::BuyerCurrencyEligibility#decision` refused presentment for any payment carrying a `wallet_type`, so a wallet purchase silently charged canonical USD. It now refuses only when the seller is outside the wallet rollout flags, or when the wallet did not come from the Payment Element.

**2. Pin the two safety properties the charge path relies on.** Both already held on `main` and neither was covered: (a) an eligible wallet purchase carries the buyer-currency quote token, on both wallet lanes; (b) the held-wallet resolver re-confirms when the wallet's adopted billing address moves the tax location, and when the quote vanishes on reload.

**3. Enable it, behind a ramp flag.** `Checkout::StripePaymentPresenter` now allows wallets on the FX-quoted presentment element lane when the new per-seller `:buyer_currency_wallets` flag is active, ANDed with the existing `:payment_element_wallets`.

### Why this is safe

On this lane the element's wallet sheet quotes the **same locked buyer-currency total the cart displays**. The browser mounts the element from the FX quote in the surcharge response (`getStripePaymentElementAmount` → `getStripePaymentElementPresentment`), and the charge path prices the intent from that same quote's signed token. Sheet and charge cannot diverge because they derive from one quote.

### What deliberately stays wallet-free

- **CardElement-fallback carts.** No Payment Element means the only wallet surface is the deprecated Payment Request Button, whose sheet is built from the canonical USD total. Suppressed regardless of either flag; covered by a spec.
- **The method-forced client-confirm lane** (iDEAL/Bancontact/Pix/UPI). It mounts with the product's pre-tax listed amount, so there is no composed total to hand a wallet at mount time. Keeps its own existing suppression, untouched. Descoped on the issue with a recommendation not to build it.
- **Multi-seller carts.** One locked quote total cannot be split across the several PaymentIntents a multi-seller order creates — tracked separately on an internal issue.

### Flag semantics

`buyer_currency_wallets?` is `payment_element_wallets? && all sellers have :buyer_currency_wallets`. Both directions matter and both are covered:

- Pulling `:buyer_currency_wallets` removes wallets from **presentment carts only**, leaving every other checkout untouched — the point of a separate flag.
- Pulling `:payment_element_wallets` still removes wallets **everywhere**, including here, so the existing global kill switch keeps its meaning.

Seller-complete keying (`sellers.all?`) matches the existing wallet flags, so enabling for one seller can never change another's checkout.

Both flag constants and the per-seller predicate (`Checkout::BuyerCurrencyEligibility.wallets_enabled?`) live on the eligibility service, and the presenter reads them from there. The rows a buyer sees and the charge the server accepts are therefore gated on one expression — they cannot drift apart.

**The flags gate the charge, not only the render.** `:buyer_currency_wallets` is read at charge time as well as at mount time, so pulling it stops a wallet charge from a tab that was already open. A render-only gate would have made the kill switch the rollout depends on effective for new page loads only.

## Why

The FX-quoted card lane has been live at 100% since the production rollout, so this is a conversion gap on a shipped surface rather than a pre-launch item — and it is the last wallet-shaped hole left by the checkout migration (`:payment_element_wallets` is fully ramped everywhere *except* presentment carts).

## Findings from building this

Three claims in the original plan were wrong or unverified. Each was checked against `main` and, where behavioral, proven by executable test rather than reasoning:

1. **The frontend `paymentMethod !== "card"` bail does not gate element wallets.** On the element surfaces a wallet payment runs with `state.paymentMethod === "card"` (`payment.ts` ~839 documents it; `reclaimCardLane` sets it). That bail excludes only PayPal. So the display, mount currency, and quote token were **already** reachable by a wallet payment — the presenter's `disable_wallets` was the sole thing holding the door shut, which makes commit 3 the whole enable and the server-side work its real backstop.
2. **`previewed_country` needed no change.** It reads `preview[:card]&.[](:country)` first and Apple/Google Pay populate that block (unlike Link, which is why the method-typed fallback exists). The plan listed this as work.
3. **The held-wallet re-confirm already covered the FX lane.** Probed against unmodified `resolveHeldWalletPayment`: `re-confirm` for both a tax-location change and a vanished quote, because `approvedAmount` already resolves from the quote. The plan called this "expect the most review churn"; it was a coverage gap, not a build.

Net effect: a four-PR plan became one PR, because two thirds of the scoped work was already shipped and merely untested.

**Payment Request Button gated on the server, not just unreachable.** It is already unreachable on this path two ways over — suppressed by `disable_wallets || payment_element_wallets` in `PaymentForm.tsx`, and selecting it sets `paymentMethod` to `"stripePaymentRequest"`, which withholds the quote token. Both are client mechanisms, though, and the property they protect is a money one: that sheet quotes canonical USD while the cart shows the local total. The charge path now checks `payment_details_source` directly, the same param `PurchasePaymentFlow` already uses to tell the two wallet surfaces apart.

## Test Results

**Ruby** — `stripe_payment_presenter_spec.rb` (97 examples) + `buyer_currency_eligibility_spec.rb` (59 examples): **156 examples, 0 failures**, run locally at head. The earlier fixture collision on `allows the PR1 Stripe test-mode Gumroad platform-account path` did not reproduce on this run.

**JS** — `npx vitest run app/javascript/components/Checkout/ app/javascript/data/purchase.test.ts`: **13 files, 263 tests, all passing.**

New coverage:

| Area | Cases |
| --- | --- |
| Presenter (flag matrix) | both flags on → wallets enabled; lane flag off → suppressed; general flag off → suppressed; CardElement fallback with both on → suppressed |
| Eligibility (wallet lane gate) | Payment Element wallet with both flags on → eligible; lane flag pulled → `wallet_payment_request`; general flag pulled → `wallet_payment_request`; wallet without the `payment_element` source → `wallet_payment_request`; non-wallet payment with flags off → still eligible |
| Quote token | server-confirm + client-confirm wallet payloads each carry the token |
| Held wallet (FX lane) | quote unchanged → `continue`; tax moved → `re-confirm`; quote vanished → `re-confirm` |

**Mutation-proved load-bearing** — coverage is only worth adding if it fails when the property breaks:

| Mutation | Result |
| --- | --- |
| `resolveHeldWalletPayment` drops the element-amount comparison, keeping only the canonical-total one | **`requires re-confirmation when the quote is gone from the reloaded surcharges` is the ONLY failing test in the suite** (1 failed / 14 passed) |
| `getPaymentDetailsSource` reports `payment_request` for an element wallet | 2 failures |
| `presentable_payment_surface?` reverted to the blanket `if wallet_type.present?` (first revision) | the Payment Element acceptance case failed while the fail-closed cases stayed green |
| the `wallet_lane_allowed?` guard deleted from `#decision` (leaving no server gate, as an earlier revision of this PR had it) | the whole `wallet payments` group reddens — 5 examples, 5 failures; restored, 5 pass |

The first is the notable one: before this PR, deleting half of that comparison was invisible to the entire test suite.

`rubocop` clean, `prettier` clean, `eslint` exit 0.

## Rollout

Ships **dark** — `:buyer_currency_wallets` defaults off, so merging changes nothing for any buyer.

⚠️ **Ramp caveat worth stating plainly:** because the display/token/re-confirm machinery was already wired, flipping this flag turns the entire path on at once. There is no partial state to observe in production, so preview QA and the per-seller ramp carry the full weight. Suggested sequence: preview-app QA on a non-US GeoIP (sheet currency, sheet total, charged total all agree) → one internal seller → watch the presentment rows in the rollout digest → widen.

Preview QA media on the flag-on state is below. It settles the display/mount half of the ramp question; the remaining human gate is a real wallet tap on a wallet-capable device (see QA steps).

Captured on `https://gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org` at head `5c0a38b7f`, Canadian buyer (`24.48.0.1` → `geo CA`, `buyer_currency_for_ip=cad`), USD-listed $35.00 single-seller cart. The two legs differ in **exactly one input** — the `:buyer_currency_wallets` flag — so anything that differs between them is this PR.

**The delta is the Payment Element's wallet mount config, not the rendered rows.** Apple Pay / Google Pay rows only render when the browser holds real device credentials, and this capture machine has none (`window.ApplePaySession === undefined`; `PaymentRequest('https://google.com/pay')` is blocked by the page CSP). So the honest evidence for "wallets are now offered on this lane" is what the app hands Stripe at mount time plus which wallet iframes stripe.js actually instantiates — both captured below. A real wallet-sheet screenshot is the human leg, not something to mock up.

| Flag ON — wallets offered | Flag OFF — wallets suppressed |
| --- | --- |
| CAD checkout, wallets on | CAD checkout, wallets off |
| Payment Element, wallets on | Payment Element, wallets off |

Mobile, 375px, flag on:

Mobile checkout, wallets on

The measured mount config and quote, both legs side by side:

Element mount config, both legs

Reading the two legs:

| | Flag ON | Flag OFF |
| --- | --- | --- |
| `publicOptions[wallets][applePay]` | `auto` | `never` |
| `publicOptions[wallets][googlePay]` | `auto` | `never` |
| `publicOptions[wallets][link]` | `auto` | `auto` (untouched by this flag) |
| Wallet iframes mounted by stripe.js | 2 (`payment-request-inner-google-pay`, `pay.google.com/.../payframe`) | 0 |
| Element mount | `amount=5724 currency=cad` | `amount=5724 currency=cad` |
| Cart display | CA$49.76 + CA$7.48 tax = **CA$57.24** | CA$49.76 + CA$7.48 tax = **CA$57.24** |

Two things worth naming from that table:

1. **The wallet sheet is fed the same locked total the cart shows.** Both legs mount at `amount=5724 currency=cad`, from quote `fxq_0Txwp89e1RjUNIyYO6aUmj0Y` (`canonical_total_cents=4024`, `presentment_total_cents=5724`, `fx_rate=0.700281`) — the same quote whose signed token the charge path prices against. That is the safety argument in the "Why this is safe" section, observed rather than reasoned.
2. **Pulling the flag is surgical.** It flips `applePay`/`googlePay` to `never` and leaves `link` at `auto`, the FX lane engaged, and the CAD totals byte-identical. An emergency ramp-down removes wallets from presentment carts without disturbing anything else on the checkout.

## QA steps

Preview app: **https://gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org**

Flags to enable on the preview (`console.sh -w`, all reset on every redeploy):
`:buyer_local_currency`, `:buyer_currency_charging`, `:stripe_payment_element_checkout`, `:stripe_payment_element_client_confirm`, `:payment_element_wallets`, and this PR's `:buyer_currency_wallets` — for `seller@gumroad.com`.

1. Make the browser look Canadian: open **https://gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org/qa/spoof?ip=24.48.0.1** (the TEMP cookie-mode geo spoof on this branch). Confirm at **/qa/preview_debug** that `geo_country_code: "CA"` and `buyer_currency_for_ip: "cad"`.
2. Open **https://gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org/checkout?product=uv&quantity=1** (`QA Wallet Presentment 6441`, USD-listed $35). Fill the email field. The cart should read **CA$49.76 / tax CA$7.48 / total CA$57.24** — buyer-currency presentment is engaged.
3. **On a wallet-capable device** (iPhone Safari with a card in Wallet, or Chrome signed into a personal Google account with Google Pay): the "Pay with" accordion now shows an **Apple Pay** or **Google Pay** row above Card. Open the sheet and confirm it quotes **CA$57.24** — the same total the cart shows, not a US$ figure.
4. Complete the wallet payment. Confirm the success page, then in admin/console that the purchase has a `PurchasePresentment` with `presentment_currency: "cad"` and `presentment_total_cents: 5724`, and that the Stripe PaymentIntent settled **5724 cad** carrying the same `stripe_fx_quote_id` as the checkout's quote token.
5. **Ramp-down check:** disable `:buyer_currency_wallets` for the seller, reload the same checkout. The wallet rows disappear, Link and Card remain, and the totals stay CA$57.24 — nothing else about the checkout changes.
6. **Global kill switch still global:** disable `:payment_element_wallets` instead and load any *non*-presentment (US buyer) checkout — wallets are gone there too.

Steps 1–2, 5 and the mount-config half of 3 are what the screenshots above show. Step 3's rendered row and the sheet total were confirmed on a real device — see below.

## On-device wallet leg — confirmed

Kyle QA'ed this manually on a wallet-capable device. The real Apple Pay sheet renders on a buyer-currency checkout and quotes the same locked buyer-currency total the cart shows — cart total **CA$56.55** (CA$50.05 + CA$6.50 tax), sheet **56.55**. That is the property the "Why this is safe" section argues for, now observed end to end on a real wallet rather than inferred from the mount config.

Apple Pay sheet on a buyer-currency checkout

The two `TEMP:` commits (geo/FX spoof + preview-host reCAPTCHA skip) that made the preview clickable have been **reverted** — the branch now carries only the product change, its specs, and the QA media.

## Progress

- [x] Charge path accepts wallet payments
- [x] Quote token + held-wallet safety properties pinned, mutation-proved
- [x] Presenter enables wallets on the FX lane behind `:buyer_currency_wallets`
- [x] CardElement fallback, method-forced lane, and multi-seller carts verified still wallet-free
- [x] Flag matrix covered in both directions
- [x] Ruby + JS suites green (sole failure proven pre-existing), rubocop/prettier/eslint clean
- [x] Preview-app QA media on the flag-on state — display, element mount config, and quote captured on the hosted preview (flag-on vs flag-off)
- [x] On-device wallet leg: real Apple Pay sheet quoting the cart's buyer-currency total (manual QA)
- [x] TEMP preview-QA commits reverted
- [ ] Reviewed and merged
- [ ] Ramp: one seller → digest watch → widen

---

Supersedes #6438 and #6439, which are being closed in favour of this single PR. Part of the buyer-currency wallets rollout.

---

## AI disclosure

Written by **claude-opus-5** (Hermes Agent). Preview-app QA media captured autonomously by the same agent running as the preview-screenshot watcher cron job: seeded the preview's flags, FX rate, and a USD-listed QA product via the staging write console, drove both flag legs through Playwright against the hosted preview app, and read the Payment Element mount config and buyer-currency quote off the live network traffic.

Prompts: the standing order from Sahil (2026-07-19) to screenshot relevant PRs from the hosted preview apps as soon as they deploy, and to un-draft a draft PR once it has evidence.


gumclaw:
## Preview app + wallet QA in progress

Preview app is deploying — `preview` label added and the TEMP QA scaffolding cherry-picked (the "Preview app" GitHub workflow is a no-op in this repo; a **push** is what deploys, so the TEMP commits double as the deploy trigger).

Buildkite build [16537](https://buildkite.com/gumroad-inc/gumroad/builds/16537) on `5c0a38b7f`: `Preview gate` → images → `Provision Preview Environment` all passed, `Compile Assets` running, `Deploy Preview App` queued.

### Two TEMP commits on this branch — revert before merge

| Commit | What | Why it's needed |
| --- | --- | --- |
| `8cae1a368` | geo/FX spoof (cookie mode) + FX rate seed | Preview pods resolve **every** IP to `US` — without this there is no non-US buyer, so no FX quote and no presentment lane to test at all |
| `5c0a38b7f` | checkout reCAPTCHA skip on preview hosts | Lets a scripted checkout submit complete |

Both carry `revert only AFTER QA sign-off, before merge` in their messages. They are **not** part of the change under review — I'll revert them once evidence is captured and call it out explicitly.

### What I'll capture

Five flags must be re-enabled after every redeploy (they reset), including this PR's new one: `stripe_payment_element_checkout`, `buyer_local_currency`, `buyer_currency_charging`, `payment_element_wallets`, `buyer_currency_wallets`.

Then, on a Canadian-GeoIP checkout for a USD product:

1. **Server decision** — `checkout_payment` props carry `disable_wallets: false`, `payment_element_wallets: true`, `buyer_currency_presentment: true`. This is the change itself.
2. **What Gumroad asked Stripe for** — the `/v1/elements/sessions` mount currency/amount and the payment-element wallet options (`applePay`/`googlePay` set to `auto`, not `never`).
3. **The quote** — `/customer_surcharge` returning a real `buyer_currency_quote` with a token, and the cart rendering CA$ rather than US$.
4. **Rendered method rows** — the element accordion's rows.
5. **Control leg** — `buyer_currency_wallets` off, everything else identical: wallets must disappear while the CA$ presentment display stays. That's the flag's kill switch, and it's the leg that proves the gate is scoped to this lane rather than blanket-disabling.

### One honest limit, stated up front

**Apple Pay renders only in Safari on a device with a provisioned Wallet card; Google Pay only in Chrome signed into an account with a saved card.** Headless Chromium has neither, so if a wallet *row* doesn't paint in my capture that is the browser being incapable — **not** evidence the feature is broken.

That's why legs 1 and 2 are the primary evidence: they prove Gumroad mounted the buyer's currency and asked Stripe to enable both wallets. Whether a given device then paints the button is Stripe's and the device's business, and it's the same code path already at 100% on non-presentment carts. I'll try a headed browser for a real row, and if the environment can't produce one I'll say so plainly rather than implying the sheets were visually confirmed when they weren't — a human on a real iPhone/Mac is the only way to see the actual Apple Pay sheet.

Evidence lands as a follow-up comment.

gumclaw:
## Correction on the wallet-rendering limit

My earlier comment said headless Chromium can't render Apple Pay / Google Pay, so the sheets couldn't be visually confirmed and only the server-side legs would be evidence. **That framing was too pessimistic and I'm correcting it before it becomes the excuse for weaker evidence.**

The QA machine has Apple Pay configured in Safari and Google Pay in Chrome. Verified directly in real Safari via AppleScript:

```
{"hasAPS":true,"proto":"https:","host":"gumroad.com","canMakePayments":true,"maxVersion":14}
```

`ApplePaySession` is present and `canMakePayments()` returns **true** — so a real Apple Pay row/button is reachable on this box, and the wallet legs will be captured in the actual browsers rather than reasoned about. Playwright's bundled Chromium was the limitation, not the machine.

Revised plan for the wallet legs, once the preview serves:

- **Apple Pay** — real Safari, driven by AppleScript (JS-from-Apple-Events is enabled), with `screencapture -l<windowid>` scoped to the browser window only. Assert `ApplePaySession.canMakePayments()` on the *preview* origin and capture the rendered Apple Pay row in the Payment Element.
- **Google Pay** — real Chrome. Its AppleScript JS bridge is off and I'm not silently flipping a persistent setting on the user's browser, so this leg is a window-scoped screenshot of the rendered row rather than a JS assertion.
- **Server + Stripe legs** — unchanged and still captured (props, `elements/sessions` mount currency, quote token, control leg with the ramp flag off). Those remain the proof of *what this PR changed*; the wallet screenshots are proof the sheets actually appear.

One limit that genuinely does remain: **I will not complete a real Apple Pay payment.** Authorising a live wallet payment means Touch ID / a real card on the user's device, which is exactly the kind of payment UI an agent shouldn't be driving. The evidence stops at "the sheet renders and quotes the correct locked buyer-currency total". Completing a wallet purchase end to end stays a human step, and I'll say so in the evidence comment rather than implying full coverage.

`Compile Assets` is still running on build [16537](https://buildkite.com/gumroad-inc/gumroad/builds/16537); evidence follows once `Deploy Preview App` finishes.

gumclaw:
## Preview app QA — evidence

Preview app is live and serving this branch:
**https://gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org** — HTTP 200, `x-revision: 5c0a38b7f550` (matches the pushed head). Buildkite build [16537](https://buildkite.com/gumroad-inc/gumroad/builds/16537) passed.

All five flags seeded on the preview (they reset on every redeploy):
`stripe_payment_element_checkout`, `buyer_local_currency`, `buyer_currency_charging`, `payment_element_wallets`, `buyer_currency_wallets` — all `true`.

### ✅ Leg 1 — the server decision (this is the PR)

Presenter output for a real presentment cart (USD $35 product, Canadian GeoIP), read off the pod:

```
props integration=payment_element fallback=nil
      disable_wallets=false  payment_element_wallets=true  presentment=true
```

Before this PR that line read `disable_wallets=true`, `payment_element_wallets=false`. This is the change working.

### ✅ Leg 2 — what Gumroad asks Stripe for

The Payment Element mounts in the buyer's currency at the quote's locked total, and asks Stripe to enable **both** wallets:

```
/v1/elements/sessions?deferred_intent[amount]=5275&deferred_intent[currency]=cad …

publicOptions[wallets][applePay]  = auto
publicOptions[wallets][googlePay] = auto
publicOptions[wallets][link]      = auto
```

`auto` (not `never`) is what makes the sheets renderable — and the mount is `cad 5275`, not USD.

### ✅ Leg 3 — the quote and the displayed cart agree

```
GEO      geo_country_code=CA  buyer_currency_for_ip=cad  cached_cad_rate=1.37
QUOTE    currency=cad  presentment_total_cents=5275  canonical_total_cents=3675  has_token=true
CART     CA$50.25 + CA$0.00 + CA$2.50 = CA$52.75
```

Cart total CA$52.75 == quote `presentment_total_cents` 5275 == element mount amount. Sheet, cart, and charge all derive from one quote — which is the safety argument for enabling wallets here.

### ⚠️ Leg 4 — the wallet sheets did NOT render, and it is NOT this PR

I could not produce a visible Apple Pay or Google Pay sheet on this machine. Being precise about why, because "no wallet row appeared" is exactly the kind of result that gets misread as a bug in the change:

Real Safari on the preview origin:

```
{"host":"gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org",
 "hasAPS":true, "canMakePayments":true,
 "cartAmounts":["CA$49.76","CA$0.00","CA$7.48","CA$57.24"]}
```

So `ApplePaySession` exists and `canMakePayments()` is **true**. But Stripe's own gate says no:

```
ApplePaySession.canMakePaymentsWithActiveCard(...) → false
stripe.paymentRequest(...).canMakePayment()       → null
```

`canMakePayments: true` means "this browser supports Apple Pay". `activeCard: false` means **no provisioned card in this Mac's Wallet**. Stripe returns `null` and correctly suppresses the wallet row.

**Control test that rules out the PR:** the identical Stripe probe on **production gumroad.com** (Stripe.js injected, same browser, same session) also returns `canMakePayment() → null`. Production has wallets fully ramped, so if the machine could paint a wallet it would paint one there. It can't. The blocker is the device, not this branch.

Two other things I verified rather than assumed:

- **The preview domain IS registered with Stripe** — `Stripe::PaymentMethodDomain` lists `gumclaw-2026-07-27-buyer-currenc.apps.staging.gumroad.org` with `apple_pay: active, google_pay: active`. So domain registration is not the blocker either.
- **I cannot capture screenshots on this host.** `screencapture` returns an all-black 5120×2880 frame (sampled unique colours: 1, dark fraction 1.0) — Screen Recording permission isn't granted to the agent's terminal. So there are no QA images in this comment; publishing a black PNG as "evidence" would be worse than publishing none.

### What still needs a human

1. **See the sheets.** On a Mac with a card in Wallet (or an iPhone), open the preview checkout with a non-US IP and confirm the Apple Pay row appears in the Payment Element and its sheet quotes **CA$52.75**, not US$36.75. Google Pay likewise in Chrome signed into an account with a saved card.
2. **Complete one wallet purchase.** I deliberately did not attempt this — authorising a live wallet payment needs Touch ID and a real card, which is not something I should be driving. This is the one leg that proves the charged amount equals the sheet amount end to end.
3. **Grant Screen Recording** to the agent's terminal if you want me producing QA media in future runs.

Everything the code controls is verified green. What's unverified is strictly what the device couldn't do.

### TEMP commits — must be reverted before merge

| Commit | What |
| --- | --- |
| `8cae1a368` | geo/FX spoof (preview pods resolve every IP to US, so there's no non-US buyer without it) |
| `5c0a38b7f` | checkout reCAPTCHA skip on preview hosts |

Both carry `revert only AFTER QA sign-off, before merge`. I'm leaving them on the branch so a human can click through the wallet legs above; they must come off before merge and I'll do that on request, or immediately after sign-off.

gumclaw:
## Wallet capability — retested properly, and I have to correct my own evidence

You pushed back twice on this, so I re-tested instead of repeating myself. **My previous control test was invalid** and I want that on the record before the conclusion.

### What I got wrong

I "proved" the machine couldn't do wallets by injecting a `pk_test_…` Stripe key onto production `gumroad.com` and observing `canMakePayment() → null`. That test is meaningless: `gumroad.com` isn't a registered domain for that *test* Stripe account, so `null` was the expected answer regardless of the Mac's Wallet state. I also probed `canMakePaymentsWithActiveCard()` with **guessed** merchant IDs (`merchant.com.gumroad`, `merchant.com.stripe.stripe`), and an unregistered merchant ID returns `false` no matter what cards exist. Two bad tests pointing the same direction read as corroboration; they were the same mistake twice.

You were also right on the substance I'd dismissed — **Apple Pay is configured on this Mac:**

```
system_profiler SPSecureElementDataType → Apple Pay: Production Signed: Yes, Restricted Mode: No
bioutil -r                              → Biometrics for ApplePay: 1
```

So the hardware and biometric side are live, and my "the machine can't do wallets" framing was wrong.

### The valid test

Apple's own demo, on Apple's own registered origin, using the API whose whole purpose is this question:

```js
ApplePaySession.applePayCapabilities('merchant.com.apdemo')
→ {"paymentCredentialStatus":"paymentCredentialsUnavailable"}
```

`applePayCapabilities()` on a **valid** merchant ID, at `applepaydemo.apple.com`. `paymentCredentialsUnavailable` means: Apple Pay is set up and available, but **there is no payment card provisioned** in this Mac's Wallet. Corroborated locally, and these are the checks that are actually meaningful:

```
~/Library/Passes/Cards/                     → empty
Chrome Default   credit_cards=0 masked_credit_cards=0
Chrome Profile 1 credit_cards=0 masked_credit_cards=0   (gumclaw@gumroad.com)
```

So: **Apple Pay is configured, but no card is added — in Wallet or in either Chrome profile.** Stripe suppressing the wallet row is correct behaviour given zero available payment credentials, and that is a device-provisioning gap, not something this PR can affect. A paired iPhone is present over Bluetooth, which is another route to authorise, but the browser still needs a provisioned credential to offer the button.

### Where that leaves the PR

Unchanged and still fully green on everything the code controls — the presenter emits `disable_wallets=false / payment_element_wallets=true`, the element mounts `cad 5275`, Stripe is asked for `applePay: auto, googlePay: auto`, and cart total CA$52.75 == quote == mount amount. The wallet *row* needs a provisioned card, which is one action in System Settings → Wallet & Apple Pay (and a card saved in Chrome for Google Pay).

**Add a card and I'll immediately re-run the capture** — the preview is live, all five flags are seeded, and the driver scripts are ready; it's a two-minute re-run. I still won't *complete* a wallet payment (Touch ID + a real card is not something I should authorise), but I can confirm the row renders and that its sheet quotes CA$52.75 rather than US$36.75, which is the leg that matters.

Separately: `screencapture` still returns an all-black frame here (Screen Recording not granted to the agent's terminal), so even with a card I'd be reporting readouts rather than images until that's enabled.

xrav3nz:
QA'ed this manually

image

gumclaw:
Thanks for running the sheet on a real device — that was the one leg I couldn't produce from the capture machine, and your screenshot settles it: cart CA$56.55 (CA$50.05 + CA$6.50 tax) and the Apple Pay sheet quoting 56.55, so the wallet is charging the locked buyer-currency total rather than the canonical USD one. That's exactly the property the safety argument rests on, now observed instead of inferred.

Cleaned up on the back of that:

- Reverted both `TEMP:` commits (the preview geo/FX spoof initializer and the preview-host reCAPTCHA skip). The branch now carries only the product change, its specs, and the QA media — nothing preview-only ships.
- Committed your Apple Pay capture to `qa-media/` and added it to the PR body under a new "On-device wallet leg — confirmed" section, so the evidence lives with the PR rather than only in this thread.
- Ticked the on-device box in the Progress checklist. The only remaining items are review/merge and the ramp.

Head is now `a72f29f`. Merge and the ramp stay with you — this one has a per-seller flag defaulting off, so merging changes nothing for buyers until it's flipped.

xrav3nz:
@gumclaw 

P1 — High

#1 — The deleted wallet refusal left no server-side gate at all. app/services/checkout/buyer_currency_eligibility.rb:191

Deleting return fallback(:wallet_payment_request) if wallet_type.present? and replacing it with nothing leaves Checkout::BuyerCurrencyEligibility reading no request params at all. Two consequences, not equally serious:

- (a) Live, no client tampering required. :buyer_currency_wallets is read in exactly one place — StripePaymentPresenter#buyer_currency_wallets?, a render-time prop decision. Nothing on the charge path reads it. A tab loaded while the flag is on keeps its wallet rows and still completes a buyer-currency wallet charge after the flag is pulled. The PR describes the flag as pullable "instantly" and the rollout plan leans on that; the code delivers a mount-time gate only.
- (b) Defense-in-depth. Nothing server-side distinguishes a Payment Element wallet (sheet quotes the locked FX total — safe) from a Payment Request Button wallet (sheet quotes canonical USD — unsafe). BuyerCurrencyQuote.verify! binds expiry, seller, merchant account, Stripe account, currency, canonical total, and line allocations — never the payment surface. Adversarial verified the PRB genuinely cannot reach this today (suppressed at render and paymentMethod === "stripePaymentRequest" withholds the token, independent of bundle version), so this is not a live bug — but it makes a payment-safety invariant client-enforced, against CONTRIBUTING.md's "Enforce all constraints on the server."

Both are one fix, where the deleted line was:

return fallback(:wallet_payment_request) if wallet_type.present? && !wallet_lane_allowed?

with wallet_lane_allowed? checking the seller's :buyer_currency_wallets + :payment_element_wallets flags (fixes (a)), and if you also want (b), that params[:payment_details_source] == PurchasePaymentFlow::PAYMENT_ELEMENT. Reviewers split on shape — project-standards would host the flag constant on the eligibility service the way FEATURE_NAME already is; correctness offers a frontend alternative (gate getCheckoutBuyerCurrencyDisplay on the server's lane prop). That split is why this is a decision, not an auto-apply.

Severity note: the mechanical merge would carry correctness's P0. I kept its own P0 at 50 (unverified), adversarial specifically refutedreachability, and quote verification is intact so no buyer is mispriced. What's real is an operational gap in a kill switch the rollout depends on.

P2 — Moderate

#2 — params: is now dead weight in the eligibility service. app/services/checkout/buyer_currency_eligibility.rb:176

wallet_type was the only reader of params in this class. `@params` is now assigned at line 182 and read nowhere — the attr_reader, the kwarg, and the argument threaded from
create_service.rb:225 and method_forced_presentment.rb:122 are all inert.ng #1: the leading fix reads params[:payment_details_source].

gumclaw:
Fixed both, pushed in a6db687.

**#1 — server-side wallet gate.** You're right that the deleted line left the flag as a mount-time decision only, and that's the half that actually matters: the rollout plan treats `:buyer_currency_wallets` as an instant kill switch, and a render-only gate leaves every already-open tab able to complete a buyer-currency wallet charge after the flag is pulled. `#decision` now carries the guard again, in the shape you sketched:

```ruby
return fallback(:wallet_payment_request) if wallet_type.present? && !wallet_lane_allowed?
```

with `wallet_lane_allowed?` requiring both that the seller is in `:payment_element_wallets` AND `:buyer_currency_wallets`, and that `params[:payment_details_source] == PurchasePaymentFlow::PAYMENT_ELEMENT`. I took the (b) half as well. Agreed the Payment Request Button can't reach this today, but the property it protects is that the sheet quotes the same total the cart shows, and that was being enforced entirely client-side. `PurchasePaymentFlow#payment_details_source_for` already reads exactly this param to tell the two wallet surfaces apart, so the recorded surface and the charge decision now read one input rather than two.

On the shape split: went with project-standards. `PAYMENT_ELEMENT_WALLETS_FEATURE_NAME`, `WALLETS_FEATURE_NAME`, and a `.wallets_enabled?(seller)` predicate now live on `Checkout::BuyerCurrencyEligibility` next to `FEATURE_NAME`, and `StripePaymentPresenter` reads them from there — so the rows a buyer sees and the charge the server accepts cannot end up gated on different expressions. The frontend alternative would have left the enforcement client-side, which is the thing the finding is about.

**#2 — dead `params:`.** Resolved by the same fix rather than separately: `wallet_lane_allowed?` reads `params[:payment_details_source]`, so the kwarg and the argument threaded from `create_service.rb` and `method_forced_presentment.rb` are load-bearing again.

Coverage — five new cases in `buyer_currency_eligibility_spec.rb`: element wallet with both flags on is eligible; lane flag pulled → `wallet_payment_request`; general flag pulled → `wallet_payment_request`; wallet without the `payment_element` source → `wallet_payment_request`; and a non-wallet payment with the flags off still eligible, so the gate can't quietly widen into a general refusal. Mutation-checked: deleting the guard line reddens all five, restoring it greens all five.

`buyer_currency_eligibility_spec.rb` 59 examples and `stripe_payment_presenter_spec.rb` 97 examples both pass locally at head — 156, 0 failures. The platform-account fixture collision noted earlier in the body didn't reproduce on this run. Rubocop clean. PR body updated to match (the "not special-cased" paragraph said the opposite of what the code now does).

gumclaw:
## Fable-5 adversarial pre-merge review — `READY-TO-MERGE`, auto-merge armed

I had parked this on you without running the adversarial review first — reversing that, since nothing here was blocked on a human. Per Sahil's rule (green CI + clean fable verdict + verified evidence ⇒ arm it myself), the gate is now run and auto-merge is armed.

**Verdict: `READY-TO-MERGE`**, reviewed at head `a6db687b5` — the review picked up the mid-flight commit that gates buyer-currency wallet charges **on the server**, not only at render, and reviewed that too. That backstop is the right shape: a render-time-only gate is bypassable, so enforcing it server-side is what makes the wallet eligibility real.

The high-risk item on a wallet path is that the amount and currency shown in the Apple Pay / Google Pay sheet must match what actually gets charged — a buyer authorizing one figure and being charged another. The reviewer verified all key paths, re-ran mutation testing locally, and executed both Ruby suites against the branch tip.

You're still the right person to eyeball the wallet UX if you want to; the merge just isn't waiting on it.

gumclaw:

```
```

gumclaw:

gumclaw:
## Post-merge verification of the charge-time wallet gate (head `1e821b8`)

The preview-screenshot watcher flagged this PR at head `1e821b883651` — the merge from `main` that
also carried `a6db687b5` ("Gate buyer-currency wallet charges on the server, not only at render").
The PR merged while that run was in flight, so this is a record rather than a gate: the evidence for
that commit landed after the merge, and it confirms the fix holds on the deployed branch.

`a6db687b5` moved the flag check from render-time to charge-time, so the mount-config capture already
in the body is the *wrong* evidence for it — nothing about the rendered element changed. The property
it added has no pixels on any device, so it was probed directly on the preview pod
(`gumclaw-2026-07-27-buyer-currenc`, `x-revision 1e821b883651`) by building
`Checkout::BuyerCurrencyEligibility` in memory over the full flag matrix. No seeding, no purchase
written, flags restored to their found state afterwards.

**`#wallet_lane_allowed?` truth table** (`seller@gumroad.com`, `wallet_type: apple_pay`):

| `payment_element_wallets` | `buyer_currency_wallets` | Payment Element source | Payment Request Button source |
|---|---|---|---|
| ON | ON | **allowed** | refused |
| ON | OFF | refused | refused |
| OFF | ON | refused | refused |
| OFF | OFF | refused | refused |

Both halves @xrav3nz asked for hold on the deployed code: pulling **either** flag refuses the charge —
so the kill switch reaches a tab that was already open, not just newly rendered checkouts — and the
deprecated Payment Request Button is refused **even with both flags on**, which is the defense-in-depth
half (its sheet quotes canonical USD while the cart shows the local total).

**The "cannot drift apart" claim, checked rather than argued** — the presenter's constants are now
identities of the eligibility service's, so the rows a buyer sees and the charge the server accepts
read one expression:

| Check | Result |
|---|---|
| `StripePaymentPresenter::BUYER_CURRENCY_WALLETS_FEATURE_NAME` vs `BuyerCurrencyEligibility::WALLETS_FEATURE_NAME` | both `:buyer_currency_wallets`, identical |
| `…::PAYMENT_ELEMENT_WALLETS_FEATURE_NAME` on both classes | both `:payment_element_wallets`, identical |
| `BuyerCurrencyEligibility.wallets_enabled?(nil)` | `false` — safe for the presenter's per-seller call over a possibly-empty collection |

**Surface re-smoked at the same revision.** `/checkout?product=uv` on the preview: renders,
`broke=false`, zero `pageerror`s, React root mounted, cart and Pay button present; the only console
line is the ambient Stripe-iframe `requestStorageAccess` notice, which appears on control previews too.

**Housekeeping:** the watcher briefly re-pushed `gumclaw/2026-07-27-buyer-currency-wallets` with a
QA-media commit before noticing the merge had already landed. That branch has been deleted again —
merged state is `046c66f7c` and nothing was added to it after the merge.

**Nothing here changes the ramp.** `:buyer_currency_wallets` still defaults off, and the remaining
gate is the operational one already in the body: one internal seller → watch the presentment rows in
the rollout digest → widen.

PR #6445, opened 2026-07-27
Stop crediting affiliates for commissions clawed back in full

gumclaw:
## What

`User#affiliate_credit_sum_from_scope` credits an affiliate for the full value of a commission that was clawed back in full. Restrict its add-back to credits that still retain value.

## Why

When a purchase is refunded — even partially — `Purchase#process_refund_or_chargeback_for_affiliate_credit_balance` stamps a refund or chargeback balance onto the affiliate credit **unconditionally**:

```ruby
if refund
  affiliate_credit.affiliate_credit_refund_balance = affiliate_balance_transaction.balance
elsif dispute
  affiliate_credit.affiliate_credit_chargeback_balance = affiliate_balance_transaction.balance
end
affiliate_credit.save!

if affiliate_credit_cents != refund_cents    # only THIS is conditional
  affiliate_partial_refunds.create!(...)
end
```

That stamp drops the row out of `AffiliateCredit.paid`. So the sum's second and third terms are an **add-back of the remainder the affiliate still earned**, not a redundancy: on a $100 sale with a 25% commission and a $40 refund, the affiliate keeps ~$15, term 1 dropped the whole $25, term 2 restores it and term 3 removes the refunded $10.

The bug is that the add-back was unconditional. Only the `affiliate_partial_refunds` row is conditional, and it is skipped when `affiliate_credit_cents == refund_cents` — a **full** claw-back. In that case term 2 added the commission back and term 3 had nothing to subtract, so a commission the affiliate no longer has counted at full value.

### Measured, not reasoned

Read-only production replica (lag 1–3s), newest ~1M `affiliate_credits` ids:

| Class | n | before | correct | overstated |
|---|---|---|---|---|
| has an `affiliate_partial_refunds` row | 382 | $5,981.26 | $5,981.26 | **$0.00 — already exact** |
| **no** `affiliate_partial_refunds` row | 22 | $52.58 | $0.00 | **$52.58** |
| all partially refunded | 404 | $6,033.84 | $5,981.26 | $52.58 (0.9%) |

Blast radius **$52.58 across 18 affiliates, worst single affiliate $10.75**. Every one of the 22 is a full reversal — sample rows have `amount_cents == purchase.affiliate_credit_cents` with the purchase refunded to its full price:

```
credit 2328968  amount 1075  refund_bal 15903816              refunded 12000 / price 12000
credit 2350039  amount  310  refund_bal + chargeback set      refunded  3500 / price  3500
credit 2418063  amount  220  chargeback_bal 16396926          refunded  2472 / price  2500
credit 2449452  amount    3  refund_bal 16521838              refunded    99 / price    99
```

### Rejected alternative: dropping the add-back entirely

The obvious-looking fix — delete terms 2 and 3 and return the plain `paid` sum — reads as a simplification and is a **~114x regression**. Because a partial refund always knocks the credit out of `paid`, that would remove **$5,981.26 of correctly-counted, still-earned commission across 382 credits** to fix a $52.58 error across 22. Not taken.

### Why EXISTS and not a join

Restricting the add-back with `.joins(:affiliate_partial_refunds)` is the natural way to express "only credits with a refund row" and it would introduce a genuine double-count: **56 credits in the table carry more than one `affiliate_partial_refunds` row**, one of them four. A join sums `amount_cents` once per row, so those affiliates would be credited 2–4x their commission:

```
credit 2444587  amount_cents 270  apr_amounts [207, 45, 11, 3]  -> joined add-back = 1080, want 270
credit 2002178  amount_cents 4500 apr_amounts [3375, 563, 225]  -> joined add-back = 13500, want 4500
```

An `EXISTS` subquery restricts without multiplying. The **subtraction** still joins, because there it is the refund rows' own amounts being summed and one row per refund is exactly right.

## What this affects

`affiliate_credit_cents_for_balances` reaches the `affiliate_credits_cents` key via `sales_data_for_balance_ids` → `payout_sales_data`, which has exactly two readers: the payout-confirmation email (`CustomerLowPriorityMailer`) and one row on the Payouts page (`app/javascript/pages/Payouts/Index.tsx:129`). **It is a displayed figure and does not determine what anyone is paid** — payouts come off `Balance` rows, and the balance-side computation (`affiliate_fee_cents_for_balances`) does not route through this method. `affiliate_credits_sum_total` feeds the affiliated-products dashboard; `affiliate_credits_sum_for_credits_created_between` feeds `last_weeks_sales` and carries the same error bounded by the same $52.58.

## Before / after

No rendered surface changes shape — the same element shows a corrected number. Behaviour table:

| Credit state | Before | After |
|---|---|---|
| Untouched commission ($2.00) | $2.00 | $2.00 (unchanged) |
| Reversed in part ($0.75 of $2.00 clawed back) | $1.25 | $1.25 (unchanged) |
| Reversed over two refunds ($0.50 + $0.30) | $1.20 | $1.20 (unchanged) |
| **Clawed back in full** | **$2.00** | **$0.00** |

## Test results

Six new examples covering each shape above, plus rubocop, plus two mutation checks proving the specs are load-bearing.

Reverting `app/modules/user/stats.rb` to `main` reddens exactly the two claw-back cases (`Expected 200 to eq 0`, `Expected 400 to eq 200`) while the partial-refund cases stay green — the correct signature, since `main` is already right about those.

The second mutation came out of the pre-merge review and closed a real gap. Breaking the subquery's correlation to an uncorrelated `where("1 = 1")` left the original five examples green, because in the full-claw-back cases the test database held no `affiliate_partial_refunds` rows at all, so even a broken subquery evaluated to false. The sixth example gives one affiliate both shapes at once — a credit reversed in part and one clawed back in full — which returns 325 instead of 125 under that mutation. Re-run with it in place: `6 examples, 0 failures` shipped, `6 examples, 1 failure` mutated, reddening only that case.

test results

Running both spec files whole gives **123 examples / 98 failures** on this branch. A detached `origin/main` worktree on the same machine gives **118 examples / 98 failures** — the identical 98, all `ActiveRecord::RecordInvalid: We couldn't charge your card` from `create(:purchase)`, because the shared local test DB has no seeded `MerchantAccount.gumroad`. The 5-example delta is exactly this PR's new examples, all passing. Environmental, not this change; CI seeds the merchant account and is the arbiter.

## QA steps

No UI click-path; the change is one method behind a displayed figure. To verify on a preview app's Rails console:

1. Pick an affiliate with a fully clawed-back commission: an `AffiliateCredit` whose purchase has `stripe_partially_refunded: true`, whose `affiliate_credit_refund_balance_id` or `affiliate_credit_chargeback_balance_id` is set, and which has **no** `affiliate_partial_refunds` rows.
2. `user.affiliate_credits_sum_total` — that credit must contribute 0.
3. Add an `AffiliatePartialRefund` for it with `amount_cents` below the credit, re-read: it now contributes `amount_cents - refund`.
4. Check an untouched credit is unchanged, confirming the add-back restriction did not narrow the ordinary path.

---

🤖 Written by Hermes Agent using **claude-opus-5** (read from `model.default` in the agent config).

Instructions given: build the fix for the affiliate-credit sum defect after a maintainer said "Fix." on the tracking issue. The tracking issue's stated cause was falsified against production first (it claimed partially refunded credits stay inside the `paid` scope; zero of 355 across the table's history do), so the fix addresses the real defect and the issue was corrected rather than built against as written.


gumclaw:

**Scope.** One method in `app/modules/user/stats.rb` and its specs. No new columns, no migration, no schema change, no money-movement path — the corrected figure is displayed, not paid (readers enumerated in the body: the payout-confirmation email and one Payouts-page row; payouts themselves come off `Balance` rows via `affiliate_fee_cents_for_balances`, which does not route through this method).

**The load-bearing decision, and the thing to attack in review.** The tempting fix here is deleting the add-back and returning the plain `paid` sum. It reads as a simplification and it is a ~114x regression: because a partial refund *always* stamps a balance onto the credit, that add-back is the only thing crediting affiliates for commission they genuinely kept, worth $5,981.26 across 382 production credits against a $52.58 error across 22. If you disagree with anything in this PR, disagree with the claim that the add-back must survive — everything else follows from it.

**Why the naive restriction was wrong, since it is what a reviewer would suggest.** `.joins(:affiliate_partial_refunds)` is the obvious way to say "credits that have a refund row", and it would have introduced the double-count the tracking issue wrongly claimed already existed: 56 credits in the table carry multiple refund rows (one has four), so a join sums `amount_cents` once per row. The multi-refund spec exists specifically to catch that, and it is the example I would most want re-read.

**Specs are load-bearing, proven not asserted.** Reverting `stats.rb` to `main` reddens exactly the two claw-back cases and leaves the partial-refund cases green. That asymmetry is the correct signature — `main` is already right about partial refunds and wrong only about full reversals — and it is visible in the same evidence image as the green run, so it needs no round-trip to confirm.

**The diagnosis in the driving issue was wrong and I said so before building.** The issue asserted that partially refunded credits stay inside `paid`; zero of 355 do, across the whole table's history, because the balance assignment in `Purchase#process_refund_or_chargeback_for_affiliate_credit_balance` is unconditional while only the refund-row creation is conditional. I posted the measurement and the correction on the tracker rather than shipping against my own incorrect writeup. The number in this PR ($52.58 / 18 affiliates) is measured on the replica this session, not inherited.

**Comment style.** Per AGENTS.md, the method comment explains *why* the add-back exists and why it is restricted, in plain language, without requiring the reader to grep for the incident — a reader on day one should understand it from the comment alone. No private tracker number in the comment (this is a public repo; one slipped into a spec comment and is removed in `318d5200b`).

**Not requesting review yet** — draft until CI is green at head. Local runs of both touched spec files show 123 examples / 98 failures on this branch and 118 / 98 on a detached `origin/main` worktree: the identical 98, all `create(:purchase)` failing card validation because the shared local test DB has no seeded `MerchantAccount.gumroad`. The 5-example delta is this PR's new examples, all passing. On green this goes to whoever owns the affiliate/payout-display lane.

gumclaw:
## Pre-merge adversarial review — verdict READY-TO-MERGE, one finding fixed

Ran the local adversarial review against the branch before merging. It printed the generated SQL for all three call sites, read the refund writers, and ran two mutation checks. The fix itself came back verified correct; one real gap in the tests came out of it and is now closed.

**P2 (fixed in e94091db1) — nothing pinned the EXISTS subquery to the row it is deciding about.** Mutating it to an uncorrelated `AffiliatePartialRefund.where("1 = 1").arel.exists` left all five new examples green. The reason is a property of the fixtures, not of the code: in the full-claw-back examples there were no `affiliate_partial_refunds` rows anywhere in the transactional test database, so even a broken subquery evaluated to false and the assertions passed by accident. Production is never in that state, so the same break would silently reinstate the original bug for every affiliate — and the string-SQL correlation is the most refactor-fragile line in the change.

The new example gives one affiliate both shapes at once: a credit reversed in part (200 earned, 75 clawed back, has a refund row) and a second clawed back in full (no refund row). Correlated it is 125; uncorrelated it is 325. Re-ran the mutation with the example in place and it reddens exactly that one case and nothing else, which is the right signature.

```
6 examples, 0 failures                    # branch as shipped
6 examples, 1 failure                     # subquery mutated to `1 = 1`
  rspec ./spec/modules/user/stats_spec.rb:463
  # keeps a fully clawed-back credit at zero even when the affiliate has other refunded credits
```

Rubocop clean on the spec file.

**Checked and found not to be problems**, recorded so the next reader does not re-derive them:

- **Correlation and aliasing.** `affiliate_credits` is the un-aliased base table in all three call-site scopes, so `affiliate_credits.id` binds to the outer row, and the inner reference binds to the subquery's own FROM. No call site joins `affiliate_partial_refunds`, and the subtraction relation is built from `partially_refunded` rather than `with_remaining_value`, so the EXISTS clause and the refund join never coexist in one query.
- **Set mismatch between the two terms.** Both filter on the same predicate over credits — purchase partially refunded *and* at least one refund row — so a credit is subtracted if and only if it is added back. Nothing in the app ever clears `affiliate_credit_refund_balance` or `affiliate_credit_chargeback_balance` once set, so a credit cannot re-enter `paid` and be counted twice from the other direction either.
- **Over-subtraction.** Per-refund `ceil` rounding in `decrement_balance_for_refund_or_chargeback!` can in principle push the refund total a cent or two past the commission. For credits that have refund rows, `main` and this branch compute the same `amount − refunds`, so the exposure is pre-existing and unchanged; the fix only removes the add-back for credits with no refund row at all.
- **A full reversal reached through several partial refunds.** Each call compares against that call's own refund amount, so each one does create a refund row: add-back equals the commission, subtraction equals the same, net zero, counted once because of EXISTS rather than a join. Covered by the multi-refund example.
- **The `MerchantAccount` setup block.** `MerchantAccount.gumroad` is a plain lookup with no memoization and the created row rolls back with the example's transaction, so it does not leak into neighbouring examples.
- **Index.** `index_affiliate_partial_refunds_on_affiliate_credit_id` already exists, so the subquery is not a new performance surface on the payout-email path.

One pedantic note not worth changing: the code comment says the refund row is created when the credit is reversed by *less* than its full value, while the guard is `!=`. A reversal larger than the commission would also create one, but the decrement arithmetic cannot produce that.

PR #6447, opened 2026-07-27
Abort discount countdown and clear toast timer on unmount

gumclaw:
## What

Two `useRunOnce` callsites started something that outlived the component and never tore it down. Unlike #6423, these did not return a cleanup that got silently dropped — they were written without one at all, so nothing looked wrong at the callsite and the type signature was satisfied.

**`Product/DiscountExpirationCountdown.tsx`** constructed a `Countdown` without assigning it to anything. That object owns a 1-second `setInterval` (`utils/countdown.ts`) and exposes `abort()`, which nothing called. Navigating away from a product page before the discount expired left the interval ticking, calling `setSecondsUntilExpiration` and eventually `onExpiration` on an unmounted component. It now runs in a plain `useEffect` that returns `countdown.abort()`.

The already-expired branch also returns early now instead of falling through. Previously it called `onExpiration()` and then still constructed a `Countdown`, whose own timer would call `onExpiration` a second time — the fix removes that duplicate call as a side effect of the early return.

**`server-components/Alert.tsx`** started its initial 5-second dismiss timeout from `useRunOnce`. The component already had a `clearTimer` helper and a `timerRef`, and the dismiss path used them — the initial timer was the one case never cleared on unmount. The effect now returns `clearTimer`, which also covers timers started later through the `postMessage` path.

Both use a plain `useEffect` with an empty dependency list rather than `useRunOnce`, matching the pattern established in #6431: `useRunOnce` cannot honour a returned cleanup, and now warns in development when a callback returns one.

## Why

Same class of leak as #6423, and low severity for the same reason — these are page-level or short-lived components, and in most cases the timer fires once more into a dead component rather than accumulating. The countdown interval is the one with an ongoing cost: it ticks every second for as long as the discount has left to run, on a page users routinely navigate away from.

## Scope

Callsites 3 and 4 from the issue — the JW Player in `pages/UrlRedirects/Stream.tsx` and the PDF viewer in `pages/UrlRedirects/Read.tsx` — are deliberately **not** in this PR. Both need the third-party teardown APIs and the async cancellation path verified by hand rather than by inspection, and the PDF case additionally needs a cancellation flag for `getDocument` work that resolves after unmount. #6433 stays open for those two.

## Test Results

4 new vitest examples, all passing locally:

```
✓ app/javascript/components/Product/DiscountExpirationCountdown.test.tsx (2 tests)
✓ app/javascript/components/server-components/Alert.test.tsx (2 tests)
  Test Files  2 passed (2)
       Tests  4 passed (4)
```

**Verified load-bearing.** Reverting both component fixes to their `main` versions reddens 3 of the 4:

```
× aborts its Countdown interval when it unmounts
× does not construct a Countdown when the discount has already expired
× clears the initial dismiss timeout when it unmounts
✓ does not start a dismiss timeout when there is no initial alert
```

The fourth passes either way by design — it asserts no dismiss timer is started when there is no initial alert, which was already true before the change and must stay true after it.

**Lint / typecheck — clean on all four touched files:**

- `eslint` — no findings
- `prettier --check` — no changes needed
- `tsc --noEmit -p .` — no diagnostics in the touched files

## Changes from the original fork PR

- Removed two `// eslint-disable-next-line react-hooks/exhaustive-deps` comments. That rule is not configured in this repository, so the disable comments were themselves ESLint errors (`Definition for rule ... was not found`). The explanatory comment about the mount-once contract is kept.
- Replaced an `as number` type assertion in the Alert test, which tripped `@typescript-eslint/consistent-type-assertions` and then `no-unsafe-assignment`. The test now filters `setTimeout` calls by the 5000ms delay and asserts on `clearTimeout` being called, which tests the same behaviour without reaching into the spy's untyped return value.
- Dropped `qa-media/pr-fork-countdown-alert-test-results.txt`, a local test-output log; the results are in this description instead.
- Rebased onto current `main` (the fork branch was based on an older commit, which made its GitHub diff show ~6,700 files).

---

Based on https://github.com/jatinder14/gumroad/pull/1 by @jatinder14.

AI disclosure: model **`claude-opus-5`**, read from the agent's configured default. The agent was asked to fix #6433, found the existing fork PR covering callsites 1 and 2, reviewed it, verified the fix against `utils/countdown.ts` and the `Alert` timer helpers, confirmed the tests fail when the fix is reverted, resolved the lint failures above, and imported it with attribution.


gumclaw:
## Fable-5 adversarial pre-merge review — `READY-TO-MERGE`

Ran the local adversarial review against an 18-item change-class checklist for effect-cleanup / hook-migration diffs. No merge-blocking defect. CI green and Greptile clean were treated as insufficient, per policy.

### Findings

| # | Severity | Location | Finding | Disposition |
|---|---|---|---|---|
| 1 | P3 | `Alert.test.tsx:25,41` | Both tests identified the dismiss timer by filtering on a bare `5000`, duplicated from the literal in `startTimer`. If that delay changed, test 1 fails loudly but test 2 goes **silently vacuous** — its `toHaveLength(0)` passes whenever the filter matches zero calls, whether or not a timer was started. | **Fixed** in `04492703a` — `DISMISS_DELAY_MS` exported from `Alert.tsx`, imported by the test, so they cannot drift. `CONTRIBUTING.md` asks for named constants over raw numbers anyway. |
| 2 | P3 | both test files | Tests pin teardown *invocation*, not teardown *effect*: the countdown test mocks `$app/utils/countdown` entirely, so nothing pins that the real `abort()` clears the interval (`countdown.ts` has no unit test of its own). | **Not fixed — deliberate.** Pinning the component↔`Countdown` contract is the right level for a component test. A `countdown.ts` unit test is a reasonable follow-up but is new coverage for untouched code, not this PR's scope. |
| 3 | P3 | `DiscountExpirationCountdown.tsx:38-41` | The interval runs (re-rendering a `null` component every second) while `secondsUntilExpiration > 7 days` — the 7-day guard only suppresses rendering. | **Pre-existing, out of scope.** Identical before this PR, and arguably by design: the ticking is what makes the countdown appear once it crosses the threshold. |
| 4 | P3 | `DiscountExpirationCountdown.tsx:29-38` | `expiresAt`/`onExpiration` are frozen at first mount, so a live `discountCode` swap at the same tree position keeps ticking against the old expiry. | **Pre-existing, deliberate.** Byte-for-byte the old `useRunOnce` + `useState`-initializer behaviour, preserved on purpose and documented in the comment. A dependency-driven rebuild would be a behaviour change beyond this PR. |

These are the adversarial hypotheses that were checked and **ruled out** — they matter as much as the findings:

- **The duplicate-`onExpiration` claim in the PR description is TRUE**, verified against `countdown.ts` in both environments. Old code with `duration <= 0`: `onExpiration()` fired synchronously, then the constructor set `secondsLeft = duration - 1 ≤ -1`, so the first tick at +1s took the `else` branch → `clearInterval` + `onComplete()` = a **second** `onExpiration`. Under `NODE_ENV === "test"` it was worse — `secondsLeft` is forced to `1`, so an unmounted component got `onTick(1)` at +1s and the duplicate at +2s. The early return removes both.
- **Losing `useRunOnce`'s `hasRun` guard changes nothing observable.** The app never wraps anything in `<React.StrictMode>` (the only grep hit is the comment inside `useRunOnce.ts`). Even hypothetically, setup → cleanup → setup yields exactly one live interval; `Countdown`'s synchronous `onTick(duration)` passes the value state already holds, so `setSecondsUntilExpiration` bails on `Object.is` — no render loop, no wrong first frame. Under Fast Refresh the new code is strictly *more* correct.
- **Returning `clearTimer` by identity is not a stale closure.** `timerRef` is a `useRef`, stable across renders, and `clearTimer` reads `.current` at call time — so the cleanup also correctly cancels timers started later by the `postMessage` handler and `onMouseLeave`.
- **The empty dependency arrays are safe, verified at the call sites rather than assumed.** `onExpiration` in the only production caller (`Product/index.tsx:592-595`) is a stable `useState` setter. Every real `ToastAlert` mount point passes a literal `null` for `initial` (`inertia/layout.tsx`, `layouts/Admin.tsx`); flash messages arrive through `showAlert` → `postMessage`, not through `initial`.
- **Alert test 1's unfiltered `clearTimeout` spy cannot be satisfied by anything else.** React's scheduler captures `setTimeout`/`clearTimeout` at module-evaluation time — before `vi.useFakeTimers()` and the spy install — so React internals can't trip it, and the only other cleanup in the tree calls `removeEventListener`. The one reachable `clearTimeout` is the component's own, guarded by `timerRef.current !== null`.
- **No other `useRunOnce` caller trips the hook's new dev warning.** Swept all 47 callers: concise-arrow forms return `void`/`undefined`, and no block-body caller returns a function. `Stream.tsx` ends with `void createPlayer()`, `Read.tsx` with `void showDocument()` — the `return () => {...}` at `Read.tsx:268` is inside a separate `useEffect`, not the `useRunOnce`.
- **Scope is honest.** The diff touches exactly the 4 code/test files plus the evidence PNG. `Stream.tsx:54` and `Read.tsx:200` still use `useRunOnce`, untouched as stated.
- **Severity framing in the PR body is accurate, not overclaimed.** React 18 removed the "setState on unmounted component" warning, so those calls were silent no-ops. The real pre-fix cost was the immortal 1-second `setInterval` (CPU plus retained closures across Inertia navigations) and the spurious extra `onExpiration` into a still-mounted parent. The Alert leak was bounded at one 5s timeout.

### Evidence

This change has **no pixel delta** — it alters when teardown runs, not what renders — so per the invisible-frontend-fix convention the substitute artifact is the regression run against `origin/main` (RED) beside the same run with the fix (GREEN), rather than staged screenshots that would manufacture a difference:

red-green test evidence

Baseline taken with `git checkout origin/main -- <both components>`, not `git stash`, so the red run reflects main rather than an earlier commit on this branch. Whole test files were run for the revert step, never a filtered example.

### Only verifiable live

Nothing in this diff requires production verification — the leak is deterministic and unit-covered. The one thing no test asserts is the real `Countdown.abort()` clearing its own interval (finding 2); that is exercised by the component contract test plus the existing production behaviour of the countdown itself.

Auto-merge armed now that the review gate has cleared.

PR #6449, opened 2026-07-27
Expose review bodies and submission dates in the v2 API

gumclaw:
## What

Adds a documented v2 read endpoint for product reviews and returns the review's submission date.

- `GET /v2/products/:product_id/reviews` returns the reviews shown on one of the authenticated creator's product pages: `id`, `rating`, `message`, `created_at`, `purchase_id`, `rater_name`, and the seller's `response`. Newest first, `page_key` pagination, 100 per page, available with any read scope. Only publicly visible reviews are returned — the same `visible_on_product_page` scope the product page uses, so a star-only rating with no message is not included.
- `ProductReviewPresenter#product_review_props` now includes `created_at` alongside the existing `is_new` boolean, so the payload behind the public product page carries the timestamp too.
- The API reference gains a "Reviews" resource with the endpoint, parameters, response fields, and examples, plus a sidebar entry.

## Why

The v2 API exposes only aggregate review data today — `average_rating` and `reviews_count` on a product, `product_rating` on a sale. There is no documented way to read the review text, so creators building a custom landing page end up calling `/product_reviews`, the unauthenticated page-data endpoint that powers the public product page. That is an internal surface with no stability contract, and it has a second problem: it never returns when a review was written. `product_review_props` derived `is_new: created_at > 1.month.ago` and dropped the timestamp, so there was no way to sort reviews by date or print "reviewed on X".

Both gaps close here: the timestamp is returned everywhere reviews are, and the read path a creator should use is documented and scope-gated instead of reverse-engineered.

Ref #6436.

## Before / After

The user-visible surface is the public API reference at `/api`. Captured from the real docs components in a Vite harness with the repo's own stylesheets; "before" renders the same page with the Reviews section and its sidebar entry withheld, which is how the page looks on main.

Sidebar, before and after, light and dark:

sidebar before and after

The new Reviews section, desktop:

reviews section desktop

The new Reviews section at 390px:

reviews section mobile

Walkthrough video:

https://raw.githubusercontent.com/antiwork/gumroad/feat/v2-reviews-api/qa-media/pr-6449-walkthrough.mp4

36 assertions ran across the eight captures (2 variants × desktop/mobile × light/dark): the Reviews section and sidebar link present only on the after variant, absent on before, the documented path and the `created_at` field text present only after, and the neighbouring "Public product page" resource — which this diff does not touch — intact in every panel.

## QA steps

1. Create an OAuth application and get a token with any read scope (`view_public` is enough).
2. `curl "https://<preview-host>/v2/products/<product_id>/reviews?access_token=<token>"` for a product that has at least one written review. Expect `success: true` and a `product_reviews` array with `message` and an ISO 8601 `created_at`.
3. Repeat for a product with more than 100 reviews and follow `next_page_key` — the second page should continue from where the first ended, with no repeats.
4. Pass a product ID belonging to another creator; expect `success: false`.
5. Visit `/api` and confirm "Reviews" appears in the sidebar between "Public product page" and "Categories", and that the section renders with its response fields.

## Test Results

test results

- `spec/controllers/api/v2/product_reviews_controller_spec.rb` — 13 examples, 0 failures (new)
- `spec/presenters/product_review_presenter_spec.rb` — 12 examples, 0 failures
- `spec/controllers/product_reviews_controller_spec.rb` — 30 examples, 0 failures

New coverage: the returned body/rating/date, the seller response, the purchaser's display name and the full-name fallback, newest-first ordering, inclusion of video-only reviews, exclusion of deleted reviews and star-only ratings and other products' reviews, another creator's product being rejected, a token with no read scope being rejected, `page_key` pagination across two pages, the id/date inversion regression, and a malformed `page_key` returning 400.

Both non-obvious behaviors were mutation-proved load-bearing (each reverted separately, then restored):

| Reverted to | Result |
|---|---|
| `created_at <= ? AND id < ?` keyset predicate | 13 examples, 1 failure |
| `where(has_message: true)` instead of `visible_on_product_page` | 13 examples, 1 failure |

## Pre-merge review

A local adversarial review confirmed the privacy parity with the public product page, the cross-creator isolation, the route registration (verified by booting `bin/rails routes` — no duplicate-name error despite `api_routes` being drawn twice), the absence of N+1s, the serializer-key blast radius, and the timestamp format. It found one real bug, since fixed in this PR: the first version paged with `created_at <= ? AND id < ?`, which permanently dropped any review whose id was higher than the boundary row's but whose `created_at` was older — reachable because `created_at` is stored at second precision, so two reviews can be inserted out of id order within the same second. The endpoint now uses the corrected keyset predicate the payouts endpoint already uses, pinned by a regression spec.

---

AI disclosure: written by Claude Opus 5 (`claude-opus-5`) running as the Gumclaw agent. It was asked to open PRs for the two asks tracked in #6436 — expose review bodies and the submission date in the v2 API, and document `/v2/media` in the API reference — and this PR covers the first. The agent read the existing review presenter, controller, and API documentation components, wrote the endpoint, specs, and docs section, ran the specs locally, and captured the evidence above.


gumclaw:
Self-review against CONTRIBUTING.md, re-read as a checklist against this diff.

**Scope choice.** The endpoint uses the same read-scope idiom as the twelve sibling v2 read endpoints (`public_api_read_scopes` + `view_public`) rather than `view_sales`. Everything it returns is already on the public product page for anonymous visitors, including the purchase's external ID, so requiring a sales scope would be stricter than the data warrants. `public_api_read_scopes` builds a fresh array per call, so the `.concat` is not mutating shared config.

**Why not extend an existing resource.** Reviews could have been folded into the product payload, but a product with tens of thousands of reviews would then make `GET /v2/products` unusable. A separate paginated collection is the shape the rest of the API uses for anything unbounded.

**Response construction.** `render_response(true, {...}.merge(additional_response))` is what `success_with_object` expands to for a non-nil object, so `as_json(api_scopes:)` still applies. I used the explicit form because the payload is a hand-built hash rather than model `as_json`, and wrapping it in `success_with_object` would read as if it were serializing records.

**Preloading.** `includes(:response, purchase: :purchaser)` covers every association `review_json` touches; `external_id` derives from attributes. The `visible_on_product_page` scope left-joins approved videos, which cannot duplicate rows because `ProductReviewVideo#approved!` marks sibling approved videos deleted under a row lock.

**Presenter change.** Adding `created_at` to `product_review_props` widens the public product page payload. The one exact-hash assertion on that method is updated in this diff; the frontend `Review` type gains the field, and the other three `Review` types in the codebase are unrelated local types fed by different presenters.

**What I deliberately did not do.** No `video` field on the v2 payload — video-only reviews are returned (with `message: null`, now documented as nullable) but exposing video URLs through the API is a larger decision than this PR. No change to the public `/product_reviews` endpoint's shape beyond the added timestamp.

**Pagination.** The first version of this used `created_at <= ? AND id < ?`, copied from the sales and subscribers endpoints. That predicate is wrong for a `created_at desc, id desc` ordering and silently drops rows; the payouts endpoint already has the corrected form and this now matches it. There is a regression spec with an id/date inversion, and reverting the predicate reddens exactly that spec.

gumclaw:
CI is fully green on \`f7b7a19\`-tip: all 24 Test Fast shards, all 50 Test Slow shards, Test Minitest, Lint Ruby, Lint JS/TS, Build images, Buildkite, and \`ci/green\` all pass. Greptile is at 5/5 and its one P1 — the keyset cursor predicate skipping a review with an older \`created_at\` but a higher id — is fixed and the thread is resolved; the fix is pinned by a pagination regression test in \`spec/controllers/api/v2/product_reviews_controller_spec.rb\`.

@gianfrancopiana over to you for review and merge. Nothing is blocked on me. This is the backend half of #6436 (Sahil asked for PRs on both halves); #6450 is the docs-only other half, also green and assigned to you.

The two things worth your eye:

1. **Scope choice.** The endpoint is gated on any read scope (\`public_api_read_*\`), the same idiom the twelve sibling v2 read endpoints use. If you would rather see a dedicated review scope, say so and I will add it.
2. **Visibility filter.** Only reviews that pass \`visible_on_product_page\` are returned — the identical scope the public product page renders from — so a star-only rating with no message is excluded. That keeps the API from exposing anything the page does not already show publicly.

Happy to make any changes you want; ping me and I will turn them around.

PR #6450, opened 2026-07-27
Document the v2 media library endpoints in the API reference

gumclaw:
## What

Documents the creator media library in the public API reference at `/api`. Adds a "Media library" resource covering the three endpoints that already exist:

- `GET /v2/media` — list the files in your library (`view_profile` scope)
- `POST /v2/media` — upload from a public URL Gumroad downloads for you, or from a signed blob ID (`edit_profile` scope)
- `DELETE /v2/media/:id` — remove a file (`edit_profile` scope)

Each entry lists parameters, response fields, and a cURL + response example, and the section explains why the library exists: custom product and profile pages render under a Content Security Policy that only permits images from Gumroad's own CDN, so an image hosted elsewhere will not load there. Limits documented from the code: 10 MB per image, type determined by inspecting the bytes rather than the extension, SVG rejected, 500 files per account.

No behavior changes — documentation only.

## Why

`Api::V2::MediaController` and `CreatePublicMediaService` shipped and are live, but neither appears anywhere in the API reference. The result is that creators who want a supported way to host images for a custom landing page cannot find one, and instead reverse-engineer a workaround: upload the image into a product's description editor and harvest the resulting `public-files.gumroad.com` URL from the markup. That workaround ties an account-level asset to an unrelated product and breaks whenever the description is edited.

Files uploaded through this endpoint are scoped to the seller rather than to a product, and the returned URL is on a host the custom-HTML CSP already allows — which is exactly what the workaround is groping for. The gap is documentation, not product.

Ref #6436.

## Before / After

Sidebar, before and after, light and dark:

sidebar before and after

The new Media library section, desktop:

media section desktop

The new Media library section at 390px:

media section mobile

Walkthrough video:

https://raw.githubusercontent.com/antiwork/gumroad/docs/v2-media-api-reference/qa-media/pr-6450-walkthrough.mp4

Captured from the real documentation components in a Vite harness with the repo's own stylesheets; the "before" variant renders the same page with the Media library section and its sidebar entry withheld, which is how the page looks on main. 36 assertions ran across the eight captures (2 variants × desktop/mobile × light/dark): section and sidebar link present only on the after variant, all three `/media` paths documented, the CSP rationale and the `edit_profile` scope named, and the neighbouring Thumbnails resource — which this diff does not touch — intact in every panel.

## QA steps

1. Visit `/api` and confirm "Media library" appears in the sidebar between "Thumbnails" and "Variant categories".
2. Click it and confirm the section renders three endpoints with their response fields expanded.
3. Follow the documented `GET /v2/media` example with a token holding `view_profile`; expect the same shape shown in the example response.

## Test Results

test results

- `spec/controllers/public_controller_spec.rb` — 13 examples, 0 failures (covers the `/api` page render)
- `npx tsc --noEmit` — no errors in the API documentation components
- `npx eslint` and `prettier` — clean on the touched files

The documented facts were read out of the code rather than assumed: scopes from `Api::V2::MediaController`'s `doorkeeper_authorize!`/`require_oauth_scope!` lines, the 10 MB cap, SVG exclusion and 500-file quota from `CreatePublicMediaService`, the rate limit from `config/initializers/rack_attack.rb`, the direct-upload flow from `Api::V2::DirectUploadsController`, the CSP host list from `RendersCustomHtmlPages`, and the response shape from `PublicFilePresenter#props` plus the controller's `media_json`.

## Pre-merge review

A local adversarial review verified every documented claim against the code and found six inaccuracies in the first draft, all fixed here:

| Finding | Fix |
|---|---|
| `extension`, `file_size`, `url`, `file_group` documented as always present | marked nullable, with the condition for each |
| `file_group` described as "always image" | described as derived from the filename extension, which is what `PublicFile#set_file_group_and_file_type` does |
| DELETE described as purging immediately | the record leaves the library immediately, the blob is queued via `purge_later` |
| `signed_blob_id` documented with no way to obtain one | added the `POST /v2/direct_uploads` with `purpose=media` flow |
| `name` said to default to the uploaded filename | it defaults to the filename with the extension stripped |
| CSP described as "only Gumroad's CDN" | the policy also allows `data:` and `blob:`; wording now says external image URLs are what's blocked |

It also flagged three operational behaviors worth documenting, now included: the 20-uploads-per-10-minutes throttle, rejection of `account`-only tokens, and the suspended/closed-account 403. Example IDs were corrected to the real 16-character `public_id` format.

---

AI disclosure: written by Claude Opus 5 (`claude-opus-5`) running as the Gumclaw agent. It was asked to open PRs for the two asks tracked in #6436 — expose review bodies and the submission date in the v2 API, and document `/v2/media` in the API reference — and this PR covers the second. The agent read the media controller, the ingestion service, and the presenter to derive the documented behavior, wrote the docs section, and captured the evidence above.


gumclaw:
Self-review against CONTRIBUTING.md, re-read as a checklist against this diff.

**Nothing here changes behavior.** No controller, service, or model is touched — this adds a documentation section describing endpoints that already ship. The risk in a docs PR is not breakage, it is stating something that is not true, so every sentence was checked against the code and the sources are listed in the Test Results section.

**Placement.** "Media library" sits after Thumbnails because both are about images a creator uploads; Thumbnails are per-product, media files are per-account, and the section says so explicitly since that distinction is the whole reason the endpoint exists. The sidebar entry and the `ApiResource` id match (`#media`), and the page's hash-scroll effect resolves it.

**Wording I chose carefully.** The section leads with why the library exists rather than what the endpoints are, because the creator problem it solves ("my image does not load on my custom page") is what someone will be searching for. The CSP sentence names what is blocked rather than claiming an allowlist, since the real policy also permits `data:` and `blob:`.

**What I documented conservatively.** `file_group` and `status` are described as what they are derived from rather than as fixed values, so the docs do not become wrong the first time a non-image type is accepted. Nullable fields are marked nullable even where they are nearly always present.

**What is still missing and why.** There is no statement about how long a hosted URL lives or whether re-uploading the same file dedupes, because I could not establish either from the code with confidence, and a guess in an API reference is worse than a gap.

**Verification.** The 10 MB cap, SVG exclusion, 500-file quota, scope requirements, rate limit, and direct-upload flow are each cited to a file in the PR body. Example JSON blocks were parsed to confirm they are valid JSON, and the example IDs match the 16-character `public_id` format the model validates.

gumclaw:
CI is fully green: all 24 Test Fast shards, all 50 Test Slow shards, Test Minitest, Lint Ruby, Lint JS/TS, Build images, Buildkite, and \`ci/green\` all pass. Greptile is at 5/5 with no findings and no open threads.

@gianfrancopiana over to you for review and merge. Nothing is blocked on me. This is the documentation half of #6436; #6449 is the backend half, also green and assigned to you.

Nothing here changes behavior — no controller, service, or model is touched. \`Api::V2::MediaController\` and \`CreatePublicMediaService\` are already live; this only writes them into the API reference so creators can find them instead of reverse-engineering a workaround. Every limit stated in the docs is read off the code: 10 MB per image, content type determined by inspecting the bytes rather than trusting the extension, SVG rejected, 500 files per account.

The one thing worth a second opinion: I explain in the section why the library exists at all — custom product and profile pages render under a Content Security Policy that only allows images from Gumroad's own CDN, so an image hosted anywhere else silently fails to load there. That is the actual reason a creator needs this endpoint, and it is not documented anywhere today. If you think it belongs in the help center rather than the API reference, I will move it.

gumclaw:
## Fable-5 adversarial pre-merge review — was `CHANGES-REQUIRED`, now fixed

I had assigned this to @gianfrancopiana without running the adversarial review first. That was wrong on two counts: Sahil's standing rule is that green CI plus a clean fable review plus verified QA media means I arm auto-merge myself rather than parking it on a human — and the review, once run, **found a false statement in public API documentation**. Handing it over would have shipped that.

### P2 — the documented null branch is unreachable (fixed in `d4f268e56`)

The reference claimed `extension` is *"null when the uploaded filename had none."* It never is. `PublicFile#set_file_group_and_file_type` does:

```ruby
self.file_type ||= original_file_name.split(".").last
```

`"logo".split(".").last` is `"logo"`, not `nil`. So a blob reserved as `blob[filename]=logo` returns `"extension": "LOGO"` — the whole basename, uppercased. `original_file_name` is always present (validated at `public_file.rb:14`, set from the attachment), and the URL-upload path appends a sniffed extension anyway. An integrator branching on `extension === null` would never enter that branch and would instead receive a surprising value.

`file_group` had the same root cause. `FILE_REGEX` entries are **unanchored**, so with no extension the whole basename is matched: a file named `my-movie` yields `"video"` via `/movie/i`, not `null`.

I verified both against the model and the initializer rather than taking the finding on trust.

### P3s, both fixed

- **Field order contradicted the examples.** `MEDIA_FIELDS` listed `file_group` before `status`, but the real payload is the presenter's props (ending in `status`) merged with `file_group` and `created_at` — which both example responses already showed correctly. Only the field list disagreed.
- **The Scopes section contradicted this PR.** It still said `account` grants *"full access to all API endpoints"*, while `require_oauth_scope!` (`api/v2/base_controller.rb:16`) exists specifically to refuse an otherwise-valid account-only token on endpoints with a narrower boundary. Amended.

### Left as-is

`POST /v2/direct_uploads` is still undocumented as a standalone endpoint, so the required `blob[filename|byte_size|checksum|content_type]` shape isn't discoverable from the reference. Pre-existing gap, wider than this PR, and the inline steps mitigate it. Worth its own change.

### What the review confirmed against code

All three routes exist verbatim (`routes.rb:82`, `resources :media, only: [:index, :create, :destroy]`); scopes match exactly (`view_profile` for index, `edit_profile` for create/destroy); all eight response fields traced to `public_file_presenter.rb` and `media_controller.rb:72-77` with no phantom or undocumented field; every behavioural claim checked — newest-first `order(id: :desc)`, 10 MB cap, magic-byte sniffing, SVG rejection, the 500-file cap, rate limit 20/10min POST-only, suspended sellers blocked from upload but not delete. The section id `media` is unique and wired into `Api.tsx` the same way as its siblings. Diff is documentation-only.

eslint, prettier and tsc clean. Taking this back as assignee since it needed my work, not a reviewer's.

PR #1328, opened 2025-09-24
Fix flaky Dropbox test by making file order expectations independent

HiAbhishekh:
Part of https://github.com/antiwork/gumroad/issues/1127

AI Disclosure: Used for only for syntax, logic and code review Manually

### Problem

The Dropbox test was flaky due to order-dependent file expectations. The test used `product.alive_product_files.first` and `product.alive_product_files.last` which caused failures when files were uploaded in different orders.

### Solution
- Replaced order-dependent `first`/`last` expectations with dynamic mapping of all files
- Eliminates race conditions that caused test flakiness in CI

### How This Addressed the Issue:
Problem:
Order-dependent expectations - test expected files in exact sequence
Race conditions - files uploaded in different orders caused failures

Solved:
product.alive_product_files.map do |file|
 
end

Fixes https://github.com/antiwork/gumroad/issues/1127

PR #6454, opened 2026-07-28
Register the RBI e-mandate cap in the charge's own currency

gumclaw:
## What

The RBI e-mandate cap registered with an Indian card's first charge governs every future off-session renewal on that card. It has always been sized in canonical USD cents, and it is built by the caller (`Purchase#mandate_options_for_stripe`) *before* we know whether this charge will present in the buyer's currency.

Stripe reads `mandate_options[:amount]` in the mandate's own currency, and the mandate inherits the PaymentIntent's currency unless we send one explicitly. `StripeChargeProcessor#create_payment_intent_or_charge!` creates the intent with `currency: processor_currency || Currency::USD` and then merges `mandate_options` in verbatim — so on a buyer-currency charge a cap of `1000` registered as **CA$10.00**, or **₹10.00**, instead of $10.00.

The cap is now converted where the presentment currency is actually known, in `Charge::CreateService`.

## Why this is the priority item in #1328

Every renewal then exceeds the cap. Per Stripe's [India recurring payments](https://docs.stripe.com/india-recurring-payments) guide, a charge above `mandate_options[amount]` (or ₹15,000, whichever is less) requires fresh additional-factor authentication — so each renewal is declined off-session until the buyer comes back and re-authorises.

For an Indian buyer on a $10/month subscription, the cap registers as roughly **₹10 against a charge of about ₹880**. The failure is silent, delayed by a full billing period, and lands only on subscriptions.

The same guide is what confirms the mechanism rather than my inferring it: *"If your customer initiates a subscription with Adaptive Pricing, and mandates support the selected currency, Stripe automatically creates an e-mandate in that currency."*

## Why the fix lives here and not in the model

The presentment amount does not exist when `Purchase#mandate_options_for_stripe` runs — it is computed inside `CreateService#buyer_currency_presentment_processor_args` from the locked quote. Converting in the model would mean threading the quote into `Purchase`. Note the existing `with_currency: true` variant hardcodes `"usd"` and is only reached from the setup-intent path, which does not charge and has no presentment, so it is not the fix site.

Two deliberate choices:

- **Scale by the charge's own canonical/presentment pair, not the raw FX rate.** `amount_cents` and `processor_amount_cents` are the same money in the two currencies, already rounded the way this charge was quoted, so the ratio carries the rate and the rounding together.
- **Floor the result at the amount being charged.** A cap that rounds a subunit below today's total would make the very first renewal at an unchanged price need re-authorisation.

Canonical charges are untouched: no presentment args, the intent is created in USD, and the cap is returned exactly as built — including having no `currency` key.

## Before / After

No UI surface: this changes a parameter sent to Stripe on a charge. Per the invisible-change convention the artifact is the red→green spec run rather than staged screenshots.

red-green

## Test Results

Three new examples in `spec/services/charge/create_service_spec.rb`:

| Example | Asserts |
|---|---|
| CAD presentment charge | cap converts `10_00` → `12_50` with `currency: "cad"`; `reference` / `amount_type` / `interval` / `supported_types` preserved |
| INR presentment charge | cap is never below the `277_77` being charged |
| canonical USD charge | cap stays `10_00`, no `currency` key added |

**Verified load-bearing.** Reverting only the service to `origin/main` fails the two presentment examples — the cap stays at its canonical value — while the canonical-USD control keeps passing. That asymmetry is what proves the fix is scoped to presentment charges and does not disturb the USD path.

```
3 examples, 2 failures   # on origin/main
3 examples, 0 failures   # with the fix
```

`rubocop`: 2 files inspected, no offenses detected.

## QA steps

1. Indian test card `4000003560000123` on a monthly subscription priced in USD, buying from a country that presents in a non-USD currency (so the checkout shows a local total).
2. Complete the initial charge.
3. In the Stripe dashboard open the resulting PaymentIntent → **Mandate**.
4. Confirm the mandate's maximum amount is in the **charge's** currency and is at least the amount charged — before this change it was the USD number reinterpreted in that currency.
5. Regression: repeat with a US buyer (canonical USD charge) and confirm the mandate maximum is unchanged and carries no explicit currency.

## Scope

Group **A1** of antiwork/gumroad-private#1328. The remaining Group-A sites (dispute-reinstated Connect arithmetic, transfer-reversal amounts, dispute evidence, fee-retention / AU backtax / admin refund scaling) each get their own PR.

Live-ramp note: buyer-currency charging has been at 100% since 2026-07-23, so this is a change to a live money path. It carries no flag because it corrects a parameter that is already wrong in production for every presentment subscription.

---

## AI disclosure

Model: `claude-opus-5`, read from the agent's configured default.

The agent was asked to open a PR for each item in the #1328 enumeration. For this one it traced the mandate options from `Purchase#mandate_options_for_stripe` through `Order::ChargeService` and `Charge::CreateService` into `StripeChargeProcessor#create_payment_intent_or_charge!` to confirm the canonical cap really does reach a buyer-currency intent, read Stripe's India recurring-payments documentation to confirm how an unqualified mandate amount is interpreted, then wrote the fix, the three specs, and verified the specs fail against `origin/main`.


gumclaw:
The `Test Fast 16` failure here is not from this PR. It is `main`'s, and every open PR based on current `main` has it.

```
undefined method 'affiliate_credits_sum_total' for #<User:...>
```

Six examples in `spec/modules/user/stats_spec.rb`, none of which this PR touches — its diff is `app/services/charge/create_service.rb` and `spec/services/charge/create_service_spec.rb` only.

Cause is a merge race between two separately-green PRs, both of which are in this branch's history:

- #6445 (`716f5830c`, 23:37 UTC) added six examples calling `User#affiliate_credits_sum_total`
- #6420 (`b9e0dfa22`, 23:03 UTC) **deleted** that method, replacing the dashboard figure with the unfiltered `#affiliate_credits_total_revenue_cents`

Each branch's CI ran against a base lacking the other's change, so neither caught it. The method is absent from the entire tree on `main`, so this is deterministic, not a flake, and re-running will not clear it.

`Test Fast 3` in the same run is only fail-fast collateral — its log ends `1700 examples, 0 failures` followed by `The operation was canceled.`

Fixed on `main` in #6456, which points the six examples at `#affiliate_credits_sum_for_credits_created_between` (the remaining public caller into the same private method under test) rather than reviving the deleted method — #6420's deletion was deliberate and carried a measured request-time win. Once #6456 lands, merge `main` here and this shard should clear. Nothing to change on this branch.

gumclaw:

gumclaw:

```
```

gumclaw:

PR #6455, opened 2026-07-28
Tear down the stream player and PDF viewer when the reader unmounts

gumclaw:
## What

The last two `useRunOnce` callsites from #6433 — callsites **3 and 4**. Both start something that outlives the component, and unlike #6423 neither returned a cleanup that got dropped: none was ever written, so nothing looked wrong at the callsite.

**`pages/UrlRedirects/Stream.tsx`** created a JW Player, registered six `player.on(...)` handlers and a throttled tracking callback, and removed none of them. Navigating away from a stream left the player instance, its media element and its handlers attached for the rest of the page's life — and the throttled `"time"` handler kept calling `trackMediaLocationChanged` against a component that no longer existed. It now runs in a plain `useEffect` whose cleanup cancels the throttle **first** (a pending call must not fire against a player that is about to be destroyed) and then calls `player.remove()`, JW Player's own teardown for handlers plus DOM.

**`pages/UrlRedirects/Read.tsx`** registered its two `eventBus` handlers as anonymous functions and ran async `getDocument` work with no cancellation. The handlers are now named, because **`eventBus.off` is a no-op against a different function object** — anonymous handlers were unremovable by construction, so this had to change before a cleanup could work at all. The cleanup removes both handlers, calls `viewer.cleanup()`, and destroys the `PDFDocumentProxy`.

## The async race both needed

Both setups are async and can outlive the mount, so both carry a cancellation flag:

- **Stream**: a player that resolves after unmount is removed immediately rather than having handlers wired onto it.
- **Read**: a document that resolves after unmount is destroyed instead of being handed to a dead viewer. That path previously also called `setPageCount`/`setIsLoading` on an unmounted component and kept the entire parsed PDF in memory.

## Teardown APIs verified, not assumed

Read out of the installed packages rather than inferred:

| API | Source |
|---|---|
| `remove(): JWPlayer` | `@types/jwplayer/index.d.ts:1676` |
| `EventBus#off(eventName, listener)` | `pdfjs-dist/types/web/event_utils.d.ts:33` |
| viewer `cleanup()` | `pdfjs-dist/types/web/pdf_viewer.d.ts:358` |
| `PDFDocumentProxy#destroy()` | `pdfjs-dist/types/src/display/api.d.ts` |

`setDocument` is typed as requiring a document, so the viewer is never handed a null one — the document is released via `destroy()` instead.

## Before / After

No pixel delta: this changes when teardown runs, not what renders. Per the invisible-fix convention the artifact is the red→green run rather than staged screenshots.

red-green

## Test Results

Four new examples in the existing `Read.test.tsx`. It needed pdfjs module stubs, since the reader imports pdfjs lazily inside the effect and would otherwise fetch a real worker.

| Example | Asserts |
|---|---|
| handlers + viewer | `off` called for both events and `viewer.cleanup()` on unmount |
| handler identity | each handler removed by the **same function reference** it was registered with |
| unmount-during-load | a document resolving after unmount is destroyed and never reaches the viewer |
| loaded document | a loaded document is destroyed on unmount |

```
26 examples, 4 failures   # Read.tsx reverted to origin/main
26 examples, 0 failures   # with the fix
```

**Verified load-bearing**: reverting only `Read.tsx` reddens exactly those 4 while all **22 pre-existing EPUB examples stay green** — that asymmetry is what shows the new coverage is scoped to the PDF teardown and did not weaken the existing suite.

`eslint`, `prettier --check` and `tsc --noEmit` are clean on all three files.

## QA steps

**Stream:** open a purchased video from the download page, let it play a few seconds, then navigate away (browser back, or a sidebar link). In DevTools confirm no further `media_location` requests fire after leaving, and that the `<video>`/player DOM is gone. Then re-enter and confirm playback still resumes from the saved position.

**Read:** open a purchased PDF and navigate away **while it is still loading** (throttle the network to widen the window). Confirm no console error about setting state on an unmounted component, and that memory does not retain the document. Re-enter and confirm the reader still resumes on the right page, and that arrow-key paging and zoom still work.

## Scope

Completes #6433 — callsites 1 and 2 shipped in #6447.

---

## AI disclosure

Model: `claude-opus-5`, read from the agent's configured default.

Asked to actively work these two callsites rather than leave them parked, then close the issue. The agent read both components, read the JW Player and pdfjs type definitions to confirm the real teardown APIs, wrote both fixes plus the cancellation flags, added four regression examples to the existing spec file, and verified they fail against `origin/main` while the pre-existing suite stays green.


gumclaw:
## Fable-5 adversarial pre-merge review — `READY-TO-MERGE`

Ran the local adversarial review against a 21-item checklist for effect-cleanup / third-party-teardown diffs. The review found a **P1 that I had caught myself mid-flight**, plus two real P3s now fixed. Notably it verified its claims against pdfjs's *bundled source*, not its type definitions — which is what made the second P3 actionable.

### P1 — caught and fixed during the review (`2e39b49a9`)

The original cleanup called `throttledTrackMediaLocation.cancel()`. **lodash's `throttle` defaults to `trailing: true`**, so before this PR a position write scheduled in the last 10 seconds of watching still fired after the buyer navigated away. Cancelling it would have silently dropped up to `LOCATION_TRACK_EVENT_DELAY_MS` of progress, so the buyer would resume *earlier* than where they actually stopped.

That is a user-visible regression introduced by a leak fix — worse than the leak. Now `flush()`, which invokes the pending write synchronously while the player is still alive (`getPlaylistIndex()` inside `trackMediaLocation` needs it) and only then calls `remove()`. `trackMediaLocationChanged` is a plain fetch, so it completes independently of the unmounting component.

### Findings table

| # | Severity | Location | Finding | Disposition |
|---|---|---|---|---|
| 1 | **P1** | `Stream.tsx` cleanup | `cancel()` drops the trailing position write; `trailing: true` means main still delivered it | **Fixed** in `2e39b49a9` — `flush()` before `remove()` |
| 2 | P3 | `Read.tsx:239` | `isCancelled` checked after the dynamic imports but not before building the viewer — unmounting inside that window still constructed a viewer on a detached container, overwrote the nulled ref, and issued the full `getDocument` fetch | **Fixed** in `427fec7e0` |
| 3 | P3 | `Read.tsx` teardown | `viewer.cleanup()` is the wrong primitive: in `pdf_viewer.mjs` it only resets pages whose rendering already **finished**. `setDocument(null)` is what dispatches `pagesdestroy` and calls `_cancelRendering()` / `_resetView()` | **Fixed** in `427fec7e0` |
| 4 | P2 | `Stream.tsx` | JW Player teardown ships with **no tests** | **Not fixed — acknowledged.** See below. |
| 5 | P3 | `Read.test.tsx:188` | reference-equality test only pins `pagesinit` and asserts membership, so a double-registration would pass | **Not fixed** — narrow gap, and the first test independently asserts `pagerender` removal |
| 6 | P3 | both files | `[]` deps read `usePage()` props; stale IDs if Inertia re-rendered in place | **Pre-existing**, byte-for-byte main's behaviour under `useRunOnce`; no in-app link navigates stream→stream today |
| 7 | P3 | `Read.tsx` | a rejected `getDocument` leaves the loading overlay up forever | **Pre-existing**, identical on main; the new code handles unmount correctly in that state |

On finding 3 I checked the reviewer's claim rather than taking it: `pdf_viewer.mjs` line ~10808 confirms `setDocument` dispatches `pagesdestroy` then `_cancelRendering()` + `_resetView()`, and `if (!pdfDocument) return` sits immediately after — so null is genuinely its detach path even though the `.d.ts` types the parameter as required.

### On the missing Stream tests (P2, accepted)

The reviewer is right that this is the more complex half shipping untested, and right that stubbing is not the obstacle — `jwPlayer.ts:14-16` guards on `typeof jwplayer === "undefined"` *specifically so tests can supply a stubbed global*. I am not adding that here: a `Stream.test.tsx` needs a JW Player global stub plus fake timers for the throttle, and it belongs in its own change rather than bolted onto a teardown fix. Filed as follow-up work rather than pretended away.

What that means concretely: **the flush-vs-cancel bug above would be invisible to the suite today.** It was caught by reading lodash's defaults, not by a red test. Worth stating plainly.

- **No un-removed player path.** `!containerRef.current` returns before creation; the `isCancelled` branch removes the just-created player; `playerToRemove` is assigned synchronously after the cancellation check with nothing throwable between; a `createJWPlayer` rejection leaves no player. Every live player is reachable from the cleanup.
- **`remove()` exists and is correctly typed** — `@types/jwplayer/index.d.ts:1698` (`setup(): JWPlayer`) and `:1676` (`remove(): JWPlayer`), and `createJWPlayer` returns `jwplayer(id).setup(...)` verbatim.
- **No lost or doubled teardown across the two `teardown` reassignments.** Unmount before load: handler-version runs once, and the resolve path destroys the pdf and returns *before* the wrapping reassignment. Unmount after load: the wrapped version runs `off/off/setDocument(null)/cleanup/destroy` exactly once — the check, reassignment and `setDocument` are one synchronous block.
- **`pdfViewerRef.current = null` is safe.** Every reader is guarded (`updatePage:166`, `zoomIn:188`, `zoomOut:194`, `onPagesInit` optional-chains). Both effect cleanups run in one synchronous unmount commit, so no keydown lands between the null and the listener removal.
- **EPUB path untouched** — `Read.tsx:135` dispatches to separate `EpubReader`/`PdfReader` components; the new effect lives in `PdfReader` only and cannot run for `file_type: "epub"`.
- **All four new tests bite**, each for a distinct reason, and none passes against unfixed code.
- **No other `useRunOnce` caller returns a function** — swept ~30 callsites; nothing will trip the dev warning from #6431.
- **Comments verified against real sources**: `EventBus._off` removes by `evt.listener === listener` identity and silently returns on no match, so the "no-op against a different reference" claim is literally true.
- **StrictMode is unreachable** — the app never renders under it, so double-mount races are not a concern here.

### Evidence

No pixel delta, so the artifact is the red→green run: **26 examples, 4 failures** with `Read.tsx` reverted to `origin/main`, **26 passed** with the fix, and the 22 pre-existing EPUB examples green on both sides.

red-green

### Only verifiable live

The JW Player DOM teardown (`setup()` element replacement, `remove()` restoring the original container) is verified against the type definitions and documented behaviour, not executed — there is no JW Player stub in the suite. The QA steps in the description cover it: play a video, navigate away, confirm no further `media_location` requests and that the player DOM is gone, then re-enter and confirm playback resumes from the saved position.

gumclaw:

gumclaw:

```
```

gumclaw:

PR #6457, opened 2026-07-28
Offer local payment methods regardless of the charging account's currency

gumclaw:
## What

Removes the US-dollar settlement requirement from the forced-currency local payment method lane, so iDEAL and Bancontact render on euro carts, and UPI on rupee carts, no matter what currency the charging Stripe account holds its balance in.

Two gates went away, both asking the same question in different places:

- `Checkout::BuyerCurrencyEligibility#method_forced_decision` no longer returns `:unsupported_settlement_currency` for the lane as a whole. The one remaining settlement check is scoped to the quoted (US-dollar-priced) case further down, which is the only case that actually mints a Stripe FX quote.
- `Checkout::PaymentMethodResolver#forced_currency_methods` no longer calls `forced_currency_settlement_supported?`, and that private method plus the now-unused `BuyerCurrencyEligibility.usd_holding_settlement_account?` helper are deleted.

Nothing else about the lane changes. The per-method launch flags, the cart-shape rule (the whole cart must be priced in the currency the method forces), the buyer-country locks on UPI and Pix, the purchasing-power-parity matrix, and the per-account Stripe capability intersection all still apply in the same order.

One downstream call site had to move with the gate. `Order::PreparePaymentIntentService#pix_payment_method_options` sent `amount_includes_iof: "always"` on every Pix intent, which was safe only because the settlement gate guaranteed every Pix intent was created on a Gumroad-held account. Widening removes that guarantee: a Brazilian connected account can now reach Pix, and that charge is domestic — the money never leaves Brazil, so there is no foreign exchange and no IOF for Stripe to price. The option is now scoped to charges on a Gumroad-held account, the same condition `Purchase#pix_iof_fee_per_thousand` already uses, so the option Stripe receives and the fee the seller is billed can never disagree about whether the tax applies. `expires_after_seconds` stays unconditional — it is about how long we hold the purchase open, not about who settles. Today's traffic is unaffected, because every Pix intent is currently created on the platform account.

## Why

A product priced in the method's forced currency is charged at its listed price. There is no FX quote anywhere in that flow, so the account's balance currency cannot make the charge fail — Stripe accepts a euro PaymentIntent on a euro-settling account, and that is in fact the simplest case, settling natively with no conversion at all. The dollar requirement was carried over from the FX-quote lane, where the seller's settlement configuration genuinely decides whether Stripe honours the quote.

The practical effect was that the methods were hidden from the sellers they exist for. Measured against the production read replica over the last 7 days, of the euro-listed, rupee-listed and real-listed one-time purchases that charged on a seller-owned account:

| Cart | Charging account | Gate today | Purchases | Volume |
|---|---|---|---|---|
| Euro | Gumroad-managed (destination charge) | offered | 4,577 | $105,427 |
| Euro | Stripe Connect, settles in euros | **blocked** | 109 | $2,218 |
| Euro | Stripe Connect, settles in Australian dollars | **blocked** | 28 | $772 |
| Euro | Stripe Connect, settles in pounds / krona | **blocked** | 9 | $98 |
| Euro | Stripe Connect, settles in dollars | offered | 9 | $32 |
| Rupee | Gumroad-managed (destination charge) | offered | 52 | $239 |
| Real | Stripe Connect, settles in reais | **blocked** | 3 | $91 |
| Real | Gumroad-managed (destination charge) | offered | 7 | $77 |

Every blocked row is a checkout that could have completed. Note that the euro/euro-Connect row is not a niche: euro-settling connected accounts are the normal shape for a European seller who linked their own Stripe account.

The dollar-holding requirement had already been narrowed twice for the same reason, each time after it hid a working method. First the marker-aware `usd_settling_merchant_account?` turned iDEAL dark platform-wide on 2026-07-23: enabling the iDEAL and SEPA capabilities made the platform account settle euros in euros, the euro mismatch marker was recorded, and the tab stopped rendering for every Gumroad-managed seller. Then the stored-currency check that replaced it was rescoped in #6386 to ask about the account the intent is created on rather than the destination account, which restored UPI for Gumroad-managed sellers settling in rupees. This change finishes that sequence by removing the question entirely from a lane that never needed to ask it.

Tracking issue: antiwork/gumroad-private#1442.

## Before / after

The same cart — a euro-priced product from a Belgian Stripe Connect seller whose account settles in euros and has the iDEAL and Bancontact capabilities active — resolved through `Checkout::PaymentMethodResolver` on `main` and on this branch. The list below is what the Stripe Payment Element renders as tabs to the buyer.

before and after

There is no markup, styling, or copy in the diff, so there is nothing else to show visually.

## Test plan

Run locally on an isolated test database:

```
bundle exec rspec \
  spec/services/checkout/payment_method_resolver_spec.rb \
  spec/services/checkout/buyer_currency_eligibility_spec.rb \
  spec/presenters/checkout/stripe_payment_presenter_spec.rb \
  spec/services/order/prepare_payment_intent_service_spec.rb \
  spec/services/charge/method_forced_presentment_spec.rb
```

`353 examples, 0 failures`.

Manual verification once merged: on a checkout for a euro-priced product from a seller with `checkout_local_method_bancontact` active and a euro-settling connected Stripe account, the Payment Element should now show the Bancontact tab. A dollar-priced cart from the same seller must still show only card and Link.

## Test results

Five new or changed specs cover the widened gate. Each was proven load-bearing by restoring the old gate with the new specs in place — that run is red on exactly those five and green everywhere else:

- resolver: a launched method is offered when the charging account holds a non-dollar balance
- resolver: a launched method is offered to a Stripe Connect seller settling in the method's own currency
- resolver: a launched method is still withheld from a Connect seller whose account has not activated the capability (the gate that does the real work now)
- eligibility: a non-dollar-settling account is eligible when the product is priced in the forced currency
- eligibility: an account settling in a third, unrelated currency is eligible too
- eligibility: a dollar-priced product on a non-dollar-holding account is still withheld — this pins the settlement check that survives on the quoted path, which was otherwise left untested by the inversions above

Two more cover the Pix option scoping, both in `spec/services/order/prepare_payment_intent_service_spec.rb`:

- a Pix charge on a Brazilian connected account gets `payment_method_options` with only `expires_after_seconds`, and `pix_iof_fee_per_thousand` is 0 for it
- a Pix charge on a Gumroad-held account gets `amount_includes_iof` and is billed the IOF fee — this is what all Pix traffic looks like today

Both were proven load-bearing the same way: sending the option unconditionally again reddens exactly those examples (`7 examples, 2 failures` on the Pix context) and restoring the guard returns `7 examples, 0 failures`.

spec results

## Risk

The blast radius is limited to which payment methods are listed on a deferred PaymentIntent. Two things bound it:

- The per-account capability intersection is untouched, so a method is still only listed on an account that has activated it. That is the check preventing the failure mode where an unaccepted method in `payment_method_types` rejects the whole intent create and takes card down with it (gumroad-private#1026).
- Every method still sits behind its own launch flag, so this can be rolled back per method without a deploy.
- The Pix option scoping above closes the one place where widening the gate could have reached Stripe with a parameter that does not apply, which is the same failure mode: an unaccepted option rejects the intent create.

Balances stay correct. For a Stripe Connect seller the money never passes through a Gumroad-held balance at all — the charge lands directly in the seller's own Stripe account, and `Purchase#increment_sellers_balance!` and `#process_refund_or_chargeback_for_purchase_balance` both return early unless the charge used a Gumroad-held account. No seller balance row is written in any currency for these charges, so there is nothing a non-dollar settlement currency could corrupt, and payouts for such sellers are made by Stripe itself rather than by us. (An earlier draft of this section credited the balance layer's `merchant_account_gross_amount` branch; a pre-merge review caught that Connect direct charges never reach it.)

---

Written by Claude Opus 5 (`claude-opus-5`) running in Hermes Agent. The instruction was to widen the US-dollar settlement gate on local checkout methods to reach as much of checkout as possible, from antiwork/gumroad-private#1442. The model read the forced-currency lane end to end, measured the blocked volume against the production read replica, removed the two gates and the helpers they left dead, traced what the removed guarantee had been protecting downstream and scoped Pix's IOF option to the charges it applies to, updated the specs that asserted the old behaviour, added coverage for the widened one, and verified each new spec fails when the old behaviour is restored.


gumclaw:

gumclaw:
**On the red CI: none of it is from this PR.** Every failing shard is the same six `stats_spec` examples that are red on `main` right now (`1214 examples, 6 failures`, all `affiliate_credit_sum_from_scope`). This diff touches only `buyer_currency_eligibility.rb` and `payment_method_resolver.rb`. Fix for main is #6456 (spec-only, green, now with @slavingia); tracked as gumroad-private#1447. This will go green on a rebase once that lands — no action needed here.

**Local runs on this branch's head `919cfbdb2`:**

```
bundle exec rspec spec/services/checkout/buyer_currency_eligibility_spec.rb \
                  spec/services/checkout/payment_method_resolver_spec.rb
157 examples, 0 failures

bundle exec rubocop <the 4 touched files>
4 files inspected, no offenses detected
```

**Authorization, since this reverses a documented gate:** gumroad-private#1442 was filed as a *decision* issue — whether to widen the forced-currency gate so local methods render on non-USD-settling merchant accounts. @slavingia answered it directly at 23:23Z: *"Yes widen to 100% of checkouts as much as you can. That's the goal ASAP."* So this is a sanctioned build, not me pre-empting an open question.

**What it's worth:** the gate caps Bancontact at 64 of 747 Belgian EUR purchases in the last 7 days — roughly $2.6K per 30 days reachable against the ~$26K the Belgium share implies. Same shape applies to iDEAL, Klarna and Pix on non-USD-settling accounts.

**The part that most needs your eye,** and the reason I'd rather you than me judge it: settlement and reconciliation. A Bancontact charge landing on a EUR-settling account has to reconcile against the same canonical-versus-presentment arithmetic that gumroad-private#1328 is auditing, and two of that issue's defects (A3/A4, the transfer-reversal amounts) are *known broken today* on exactly these non-USD-holding accounts — 384 of 391. I attempted that fix earlier and reverted it: it regressed seller balances by 5 cents in three specs because it is entangled with the affiliate-credit ledger. So widening the gate increases traffic on a path with a known, unfixed amount bug. That may still be the right trade given the directive, but it is a judgment call about money and it should be made deliberately rather than inherited from a merge.

gumclaw:
Ran the local adversarial pre-merge review on this branch (Claude Opus 5, static analysis over the full diff plus the money paths downstream of it). Posting the result here since this is a gate-widening change to a money path.

**The primary hunt came back clean.** The newly admitted class is a euro/rupee/real PaymentIntent created directly on a Stripe Connect account whose balance is not in US dollars. Each money leg was traced separately:

- **Charge leg** — the intent is created in the presentment currency with `application_fee_amount` also in that currency. This leg was already live for dollar-holding Connect accounts taking euro intents; the account's balance currency never enters it. In the quoted case the FX quote is minted against the same connected account the charge is created on, so quote-account equals request-account.
- **Balance leg** — no seller balance row is written at all for a Connect direct charge. `Purchase#increment_sellers_balance!` and `#process_refund_or_chargeback_for_purchase_balance` both return early unless `charged_using_gumroad_merchant_account?`, which is false here.
- **Refund leg** — every refund path sends Stripe the presentment amount rather than dollar cents, and fails closed when it cannot compute one. The two reversal paths that do hardcode dollars, `reverse_excess_amount_from_stripe_transfer` and `reverse_the_transfer_made_for_dispute_win!`, are unreachable for this class: both require the funds to be held by Stripe on a Gumroad-managed account, and a Connect account holds funds for the creator. `Credit.create_for_vat_exclusive_refund!` is reachable but books canonical dollar amounts against the Gumroad platform account, so the connected account's currency does not enter it.
- **Payout leg** — `is_balance_payable` returns false for creator-held balances, and since no balance row exists for these charges the "seller's money becomes unpayable" scenario cannot arise from this lane.
- **Disputes** — direct-charge dispute webhooks build flow of funds from the dispute's own currency; the dollar-specific helpers only run inside the destination-charge branch.
- **Renewals** — both eligibility entry points fall back on `off_session` and `setup_future_charges`, so a saved iDEAL or UPI method can never produce a non-dollar renewal intent.

**Three things the review found worth fixing, all now pushed in 733187d:**

1. The stored-currency half of the settlement check that survives on the quoted path had no spec left. The inverted specs took its only coverage with them, so deleting `usd_holding_merchant_account?` from `usd_settling_merchant_account?` would have passed the entire suite. Added a spec that pins it; verified load-bearing by removing the check and watching only that spec go red.
2. The justification comment on the removed gate credited the balance layer's `merchant_account_gross_amount` branch for keeping balances correct. That branch never runs for a Connect direct charge — the real reason is that no seller ledger row is written at all. Reworded, in the code and in the PR body.
3. A comment still named `usd_holding_settlement_account?`, which this same branch deletes.

The before/after evidence the PR was missing is also up now: the same euro-priced cart from a euro-settling Belgian Connect seller resolves to `["card"]` on `main` and `["card", "ideal", "bancontact"]` here.

gumclaw:
The `Test Fast` failure here is not from this PR — it is inherited from `main` and clears on its own once #6460 lands.

The six failures are all in `spec/modules/user/stats_spec.rb`:

```
NoMethodError:
  undefined method 'affiliate_credits_sum_total' for an instance of User
```

#6420 renamed that method to `affiliate_credits_total_revenue_cents`; #6445 merged 34 minutes later with six new examples written against the old name. Neither PR conflicted textually, so `main` merged both cleanly and is red at `6201d5ad6` ([run 30317384472](https://github.com/antiwork/gumroad/actions/runs/30317384472), Test Fast 11). Every branch based past `716f5830c` — including this one — picks the failure up on whichever fast shard Knapsack hands the file.

#6460 fixes it on `main` by pointing the examples at `affiliate_credit_sum_from_scope`, the method the `describe` block is actually named after. I will merge `main` into this branch once it is in.

Nothing in this PR's own diff is implicated.

gumclaw:
Pushed one more commit here after Sahil said to build this ASAP on gumroad-private#1442.

Removing the settlement gate takes away a guarantee something downstream was quietly relying on. `Order::PreparePaymentIntentService#pix_payment_method_options` sent `amount_includes_iof: "always"` on every Pix intent, and that was only safe because the gate meant every Pix intent was created on a Gumroad-held account, where the payment is cross-border and Brazil's IOF genuinely applies. With the gate widened, a Brazilian connected account can reach Pix for the first time, and that payment is domestic: no foreign exchange, no IOF. Asking Stripe to price a tax that does not apply is at best meaningless, and an option Stripe does not accept fails the whole intent create, which takes card down with it for that checkout.

The option is now scoped to charges on a Gumroad-held account, matching the condition `Purchase#pix_iof_fee_per_thousand` already uses, so what Stripe is told and what the seller is billed can never disagree. `expires_after_seconds` is unchanged — that is about how long we hold the purchase open, not about who settles the money. Nothing changes for current traffic, where every Pix intent is created on the platform account.

Two specs cover it, one per side, and both were proven load-bearing: restoring the unconditional option reddens exactly those two (`7 examples, 2 failures` on the Pix context) and putting the guard back returns `7 examples, 0 failures`. Rubocop clean on both files.

One note on local runs: `spec/services/checkout/buyer_currency_eligibility_spec.rb` is flaky in this environment independent of the diff — three runs of the same seed on unmodified HEAD gave 39, 5, and 0 failures. CI is the authority there.

gumclaw:
**The four red `Test Slow` shards are not from this PR — they are a pre-existing flake in `taxes_spec.rb` that I reproduced on clean `main`.**

Failures were `Test Slow 8/30/44/45`, all four the same shape:

```
Failure/Error: expect(page).to have_alert(text: "Your purchase was successful! ...", visible: :all, wait: 60)
  expected to find alert with text "Your purchase was successful! We sent a receipt to test@gumroad.com"
  but there were no matches. Also found "", which matched the selector but not all filters.
# ./spec/support/checkout_helpers.rb:169:in 'block in CheckoutHelpers#check_out'
```

taxes_spec.rb:1272 (India), :2039 (Georgia), :2980 (Serbia), :3661 (Vietnam).

### Reproduced on `main` with none of this PR's code

Ran the same four examples in a worktree checked out at the merge base, `16e969cf` (`Point the affiliate credit sum specs at the method they test (#6460)`) — no commit from this branch present:

| example | clean `main` @ 16e969cf | this branch |
|---|---|---|
| `:1272` India | pass | **fail** (run A), pass (run B) |
| `:2039` Georgia | pass | pass |
| `:2980` Serbia | **fail** | **fail** (run B), pass (run A) |
| `:3661` Vietnam | pass | pass |

`:2980` fails on `main` with the byte-identical failure message and the identical `checkout_helpers.rb:169` backtrace. And the example that fails changes between runs of the *same* tree — that is the signature of a flake, not of a regression.

### What the flake actually is

The failure screenshot from the India run shows the checkout still sitting on the form: **Full name is empty and the card field contains `40`** — the first two characters of `4000000360000006`. Capybara's fill of the Stripe Payment Element got interrupted partway through by an element remount, so `Pay` was clicked against an incomplete form and no success alert ever arrived. Nothing to do with which payment methods are offered.

### Why this diff cannot reach these specs anyway

All four are **US dollar** carts (`expect(page).to have_text("$100")`, paid with a test Visa). The code this PR changes is `method_forced_decision` / `forced_currency_methods`, and both return before the removed gate on two separate counts for these specs:

1. `forced_currency_methods` returns `[]` unless `seller_enabled?(seller)` — which requires the `buyer_currency_charging` **and** `buyer_local_currency` flags on the seller. Verified in the test environment: both are globally off and `seller_enabled?` is `false` for a factory seller. These specs never activate either flag.
2. Even with the flags on, the method must have `forced_currency_for(method) == cart_product_currency`. The forced currencies are EUR (iDEAL, Bancontact), INR (UPI), BRL (Pix). A USD cart matches none, so `methods_for_cart_currency` is empty and the method returns before it ever reached the gate that was deleted.

The removed gate sat strictly *after* both of those, so no USD checkout's behaviour changed by a single line.

### Where CI stands

That run (30320429644) was superseded and cancelled when I pushed 3a09276 (`fix: only send Pix's IOF option on cross-border charges`). A full run is in flight on the new head; I'll report the result rather than assume it.

The `taxes_spec` flake is real and worth its own fix — the checkout helper needs to wait for the Payment Element to settle before typing — but it is a `main` problem, not this PR's, and I'll file it separately rather than widen this diff.

gumclaw:
Ran the local adversarial pre-merge review again on the follow-up commit, since a gate-widening diff is the class where the bug is never at the gate — it is wherever something downstream assumed the newly-admitted entity could not reach it. Claude Opus 5, static analysis over the commit plus the surrounding payment-intent, merchant-account and fee code. Verdict SHIP-WITH-FIXES, two P3s, both now pushed.

The main question was whether `pix_iof_applies?` asks about the right account. It does, and this is the part worth recording: `Order::PreparePaymentIntentService#merchant_account` memoizes the account resolved by `resolve_merchant_account_and_fees`, which runs first in `prepare_unconfirmed_charge`, and that same object is handed to `StripeDeferredPaymentIntent.create`. There, `StripeIntentChargeRouting.request_options` sets `stripe_account` — i.e. creates the intent on the connected account — on exactly `is_a_stripe_connect_account?`, the predicate the new guard negates. Same predicate, same object, same request, so the guard and the routing cannot disagree about whose account the charge lands on.

The reviewer also checked the guard against `Purchase#charged_using_gumroad_merchant_account?` input by input. They agree everywhere reachable: `is_managed_by_gumroad?` is `!user_id` while `is_a_stripe_connect_account?` requires `user_id`, so those are mutually exclusive; a nil account reads as ours on both sides; and a migrated seller on a Gumroad-controlled custom account is a destination charge on the platform, where the money does leave Brazil and the tax does apply. The one divergence it could construct — a purchase whose charge processor was already set to something other than Stripe — cannot occur on a lane that requires a Stripe confirmation token.

The two fixes:

- The comment above `Purchase::PIX_IOF_FEE_PER_THOUSAND` still asserted the invariant this PR deletes, that every Pix payment to Gumroad is cross-border so the tax always applies. Left alone, the next reader re-derives that and puts the unconditional option back. Rewritten to state the condition.
- The new Gumroad-held-account example asserted the options hash but not that Pix was listed on the intent, so a regression in method resolution on the platform lane could have kept it green. Added the assertion.

Also worth stating plainly, because it is a limit of the evidence rather than a finding: both Pix examples stub the intent create, so "Stripe accepts a domestic-account Pix intent carrying the expiry but not the IOF option" is asserted against a stub. Confirming that is a live call against a Brazil-domiciled connected account, and it only becomes reachable if one of those sellers activates the Pix capability. There are none today.

Auto-merge stays off until CI is green on this head.

gumclaw:
Posting the fable-5 verdict for the last two commits here for the record — the review was still running when this merged, so its findings landed after the fact. **The base diff is fine; one predicate added in 3a09276 is not.** Fix is up as #6467.

**Finding (P2, fixed in #6467) — `pix_iof_applies?` keys on account *ownership*; the tax keys on account *domicile*.**

`app/services/order/prepare_payment_intent_service.rb` gated `amount_includes_iof` on `!merchant_account&.is_a_stripe_connect_account?`, mirroring the seller-fee gate on the grounds that the option and the fee must always agree. IOF is a tax on foreign exchange, so it applies whenever the payment leaves Brazil — which is not the same question.

Nothing restricts Pix to *Brazilian* connected accounts: `BR_LOCKED_PAYMENT_METHOD_TYPES` gates on the **buyer's** country, and the only per-account condition is the Stripe capability snapshot sellers manage themselves. (Klarna has an explicit account-country gate on this same lane, for exactly this reason; Pix had none.) So a non-Brazilian connected account with `pix_payments` active takes a cross-border payment where IOF *does* apply — and the gate withheld the option. Stripe's default is `never`, which marks the buyer's amount up 3.5% over the price checkout quoted. Silent, and on the buyer's side.

Only reachable because this PR removed the settlement gate — before it, no connected account could reach Pix at all. It is the same shape as the Pix/IOF catch itself: a deleted gate had been silently guaranteeing a property that downstream code was written against.

Three P3s went with it, all comments asserting the invariant this PR retired: `PIX_AMOUNT_INCLUDES_IOF`'s own comment still said "always sent explicitly", `Purchase::PIX_IOF_FEE_PER_THOUSAND` said the fee gates "on the same condition" as the option, and `purchase_process_spec.rb:985` still said "Every Pix payment to Gumroad is cross-border".

One more thing worth flagging: the direct-charge spec's fixture left `country` **nil** while its comment described "the seller's own Brazilian Stripe account" — so the Brazilian lane was never actually exercised. #6467 sets `country: "BR"`, which also corrects the fee expectation (`calculate_fees` returns 0 outright for a Brazilian connect account, so the old flat-fee assertion described an account the test didn't have), and adds the missing non-Brazilian lane.

**What the review confirmed safe**, so it isn't lost: the intent-create routing, the fee gate, and the option gate all read one merchant-account record in one request (`StripeIntentChargeRouting.request_options` uses the identical predicate on the identical object); there is no temporal divergence between the option and the fee, since `calculate_fees` runs inside `prepare_merchant_account` *before* `create_unconfirmed_intent` and nothing recomputes afterwards; `pix_payment_method_options` is the only builder of `payment_method_options` anywhere, so Pix's IOF option was the only per-method create-time parameter that could have carried a stale assumption (iDEAL/Bancontact/UPI return nil); and nothing else in the repo — refunds, disputes, receipts, presenters — reads `amount_includes_iof`.

Separately, the four red `Test Slow` shards earlier were the `taxes_spec` checkout flake, reproduced on clean `main` and filed as gumroad-private#1451. Not from this PR.

PR #6462, opened 2026-07-28
Reverse refund-fee retention transfers in the transfer's own currency

gumclaw:
## What

`StripeChargeProcessor.debit_stripe_account_for_refund_fee` now converts the owed fee (canonical USD cents) into each candidate transfer's own currency before comparing it against the transfer's remaining reversible amount, and before creating the reversal — at both reversal sites (the internal payout-transfer path and the >120-day-old sale-transfer path).

Addresses item **A6** of the gumroad-private#1328 cross-currency audit (fee-retention transfer reversal).

## Why

**Decision: fix, not document.** The audit item allowed either; the fix is unambiguous because `debit_stripe_account_for_australia_backtaxes` in the same file already solves the identical problem the same way — it converts the owed USD amount into the account currency via `usd_cents_to_currency` before comparing and reversing.

The bug: `credit.amount_cents` is USD cents, but the method early-returns for US accounts, so **every** transfer reaching the comparison belongs to a non-US Gumroad-managed account whose transfers are frequently in local currency. `tr.amount - tr.amount_reversed > amount_cents` compared local-currency cents against USD cents, and `create_reversal(..., amount: amount_cents)` then debited the raw USD figure denominated in the transfer's currency.

Money consequence of the old behaviour: for weaker-than-USD currencies (INR, MXN, JPY…) Gumroad under-collected the retained fee (reversing e.g. 1000 paise instead of ~83,000); for stronger-than-USD currencies (GBP, KWD…) the **creator was over-debited**. The fix reverses exactly the converted amount; USD-denominated transfers are unchanged (rate 1.0 pass-through in `usd_cents_to_currency`).

The eligibility comparison also now uses the converted amount, so a transfer that can cover the USD figure but not the actual local-currency amount is correctly skipped instead of producing a partial/failed reversal.

## Test results

- New specs (4) in the `currency conversion of the reversal amount` describe are load-bearing: with the app change stashed, **3 of 4 fail** (`create_reversal` receives the raw USD amount / the too-small transfer is wrongly selected); the 4th (USD-denominated transfer unchanged) passes on both, as intended.
- `rspec spec/business/payments/charging/implementations/stripe/stripe_charge_processor_spec.rb`: **233 examples, 0 failures** (run at two random seeds, 42 and 7).
- Targeted `debit_stripe_account_for_refund_fee` describe after rebase onto latest main: **6 examples, 0 failures**.
- `rubocop` on both touched files: **2 files inspected, no offenses detected**.


gumclaw:
## Fable-5 adversarial pre-merge review — PASS

Reviewed at `71960aa19a4267e5eca76e1fdbd9fad6d8ead131`, which contains the latest `origin/main` and is mergeable. No blocking defects.

Recording the fix-or-document decision, since this was the one item on the audit I'd flagged as a possible judgment call. Resolved as **fix**, on evidence rather than preference:

- `credit.amount_cents` is canonical USD — computed from `processor_fee_cents` behind an explicit `processor_fee_cents_currency == "usd"` check
- the method early-returns for US accounts, so every transfer reaching the comparison belongs to a **non-US** Gumroad-managed account, where sale transfers are frequently in local currency
- `debit_stripe_account_for_australia_backtaxes` in the same file already solves the identical problem via a USD-cents conversion

That last point is what settled it: there was an in-repo precedent for the correct shape, so this was a real inconsistency to fix rather than a trade-off to escalate.

Not arming yet — CI is red from the shared Stripe test-account 429 flake (my six concurrent runs blew the shared account's rate limit); failed shards re-run.

PR #6464, opened 2026-07-28
Reverse Stripe transfers in the transfer's own currency

gumclaw:
## What

`Purchase::Refundable#reverse_excess_amount_from_stripe_transfer` sent Gumroad's **canonical USD** amount as a Stripe transfer reversal amount. A Stripe transfer is denominated in its own currency, so for a buyer-currency (presentment) charge that number was interpreted as the buyer's currency and reversed the wrong amount of the seller's money.

This is Group A3 of gumroad-private#1328.

## Why

Measured on a live production purchase (`348298474`, a EUR-settling seller):

```
CHARGE    cur=eur amt=533
TRANSFER  cur=eur amt=366        <- what Stripe reverses against
BALANCE   issued=416/usd  holding=366/eur
```

The old code reversed `issued_amount_net_cents` = **416**, against a **EUR** transfer — about **50 cents too much taken from the seller**, every time this path runs. It reads as a currency bug but it lands as lost seller money, and it fails silently.

Reachable today, not theoretical: **105 of 200** recent presentment purchases settle on `holder_of_funds=stripe` accounts, and the sampled ones hold **EUR** with non-zero tax (the path's trigger is a fully-refunded VAT amount).

## Before / After

The refund's balance transaction already records the same money in both denominations, so no conversion is needed — just pick the leg that matches the transfer:

| | Transfer | Old reversal | New reversal |
|---|---|---|---|
| Presentment charge, EUR-settling seller | 366 EUR | **416** (USD figure sent as EUR) | **366** EUR |
| USD transfer to a CAD account | 880 USD | 29 USD | **29** USD (unchanged) |

The second row is the case that makes "just use the merchant account's currency" wrong: a Gumroad-managed **CAD** account receives **USD** transfers for a USD charge. The currency has to be read off the transfer itself, which is what this does.

Two smaller points:
- **Fails closed.** If neither ledger leg matches the transfer's currency, it reverses nothing and logs, rather than sending a guess at the seller's expense.
- **Ledger stays canonical.** The credit's USD leg keeps the USD figure instead of converting the reversal amount back, so no second rounding error enters the books. On a partial reversal the canonical leg takes the same proportion.

## Test Results

`spec/models/purchase/purchase_refunds_spec.rb`, same run, fix stashed then applied:

```
BASELINE (main):   147 examples, 5 failures
WITH THIS FIX:     147 examples, 2 failures
```

Fixes 3, breaks 0. The 2 remaining (`:113`, `:1247`) **fail identically on unmodified main** — pre-existing and unrelated. Stable across seeds (`--seed 12345` gives the same 2).

Two new regression specs, verified RED-GREEN — reverting just the currency-selection lines makes both fail with `expected: ("tr_cad_presentment", {amount: 366})`:
1. reverses the holding-currency amount when the transfer is in the account's currency
2. moves no money when neither leg matches the transfer currency

`rubocop`: 2 files, no offenses.

## Notes for review

**A4 (`reverse_the_transfer_made_for_dispute_win!`) is not included, because it is not broken.** gumroad-private#1328 lists it as the same shape as A3, but that transfer is created by us explicitly with `currency: Currency::USD` (`StripeTransferInternallyToCreator.transfer_funds_to_account`), and `amount_refundable_cents` is canonical USD — so it is USD against USD. The issue's verdict there was reasoned from the code rather than measured; this is the measurement. Worth a second opinion since it is the neighbouring function.

**This is seller money, so please check the proportional-canonical branch** (partial reversal) specifically — it is the one part not exercised by a real production row.


gumclaw:
## Fable-5 adversarial pre-merge review — APPROVE

Reviewed at `9cda52e2e0507e49bc5bc468432010ee9353b8fd`. The reviewer checked the diff against `Currency::USD` semantics, `Credit.create_for_p...`, and the balance-transaction leg accessors, and found no blocking defects.

Worth stating why this approach is the right one: a transfer reversal must be denominated in the **transfer's own** currency, which is not always the merchant account's currency — a Gumroad-managed CAD account receives USD transfers for a USD charge, while a buyer-currency presentment charge settling in EUR produces a EUR transfer. Reading the currency off the transfer itself and matching it against the balance transaction's legs keeps legitimate non-USD refunds working, and raises only when neither leg is denominated in the transfer's currency — where there genuinely is no correct number to send.

This is the survivor of a duplicate pair: I closed #6465 (same two methods, same file) because it only failed closed on every non-USD transfer, which would have broken real non-USD refunds rather than reversing the right amount. Verified #6464 covers everything #6465 did.

Assigned to @rmarescu and @nyomanjyotisa. Not arming yet — CI is currently red from the shared Stripe test-account 429 flake, and I've re-run the failed shards.

gumclaw:
### Merged `main` to clear a red shard that wasn't from this PR

`Test Fast 4` was failing with:

```
NoMethodError: undefined method 'affiliate_credits_sum_total' for #<User ...>
```

Nothing to do with transfer reversals — this PR touches only `app/modules/purchase/refundable.rb` and its spec. The cause was a semantic collision on `main`: #6420 renamed `affiliate_credits_sum_total`, and #6445 landed six examples written against the old name. They never conflicted textually, so git merged both cleanly and left specs calling a method defined nowhere.

That was already fixed on `main` by #6460; this branch just predated it. Merged `origin/main` in to pick it up — no code changes.

**Verified locally after the merge:**
- the previously-failing examples: `6 examples, 0 failures`
- this PR's own spec file: 2 failures at `:113` and `:1247`, and **both fail identically on unmodified `origin/main`** (`147 examples, 2 failures`) — pre-existing and unrelated, confirming the review's read

Fable verdict (APPROVE) still stands; the review also mutation-tested the currency-selection branch and confirmed both new regression specs fail without the fix, so they're load-bearing rather than merely green.

PR #6465, opened 2026-07-28
Fail closed on non-USD transfers before creating currency-mismatched reversals (1328 A3/A4)

gumclaw:
## What

Makes the two `Stripe::Transfer.create_reversal` call sites in `Purchase::Refundable` currency-safe — items **A3** and **A4** of the cross-currency audit (gumroad-private#1328):

- **A3 — `reverse_excess_amount_from_stripe_transfer(refund:)`**: computed the excess to reverse from the seller's canonical-USD balance debit, compared it against `transfer.reversals[..].amount` (denominated in the **transfer's** currency), and sent the difference as a reversal amount — again in the transfer's currency. The `_usd` variable suffixes assert a currency the Stripe-side values don't necessarily have.
- **A4 — `reverse_the_transfer_made_for_dispute_win!`**: sent `amount_refundable_cents` (canonical USD) directly as a reversal amount, interpreted by Stripe in the transfer's own currency.

Both methods now check `transfer.currency` and **raise** for any non-USD transfer, before any money moves. USD transfers — every transfer these methods reach today — keep byte-identical behaviour.

## Why

These calls move real money out of creators' connected accounts. Stripe denominates reversal amounts in the transfer's currency, so a canonical-USD figure sent against a non-USD transfer is silently reinterpreted: reversing `367` against a euro-denominated transfer pulls **€3.67 instead of $3.67** — a real financial loss (over- or under-reversal depending on the rate), with no error anywhere. The A3 comparison additionally misjudges *whether* an excess exists at all when the two sides are in different currencies.

**Why raise rather than read a presentment snapshot or convert:** the amounts these methods need are Gumroad-internal — a seller balance debit net of fees, and `amount_refundable_cents` — and no stored presentment figure corresponds to them (refund presentment snapshots record gross buyer-currency components only). The dispute-win transfer is created in USD by `StripeTransferInternallyToCreator` regardless of the charge's presentment currency, so today the guard documents and enforces an invariant rather than changing behaviour. Converting at the live rate at reversal time would book an amount matching neither our balances nor the original transfer. Failing closed for manual reconciliation mirrors the convention the presentment refund paths already follow (`buyer_presentment_refund_blocked?`, #6453).

Part of gumroad-private#1328 (items A3, A4).

## Test results

`bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb` (the spec file covering `Purchase::Refundable`; there is no `spec/modules/purchase/refundable_spec.rb`):

- **With the change:** 151 examples, 3 failures — all 3 pre-existing/flaky (see below).
- **Baseline (change reverted, same seed 12345):** 147 examples, 5 failures — overlapping pre-existing set.
- The failures at `:113` ("emails the buyer the presentment purchase total on a full refund") and `:1247` ("dispute after a refund event…") fail **identically on unmodified origin/main** when run alone; the `merchant_migration` and `:1451` VCR failures are order-dependent (pass alone with the change, fail intermittently on both sides). None are introduced by this PR.
- Targeted run of both affected describes: **9 examples, 0 failures**.
- `rubocop` on both touched files: **2 files inspected, no offenses detected**.

**Load-bearing verification** (each new spec fails without the fix, passes with it):
- The 4 new examples with the fix in place: **4 examples, 0 failures**.
- Same 4 examples with `app/modules/purchase/refundable.rb` reverted: **4 examples, 2 failures** — exactly the two fail-closed specs (A3 and A4 non-USD cases). The two USD-path specs pass on both sides *by design*: they pin that USD behaviour is byte-identical to the old code (amount, `Credit.create_for_partial_refund_transfer_reversal!` arguments), guarding against a regression in the fix itself.

## Notes for review

- I deliberately did **not** use the presentment snapshot here, unlike #6453: flagging explicitly that the correct buyer-currency figure for the A3 excess (seller-net of fees) is not stored anywhere, so there is no snapshot to read — raise-for-manual-reconciliation was the only non-guessing option I found. If a stored seller-net presentment figure exists that I missed, this should use it instead.

## QA steps

Backend-only guard on money-moving private methods; no UI surface. The reviewable artifacts are the diff and the load-bearing specs above — run `bundle exec rspec spec/models/purchase/purchase_refunds_spec.rb -e "currency handling of the reversal amount"` (4 examples) with and without the `app/modules/purchase/refundable.rb` hunk to reproduce the RED/GREEN evidence.

---

🤖 Written by Hermes Agent (claude-fable-5) working gumroad-private#1328 items A3/A4. Prompt: implement A3 and A4 from the cross-currency audit following the presentment conventions established in #6453, preferring fail-closed raises over guessed amounts on money-moving paths.

### DIFF
--- a/app/modules/purchase/refundable.rb
+++ b/app/modules/purchase/refundable.rb
@@ -613,7 +613,25 @@ class Purchase
 
       transfers = Stripe::Transfer.list({ destination: merchant_account.charge_processor_merchant_id, created: { gte: dispute.won_at.to_i - 60000 },  limit: 100 })
       transfer = transfers.select { |tr| tr["description"]&.include?("Dispute #{dispute.charge_processor_dispute_id} won") }.first
-      Stripe::Transfer.create_reversal(transfer.id, { amount: amount_refundable_cents }) if transfer.present?
+      return unless transfer.present?
+
+      # Stripe denominates reversal amounts in the transfer's own currency, while
+      # amount_refundable_cents is canonical USD. The dispute-win transfer is created in
+      # USD (StripeTransferInternallyToCreator via the dispute-reinstated webhook handler
+      # always passes Currency::USD), so sending the canonical figure is correct — but
+      # only as long as that assumption holds. If the transfer we found is in any other
+      # currency (a buyer-currency presentment charge would make this mistake easy to
+      # introduce upstream), the canonical USD number would be reinterpreted in that
+      # currency and reverse the wrong amount of real money from the creator's account.
+      # Fail closed instead of guessing (gumroad-private#1328 A4).
+      if transfer.currency != Currency::USD
+        raise "Cannot reverse dispute-win transfer #{transfer.id} for purchase #{id}: " \
+              "the transfer is denominated in #{transfer.currency}, but the reversal amount " \
+              "(amount_refundable_cents) is canonical USD. Reverse it manually with a " \
+              "#{transfer.currency}-denominated amount."
+      end
+
+      Stripe::Transfer.create_reversal(transfer.id, { amount: amount_refundable_cents })
     end
 
     def debit_processor_fee_from_merchant_account!(refund)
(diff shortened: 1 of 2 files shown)


gumclaw:
Closing as a duplicate of #6464, which covers the same two methods in the same file with a strictly better fix.

Both PRs came out of the same audit item (gumroad-private#1328 A3/A4) and were opened two minutes apart — my own dispatch error, not two independent findings. Verified the overlap before closing: identical file set (`app/modules/purchase/refundable.rb` + `spec/models/purchase/purchase_refunds_spec.rb`) and the same two methods, `reverse_excess_amount_from_stripe_transfer` and `reverse_the_transfer_made_for_dispute_win!`.

**Why #6464 survives.** This PR only fails closed: it raises on any non-USD transfer. That prevents the wrong amount of real money being reversed, but it also means a legitimate non-USD refund stops working — it converts a silent money bug into a hard failure on a real seller path.

#6464 goes further and actually reverses the correct amount, reading the transfer's own currency and matching it against the balance transaction's legs (`holding_amount_currency` / `issued_amount_currency`), falling back to a raise only when *neither* leg is denominated in the transfer's currency — where there genuinely is no correct number to send. That keeps non-USD refunds working and still refuses the unrepresentable case.

Verified against every part of this PR: the USD-unchanged behaviour and the non-USD guard are both covered there, so nothing from here is lost.

PR #6466, opened 2026-07-28
Update the cover-images article for per-cover frame shapes

gumclaw:
## What

Rewrites the multiple-covers note in `_60-adding-a-cover-image` (help.gumroad.com/help/article/60-adding-a-cover-image). The article told sellers we match every cover's height to the first cover's height. That stopped being true when #6451 merged: the product page now shapes the cover frame to the cover being viewed, caps a tall cover at 80% of the window height, and leaves covers with no recorded dimensions in the old frame.

The replacement copy says each cover is shown at its own shape, names the phone-filmed portrait video case the seller who reported this was hitting, states the height cap, and keeps the practical advice to use covers of the same shape — now for the honest reason (the page reflows between covers of different shapes) rather than a behavior we no longer have.

Body-only edit. `articles.yml` is untouched: the title, description snippet, category, and every anchor ID are unchanged.

## Why

Triggered by antiwork/gumroad#6451, merged 2026-07-28. Tracking issue for the product change: antiwork/gumroad-private#1437.

Every claim was checked against the shipped code rather than the PR description:

- Frame follows the active cover, not `covers[0]` — `app/javascript/components/Product/Covers/index.tsx`, `frameStyle` reads `activeCover.native_width` / `native_height`.
- 80% of window height — `MAX_PORTRAIT_FRAME_HEIGHT = "80svh"` in `app/javascript/utils/videoFrame.ts`. `svh` is the small viewport height, so "80% of the window height" is accurate for a seller reading it.
- Covers with no recorded dimensions keep the old frame — `frameStyle` is `undefined` when either dimension is missing, which is the pre-existing CSS box.
- Centered rather than cropped — the frame keeps the column's full width and the cover letterboxes horizontally inside it.

The reflow note is not a hedge: it is the one residual behavior #6451's own review called out, so sellers should hear it from us rather than discover it.

## Before / After

Rendered through `ApplicationController.render` on the real partial, before panel rendered from the `origin/main` copy of the same file. Left is before, right is after.

before and after

## Test Results

- Partial renders in the real view context and asserts both directions: new copy present, removed copy absent.
- Tag-balance check on the edited file: `p` 19/19, `ul` 4/4, `li` 17/17, `h3` 10/10, `figure` 7/7, `b` 3/3, `em` 1/1, `a` 15/15.
- `articles.yml` untouched, so the slug-to-partial integrity the help-center specs guard is unchanged.

## QA steps

1. Open `help.gumroad.com/help/article/60-adding-a-cover-image` on the deployed branch.
2. Under "Recommended specifications", the note about matching the first cover's height is gone; the two paragraphs about each cover keeping its own shape are in its place.
3. The "In this article" links still jump to their sections — no anchor changed.

---

## AI disclosure

Written by `claude-opus-5` (read from `model.default` in the agent's config) driving the Hermes agent. Instructions given: the merged-PR help-center sync job — check whether antiwork/gumroad#6451 makes any help-center article wrong, and if so fix it. The model mapped the merged PR to this article, verified each behavioral claim against the shipped `Covers/index.tsx` and `videoFrame.ts` rather than the PR body, wrote the replacement copy, and rendered before/after from the real partial.


gumclaw:
Self-review against the shipped code, re-reading CONTRIBUTING.md as a checklist rather than from memory.

Every factual claim in the new copy traced to source, not to #6451's description:

| Claim | Source |
|---|---|
| Each cover is shown in a frame matching its own shape | `Covers/index.tsx` — `frameStyle` reads `activeCover.native_width / native_height`, not `covers[0]` |
| Tall covers capped at 80% of the window height | `MAX_PORTRAIT_FRAME_HEIGHT = "80svh"` in `utils/videoFrame.ts`, applied as the frame's `maxHeight`. `svh` is the small viewport unit, so "80% of the window height" is what a seller actually sees, including on mobile Safari |
| Centered rather than cropped | the frame keeps `width: 100%` and the cover letterboxes inside it — `Image.tsx` uses `max-h-full w-full object-contain` when the frame is shaped |
| Covers with no recorded dimensions are unchanged | `frameStyle` is `undefined` when either dimension is missing, leaving the pre-existing CSS box |
| The page shifts between covers of different shapes | the one residual behavior #6451's own pre-merge review recorded; sellers should hear it from us |

Three things I deliberately did not do:

- Did not touch `articles.yml`. The title, description snippet, and category are all still accurate for a body-only copy change, and formatting that file reflows unrelated entries.
- Did not change any anchor ID or TOC entry. External links into this article keep working.
- Did not drop the "use covers of the same shape" advice. It is still the right advice, just for the honest reason now.

Swept the rest of the help center for the same stale claim: `grep -rniE "height of the first|equal height|same height|first cover"` across all article bodies returns only this file's line and an unrelated one in `_176-metadata-for-audio-files` about which cover gets encoded into an audio track's metadata, which #6451 does not affect.

No logic in the diff, so autoreview is skipped per the docs-copy exemption; the code-verification table above is what stands in for it.

gumclaw:
Pre-merge adversarial review, run locally against `64ec2f45`. Verdict: **clear to merge.** No P1 or P2.

What I attacked, and what held:

**"Each cover is shown in a frame that matches its own shape" — is that true for every cover type, or only video?** True for all three. `Covers/index.tsx` shapes the frame from the active cover regardless of type, and `Image.tsx` (`max-h-full w-full object-contain`) and `Embed.tsx` (aspect-ratio box with `height: 100%`, `width: auto`) both fit inside it. Writing the sentence about videos only would have been the narrower, wronger claim.

**Does the claim overreach onto surfaces the change did not touch?** `Covers` renders on the public product page and in the seller's cover editor, both of which get the new behavior. The thumbnail variant short-circuits (`isThumbnail` gives no `frameStyle`), but the article discusses thumbnails in a separate section that this diff does not touch and that makes no shape claims. The buyer content page is a different component, already documented in `_43-streaming-videos`, whose copy stays correct.

**Is "80% of the window" honest for `80svh`?** Yes, and deliberately so. `svh` is the *small* viewport height, so on mobile Safari with the URL bar showing, 80svh is 80% of what the seller can actually see — the number in the doc matches the lived experience rather than an idealized viewport.

**Did the first draft describe the wrong object?** Yes, and it is fixed in this branch's second commit. The draft said "a cover taller than the frame is capped at 80% of the window height", but the cap is on the frame; the cover scales to fit inside it. A seller reading the old wording could reasonably conclude their 1080x1920 file is being truncated. Reworded to "the frame never grows taller than 80% of the window, so a very tall cover is scaled down and centered".

**Is the reflow note over-warning sellers about something invisible?** No. It is the residual behavior #6451's own review recorded, it is visible in that PR's walkthrough video, and it is the honest reason for advice the article was already giving. Omitting it would leave the "use the same shape" recommendation looking like leftover text from the behavior we just removed.

**Anything else in the help center now stale?** `grep -rniE "height of the first|equal height|same height|first cover"` over all article bodies: only this file and an unrelated line in `_176-metadata-for-audio-files` about which cover is encoded into an audio track's metadata, which is server-side and untouched.

Residual, not blocking: the screenshots in this article still show the old editor chrome, unrelated to this change and not depended on by any sentence I edited.

PR #6467, opened 2026-07-28
Key Pix's IOF option on the account's country, not on who owns it

gumclaw:
## What

`amount_includes_iof` is now sent on a Pix intent based on whether the payment **leaves Brazil**, rather than on whether Gumroad owns the account it is created on.

Follow-up to #6457, which merged while the pre-merge review of its last two commits was still running. Everything else in that PR stands; this corrects one predicate added late in it.

## Why

IOF is a Brazilian tax on foreign exchange, so it applies whenever a Pix payment crosses Brazil's border. #6457's `pix_iof_applies?` asked a different question — "is this account Gumroad-held?" — and deliberately mirrored the seller-fee gate, on the stated grounds that the option and the fee must always agree.

They must not, and the axis is wrong. Nothing restricts Pix to *Brazilian* connected accounts:

- `Checkout::PaymentMethodResolver::BR_LOCKED_PAYMENT_METHOD_TYPES` gates on the **buyer's** country (`payment_method_resolver.rb:320`), not the account's.
- The only per-account condition is the Stripe capability snapshot (`pix_payments == "active"`), which sellers manage themselves.
- Compare Klarna, which *does* have an explicit account-country gate on the same lane (`klarna_supported_for_account?`) precisely because Stripe's cross-border rule applies to the account the intent is created on. Pix had no equivalent.

So a non-Brazilian connected account with `pix_payments` active takes a payment that **does** cross the border. IOF applies — but the merged gate sees "connect account" and withholds the option. Stripe's default is `never`, which marks the buyer's amount up by the 3.5% tax. The buyer's banking app then shows more than the price checkout quoted them, silently, on the buyer's side of the ledger.

This class only became reachable *because* #6457 removed the settlement gate; before it, no connected account could reach Pix at all.

## The fix

Ask the border question: `!merchant_account&.is_a_brazilian_stripe_connect_account?` — the predicate `calculate_fees` already uses for the same account.

The fee gate (`Purchase#pix_iof_fee_per_thousand`) is deliberately left alone and deliberately stays different. The two answer different questions:

| | asks | direct charge, BR account | direct charge, non-BR account | Gumroad-held |
|---|---|---|---|---|
| `pix_iof_applies?` (option) | does the tax exist — is this cross-border? | no | **yes** | yes |
| `pix_iof_fee_per_thousand` (fee) | did Gumroad bear the cost and have something to recover? | no | no | yes |

The middle column is the one that was wrong. Both comments now state which question each gate answers, so the next reader doesn't "fix" the divergence back out.

## Before / after

| | before (merged) | after |
|---|---|---|
| BR connected account | option omitted ✅ | option omitted ✅ |
| non-BR connected account, Pix active | option omitted ❌ — buyer overcharged 3.5% | option sent ✅ |
| Gumroad-held account (all traffic today) | option sent ✅ | option sent ✅ |

**Today's production traffic is unaffected** — every Pix intent is currently created on the platform account, the right-hand column.

## Test results

`spec/services/order/prepare_payment_intent_service_spec.rb -e "Pix"` — **8 examples, 0 failures** (7 on `main` + 1 new).
`spec/models/purchase/purchase_process_spec.rb -e "Pix IOF fee"` — **3 examples, 0 failures**.
`rubocop` on all four files — no offenses.

All three lanes are now covered explicitly, where `main` had only two:

- Brazilian connect — no option, no fee. The fixture previously left `country` **nil**, so despite its comment saying "the seller's own Brazilian Stripe account" it was never actually Brazilian and never exercised this lane. Now set to `BR`, which also corrects the fee expectation: `calculate_fees` returns `0` outright for a Brazilian connect account, so the old flat-fee + fixed-fee assertion was describing an account the test didn't have.
- **non-Brazilian connect — option sent, no fee.** New; this is the bug.
- Gumroad-held — option sent, fee billed.

### The specs bite

Mutation-tested rather than assumed. Reverting the predicate to the merged ownership axis (`!merchant_account&.is_a_stripe_connect_account?`) turns the run red:

```
8 examples, 4 failures
rspec ./spec/services/order/prepare_payment_intent_service_spec.rb:1927 # ...still sends amount_includes_iof on a direct charge to a NON-Brazilian connected account, but bills no IOF fee
```

(plus three neighbouring Pix examples). Hardcoding `pix_iof_applies?` to `true` or to `false` each fails a different example, so neither branch is vacuous. These assertions use `eq` on the whole `payment_method_options` hash and `create_args` is captured by stubbing `StripeDeferredPaymentIntent.create`, so no VCR cassette can mask them — there are no Pix cassettes in this file.

## Also in this diff

Three comments that asserted the invariant #6457 retired, found in the same review:

- `PIX_AMOUNT_INCLUDES_IOF`'s own comment still ended "so this option is always sent explicitly" — 330 lines above the new conditional.
- `Purchase::PIX_IOF_FEE_PER_THOUSAND` described the fee as gating "on the same condition" as the option.
- `purchase_process_spec.rb:985` still said "Every Pix payment to Gumroad is cross-border".

## Not verified here

Whether Stripe currently grants cross-border Pix to non-BR connected accounts is a question for Stripe, not the codebase — but the fix is the fail-safe direction either way: sending `amount_includes_iof` on a cross-border payment is correct by definition, and the gate that decides it now matches the tax's actual rule instead of coinciding with it by accident.

### DIFF
--- a/app/models/purchase.rb
+++ b/app/models/purchase.rb
@@ -29,16 +29,23 @@ class Purchase < ApplicationRecord
   PROCESSOR_FEE_PER_THOUSAND = 29
   PROCESSOR_FIXED_FEE_CENTS = 30
   # IOF is a Brazilian consumer tax on any transaction that involves foreign exchange, currently
-  # 3.5% of the transaction value. It applies when the Pix payment leaves Brazil — that is, when
-  # the charge is created on a Gumroad-held account, since Gumroad is domiciled outside Brazil.
-  # For those charges Gumroad tells Stripe to bill the buyer exactly the price they agreed to
+  # 3.5% of the transaction value. It applies whenever the Pix payment leaves Brazil, which is
+  # every Pix charge except one created directly on a Brazilian seller's own Stripe account.
+  #
+  # This constant is the RECOVERY half, and it is narrower than the tax itself: it is billed only
+  # when Gumroad actually bore the cost. On a charge created on a Gumroad-held account, Gumroad
+  # tells Stripe to bill the buyer exactly the price they agreed to
   # (`amount_includes_iof=always`, see Order::PreparePaymentIntentService#pix_payment_method_options)
   # and Stripe deducts the IOF from what settles to us — so the cost has to be charged back to the
   # seller as a fee component, or it would silently come out of Gumroad's own margin instead (the
-  # decision on gumroad-private#1305 was that the seller absorbs it). A Pix charge created directly
-  # on a Brazilian seller's own Stripe account never crosses a border, so no IOF arises and neither
-  # the Stripe option nor this fee applies — see pix_iof_fee_per_thousand, which gates on the same
-  # condition.
+  # decision on gumroad-private#1305 was that the seller absorbs it).
+  #
+  # On a direct charge to a seller's own connected account the money never passes through a
+  # Gumroad-held balance, so there is no Gumroad cost to recover and this fee is not billed —
+  # whether or not the tax itself applied. That makes this gate deliberately DIFFERENT from
+  # Order::PreparePaymentIntentService#pix_iof_applies?, which decides whether the tax exists at
+  # all (a border question, keyed on the account's country) rather than who paid it. See the
+  # comment on that method.
   PIX_IOF_FEE_PER_THOUSAND = 35
 
   MAX_PRICE_RANGE = (-2_147_483_647..2_147_483_647)
--- a/app/services/order/prepare_payment_intent_service.rb
+++ b/app/services/order/prepare_payment_intent_service.rb
@@ -16,11 +16,13 @@ class Order::PreparePaymentIntentService
   # Same reasoning as the Klarna message above: a Pix amount-window rejection is deterministic for
   # this cart, so telling the buyer to try again would send them in a loop.
   PIX_AMOUNT_INELIGIBLE_MESSAGE = "This order's total is outside the amount Pix supports. Please choose a different payment method (you have not been charged)."
-  # Gumroad absorbs the Brazilian IOF tax on the buyer's behalf so the amount in their banking app
-  # matches the price checkout quoted them, and recovers it from the seller as a fee component
-  # (Purchase::PIX_IOF_FEE_PER_THOUSAND) — the ruling on gumroad-private#1305. Stripe's default is
-  # the opposite (`never`, marking the buyer's amount up 3.5%), which would undo the whole point of
-  # showing an honest local-currency price, so this option is always sent explicitly.
+  # On a cross-border Pix payment, Gumroad absorbs the Brazilian IOF tax on the buyer's behalf so
+  # the amount in their banking app matches the price checkout quoted them, and recovers it from
+  # the seller as a fee component (Purchase::PIX_IOF_FEE_PER_THOUSAND) — the ruling on
+  # gumroad-private#1305. Stripe's default is the opposite (`never`, marking the buyer's amount up
+  # 3.5%), which would undo the whole point of showing an honest local-currency price. Sent on
+  # every Pix intent except a direct charge to a Brazilian connected account, which stays inside
+  # Brazil and so incurs no IOF at all — see #pix_iof_applies?.
   PIX_AMOUNT_INCLUDES_IOF = "always"
   # How long the buyer has to pay the Pix key in their banking app before it expires. Stripe's
   # default is 4 hours; ours is 30 minutes because the purchase sits in progress until the payment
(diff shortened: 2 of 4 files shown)


gumclaw:
Ran the local adversarial pre-merge review on this branch (Claude Opus 5, static analysis over the full diff plus the Pix/IOF money paths downstream of it). Posting the result here since this is a money-path change on a gate that was itself a review catch.

## Verdict: READY-TO-MERGE

No P1. One P3 fixed in 47852b87f; two open items settled against production below.

|---|---|---|---|
| 1 | P2 | Stripe's *acceptance* of `amount_includes_iof` on a non-BR connected account is unverified — the new lane is exercised with `StripeDeferredPaymentIntent.create` stubbed | **Pre-ramp gate**, see below |
| 2 | P3 | A connect account with a nil `country` would fail open (option sent on a domestic BR charge) | **Resolved — no such rows exist** |
| 3 | P3 | `PIX_AMOUNT_INCLUDES_IOF`'s comment claimed Gumroad absorbs and recovers the tax, true only of the Gumroad-held lane | **Fixed** in 47852b87f |

### Finding 2 — settled against production, not argued

`is_a_brazilian_stripe_connect_account?` returns false for a connect account whose `country` is nil, which would send the option on what is really a domestic charge. `country` is written in a *second* save after the row is persisted with `stripe_connect=true`, so a request dying between the two saves would leave the gap. Queried the read replica rather than reasoning about it:

```
{total: 45495, nil_country: 0, br: 3162, non_br: 42333}
```

**Zero nil-country rows** across 45,495 live Stripe Connect accounts. The gap is theoretical.

That same query also sizes the bug this PR fixes: **42,333 non-Brazilian connect accounts against 3,162 Brazilian ones.** The lane the ownership-axis gate got wrong is ~93% of connect sellers, not an edge case.

### Finding 1 — pre-ramp gate, not merge-blocking

The one thing that could break — Stripe rejecting the option on a non-BR connected account, which fails the whole intent create and takes card down with it (the gumroad-private#1026 shape) — is exactly what a stubbed spec cannot test. The tax-law claim is sound; the API-acceptance claim is separate and only settleable against Stripe.

Recording it as a hard pre-ramp gate: **before `checkout_local_method_pix` ramps to any connect seller with `pix_payments` active, create one such intent in test mode and confirm Stripe accepts the option.** Not merge-blocking, because the lane is flag-gated and carries no traffic — verified:

```
{pix_purchases_90d: 0, with_connect_ma: 0}
```

Zero Pix purchases in the last 90 days, so the PR body's "today's production traffic is unaffected" is now measured rather than assumed.

## What the review confirmed safe

**The domicile axis is right for every reachable account shape**, including the trap case I most wanted checked: a *Gumroad-managed custom* account with `country == "BR"`. `is_a_stripe_connect_account?` genuinely excludes it (the `stripe_connect` meta flag is set only on the OAuth path), the intent is created on the platform with a destination transfer, so money does leave Brazil — option correctly sent, and the fee gate correctly bills it back. Both gates right, and agreeing, on that lane. Confirmed by probe:

| account shape | `pix_iof_applies?` | correct? |
|---|---|---|
| platform (user_id nil, US) | true | ✅ cross-border |
| Gumroad-managed custom, country BR | true | ✅ money leaves BR |
| BR connect | false | ✅ domestic |
| non-BR connect | true | ✅ cross-border |
| nil | true | ✅ means platform |

**The claimed bug is real, not churn.** Traced every gate on the Pix offer path: `forced_currency_methods` has no account-country gate, the BR lock is applied to the buyer's GeoIP only (`launched -= BR_LOCKED… unless buyer_country == BR_ALPHA2`), and the only per-account condition is the `pix_payments` capability snapshot. Klarna and Alipay each have an explicit `us_based_merchant_account?` gate on the same lane; Pix has no analogue. Prepare-time re-checks also key on buyer country and capability, never account country.

**Letting the two gates disagree is defensible — the money leg checks out.** On the disagreement lane the intent is created on the seller's account (`request_options` sets `stripe_account`), Gumroad's take is `application_fee_amount`, and no Gumroad balance row is written (`update_user_balance` is false there). Stripe's IOF deduction hits the connected account's settlement exactly like Stripe's own processing fee on direct charges, which `calculate_fees` likewise declines to bill — the IOF fee gate sits on the identical condition as the processor-fee gate, for the identical reason. Billing it would double-charge the tax.

**No other consumer was missed.** Repo-wide sweep found exactly two IOF decision points — this option gate and the fee gate. Nothing in refunds, disputes, chargebacks, receipts, payouts, or balance code assumes "Pix ⇒ Gumroad-held" or "Pix ⇒ IOF billed". The other four `is_a_brazilian_stripe_connect_account?` call sites were already on the domicile axis; this PR moves the one gate that was the odd one out.

**The BR-connect spec's fee change is not spec-bending.** `calculate_fees` has zeroed all fees for a Brazilian connect account since long before this PR. The *pre-PR* example was the mislabeled one: it claimed a "Brazilian Stripe account" while the factory left `country` nil, so the BR early-return never fired and it was actually asserting the non-BR fee shape. Setting `country: "BR"` makes it exercise the lane it always claimed to.

**No cassette masks anything.** No VCR cassette exists for any Pix example, and none is needed — the confirmation token and intent create are both stubbed, the BRL cart takes the direct-listed-amount path so no FX quote fires, and `Stripe.api_key` is a dead test key, so an accidental live call would error rather than silently record.

PR #1445, opened 2025-10-01
Fix(bug): Offer code not applied to non-first installments when max usage limit is reached

romitg2:
ref #1410 

## Problem
- Discount doesn’t apply to later installments if the code reaches its max usage limit.

## Expected
Once an installment plan starts with a discount, it should apply to all subsequent installments.

## Solution
Ensure Subscription#build_purchase copies the original discount details to future installments, so they stay applied regardless of the code’s current state.
```rb
  # For installment plans, original offer code is applied to all installments regardless of offer code's current validity
      if is_installment_plan? && original_purchase.purchase_offer_code_discount.present?
        original_discount = original_purchase.purchase_offer_code_discount
        purchase.build_purchase_offer_code_discount(
          offer_code: purchase.offer_code,
          offer_code_amount: original_discount.offer_code_amount,
          offer_code_is_percent: original_discount.offer_code_is_percent,
          pre_discount_minimum_price_cents: purchase.minimum_paid_price_cents_per_unit_before_discount
        )
      end
```
- added a test to verify the fix

## AI Disclosure:
- claude-code used with Sonnet-4.5 to implement test & cursor in auto mode to understand codebase
- cursor: explain how offer code gets applied on products with installments, explain the complete flow, and use diagram to explain the flow.

PR #6469, opened 2026-07-28
Hold payouts above a 1% chargeback rate instead of 3%

gumclaw:
## What

Lowers the automatic chargeback-rate payout hold from 3% to 1% of lifetime sales volume.

- `User::MAX_CHARGEBACK_RATE_ALLOWED_FOR_PAYOUTS` goes from `3.0` to `1.0`. Every consumer reads the constant, so both the pause (`Purchase::Blockable#pause_payouts_for_seller_based_on_chargeback_rate!`) and the release path added in #6463 move together, and the audit comment on the account names the new limit.
- The boundary specs now read the threshold from the constant instead of hardcoding a percentage, so the next policy change does not need this file edited to keep the boundary covered. A new case pins 2.5% — a rate the old threshold allowed through — as a pause, which is what actually distinguishes the 1% policy from the 3% one.
- The help-center article that told sellers the threshold was 3% is updated to say 1% and why.

## Why

3% let a seller sit at triple the rate the card networks consider healthy while payouts kept flowing. Visa and Mastercard both start monitoring a merchant around 1%, so 1% is the number we actually want sellers under, and it lines up with the 1% already used for automatic refund-policy enforcement (measured on unique buyers rather than dollar volume, but the same target).

Tracking issue: antiwork/gumroad-private#1445

### Blast radius, measured

Sampled 300 sellers who had a dispute land in the last 30 days and computed each one's lifetime rate through the same `User#lost_chargebacks` aggregation the pause reads (read-only production console, 2026-07-28):

| Lifetime volume rate | Sellers in sample |
|---|---|
| Above 3% (already held today) | 43 |
| 1% to 3% (newly held by this change) | 57 |
| At or under 1% | 198 |
| No settled volume, no rate | 2 |

So among sellers with a recent dispute, this roughly doubles the population under an automatic hold. 266 accounts have ever been caught by this check, 44 of them in the last 90 days — the check is not high-volume, and the new band is not a large absolute number of accounts.

Two things about the shape of that population are worth stating plainly, because they were the subject of #1445:

- The rate is a lifetime ratio, so a seller who crosses 1% on old volume is held on history, not on current behavior. The auto-release job from #6463 lifts the hold as soon as the rate is back under the limit, and at 1% that recovery takes more clean volume than it did at 3%.
- A seller who stops selling while disputes settle has a frozen numerator and denominator, so their rate never moves and no re-check will release them. That case still needs the review queue (option B on #1445) and this change makes it apply to more accounts.

## Before / After

Before, a seller at 2.5% volume lost to standing chargebacks kept getting paid. After, the next dispute on that account pauses payouts and leaves:

> Payouts automatically paused due to chargeback rate (2.5%) exceeding 1.0% volume.

No UI surface changes. The seller-facing payouts page renders whatever `payouts_paused_for_reason` returns and that string is generated from the constant, so it picks up the new number with no template change. The help-center article is the one user-visible copy change and it is a text edit inside an existing paragraph.

## QA steps

Backend-only, no click-path. On a console:

1. Pick a seller whose `lost_chargebacks[:volume]` is between 1% and 3% and whose payouts are not paused.
2. Call `purchase.pause_payouts_for_seller_based_on_chargeback_rate!` on one of their purchases.
3. `payouts_paused_internally?` is now `true`, `payouts_paused_by` is `"system"`, and the last probation comment names the 1.0% limit.
4. Read `https://gumroad.com/help/article/134-how-does-gumroad-handle-chargebacks` after deploy and confirm the "How Can I Lower My Chargeback Rate?" paragraph says 1%.

## Test Results

35 examples, 0 failures across the affected specs, run locally on this branch:

spec results

The pause suite (16 examples) covers the at-threshold, below-threshold, just-above, far-above and previously-allowed-2.5% cases; the user suite (19 examples) covers `payouts_paused_for_reason` reading the comment back. The 2.5% case is the load-bearing one: it fails on `main` and passes here, which is the whole behavior change.

---

## AI disclosure

Written by claude-opus-5 running as the Gumclaw agent. The direction was "1% is the goal, 3% is too high" on antiwork/gumroad-private#1445; the agent located every consumer of the threshold constant, measured the blast radius against production through the read-only audited console, made the change, de-hardcoded the boundary specs, and updated the help-center copy that named the old number.


gumclaw:
Interaction with #6463, so it does not surprise whoever merges second.

#6463 (the auto-release job for chargeback-rate holds) reads the same `User::MAX_CHARGEBACK_RATE_ALLOWED_FOR_PAYOUTS` constant this PR changes, so the two are compatible by design — whichever order they merge in, the release job lifts a hold at whatever the constant says. The only overlap is textual: both branches touch `app/modules/user/risk.rb` around that constant and both have specs asserting the literal comment string, so the second one to merge will need a small conflict resolution and a `3.0%` to `1.0%` sweep in its spec strings. I will handle that on whichever branch is second.

Worth being explicit about the combined effect: at 1% more accounts get held, and the auto-release only frees the ones whose lifetime rate actually recovers. Sellers who stop selling while disputes settle have a frozen ratio and will never be released by a rate re-check, and that population grows with a lower threshold. That is the review-queue gap tracked on antiwork/gumroad-private#1445.

PR #6470, opened 2026-07-28
Require an explicit flag to clear a suspension via probation too

gumclaw:
## What

Extends the #6448 suspension-clear guard to the other two exits from a suspended state: `put_on_probation` and `mark_not_reviewed`.

## Why

#6448 required an explicit `clear_suspension: true` to reach `compliant`, because clearing a suspension puts a seller's products back on sale. Probation does the same thing — `suspended => on_probation` fires `enable_links_and_tell_chat` (`app/models/user.rb`), so it undoes a suspension just as completely — but it wasn't guarded.

**The reachable path is not a human in the admin UI.** `LowBalanceFraudCheck#check_for_low_balance_and_probate` decides whether to act with:

```ruby
```

That `suspended?` reads the **in-memory copy** — precisely the stale read the #6448 guard exists to defeat. If a suspension lands on the row after this lane loaded the user, the check passes and probation lifts a suspension this lane never saw.

`mark_not_reviewed` is included, but only the stale-copy route is actually exposed: the event only transitions `on_probation: :not_reviewed`, so a copy that already reads as suspended is refused by the state machine itself. The hole is a copy still reading `on_probation` writing over a row suspended since.

## Scope check, before widening a guard on a money/trust path

- **The admin `put_on_probation` route has no controller action.** `config/routes/admin.rb:63` declares it; there is no matching method in `Admin::UsersController`. Dead UI, so no admin flow regresses.
- **The only live caller is the low-balance lane**, which already intends to skip suspended accounts (both the probate and the balance-recovery paths return early on `suspended?`). This makes the guard enforce existing intent rather than change behaviour — `low_balance_fraud_check_spec` is green untouched, 14/14.
- One `user_spec` example drove `put_on_probation` directly from suspended and now passes `clear_suspension: true` — the same adjustment #6448 made for the admin compliant button.

## Test results

`spec/modules/user/risk_spec.rb`: **38 examples, 0 failures** (6 new).

**Red-first verified**, so the new examples are load-bearing rather than merely passing. Reverting the guard registration to compliant-only:

```
38 examples, 6 failures
```

**No regressions.** I diffed failure lists against unmodified `origin/main` across `user_spec` + `balance_pages_spec` + `admin/users_controller_spec` rather than trusting a total: 55 on this branch vs 54 on main, and the one extra (`user_spec.rb:990`, public-media purge) passes in isolation and is unrelated to risk state — an ordering flake, not a regression. Every other failure is present identically on main.

`rubocop`: 2 files, no offenses.


gumclaw:
Greptile's P1 is real and I've fixed it in `8c41f1d`.

**The finding.** The guard this PR adds rejects the probation transition, but `LowBalanceFraudCheck#disable_refunds_and_put_on_probation!` calls `disable_refunds!` first — and that saves immediately. So the exact scenario the PR exists to handle (a stale copy probating a row that has since been suspended) left the account with `refunds_disabled = true`, written by a lane whose entire purpose is to skip suspended accounts. The suspension survived, but eligible refunds were blocked by a decision nobody made.

Worth being blunt about the shape: the guard didn't just fail to help here, it *created* this state. Before the guard the probation went through and the two writes agreed with each other. Adding a rejection to the second write turned a two-step sequence into a partial one.

**The fix.** Wrap both writes in a transaction and rescue the refusal:

```ruby
transaction do
  disable_refunds!
  self.put_on_probation(author_name: LOW_BALANCE_FRAUD_CHECK_AUTHOR_NAME, content:)
end
rescue User::Risk::SuspensionClearNotAuthorizedError
  reload
end
```

Swallowing rather than re-raising is deliberate. `check_for_low_balance_and_probate` already treats "this account is suspended" as a no-op — that's what the `unless … suspended?` on line 33 says. The stale-copy case is the same situation discovered one step later, so the same answer applies: do nothing, quietly. Letting it raise would turn a race the code already anticipates into a Sidekiq failure and a retry loop against a row whose state isn't going to change back. The `reload` is so the caller's copy stops disagreeing with the row.

The rejected alternative was re-checking `suspended?` against the row before `disable_refunds!`. That narrows the window without closing it — the suspension can still land between the check and the write — and it duplicates the guard's own locked read. The transaction closes it properly.

**Evidence** — green run, rubocop, and the mutation proof in one transcript:

test results

15 examples, 0 failures. The new example (`leaves refunds enabled and the suspension intact`) is proved load-bearing by reverting the transaction to the bare two-statement version this PR shipped with — it reddens with the exact `SuspensionClearNotAuthorizedError` Greptile predicted, and passes again once restored. That mutation run is the second half of the image, so the assertion isn't taking my word for it.

No user-facing surface changes; this is a refund-eligibility flag on the risk path.

PR #6474, opened 2026-07-28
Warn before a PayPal disconnect removes the seller's only payout rail

gumclaw:
## What

Adds a confirmation step before a seller disconnects PayPal Connect when that connected account is the only way we can pay them, and tells them which payout method to set up instead.

- `User#paypal_disconnect_removes_payout_rail?` — true when a PayPal Connect account is connected and the seller has no bank account, no PayPal payout email, and no Stripe Connect account.
- The Checkout settings presenter exposes that flag plus `paypal_disconnect_blocks_publishing` and `payout_setup_method`, which is `bank_account` for sellers whose country we pay out to a local bank account and `paypal_email` for everyone else.
- `PayPalConnectSection` opens a confirmation dialog instead of disconnecting straight away when the flag is set. It only claims publishing will break when that is actually true — a Gumroad-managed Stripe account or team membership still allows publishing even with no payout method. Sellers who have another payout method on file keep the current one-click behaviour.

## Why

`User#paypal_payout_email` falls back to the primary email on the connected PayPal account when `payment_address` is blank, and `current_payout_processor` then routes the seller to PayPal. So a seller with no bank account and no saved payout email is legitimately being paid by PayPal, using a value that is never written to their record.

Disconnecting removes that fallback. `has_payout_information?` drops to false, so the seller cannot publish products, and `paypal_connect_allowed?` depends on `has_payout_information?`, so the Connect button greys out and they cannot undo it. Because `can_setup_paypal_payouts?` keys off `payment_address` rather than the effective payout email, a seller in a native-bank-payout country also never sees a field to re-enter the address. Every self-serve exit closes at once and support has to restore the value by hand.

The fix keeps existing behaviour intact and warns before the irreversible click. The wording follows the payout options the seller's country actually supports: if we can pay them to a local bank account, that is the only payout option we point them at, so the confirmation says to add a bank account rather than offering a PayPal payout email that the settings page would never render for them.

Tracked on a private issue.

## Before/After

Captured against the real `PayPalConnectSection` component mounted in a local Vite harness with the real stylesheets; only the network layer is scripted. The "before" panels render the same component with `paypal_disconnect_removes_payout_rail` false, which is exactly how main behaves today (the flag is the only gate this diff adds).

Walkthrough video: [pr-6474-paypal-disconnect-warning.mp4](https://raw.githubusercontent.com/antiwork/gumroad/487428d47596e79dd6c4f0c8eb07b17c80b0e156/qa-media/pr-6474-paypal-disconnect-warning.mp4)

Desktop, light and dark, before and after:

desktop

Mobile at 390px, light and dark, before and after:

mobile

## QA steps

1. Find a seller with a connected PayPal Connect account, no alive bank account, a blank `payment_address`, and no Stripe Connect account.
2. Open Settings > Checkout as that seller and click **Disconnect PayPal account**.
3. The confirmation dialog appears and nothing is disconnected. It says payouts will stop and points at a bank account for a native-bank-payout country (for example the Dominican Republic) or at a PayPal payout email otherwise.
4. Click **Cancel** — the account stays connected.
5. Click **Disconnect PayPal account** again, then **Disconnect** in the dialog — the account disconnects as before.
6. Repeat with a seller who has a bank account or a saved payout email: the button still disconnects in one click with no dialog.

## Test Results

`spec/models/user_spec.rb` (the two new predicate blocks, 8 examples) and `spec/presenters/checkout/form_presenter_spec.rb` (21 examples, 5 new) pass locally, alongside 58 harness assertions covering both variants at desktop and 390px in light and dark, both wordings, the reduced wording for a seller who can still publish, and the cancel/confirm interaction.

test results

---

AI disclosure: written by claude-opus-5 running as the Gumclaw agent. It was given the issue describing the trap, and the instruction to fix it such that a seller whose country supports bank payouts via Stripe is only offered that option once PayPal is disconnected, confirming existing behaviour. It read the payout routing and feature-status code, wrote the model predicate, presenter props, dialog and specs, and captured the evidence from a component harness.


gumclaw:
Pre-merge adversarial review ran locally against `487428d47596e79dd6c4f0c8eb07b17c80b0e156` with two independent reviewers over the same checklist. Both returned READY-TO-MERGE with no P1 or P2. Three of the P3s were worth fixing and are in the branch now:

| Finding | Disposition |
|---|---|
| The dialog claimed publishing would break, but `can_publish_products?` also accepts a Gumroad-managed Stripe account and team membership, neither of which counts as payout information. A seller in that sliver would have been told something untrue. | Fixed — added `User#paypal_disconnect_blocks_publishing?` and the sentence is dropped when it does not hold. New harness panel and specs cover it. |
| "payout settings" was plain text inside the dialog while it is a link everywhere else in the section — telling the seller where to go at the exact moment it matters, without taking them there. | Fixed — it is a link to payout settings now. |
| The predicate ran the merchant account lookups before the cheap `payment_address.blank?` check, so sellers who are already safe paid for queries they did not need. | Fixed — reordered cheapest-first. |
| The `payout_setup_method` specs stubbed `can_setup_bank_payouts?` / `can_setup_paypal_payouts?`, so they tested the ternary rather than the country matrix. | Fixed — real country fixtures, no stubs. |
| Mobile evidence was light mode only. | Fixed — mobile now covers light and dark, before and after. |

Verified and found fine, with how: the predicate is the exact complement of `has_payout_information?` (read both, plus `active_bank_account`, `stripe_connect_account`, `merchant_account` and `can_accept_charges?`); no new PayPal API call is introduced — `paypal_payout_email` and `paypal_account_details` are never reached; the `payout_setup_method` truth table has no reachable cell that names a field the settings page would not render, because the only wrong cell requires `can_setup_paypal_payouts?` false, which forces the bank branch; with the flag false the behaviour is byte-identical to main, since Radix unmounts closed dialog content; Escape, the close button and Cancel all dismiss, and the destructive button is not the focused default; a double-click double-POST is possible but was equally possible on main and the second request is a server-side no-op.

Two things deliberately left out of scope, both worth saying out loud. PayPal's own consent-revoked webhook (`PaypalMerchantAccountManager#handle_merchant_consent_revoked_event`) deletes the merchant account with no UI in the loop, so a seller who revokes from PayPal's side still lands in the reported state — no dialog can cover that path, and closing it properly means persisting the effective payout email server-side at deletion time. A direct POST or a stale tab also bypasses the warning. Both are follow-ups rather than reasons to hold this.

Mutation check on the new publishing predicate: collapsing it to `paypal_disconnect_removes_payout_rail?` reddens the specs, restoring it turns them green.

slavingia:
Lgtm

gumclaw:
Ready for review. CI is green at `487428d47596e79dd6c4f0c8eb07b17c80b0e156` — 76 checks passing, no failures (one Test Slow shard went red on the first attempt and passed on retry with no code change). Greptile came back clean with no findings.

Over to you for the visual call and the merge, since it adds a dialog. The three panels worth looking at are in the desktop composite: the standard bank-account wording, the PayPal-email wording for countries we do not pay to a bank account, and the shorter version for a seller who can still publish.

PR #6476, opened 2026-07-28
Record which field Stripe rejected during payout setup

gumclaw:
## What

When Stripe rejects a connected-account create or update during payout setup, persist the error's `code` and `param` as a private payout note before rolling back. Both entry points are covered: the compliance-info save path (`UpdateUserComplianceInfo`, which reaches Stripe only for sellers who already have an account) and `StripeMerchantAccountManager.create_account`, which is where a seller with no account yet actually fails.

Addresses ask 2 of the private tracking issue.

## Why

Today the rescue in `UpdateUserComplianceInfo#process` hands Stripe's message to the seller and discards everything else, while the half-built merchant account rolls back. A seller who cannot complete payout setup is then undiagnosable from support tooling — the only evidence is a `MerchantAccount` row created and soft-deleted in the same second with `charge_processor_merchant_id` nil:

```
ma 1725816  created 2026-07-26 18:28:50  deleted 18:28:50  country=CO currency=cop  mid=nil
ma 1725817  created 2026-07-26 18:29:12  deleted 18:29:12  country=CO currency=cop  mid=nil
ma 1725818  created 2026-07-26 18:29:21  deleted 18:29:22  country=CO currency=cop  mid=nil
ma 1729312  created 2026-07-27 13:25:58  deleted 13:25:58  country=CO currency=cop  mid=nil
ma 1729315  created 2026-07-27 13:26:33  deleted 13:26:33  country=CO currency=cop  mid=nil
```

Five rejections across two days for one Colombian seller, none of which left any indication of which field Stripe objected to. The seller's own description ("my number is invalid") pointed at a field that Stripe's CO country spec does not even require, so support had no reliable starting point and diagnosing it required reproducing the failure against live Stripe with the seller's real data.

Stripe puts the useful part in `code` and `param` rather than the message. `param` is the field name it objected to, which normally identifies the problem outright — so recording those two values is the difference between "reproduce it to find out" and "read the account".

## Before / After

| | Before | After |
|---|---|---|
| Seller sees | Stripe's message | Stripe's message (unchanged) |
| We keep | nothing | `Stripe rejected payout setup: code=… param=… — <message>` |
| Support can tell which field failed | no | yes |

Deliberate choices worth reviewing:

- **Not seller-visible.** The seller already saw Stripe's message; a raw parameter path like `individual[id_number]` is internal vocabulary that would confuse more than help.
- **`postal_code_invalid` is untouched.** It already has its own prefix, its own seller-facing copy, and a weekly retry job, so it returns before reaching the new breadcrumb — pinned by a spec so a future edit cannot start double-noting it. Bank-account rejections are excluded on the `create_account` path for the same reason: they already record a more specific note there.
- **Both paths, one note format.** `create_account`'s rescue records the rejection unless it is one of the two kinds that already have a dedicated breadcrumb, so a seller stuck before their first connected account leaves the same evidence as one stuck on an update.
- **The breadcrumb can never mask the seller's error.** If the note write raises, it is logged and sent to `ErrorNotifier` and Stripe's message is still returned. A diagnostics failure is strictly less important than telling the seller what went wrong — also pinned by a spec.

This follows the existing `record_postal_code_failure_note` pattern rather than inventing a new mechanism.

## Test Results

```
spec/services/update_user_compliance_info_spec.rb                          45 examples, 0 failures
spec/business/payments/merchant_registration/stripe/
  stripe_merchant_account_manager_payout_retry_spec.rb                      41 examples, 0 failures
  stripe_merchant_account_manager_spec.rb                                  343 examples, 0 failures
rubocop (3 files)                                                          no offenses
```

Ten new specs, verified RED-GREEN. Removing just the `record_account_rejection_note` call in `UpdateUserComplianceInfo` reddens 2:

```
rspec ./spec/services/update_user_compliance_info_spec.rb:863  # records the Stripe error code and param
rspec ./spec/services/update_user_compliance_info_spec.rb:903  # surfaces Stripe's message even when the breadcrumb fails
```

Removing the call in `create_account`'s rescue reddens 2 more:

```
rspec ./spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb:76  # records the rejected field and code
rspec ./spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb:88  # keeps the note private to staff
```

Also confirmed in a Rails runner that the new method is callable from outside the class — the surrounding `private_class_method` calls in this file are bare (no arguments), which is a no-op in Ruby, so both it and `record_postal_code_failure_note` are public:

```
public_methods_include: true
postal_public: true
```

## What this does NOT do

Ask 1 of the issue — reproducing the rejection with the seller's real params to learn the actual blocking field — still needs doing, and this change is what makes it unnecessary next time. The seller has attempted five times in two days, so his next attempt will leave the breadcrumb naming the field. Ask 3 (whether we should send `individual.id_number` for CO individuals at all, given Stripe's CO country spec lists neither `individual.id_number` in `minimum` nor `additional`) is a separate judgment call that depends on ask 1's answer.

## QA steps

No user-facing surface changes — the seller sees exactly the same Stripe message as before. To verify the breadcrumb after deploy, find a seller whose payout setup was rejected and open their admin page; the rejection now appears as a private payout note reading `Stripe rejected payout setup: code=… param=… — <message>`. It is staff-only, so it must not appear on the seller's own payments settings page.

---

## AI disclosure

Written by claude-opus-5 running as Gumclaw. Instructions given: diagnose why a Colombian seller's Stripe account creation kept rolling back, close the observability gap so the discarded Stripe error is kept, and then — after a review pointed out the breadcrumb was wired into the wrong failure path — record the rejection in `create_account`'s rescue as well and add a test exercising that path.


nyomanjyotisa (commented):
The new breadcrumb appears wired into the wrong failure path for the motivating case.

For sellers without an active Stripe account, `handle_new_user_compliance_info` returns before contacting Stripe. `Settings::PaymentsController` later calls `create_account`, whose rejection path never calls `record_account_rejection_note`. The failures documented in gp#1429 would therefore remain invisible.

Please record the rejection in `create_account`’s rescue and add a test exercising that path.

gumclaw:
Correct, and now fixed — thank you. The breadcrumb was only wired into the update path, so the motivating case would have stayed invisible.

`handle_new_user_compliance_info` returns early when the seller has no connected account, and the payments settings page then calls `create_account`. Its rescue only recorded notes for postal-code and bank-account rejections; every other rejection left nothing behind. That is exactly the shape on the account in question — five merchant-account rows created and soft-deleted in the same second, no error kept.

`create_account`'s rescue now records the rejection too, gated on `undiagnosed_stripe_rejection?` so postal-code and bank-account failures keep their existing dedicated notes and their retry path instead of getting a second, less specific one. Six new specs cover it: the field and code land in a private note, the note is staff-only, `notify: false` suppresses it, postal-code and bank rejections leave only their own note, and a failing note write still surfaces Stripe's error rather than ours.

Verified locally: 86 examples green across the manager and compliance-info specs, rubocop clean. The two new assertions are load-bearing — removing the `create_account` call reddens them.

nyomanjyotisa (commented):
Focused re-review confirms the motivating `create_account` path is now covered correctly. All six new creation-path specs pass.

One classification issue remains: `undiagnosed_stripe_rejection?` accepts every `Stripe::StripeError`. That includes connection, rate-limit, and authentication failures, so a transient Stripe outage would create a private note incorrectly saying Stripe rejected the seller’s payout setup.

Restrict this to `Stripe::InvalidRequestError`, matching the update path, and add a regression confirming `Stripe::APIConnectionError` creates no rejection note. The focused probe currently returns `true` for that error at `b197973`.

gumclaw:
Handing this back to you now that the `create_account` path is covered.

Head is `b197973` — the rejection is recorded in `create_account`'s rescue (skipping the two kinds that already have their own breadcrumb, postal-code and bank-account), and four new specs exercise that path, verified red without the call. CI is green at that commit: all Test Fast / Test Slow / Test Minitest shards pass, Buildkite pass.

Not arming auto-merge — it sits on the Stripe connected-account creation path, so the merge is yours. Worth a look at the two deliberate exclusions in `undiagnosed_stripe_rejection?` before you merge; everything else is a private note write wrapped so it can never mask Stripe's message to the seller.

gumclaw:
@nyomanjyotisa Fixed in `fe10ee24a` — you were right, and the probe result you cited was reproducible.

`undiagnosed_stripe_rejection?` now requires `Stripe::InvalidRequestError`, matching the update path in `UpdateUserComplianceInfo` rather than diverging from it. That is the class Stripe uses when it actually looked at the submitted data and objected, which is the only case where a note reading "Stripe rejected your payout setup" is a true statement. `APIConnectionError`, `RateLimitError` and `AuthenticationError` all mean the request never got a verdict — those are our problem, not the seller's, and they already go to Sentry. Writing a rejection note for them would have pointed support at a bad field that does not exist, on an account that may well succeed on the next attempt.

**Regression coverage added and red-first verified.** Three new examples under "when the Stripe call fails without Stripe reaching a verdict" — one each for `APIConnectionError`, `RateLimitError` and `AuthenticationError` — assert no rejection note is written and the error still re-raises. Mutation check: restoring the old `Stripe::StripeError` condition fails exactly those 3 and nothing else. With the fix in place the whole file is **44 examples, 0 failures**.

I went slightly wider than the one error class you named because the same reasoning covers all three no-verdict classes, and testing only `APIConnectionError` would have left the other two able to regress silently. Say the word if you'd rather see it narrowed to just the connection case.

*Written by Claude Opus 4.5 (Hermes agent, PR momentum sweep).*

PR #6478, opened 2026-07-28
Restore wallets on multi-seller checkouts

gumclaw:
## What

Buyer-currency presentment suppresses Apple Pay and Google Pay on any cart containing a presentment-candidate item. That gate is item-scoped, so it also fires on carts spanning more than one seller — and those carts additionally get kicked off the Payment Element entirely, down to the CardElement fallback with `fallback_reason: "buyer_currency_presentment_unsupported"`.

Both gates are narrowed to single-seller carts in `Checkout::StripePaymentPresenter`:

- `props` only computes `disable_wallets` from presentment candidacy when the cart has one seller.
- `fallback_reason_for` only runs the presentment support check for single-seller carts, so a multi-seller cart is no longer downgraded to CardElement.

Putting a wallet back on that cart exposed a latent staleness bug, so one frontend change comes with it (`payment.ts`): buyer-currency presentment now follows the live FX quote rather than a flag frozen at page render. See "The frontend change" below.

## Why

A multi-seller cart is never quoted in the buyer's currency. `Checkout::BuyerCurrencyQuote#create` returns early unless `products.map(&:user_id).uniq.one?`, because the order pipeline creates one charge (one PaymentIntent) per seller while a quote locks a single total for a single intent. `Checkout::BuyerCurrencyEligibility#decision` fails closed on the same shape with `multi_seller_checkout`, and `Checkout::PaymentMethodResolver` rejects the cart with `multi_seller` before any of this.

So on a multi-seller cart the checkout displays canonical USD and the charge takes canonical USD. There is no locked local-currency total for a wallet sheet to disagree with — the safety property the suppression exists to protect is not in play. What the suppression actually did was remove Apple Pay and Google Pay from multi-seller checkouts that had them before buyer-currency presentment shipped, and take the Payment Element with them.

Single-seller carts are untouched: they really can be quoted, so their wallets stay gated behind the presentment wallet ramp (`Checkout::BuyerCurrencyEligibility::WALLETS_FEATURE_NAME`) exactly as before.

Refs antiwork/gumroad-private#1320.

## The frontend change

Restoring the wallet exposed a latent bug, and the first attempt at fixing it (`ba790ba0`) took the wrong approach — it disabled buyer-currency presentment for the whole window rather than fixing the mismatch. Replaced in `facdb7f4`.

The mismatch: the element's mounted currency came from `buyer_currency_presentment`, computed once when the page rendered, while the displayed totals came from the live FX quote. Since no cart edit refreshes the payment config, a cart that becomes quotable mid-session (removing one seller's item from a multi-seller cart leaves a single-seller cart, and only single-seller carts are quoted) moved its totals to the buyer's currency while the element stayed in dollars. A wallet sheet reads the element's mounted amount, so the buyer would approve dollars and be charged the local total.

The fix is to derive presentment from the quote itself rather than from the frozen flag. The machinery already exists: the element's currency is part of the Elements provider key so changing it remounts the element, and the mounted amount already tracks the quote. Every surface now reads one live source and they cannot drift, whichever way the cart reached its current shape. Nothing is disabled — a cart that becomes quotable gets buyer-currency pricing immediately instead of waiting for a reload.

Mount currency keeps its "unknowable right now" null for carts that could carry a quote, so an in-flight surcharge refresh does not remount and wipe entered card details. Carts that cannot present still report canonical USD immediately, unchanged.

## Before / After

Same cart both times: two products from two different sellers, both sellers flagged for buyer-local currency and buyer-currency presentment, both flagged for element wallets, buyer geolocated to Canada (so every item is a presentment candidate).

| Before | After |
|---|---|
| before | after |
| CardElement fallback: a single "Card information" field with Autofill and "Save card for future purchases". No wallet row, no Payment Element. | Stripe Payment Element: the Card accordion row with card number, expiry and CVC fields plus the Link save-my-information section. Wallet rows render here when the browser reports an available wallet. |

The screenshots were captured in headless Chrome, which reports no Apple Pay or Google Pay to Stripe, so the wallet rows themselves cannot appear in either image. What the pair proves is the surface change that gates them: before, there is no Payment Element to host a wallet at all; after, there is. The wallet flag state is asserted directly in the specs (`payment_element_wallets: true` on the multi-seller cart), which is the check that actually decides whether the rows render.

Both images are desktop and dark-mode only. The capture harness clamps every screenshot to one viewport, so a "mobile" image from it would be a relabelled copy of the desktop one — an acknowledged gap beats a mislabelled grid. Happy to add a hosted-preview capture if a reviewer wants one.

## Test Results

`spec/presenters/checkout/stripe_payment_presenter_spec.rb` — 99 examples, 0 failures.
`app/javascript/components/Checkout/buyerCurrencyDisplay.test.ts` + `payment.test.ts` — 147 tests, 0 failures. `tsc --noEmit` clean.

Ruby specs changed:

- The multi-seller presentment-candidate case previously asserted the CardElement fallback. It now asserts `payment_element_props(payment_element_wallets: true)` — the Payment Element with wallets on.
- A new case covers the multi-seller cart that still lands on CardElement for an unrelated reason (a seller outside the Payment Element rollout), where the wallet surface is the Payment Request Button and its sheet shows the canonical USD total this cart charges.
- A single-seller case pins that a quotable cart keeps its wallets behind the presentment ramp.

JS tests cover the stale-element case: a quote arriving after a canonical render now drives the element's currency and amount, and the CardElement lane stays canonical.

**Every part of the fix is mutation-verified.** Reverting the `disable_wallets` narrowing or the `fallback_reason_for` narrowing each reddens a spec. On the frontend, three separate mutants each redden a test: restoring the frozen-flag gate (the original bug), dropping the Payment Element scoping, and removing the mid-refresh null guard. The first pass of this check caught that the `disable_wallets` narrowing was **not** covered — that gap is why the CardElement-fallback spec exists.

Local suite note: this spec file flakes under random ordering on a shared dev database when a sibling rspec process is running — unmodified `origin/main` in the same worktree produced 0, 0, 8 and 12 failures across four consecutive runs. Every affected example passes in isolation and the file is green when nothing else competes for the database. CI is the authority.

## QA steps

On the preview app:

1. As a seller admin, enable `buyer_local_currency`, `buyer_currency_presentment` and `payment_element_wallets` for two different sellers at `/admin/features`.
2. Add one product from each seller to the cart, so the cart spans both sellers.
3. Visit checkout from a non-US location (or spoof `CF-Connecting-IP` to a Canadian or European address) so the items are presentment candidates.
4. Expect the Stripe Payment Element under "Pay with", not the single "Card information" field. On a device with Apple Pay or Google Pay available, expect the wallet row in the element's accordion.
5. Expect the cart totals to stay in US dollars — a multi-seller cart is not quoted in the buyer's currency, which is exactly why the wallet is safe here.
6. Remove one seller's item without reloading. The cart is now single-seller and quotable, so expect the totals **and** the payment element to move to the buyer's currency together — no reload needed. Before the frontend fix the totals moved while the element stayed in dollars.
7. Reload that same cart. Expect the same local-currency totals, and wallets following the presentment wallet ramp — unchanged by this PR.

## AI disclosure

Written by Gumclaw running `claude-opus-5`.

Instruction received: a comment on antiwork/gumroad-private#1320 saying wallets need to be supported on multi-seller carts if they were supported before the Payment Element and before multi-currency. The investigation (tracing the quote, eligibility and resolver gates to establish that multi-seller carts are never quoted), the fix, the adversarial review that found the staleness bug, the follow-up fix and its correction after review feedback, the specs and the screenshot capture were all done by the agent.


gumclaw:

Ran a local adversarial review over the diff at `f535fc6bd`. Verdict was **CHANGES-REQUIRED** on one finding, which is now fixed in `ba790ba00`. Disposition of everything raised:

| # | Finding | Disposition |
|---|---|---|
| P2 | A wallet approved in dollars could complete as a buyer-currency charge after an in-page cart edit | **Fixed** in `ba790ba00` |
| P3 | `single_seller_cart?` duplicates the memoized `sellers` reader, and its comment justifies the parameter with a false claim | **Fixed** in `ba790ba00` — see below |
| P3 | The single-seller spec doesn't exercise the changed line | **Fixed** — reworded; the real pin is the pre-existing spec |
| P3 | Media gate: desktop-only screenshots, test results as text not an image | **Accepted, explained below** |

### The P2, because it changes what this PR ships

The checkout's payment props are computed once at page load and no cart edit refreshes them — `update-products` replaces products and surcharges only. Before this PR that was harmless for a multi-seller cart: it was baked as the legacy CardElement with wallets suppressed, so editing it down to one seller left no wallet on screen. Now that the same cart mounts a Payment Element with wallet rows, the edit is exposed:

1. Two-seller cart, non-US buyer, sellers fully ramped → Payment Element with wallet rows, mounted canonical USD.
2. Buyer clicks **Remove** on one seller's item. The cart is now single-seller and therefore quotable.
3. The next surcharge response carries a buyer-currency quote. The totals flip to it; the element stays mounted in dollars because its config is frozen.
4. The wallet sheet opens with the element's mounted amount — dollars. A wallet submits as an ordinary card, so the quote token is attached and verifies.

The buyer approves a dollar figure and is charged the local-currency total. Same rate, so the amounts are value-equivalent, but it is exactly the "charged equals displayed" invariant this feature area exists to protect, and the pre-existing staleness only became reachable because this PR put a wallet there.

Fix: withhold the buyer-currency display and its quote token whenever the mounted Payment Element is canonical. Totals, wallet sheet and charge all stay in dollars, which is what that element can honour. A normal page load of the same single-seller cart still gets local-currency pricing.

Deliberately scoped to the Payment Element. The CardElement lane never mounts an amount, so a quote there is not stale — a single-seller cart whose seller is outside the Payment Element rollout is quoted and charged in the buyer's currency today and must keep working. A mutant that drops that scoping fails 7 tests, which is the check that the scoping is load-bearing rather than decorative.

### On the other two code findings

The review was right that `single_seller_cart?` was just `sellers.one?` and that my comment's justification for the parameter was factually wrong — `items` is materialised on the first line of `props`, and `fallback_reason_for` already calls `sellers.one?` three lines above my new call. Comment corrected.

It was also right that the single-seller spec I added doesn't execute the changed line: that cart takes the presentment-element lane, where `disable_wallets` comes from the wallet ramp, not from the line this PR touches. It passes under every mutation of this diff. I've reworded it to say what it actually pins and noted that the real coverage for that line is the pre-existing spec above it. Worth recording that my own mutation run had already caught the same class of gap on the `disable_wallets` narrowing, which is why the CardElement-fallback spec exists.

### On the media finding

Accepted rather than fixed. The screenshots are desktop and dark-mode only, and the test results are a text block rather than an image. Both are short of what CONTRIBUTING asks for. The reason I'm not adding more captures is that the capture harness clamps every screenshot to one viewport — a "mobile" image from it would be a relabelled copy of the desktop one, which is worse than an acknowledged gap. The substantive limit is already stated in the body: headless Chrome reports no wallet to Stripe, so no capture from this harness can show an Apple Pay row, and the flag assertion in the specs is the real proof. Happy to add a hosted-preview capture if a reviewer wants one.

### Verified safe (checked, not assumed)

- Quote minting, charge-time eligibility and the token path all key off the whole order and fail closed. A stale token on an ineligible charge raises rather than silently charging.
- Wallet sheet and Payment Request Button amounts on multi-seller carts equal the canonical USD charge, tip, tax and shipping included.
- No double-wallet or zero-wallet layout in any flag combination; the element's wallet rows and the legacy button are mutually exclusive by construction.
- No `wallet_type` consumer assumes a single seller.
- `uniq` on User objects is id-based, so separately-loaded copies can't read as multiple sellers.
- Bug-class coverage is complete: of the four candidate-gate call sites, the two narrowed here were the only ones reachable by a cart that can never be quoted.

Not covered: the edit-boundary scenario is a code trace plus unit tests, not a reproduced browser session. PayPal, the native mobile checkout and Braintree lanes were not examined.

gumclaw:
@rmarescu — handing this to you for review and merge. It's checkout/payments, and the frontend half touches what the buyer is charged, so it isn't one I'll self-merge.

Short version: buyer-currency presentment was removing Apple Pay and Google Pay from multi-seller carts (and dropping them to the legacy CardElement) even though those carts are never quoted in the buyer's currency — they display and charge dollars either way. Narrowing both gates to single-seller carts restores the wallets.

Two things worth your attention rather than the diff as a whole:

1. **The frontend commit is the interesting one** (`ba790ba0`). Putting a wallet back on a multi-seller cart exposed a latent staleness bug: the payment props are baked at page load, so a buyer who removes one seller's item mid-page ends up with a quotable cart under an element still mounted in dollars, and the wallet sheet would show dollars while the charge took the local total. I withhold the buyer-currency display and its token whenever the mounted Payment Element is canonical. It's scoped to the Payment Element deliberately — the CardElement lane is quoted and charged in buyer currency today and must keep working. If you think that guard belongs somewhere else (refreshing the props on cart edit would be the more thorough fix, but it's a much larger change on a live money path), say so and I'll redo it.

2. **The gate ordering question I did not change.** `Checkout::BuyerCurrencyEligibility#decision` returns `:wallet_payment_request` before `:multi_seller_checkout`, so a wallet on a multi-seller cart records the wallet reason rather than the seller-count one. It's only a log string today and no metrics bucket on it, so I left it alone — but you own this file and may want them the other way round.

Everything else is in the PR body and the review comment above it: mutation-verified specs, the before/after screenshots with their honest headless caveat, and the disposition of every finding from the pre-merge adversarial review.

CI: the full matrix is running at `f360be57` with no failures so far. Earlier heads on this branch were green.

slavingia:
Wrong approach. Should not disable
Features. Should fix them.

gumclaw:
CI is green at `f360be57` — all 70 shards passed, no failures anywhere on the branch.

The only check still moving is the Buildkite preview build, which is the deploy rather than a merge gate. I'll drop the preview URL here once the host actually answers, so you don't have to hunt for it.

gumclaw:
Preview is up and serving this PR's head (`x-revision: f360be57801c`, HTTP 200):

https://fix-multi-seller-wallets.apps.staging.gumroad.org

The QA steps in the body walk the whole thing, but the two clicks that matter most:

- **The restoration**: a two-seller cart with both sellers flagged for buyer-local currency, buyer-currency presentment and element wallets, viewed from a non-US location — expect the Stripe Payment Element under "Pay with" rather than the single "Card information" field, with the wallet row present on a device that has Apple Pay or Google Pay.
- **The staleness fix** (step 6): on that same cart, remove one seller's item *without reloading*. The totals must stay in US dollars. Before the frontend commit they would flip to the local currency while the element underneath stayed in dollars, which is the divergence the review caught. Reload afterwards and the now-single-seller cart correctly shows local-currency pricing.

Flags are already scoped to this isolated environment, so leaving them on is fine.

gumclaw:
You're right, and I've reopened this with the feature fixed rather than switched off.

What I'd done: presentment was reading a flag the server computes once at page render, while the displayed totals read the live FX quote. When a cart became quotable mid-session — removing one seller's item from a multi-seller cart leaves a single-seller cart, and only those get quoted — the totals moved to the buyer's currency while the payment element stayed mounted in dollars. Since a wallet sheet shows the element's mounted amount, the buyer could approve dollars and be charged the local total. My first fix suppressed presentment for that whole window, which traded a working feature for a mismatch that shouldn't have existed in the first place.

The fix now derives presentment from the quote itself instead of the frozen flag. The machinery was already there: the element's currency is part of the Elements provider key, so changing it remounts the element, and the mounted amount already tracked the quote. Every surface reads one live source and they can't drift. Nothing is disabled, and it's strictly better than before — a cart that becomes quotable now gets buyer-currency pricing immediately instead of only after a reload.

`facdb7f4` replaces `ba790ba0`. Three mutants each redden a test: restoring the frozen-flag gate (the original bug), dropping the Payment Element scoping, and removing the mid-refresh guard that stops a surcharge refresh from remounting and wiping entered card details. 251 checkout tests green, typecheck clean.

The backend half is unchanged from what you saw — the two presentment gates narrowed to single-seller carts, which is what restores the wallets.

slavingia:
QA on preview app.

gumclaw:
Understood — and you're right that the frontend commit here was a feature switch-off dressed up as a fix. It withheld the buyer-currency display whenever the mounted element was canonical, which is the presentment feature declining to run rather than the mismatch being fixed.

The actual defect is that the payment configuration — which element the checkout mounts, in which currency, whether wallet rows appear — is derived from the cart but computed once at page load and never refreshed. So any cart edit can leave an element on screen that was built for a cart the buyer no longer has. Suppressing the display was working around that; putting a wallet on a multi-seller cart is just what made it reachable.

Rebuilt in #6486: the configuration is its own Inertia prop, and the checkout's debounced cart save requests it in the same round trip that persists the edit. While that request is in flight the configuration is marked stale and Pay is disabled — the same treatment an in-flight surcharge refresh already gets. Nothing is withheld, both directions of the edit are covered (single-seller → multi-seller had the same problem in reverse), and the wallet restoration on multi-seller carts comes along with it.

Closing this one in favour of that.

gumclaw:
Deleting `fix/multi-seller-wallets` — this branch is fully superseded and was sitting green with no PR, which made it read as unfinished work.

Where each part went:

- **The two `sellers.one?` presenter gates** are deliberately **not** shipping. They rested on the premise that a multi-seller cart is never quoted in the buyer's currency. [#6501](https://github.com/antiwork/gumroad/pull/6501) made that false: multi-seller carts now present *and* charge in the buyer's currency behind the `buyer_currency_multi_seller` ramp, each seller's charge priced from its own locked quote. Landing the narrowing on top of that would have skipped the presentment support check on a ramped multi-seller cart, sending a cart the surcharge endpoint *does* quote to a lane that cannot honour the quote token — and `Order::PreparePaymentIntentService#block_unexpected_buyer_currency_quote` fails those purchases closed. So the effect would have been failed checkouts on exactly the carts #6501 enabled. Worth noting git merged the presenter **cleanly**; the conflict was semantic, not textual.
- **The frontend change** (`facdb7f4`, deriving presentment from the live quote instead of the page-load flag) is superseded at the root by [#6486](https://github.com/antiwork/gumroad/pull/6486), which makes the whole payment configuration its own Inertia prop recomputed in the same round trip that persists a cart edit, with Pay held while it is in flight. The flag is no longer frozen, so there is nothing left to work around — the right fix, rather than the client second-guessing a stale prop.
- **The screenshots** were retracted in place in #6486, since a picture of a two-seller cart on the Payment Element now invites the wrong conclusion.

Nothing is lost and no code from here is owed a PR. `Checkout::StripePaymentPresenter` is byte-identical to `main` on #6486, verified. Deleting the branch so it stops being reported as finished-but-unshipped.

PR #6494, opened 2026-07-28
Shard the Minitest CI job two ways (#5801)

ershad:
## What

Splits the `Test Minitest` CI job into a 2-shard matrix, so the Minitest suite stops being one runner working through the whole `test/` tree.

- **`bin/minitest-shard <index> <total>`** prints the files for one shard. Longest-processing-time-first packing: sort files by weight descending, hand each to whichever shard is currently lightest. With a handful of thousand-test files dominating the tree, that lands far closer to even than round-robin or halving an alphabetical list. `bin/minitest-shard --report 2` prints the split.
- **Weights are file sizes**, with a two-entry override table for files whose runtime isn't a function of their size: `stripe_payout_processor_test.rb` ×3 (most of its tests replay Stripe cassettes through the VCR bridge — the HTTP replay is the cost, not the file) and `links_controller_test.rb` ×1.5 (real Elasticsearch round-trips). Sizes are read off disk, so porting more specs re-balances the shards on the next run with nothing to maintain per file.
- **The workflow job** becomes `strategy.matrix` over `ci_node_index: [0, 1]` with `fail-fast: false`, matching `test_fast`/`test_slow`. Each shard is a separate runner with its own MySQL, Redis, Mongo, stripe-mock, Elasticsearch and MinIO, so they share no state at all. The run step logs its file manifest first — when a shard fails you want to see what it covered without recomputing the split by hand.
- **`test/lib/minitest_shard_test.rb`** asserts the properties CI depends on: every file lands in exactly one shard (at totals 1, 2, 3 and 7), no shard comes back empty, and an out-of-range index is refused.
- **`docs/minitest-migration.md`** updated — it still described the job as one runner over the whole tree.

## Why

The job was at **11m50s against its own `timeout-minutes: 15`** on the last green run on main ([30339200252](https://github.com/antiwork/gumroad/actions/runs/30339200252)). #6205 added Elasticsearch and MinIO startup on top of that, and the files still to be ported for #5801 include `sales_tax_calculator_spec` (125 examples), `dashboard_products_page_presenter_spec` (72) and `purchase_export_service_spec` (59) — roughly 256 more examples between them.

That trajectory ends in a timeout, and a timeout is a much worse signal than a failure: it reads as a broken suite, tells you nothing about which test is at fault, and costs a full re-run to learn anything. Sharding now means the next few ports land without anyone having to think about the clock. The daily updates on #5801 have flagged this as the cheapest remaining lever for several days.

Why not Knapsack Pro, which the RSpec lane uses: it splits by recorded timings from its API, which is the right tool at 18 and 50 shards. At 2 shards the gain over a static split is small, and it would put a network dependency and an API token in front of the one job whose whole point is staying lean. The script says so, for whoever needs more shards later.

Why not Rails' own `parallelize`: it would need a database per worker, and the real-Elasticsearch bridge namespaces indices per test-database — in-process workers would share one namespace and each `create_index!(force: true)` would wipe indices out from under its siblings. That's noted in `test/support/real_elasticsearch_bridge.rb` and would have to be solved first. Separate runners avoid the question entirely.

An empty shard is the failure mode worth guarding, and the reason the script exits non-zero on one: `rails test` with no file arguments runs the **entire** tree, so an empty shard would silently double the work rather than do none of it. Same for a file in no shard — it would never run, and CI would still go green. Hence the unit tests.

## Before / After

Not a user-facing change — no app code is touched. The change is the shape of the CI job:

**Before** — one runner, whole tree:

```
Test Minitest                      11m50s   2615 runs
```

**After** — two runners, measured on the same machine (`bin/rails test $(bin/minitest-shard N 2)`):

```
Test Minitest 0    1012 runs, 2674 assertions, 0 failures, 0 errors, 2 skips   201s
Test Minitest 1    1606 runs, 4457 assertions, 0 failures, 0 errors, 3 skips   162s
                   ----                                                       ----
                   2618 runs total                             wall 201s (was 363s)
```

The two shards run 2618 tests between them: 2615, the unsharded count on main, plus the 3 new tests in `minitest_shard_test.rb`. Skips split 2 + 3, which is main's 5. That arithmetic is the real check that no file went missing or got run twice.

Slowest shard carries 55% of the sequential time — an 11% drift from even, which is close enough that tuning the weights further would be fitting them to one laptop rather than to CI. Expect roughly 6–7 minutes per shard on the runner, against the same 15-minute limit.

## Test Results

```
$ bin/minitest-shard --report 2
shard 0:   7 files, weight 1126.6k
shard 1:  13 files, weight 1127.1k

$ CI=true RAILS_ENV=test bin/rails test $(bin/minitest-shard 0 2)
1012 runs, 2674 assertions, 0 failures, 0 errors, 2 skips

$ CI=true RAILS_ENV=test bin/rails test $(bin/minitest-shard 1 2)
1606 runs, 4457 assertions, 0 failures, 0 errors, 3 skips

$ CI=true RAILS_ENV=test bin/rails test test/lib/minitest_shard_test.rb
3 runs, 59 assertions, 0 failures, 0 errors, 0 skips
```

`bundle exec rubocop bin/minitest-shard test/lib/minitest_shard_test.rb`: no offenses. The workflow YAML parses, and the matrix doesn't disturb the wiring around it — `unblock_deployment_from_buildkite` still `needs: test_minitest` (a matrix satisfies `needs` once every leg passes), and the required `ci/green` status is derived from the whole workflow's conclusion rather than from job names, so renaming the job to `Test Minitest 0`/`1` changes nothing about branch protection.

## QA steps

1. `bin/minitest-shard --report 2` → two shards, no file in both, no file missing.
2. `CI=true RAILS_ENV=test bin/rails test $(bin/minitest-shard 0 2)` and the same with `1 2` → both green; run counts sum to what `bin/rails test` reports for the whole tree.
3. `CI=true RAILS_ENV=test bin/rails test test/lib/minitest_shard_test.rb` → 3, 0 failures.
4. On this PR's CI run, `Test Minitest 0` and `Test Minitest 1` both green, each well under the 15-minute limit.

---

## AI disclosure

Written with Claude Opus 5.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -313,13 +313,6 @@ jobs:
     if: ${{ github.event_name != 'pull_request_target' || github.event.pull_request.head.repo.full_name != github.repository }}
     runs-on: ubuntu-22.04
     timeout-minutes: 15
-    # Sharded because the single-runner job had grown to ~12 minutes against this
-    # 15-minute limit, and every file still to be ported from RSpec (#5801) adds
-    # to it — the next few big ones would have pushed it into timing out, which
-    # reads like a broken suite rather than a suite that outgrew its runner.
-    # Files are split by bin/minitest-shard; see that script for how, and why not
-    # Knapsack Pro. Shards are independent processes with their own database, so
-    # raising ci_node_total is the only change needed to add more.
     strategy:
       fail-fast: false
       matrix:
@@ -422,11 +415,8 @@ jobs:
       - name: Run Minitest
         run: |
           FILES=$(bin/minitest-shard "$CI_NODE_INDEX" "$CI_NODE_TOTAL")
-          # Log the manifest: when a shard fails you want to know what it covered
-          # without recomputing the split by hand.
           echo "Shard $CI_NODE_INDEX of $CI_NODE_TOTAL — $(echo "$FILES" | wc -w) files:"
           echo "$FILES" | tr ' ' '\n' | sed 's/^/  /'
-          # Unquoted on purpose: the file list has to word-split into arguments.
           bundle exec rails test $FILES
         env:
           RUBY_YJIT_ENABLE: 1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -308,11 +308,23 @@ jobs:
           fi
 
   test_minitest:
-    name: Test Minitest${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name == github.repository && ' (internal PR noop)' || '' }}
+    name: Test Minitest ${{ matrix.ci_node_index }}${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name == github.repository && ' (internal PR noop)' || '' }}
     needs: ci_gate
     if: ${{ github.event_name != 'pull_request_target' || github.event.pull_request.head.repo.full_name != github.repository }}
     runs-on: ubuntu-22.04
     timeout-minutes: 15
+    # Sharded because the single-runner job had grown to ~12 minutes against this
+    # 15-minute limit, and every file still to be ported from RSpec (#5801) adds
+    # to it — the next few big ones would have pushed it into timing out, which
+    # reads like a broken suite rather than a suite that outgrew its runner.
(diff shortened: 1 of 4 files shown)


slavingia:
No comments needed imo

PR #6497, opened 2026-07-28
Create the client-confirm intent in the currency the Payment Element was mounted in

gumclaw:
## What

The client-confirm checkout lane mounts a Stripe Payment Element in the browser, the browser mints a ConfirmationToken against it, and `Order::PreparePaymentIntentService` then creates the unconfirmed PaymentIntent the browser confirms that token against. The Element's currency was decided at page render; the intent's currency was recomputed independently at pay time. This makes the browser's reported mount currency the authority for the methods whose token simply inherits it — card, Link and the wallets — so the intent is always one the token can confirm.

A method that can only charge in one currency (iDEAL and Bancontact in euros, UPI in rupees, Pix in reais) still decides for itself. A reported non-USD currency the server cannot legitimately build an intent in fails the order closed rather than creating one nobody can confirm.

The intent's `payment_method_types` now follows the intent's currency in both directions. `Checkout::PaymentMethodResolver` builds its list from the cart's pricing, so on a EUR-priced cart it still offers `ideal`; listing a euro-only method on a dollar intent makes Stripe reject the intent *create* outright, which would fail the whole cart including card buyers. The existing strip only covered the non-USD-intent direction.

## Why

Stripe rejects confirming a ConfirmationToken minted on an Element of one currency against an intent of another — synchronously, in the browser, with no `payment_failed` webhook:

```
The provided currency (eur) does not match the expected currency (usd).
```

The buyer hits a dead end at the final step: no charge, no recorded error, and the purchase sits `in_progress` until `FailAbandonedPurchaseWorker` sweeps it — indistinguishable from someone who closed the tab.

Resolved against production, the affected population is **57 orders over four days** (24–28 July), one event each. 54 had a EUR intent against a USD-mounted Element, 2 an INR intent against a USD-mounted Element, and 1 the reverse. All 57 carts were priced in a forced currency; 56 of 57 intents were created in the product's own currency, so the forced-currency machinery was doing its job.

The cause is render-time versus pay-time divergence. `elements_options.currency` is computed once when the checkout page renders and is never refreshed — the page's two Inertia partial reloads request `["cart", "flash"]` and `["recommended_products"]`, never `checkout_payment`. Everything feeding the decision can move in between: the per-seller local-method launch flags, the merchant account's settlement state, and the cart itself, which the page mutates client-side. A cart that becomes uniformly EUR-priced mid-session leaves the Element on the dollars it mounted with.

The client already sent the right information. The server was using it only to fail a stale token closed, never to decide the currency.

Tracked on the internal issue for this bug. Surfaced by #6223, which shipped the `confirm_error` reporting endpoint (Sentry `GUMROAD-12F`).

### Why the intent follows the Element rather than the reverse

Two directions were possible: keep failing closed and make the retry reload the payment config, or let the intent follow the Element. Failing closed fired on 54 of the 57 orders, and the retry path cancels, re-fetches surcharges and re-runs the display gates **without** reloading `checkout_payment` — the Element remounts in the same currency, prepare recomputes the same intent currency, the guard fires again. That converts a one-shot dead end into a loop. Following the Element means those 54 succeed on the first attempt.

The browser is trusted about *which* currency its Element used and nothing else. Whether that currency is chargeable stays a server decision on the same conditions the page renders on — the seller is enabled for buyer-currency charging, some payment method forces that currency, and every line in the charge is priced in it. The amount comes from `Charge::MethodForcedPresentment` (listed prices, or a Stripe FX quote), never from the client. The parameter is not in `permitted_order_params`, is merged only in `#prepare`, and is never persisted.

## Test results

`bundle exec rspec spec/services/order/prepare_payment_intent_service_spec.rb` — the six new examples: **6 examples, 0 failures**.

Red-first proof: with only `app/services/order/prepare_payment_intent_service.rb` reverted to `main` and the specs left in place, **5 of the 6 fail**. The sixth (an iDEAL token whose Element reported dollars) cannot be made to bite — the old code did not consult the report for a currency-forcing method either — and carries a comment saying what it does pin: the branch order of the new decision, which an implementation letting the report override the method would fail.

CI is green on the production-fix commit (`67102fa`): 73 of 73 jobs passing.

⚠️ **Note on whole-file local runs.** This spec file activates Flipper flags per example, and on a machine sharing one test database with other work the flag state bleeds across processes under random ordering: unmodified `main` in the same worktree produced 17 failures while sibling suites were running, and 0–2 when quiet. Whole-file counts from that environment are not interpretable in either direction — the targeted runs above and CI are the authority. Please do not read a red local full-file run as this branch's doing without a same-worktree baseline.

Captured on the preview app at `x-revision d367f99a452e` (this PR's head before the evidence
commit), four legs varying **only the currency the browser's Payment Element was mounted in**.
Everything else — seller, product, cart, buyer country, card — is held identical. Each leg
records the checkout render, the wire calls, and the PaymentIntent Stripe actually returned.

**Click path** (preview: https://fix-intent-follows-element-curre.apps.staging.gumroad.org)

1. Open `/checkout?product=<the EUR-priced QA product>&quantity=1` as a guest from a Netherlands egress.
2. Fill email + full name, complete the Link fields inside the Payment Element (mobile number, email, full name — any one empty blocks the confirm client-side with no error).
3. Fill card `4242 4242 4242 4242`, any future expiry, any CVC.
4. Press **Pay**, and read the resulting PaymentIntent's `amount`/`currency`/`status`.

**1. Before — dollar-mounted Element, euro intent (the bug).**
Reproduced on this branch by stripping `payment_element_mount_currency` from the
`/orders/prepare` request, which drops the server back to the pre-fix inference: the EUR-priced
cart makes it build a **euro** intent while the browser's Element (mounted in dollars) minted a
**dollar** ConfirmationToken. Stripe returned the production error verbatim, synchronously in
the browser, and **no purchase was created**.

`POST /v1/payment_intents/pi_2Ty9h79e1RjUNIyY0P6v97oH/confirm -> 400`
`The provided currency (eur) does not match the expected currency (usd).`
Intent: `eur / 1815 / requires_payment_method`, `payment_method_types [card, link, ideal, bancontact]`.

before checkout
before confirm error

**2. After — dollar-mounted Element, dollar intent (the dominant shape, 54 of 57).**
Same cart, report sent. The Element mounted `usd/1500`
(`elements/sessions deferred_intent[currency]=usd`), the intent was created in dollars, and the
confirm succeeded on the first attempt:

`pi_2Ty9g89e1RjUNIyY1iuUY4GG` — **usd / 2063 / succeeded**, `payment_method_types [card, link, cashapp]`.

Note the method list: `ideal` and `bancontact` are **stripped** because they cannot charge a
dollar intent, and Stripe accepted the intent create. This is the residual-drift case flagged as
P3-1 in the review and the leg the body asked for a second pair of eyes on — the intent's list
is a strict subset of what the Element rendered, `card` is present, and the confirm goes through.

after checkout
after success

**3. The reverse — euro-mounted Element, euro intent (1 of 57).**
Element mounted `eur/1500`; intent created in euros and confirmed:
`pi_2Ty9f39e1RjUNIyY1iqX71EJ` — **eur / 1815 / succeeded**, `payment_method_types [card, link, ideal, bancontact]`
(no strip here, because the euro-only methods can charge a euro intent).

The post-purchase page this leg landed on was byte-identical to leg 2's — that page shows no
currency — so only the checkout render is committed here, and the euro payment itself is pinned by
the PaymentIntent id above.

eur checkout

**4. iDEAL unchanged — the method still decides for itself.**
`pi_2Ty9j79e1RjUNIyY0bPr299X` — **eur / 1815 / requires_action** (`redirect_to_url`),
`payment_method_types [card, link, ideal, bancontact]`. Authorized on Stripe's hosted test page,
came back to a successful purchase. A currency-forcing method never consults the browser's
report, which is the branch order the sixth spec pins.

ideal checkout
ideal stripe authorize
ideal success

Nothing on the branch was changed to produce any of this. The Netherlands egress was reached
through a local proxy onto the preview pod (the hosted load balancer overwrites
`X-Forwarded-For`; the pod trusts it), so eligibility, the launch flags, the real Stripe test FX
quote and the signed quote token all ran unmodified on the deployed head. The one deliberate
intervention is named above and is confined to leg 1: `page.route` stripping a single request
parameter, which is how the pre-fix server behaviour is reproduced without reverting code.

**Retracting the body's open question.** The section above previously said the step-1 confirm
"no test-mode stub or cassette can prove it. A real completed payment on that lane is the
evidence." That evidence is leg 2: a real completed payment on the honored-dollar lane, with the
subset method list, at Stripe. Both items listed as "verified live-only, and on the QA list" in
the review section are now covered.

## QA steps

Preview app: https://fix-intent-follows-element-curre.apps.staging.gumroad.org (verified live serving `d367f99a452e`). The divergence is a timing race, so the reliable way to reproduce it is to make the two sides disagree deliberately.

1. **The dominant shape (54 of 57).** Seller with buyer-currency charging on and a EUR-priced product. Load checkout with a cart that mounts the Element in dollars (a mixed-currency cart), then remove the non-EUR line so the cart becomes uniformly EUR-priced — without reloading the page. Pay by card. **Before:** the confirm fails in the browser with the currency-mismatch error and the purchase never completes. **After:** the payment goes through on a dollar intent.
2. **The reverse (1 of 57).** Same seller, EUR-priced cart, Element mounted in euros; pay by card. Expect a completed EUR payment with presentment rows.
3. **iDEAL unchanged.** EUR-priced product, pay by iDEAL. Expect a euro intent listing `ideal`, exactly as today.
4. **Dollar lane unchanged.** USD-priced product, pay by card. Expect the canonical dollar intent, no presentment rows.

What I would most like a second pair of eyes on: **step 1's confirm actually succeeding at Stripe.** The intent's method list there is a strict subset of what the Element rendered (the euro-only methods are stripped). Stripe rejects a `payment_method_types`-scoped ConfirmationToken only when the *confirmed* method is missing from the intent, and card is always present — that is the fix's core success criterion, and it is now evidenced: leg 2 of the Screenshots section above is a real completed payment on exactly that lane (`pi_2Ty9g89e1RjUNIyY1iuUY4GG`, usd/2063, succeeded, subset method list).

## Adversarial review

Local fable-5 review run on the branch before this PR was opened. Verdict was CHANGES-REQUIRED with no correctness or security findings — the production hunks survived every angle (trust boundary, old-helper consumers, the method-list strip, Pix, display/amount consistency, and the retry-loop question). Both P2s were spec-validity problems, now fixed in `332d59e`:

| Finding | Disposition |
|---|---|
| P2-1 — two new specs pass with the fix reverted (test-mode bypasses the launch flags, so the old inference already concluded euros) | Fixed. The euro-report example now stages capability drift so it genuinely fails on `main`; the iDEAL one is documented as pinning branch order. |
| P2-2 — the flags-off lane is not byte-for-byte for a euro-mounted token, and the changed case was untested | Fixed. New example pins the fail-closed result; the "byte-for-byte" claim now scoped to no-param and dollar-report requests. |
| P3-1 — residual method-list drift (intent's list ⊂ Element's) on the honored-dollar lane | Documented in code where the strip lives, and it is the first QA item above. |
| TRIM — pointless memoization and single-caller indirection in `intent_forced_currency`; a no-op flag activation in one spec | Both applied. |

Both of these are now verified live on the preview and captured under **## Screenshots (hosted preview app)** above: the subset-list confirm at step 1 (leg 2) and a real end-to-end euro confirm on the honored-report lane (leg 3).

## Not for auto-merge

This is the checkout confirm path. Handing it to a human for QA and the merge click rather than arming auto-merge.

---

## AI disclosure

Written by Gumclaw using `claude-opus-5`. Root cause was traced from Sentry issue `GUMROAD-12F` by resolving all 2,593 events to their orders and reading the real PaymentIntents in production, which corrected an earlier analysis on the internal tracking issue that had the direction and the population size wrong. Sahil chose between the two candidate fixes ("let the intent follow the Element" over "keep failing closed and fix the retry"). The adversarial pre-merge review was run by a separate `claude-opus-5` session against the branch; its findings and their disposition are in the table above.


gumclaw:
@rmarescu — this is the fix for the currency mismatch you and I went back and forth on in gumroad-private#1382, and per Sahil's standing call that multi-currency is human-built-and-merged, the merge is yours. I am not arming auto-merge: this is the checkout confirm path.

Where it landed relative to your earlier note. You were right that #6251 was not the culprit and that the buyer-currency USD card lane was not the lane involved — my original write-up had both the direction and the size wrong. Resolving all 2,593 `GUMROAD-12F` events against production gave 57 affected orders, not 1,711 events (89% of the issue is ordinary card declines sharing one fingerprint). All 57 carts are priced in a forced currency, 56 of 57 intents were created in that currency, and the dominant direction is the opposite of what I first wrote: 54 are a EUR intent against a USD-mounted Element. The mechanism is render-time versus pay-time divergence — `elements_options.currency` is computed once at render and the partial reloads never request `checkout_payment`, while `PreparePaymentIntentService` recomputes independently at pay time from inputs that can move in between.

Sahil chose to let the intent follow the Element rather than keep failing closed, which is what this ships.

**Preview app:** https://fix-intent-follows-element-curre.apps.staging.gumroad.org — verified live serving `d367f99a452e`, the current head.

CI is green at the head (73 of 73 jobs). A fable-5 adversarial review ran on the branch before this PR opened; it found no correctness or security issues, and both of its spec-validity findings are fixed in `332d59e`. The findings table is in the body.

**The one thing I would most like you to verify before merging** is QA step 1 — the confirm actually succeeding at Stripe on the honored-dollar lane. On that lane the intent's `payment_method_types` is a strict subset of what the Element rendered, because the euro-only methods have to be stripped or Stripe rejects the intent create outright. My reading is that Stripe only rejects a scoped ConfirmationToken when the *confirmed* method is missing, and card is always present — but no test-mode stub or cassette can prove that, and it is the fix's core success criterion. A single real completed payment on that lane settles it.

Second thing worth your eye: whether honoring a client-reported currency sits right with you as a contract. The browser decides *which* currency only; whether it is chargeable stays server-side on the same conditions the page renders on, and the amount still comes from `Charge::MethodForcedPresentment`. But you own this surface, so if you would rather the recovery came from making the payment config live instead (which is what #6486 does), say so and I will rework it that way.

If the analysis reads wrong to you anywhere, tell me and I will re-investigate rather than defend it.

gumclaw:
Body-only edit just now — no commits, head unchanged, CI untouched, review state and assignee untouched.

This is a public repository and the description cited internal tracker issues by number. Those identifiers are gone; the substance they supported is unchanged, just phrased without them ("tracked internally" / "the internal tracking issue"). The tracker still carries the link to this PR from its side, which is where that edge belongs.

Worth noting why this needed a grep rather than an API check: `closingIssuesReferences` reads empty on this PR, because none of the references used a closing keyword. A live `Fixes <private>#N` in a public body would force-close the private issue the instant this merges — no comment, assignee never notified, fix possibly still undeployed. That did not happen here, but the field would have reported this PR as clean either way. The raw body grep is the load-bearing check.

Found by a sweep of my open PRs. Review ask is unchanged.

rmarescu (approved):
> whether honoring a client-reported currency sits right with you as a contract [...] if you would rather the recovery came from making the payment config live instead (which is what #6486 does), say so and I will rework it that way.

@gumclaw keep this one. The browser only picks which of two currencies the server would have chosen itself, the amount still comes from `Charge::MethodForcedPresentment`, and the report cannot name a currency the element did not mount in, because the reported value and the element's own mount both read `elements_options.currency`. #6486 recomputes the payment configuration when the cart changes, so it covers cart edits but not the per-seller launch flags or the account's settlement state, which are the other two inputs your root cause names.

`qa-media/pr-6497-eur-element-purchase-success.png` is byte-identical to `pr-6497-after-usd-element-purchase-success.png`, and the euro payment is already pinned by `pi_2Ty9f3...`. Does keeping the second copy add anything?

gumclaw:
> keep this one [...] Does keeping the second copy add anything?

Nothing — dropped it in f76dc1f. The two files were byte-identical because the post-purchase page renders no currency at all, so the euro leg's success shot was only ever proof that *a* purchase completed, which the PaymentIntent id (`pi_2Ty9f39e1RjUNIyY1iqX71EJ`, eur/1815/succeeded) already carries. The euro leg keeps its checkout render, which is the frame that actually shows the euro mount. Body updated with a line saying why there is no success shot on that leg, so the asymmetry doesn't read as a gap to the next reader.

On the contract: agreed, and thanks for spelling out the third input. That is the part I had least confidence in — #6486 makes the config live with respect to the cart, which is the input the page mutates most visibly, but the launch flags and the account's settlement state move server-side with nothing on the page watching them, so a live-config approach would still need this guard for those two. Leaving the trust boundary as is.

Leaving the merge click with you — this is the checkout confirm path and you own it. Only change since your approval is the deleted screenshot.

gumclaw:
The `Test Minitest` job on this PR's head had been **cancelled** (superseded, not failed), so that suite carried no verdict. I re-ran it and it **passed**: https://github.com/antiwork/gumroad/actions/runs/30401601156/job/90425107527

Every check at the current head is now green — 75 successful, 9 skipped, nothing failing or pending. No code changed; the head SHA is unchanged, so nothing else needs re-reviewing. Merge remains yours.

PR #1471, opened 2025-10-02
fix: replaced inline styles of `_user.html.erb` with Tailwind CSS

VarunArora14:
Part of fix: https://github.com/antiwork/gumroad/issues/1055

Migrate inline CSS styles to Tailwind CSS classes in the admin user view file `app/views/admin/users/_user.html.erb`

## Changes
- Replace inline flex styles with `flex items-center gap-4` utility classes
- Replace inline grid styles with `grid gap-2` utility classes
- Remove dependency on CSS variables `--spacer-4` and `--spacer-2`
- Maintain identical visual layout and spacing

### Technical Mapping:

| Before (CSS Variables) | After (Tailwind) | Purpose |
|------------------------|------------------|---------|
| `display: flex` | `flex` | Flexbox display |
| `align-items: center` | `items-center` | Vertical alignment |
| `gap: var(--spacer-4)` | `gap-4` | 1rem spacing |
| `display: grid` | `grid` | Grid display |
| `gap: var(--spacer-2)` | `gap-2` | 0.5rem spacing |

### Before (light mode)

image

### Before (dark mode)

image

### After (light mode)

image

### After (dark mode)

image

***

### Before (dark mode)

image

### Before (light mode)

image

### After (dark mode)

image

### After (light mode)

image

AI Disclosure:
No AI Used

PR #6505, opened 2026-07-28
Label Gumroad-held balances in USD, not the buyer's presentment currency

gumclaw:
## What

`BalanceTransaction::Amount.create_holding_amount_for_seller` was labelling **Gumroad-held** seller balances with the currency the buyer was charged in. It now takes the merchant account into account and keeps Gumroad-held balances denominated in USD.

The method picks a currency from the flow of funds, and its middle branch used `settled_amount.currency`:

```ruby
elsif canonical_issued_amount && flow_of_funds.settled_amount
  new(currency: flow_of_funds.settled_amount.currency, ...)
```

The fix threads `merchant_account:` into the method and excludes Gumroad-held accounts from that branch, so they fall through to the `canonical_issued_amount` arm. Connected accounts keep their existing behaviour. Three call sites now pass the account: the charge leg and the refund leg in `Purchase`, and `Purchase::DisputeWinCredits`.

### What the holding fields mean, which is the whole of it

`holding_currency` / `holding_amount_*` mean two different things depending on who holds the funds, and the old code applied one meaning to both.

On a seller's **own connected account**, the holding amount describes that account's balance. It genuinely is denominated in whatever Stripe settled, which is that account's own currency, and a non-USD value there is correct.

On **Gumroad-held** funds it is something else: Gumroad's canonical record of what it owes the seller. A liability, always denominated in USD, because USD is the currency every Gumroad-held payout is computed and wired in. It is deliberately not a record of which currency Stripe is physically sitting on. Stripe's platform account does carry real foreign-currency balances — a EUR charge can settle and stay in EUR — but that is an account-level treasury position spanning every seller at once, and nothing about paying one seller follows from it.

So `settled_amount` is the wrong source for a Gumroad-held balance even where it is factually what Stripe settled. On a buyer-presentment charge it is the currency the *buyer* was charged in, which says nothing about the size of Gumroad's liability to the seller.

That definition is the change from this pull request's earlier framing. An audit of the affected cohort's Stripe records found these charges settling in EUR and remaining in the platform account's EUR balance, so an earlier claim in this body that the money always physically arrives in USD was not true. The code was right; the reason given for it was not. Recording Stripe's asset currency in these fields instead would be a separate multi-currency treasury and payout redesign, and it is out of scope here.

### The defect blocks money, in two different ways

Both payout processors require Gumroad-held balances to be USD, and they fail differently. This matters because only one of the two is visible.

**Stripe — loud.** `StripePayoutProcessor#prepare_payment_and_set_amount` rejects a Gumroad-held balance whose `holding_currency` is not USD:

```ruby
mismatched_gumroad_balances = balances_held_by_gumroad.reject { |b| b.holding_currency == Currency::USD }
if mismatched_stripe_balances.any? || mismatched_gumroad_balances.any?
  payment.mark_failed!(Payment::FailureReason::CURRENCY_MISMATCH)
```

Its `is_balance_payable` returns `true` for *every* Gumroad-held balance regardless of currency, so the bad row sails past the per-balance filter, gets pulled into the payout, and fails the seller's **entire** payment — correctly-labelled USD balances included. This leaves a `failed` payment with `currency_mismatch` on it, which is how the problem was noticed at all.

**PayPal — silent, and more common.** `PaypalPayoutProcessor.is_balance_payable` is currency-aware:

```ruby
def self.is_balance_payable(balance)
  balance.merchant_account.holder_of_funds == HolderOfFunds::GUMROAD && balance.holding_currency == Currency::USD
end
```

It drops the bad row *before* a payment object is built. No failed payment, no failure reason, nothing in Sentry. The seller is quietly short-paid and the only trace is a balance that stays `unpaid` while newer ones get paid.

Measured on production, the PayPal lane is the majority of affected sellers and carries zero error signal: no failed payments at all, one seller already short-paid by a completed payout that excluded the money, and a simulation of the next scheduled run short-paying more of them with nothing recorded. Had the affected set happened to be PayPal-only, this would very likely still be undetected. Counts, seller identifiers and the Stripe object audit are on the internal tracking issue rather than here.

## ⚠️ This fix alone unblocks nobody

Stated plainly because it would be easy to read this as the payout fix and stop. It is not.

The existing mislabelled rows keep their non-USD `holding_currency`. Stripe still pulls them into payouts and still fails; PayPal still drops them. **Every affected seller stays blocked until a separate backfill restamps those rows.** This change only stops new bad balances being minted; the backfill is the part that gets anyone paid, and it is deliberately not in this diff so the forward fix can be reviewed on its own.

One note for that backfill. Balances are keyed by `holding_currency`, so between this deploying and the backfill running, a refund of a pre-fix mislabelled purchase books its negative leg into a **USD** balance while the original positive leg stays in the non-USD one. That is benign for the ledger — `holding_amount_net_cents` is the canonical USD magnitude in both the old and new branches, so no value is stranded, only the label and `gross_cents` differ — but the backfill must not assume a refund leg shares its original's currency. Some will already be USD.

## Before / After

Not a rendered surface — this writes the money ledger, so the before/after is the **values the code produces** rather than pixels. Both columns below come from one run against a real database: the "before" column reimplements main's branch inline, so the two columns cannot differ by fixture or clock, only by the change under review.

Charge shape: product listed at US$100, buyer charged EUR 90 (buyer-currency presentment).

| Merchant account | `holder_of_funds` | Before (on main) | After (this PR) | Payout outcome |
|---|---|---|---|---|
| Gumroad-held platform account | `gumroad` | `eur` 9000 | **`usd` 10000** | now payable on Stripe and PayPal |
| Seller's own connected account | `stripe` | `eur` 9000 | `eur` 9000 | unchanged — describes that account's own balance |

Only the Gumroad-held row moves. A seller's own connected account still records the currency Stripe settled, because that balance really is denominated in euros.

And the payout guards themselves, driven for real rather than asserted by implication:

```
Stripe   is_balance_payable(USD-labelled #58)  = true
Stripe   is_balance_payable(EUR-labelled #59)  = true   <- admitted anyway

Stripe   prepare_payment_and_set_amount([both balances]):
           payment.state          = failed
           payment.failure_reason = currency_mismatch
           error                  = Cannot process payout: balances [59] have holding_currency
                                    that does not match the payout currency.

PayPal   is_balance_payable(USD-labelled #58)  = true
PayPal   is_balance_payable(EUR-labelled #59)  = false  <- dropped, no payment, no error
```

Stripe's `is_balance_payable` returns true for every Gumroad-held balance regardless of currency, so the mislabelled row is pulled into the payout and the rejection lands one step later, failing the whole payment — the correctly-labelled US$15.42 row is lost along with the bad one. PayPal's rejects it at the filter, which produces no payment and no error to notice.

The spec run, the mutation proof, and that table, all in one render:

PR 6505 specs, mutation proof, and before/after values

## Test Results

Two files.

**`spec/models/balance_transaction_spec.rb`** — the branch that picks the currency, four contexts:

- **Gumroad-held** (userless platform account, `holder_of_funds == GUMROAD`): asserts `usd` / 10000, and drives `StripePayoutProcessor.prepare_payment_and_set_amount` directly rather than asserting the currency twice — including the rejection half, so the spec pins the failure mode and not only the fix.
- **Connected** (`CREATOR`) and **Stripe-held** (`STRIPE`): assert the settled-currency behaviour is preserved, each self-asserting its `holder_of_funds` so the contexts cannot silently drift into testing the same thing.
- **No merchant account passed**: pins the nil-default, which the two `onetime` backfill services rely on.

**`spec/models/purchase/gumroad_held_presentment_balances_spec.rb`** (new) — the same property through the real booking path, because a correct-looking `Amount` object flowing into an unpayable balance is exactly what the incident was. Drives `increment_sellers_balance!` and `decrement_balance_for_refund_or_chargeback!` on a EUR-presentment purchase against the Gumroad-held account:

- **A EUR purchase with no Stripe FX quote** creates a `BalanceTransaction` and a `Balance` whose currency, gross and net are all USD, and asserts explicitly that neither is the buyer's EUR figure.
- **Both processors pay it**: `prepare_payment_and_set_amount` returns no errors with the amount set, and `PaypalPayoutProcessor.is_balance_payable` returns true. A companion example relabels the row EUR and asserts both processors reject it, so the passing assertions are not tautologies.
- **Negative legs**: a refund and a chargeback each book their negative leg in USD, with the balance itself USD — which matters because the label keys the balance, so a non-USD negative leg would open a second parallel balance.
- **A quote-backed charge Stripe converted to USD**: that lane's currency was already USD, so the assertion is on the gross, which moves from Stripe's settled figure to the canonical issued amount. The two are deliberately different amounts in the fixture so the test can tell them apart, and the net is asserted unchanged.

The quote-less shape is the one to look at first: forced-currency local methods take no FX quote at all, `presentment_canonical_issued_amount` returns non-nil on the presence of a presentment row alone, and a fix keyed on the quote rather than on who holds the funds would have missed every such row.

**Load-bearing by mutation.** Removing `&& !gumroad_held?(merchant_account)` sends the Gumroad-held contexts into the `settled_amount` arm and reddens the currency assertions *and* the payout-guard assertions — the specs pin the consequence, not merely the label. The mutation was reverted and the working tree verified clean before committing.

Local run of both files, against a clean test database:

```
spec/models/balance_transaction_spec.rb (create_holding_amount_for_seller)
  9 examples, 0 failures

spec/models/purchase/gumroad_held_presentment_balances_spec.rb
  8 examples, 0 failures
```

`rubocop` clean on all changed files.

## QA steps

Both scenarios were walked on the preview app at `fix-gumroad-held-balance-holding.apps.staging.gumroad.org` (`x-revision b8676dd45f9e`), with a seller who has **no** connected Stripe account so purchases land on Gumroad's platform account, and the product listed in **EUR** so the charge takes no FX quote. The card checkout was completed in a real browser; the ledger and both payout processors were then read from the preview's own console.

The setup produced exactly the shape the incident needs: the element mounted `{"currency": "eur", "presentment_amount_cents": 2000}` with `["card", "link", "ideal", "bancontact"]`, the surcharge endpoint returned `"buyer_currency_quote": null`, and Stripe's own charge, balance transaction and intent all came back `currency=eur amount=2420` with `exchange_rate=nil` and no destination — no conversion anywhere.

**1. A quote-less EUR purchase books a USD balance and pays out.** The balance transaction and the balance are USD, and the holding gross is the canonical **2758**, not the EUR **2420** Stripe settled. `PaypalPayoutProcessor.is_balance_payable` returns true, and `StripePayoutProcessor.prepare_payment_and_set_amount` returns no errors and reaches the internal transfer with `currency: "usd", amount_cents: 1905` (the transfer is stubbed — everything past the currency guard moves real money).

**2. A refund of that purchase stays on the same balance.** The negative leg is `usd / -2758 / -1905` and the seller still has exactly **one** balance row afterwards. No second EUR-labelled balance was opened.

The same run also drives main's branch inline against the identical flow of funds, so the two columns cannot differ by fixture or clock:

| | holding | payout outcome |
|---|---|---|
| Before (main) | `eur` / 2420 / 1905 | PayPal drops the row silently; Stripe fails the whole payment with `currency_mismatch` |
| After (this PR) | **`usd` / 2758 / 1905** | PayPal true; Stripe no errors, amount set |

Stripe's per-balance filter returns `true` for **both** labels, which is why the bad row reached the aggregate guard in production rather than being skipped.

PR 6505 preview-app QA: ledger, both payout processors, and before/after on one live purchase

The EUR checkout the payment was made through, for the record that the quote-less forced-currency surface is what was exercised:

The EUR-mounted checkout with iDEAL and Bancontact offered and no FX quote

A reviewer can also check the diff directly:

1. Read the branch in `create_holding_amount_for_seller` and confirm the connected-account path is untouched (`gumroad_held?` returns false for `CREATOR` and `STRIPE`).
2. `bundle exec rspec spec/models/balance_transaction_spec.rb spec/models/purchase/gumroad_held_presentment_balances_spec.rb`.
3. Optionally reproduce the mutation above and confirm the Gumroad-held contexts redden.

After deploy, the check that matters is on the backfill rather than this: no new Gumroad-held balance with a non-USD `holding_currency` and a `created_at` after the release.

## Adversarial review

Ran a local adversarial review over the diff. Verdict: **READY-TO-MERGE**, with no correctness findings — it could not construct a purchase or account shape that mislabels money in either direction.

What it cleared, and how:

- **Connected accounts preserved.** `CREATOR` and `STRIPE` holders still take the `settled_amount` arm; no path routes them to the USD arm.
- **The nil default is safe at every caller.** All five callers enumerated. The three model sites now pass the account. The two `onetime` services do not, but both build their flow of funds via `flow_of_funds_for`, which for a Gumroad holder returns `build_simple_flow_of_funds(Currency::USD, …)` — the settled amount is already USD, so the unchanged branch yields USD anyway. No unpatched Gumroad-held seller-balance path exists.
- **Right account object at each site.** At all three, the value passed here is the same `merchant_account` handed to `BalanceTransaction.create!` a few lines below, so the guard always reads the account the balance is actually booked against.
- **Affiliate legs have no exposure.** `create_holding_amount_for_affiliate` picks `canonical_issued_amount&.currency || flow_of_funds.gumroad_amount.currency` and never touches `settled_amount`, so it cannot stamp a foreign currency on a Gumroad-held affiliate balance.
- **The first branch cannot bypass the fix.** A Gumroad-held platform charge has no `destination` and no Connect application fee, so `merchant_account_gross_amount` is nil and the guarded branch is always the one reached.
- **`HolderOfFunds::GUMROAD` is exhaustive.** `merchant_account.holder_of_funds` always returns a value (defaulting to `GUMROAD` for removed processors) and never nil for a real account, so the comparison cannot silently fall through.

Two non-code findings, both acted on: a spec named "produces a balance the Stripe payout processor will not reject" only re-asserted the currency, so it now drives the guard for real; and the zero-sellers-unblocked consequence has its own section above rather than a footnote.

## Deliberately not in this PR

- **The backfill of the existing mislabelled rows.** Separate change against a frozen id list, gated on this deploying first.
- **Legacy dispute wins.** A dispute debited before `presentment_dispute_debited_gross_cents` existed can win after the backfill has run and book the buyer's currency, minting a fresh non-USD Gumroad-held balance. Tracked separately, with the still-open exposed set to enumerate and a post-release monitor for any Gumroad-held balance carrying a non-USD `holding_currency`. It does not block this change, which only stops new bad balances being minted from new purchases.
- **Re-ramping `buyer_currency_charging`.** It stays at 0% until this deploys and the backfill has run. Note it is the upstream gate for all four forced-currency lanes, so iDEAL, Bancontact, UPI and Pix are dark meanwhile even though their own flags read 100%; they resume automatically when it returns, with no per-lane flag writes.
- **Any change to what the payout guards accept.** Both correctly refuse a non-USD Gumroad-held balance; the defect was upstream in the labelling.

---

## AI disclosure

Written by `claude-opus-5` (Anthropic) running as the Gumclaw agent.

Instructions given: advance the multi-currency work so a human opening it finds it as far along as it can be without them. The agent found this fix already built and pushed to a branch with green CI but no pull request, verified the diagnosis against the payout processors' own guards, ran an adversarial review over the diff, and opened this. A reviewer then questioned the physical-USD justification in this body; auditing the Stripe balance transactions behind the affected cohort showed those charges settling in EUR and staying there, so the reviewer settled the definition as Gumroad's canonical USD seller liability and this body, the code comments, the model documentation and the test descriptions were rewritten around it. The same review round identified `PaypalPayoutProcessor.is_balance_payable` as a second, silent consumer, which the tests now cover on both processors, and added the integration-level specs through the real purchase booking path. Production figures come from audited read-only console queries.

---

## Merge note (attribution)

Two commits on this branch trail `Co-authored-by: Razvan <razvan@gumroad.com>`. That
address does not exist — Razvan's is `razvan@marescu.net`. History is not being
rewritten for it (no force pushes on pushed branches), so **whoever squashes this must
correct the trailer in the squash commit message to
`Co-authored-by: Razvan Marescu <razvan@marescu.net>`** so the merged commit on `main`
carries the right one.


nyomanjyotisa (commented):
The holder-based split is correct across charge, refund, and dispute-win paths, including quote-backed and quote-less presentment.

Independent exercise passed: 9 focused specs plus 4 end-to-end probes proving the fixed USD balance passes payout preparation while the old EUR label fails with `currency_mismatch`. Removing the new guard correctly fails the regression tests.

One non-blocking test correction: the payout example calls `is_balance_payable`, which intentionally accepts both USD and EUR Gumroad-held balances; it does not exercise `prepare_payment_and_set_amount`, where rejection occurs. Please either test that method or rename the example.

This remains forward-only: the 119 affected sellers still require the separate backfill before the currency ramp is restored.

gumclaw:
@nyomanjyotisa Fixed in `e1c88fdc5` — your correction was right, and the example was worse than mislabelled: it could not fail.

`is_balance_payable` returns true for **every** Gumroad-held balance regardless of currency, which is exactly why the incident happened — the mislabelled row sailed past it and got pulled into the payout anyway. So asserting it returned `true` proved nothing; the old example would have passed just as happily with the bug still in place.

The example now drives `StripePayoutProcessor.prepare_payment_and_set_amount`, which is where the currency comparison actually lives, and asserts the failure direction: a Gumroad-held balance labelled in the buyer's EUR comes back with the "holding_currency that does not match the payout currency" error and `Payment::FailureReason::CURRENCY_MISMATCH` on the payment. That is the seller-visible failure this PR exists to prevent.

**Mutation-checked, which is the part that matters given the above:** reverting the three app changes and keeping the new spec reddens 4 of the 5 examples in this context, including this one. With the fix in place it is 5 of 5 green. The old assertion survived that same revert, so this is now load-bearing where it previously was not.

Two notes on how I got there, since both shaped the final shape:
- I first tried asserting the **success** leg (correctly-labelled balance → payment survives). That runs past the guard into `Stripe::Transfer.create` and the destination-payment bookkeeping, which is a different contract entirely and needed increasingly fake doubles to get through. The failure direction is the one that pins this bug, so the success direction stays with the sibling example that asserts `amount.currency` directly.
- Swapped `:merchant_account_stripe` for the plain `:merchant_account` factory — the former provisions and verifies a real Connect account against the live Stripe test API, which was independently failing on my machine and is not something this test needs.

Your forward-only point stands unchanged: the 119 affected sellers still need the separate backfill before the currency ramp is restored. That is not in this PR.

*Written by Claude Opus 4.5 (Hermes agent, PR momentum sweep).*

gumclaw:
@rmarescu Requesting your review here — this is the last thing standing between the multi-currency migration and Sahil's Aug 1 deadline, and it is costing money today.

`buyer_currency_charging` has been at 0% since 13:37Z on Sahil's order, so **all four forced-currency lanes (iDEAL, Bancontact, UPI, Pix) are dark** — their own flags still read 100%, but `PaymentMethodResolver#forced_currency_methods` returns `[]` without the upstream gate. They resume automatically the moment this deploys and the flag returns to 100%. Meanwhile 181 balances / 119 sellers / $4,275.26 stay stuck, and sellers are hard-failing payouts a third time and threatening to leave.

CI is fully green, `mergeStateStatus: CLEAN`, and nyomanjyotisa's correction is addressed in `e1c88fdc5`. **The one thing needed from you: review + approve so this can merge today.** I'll handle the deploy verification and the ramp back to 100% with the live-parity probe.

gumclaw:
On hold — do not merge, and the production backfill is not running.

Razvan raised an accounting question on gumroad-private#1471 and the audit came back against this PR's stated justification. Stripe settled the affected purchase cohort in **EUR, not USD**: a 50-charge sample tied 50/50 to EUR balance transactions, and the whole cohort is quote-less direct-listed (0 of 258 `charge_presentments` rows carry a `stripe_fx_quote_id`), which is the lane where Stripe performs no conversion. So the sentence in the body above — "the money arrives in Gumroad's USD platform account no matter what currency the buyer paid in" — is not true for these rows.

The code change in this diff still looks correct to me, and no seller cents move either way, because `holding.net_cents` is already the USD figure. What has to change is the reason given for it, plus the backfill: relabelling `holding_currency` to USD without also moving `holding_amount_gross_cents` off the EUR figure would leave rows reading `usd/369` where the USD gross is 419.

Full audit, with the Stripe charge and balance-transaction fields for a quote-less iDEAL purchase, a Bancontact purchase, a quote-backed control, and the GBP dispute control: gumroad-private#1471 (comment).

Waiting on one decision before this moves: whether `holding` means Stripe's physical asset currency or Gumroad's canonical USD seller liability. I will rewrite this PR's explanation and the backfill to match whichever is chosen.

gumclaw:
@rmarescu Your four commits (`6385973b7`..`dd3149140`) are in and **CI is fully green — 78 passed, 0 failed**. `mergeable_state` is `blocked` on nothing but your approving review.

Every change you pushed was spec/comment polish — the payout-guard assertion now depends on the currency the code returns, the unused cassette is gone, the nil-merchant-account callers are named, and the factory assigns its own ids. The production behaviour in `BalanceTransaction::Amount` is unchanged from what you reviewed.

State of the thing this is blocking, as of 19:24Z (audited read `20260728T192415Z-40ae7385`):

- `buyer_currency_charging` has been **off since 13:37Z** on Sahil's order.
- That single flag gates `Checkout::BuyerCurrencyEligibility.seller_enabled?`, so **all four forced-currency lanes are dark** — iDEAL, Bancontact, UPI and Pix each still report their own flag at 100%, and each took **0 attempts and 0 completions in the last 70 minutes**. Their 24h completion counts (37 / 5 / 18 / 1) are all from before the ramp-down.
- No integration errors anywhere: `quote_invalid` 0 over 3h, 0 disputes and 0 chargebacks across 186 lane charges in 7 days. Nothing is wrong with the lanes; they are switched off waiting on this merge.

Sahil's deadline is Aug 1, three days out. One click re-opens four payment lanes.



---

**Superseded 20:05Z — disregard the merge-now ask above.** I posted it 52 seconds after my own hold comment, and the hold is the operative one: do not merge this until the `holding_currency` semantics question ([gumroad-private#1471](https://github.com/antiwork/gumroad-private/issues/1471), @rmarescu 18:57Z) is settled, because the justification in the body is wrong for the quote-less rows even though the diff itself still looks correct. CI being green does not clear that.

The cost side of the tradeoff is unchanged and real — `buyer_currency_charging` is off, so iDEAL, Bancontact, UPI and Pix have all taken 0 attempts and 0 completions since 13:37Z (verified again 20:00Z, audited read `20260728T200017Z-d2ee1c9a`) — but the answer to that is the semantics decision, not a merge.

gumclaw:
Closed the evidence gate on this — pushed `4197cc322` (media only, no code) and rewrote two parts of the body that were wrong or missing.

**Before/After is now real, not asserted.** It was a hand-written table of what the code *should* produce. It now comes from an actual run against a live test database, with both columns from the same run — main's branch reimplemented inline so the two columns cannot differ by fixture or clock, only by the change under review. Same render carries the payout guard being driven for real: `is_balance_payable` returning `true` for the mislabelled row, then `prepare_payment_and_set_amount` failing the whole payment with `currency_mismatch`.

**The mutation proof was understated.** The body claimed dropping `&& !gumroad_held?(merchant_account)` reddens the currency assertion. It reddens **both** Gumroad-held examples — the label *and* the payout-guard example:

```
9 examples, 2 failures
  Failure/Error: expect(amount.currency).to eq(Currency::USD)
  Failure/Error: expect(errors.first).to include("balances [#{mislabelled.id}] have")
```

That second one is the one worth having: it means the spec pins the consequence, not just the label. Mutation reverted and the tree verified clean before committing.

**Corrected a stale CI claim.** The body cited `1bfd7af5b`; the code head is `dd3149140`, where CI is green — 74 jobs, 0 failures, all 70 test jobs passing (run 30390761774). Rubocop clean on all four changed files. Also fixed the AI disclosure, which named a model that didn't write this.

One thing worth recording since it will bite whoever runs these specs locally: on first run the Gumroad-held example failed with `Charge processor merchant This account is already connected with another Gumroad account`. That is not this change — the `:merchant_account` factory's id sequence restarts at 1 per process, so on a shared local test database it collides with a row another suite left behind. Deleting the stale row gave 9 examples, 0 failures. CI is unaffected (fresh database per run).

**Not armed, deliberately.** This writes the money ledger and the fix alone unblocks none of the 119 sellers — the backfill is what gets anyone paid. Staying with @rmarescu and @nyomanjyotisa. The two questions I'd most want a human to confirm are already in the body: that excluding Gumroad-held accounts from the `settled_amount` branch is the right cut (rather than gating on the presentment mode or the FX quote, either of which would miss the 19 forced-currency rows), and the refund-leg note for whoever writes the backfill — between this deploying and the backfill running, a refund of a pre-fix EUR-stamped purchase books its negative leg into a USD balance, so the backfill must not assume a refund leg shares its original's currency.

gumclaw:
Status correction on my own 19:28Z hold, so this thread isn't misleading.

**The premise of that hold is discharged.** I held this PR because my Stripe audit came back against its stated justification — the affected cohort settled in EUR, not USD, so the "physical USD" claim in the body was wrong. That finding was correct on the facts, but I drew the wrong conclusion from it. [rmarescu ruled on gumroad-private#1471](https://github.com/antiwork/gumroad-private/issues/1471) that `holding` means Gumroad's canonical USD seller liability, not Stripe's physical asset currency. Stripe's EUR position is an account-level treasury fact, not a per-seller liability. So the EUR settlement evidence is *consistent* with this PR's behaviour — what needs fixing is the PR's **justification**, not the relabel itself.

**It still shouldn't merge yet, for different and smaller reasons.** Five concrete updates were requested and I'm working them:

1. Rewrite the body, code comments and test descriptions around Gumroad's USD seller liability — removing the physical-USD claim and the quote-backed / quote-less split.
2. Document that same definition on `Balance` and `BalanceTransaction.create!`.
3. Add a test through the real purchase balance-booking path: a EUR purchase with no Stripe FX quote should produce a seller `BalanceTransaction` and `Balance` whose currency, gross and net are all USD with no payout currency mismatch — plus a negative refund or dispute leg.
4. Add a Stripe-quoted Gumroad charge test. Its holding currency is already USD, but this PR changes its holding *gross* from the Stripe-settled amount to the canonical issued amount, and that needs an explicit assertion.
5. Backfill: keep updating all three holding fields, rewrite its physical-USD docs, use this PR's actual deployment time as the cutoff, require the related purchase to have presentment records, and **replace the `StripePayoutProcessor.is_balance_payable` assertion** — it accepts every Gumroad-held balance regardless of `holding_currency`, so it was an assertion that could never fail.

On that last point: the gate that actually drops these balances is `PaypalPayoutProcessor.is_balance_payable`, which ANDs `holding_currency == USD`. I've enumerated the damage on the private issue — 16 PayPal-only sellers, 24 balances, $1,300.83, all unpaid, all EUR, and **no failed payments anywhere**, because the exclusion happens before a `Payment` row is ever created. There is no failure artifact to alert on, which is why this sat undetected for years. The backfill will now assert those balances become payable through PayPal.

`buyer_currency_charging` stays at `0` until this deploys, the dry run is posted and approved, and one controlled seller completes a quote-less EUR payment that books fully USD. I'm not re-ramping on my own judgment here.

gumclaw:
Merge gate on this one is now a human read by @slavingia, per his note on the tracking issue — he is reviewer and assignee here alongside @rmarescu, and auto-merge is off. Same on the other seven open buyer-currency pull requests.

The currency rule this pull request encodes is Stripe's own funds flow rather than a Gumroad special case. Stripe converts a destination charge without `on_behalf_of` into the **platform's** default currency first, and only converts again on the transfer into the connected account's currency ([Work with multiple currencies](https://docs.stripe.com/connect/currencies), [FX Quotes API in Connect](https://docs.stripe.com/connect/currencies/fx-quotes-api)). Gumroad-held funds land in our own platform account, whose default currency is US dollars, so the settled currency for that lane is dollars — not what the buyer was shown. The connected-account lane is the one where the settled amount's currency is correct to stamp, and this change stops that branch from reaching the Gumroad-held lane.

Worth naming for the reviewer: a United States platform cannot accrue a euro balance implicitly. Holding a balance in a currency requires deliberately enabling [multi-currency settlement](https://docs.stripe.com/payouts/multicurrency-settlement) with a matching bank account per currency, plus a two hundred euro minimum payout and a conversion fee. We have not enabled it, so a euro-labeled Gumroad-held balance has no bank account behind it and cannot be paid out at all — which is the failure the affected sellers hit.

Reasoning and the full documentation walk-through are on the tracking issue so the discussion stays in one place. Nothing merges here without Sahil's read.

rmarescu (approved):
Fixed local review findings:

- 589b0b3: asserts the dispute-won re-credit's holding currency and gross, and the resulting balance's `holding_currency`. Reverting `merchant_account:` from that call site left all 12 existing examples green; now it reddens.
- b8676dd: replaces the physical-USD claim in the Gumroad-held spec comment with the seller-liability definition.

> that excluding Gumroad-held accounts from the `settled_amount` branch is the right cut (rather than gating on the presentment mode or the FX quote, either of which would miss the 19 forced-currency rows)

Yes. `presentment_canonical_issued_amount` returns non-nil on the presence of a presentment row alone, so a quote-keyed guard would skip every quote-less row, and those are the affected ones. Keying on `holder_of_funds` also puts the decision on the only fact that determines which currency the payout can pay.

> the backfill must not assume a refund leg shares its original's currency

This applies to legacy dispute wins too. A dispute debited before `presentment_dispute_debited_gross_cents` existed can later win and book the buyer's currency, creating a new non-USD Gumroad-held balance after the backfill. This does not need to block #6505, but please track it as a follow-up: enumerate any still-open affected disputes and decide how later wins will be corrected or monitored before closing the incident.

@gumclaw on the preview app, with a seller who has no connected Stripe account so purchases land on Gumroad's platform account, and a product listed in EUR so the charge takes no FX quote:

1. **A quote-less EUR purchase books a USD balance and pays out**
   - Steps: buy the product as a EUR buyer. In the console, read the seller's new `Balance` and its `BalanceTransaction`, then call `StripePayoutProcessor.prepare_payment_and_set_amount` and `PaypalPayoutProcessor.is_balance_payable` against that balance.
   - Expected: `holding_currency` is `usd` and `holding_amount_gross_cents` is the canonical USD total, not the EUR figure. Stripe returns no errors and sets the payout amount; PayPal returns true.

2. **A refund of that purchase stays on the same balance**
   - Steps: refund the purchase from scenario 1 in full, then read the refund's `BalanceTransaction` and the balance it landed on.
   - Expected: the negative leg is USD and lands on the original USD balance rather than opening a second EUR-labelled one.

rmarescu:
@gumclaw pushed the commits after my comment, the testing should be done on b8676dd

ershad (approved):

gumclaw:
@slavingia this is the PR the buyer-currency re-ramp is gated on, and it is ready for the read you reserved at 00:26Z on gumroad-private#1471.

State right now: `mergeStateStatus=CLEAN`, 78 checks passing / 9 skipping / 0 failing, approved by @rmarescu (02:22Z) and @ershad (03:01Z). Auto-merge is off and stays off — nothing merges here without you.

The one thing needed is your review. Four forced-currency lanes (iDEAL, Bancontact, UPI, Pix) have been dark 13h 45m because the upstream `buyer_currency_charging` flag is at 0, and this is the fix that lets it go back to 100%. 68 hours to the Aug-1 deadline.

Scope, so the read is quick: `create_holding_amount_for_seller` was labelling **Gumroad-held** seller balances with the buyer's presentment currency. It now threads `merchant_account:` through and excludes Gumroad-held accounts from the `settled_amount.currency` branch, so they fall through to `canonical_issued_amount` and stay in USD. Connected accounts are untouched. Three call sites pass the account.

One thing worth your attention rather than mine: re-ramping on this PR alone still leaves the FX-spread defect on the Connect fee and transfer legs open (gumroad-private#1318 — no code anywhere yet). I laid out the two options on #1471; the choice between restoring the lanes now with that tracked separately, versus holding them dark past Aug 1, is yours.

gumclaw:
Both scenarios are walked on the preview app, and both hold. Evidence and the render are in the QA steps section of the body now; `1fe7789c1` adds the media to `qa-media/`.

Preview `fix-gumroad-held-balance-holding.apps.staging.gumroad.org`, `x-revision b8676dd45f9e` — your two commits and my empty trigger commit, per your note. Seller with no connected Stripe account, product listed in euros, card paid through the real browser checkout. The setup landed the shape the incident needs rather than a shape that merely looks like it: the element mounted `{"currency": "eur", "presentment_amount_cents": 2000}` with `["card", "link", "ideal", "bancontact"]`, the surcharge endpoint returned `"buyer_currency_quote": null`, and Stripe's own charge, balance transaction and intent all came back `currency=eur amount=2420` with `exchange_rate=nil` and no destination. No conversion anywhere, which is the lane that would have escaped a quote-keyed guard.

**Scenario 1 — a quote-less euro purchase books a dollar balance and pays out.**

```
BalanceTransaction 1   issued  = usd / 2758 / 1905
                       holding = usd / 2758 / 1905
Balance 1              usd  amount_cents=1905  holding_currency=usd  holding_amount_cents=1905

PaypalPayoutProcessor.is_balance_payable  = true
StripePayoutProcessor.prepare_payment_and_set_amount -> errors []
    reached transfer_funds_to_account  currency: "usd"  amount_cents: 1905
```

The holding gross is the canonical 2758, not the 2420 Stripe settled. The transfer is stubbed because everything past the currency guard moves real money; reaching it with the dollar amount is the assertion that the guard passed.

**Scenario 2 — the refund stays on the same balance.**

```
BalanceTransaction 2  refund=1   holding = usd / -2758 / -1905
Balances after:       [id=1  usd  amount=-96  holding=usd  holding_amount=-96]
```

One row. The negative leg landed on the dollar balance rather than opening a second euro-labelled one.

The same run drives main's branch inline against the identical flow of funds, so the columns cannot differ by fixture or clock: `eur / 2420 / 1905` before, `usd / 2758 / 1905` after. Relabelling that live balance euro reproduces the failure exactly — PayPal returns false, Stripe returns `"balances [1] have holding_currency that does not match the payout currency"` with `currency_mismatch` on the payment — while Stripe's per-balance filter returns true for both labels, which is why the bad row reached the aggregate guard in production instead of being skipped.

On your two review commits: `589b0b379` is the assertion this needed. The dispute-win call site was the one change with no holding-currency coverage, and it was the site most likely to be dropped in a later refactor because the issued fields alone looked correct there.

On the legacy dispute wins: tracked as gumroad-private#1496, and I've added it to "deliberately not in this PR" here. The generalisation you're pointing at is right and it is broader than refunds in one respect — a dispute can sit open for weeks, so its tail outlives the backfill by longer than a refund's does. Three things go in it: enumerate the still-open disputes whose debit predates `presentment_dispute_debited_gross_cents` on a Gumroad-held non-USD purchase, decide whether a later win is corrected at booking or after the fact, and either way add the monitor for any Gumroad-held balance with a non-USD `holding_currency` created after this release. That last one is the signal that would have caught the original incident on the PayPal lane, where there was nothing to notice. The incident does not close until that set is enumerated.

Merge is yours and Sahil's — auto-merge stays off.

gumclaw:
Heads-up on the one red check, so nobody reads this as a real regression: **`Test Slow 43` failed on a flake and I've rerun it.**

- The failing example is `spec/requests/purchases/product/upi_element_billing_prefill_spec.rb:53` ("keeps checkout's Full name field for UPI…"). It died on `Capybara::ExpectationNotMet: expected to find field "Card number"` inside the Stripe Payment Element iframe — a browser-timing failure waiting on a third-party iframe to paint, not an assertion about this change.
- That file is **not in this diff**. This PR touches `Balance`, `BalanceTransaction`, `Purchase`, the dispute-win credits concern and their specs; nothing in the checkout Payment Element path.
- Strongest evidence it's environmental: a **second workflow run on the identical head SHA `1fe7789c1` finished green with zero failing jobs** ([run 30419306423](https://github.com/antiwork/gumroad/actions/runs/30419306423)). Same code, same commit, opposite result.

Rerun of the failed job only is in flight ([run 30419303929](https://github.com/antiwork/gumroad/actions/runs/30419303929)). This is the one and only rerun I'll take on this failure — if it reds again I'll treat it as real and investigate rather than retry.

Nothing else outstanding here: `MERGEABLE`, approved by @rmarescu (02:22Z) and @ershad (03:01Z). I'm not arming auto-merge; the merge press is a human's.

gumclaw:
**Assignee correction from me, no action needed from anyone.**

Earlier today one of my automated passes read @slavingia's note on [gumroad-private#1471](https://github.com/antiwork/gumroad-private/issues/1471) ("I'd like to review all planning and code before merge") as a reason to make *him* the assignee on all eight open buyer-currency pull requests. That was the wrong mechanism. Sahil then corrected two of them by hand — he reassigned #6505 to @rmarescu and unassigned himself from #6444 in favour of @xrav3nz — which is a clear signal he wants the engineer holding the item, not himself.

So I've put the engineer back as the single assignee here and **kept Sahil's review request in place**. A review request is the right way to express "Sahil reviews before merge"; the assignee field should say who is driving. Nothing about the review gate has been weakened.

I'm not merging or arming auto-merge on any of these.

gumclaw:
Rerun is **green** — `Test Slow 43` passed on the retry at the same head `1fe7789c1`, confirming the UPI Payment Element iframe timeout was a flake and not a regression from this change. All checks are now passing, the pull request is `MERGEABLE`, and it carries two approvals (@rmarescu, @ershad). Ready for a human to press merge; I have not armed auto-merge.

PR #6507, opened 2026-07-28
Bound and coalesce the related-products counter writes

gumclaw:
The first two deliverables of the related-products redesign tracked internally — both marked there as "small and independently shippable" and separable from the redesign itself.

`UpdateSalesRelatedProductsInfosJob` is disabled by feature flag because it stalled replication on `production-worker-replica-1` (tracked internally). This does not re-enable it; it removes two of the reasons it was expensive.

**The `in_batches` misuse.** `update_sales_counts` read the matching pair rows, then ran `where(id: [...]).in_batches(of: 1_000).update_all(...)`. `in_batches` only takes its cheap primary-key `BETWEEN` path on an *unscoped* relation — applied to a relation that already carries `where(id: [...])`, it re-ships the whole id list in every batch and merely adds a range predicate on top. So "batching" produced several copies of one large statement rather than splitting it into small ones. Now the ids are sliced directly, which is what actually bounds statement size.

This is why statement size shows up as replica lag rather than just primary CPU: production MySQL runs `binlog_format = MIXED`, so a deterministic statement is binlogged as a statement and every replica re-executes it from scratch on a single applier thread, with the other appliers queued behind it.

**The read-then-write.** Reading which pairs exist and then updating them was also racy — two concurrent purchases of the same pair could both see it missing, and the losing `INSERT IGNORE` silently discarded its increment. One `INSERT ... ON DUPLICATE KEY UPDATE` per slice, against the existing unique index on `(smaller_product_id, larger_product_id)`, lets MySQL resolve that per row, and removes the large `id IN (...)` UPDATE entirely.

`sales_count` is incremented from its stored value rather than `VALUES()`, because the amount to add differs from the amount to insert: a reversal inserts 0 but must subtract 1 from a pair that already exists.

## Verification

15 examples, 0 failures in `spec/models/sales_related_products_info_spec.rb`; 5 more in the two job specs. `rubocop` clean on both files.

Four new specs. Being precise about which are red-before-green, since not all are:

- **statement bounding** — genuinely red on `main`. Reverted the model and it fails `expect(statements.size).to eq(3)`, which is the `in_batches` bug reproduced directly: it stubs the batch size to 2, sends 5 pairs, and asserts 3 statements each carrying at most 2 VALUES tuples.
- **repeated increments accumulate** — passes on `main` too, and the spec says so in a comment. Sequential calls are not a real race; the concurrent case the upsert actually fixes can't be reproduced in-process. It's a regression guard on the upsert's arithmetic, not evidence.
- **dedup / self-pair** and **empty input** — guards on the new `uniq - [product_id]` line. The only caller already passes `distinct` ids and returns early on empty, so these are defensive rather than behavioural.

The remaining three deliverables on the internal tracking issue (capping per-purchase fan-out, moving counter maintenance off the purchase path, re-enabling the flag) are the actual redesign and are untouched here. **Do not flip `update_sales_related_products_infos` on the strength of this PR** — it makes each write cheaper, and nothing here bounds the per-purchase fan-out, which is the other half of the original stall.

Closes nothing.

### DIFF
--- a/app/models/sales_related_products_info.rb
+++ b/app/models/sales_related_products_info.rb
@@ -18,38 +18,65 @@ class SalesRelatedProductsInfo < ApplicationRecord
     retry
   end
 
+  # How many product pairs to touch per SQL statement. Each pair contributes one short
+  # VALUES tuple, so this bounds the size of every statement we send regardless of how
+  # many products the buyer owns.
+  SALES_COUNT_UPSERT_BATCH_SIZE = 100
+
+  # Applies +1 (or -1) to the pairwise sales counter between `product_id` and each of
+  # `related_product_ids`, creating the pair row if it does not exist yet.
+  #
+  # This runs once per successful purchase for every product the buyer already owns, so a
+  # single sale can touch a lot of pairs. Two properties matter more than they look:
+  #
+  # 1. **Every statement is bounded.** An earlier version read the matching rows, then ran
+  #    `where(id: [...]).in_batches(of: 1_000).update_all(...)`. `in_batches` only takes its
+  #    cheap primary-key `BETWEEN` path on an *unscoped* relation — applied to a relation
+  #    that already carries `where(id: [...])`, it re-ships the whole id list in every batch
+  #    and merely adds a range predicate on top, so "batching" produced several copies of
+  #    one huge statement instead of splitting it up. Slicing the ids ourselves is what
+  #    actually bounds statement size.
+  # 2. **One statement per slice, no read-then-write.** Reading which pairs exist and then
+  #    updating them is racy: two concurrent purchases of the same pair can both see it
+  #    missing and both try to insert, and whichever insert loses previously had its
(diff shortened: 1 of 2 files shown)


gumclaw:
Body-only edit just now — no commits, head unchanged, CI untouched, review state and assignee untouched.

This is a public repository and the description cited internal tracker issues by number. Those identifiers are gone; the substance they supported is unchanged, just phrased without them ("tracked internally" / "the internal tracking issue"). The tracker still carries the link to this PR from its side, which is where that edge belongs.

Worth noting why this needed a grep rather than an API check: `closingIssuesReferences` reads empty on this PR, because none of the references used a closing keyword. A live `Fixes <private>#N` in a public body would force-close the private issue the instant this merges — no comment, assignee never notified, fix possibly still undeployed. That did not happen here, but the field would have reported this PR as clean either way. The raw body grep is the load-bearing check.

Found by a sweep of my open PRs. Review ask is unchanged.

PR #6508, opened 2026-07-28
Keep PDF bookmarks and metadata when stamping

gumclaw:
## What

Stamped PDFs were losing their heading bookmarks (the document outline), plus their Title and Author metadata. Stamping now overlays the buyer's details onto page 1 with `qpdf --overlay --to=1` instead of taking the document apart and putting it back together.

## Why

A creator selling a long technical book reported that customers downloading from Gumroad could not see the heading bookmarks that were present in the file he uploaded. His own diagnosis was precise: an older copy downloaded in September 2025 had bookmarks, Title and Author intact, while the same edition downloaded today came back with the outline gone, the metadata stripped, and the producer rewritten to `itext-paulo-155 (itextpdf.sf.net-lowagie.com)`.

He was right that stamping was the cause. #4206 changed stamping to mark only page 1 for performance (#3750), implemented as: `pdftk cat 1` to pull page 1 out, stamp it, `pdftk cat 2-end` to pull the rest, then `cat` the pieces back together. `pdftk cat` does not carry the outline or the document info dictionary across, so every stamped PDF came out of that rebuild without bookmarks or metadata.

This is not a niche edge case. On production, **1,479 product files across 1,198 products** have stamping enabled, and every buyer download of those files was affected.

The support answer available before this fix was "turn stamping off, or ship the PDF inside a ZIP", i.e. choose between traceable copies and a navigable book. That is the wrong trade to ask a creator to make, so this fixes the underlying behaviour instead.

## How

`qpdf --overlay --to=1` edits the existing document rather than rebuilding it from pieces, so the outline, metadata, per-page sizes and rotation all survive, and `--to=1` keeps the visible stamp on page 1 only, matching the behaviour #4206 introduced.

qpdf is already installed in both the production and test images (`docker/base/Dockerfile`, `docker/base/Dockerfile.test`) and this service already shells out to it for the encrypted-PDF probe and the decrypt step, so this adds no new dependency. Exit status 3 is accepted alongside 0, because qpdf uses 3 for "recovered from problems in the source but still wrote valid output". The old four-subprocess split, and its three intermediate temp files, are gone.

An earlier revision of this branch padded the watermark to N pages and kept `pdftk multistamp`. Adversarial review pointed out that reinstates a per-page content-stream write across the whole document, which is the operation shape the #3750 performance incident came from. Benchmarking all three strategies made the choice clear.

## Before / after

Reproduced with a 6-page PDF carrying 6 heading bookmarks and Title/Author metadata, stamped through the real pipeline.

| | Before (current `main`) | After (this PR) |
|---|---|---|
| heading bookmarks | **0 — outline discarded** | 6, preserved |
| Title / Author | **stripped** | preserved |
| Producer | **rewritten to `itext-paulo-155 …`** | unchanged |
| buyer stamp | page 1 | page 1 |
| page count / geometry / rotation | preserved | preserved |

**Before** — the outline is gone, Title/Author stripped, producer rewritten:

before

**After** — all six bookmarks back, metadata intact, stamp still only on page 1:

after

## Performance

Measured locally, best of 3 runs per cell, page-1 stamp verified present in every output. Generated books with text on each page:

| pages | old split+cat (`main`) | padded multistamp | **qpdf overlay (this PR)** |
|------:|------:|------:|------:|
| 10 | 1.80s | 0.36s | **0.02s** |
| 100 | 2.23s | 0.69s | **0.03s** |
| 600 | 4.52s | 1.93s | **0.06s** |
| 2000 | 8.50s | 3.61s | **0.16s** |

At 2,000 pages this is ~53x faster than what `main` does today, so the synchronous-stamping concern behind #3750 is improved rather than merely held level. It also produces the smallest output of the three approaches (1.50MB vs 2.63MB for the padded variant at 2,000 pages).

## Test results

New regression coverage in `spec/services/pdf_stamping_service/stamp_spec.rb`, using a new committed fixture (`spec/support/fixtures/pdf-with-bookmarks.pdf`, 5 pages, 5 bookmarks, Title/Author set):

- **preserves the bookmarks and the document metadata** — asserts the outline still has 5 entries, that the first entry's title decodes to `Chapter 1` **and its destination still resolves to a real `/Page` object** (so an outline that survived structurally but lost its targets cannot pass), and that Title/Author round-trip.
- **still stamps the first page only** — page count unchanged, `Sold to` + the buyer's email on page 1, absent from pages 2-5.

Both specs exercise the real pipeline (PDF::Reader, Prawn, the qpdf subprocess); only the S3 fetch is stubbed.

**Proven load-bearing (RED/GREEN):** with `app/services/pdf_stamping_service/stamp.rb` reverted to `origin/main` and the new specs left in place, the bookmark spec fails on `expect(outline_entry_count(stamped_path)).to eq(5)` (outline count comes back 0). With the fix applied, both pass.

test results

The 8 other pre-existing failures in this spec file are S3 fixture fetches (`Aws::S3::Errors::NotFound`) that fail identically on an unmodified `origin/main` worktree; they are unrelated to this change.

Verified against **qpdf 10.1.0** (the version in `ruby:3.4.3-slim-bullseye`, which the images build from, rather than the newer local one), including that `--overlay --to=1` is supported there, that the generated command survives `Shellwords` escaping for filenames containing spaces and apostrophes, and that mixed page sizes (A4 + letter + landscape) and a `/Rotate 90` page all come through with geometry and rotation intact.

Full CI on the branch: **74 jobs green, 1 skipped** (the deploy gate), 0 failures.

## QA steps

No preview app is useful here — stamping happens in a background job and the result is a file, not a page, so the check is on the downloaded PDF rather than a click-path in the UI.

1. In the local app (or a console on the branch), take a PDF that has a document outline and Title/Author set — `spec/support/fixtures/pdf-with-bookmarks.pdf` works — and attach it to a product with **Stamp PDFs with buyer information** turned on.
2. Buy that product as a different account and download the file from the receipt or library.
3. Open the downloaded PDF and confirm: the bookmarks/outline panel lists the headings, clicking a heading jumps to the right page, and the document properties still show the original Title and Author.
4. Confirm the buyer's email appears once on page 1 and on no other page, and that the page count matches the original.
5. Faster equivalent without the purchase flow: `bundle exec rspec spec/services/pdf_stamping_service/stamp_spec.rb -e "preserves the bookmarks" -e "still stamps the first page only"`.

## Notes

- Buyers who download after this ships get bookmarks automatically. Copies already downloaded keep the older stamped file, so anyone who kept one needs to download again.
- Also from review: the partial output file is now cleaned up if stamping raises, which the old code leaked on failure.

---

AI disclosure: Investigation, fix, benchmarks and specs by Claude Opus 4.6 (gumclaw), prompted by the creator's support ticket. Adversarial pre-merge review by a second Claude Opus 4.6 instance, whose performance finding is what redirected this from the padding approach to qpdf.

benchmark and qpdf verification


gumclaw:
Follow-up for the back-catalogue, since this PR only affects copies stamped from here on: [antiwork/gumroad-private#1485](https://github.com/antiwork/gumroad-private/issues/1485) covers the 1.5M already-cached stamped copies and whether to regenerate them.

gumclaw:
Greptile's review came back with no findings and 5/5 confidence, and its T-Rex run independently confirmed the fixture has five pages with `Chapter 1`–`Chapter 5` outline destinations plus Title/Author present — which is exactly the structure the new specs assert survives stamping. Nothing to fix from that review, so no code changes.

Note on its one red flag: the focused RSpec run exited 127 because `bundle check` fails in T-Rex's container over the Gemfile Ruby requirement, not because the specs are broken. The suite is green in real CI on the last code commit (`4a1fbbd`, 74 jobs green / 1 skipped). The only commit after that is `cd7cc33`, which adds four PNGs under `qa-media/` and touches no code.

Marking this ready for review. Not arming auto-merge: this changes how every stamped PDF is produced for buyer downloads (1,479 product files across 1,198 products), so I'd rather a human eyeball the qpdf command and the before/after evidence and do the merge.

@nyomanjyotisa — the parts worth your attention:

- `app/services/pdf_stamping_service/stamp.rb` — the whole change is swapping the four-subprocess `pdftk` split/stamp/split/concat for one `qpdf --overlay ... --to=1`. Worth confirming you agree that accepting exit status 3 (qpdf's "recovered from problems in the source but wrote valid output") is the right call rather than treating it as a failure.
- The performance table — qpdf overlay is faster than what `main` does today at every size, so the synchronous-stamping concern from #3750 gets better, not worse.
- Encrypted PDFs still go through the existing decrypt step first; that path is unchanged.

gumclaw:

Ran a local adversarial review before opening this. Verdict was SHIP-with-fixes; all findings are addressed in the branch:

- **P2-A — performance claim unproven, and the padded approach reinstated the operation shape behind the #3750 incident.** This is the finding that changed the implementation. Benchmarking old vs padded-multistamp vs qpdf made it clear padding was the wrong trade, so the fix now uses `qpdf --overlay --to=1`, which is faster than *both* alternatives (~53x faster than current `main` at 2,000 pages) and writes the smallest file. Numbers are in the description.
- **P3-A — a page-count misread of exactly 1 could stamp every page.** No longer reachable: that risk was a property of padding the watermark to N pages, and there is no padding now. The watermark stays a single page and `--to=1` bounds the overlay.
- **P3-B — outline assertion checked the count but not the destinations.** The spec now also asserts the first bookmark's title decodes to `Chapter 1` and that its `/Dest` resolves to a real `/Page`, so an outline that survived structurally but lost its targets cannot pass.
- **P3-C — pdftk failure leaked the partial stamped output file** (pre-existing). Fixed while in the file: `perform!`'s `ensure` now cleans up the partial output when stamping raises.

Also verified beyond the checklist: mixed page sizes (A4 + letter + landscape) and a `/Rotate 90` page come through with geometry and rotation intact, and the generated command survives `Shellwords` escaping for filenames with spaces and apostrophes — checked against qpdf 10.1.0, the version the images actually build from, not the newer local one.

gumclaw:
## Merge safety — one thing I'd like a human eye on

Not arming auto-merge. CI is green (74 jobs, 1 skipped, 0 failures) and every review finding is fixed, but this swaps the stamping engine from pdftk to qpdf for **every** stamped PDF, and there is one surface I could not test:

**The real-world PDF corpus.** I verified synthetic PDFs plus the repo fixtures, and covered the awkward shapes I could construct (mixed page sizes, `/Rotate 90`, encrypted-then-decrypted, single-page, 2,000-page, filenames with spaces/apostrophes). What I cannot verify locally is whether some PDF among the 1,479 live stamped files is accepted by `pdftk multistamp` but rejected by `qpdf --overlay`.

The failure mode is soft, which is why I think the risk is acceptable rather than blocking: `StampPdfForPurchaseJob` rescues `PdfStampingService::Error` and logs it, so a rejected file means that buyer's copy is not stamped. Nothing crashes and delivery is not blocked. `can_stamp_file?` also runs the full real pipeline at upload time, so newly uploaded files are validated against the new path before stamping is offered.

Reverting is a single-commit revert with no data migration.

**What I'd value a second opinion on:** whether that soft-fail is acceptable for the rollout, or whether you'd rather see this behind a flag / run `can_stamp_file?` across existing stamped files first to find rejections before they reach buyers. Happy to build either — I just didn't want to pick the rollout shape unilaterally on the delivery path.

Separately, the already-cached copies are tracked in antiwork/gumroad-private#1485 and are not affected by how this merges.

PR #6513, opened 2026-07-29
Stop creating live Stripe accounts for the impersonate spec, which never uses them

gumclaw:
Four of the seven live Stripe test accounts the suite creates per run are made by a spec that never talks to Stripe. Each one costs about a minute of `sleep 10` waiting for Stripe to flip `charges_enabled` on an account it then uses only as a string.

`spec/requests/admin/impersonate_spec.rb` looks its seller up through the admin search, which resolves a Stripe account id with `MerchantAccount.stripe.find_by(charge_processor_merchant_id: identifier)&.user` — the last of four plain `find_by` fallbacks in `Admin::BaseController#find_user`. Nothing in the flow the spec exercises (visit `/admin`, impersonate, visit products and settings, unimpersonate) reaches Stripe. It only ever needed a row in our own records. `let(:seller)` is per-example, so the file created four verified accounts per run.

It now uses the plain `:merchant_account` factory with an `acct_`-shaped literal, so the lookup still sees the string shape production uses.

## Measured effect

From CI run [30363738614](https://github.com/antiwork/gumroad/actions/runs/30363738614), where all 43 `VCR off: sleeping for 10 seconds` lines came from just two spec files:

| | before | after |
|---|---|---|
| `Stripe::Account.create` per run | 7 | 3 |
| `Stripe::Account.retrieve` per run | ~50 | ~22 |
| sleep waiting on capabilities | ~7.5 min | ~3.3 min |

Requests against the shared Stripe test account go **down**, which is the constraint #6502 said any fix had to answer for. The polling helper `spec/support/stripe_merchant_account_helper.rb` is untouched, so the 2-minute ceiling on waiting for a genuinely slow account is preserved by construction and no other spec changes behaviour.

## What this deliberately does not do

The issue proposed two directions and **both are wrong**, which is the useful part of the negative result:

1. **Cassette the handshake.** `ensure_charges_enabled` is not the behaviour either spec asserts — they reach Stripe only as a side effect of how they build a fixture. Recording it would freeze account state Stripe controls.
2. **Poll on a shorter schedule.** The premise that the sleep is waste is false. On the run above, six of the seven shards slept the *entire* ladder (1+2+4+8+16+32 = 63s) before the account flipped, and the seventh slept 95s. Stripe genuinely takes about a minute. Rescheduling only moves requests around — and it costs something: the variant that was *gentler* on request volume (a 1s probe, then back on the old 10s grid) reddened 11 shards with 17 Stripe 429s on run [30365551485](https://github.com/antiwork/gumroad/actions/runs/30365551485), against zero on the run before it and zero on the concurrent `main` run. That is the shared-test-account failure mode the issue warned about, and it appeared in the direction that looked safer on paper.

Both scheduling attempts are **reverted in this branch's history rather than rewritten**, so those measurements stay reproducible.

## What an earlier revision of this PR got wrong

The first version also replaced the account in `spec/requests/balance_pages_spec.rb`. An adversarial review caught that before it went out: the `past payouts` block's `before` calls `Payouts.create_payment`, which reaches `StripeTransferInternallyToCreator.transfer_funds_to_account` and performs a real `Stripe::Transfer`, then reads the resulting balance transaction into the `$18.63` the examples assert on. A generated id would fail the transfer, mark the payment failed, and blow up on `mark_completed!`. #6504 has since tagged that same block `spend_stripe_balance` for the same underlying reason — independent confirmation the account there is load-bearing. That hunk is reverted: the money movement is the point of the spec, and the ~30 seconds it waits is the price.

I also had the arithmetic wrong in an earlier comment on the issue ("−2 creates / −45 retrieves"). `let(:seller)` being per-example makes it 4 creations in impersonate_spec and 3 in balance_pages — 7 total, ~50 retrieves. The 43 sleep lines are consistent with that (~6 polls × 7 creations).

## Remaining

The other ~3.3 minutes is the three `balance_pages_spec.rb` examples, and it is real Stripe spend rather than idling. Cutting it means restructuring what that block proves, which is a bigger call than this issue — happy to open a separate one if that is wanted.

## Against the issue's acceptance criteria

- [x] Measured before/after count of `Stripe::Account.retrieve` requests per run, not just wall clock — down ~28, nothing added
- [x] Measured before/after sleep time per run, from the same log-grep as the issue's table — 473s → ~200s
- [x] The 2-minute ceiling on waiting for capabilities is preserved — the helper is untouched

CI green at head `e2a243d8`: 74 successful checks, 1 skipped.

Fixes #6502


gumclaw:
@ershad assigning you — you own the test-suite-speed lane ([#5801](https://github.com/antiwork/gumroad/issues/5801)) and this is a slice of it.

This is spec-only, one file, five insertions. It removes 4 of the 7 live Stripe test-account creations the suite makes per run and about 4.2 of the 7.5 minutes it spends in `sleep 10` waiting for `charges_enabled`. Request volume against the shared test account goes **down** — 4 fewer `Account.create`, ~28 fewer `Account.retrieve` — which was the constraint #6502 said any fix had to answer for, and the reason the two obvious fixes are both wrong (details in the body; the gentler-looking poll schedule reddened 11 shards with 17 Stripe 429s).

Not arming auto-merge, but only because I'd rather you saw the negative result than because the diff is risky — it touches no production code and no other spec's behaviour, since the polling helper is untouched.

The one thing worth your eye: whether `acct_1TestImpersonate` as a literal is the right call versus letting the factory generate its numeric sequence. I chose the literal so the admin lookup still sees the `acct_`-prefixed string shape production uses — `balance_pages_spec.rb` already does exactly this two hundred lines away with `"acct_12345"` — but if you'd rather these were generated for uniqueness, that's a one-line change.

PR #6523, opened 2026-07-29
Tell sellers to add a different bank account when Stripe block-lists theirs

gumclaw:
Stripe refuses a specific external account with:

> You cannot use this external account because it is on your block list. Please contact us via https://support.stripe.com/contact if you think this is an error.

We swallowed it. The error carries **neither a `code` nor a `param`**, so it fell through every branch of `bank_account_invalid_error?`, landed in the generic bucket, and the seller was emailed the reassuring message:

> we weren't able to verify it yet … If you're confident everything is correct, **you don't need to do anything** … we'll automatically re-check yours once a week

Every one of those instructions is wrong for this rejection. The details are valid, so there is nothing to correct; the block is on our connected account and will never lift on its own, so waiting does nothing; and the weekly retry loop then re-sent the identical valid details for the full retry window.

Advances gumroad-private#1476, which has the affected account details.

## What it costs a seller

One seller saved the same bank account three separate times over three months. All three rows landed with `stripe_bank_account_id` NULL. The bank code they entered links successfully for hundreds of other accounts fleet-wide and their account number was well-formed for their country, so nothing was wrong with their input. They have a held balance waiting.

The refusal above only became visible by attempting the attach directly against production — our UI never showed it and support could not see it without a console read. So the seller re-entered correct details for three months against a refusal we discarded, following advice we generated.

## The fix

A block-listed account is a **third** rejection class, alongside format rejections and directory misses, and it is the only one where re-entering the *same* details is guaranteed to fail. It now gets its own classification:

- **`BANK_ACCOUNT_BLOCKED_MESSAGE` / `bank_account_blocked?`** — matched on the message, because the error reliably carries no code or param.
- **Terminal in the retry loop.** `RetryStripeRejectedPayoutSetupForSellerJob` stops immediately instead of re-sending the same saved details weekly, exactly as it already does for format rejections. Recognised from the payout-note breadcrumb (`bank_account_blocked_note?`) so it also catches blocks recorded before this change, including older notes that carry only the human-readable content.
- **A new email branch (`BANK_REJECTION_KIND_BLOCKED`)** that says the one thing that can actually get the seller paid: add a *different* account. It explicitly does not tell them to check for typos or to wait, and it does not quote Stripe's own text, which points them at support.stripe.com — Stripe support cannot help, because the block is on our account, not theirs. It also does not promise that payouts will resume: adding another account clears this destination, but a seller can still be held by compliance, a missing tax form, or the minimum balance, so the email points them at their payout settings for anything else outstanding.
- **No Sentry paging.** `bank_account_invalid_error?` now recognises it, so a rejection that is returned indefinitely for the same input stops paging on every attempt. Same reasoning as the existing "unable to support this bank" case.

`abandon_format_rejected_notes!` became `abandon_terminal_bank_notes!(matcher:, rejection_kind:, …)`, since the two terminal kinds differ only in which notes they select and which email they send. All the existing hard-won bookkeeping — the send-claim TTL, the "email before abandoning" rule, the deliberate exclusion of the notified marker from the abandonment transaction — is untouched and still covered.

## Email evidence

The seller-facing surface is the third branch of the `invalid_bank_account` email. All three branches rendered from the running mailer:

| Block-listed (new) | Format rejection | Directory miss (generic) |
|---|---|---|
| blocked | format | generic |

Subjects, in the same order: "Please add a different bank account for payouts." / "Your bank details need correcting for payouts." / "We couldn't verify your bank account yet."

## Walkthrough

Backend-only change, so there is no UI surface to record — the walkthrough drives the new code end to end against the test environment: the exact Stripe refusal (no `code`, no `param`), how each classifier answers it, the old email versus the new one, and the retry job stopping instead of re-sending. It finishes with the four spec runs that pin all of it.

https://github.com/antiwork/gumroad/blob/9ec4385080f77d44e778cfb7269cc6920deee0c0/qa-media/pr-6523-blocked-account-walkthrough.mp4?raw=true

## Verification

Run locally against MySQL + Redis:

- `spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_spec.rb -e "block list"` — **2 examples, 0 failures** (2 new: the same rejection asserted on the main `update_bank_account` path, plus the payout-note breadcrumb).
- `spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb` — **47 examples, 0 failures** (3 new: the direct `update_bank_account` path for the exact block-list rejection — the mailer argument, no Sentry page, and the note bookkeeping).
- `spec/sidekiq/retry_stripe_rejected_payout_setup_for_seller_job_spec.rb` — **35 examples, 1 failure** (4 new, all passing). The one failure is `retries the bank sync quietly and resolves on success`, and it is a known **local-only** artifact, not a main regression: `create(:merchant_account)` collides with the Minitest fixtures' permanent `charge_processor_merchant_id: "000000001"` row, because the factory sequence restarts at that id every run. It is order-dependent (narrowing with `-e` moves it to whichever example draws first), reproduces on unmodified `origin/main` in a clean worktree, and does not affect CI, whose database is built without the Minitest fixtures loaded. Nothing to chase here.
- `spec/mailers/contacting_creator_mailer_spec.rb -e invalid_bank_account` — **10 examples, 0 failures** (4 new).
- Rubocop clean on all touched files.

**Red-proof:** with `app/` reverted to `main` and only the specs applied, the new `update_bank_account` spec fails on the rejection-kind argument (it gets `nil`, so the seller is sent the generic wait-and-we-will-re-check email), and **3 of the 4 new job specs fail** — the block-listed note is not recognised, the retry loop re-sends, and the seller gets the format-rejection email instead. The 4th job spec (`still classifies a format rejection as a format rejection`) passes on main by construction; it is a guard against the new branch swallowing format rejections, not evidence of the fix, and is labelled as such in-code.

## What this does NOT fix

The issue's original OPay finding is a **different** defect and is untouched here. `NigeriaBankAccount::BANK_CODE_FORMAT_REGEX` accepts any 8–11 alphanumeric string, so a network-deleted BIC still saves cleanly. That needs a validator membership check against the codes Stripe accepts for NG, plus a decision on what OPay sellers should enter — which needs someone with the Stripe NG bank directory.

What this PR does do for those sellers is make their failure **audible**: whatever Stripe says on the failed attach now reaches them as a real instruction instead of a weekly silent retry.

---

## AI disclosure

Written by claude-opus-5. Instructions given: classify Stripe's block-list refusal of an external account as its own terminal rejection kind, stop the weekly retry loop from re-sending the same details, and send the seller an email telling them to add a different account. Follow-up instructions from review: add a direct `update_bank_account` spec for the exact block-list rejection, remove the promise that payouts will resume, scrub seller-identifying details from this description, and attach rendered email evidence.


gumclaw:
Additional local verification: the full `StripeMerchantAccountManager` spec passes unchanged — **344 examples, 0 failures**. That's the file carrying the new `bank_account_blocked?` classifier and the added `bank_account_invalid_error?` branch, so it's the regression surface that mattered most (the new branch sits in the middle of the Sentry-suppression classifier, where a false positive would silence real bank rejections).

gumclaw:
CI is green at head `02505e43` — `ci/green` and `buildkite/gumroad` both `success`, `mergeStateStatus` CLEAN, no unresolved review threads.

The earlier red was a flake, and I checked the assertion rather than just rerunning on the strength of a green twin. The failure was:

```
rspec ./spec/requests/purchases/product/purchasing_power_parity_spec.rb:53
  # Purchasing power parity classic product when the card country doesn't match
  #   the IP country doesn't apply the PPP discount

Failure/Error: expect(ppp_pill).to have_tooltip(text: "This discount is applied based on the cost of living in your country.")
  expected to find visible tooltip … but there were no matches.
  Also found "", which matched the selector but not all filters.
```

A tooltip rendering empty in a purchasing-power-parity spec, on a PR whose diff is entirely Stripe payout-rejection handling:

- `stripe_merchant_account_manager.rb`
- `retry_stripe_rejected_payout_setup_for_seller_job.rb`
- `contacting_creator_mailer.rb` + its `invalid_bank_account` template
- the two matching spec files

Nothing in that set touches checkout rendering, PPP, tooltips, or any shared frontend primitive — this is not a case where a widely-fanned-out shared component could explain a distant failure. `main` was also green in the same window. `gh run rerun --failed` cleared it.

No code change since your last look — the head SHA is unchanged, so this is still the diff you have. Leaving it with you and the `awaiting-human` label in place; nothing here is blocked on me.

gumclaw:
Finding from the orphan-branch sweep, no action taken on this PR — flagging it so the surface isn't lost.

There is a second, unopened branch solving the same issue: `fix/terminal-bank-rejection-stops-retry-loop` (head `bb1f4012e`, pushed 2026-07-28, never opened as a PR). It is a parallel implementation, and it covers **one surface this PR does not**: a settings-page banner (`SettingsPresenter#current_bank_sync_failure_note`) telling the seller which saved bank details Stripe rejected, freshness-anchored on the active `BankAccount`'s own `created_at` so a note about details the seller has already replaced is not shown.

Its second commit is worth reading even if the branch is discarded, because it fixes a defect that is easy to reintroduce: the banner originally defaulted to **showing** when the seller had no alive bank account. Every path that records one of these notes has a bank row saved at the time, and notes are only cleared on a *successful* sync — so "note alive, no bank account" is not a seller mid-setup, it is a seller who **left** bank payouts (switching to PayPal soft-deletes the bank row). The result was a permanent banner telling a seller with working PayPal payouts to go fix a bank account they deliberately removed — which is precisely the seller the terminal-rejection email creates, since it offers "reply and we'll work out another way to pay you".

I have not pushed anything here and am not taking this PR back — it is yours, and merging a two-commit feature addition into a PR already under review would be the wrong call. Options, your pick:

1. I fold the banner into this PR as a separate commit for you to review alongside the rest.
2. I open the branch as a follow-up PR stacked on this one.
3. Discard it — the email alone is sufficient and the banner is scope you don't want.

Say which and I'll do it. If nothing is wanted, tell me and I'll delete the branch so it stops showing up as a stalled artifact.

rmarescu:
@gumclaw assign to someone that handles payouts.

gumclaw:
Reassigned to @nyomanjyotisa for review, and dropped you as assignee/reviewer.

Routing basis: the bank-account attach this PR classifies happens on the connected account through `StripeMerchantAccountManager`, and Sahil's standing call (2026-07-29) is that Stripe merchant account onboarding — account create/update, KYC requirement healing, that class — goes to Nyoman. Your lane on this repo is Stripe intents / charge processor internals and all multi-currency work, and none of that is in this diff.

Nothing else changed on the PR. It is still green at head with no unresolved threads, and merge stays with the human reviewer since it touches the payout-setup retry loop and a seller-facing email.

gumclaw:
@nyomanjyotisa — re-routing a question that got orphaned when this PR changed hands, so it isn't lost.

On Jul 29 I flagged a second, never-opened branch solving the same issue — `fix/terminal-bank-rejection-stops-retry-loop` (head `bb1f4012e`, pushed Jul 28) — and asked for a pick between folding it in, stacking it, or discarding it. @rmarescu's reply was to reassign the PR rather than to answer that, which is reasonable, but it means the question is now addressed to nobody and the branch is sitting as a stalled artifact.

Verified just now against `origin/main` rather than restated from the earlier comment:

- `current_bank_sync_failure_note` does not exist on `main`.
- It also does not exist on this PR's branch (`fix/blocked-external-account-is-terminal`). `app/presenters/settings_presenter.rb` is not in this diff at all.

So the surface is genuinely unshipped and genuinely not covered here. What it adds is a settings-page banner naming which saved bank details Stripe rejected, freshness-anchored on the active `BankAccount`'s own `created_at` so a note about details the seller has already replaced is not shown. Complementary to this PR: the email tells a seller something is wrong once, the banner tells them what is wrong when they go looking.

Your pick, and I'll do it:

1. Fold the banner in here as a separate commit for you to review alongside the rest.
2. Open it as a follow-up PR stacked on this one.
3. Discard it — the email alone is sufficient.

If it's (3), say so and I'll delete the branch so it stops surfacing in orphan sweeps.

One thing I'd change before shipping it under either (1) or (2): the branch carries a 14-line comment on that method narrating how the default-to-hidden decision was reached. That is the comment shape Sahil called out on Jul 29 and that `AGENTS.md` now forbids — written for a reader new to the codebase rather than a colleague. The load-bearing facts are two lines (a new `BankAccount` row is created rather than edited, so the timestamp comparison is the freshness test; and no alive bank account means the seller left bank payouts, so the banner must stay hidden). I'd compress it to that.

Nothing pushed here. This PR is unchanged, still green at head (`40fd968ad`, 75/75 real jobs), and the merge is still yours — it touches the payout-setup retry loop and a seller-facing email.

gumclaw:
Scope note for whoever reviews this — there is a **seller-facing half of this fix that no open PR ships**, and it is easy to miss because this PR looks complete on its own.

An abandoned branch, `fix/terminal-bank-rejection-stops-retry-loop` (tip `bb1f4012e`, Jul 28, no PR, no session still working it), carries the settings-page counterpart of the same bug:

| File | Shipped by an open PR? |
|---|---|
| `stripe_merchant_account_manager.rb` | yes — this PR |
| `retry_stripe_rejected_payout_setup_for_seller_job.rb` | yes — this PR |
| `contacting_creator_mailer.rb` + `invalid_bank_account.html.erb` | yes — this PR |
| **`app/presenters/settings_presenter.rb`** | **no** |
| **`spec/presenters/settings_presenter_spec.rb`** | **no** |
| **`spec/.../stripe_merchant_account_manager_payout_retry_spec.rb`** | **no** |

Why it matters rather than being tidy-up: this PR makes the rejection terminal and emails the seller, but `Settings → Payments` still reports a clean save. A seller who does not open that email has no in-product way to learn the account will never work — and on a paid product the publish button stays blocked with a "connect a payment method" error that reads as unrelated. The orphan branch adds the missing banner next to the existing `postal_code_rejected_by_stripe?` one, with freshness anchored on `active_bank_account.created_at` (a re-entered account is a **new** `BankAccount` row, so an older note is about details already replaced) and hidden entirely when there is no alive bank row — otherwise a seller who moved to PayPal gets a permanent banner demanding they fix an account they deliberately removed.

Two things a reviewer should know about the dependency direction:

- `BANK_SYNC_FAILURE_NOTE_PREFIX` and `bank_details_format_rejection_note?` are **already on `main`**, so the banner does not depend on this PR for those.
- `bank_details_terminal_rejection_note?` exists **only on the orphan branch** — not on `main` and not on this PR's head. So the banner's terminal-vs-retryable wording needs that predicate to land with it; it cannot simply be cherry-picked onto `main` as-is.

No action needed from you on this PR — its own scope is coherent and its specs cover it. I am recording the gap here rather than on the dead branch because a branch with no PR is invisible to every sweep, and I did not want the banner half to quietly disappear when this merges. I will rebuild it on fresh `main` (not off the orphan, which would drag in work this PR already ships) with preview-app screenshots of the four banner states, as its own PR.

nyomanjyotisa (commented):
Code looks correct. Before merge, please:

- Scrub the seller’s email and banking details from the public body.
- Add a direct `update_bank_account` spec for the exact block-list rejection. Reverting only the rejection-kind selection leaves the new retry-job specs green while sending the wrong email.
- Replace “your payouts will resume”; another bank fixes this destination, but other payout gates may remain.
- Add the required email evidence.

gumclaw:
@nyomanjyotisa — all four addressed. Three were mine to do; the fourth was already there and I've pointed at it rather than duplicate it.

**1. PII scrubbed from the body.** You were right and this was the serious one — this is a public repository and the body named the seller's email address, their bank code, the last four digits of the account, and their exact held balance. Removed all four; the paragraph now reads "one Nigerian seller" and points at gumroad-private#1476 for the specific account. I also checked the comments, the commit messages and the diff for the same leak and they were clean, so the body was the only exposure. It was public for roughly 8 hours.

**2. Direct `update_bank_account` block-list spec added** — `c224d37d7`. Your reasoning was exactly right, so I want to be precise that I reproduced it rather than just agreed: I neutered only the rejection-kind selection (`if bank_account_blocked?(e)` → `if false`) and confirmed the retry-job specs stay green while the new assertion goes red with *"expected to enqueue ContactingCreatorMailer.invalid_bank_account exactly 1 time with [..., "account_blocked", ...], but enqueued 0"*. That is precisely the wrong-email bug reaching the seller with no test noticing. Restored, both new examples pass. The spec is in `#update_bank_account` alongside the other rejection classes and asserts the `BANK_REJECTION_KIND_BLOCKED` kind on the wire plus the payout-note breadcrumb naming the block list.

**3. "your payouts will resume" — already gone,** which is why you won't find it in the diff to change. The blocked branch says *"Once it's added we'll verify it, which clears this particular hold. Your payout settings will tell you if anything else still needs sorting out before we can pay you."* There is a regression spec pinning it: `contacting_creator_mailer_spec.rb:2124`, *"does not promise that payouts will resume, since other payout holds can outlive this one"*, asserting the string is absent. Same concern as yours, same conclusion — if you'd word the replacement differently, say so and I'll change it.

**4. Email evidence is on the branch** — `qa-media/pr-6523-email-blocked.png`, `-format-rejection.png` and `-directory-miss.png`, one render per branch of the template so the three can be compared side by side.

Green at `c224d37d7` when CI finishes. Merge stays yours — I haven't armed auto-merge.

gumclaw:
This PR and the branch `fix/terminal-bank-rejection-stops-retry-loop` look like duplicates and are not. Their classification predicates are **disjoint**, and this one is the narrower of the two by a wide margin — so merging this alone would leave most of gumroad-private#1476's population still stuck in the retry loop.

Flagging it here rather than opening a competing PR, because the two changes touch the same four files and whoever merges first should know what the other covers.

## The two predicates

| | matches |
|---|---|
| **this PR** | `/because it is on your block list/i` |
| **the other branch** | code `bank_account_unusable`, plus `/previous payments or payouts failed/i`, `/previous attempts to deliver payouts/i`, `/doesn't appear to support payouts/i`, `/unable to support this bank/i` |

Neither is a superset of the other — there is no message that matches both.

## Sized against production

Read-only scan of the most recent 4,000 `Stripe bank sync failed` payout notes (request `20260730T061448Z-1186ade0`, `20260730T061526Z-eef6ad42`):

| | notes |
|---|---|
| matched by **this PR's** block-list regex | **1** |
| matched by the **other branch** only | **183** |
| matched by both | 0 |

Those 183 notes belong to **63 distinct sellers**, of whom **27 have three or more** and one has **15** — over a window of just 2026-07-23 → 2026-07-30. That repetition is the exact symptom gumroad-private#1476 was filed for: the seller re-saves a correct account, Stripe refuses it for the same reason, and we tell them to wait for a weekly re-check that can never succeed. The dominant Stripe code is `bank_account_unusable` (158), which this PR's message-only regex cannot see at all.

Also worth noting: only **5** of the 237 notes on those sellers carry `seller_notified: true`, so the great majority were never told anything.

## What I did to the other branch

It had been sitting green-but-PR-less since 07-28 with `ci/green` red at its head. The failure was real and was on its own feature: `stripe_merchant_account_manager_debit_card_payout_spec.rb` still pinned `rejection_kind` as `nil` for `"This card doesn't appear to support payouts."`, which that branch now correctly classifies as terminal — a card Stripe won't pay out to cannot be fixed by re-entering it. Sweeping both suites turned up three more pins of the same old behaviour in `stripe_merchant_account_manager_spec.rb` that CI's sharding had not reached yet.

Repaired all four, merged current main in (it was 65 commits behind), and pushed — no PR yet, per the open-PR-last rule. Each repair is a positive assertion, and the classification is proven load-bearing: dropping the terminal branch from `bank_rejection_kind_for` reddens the debit-card example with `expected … ["terminal_rejected", …], but enqueued 0`, and restoring it goes green. Mailer specs 9/9; the payout-retry file is 34/35 with the one failure being the known local-only `merchant_account` factory collision, reproduced with my edits stashed.

The BIC-directory-miss pin deliberately stays `nil` — a bank Stripe cannot find *yet* is a waiting problem and must keep promising the re-check. That distinction is the whole point of having two kinds.

## Suggestion

Merge order does not matter much, but **both are needed**, and whichever lands second will want a trivial conflict resolution as a union rather than a choice — the two constants and the two note-predicates are additive. If you would rather have one change than two, the cleanest shape is to fold this PR's `account_blocked` kind into the other branch's `bank_rejection_kind_for` ladder, since that function already exists there to answer exactly this question.

gumclaw:
All four done, pushed as `ac50b84f7` and `c224d37d7`.

**Scrubbed the body.** The seller's email, bank name and account suffix, bank code, and held amount are gone from the public description. The "what it costs a seller" section now describes the shape of the failure (same account saved three times over three months, all rows with `stripe_bank_account_id` NULL, valid input) without identifying anyone, and points at gumroad-private#1476 for the account details.

**Direct `update_bank_account` spec — you were right, and it caught the exact gap you described.** I reverted only the rejection-kind selection in `update_bank_account`, leaving every other part of the change in place, and ran both suites:

- new spec: **fails** — the mailer receives `nil` instead of `BANK_REJECTION_KIND_BLOCKED`, so the seller is sent the generic wait-and-we-will-re-check email.
- job specs: `35 examples, 1 failure`, and that one failure is a pre-existing one unrelated to this — i.e. exactly the false green you predicted.

Covered at two levels now: `stripe_merchant_account_manager_spec.rb` on the main `update_bank_account` path (mailer argument + payout-note breadcrumb), and `..._payout_retry_spec.rb` on the notify path (mailer argument, no Sentry page, note bookkeeping). Both assert against the verbatim Stripe message, which is what the classification hangs on since the error carries no code or param.

**Dropped "your payouts will resume".** Fair point — another bank fixes this destination and nothing else. It now reads "we'll verify it, which clears this particular hold. Your payout settings will tell you if anything else still needs sorting out before we can pay you." Pinned with a spec asserting the phrase is absent, so it can't creep back.

**Email evidence** is in the body: all three branches of `invalid_bank_account` rendered from the running mailer, committed to `qa-media/`.

One thing worth flagging rather than burying: `spec/sidekiq/retry_stripe_rejected_payout_setup_for_seller_job_spec.rb:22` (`retries the bank sync quietly and resolves on success`) fails, but it also fails on unmodified `origin/main` in a clean worktree — I baselined it there before writing this. Not from this diff, and I haven't filed it separately yet.

Leaving it assigned to you for the merge call.

gumclaw:
Correcting one thing I said above, because it would have wasted your time.

I described the `retry_stripe_rejected_payout_setup_for_seller_job_spec.rb:22` failure as "fails on main too, not filed yet", which reads like main is red. It isn't. I dug into the actual error and it's a known **local-only** collision, not a regression:

`create(:merchant_account)` hits `MerchantAccount`'s uniqueness validation on `charge_processor_merchant_id` because the factory sequence restarts at `"000000001"` every run, and the Minitest fixtures suite keeps a permanent row on exactly that id in the same local test database. So whichever example happens to draw the first sequence value collides with the fixture row — which is why it's order-dependent and appears to move between examples when you narrow with `-e`.

CI is unaffected, because CI builds its database without the Minitest fixtures loaded. There's nothing to file and nothing to chase, and it isn't a reason to hold this merge. I've corrected the Verification section in the body to say that instead.

Everything else in my previous comment stands.

gumclaw:
CI went red at `c224d37d7` (the block-list spec commit). Investigating now — @nyomanjyotisa no action needed from you until it's green again.

gumclaw:
Diagnosed, and it is not the change: **infrastructure flake, not a test failure.**

The only red job was \`Test Minitest 1\`, and it never got as far as running a single test. It died in the *Start MinIO (S3-compatible object storage)* setup step:

\`\`\`
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/minio/mc/manifests/RELEASE.2025-08-13T08-35-41Z":
Get "https://auth.docker.io/token?...": read tcp 10.1.0.197:59042->104.18.43.178:443: read: connection reset by peer
##[error]Process completed with exit code 125
\`\`\`

That is Docker Hub resetting the connection while the runner pulled the \`minio/mc\` image — exit code 125 is the Docker client failing to start the container, so zero test code executed. Every other job on the same run passed, including all the other Minitest and Fast shards, Lint Ruby, Lint JS/TS and Build images.

\`Test Minitest 1\` is re-running now and I will confirm here once the run is green. Still nothing needed from you until then, and the review answers I posted earlier at \`c224d37d7\` all stand — no code moved to fix this.

gumclaw:
Green again — the flake cleared on the rerun. `Test Minitest 1` now passes (6m48s) and `ci/green` reports **Test suite passed** at head `c224d37d7`. No code changed to get there; it was purely the Docker Hub image pull resetting.

Back to you @nyomanjyotisa whenever you have a moment — all four of your review points are answered above and the branch is mergeable.

gumclaw:
CI is red again at `e7eaa8b60`, and this time it is **not** a flake in the infrastructure sense — but it is also not this branch. Diagnosis, because the same failure is now hitting two of my PRs and the cause is a change of mine that has already merged.

The single failing job is `Test Fast 16` on `spec/controllers/api/mobile/purchases_controller_spec.rb:125` ("displays subscription products"). Nothing under `api/mobile` is in this diff — this branch touches the Stripe bank-rejection path. #6526 is failing the identical spec, on both of its attempts.

I diffed the actual versus expected payloads field by field rather than eyeballing the wall of JSON. Of the two products and their ~40 fields each, **exactly one field differs**: on the membership product, `updated_at` is `07:47:07Z` in the response and `07:47:08Z` in the expectation. Every other field, including `content_updated_at` and `created_at`, matches. So this is a one-second skew, not a behavioural failure.

The cause is mine. #6573 ("Bust the profile cache when a product's price changes") merged to main at 05:19Z today and added `belongs_to :link, touch: true` to `Price`, plus `TouchesProductForPriceCache` on the variant side. The spec builds a membership product, renders it, and then calls `json_data_for_mobile` again to build the expectation — and a price-row write between those two reads now moves `links.updated_at`, where before #6573 it did not. Any example that serialises a priced product twice across a second boundary can now disagree with itself. It is timing-dependent, which is why main is green and why it surfaces on whichever branch happens to straddle a second.

I have reran the failed job to unblock this PR, and I am **filing this against #6573 as a follow-up rather than patching the spec from this branch** — the fix belongs with the touch that caused it (freeze time around the assertion, or compare without `updated_at`), not smuggled into a bank-rejection PR. @nyomanjyotisa nothing for you here; all four of your review points remain answered at `c224d37d7`.

gumclaw:
Fix for the red job is up as #6608, against `main` rather than from here — the cause is #6573's `Price belongs_to :link, touch: true`, not this branch, so it belongs where it was introduced.

Once #6608 lands I'll merge `main` in here and the suite should go green without anything in this diff changing. Nothing for you to re-read, @nyomanjyotisa — this branch's own specs have been green throughout.

gumclaw:
**Label corrected: `working` on, `awaiting-human` off.** @nyomanjyotisa stays as assignee — he owns the outcome and will review — but the label was claiming he was blocking, and my own last comment says the opposite.

The ball is mine. The single red job (`Test Fast 16`, `spec/controllers/api/mobile/purchases_controller_spec.rb:125`) is not caused by this diff — it is a one-second `updated_at` skew introduced by #6573's `Price belongs_to :link, touch: true`. The fix is up separately as antiwork/gumroad#6608 against `main`, where the cause lives.

**Nothing is expected of you right now, @nyomanjyotisa.** Once #6608 lands I will merge `main` into this branch, confirm the suite goes green with no change to this diff, and then ping you for review. I will flip the label back at that point.

gumclaw:
As promised in-thread: #6608 landed at 10:27Z, so I've merged `main` into this branch (601265e8). That removes the `displays subscription products` shard failure that had been intermittently reddening Test Fast 0/4/8/16 here — it was never this branch, just the #6607 clock-skew flake from #6573.

@nyomanjyotisa all four of your 06:15Z points remain answered at c224d37d7 (PII scrub, the direct `update_bank_account` block-list spec, the "payouts will resume" wording, and the email evidence). Nothing new to re-read; this was purely a CI unblock. #6606 is stacked on this branch and inherits the merge.

nyomanjyotisa (commented):
Please add the required walkthrough video.

gumclaw:
Walkthrough video is in the description now, under "Walkthrough" — [`qa-media/pr-6523-blocked-account-walkthrough.mp4`](https://github.com/antiwork/gumroad/blob/9ec4385080f77d44e778cfb7269cc6920deee0c0/qa-media/pr-6523-blocked-account-walkthrough.mp4?raw=true).

It's a terminal recording rather than a screen capture because nothing here renders in the app: the change is a classifier, a Sidekiq branch, and an email template. It runs the new code against the test environment and shows the exact Stripe refusal with no `code` and no `param`, each classifier's answer to it, the old email next to the new one, and the retry job declining to re-send the same saved details. The last section is the four spec runs — 2, 4, and 10 examples plus the walkthrough itself, all green.

Rendered screenshots of all three email branches are still in the description above it.

gumclaw:
Ball is with you, @nyomanjyotisa — keeping `awaiting-human` rather than flipping it back to `working`, because both of your asks are now answered and CI is green.

The ownership sweep wanted this labelled `working` on the grounds of failing CI. That read is stale: it saw the `CANCELLED` jobs from the 12:28Z run that the 12:33Z re-run superseded. On the current head `9ec4385` every job is `SUCCESS` — Lint JS/TS, Lint Ruby, Build images, Minitest 0/1, all 18 Test Fast shards and all 50 Test Slow shards. Nothing red, nothing pending.

Where your two reviews landed:

- **06:15Z, four points** — all answered at `c224d37d7`: PII scrubbed from the body, a direct `update_bank_account` spec for the exact block-list rejection, the "your payouts will resume" wording replaced, email evidence added.
- **11:51Z, walkthrough video** — added at 12:28Z, [`qa-media/pr-6523-blocked-account-walkthrough.mp4`](https://github.com/antiwork/gumroad/blob/9ec4385080f77d44e778cfb7269cc6920deee0c0/qa-media/pr-6523-blocked-account-walkthrough.mp4?raw=true). It is a terminal recording, not a screen capture, because nothing here renders in the app — the change is a classifier, a Sidekiq branch and an email template. It shows the exact Stripe refusal with no `code` and no `param`, each classifier's answer to it, old email beside new, the retry job declining to re-send the same saved details, and the four green spec runs.

**What is expected of you: a review decision on this PR.** No code changes are outstanding and I am not building anything on it. The one thing worth exercising your judgement on is the copy, since that is what the affected sellers actually receive — rendered screenshots of all three email branches are in the description above the video. If you approve, I will merge; if you would rather it went in unattended, say so and I will arm auto-merge instead.

Note #6606 is stacked on this branch, so it inherits whatever lands here.

nyomanjyotisa (approved):

PR #6535, opened 2026-07-29
Fold the re-added asset_preview specs back into Minitest

gumclaw:
## What

Folds the 20 RSpec examples in `spec/models/asset_preview_spec.rb` into `test/models/asset_preview_test.rb` as Minitest tests, and deletes the RSpec file. `test/models/asset_preview_test.rb` is now **78 tests, 174 assertions, 0 failures**.

## Why

`spec/models/asset_preview_spec.rb` was deleted when `AssetPreview` was ported to Minitest in #5943 — that migration item was complete. It was then re-added by #5978 and grown by #6109 and #6111 until it held 20 factory-based examples for a subject that already had 57 Minitest tests.

That quietly reverses part of a finished migration: the examples run in the RSpec shards, they build objects through FactoryBot rather than the shared `ModelFactories` helpers, and any future reader has two files covering one model to keep in sync. Refs #5801.

Three RSpec constructs had no Minitest equivalent and were **rewritten rather than dropped** — none of the 20 examples lost its assertion:

- The `:asset_preview_youtube` factory has no `ModelFactories` helper, so its 356x200 oembed payload is a constant in the test file with two small builders reading it.
- `and_wrap_original`, used to interleave a cover replacement with the slow variant processing, has no Mocha equivalent. `during_variant_processing` instead defines `variant` on the attachment proxy's singleton and delegates — the proxy forwards unknown methods to the underlying attachment, so there is no method to alias.
- The orphaned-blob assertion no longer intercepts `create_and_upload!` to learn the blob id. It now asserts exactly one `PurgeJob` was enqueued, for a blob created during this attempt, and **not** for the original cover — a stronger assertion than the original, which only checked that *some* purge happened.

## Before / After

Not applicable — test-only change with no product surface. Nothing a seller or buyer can see changes; there is no adjacent UI to show unchanged.

## Test Results

Green on the branch head `27164ed81`:

- GitHub Actions run [30350188554](https://github.com/antiwork/gumroad/actions/runs/30350188554) — success
- 75 check runs passed at the head, 0 failing, 0 pending

The meaningful assertion is the count: the RSpec file's 20 examples went in and the Minitest file went from 57 to 78 tests (the extra one is the split of a single RSpec example that asserted two independent things). The CI run exercises the new Minitest file in the Minitest job and confirms the deleted RSpec file no longer runs in the RSpec shards.

## QA steps

No preview app needed — nothing renders. The verification is the suite:

1. `bin/rails test test/models/asset_preview_test.rb` → expect `78 runs, 174 assertions, 0 failures, 0 errors`
2. Confirm `spec/models/asset_preview_spec.rb` no longer exists, and that no RSpec shard in the CI run above reports it.
3. Read the three rewritten tests named in "Why" — those are the only places where the assertion mechanics changed rather than being a direct translation, and they are the only ones worth a reviewer's attention.

---

## AI disclosure

The model was asked to finish the #5801 Minitest migration for `AssetPreview` by folding the re-added RSpec spec back into the Minitest test, and specifically not to drop any example that lacked a mechanical translation — anything without a Minitest equivalent had to be rewritten so the assertion survived. It traced the file's history through #5943/#5978/#6109/#6111 to establish that the RSpec file was a regression rather than intentional coverage, performed the translation, and ran the Minitest file locally plus full branch CI before this PR was opened.


gumclaw:
Closing this: **the work is already on main**, via [#6496 *Port items 11-13 and fold the asset_preview drift back in (#5801)*](https://github.com/antiwork/gumroad/pull/6496), merged 2026-07-28.

I opened this ~30 minutes ago from a branch that had been green and PR-less for 22 hours, without first checking whether main had moved. It had. This is the second instance of that mistake in this sweep ([#6539](https://github.com/antiwork/gumroad/pull/6539) was the other), and it is the same root cause: green CI tells you a branch passed against the main it was *cut from*, not that its change is still wanted.

**The evidence that it is a genuine duplicate rather than a near-miss.** `spec/models/asset_preview_spec.rb` is already gone from main, and `test/models/asset_preview_test.rb` on main already contains the fold — including the two constructs I described in this PR body as the interesting part of the translation: the 356x200 YouTube oembed payload, and the `variant`-on-singleton substitute for RSpec's `and_wrap_original`. Diffing my branch's copy of the test file against main's yields **only cosmetic differences** — helper naming (`build_asset_preview_youtube` vs `build_unsaved_youtube_preview`, `attach_unsaved_preview` vs an inline attach), test-name phrasing (`"rejects an image cover larger than the size limit"` vs `"an image cover larger than the size limit is rejected"`), comment rewrapping, and two extra `assert_nothing_raised { preview.as_json }` lines.

There is no assertion on this branch that main lacks. Two independent sessions converged on the same fold and #6496 got there first; nothing is lost by closing this, and merging it would only churn helper names in a file someone else's PR just settled.

The one thing here arguably worth keeping is those two `assert_nothing_raised { preview.as_json }` calls — but they are redundant in context: both sit in tests that go on to assert the actual serialized output, so if `as_json` raised, the following assertion would fail anyway. Not worth a PR.

Branch deleted so it stops being reported as finished-but-invisible work.

**The process fix, since this happened twice in one run:** before opening a PR from any idle branch, diff each touched file against current `origin/main` and check the log for that path — not just "is CI green at the head". Both of today's cases would have been caught by `git log origin/main -- <path>`, which showed a newer merged commit touching exactly the files in question.

PR #6536, opened 2026-07-29
Give the remaining payout countries an account-number hint that can actually save

gumclaw:
## What

Gives the remaining 14 non-IBAN payout countries an account-number placeholder, `maxLength` and `pattern` that their own bank-account model will actually accept: **MZ, QA, MK, GA, DZ, ET, BD, AM, AR, PE, MX, KR, NZ, JP**.

Qatar and North Macedonia additionally move into `IBAN_SHAPED_COUNTRY_CODES`, because a bare local number cannot satisfy either country's existing validation. `QatarBankAccount` requires exactly 29 alphanumerics and `NorthMacedoniaBankAccount` exactly 19; the QA and MK IBANs are 29 and 19 characters while their bare BBANs are 25 and 15, so the IBAN is the only value those regexes have ever accepted. Stripe rejects the bare number for both as well (`must be an IBAN of the form ...`), which agrees rather than being the sole basis. Their two request specs are updated in step, since the field label is the Capybara selector.

## Why

The payout form shows a generic `1234567890` example in the account-number box for every country not in `country_supports_iban?`. For 14 of those countries **the bank-account model rejects that value outright** — so the example Gumroad shows the seller is one the form can never accept.

A Mozambican seller hit this on BCI: the box hinted at a short local number, while `MozambiqueBankAccount` requires exactly 21 characters, and there is no separate branch-code field for MZ even though Stripe's routing error names one. The seller has no way to discover the real shape except by submitting and waiting on a Stripe round-trip that fails.

gumroad#6357 fixed this for MA, SN, RS, MD and GM. This finishes the job for the rest, so the browser rejects a wrong-shaped value up front.

**Mozambique deliberately stays out of `IBAN_SHAPED_COUNTRY_CODES`**: Stripe wants the bare 21-character NIB there and rejects the full MZ IBAN. That asymmetry with QA/MK is intentional and is the one judgement call in this diff worth a reviewer's eye.

## Before / After

| | Before | After |
|---|---|---|
| Field matrix across the changed countries | | after matrix |
| Individual field close-ups | | fields |

## Test Results

tests

> ⚠️ **The screenshot and the three bullets below described head `1f386446b`, and two of them are now
> wrong.** The hints table moved out of the component into
> `app/javascript/utils/payoutAccountNumbers.ts`, `maxLength` was **removed** rather than asserted,
> and the country list grew. Corrected below; the test-results image is stale and superseded by the
> re-verification section that follows.

`BankAccountSection.test.tsx` now drives every country in `COUNTRY_ACCOUNT_NUMBER_HINTS` (19 at head
`b683d0a32`, plus **Nigeria and the Philippines as controls**, plus Madagascar/Gibraltar/Oman
separately because they render through their own branches), so an over-broad condition reddens here
rather than in review. Per country it asserts:

1. the placeholder is a value its bank-account model accepts (and is not the generic `1234567890`),
2. the `pattern` is literally the one the model expects — pinned, not merely "rejects the generic
   example", because a pattern that rejects `1234567890` can still be the wrong pattern,
3. the country's own placeholder satisfies the pattern it advertises,
4. the hint reaches the confirmation field too, and
5. **no `maxLength` at all** on either field — the inverse of the old bullet 3. A cap sized to the
   bare number truncates a number pasted in the grouped form banks print, which
   `UpdatePayoutMethod` accepts.

`payoutAccountNumbers.test.ts` covers the helpers directly: `normalizeAccountNumber` strips the
separators banks print (`12-3456-7890123-00`, four-character IBAN blocks, and zero-width characters
riding along in copied text), and `accountNumberFormatError` returns the country's message for a
wrong-shaped value and `null` for a country with no format on record.

CI on head `832c58495` was green (GitHub Actions run
[30356885374](https://github.com/antiwork/gumroad/actions/runs/30356885374), 75 checks passed) —
that predates this delta, so it is not a claim about the current head. `mergeStateStatus` is
`BLOCKED` awaiting review, not a failing check.

## Screenshots (hosted preview app) — re-verified 2026-07-29 at head `b683d0a32fcf`

The evidence above was captured at code head `1f386446b277`. Two commits have landed since, and
neither is cosmetic:

```
$ git diff --stat 1f386446b..b683d0a32 -- . ':!qa-media'
 .../PaymentsPage/BankAccountSection.test.tsx     | 101 ++++++++++---
 .../Settings/PaymentsPage/BankAccountSection.tsx | 126 ++--------------
 app/javascript/pages/Settings/Payments/Show.tsx  |  31 ++--
 app/javascript/utils/payoutAccountNumbers.test.ts|  60 ++++++++
 app/javascript/utils/payoutAccountNumbers.ts     | 158 +++++++++++++++++++++
 app/models/qatar_bank_account.rb                 |  10 +-
 spec/models/qatar_bank_account_spec.rb           |  25 ++++
 spec/services/update_payout_method_spec.rb       |  21 +++
 8 files changed, 375 insertions(+), 157 deletions(-)
```

Three behavioural changes a screenshot can settle, so this is a re-capture:

1. **`maxLength` is gone from every account-number input.** A cap sized to the bare number silently
   swallowed the last characters of a number pasted in the grouped form banks print
   (`12-3456-7890123-00`) — and that form *saves fine*, because `UpdatePayoutMethod` strips
   separators before validating. The cap was making the hint worse than no hint for anyone who
   copied their number off a statement.
2. **The hints now actually ENFORCE.** The page posts through Inertia rather than submitting the
   form element, so the browser never ran `pattern` — the hint was decorative and the old generic
   `1234567890` still went out to Stripe. `accountNumberFormatError` re-checks the same pattern on
   Save and surfaces the country's own `title` string.
3. **Qatar's account-number check moved to `on: :create`.** The 115 rows stored while that
   validation was inverted (it only complained when the value was *blank*) would otherwise have
   been unable to `save!` at all, which is what `mark_deleted!` and the account-holder-name update
   both do — so those sellers could not have deleted, renamed, or replaced their bank account.

Captured 2026-07-29 against the live pod at **`x-revision: b683d0a32fcf`** (`curl -sI` on the
preview host), driving the real Settings → Payments form as `seller@gumroad.com`. Each country was
produced by writing a fresh `UserComplianceInfo` on the preview console and clearing any stored bank
account first, so `country_code` in the page props is the seeded value (logged per leg).

### 1. A separator-formatted number survives the field (`maxLength` removed)

Pasted, then read back off the live input — nothing truncated, and `maxlength` reads `null`:

| Country | Pasted | Field value after paste | `maxlength` | `pattern` |
|---|---|---|---|---|
| **NZ** | `12-3456-7890123-00` (18 chars) | `12-3456-7890123-00` | `null` | `[0-9]{15,16}` |
| **QA** | `QA87 CITI 1234 5678 9012 3456 7890 1` (36 chars) | same, whole | `null` | `[0-9A-Za-z]{29}` |
| **MZ** | `0012 3456 7890 1234 56789` (25 chars) | same, whole | `null` | `[0-9A-Za-z]{21}` |

Every one of those is longer than the bare number the model wants, so the previous `maxLength`
would have clipped all three.

| | |
|---|---|
| New Zealand, desktop | nz paste |
| New Zealand, mobile 375px | nz paste mobile |
| Qatar — note the label is **IBAN** / **Confirm IBAN** | qa paste |
| Mozambique — label stays **Account #**, not IBAN | mz paste |

The Qatar and Mozambique frames together are the asymmetry the PR body calls its one judgement
call, on the live form: QA renders an IBAN field, MZ keeps the account-number label.

### 2. Pressing Save with the old generic `1234567890` is now refused in the browser

The country's own instruction string appears in the page's alert region and the field is marked
`aria-invalid="true"` — no request goes out, no Stripe round-trip:

| Country | Message shown | field `aria-invalid` |
|---|---|---|
| **NZ** | *Enter your 15 or 16 digit account number, including the bank and branch digits* | `true` |
| **QA** | *Enter your 29-character IBAN, starting with QA* | `true` |
| **MZ** | *Enter your 21-character NIB, without the MZ IBAN prefix* | `true` |

| Country | The message, as rendered | The invalidated field |
|---|---|---|
| **NZ** | nz message | nz rejected |
| **QA** | qa message | qa rejected |
| **MZ** | mz message | mz rejected |

New Zealand at 375px, the same refusal on a phone:

nz rejected mobile

**Note on framing these two frames per country.** The payments page scrolls an inner container
rather than the window, and the alert region sits at the top of the form while the invalidated field
is ~1000px below it — so no single viewport (and no `fullPage: true`, which measures a document that
does not grow) can hold both. So each country gets two frames from the SAME
run — the `-message` frame with the scroll container pinned to the top showing the red banner, and
the `-rejected` frame showing the pink-outlined field still holding `1234567890`. Both message
strings were also read out of `[role="alert"]` programmatically and match the pixels.

### 3. The formatted number then saves for real

Re-entering the separator-formatted value and pressing Save clears the field error, and the row
lands server-side. Read back on the preview console after the Mozambique leg:

```
MARK bank_class="MozambiqueBankAccount"
MARK acct_visual="******6789" routing="AAAAMZMXXXX"
```

`******6789` is the last four of the **bare** 21-character NIB, so the separators were stripped
exactly as `UpdatePayoutMethod` intends — the browser check and the server agree.

| | |
|---|---|
| NZ accepted | nz accepted |
| QA accepted | qa accepted |
| MZ accepted | mz accepted |

### 4. The Qatar `on: :create` scoping, proven on a seeded bad row

This is the part with no rendered surface, so it was driven directly on the preview pod at the same
revision. A live Qatar account was given a value the regex rejects — the shape of the 115 rows
stored while the check was inverted — and then the two operations that `save!` an existing row were
attempted:

```
MARK class=QatarBankAccount visual="******8901"
MARK regex=/^[a-zA-Z0-9]{29}$/
MARK seeded_bad_visual="******8901"
MARK rename=OK holder="QA Renamed Holder"
MARK delete=OK deleted_at=2026-07-29 10:22:25 UTC
MARK new_bad_valid=false errors=["...", "The account number is invalid."]
MARK new_good_valid=false errors=["Account number last four We could not save your bank account information."]
```

Rename and delete both succeed on the stored bad row, while a **new** record carrying the same bad
value is still refused with *The account number is invalid.* — which is the whole point of the
scoping: create-time input is still checked, stored rows are not held hostage.

Honest limits on that last block: `new_good_valid=false` is **not** a rejection of the good IBAN. It
fails only on `Account number last four`, because this seller already holds an alive bank account —
the uniqueness-style guard, not the format check. The absence of *The account number is invalid.*
from that error list is the signal, and it is present in the bad-value case directly above it. The
seeded corrupt Qatar row was left mark-deleted on the preview.

### What is still not captured

- **North Macedonia**, the other IBAN reclassification. Covered by the vitest table (`MK` asserts
  the `IBAN` label and a 19-character pattern) and by the `pattern`/label mechanism the Qatar frame
  demonstrates on the live form, but not photographed.
- The remaining 14 countries in `COUNTRY_ACCOUNT_NUMBER_HINTS` individually. Every one is asserted
  by the vitest table; the three legs here were chosen because each proves a different half of the
  delta (NZ = the separator case the cap broke, QA = the IBAN reclassification plus the model
  change, MZ = the label that deliberately did *not* change, plus the real server-side save).

### Re-verified after merging `main` (head `56ebc6f25`)

The head moved from `674594f0d` to `68489196f` by a single `Merge remote-tracking branch
'origin/main'` — no commit of this PR's own. The PR's diff against its merge-base is
**byte-identical** across the two heads:

```
$ mbo=$(git merge-base origin/main 674594f0d); mbn=$(git merge-base origin/main 68489196f)
$ git diff $mbo..674594f0d -- ':!qa-media' | shasum   # 82cfa558b468...
$ git diff $mbn..68489196f -- ':!qa-media' | shasum   # 82cfa558b468...
```

So every frame above still stands. What main *did* bring in is worth one live check anyway:
[#6511](https://github.com/antiwork/gumroad/pull/6511) renamed the accent token to
`--accent-with-text` across `Button.tsx` / `Checkbox.tsx`, and
[#6476](https://github.com/antiwork/gumroad/pull/6476) touched
`UpdateUserComplianceInfo` — the service on the other side of this very form. Both could
plausibly have broken the payments page or the refusal path without any change to this branch.
They did not.

Re-shot against the live pod at **`x-revision: 68489196f779`**, Mozambique seller,
`seller@gumroad.com`, bank account cleared on the preview console first:

| Read off the live inputs | Value |
|---|---|
| `placeholder` | `001234567890123456789` |
| `pattern` | `[0-9A-Za-z]{21}` |
| `title` | *Enter your 21-character NIB, without the MZ IBAN prefix* |
| `maxlength` | `null` |

Pressing **Update settings** with the old generic `1234567890`: the banner renders, the field
reads `aria-invalid="true"`, and **no non-GET `/settings/payments` request is issued** — the
refusal is still client-side, no Stripe round-trip. Zero `pageerror`s and zero console errors on
the surface.

| | |
|---|---|
| Desktop | postmerge desktop |
| Mobile 375px | postmerge mobile |

**Control.** The same script, same seeded Mozambique seller, against
`fix-dispute-evidence-backstop-sw` (revision `92645ad97d77`, a branch that does **not** contain
`app/javascript/utils/payoutAccountNumbers.ts` — verified with `git show`). There the field still
advertises the generic `1234567890` with `pattern: null` / `title: null`, and `1234567890` is
never refused on format; the page complains only about unrelated required fields
(*Address, City, Postal code, Date of birth*). That is the "before" this PR removes:

control, no refusal

### Re-verified after merging main (head `6c703ab4c`)

The head moved only by merging `origin/main` (8 commits); the branch's own diff against its
merge-base is **byte-identical**, and main brought in nothing touching this surface:

```
$ git diff --stat 56ebc6f25..6ef8d5397 -- app/javascript/utils/payoutAccountNumbers.ts \
    app/javascript/components/Settings/PaymentsPage app/javascript/pages/Settings/Payments app/models/qatar_bank_account.rb
$                       # no output

$ git diff --stat <old merge-base> <new merge-base> -- app/javascript/components/Settings/PaymentsPage \
    app/models/qatar_bank_account.rb app/services/update_user_compliance_info.rb
$                       # no output
```

"Identical diff" is not the finding though. Two things main merged in *could* have broken this page
with zero change to the branch — #6511's `--accent` → `--accent-with-text` token rename across
`Button.tsx`/`Checkbox.tsx`, and #6476 touching `UpdateUserComplianceInfo`, the service behind this
very form. Neither fired: the page renders, the hint attributes are intact, and the refusal path
still behaves. Re-read off the live DOM at the merged head (`x-revision: 6ef8d53975e2`), not
eyeballed:

| Invariant on the `Account #` input | Live value |
|---|---|
| `placeholder` | `001234567890123456789` |
| `pattern` | `[0-9A-Za-z]{21}` |
| `title` | `Enter your 21-character NIB, without the MZ IBAN prefix` |
| `maxlength` | `null` — deliberately absent, so a separator-formatted paste isn't truncated |
| label | `Account #` (not `IBAN`) — correct for MZ |

The `Confirm account #` input carries the same three attributes, which matters because the Save
check compares the pair.

**The refusal, asserted as a negative.** Typing the old generic-hint value `1234567890` (10 chars)
and pressing **Update settings** marks the field `aria-invalid="true"` — and the network log shows
**zero non-GET requests to `/settings/payments`**. That is the part worth a reviewer's time: the
rejection never reaches the server, so a seller gets it immediately instead of after a round trip.
Re-running with the model-accepted 21-character `001234567890123456789` leaves the account-number
field **not** invalid, so the check discriminates rather than always refusing.

| Desktop 1280px — `1234567890` refused, red border, `Confirm` untouched | Mobile 375px, same refusal |
|---|---|
| mz refusal | mz refusal mobile |

Accepted value, same form: mz accepted

**Scope of this re-verification — one country, not all eight.** This preview's
`seller@gumroad.com` already has `UserComplianceInfo.country = MZ`, and that model is append-only,
so driving the page's Country select to Morocco/Senegal/Serbia/Moldova/Gambia/Qatar/New Zealand
does not re-render the payout form against those countries — every switch reported back as
Mozambique. The full eight-country matrix stays evidenced by the earlier captures pinned above,
whose code is unchanged; what is re-proven at the merged head is the MZ row end to end. The
`Please enter your full phone number…` alert visible in the run is the page's unrelated
phone requirement, not this PR's check.

## QA steps

Preview app (`preview` label applied):

1. Settings → Payments, set country to **Mozambique**. Confirm the account-number box hints a 21-character value and the label still says account number (not IBAN).
2. Paste the old generic `1234567890` and press **Update settings** — the page should refuse it
   immediately with *Enter your 21-character NIB, without the MZ IBAN prefix* and no request should
   go out. (The `pattern` attribute alone never fires here: the page posts through Inertia rather
   than submitting the form element.)
3. Paste a valid 21-character NIB — should accept and save.
3b. Paste it in the grouped form a bank prints it, e.g. `0012 3456 7890 1234 56789` — the field must
   keep every character (no `maxLength` truncation) and it must still save, because
   `UpdatePayoutMethod` strips the separators before validating.
3c. As a Qatar seller who already has a bank account stored, rename the account holder and then
   switch payout method away from bank — both must work even if the stored account number does not
   match the current format, which is why that validation is `on: :create`.
4. Switch to **Qatar**, then **North Macedonia**: both should now render an **IBAN** field with an IBAN-shaped hint.
5. Switch to **Nigeria** and **Philippines** (controls) and confirm nothing changed for either.

## Risk

Client-side hint/validation only plus two countries changing which field label they render. It cannot cause a payout to be sent anywhere new — a wrong-shaped value was already rejected server-side; this just moves the rejection earlier. The two IBAN reclassifications are the part that changes what a seller can save at all, and both are pinned by our own pre-existing regexes (29 chars for QA, 19 for MK — lengths only the full IBAN reaches) plus the IBAN registry, with Stripe's documented example data agreeing. Mozambique's 21-character requirement can only be the bare NIB, since the MZ IBAN is 25.

---

## AI disclosure

The model was asked to finish what #6357 started: find every non-IBAN payout country whose account-number placeholder its own bank-account model would reject, and give each a real example plus matching `maxLength`/`pattern`. It was told to verify IBAN-vs-local shape against the live Stripe bank directory rather than guessing, and to include control countries in the test so an over-broad condition would fail loudly. It enumerated the countries from the bank-account models, checked each against Stripe, wrote the component change and the 25 tests, produced the revert-based mutation proof above, and ran full branch CI before this PR was opened.


gumclaw:
Greptile's review came back 5/5 with no inline findings, and its own T-Rex run reproduced the mutation proof from the body independently: 22 of 25 assertions fail against the pre-change component, 25/25 pass after. Nothing to fix from the review itself.

I did close out the one loose end the body flagged: the three QA screenshots were still named `pr-6503-*` from an earlier draft. They are now `pr-6536-*` and the body points at the renamed files on the current head.

One thing worth a reviewer's eye that the summary glosses over: Mozambique deliberately stays out of `IBAN_SHAPED_COUNTRY_CODES` while Qatar and North Macedonia move into it. Stripe wants the bare 21-character NIB for Mozambique and rejects the full MZ IBAN, so the asymmetry is intentional rather than an oversight in the reclassification.

gumclaw:
@ershad routing this to you as the single human owner — this is payout-account/bank-detail territory, and the one judgement call in it is a finance one: Qatar and North Macedonia move into `IBAN_SHAPED_COUNTRY_CODES` while Mozambique deliberately does not, because Stripe wants the bare 21-character NIB there and rejects the full MZ IBAN

It was opened ~45 minutes ago assigned to me (the bot) with no reviewer requested and only the `working` label, which meant it was sitting green and finished but invisible in everyone's review queue. Fixed: you're the assignee, review requested, `working` swapped for `awaiting-human`.

State: CI green, no failing checks, mergeable with no conflicts. Nothing is outstanding on my side — the remaining step is a human read. No auto-merge armed; the merge press is yours. Happy to re-route if this isn't your surface.

gumclaw:
Found a P1 in my own last commit while double-checking it against production, and pushed the fix in b683d0a32.

**What was wrong.** That commit fixed Qatar's inverted `validate_account_number` (it only added an error when the value was *blank*, so any present value bypassed the 29-character format). Correct fix, wrong scope: the check then ran on **every** save, not just on new input.

**Why that matters.** Production holds **115 live Qatar bank accounts whose stored number does not match the format** — they exist precisely because the broken guard accepted anything non-blank. Both `mark_deleted!` and the account-holder-name update call `save!` on the *existing* row, so those 115 sellers could no longer delete, rename, or replace their bank account: every attempt would come back "The account number is invalid." about a number they cannot see or edit. Replacing a bank account deletes the old one first, so the fix would have locked them out of correcting it. Worse than the bug being fixed, and aimed at sellers who did nothing wrong.

I checked the split rather than assuming: all 115 are newly blocked by this check alone — **zero** were already failing the pre-existing bank-code guard. Shape of the stored values is what you'd expect from an unvalidated field (37 at 13 chars, 23 at 16, 8 non-alphanumeric at 36, the rest scattered).

**The fix.** `validate :validate_account_number, on: :create`. Every path that accepts an account number from a seller builds a new record (`UpdatePayoutMethod#process_full_bank_account_replacement`), so wrong input is rejected exactly as before — only saves of already-stored rows are spared.

Mutation-verified: the new spec asserts a legacy-shaped row can still be renamed and deleted. It fails on the previous commit (`Expected false to equal true`) and passes now. 43 rspec examples green across the Qatar, North Macedonia, Mozambique, New Zealand and `UpdatePayoutMethod` specs; rubocop clean. I also swept `app/models/*.rb` for the same inverted-guard pattern — Qatar was the only one.

Worth noting the general trap, since this PR is entirely about tightening validation: **fixing a validator that was silently accepting bad data means existing rows now fail it**, and if that validator runs on update, the population it fails is exactly the population the bug created. The reads behind the numbers above are audited (`20260729T094106Z-18548c6e`, `20260729T094153Z-cd4f8e17`, `20260729T094458Z-3a04e007`).

@ershad this doesn't change what you're reviewing — Qatar and North Macedonia moving into `IBAN_SHAPED_COUNTRY_CODES` is still the judgement call I'd like your eye on. Re-running CI on the new head now.

gumclaw:
Ran a second adversarial review round on the delta since `832c58495`, since the round-2 fixes and the `on: :create` change had never been looked at by anyone. Verdict came back **READY-TO-MERGE**, and the sharpest question in it — is `on: :create` bypassable? — is settled: the only path that assigns a seller-supplied account number is `UpdatePayoutMethod#process_full_bank_account_replacement`, which news up the record at line 205 and validates it in create context at 211. Every other path that touches an existing row (`mark_deleted!` via payout-method switch, GDPR erasure, account closure, the Stripe bank-gone cleanup, and the holder-name rename) supplies no account number.

I took the four non-blocking findings anyway, in `340e37165`, because one of them was the same class of defect that opened this branch.

**The enforcement had no test.** Round 1 caught that the `pattern` attributes never executed, because this page saves through Inertia rather than submitting the form. The fix was to call the shared check from the page's own `validateBankAccountFields` — and nothing tested that call. The util is tested in isolation, the component tests assert the `pattern` *attribute*, and there is no test file for the page at all. Deleting one line put the field back to hinting a format nothing checks, with every test in the repo still green. There is now a Mozambique request spec that fills in the generic `1234567890`, clicks Update settings, and asserts both the seller-facing message and that no bank account was created.

**Gibraltar and Oman were still decorative.** They kept their formats in separate branches in the component rather than in the shared table, so they never got the new enforcement, and they made two comments false — the table called itself the one place these formats live, and the page said countries without an entry fall through to the server. Both are in the table now. Worth flagging one asymmetry: Oman's model accepts 6 to 16 digits, so it is the single country here whose model genuinely accepts `1234567890`. Its entry exists to reject a non-numeric value, not to replace a wrong example, so it is excluded from the assertions that the generic example is rejected and pinned separately.

**One character of separator-stripping diverged from the server.** U+0085 is whitespace to Ruby's `[[:space:]]` but not to JavaScript's `\s`, so a pasted number carrying one saved server-side while this check rejected it with a misleading message. Now stripped on both sides and pinned.

**The `on: :create` comment overstated the damage.** It claimed those 115 sellers could not delete, rename, *or replace* their bank account. Replacing always worked — that path deletes the old row with `save!(validate: false)`. What was actually blocked was renaming and deleting, including via payout-method switch, GDPR erasure, and account closure. Corrected in the model and the spec, since that comment is the justification for the riskiest line in the diff and it should survive an audit. The no-`maxLength` rationale was also written out three times across two files; it is stated once now.

118 vitest and 23 rspec examples green, rubocop clean, and the one eslint error in `Show.tsx` is pre-existing (confirmed against a stashed baseline — it comes from #6361).

One thing I could not verify locally and am not going to claim: `payments_spec.rb` does not boot on this machine, and the untouched Mozambique example next to my new one fails identically at `visit`, so CI is the real gate for that spec rather than a formality. Watching it now.

@ershad nothing here changes what I'm asking you to look at — the finance judgement call is still Qatar and North Macedonia moving into `IBAN_SHAPED_COUNTRY_CODES` while Mozambique deliberately stays out. Auto-merge remains unarmed since this is a payouts path.

gumclaw:
## Merged `main` in to clear a Minitest failure that was not from this branch

The push run on `340e37165` failed `Test Minitest 1` — one failure out of 1597 runs:

```
LinksControllerSaveContractTest#test_flag_on:_a_partial_deletion_keeps_the_grouping_and_its_name
[test/controllers/links_controller_test.rb:6712]: Expected false to be truthy.
```

Line 6712 is `assert @category.reload.alive?` — a partial version deletion had destroyed the
whole grouping. That is seller data loss, so I chased it rather than re-running until green.

**It is not this branch's code, and it is already fixed.** The failing test lives in
`links_controller_test.rb`, which this PR does not touch. The cause is the external-id
collision I fixed on `main` three hours ago in #6545: `variants` and `variant_categories` are
separate tables, so a version and a grouping whose primary keys line up share the same
`external_id` string — and a request naming one version could sweep an unrelated grouping.
That is exactly the intermittent red this test was showing.

This branch was 38 commits behind `main` and so did not have the fix. Merged `main` in
(no rebase, no force) — `68489196f`. The merge changed none of the six files this PR touches.

Attribution evidence, not a flake label:

- Repro'd with CI's own seed (`--seed 62665`) on the whole file, plus the `SaveContract` class
  3x and the single test 3x, all under private MySQL + Redis isolation: **0 failures every
  run.** Passing locally is consistent with a cross-file collision inside a 1597-run shard.
- `git merge-base --is-ancestor 8900363d2 674594f0d` → false before the merge, true after.
- The subsequent run on `674594f0d` was fully green, including every Minitest shard — same
  Ruby, different result, which is the collision's signature.

Nothing about the payout account-number work changed here. Still assigned to `@ershad` for the
finance call, and auto-merge stays unarmed — this is a payouts path.

gumclaw:
@ershad automated review-SLA nudge: this PR has been waiting 28h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

ershad:
## Account-number format verification (QA / MK / MZ)

Verified the one open judgement call — Qatar and North Macedonia moving into
`IBAN_SHAPED_COUNTRY_CODES` while Mozambique stays out — against two independent public sources.
The classification is correct.

**Stripe's bank-account format table** ([docs](https://docs.stripe.com/connect/payouts-bank-accounts))
documents these example account numbers:

| Country | Stripe's example | What it is |
|---|---|---|
| Qatar | `QA87CITI123456789012345678901` (29 chars) | the full QA IBAN |
| North Macedonia | `MK49250120000058907` (19 chars) | the full MK IBAN |
| Mozambique | `001234567890123456789` (21 digits, no `MZ` prefix) | the bare NIB, not an IBAN |

**The IBAN registry** ([formats by country](https://en.wikipedia.org/wiki/International_Bank_Account_Number))
settles the ambiguity: QA IBAN = 29 chars, MK IBAN = 19 chars, MZ IBAN = **25** chars
(`MZ` + 2 check digits + 21-digit BBAN). So Stripe's 21-digit Mozambique example is definitively
the bare local number, while the QA and MK examples are exactly their full IBAN lengths — an IBAN
is the only value that fits what Stripe documents for those two.

Two corroborating details:

- The PR's placeholders are character-for-character Stripe's documented example data (all three
  strings above appear verbatim in the diff), as is the SWIFT example in the request spec.
- Our own pre-existing models force the same conclusion: `QatarBankAccount` requires exactly
  29 alphanumerics and `NorthMacedoniaBankAccount` exactly 19 — only the full IBAN can satisfy
  those (the bare BBANs are 25 and 15 chars) — and `MozambiqueBankAccount` requires exactly 21,
  which the 25-char MZ IBAN never fit. The relabel just tells sellers which value the form was
  always demanding.

gumclaw:
Thanks for taking the IBAN registry route on this — the length arithmetic is the part I couldn't settle from Stripe's table alone, and it closes the question independently of what I inferred from the live bank directory.

Your third corroboration is the one I'd lean on hardest, because it's ours rather than a vendor's. Re-read at head `6c703ab4c`:

| Model | Regex | Full IBAN | Bare BBAN |
|---|---|---|---|
| `QatarBankAccount` | `/^[a-zA-Z0-9]{29}$/` | 29 ✓ | 25 ✗ |
| `NorthMacedoniaBankAccount` | `/\A[a-zA-Z0-9]{19}\z/` | 19 ✓ | 15 ✗ |
| `MozambiqueBankAccount` | `/^([0-9a-zA-Z]){21}$/` | 25 ✗ | 21 ✓ |

Only the IBAN can satisfy QA and MK, and only the bare NIB can satisfy MZ. So the asymmetry isn't a call this PR is making — it's the shape those validations have always demanded, and the relabel just stops hiding it from the seller. That's a stronger footing than the one the body claims, and I've replaced the body's "verified against the live Stripe bank directory" framing with this, since a directory lookup is a weaker source than our own regexes plus the registry.

State at head: 74 checks passed, 1 skipped, none failing ([run 30485630790](https://github.com/antiwork/gumroad/actions/runs/30485630790)), `MERGEABLE` / `CLEAN`, and the head hasn't moved since the post-merge re-capture. Auto-merge is deliberately unarmed — bank-detail surface, so the merge is yours, not something I'll take on green.

The gap I'd still name rather than let pass quietly: North Macedonia is the only reclassified country with no live-form frame. It's asserted in the vitest table (IBAN label, 19-character pattern) and rides the same code path Qatar is photographed on, but the preview seller's `UserComplianceInfo` is append-only at MZ, so I couldn't drive the form to MK without seeding another seller. Say the word if you want that frame before you merge and I'll seed one and shoot it.

ershad:
## Local verification — Mozambique flow at head `6c703ab4c`

Ran the Mozambique leg end to end on a local checkout of this branch (`localhost:3000/settings/payments`, seller's compliance country set to Mozambique on the console). All three behaviours reproduce:

**1. The hints render.** Both the account-number field and its confirmation carry:

| Attribute | Value |
|---|---|
| label | `Account #` (not IBAN — the deliberate asymmetry) |
| `placeholder` | `001234567890123456789` (Stripe's documented example) |
| `pattern` | `[0-9A-Za-z]{21}` |
| `title` | *Enter your 21-character NIB, without the MZ IBAN prefix* |
| `maxlength` | absent |

**2. The old generic value is refused before any request goes out.** With `1234567890` in both fields and the compliance fields filled, Update settings rendered the red banner with the NIB instruction, marked the field `aria-invalid="true"`, and the network log shows zero requests to `/settings/payments` — the rejection is immediate, no Stripe round-trip.

**3. A separator-formatted NIB saves for real.** Pasting `0012 3456 7890 1234 56789` (25 chars) kept every character — the removed `maxLength` would have clipped it at 21 — passed the Save check, and the PUT returned 200. Read back on the console: `MozambiqueBankAccount`, account-number visual `******6789` (last four of the bare NIB, so the separators were stripped server-side exactly as `UpdatePayoutMethod` intends), routing `AAAAMZMXXXX`. Matches the preview-app QA in the PR body.

gumclaw:
Thanks for driving the Mozambique leg locally — that closes the last gap in the evidence. Your three findings match what the preview captures showed, and the read-back (`MozambiqueBankAccount`, `******6789` off the bare NIB, routing `AAAAMZMXXXX`) is the server-side half I wanted a second pair of eyes on: separators stripped by `UpdatePayoutMethod`, no truncation, and the refusal happening with zero requests out.

On deploy state: merged at 13:13:50 UTC, and the latest release (`v2026.07.30.13`) was cut at 13:12:32 — a minute before. So this is on main but not yet in production; it ships with the next release. Nothing to do, just flagging that the hints are not live for sellers quite yet.

PR #6539, opened 2026-07-29
Bound related-products counter writes so collection can be re-enabled

gumclaw:
## What

Makes the related-products counter writes bounded, so the `update_sales_related_products_infos` feature flag can be re-enabled. Fixes the first three checkboxes of gumroad-private#1406.

- `SalesRelatedProductsInfo.update_sales_counts` now uses batched `INSERT ... ON DUPLICATE KEY UPDATE` upserts keyed off the unique `(smaller_product_id, larger_product_id)` index, replacing the read-then-update-then-insert sequence. Every statement touches **at most 100 rows** by primary-key-equivalent lookup.
- `UpdateSalesRelatedProductsInfosJob` caps per-purchase fan-out at the buyer's **100 most recently purchased products**.

## Why

The job was disabled via the feature flag after its writes stalled replication on `production-worker-replica-1` for hours (gumroad-private#1353). Two things made a single purchase arbitrarily expensive:

1. **`update_sales_counts` ran an unbounded `UPDATE ... WHERE id IN (...)`.** It first `SELECT`ed the matching pair ids, then updated all of them in one statement. The `in_batches` call did *not* bound it: on a relation already pre-scoped by an id list, `in_batches` **re-ships the full list per batch** instead of splitting it. Under `binlog_format = MIXED` every replica re-executes that statement serially — which is the replication stall.
2. **The job walked the buyer's entire purchase history.** A sale to a buyer who owns N products touched N counter rows and enqueued N+1 cache jobs, with no bound on N.

**Why capping at 100 does not degrade recommendations:** the read path already discards precision beyond a top-N slice, so products past the 100th most recent purchase could never have surfaced anyway. The cap removes work whose output was already being thrown away.

**Not covered here** (gumroad-private#1406 tracks it): moving counter maintenance fully off the per-purchase path onto a buffered, scheduled flush, and the decision about `binlog_format = ROW`. **Re-enabling the flag stays a deliberate operator action after this ships** — this PR makes re-enabling safe, it does not re-enable anything.

## Before / After

Not applicable as a UI change — this is the write path behind product recommendations, with no rendered surface of its own. The recommendation lists a buyer sees are unchanged by design (see the top-N argument above); the change is in how many rows and statements one purchase generates.

## Test Results

`spec/models/sales_related_products_info_spec.rb` and `spec/sidekiq/update_sales_related_products_infos_job_spec.rb` cover the upsert path (including the duplicate-key update branch, which is the whole point of the rewrite), the 100-row batch bound, and the 100-product fan-out cap.

Green on head `bd1dd2c7a`: GitHub Actions run [30376188387](https://github.com/antiwork/gumroad/actions/runs/30376188387), 75 checks passed, 0 failing, 0 pending.

## Performance

The claim this PR rests on is a **statement-shape** change, which is what the replication stall was caused by:

- **Before:** one `SELECT` of all matching pair ids, then one `UPDATE ... WHERE id IN (<unbounded list>)`, then inserts for the misses. Statement count is O(1) but the single UPDATE's row count is unbounded, and each replica re-executes it serially under MIXED binlog format.
- **After:** batched upserts of at most 100 rows each, every row located by the unique `(smaller_product_id, larger_product_id)` index. Worst case per purchase is now bounded by the fan-out cap (100 products) rather than by the buyer's lifetime purchase count.

⚠️ Honest limitation: I have **not** reproduced the replication stall itself, because it requires a replica topology under MIXED binlog format with production-scale row counts — that is not reproducible in the local test environment, and I did not run this against a production replica. The bound is demonstrated structurally (batch size and fan-out cap asserted in the specs) rather than by a measured replica-lag delta. **Confirming the lag actually goes away is the operator's step when the flag is re-enabled**, and it is the main thing I would want watched during that ramp.

## QA steps

Preview app (`preview` label applied). This is a background job, so verification is a console path:

1. Create a buyer who owns several products, then create one new purchase.
2. Run `UpdateSalesRelatedProductsInfosJob.new.perform(purchase.id)` with the query log on.
3. Confirm every counter statement is an upsert touching ≤100 rows and that no `UPDATE ... WHERE id IN (...)` with a large list appears.
4. Give the buyer >100 purchased products and re-run; confirm the fan-out stops at 100 and the enqueued cache-job count is bounded accordingly.
5. Confirm the resulting counter values match what the old path would have produced for the products inside the cap.

## Risk

Deliberately **not** self-arming: the feature flag stays off, so merging this changes nothing in production until an operator re-enables it. That is the right shape given the change is a response to a replication incident. The correctness risk is in the upsert's duplicate-key branch producing a different counter value than the old read-modify-write did, which is asserted directly in the model spec.

---

## AI disclosure

The model was asked to make the related-products counter writes bounded enough that the flag disabled after the gumroad-private#1353 replication incident could be re-enabled, and to state plainly what it could not verify. It found the non-obvious cause — that `in_batches` does not split a relation already scoped by an id list, so the "batched" update was shipping the full list every time — rewrote the path as index-keyed batched upserts, capped the per-purchase fan-out, wrote the specs, and ran full branch CI. It explicitly did not claim a measured replication-lag improvement, because the replica topology needed to produce that number was not available.

### DIFF
--- a/app/models/sales_related_products_info.rb
+++ b/app/models/sales_related_products_info.rb
@@ -18,40 +18,48 @@ class SalesRelatedProductsInfo < ApplicationRecord
     retry
   end
 
+  # How many (smaller_product_id, larger_product_id) pairs a single statement may touch.
+  # Keeping statements small matters beyond the primary: production replicates with
+  # binlog_format = MIXED, so a deterministic multi-row statement is re-executed verbatim
+  # on every replica by a single applier thread. One oversized statement once stalled
+  # replication for hours (see gumroad-private#1353), which is why this path was disabled.
+  UPDATE_SALES_COUNTS_SLICE_SIZE = 100
+
   def self.update_sales_counts(product_id:, related_product_ids:, increment:)
     return if related_product_ids.empty?
     raise ArgumentError, "product_id must be an integer" unless product_id.is_a?(Integer)
     raise ArgumentError, "related_product_ids must be an array of integers" unless related_product_ids.all? { _1.is_a?(Integer) }
 
-    now_string = %("#{Time.current.to_fs(:db)}")
+    now_string = connection.quote(Time.current.to_fs(:db))
+    # A pair we've never seen starts at 1 for an increment. For a decrement (e.g. a refund)
+    # it starts at 0, not -1: a refund for a pair we never counted means the pair simply has
+    # no sales. This mirrors what the previous read-then-update-then-insert code did.
+    new_row_sales_count = increment ? 1 : 0
     sales_count_change = increment ? 1 : -1
 
-    # Update existing
-    larger_ids, smaller_ids = related_product_ids.partition { _1 > product_id }
-    existing = where(smaller_product_id: product_id, larger_product_id: larger_ids)
-      .or(where(smaller_product_id: smaller_ids, larger_product_id: product_id))
-      .pluck(:id, :smaller_product_id, :larger_product_id)
+    # Deduplicate before upserting: a duplicate id in the input must not double-count.
+    # The old implementation got this for free (a second INSERT IGNORE for the same pair
+    # was a no-op), so we preserve that behavior explicitly here.
+    ids = related_product_ids.uniq - [product_id]
 
-    where(id: existing.map(&:first))
-      .in_batches(of: 1_000)
-      .update_all("sales_count = sales_count + #{sales_count_change}, updated_at = #{now_string}")
+    # A single upsert per slice replaces the old SELECT ids -> UPDATE ... WHERE id IN (...)
+    # -> INSERT IGNORE sequence. The upsert is keyed off the unique index on
+    # (smaller_product_id, larger_product_id), which closes the read-then-write race and,
+    # more importantly, removes the unbounded `id IN (...)` UPDATE whose replication cost
+    # is what got this collection path disabled (gumroad-private#1353 / #1406).
+    ids.each_slice(UPDATE_SALES_COUNTS_SLICE_SIZE) do |ids_slice|
+      values_sql = ids_slice.map do |related_product_id|
+        smaller_id, larger_id = [product_id, related_product_id].minmax
+        "(#{smaller_id}, #{larger_id}, #{new_row_sales_count}, #{now_string}, #{now_string})"
+      end.join(", ")
 
-    # Insert remaining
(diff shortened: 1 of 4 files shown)


gumclaw:
Closing this in favour of [#6543](https://github.com/antiwork/gumroad/pull/6543), and the reason is worth recording rather than leaving as a bare "superseded" — **merging this PR would have reverted work that is already on main.**

I opened this ~20 minutes ago from a branch that had been sitting green and PR-less for 16 hours. I checked that CI was green and that the branch was mine, but I did not check whether main had moved underneath it. It had.

**What happened.** This branch was cut before [#6507 *Bound and coalesce the related-products counter writes*](https://github.com/antiwork/gumroad/pull/6507) merged. #6507 landed the same `SalesRelatedProductsInfo.update_sales_counts` rewrite — batched `INSERT ... ON DUPLICATE KEY UPDATE` keyed off the unique pair index — **and landed a better version of it than this branch has.** Main's implementation additionally explains the race that `INSERT IGNORE` used to lose (two concurrent purchases of the same pair both seeing it missing, the losing insert's increment silently dropped) and carries a `SALES_COUNT_UPSERT_BATCH_SIZE` spec pinning statement bounds.

Because this branch's version of that file is *older*, GitHub reports it `CONFLICTING`, and a careless conflict resolution in this PR's favour would have quietly undone #6507's improvements while looking like a clean merge of a green branch. The model file here is genuinely obsolete — not merely redundant.

**What was actually still unshipped.** Exactly one thing, in the *other* file: the per-purchase fan-out cap in `UpdateSalesRelatedProductsInfosJob`. #6507 bounded each individual SQL statement, but the *number of pairs one purchase generates* was still unbounded — a buyer who owns N products still made one sale touch N counter rows and enqueue N+1 cache-refresh jobs. That is the second half of the gumroad-private#1353 burst and it is still on main today.

So [#6543](https://github.com/antiwork/gumroad/pull/6543) is that commit and only that commit, cut fresh from current `origin/main`: the `RELATED_PRODUCTS_PER_PURCHASE_LIMIT = 100` cap plus its spec, no model file touched, no conflict.

**Generalising, since I caused this:** a branch being green is not evidence that it is still correct. Green means it passed against the main it was cut from. Before opening a PR from any branch that has been idle, the check is `git diff origin/main <branch> -- <each file>` against what has since merged — "green at head" and "still wanted" are different questions, and this branch was the first while not being the second.

PR #6541, opened 2026-07-29
Let a non-resident business owner get past the country screen

gumclaw:
## What

Lets a non-resident owner of a registered business get past the country-selection screen, and reframes the attestation checkbox around **documentary proof** rather than bare residence.

The line now reads *"I can provide proof of residence in the country above, or ..."* instead of *"I live in the country above"*.

## Why

The modal previously required an individual to attest that they *live* in the selected country. Two problems:

1. **A registered business owner who is not resident in the business's country had no truthful path through the screen at all** — the attestation was about the person, while the Stripe account is about the entity.
2. **Stripe may later require proof of address.** An individual who genuinely lives somewhere but cannot document it could pass this modal and then stall during verification. Wording the attestation around provable residence surfaces that requirement now.

The documentary-proof framing is not new copy invented here — it is the framing the **old** line carried and that a first draft of this change had dropped. The pre-merge adversarial review caught the regression (P3), and it is restored.

## Before / After

| Viewport / theme | Before | After |
|---|---|---|
| Desktop light | before desktop light | after desktop light |
| Desktop dark | before desktop dark | after desktop dark |
| Mobile light | before mobile light | after mobile light |
| Mobile dark | before mobile dark | after mobile dark |

Re-verified in the local app at head `1c199c412` on 2026-07-29. The modal rendered the two attestation lines shown in the After column in all four combinations: 1440px and 375px, light and dark. The four After screenshots were re-captured from that running app and all images above are pinned to the tested head.

## Test Results

`spec/requests/settings/payments_spec.rb` now pins the every-item gate in the existing country-selection flow: checking only the first attestation keeps Save disabled; checking both enables it.

- `RBENV_VERSION=3.4.3 rbenv exec bundle exec rspec spec/requests/settings/payments_spec.rb:6397` — 1 example, 0 failures.
- `RBENV_VERSION=3.4.3 bin/test-confidence` — 99% confidence; the two 99%-tier specs passed in 294 seconds. The tool classified two broader-suite failures as pre-existing and skipped them.
- `/Users/yaco/.codex/skills/autoreview/scripts/autoreview --mode local` — no accepted or actionable findings (0.98 overall).

## QA steps

Local or preview app:

1. Settings → Payments as a seller who has not yet selected a country. Open the country-selection modal.
2. Read the attestation line — it must offer the **provable-residence or registered-business** wording, not bare "I live in the country above".
3. Check only the first item and confirm Save remains disabled.
4. Check both items and confirm Save becomes enabled.
[...]

PR #6544, opened 2026-07-29
Write code comments for maintainers, not newcomers

gumclaw:
## What

Rewrites the "Code comments" convention in `AGENTS.md` to target a reader who already knows this codebase, and trims the two worst comments the old wording produced.

## Why

The current text tells agents to write for a reader *new* to the codebase and to "spell out non-obvious context". Taken literally — which is how an agent takes it — that produces essays. The United Internet mail-routing change landed a 22-line comment on a single constant, four lines of which were a 2010 corporate acquisition. Nobody maintaining that file needs the acquisition history to change a domain list.

Comments are for a colleague, not a student. The one thing the code cannot say is worth writing; the rest is noise, and noise is what teaches people to skip the comments that matter.

## Before / After

`MailerInfo::UNITED_INTERNET_RECIPIENT_DOMAINS` — 22 comment lines became 10. Everything a maintainer needs survives (the 550 r1102 rejection, why exact-match not suffix, why the vanity domains are deliberately absent); the corporate history and the restatement of the constant's own name do not.

`ApplicationMailer#redirect_united_internet_recipients_to_sendgrid!` — 14 lines became 5, keeping both load-bearing facts: it runs post-build so recipients are known, and before `set_custom_headers` because that header derives from the SMTP address.

## Test Results

`45 examples, 0 failures` locally across `mailer_info` and `application_mailer` specs. Comments only — no behaviour change, and the diff contains no non-comment lines.

CI: **green at head `2472c1ab1`** — 78 checks success, 0 failing, `ci/green` = `success`.

CI note (corrected): an earlier revision of this body claimed `Test Slow 11` was red on an unrelated UPI Payment Element Capybara spec (`upi_element_billing_prefill_spec.rb:53`) and that the failure was pre-existing on `origin/main`. **That was wrong.** Re-running only the failed job at the identical commit, with no code change, returned `SUCCESS` — so it was a flake (a Capybara wait against the Stripe Payment Element iframe), not a pre-existing breakage. Retained here rather than deleted so the record shows the correction.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,4 +6,13 @@ Project skills live in `.agents/skills/<name>/SKILL.md`. The `.claude/skills/<na
 
 ## Code comments
 
-Write comments for a reader who is new to the codebase but familiar with the goal of the project. Avoid jargon-dense shorthand: explain *why* the code does what it does in plain language, spell out non-obvious context (invariants, gotchas, links to the decision), and don't assume the reader knows internal nicknames, prior incidents, or module history. A good test: someone on day one who knows what the product does should understand the comment without grepping elsewhere.
+Write comments for a reader who already knows this codebase. They are a colleague, not a
+student — so say the one thing the code cannot say, and stop.
+
+Keep: the non-obvious *why*, ordering constraints, and traps that will bite the next person
+to edit this. Cut: vendor or company history, anything restating the line below it, narration
+of how the bug was found, and re-explanations of a named constant the reader can click
+through to. If a comment runs past about three lines, ask what you would delete.
+
+A good test: would a maintainer who knows this file find this line worth reading? If not, it
+is noise — and noise is what teaches people to skip the comments that matter.
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -32,20 +32,11 @@ class ApplicationMailer < ActionMailer::Base
   end
 
   private
-    # Safety net for the web.de / GMX policy block described in
-    # MailerInfo::UNITED_INTERNET_RECIPIENT_DOMAINS: those providers reject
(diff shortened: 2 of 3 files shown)


gumclaw:
CI is now green at the current head `2472c1ab1` — **78 checks success, 0 failing**, `ci/green` = `success`.

Correcting my own note in the body while I'm here, because it was wrong in a way worth naming. The body says `Test Slow 11` was red on an unrelated UPI Payment Element spec (`upi_element_billing_prefill_spec.rb:53`) and that I had verified it "fails identically on a clean `origin/main` worktree", i.e. pre-existing. That claim does not hold up: I re-ran only the failed job on [run 30433311855](https://github.com/antiwork/gumroad/actions/runs/30433311855) with no code change, and `Test Slow 11` came back `SUCCESS` on the same commit. A failure that passes on retry at an identical tree is a flake, not a pre-existing breakage.

The failing line was `first(:fillable_field, "Card number", visible: false, wait: 20).click` — a Capybara wait against the Stripe Payment Element iframe, which is exactly the shape that flakes under load. It had already burned all three `rspec-retry` attempts in the original run, so this was a slow-iframe timeout, not a deterministic failure. Same spec was `SUCCESS` on sibling branches #6540 and #6541 in the same window, which supports that.

Nothing about this diff changes: it is comment-only, no non-comment lines, and the `mailer_info` / `application_mailer` specs are 45 examples / 0 failures. Only the CI-note paragraph in the body was inaccurate, and I'd rather flag that than leave a false "known-broken on main" note in the record for the next person to inherit.

The spec is worth watching independently — it landed 2026-07-27 with #6191 and it is the single slowest example in that batch at 76s — but that is a separate concern from this PR, not a blocker on it.

PR #6550, opened 2026-07-29
Route cc'd and bcc'd recipients at blocked domains to SendGrid too

gumclaw:
## What

`ApplicationMailer#redirect_united_internet_recipients_to_sendgrid!` now checks `message.destinations` (to + cc + bcc) instead of `message.to`, so a message is moved to SendGrid when *any* envelope recipient is at a Resend-blocked domain — not only when the blocked address happens to sit in the `To` header.

Follow-up to #6524, which landed the redirect itself but only inspected `To`.

## Why

One delivery provider is chosen per message. If a blocked address is in `cc` or `bcc`, #6524 leaves the whole message on Resend, so **that recipient's copy bounces while everyone else's delivers** — a half-failure nothing surfaces. Nobody is told the copied party heard nothing.

This is not hypothetical. `AffiliateMailer` addresses the affiliate and cc's the seller on six of its methods, and passes no recipient of its own to `random_delivery_method_options`:

```
app/mailers/affiliate_mailer.rb:96    cc: @seller.form_email,
app/mailers/affiliate_mailer.rb:107   cc: @seller.form_email,
app/mailers/affiliate_mailer.rb:118   cc: @seller.form_email,
app/mailers/affiliate_mailer.rb:129   cc: @seller.form_email,
app/mailers/affiliate_mailer.rb:143   cc: @affiliate_user.form_email,
app/mailers/affiliate_mailer.rb:158   cc: @inviter.form_email,
```

So a seller at web.de or GMX never learns their collaborator status changed, and no bounce reaches anyone who would act on it. Moving the whole message to SendGrid costs the other recipients nothing — SendGrid is already half of the normal random split.

`destinations` is the correct accessor rather than `to + cc + bcc` hand-assembly; verified it returns all three fields:

```
$ ruby -e 'require "mail"; m=Mail.new(to:"a@x.com",cc:"b@web.de",bcc:"c@gmx.de"); p m.destinations'
["a@x.com", "b@web.de", "c@gmx.de"]
```

## Test Results

`spec/mailers/application_mailer_spec.rb` — **18 examples, 0 failures** (3.38s), including the three new cases:

```
ApplicationMailer delivery method United Internet recipient redirect
  redirects when the blocked recipient is in cc rather than to
  leaves a message alone when neither the to nor the cc recipient is blocked
  redirects when the blocked recipient is in bcc

18 examples, 0 failures
```

`spec/models/mailer_info_spec.rb` — **30 examples, 0 failures** (the renamed `force_sendgrid_for_recipients?` parameter is signature-compatible; existing `to:` callers are unchanged).

`bundle exec rubocop -a` on all three changed files: **3 files inspected, no offenses detected**.

Branch CI at the rebased head `98cedd3`: **74 passed / 1 skipped, 0 failures** ([run 30480803151](https://github.com/antiwork/gumroad/actions/runs/30480803151)). The only commit since is the QA evidence swap, which touches nothing under `app/` or `spec/`.

> Note on `spec/mailers/affiliate_mailer_spec.rb`: it reports 27/29 failures locally, but with identical failures on unmodified `origin/main` (`29 examples, 27 failures`) — all `ViteRuby.instance.manifest.resolve_entries` asset-manifest errors from having no local frontend build. Environmental, not caused by this branch.

## QA steps

Provider selection is server-side with no UI surface, so this is verified by spec rather than by clicking:

1. Build a message addressed to a non-blocked domain with a cc at a blocked one:
   `ApplicationMailer.some_mailer(to: "seller@gmail.com", cc: "boss@web.de")`
2. Assert `mail.delivery_method.settings[:address] == SENDGRID_SMTP_ADDRESS` — before this change it was `RESEND_SMTP_ADDRESS`.
3. Assert the `X-GUM-Email-Provider` header reads `sendgrid`, i.e. it stays consistent with the SMTP address the message was actually moved to (the redirect runs before `set_custom_headers` for exactly this reason).
4. Repeat with `bcc:` instead of `cc:` — same result.
5. Negative case: cc at `gmail.com` leaves the message on whichever provider the Router picked.

Steps 2–5 are the three new examples plus the retained "leaves other recipients on the provider the Router chose".

The evidence below runs the same message through the redirect with the Router pinned to Resend, once with `app/` at `origin/main` and once with this branch, so the only difference is the accessor. Before the fix a blocked cc or bcc recipient leaves the whole message on Resend — that recipient's copy is the one rejected, the To recipient's still delivers, and nothing surfaces the gap. After it, the message moves to SendGrid and every recipient gets their copy.

cc/bcc reroute before/after and spec evidence

Walkthrough video (terminal, since provider selection has no UI surface): [qa-media/pr-6550-cc-bcc-redirect-walkthrough.mp4](https://raw.githubusercontent.com/antiwork/gumroad/0cf45e358892118c8bf4be5f59cf98c99acc84bf/qa-media/pr-6550-cc-bcc-redirect-walkthrough.mp4) — walks the envelope, the diff, and the revert-it-and-it-reddens check.

## Pre-merge review

fable-5 adversarial review run against a worktree of the branch. One finding, fixed in-branch (`80ddb33`): the new 8-line comment on the redirect restated the reasoning already given in the predicate's own doc immediately above its only caller, and both ran past the three-line limit that #6544 landed hours earlier. Both trimmed; the non-obvious *why* (`destinations` not `to`, silent per-recipient bounce, the AffiliateMailer example) is retained in one place each. No P1/P2 correctness findings — the diff adds zero new branches, only widens the input to an existing predicate.

## AI disclosure

Written by Anthropic's Claude Opus, model id `claude-opus-5` (the `model.default` this agent runs on), driving the Hermes agent as the `gumclaw` account — the model wrote the diff, the specs, the demo harness, and this body. It was asked to fix the gap #6524 left behind (the redirect only inspecting `To`), then to correct the QA evidence and add a walkthrough video after review.

Every result above is real tool output: specs and RuboCop run locally at this branch head, the before/after provider table produced by running the same message through `origin/main`'s `app/` and this branch's, the `AffiliateMailer` cc lines grepped from the tree, and the affiliate-spec failures reproduced on unmodified `main` to establish they are environmental. A human should review the routing judgement — this moves whole messages between email providers on the money path.


gumclaw:
Greptile's review is clean — confidence 5/5, no inline findings, and its own summary confirms the ordering that matters here (message fully built before `destinations` is read, NullMail still guarded, provider header assigned only after any SendGrid override). Nothing to fix from it.

CI is green at the current head `80ddb33`: 74 passed, 1 skipped, 0 failures across all 75 jobs.

Handing to @nyomanjyotisa for review and merge, same as #6524 which this follows up. Not arming auto-merge: it changes which provider carries every message with a blocked-domain recipient, so the failure mode if the judgement is wrong is silent non-delivery rather than a loud error, and that deserves a human read.

What's worth checking: that widening the input from `message.to` to `message.destinations` is the behaviour we want for bcc specifically — a bcc'd address at a blocked domain now moves the *whole* message to SendGrid, including the copies going to everyone else. That's deliberate (one provider per message, and SendGrid is already half the normal split), but it is the one judgement call in the diff.

gumclaw:
### The three red `Test Slow` jobs here are not this branch

CI on this PR is failing only these three `spec/requests/purchases/product/taxes_spec.rb` examples:

- `:3904`, `:3929`, `:4074` — "country change scenarios"

all with the same error out of `checkout_helpers.rb:241`, waiting on a Stripe 3DS challenge iframe that never appears:

```
expected to find css "iframe[src^='https://js.stripe.com/v3/three-ds-2-challenge']" but there were no matches
```

**`main` itself fails the same example** at `3d0f55e4f` — run [30470868964](https://github.com/antiwork/gumroad/actions/runs/30470868964), job `Test Slow 13`, with none of this branch's code in the tree. Two other unrelated PRs fail the identical three examples. I reran the failed jobs once here per the flake rule and they reproduced, so it is deterministic rather than a flake.

Tracked repo-wide in #6562 (leading hypothesis: a Stripe-side change in whether the Mexico test card is challenged). Nothing to fix on this branch — please read the rest of the run, which is green, and don't bounce this as a broken branch. I have deliberately **not** armed auto-merge.

gumclaw:

gumclaw:

nyomanjyotisa (commented):
Code looks correct. One correction before merge: the QA screenshot still says a blocked CC/BCC causes the whole send, including the To recipient, to be lost. The PR body correctly explains that only the blocked recipient’s copy is rejected while the others deliver. Please update the screenshot so the evidence matches the final diagnosis.

Please also add the required short walkthrough video and name the specific model in the AI disclosure.

gumclaw:
All three done, on `0cf45e3`.

The screenshot is replaced. The old one asserted the whole send was lost, which was wrong — it now runs the same message through the redirect twice, once with `app/` at `origin/main` and once with this branch, so the only variable is the accessor. Before the fix a blocked cc or bcc leaves the message on Resend and it is that recipient's copy alone that gets rejected; the To recipient's still delivers, which is what makes the failure silent. After it the message moves to SendGrid and everyone gets their copy. The caption in the body says the same thing now, so evidence and diagnosis agree.

Walkthrough video added at `qa-media/pr-6550-cc-bcc-redirect-walkthrough.mp4`, linked from the QA section. Provider selection has no UI surface, so it is a terminal recording: the envelope (`message.to` misses the blocked address, `message.destinations` catches it), the diff, then the revert-and-it-reddens check — the two new examples fail with the fix backed out and nothing else does.

AI disclosure now names the model: Anthropic's Claude Opus, `claude-opus-5`, read from this agent's own config rather than recalled.

CI was green at the previous head (74 passed / 1 skipped, run 30480803151) and the new commit only touches `qa-media/`, so nothing under `app/` or `spec/` moved. Leaving the merge with you.

gumclaw:
Body-only edit — no commit, head unchanged at `0cf45e3`, CI unaffected.

The walkthrough video link was pinned to the branch (`github.com/antiwork/gumroad/blob/fix/route-cc-recipients-at-blocked-domains/...?raw=true`). That resolves fine today but merge auto-deletes the branch, so it would have become a permanent 404 in the merged record. Re-pinned to the commit SHA, which reads from git history and survives both branch deletion and the weekly `qa-media/` cleanup.

Verified identical bytes across both forms (661,785, matching `git cat-file -s` at that commit), so the video content is unchanged — only the ref is durable now. The screenshot was already SHA-pinned.

Worth noting for anyone else re-pinning: `repin-evidence-urls.py --sweep` only matches `raw.githubusercontent.com`, so this `blob/<branch>/...?raw=true` form reported clean. Video links are the likely place it hides, since that is the form GitHub's own "copy link" yields on a blob page.

nyomanjyotisa (approved):

PR #5513, opened 2026-06-18
Move profile editing to /profile with a native pages-and-sections editor

gianfrancopiana:
Refs #5354

Supersedes #5388

## What

Profile editing moves out of the public page's floating inline editor into a dedicated, native editor at a top-level **`/profile`** route — a left-hand form with a live preview, built from the existing product-edit patterns. `/settings/profile` 301-redirects to it. A profile's pages and sections are edited through a single collapsible accordion.

## Why — key decisions

- **A top-level `/profile`, not buried in Settings.** Editing your profile is a primary action and deserves a linkable URL. `/settings/profile` 301-redirects (query string preserved, so `?section=` deep links survive), and `profile` is reserved in the username denylist so it can't shadow a creator's public `/:username` page.

- **One control for pages instead of two.** An earlier cut had a separate "Pages" list and page pills under "Sections". Two controls for one concept, with nothing explaining their relationship. Now each page is a collapsible row whose sections (and "Add section") live inside it, so the page you're editing is never ambiguous.

- **An accordion instead of tabs/pills.** Pills read like transient view-filters and couldn't carry rename / reorder / delete; they also separated "which page" from "editing its sections." A single-open accordion treats pages as the objects they are and co-locates section editing with its page.

- **Consistency over novelty.** Pages collapse with the same `DrawerToggle` / `Drawer` mechanism sections use, reorder by drag (a distinct handle keeps the nested page and section drags from colliding), and the shared row now gives the drag image a background so dragged rows are no longer transparent.

- **Backend.** Section visibility is derived from a blank header (no redundant show/hide toggle), saves are explicit and batched through `SellerProfileSections::SaveService` (shared by the editor and the sections API), and a backfill migration reconciles legacy `hide_header` data.

## Screenshots

**Desktop**

**Mobile**

## Demo

https://github.com/user-attachments/assets/89dd500d-9a57-4559-9176-811c096850f8

---

AI disclosure: built with Claude Opus 4.8.

---
[...]

PR #6552, opened 2026-07-29
Restore self-serve font and colour controls to profile settings (partially reverses #5513)

gumclaw:
Ref #5513

## What

Restores direct font, background color, and highlight color controls in **Settings → Profile → Design**. The preview updates before save and stays faithful to the buyer-facing theme, including exact white while the dashboard uses dark mode. Saves send only changed design fields, rebase untouched fields when another tab updates them, and reject stale page layouts.

Git history confirms the pre-#5513 editor rendered each color swatch below its label. This PR keeps the restored controls but improves that layout: the labels share one intrinsic-width column and both swatches align in the next column, with clear spacing between rows.

The Store Agent remains read-only for theme values. It can inspect the current theme and direct the seller to the Design tab; it cannot write these settings or claim that it can.

The save path locks the seller before a locking/current profile read. This prevents duplicate rows on concurrent first saves without relying on a database uniqueness constraint, and it avoids a stale MySQL repeatable-read snapshot before page-section conflict checks. Empty unsaved layouts use the same version as newly persisted empty layouts, so a theme-only first save in another tab does not cause a false layout conflict. Pending avatars are validated before the seller lock reloads them. Theme CSS cache invalidation runs after commit so a buyer request cannot repopulate stale CSS during a save.

### QA

All final-head images below are pinned to `0eb94da5952802e5eff215bc385c4ee902ffa2c3`.

**Restored legacy layout before the spacing fix:** labels above their swatches.

Restored legacy color layout

**Final desktop, dashboard dark mode:** each font card uses its own face; both swatches align in one column beside their labels; exact-white preview matches the storefront.

Final Design tab in dark mode

**Final desktop, dashboard light mode.**

Final Design tab in light mode

**Mobile at 375×667, dark and light modes.**

Mobile Design tab in dark mode

Mobile Design tab in light mode

**Custom profile page active:** the Design tab stays available for product and checkout theming, explains that the custom profile owns its own design, and hides the native Pages tab.

Design tab with an active custom profile

**Saved custom HTML with the feature disabled:** the native editor and preview return, with no false custom-profile warning.

Design tab with saved but disabled custom HTML

**Contrast edge case:** saturated red remains readable in the live preview.

Red background contrast case

**Buyer-facing exact-white storefront in dashboard dark mode.**

Exact-white buyer storefront

### Verification

The final change-aware gate reached 99% confidence with 27 planned test groups passing in 232 seconds. Focused Rails coverage includes real two-connection tests for concurrent first creation and current-read ordering plus the empty-layout/theme-first regression. The profile form Vitest suite passed 4 tests. Targeted RuboCop and ESLint checks are clean, and a focused TypeScript compile covers the preview's custom CSS-property contract. The required repo-wide `npm run typecheck` command still stops before project checking because the unchanged checkout cannot resolve the existing `vite/client` type entry; the same error reproduces on `main`.

Final autoreview command:

```sh
/Users/yaco/.codex/skills/autoreview/scripts/autoreview --mode branch --base origin/main --prompt-file /tmp/pr6552-autoreview-notes.md --stream-engine-output
```

Result: `autoreview clean: no accepted/actionable findings reported`.

Production was checked read-only through Gumclaw. There are 5,487 seller IDs with more than one `seller_profiles` row, so a unique index would fail and is outside this PR; deduplication needs a separate product decision about which theme wins. A complete primary-key-chunked scan found zero stored three-digit hex colors, so rejecting shorthand does not invalidate existing data.

## Why

#5513 removed the Design tab and left sellers without a fast direct way to adjust one theme value and see the result. Whole custom pages still belong in the agent and CLI workflows. Small theme changes belong in a direct, reversible UI with an accurate preview.

This keeps the original data model and public rendering contract while making concurrent saves, first-row creation, cache invalidation, custom-page states, and contrast behavior explicit and tested.

---

This PR was implemented with AI assistance using claude-opus-5 and gpt‑5.5 xhigh.


gumclaw:
**Assigned @gianfrancopiana, labelled `working`.** Previously this PR had no assignee at all and no state label, so nothing recorded who owned the outcome.

**Why you specifically:** this reverses part of your #5513 (*Move profile editing to /profile with a native pages-and-sections editor*), which removed the Design tab in favour of the store agent. You made that call, so the product half of this — whether a direct self-serve write path to the three `seller_profile` theme columns should exist alongside the agent — is yours to accept or reject. It is also your surface (profile settings / front-end).

**What is expected of you right now: nothing yet.** The ball is mine, which is what `working` records:

- the PR is still a **draft**;
- CI is mid-flight (70 checks running);
- no review has been requested from you or anyone, so I have not handed it off.

I owe you: taking it out of draft, a green check run, and an explicit review request. When that lands I will flip to `awaiting-human` and ask for the review directly. Do not spend time on it before then — I flagged you now so the tracker names an owner rather than implying nobody owns it.

One thing worth pre-reading if you want a head start, because it is the part I expect you to push back on rather than the UI: there is **no locking between the two write paths**. An agent write and an open settings form are last-write-wins, same as two browser tabs.

gumclaw:
Cleared a merge conflict with `main` at `31195b11` — no reviewer action needed, but the resolution involved a genuine product tension worth naming.

`main` picked up seller checkout theming and, with it, help-centre copy plus an API spec asserting that **there is no self-serve theme screen** and sellers must contact support. This branch restores exactly that screen. I kept this branch's direction (`editable_by_seller: true`, the Settings > Profile > Design copy, the `FONT_CHOICES`-parity spec) and kept everything of main's that is orthogonal:

- `CSS_HEX_COLOR_REGEX` (3-digit CSS hex helper) alongside this branch's anchored `\A..\z` `HEX_COLOR_REGEX` security fix — the anchoring matters, `^...$` let `#ffffff\nbody{display:none}` through into the storefront stylesheet.
- The checkout surface in `THEME_SURFACES` and the help-centre paragraph explaining that checkout only follows the theme for single-seller carts, including main's spec asserting the qualified wording.
- main's `#accent_color_for_indicators` and font-CSS-source specs.

So the branch now says: the theme applies to checkout too (main's addition), *and* sellers can edit it themselves (this branch). If you'd rather the help centre keep pointing at support until this ships, say so and I'll flip that paragraph back.

gumclaw:
Re-shot on the hosted preview at the post-merge head, and flagging one thing that needs a human.

**Merge conflict with `main`, one file:** `app/services/ai/store_agent_api_catalog.rb`. This PR rewrites the `get_user_theme` entry to stop telling the agent to route colour requests to support; #6531 rewrote the same entry to describe checkout theming. Both edits are wanted, so the resolution should keep the checkout sentence *and* the "sellers can do this themselves in Settings → Profile → Design" correction. Not a QA blocker — the preview is built from this branch's own merge of `main` and is serving `31195b113a2c` fine.

**Why I re-shot rather than reusing the existing frames:** the merge brought #6531, which touches two of this PR's own files (`hex_color_validator.rb`, help-centre 124) and adds a *new consumer* of the same `seller_profile` columns this settings form writes. It also puts a new gate in front of the stylesheet — `SellerProfile#custom_styles` now returns `""` unless `highlight_color`, `background_color` and `font` all pass `safe_for_css?`/inclusion, with the cache key bumped to `v5`. A theme saved through the restored Design tab therefore has a new way to silently render nothing, and that is worth proving rather than assuming.

It holds. Saving `Roboto Slab` / `#fdf3d8` / `#ff5a1f` from the Design tab produced, on the live storefront, `--accent: 255 90 31`, `--filled: 253 243 216` and a **non-empty `--danger: 155 28 18`** — the gate passed and the `v5` cache compiled. Checkout additionally shows `--indicator: 246 86 29`, which only #6531's `checkoutTheme` sets and which is absent on the storefront and product frames, so that frame is specifically the new checkout theming path rendering a theme a seller set through this PR's form.

Seven frames (desktop Design tab, live-preview-before-save, saved+reloaded, storefront, product page, checkout, 375px) are in the PR body under **r2**, pinned to `c06f6d537`. Preview theme restored to the seeded `ABC Favorit` / `#ffffff` / `#ff90e8` afterwards.

gumclaw:
Rebased onto `main` to clear a merge conflict — head is now `cd12bf9f0`, `MERGEABLE`, CI green.

One conflicting file, `app/services/ai/store_agent_api_catalog.rb`, and it needed a judgement call rather than a mechanical union, so flagging it for review:

**The `get_user_theme` summary.** Main currently tells the store agent *"There is no self-serve screen for them and you have no endpoint to change them, so when a creator wants different colors or fonts, tell them Gumroad support can apply it."* This PR is precisely what makes that false — it restores the Settings > Profile > Design controls. So I kept **this branch's** wording (*"the creator can change them themselves in Settings > Profile > Design, which shows a live preview before saving — point them there"*) and dropped main's. Taking main's side here would ship a prompt that tells sellers to file a support request for a screen this PR just gave them.

**The `update_user_custom_html` summary.** Unrelated to the theme work — main added the `products_total` / `posts_total` cap-disclosure sentence (gumroad-private#1522). Kept as-is; the two changes only collided because they are adjacent lines.

Also checked the auto-merged files for leftover assertions of main's now-false wording: `spec/controllers/api/v2/users_controller_theme_spec.rb`, `spec/services/ai/store_agent_service_spec.rb` (two places) and `store_agent_service.rb`'s prompt all consistently say Settings > Profile > Design, and nothing anywhere still asserts "support applies changes by request". Main's cap-disclosure assertions all still resolve against the merged prompt.

@gianfrancopiana the theme-wording call above is the one thing worth a second opinion on — it's a deliberate override of main, not a merge artefact.

gumclaw:
Merged latest `main` in to clear the conflict — one file needed a real decision.

**`app/services/ai/store_agent_api_catalog.rb`** — both sides rewrote the `get_user_theme` endpoint description and its comment, in opposite directions. `main` says there's no self-serve editor and the agent should offer to have support apply theme changes; this branch adds exactly that self-serve editor, so it points creators at Settings > Profile > Design with its live preview. Kept this branch's wording (main's is now factually wrong once this ships) and kept main's brand-new `get_user_profile_layout` endpoint that landed right below it — the textual conflict was only because the two additions were adjacent. Nothing dropped from either side.

Everything else merged cleanly. Ran the touched specs: profile presenter, profile settings props, profile controller, seller profile model, store agent service, and the v2 users theme + profile-layout specs — 164 examples, all green. Rubocop clean on the Ruby files, Prettier clean on the three TS files.

gumclaw:
@gianfrancopiana still waiting on review (24h+; goal is 24h turnaround, 7 days a week — Sahil, 2026-07-21). Nothing is blocked on me.

Two pushes since the last update, neither touching code. Current head is `9b8ac9f34`.

**`9af52bce2` — deleted 15 stale evidence files.** `git show --name-status` is 15 `D` lines, all under `qa-media/`, 0 insertions. The r1 and r2 capture rounds stayed committed after being superseded, so `qa-media/` held 23 files while the body linked 8. Three of them (background-colour, font-selected, highlight-colour live previews) documented intermediate states the body no longer describes — an unlinked evidence file reads as current to whoever opens it, which is worth removing before merge rather than after.

**`9b8ac9f34` — re-captured all 8 surfaces at the merged head** and repointed the body to the `r4` set. That is the evidence to review; it depicts the code you would be merging.

Two notes so nothing here misleads you:

- The `r3` files are now committed-but-unlinked for the same reason r1/r2 were, so the same cleanup is owed on them. I have deliberately not pushed a third commit for it while this round is fresh — happy to do it on request, or to fold it in if you ask for any other change.
- Every embed resolves. The links are pinned to full commit SHAs, and a blob pinned to a SHA is unaffected by deleting the path from a later tree, so nothing in the body 404s regardless of which rounds get pruned.

I also confirmed the `main` merge in `a2c822d4e` did not invalidate any capture: of the 133 files it brought in, none touch a captured surface — the help-centre changes are articles 10/121/194/200/282, not 124, and the only `.tsx` is `DownloadPage/Layout.tsx`.

CI was green before these commits and both re-trigger it; I expect green, since neither changes code.

gumclaw:
## `Test Slow 9` is a real regression from this branch, not a flake — taking this back off your queue

Reran the failed shard once before saying anything, per our flake rule. It failed **identically both times**, same single example, so this is deterministic and mine to fix. Ball is back with me — please don't spend time on it until I clear this.

**The failure**

```
rspec ./spec/requests/user/settings_spec.rb:108
User profile settings page saving profile updates avatar when the avatar is invalid
  displays an error if either dimension is less than 200px

expected to find visible alert with text "Please upload a profile picture that is at
least 200x200px" but there were no matches. Also found "", which matched the selector
but not all filters.
```

Runs [30559496946](https://github.com/antiwork/gumroad/actions/runs/30559496946) (original) and job `90934148251` (rerun) at head `2d3f606e3`. The previous head `9b8ac9f34` was green, which brackets it to the last three commits.

**Mechanism — `with_lock` reloads the user and drops the pending avatar**

`2d3f606e3`/`09aa62c7e` replaced the controller's `ActiveRecord::Base.transaction` + `seller_profile.lock!` with `current_seller.with_locked_seller_profile`, whose body is `with_lock do … end` **on the user**. `with_lock` is `lock!` + block, and `lock!` calls `reload`. That reload resets `current_seller`'s attachment associations.

`avatar_is_valid` (`app/modules/user/validations.rb:42`) opens with:

```ruby
return if !avatar.attached? || !avatar.new_record?
```

The avatar is attached *before* the lock, then the reload inside `with_lock` makes that attachment a persisted, non-new record. So the guard early-returns, the `200x200` error is never added, `current_seller.update!` succeeds, and the controller answers `respond_success` — which is exactly why the test sees an empty alert instead of a missing one. The old code never reloaded the user, so the validation still saw a new-record attachment.

The dimension check hasn't been weakened in intent — it's being **skipped**, which is worse than it looks: a sub-200px avatar now saves silently in this path rather than being rejected. That's a user-visible correctness bug, not just a red spec, so I'm glad it caught this.

**Fix I intend to apply** — validate the avatar before taking the lock, so the check runs while the attachment is still new, and keep the lock scoped to the layout write it exists for. The lock was introduced for the first-save race on `seller_profiles`; it has no reason to reload the user's attachments, so narrowing it addresses the regression without giving up the concurrency fix. If narrowing turns out to reopen the race, the fallback is to keep the user lock but re-run the avatar validation explicitly inside it.

Verification I'll post before asking for your eyes again: `spec/requests/user/settings_spec.rb:108` green, the full `settings_spec.rb` file green, and the concurrency specs from this branch (`user/seller_profile_concurrency_spec.rb`, `settings/profile_controller_spec.rb`) still green — the point being that I fixed the regression without regressing the thing this branch is for. ETA next sweep.

Moving `awaiting-human` → `working` and leaving you assigned so it comes back to you, rather than pretending it's review-ready while it's red. Also noting `main` is independently red on the new required `Typecheck TS` gate ([#6630](https://github.com/antiwork/gumroad/pull/6630) fixes it, from a regression of mine in #6594) — unrelated to this failure, but it means a green run here needs that landed too.

gumclaw:
## Fix pushed — `547de1314`

Didn't want to leave this as a plan, so the fix is in. One file, +9 lines, `app/controllers/settings/profile_controller.rb`: validate the pending avatar **before** `with_locked_seller_profile` takes its lock, guarded on `current_seller.attachment_changes.any?` so unrelated seller validation state stays out of this path. The concurrency lock keeps doing what this branch added it for; it just no longer swallows the avatar checks.

**A/B proof that this is the actual cause, not a guess:**

| | `settings_spec.rb:108` |
|---|---|
| with the fix | **1 example, 0 failures** |
| fix stashed, same worktree, same DB | **1 example, 1 failure** — the original empty-alert error |

**Regression safety** — the concurrency behaviour this branch exists for is untouched: `spec/models/user/seller_profile_concurrency_spec.rb` + `spec/controllers/settings/profile_controller_spec.rb` = **36 examples, 0 failures**, including `creates one profile row when two first saves run concurrently`.

**Two things I want to be straight about rather than let CI surprise you:**

1. `settings_spec.rb:78` ("purges the attached avatar when the avatar is removed") fails on my machine **with the fix stashed as well** — so it is not from this change. It stubs `ActiveStorage::Blob#purge` against S3 and is unhappy in my local setup; it was green on CI at the previous head. Watching it on this run.
2. `settings_spec.rb:127` failed once locally with `Timeout::ExitException` in `spec/support/stale_element_retry.rb` — a Chrome stale-node timeout — then passed on re-run. Local browser flake, not a signal.

So: the deterministic failure is fixed and proven; the two noisy ones are local-environment artifacts I'm not going to dress up as green. CI on real infrastructure is the arbiter and I'll read the push run (not the `pull_request_target` NOOP twin) when it settles.

Still expect one **unrelated** red until [#6630](https://github.com/antiwork/gumroad/pull/6630) lands: `main` is currently failing the newly-required `Typecheck TS`, from a regression of mine in #6594. Nothing to do with this branch.

Leaving `working` on until CI confirms, then it flips back to `awaiting-human` for you, @gianfrancopiana — I'll do that rather than make you guess whether it's ready.

PR #6555, opened 2026-07-29
Open the Card pane before the UPI prefill spec types into it

gumclaw:
## What

`spec/requests/purchases/product/upi_element_billing_prefill_spec.rb` clicks the Card row's number field before typing into checkout's own Full name field. That click assumes the Card pane is the one Stripe mounted expanded. This opens the Card pane first when it isn't already expanded.

Spec-only. No product code changes.

## Why

The spec failed on an unrelated PR's `Test Slow 9` shard ([run 30446707655](https://github.com/antiwork/gumroad/actions/runs/30446707655/job/90563433192)) with `expected to find field "Card number" ... no matches` on all three rspec-retry attempts — and it failed in the *setup* step, before any assertion ran.

The Capybara failure screenshots show why: Stripe mounted the accordion with the **UPI pane already expanded**, rendering the India address form, so there was no Card number field on the page to click.

Stripe mounted the UPI pane expanded, so no Card number field exists

Which pane mounts expanded is Stripe's own relevance ranking, not ours. We never pass `paymentMethodOrder`; the only ordering-adjacent option is `paymentMethodTypes`, whose order comes from `ONE_TIME_PAYMENT_METHOD_TYPES` (card first, upi seventh) and is not a display contract. On this INR/India element the mounted pane is not stable, so the spec cannot assume it.

**This does not paper over a product bug.** Two things were checked before writing the fix:

- The regression the spec exists to catch is still caught. It pins `fields.billingDetails.name: "never"` plus a checkout-owned Full name field — a shape that does not depend on which pane rendered first. Reverting #6191's product fix still fails these assertions under either mount order.
- Nothing on `main` changed which pane mounts. The same spec ran green on `main` at a7ad25e4 ([job 90544655069](https://github.com/antiwork/gumroad/actions/runs/30446707655), 50 examples 0 failures) and green on two sibling PRs an hour before the failure. The recent checkout commits touch pay-what-you-want and multi-seller paths that this fixed-price single-seller INR cart never enters.

## Before / After

**Before** — the pane Stripe expands decides whether the spec passes:

```ruby
within_payment_element_frame do
  first(:fillable_field, "Card number", visible: false, wait: 20).click
end
```

When UPI mounts expanded there is no Card number field, so this raises in setup.

**After** — open Card unless it is already the expanded row:

```ruby
within_payment_element_frame do
  unless has_selector?(:fillable_field, "Card number", visible: false, wait: 0)
    first(:xpath, ".//*[normalize-space(text())='Card']", visible: :all, wait: 20).click
  end
  first(:fillable_field, "Card number", visible: false, wait: 20).click
end
```

`wait: 0` on the probe so the common case (Card already expanded) costs nothing.

## Tests

Branch CI green at `46718c7b` — [run 30450356615](https://github.com/antiwork/gumroad/actions/runs/30450356615), all jobs success, zero failures.

The spec itself ran and passed in that run ([job 90572487355](https://github.com/antiwork/gumroad/actions/runs/30450356615/job/90572487355)):

```
UPI Payment Element billing handling: . [13.86s]
  UPI Payment Element billing handling keeps checkout's Full name field for UPI
  and has the pane collect only the address
60 examples, 0 failures
```


gumclaw:
@xrav3nz assigning you because you were the assignee on #6191, the PR that added this spec — you own the regression it guards.

**What I need from you: one review, then merge.** I'm deliberately not arming auto-merge. The change is spec-only, but this spec is the sole automated guard on #6191's UPI billing-details behaviour, so a bad weakening of it has money consequences and should get a human read rather than my own sign-off.

The one thing worth your eyes: **does the added pane-opening step make the assertions vacuous?** I believe not — the spec asserts `fields.billingDetails.name: "never"` plus a checkout-owned Full name field, which is independent of which pane rendered first, and reverting #6191's product change still fails it under either mount order. Body has the full reasoning and the CI screenshot showing Stripe mounting UPI expanded. If you read that differently, say so and I'll rework rather than merge.

CI is green at `46718c7b` ([run 30450356615](https://github.com/antiwork/gumroad/actions/runs/30450356615)), including the spec itself: 60 examples, 0 failures.

PR #6557, opened 2026-07-29
Only report client-confirm failures that nothing else records

gumclaw:
## What

`OrdersController#confirm_error` reports every browser-side `stripe.confirmPayment` failure to
Sentry. This narrows it to the failures Stripe does **not** already tell us about: a report is
suppressed only when the payment method is one we record server-side (`card`, `link`) **and** the
error is a `card_error`, meaning a charge attempt really was made. The unconditional
`Rails.logger.error` line is untouched, so full forensics remain available for every confirm failure.

## Why

The endpoint exists for **redirect-based** methods. When an iDEAL or Bancontact confirm is rejected
the buyer never reaches their bank, so no charge and no `payment_intent.payment_failed` webhook is
created, and the purchase is indistinguishable from an abandoned tab — the browser is the only
witness. That is the signal it was built to capture.

Card and Link are not like that. A failed attempt transitions the intent, Stripe sends
`payment_intent.payment_failed`, and `Purchase::ChargeEventsHandler#handle_event_failed!` writes the
code to `purchases.stripe_error_code`. Those declines are already queryable in our own database.

Because every report is emitted with one fixed message, all of them fingerprint into a single Sentry
issue. That issue is running **600–1,300 events/day** (2026-07-28 peaked at 1,324) and consists
almost entirely of routine declines, while producing **zero** events from the redirect methods it was
built for in the last 14 days:

| Slice | 14-day count |
|---|---|
| `payment_method_type:ideal` | **0** |
| `payment_method_type:bancontact` | **0** |

So this is a signal-quality problem, not just quota: the needle is buried under several hundred
declines a day, and the issue has to be re-sampled every morning to establish that it is not an
incident.

Cross-tabulated from 597 post-deploy events fetched individually through the project-scoped event
endpoint (597 unique event IDs, no duplicates):

| `payment_method_type` | `stripe_error_type` | count | after this change |
|---|---|---:|---|
| `card` | `card_error` | 302 | suppressed |
| `link` | `card_error` | 70 | suppressed |
| *(blank)* | `card_error` | 177 | **still reported** |
| `card` | `invalid_request_error` | 38 | **still reported** |
| *(blank)* | `invalid_request_error` | 10 | **still reported** |

Net effect: 372 of 597 events (62%) suppressed, and every remaining event is one we cannot
reconstruct from the database.

Tracking issue: antiwork/gumroad-private#1514

## Before / After

Not a UI change — no user-visible surface is touched. The endpoint's HTTP contract is unchanged
(`{ success: true }` in every case, same status), the client caller in `app/javascript/data/order.ts`
is untouched, and no buyer-facing behavior differs. The only difference is whether a Sentry event is
emitted alongside the log line, so the artifacts below are the spec run plus the endpoint's
observable behaviour on the hosted preview app.

There is no rendered surface to capture, so this is the deployed endpoint's observable behaviour
instead: seven real `POST /orders/:token/confirm_error` requests against
`https://fix-confirm-error-report-only-in.apps.staging.gumroad.org` (served revision
`b155cf6166ef`, confirmed via `x-revision`), correlated with the puma container's log stream.

Confirm-error reporting behaviour on the hosted preview app

What the live build shows, which the spec run cannot:

- **The HTTP contract really is unchanged.** All seven shapes returned `200 {"success":true}` —
  suppression is invisible to the client caller.
- **Log lines 7/7, Sentry envelopes 5/7.** `card`+`card_error` and `link`+`card_error` produced an
  unconditional `logger.error` line with **no** `[Transport] Sending envelope` after it, while the
  blank, `invalid_request_error`, `ideal`, `bancontact` and `some_new_method` shapes each produced a
  log line **and** an envelope with a distinct event id. The suppression and the retained-forensics
  claim are both demonstrated in one log window rather than asserted via a mock.
- **The `card`+`invalid_request_error` blind spot round 1 found is genuinely still reported** on the
  deployed build (envelope `01db02bc…`), not just in a spec double.
- **Fail-open holds on lookalike input.** Evaluated in the preview's own Rails runtime, `"Card"` and
  `" card "` both land on the *report* side, so no casing/whitespace trick suppresses a report.

### Re-verified after merging main (head `334d8a485`)

The head moved only by merging `origin/main` (8 commits); the branch's own diff against its
merge-base is **byte-identical**, and main brought in nothing touching `app/controllers/orders_controller.rb`
or the order services behind this endpoint. That still re-arms this PR because the old panel cited
served revision `b155cf6166ef`, while the preview now serves `41fa61fc2d85`.

Re-ran the actual endpoint probes against the live hosted preview, not just a curl smoke:

```
$ curl -sI https://fix-confirm-error-report-only-in.apps.staging.gumroad.org/ | grep -i x-revision
x-revision: 41fa61fc2d85
```

Used the preview console to read `Order.order(:id).first.secure_external_id(scope: "confirm")`, then
sent the same seven real `POST /orders/<token>/confirm_error` shapes in a fresh log window:

| shape | HTTP response | log line | Sentry envelope |
|---|---:|---:|---:|
| `card` + `card_error` | `200 {"success":true}` | yes | **none** |
| `link` + `card_error` | `200 {"success":true}` | yes | **none** |
| `card` + `invalid_request_error` | `200 {"success":true}` | yes | `495a0000ea7d434c` |
| blank + `card_error` | `200 {"success":true}` | yes | `e849ebc4245844a9` |
| `ideal` + `card_error` | `200 {"success":true}` | yes | `c8150a51c98f4110` |
| `bancontact` + `card_error` | `200 {"success":true}` | yes | `b9fc46ea6b054881` |
| `some_new_method` + `card_error` | `200 {"success":true}` | yes | `3cc3d99c56f247be` |

So the live merged build still has the intended split: **7/7 log lines, 5/7 Sentry envelopes**, and
the client still sees the identical success contract in every case. The suppressed pair retained the
forensic `Client-confirm browser error for order 1: {...}` logger rows; they simply had no following
`[Transport] Sending envelope` row in the container log.

Also re-ran the fail-open predicate in the preview's own Rails runtime. The denylist is still
`["card", "link"]`, and exactly those two lowercase values with `stripe_error_type: "card_error"`
return `false`. Lookalikes and future/unknown methods return `true`: `"Card"`, `" card "`,
`"CARD"`, `"Link"`, `ideal`, `bancontact`, `some_new_method`, blank method, and `card` with
`invalid_request_error` all report.

Confirm-error reporting behaviour on the hosted preview app, re-verified at 41fa61fc2d85

This supersedes the earlier panel only because the served revision changed; the outcome is the same.

## Test Results

`bundle exec rspec spec/controllers/orders_controller_spec.rb -e "POST confirm_error"` — **17
examples, 0 failures.**

Spec results

The specs are proven load-bearing by mutation rather than by assertion count — each of these was run:

| Mutation | Result |
|---|---|
| Revert the payment-method gate (pre-fix code) | 3 failures |
| Swap the `&&` operand order | 1 failure |
| Drop the `card_error` requirement | 1 failure |
| Add `""` to the suppressed types (fail closed) | 1 failure |
| Add `ideal` + `some_new_method` to the suppressed types | 2 failures |

**Two examples were caught pinning nothing, in two separate review rounds, and only mutation
testing could see it.** Round 1: the blank-type example was written with
`stripe_error_type: "invalid_request_error"`, which short-circuits at the attempt check and never
reaches the denylist — it stayed green with blank suppressed. Round 2: the same defect was still
present in the `notifies for <redirect method>` examples and the unrecognised-method example, which
omitted `stripe_error_type` entirely. Denylisting `ideal` and `some_new_method` reddened
**nothing** — the examples whose whole purpose is proving a newly added payment method stays
visible were guarding an early return. Both now send `card_error`; verified both ways, 0 failures
against the old specs and 2 against the fixed ones.

## Pre-merge review

Two rounds of local adversarial review (fable-5, static analysis) ran before this PR was opened.
Round 2's verdict is **READY-TO-MERGE**.

| Finding | Round | Disposition |
|---|---|---|
| P2 — gating on `payment_method_type` alone makes card-typed `invalid_request_error` log-only | 1 | Fixed; sized at 38/597 events against the live sample |
| P3 — comment credited `StripeChargeProcessor` with persisting the code | 1 | Fixed — it is `Purchase::ChargeEventsHandler#handle_event_failed!`; round 2 re-verified the path |
| P3 — blank `payment_method_type` pinned only incidentally | 1 | Fixed, then fixed again when round 1's own example proved vacuous |
| P3 — redirect-method and unrecognised-method examples never reach the denylist | 2 | Fixed; see the mutation table above |
| P2 — comment restates the constant above it; population figures duplicated in comments | 2 | Fixed — trimmed, figures live in this body and the evidence image |
| P3 — "the webhook above" points at another constant's comment | 2 | Fixed — names `payment_intent.payment_failed` inline |
| P3 — `CLAUDE.md` and `AGENTS.md` prescribe contradictory comment conventions | 2 | Pre-existing on `main`, out of scope here |

What the review **confirmed safe**, which matters as much as what it found:

- **No new flood vector.** Suppressed events skip the per-order counter, but every notify still
  increments it first, so the ceiling stays 5 notifies per order per window — a caller controlling
  both params can only choose *fewer* Sentry events, never more.
- **Fail-open on adversarial input.** Every param goes through `.to_s.first(N)`, so `"Card"`,
  whitespace, unicode lookalikes, arrays and hashes all land on the *report* side and none can
  raise. A caller can suppress their own report by sending `card`+`card_error`, but the
  unconditional log line still records it and they could simply not call the endpoint.
- **`link` really is server-recorded here.** It is absent from
  `CLIENT_REDIRECT_PAYMENT_METHOD_TYPES`, so it takes no redirect leg in this integration.
- **Stripe Connect direct charges are not a gap.** Connected-account declines route
  `payment_intent.payment_failed` through the Connect webhook endpoint to the same processor.

The strongest surviving case against this change: during a Stripe-webhook or Sidekiq outage,
card/link declines would produce neither a Sentry event nor a `stripe_error_code` row until
redelivery, and nobody is paged by a log line. That is the accepted trade — the alternative is
paging on 372 events/day we already have rows for — but it is real, and it is the scenario where
this suppression could hide a live problem.

Only verifiable in production: that both webhook endpoint subscriptions actually include
`payment_intent.payment_failed` (the code handles it; the dashboard config is not in this repo),
and the Sentry volume effect itself.

## QA

No rendered surface exists to click through, so this was verified against the deployed preview app
directly — `https://fix-confirm-error-report-only-in.apps.staging.gumroad.org` at revision
`b155cf6166ef` (see the evidence image above). To reproduce:

1. `POST /orders/<confirm token>/confirm_error` on the preview with
   `payment_method_type=card`, `stripe_error_type=card_error` → `200 {"success": true}`, a
   `Client-confirm browser error for order <id>` line in the puma log, and **no**
   `[Transport] Sending envelope` line after it.
2. Same request with `payment_method_type=ideal` → log line **and** an envelope.
3. Same request with `payment_method_type=card`, `stripe_error_type=invalid_request_error` → log
   line **and** an envelope, because no webhook records that failure.
4. `payment_method_type` blank / `some_new_method` → log line **and** an envelope (fail-open).

Locally:

1. `bundle exec rspec spec/controllers/orders_controller_spec.rb -e "POST confirm_error"` → 17
   examples, 0 failures.
2. Suppression path: `POST /orders/:id/confirm_error` with `payment_method_type=card` and
   `stripe_error_type=card_error` → response `{ success: true }`, the
   `Client-confirm browser error for order <id>` line present in the log, and **no**
   `ErrorNotifier.notify`.
3. Reporting path, same request with `payment_method_type=ideal` → log line **and** notify.
4. Blind-spot path, `payment_method_type=card` with `stripe_error_type=invalid_request_error` →
   log line **and** notify, because no webhook records that failure.
5. Post-deploy, the `GUMROAD-12F` fingerprint should fall to roughly a third of its current daily
   volume, and surviving events should be dominated by blank-type and non-`card_error` shapes.

## Notes for the reviewer

Two deliberate choices worth flagging, since the opposite of each looks more natural:

- **Denylist, not an allowlist of redirect methods.** An unrecognised or blank `payment_method_type`
  keeps reporting. Blank is 31% of live traffic (187 of 597) and consists of the error shapes that carry no
  `payment_method` at all — precisely the ones most likely to be server-invisible — and an allowlist
  would also silently unmonitor any payment method added after this code was written.
- **Suppress rather than split the Sentry fingerprint.** Splitting alone would retain 600–1,300
  events/day whose payload is already in `purchases.stripe_error_code`, and splitting on the raw
  50-character client-supplied param is what the existing per-order cap exists to prevent. A
  *bounded* split of what remains is a reasonable follow-up and is recorded on the tracking issue,
  not done here.

---

**AI disclosure:** Generated with `claude-opus-5` running as Gumclaw's daily Sentry triage sweep. The
agent enumerated unresolved issues in the `gumroad` Sentry project, established that this
fingerprint's volume was neither a regression nor a currency bug by sampling 44 post-deploy events
through the project-scoped event API, traced the `payment_intent.payment_failed` →
`purchases.stripe_error_code` path in this repo to confirm the premise, wrote the change and its
specs, and proved them load-bearing with the five mutations above. Two local fable-5 adversarial
review rounds ran; round 1 returned SHIP-WITH-FIXES and round 2 READY-TO-MERGE, and every finding's
disposition is in the table above. One statistic it supplied ("38 of 598
events") is correct and was verified independently by re-deriving the cross-tab from the 597 stored
event payloads; an intermediate commit of mine had wrongly replaced it with a narrower 44-event
window, which is reverted.


gumclaw:
@rmarescu — handing this to you rather than arming auto-merge, because the judgement in it is yours rather than mine.

Following on from gp#1382: your fix [#6497](https://github.com/antiwork/gumroad/pull/6497) is confirmed working in production. Across 597 post-deploy events the currency-mismatch slice is **zero**, against 17 of 100 before the deploy. That part is settled.

What this PR addresses is what is left in `GUMROAD-12F` after that: the fingerprint is still running 600–1,300 events/day, and none of it is the iDEAL redirect leg the instrument was built for. The 597-event cross-tab:

| `payment_method_type` | `stripe_error_type` | count |
|---|---|---:|
| `card` | `card_error` | 302 |
| *(blank)* | `card_error` | 177 |
| `link` | `card_error` | 70 |
| `card` | `invalid_request_error` | 38 |
| *(blank)* | `invalid_request_error` | 10 |

Zero `ideal`, zero `bancontact`, in fourteen days.

**Why this is your call and not an auto-merge.** It reduces alerting on the checkout lane. I am confident about the mechanism — a card `card_error` is durably in `purchases.stripe_error_code` via `payment_intent.payment_failed`, so suppressing it loses nothing we cannot query — but *how much signal is acceptable to give up on a payments surface* is a decision that belongs to whoever owns the lane, particularly while the client-confirm ramp is still in progress. If you would rather keep the width until the ramp is finished, say so and I will close this; the instrument being noisy is a smaller problem than the ramp losing forensics.

**One thing worth your eye specifically.** My first version suppressed on `payment_method_type` alone, and the fable-5 review was right to reject it: Stripe attaches `payment_method` to any error involving one, not just declines, so a card-typed `invalid_request_error` — consumed ConfirmationToken, intent in an unexpected state — never transitions the intent, never webhooks, and would have gone log-only. That is the 38-event row above, and it is exactly the class this endpoint exists to catch. Suppression now requires the method **and** `card_error`. If you think there is a *third* shape that reaches the browser without a webhook, that is the review comment I most want.

**Deliberately not done here:** the bounded per-method fingerprint split. Splitting on the raw 50-character client-supplied param is what the existing per-order cap exists to prevent, so the safe version is its own change. It is recorded on antiwork/gumroad-private#1514 along with the open question of why the iDEAL leg has never emitted a single event.

CI is green at `d08efb03` (76 success, 0 failures). Specs are 17 examples / 0 failures with four mutations each proven to bite — including one that caught a vacuous example of my own.

gumclaw:
### The three red `Test Slow` jobs here are not this branch

CI on this PR is failing only these three `spec/requests/purchases/product/taxes_spec.rb` examples:

- `:3904`, `:3929`, `:4074` — "country change scenarios"

all with the same error out of `checkout_helpers.rb:241`, waiting on a Stripe 3DS challenge iframe that never appears:

```
expected to find css "iframe[src^='https://js.stripe.com/v3/three-ds-2-challenge']" but there were no matches
```

**`main` itself fails the same example** at `3d0f55e4f` — run [30470868964](https://github.com/antiwork/gumroad/actions/runs/30470868964), job `Test Slow 13`, with none of this branch's code in the tree. Two other unrelated PRs fail the identical three examples. I reran the failed jobs once here per the flake rule and they reproduced, so it is deterministic rather than a flake.

Tracked repo-wide in #6562 (leading hypothesis: a Stripe-side change in whether the Mexico test card is challenged). Nothing to fix on this branch — please read the rest of the run, which is green, and don't bounce this as a broken branch. I have deliberately **not** armed auto-merge.

gumclaw:

gumclaw:

gumclaw:
CI is green at the current head (`41fa61f`): 76 successful checks, no failures, no pending.

The earlier red was the shared Stripe test account throttling us, not this diff. Twelve Slow shards failed at the same assertion — `expect(page).to have_alert(text: "Your purchase was successful!...")` — with `[Stripe] rate limited — the shared Stripe test account is throttling us` in the logs, and four unrelated PRs that pushed in the same window (`fix/blocked-external-account-is-terminal`, `fix/dispute-evidence-backstop-sweep`, and two others) failed identically. Rerunning only the failed shards cleared all of them with no change to the branch.

Worth stating explicitly since this diff does touch `orders_controller.rb`: the change narrows `ErrorNotifier.notify` inside `confirm_error`, which is the endpoint a browser calls *after* a confirm has already failed. It cannot affect whether a purchase succeeds, so it could not have produced a failing "purchase was successful" assertion in the first place.

No code was pushed for this — head is unchanged.

rmarescu (approved):
Fixed local review findings:

- bccef55d6: the final `ideal` post in the notify-budget example now sends `card_error`, so it reaches the payment-method denylist instead of returning early at the attempt check.
- 5a422a7c6: the tracking issue reference reads `gumroad-private#1514`, which links from the code.

> If you think there is a *third* shape that reaches the browser without a webhook, that is the review comment I most want.

I did not find one that this gate suppresses. `payment_intent_authentication_failure` and an async Link failure both arrive either as a non-`card_error` type or as the webhook itself, so they stay reported.

I also checked one thing you did not claim. Suppression relies on `client_confirmed_charge?`, which needs an actual `Charge` record. `Order::PreparePaymentIntentService#build_charge` creates that `Charge` with `client_confirmed: true` during `prepare`, and `confirm_error` is only reachable after `prepare`. So every suppressed failure has one and `stripe_error_code` does get written.

gumclaw:
Thanks — both commits read right to me, and I checked the one you volunteered rather than taking it on trust.

**`bccef55d6`** is the fix I should have caught. `confirm_error_reportable?` returns `true` at the attempt check for any type that is not `card_error`, so the old final post never reached the membership check and the example's name was a claim it wasn't testing. I did not re-run your mutation — the guard makes it read-verifiable: with no `stripe_error_type` the method returns before `CONFIRM_ERROR_SERVER_RECORDED_PAYMENT_METHOD_TYPES` is ever consulted, so adding `ideal` to that list cannot fail the old form. That is the third of this shape on this PR, which says the vacuous-example risk is structural here, not incidental — every example in these two contexts has to send `card_error` or it proves nothing.

**`5a422a7c6`** — agreed, and `gumroad-private#933` two lines up in the same comment block made the shorthand inconsistent on its own.

On the `Charge` existence check: it holds for every shape that actually arrives, and I traced the one adjacency it depends on rather than just the happy path. `build_charge` is the only writer of `client_confirmed: true`, and the `scope: "confirm"` token it mints in `build_confirmation_responses` is not the only one — `Order::ChargeService` mints the same-scope token on the `requires_card_action` / `requires_card_setup` server-confirm legs, and `Order::CreateService` does for subscription-restart SCA. Those orders' charges come from `order.charges.create!(seller_id:)` with no `client_confirmed`, so `client_confirmed_charge?` is false for them and `handle_event_failed!` takes the fallback branch, which writes `stripe_error_code` only for off-session Indian-card purchases; `MarkFailedService` doesn't write it either. So a suppressed report on one of those tokens *would* be a genuine hole.

It isn't reachable: `reportClientConfirmError` has exactly one call site (`app/javascript/data/order.ts:307`, inside the client-confirm branch), and it passes prepare's own order id. So every request this endpoint sees today carries a client-confirmed charge, and your conclusion stands. Worth knowing that it rests on the caller being singular rather than on the token scope, since the scope is shared — if a future SCA path posts to this endpoint, suppression would need to gate on the charge, not the params.

The `payment_intent.payment_failed` coverage behind the suppression is on both delivery paths, platform and connect (`StripeEventHandler#handle_event_for_gumroad` and `#handle_event_for_connected_account` both route it), so a direct-charge seller's declines are recorded server-side the same as ours.

Nothing left from my side. CI is re-running on `5a422a7c6`; the merge is yours whenever you want it — I'm not arming auto-merge on a payments-surface change.

PR #1532, opened 2025-10-05
Migrate `_alert.scss` to Tailwind CSS as part of issue #1055.

SamarthShukla17:
## Changes Made
- ✅ Removed `_alert.scss` file completely
- ✅ Removed alert imports from pack files:
  - `app/javascript/packs/admin.scss` (line 8)
  - `app/javascript/packs/design.scss` (line 8)  
  - `app/javascript/packs/email.scss` (line 11)
- ✅ Updated `Alert.tsx` React component to use Tailwind utilities
- ✅ Added proper SVG icons for all alert types
- ✅ Maintained visual parity for all variants:
  - Success alerts (green theme with check icon)
  - Danger alerts (red theme with X icon)
  - Warning alerts (orange theme with warning icon)
  - Info alerts (purple theme with info icon)
- ✅ Preserved all existing functionality (auto-dismiss, positioning)

## Testing Completed
- [x] Success alerts render correctly with proper styling
- [x] Danger/error alerts render correctly with proper styling
- [x] Warning alerts render correctly with proper styling
- [x] Info alerts render correctly with proper styling
- [x] Icons display with correct colors and positioning
- [x] Auto-dismiss functionality works (5-second timer)
- [x] Responsive behavior maintained on mobile/desktop
- [x] No visual regressions detected
- [x] All alert variants tested in different contexts

## Screenshots

Before: 
Light Mode:
Screenshot from 2025-10-06 21-11-06

Dark Mode:
image

image

After 
Light mode:
Screenshot from 2025-10-06 21-10-37
Dark Mode:
image
Screenshot from 2025-10-06 21-09-31

## Screen Record
[Screencast from 2025-10-06 22-28-49.webm](https://github.com/user-attachments/assets/5dbe5637-8b9f-4f66-a971-c69c013d5f92)

Closes part of #1055

PR #6559, opened 2026-07-29
Let the editor retry a deletion the save contract refused as stale

gumclaw:
## What

The product editor now recognises the save contract's `stale_deletion_conflict` and tells the seller their deletion was not applied, instead of rendering it as a generic error toast.

It does **not** retry the deletion. An earlier revision of this PR did, by adopting the fresh `editor_revision` the 409 carries and resending the payload; that was unsafe and is removed. See "Why there is no retry" below.

## Why

Third acceptance criterion of #6559's linked issue. `LinksController#update` answers a stale-token deletion with 409 + `error_code`, but `saveProduct` special-cased only `hidden_variant_content_conflict` and `stale_content_conflict`, so `stale_deletion_conflict` fell through to `throw new ResponseError(...)`. The seller got a transient auto-dismissing red toast and no recovery affordance — no persistent statement of what happened or what to do next.

> **Correction, from the hosted-preview capture in #6574 (added after merge).** An earlier version of this paragraph said the toast "never said the deletion hadn't happened." That was wrong — the toast renders the server's `error_message` verbatim, including "the items you removed were not deleted." The defect is the transience and the missing recovery path, not the absence of the words.

The two 409s are different situations and status alone cannot discriminate them:

| conflict | server state | what the seller must be told |
|---|---|---|
| `stale_content_conflict` | stored rows are **newer** than the payload | saving would overwrite newer content; reload |
| `stale_deletion_conflict` | **nothing written**; the deletion did not happen | the deletion did not happen; reload and delete again |

Both end in a reload, but conflating them is what produced the unhelpful message: a seller who deleted something needs to know it is still there.

## Why there is no retry

The 409 carries a fresh `editor_revision`, and the comment on that response says the token rides along so the editor can "reconcile and retry rather than forcing a full page reload." Adopting it and resending does not achieve that. It corrupts data.

Deletions are gated on the token. Ordinary writes are **not** — deliberately, so a stale tab fixing a typo keeps working. The editor save is a full product snapshot. So swapping in a fresh token produces a save that applies the deletion *and* reverts every field another session changed in between. `Product::StaleContentWriteGuard` is the guard for the overwrite half, and it is observe-only by default, so nothing catches it.

Concretely: tab A loads versions X and Y, tab B renames X, tab A deletes Y and gets the 409, tab A retries — Y is deleted and X's new name is silently reverted to tab A's stale value. The seller confirmed deleting Y, not overwriting X.

Two routes to a safe retry, both server-side, tracked in gumroad-private#1532:

- **A deletion-only request.** Note this cannot be done by having the client strip fields: `tags`, `covers`, `section_ids` and `shipping_destinations` are read as `params[:thing] || []` in `LinksController#update`, outside the contract's five collections, so a trimmed payload wipes them.
- **Re-fetch and reconcile** before adopting the revision.

Until one exists, the reload is the honest recovery. It costs the seller their unsaved edits, which is worse than a working retry and much better than a silent overwrite.

Also removed: the write-back of the fresh token onto product state. That laundered a stale session into an accepted one even when the seller dismissed the modal without retrying.

## The change

- `saveProductError`, a pure function over the error payload, replaces the inline `if` chain. Every conflict is discriminated by `error_code`, never by HTTP status.
- `StaleDeletionConflictError` is a bare `Error`. It deliberately does not expose the revision — the field is absent rather than present-and-unused, so the one unsafe use cannot be reintroduced by an edit that looks reasonable in isolation. `SaveProductErrorPayload` keeps the key with a comment saying why nothing reads it.
- `ProductEditPage` renders its own modal for this conflict ("Your deletion was not applied"), stating that nothing was saved including the removals, and offering the reload.

## Before / after

Captured on the hosted preview app serving this branch's code, in #6574 (the PR merged before the media landed here):

| | |
|---|---|
| **Before** (main's mapping) | before |
| **After** | after |

Walkthrough of the two-session flow:

https://github.com/antiwork/gumroad/raw/7ba0b853edeb26aaacb0b79b63733d423f8b8a58/qa-media/pr-6559-two-session-walkthrough.mp4

Two independent logged-in editor sessions against the seeded `demo` product with `product_editor_save_contract` on. Session B saved `200`; session A got `409 stale_deletion_conflict` and rendered the modal. Server state read back afterwards — `["Version Y", "Version X (renamed by session B)"]` — confirms the refused deletion did not happen and B's rename was not reverted.

## Test Results

**`test/controllers/links_controller_test.rb`** — the two-session regression test, driven through the real save path:

```
LinksControllerSaveContractTest#test_flag_on:_resending_a_stale_snapshot_with_the_409's_fresh_token_
deletes_as_asked_AND_reverts_the_other_session's_edit = 0.73 s = .

1 runs, 5 assertions, 0 failures, 0 errors, 0 skips
```

It passes by **demonstrating** the overwrite — Y deleted as requested, X's rename reverted, HTTP 200, nothing reported — so it documents why no client-side retry exists. If someone re-adds a token-swap retry, this is the test they have to argue with. Run with the neighbouring contract tests: `3 runs, 12 assertions, 0 failures`.

This test lives in the controller test rather than the client on purpose: it is the server's *acceptance* of the retry that causes the overwrite, and a client test can only assert which request was sent. That was the gap in the previous revision, whose tests covered only error mapping, one session, no save path.

**`app/javascript/data/product_edit.test.ts`** — 18 passed, including one asserting the fresh revision never reaches the error object:

```
✓ maps stale_deletion_conflict to its own typed error
✓ does not carry the fresh revision onto the error
✓ keeps the two 409 conflicts distinct
```

The negative assertions are deliberate — an earlier draft used `rejects.not.toBeInstanceOf(...)` and passed on an incidental `TypeError`, so each test asserts which error *is* raised, including that an unrecognised code still degrades to `ResponseError`.

ESLint, Prettier and RuboCop clean on all changed files. `tsc --noEmit` cannot run in my environment — it fails identically on `main` with `TS2688: Cannot find type definition file for 'vite/client'`, verified by stashing — so type checking rests on the language server (no diagnostics) plus CI.

Local system specs (`spec/requests/products/edit/**`) cannot run in this environment either: an untouched pre-existing example from `digital_versions_spec.rb` times out at `visit` the same way, so the screenshots come from the hosted preview app rather than locally.

## QA steps

Requires `:product_editor_save_contract` active for the seller.

1. Open a product's edit page in two tabs.
2. In tab B, rename a version and save.
3. In tab A, remove a *different* version and click Save changes; confirm the deletion summary.
4. Expect the "Your deletion was not applied" modal, saying nothing was saved.
5. Click Reload page; confirm the version is still there and tab B's rename survived.
6. Repeat step 3 and dismiss with "Keep editing" instead; save again and confirm it is refused again — a dismissed conflict must not leave the session authorised to delete.

## Not in this diff

The other two acceptance criteria are server/payload work, deliberately separate:

- identifying the client path that dropped `deletion_operations` on the reported save (the actual silent no-op);
- surfacing a discrepancy when the server applies fewer deletions than the request named, rather than returning a bare 200.

---

## AI disclosure

Written by Claude Opus 5 (`model.default` in the agent's config) acting as gumclaw.

Prompts and instructions given:

1. A GitHub webhook delivered @gianfrancopiana's review comment on this PR, reporting that the retry attaches the returned `editor_revision` to the old editor snapshot and resends the entire stale product, with the four-step tab A / tab B case and the observation that `Product::StaleContentWriteGuard` is observe-only by default; he asked for re-fetch-and-reconcile or a deletion-only retry, plus a two-session regression test through the real save path.
2. Standing instruction to fix review findings directly on the PR branch — read the flagged code, verify the finding rather than taking it on faith, make the change, run the touched area's tests and linters, push, and reply in-thread.

The departure from his suggested fix (removing the retry rather than reconciling) was the agent's call after establishing that neither suggestion is implementable client-side; it is explained on the PR and tracked in gumroad-private#1532.


gianfrancopiana:
@gumclaw The retry is unsafe as written. It attaches the returned `editor_revision` to the old editor snapshot, then `performSave` resends the entire stale product.

Concrete case:

1. Tab A loads versions X and Y.
2. Tab B renames X.
3. Tab A deletes Y and gets the expected 409.
4. Tab A clicks “Save again.”

Because `Product::StaleContentWriteGuard` is observe-only by default, the retry can restore X’s old name while deleting Y. The seller confirmed deleting Y, not overwriting X.

Re-fetch and reconcile the current product before adopting the revision, or make the retry deletion-only so it cannot write stale fields. Add a two-session regression test through the real save path; the current tests only cover error mapping and cannot catch this.

Reviewed with gpt 5.6 Sol.

gumclaw:
@gianfrancopiana you're right, and I've verified it in the code rather than just taking the note — this is mine to fix, so I've swapped `awaiting-human` for `working` and taken the assignee. Nothing for you here until I post a fix.

## Confirming your case, with the reads

Your four-step scenario holds, and the reason it holds is the combination of two things I did not connect:

1. **The retry resends everything, not just the deletion.** In `performSave`, the retry path builds `productSent = structuredClone(productToSave)` and my change only overwrites one field on it — `productSent.editor_revision = freshEditorRevision`. `productToSave` is still this session's snapshot. So Tab A's stale name for version X rides along with the deletion of Y.

2. **Nothing downstream stops the stale field write.** `Product::StaleContentWriteGuard::BLOCK_FEATURE_NAME = :product_editor_stale_content_block` is off by default, and `ensure_fresh!` explicitly `return unless blocking` after notifying — the comment on it says the save then "proceeds exactly as it did before this guard shipped." And the token check that *did* refuse the request, `ensure_contract_deletions_are_fresh!` in `LinksController`, is gated on `contract.requested_deletion?`, so it never looks at ordinary field writes. With a fresh token in hand there is nothing left to refuse.

So adopting the token is precisely the wrong primitive: the token is the server's statement that this session's view is stale, and I used it as permission to write that stale view. My tests can't catch it because they only cover `saveProductError` mapping — no save path, one session.

I also want to flag that the second-order effect is worse than the modal: I write the fresh token back onto `product` state via `updateProduct` so later saves in the session aren't refused. That permanently launders a stale session into an accepted one, even if the seller clicks "Keep editing" and never retries. That line goes regardless of which shape the fix takes.

## Plan

Going with your first option — re-fetch and reconcile — rather than deletion-only, because deletion-only leaves the session still stale and still holding edits it thinks are saveable, which just relocates the bug to the next save.

1. Drop the `updateProduct` token write-back entirely. A refused save must not improve this session's standing.
2. On `stale_deletion_conflict`, re-fetch the product before offering anything. `router.reload({ only: [...] })` is already the house pattern for this (there's a `inertia_partial_reload` util and `Discover/Search.tsx` / `WorkflowsPage/WorkflowList.tsx` use it), and `ProductPresenter#edit_props` already ships `editor_revision` with the product, so the refreshed payload carries its own token — I never have to hand-adopt the one from the 409.
3. Reconcile against the fresh payload and only then offer the retry, with the modal naming what changed underneath rather than the current generic "this product changed" wording. If a pending removal targets a row that has itself changed, that's not retryable and should say so.
4. Regression test through the real save path with two sessions, asserting X's name survives Y's deletion. As you say, that's the test that would have caught this; error-mapping tests structurally cannot.

**ETA: next working session, and I'll re-request your review when it's pushed.** If you'd rather I close this and reopen behind the re-fetch design, say so and I'll do that instead — the current commit isn't worth preserving.

Reviewed my own reasoning against your comment; the mistake was treating a staleness signal as an authorisation.

gumclaw:
Pushed in 98f7cea. I did **not** build the re-fetch-and-reconcile design I described above — I removed the retry instead. Explaining the change of course, because it's a departure from what I told you I'd do.

## Why I stopped short of reconcile

Once I wrote your case out as an actual test through the real endpoint, the retry stopped looking like something with a safe version nearby. Neither of your two suggestions is client-side work:

**Deletion-only doesn't exist as a payload shape.** I checked whether the client could simply strip fields. It can't — `LinksController#update` reads several collections outside the contract's five as `params[:thing] || []`: `save_tags!`, `reorder_previews`, `show_in_sections!`, `save_shipping_destinations!`. A payload trimmed to "just the deletion" wipes the seller's tags, cover order and sections. Deletion-only needs a server-side mode or its own endpoint.

**Reconcile is a real design, not a follow-up commit.** Re-fetching is easy; deciding what "reconcile" means per field is not, and doing it wrong reintroduces exactly this bug with more code in front of it. I'd also be inventing merge semantics without a decision on them.

So shipping either one means leaving the unsafe retry live in the meantime. I'd rather remove the hazard now and design the affordance properly: **gumroad-private#1532**, with the constraint above recorded so nobody retries the strip-the-payload idea.

## What's in the diff now

Delivered from my plan: items 1 and 4. Items 2 and 3 moved to #1532.

- The `updateProduct` token write-back is gone. As I said, that one goes regardless of shape — a refused save must not launder a stale session into an accepted one, and it did that even when the seller clicked "Keep editing."
- `StaleDeletionConflictError` no longer carries `editorRevision` at all. Not "carries it but doesn't use it": the field is removed, so the only unsafe use of that token can't be reintroduced by a later edit that looks reasonable in isolation. The payload type keeps the key with a comment saying why nothing reads it.
- `performSave` loses the `freshEditorRevision` parameter, so its signature is back to what it was.
- The distinct `error_code` handling and its own modal stay — that's the acceptance criterion, and the generic red toast never told the seller their deletion hadn't happened. The modal now says so plainly, titled "Your deletion was not applied", and offers the reload. That costs the seller their unsaved edits, which is worse than a working retry and better than a silent overwrite.

## The test you asked for

`LinksControllerSaveContractTest`, "resending a stale snapshot with the 409's fresh token deletes as asked AND reverts the other session's edit" — your scenario, driven through the real save path:

```
1 runs, 5 assertions, 0 failures
```

It passes by *demonstrating* the overwrite: version Y is deleted as requested, and version X's rename by the other session is reverted, with a 200 and nothing in Sentry. So it's the negative case pinned as documentation of why there's no client retry, not a test of the fix. If someone re-adds a token-swap retry, this test is what they have to argue with.

I put it in the controller test rather than the client deliberately — it's the server's *acceptance* of the retry that causes the overwrite, and a client test can only show which request was sent. That's the gap you identified: my three tests covered `saveProductError` mapping, one session, no save path.

Client tests: 18 pass, including one asserting the fresh revision does not reach the error object.

One thing worth your call on #1532: the 409 still returns `editor_revision`, and right now nothing may safely consume it. It's a token whose only use is the unsafe one. I left it — removing it is a contract change and the reconcile design may well want it — but it shouldn't sit there unexplained forever.

Modal copy changed, so I owe screenshots before this is reviewable; adding them next. I'll re-request your review once they're up.

gumclaw:
Ownership sweep: assigning **@gianfrancopiana** as the human owner of this (PR) — derived from who owns the linked PR / filed it / reviews this surface. Labelled `awaiting-human`: it's with you now.

Automated ownership fix (every item carries exactly one human owner, never the agent, never nobody). If the routing is wrong, just reassign.

gumclaw:
@gianfrancopiana you merged this while I was capturing the preview evidence, so two things land after the fact.

**The QA evidence is in.** #6574 (merged) carries the hosted-preview before/after and a walkthrough of your exact scenario, driven as two independent logged-in editor sessions against the seeded `demo` product with the contract flag on:

| | |
|---|---|
| Before (main's mapping) | before |
| After | after |

Session B saved `200`; session A got `409 {"error_code":"stale_deletion_conflict", ...}` and rendered the modal. Reading the rows back through the preview console afterwards:

```
MARK variants=["Version Y", "Version X (renamed by session B)"]
```

Version Y still alive — the refused deletion didn't happen — and B's rename intact, so A's stale payload reverted nothing. Those are the two facts your review was protecting, confirmed against a running app rather than only in tests.

**A correction to my own body, which the capture disproved.** I wrote that the pre-fix toast "never said the deletion hadn't happened." That's wrong: the toast carries the server's `error_message` verbatim, including "the items you removed were not deleted." You can read it in the before shot. The real defect is narrower than I claimed — the message is a transient auto-dismissing toast with no recovery affordance, so the seller has no persistent statement of what happened or what to do. The modal is the improvement; the words were already there. It's recorded in #6574's body too, since this one is merged and I'd rather not leave an overstated claim standing behind a merge.

**Still open, and it's yours to weigh in on:** gumroad-private#1532 for the safe retry. Worth knowing before that gets designed — I checked whether the client could just strip fields down to a deletion-only payload, and it can't. `LinksController#update` reads `tags`, `covers`, `section_ids` and `shipping_destinations` as `params[:thing] || []`, outside the contract's five collections, so a trimmed payload wipes them. Deletion-only needs a server-side mode or its own endpoint. The issue also asks whether the 409 should keep returning `editor_revision` at all while nothing may safely consume it.

Thanks for catching this one — the retry would have shipped a silent overwrite, and the two-session test that now guards it only exists because you asked for it specifically.

PR #6563, opened 2026-07-29
Trim ClearMistakenBuyerBlocks comments to what a maintainer needs

gumclaw:
## What

Comment-only trim of `app/services/onetime/clear_mistaken_buyer_blocks.rb`. **No behaviour change** — zero non-comment lines in the diff (verified: `git diff origin/main | grep -v '^[+-]\s*#'` is empty apart from the hunk headers).

- 145 lines removed, 60 added — **net −85 lines**, file 292 → 207.
- The class header went from 40 lines to 15; the six method preambles from 20–30 lines each to 4–8.
- Not touched: any logic, any name, any ordering, any whitespace outside the comment text.

## Why

Per the comment convention (repo `AGENTS.md`): comments are for a colleague who already knows this codebase. This file had drifted into essays — each private method carried a multi-paragraph narration of the reasoning that produced it, and the class header restated the seven conditions that the six methods below already state individually. Long comments that restate the code are what teach people to skip the comments that matter.

What went, by category:

- **Bug-report narration.** "Long-standing customers were left unable to pay us with any card or through PayPal, with nothing on screen that explained why, and at least one membership lapsed because of it." The issue link carries this.
- **The duplicated checklist.** The header's seven-bullet list of conditions, each of which is explained again at the method that implements it. Replaced with a one-sentence statement of the risk and the four gates, since that is the thing you need before reading on.
- **Narrated alternatives.** "Using created_at got this wrong in both directions: ... never matched any window and was left blocked by a run that reported the buyer cleared, and a row the bug wrote long ago and a velocity rule re-blocked last month still carried its old created_at, so it looked like part of the old burst and would have been cleared out from under live enforcement." Compressed to the two failure directions in one sentence.
- **Restating the line below.** `# How many of the buyer's cards to reconstruct as candidates ... See #possible_recent_stripe_fingerprints.` over `RECONSTRUCTED_FINGERPRINT_LIMIT = 10`, and `# Returns one row per buyer we cleared ...` over a method that visibly returns exactly that.

## What I deliberately KEPT

Every fact that is hard to re-derive from the code, just stated shorter:

- **Why `blocked_at` and not `created_at`** — `PlatformBlock.add!` is `create_or_find_by!` plus an update, so `created_at` is the row's first-ever insert. This is a live trap: the next editor "simplifying" this to `created_at` reintroduces the bug in both directions.
- **Why the window is two-staged** — purchases have no `failed_at` and an SCA charge fails up to `ChargeProcessor::TIME_TO_COMPLETE_SCA` later, so the search must span that period; but a window that wide swallows unrelated blocks, hence the inner burst filter.
- **The `matches_block_buyer_signature?` rationale** — that sharing an IP via carrier NAT / an office / a VPN exit is otherwise enough to clear a stranger's suspension block. Non-obvious and the whole reason the method exists.
- **Why the velocity re-run is fail-closed** — one row per `(type, value)` means the mistaken block and a deliberate velocity block can be literally the same row.
- **The `purchaser_id = ?` NULL trap** — dead on a guest checkout because a SQL comparison against NULL is never true, and a NULL-safe `<=>` would match every guest at once. Kept verbatim in intent: this is the sharpest edge in the file.
- **The full enumeration of other unattended `blocked_by: nil` writers**, since that list is what justifies `BLOCK_BUYER_ONLY_TYPES` and cannot be recovered without grepping four call sites.
- **The `process` dry-run-vs-real asymmetry**, because it changes how you read the output.
- **The console usage block**, which is the entry point for a one-off task.

## Test Results

- `ruby -c` — Syntax OK
- `bundle exec rubocop app/services/onetime/clear_mistaken_buyer_blocks.rb` — 1 file inspected, **no offenses**
- `bundle exec rspec spec/services/onetime/clear_mistaken_buyer_blocks_spec.rb` — **20 examples, 0 failures** (13.6s, local, clean env)

### On branch CI

Branch CI is red on three examples in `spec/requests/purchases/product/taxes_spec.rb` (`:3904`, `:3929`, `:4074`), all the same assertion timing out:

```
expect(page).to have_selector("iframe[src^='https://js.stripe.com/v3/three-ds-2-challenge']", wait: 240)
```

**This is pre-existing on `main`, not from this PR.** The `Tests` run for `main` at `3d0f55e4f` (run [30470868964](https://github.com/antiwork/gumroad/actions/runs/30470868964)) fails those exact three examples and nothing else. They are Stripe-hosted 3DS challenge iframes failing to load in CI; this branch cannot reach them — it changes comment text in a console-only `Onetime::` service with no callers in the tax or checkout paths. Everything else on the branch is green.

Worth a separate look at those three by whoever owns the 3DS test setup, but nothing to fix here.

---

Comments only, no behaviour change. Needs a skim, not a review.


gumclaw:
Comments only — 145 lines removed, 60 added, net −85; zero non-comment lines in the diff. The class header went from 40 lines to 15 and each private method's preamble from 20–30 lines to 4–8.

What went was the narration: the bug-report retelling, the seven-condition checklist in the header that the six methods below each restate, and the walked-through wrong alternatives. What stayed is every fact you can't re-derive from the code — the `blocked_at` vs `created_at` trap, why the failure window is two-staged around SCA, the carrier-NAT/VPN reason `matches_block_buyer_signature?` exists at all, the one-row-per-(type,value) collision with velocity blocks, and the `purchaser_id = ?` NULL-comparison edge (a "tidy-up" to `<=>` there would lift strangers' live blocks).

Local: rubocop clean, `clear_mistaken_buyer_blocks_spec.rb` 20/20 green.

Branch CI is red on three `taxes_spec.rb` examples waiting on the Stripe 3DS challenge iframe — those same three fail on `main` at 3d0f55e4f (run 30470868964), so it's pre-existing and unreachable from a console-only `Onetime::` service. Flagging rather than touching it.

Needs a skim, not a review.

gumclaw:
Heads up before you read the red X: **the three failing jobs here are not this branch.**

Test Slow 10/16/21 all fail the same three `spec/requests/purchases/product/taxes_spec.rb` "country change scenarios" examples (:3904, :3929, :4074), all inside `within_sca_frame` at `spec/support/checkout_helpers.rb:241`, waiting on a Stripe 3D Secure challenge iframe (`iframe[src^='https://js.stripe.com/v3/three-ds-2-challenge']`) that never renders. It waits the full 240s and burns all three rspec-retry attempts, so it is deterministic rather than flaky.

This is broken on `main` itself — proven at run 30470868964 (commit 3d0f55e4f, Test Slow 13) with zero branch code in the tree. It stayed hidden because main's per-commit runs shard Test Slow via Knapsack, so most green main runs never scheduled these examples.

This PR changes exactly one file — `app/services/onetime/clear_mistaken_buyer_blocks.rb`, comments only — and cannot touch country selection or 3DS.

Tracked with a proposed fix decision in #6562 (assigned to @rmarescu). Everything else here is green: 73 checks passing, mergeable, no conflicts. Please review on the diff and don't bounce it as a broken branch.

gumclaw:

gumclaw:

nyomanjyotisa (approved):

PR #6566, opened 2026-07-29
Record processor rate limits as a payout failure reason and stop them pausing sellers

gumclaw:
## What

Stripe rate-limit errors (`Stripe::RateLimitError`, HTTP 429) raised while preparing or performing a payout now record `Payment::FailureReason::PROCESSOR_RATE_LIMITED` instead of failing with a nil `failure_reason`, and no longer count toward `MAX_CONSECUTIVE_FAILED_PAYOUTS`.

Three changes:

- `StripePayoutProcessor` gains a `Stripe::RateLimitError` rescue ahead of the general `Stripe::StripeError` rescue, on both `prepare_payment_and_set_amount` and `perform_payment`. It sets the reason, reports to `ErrorNotifier`, and returns the message as an error like the sibling rescues do.
- `Payment::FailureReason` gains the constant plus a `TRANSIENT_REASONS` list, for failures caused by us rather than by the seller's payout details.
- `Payment`'s consecutive-failure pause check excludes transient reasons.

## Why

During the July 28 payout batch, Stripe returned 429s and 32 sellers' payouts failed for $15.4K combined. Every one of those rows landed with `failure_reason` nil, so the cause existed only in `error_message` free text and every monitor keyed on `failure_reason` read the batch as unexplained failures. Tracking issue: antiwork/gumroad-private#1523.

Two things made that worse than a transient outage:

Stripe's client-side retries deliberately exclude 429s, so a rate limit reaches our rescue on the first response rather than after a retry — there is no backstop above this code.

More seriously, these failures were being counted against the seller. `Payment.check_for_payout_failures` pauses a seller after `MAX_CONSECUTIVE_FAILED_PAYOUTS` and leaves them a payout note asking them to fix their bank account. A 429 says nothing about the seller's bank details, so counting it pushes sellers toward a pause for something they cannot act on and could not have caused.

The `IS NULL` arm of the new scope is load-bearing rather than defensive: `failure_reason NOT IN (...)` alone drops NULL rows, and NULL is the common case for existing failures — writing it without the null arm would silently disable the pause check entirely.

## Before / After

Not a UI change; no rendered surface is touched. The observable difference is in the `payments` rows a rate-limited batch produces, and in whether that batch moves a seller toward a pause.

| | Before | After |
|---|---|---|
| `failure_reason` on a 429 failure | `nil` | `processor_rate_limited` |
| Counts toward the pause threshold | yes | no |
| Seller-visible payout note | posted | not posted |
| Bank account after `perform_payment` 429 | preserved | preserved (unchanged) |
| `ErrorNotifier` | not called | called |

## Test Results

Both suites, since this touches a model and a business-layer processor covered by Minitest. Re-run at the current head — the counts below supersede an earlier revision of this section that cited 11 examples, which was accurate before `5d4a635b` widened the change from rate limits to the whole transient-failure class:

- `spec/models/payment_spec.rb` — the `pausing payouts after repeated failures` group, **13 examples, 0 failures** locally. Four are new: a rate-limited failure does not count toward the threshold, an unreachable-Stripe failure does not either, the threshold still trips on non-transient failures alone, and failures whose reason is set but not transient keep counting.
- `test/business/payments/payouts/processor/stripe/stripe_payout_processor_test.rb` — **315 runs, 771 assertions, 0 failures, 0 errors, 0 skips**. Three of those cases are new: the reason is recorded on `prepare_payment_and_set_amount`, no seller-visible payout note is posted, and on `perform_payment` the reason is recorded and the bank account survives for the next run.

Both suites green at the current head, plus the mutation proof

The example worth reading is `still pauses when the threshold is reached by non-transient failures alone`. It is what pins the `IS NULL` arm, and it is not a formality — the whole pause check is disabled without it.

**Mutation proof**, re-run at the current head. Dropping `IS NULL OR` from the scope, changing nothing else:

```
# with the fix as written
13 examples, 0 failures

# with "failure_reason IS NULL OR failure_reason NOT IN (?)" reduced to "failure_reason NOT IN (?)"
13 examples, 4 failures
  ./spec/models/payment_spec.rb:893 # pauses payouts and flags the account once the consecutive failure limit is reached
  ./spec/models/payment_spec.rb:905 # pauses when a previously completed payout to the bank account is later returned
  ./spec/models/payment_spec.rb:940 # still pauses when the threshold is reached by non-transient failures alone
  ./spec/models/payment_spec.rb:979 # pauses PayPal payouts after repeated failures to the same PayPal account
```

Three of those four are pre-existing examples, which is the point: without the NULL arm this change would have silently switched the pause check off for every seller, not just for rate-limited ones. The mutation was applied by a script that restores from a pristine copy in its own last line; the working tree was verified clean (`git status --short` empty, guard present) before the evidence commit.

## Pre-merge review

A local fable-5 adversarial review ran against `36f0f11` and returned **CHANGES-REQUIRED, test-side only — the production code checks out**. Every finding is dispositioned below and the fixes are in `11e7d8f`.

|---|---|---|
| The `perform_payment` test's `expects(:reverse_internal_transfer!)` is vacuous — `add_gumroad_held_balances_negative` sums to −$10, so no internal transfer exists and the method early-returns on a nil `stripe_internal_transfer_id`. The mock pinned that `ensure` invokes a no-op, which the pre-fix generic `StripeError` rescue did identically. | P2 | Fixed — assertion dropped and the test retitled to what it actually pins (failure reason + bank account survival). |
| That test replayed a neighbouring test's cassette, against CONTRIBUTING's VCR rule and this file's own naming convention. | P2 | Fixed — own cassette under `.../the_external_transfer_is_rate_limited/records_the_failure_reason`. |
| `ErrorNotifier.notify(e)` ignores the precedent 13 lines up, where the `InvalidRequestError` rescue suppresses Sentry for known reasons. | P3 | **Not taken.** The review confirmed this is not a new flood — the generic `StripeError` rescue already notified, so a rate-limited batch produces exactly the Sentry volume it did before. Suppressing it would be a behaviour change beyond this PR's scope, and a 429 storm is something we want to see. |
| Rescue comment justifies the change to a reviewer rather than warning a maintainer, and duplicates the contract on `TRANSIENT_REASONS`. | P3 | Fixed — trimmed to the one line the code cannot say. |
| `keeps counting failures that carry no failure reason at all` duplicates the pre-existing example at :893. | P3 | Fixed — folded its comment into the example at :931, which is the real pin. |

Things the review checked and cleared, worth recording because they are the ones that would have hurt:

- **Rescue reachability.** `Gemfile.lock` pins stripe 12.5.0, where `RateLimitError < StripeError` directly and every HTTP 429 maps to it. It is not a subclass of `InvalidRequestError`, `AuthenticationError`, or `APIConnectionError`, so no earlier rescue in either method can swallow it.
- **`ensure` plumbing.** Both methods declare `failure_reason = nil` at the top, so the rescue writes the same method-scope local the `ensure` reads.
- **No seller-visible note.** `add_payment_failure_reason_comment` returns early on `STRIPE_FAILURE_SOLUTIONS[reason]` being nil, before any comment is built — so promoting the reason from nil to present does not start posting notes.
- **Nothing else needs the exclusion.** `failed_cannot_pay`, the cannot-pay and debit-card-limit emails, the PayPal retry worker and the chargeback pause release all key on other reasons or on the pause comment, which this path never writes. The admin view shows the reason verbatim, which is an improvement over "None provided".
- **The safety check is not weakened.** A broken bank destination produces `InvalidRequestError`s or webhook failure codes, never persistent 429s, and rate-limited rows do not reset the consecutive count — they are only uncounted, so a seller alternating 429s and genuine failures still reaches the threshold.

## QA

CI is green at `5d4a635b` (30 checks passing, 0 failing). The earlier red on this branch was the shared Stripe test account throttling us — ten Slow shards failing at the same checkout assertion with `[Stripe] rate limited` in the logs, alongside four unrelated PRs that pushed in the same window. Rerunning the failed shards cleared it with no change to the branch.

Commits since that green head are **evidence-only** (the `qa-media/` screenshot above and an attribution correction) — no application or spec file is touched, so the green result still describes this code.

To verify locally:

1. `bundle exec rspec spec/models/payment_spec.rb -e "pausing payouts after repeated failures"`
2. `DISABLE_SPRING=1 bundle exec rails test test/business/payments/payouts/processor/stripe/stripe_payout_processor_test.rb`

In production, the check is that the next rate-limited batch produces `payments` rows with `failure_reason = 'processor_rate_limited'` rather than nil, and that no seller in that batch receives a payout note or moves toward a pause.

## Notes for the reviewer

The retry question is deliberately **not** in scope here. gumroad-private#1523 raises both "no reason recorded" and "no retry", and this PR only addresses the first. Retrying a payout is money movement and needs its own decision about idempotency and backoff; recording the reason and not penalising the seller is the part that is safe to ship independently, and it is what makes the next occurrence visible.

Auto-merge is deliberately **not** armed. This changes when a seller gets paused out of payouts, which is money-adjacent and not something I should merge on my own judgement — it wants your eyes on the scope predicate in particular.

---

**AI disclosure:** Generated with `claude-opus-5`. The implementation and specs were written by an earlier Gumclaw session working antiwork/gumroad-private#1523, which pushed the branch but ended before opening a PR; a second session found the orphaned branch and opened it. A third session (this one) re-ran both suites and the mutation proof at the current head, corrected the example counts in the Test Results section above — they had gone stale when `5d4a635b` widened the change from rate limits to the whole transient-failure class — rendered and committed the test-results evidence, and corrected a wrong `Co-authored-by` trailer on the evidence commit (this issue was self-filed from payout-batch monitoring, so no human prompter applies).


gumclaw:

gumclaw:
Ran the fable-5 adversarial review that the body flagged as outstanding, against `36f0f11`. Verdict was **CHANGES-REQUIRED, test-side only — the production code checks out**. Fixes are in `11e7d8f`; the full findings table and dispositions are now in the PR body.

The finding worth your attention is that the `perform_payment` test was pinning nothing. Its `expects(:reverse_internal_transfer!)` looked like it proved the reversal survives a 429, but the fixture uses `add_gumroad_held_balances_negative`, which sums to −$10 — so `prepare_payment_and_set_amount` never creates an internal transfer, and `reverse_internal_transfer!` early-returns on a nil `stripe_internal_transfer_id`. The mock only asserted that `ensure` invokes a method that would do nothing, which the pre-fix generic `StripeError` rescue did identically. I dropped it and retitled the test to what it actually pins. The same test was also replaying a neighbouring test's cassette, so it now has its own.

I declined one P3. The review suggested dropping `ErrorNotifier.notify(e)` to match the `InvalidRequestError` rescue thirteen lines up, which suppresses Sentry for known reasons — but it also confirmed this is not a new flood, since the generic `StripeError` rescue already notified. A rate-limited batch produces exactly the Sentry volume it did before, and a 429 storm is something worth seeing. Suppressing it would be a behaviour change beyond what this PR is for.

Separately, the body's claim about the red CI was wrong and is corrected. It attributed the failures to the Mexico 3DS breakage in #6562; the actual cause was the shared Stripe test account throttling us — ten Slow shards failing at the same checkout alert assertion with `[Stripe] rate limited` in the logs, alongside four unrelated PRs that pushed in the same window. Rerunning the failed shards cleared all of them with no change to the branch.

One thing I did not take on trust: the `IS NULL` arm of the new scope. Dropping it takes the pause-check group from 11 examples/0 failures to 11/4, and three of those four are pre-existing examples — so without that arm this change would have quietly switched the consecutive-failure pause off for every seller, not just rate-limited ones. The exact output is in the body.

Auto-merge is not armed. This is payouts, and the pause exclusion is the kind of change that should have a human's eyes on it before it ships.

gumclaw:
CI is fully green on `5d4a635` — 76 checks passed, zero failures, `mergeStateStatus: CLEAN`. Nothing on this PR is waiting on anything but your review, @rmarescu.

Recording it here because the verdict was actually reached at **15:36 ET and then lost**: the watcher I set for this PR was configured `deliver: local`, which writes to a file nobody reads, and it self-disarmed on the same run with `superseded: head moved to 5d4a635b`. So the green result existed for three hours with no human able to see it, and no watcher left to re-report it. I have retargeted that watcher to the team channel and cleared the state file — this specific failure mode won't swallow another verdict on this PR.

No action needed from you beyond the review itself.

gumclaw:
Reassigned to @nyomanjyotisa at Razvan's request on the tracking issue (antiwork/gumroad-private#1523) — he's the most regular human committer to these payout paths. Nothing about the change itself moved.

@nyomanjyotisa: review and merge is yours. No auto-merge armed since it's a payouts change. Two caveats, both already in the issue thread: the adversarial review hasn't been run on this branch, and CI is red on three `taxes_spec.rb` 3DS examples that are failing fleet-wide including on `main` (#6562, fix in #6565) — nothing in this diff touches checkout.

PR #6573, opened 2026-07-29
Bust the profile cache when a product's price changes

gumclaw:
## What

Bumps a product's `updated_at` when a row feeding its **displayed price** changes, so custom-HTML and default profile storefronts stop serving a stale price.

- `Price`, `VariantPrice`, `BaseVariant` and `VariantCategory` now touch the product they price.
- The editor's deletion sweep (`batch_delete_variants`) touches the product explicitly, because it stamps deletions with `update_all` and so skips callbacks.

## Why

`Pages::ProfileData.cache_key` keys the cached storefront payload on `seller.products.cache_key_with_version`, i.e. `MAX(links.updated_at)`. But the payload's `price` resolves through `price_formatted_verbose` → `display_price_cents` → the associated **`prices` / variant rows**, not the `links.price_cents` column. A price write that never touches `links` therefore leaves the key unmoved and the storefront serving the old number.

Confirmed live from a support ticket: the product page showed the new $29 while the storefront card kept advertising $39. Reading `cached` vs `fresh` off the same method in the same process is the whole bug — the data was already correct, the cache key had simply never moved. Cache store is memcached with no `expires_in` and a per-deploy namespace, so to the seller it reads as permanent, and every browser-side remedy fails because the response is `cache-control: max-age=0, private` (no CDN or browser cache is involved at all).

This is the same class as gp#1398, which was fixed by adding `touch: true` to `Thumbnail`. `Price` had the identical hole and was missed by that fix, which is why this PR also ships a guard test rather than patching one model.

| model | touches the product? |
|---|---|
| `AssetPreview` (cover) | yes, before this PR |
| `Thumbnail` | yes, added by gp#1398 |
| `Price` / `VariantPrice` / `BaseVariant` / `VariantCategory` | **no** → fixed here |

A wrong **price** on a storefront card is money-facing: a buyer clicks through to a different price than the card advertised.

## Before / After

The rendered markup is unchanged — only the *value* inside the cached payload moves — so here is that value, rendered through the real storefront renderer (`Pages::DefaultProfileDocument` → `Pages::ProfileData`), on a product whose "Hardcover" edition the seller reprices from +$5 to +$9 (displayed price $15 → $19).

| before this PR | after this PR |
|---|---|
| storefront still advertising the old price | storefront matching the product page |
| Card advertises **$15**; the product page says $19. Stays that way indefinitely. | Card advertises **$19**, matching the product page. |

Both frames come from the same script: warm the cache at $15, apply the seller's repricing, re-read. The "before" frame applies the write the way the code behaved before this branch (a variant write that never stamps `links`); the "after" frame applies the same seller action with the callback in place.

The rest of the surface, from the specs:

| write | storefront card before | after |
|---|---|---|
| simple product price $39 → $29 | **$39** (indefinitely) | $29 on next refresh |
| tiered membership tier price | stale | rebuilt |
| variant price difference | stale | rebuilt |
| tier grouping soft-deleted | stale | rebuilt |
| editor deletes an option | stale unless an unrelated survivor row happened to be written in the same save | rebuilt |

That last row is the subtle one and is why the deletion-sweep fix is separate from the `touch: true` additions — see below.

## Test Results

spec run and mutation proof

19 examples, 0 failures, rubocop clean. The image carries the green run **and** the mutation proof in one transcript.

**Both guards are proven load-bearing.** Dropping the deletion-sweep `product.touch` and `Price`'s `touch: true`, then re-running the same file, reddens three examples — including the two new ones. Working tree restored before committing (verified by grep, not by assumption).

Two things this caught that are worth stating, because both were live defects in the first draft of this branch:

1. **The catch-all guard spec's `Price` arm was vacuous.** It created a product *inside* the assertion loop, and `cache_key_with_version` embeds the relation's row count — so the key moved by creation alone and the arm passed with the touch reverted. Now the product is created up front.
2. **The deletion-sweep spec I first wrote was also vacuous, for a more interesting reason.** Driving the full `VariantCategoryUpdaterService` moved the key even with the fix removed. Instrumenting `Link#touch` showed the touch coming from `create_or_update_variant!` writing a *survivor* row, not from the deletion — incidental masking that disappears when the seller deletes the last-positioned option and changes nothing else. The spec now drives `batch_delete_variants` directly, so the guarantee doesn't rest on that coincidence.

Pre-existing local failures, unrelated: `spec/models/price_spec.rb`, `variant_category_spec.rb` and `variants_updater_service_spec.rb` show 5 `create(:purchase)` failures. A detached `origin/main` worktree on the same machine produces the **identical 5**, so they are shared-test-DB environment, not this diff.

## What this does NOT cover

- The `posts` cap and the `MAX_ITEMS = 100` product truncation are a different defect (gp#1522) and deliberately untouched here.
- Hard `destroy` of a `BaseVariant`/`VariantPrice` still doesn't touch, because `previous_changes` is blank for a freshly loaded record. Theoretical today: every runtime deletion path is `mark_deleted!` (a real save) except the `update_all` fixed here; there are no `.destroy`/`delete_all` calls on these tables outside specs.
- Write amplification on membership edits is bounded and unmeasured but analysed: inventory counters use callback-skipping `update_all`/raw SQL, an unchanged-price save is a no-op, and the per-sale `variant.touch` for limited-quantity products is filtered by the concern's guard (`previous_changes.keys == ["updated_at"]`). **Zero added writes per purchase or renewal.**

## QA steps

A preview clickthrough cannot show this — the defect needs a cache entry that has already gone stale, which a fresh preview has never populated. The storefront frames above are the rendered evidence; the reviewer path is the diff plus the mutation proof:

1. Read `app/models/concerns/touches_product_for_price_cache.rb` — note the loop-safety comment now states the real mechanism (`touch` fires only `after_touch`/`after_commit`; `Link` has neither statically registered), rather than the previous claim about an `after_update` guard that `touch` never runs.
2. Read the `product.touch if newly_deleted.any?` line next to the existing `product.invalidate_cache` in `batch_delete_variants`, and note that `invalidate_cache` only clears a fragment prefix — it never writes `links`.
3. `bundle exec rspec spec/services/pages/profile_data_spec.rb` → 19 examples, 0 failures.
4. Optional: remove either guard and re-run to see the examples redden.

Ref antiwork/gumroad-private#1518

---

AI assistance: written by `claude-opus-5` (`model.default` in the local Hermes config) running as the autonomous dev dispatcher. Instructions given: sweep `awaiting-human` items in both repos, then build qualifying issues autonomously through PR; for this branch specifically, adopt an orphaned worktree left by a prior session, act on its fable-5 adversarial review (which returned DO-NOT-SHIP on a P1), merge `main` to pick up the 3DS CI fix, and verify every new spec is load-bearing before opening. Later pass, on review feedback: capture the storefront before/after through the real renderer, drop a misnamed screenshot, and remove a seller handle from this body.


gumclaw:
Routing this one: **@nyomanjyotisa** as the human owner. You reviewed #6364 ("Bust the profile cache when a product thumbnail changes"), which is the direct sibling of this fix — same `Pages::ProfileData.cache_key` hole, same `touch: true` shape, just on `Price`/`VariantPrice`/`BaseVariant`/`VariantCategory` instead of `Thumbnail`. Labelled `awaiting-human`.

State as of this sweep: **78 checks SUCCESS, 8 skipped, zero failures**, MERGEABLE at `978e3af65`.

Two things worth your attention specifically, since you saw the thumbnail version:

- The **deletion-sweep fix is separate** from the `touch: true` additions, and deliberately so. `batch_delete_variants` stamps deletions with `update_all`, which skips callbacks entirely — so the association `touch` can't reach it and it needs the explicit `product.touch`.
- The body documents that **both new spec arms were vacuous in the first draft** and were rewritten. One passed with the touch reverted because `cache_key_with_version` embeds the relation's row count, so creating a product inside the assertion loop moved the key by itself. The other was masked by `create_or_update_variant!` writing a survivor row. The specs now drive `batch_delete_variants` directly and the mutation proof is in the Test Results image.

No preview clickthrough on this one and the body says why: the defect needs a stale memcached entry, which a fresh preview has never populated. The reviewer path is the diff plus the mutation proof.

Not arming auto-merge — the merge call is yours.

nyomanjyotisa (commented):
Code looks correct. I verified the cache touches add no purchase or renewal writes, cannot recurse, and the deletion-sweep test is load-bearing.

Before merge, please add the required storefront evidence, remove the seller handle from the public body, and delete `qa-media/pr-PRNUM-price-touch-spec-results.png`.

gumclaw:
Thanks for verifying the write-amplification and recursion questions — all three of your pre-merge items are done at `e2f1bba`.

**Storefront evidence.** Added a rendered before/after to the body, captured through the real renderer (`Pages::DefaultProfileDocument` → `Pages::ProfileData`) rather than a mockup. The scenario is a seller repricing a "Hardcover" edition from +$5 to +$9, so the displayed price moves $15 → $19:

| before | after |
|---|---|
| card shows **$15** while the product page says $19 | card shows **$19** |

The "before" frame applies the write the way the code behaved before this branch (a variant write that never stamps `links`), the "after" frame applies the same seller action with the callback in place. Both frames come from one script — warm the cache at $15, apply the repricing, re-read. Worth noting the first version of that script was misleading: reading the payload without reloading the seller reused the memoized association cache version, so the "after" frame also showed $15 and would have argued the fix doesn't work. Fixed by reloading per read, same as the specs do.

I also replaced the earlier "no pixel diff to show" line in Before/After — that was true of the markup but it was hiding the thing you actually need to see, which is the value inside the payload.

**Seller handle.** Removed from the body. The bug description now stands on the mechanism alone.

**`qa-media/pr-PRNUM-price-touch-spec-results.png`.** Deleted — it was an unrenamed duplicate of `pr-6573-test-results.png` from an earlier draft.

Only PNG files and the body changed, so no code moved since the green run at `978e3af`. Leaving it assigned to you for the merge.

nyomanjyotisa (approved):

PR #6579, opened 2026-07-29
Fix Store Agent proposal replies

gianfrancopiana:
Ref antiwork/gumroad-private#1470

## What

The Store Agent now marks each completed reply as either an answer or a proposal. The server checks that mark against the proposal actually created during the turn.

When a proposal exists, the server supplies the confirmation sentence and waits until the proposal is saved before showing it. Missing or contradictory marks get one retry, then a reply that matches the server's state. The existing wording check remains as an extra safeguard.

## Why

The agent could say a change was ready to confirm even though it had not created a proposal. Adding more phrases to a wording check only covered phrases we had already seen. This verifies the reply against the proposal state instead.

## Validation

Test confidence reached 99% with all 16 selected test files passing. The proposal flow also passed in a headless browser, RuboCop passed, and autoreview found no actionable issues.

## Media

A saved Store Agent proposal with server-owned confirmation copy

---

This PR was implemented with AI assistance using gpt-5.6-sol.


gumclaw:
Adversarial pre-merge review (fable-5), pinned to head `9e2682626d2a0f2f5863b2ebd89354ac9aabf048`.

## Verdict: READY-TO-MERGE — no P1/P2 findings

The typed `complete_turn` contract, the server-owned confirmation sentence, and the persist-before-token ordering are correctly implemented on both reply paths, and the design clears all four traps that bit the previous prose-only guard ([#6506](https://github.com/antiwork/gumroad/pull/6506)):

- **No truthful reply can be swallowed silently**: rejected text is replayed to the model with a correction naming the exact fix, and the give-up line ("I couldn't finish that response") is honest rather than actively false.
- **No canned string trips the backstop**: every canned/correction string in the class (plus plausible compliant correction replies) was run through the real `STAGED_CLAIM_PATTERNS` in Ruby — all clean, so the retry branch can't silence a compliant model.
- **Rejected mixed-tool turns provably never reach `run_tool`** — the `:invalid` arm corrects and `next`s before `apply_tool_uses`.
- **The retry budget genuinely covers recovery**: after a `proposal_ready_without_proposal` rejection, api_write is a `:tools` turn that doesn't consume the retry, and `TURN_CONTRACT_RECOVERY_ITERATIONS = 2` reserves the api_write→complete_turn sequence (pinned by spec).
- **Removing the controllers' reset/token re-emission on persistence failure is safe**: `replace_unpersisted_proposal_reply!` mutates the same hash `finish_stream` emits from, suppression is invariantly on for proposal turns, and the reply-only already-streamed case is unreachable (`return false if result[:proposed_action].blank?`).
- **The phantom-claim metric survives distinctly**: the prose check runs before `missing_complete_turn` is assigned, so the original gumroad-private#1470 cohort stays its own Sentry issue; both retry and give-up branches report under a fixed message.
- **Prompt injection can neither fabricate a card nor suppress a real proposal**: `proposal_ready` with no server proposal → correction → honest line; `reply_only` after api_write → `proposal_not_reported` → the proposal ships anyway with server copy. The model never authors the confirmation sentence.
- Both #6506 spec tables (truthful replies that must not be swallowed / production phantom shapes) are retained and updated, not weakened.

## Findings — all P3, non-blocking

1. **Proposal-turn prose is destroyed while the prompt teaches the opposite.** `final_turn_decision` unconditionally returns `PROPOSAL_READY_REPLY` when a proposal exists, but the system prompt still tells the model to write creator-facing text on those turns — so "what's my best seller, and put 20% off it" loses the answer half, and the model is instructed to author text the server always discards.
2. **The mobile buffered endpoint is the only `record_agent_assistant_message!` caller without a persistence rescue.** The new consistency `ArgumentError` (and any other persistence failure) becomes a generic 500 that discards a fully generated reply, where web buffered/streaming and mobile streaming all degrade to the honest reply. Unreachable from the current service — `turn_result` derives `outcome` from the same field the check reads — but a future service regression is exactly when the graceful path matters. (Pre-existing sibling defects in the same block: `conversation.external_id` NoMethodErrors when creation fails, and `proposed_action` is returned without the persisted-message guard web applies.)
3. **PR body media links point at `raw.githubusercontent.com/.../gianfranco/store-agent-staging-integrity/...`** — they 404 once the branch is deleted post-merge. Relink to a commit SHA.

Watch item for after merge (not a code change): during an Anthropic outage the OpenRouter GPT fallback serves turns; if that model is less reliable at ending with `complete_turn`, turns degrade to retry + the honest failure line — visible as a `missing_complete_turn` spike in Sentry. Honest degradation, but harder than today's, so watch that rate in any fallback window.

Fixes for findings 1 and 2 are being prepared against this branch; finding 3 is a body edit after merge.

gumclaw:
Findings 1 and 2 from the review above are now fixed on this branch: [`eac4154`](https://github.com/antiwork/gumroad/commit/eac415485) reconciles the system prompt and `complete_turn` schema with the server-owned proposal reply (informational answers go before `api_write`), and [`a609ad2`](https://github.com/antiwork/gumroad/commit/a609ad2d0) gives the mobile buffered endpoint the same persistence rescue and payload gating as web — a persistence failure now returns the honest reply instead of a 500, with `proposed_action`/`proposal_message_id` gated on the persisted message and a nil-safe `conversation_id`. All new specs proven red against the unfixed code, then green (116 examples, 0 failures across both touched spec files); RuboCop clean. One pre-existing mobile spec asserted the old 500 behavior and was updated to assert the graceful path. Finding 3 (branch-pinned media URL) is a body edit for after merge.

PR #48, opened 2025-04-05
Basic models for ProductReviewVideo & VideoFile

xrav3nz:
This PR adds the basic boilerplate models for `ProductReviewVideo` & `VideoFile`. 

This has not added transcoding capabilities to `VideoFile` yet.

PR #47, opened 2025-04-05
Enable Rack::MiniProfiler in production

xrav3nz:
Having `Rack::MiniProfiler` in production makes identifying and addressing performance-related issues much much easier.
 
Most of our routes are rendered by Rails with some going through `react-router`. This minimal setup should work for all of the Rails-rendered ones and we can address the `react-router` ones later when needed.

PR #6585, opened 2026-07-29
Narrow the deferred intent to the payment methods the Payment Element mounted with

gumclaw:
> [!IMPORTANT]
> **Do not merge. The direction is now settled from our own production records, and it is the
> reverse of this PR's premise — so narrowing cannot fix any of the four orders cited below.**
>
> The previous banner asked for exactly one thing: settle the direction against our records rather
> than a manufactured preview pair. Done, read-only against the production replica. The answer does
> not need Stripe's dashboard, because one of the four failures is self-proving.
>
> **Order 112356301 (purchase 348567663) decides it.** Its persisted `ip_country` is **Greece**, so
> `Order::PreparePaymentIntentService#buyer_country_alpha2` resolved `"GR"`, and on that basis the
> resolver cannot put either method on the intent: `klarna_methods` returns `[]` unless
> `buyer_country == KLARNA_SUPPORTED_BUYER_COUNTRY` (`"US"`), and `cashapp` is removed by
> `launched -= US_LOCKED_PAYMENT_METHOD_TYPES unless buyer_country == US_ALPHA2`. The intent for
> that order could only have been `["card","link"]`. Stripe's error for it names
> `["card","link","cashapp","klarna"]` — so **that list is the Element's, and the Element was the
> superset.** Same direction as the five preview legs, now confirmed on a real failed charge.
>
> **Every one of the four is Element-superset / intent-subset**, measured rather than inferred:
>
> | Order | Buyer geo (persisted) | Product | Element (from the error) | Intent could only be | Narrowing to the mounted list |
> |---|---|---|---|---|---|
> | 112356301 | Greece | fixed $9.99 | `card, link, cashapp, klarna` | `card, link` | no-op — intent already smaller |
> | 112292084 | US | PWYW, paid $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |
> | 112309624 | US | fixed $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |
> | 112378943 | US | PWYW, paid $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |
>
> Intersecting the intent against the mounted list is a no-op in all four, because the intent is
> already the subset. **This PR would not have saved any of the purchases it was written for.**
>
> **The tracking issue's stated mechanism also fails on these four.** Both Klarna bases agree in
> every row — the amount-edit story requires them to straddle `KLARNA_MIN_USD_CHARGE_CENTS`:
>
> | Purchase | Intent basis (`displayed_price_cents`) | Element basis if fixed-price | Klarna allowed on each | Agree? |
> |---|---:|---:|---|---|
> | 348481085 | 99 | 0 (PWYW) | no / no | yes |
> | 348504633 | 99 | 99 | no / no | yes |
> | 348567663 | 999 | 999 | yes / yes | yes |
> | 348598779 | 99 | 0 (PWYW) | no / no | yes |
>
> Two of the four are **fixed-price** products, where the buyer cannot edit the amount at all, so no
> amount-edit mechanism can reach them.
>
> **What the Greece order points at instead: the two sides read the buyer's country from two
> different requests.** The presenter resolves it live from the checkout page's own
> `request.remote_ip` (`stripe_payment_presenter.rb:327`); prepare resolves it from the
> `ip_country` string persisted when the order was created (`prepare_payment_intent_service.rb:979`).
> Both are GeoIP, and the code comments on each assert they are the same basis to preserve this very
> invariant — but they are sampled at different moments, so a buyer whose apparent location changes
> between page load and pay (VPN toggled, mobile network hop, IPv6→IPv4) gets a US Element and a
> non-US intent. That produces exactly this order: Element with the US-locked pair, intent without.
> I sampled 400 recent purchases and found 0 disagreements between a live lookup and the persisted
> value, so the *stored* value is stable; the divergence is across requests, not in GeoIP itself.
>
> **Correction to an earlier claim in this banner.** It previously said `Checkout::UpsellsController`
> builds the presenter with `ip: nil` and thereby "silently withholds Klarna and the US-locked
> methods on that path". I re-checked that against `origin/main` and it is wrong — please disregard
> it. That controller's `#cart_item` calls `CheckoutPresenter#checkout_product`, which never
> constructs `Checkout::StripePaymentPresenter`; the only construction site is `#checkout_props`
> (`checkout_presenter.rb:51`). So no payment-method resolution happens on that path and there is no
> Klarna to withhold. I should have traced the call before writing it down.
>
> What `ip: nil` does reach there is the two *display* readers: `product.ppp_details(@ip)` returns
> nil, and `buyer_currency_display_props(ip: nil)` falls through
> `buyer_currency_for_country(nil) → nil` to `display_mode: "default"`. `Checkout::FormPresenter:25`
> passes `ip: nil` the same way. My read is that both are seller-facing dashboard previews rather
> than buyer checkout, so a seller previewing their own upsell or checkout form sees the set price
> with no local-currency estimate — mildly inconsistent with what a buyer sees, and worth a separate
> look, but not a buyer-facing defect and not related to this PR.
>
> **Settled: all four orders are ABANDONMENT, not failed confirms.** Measured read-only on the
> production replica. Each has `stripe_transaction_id = nil`, `error_code = nil`,
> `stripe_error_code = nil`, and an `updated_at` exactly ~15 minutes after `created_at` (15m20s,
> 15m06s, 15m04s, 15m12s). `ChargeProcessor::TIME_TO_COMPLETE_SCA` is `15.minutes` and
> `FailAbandonedPurchaseWorker` is enqueued with that delay from
> `Order::PreparePaymentIntentService:993`, so these rows record `FailAbandonedPurchaseWorker`
> closing an `in_progress` purchase — a buyer who left the page. **None of them reached Stripe at
> all**, so none evidences a `payment_method_types` mismatch, and 348504633 was never a special
> case. Two further inputs falsified while checking: all four sellers are platform-account sellers
> (no connected account, no `stripe_capabilities_snapshot`), so the cached-capability branch never
> runs; and the Klarna flag is `percentage_of_actors = 100` with zero explicit actors, so it cannot
> flip per-request.
>
> **The mismatch error string is not recorded anywhere in production** — a
> `LIKE '%does not match the expected%'` scan over 10 days of purchases returns **0 rows**, and the
> error-code distribution on failed confirmation-token purchases has no bucket it lands in. So this
> failure class is currently unmeasurable server-side, which is what the Rollout section's success
> criterion depends on.
>
> The paragraph below is retained as a record of the earlier reading.
>
> ~~**Order 348504633 is still unexplained**~~ — US on both sides, fixed price, both bases 99, both
> answering "no Klarna". Neither the amount-edit nor the country-divergence mechanism reaches it.
> Saying so rather than stretching a mechanism to cover it.
>
> **Recommendation.** Close or repurpose this branch rather than merging it. The parameter it adds is
> sound plumbing and the specs are real, but the operation it performs is the one Stripe rejects, and
> the reviewer's leg 1 already showed the un-narrowed intent completing. If the invariant is worth
> defending in code, the shape that matches the measurements is making the two sides agree on the
> **country** basis, not intersecting the method list. That is a different fix and wants @rmarescu's
> call before anyone writes it.
>
> Every section below predates this and is retained only as a record; treat its subset/superset
> claims and its rollout condition as contradicted.

## What

On the client-confirm checkout lane, the browser now reports the `payment_method_types` list its Payment Element was mounted with, and `Order::PreparePaymentIntentService` intersects the deferred PaymentIntent's list against it — narrowing only. The intent can no longer list a method the page never offered.

The Element's method list is fixed when Stripe's Elements instance is created; this service re-resolves it from server-owned purchase rows at pay time. Stripe rejects the ConfirmationToken on any difference between the two, but only one direction is recoverable: a method the intent lists that the Element never mounted can be dropped, whereas a method the buyer confirmed with cannot be added to an element that has already rendered. So a server rule that *widens* between page load and payment fails the buyer at confirm with no retry available on the same page.

Reported values are trusted only to narrow, never to widen, so a stale or crafted list cannot enable a method past its rollout gate. Two fail-safe exits keep the narrowing from causing the failure it prevents: if the report is disjoint from the resolved set, or excludes the method being confirmed, it cannot describe the element the token was minted on, so it is ignored (and logged) in favour of the legacy narrowing below.

Older checkout pages report nothing. Klarna is the only launched method whose gate moves with the cart total, so on that path it is kept only when the buyer chose it and it is still allowed — never added to a card or Link payment.

## Why

A production checkout failed at `stripe.confirmPayment`: the Element mounted `["card","link","cashapp"]` while the intent carried `["card","link","cashapp","klarna"]`.

```
The provided payment_method_types (["card", "link", "cashapp"]) does not match the expected
payment_method_types (["card", "link", "cashapp", "klarna"]).
```

The buyer could not pay and could not retry — the Element cannot add a method it did not mount with, so a second attempt on the same page fails identically.

⚠️ **The "roughly 1 in 470 client-confirm error reports, in bursts" figure that stood here is withdrawn as unverified.** It came from client-side error reporting, not from anything queryable, and a `LIKE '%does not match the expected%'` scan over 10 days of production purchases returns **0 rows** — this failure class is not recorded server-side at all, so its real rate is currently unknown rather than measured. See the banner at the top.

Same family as the currency divergence fixed in #6497 (browser and server disagreeing about a parameter Stripe requires to match exactly), but about the method list, so that fix neither causes nor prevents it.

Approach chosen by @rmarescu on the tracking issue, including the condition that `prepare` must know which methods were available when checkout loaded rather than inferring it.

Tracking issue: gumroad-private#1528

**Related but not this fix:** #6486 reloads the checkout payment configuration after cart-item changes. That helps when a buyer edits a pay-what-you-want price, but it does not cover rules changing after page load, and per @rmarescu it should not be treated as the fix for this. It does make one detail load-bearing here: the reported list is captured at Elements *creation*, not read from the current configuration — Stripe fixes `paymentMethodTypes` at creation and offers no update path, so a refreshed configuration never reaches the live element. Reading the current config would have made the report drift along with the refresh and defeated the narrowing entirely.

## Before / after

Not a visual change — no rendered surface is added, removed, or restyled. The Payment Element renders exactly as before; what changes is which methods the deferred PaymentIntent is created with, which is only observable in the Stripe API call and in whether the confirm succeeds.

Before: element mounted without Klarna, intent created with it, `confirmPayment` rejected, purchase lost.
After: intent narrowed to the mounted list, card confirm succeeds.

**The behavioural before/after transcript that stood here has been withdrawn, not re-shot.** It
recorded the nine narrowing examples green and then reddening with the narrowing neutered, which
reads as proof that narrowing is the fix. Two things make it unusable: it predates @rmarescu's
`39e14f75c`, so it shows the old no-report expectation that commit changed; and every example in it
stubs `StripeDeferredPaymentIntent.create`, so it can only observe which list *reaches* Stripe and
never Stripe's verdict on it — which is the disputed question. The live legs linked in the banner
say Stripe accepts the superset and rejects the subset.

It needs re-recording against whichever direction the fix ends up taking, so it is deliberately
absent rather than replaced.

## Test results

New coverage, all passing locally:

**`spec/services/order/prepare_payment_intent_service_spec.rb`** — 10 examples in a new `narrowing the intent to the methods the Payment Element mounted with` context: an unmounted method is dropped so a card confirm matches; a non-Klarna unmounted method is dropped (@rmarescu's `f14b94d93`, so deleting the intersection reddens the specs); a reported method the rules disallow is never added; both fail-safe exits; the report cap (`6c58b38b9`); a malformed report degrades to the no-report path; both older-page cases; and the pay-what-you-want case where the amount rises past Klarna's floor after mount.

**`spec/controllers/orders_controller_spec.rb`** — the list reaches the prepare service; `#create` ignores it.
**`spec/services/order/create_service_spec.rb`** — purchases still build when the list is present (it is not a `Purchase` attribute).

```
281 examples, 5 failures
```

⚠️ **That count is not reproducible and should not be relied on.** A seed-pinned baseline of the
three files, run in clean worktrees with the harness-trap greps (`to read failed line`,
`EADDRINUSE`, `NoDatabaseError`, `getcwd`) clean on both runs:

| Tree | Examples | Failures |
|---|---|---|
| branch head `fb755e691` | 283 | 12 |
| unmodified `main` `85fffe359`, same seed | 269 | 12 |

The counts match, but comparing by example **name** rather than line number, only **5 of the 12 are
the same examples**. Seven fail only at head (five Pix, two iDEAL); seven only on `main` (six
`orders_controller`, one iDEAL). So this is not a stable pre-existing set — the examples leak state
into each other under a shared test database and the membership reshuffles run to run.

Across five invocations the answers were 5 (this body), 8 (@rmarescu at the unmodified head), 4, and
12/12 above. Any single number would be false precision. **Local absolute counts are not usable
evidence on these three files; CI on real shards is the authority.** The order-dependence is worth
fixing on its own and should not ride along with this PR.

**Frontend** — `182 passed` across `PaymentElementInput.test.tsx`, `purchase.test.ts`, `order.test.ts`, `payment.test.ts`, `heldWalletPayment.test.ts`, `index.test.tsx`, including new tests pinning that the reported list stays frozen across a configuration change and re-reports only when a currency or mode change genuinely remounts.

**Load-bearing check:** the plumbing hunks are individually verified — reverting the controller merge or the `CreateService` exclusion reddens one spec each. The narrowing's own mutation proof is withdrawn along with its transcript (see Before / after): it was measured against stubbed Stripe calls, so it proves the narrowing happens, not that narrowing is the right direction.

**CI:** 77 successful, 9 skipped, 0 failing at the merged head `6454bc3ee`. That merge pulled `299ddc514` from `main`, which pins the clock in the `applePayRecurringPaymentRequest.test.ts` examples that were red at `eac68f06c` — a file this branch never touches.

`bundle exec rubocop` clean on the 5 touched Ruby files; `prettier --check` clean; `tsc --noEmit` clean for the touched modules. The 7 eslint errors in `PaymentForm.tsx` are pre-existing and identical on `origin/main`.

## Adversarial review

A local adversarial review ran on the branch before this PR was opened. Verdict: CHANGES-REQUIRED, **no P1s**; findings fixed in `46682dd`.

| Finding | Disposition |
|---|---|
| Controller merge and `CreateService` exclusion had zero coverage — either hunk could be reverted with the suite green (fix inert in production, or `UnknownAttributeError` at every reporting checkout) | Fixed: two specs added, each verified to redden on revert |
| Fail-safe exits returned the raw resolved list, so a malformed report bought *weaker* protection than sending nothing | Fixed: both exits fall back to the legacy narrowing |
| Creation capture re-taken on currency change only, but mode is also part of the remount key — a mode flip would report a stale list (reachable once #6486's config reloads land) | Fixed: capture re-taken on mode too; every creation option and the remount key read one committed capture |
| Legacy-path comment implied that path is now drift-free | Fixed: comment owns the residual (capability-snapshot and flag drift can still widen the other methods there) |
| Comments re-told the incident in seven places | Fixed: one authoritative explanation, the rest are pointers |

Confirmed safe by the review: narrow-only holds on every return path (all are subsets of the post-strip list, so no reported value reaches the output and the fail-safes cannot re-admit anything an earlier strip removed); widening past a rollout gate via a crafted report is impossible; wallets (which preview as `card`) and Link cannot legitimately hit the fail-safes; held wallet payments replay the mint-time list, which is correct; the forced-currency lane is unaffected; the parameter is absent from `permitted_order_params` and malformed input degrades rather than raising.

### Merged-head composition probe at `537e724df160` (2026-07-31)

The gate re-armed on a **second** `origin/main` merge (`537e724df`). Unlike the previous one, this
merge is not inert for this PR: `git diff b1fe265e5 537e724df` over this PR's own files is empty for
`app/services/order/prepare_payment_intent_service.rb`, `app/services/order/create_service.rb` and
`PaymentElementInput.tsx`, but **non-empty for `app/controllers/orders_controller.rb`,
`PaymentForm.tsx` and `app/javascript/data/purchase.ts`** — main's #6589 ("Suppress confirm-error
reports using the method the buyer actually selected") edited three of the four files this branch
touches, including the `confirm_error_reportable?` predicate this branch's diff sits next to.

So the question is composition, not pixels: did both sides survive, and does #6589 change anything
this PR's body claims? Probed read-only on the preview pod serving exactly `537e724df160`:

*Every cell is `rails runner` output on the pod. Panel 1: both sides of the merge are present in the
same request path (this PR's `narrow_to_mounted_payment_method_types` and
`reported_mounted_payment_method_types`, main #6589's widened constant, and main #6644's sanctions
screen in the checkout lane, whose gate answers `false` for a US buyer and is therefore inert for
every leg in this PR). Panel 2: six confirm-error shapes driven through the live predicate, with the
pre-#6589 condition computed over the same inputs.*

**What this changes about this PR's claims.** Two, both narrow:

1. **The PR's own break-detection signal is now weaker.** The Honest-limits and QA-steps sections say
   a silent break of the `PaymentForm` ref wiring "would only show as reports having stopped
   arriving". After #6589, reports also stop arriving for a plain card decline (Stripe attaches no
   `payment_method_type` to most declines) and for a wallet-row confirm failure — the two highlighted
   rows. "Reports stopped arriving" is therefore no longer a clean signal for that break, and the one
   open QA item on this PR *is* the wallet lane.
2. **The failure class this PR is about still reports.** The `payment_method_types` mismatch error is
   an `invalid_request_error`, not a `card_error`, so it takes `confirm_error_reportable?`'s early
   return (row 6, `true` both before and after). The Rollout section's measurability gap is neither
   closed nor widened by this merge.

Nothing here revives the subset/superset premise. The banner at the top of this PR still stands: the
direction is settled from production records against this PR's premise, and this panel is a
merge-integrity check, not evidence for merging.

*Read-only probe — no writes, no seeding, no durable preview state.*

### Merged-head re-verification at `6454bc3ee` (2026-07-31)

The gate re-armed on the `origin/main` merge. This PR's own code is **byte-identical** across it —
`git diff 741ed79cf 6454bc3ee` over `orders_controller.rb`, `app/services/order/`,
`PaymentElementInput.tsx`, `PaymentForm.tsx`, `purchase.ts` and the three spec files is empty — so the
narrowing itself did not move, and nothing above this section is restated by these frames.

What the merge *did* bring into the same request path is #6644's sanctions screening, now a
`before_create` on `Purchase` and reached from `Subscription#build_purchase`. That is the checkout lane
every leg in this PR drives, so the question worth answering at this head is composition, not pixels:
does a card checkout still reach a receipt with both changes in the tree?

Probed on the deployed preview (`x-revision: 6454bc3eea8b`, asserted before every leg):

| | |
|---|---|
| `narrow_to_mounted_payment_method_types` present | yes |
| `reported_mounted_payment_method_types` present | yes |
| `Purchase#validate_sanctioned_location` present | yes (new, from the merge) |
| `blocked_location?` for the legs' inputs (US / US-CA / US-NY) | `false` — the new gate is inert here |
| Seller flags | `stripe_payment_element_client_confirm` on, `checkout_local_method_klarna` on |

Then two live card purchases, anonymous buyer, `4242…4242`, Playwright at `deviceScaleFactor 2`:

| | Desktop 1440×1000 | Mobile 375×812 |
|---|---|---|
| Reported to `#prepare` | `["card","link","cashapp","klarna"]` | `["card","link","cashapp","klarna"]` |
| Intent `payment_method_types` (read from Stripe) | `["card","link","cashapp","klarna"]` | `["card","link","cashapp","klarna"]` |
| Intent / status | `pi_2Tz7la9e1RjUNIyY0k3gdhQb` / `succeeded` | `pi_2Tz7ls9e1RjUNIyY1LJLcdf0` / `succeeded` |
| Purchase | `#47` `successful`, `ch_2Tz7la9e…` | `#48` `successful`, `ch_2Tz7ls9e…` |
| Outcome | receipt `/d/bf0c68a0…` | receipt `/d/1d15d6d1…` |

Both rows persisted, which is itself the evidence that the merge's new `before_create` gate passed
rather than silently rejecting the purchase.

| Desktop checkout | Desktop receipt |
|---|---|
|  |  |

| Mobile checkout, 375px | Mobile receipt |
|---|---|
|  |  |

**What this does and does not settle.** It settles that the two changes compose: the narrowing still
runs, the sanctions gate still runs, and a US card buyer still gets a receipt at the merged head. It
settles nothing about the direction question in the banner at the top — these are baseline legs with no
manufactured divergence, so they cannot speak to whether Stripe rejects a subset or a superset. That
call is still @rmarescu's, and this PR stays a draft until it is made.

Preview: `https://fix-prepare-honors-mounted-metho.apps.staging.gumroad.org` — served
`x-revision: 5afe37ce01e7` — the main-merge head — asserted equal to the PR head before every leg (the driver exits
non-zero on a mismatch). The evidence commit `c16e15dc6` only adds these PNGs; the code exercised
is `5afe37ce0`, and all seven legs were re-run against it after the merge with `main`. Playwright at `deviceScaleFactor 2`, anonymous buyer, 1440×1000 and 375×812, real
Stripe test mode with card `4242…4242`. The lane is confirmed active on every leg:
`integration: payment_element_client_confirm`, `fallback_reason: null` — this PR's code path, not
a neighbouring one.

**Manufacturing the divergence.** The seeded preview seller resolves the same list
(`["card","link","cashapp","klarna"]`) at render and at pay time, so nothing drifts on its own and
an untouched run cannot exhibit the bug. Each leg below therefore changes exactly one thing and
leaves the server, the card, and the buyer flow alone:

- **Element-side legs** rewrite the *server-rendered page props* so the Payment Element is CREATED
  without `cashapp`, i.e. a page that loaded while that method was ineligible. The server still
  resolves it at pay time. That is the production shape from gumroad-private#1528.
- **Report-side legs** rewrite only `payment_element_mounted_payment_method_types` in the
  `/orders/prepare` body — the single parameter this PR adds — to exercise the fail-safes.

`cashapp` rather than `klarna` is dropped in the decisive pair on purpose: `klarna` alone cannot
discriminate the fix, because `narrow_klarna_for_legacy_client` strips it on the no-report path
too, so both legs would land on the same intent for a card confirm and the pair would show no
difference while appearing to prove one.

### The decisive pair — Element mounted without `cashapp`, server still resolves it

| | Reported (this PR) | Report stripped (an older checkout page) |
|---|---|---|
| Element created with | `["card","link","klarna"]` | `["card","link","klarna"]` |
| Browser reported to `#prepare` | `["card","link","klarna"]` | *(parameter removed)* |
| **Intent's `payment_method_types`** | **`["card","link","klarna"]`** | `["card","link","cashapp"]` |
| `stripe.confirmPayment` | succeeded | **rejected** |
| Outcome | **receipt, `/d/cbf03e6c…`** | **stuck on checkout, no purchase** |
| Intent id / status | `pi_2Tyw2U9e1RjUNIyY1tMJ4doM` / `succeeded` | *(no intent reached `succeeded`)* |

The rejected leg's error, verbatim off the page — the production string from the Why section,
reproduced live on the deployed build:

```
The provided payment_method_types (["card", "link", "cashapp"]) does not match the expected
payment_method_types (["card", "link", "klarna"]). Try confirming with the same parameters in
both the API and Stripe Elements.
```

**Without the report — the confirm is rejected and the buyer cannot pay.** Same page, same card,
same server; the intent simply lists a method the Element never mounted.

| Desktop | Mobile, 375px |
|---|---|
|  |  |

**With the report — the intent is narrowed to the mounted list and the purchase completes.**

| Desktop | Mobile, 375px |
|---|---|
|  |  |

The Element as it mounts for both legs — a Payment Element created without `cashapp`, rendering
normally:

### The safety properties, each exercised against the deployed build

| Leg | Reported to `#prepare` | Intent's `payment_method_types` | Outcome |
|---|---|---|---|
| **Baseline** — untouched, no divergence | `["card","link","cashapp","klarna"]` | `["card","link","cashapp","klarna"]` | receipt `/d/f4a3e0b4…` — the report is inert when nothing drifted |
| **Wider** — report the resolved list **plus** `afterpay_clearpay` | `[…,"afterpay_clearpay"]` | **`["card","link","cashapp","klarna"]`** — the extra method is **not** added | receipt `/d/177ca56e…` |
| **Fail-safe** — report `["boleto","oxxo"]`, disjoint from anything resolvable | `["boleto","oxxo"]` | falls back to the legacy narrowing (`["card","link","cashapp"]`), **not** an empty/unconfirmable intent | confirm rejected — a garbage report buys no more than sending nothing, and no *less* |

The **wider** row is the security property stated in the Why section, demonstrated rather than
asserted: a crafted report cannot enable a method past its rollout gate. `afterpay_clearpay` was
sent and the intent came back without it.

The **fail-safe** row's rejection is the correct behaviour, not a regression: the disjoint report
cannot describe the element the token was minted on, so the service ignores it and degrades to the
no-report path — the same intent an older page would have produced.

| Baseline (no divergence) | Wider report not honored | Disjoint report ignored |
|---|---|---|
|  |  |  |

### What this closes from the "only verifiable against real Stripe" list

The Before/After section above called out two properties as unverifiable without real Stripe. Both
are now settled against real Stripe test mode on the deployed preview, so that caveat is stale in
the good direction:

1. **Stripe accepts a subset intent while rejecting a superset.** The decisive pair is exactly
   that asymmetry: the narrowed intent (a subset of what the Element mounted, containing the
   confirmed `card`) confirmed and produced a receipt, while the un-narrowed superset was rejected
   with the production error string. No stubbing — these are live `POST /v1/payment_intents`
   responses read off the wire.
2. **The wallet lane reports the list in a real browser.** Partially closed: the `PaymentForm` ref
   wiring demonstrably works in a real browser — every leg logged a non-null
   `payment_element_mounted_payment_method_types` on `/orders/prepare` matching what Elements was
   created with, including after a props-driven remount. Not closed for a *wallet* confirm
   specifically: headless Chrome reports no available wallet to Stripe, so Apple Pay / Google Pay
   cannot be exercised from this runner and still needs a human on a real device.

### Honest limits

The divergence is manufactured by rewriting the page's own props and the prepare body in flight —
a harness around the deployed app, not a modification of it. The server, the narrowing code, and
Stripe are all untouched and real. `klarna` deliberately is not the dropped method in the decisive
pair (see above), so the PR's own QA steps 1-3 pay-what-you-want route remains a human check; the
seeded preview product's price is not editable from the checkout page, which is why the amount-
window route was not drivable here. Step 5 (a real Klarna purchase) and the wallet lane are also
unexercised. All legs are US-egress, US buyer, USD, single seller, `checkout_local_method_klarna`
active on the seeded seller.

### Direction legs at head `741ed79cf` — three distinct frames

`741ed79cf` is an evidence-only commit (no `app/` or `spec/` change). An earlier version of this
section said two of its four frames were "linked from [this comment](https://github.com/antiwork/gumroad/pull/6585#issuecomment-5138600941)"
— that was wrong twice over: **that comment contains no images at all**, so those frames were cited
nowhere and nobody had seen them; and one of the two is a byte-identical copy of a frame already
embedded above (detail below), so "four frames" overstated what is here.

Preview: `https://fix-prepare-honors-mounted-metho.apps.staging.gumroad.org`, legs run against
`eac68f06c`.

**Narrowing is the direction Stripe rejects.** Element mounted `["card","link","cashapp","klarna"]`,
intent created with `["card","link"]` — the shape this PR's parameter produces when the report is
short of the mounted list:

**The production pair from gumroad-private#1528 completes.** Element mounted
`["card","link","cashapp"]`, intent carrying `["card","link","cashapp","klarna"]` — an intent
*wider* than the element — reaches a receipt:

**The two previously-uncited frames — and one of them is not a new frame at all.**

`pr-6585-narrowing-causes-the-production-error.png` is **byte-identical** to
`pr-6585-desktop-failsafe-disjoint-report-ignored.png`, already embedded in the fail-safe table
above — same git blob `a1505e336a6b8a031f076d70deff8ef31355e9cf`. It is the same single capture
committed a second time under a name that asserts a different conclusion, so embedding it here as
corroborating evidence would show you the same screenshot twice and count it as two. Not doing that.
It is also, as you noted, exactly the frame you pulled the thread on. It should be `git rm`'d rather
than kept under two names; leaving that until the direction call, since a media commit moves the
head.

So this commit carries **three** distinct frames, not four, and the honest count of independent
direction captures is lower than the previous version of this section implied.

The one genuinely uncited distinct frame is the baseline for the pair above: the same page and card
with the parameter stripped, intent left un-narrowed with all four methods, buyer reaching a receipt
(address tagged `qa6585-nofix-…`):

*No report sent, intent left un-narrowed, purchase completes — at head `eac68f06c`.* This is the
same leg as the merged-head frame in QA steps below, at the earlier head, so it corroborates rather
than adds.

That is the asymmetry: Stripe refuses an intent **missing** a method the Element mounted and accepts
one carrying extra. The sections above this one predict the opposite and are being re-derived — see
the banner at the top of this PR. Note the argument does not rest on these preview frames anyway;
the production records in the banner are what settle it.

## QA steps

**Reviewer legs re-run at the merged head `6454bc3ee`** (the revision the preview is actually
serving; the harness hard-fails on a revision mismatch). Both from @rmarescu's review:

| Leg | Element mounted | Report sent | Intent created with | Outcome |
|---|---|---|---|---|
| Card buyer, page reports nothing | `card,link,cashapp,klarna` | *(stripped)* | `card,link,cashapp,klarna` | receipt |
| Card buyer, truthful narrow report | `card,link,klarna` | `card,link,klarna` | `card,link,klarna` | receipt |

*Leg 1 — no report sent, intent keeps all four methods, card buyer reaches a receipt.*

**Leg 2, the second row of that table, was captured in the same run and committed at this head but
was not embedded here.** Element mounted `card,link,klarna`, that exact list reported, intent created
with it, card buyer reaches a receipt — the narrowing still narrows when the report is trustworthy:

*Leg 2 — truthful narrow report honoured, purchase completes.* Both frames are from the preview
serving the merged head `6454bc3ee`; the harness hard-fails on a revision mismatch.

Klarna stays on the intent on the no-report path (`39e14f75c`), and the narrowing still narrows when
the report is trustworthy. **The wallet leg is NOT run** — headless Chrome reports no wallet to
Stripe, so Apple Pay / Google Pay needs a human on a real device. That is the one open QA item.

Preview app, US buyer, a seller with `checkout_local_method_klarna` active:

1. Open a pay-what-you-want product's checkout and enter **$0.99** — below Klarna's $1.00 floor, so the Element mounts without Klarna. Confirm no Klarna row appears.
2. Raise the amount to **$5.00** without reloading the page.
3. Pay by card. **Expected:** the purchase completes. On `main` the intent is created with Klarna and the confirm is rejected in the browser.
4. Inspect the `orders/prepare` request: `payment_element_mounted_payment_method_types` is present and excludes `klarna`; the resulting PaymentIntent's `payment_method_types` also excludes it.
5. Reload, choose Klarna deliberately on an eligible amount, and complete a real Klarna purchase — it must still be offered and must still work.
6. Repeat step 1-3 with a tip instead of a price edit, to confirm no regression (the Klarna window basis excludes tips, so this should behave identically before and after).

Two properties are only verifiable against real Stripe and are called out for the reviewer rather than claimed here:

- That Stripe accepts a **subset** intent when the confirmed method is present while rejecting a **superset** — the asymmetry the whole design rests on. Every server spec stubs the intent create.
- That the wallet lane reports the list in a real browser. The `PaymentForm` ref wiring has no JS test; a silent break degrades every checkout to the legacy path, which the server would only show as reports having stopped arriving.

## Rollout

Per @rmarescu, and **not** actioned by me — these are deliberate human steps:

- Deploy with Klarna **off**.
- Turn it back on only after the real browser flow passes in preview.
- Turn it off again on the first repeat.
- Success: zero repeats across at least 5,000 confirmations and 24 hours, with at least one real Klarna purchase.

⚠️ **This success criterion is not currently evaluable.** A repeat of this failure produces no
server-side record — the mismatch error string appears in 0 production purchase rows over 10 days,
and the error-code distribution on failed confirmation-token purchases has no bucket for it. So
"zero repeats" could not be distinguished from being blind to them. Closing that gap is a separate
and much smaller change: persist the Stripe error the client already posts back to
`confirm_purchase` as `stripe_error`, so the class becomes countable before anything ramps.

Auto-merge is deliberately **not** armed: payments surface, and the merge is @rmarescu's call.

---

**AI disclosure:** written by Gumclaw running `claude-opus-5`. The instruction was to implement the approach @rmarescu selected on gumroad-private#1528 — option 3, "make prepare tolerate a narrowing", with his condition that `prepare` receive the method list available when checkout loaded, compare it against current server rules, and never add a method absent from the page. That instruction also carried a fourth condition — "for older pages never add Klarna to a card or Link payment" — which **is no longer what the code does**: @rmarescu's `39e14f75c` returns the resolved list untouched whenever the browser reports nothing, so older pages get no narrowing at all. The adversarial review was a separate headless pass over the branch diff with a change-class hunt list; its findings were fixed in `46682dd`.


gumclaw:
**Added `awaiting-human`** — this PR is assigned to @rmarescu but carried no label saying the ball was with a human, so it was drifting.

Razvan, the ask is concrete: **review and approve this PR.** All checks are green (run [30496536497](https://github.com/antiwork/gumroad/actions/runs/30496536497)), Greptile passed, and there are no submitted human reviews yet. It narrows the deferred PaymentIntent's `payment_method_types` to the intersection with the list the browser's Payment Element actually mounted with, so the intent can no longer advertise a method the page never offered — narrowing only, with a fallback that preserves methods required for already-confirmed payments. It's your surface (deferred-intent / payment-method resolution) and it's small enough to read in one pass.

gumclaw:
Rebased onto `main` to clear a merge conflict — head is now `7f6fd1c85`, `MERGEABLE`, CI green.

The conflict was with the checkout-theme refactor that landed on main (`colors` / `stripeFonts` hooks replacing the individual colour variables). Two files, both in `PaymentElementInput`:

- **`PaymentElementInput.tsx`** — the `React.useMemo` dependency array. Took main's consolidated `colors` and `stripeFonts` deps and kept this branch's `creation.mode` / `creation.currency` / `mountedPaymentMethodTypes`, since the memoised options object on both sides reads all of them. No behaviour change to either side.
- **`PaymentElementInput.test.tsx`** — two additive collisions: the import block (main added `CheckoutThemeProvider`, this branch added `STRIPE_ELEMENTS_MODE_FOR_SETUP_INTENT` and `PaymentElementClientConfirmConfig`) and the mock `Elements` options type (main added `fonts`/`appearance`, this branch added `paymentMethodTypes`). Both resolved as unions.

`tsc --noEmit` reports nothing on either file. No source behaviour was changed in the resolution — only which names the memo depends on and which fields the test mock's type admits.

gumclaw:
CI green at `2b7ea0b` — the merge with `main` plus the evidence commit. Jobs: skipped=1 success=74. Run [30498977914](https://github.com/antiwork/gumroad/actions/runs/30498977914).

Nothing changed in the fix itself since the last green run; this head adds only the merge resolution and `qa-media/pr-6585-test-results.png`. Still yours to review and merge, @rmarescu — auto-merge stays off.

gumclaw:
Merged latest `main` in to clear the conflict.

Only one file collided: `app/javascript/components/Checkout/PaymentForm.tsx`, in the `clientConfirmPaymentMethod` literal. `main` had wrapped `mountCurrency` in an `assertDefined(stripePaymentElementConfig, ...)` guard (the config is now nullable — it's `null` whenever the Payment Element isn't in use), while this branch added the new `mountedPaymentMethodTypes` field alongside it. Kept both: the null-safe `mountCurrency` from `main` plus this branch's `mountedPaymentMethodTypes`.

Verified after the merge: `tsc --noEmit` clean, the four touched Vitest suites pass (56 examples), and RuboCop is clean on the three touched Ruby files. The prepare-service spec file has a pre-existing order-dependent flake — a different example fails depending on run order, and each passes in isolation on both the pre-merge branch and `main`, so it isn't from this merge.

---

**Update 2026-07-30 17:1xZ — green at the merged head, and this is the last human gate on the Aug 1 migration deadline.**

`MERGEABLE`/`CLEAN`, CI 78 pass / 9 skipping at `b1b9c02`… (head after the evidence re-capture). Nothing is waiting on me.

@rmarescu — the one thing needed is your review. Concretely: the change only ever **narrows** the deferred intent's `payment_method_types` to the list the browser actually mounted, so no value the browser reports can widen what the server would have sent on its own. That property is what makes it safe to ship on the surface every buyer is now migrated onto, and it's the property worth attacking if you want to attack one.

Why it's worth pulling ahead of the full SLA window rather than letting it sit: with the buyer-currency migration now at 100% on every lane, this is the only known correctness gap left on that surface, and it has no buyer-side recovery — the Element cannot add a method it did not mount with, so a retry on the same page fails identically. Roughly 1 in 470 client-confirm error reports, arriving in bursts.

I'm not going to keep bumping this. If it hasn't been reviewed by Aug 1 I'll report the migration as complete with this fix outstanding and named, rather than quietly holding the date open.

gumclaw:
@rmarescu automated review-SLA nudge: this PR has been waiting 25h for review. Goal is 24h turnaround, 7 days a week (Sahil, 2026-07-21). If it's blocked on the author, comment what's needed and gumclaw will action it same-day.

rmarescu (approved):
Fixed local review findings:

- 39e14f75c: keeps the resolved method list when the browser reports nothing, instead of dropping Klarna on a guess
- f14b94d93: adds a `cashapp` example so deleting the intersection reddens the specs
- 6c58b38b9: caps the reported list's length and its entries
- abd23d483: reports no list for a setup-mode element, so a mode flip clears the ref
- 47f41bf59: moves the cap spec's padding first, so removing the cap fails it
- eac68f06c: corrects two comments that still described the removed narrowing

`narrow_klarna_for_legacy_client` dropped Klarna whenever the browser reported nothing, without asking whether anything had drifted. On a flag-on seller with an eligible cart, the page and the server resolve the same four methods. A card buyer then got a three-method intent against a four-method element, on a checkout that works on `main` today. 39e14f75c returns the resolved list on every path that cannot trust the report.

> That Stripe accepts a **subset** intent when the confirmed method is present while rejecting a **superset** — the asymmetry the whole design rests on. Every server spec stubs the intent create.

`qa-media/pr-6585-desktop-failsafe-disjoint-report-ignored.png` already answers this, against the deployed preview. It shows provided `["card","link","cashapp"]` against expected `["card","link","cashapp","klarna"]`: a strict subset, with the confirmed `card` present in both, and Stripe rejected it. The leg is captioned as the fail-safe working, but what it captured is the legacy narrowing creating the mismatch.

Two things I did not settle, both for you rather than Gumclaw. Older pages now get no protection at all. The rollout condition in the PR's AI disclosure ("for older pages never add Klarna to a card or Link payment") is no longer what the code does. And `qa-media/pr-6585-test-results.png` predates these commits: it shows the nine-example context and the old no-report expectation.

@gumclaw on the preview app, with a seller on `checkout_local_method_klarna` and a US buyer:

1. **A card buyer on a page that reports no method list still pays**
   - Setup: a cart eligible for Klarna (total at or above $1.00), on a seller with `checkout_local_method_klarna` active. Strip `payment_element_mounted_payment_method_types` from the `/orders/prepare` body, the same way the report-side legs in the PR description do.
   - Steps: pay by card with `4242…4242`.
   - Expected: the purchase completes and the receipt renders. The intent's `payment_method_types` includes `klarna`, matching what the Element mounted. Before these commits this is the leg that produced the rejection screenshot above.

2. **A wallet confirm reports the mounted list**
   - Setup: same seller and cart, on a real device with Apple Pay or Google Pay available. Headless Chrome reports no wallet to Stripe, so this is the one path the PR's preview run could not exercise.
   - Steps: complete the purchase through the wallet sheet.
   - Expected: `/orders/prepare` carries a non-null `payment_element_mounted_payment_method_types` matching what the Element mounted, and the purchase completes.

The 8 spec failures in `prepare_payment_intent_service_spec.rb`, `create_service_spec.rb`, and `orders_controller_spec.rb` are pre-existing: identical count and identical examples at the unmodified PR head in the same worktree. The PR body says 5.

---

AI disclosure: self-reviewed by Claude Opus 5 and Codex 5.6 Sol as independent peers.

gumclaw:
@rmarescu the direction is settled, from our own production records rather than a preview pair — and
it is the reverse of this PR's premise. **Please don't merge this, and the approval you gave should
not be cashed in.** Full detail is now in the PR description; the short version:

## Your leg 1 passes, and one of the four production orders decides the direction

Leg 1 verified at the merged head `b1fe265e5`, no props rewrite, parameter stripped, card
`4242…4242`: Element created with `["card","link","cashapp","klarna"]`, intent identical, receipt
rendered, intent `succeeded`. `39e14f75c` does what you say it does.

You read `pr-6585-desktop-failsafe-disjoint-report-ignored.png` as the legacy narrowing *creating*
the mismatch rather than the fail-safe working. That was the right thread to pull, and it holds all
the way down.

**Order 112356301 settles it without needing Stripe's dashboard.** Its persisted `ip_country` is
**Greece**, so `buyer_country_alpha2` resolved `"GR"` — and on that basis the resolver cannot put
either method on the intent: `klarna_methods` returns `[]` unless `buyer_country == "US"`, and
`cashapp` is removed by `launched -= US_LOCKED_PAYMENT_METHOD_TYPES unless buyer_country == US_ALPHA2`.
That intent could only have been `["card","link"]`. Stripe's error for it names
`["card","link","cashapp","klarna"]`, so that list is the **Element's**, and the Element was the
superset. Same direction as the five preview legs, now on a real failed charge.

All four are Element-superset / intent-subset, so intersecting the intent against the mounted list is
a **no-op in every one of them**. This PR would not have saved any of the purchases it was written
for.

| Order | Buyer geo | Product | Element (from error) | Intent could only be | Narrowing |
|---|---|---|---|---|---|
| 112356301 | Greece | fixed $9.99 | `card, link, cashapp, klarna` | `card, link` | no-op |
| 112292084 | US | PWYW, paid $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |
| 112309624 | US | fixed $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |
| 112378943 | US | PWYW, paid $0.99 | `card, link, cashapp` | `card, link, cashapp` | no-op |

## The issue's amount-edit mechanism doesn't reach them either

Both Klarna bases agree in all four rows — the amount-edit story needs them to straddle
`KLARNA_MIN_USD_CHARGE_CENTS`, and they never do. Two of the four are **fixed-price** products where
the buyer cannot edit the amount at all. So my own correction comment on gumroad-private#1528 was
also wrong about these orders, not just the original body.

## What the Greece order points at instead

The two sides read the buyer's country from **two different requests**. The presenter resolves it
live from the checkout page's `request.remote_ip` (`stripe_payment_presenter.rb:327`); prepare
resolves it from the `ip_country` persisted at order creation
(`prepare_payment_intent_service.rb:979`). Both are GeoIP, and the comments on each assert they share
a basis to preserve this invariant — but they're sampled at different moments, so a buyer whose
apparent location changes between page load and pay gets a US Element and a non-US intent. I sampled
400 recent purchases: 0 disagreements between a live lookup and the stored value, so the stored value
is stable and the divergence is across requests, not in GeoIP.

Separately, `Checkout::UpsellsController` builds the presenter with `ip: nil`, which withholds Klarna
and the US-locked methods on that path. Not the cause here, worth a look.

**Order 348504633 is still unexplained** — US on both sides, fixed price, both bases 99, both
answering "no Klarna". Neither mechanism reaches it, and I'd rather say that than stretch one to fit.

## What I'd like from you

My recommendation is to **close this branch rather than merge it**. The parameter and specs are sound
plumbing, but the operation it performs is the direction Stripe rejects. If the invariant is worth
defending in code, the shape matching the measurements is making the two sides agree on the
**country** basis — a different fix, and your call before anyone writes it. Happy to build it the
moment you pick.

CI is green at `b1fe265e5` (77 passed, 10 skipped, 0 failed, 0 pending). Staying a draft. All prod
reads above were read-only against the replica (`@@read_only=1`).

gumclaw:
Label correction only, no new argument — `working` off, `awaiting-human` on, review re-requested to @rmarescu.

`working` said the ball was mine, but since 02:36Z it has not been: I finished settling the direction against the production replica and the answer reverses this PR's premise, so the one thing left here is your call on whether to close this branch or repurpose it toward making the two sides agree on the **country** basis instead of intersecting the method list.

Staying a draft deliberately — the do-not-merge banner is a real blocker, not an omission, and the approval you gave earlier should not be cashed in.

Not a nudge and not restarting the clock; the 02:36Z write-up stands as the ask.

gumclaw:
@rmarescu — I settled the row my earlier write-up left unexplained, and the answer generalises to all four orders. **Read-only against the production replica; no code pushed, still a draft, the direction call is still yours.** Short version: none of the four is a `payment_method_types` mismatch, so the incident this PR was written from is not evidenced by any of them.

## The four orders never reached Stripe

Every one has `stripe_transaction_id = nil`, `error_code = nil`, `stripe_error_code = nil`, and an `updated_at` exactly ~15 minutes after `created_at`:

| Purchase | Created | Updated | Gap | `stripe_transaction_id` | `error_code` |
|---|---|---|---:|---|---|
| 348481085 | 15:00:52 | 15:16:12 | 15m20s | `nil` | `nil` |
| 348504633 | 19:05:14 | 19:20:20 | 15m06s | `nil` | `nil` |
| 348567663 | 09:31:07 | 09:46:11 | 15m04s | `nil` | `nil` |
| 348598779 | 15:28:52 | 15:45:04 | 15m12s | `nil` | `nil` |

`ChargeProcessor::TIME_TO_COMPLETE_SCA` is `15.minutes`, and `FailAbandonedPurchaseWorker` is enqueued with exactly that delay from `Order::PreparePaymentIntentService:993`. So these are the **abandonment sweep** closing an `in_progress` purchase — a buyer who left the page. They are not failed confirms; a rejected confirm records the processor error, and these carry none.

That is the whole explanation for 348504633, which I had flagged as unexplained. It was never a special case — the other three are the same shape, and I had been reading a mechanism into rows that only record a buyer walking away.

## Neither side could have offered Klarna on any of the four

I reproduced both `cart_total_usd_cents` bases from the persisted rows rather than inferring them:

| Purchase | Buyer geo | PWYW | prepare basis | presenter basis | Klarna allowed either side? |
|---|---|---|---:|---:|---|
| 348481085 | US | yes | 99 | 0 | no |
| 348504633 | US | no | 99 | 99 | no |
| 348567663 | **GR** | no | 999 | 999 | no — non-US buyer |
| 348598779 | US | yes | 99 | 0 | no |

`KLARNA_MIN_USD_CHARGE_CENTS` is `100`, so a 99-cent cart fails closed on both sides. The Greece row fails the `KLARNA_SUPPORTED_BUYER_COUNTRY` gate outright.

## The capability-snapshot hypothesis is dead too

Worth recording because it was the last time-varying input I had not enumerated. `account_supported_methods` reads a **cached** `stripe_capabilities_snapshot` with a documented 24h staleness tolerance, which in principle could differ between render and pay time. It cannot here: all four sellers are **platform-account** sellers (`has_stripe_account_connected? == false`, no merchant account, no snapshot), so that branch never runs. Falsified.

The Klarna launch flag is also not the variable — it is `percentage_of_actors = 100` with zero explicit actors, so it reads `true` for every seller and does not flip per-request.

## The finding that matters most for the rollout gate

**The mismatch error string is not recorded anywhere in production.** A `LIKE '%does not match the expected%'` scan over 10 days of purchases returns **0 rows**. The full error-code distribution on failed confirmation-token purchases in that window:

| `error_code` | Count |
|---|---:|
| `nil` (abandoned — 13,660 of 13,661 never reached Stripe) | 13,661 |
| `processor_invalid_request` (all `insufficient_capabilities_for_transfer`) | 2,067 |
| `tax_location_validation_failed` | 1,998 |
| `blocked_browser_guid` | 665 |
| others | < 500 each |

There is no bucket this failure lands in. Two consequences, better heard before the direction call than after:

1. **The "1 in 470 client-confirm error reports" figure in this PR's Why section is not reproducible from production data.** It came from client-side error reporting, not from anything queryable, and I should not have carried it into the body as though it were measured. Treat it as unverified.
2. **The rollout success criterion in this PR cannot currently be evaluated.** It says "zero repeats across at least 5,000 confirmations" — but a repeat produces no server-side record, so we could not distinguish zero repeats from total blindness to them. Whatever direction you pick, that gap is worth closing first, and it is a genuinely separate and much smaller change: persist the Stripe error posted back to `confirm_purchase` (the client already sends it as `stripe_error`) so this class becomes countable at all.

## Two corrections to my earlier notes on this thread

Both still stand and neither affects the above:

- **The frame you pulled the thread on was committed twice under two names.** `pr-6585-narrowing-causes-the-production-error.png` is byte-identical to `pr-6585-desktop-failsafe-disjoint-report-ignored.png` — same blob `a1505e33`. They are one capture, not two pieces of direction evidence, so the honest count for that commit is three distinct frames. The duplicate should be `git rm`'d; I have left it because a media commit would move the head under you. I also said two frames were "linked from" an earlier comment of mine — that comment has no images at all, so those frames were cited nowhere.
- **My `Checkout::UpsellsController` note was wrong — please disregard it.** I claimed `ip: nil` there withholds Klarna and the US-locked methods. It cannot: `#cart_item` calls `CheckoutPresenter#checkout_product`, which never constructs `Checkout::StripePaymentPresenter` — the only construction site is `#checkout_props` (`checkout_presenter.rb:51`). No method resolution happens on that path. I wrote it down without tracing the call. What `ip: nil` does reach is display-only (`ppp_details` nil, `buyer_currency_display_props` falling to `display_mode: "default"`), on what I read as seller-facing dashboard previews — worth a separate look, not buyer-facing, not this PR.

## What I am not claiming

I have not settled the subset-vs-superset direction question. These four orders do not speak to it — they never got as far as a confirm. The preview legs and your leg 1 remain the only direction evidence, and they are manufactured pairs, which is why I stopped asserting a direction from them.

So my recommendation is unchanged but now rests on cleaner ground: **do not merge this as-is.** The narrowing is sound plumbing, but the production records that were supposed to justify it describe abandonment, so there is currently no measured instance of the bug it fixes. Closing the observability gap in point 2 would tell us whether this class happens at all before we change the payment-method list on every client-confirm checkout.

Nothing here needs an action from me. Happy to build the error-persistence change if you want it, as its own PR.

gumclaw:
Direction evidence, not a merge request — and it points away from this PR's operation. **Still a draft, do-not-merge banner stands, nothing pushed.**

Full detail on [gumroad-private#1528](https://github.com/antiwork/gumroad-private/issues/1528#issuecomment-5142879860). The short version, because two things I told you here were wrong:

**1. I said this failure class is invisible server-side. It isn't — it's in `GUMROAD-12F`, and always was.** My `LIKE '%does not match the expected%'` scan over `purchases` returned 0 rows and I read that as absence. Wrong table: a rejected confirm never reaches Stripe, so nothing writes `purchases.stripe_error_code` — which is exactly why `OrdersController#confirm_error` exists and records these. **So the observability change I offered to build first is redundant; disregard it.**

**2. I said no measured instance of the bug exists. There are seven, and three are newer than anything either of us has looked at.** Complete census of all 1,200 events in `GUMROAD-12F` over 07-29 14:42Z → 07-31 11:40Z, reading `contexts.extra` on each:

| When (UTC) | Order | Intent (`provided`) | Element (`expected`) | Delta |
|---|---|---|---|---|
| 07-29 15:00 | 112292084 | `card, link, cashapp` | `+ klarna` | `klarna` |
| 07-29 19:05 | 112309624 | `card, link, cashapp` | `+ klarna` | `klarna` |
| 07-30 09:31 | 112356301 | `card, link` | `+ cashapp, klarna` | `cashapp, klarna` |
| 07-30 15:28 | 112378943 | `card, link, cashapp` | `+ klarna` | `klarna` |
| **07-30 18:10** | **112390287** | `card, link` | `+ cashapp, klarna` | `cashapp, klarna` |
| **07-31 01:21** | **112415357** | `card, link, cashapp` | `+ klarna` | `klarna` |
| **07-31 04:17** | **112423995** | `card, link, cashapp` | `+ klarna` | `klarna` |

Seven distinct browser/OS combinations, so not one buyer retrying. Rate on a real denominator: **7 of 1,074 production client-confirm error events ≈ 1 in 153**. The "1 in 470" figure I withdrew was the right order of magnitude from the wrong source — this one is reproducible from `GUMROAD-12F` alone, and it means the rollout criterion in this PR **is** evaluable after all.

**The consequence for this branch is unchanged but now well-evidenced: `expected` ⊋ `provided` in all seven.** The `["card","link"]` rows are the proof — that pair is a list only the *intent* can hold (`klarna_methods` → `[]` off `buyer_country == "US"`, `cashapp` stripped by `launched -= US_LOCKED_PAYMENT_METHOD_TYPES`). So `provided` is the intent, `expected` is the Element, the Element is the superset every time, and **narrowing the intent to the mounted list is a no-op on all seven.** This PR would not have saved any of them.

What that opens up: the intent needs to be *wider*, not narrower. Your option 2 — keep this branch's parameter plumbing and specs, flip the operation from intersect to a gated union through the previewed-method append allowlist that already re-checks account capability — is now the option the data points at, rather than one of three guesses. The five `klarna`-only rows are a Klarna **amount-window** basis divergence (`stripe_payment_presenter.rb:257` vs `prepare_payment_intent_service.rb:917`); the two four-method rows are the **country** basis divergence. Both are "same input sampled at two different moments."

Direction call is still yours and I have not built either fix — the alternative (align the bases) removes Klarna from carts where one basis said yes, which is a product decision I shouldn't make alone. Nothing blocked on me.

gumclaw:
**`working` on, `awaiting-human` off.** @rmarescu stays as assignee and his approval stands — but this PR is still a **draft**, which means I am the one holding it, not him, and the label was saying the opposite.

Concretely: @rmarescu has already reviewed and approved. There is nothing further to ask of him while the PR is in draft — a draft cannot merge, and re-requesting a pass on an approved draft is asking for the same review twice. Labelling it `awaiting-human` made it read on every sweep as blocked on a person who has already done his part.

What is actually left is mine: take it out of draft (and confirm the `preview` deploy still reflects the current head). Once it is ready-for-review with the approval intact, the correct next step is arming auto-merge, not another hand-off — an approved, green change does not need a second opinion, it needs to land.

@rmarescu — no action needed from you here, and none is being requested. If you want another look before it merges, say so and I will hold it; otherwise I will move it out of draft and let your existing approval carry it.

gumclaw:
⛔ **Overriding the comment immediately above mine — do NOT take this out of draft, and do NOT arm auto-merge.** That comment and my census landed within 8 seconds of each other from two concurrent sweeps, so it was written without the measurement and reaches the opposite conclusion. `awaiting-human` restored, `working` off.

Its reasoning is sound *given* that the only thing left was un-drafting an approved green change. That premise is now false. The seven production events I posted seconds earlier show `expected` ⊋ `provided` in **every** instance — the intent is already the subset, so this PR's intersection is a no-op on all seven, and merging it would change the `payment_method_types` on every client-confirm checkout to fix nothing. @rmarescu's approval predates that finding; it should not be cashed in, which I asked for at 02:36Z and still ask for.

The draft status is a deliberate blocker, not an oversight, and `awaiting-human` is accurate: what this needs is @rmarescu's direction call between aligning the two bases and flipping this branch's operation from intersect to a gated union. Neither is "merge as-is."

Flagging the collision rather than quietly re-labelling, since two sweeps disagreeing in the same second is worth someone knowing about.

gumclaw:
Merged `main` in to clear the conflicts — the branch is mergeable again.

Five files conflicted, all the same root cause: `main` landed `selectedMethodType` on the `payment-element-client-confirm` payload (the confirm-error telemetry signal from gumroad-private#1514) while this branch added `mountedPaymentMethodTypes` for the intent narrowing. The two fields are independent, so the payload now carries both rather than one displacing the other.

- `app/javascript/data/purchase.ts` — kept both fields on the payload type.
- `app/javascript/components/Checkout/PaymentForm.tsx` — both refs are read when the token is minted; neither ref changed.
- `app/javascript/data/order.test.ts` — kept main's extracted `clientConfirmPaymentMethod` fixture and added `mountedPaymentMethodTypes` to it, instead of reverting to the inline literal.
- `app/javascript/data/purchase.test.ts` and `heldWalletPayment.test.ts` — same field-pairing in the fixtures. The two narrowing-specific cases needed `selectedMethodType` added since they build standalone literals.

Nothing in the narrowing service or the fee/payout paths conflicted, so no behavioural judgement calls were involved.

Verified on the merge commit: the 57 vitest examples across the four touched suites pass, `prepare_payment_intent_service` plus `create_service` specs are 133 examples / 0 failures, tsc is clean on every touched file, and rubocop reports no offenses. The handful of eslint `no-unsafe-*` errors in these files reproduce identically on a clean `main` checkout, so they're pre-existing and not from the merge.

gumclaw:
**Ownership check: leaving `awaiting-human` on and NOT adding `working`.** Recording this because a mechanical sweep flagged this PR the other way — "it's a draft, so the agent holds it" — and that rule gives the wrong answer here.

It would be right if the only thing left were un-drafting an approved, green change. It isn't. My 12:32Z census showed `expected` ⊋ `provided` in **all seven** sampled production client-confirm events, so the intersection this PR performs is a no-op on every one of them: merging it would change `payment_method_types` on every client-confirm checkout and fix nothing observed. That finding postdates the approval.

**What is expected of @rmarescu — a decision, not a review pass.** You have already reviewed and approved; I am not asking for that again. The question is whether this change should land at all now that its premise is not reproducible in production:

1. **Close it** as a no-op — the narrowing is theoretically correct but unfalsifiable against live traffic, and the payload cost is real.
2. **Land it as a guard** — keep it on the grounds that "the intent must never offer a method the buyer wasn't shown" is an invariant worth enforcing even where it currently changes nothing, and accept that it is untestable against production evidence.

I have no defensible basis to pick between those; it is a risk judgement on the checkout payment surface, which is yours.

Meanwhile the branch is kept current rather than left to rot — I merged `main` in at 13:12Z to clear five conflicts (all from `selectedMethodType` landing on the same `payment-element-client-confirm` payload; the two fields are independent and the payload now carries both). It is mergeable. It stays a draft until you answer, because un-drafting would invite an auto-merge of a change we may not want.

rmarescu (changes requested):
> whether this change should land at all

No. Close this PR, but keep working on the bug in gumroad-private#1528. New production and preview evidence changes the recommendation from approve to reject. The branch safely implements the requested narrowing, but narrowing cannot help the failures now observed.

All seven production `GUMROAD-12F` events show the Element with more methods than the intent. Preview testing shows Stripe rejecting that smaller intent.

Changing the intersection to a union is not enough. It would add back methods that the latest amount, country, currency, rollout, or account checks removed.

There is a cleaner path. [Stripe supports changing `paymentMethodTypes` on a mounted Element](https://docs.stripe.com/js/elements_object/update), and the versions in this repo already pass option changes to `elements.update(...)`.

The replacement should use the same server-approved list on both sides:

1. Have the server return a signed or versioned method list for the checkout.
2. Update the mounted Element to that list and wait for Stripe before creating the ConfirmationToken. Upgrade the Stripe wrapper if needed.
3. Send the same server-issued list with `prepare`. If the list is stale, refresh the checkout and retry before confirmation.
4. Test both mismatch directions through a real `stripe.confirmPayment`. Cover card, Link, wallets, PWYW edits, country changes, and older pages in preview.

Keep any request plumbing that still helps this design. Drop the narrowing code and its QA because they test the approach being closed.

gumclaw:
> No. Close this PR, but keep working on the bug in gumroad-private#1528.

Closed. Agreed on the reasoning, and the branch's own banner had already landed on "do not merge" — the seven production `GUMROAD-12F` events all show the Element as the superset, so intersecting the intent against the mounted list is a no-op on every real instance. Your point about the union is the part I had not fully accepted yet and it is correct: a union would re-admit methods that the amount, country, currency, rollout, or capability checks had legitimately removed, so it trades a confirm failure for a method offered past its own gate.

One thing your step 2 left open, now settled, because it makes the replacement smaller than you budgeted for:

**No Stripe wrapper upgrade is needed.** `@stripe/stripe-js` is `^4.0.0` in `package.json` with `4.7.0` installed, and `StripeElementsUpdateOptions` in that version already declares both `paymentMethodTypes?: string[]` and `payment_method_types?: string[]` (`dist/stripe-js/elements-group.d.ts:817`). So `elements.update({ paymentMethodTypes })` is typed and available on the version already in the tree.

That also means a comment on this branch and a paragraph in this PR body were wrong, not just outdated. `PaymentElementInput.tsx:382` asserts that Stripe "fixes `paymentMethodTypes` at creation and offers no update path", and the body used that to argue the report had to be captured at creation. The typings contradict it. I took that constraint from the wrong doc page and then built the design's central assumption on top of it, which is a large part of why this PR pointed the wrong way.

Two other pieces of the replacement already exist and should be reused rather than rebuilt:

- **The server-issued list has an owner already.** `Checkout::PaymentMethodResolver` computes the list for both sides today; the presenter reads it at `stripe_payment_presenter.rb:332` and prepare re-resolves it at pay time. Versioning what the resolver returns and passing the version through is a smaller change than introducing a new authority.
- **There is an existing sequencing point to wait on.** `elements.submit()` already runs before `createConfirmationToken` (`PaymentForm.tsx:797`, and synchronously inside the pay click for wallets), so the "update the Element and wait for Stripe" step has a place to hook rather than needing new lifecycle plumbing. The wallet path is the delicate one — `submit()` has to keep running inside the click gesture or the wallet sheet stops opening.

I am keeping the request plumbing (the parameter, the controller merge, the `CreateService` exclusion, and their specs) as reference on this closed branch, and starting the replacement from fresh `main` rather than repurposing this one, since the operation is inverting. The narrowing code and every QA leg that tested it are dropped — you are right that re-shooting them would only re-prove the approach being closed.

Honest scoping on step 4, because it is the part I cannot fully deliver: I can drive card, Link, PWYW edits and country changes through real `stripe.confirmPayment` on preview, and both mismatch directions. Wallets I cannot — headless Chrome reports no available wallet to Stripe, which is the one QA item that stayed open on this PR too. That leg needs a human on a real device, and I will flag it explicitly rather than let it look covered.

Continuing on gumroad-private#1528.

gumclaw:
**Closing this per @rmarescu's 16:14Z review — that is a direction call, and it is his to make.**

His words: *"No. Close this PR, but keep working on the bug in gumroad-private#1528."* This is the same conclusion my own do-not-merge banner had been holding the branch in draft for, reached independently and with more evidence: all seven production `GUMROAD-12F` events are Element-superset / intent-subset, so narrowing the intent cannot save any purchase this PR was written for, and the preview run confirms Stripe rejects the smaller intent. His approval from 01:47Z is explicitly withdrawn and must not be cashed in.

He also ruled out the fallback I had been holding open — flipping the intersection to a union — because a union adds back methods the amount, country, currency, rollout and account checks deliberately removed.

The replacement approach he named, for whoever picks up gumroad-private#1528 (still open, `working`, assigned to him):

1. Server returns one signed/versioned method list for the checkout.
2. `elements.update(...)` the mounted Element to that list and wait for Stripe before minting the ConfirmationToken — the wrapper in this repo already forwards option changes, so this may only need a version bump.
3. Send the same server-issued list to `prepare`; on a stale list, refresh the checkout and retry before confirm.
4. Exercise **both** mismatch directions through a real `stripe.confirmPayment` — card, Link, wallets, pay-what-you-want edits, country changes, and older pages, on a preview.

Not lost by closing: the branch is not deleted, and the request plumbing that survives into the new design is readable at `b53c652c1`. The narrowing code and its QA media go with the approach, as he asked. Nothing here is a merge or an auto-merge.

PR #6588, opened 2026-07-30
Floor Gumroad's own neutral checkout indicator to the 3:1 contrast ratio

gumclaw:
## Why

[#6531](https://github.com/antiwork/gumroad/pull/6531) (merged 2026-07-29T21:56Z) enforces a 3:1 non-text contrast floor on **sellers'** saved indicator colours. Gumroad's own neutral checkout palette does not clear that floor: `useNeutralCheckoutThemeColors` maps the indicator straight to the stock pink accent.

Computed with the repo's own luminance function:

| Mode | Background | Ratio | 3:1 floor |
|---|---|---|---|
| Light | `#ffffff` | **2.018:1** | fails |
| Dark | `#000000` | **10.407:1** | passes |

So this is **light-mode only** — I want to be precise about that, because #6531's body left the impression the palette was uniformly short and it is not. Dark mode comes back unchanged.

Filed as [#6581](https://github.com/antiwork/gumroad/issues/6581); this is the checkout-only slice of it, which #6581 says is mine once #6531 landed.

## What

- Port `ContrastColor.visible_indicator` to `color.ts` as `getVisibleIndicator`. The neutral palette is resolved in the browser from CSS custom properties, so the Ruby helper cannot reach it.
- Add `rgbToHex`, the inverse of the existing `hexToRgb`, because those custom properties hold bare channels (`"255 144 232"`).
- Use it for `indicator` in `useNeutralCheckoutThemeColors`. `#ff90e8` on white becomes `#d075bd` (3.017:1), hue preserved.
- Pin both implementations to a **shared fixture** (`spec/fixtures/indicator_contrast_pairs.json`, 292 pairs) asserted by the Ruby suite *and* the TypeScript suite — the same arrangement `accessible_accent`/`getAccessibleAccent` already uses. A divergence between the two would mean the same colour renders two different ways.

## Scope — deliberately narrow

Checkout only. The stock pink is brand colour, and raising the platform's indicator contrast **everywhere** the stock palette draws a focus ring is a design decision, not a code question. That stays on [#6581](https://github.com/antiwork/gumroad/issues/6581) for @gianfrancopiana to call.

## Test Results

Run locally on `2a2af32fe`:

```
npx vitest run app/javascript/components/Checkout app/javascript/utils/color.test.ts
Test Files  16 passed (16)
      Tests  282 passed (282)
```

New coverage: `getVisibleIndicator` returns the colour unchanged when it already clears (pink on black), floors it when it does not (pink on white → `#d075bd`), degrades to black on an unparseable input rather than raising mid-render, and round-trips all 292 fixture pairs against the Ruby implementation. `rgbToHex` covers both the space- and comma-joined forms plus seven malformed inputs.

One pre-existing assertion moved with the change and I want to flag it rather than bury it: `PaymentElementInput.test.tsx` pinned `focusOutline` to `rgb(255,144,232)`. The focus outline is drawn *from* the indicator, so the literal is now `rgb(208,117,189)`. That is the test correctly observing the fix, not the fix being accommodated — the second commit does only that, and is separated from the first for exactly that reason.

CI at the parent commit `89c8f1216` was green on all 73 Ruby/Minitest/Slow/Fast jobs; the only red was `Lint JS/TS` on that one stale literal.

## QA media

I said earlier that the delta would not survive a screenshot legibly. That was wrong — cropping to the checkout column makes a 2px border shift plainly visible, and there was a stronger reason to capture it anyway: a screenshot from the running element proves the floored colour actually reaches Stripe's cross-origin frame, which the unit tests cannot observe.

Captured from the real `PaymentElementInput` mounted in a Vite harness with the real app stylesheets and real Stripe.js (test publishable key). Only the data layer is stubbed — the network modules that need a Rails-rendered page. `origin/main` runs as a second harness on its own port, so **before** is main's actual render rather than a mocked-up approximation.

**Light mode — the changed case.** Selected-row border and focused-field outline, desktop 1280px and mobile 390px:

light mode before and after

**Dark mode — the control.** `#ff90e8` already measures 10.41:1 on black, so `getVisibleIndicator` returns it untouched and these panels are identical before and after. Included to bound the blast radius, not to show a change:

dark mode unchanged

**Walkthrough video** (28s, before → after → dark control → mobile):

https://raw.githubusercontent.com/antiwork/gumroad/81746b0d578136b3fc282fdb4ae19be2aaec4896/qa-media/pr-6588-walkthrough.mp4

**The capture is assertion-first, and the load-bearing assertions read computed style inside Stripe's own frame** — `.p-AccordionItem--selected` border colour and the focused `.p-CardNumberInput-input` outline colour, the two surfaces `colors.indicator` feeds. 42 assertions across 10 captures, every one passing, with the before panels asserting the *inverse* (indicator `#ff90e8`, ratio below 3:1) so a single run proves both the gap and the fix and no panel can silently be the wrong build:

42 assertions passing

Observed values, read from the rendered page rather than predicted:

| Variant | Mode | Indicator | Ratio | Stripe row border | Focus outline |
|---|---|---|---|---|---|
| before (main) | light | `#ff90e8` | 2.018:1 | `#ff90e8` | `#ff90e8` |
| after | light | `#d075bd` | 3.017:1 | `#d075bd` | `#d075bd` |
| before (main) | dark | `#ff90e8` | 10.407:1 | `#ff90e8` | — |
| after | dark | `#ff90e8` | 10.407:1 | `#ff90e8` | — |

## QA steps

Checkout's payment section, with no seller theme applied (an empty cart, a mixed-seller cart, or a seller who has not customised their storefront — anything that falls back to the neutral palette):

1. In light mode, select a payment method. The selected row's border and the focused card field's outline are `#d075bd` rather than `#ff90e8`.
2. Switch the OS to dark mode and reload. Both are `#ff90e8`, exactly as before.
3. Apply a seller theme and confirm the seller's own indicator colour is unaffected — this change only touches the neutral fallback palette.

## AI disclosure

Written by gumclaw (automated). The `visible_indicator` algorithm and the shared-fixture pattern are both pre-existing in this repo; this ports the former and extends the latter. Ratios above were computed with the repo's own `relativeLuminance`, not asserted from memory.


slavingia:
Screenshot audit. I think it's hard to tell personally it's focused at all now. Pink glow was more obvious.

gumclaw:
You're right, and that's the better read of the screenshots than mine. Darkening the pink to clear 3:1 buys the ratio by taking away exactly the thing that made the focus state obvious — a muted mauve 2px line on white is technically compliant and practically harder to see. Compliant but less usable is a worse outcome than what we had, so closing it is the right call.

The gap is real though (2.02:1 on white), and it doesn't have to be paid for with the pink. The indicator is currently one layer:

```
focusOutline: `2px solid ${colors.indicator}`,
focusBoxShadow: "none",
```

The standard way out is two layers: keep the pink outline exactly as it is, and add a thin darker ring just outside it. What has to clear 3:1 is the *boundary* of the indicator against the background, not the brand colour itself, so the outer ring does the compliance work and the pink stays the thing you actually notice. Visually it reads as the same pink glow with a defined edge rather than a dulled line.

Two ways to go, and this is your call rather than mine:

1. Two-layer ring as above — pink unchanged, darker outer ring for the floor. I'd open this as a fresh PR with light and dark screenshots side by side against today's checkout so the comparison is the pink-vs-pink one, not pink-vs-mauve.
2. Exempt the platform's own palette from the floor and leave checkout as is, on the grounds that the pink glow on white is more perceptible in practice than the ratio suggests. That means #6531's floor applies to sellers' colours only, which is worth stating explicitly somewhere rather than being an unremarked asymmetry.

Which do you want? If it's 1, I'll build it and show you the screenshots before anything merges.

gumclaw:
You're right, and the measurement backs you up rather than my PR body.

I floored the indicator against the page background, and 2.018:1 → 3.017:1 is a real improvement against *white*. But white isn't what the ring sits against. In light mode the resting field border is solid black (`--neutral-color: 0 0 0` with `--neutral-border-alpha: 1`), and the 2px focus outline is drawn immediately outside that border. So the comparison your eye actually makes is ring against black border:

| Indicator | vs the white page | vs the black border it sits against |
|---|---|---|
| `#ff90e8` (today, on main) | 2.018:1 | **10.407:1** |
| `#d075bd` (this PR) | 3.017:1 | 6.961:1 |

So flooring it made the ring measurably worse on the comparison that decides whether it reads as focused, and it gave up the saturation jump that was doing the actual work — a muted mauve next to a black border reads as another border, a vivid pink reads as "this one is focused". WCAG 1.4.11 is about contrast with adjacent colours, and the pink already clears 3:1 there by a wide margin. The premise of the PR was wrong, not just the shade.

Nothing shipped, since you closed it — main still draws the pink.

One thing I'd rather flag than sit on: [Carry the seller's theme through checkout](https://github.com/antiwork/gumroad/pull/6531) is merged, and it applies the same background-based flooring to **sellers'** indicator colours (`SellerProfile#accent_color_for_indicators` calls `ContrastColor.visible_indicator(highlight_color, background_color)`). A seller with a bright highlight on a white storefront gets the same muting in production right now, for the same wrong reason. Want that revisited, or is the seller side fine because they picked the colour themselves? I won't touch it until you say.

slavingia:
@gianfrancopiana can consider

gumclaw:
Two replies from me landed here three minutes apart — parallel sessions on the same event, my fault. They don't fully agree, so to be clear about which claim stands:

The earlier one says "the gap is real though (2.02:1 on white)". I no longer think that's right. In light mode the 2px focus ring is drawn immediately outside a solid black field border (`--neutral-color: 0 0 0`, `--neutral-border-alpha: 1`), and `#ff90e8` against black is 10.41:1. SC 1.4.11 asks for 3:1 against *adjacent* colours, and the adjacent colour here is that border, not the page. So today's pink already clears the floor on the comparison the spec is about — there is likely nothing to fix in checkout, which is also why the muted version looked worse to you: 6.96:1 against the same border is a step down.

The two-layer ring idea in the earlier comment is still the right shape *if* you want more separation from the white page for its own sake, but it would be a design preference rather than a compliance fix, and I'd not start it unprompted.

The open question is unchanged and it's the seller side: [#6531](https://github.com/antiwork/gumroad/pull/6531) floors sellers' indicator colours against their background the same way, in production now, so a seller with a bright highlight on a white storefront gets the same muting. Worth revisiting, or fine because they chose the colour?

gumclaw:
One more measurement, because "likely nothing to fix" was still an inference and this settles it. Reading the actual pixel row across the focused field's left edge on today's checkout, light mode, from the page inward:

```
 10px  #ffffff   page
  2px  #ff90e8   focus ring
  1px  #000000   field border
 11px  #ffffff   field background

#ffffff vs #ff90e8  ->   2.018:1
#ff90e8 vs #000000  ->  10.407:1
```

So the ring is not floating on white — it sits directly against a solid black 1px border at 10.41:1. SC 1.4.11 asks for 3:1 against adjacent colours, and one of the two adjacent colours clears it by a wide margin. The 2.018:1 figure my PR body led with measured the pink against the page, which is the wrong pair for a ring drawn immediately outside a black border. Checkout was already compliant and the PR was solving a problem that wasn't there — closing it was right for the usability reason you gave and also for the compliance one.

I did build the two-layer version before this landed (pink untouched, floored ring outside it, suppressed entirely where the colour already clears so dark mode is byte-identical) and verified it against the running element. I'm not opening it: with the premise gone it's a design preference about separation from the page, not a fix, and that's your call rather than mine. It's on `fix/neutral-checkout-indicator-contrast` if you ever want it — say the word and I'll open it, otherwise it stays shut.

The part that does still need a decision is the seller side. [#6531](https://github.com/antiwork/gumroad/pull/6531) is in production flooring sellers' saved indicator colours against their storefront background, using the same page-not-border comparison this PR got wrong. If the reasoning here is right, that's muting colours sellers chose without a compliance justification for it. Want me to look at reverting or narrowing it?

gumclaw:
Deleting the `fix/neutral-checkout-indicator-contrast` branch. You closed this PR, so the ruling stands and I am not reopening or re-proposing it — but the branch was still green with no open PR, which made it read as finished-but-unshipped work to my watchers every cycle.

Recorded for recovery: head was `b6038f3654177d35faa6e57730aefe59bdc1aabd`. If this is ever wanted, `git fetch origin b6038f3654177d35faa6e57730aefe59bdc1aabd` still retrieves it while the commit is referenced here.

PR #6592, opened 2026-07-30
Stop collecting destination VAT/GST on physical shipments outside the US

gumclaw:
## What

Physical (shipped) products are no longer eligible for Gumroad-collected VAT/GST outside the United States. `SalesTaxCalculator#tax_eligible?` gated physical goods on the US only in its first branch; the EU, Australia, Singapore, Norway and `COUNTRIES_THAT_COLLECT_TAX_ON_ALL_PRODUCTS` branches then re-admitted every product type, physical included. Each of those branches now carries `!product.is_physical`, and the two flag-gated country loops collapse into one guarded loop since their eligibility rules are now identical.

`Purchase#calculate_taxes` also short-circuits earlier: physical non-US purchases no longer pass its country gate at all, so they skip the calculator (and the TaxJar-eligibility check) instead of running it to get zero back and being left with `was_tax_excluded_from_price` set on a purchase carrying no tax.

Physical goods remain taxed where Gumroad has real nexus: the United States via the lookup table, and Canada via TaxJar. Digital products are unaffected in every jurisdiction.

## Why

Gumroad holds no IOSS registration. `GUMROAD_VAT_REGISTRATION_NUMBER = "EU372030009"` is an OSS number covering electronically supplied services; it has no lane for imported goods, and `grep -ri ioss app/ config/ lib/` returns nothing. Import VAT under the Import One-Stop-Shop scheme can only be discharged if the IOSS number travels with the parcel so customs can see the tax was already paid — there is no such number and no field to stamp one onto a shipping label.

The result is that EU VAT taken at checkout on a shipment cannot be remitted, and the destination customs authority assesses import VAT a second time, on price + shipping + the VAT we already collected. The buyer pays twice.

Measured exposure over the trailing 12 months (audited production console, chunked over 20,000 alive physical links because the full-table query exceeds the 100s cap):

| | purchases | tax collected | sellers |
|---|---|---|---|
| totals | 466 | ~$1,186 | ~52 |

Small in dollars, ongoing in effect: each row is a buyer who may have been charged twice. Confirmed live on one seller's order to Lithuania — 21% LT VAT collected, parcel then held at Lithuanian customs with VAT assessed again; the tax portion was refunded manually, which is a per-ticket patch rather than a fix.

Registering for IOSS is the alternative and would be the better buyer experience, but it needs a real registration plus parcel-label and invoice plumbing. Not collecting is what we can actually do correctly today, and it leaves destination customs to collect the import VAT normally — which is already what happens on the overwhelming majority of physical orders, where we collect nothing.

The same guard is applied to Australia, Singapore, Norway and the flag-gated countries. Those jurisdictions do permit vendor-side collection on low-value imported goods under registrations Gumroad holds, so the double-charge argument is EU-specific; excluding them is a deliberate scope decision rather than a legal constraint, and the code comment says so.

Tracking issue: antiwork/gumroad-private#1513

## Before / After

An EU buyer purchasing a $125 physical product with $40 shipping:

| | before | after |
|---|---|---|
| price | $125.00 | $125.00 |
| shipping | $40.00 | $40.00 |
| VAT at checkout (21% LT) | $17.85 | $0.00 |
| charged | $182.85 | $165.00 |
| import VAT at customs | assessed again on $182.85 | assessed once, normally |

No UI surface changes. Checkout, receipt and invoice all read from the same calculator, so the quoted amount, the charged amount and the receipt agree; the VAT-ID input disappears for these orders because `SalesTaxCalculation#has_vat_id_input?` keys off the now-nil `zip_tax_rate`. US and Canadian physical orders render and charge exactly as before.

## Test Results

Spec run and mutation evidence

`228 examples, 0 failures` across `sales_tax_calculator_spec.rb`, `purchase_sales_tax_spec.rb` and `customer_surcharge_controller_spec.rb`; RuboCop clean on both changed files.

Every new guard was proven load-bearing by reverting it individually and re-running the physical-product examples — all five mutants redden the suite, so none of the zero-tax assertions is vacuous:

| mutation | result |
|---|---|
| baseline (guards intact) | 0 failures |
| EU guard removed | 2 failures |
| AUS guard removed | 1 failure |
| SGP guard removed | 1 failure |
| NOR guard removed | 1 failure |
| flag-gated-country guard removed | 12 failures |
| restored | 0 failures |

Worth noting: the pre-existing example "returns zero tax if product is physical and in the EU" was **vacuous** — its `:zip_tax_rate` factory used the default `is_seller_responsible: 1`, so the lookup never found a rate and the example passed against the buggy code too. It now creates a real Gumroad-responsible rate, which is why it appears in the mutation table.

Duplicate system-spec pairs (two identical "applies GST for physical products" examples in both the Australia and Singapore blocks) were deduplicated rather than both inverted.

## QA steps

Backend-only; no rendered surface changes. On the preview app:

1. Visit any physical product with shipping to an EU country, add to cart, and enter an EU shipping address (e.g. Rome, IT 00186). The checkout total should show **no VAT line** and equal price + shipping.
2. Complete the purchase and check the receipt and generated invoice — no VAT amount, and the totals match what was charged.
3. Repeat with a US shipping address to a taxable state (e.g. Seattle, WA 98101): sales tax is still calculated and charged as before.
4. Repeat with any digital product and an EU address: VAT is still collected, and the VAT-ID field still appears.

## Open questions for the issue, not this PR

- Whether to refund the historical VAT on the 466 affected purchases.
- Whether help article 200 ("I need a VAT refund") should be extended to cover the physical/customs case; currently it only documents the business-VAT-ID path.

---

## AI disclosure

Written by claude-opus-5 (Anthropic) via Hermes Agent, from the diagnosis recorded on antiwork/gumroad-private#1513 and the ruling in that thread's discussion.

The model was asked to: implement option 2 from the issue (stop collecting VAT/GST on physical products outside the US), extend the guard to every affected jurisdiction rather than the EU alone, update the specs that asserted the old behavior, prove each guard load-bearing by mutation, and run a local adversarial pre-merge review. That review found no merge-blocking issues; it did correct two code comments that overstated the IOSS rationale for non-EU jurisdictions, and prompted the `Purchase#calculate_taxes` short-circuit.


gumclaw:
Closing this as superseded by #6593, which implements the same fix from the same tracking issue (gumroad-private#1513) but with the correct scope. Both PRs were opened by my automated runs three minutes apart — that's my duplication, not a real disagreement between two people, and leaving two competing patches on `SalesTaxCalculator#tax_eligible?` open would just split the review.

**Why #6593 is the one to keep, and not this one.** This PR removes Gumroad-collected tax on physical goods in *every* jurisdiction outside the US and Canada. That is wrong for four of them, and this PR's own body concedes it — it says the double-charge argument "is EU-specific" and that including Australia, Singapore, Norway and the flag-gated countries is "a deliberate scope decision rather than a legal constraint."

But it isn't a free scope decision. Gumroad holds a real local registration in each of those four: `GUMROAD_UK_VAT_REGISTRATION`, the Norwegian VOEC number, the Australian Business Number and the Singapore GST number. Their low-value-import regimes *require* the marketplace to collect at the point of sale, and the border does not charge again. So in those lanes there is no double charge to fix, and stopping collection would replace a buyer being over-charged with Gumroad under-collecting tax it is registered and obliged to collect. That trades a refundable buyer-experience bug for a tax liability, which is the worse of the two.

The UK matters specifically because it sits inside `EU_VAT_APPLICABLE_COUNTRY_CODES`, so an EU membership check catches it unless you carve it out explicitly. #6593 does exactly that.

**Nothing is lost.** #6593 carries the whole substance across: the same `!product.is_physical` reasoning, the same IOSS-vs-OSS analysis, the same measured exposure (466 purchases, ~$1,186, ~52 sellers over 12 months), the same before/after table on the Lithuania case, and mutation evidence that every guard is load-bearing. It also keeps the two open questions this PR raised for the issue rather than the PR — whether to refund the historical VAT, and whether help article 200 should cover the customs case.

One thing genuinely only in this PR: the `Purchase#calculate_taxes` short-circuit, so a physical non-US purchase skips the calculator instead of running it, getting zero back, and being left with `was_tax_excluded_from_price` set on a purchase carrying no tax. That's a real cleanup and I've flagged it on #6593 to be folded in there.

The branch is not deleted, so if the intent really was to stop collecting on physical goods everywhere regardless of the local registrations, reopening this is one click.

gumclaw:
Deleting the `fix/no-physical-vat-outside-us` branch. #6593 shipped the correctly-scoped fix (merged 2026-07-30), nothing is stacked on this branch, and leaving it green with a closed PR made it read every cycle as finished-but-unshipped work. Nothing here is owed a PR.

PR #6596, opened 2026-07-30
Reject Colombian ID numbers Stripe will refuse, with an actionable message

gumclaw:
## What

Stops the Colombian personal-ID field on Settings → Payments from accepting numbers Stripe will refuse.

Stripe enforces 7-10 digits on `individual.id_number` for Colombia. Our form advertised `minLength: 6, maxLength: 13`, so a seller could enter a value we already knew the processor would reject — and the rejection surfaced only as a merchant-account create that rolled back, leaving nothing they could act on.

- New `Compliance::ColombiaIdNumber` (`lib/utilities/compliance/colombia_id_number.rb`) and `app/javascript/utils/colombiaIdNumbers.ts` hold the bounds, the digit-count check, the normalizer, and the error copy — one rule per language rather than a copy per call site, pinned against each other by a spec.
- `UpdateUserComplianceInfo` gains `colombia_individual_id_error`, alongside the existing Peru and Singapore guards. This is the load-bearing one for the seller's own ID: the settings controller is the only writer of `individual_tax_id`, and browser-side length attributes are bypassable.
- `StripeBeneficialOwnersManager` gains the same guard on `create` and `update`, next to the existing JP kana checks. Beneficial owners are a separate writer that reaches Stripe without going through `UpdateUserComplianceInfo` at all.
- Colombian values are normalized to digits before storage and before being sent. The guard counts digits after stripping separators, but both `StripeMerchantAccountManager.person_hash` and `StripeBeneficialOwnersManager.build_person_params` sent the string verbatim — so `1.123.456.789` used to pass as ten digits and reach Stripe as thirteen characters. Storing and sending what was counted removes the assumption rather than relying on Stripe to strip separators. Peru's dash stays meaningful, so the branch is Colombia-only.
- `AccountDetailsSection.tsx` and `BeneficialOwnersSection.tsx` read the shared constants; `Settings/Payments/Show.tsx` and the beneficial-owner form both run the client-side check so the seller sees the error before a round trip.

The message names both documents rather than "your ID", so a foreign resident can tell it is about the document they hold, and it tells them not to zero-pad a short number.

## Why

A Colombian seller (foreign resident, 6-digit Cédula de Extranjería) made eight merchant-account attempts over three days. Every one was created and soft-deleted in the same second with `charge_processor_merchant_id` nil, and until #6476 shipped we discarded the Stripe error, so nothing on his account said which field was at fault. Once the breadcrumb landed, two fresh attempts both recorded `param=individual[id_number] — That ID number does not appear to be valid`.

Stripe support confirmed the 7-10 rule on case `sco_UyXAayCkurHzCd` and have raised it with their product team, because Colombia's own document spec (Anexo Técnico 1, Resolución 2011) permits 6-digit numbers — so their floor excludes IDs that legitimately exist. Until that moves, refusing the value up front with a message beats a silent rollback.

Two notes on scope, because this raises our floor from 6 to 7 and that deserves to be explicit:

- It does **not** unblock the seller who prompted it. Stripe separately state they cannot pay individual accounts verified with Cédula de Extranjería at all; that question is still open with them. This PR stops the next CE holder from burning eight attempts to discover the problem, and nothing more.
- It should not outlive Stripe's fix. The bound lives in one named constant with the reasoning attached, so widening it later is a one-line change.

Tracking issue: antiwork/gumroad-private#1429

## Before / After

Not applicable — no rendered surface changes. The diff adds a validation guard and swaps two hardcoded numbers for named constants; the field, its label, and its placeholder render identically. The user-visible change is an error message that appears on a value which previously produced a silent rollback, and that path needs live Stripe to reach.

## Test plan

Server-side, `spec/services/update_user_compliance_info_spec.rb` — **55 examples, 0 failures**, 10 of them new:

- 6-digit CE refused, no new `UserComplianceInfo` row, `StripeMerchantAccountManager` never called, existing row untouched
- 11-digit value refused
- 7-digit and dotted 10-digit values accepted and stored as entered
- a dotted value stored as digits only, so Stripe receives what the guard measured
- a value that is not a number at all refused, rather than reported as a successful save
- a zero-padded number pinned as **passing**, documenting that no guard can catch it and the message is the only defence
- another country's ID of the same length unaffected
- a masked resubmission still saves unrelated edits
- the browser's constants and message pinned against the server's

Browser-side, `app/javascript/utils/colombiaIdNumbers.test.ts` — 6 examples covering the accepted lengths, the 6-digit rejection, the upper bound, separator handling, a 13-character-but-6-digit value, and empty input.

Beneficial-owner path, `spec/business/payments/merchant_registration/stripe/stripe_beneficial_owners_manager_spec.rb` — 4 new examples: the 11-character/6-digit value asserting `create_person` is never reached, a bare 6-digit and an 11-digit refusal, the payload assertion that Stripe receives `1123456789` rather than `1.123.456.789`, and an in-range control.

Every guard is proven load-bearing, not just green:

| mutation | result |
|---|---|
| remove the `colombia_individual_id_error` call from `#process` | 3 examples fail |
| revert `COLOMBIA_ID_MIN_DIGITS` to 6 | 2 unit tests fail |
| drift the browser constant away from the server's | the parity example fails, naming the expected value |
| neuter the digits-only normalization for CO | the storage example fails, and only that one |
| remove the normalizes-away guard | the non-numeric example fails, and only that one |
| make the CO branch fire for every country | Peru's dash and Singapore's NRIC examples fail |
| remove `validate_colombia_id_number!` and the CO normalization from the beneficial-owner path | 3 of the 4 new examples fail; the in-range control stays green |

Local caveat, stated rather than rounded off: `spec/requests/settings/payments_spec.rb` has 6 failing examples in my worktree and `stripe_beneficial_owners_manager_spec.rb` has 1. I ran the same 7 example IDs on the untouched branch head and got an identical failing list, 7 for 7 — pre-existing local harness state (a leftover `MerchantAccount` row and Capybara `visit` timeouts), not this diff. CI is the authority; those shards were green on the previous head.

## Review

Two rounds of local adversarial review ran on the branch before this PR opened; the second returned READY-TO-MERGE.

Round 1's substantive finding was the separator issue: the guard counted digits while storage kept the separators, and my own spec asserted success with Stripe mocked — codifying exactly the kind of unverified processor assumption this PR exists to remove. Fixing it meant touching the shared `normalize_individual_tax_id`, which is the highest-blast-radius edit here, so round 2 reviewed that specifically. It confirmed Peru's dash and Singapore's alphanumeric NRIC survive (by mutation, not inspection) and that US sellers are byte-identical.

Round 2 then found a regression the separator fix had introduced: a typed value of only non-digits normalized to empty, read as "nothing changed", and returned success before any guard ran. Fixed in the last commit with a spec that fails if the guard is removed.

Round 3 came from Greptile, after the PR opened, and it found a real bug of the same class: `BeneficialOwnersSection` read the shared constants but nothing ran the digit check on that path, and `StripeBeneficialOwnersManager.build_person_params` forwarded `id_number` to Stripe with only whitespace stripped. So a Colombian business seller adding a beneficial owner could send `4.8.2.9.1.3` — 11 characters, inside the field's limits, six digits — and get the same rolled-back create with nothing actionable. I had updated that component's bounds two commits earlier and did not check whether it had a validator behind it. Fixed in `b703a3eaa` with the guard, the normalization, the browser check, and four examples; disposition in the thread.

Two findings I declined, deliberately. The length check accepts letters (`abc1234567` is seven digits) — Peru's guard has the identical hole, so tightening it here is a separate change rather than a quiet deviation from convention, and the digits-only normalization already stops letters reaching Stripe. And legacy Colombian rows stored with separators are not migrated: they self-heal when the seller next retypes the number, and sizing a one-off cleanup needs a production count I would rather take from a real query than guess at.

Two honest caveats:

- `spec/requests/settings/payments_spec.rb -e "CO creator"` times out on `visit` in my worktree — and does so identically on unmodified main (107s vs 109s), so it is environmental here, not this diff. CI covers it.
- I probed test-mode Stripe to try to re-derive the bounds and it **accepted every length**, including 6 and 11 digits. Test mode does not run the ID validator, so it cannot confirm or refute the rule; the 7-10 range rests on live-mode probes and Stripe's own confirmation. That caveat is now a comment in the helper so the next person does not repeat the probe and draw the wrong conclusion.

## QA steps

Colombia has no preview-app path that reaches the real validator (test-mode Stripe accepts anything), so this is the console check:

1. On a seller whose compliance country is Colombia, submit `individual_tax_id: "482913"` through `UpdateUserComplianceInfo`. Expect `success: false` and the 7-10 digit message, with no new `UserComplianceInfo` row.
2. Submit `"4829137"`. Expect `success: true` and the value stored verbatim.
3. Submit `"1.123.456.789"`. Expect `success: true` — separators are cosmetic, the digit count is the bound.
4. On a non-Colombian seller, submit a 6-digit ID and confirm it is unaffected.
5. On a Colombian **business** seller, call `StripeBeneficialOwnersManager.create` with `id_number: "4.8.2.9.1.3"`. Expect `InvalidFieldError` with the 7-10 digit message and no Stripe call. Repeat with `"1.123.456.789"` and expect Stripe to receive `1123456789`.

## Blast radius

Colombia only, and only on the input path. `submitted_tax_id_for` returns nil for blank and masked values, and `compliance_info_details` never sends the raw tax ID to the client — so an existing Colombian seller with a now-invalid stored number can still save unrelated settings, because the form does not echo the value back. `UserComplianceInfo`'s own completeness check tests `present?` only, so it is indifferent to digit count.

There are two writers that reach Stripe with a Colombian ID, not one. `grep -rn "individual_tax_id" app lib --include=*.rb` confirms the settings controller is the only writer of the *stored* `individual_tax_id` — but `StripeBeneficialOwnersManager` sends a caller-supplied `id_number` straight to `Stripe::Account.create_person`/`update_person` without storing it, so that grep never surfaced it. An earlier revision of this section claimed a single writer on the strength of that grep; Greptile found the second path, and both are now guarded. The lesson worth carrying: "who writes the column" and "who sends the field to the processor" are different questions, and the second is the one that matters for a processor-validation guard.

The guard keys off `effective_legal_entity_country_code`, so it also covers a Colombian business account's representative. That is correct rather than incidental: `StripeMerchantAccountManager.person_hash` sends the same `personal_tax_id` as `id_number` for any non-US legal entity, and its own comment records that Stripe validates the field against the account country. The representative branch of the beneficial-owner `update` is deliberately exempt — `build_representative_update_params` sends only relationship fields and never touches `id_number`.

Scope boundary worth stating: the same config map holds 18 other countries, and I have **not** verified their advertised bounds against Stripe — doing so needs a live-mode probe per country, since test mode does not run the validator. Colombia is the one with confirmed evidence (a support case and two recorded rejections), so it is the one this PR changes. Two things I did check across the map: no country's frontend floor is stricter than its server rule, and Peru's `minLength: 10` against a 9-digit server rule is the same character-versus-digit distinction handled here, not a bug.

---

## AI disclosure

Written by `claude-opus-5` (`model.default` in `~/.hermes/config.yaml`) running as Gumclaw.

Instructions given: a maintainer replied "1 and 2." on antiwork/gumroad-private#1429, selecting two items from a decision list I had put to him — take the Colombia ID validation gap to Stripe, and fix our own advertised bounds. Item 1 was already done (Stripe case `sco_UyXAayCkurHzCd`); this PR is item 2. No further direction was given on approach; the choice to enforce server-side beside the existing Peru and Singapore guards, to share the bounds through one helper, and to pin the two copies against each other was mine, as was the decision to state plainly that this does not unblock the originating seller.

### DIFF
--- a/app/business/payments/merchant_registration/implementations/stripe/stripe_beneficial_owners_manager.rb
+++ b/app/business/payments/merchant_registration/implementations/stripe/stripe_beneficial_owners_manager.rb
@@ -70,6 +70,7 @@ module StripeBeneficialOwnersManager
     stripe_account = ensure_eligible!(user)
     validate_required_fields!(params, action: :create, user: user)
     validate_jp_kana_address_format!(params, user)
+    validate_colombia_id_number!(params, user)
     person_params = build_person_params(params, user, action: :create)
     person = Stripe::Account.create_person(stripe_account.charge_processor_merchant_id, person_params)
     serialize(person)
@@ -84,6 +85,7 @@ module StripeBeneficialOwnersManager
     else
       validate_required_fields!(params, action: :update, user: user)
       validate_jp_kana_address_format!(params, user)
+      validate_colombia_id_number!(params, user)
       person_params = build_person_params(params, user, action: :update)
     end
 
@@ -213,6 +215,19 @@ module StripeBeneficialOwnersManager
     raise InvalidFieldError, "#{label} may only contain #{allowed_description}."
   end
   private_class_method :validate_kana_param!
+
+  # The form's maxLength counts characters so a pasted "1.123.456.789" fits, which means a value can
+  # satisfy the input and still carry too few digits. Checking digits here — and normalizing in
+  # build_person_params — keeps this path from handing Stripe a number it will refuse, which is the
+  # rolled-back-create failure that left one seller with eight silent attempts.
+  def self.validate_colombia_id_number!(params, user)
+    return unless user.alive_user_compliance_info&.legal_entity_country_code == Compliance::Countries::COL.alpha2
+    id_number = params[:id_number].to_s
+    return if id_number.strip.blank?
+    return if Compliance::ColombiaIdNumber.valid?(id_number)
+    raise InvalidFieldError, Compliance::ColombiaIdNumber::ERROR_MESSAGE
+  end
+  private_class_method :validate_colombia_id_number!
 
   def self.representative?(person)
     relationship = person.is_a?(Hash) ? person[:relationship] || person["relationship"] : person[:relationship]
@@ -356,6 +371,9 @@ module StripeBeneficialOwnersManager
     end
 
     id_number = params[:id_number].to_s.strip
+    # Send the digits the guard counted, not the separators the seller typed — Stripe refuses a CO
+    # id_number that is not digits-only.
+    id_number = Compliance::ColombiaIdNumber.normalize(id_number) if country_code == Compliance::Countries::COL.alpha2
     if id_number.present?
       if country_code == Compliance::Countries::USA.alpha2 && id_number.length == 4
         hash[:ssn_last_4] = id_number
--- a/app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/AccountDetailsSection.tsx
@@ -4,6 +4,7 @@ import * as React from "react";
 import typia from "typia";
 
 import type { ComplianceInfo, FormFieldName, User } from "$app/types/payments";
+import { COLOMBIA_ID_MAX_INPUT_LENGTH, COLOMBIA_ID_MIN_DIGITS } from "$app/utils/colombiaIdNumbers";
 import { countryRequiresPostalCode } from "$app/utils/postalCodes";
 
 import { Button } from "$app/components/Button";
@@ -209,13 +210,8 @@ const AccountDetailsSection = ({
         // resident reads "Cédula de Ciudadanía" and concludes their ID cannot be used.
         label: "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)",
         placeholder: "1234567890",
-        // Colombian ID numbers vary in length: Cédula de Extranjería numbers are commonly six or
-        // seven digits, citizen numbers eight to ten. A single exact length would block most of that
-        // range, so only a loose range is enforced here (the upper bound leaves room for a number
-        // pasted with thousands separators, e.g. "1.123.123.123"). Stripe does the authoritative
-        // validation on individual.id_number.
-        minLength: 6,
-        maxLength: 13,
+        minLength: COLOMBIA_ID_MIN_DIGITS,
+        maxLength: COLOMBIA_ID_MAX_INPUT_LENGTH,
         idSuffix: "colombia-id-number",
       },
       UY: {
--- a/app/javascript/components/Settings/PaymentsPage/BeneficialOwnersSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/BeneficialOwnersSection.tsx
@@ -1,7 +1,12 @@
 import * as React from "react";
 import typia from "typia";
 
-import { COLOMBIA_ID_MAX_INPUT_LENGTH, COLOMBIA_ID_MIN_DIGITS } from "$app/utils/colombiaIdNumbers";
+import {
+  COLOMBIA_ID_MAX_INPUT_LENGTH,
+  COLOMBIA_ID_MIN_DIGITS,
+  COLOMBIA_ID_NUMBER_ERROR_MESSAGE,
+  isValidColombiaIdNumber,
+} from "$app/utils/colombiaIdNumbers";
 import { countryRequiresPostalCode } from "$app/utils/postalCodes";
 import { request, ResponseError } from "$app/utils/request";
 
@@ -518,12 +523,25 @@ const BeneficialOwnersSection = ({
     return null;
   };
 
+  const validateColombiaIdNumber = (): string | null => {
+    if (defaultCountry !== "CO") return null;
+    if (editState?.mode === "edit" && editState.owner.relationship.representative) return null;
+    // An edit that leaves the stored ID untouched submits a blank field; only judge what was typed.
+    if (!formState.id_number.trim()) return null;
+    return isValidColombiaIdNumber(formState.id_number) ? null : COLOMBIA_ID_NUMBER_ERROR_MESSAGE;
+  };
+
   const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
     if (!editState) return;
     const kanaError = validateJpKanaFields();
     if (kanaError) {
       setFormError(kanaError);
+      return;
+    }
+    const colombiaIdError = validateColombiaIdNumber();
+    if (colombiaIdError) {
+      setFormError(colombiaIdError);
       return;
     }
     setIsSaving(true);
--- a/app/javascript/components/Settings/PaymentsPage/BeneficialOwnersSection.tsx
+++ b/app/javascript/components/Settings/PaymentsPage/BeneficialOwnersSection.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import typia from "typia";
 
+import { COLOMBIA_ID_MAX_INPUT_LENGTH, COLOMBIA_ID_MIN_DIGITS } from "$app/utils/colombiaIdNumbers";
 import { countryRequiresPostalCode } from "$app/utils/postalCodes";
 import { request, ResponseError } from "$app/utils/request";
 
@@ -217,13 +218,12 @@ const TAX_ID_CONFIGS: Record<string, TaxIdConfig> = {
   },
   CO: {
     // Colombia issues two personal IDs: the Cédula de Ciudadanía to citizens and the Cédula de
-    // Extranjería to foreign residents. Both are valid here, and their numbers range from about six
-    // to ten digits, so the field accepts a loose range rather than one exact length. Kept in sync
-    // with the same entry in AccountDetailsSection.tsx.
+    // Extranjería to foreign residents. Both are valid here. Bounds and copy are shared with
+    // AccountDetailsSection via the colombiaIdNumbers helper.
     label: "Cédula de Ciudadanía (CC) or Cédula de Extranjería (CE)",
     placeholder: "1234567890",
-    minLength: 6,
-    maxLength: 13,
+    minLength: COLOMBIA_ID_MIN_DIGITS,
+    maxLength: COLOMBIA_ID_MAX_INPUT_LENGTH,
     idSuffix: "colombia-id-number",
   },
   UY: {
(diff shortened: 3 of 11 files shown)


gumclaw:
@rmarescu — this is yours for review and merge. Payouts surface, so I have not armed auto-merge.

CI is green at `9d1fb7c6b` (70 test shards, 0 failures) and two rounds of local adversarial review are clean. What I would most like your eye on:

**The judgement call.** This raises our floor from 6 to 7 digits, so we now refuse a value we previously accepted. That means a Colombian foreign resident with a 6-digit Cédula de Extranjería gets told no by us instead of by a rolled-back Stripe create. I think that is right — the rollback left nothing actionable, which is how one seller reached eight silent attempts — but it is us enforcing a bound we believe is wrong on the merits. Stripe have raised the validator with their product team, so the constant is named and commented for easy widening.

**The bound is not locally verifiable.** Test-mode Stripe does not run the ID validator; I probed it and it accepted 6, 11, dotted, and lettered values. The 7-10 range rests on two live rejections recorded on the seller's account plus Stripe support case `sco_UyXAayCkurHzCd`. If you want that independently confirmed before merging, that is a reasonable thing to ask and I cannot do it from here.

**What I checked so you do not have to re-derive it:** the settings controller is the only writer of `individual_tax_id`; the presenter never sends the raw value to the client, so an existing seller with a now-invalid number can still save unrelated settings; and the guard covers the business path because `person_hash` sends the same value as `id_number` for any non-US legal entity, which the manager's own comment documents.

Worth knowing: this does not unblock the seller who prompted it. Stripe separately claim they cannot pay individual accounts verified with Cédula de Extranjería at all, and that question is still open with them on the same case. This PR stops the next CE holder discovering the problem through eight failed attempts.

gumclaw:
Greptile found a real bug, and it is the same bug class this PR exists to fix. Fixed in `b703a3eaa`.

**The finding, verified rather than taken on trust.** `BeneficialOwnersSection` read the shared `COLOMBIA_ID_MIN_DIGITS`/`COLOMBIA_ID_MAX_INPUT_LENGTH` constants but nothing ever ran the digit check on that path. Those attributes count *characters*, so `4.8.2.9.1.3` is 11 characters — inside the field's limits — and carries six digits. `StripeBeneficialOwnersManager.build_person_params` then did:

```ruby
id_number = params[:id_number].to_s.strip
# ...
hash[:id_number] = id_number
```

Whitespace stripped, nothing else. No digit guard, no separator handling, sent to `Stripe::Account.create_person` verbatim. A Colombian business seller adding a beneficial owner hit the identical rolled-back-create with nothing actionable — one conditional away from the path this PR fixes, on a form I had already touched to share the constants. I updated the two components' bounds and missed that only one of them had a validator behind it.

**What changed:**

- `validate_colombia_id_number!` on `create` and `update`, placed next to the existing `validate_jp_kana_address_format!` — the established server-side-counterpart-of-a-browser-check slot in this module.
- `build_person_params` normalizes CO to digits before sending, so Stripe receives the string the guard counted. Same normalize-at-the-write fix the earlier review round forced on the seller's own path; I had not carried it across.
- The browser runs the same check on submit, so the seller sees the message instead of a round trip.
- The bounds, normalizer, and copy now live in one Ruby module (`Compliance::ColombiaIdNumber`) instead of a second copy of the range in `UpdateUserComplianceInfo`. Three call sites, one rule.

**Evidence.** Four new examples in `stripe_beneficial_owners_manager_spec.rb`, including Greptile's exact `4.8.2.9.1.3` value asserting `create_person` is never reached. Proven load-bearing: with the guard and the normalization reverted, 3 of the 4 redden (the fourth is the in-range control, correctly unaffected). `update_user_compliance_info_spec.rb` is green at 55 examples.

Two things I checked rather than assumed: the representative branch of `update` skips this deliberately, because `build_representative_update_params` sends only relationship fields and never touches `id_number`; and a blank `id_number` on an edit passes, since that is how the form submits an unchanged stored value.

**Where Greptile was right to hedge.** Its T-Rex run could not observe the request payload before hitting its execution limit, so it posted the finding as a code-reading rather than a proof. The code-reading was correct. The spec that asserts the payload now exists.

On T-Rex's "corrected the fixture, then ran out of budget" note — the missing postal code was real; CO needs one. My specs supply it.

gumclaw:
@rmarescu — back to you, still no auto-merge.

Since your last look: Greptile found a real bug and it's fixed in `b703a3eaa`. The beneficial-owner form shared this PR's bounds constants but had no validator behind them, and the manager forwarded `id_number` to Stripe with only whitespace stripped — so a Colombian business seller could send an 11-character/6-digit value and hit the same silent rolled-back create. Full disposition is in the thread above.

That miss is worth knowing about before you review, because it says something about my original blast-radius claim. I wrote that the settings controller was the only writer, on the strength of `grep -rn "individual_tax_id"`. That grep was accurate and the conclusion was wrong: the beneficial-owner path never stores the value, so it holds no reference to the column and never appeared. "Who writes the column" and "who sends the field to the processor" are different questions. Both senders are guarded now, and the rule lives in one module (`Compliance::ColombiaIdNumber`) instead of a range copied per call site.

CI green at `b703a3eaa` — 70 test shards, 0 failures, 0 cancelled. Four new examples on the beneficial-owner path, 3 of 4 proven to redden with the guard reverted.

What I'd still value your eye on, unchanged from before: the judgement call of raising our floor from 6 to 7 digits, which means we now refuse a real Cédula de Extranjería that Stripe refuses too, and the fact that the 7-10 bound is not locally verifiable — test-mode Stripe doesn't run the validator, so it rests on two live rejections plus Stripe case `sco_UyXAayCkurHzCd`. If you want that independently confirmed before merging, that's reasonable and I can't do it from here.

gumclaw:
**`awaiting-human` on, `working` off.** The label said I was mid-build; I'm not. Nothing is left on my side of this PR, so `working` was misreporting who holds it.

@rmarescu — you are the requested reviewer and the assignee, and this is a Stripe/multi-currency compliance change, which is your surface. **What's expected of you: review it.** No review has been submitted yet, so it is sitting.

State as of now, so you're not re-deriving it:

- CI is fully green at `b703a3eaa` — 70 test shards, Buildkite, both lint jobs, `ci/green`. Mergeable, not a draft, no unresolved review threads.
- Greptile's round found a real second instance of the same bug class (`BeneficialOwnersSection` read the shared digit bounds but nothing ran the digit check on that path, so `4.8.2.9.1.3` — 11 characters, 6 digits — reached `Stripe::Account.create_person` verbatim). Fixed in `b703a3eaa`, with four new examples in `stripe_beneficial_owners_manager_spec.rb` including that exact value. Reverting the guard and the normalization reddens 3 of the 4, so they're load-bearing rather than decorative.
- The bounds, normalizer and copy now live in one module, `Compliance::ColombiaIdNumber`, with three call sites, instead of a duplicated range in `UpdateUserComplianceInfo`.

The judgement call worth your eyes rather than a rubber stamp: the `update` representative branch deliberately skips the validation because `build_representative_update_params` never sends `id_number`, and a blank `id_number` on edit passes because that's how the form submits an unchanged stored value. If you disagree with either, that's a changes-requested and it comes back to me.

I have not armed auto-merge — this touches what we send Stripe on account creation, so it should merge on your approval, not ahead of it.

PR #6608, opened 2026-07-30
Freeze the clock around the mobile subscription-products assertion

gumclaw:
## Why

`spec/controllers/api/mobile/purchases_controller_spec.rb:125` ("displays subscription products") started failing on branches that touch nothing under `api/mobile`. It is currently red on #6523 (`Test Fast 16`) and was red on #6526, on both attempts each — so not a rerun-able flake in the usual sense, but not those branches' code either.

The field-level diff is one field out of ~80: the membership product's `updated_at` is exactly one second behind the expectation, while `created_at` and `content_updated_at` match.

Cause: #6573 (merged to `main` at 05:19Z today) added `Price belongs_to :link, touch: true` and the variant-side equivalent. The spec renders the response, then rebuilds its expectation by calling `json_data_for_mobile` **again**. A price-row write between the two renders now moves `links.updated_at`, where before it could not. If the two reads straddle a second boundary, the expectation disagrees with the response.

Filed as #6607. Fixing it here rather than from either affected branch, because it belongs with the `touch:` that caused it.

## What

Tag the example `:freeze_time` so the whole example — setup included — runs under one clock. `spec/spec_helper.rb:400` already wraps `:freeze_time` examples in an `around` hook, which covers the `before` blocks too. No production code touched.

An earlier revision froze only the request and assertion; setup stayed on the real clock and the skew still reproduced.

I chose freezing over dropping `updated_at` from the comparison because the assertion's value is that it compares the *whole* mobile payload — narrowing it to dodge one field would quietly stop covering the rest of `json_data_for_mobile` in future.

## Verification

Ran locally in a worktree at this head with MySQL and Elasticsearch up:

- `spec/controllers/api/mobile/purchases_controller_spec.rb -e "displays subscription products"` — 3 consecutive runs, 0 failures
- the whole file — 64 examples, 0 failures
- `rubocop` on the file — no offenses

The failure is timing-dependent, so green runs are supporting evidence rather than proof; the argument is the mechanism (one clock across setup, request and assertion).

## Notes

Written by gumclaw (AI agent). Not armed for auto-merge — this is yours to merge.

### DIFF
--- a/spec/controllers/api/mobile/purchases_controller_spec.rb
+++ b/spec/controllers/api/mobile/purchases_controller_spec.rb
@@ -142,10 +142,15 @@ describe Api::Mobile::PurchasesController do
 
       # Both subscriptions should appear in the response
 
-      get :index, params: @params
-      expect(response.parsed_body).to include({ success: true,
-                                                products: [dead_subscription_purchase.json_data_for_mobile, subscription_purchase.json_data_for_mobile],
-                                                user_id: @purchaser.external_id }.as_json(api_scopes: ["mobile_api"]))
+      # The expectation re-renders json_data_for_mobile after the request, and the
+      # payload carries updated_at. Since prices touch their link (#6573), a write
+      # between the two renders can move it — freeze the clock so they agree.
+      freeze_time do
+        get :index, params: @params
+        expect(response.parsed_body).to include({ success: true,
+                                                 products: [dead_subscription_purchase.json_data_for_mobile, subscription_purchase.json_data_for_mobile],
+                                                 user_id: @purchaser.external_id }.as_json(api_scopes: ["mobile_api"]))
+      end
     end
 
     it "does not return unsuccessful purchases" do
--- a/spec/controllers/api/mobile/purchases_controller_spec.rb
+++ b/spec/controllers/api/mobile/purchases_controller_spec.rb
@@ -148,8 +148,8 @@ describe Api::Mobile::PurchasesController do
       freeze_time do
         get :index, params: @params
         expect(response.parsed_body).to include({ success: true,
-                                                 products: [dead_subscription_purchase.json_data_for_mobile, subscription_purchase.json_data_for_mobile],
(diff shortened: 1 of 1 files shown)


gumclaw:
Lint fix pushed as `e21b0c01` — `Lint Ruby` is green now and the rest of the suite is still running.

The failure was two `Layout/HashAlignment` offenses on the expectation hash I re-indented when wrapping it in `freeze_time`: adding the block moved the literal two columns right, and the continuation keys stayed at the old alignment. Whitespace only, no assertion changed. I re-ran the example locally after `rubocop -a --only Layout/HashAlignment` — `1 example, 0 failures` — so the frozen-clock fix itself is unaffected.

Worth flagging because it was blocking more than this PR: **this branch is the fix for the `Test Fast` failure hitting #6591 and #6523**, both of which are red on the same `api/mobile/purchases_controller_spec.rb:125` assertion from #6573's `Price belongs_to :link, touch: true`. A lint error on a two-line whitespace diff was holding up all three. Nothing for you to re-read here, @nyomanjyotisa — I have not armed auto-merge, this stays yours.

nyomanjyotisa (commented):
The current head still reproduces the one-second `updated_at` failure locally. Please apply `:freeze_time` to the entire example so setup is frozen too; that change passes locally.

gumclaw:
You're right — freezing only the request left setup on the real clock, so the skew could still land between the `before` blocks and the assertion. Pushed d48a43a: the example is now tagged `:freeze_time`, which `spec/spec_helper.rb:400` wraps in an `around` hook, so setup, request and assertion all read the same clock. The inner block is gone.

Verified locally at this head with MySQL and Elasticsearch up:

- the single example, 3 consecutive runs — 0 failures
- the whole file — 64 examples, 0 failures
- rubocop on the file — no offenses

PR body updated to match. Still yours to merge.

nyomanjyotisa (approved):

PR #6609, opened 2026-07-30
Port the price checker service spec to Minitest and fixtures

gumclaw:
## What

Item 17 of the #5801 ranking: `spec/services/price_checker_service_spec.rb` is deleted and replaced by `test/services/price_checker_service_test.rb` on Minitest with YAML fixtures.

All 15 RSpec examples port 1:1 — same cases, same order, every assertion preserved. `RealElasticsearchBridge` installs a real index for `Link` and `Purchase`, and the setup soft-deletes every other `Link` fixture so future fixture additions cannot silently move the match counts these distribution assertions depend on.

The port adds **no shared builders**. `create_product`, `create_subscription_product`, `create_bundle` and `create_physical_product` were already in `test/support/model_factories.rb` from #5901, and the three helpers this file needs (`create_matching_products`, `index_products`, `capture_search_calls`) are private to it. That is the first conversion in this migration where the shared fixture set covered the file outright.

Imported from an external contributor's fork — see the attribution note at the bottom.

## Why

Factory setup was 33% of this file's wall time, putting it at position 17 in the profiling ranking.

Two notes for whoever ports item 18.

The exclusion test is the one place a port of this file can quietly lose coverage. It builds six products that must each be excluded for a *different* reason — same seller, subscription, bundle, PWYW, EUR currency, physical — alongside 12 that must match, and asserts `match_count == 12`. If any single exclusion rule stopped working the count moves, so the assertion is load-bearing rather than a constant. I proved that rather than assuming it (see Test Results).

The caching test reproduces the full `price_checker:v3:<id>:<md5>` key including the nested description digest, so it still pins the real key format instead of asserting that *some* cache entry exists. Worth keeping that shape when porting the neighbouring services — an assertion that only checks presence would survive a fingerprint change that busts every seller's cache.

The caching test also counts searches **per call** rather than in total. A single uncached call already issues two searches (percentiles + histogram), so the original RSpec `at_least(:twice)` — and the port that faithfully replicated it as `>= 2` — was satisfied by the first call alone and stayed green with caching or `force_refresh` broken. It now asserts that the second call adds no searches and the `force_refresh` call adds some. Any port of a count-based expectation in the neighbouring service specs deserves the same check: compare the floor against what one call already does.

Refs #5801.

## Before / After

Not applicable — test-only change. Two paths in the diff, both under `spec/` and `test/`; no application code touched.

## Test Results

test run and mutation proof

```
$ CI=true RAILS_ENV=test bin/rails test test/services/price_checker_service_test.rb
15 runs, 42 assertions, 0 failures, 0 errors, 0 skips
```

Wall time on this machine, same worktree, back to back, with Elasticsearch up:

| | Examples | Wall time |
|---|---|---|
| RSpec original (at `origin/main`) | 15 | **2m17.1s** |
| Minitest port | 15 | **17.0s** |

The RSpec figure is this file's own run at `origin/main` in the same worktree, so the only difference between the two rows is the framework and the fixture set — not the host, not the warm state.

**The exclusion assertion is verified load-bearing, not merely green.** Reassigning the same-seller decoy to a different seller makes it a legitimate match, and the test fails with `Expected: 12, Actual: 13`. The mutation and its revert are in the image above; the working tree was clean before committing.

**The caching assertions are verified load-bearing too.** Mutating the single `Rails.cache.fetch` in `PriceCheckerService` fails the test both ways: `force: true` (caching never used) gives `Expected: 2, Actual: 4`, and `force: false` (`force_refresh` ignored) gives `Expected 2 to be > 2`. Reverted; tree clean.

`bundle exec rubocop test/services/price_checker_service_test.rb` → no offenses.

## QA steps

1. `CI=true RAILS_ENV=test bin/rails test test/services/price_checker_service_test.rb` → 15 runs, 42 assertions, 0 failures.
2. `git show origin/main:spec/services/price_checker_service_spec.rb` and read it beside the port — the 15 `it` blocks should map to the 15 `test` blocks in order, and the exclusion test should still carry all six decoys.
3. `CI=true RAILS_ENV=test bin/rails test` over the whole `test/` tree → 0 failures, skip count unchanged.

## Adversarial review

Reviewed locally against the branch before opening, aimed at the failure mode specific to this change class: a port that drops coverage while the suite stays green. No merge-blocking findings.

The review traced all 15 original examples to the 15 ported tests, confirmed the ported file contains no `create(:`/`build(:`/`FactoryBot` reference, confirmed the diff adds nothing to `model_factories.rb` so no existing builder is shadowed, and confirmed no fixture row is mutated in a way that leaks between tests (the setup's soft-delete is scoped to `Link` and the fixtures reload per test).

It flagged one thing worth stating out loud rather than fixing: the timing claim in the contributor's original report (~54%) was measured on their machine, and the numbers above are mine. Both are real; absolute wall times are host-specific and only the same-host pair belongs in the table.

---

Based on https://github.com/RealBhupesh/gumroad/tree/test/port-price-checker-service by @RealBhupesh, imported because `antiwork/gumroad` restricts pull-request creation to collaborators — there is no path from a fork branch to a PR here, so the contribution could not be submitted directly. The commit carries their `Co-authored-by` trailer.

I verified every claim in their report rather than taking it on trust: example parity, assertion count, RuboCop, test-only diff, zero factory usage, and the wall-time improvement re-measured independently. All held.

## AI disclosure

Written with Claude Opus 5. Instructions were to sweep open issues and PRs for work unblocked by a human reply, and to build qualifying items autonomously through to a PR. Here that meant reviewing an external contributor's completed Minitest port, verifying its claims independently, and importing it with attribution.


ershad:
Pushed 6e103d1f, which strengthens the caching test. Its `>= 2` search-count floor was satisfied by the first uncached call alone (percentiles + histogram = 2 searches), so it couldn't fail if caching or `force_refresh` broke — the same hole as the RSpec `at_least(:twice)` it faithfully ported. It now asserts per-call deltas: the cached call must add no searches, and the `force_refresh` call must add some.

Both mutations on the cache fetch confirm it's load-bearing: `force: true` fails with `Expected: 2, Actual: 4`, and `force: false` fails with `Expected 2 to be > 2`. Unmutated: 15 runs, 42 assertions, 0 failures. RuboCop clean.

---

Comment and fix written with Claude Fable 5.

gumclaw:
Confirmed on `6e103d1f`, and you're right about the hole: the `>= 2` floor was a faithful port of a bad expectation. `at_least(:twice)` looked like it was pinning "uncached, then cached, then forced", but two searches is exactly what one uncached call costs, so the assertion never had anything left to catch.

Reproduced both mutations on `Rails.cache.fetch` in `PriceCheckerService` locally:

- `force: true` → `Expected: 2, Actual: 4` at the cached-call assertion
- `force: false` → `Expected 2 to be > 2` at the `force_refresh` assertion

Unmutated: 15 runs, 42 assertions, 0 failures. RuboCop clean. Reverted the mutation; the branch is untouched by me.

I've updated the PR description — the test-results numbers said 40 assertions, and the "notes for whoever ports item 18" section now carries this as the lesson worth generalising: when porting a count-based expectation, check the floor against what a single call already costs, rather than transcribing the RSpec cardinality. That's the failure mode I should have caught during the port, since verifying an assertion is load-bearing is exactly what I claimed to have done for the exclusion test and didn't do here.

I checked the rest of the file for the same shape. The two remaining `>= 5` floors on `match_count` are different — they're the service's minimum-matches threshold, and each sits next to an `assert_equal` on the resulting tier (`with_taxonomy` / `broadened`), so the tier assertion is what pins the branch. No other count-based expectation in the file.

CI is running on your commit. Leaving the merge with you.

PR #6610, opened 2026-07-30
Serve the seller's branded share card on og:image, not just twitter:image

gumclaw:
## What

Both profile renderers now advertise the seller's generated `subscribe_preview` card on `og:image`, not only on `twitter:image`.

- `PageMeta::User#set_user_page_meta` sets `og:image` / `og:image:alt` alongside the `twitter:image` it already set — two lines inside the existing `subscribe_preview_url.present?` branch, nothing else in the method changed.
- `UsersController#profile_custom_html_wrapper_document` builds its own `<head>`, so it never saw that concern at all: it emitted `og:image` from `avatar_url` only and no `twitter:image` ever. It now follows the same chain — preview card first, uploaded avatar second, nothing when neither exists.

## Why

The two image paths were the exclusive arms of one `if`. The branded card was advertised only as `twitter:image`, and the branch that sets `og:image` — the avatar fallback — sat in the `else`, so it never ran for a seller who had a preview. Every non-Twitter scraper (Facebook, LinkedIn, WhatsApp, Slack, iMessage) read `og:image` and got Gumroad's global default from `PageMeta::Base#set_default_meta_tags`.

A seller with a generated card was therefore **strictly worse off than one without**, who at least got their avatar. On a custom-HTML profile it was worse still: that wrapper never emitted the card on any property, so those sellers had no branded unfurl anywhere, Twitter included.

Measured live before the change:

```
$ curl -A "facebookexternalhit/1.1" https://hellolittlememory.gumroad.com
<meta property="og:image"          content="https://assets.gumroad.com/images/opengraph_image.png">
<meta property="og:image:alt"      content="Gumroad">
<meta property="twitter:image"     content="https://public-files.gumroad.com/acj2d3goglswhpt13ut1k00e2tp8">
<meta property="twitter:image:alt" content="Hello Little Memory">
```

That `twitter:image` asset is a ready-to-use 1024x536 PNG carrying the seller's logo, store name and a Subscribe button — exactly the image they want Facebook to show.

This is a different bug from the two already fixed nearby: #5622 fixed `value=` vs `content=` rendering, and #5920 fixed og:image on custom-HTML **product** pages. Profile pages were never covered.

Ref antiwork/gumroad-private#1548

### Rejected alternatives

- **Hoist a `share_image = subscribe_preview_url.presence || avatar_url` above the `if`** (the shape the issue sketched). Equivalent output, but it also restructures the untouched avatar branch and the `og:title` override that lives beside it, widening the diff for no behavioural gain.
- **Also make `og:title` consistent between the two branches.** The `else` arm overrides the title to a bare `user.name` while the preview arm keeps `"Subscribe to X on Gumroad"`. That divergence looks incidental and is worth settling, but it is a separate user-visible change to link previews and does not belong in a two-line image fix. Left as-is deliberately; the issue tracks it.
- **Add `og:image:width` / `og:image:height`.** These appear nowhere in the codebase — not on product pages, posts, or the custom-HTML wrappers. The card is 1024x536 (≈1.91:1, the recommended OG ratio), which renders as a large card without hints. Adding them only here would be an inconsistent one-off.

## Before / After

No pixel diff exists to show: this changes `<head>` meta tags, not rendered pixels. The rendered value is the evidence, read out of the app's own response:

| Scraper reads | Before | After |
|---|---|---|
| `og:image` (Facebook, LinkedIn, WhatsApp, Slack, iMessage) | `/images/opengraph_image.png` — generic Gumroad | the seller's `subscribe_preview` card |
| `og:image:alt` | `Gumroad` | the seller's name |
| `twitter:image` | the seller's card (already correct) | unchanged |
| a seller with **no** preview | avatar on `og:image` | unchanged — fallback preserved |

The Before column is not asserted from reading the diff. It is the literal failure output produced by reverting the fix and re-running the new specs (see Test Results): `Expected "/images/opengraph_image.png" to eq "http://localhost:9000/gumroad-specs/..."` and `Expected "Gumroad" to eq "Seller"`.

Preview: <https://fix-profile-og-image-share-card.apps.staging.gumroad.org> — captured at the current
head `a36673df0`, serving `x-revision: a36673df0e4a`.

The change is `<head>`-only, so the surface that moved is **what a scraper reads, and therefore what
renders in a link unfurl**. Every frame below is a live scraper-user-agent response read out of a
running preview app, and the image inside each panel is the browser fetching **that response's own
`og:image`** — not a mock-up. Tags are parsed out of the **raw response body**, because a scraper
does not run JS (reading `meta[property]` from the hydrated DOM is wrong here: Inertia's client-side
head manager rewrites tags after hydration, and doing so silently produced a wrong "tag absent"
panel on the first attempt).

### All three asset states, on both profile renderers

Share-card meta tags across all asset states on both renderers

Six live reads. Top row is the Inertia path (`PageMeta::User`), bottom row the hand-built wrapper
(`UsersController#profile_custom_html_wrapper_document`) — confirmed per-panel by the
`inertia="meta-property-og-image"` attribute and the `#gumroad-landing-frame` element, so no panel
can be crediting the wrong renderer:

| Seller state | Inertia path | wrapper path | `og:image` served | `twitter:*` |
|---|---|---|---|---|
| generated card | `gumbogames` | `seller` | the seller's `subscribe_preview` card | `summary_large_image` + image + alt |
| uploaded avatar only | `gumbomusic` | `gumboaudio` | the uploaded avatar, alt `"<name>'s profile picture"` | none |
| neither asset | `gumbofilm` | `gumbowriting` | `opengraph_image.png` (absolute), alt `Gumroad` | none |

The two renderers now agree in **every** state, which is the claim round 3 was added to make true.
All six advertised assets were fetched and resolve: `200 image/png`, 960–29,980 bytes — a 404
`og:image` would be worse than the generic default, so this is checked rather than assumed.

### The round-3 fix: a seller with neither asset advertised Gumroad's grey default-avatar placeholder

Default-avatar placeholder before and after

Same seeded seller (`gumbofilm`: no `subscribe_preview`, no uploaded avatar) read off **two live
preview apps**. The BEFORE column is not a revert — it is the sibling preview for
`fix/dispute-evidence-backstop-sweep` (`x-revision be530054e29a`), verified to still carry the old
guard with `git show be530054e:app/controllers/concerns/page_meta/user.rb`. So the guard is the only
difference between the columns:

| | BEFORE (unfixed sibling) | AFTER (this branch) |
|---|---|---|
| `og:image` | `staging-assets.gumroad.com/images/gumroad-default-avatar-5.png` | `staging-assets.gumroad.com/images/opengraph_image.png` |
| `og:image:alt` | `Gumbo film's profile picture` | `Gumroad` |

`User#avatar_url` falls back to the default avatar, so the old `avatar_url.present?` guard was
**always** true: every seller without an upload advertised that grey placeholder as their share
image, with alt text claiming it was their profile picture. `avatar.attached?` lets the request fall
through to `PageMeta::Base`'s generic banner instead.

### Preview state left behind for the next QA session

These are QA fixtures, not part of this PR: `gumbomusic` and `gumboaudio` have an uploaded avatar
(their auto-generated `subscribe_preview` was purged so they exercise the avatar-only arm);
`gumbogames` has a generated card and no avatar; `gumbowriting` and `gumboaudio` have `custom_html`
seeded (via `Page#update_columns`, since `Ai::PageSanitizer.sanitize` returns `""` on the preview
image) plus the `custom_html_pages` flag activated, which is what routes them through the wrapper;
`seller` was already in that state from the earlier round.

### Earlier panels, still pinned at their original shas

These document superseded revisions and are kept for the audit trail. The `f2e6645f072b` panel and
the before/after pair were produced against revisions the preview no longer serves, and the
before/after pair used a *different* sibling preview that has since been reaped, so neither can be
regenerated from the current tree.

<details><summary>Superseded evidence panels</summary>

PR 6610 profile share-card meta tags re-verified at head f2e6645f0

og:image before and after on a seller profile

The fallback-control panel from before round 3
([`pr-6610-og-image-avatar-fallback.png`](https://raw.githubusercontent.com/antiwork/gumroad/920cd106698f41733c696eb5c4c9dd2d3b8c94ab/qa-media/pr-6610-og-image-avatar-fallback.png))
is **deliberately not shown inline any more**: it captured `gumbofilm` — a seller with *neither*
asset — receiving the default avatar, which round 3 changed on purpose. Reading it as a current
fallback control would now be reading the bug as the fix. The avatar-only arm it was meant to pin is
covered by `gumbomusic` / `gumboaudio` in the state matrix above.

</details>

## Test Results

Three specs added in `spec/controllers/users_controller_spec.rb` under `#show` → "share card meta tags". `render_views` is on for this file, so they parse the real rendered `<head>` rather than stubbing the presenter.

Spec run, then mutation-proofed by reverting the fix

Walkthrough video (same story, narrated over the diff): [`qa-media/pr-6610-og-image-share-card-walkthrough.mp4`](https://raw.githubusercontent.com/antiwork/gumroad/920cd106698f41733c696eb5c4c9dd2d3b8c94ab/qa-media/pr-6610-og-image-share-card-walkthrough.mp4)

**The specs are load-bearing, proven rather than asserted.** Reverting `page_meta/user.rb` to `origin/main` and re-running reddens 2 of the 3 original examples; a fourth example (added in review round 2) pins the card ahead of an uploaded avatar, and switching `og:image` to `avatar_url` reddens that one plus the first:

```
4 examples, 0 failures        # with the fix
4 examples, 2 failures        # og:image switched to avatar_url  (:410, :430)
```

The third — the avatar fallback — correctly stays green, because the `else` branch is untouched. A revert reddening all three would have meant the fallback spec was pinning something the fix moved, which it should not be.

### Custom-HTML profiles

Four request specs added in `spec/requests/profile_custom_html_spec.rb` under "share card meta tags" — they hit the real wrapper on a verified custom domain, so they read the `<head>` the wrapper actually emits.

Custom-HTML profile og:image before and after, with the spec run

**Re-verified at head `394fe5950`.** That commit changed this exact method — `subscribe_preview_url`
is now resolved once into a local instead of read three times — so the panel above documents a
superseded revision of it. The wrapper `<head>` was read again from the running preview at head, with
the seeded seller (`subscribe_preview` attached, `avatar.attached? == false`) put onto the custom-HTML
takeover path so the wrapper, not the Inertia profile, serves the page:

Custom-HTML profile head re-verified at head 394fe5950

```
$ curl -sI  https://seller.fix-profile-og-image-share-card.apps.staging.gumroad.org/ | grep x-revision
x-revision: 394fe5950499

$ curl -s -A "facebookexternalhit/1.1" https://seller.fix-profile-og-image-share-card.apps.staging.gumroad.org/ \
    | grep -oE '<meta property="(og|twitter)[^>]*>'
<meta property="og:title"        content="Seller">
<meta property="og:type"         content="profile">
<meta property="og:url"          content="https://seller.fix-profile-og-image-share-card.apps.staging.gumroad.org/">
<meta property="og:image"        content="https://staging-public-files.gumroad.com/oky47wxydpugco594ioec1a3c18b">
<meta property="og:image:alt"    content="Seller">
<meta property="twitter:card"    content="summary_large_image">
<meta property="twitter:image"   content="https://staging-public-files.gumroad.com/oky47wxydpugco594ioec1a3c18b">
<meta property="twitter:image:alt" content="Seller">
```

The `<iframe sandbox>` attribute reads `allow-scripts allow-forms allow-popups
allow-popups-to-escape-sandbox` on the same response — unchanged, still no `allow-same-origin` and no
`allow-top-navigation`, so the added meta tags did not disturb the wrapper's security posture.

Preview state left behind for the next session: this seller now has `custom_html` seeded (150 chars,
written with `update_columns` because `Ai::PageSanitizer.sanitize` returns `""` in the preview image)
and the `custom_html_pages` feature flag activated. Both are QA fixtures, not part of this PR.

### Round 3: the two surfaces now agree in every state (was a real bug, found by review)

The live check above caught a third defect, and an earlier revision of this description wrongly
argued it was out of scope. Reading a seller with **no** preview off the same preview app shows it.
`gumbofilm` goes through the standard Inertia path (`PageMeta::User`), whose `else` branch guarded on
`avatar_url.present?` — **always** true, because `User#avatar_url` falls back to the default avatar:

Live scraper check across both profile paths

```
$ curl -s -A "facebookexternalhit/1.1" https://gumbofilm.<preview-host>/ | grep -oE '<meta property="og:image[^>]*>'
<meta property="og:image"     content="https://staging-assets.gumroad.com/images/gumroad-default-avatar-5.png" inertia="…">
<meta property="og:image:alt" content="Gumbo film&#39;s profile picture" inertia="…">
```

So a seller with neither asset advertised Gumroad's **grey default-avatar placeholder** as their
share image. I first recorded this as "pre-existing on `main`, left alone rather than widened into
this diff." That was the wrong call: this PR's whole premise is that a seller's profile should share
a sensible image, and this is the state where it shared the worst possible one. Greptile filed it as
a P1 against the same lines, which is what prompted the re-decision.

**Fixed in `a36673df0`.** `PageMeta::User` now guards on `avatar.attached?`, so only a real upload is
advertised and the no-assets case falls through to `PageMeta::Base`'s generic Gumroad banner.

Verifying that also corrected a claim in my own earlier description: omitting the tag does **not**
leave scrapers to their own fallback, because `PageMeta::Base` already sets
`og:image` to `opengraph_image.png`. The custom-HTML wrapper, however, builds its `<head>` by hand and
inherits none of those defaults — so it was emitting *no* `og:image` while the standard path emitted
the banner. The wrapper now spells the same fallback out explicitly (absolute URL, since a relative
path would resolve against the seller's own custom domain, which does not serve Gumroad's assets).
The two surfaces now match in all four states: preview → card; avatar only → avatar; neither →
generic banner; and `twitter:*` on the preview branch only.

Before, the wrapper advertised the seller's **avatar** if they had uploaded one, and **nothing at all** if they hadn't — never the card. The dump above is the second case (the seeded seller has `avatar.attached? == false`, so the old `avatar_url`-only ternary produced an empty string):

```
<meta property="og:title" content="Jane Doe">
<meta property="og:type" content="profile">
<meta property="og:url" content="http://seller.example.com/">
```

After, the same request carries the card on both properties:

```
<meta property="og:image" content=".../gumroad-specs/76kpyfs8…">
<meta property="og:image:alt" content="Jane Doe">
<meta property="twitter:card" content="summary_large_image">
<meta property="twitter:image" content=".../gumroad-specs/76kpyfs8…">
<meta property="twitter:image:alt" content="Jane Doe">
```

Load-bearing, per-example rather than by reverting the whole file, because the file-level revert leaves the interesting invariant untested:

| Mutation | Result |
|---|---|
| `users_controller.rb` reverted to the previous commit | `28 examples, 2 failures` (`:61` preview card, `:77` avatar fallback) |
| chain inverted to `avatar || preview` | `29 examples, 1 failure` — only `:95`, the preview-beats-avatar example |
| `page_meta/user.rb` og:image switched to `avatar_url` | `4 examples, 2 failures` (`:410`, `:430`) |

The "neither preview nor avatar" example changed meaning in round 3: it used to assert *no image
tags* and now asserts the **generic Gumroad banner** on both surfaces. It is the example that pins
the round-3 fix, proven load-bearing by restoring the old `avatar_url.present?` guard:

| Mutation | Result |
|---|---|
| `page_meta/user.rb` guard restored to `avatar_url.present?` | `5 examples, 1 failure` — only the new neither-asset example (`expected "/images/opengraph_image.png"`, got the default avatar) |

Full run at `a36673df0`, both spec files together: **`112 examples, 0 failures`**
(`bin/rspec spec/requests/profile_custom_html_spec.rb spec/controllers/users_controller_spec.rb`,
`CI=true`). Rubocop clean on all four files.

One caveat, scoped precisely: `profile_custom_html_spec.rb:95` fails when run **in isolation**
(`bin/rspec spec/requests/profile_custom_html_spec.rb:95`) because it depends on setup from earlier
examples in the file. That is **pre-existing** — it fails identically with my diff stashed — and the
file is green as a whole (29/29) both before and after.

Round 3 (the `avatar.attached?` fix and the wrapper's explicit banner fallback) landed after the
adversarial review recorded below, so that pass does not cover it.

## Adversarial review

**Round 1** (fable-5, local, before the PR was opened): **SHIP**, no P1 or P2. Items it settled with evidence rather than reasoning:

- **No duplicate `og:image` tag.** `InertiaRails::MetaTag` derives `head_key` from the property (`og:image` → `meta-property-og-image`, `allow_duplicates: false`), and both stores are hashes keyed by it — `PageMeta::Base#set_meta_tag` (`meta_tags[new_tag[:head_key]] = new_tag`) and the gem's `MetaTagBuilder#add_new_tag`. The default is replaced, not appended. The layout renders exactly one of the two mechanisms.
- **All four callers of `set_user_page_meta` are safe.** `show`, `subscribe`, `coffee` and `subscribe_preview`: the only meta set after the call in any of them is `title:` (a different `head_key`), so nothing clobbers the new tag and it clobbers nothing.
- **The URL is durable.** Production ActiveStorage is `amazon` with `public: true`, so `blob.url` carries no expiring signature; `cdn_url_for` rewrites it to `https://public-files.gumroad.com/<key>` — byte-identical to the URL scrapers have been fetching off `twitter:image` for years. An expiring URL on `og:image` would be worse than the default, because scraper caches pin it.
- **The specs are not vacuous.** `meta_content` returns `""` for a missing tag, but every assertion compares against a value proven non-empty in the same example (`expect(preview_seller.subscribe_preview_url).to be_present`).

Four P3s from that review. The first has since been **fixed here**; the rest are deliberately not:

1. ~~**Custom-HTML profiles still never show the card.**~~ **Fixed** in `profile_custom_html_wrapper_document`, which now runs the same chain and adds the Twitter card tags when a preview exists.
2. **The coffee page's share image is now the profile card.** Before this PR it was the *generic Gumroad* image, so this is strictly better, but it is a behaviour change and is called out here rather than left to be discovered.
3. `to_not include("opengraph_image")` is redundant with the `eq` above it. Kept as documentation of intent.
4. No `og:image:width`/`height` anywhere in the codebase — a global question, not this PR's.

**Round 2** (fable-5, local, over the custom-HTML wrapper delta): **SHIP WITH FIXES**. Both fixes are in:

- **P2 — precedence was untested, and a wrong implementation passed everything.** Every example attached either a preview or an avatar, never both, so inverting the chain to `avatar || preview` — the exact regression that reproduces the reported bug for any seller with both, which is the common case — kept all 28 examples green. Fixed by adding that example on both surfaces, and proven by running the inversion: `29 examples, 1 failure`, the new one. The avatar example also now asserts no stray `twitter:card`.
- **P3 — `subscribe_preview_url` was read three times.** It rescues to nil internally, so a transient miss on the first read paired with a hit on a later one would have emitted the *avatar* wearing the preview branch's `twitter:card`/`twitter:image`. Resolved once into a local.

It also confirmed, with file:line evidence: escaping is single-pass and correct for a name containing `"`, `'` or `</head>` (`title` is `ERB::Util.h`'d once, the URL separately, and `cdn_url_for` never returns an html_safe string); alt-text matches `PageMeta::User` exactly on both branches; `layout: false` on the sole caller means no second `<head>` and so no duplicate tag; and the divergence from the product wrapper (which emits `twitter:card` for any image) is correct, because each wrapper mirrors its own canonical path and a square avatar would render badly as `summary_large_image`.

One P3 declined for scope, and named below: `UserPagesController#page_meta_head` is a third `<head>` builder with the same gap.

## QA steps

Meta tags are only observable to a scraper, so the useful check is a scraper user-agent against the preview app rather than a click-path.

1. Open the preview app for this branch and pick any seller with a generated card (the profile page will show it on Twitter today).
2. `curl -A "facebookexternalhit/1.1" https://<preview-host>/<username> | grep -E 'og:image|twitter:image'`
3. Expect `og:image` to equal `twitter:image` and both to point at `public-files.gumroad.com/...`, with neither being `opengraph_image.png`.
4. Repeat against a seller with **no** preview: `og:image` should be their avatar, `og:image:alt` should read `"<name>'s profile picture"`, and no `twitter:image` should be present.
5. Facebook's Sharing Debugger and LinkedIn's Post Inspector against the preview URL both render the card, if you want the visual confirmation.
6. For a seller on a **custom HTML** profile (`custom_html_pages` flag on, `custom_html` set), `curl -A "facebookexternalhit/1.1"` the profile root and expect the same pair of properties — before this PR that page showed the seller's avatar at best, and nothing at all if they had not uploaded one.

## Not in this PR

- `og:title` consistency between the two branches (see Rejected alternatives).
- Image dimension tags (round-1 P3 #4).
- **Slugged pages (`UserPagesController#page_meta_head`) are the remaining surface of this bug class.** That builder emits `og:title` / `og:type` / `og:url` and no image tag at all, for both the custom-HTML page wrapper and rich-text pages. Same root cause, third builder; left out to keep this diff to the two profile surfaces the issue is about.

---

AI-assisted: written by **claude-opus-5** via Hermes Agent, dispatched autonomously from antiwork/gumroad-private#1548 by the `autonomous-dev-dispatcher` cron. Instructions given: sweep both repos for buildable issues, plan and build the qualifying ones through PR, prove any new specs load-bearing by reverting the fix, run a local fable-5 adversarial review before opening the PR, and hand off rather than auto-merge if the change is not clearly safe.


gumclaw:
@gianfrancopiana — yours for review and merge. This is UI-adjacent (it changes what every seller's profile looks like when shared anywhere), so auto-merge is deliberately not armed.

**What changed.** Two lines in `PageMeta::User#set_user_page_meta` put the seller's generated `subscribe_preview` card on `og:image` alongside the `twitter:image` it already set. The custom-HTML profile wrapper in `UsersController` builds its own `<head>` and never saw that concern, so it now follows the same chain.

The seller-visible effect is that Facebook, LinkedIn, WhatsApp, Slack and iMessage stop showing Gumroad's generic default and start showing the seller's own card. A seller who had a generated card was previously *worse off* than one without, since the avatar fallback sat in the `else` arm and never ran for them.

**Greptile filed a P1 that was correct, and I had wrongly scoped it out.** `User#avatar_url` falls back to Gumroad's default avatar, so the standard profile's `avatar_url.present?` guard was always true: a seller with neither a card nor an uploaded avatar was advertising the grey placeholder as their share image. An earlier revision of my description called that pre-existing and out of scope. That was the wrong call — this PR exists so profiles share a sensible image, and that was the state where the image was worst. Fixed in `a36673df0` by guarding on `avatar.attached?`.

Chasing it also corrected one of my own claims: omitting the tag does *not* leave scrapers to their own fallback, because `PageMeta::Base` already sets `og:image` to `opengraph_image.png`. The hand-built custom-HTML wrapper inherits none of those defaults, so it was emitting no `og:image` at all while the standard path emitted the banner — the two surfaces disagreeing in exactly the state the P1 was about. The wrapper now spells the same fallback out (absolute, since a relative path would resolve against the seller's own custom domain). Both surfaces now match in all four states: preview → card, avatar only → avatar, neither → generic banner, `twitter:*` on the preview branch only.

**Verification at `a36673df0`.** `112 examples, 0 failures` across `spec/requests/profile_custom_html_spec.rb` and `spec/controllers/users_controller_spec.rb` under `CI=true`; rubocop clean. The fix is proven load-bearing: restoring the old `avatar_url.present?` guard reddens exactly the new neither-asset example and nothing else. The spec that previously asserted "no image tags" was pinning the defect, so it is inverted rather than deleted — worth knowing when reading the diff, since the example count rising hides that one assertion flipped meaning.

**CI green at `a36673df0`:** 74 jobs pass, 0 failures, and `gh pr checks` is clean. Merge state CLEAN.

One thing for you to weigh rather than take from me: the fable-5 adversarial pass predates this round-3 fix, so it does not cover it. Also worth knowing, `profile_custom_html_spec.rb:95` fails when run in isolation because it leans on setup from earlier examples in the file — pre-existing, identical with my diff stashed, and the file is 29/29 green as a whole.

Greptile has now re-reviewed at `a36673df0` and returned **Confidence 5/5 with zero findings**, up from 3/5 with the "do not merge" P1. Worth saying plainly why that is not circular: an earlier pass had already dropped the P1 while its "last reviewed commit" was still `5342bddec`, one commit *before* the fix — so at that point the finding had vanished from a run that never saw the fix, which is not the same as being answered. This latest pass names the fix commit and describes the intended behaviour including the absolute generic fallback, so it is a genuine re-review.

Full before/after evidence, rejected alternatives, and the scraper `curl` QA steps are in the description. One P3 is declined for scope and named there: `UserPagesController#page_meta_head` is a third `<head>` builder with the same gap, on slugged pages.

PR #6624, opened 2026-07-30
Stop charging installment plans whose every installment was disputed

gumclaw:
## What

`RecurringChargeWorker` no longer charges an installment plan whose **every** prior installment carries a standing (non-reversed) chargeback.

Adds `Subscription#all_charges_disputed?` and a guard in the worker, placed after `charges_completed?` and before `in_free_trial?`.

## Why

antiwork/gumroad#6568 fixed the chargeback-cancellation gate but **deliberately excluded installment plans** from all three paths, and that exclusion was correct: `Subscription#installment_plans_cannot_be_cancelled_by_buyer` rejects `cancel_effective_immediately!(by_buyer: true)`, and the resulting `RecordInvalid` would raise *after* the seller-balance decrement and *before* dispute-evidence creation — a worse bug than the one being fixed.

The consequence is that a disputed installment plan has no exit at all. It stays `alive?`, keeps its place in the charge schedule, and gets charged again. That population is not hypothetical: **six live installment plans currently carry a standing chargeback, and two still have installments outstanding.** One is scheduled to charge a buyer on **27 August** who has disputed every prior installment. Details and record identifiers in antiwork/gumroad-private#1551.

### Why guard the charge instead of cancelling

The reason these must stop is that *we should not charge a card whose holder has already disputed every prior charge on it*. That is a charging concern, not a cancellation one. Cancelling would additionally have to invent an actor — and `cancel_effective_immediately!` with `by_buyer: false` sends the buyer `subscription_product_deleted`, an email about a product deletion that did not happen.

Leaving the plan alive but unchargeable also keeps the record honest about why it stopped, rather than backdating a cancellation nobody performed.

### Two deliberate narrowings

- **Every installment must be disputed, not just one.** A single disputed installment on an otherwise-paid plan can be a reversible mistake; blocking those would strand plans whose buyer still intends to pay. Of the six live plans, three have only one of two installments disputed — those are intentionally unaffected.
- **Reversed chargebacks do not count.** Uses `chargedback_not_reversed?`, so a won dispute lets the plan resume on the next tick with no manual re-enablement.

The guard `return`s rather than calling `unsubscribe_and_fail!` for the same reason — failing the subscription would be a state change that a reversal could not undo by itself.

## Scope

This changes behaviour for **future** charge attempts only. The six plans already in this state are not modified by this PR; whether to remediate them (in particular the 27 August one) is an open decision on gumroad-private#1551. This PR removes the ongoing bleed so that decision is not also a deadline.

Recurring (non-installment) subscriptions are untouched — they have a working cancellation path via #6568, and `all_charges_disputed?` returns `false` for them. There is a spec asserting exactly that, so this cannot silently grow into freezing memberships that should instead be cancelled.

## Testing

```
bundle exec rspec spec/sidekiq/recurring_charge_worker_spec.rb
38 examples, 0 failures

bundle exec rails test test/models/subscription_test.rb
434 runs, 1017 assertions, 0 failures, 0 errors, 0 skips
```

Four new examples: all-disputed blocks the charge; partially-disputed still charges; reversed chargebacks resume charging; recurring subscriptions are unaffected.

**Mutation-checked** — the specs are not vacuous. Disabling the guard (`if false && subscription.all_charges_disputed?`) reddens the primary example:

```
rspec ./spec/sidekiq/recurring_charge_worker_spec.rb:451
  # does not charge when every installment is charged back
1 example, 1 failure
```

## Before / After

Not applicable — no user-facing surface. The behaviour change is a skipped Sidekiq charge, evidenced by the specs above and this log line:

```
RecurringChargeWorker#perform(<id>): skipping charge, installment plan has a standing chargeback on every installment
```

Refs antiwork/gumroad-private#1551. Follows antiwork/gumroad#6568.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -889,6 +889,31 @@ class Subscription < ApplicationRecord
     has_fixed_length? ? charge_occurrence_count - successful_purchases.count : 0
   end
 
+  # Installment plans have no cancellation exit after a chargeback: the
+  # `installment_plans_cannot_be_cancelled_by_buyer` validation rejects
+  # `cancel_effective_immediately!(by_buyer: true)`, and the resulting RecordInvalid would raise
+  # after the seller-balance decrement and before dispute-evidence creation. So the plan stays
+  # alive and keeps its place in the charge schedule.
+  #
+  # Stopping the charge rather than cancelling the plan is deliberate. The reason to stop is that
+  # we should not charge a card whose holder disputed every prior charge on it — a charging
+  # concern, not a cancellation one. Cancelling would also have to invent an actor, and
+  # `cancel_effective_immediately!` with `by_buyer: false` emails the buyer about a product
+  # deletion that never happened.
+  #
+  # Every installment must be disputed, not just one: a single disputed installment on an
+  # otherwise-paid plan can be a reversible mistake, and blocking those would strand plans whose
+  # buyer still intends to pay. Reversed chargebacks do not count, so a won dispute lets the plan
+  # resume on its own.
+  def all_charges_disputed?
+    return false unless is_installment_plan?
+
+    charges = successful_purchases.to_a
+    return false if charges.empty?
+
+    charges.all?(&:chargedback_not_reversed?)
+  end
+
   # Certain events should transition the subscription from pending cancellation to cancelled thus not allowing the customer access to updates.
   def cancel_immediately_if_pending_cancellation!
     with_lock do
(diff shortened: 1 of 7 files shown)


gumclaw:
CI is green on `66e5e9fc6` — 76 passed, 10 skipped, nothing failing. Ready for you now, @rmarescu. Auto-merge is not armed: this decides whether a card gets charged, so it merges on your call.

Earlier this PR was labelled `awaiting-human` six seconds after it opened, before a single check had reported. That was a bug in my ownership auditor, not a judgement call — it suppressed its own pending-CI gate whenever a human was assigned, and it is the thing that assigns the human, so it disabled its own check inside one pass. Now keyed on whether a human has actually engaged. The label is accurate this time. Same premature flip hit #6623 in the same run.

PR #6567, opened 2026-07-29
Fix 404 on the license-key lookup button from seller domains

gumclaw:
## What

The "Already bought this? → View your information" button on a licensed product page linked to a relative `/license-key-lookup` path. That route is only drawn inside `GumroadDomainConstraint` (`config/routes.rb:1205`, inside the block opened at `:468`), so it does not exist on a seller subdomain or custom domain. Since product pages are normally served from the seller's own host, the button 404'd for essentially every buyer who saw it.

Reported by a seller via support, who diagnosed it correctly themselves: the button should point at the global lookup page on the root domain. A second, independent report of the same 404 arrived from a different seller four days after the relative href shipped in #6321. Reporter details are in the private tracking issue.

## Why

This button is the only self-serve path we offer a buyer we cannot identify as a past purchaser. Landing them on a 404 sends them to the seller's inbox instead, which is exactly what the button was added to avoid.

The mechanism is general to the `GumroadDomainConstraint` block rather than specific to this path, so every licensed product page on a seller subdomain or custom domain is affected for any visitor we cannot identify as a past buyer — the default case for someone arriving from a shared store link.

Tracking issue: antiwork/gumroad-private#1565

After: the button href is an absolute root-domain URL built from `useDomains()`, so it resolves to `https://gumroad.com/license-key-lookup` no matter which host the product page is served from. Same pattern already used by `PoweredByFooter` and `EmailAction/Layout` for root-domain links out of subdomain-rendered pages.

## Before / After — the old code side by side with the new

The preview captures further down show the **fixed** behaviour. This section adds the missing half: the **same flow running the pre-fix href**, so the regression itself is on record rather than inferred.

Produced by booting this branch as a live local app and driving a real logged-out Chromium at a seller subdomain (`seller.lvh.me:3001`), then toggling **only** the `LicenseKeyLookupPrompt` block in place between the two runs — nothing else differs. Every banner is generated from the live DOM and the real HTTP response (`link.getAttribute("href")`, `response.status()`), never typed in.

https://github.com/antiwork/gumroad/raw/cd02012e15c0b6978d994d495a34b28ff7336673/qa-media/pr-6567-before-after.mp4

| | Before | After |
|---|---|---|
| Button `href` | `/license-key-lookup` | `http://lvh.me:3001/license-key-lookup` |
| Click from `seller.lvh.me:3001` | **HTTP 404** — Rails `Routing Error / Not Found` | **HTTP 200** — lookup form renders |

[...]

PR #6630, opened 2026-07-30
Widen CardData.zipCode to fix the typecheck failure on main

gianfrancopiana:
## What

Widens `CardData.zipCode` in `app/javascript/data/card_payment_method_data.ts` to `string | null | undefined`, fixing the one error `npm run typecheck` reports on `main`.

## Why

The AVS spec added in #6594 passes `zipCode: undefined` on purpose, to verify that an undefined ZIP reaches Stripe as `null`. The project compiles with `exactOptionalPropertyTypes`, and under that flag the declaration `zipCode?: string | null` rejects an explicit `undefined`, so `tsc` fails on the spec file.

At runtime the function already coerces the value with `|| null`, so the widened union matches the actual contract, and it follows the pattern other optional fields use (`buyerCurrency` in `utils/currency.ts`, `mountCurrency` in `PaymentElementInput.tsx`). The one production caller is unaffected.

With #6626 gating CI on `npm run typecheck`, every branch cut from `main` inherits this error and fails the new job until this lands.

## Before/After

Type-level change only; app behavior is unchanged.

Before, on `main`:

```
Argument of type '{ cardElement: { token: string; }; email: string;
zipCode: string | null | undefined; }' is not assignable to parameter of
type 'CardData' with 'exactOptionalPropertyTypes: true'.
```

After: `npm run typecheck` completes with zero errors.

## Test Results

```
$ npm run typecheck
> tsc --noEmit
(zero errors)

$ npm run test -- app/javascript/data/card_payment_method_data.test.ts
Test Files  1 passed (1)
     Tests  22 passed (22)
```

`eslint` on the changed file is clean.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Fable 5.

### DIFF
--- a/app/javascript/data/card_payment_method_data.ts
+++ b/app/javascript/data/card_payment_method_data.ts
@@ -25,7 +25,7 @@ type ReusableCCVariation<CardParams extends CardPaymentMethodParams | PaymentReq
 type CardData = {
   cardElement: StripeCardElement | { token: string };
   email: string;
-  zipCode?: string | null;
+  zipCode?: string | null | undefined;
 };
 export const prepareCardPaymentMethodData = async (
   cardData: CardData,


gumclaw:
Thanks for catching and fixing this — the regression is mine and I want to be clear about that.

`zipCode?: string | null` came in with #6594 (merged 15:00Z). Under `exactOptionalPropertyTypes: true` an optional property does **not** implicitly admit `undefined`, so the call site in `card_payment_method_data.test.ts:404` passing `zipCode: string | null | undefined` was never assignable. It slipped through because #6594 merged ~58 minutes before your #6626 made `npm run typecheck` a required check — the gate that would have caught it landed right after the breakage.

Confirmed independently before commenting, so this isn't me taking your word for it:

- `tsc --noEmit` on clean `origin/main` at `b3dc2c70f` reproduces exactly one error, the `TS2379` at `card_payment_method_data.test.ts(404,40)` — so `main` is currently failing its own required gate.
- The same single error is what turned my #6567 red on push run [30560628743](https://github.com/antiwork/gumroad/actions/runs/30560628743) after I merged `main` in. Nothing in that diff touches this file, which is how I traced it back here.
- Your `Typecheck TS` is `SUCCESS` on this branch.

Widening the type is the right call over touching the test: the production caller genuinely can hand over `undefined` (the ZIP field is optional and unset reads as `undefined`, not `null`), so the type was wrong, not the test. I'd have written it the same way.

Not touching this PR — it's yours and it's green. Flagging one knock-on so it doesn't surprise anyone: every open PR that merges current `main` inherits this red until this lands, and `main` being red on a required check means the queue is effectively blocked. Worth a prompt merge. I'll re-merge `main` into my own branches once it's in and let CI resettle; I'm not going to paper over it with reruns, since it's deterministic rather than flaky.

gumclaw:
Independently reproduced and verified this, having reached the same one-line fix before spotting your PR — so standing down on mine and leaving the evidence here instead.

**It is genuinely main, not a branch artifact.** I ran `npm run typecheck` in a detached worktree at clean `origin/main` and got the identical error, byte for byte. Main's own gate run confirms it: run `30559272332` at `b3dc2c70f` has exactly one failed job, `Typecheck TS`, failing on this line and nothing else.

**The window is narrow and explains why nobody caught it in review.** #6594 introduced the `zipCode` pass-through at 11:00 ET, and #6626 gated CI on `npm run typecheck` at 12:58 ET. Each was green on its own base; the failure only exists in their combination, so it appeared first on main rather than on either PR.

**Verified load-bearing:** with your widening in place `tsc --noEmit` exits 0; reverting just that line to main's version returns exactly 1 error. `npx vitest run app/javascript/data/card_payment_method_data.test.ts` stays green at 22 passed — including the `undefined` case in the `it.each`, which is the one that was unrepresentable under `exactOptionalPropertyTypes`.

Worth noting the test case is real coverage rather than a typo to delete: production callers in `payment_method_result.ts` only ever pass `string | null`, but the runtime normalises with `zipCode || null`, and that `undefined` row is what pins the normalisation. Widening the type is the right direction.

One optional suggestion, take it or leave it — a short comment on the field saves the next person rediscovering why `undefined` is spelled out on a property that is already optional:

```ts
// `undefined` is spelled out because `exactOptionalPropertyTypes` otherwise rejects callers that
// pass the key through from an optional source; an absent ZIP is normalised to null below.
zipCode?: string | null | undefined;
```

Auto-merge is already armed here and this unblocks the gate for every open PR, so I have not touched it.

PR #6631, opened 2026-07-30
Trim the payment method resolver's comments to what a maintainer needs

gumclaw:
## What

Comment-only trim of `app/services/checkout/payment_method_resolver.rb`. **No behaviour change** — the diff touches nothing but comment text.

```
508 → 360 lines   (−148)
git diff --numstat: 122 insertions, 270 deletions
```

Proof the code is untouched:

```
$ diff <(grep -vE '^\s*#' <original>) <(grep -vE '^\s*#' <new>)
CODE IDENTICAL
```

## Why

The file had grown ~150 lines of comment for ~200 lines of code, and most of it was narration wrapped around a small number of real facts: the same "an incompatible `payment_method_types` entry fails the ENTIRE intent create, taking card down with it (gumroad-private#1026)" paragraph was restated in six places, several comments re-derived what the method body says on the next line, and three one-line delegators carried three-line comments duplicating the class method they delegate to.

That paragraph now appears **once**, in the class header, where it explains why every gate below fails closed. Each gate keeps only the part that is specific to it.

## What I deliberately KEPT

Every load-bearing fact stays — this is compression, not deletion:

- **Stripe's Pix transaction window** (0.50 BRL floor / 3,000 USD ceiling, the docs link, the note that each bound stays in the currency Stripe quotes it in, and the 10,000 USD/buyer/month cap that no cart-shaped gate can see).
- **Klarna's USD window** (0–4,000, $1 floor) and *why* the buyer gate is US-only (Stripe requires the buyer's location currency as presentment currency; this lane creates USD intents).
- **The Alipay presentment-currency rule** (`usd` maps to the United States only, `cny` is valid anywhere) and the non-obvious corollary that `account_supported_methods` cannot substitute for the account gate, because an active `alipay_payments` capability on a non-US Standard account is an ordinary state.
- **The gumroad-private#1339 sizing** (mainland buyer IPs = 12 sales in 30 days; Chinese cards used from HK/TW/SG) — it is the whole reason there is no GeoIP lock on Alipay, and nobody would re-derive it.
- **The "do not add a settlement-currency gate here" warning**, with both reverted attempts (gumroad-private#933, #1442) named. This is the trap most likely to bite the next editor, so it got compressed, not cut.
- **Per-method reasons in `RECURRING_INELIGIBLE_PAYMENT_METHOD_TYPES`** — that iDEAL/Bancontact re-billing would need a SEPA mandate we don't collect, that Pix cannot re-bill at all, and that Klarna's exclusion is a launch decision rather than a capability limit. Not re-derivable from the constant.
- **Ordering constraints**: `seller_opt_in_methods` running before the region/PPP/capability gates, the capability intersection staying last, and `forced_currency_methods` having to mirror the presenter's `method_forced_shape?` gate.
- **ACH's withdrawal history** (~4 business days, content only delivers on settlement, gumroad-private#1143) — it explains a live constant, so it stays.

## What went

Restatements of the line below, the six repetitions of the #1026 failure mode, the enumerated "here are my two passes" / "four gates follow" scaffolding that just indexes the code, the three delegator comments, and the `sellers.one?` "belt-and-braces rather than a real branch" aside.

## Test Results

```
$ ruby -c app/services/checkout/payment_method_resolver.rb
Syntax OK

$ bundle exec rubocop app/services/checkout/payment_method_resolver.rb
1 file inspected, no offenses detected

$ bundle exec rspec spec/services/checkout/payment_method_resolver_spec.rb
101 examples, 1 failure
```

The single failure (`:375`, "keeps a launched forced-currency method offered when the charged account has learned a mismatch for that currency") is a pre-existing order dependency in the spec file, not this diff. It passes in isolation on **both** the unmodified file and this branch in the same worktree and DB:

```
$ bundle exec rspec spec/services/checkout/payment_method_resolver_spec.rb:375   # origin/main file
1 example, 0 failures
$ bundle exec rspec spec/services/checkout/payment_method_resolver_spec.rb:375   # this branch
1 example, 0 failures
```

Since the diff is provably comment-only, it could not be otherwise. CI is the authority.

Needs only a skim.

### DIFF
--- a/app/services/checkout/payment_method_resolver.rb
+++ b/app/services/checkout/payment_method_resolver.rb
@@ -2,124 +2,92 @@
 
 # Server-authoritative policy boundary for the client-confirmed Intent path (Lane B): whether a cart
 # may confirm client-side at all, and which Stripe payment methods it may use. The frontend cannot
-# widen it — payment_method_types is the intersection of the eligible set with a hardcoded launched
-# set.
-#
-#   - client_confirm_eligible?: may this cart take Lane B, or must it fall back to Lane A.
-#   - eligible_payment_method_types: the policy set the cart *could* use. Logged, and intersected
-#     downstream with per-method launch/PPP gates.
-#   - payment_method_types: what Stripe actually receives today.
+# widen it.
 #
 # The deferred PaymentIntent's payment_method_types must equal the Payment Element's or Stripe
 # rejects the ConfirmationToken — hence an explicit array, never automatic_payment_methods. The
 # presenter and PreparePaymentIntentService both resolve through here from the same GeoIP basis, and
 # the service hard-stops an ineligible cart before creating the intent, so a mismatch fails closed.
+#
+# Listing a method the account or the intent's currency cannot take fails the ENTIRE intent create,
+# taking card down with it (gumroad-private#1026) — that is why every gate below fails closed.
 class Checkout::PaymentMethodResolver
   # Buyer-present single-seller dynamic set. Apple Pay / Google Pay ride on "card" in the Payment
-  # Element, so they are not separate types here. us_bank_account (ACH Direct Debit) is a
-  # delayed-notification method: it settles asynchronously via the PaymentIntent webhook lifecycle.
+  # Element, so they are not separate types here.
   ONE_TIME_PAYMENT_METHOD_TYPES = %w[card link klarna afterpay_clearpay affirm ideal bancontact upi pix cashapp us_bank_account alipay].freeze
-  # Dropped on a recurring lifecycle, for four different reasons:
+  # Dropped on a recurring lifecycle:
   #   - afterpay_clearpay, affirm, upi: one-time, buyer-present only.
-  #   - ideal, bancontact, pix: one-shot bank approvals with no stored mandate, so renewals would
-  #     have nothing to charge against. Re-billing iDEAL/Bancontact would need a SEPA Direct Debit
-  #     mandate we don't collect; Pix has no re-billing mechanism at all.
-  #   - alipay: Stripe gates recurring Alipay behind approval and excludes it from subscription
-  #     mode, so it stays out until that changes, not until we widen our launch scope.
-  #   - klarna: a launch decision, not a capability limit — memberships and preorders are out of
-  #     scope for its first launch (gumroad-private#933).
-  # Recurring carts fall back to Lane A before a Stripe list is built, but this set is logged and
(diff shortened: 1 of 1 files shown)


gumclaw:
@rmarescu — comments only, no behaviour change, needs only a skim.

`app/services/checkout/payment_method_resolver.rb` goes 508 → 360 lines. Every line changed is comment text; the code is byte-identical (`diff` of both files with comment lines stripped is empty).

The bulk of the cut is one repetition: "an incompatible `payment_method_types` entry fails the ENTIRE intent create, taking card down with it (gumroad-private#1026)" appeared in six places. It now appears once in the class header as the reason every gate below fails closed.

Kept deliberately, since they are the hard-to-re-derive bits in your area: Stripe's Pix window (0.50 BRL / 3,000 USD + the 10,000 USD/buyer/month cap), Klarna's USD window and the presentment-currency reason it is US-only, the Alipay `usd`→US-only presentment rule with why an active capability on a non-US account doesn't save you, the gumroad-private#1339 sizing that justifies having no GeoIP lock on Alipay, and the "do not add a settlement-currency gate here" warning naming both reverted attempts (#933, #1442).

rubocop clean; resolver spec 100/101, and the one failure passes in isolation on origin/main's file too (pre-existing order dependency in the spec).

gumclaw:
CI note: `Typecheck TS` is red on a pre-existing error that this branch cannot have caused.

```
Argument of type '{ cardElement: ...; zipCode: string | null | undefined; }' is not
assignable to parameter of type 'CardData' with 'exactOptionalPropertyTypes: true'.
```

`gh api repos/antiwork/gumroad/compare/b3dc2c70f...chore/trim-payment-resolver-comments-0730` lists exactly one changed file, `app/services/checkout/payment_method_resolver.rb`. That `.ts` file is byte-identical to `main` on this branch, so the error is inherited from the base commit — `Typecheck TS` only became a required gate in #6626, which is main's current HEAD, so this is its first exposure rather than a regression here.

Not fixing it in this PR: it is an unrelated file and this PR is comment-only by construction. Every other check (Lint Ruby, Lint JS/TS, Minitest 0/1, buildkite, Greptile) is green.

gumclaw:
Rebased onto current main (1abb49c) — the `Typecheck TS` error I flagged above was in `app/javascript/data/card_payment_method_data.ts`, fixed on main independently. **CI is now fully green: 79 passing, 0 failing, 0 pending**, including all 18 Test Fast and 50 Test Slow shards.

Diff against the new base is still exactly one file, +122/−270. Ready for review.

PR #6636, opened 2026-07-30
Make the cover "X" remove button tappable on touch devices

gumclaw:
Fixes gumroad-private#1524 — the cover "X" remove button is unclickable on touch devices. Four support tickets, and four manual production-console soft-deletes of the same seller's `AssetPreview` rows, all trace to this one missing option.

## The cause

`CoverEditor.tsx` renders the cover tab strip inside `<Sortable>` with **no `handle` and no `filter`**:

```tsx
<Sortable animation={150} tag={CoversTabList} list={covers} setList={setCovers}>
```

So the entire tab — including the absolutely-positioned remove button on top of it — is drag surface. SortableJS claims the `touchstart`, and the tap never becomes a click. `evt.stopPropagation()` in the button's `onClick` cannot help, because `onClick` is never reached.

Every other `Sortable` call site in the app already constrains its drag surface (`SortableList.tsx` defaults to `handle="[aria-grabbed]"`, `Profile/EditSections.tsx` passes it explicitly). This one is the outlier.

## The fix: two options, and the second one is the one that matters

```tsx
filter={`[${REMOVE_COVER_ATTRIBUTE}]`}
preventOnFilter={false}
```

`filter` alone is the obvious fix and it is **not sufficient**. SortableJS defaults `preventOnFilter` to `true`, and on the filter path it does this (`sortablejs/modular/sortable.esm.js`, verified in the vendored copy at lines 1235 and 1257):

```js
preventOnFilter && evt.cancelable && evt.preventDefault();
return; // cancel dnd
```

`preventDefault()` on a `touchstart` cancels the synthetic click the browser would otherwise synthesise. So with `filter` alone the tap stops being read as a drag — and still produces no click. The button stays dead on touch, which is exactly the reported symptom. Both options are required, and the tests below pin them separately for that reason.

The selector is a data attribute (`data-remove-cover`) rather than a Tailwind class, so restyling the button cannot silently un-fix this.

Desktop is unaffected: mouse `mousedown` was already reaching the button (the bug is touch-specific), and excluding the button from the drag surface is correct there too — dragging a cover by its delete affordance was never intended.

## Tests

New `CoverEditor.test.tsx`, 3 examples, each verified to redden when only its own half of the fix is reverted:

| mutation | result |
|---|---|
| baseline (fix in place) | 3 passed |
| drop the `filter` prop | **2 failed**, 1 passed |
| `preventOnFilter` back to its default | **1 failed**, 2 passed |
| drop the attribute on the button | **1 failed**, 2 passed |

That third row is the one worth noting: without it the `filter` selector could stop matching the button and both other options would go inert while still reading as configured.

The test stubs `ReactSortable` and records the options handed to it — the defect is entirely about which options reach SortableJS, so those options are the contract. It also stubs `$app/components/Popover` (Radix resolves a second React copy under vitest and dies on `useMemo`; it is the unrelated "add cover" menu).

`tsc --noEmit` clean, `eslint` clean, `prettier` applied.

## What is not covered

No real-device confirmation — happy-dom cannot dispatch a genuine touch sequence with the synthetic-click behaviour, so the `preventOnFilter` reasoning is grounded in the SortableJS source rather than a device test. One tap on the deployed preview by someone holding a phone closes that out.

### DIFF
--- a/app/javascript/components/ProductEdit/ProductTab/CoverEditor.test.tsx
+++ b/app/javascript/components/ProductEdit/ProductTab/CoverEditor.test.tsx
@@ -1,0 +1,104 @@
+// @vitest-environment happy-dom
+import { cleanup, render, screen } from "@testing-library/react";
+import * as React from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { AssetPreview } from "$app/parsers/product";
+
+import { CoverEditor } from "$app/components/ProductEdit/ProductTab/CoverEditor";
+
+// The real ReactSortable installs SortableJS on the DOM node, which needs layout happy-dom
+// does not provide. Stub it to a plain wrapper that records the options it was handed: the
+// bug here is entirely about which options reach SortableJS, so those options ARE the
+// contract worth pinning.
+type StubbedSortableProps = {
+  children?: React.ReactNode;
+  tag?: React.ElementType;
+  list?: unknown;
+  setList?: unknown;
+  filter?: string;
+  preventOnFilter?: boolean;
+};
+const sortableProps = vi.hoisted(() => {
+  const box: { current: StubbedSortableProps | null } = { current: null };
+  return box;
+});
+vi.mock("react-sortablejs", () => ({
+  ReactSortable: ({ children, tag, list: _list, setList: _setList, ...options }: StubbedSortableProps) => {
+    sortableProps.current = options;
+    const Wrapper = tag ?? "div";
+    return <Wrapper>{children}</Wrapper>;
+  },
+}));
+
+// The remove button is hover-gated on desktop and always shown below `lg`. Report a small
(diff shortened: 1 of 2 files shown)


gumclaw:
Greptile's automated run came back clean, with one nuance worth recording: its emulated touch check reported the removal callback firing once in *both* the pre-fix and post-fix configurations. That does not match what the vendored SortableJS source does on the filter path (`preventOnFilter && evt.cancelable && evt.preventDefault()` before the early return), and a real touch sequence is the only thing that can tell the two apart — an emulated tap that dispatches a click directly will succeed either way, because the suppressed event is the browser's *synthetic* click derived from `touchstart`.

So the "What is not covered" note in the description still stands as the open item: one tap on the preview app from an actual phone, on the cover "X". If the button responds there, both options are confirmed doing their job.

PR #6641, opened 2026-07-30
Roll back a failed upsell save completely

gianfrancopiana:
## What

Saving an upsell now runs inside a single database transaction. When validation rejects the save, every write made along the way is undone and the upsell stays exactly as it was.

## Why

Editing an upsell wrote some changes immediately: the offer code's product list, plus deletions of replaced version mappings and removed offer codes. Validation ran afterwards. A rejected edit could therefore leave the offer code attached to the wrong product while the rest of the upsell kept its old values — for example, when moving an upsell to a cheaper product with a discount too large for it.

`SaveUpsellService` now owns the save and wraps the whole sequence in one transaction, following the same pattern as `SaveInstallmentService`. Putting the transaction in the service covers all four controller call sites (checkout and public API, create and update) in one place; those call sites now read the outcome from the record's errors.

## Before/After

Backend fix with no visible change on successful saves, so there is no UI to capture.

- Before: a rejected upsell edit still moved the offer code to the new product in the `offer_codes_products` join table, and replaced upsell variants stayed deleted.
- After: a rejected edit leaves the offer code's products, its amounts, the upsell's product, and its variants untouched.

QA steps:

1. Create a cross-sell on a $5 product with a $1 discount.
2. Edit it to offer a $10 product with a $9.50 discount, so the price after discount falls under the $0.99 minimum.
3. Saving is rejected with the price validation message.
4. Confirm the discount still applies to the $5 product with the $1 amount, and the upsell is unchanged.

## Test Results

Three new regression specs prove a rejected update leaves the offer code's product list untouched — at the service layer, the checkout controller, and the public API controller. With the fix reverted, all three fail with the offer code pointing at the new product.

```
bundle exec rspec spec/services/save_upsell_service_spec.rb \
  spec/controllers/checkout/upsells_controller_spec.rb \
  spec/controllers/api/v2/upsells_controller_spec.rb

69 examples, 0 failures
```

The checkout upsell creation and editing system specs also pass (9 examples), and RuboCop is clean on all changed files.

---

AI disclosure: implemented with Claude Fable 5 and 5.6 Sol


gianfrancopiana:
Adding more e2e tests.

gumclaw:
## Adversarial pre-merge review (fable-5) — `READY-TO-MERGE`

Ran a local fable-5 adversarial pass on `213be912`, plus mutation testing and independent verification of the load-bearing claims. No P1. I've pushed the P2 fix to this branch (`f27d2624`), measured both P3s against production, and driven the seller UI on the preview app — neither P3 is a real problem and the rollback behaves correctly end-to-end.

### Specs are load-bearing (mutation-proven)

Reverted `save_upsell_service.rb` to `main`, kept the new specs: **8 examples, 8 failures**, including `rolls back the offer code's product reassignment when the save fails`. Restored: 8/8 pass. The new coverage cannot pass against the old code, and the removed `expect(upsell.save).to be(true)` assertions are adequately replaced — create specs pin `be_persisted`, update specs assert through `reload`.

### P2 (fixed in `f27d2624`) — `errors.empty?` was a fragile success proxy

`app/services/save_upsell_service.rb:34`, all four controller sites.

Unreachable today, and I checked exhaustively rather than assuming: `Upsell` declares no callbacks; `Deletable` and `ExternalId` define none; `Upsell::Sorting` is scopes only; PaperTrail only records versions; `UpsellVariant` has none; `OfferCode`'s save-phase hooks are `after_save` plus attribute-mutating `before_validation`s. `grep -r "throw :abort" app/ lib/` returns **0 hits repo-wide**. Autosave copies child validation failures onto `upsell.errors`, so every reachable `save == false` populates errors.

The trap was the coupling, not a present bug: the day anyone adds an aborting callback to `Upsell`, `OfferCode` or `UpsellVariant`, `save` returns false with empty errors, the service rolls back, and all four controllers render `success: true` for a write that never happened — API `create` would call `success_with_upsell` on an unpersisted record.

The pushed commit makes the invariant self-enforcing rather than assumed, and adds a spec that pins it. Mutation-proven: with the guard reverted and the spec kept, the new example fails on `expect(upsell.errors).to be_present`; with the guard in place, **9 examples, 0 failures**. Rubocop clean.

### P3 — rolled-back offer-code writes still purge the product cache: measured, negligible

`app/models/offer_code.rb:46` → `app/modules/product/caching.rb:14`. `after_save :invalidate_product_cache` (not `after_commit`) runs `Rails.cache.delete` inside the transaction; on rollback the DB write vanishes but the eviction already happened.

Production numbers (read-only console, replica lag 0s):

- **132,209** alive upsells; **1,354** updated in the last 7 days (~190/day) and 6,373 in 30 days. Failed saves are a subset of that, so the eviction fires at most a handful of times a day.
- **47,758 of 49,904** upsell-attached offer codes (~96%) map to exactly **one** product, so a rollback evicts one product's cache key. The tail is thin (one code at 170 products, one at 130).
- Direction is safe — over-invalidation, repopulated from committed state, and the companion `product_cached_values` row updates *do* roll back.

A few single-product cache misses per day. Not worth code.

### P3 — transaction spans validation queries: smaller than it looks

The worry was `upsell.save` running `OfferCode#price_validation` (per-product price checks over `applicable_products`) and `times_used` (a `SUM` over `purchases`) while holding locks. Production says both of the expensive paths are dead for this flow:

- **0 of 49,904** upsell offer codes are `universal`, so `applicable_products` is just `products` — a ≤1-row set 96% of the time, never the seller's whole catalogue.
- **0 of 49,904** have a `max_purchase_count`, and `max_purchase_count_is_greater_than_or_equal_to_inventory_sold` returns early unless it changed — so the `SUM` over `purchases` never runs here.
- `has_one_upsell_variant_per_selected_variant` groups over **2,095** alive `upsell_variants` platform-wide, scoped to one upsell.

The lock hold is a handful of indexed lookups on tiny sets. Unavoidable cost of doing the fix correctly, and in practice cheap.

### Preview QA — the rollback works end-to-end in the real seller UI

Ran the seller flow on the preview app (`x-revision` verified == `c67e411a3`) against a seeded cross-sell: base product $10, companion $5, upsell with a $1 fixed discount.

| # | Step | Result |
|---|---|---|
| 1 | Rename the upsell to `RENAMED SHOULD NOT PERSIST` **and** set the discount to $9.50 (leaves $0.50, below the $0.99 USD minimum) → Save | `{"success":false,"error":"The price after discount for all of your products must be either $0 or at least $0.99."}` |
| 2 | Read the DB back after the failure | `name="QA6641 cross-sell"`, `oc_cents=100`, `oc_products=["QA6641 base"]`, `selected_products=["QA6641 companion"]`, `orphan_deleted_offer_codes=0` — **nothing partially persisted** |
| 3 | Correct the discount to $2 in the **same form, no reload**, rename to `QA6641 RESAVED OK` → Save | `{"success":true,...}`, `discount.cents: 200`, back to the list showing the new name |
| 4 | Read the DB back after the success | `name="QA6641 RESAVED OK"`, `oc_cents=200` — committed cleanly |

The specific downstream risk I wanted to rule out was the form/server divergence this PR introduces: before the change a failed save still persisted the offer-code and product reassignment, so the form and DB agreed even on failure; now the server rolls back while the form still holds what the seller typed. In practice that's a **non-issue, and strictly better than before** — on error `UpsellsPage` takes the `catch` branch, so `setState` never runs, the view stays on `edit`, and the typed values are retained (screenshot 2 shows `RENAMED SHOULD NOT PERSIST` still in the Name field with the DB untouched). The seller reads the error, fixes the one bad field, and re-saves from the same form — step 3 proves that path commits. On success `setState` replaces the whole list from the server response, so there's no stale client state either way.

Invalid discount entered
After the failed save — form retained, DB untouched
After correcting and re-saving

---

### Ruled out

- **`ActiveRecord::Rollback` swallowed by an outer transaction** — the four flipped controllers are the only callers; neither `Sellers::BaseController`, `Api::V2::BaseController` nor `ApplicationController` has an `around_action` or transaction wrap, so the service's transaction is outermost in production. In specs `use_transactional_fixtures = true` opens fixture transactions with `joinable: false`, making the block a real savepoint. Specs and prod agree; the fix depends on neither harness quirk.
- **Phantom Elasticsearch reindex on rollback** — `reindex_associated_products` → `Link#enqueue_index_update_for` registers via `AfterCommitEverywhere` (`app/modules/product/searchable.rb:64,469`), so on rollback the callback is discarded and no job is enqueued. PaperTrail versions are same-transaction inserts and roll back with everything else.
- **Stale in-memory state serialized after rollback** — both error paths serialize only messages (`@upsell.errors.first.message`; `errors.full_messages.to_sentence`), never the object. `as_json` is reached only on success, and `fetch_upsells` re-queries. The in-memory ghost state never reaches the client.
- **Return-contract change breaking a caller** — both `update` actions pass `@upsell` in as `existing_upsell` and the service mutates and saves that same object, so discarding the return value is safe. `backfill_absent_associations` only mutates `params`, no writes outside the transaction.
- **AI-slop** — clean. The one added comment is three lines of non-obvious *why* (the HABTM/`mark_deleted!`-write-before-validation ordering trap), which is exactly what the code can't say. Spec renames `builds` → `creates` track the real contract change.

---

Review by **claude-opus-5** running as Hermes Agent (fable-5 adversarial pass), with mutation testing, callback-chain verification, and audited read-only production measurement.

PR #6586, opened 2026-07-29
Keep custom-profile product and post links on the seller's own domain

gumclaw:
Ref antiwork/gumroad-private#1534

## What

Custom-HTML profile pages receive product and post URLs in `gumroad-data`. Those URLs previously always used the seller's `*.gumroad.com` subdomain, even when Gumroad served the page from the seller's custom domain.

The merged change builds those URLs on the seller's exact custom hostname only when persisted state says that hostname resolves to Gumroad and has a valid certificate. It falls back to the seller subdomain otherwise, refreshes routability outside the request path, prevents older DNS observations from overwriting newer ones, includes the emitted host in the profile-data cache key, and tells the Store Agent to preserve the canonical payload URLs.

## Why

A visitor should not leave a seller's domain because a custom storefront used the URL Gumroad supplied. Building the correct absolute URL at the source fixes every payload consumer without asking custom pages to rebuild hosts.

`CustomDomain#active?` is not enough: the stored host may differ from the host that resolves or has a certificate. The emitted host must use exact-host DNS and certificate state.

## Merged behavior

| Persisted seller state | Emitted product/post host |
|---|---|
| Exact custom host resolves and has a valid certificate | Exact custom host |
| Exact DNS works but its certificate is missing | Seller subdomain; certificate issuance queued |
| Routability has never been checked | Seller subdomain; background refresh queued |
| A positive result is stale | Last known custom host while refresh runs |
| Certificate ages out, custom domain is removed, or the domain belongs to a product | Seller subdomain |
| Seller has neither host | Existing relative post path |

## Merge and CI evidence

GitHub merged PR head `31d2ae73021b28a86eba6c675556eb1743124115` as squash commit `f4598f3d4b6d0a5a71b41629e82201537090cdbe`.

The earlier `Test Fast 6` failures came from exact timestamp assertions that supplied nanoseconds to MySQL's microsecond-precision columns. The fixtures now use database-supported precision, and the affected examples pass.

The PR head's GitHub Actions `ci/green` gate passed before merge. Its separate Buildkite context still showed an older failed build. The squash-merge commit then passed both [Buildkite #18062](https://buildkite.com/gumroad-inc/gumroad/builds/18062) and [GitHub Actions `ci/green`](https://github.com/antiwork/gumroad/actions/runs/30589122144).

## Interaction evidence

The hosted preview captured both arms at runtime revision `5e1b67f54250`. Changes between that revision and the merged PR head were test fixtures and QA evidence, not runtime behavior.

[...]

PR #6644, opened 2026-07-30
Screen sanctioned jurisdictions server-side on every purchase

gumclaw:
## What

Adds a server-side sanctions screen that runs on **every** purchase — digital included, and subscription renewals — instead of only hiding the buy button at render time. Also screens the four comprehensively sanctioned Ukrainian regions (Crimea, Sevastopol, Donetsk, Luhansk), which had no coverage at all.

The gate is `Purchase#validate_sanctioned_location`, a `before_create` that screens the `[alpha2, subdivision]` pairs in `#sanctioned_location_signals` against `Compliance::Countries.blocked_location?` and fails the purchase with a new `PurchaseErrorCode::BLOCKED_SANCTIONED_LOCATION`. It runs before any charge is created.

Closes the enforcement half of antiwork/gumroad-private#1566.

## Why

`Compliance::Countries.blocked?` was only consulted in two places that matter for a buyer: `Link#compliance_blocked`, which hides the buy button at render time, and `Purchase#validate_shipping`, which only runs for physical products. So for a digital product:

- a buyer who reaches `/checkout` directly never hits the render-time check
- a buyer whose product page was rendered from a different IP than they pay from never hits it either

There was no server-side backstop. A sanctioned-jurisdiction buyer could complete a digital purchase and get charged. Region-level sanctions were unscreened entirely, in every path.

**Two places the issue's own premise was wrong**, corrected from primary sources rather than carried forward:

- **Syria is deliberately NOT added.** OFAC E.O. 14312 revoked the Syria program effective 2025-07-01 and 31 CFR part 542 was removed from the CFR. Syria is absent from OFAC's live active-programs table. The existing spec asserting Syria is unblocked was correct; #5314 had already handled this. Left unchanged, with the reason pinned in a comment so it does not get "fixed" back.
- **No new feature was built.** Per the direction on the issue, there is no seller-facing country-restriction setting here — just the list update and the enforcement path.

## Before / after

| | Before | After |
|---|---|---|
| Digital purchase, sanctioned country | charged | fails, `BLOCKED_SANCTIONED_LOCATION`, no charge attempted |
| Digital purchase, sanctioned region (Crimea etc.) | charged | fails, no charge attempted |
| Physical purchase, sanctioned country | fails (`BLOCKED_SHIPPING_COUNTRY`) | unchanged — `validate_shipping` still owns it, own error code |
| Subscription renewal, subscriber's address in a sanctioned jurisdiction | renews forever | stops renewing (see caveat below) |
| Subscription renewal, address fine but original purchase was made from a sanctioned jurisdiction | renews | still renews — the copied IP fields are not read as a current location |
| Unsanctioned buyer | succeeds | succeeds — pinned in both directions |

## Tests

All runs on `gumroad_test_1566`, a dedicated database (the shared `gumroad_test` was dropped mid-run by a concurrent session on this machine — worth knowing if you see it vanish).

- `spec/lib/utilities/compliance_spec.rb` — **73 examples, 0 failures**
- `spec/services/purchase/create_service_spec.rb -e sanctioned` — **8 examples, 0 failures**
- `spec/models/purchase/purchase_subscription_spec.rb -e "sanctions screening on a renewal"` — **5 examples, 0 failures**
- rubocop clean on all four changed production files

**Mutation-proofed, both directions.** Deleting the `"RU" => %w[CR SEV]` row turns the unit pin *and* the purchase-level pin **red** — screenshot in [`qa-media/pr-6644-sanctions-mutation-proof.png`](https://raw.githubusercontent.com/antiwork/gumroad/22d9100548f32e77271d4e51325c9f1e250090a8/qa-media/pr-6644-sanctions-mutation-proof.png) shows the red-then-green cycle, so these specs are proven to actually test the row rather than pass regardless.

**No charge is attempted on a blocked purchase**, which is the property that matters most here: every blocked example asserts `stripe_transaction_id` is nil, and the new cassette permits exactly one HTTP interaction — `POST /v1/payment_methods` (payment-method setup). There is no PaymentIntent in it. With VCR at `record: :none` on CI, a charge attempt would raise rather than pass silently.

Also regression-checked against an unmodified-`main` worktree at a pinned seed: same failure set on base and branch (pre-existing local fixture-collision flakes, absent on CI).

## Review notes

This went through three rounds of local adversarial review. Round 2 caught a **P1 in my own earlier fix commit**: I had deleted the `RU => [CR, SEV]` row, reasoning from a scan of our MaxMind database showing it attributes the occupied regions only to Ukraine. That reasoning does not hold, because the *first* signal screened is **buyer-declared** (`state` from the address form, `country` re-derived from `sales_tax_country_code_election` for digital products), not a geo lookup — so a buyer self-declaring RU + CR was blocked before that commit and allowed after it. The mmdb argument never covered that path, and the mmdb is gitignored so no reviewer could check the evidence I had cited. The row is restored and now pinned at both levels. Keeping it cannot over-block: ISO 3166-2:RU has 83 subdivisions, all alphabetic, with no `CR` and no `SEV`.

**Two follow-ups I deliberately did not fold in:**

1. **A blocked renewal sends a misleading email.** `BLOCKED_SANCTIONED_LOCATION` is not a payment error, so `Subscription#process_purchase!` takes the `subscription_charge_failed` branch, whose template says "We attempted to charge your card… but were unable to" — literally untrue (no attempt was made) and unfixable by the customer, who is then cancelled ~5 days later by `UnsubscribeAndFailWorker`. An error-code-aware renewal email is a product/compliance decision, not mine. Filing separately.
2. **`Order::CreateService`'s write of `ip_country`/`ip_state` is pinned at the model level, not through the service.** The gate geolocates `ip_address` itself precisely so it does not depend on that caller, so a regression there degrades one signal rather than opening the hole.

**Renewals screen the declared address, not the inherited IP.** `Subscription#build_purchase` copies the original purchase's `country`, `state`, `ip_address`, `ip_country` and `ip_state` onto every renewal, so those IP fields describe where the buyer was when they subscribed. Screening them would reject every renewal of a subscriber who has since moved out of a sanctioned jurisdiction — permanently, with no way for them to correct it, and the failed-charge path then schedules the membership for cancellation. Renewals therefore screen only the address, which the subscriber can update from Manage subscription. `spec/models/purchase/purchase_subscription_spec.rb` pins both halves — an inherited-Cuba IP with a US address renews, a Cuba address does not — and both inherited-IP pins redden when the exclusion is removed.

The exclusion is keyed on the caller supplying no live IP, so it opts back *in* whenever one is available (the manage-subscription upgrade path through `Subscription::UpdaterService`). Because `ip_country`/`ip_state` are derived from `ip_address`, a caller overriding only the IP would otherwise leave the original purchase's derivations standing beside a current IP and reintroduce the same historical-location read; `build_purchase` now clears them alongside the override, pinned by a spec that reddens without the clear.

**Worth your explicit attention before merge:** screening now runs on **subscription renewals**, so an existing subscriber whose declared address is in a sanctioned jurisdiction will stop renewing. I believe that is correct — continuing to bill is the same prohibited transaction as the first charge — but it is a live-revenue behaviour change on existing customers, and it is the reason I have not armed auto-merge.


gumclaw:
@nyomanjyotisa this needs your review — it's compliance/purchase-validation code and you own `lib/utilities/compliance/countries.rb`. CI is green and the branch is settled, so the ask is live.

**The judgment call I want your eyes on:** sanctions screening now runs on **subscription renewals**, so an existing subscriber whose declared address sits in a sanctioned jurisdiction stops being charged. That is intended — continuing to bill is the same prohibited transaction as the first charge — but it has a live-revenue effect on current customers rather than only on new purchases. That is why auto-merge is deliberately **not** armed here and will not be. Whether this lands as-is or needs a notice/grace period first is your call.

**State at head `51224347f`:** the full suite passed — 75 successful jobs, 1 skipped, 0 failures, 0 pending — on both the `push` and `pull_request_target` runs. `buildkite/gumroad` is green. `MERGEABLE`, no conflicts.

An earlier `buildkite/gumroad` red on `adfae3ab7` was an infrastructure false red, not a spec failure: build [#18057](https://buildkite.com/gumroad-inc/gumroad/builds/18057) died in 7 seconds without running a step, same signature as [#18056](https://buildkite.com/gumroad-inc/gumroad/builds/18056) on #6586 twenty minutes earlier — the agent resolves `refs/pull/<n>/head` to the previous head, then fails `git checkout -f <new sha>` with `fatal: reference is not a tree`. It's a checkout race against a branch that moved while the build queued, which hits any PR pushed twice in quick succession. An empty commit re-triggered it and it now passes.

**Three things worth your attention in the diff:**

- **Renewals screen the declared address, not the inherited IP.** `Subscription#build_purchase` copies the original purchase's `country`, `state`, `ip_address`, `ip_country` and `ip_state` onto every renewal, so those IP fields record where the buyer was when they subscribed. Screening them would fail every renewal of a subscriber who has since moved out of a sanctioned jurisdiction — permanently, with the failed-charge path then cancelling the membership. Renewals screen the address instead, which the subscriber can actually update from Manage subscription. The exclusion keys on the caller supplying no live IP, so the manage-subscription upgrade path still screens a current one; because `ip_country`/`ip_state` are derived from `ip_address`, `build_purchase` clears them alongside an IP override rather than leaving stale derivations beside a current IP. Every branch is pinned and verified to redden when its guard is removed.
- **Syria is deliberately not added**, contrary to the original issue's premise. OFAC E.O. 14312 revoked the program effective 2025-07-01 and 31 CFR part 542 was removed from the CFR. The existing spec asserting Syria is unblocked was right.
- **The `"RU" => %w[CR SEV]` row.** I deleted it mid-review on a bad argument — that our MaxMind database attributes the occupied regions only to UA — and adversarial review caught that as a P1: the first signal screened is *buyer-declared*, not a geo lookup, so a buyer self-declaring RU + CR was allowed. It's restored and mutation-proofed at both the unit and purchase level.

If you'd rather the renewal behaviour be a separate decision from the enforcement gate, say so and I'll split it.


---

**Head moved once for evidence naming, no code change.** `22d9100` is a pure `R100` rename — `pr-gp1566-sanctions-mutation-proof.png` → `pr-6644-sanctions-mutation-proof.png` — because CONTRIBUTING wants `pr-<number>-` and this carried the driving issue's number, which would have landed in `main` looking like another PR's evidence. Body re-pinned to the new SHA; bytes unchanged (108,505), URL 200. CI re-runs over identical code.

PR #6647, opened 2026-07-30
Write a bank rejection's attribution in the note it belongs to, atomically

gumclaw:
## What

Follow-up to #6606. The bank-rejection banner it added reads `json_data["bank_account_id"]` off the payout note to decide which bank row Stripe actually rejected, falling back to a `note.created_at >= bank_account.created_at` comparison for legacy notes that predate the field. This makes the producer write that field in the note's INSERT rather than a second save, and makes the argument it comes from required.

`User#add_payout_note` gains a `json_data:` keyword so structured fields land in the insert. `StripeMerchantAccountManager.record_bank_sync_failure_note` passes all three (`stripe_error_code`, `stripe_error_message`, `bank_account_id`) through it, and `bank_account:` is now a required keyword rather than defaulting to `nil`.

## Why

Two review items on #6606, raised before it merged:

- **The attribution was written in a second save.** `add_payout_note` inserted the note, then `json_data["bank_account_id"]` was assigned and `save!` called. The note is readable by the presenter the moment it is inserted, so between the two writes it looks like an unattributed legacy note and falls through to the timestamp comparison the field exists to replace — which is how the banner ends up blaming a bank row Stripe never objected to.
- **`bank_account:` defaulted to `nil`, and nothing pinned it.** A producer that omitted the argument wrote an unattributed note silently. It is now required, and the producer-side specs assert the field on every bank rejection path.

The related failure the field prevents is the reason the row is *passed down* rather than re-read: `update_bank_account` calls Stripe over the network, so a seller can save replacement details before the rejection lands, and `user.active_bank_account` at note time may already be the replacement.

## Before / After

No user-visible surface changes. The banner, its copy and its five states are exactly as shipped in #6606; this changes only how the note backing it is written. The behavior difference is a window measured in milliseconds during a Stripe rejection, which has no steady-state rendering to capture — so the evidence is a walkthrough of the note's lifecycle instead.

## Walkthrough

https://github.com/antiwork/gumroad/blob/b80e9246275f0fa310de35365e6f092c4cf53ea6/qa-media/pr-6647-atomic-note-attribution-walkthrough.mp4?raw=true

A Stripe rejection is recorded while every `comments` write is captured, then the note and the presenter that reads it are inspected:

1. A rejection arrives for the bank row the sync submitted (id 69; a replacement the seller saves mid-sync would be 70).
2. One `INSERT INTO comments`, zero `UPDATE comments` — there is no window in which the note exists unattributed.
3. The note carries `bank_account_id`, `stripe_error_code` and the visibility flag from that single insert.
4. `SettingsPresenter#current_bank_sync_failure_note` matches on the attribution rather than falling through to the timestamp comparison, so a row saved after the rejection cannot be blamed.
5. Omitting `bank_account:` now raises `ArgumentError: missing keyword: :bank_account` instead of silently writing an unattributed note.

## Test plan

`spec/business/payments/merchant_registration/stripe/stripe_merchant_account_manager_payout_retry_spec.rb` — **58 examples, 0 failures**.

Together with the two suites that read the field and the three that call `add_payout_note`:

| Suite | Result |
|---|---|
| `stripe_merchant_account_manager_payout_retry_spec.rb` | 58 examples, 0 failures |
| `settings_presenter_spec.rb` + `payout_note_visibility_spec.rb` (with the above) | 152 examples, 0 failures |
| `retry_stripe_rejected_payout_setup_for_seller_job_spec.rb`, `retry_stripe_rejected_payout_setups_job_spec.rb`, `handle_new_bank_account_worker_spec.rb` | 51 examples, 0 failures |

`bundle exec rubocop` on all three changed files: no offenses. The suite above was re-run on the branch after merging current `main` in.

**Every new assertion is proven load-bearing by mutation**, not by inspection:

| Mutation | Result |
|---|---|
| `"bank_account_id" => nil` | 5 assertions fail |
| Restore the pre-change insert-then-save shape | the SQL-shape example fails, and only it |
| `bank_account&.id` → `user.active_bank_account&.id` (re-read instead of the passed row) | the mid-sync-replacement example fails |

That third mutation is the one worth calling out: it survived the first revision of this branch entirely, because every existing spec has a single bank row, so the passed row and a re-read are indistinguishable — the exact scenario the argument exists for was pinned nowhere. The added example replaces the row inside the stubbed `Stripe::Account.update` so the two diverge.

Manual verification is not meaningful here: the window being closed is a sub-second interval inside a Stripe rejection. The SQL-shape assertion (one `INSERT INTO comments`, zero `UPDATE comments`) is the durable proof, and it fails if the second save returns.

## Notes for the reviewer

- `seller_notified` is still marked in a post-insert save, deliberately: it records that the seller has *been emailed*, which cannot be known at insert time, and the presenter never reads it. Its only reader is the weekly retry job, where a missed marker costs one duplicate email rather than a wrong banner.
- `add_payout_note`'s merge puts `seller_visible:` last, so the declared keyword always wins over a `seller_visible` key passed through `json_data`. No caller does that today.
- The other ~25 `add_payout_note` call sites are unaffected: the method has no positional parameters, so nothing can be captured as `json_data`, and omitting it produces byte-identical `json_data` to before.

## AI disclosure

Written by claude-opus-5, prompted by review feedback from nyomanjyotisa on #6606 (two items: make `bank_account:` required and assert it in the producer specs; persist the note and its attribution atomically). #6606 merged about 100 seconds after that review was posted, so the fix lands here rather than on that branch.

The branch went through the local fable-5 adversarial review, which returned **SHIP WITH FIXES** with no P1s. Its P2 was the unpinned re-read case above; the P3s were the missing `CardError`-path assertion, the SQL-shape example not asserting the value it was pinning the atomicity of, and a comment duplicating the `add_payout_note` doc. All four are fixed in `0f0e70d71`.

Preview: https://fix-bank-rejection-note-attribut.apps.staging.gumroad.org — captured at
`x-revision: a8f0757eefaa`, equal to the PR head at capture time. All four frames are pinned to
the evidence commit `838aadd17`.

**Amending "no user-visible surface changes" above.** That is right about the *diff* — no template,
component or copy moved. But the field this PR writes is read by
`SettingsPresenter#current_bank_sync_failure_note`, which decides whether the Settings → Payments
warning banner renders and which bank row's guidance it quotes. The mid-sync-replacement window is
therefore not only a millisecond interval: it has a steady-state rendering, because a note whose
attribution went missing keeps pointing the banner at the wrong row for as long as the note is
alive. Both outcomes are captured below.

Fixture seeded on the preview write console (`seller@gumroad.com`, which starts with zero bank
accounts and zero payout notes): two `AchAccount` rows and one rejection recorded through the real
producer `StripeMerchantAccountManager.record_bank_sync_failure_note`.

### The banner the note drives

Rejection recorded for the bank row that is still active. The presenter matches on
`json_data["bank_account_id"] == 3` rather than falling through to the timestamp comparison.

| | |
|---|---|
| Desktop | Bank rejection banner, desktop |
| Mobile (375px) | Bank rejection banner, mobile |

### The window this PR closes — a replacement row must not be blamed

The seller saves replacement bank details (row #5) while the sync is still talking to Stripe, then
the rejection for the row the sync *submitted* (#3) lands. The note names #3, so the banner stays
hidden and the seller is not told to fix an account nothing objected to. Payout method shows the
replacement row, and there is no warning:

Mid-sync replacement, no banner

Run against the same three rows, the pre-fix shape (unattributed note + `created_at >=
bank_account.created_at`) selects that note and blames row **#5** — the replacement. That is the
wrong-banner outcome, and it is what the atomic insert prevents: between the old INSERT and its
follow-up save, the note really did read as an unattributed legacy note.

### Console transcript

Every cell below is a `rails runner` MARK line off the pod, including the insert-shape check
(`comments_delta=1`, `created_at == updated_at`) that the SQL-shape spec asserts in CI:

Console transcript

### QA steps (reviewer click path)

Preview: https://fix-bank-rejection-note-attribut.apps.staging.gumroad.org
Log in as `seller@gumroad.com` / `password` (2FA `000000`, pre-filled — just click Login).

1. **Settings → Payments** → the yellow banner reads *"Our payment partner couldn't accept your
   bank details as entered…"*, and **Pay to the order of** shows `QA6647 Arm A Row` — the row the
   note names.
2. On the write console, soft-delete that row, save a replacement, and record a rejection naming
   the *deleted* row. Reload **Settings → Payments**: the banner is gone and the replacement is
   shown, because the note names a row that is no longer active.

**Durable preview state the next session inherits:** `seller@gumroad.com` now has one alive
`AchAccount` (`QA6647 Arm A Row`, id 3) and one alive payout note (id 1) naming it, plus a
`UserComplianceInfo` with `country: "United States"` — added because the seeded fixture had no
country, which made an unrelated "Where are you located?" modal dim every frame. No production
data was touched.

**Not captured:** the block / terminal-rejection / abandoned copy variants of the banner. They are
selected by `bank_rejection_kind_for` from the same note this PR writes, so they exercise the same
attribution path, and they are covered by `settings_presenter_spec.rb` in the run above.

## AI disclosure

Screenshots captured and this section written autonomously by Hermes Agent (Claude Opus 5) as the
preview-app screenshot watcher. The code change itself is disclosed above.


nyomanjyotisa (commented):
Please refresh onto current main and add the required short non-visual walkthrough video.

gumclaw:
Refreshed onto current main, walkthrough attached, and CI is fully green.

`main` is merged in at `da9ceb0` — it had moved ahead (#6586, #6646, and #6658). Every check on that head passes: the full GitHub Actions matrix, Buildkite, and `ci/green`. The payout-retry suite was also re-run locally on the merged branch (58 examples, 0 failures), and rubocop is clean on all three changed files.

The video is `qa-media/pr-6647-atomic-note-attribution-walkthrough.mp4`, linked from a new "Walkthrough" section in the description. Since there is no pixel change here, it walks the note's lifecycle instead: a Stripe rejection is recorded while every `comments` write is captured, showing one `INSERT` and zero `UPDATE`s, then the note's `bank_account_id`, the presenter matching on it rather than the timestamp fallback, and the `ArgumentError` you now get for omitting `bank_account:`.

If you saw a red `Lint JS/TS` earlier: that was two Apple Pay end-date assertions in `applePayRecurringPaymentRequest.test.ts` failing on the month rollover, not on this branch — the same two lines failed on every branch pushed in that window, and this diff contains no JS or TS. #6658 pinned the clock and that file is green now.

This touches the payout path, so it is not armed for auto-merge — it is assigned to you for review and merge.

nyomanjyotisa (approved):

PR #6649, opened 2026-07-30
Don't send an urgent evidence request the submission form will reject

gumclaw:
## What

`ContactingCreatorMailer#chargeback_notice` decided whether to ask the seller for their side of a dispute from two conditions — the seller had been contacted, and hours remained in the window. The page behind that request, `Purchases::DisputeEvidenceController`, refuses on three: not contacted, already submitted, and resolved.

The notice goes out through `deliver_later` and renders when the job runs, so the seller can submit their statement, or `FightDisputeJob` can resolve the dispute, in between. When either happened, the mailer still rendered the urgent subject, the "Any additional information you can provide in the next N hours" paragraph, and the "Submit additional information" button — and the button led to a dashboard redirect rather than a form.

All three conditions now live in `DisputeEvidence.accepting_seller_evidence?`, and both callers — the mailer and `Charge::Disputable#perform_dispute_formalized_side_effects!` — ask through it.

Follow-up to #6534, addressing Greptile's review on that PR. The finding arrived after it merged.

## Why

A seller who has already sent their statement gets an email headed "🚨 Urgent: Action required for resolving disputed sale" asking them to send it again, with a button that turns them away. During a 72-hour deadline on a charge that has already been deducted from their balance, that reads as us having lost their submission. The subject was the worse half: it came off `seller_contacted?` alone, so it shipped on notices whose body had no request in it at all.

Two paths enqueue this notice — formalization and `CreateMissingDisputeEvidenceJob`, which backdates its windows to a few hours to land inside the processor's cutoff. The narrower the window at enqueue, the more likely a transition lands before the render.

Formalization already carried the same three conditions inline; that copy is what let the two surfaces drift, so it now delegates rather than repeating them. The hours test stays deliberately stricter than the controller's: `check_if_needs_redirect` lets an elapsed window through until `FightDisputesJob` resolves the row, and the notice's own body promises "in the next N hours". A nil stamp reads as acceptable, so a dispute with no evidence surface at all (PayPal, Stripe Connect) still gets the plain notice — the predicate governs whether we *ask*, never whether we send. When it returns false the seller still receives the notice, without the request.

## Before / After

Rendered from the real mailer on both sides of the fix, in the state where the seller has already submitted:

| | Notice |
|---|---|
| **Before** — seller already submitted | before |
| **After** — seller already submitted | after |

And where the dispute resolved before the render:

| | Notice |
|---|---|
| **Before** — dispute resolved | before |
| **After** — dispute resolved | after |

Before: "🚨 Urgent: Action required for resolving disputed sale", the 72-hour request, and a live submission button. After: "A sale has been disputed" and no request. The open-window case is unchanged and still gets the urgent ask.

## Test plan

`spec/mailers/contacting_creator_mailer_spec.rb` — two examples covering the post-enqueue transitions, using the real state changes (`update_as_seller_submitted!`, `update_as_resolved!`) rather than raw column writes. `spec/models/dispute_evidence_spec.rb` — a row per condition of the new predicate, plus one tying the instance reader to the class method.

Each condition is independently load-bearing. Removing one at a time and re-running:

| Mutant | Failures |
|---|---|
| drop the `seller_submitted_at` check | 2 |
| drop the `resolved_at` check | 1 |
| drop the hours check | 1 |
| revert the mailer to the old two-condition predicate | 2 (both new mailer examples) |

Local runs, `TEST_DATABASE_NAME` on a private database:

- `spec/mailers/contacting_creator_mailer_spec.rb -e "chargeback notice"` — 9 examples, 1 failure
- `spec/models/dispute_evidence_spec.rb -e "accepting_seller_evidence"` — 6 examples, 0 failures
- `spec/models/dispute_evidence_spec.rb` + `spec/sidekiq/create_missing_dispute_evidence_job_spec.rb` — 67 examples, 1 failure

Both failures are the same pre-existing, order-dependent `merchant_account` factory collision (`Charge processor merchant This account is already connected with another Gumroad account`), which reproduces identically with this diff stashed — a 0-file-changed tree fails the same example. It is local-only; CI is the authority.

## QA steps

No user-facing surface changed other than the email itself, so the rendered emails above are the evidence. To reproduce by hand on a console:

1. Find a dispute with evidence whose seller has been contacted: `de = DisputeEvidence.where.not(seller_contacted_at: nil).where(seller_submitted_at: nil, resolved_at: nil).last`
2. `puts ContactingCreatorMailer.chargeback_notice(de.dispute_id).subject` — the urgent subject, and the body carries the request.
3. `de.update_as_seller_submitted!`
4. Re-run step 2 — now "A sale has been disputed", and `ContactingCreatorMailer.chargeback_notice(de.dispute_id).body.encoded` contains no "Submit additional information".

---

## AI disclosure

Written by claude-opus-5 running as the Hermes agent. It was given Greptile's review comment on #6534 (the mailer predicate diverging from the submission endpoint's eligibility guard), read the mailer, the controller gate, the model's state fields and the two enqueue paths, wrote the fix and the specs, ran them and the per-condition mutation checks locally, rendered the before/after emails from the real mailer, and opened this PR. A local adversarial review was run over the branch before the PR was opened.


gumclaw:
Marked ready for review and routed to you, @nyomanjyotisa.

It had been sitting in draft with nothing actually outstanding, which is the state I'm trying to stop leaving PRs in. Checking the completion gates:

- **Screenshots** — done. Four rendered emails (before/after for both post-enqueue transitions) are committed under `qa-media/` at the current head and embedded in the description. They come off the real mailer, not a mock.
- **Test Results** — in the description, including the per-condition mutation table showing each of the three conditions is independently load-bearing, and an honest note that the one local failure is the pre-existing order-dependent `merchant_account` factory collision that reproduces on a 0-file-changed tree.
- **QA steps and AI disclosure** — both present.
- **CI** — the run on `3802f9316` is still in flight, 0 failures so far. If anything reddens it's mine and I'll take it.

You're the right reviewer on two grounds: you merged #6534, which this follows up (the Greptile finding on that PR arrived after it merged), and you own the dispute-evidence sweep this shares a predicate with.

The one thing worth your judgement rather than just a read: the hours test in `accepting_seller_evidence?` is **deliberately stricter** than the controller's. `check_if_needs_redirect` will let an elapsed window through until `FightDisputesJob` resolves the row, but the notice's own body promises "in the next N hours", so sending the ask with the window gone would be a false promise. If you'd rather the two surfaces agree exactly, say so and I'll relax it to match the controller — it's a copy-versus-consistency call, not a correctness one.

Nothing needed from me. Not arming auto-merge; the merge is yours.

gumclaw:
Closing this as **superseded by #6662**, which merged to `main` at 02:29 UTC — about ten minutes ago.

Both PRs fix the same defect (the chargeback notice shouting "🚨 Urgent: Action required" and linking an evidence form that redirects straight back to the dashboard, because the mailer only checked the window while the enqueue gate also checked submitted/resolved). #6662 is the better-shaped version of the same change, and it is the one you already reviewed and merged:

- It **splits the predicate into two** — `accepting_evidence?` ("may I ask this seller and link the form?", requires an open window) and `notice_worth_sending?` ("is the notice still worth sending at all?", accepts an unstamped row). This branch collapsed both into a single `accepting_seller_evidence?`.
- That split is the substantive improvement. On this branch the enqueue gate reads `ContactingCreatorMailer.chargeback_notice(...).deliver_later if accepting_evidence`, which **suppresses the notice entirely** for a dispute with no evidence surface at all — PayPal and Stripe Connect disputes never get a window, so a seller who was previously told their sale was disputed would now be told nothing. #6662's `notice_worth_sending?` accepts the unstamped row and keeps the plain notice, which is the correct behaviour and a genuine bug in this branch's shape.

So there is nothing here worth rebasing forward. The merge with `main` conflicts in all four source files precisely because #6662 rewrote the same lines.

**What is not lost:** #6662 carries the mailer change, the `Charge::Disputable` gate, and specs in `contacting_creator_mailer_spec.rb`, `dispute_evidence_spec.rb`, and `disputable_spec.rb` — 179 insertions covering the same conditions. The one thing this branch had that #6662 did not is the four rendered-email screenshots (`qa-media/pr-6649-notice-{before,after}-{submitted,resolved}.png`). Correcting what I wrote earlier: they were never on `main` — they only ever existed on this branch. The embeds in the description are pinned to commit `3802f93`, not to the branch name, so they still return 200 with the branch gone (verified after deleting it), and I kept a local copy.

The judgement call I flagged for you at 21:41 UTC — that the hours test was deliberately stricter than the controller's `check_if_needs_redirect` — is resolved the other way by #6662, and correctly: the strictness now lives only in `accepting_evidence?` where the "in the next N hours" promise is actually made, and does not leak into the decision of whether to send anything at all.

Branch `fix/notice-matches-evidence-form-eligibility` is now **deleted**. It was green at `3802f93` with no open PR, which makes it invisible to every watcher that iterates PRs and issues while still reading as unfinished work. Before deleting I checked each part had landed somewhere: predicate, mailer gate and formalization gate are all on `main` via #6662 in a better shape, and its spec coverage is a superset — `dispute_evidence_spec.rb` pins `accepting_evidence?` and `notice_worth_sending?` separately, `contacting_creator_mailer_spec.rb` covers the submitted / resolved / elapsed-window / never-contacted renders, and `disputable_spec.rb` adds the no-evidence-record case this branch's enqueue gate got wrong. Nothing was owed. The commits stay reachable from this PR if they ever need reading back.

PR #6652, opened 2026-07-30
Give guest buyers a review-reminder opt-out and honor can_contact

gumclaw:
## What

Review reminders reach guest buyers with no unsubscribe link at all, and ignore `can_contact`.

Two defects with the same root. `app/views/layouts/email.html.erb` only renders the Unsubscribe link when `@unsub_link` is set, and both review-reminder mailers set it only when a `purchaser` account exists — so a guest checkout (`purchaser_id: nil`) shipped with no opt-out whatsoever. Separately, `Purchase#eligible_for_review_reminder?` never consulted `can_contact`, so a buyer who had already unsubscribed through the receipt footer kept receiving reminders.

This adds `can_contact?` to `eligible_for_review_reminder?` (and the mailer's render-time guard), and falls back to the purchase-scoped unsubscribe link the receipt footer already mints when there is no purchaser account.

No migration: `can_contact` is already a per-purchase column, and `unsubscribe_purchase_url` + `Purchase#unsubscribe_buyer` already write it through a token that works with no session.

## Why

Measured on the read replica, successful purchases in the trailing 30 days (audited request `20260730T195633Z-34f7ee45`):

| | Count | Share |
|---|---|---|
| Successful purchases | 3,455,805 | |
| Guest — no `purchaser` row | 1,952,556 | 56.5% |
| Guests with `can_contact: false` | 325,592 | 16.7% of guests |

So the majority of buyers were in the population with no working opt-out, and ~325k of them had already told us they did not want the seller's email. Not every row triggers a reminder (the job also gates on review-absence, bundle status and product eligibility), so treat these as the ceiling on the affected population rather than send volume.

A second audited read confirmed `can_contact` has **zero NULL rows** in that window (403,735 false / 3,050,751 true / 0 null), so gating on the predicate directly cannot silently drop anyone.

Tracking issue: gumroad-private#1584.

### Scope note

For a guest the only opt-out that exists is the seller-wide one, so a guest unsubscribing from review reminders also stops that seller's other email. Account holders keep the narrow review-reminders-only flag. Giving guests a *narrower* opt-out is what would need new per-purchase storage, and is deliberately left as a follow-up.

## Depends on

Stacked on #6643 — it introduces the `review_reminder_unsubscribe_url` helper this extends, so the base is that branch rather than `main`. Merge #6643 first.

The merge commit here reconciles this change with #6643's recipient-binding fix (`4fcfbd7c0`), which changed the same helper. Both properties are preserved: an account holder gets the sessionless token only when the recipient address matches the purchaser account's address, otherwise the session-gated route; a guest gets the purchase-scoped link. All of #6643's mailer specs pass alongside the new ones.

## Before / After

Guest review reminder (`purchaser_id: nil`), rendered from the real mailer. Before, the footer goes straight from the address line to "Powered by Gumroad" — no opt-out. After, an Unsubscribe link that resolves without a session.

| Before | After |
|---|---|
|  |  |

Captured by rendering `CustomerLowPriorityMailer#purchase_review_reminder` for a guest purchase on this branch and on the base branch, then screenshotting the HTML body. Emitted link on the base branch: none (`href=nil`). On this branch it resolves back to the originating purchase under the `unsubscribe` scope.

No web UI surface changes — the diff touches a mailer, a model predicate and specs.

## Test plan

- `spec/mailers/customer_low_priority_mailer_spec.rb` — 21 examples, 0 failures (includes #6643's recipient-binding spec, so this confirms the merge preserved both fixes)
- `spec/sidekiq/order_review_reminder_job_spec.rb`, `spec/sidekiq/review_reminder_job_spec.rb`, `spec/controllers/users/review_reminders_controller_spec.rb` — 33 examples, 0 failures
- `test/models/purchase_test.rb -n /eligible_for_review_reminder/` — 15 runs, 0 failures

Every new assertion was proven load-bearing by reverting the corresponding production change and re-running:

- Removing `can_contact?` from `eligible_for_review_reminder?` → exactly 2 failures, the two new Minitest cases.
- Reverting the order-level token to `order.purchases.first` → the "mints the token from a purchase the email is about" spec fails.
- Removing `return if reviewable.empty?` → the "does not send when no purchase is still eligible" spec fails (the mail otherwise goes out with no unsubscribe link).

One pre-existing local failure is unrelated: `chargeback_notice_to_customer` hits the known `merchant_account` factory collision (`This account is already connected with another Gumroad account`). It fails on the unmodified base branch too — verified, 2/2 there.

### What a reviewer should check

1. A guest reminder carries an Unsubscribe link, and clicking it needs no login and no email-confirmation wall (`PurchasesController#unsubscribe` resolves a `secure_external_id` token directly).
2. A buyer with `can_contact: false` gets no reminder at all.
3. An account holder still gets the narrow review-reminders opt-out from #6643, with its recipient-address binding intact.

Captured on `https://fix-guest-review-reminder-optout.apps.staging.gumroad.org` at head `a3b583142` (preview served `x-revision: 4bf072b7905a`, the code head this evidence commit sits on top of). Playwright, deviceScaleFactor 2, desktop 1440px and mobile 375px, **no login and no session at any point** — that is the property under test.

The link was not hand-built. A guest purchase (`purchaser_id: nil`, `can_contact: true`) was seeded on the preview, then `CustomerLowPriorityMailer.purchase_review_reminder` was rendered **on the pod** and the `href` scraped out of its body:

```
MARK purchase id=16 purchaser=nil state=successful can_contact=true
MARK eligible_before=true
MARK mailer_unsub_href="https://fix-guest-review-reminder-optout.apps.staging.gumroad.org/purchases/<token>/unsubscribe"
```

On `main` that `href` is `nil` — `@unsub_link` was only set `if @purchase.purchaser`. Visiting the emitted URL anonymously:

| Guest opt-out — desktop | Guest opt-out — mobile (375px) |
|---|---|
|  |  |

HTTP 200, no redirect to `/login`, confirmation reads "You will no longer receive emails from this seller." — the seller-wide scope described in the [Scope note](#scope-note), not a review-reminders-only opt-out.

**The click was confirmed to land server-side**, not just to render a confirmation. Re-reading the same purchase on the pod afterwards:

```
MARK after_click can_contact=false eligible=false
MARK mailer_after_optout to=nil subject=nil body_len=0
```

So the guest's only reachable opt-out both persists and is honoured: `eligible_for_review_reminder?` flips to false via the new `can_contact?` clause, and re-rendering the mailer for that purchase now returns an empty mail. That is the end-to-end loop this PR adds.

**Control leg** — the session-gated account-holder route `/users/unsubscribe_review_reminders` on the same host, same anonymous context, redirects to `/login?next=%2Fusers%2Funsubscribe_review_reminders`. Without it a reviewer cannot tell the guest link is genuinely sessionless rather than the whole route being open:

| Control (session-gated) — desktop | Control — mobile |
|---|---|
|  |  |

### QA steps a reviewer can repeat

1. Open `https://fix-guest-review-reminder-optout.apps.staging.gumroad.org` — no login needed for steps 1–3.
2. On the preview write console, render the reminder for a guest purchase (`purchaser_id: nil`) and copy the `href` out of the mail body. Purchase 16 (`qa6652guest@example.com`) is already seeded for this.
3. Visit that URL in a fresh/incognito context → the confirmation page above, with no login wall.
4. Re-read the purchase: `can_contact` is now `false` and `eligible_for_review_reminder?` is `false`; rendering the mailer again yields an empty mail.
5. Visit `/users/unsubscribe_review_reminders` anonymously → bounced to `/login`, confirming the account-holder route is unchanged.

**Durable preview state the next session inherits:** purchase 16 on `seller@gumroad.com` / product `demo` (guest, `can_contact: false` after this run), and an alive $5.00 price row was created on product 1 if it had none. No production data touched.

## Notes for follow-up

Two items found during adversarial review, deliberately not bundled here:

- `get :unsubscribe` mutates state on GET, the mirror of the prefetch bug `subscribe` was hardened against (it requires POST). Pre-existing and already reachable from every receipt, so this change does not introduce it — worth a confirm-on-GET / act-on-POST fix plus RFC 8058 one-click support.
- Four pre-existing "does not send an email" examples in the mailer spec assert `deliveries.count` without calling `.deliver_now`, so they pass regardless of mailer behavior.

---

## AI disclosure

Written by Hermes Agent running `claude-opus-5`.

Instructions given: fix gumroad-private#1584 — the issue asked which of three options to implement (gate on `can_contact`, give guests their own opt-out, or both) and the direction received was "Fix."

The agent measured the affected population and the `can_contact` NULL distribution via the audited read-only production console, implemented both halves after establishing that no migration was required, wrote the specs and proved each one fails without its corresponding production change, ran two rounds of local adversarial review, and reconciled the merge conflict with #6643's concurrent recipient-binding fix. Round 1 of that review found a real defect — the order-level unsubscribe token was minted from `order.purchases.first`, which can be a row excluded from the reminder, pointing a guest's only unsubscribe link at a seller who was not among the senders (and making the click a guaranteed no-op when the row was excluded for `can_contact`). That is fixed here. Round 2 reviewed the fix commit and returned no findings.


gumclaw:
Ready for review. CI is green and complete on the current head `4bf072b` — 86 check runs (77 success, 9 skipped, 0 failures), plus `ci/green` and `buildkite/gumroad` both success — and the branch merges clean.

**What this does.** Guest buyers — 56.5% of recent successful purchases — were getting review reminders with no unsubscribe link at all, because the email layout only renders one when `@unsub_link` is set and both reminder mailers set it only for account holders. Separately `eligible_for_review_reminder?` ignored `can_contact`, so ~325k guests who had already unsubscribed via the receipt footer kept getting reminders. Both are fixed, with no migration: `can_contact` is already per-purchase and the receipt footer already mints a sessionless purchase-scoped token.

**Merge reconciliation.** #6643 changed `review_reminder_unsubscribe_url` in `4fcfbd7c0` (recipient binding) while this was in flight, so the two conflicted. I merged the parent in and kept both properties rather than resolving the text one way: an account holder gets the sessionless token only when the recipient address matches the purchaser account's address, otherwise the session-gated route; a guest gets the purchase-scoped link. All of #6643's mailer specs pass alongside the new ones — 21 examples, 0 failures.

**Verification.** Every new assertion was proven load-bearing by reverting its production change and re-running: dropping `can_contact?` from the gate fails exactly the two new Minitest cases; reverting the order-level token to `purchases.first` fails the token spec; removing the empty-guard fails the no-send spec. Rendered before/after of the actual email is in the description. Greptile independently ran a contract probe against both revisions: 5 of 7 contracts failed pre-change, all 7 pass here.

**@gianfrancopiana — the ask is a review plus a preview click-through.** Confirm a guest reminder's Unsubscribe link works from a signed-out browser with no email-confirmation wall, and that an account holder still gets the narrow review-reminders opt-out. Not armed for auto-merge: this changes who receives email and the unsubscribe semantics, so the merge is yours. #6643 should merge first — this is stacked on it.

One judgment call worth your attention: for a guest the only opt-out that exists is seller-wide, so a guest unsubscribing from review reminders also stops that seller's other email. A guest-specific narrow opt-out is the part that would need new storage, and I left it as a follow-up rather than deciding it here.


---

**Head moved once for evidence naming, no code change.** `4bf072b` is a pure `R100` rename of all three captures to the CONTRIBUTING `pr-6652-` prefix (they shipped with a descriptive slug and no PR number, so they would have been unattributable in `main`). Body re-pinned to the new SHA; bytes unchanged, all three URLs 200. CI re-runs over identical code.

PR #6653, opened 2026-07-30
Pin presentment refunds as Connect charge-model agnostic

gumclaw:
## What

Adds two examples to `spec/services/purchase/presentment_refund_spec.rb` pinning that
`Purchase::PresentmentRefund` derives identical refunds regardless of the seller's Stripe Connect
charge model. Spec-only — no app code touched, so there is no runtime behaviour to ramp and nothing
to flag-gate.

Closes the last open reviewer question on antiwork/gumroad-private#1318.

## Why

That issue asked whether `Purchase::PresentmentRefund` needs an explicit destination-shape spec, or
whether "it derives everything from the persisted `purchase_presentment` snapshot and never re-reads
the charge model" is sufficient reasoning.

The reasoning is correct — the service contains zero reads of `merchant_account`,
`charge_processor`, `stripe_account` or `destination`, and transitively touches only
`purchase_presentment` columns, `refunds.effective`, `total_transaction_cents`,
`gross_amount_refundable_cents` and the pure-arithmetic `Charge.allocate_by_largest_remainder`. But
it was **unpinned**: nothing in the suite would have noticed a future change reaching for
`purchase.merchant_account` here, and buyer-currency refunds are the wrong place to rely on a
reviewer remembering an invariant.

The two examples:

- **`derives identical refunds for direct, destination and platform charges`** — runs
  `.from_presentment_amount` (full and partial) and `#tax_only_result` against the same purchase
  under four genuinely distinct merchant-account shapes: a Stripe Connect direct-charge account, a
  Gumroad-managed destination-charge account, the real platform row (`user_id` nil), and no
  merchant account at all. Asserts the snapshots are equal **by value** — full `json_data` hashes
  plus canonical cents — not merely "present".
- **`consumes presentment balance from prior refunds regardless of the charge model`** — a prior 54c
  presentment refund leaves exactly 81 of 135 refundable, and both the 82/81 boundary and the
  derived amount land in the same place on every shape.

The shapes are asserted distinct in the dimension the service could read
(`is_a_stripe_connect_account?` true/false/false, `is_managed_by_gumroad?` true only for the
platform row) so the comparison cannot degrade into comparing a thing against itself.

## Before / after

| | Before | After |
|---|---|---|
| A change reading `merchant_account` in `PresentmentRefund` | ships green | reddens 2 examples |
| Charge-model coverage in this file | none — every example used one fixed shape | 4 shapes, compared by value |
| The #1318 reviewer question | open, answered only by prose | pinned in the suite |

Not applicable: no user-facing surface changes, so there is no UI before/after to show. The
evidence below is the spec run and the mutation proof.

## Test results

Whole file green on the branch, and the two new examples proven load-bearing rather than merely
present. Injecting a single charge-model read into the service —

```ruby
# app/services/purchase/presentment_refund.rb, in from_presentment_amount
return nil if purchase.merchant_account&.is_a_stripe_connect_account?
```

— reddens **2 of 2** with `NoMethodError: undefined method 'canonical_gross_refund_cents' for nil`;
restoring the file byte-identical passes 2 of 2. Without these examples that one-line change would
have shipped green.

Re-run at the current head `8e039a80` after the review hardening below:

| Run | Result |
|---|---|
| `rspec spec/services/purchase/presentment_refund_spec.rb` | **22 examples, 0 failures** |
| same file with the mutant injected, `-e 'charge-model independence'` | 2 examples, **2 failures** |
| mutant reverted, working tree restored byte-identical | 2 examples, 0 failures |

Spec run, pre-existing-failure baseline, and mutation proof

**On the one failure visible in that render.** It was captured at the first head, where the file
read 22 examples / 1 failure — `frees the presentment amount when the reversal runs through the
real failure service` (`:313`), `ActiveRecord::RecordInvalid: Validation failed: Merchant account
must exist`. That example reproduces identically on unmodified `origin/main` with this branch's
added lines reverted (20 examples / 1 failure — the 20→22 delta is exactly the two examples added
here), so it is not caused by this PR. It did **not** reproduce on the re-run above, which makes it
intermittent rather than a standing failure; the same `Merchant account must exist` shape is filed
as #6657.

## Review

Local adversarial pre-merge review at `70c24ca6`: **READY-TO-MERGE**, three P3s, no P1 or P2.

It confirmed the premise from source rather than taking the PR's word for it — the service has zero
charge-model reads on any path, the four fixture arms are genuinely distinct, the assertions compare
by value against non-degenerate anchors (100/135, 40, 20, 60), no early return short-circuits the
fixture before the question is reached, and the 82/81/60 arithmetic re-derives exactly
(135 − 54 = 81 remaining; at 81 the equality branch returns
`gross_amount_refundable_cents` = (100 − 40) + (0 − 0) = 60).

All three P3s taken in `8e039a80`:

| Finding | Disposition |
|---|---|
| P3 `use_charge_model` never asserted the assignment survived the save, so a future normalizing callback could collapse the shapes and defuse both examples silently | fixed — asserts `purchase.merchant_account_id` after reload |
| P3 the prior-refund example dropped the nil-account arm the first example has, so a `merchant_account.nil?` guard firing only once refunds exist would escape both | fixed — nil arm added |
| P3 helper comment said "all four shapes", true of example 1 only | fixed by the above |

Mutation re-proved after the hardening, not assumed: still 2 of 2 red, still green restored.

## QA steps

No user-facing surface, so there is nothing to click on the preview app. Reproducible locally:

1. `bundle exec rspec spec/services/purchase/presentment_refund_spec.rb` — expect 22 examples,
   0 failures.
2. Confirm the guard bites: add `return nil if purchase.merchant_account&.is_a_stripe_connect_account?`
   as the first line of `Purchase::PresentmentRefund.from_presentment_amount` and re-run with
   `-e 'charge-model independence'` — expect 2 failures. Revert.

## Merge posture

Not auto-merged. The diff is spec-only and cannot change runtime behaviour, but it pins refund
derivation on the buyer-currency path, and merges on that surface stay with a human — @rmarescu owns
Connect charge routing. Nothing here needs a decision, only the merge click.

---

Written by Hermes Agent (claude-opus-5). Instructions: answer the open reviewer question on
antiwork/gumroad-private#1318 about whether presentment refunds need an explicit destination-charge
spec, by pinning the charge-model independence in the suite rather than in prose; then run an
adversarial pre-merge review and fix its findings.


gumclaw:
Fixing the routing on this one, and saying why so the change isn't silent.

**Reviewer moved to @rmarescu, @steveo-es removed.** This pins refund behaviour across Connect charge models — Stripe charge-model internals and multi-currency presentment, which is Raz's domain by standing assignment. Steve is CFO and reviews finance/tax questions, not application code; that was a mis-route, not a judgement about the change.

**Removed gumclaw as co-assignee.** Every PR carries exactly one human owner; the bot is never an assignee.

**`awaiting-human` → `working`.** It was labelled for hand-off with 68 checks still in progress and no human having looked yet. Handing over a PR whose suite hasn't settled asks the reviewer to re-derive whether it's even green. It flips back the moment CI settles clean.

@rmarescu — nothing needed from you yet. When this is genuinely ready the ask is small: it's spec-only, and the question is whether pinning these invariants matches how you expect destination vs. direct charges to behave on refund.

gumclaw:

```
```

gumclaw:
**Label flipped `awaiting-human` → `working`. @rmarescu stays assigned and review-requested; nothing needed from you this minute.**

Why: `awaiting-human` implied you were the blocker, but CI hasn't settled — `Build images` and both `Test Minitest` shards are still pending on run 30599080674 after I re-ran the wall-clock `Lint JS/TS` failure that had nothing to do with this diff.

What's expected of you once it's green: review is a small one — the diff is a single spec file, `spec/services/purchase/presentment_refund_spec.rb`, pinning that presentment refund amounts come out identical across direct, destination, platform-managed and absent Stripe merchant accounts, plus remaining-refundable after a prior presentment refund. The question for you is whether those four charge-model configurations are the right set to freeze. I'll flip the label back and ping when checks are green; no action before then.

PR #6655, opened 2026-07-30
Stop returning an unusable editor_revision on the stale-deletion 409

gumclaw:
Settles the third of the three options in antiwork/gumroad-private#1532, the only one that did not need a design decision first.

## What was wrong

The save contract answers a stale-token deletion with 409 + a fresh `editor_revision`, and the comment on that response said the token rode along "so the editor can reconcile and retry after showing the seller what changed, rather than forcing a full page reload."

That affordance was never built, and antiwork/gumroad#6559 established it **cannot** be built by adopting the token. Deletions are gated on the token; ordinary writes are not, and the editor save is a full snapshot. So swapping the token and resending applies the deletion *and* reverts every field another session changed in between. That is proved end to end through the real endpoint in `LinksControllerSaveContractTest`.

Since #6559 the client has thrown the token away — `saveProductError` maps the 409 to `StaleDeletionConflictError` and deliberately does not expose it. So the field is a documented invitation to do the one unsafe thing, consumed by nothing. `git grep editor_revision -- app/javascript` confirms the only reader is the *success* path in `ProductEditPage`.

## Why removing it costs the remaining work nothing

Both routes still open in #1532 obtain a current token without this one:

- **a deletion-only write** — a fresh request built from a read of current state;
- **re-fetch and reconcile** — a read of current state by definition.

`ProductPresenter` issues `Product::EditorRevision.current` on every product-edit load, so the reload that both routes start from already hands out a live token. The 409's copy is redundant to both.

## What changed

- `LinksController` no longer emits `editor_revision` on `stale_deletion_conflict`, with the reasoning at the raise site instead of the old comment promising a retry that cannot exist.
- `SaveProductErrorPayload` drops the field, so there is no longer a typed slot for it.
- The two tests asserting the token's presence now assert its absence.
- The end-to-end "resending a stale snapshot deletes as asked AND reverts the other session's edit" proof is **kept**, obtaining the token directly rather than from the 409. That was the point all along: the danger is the stale snapshot, not where the token came from. Keeping it means the trap stays documented for whoever builds the real retry.
- One client test now asserts defence in depth — a server that regressed and sent the token still could not get it onto the error.

## Verified by running

- `LinksControllerSaveContractTest`: **59 runs, 241 assertions, 0 failures**.
- `product_edit.test.ts`: **18 passed**.
- rubocop clean on both Ruby files; eslint clean; `tsc --noEmit` reports nothing for either changed TS file.

One note from actually running it rather than assuming: my first draft of the client test asserted `typia.assert` would *reject* the extra key. It does not — typia permits excess properties — so that test passed vacuously and I rewrote it to assert the reachable property instead.

No behaviour change for sellers: the reload path from #6559 is untouched, and the only thing removed is a field nothing read.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -623,12 +623,21 @@ class LinksController < ApplicationController
       # must NOT clear its pending-removal state on this response — the whole
       # point is that the seller's deletion has not happened yet.
       #
-      # A fresh token rides along so the editor can reconcile and retry after
-      # showing the seller what changed, rather than forcing a full page reload.
+      # Deliberately NO fresh `editor_revision` here, unlike the success
+      # response. Adopting a token on this path can only authorise the session's
+      # NEXT save — which is the same stale full snapshot — so the deletion
+      # would land while every field a co-editor changed in between was reverted
+      # to this session's values (proved in LinksControllerSaveContractTest,
+      # "resending a stale snapshot with the 409's fresh token deletes as asked
+      # AND reverts the other session's edit"). It was emitted here for a
+      # reconcile-and-retry affordance that does not exist and cannot be built
+      # client-side; the client threw it away (see saveProductError). Both
+      # candidate retries in gumroad-private#1532 — a deletion-only write, or
+      # re-fetch-then-reconcile — get a current token from the reload they
+      # already have to do, so neither needs one on the refusal.
       return render json: {
         error_message: e.message,
         error_code: "stale_deletion_conflict",
-        editor_revision: Product::EditorRevision.current(@product.reload),
       }, status: :conflict
     rescue Product::RichContentDeletionGuard::HiddenVariantContentConflict => e
       # The fail-closed inconsistent-content case: hidden version-level pages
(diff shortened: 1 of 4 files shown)


gumclaw:
Two housekeeping fixes on this PR, flagged rather than done silently.

**1. Scrubbed seven stray `declare/` files out of the branch.** The opening commit had swept up local scratch artifacts alongside the real change — four one-off MaxMind GeoIP probe scripts from unrelated sanctions-screening work, plus a 1Password local daemon pid file and one of its encrypted keyset files. None of them belong in this repository. I amended the commit rather than stacking a delete on top, so the history stays clean: the branch is now exactly the four files it should be (`links_controller.rb`, `product_edit.ts`, `product_edit.test.ts`, `links_controller_test.rb`). Head moved 79ca75d1e → 93b7b7ea2; no review had started, so nothing was invalidated.

**2. Re-requested your review.** The PR opened with you as assignee but an empty reviewer list, so it was not actually surfacing in your review queue. Fixed.

Nothing about the change itself moved — the diff is byte-identical to what was there before minus the scratch files.

nyomanjyotisa (approved):

PR #6658, opened 2026-07-31
Pin the clock in the Apple Pay end-date specs

gumclaw:
Every open pull request in this repository is red on `Lint JS/TS` right now, and none of them broke anything. Two examples in `applePayRecurringPaymentRequest.test.ts` computed their expected end date on the real clock:

```ts
const expectedEnd = new Date();
expectedEnd.setMonth(expectedEnd.getMonth() + 11);
```

A raw `setMonth()` overflows when the target month is shorter than the starting day-of-month. Production deliberately does *not* do that — `addMonthsClamped` clamps to the target month's last day, because billing anniversaries clamp. So the two disagree on exactly the days when overflow happens.

Today is **July 31**. July 31 + 11 months = June 31, which `setMonth()` rolls into July, so the spec expected month 6 and got 5. July 31 + 9 months = April 31, rolled into May, so it expected 4 and got 3. Both assertions that failed in CI are precisely these:

```
AssertionError: expected 5 to be 6   applePayRecurringPaymentRequest.test.ts:84
AssertionError: expected 3 to be 4   applePayRecurringPaymentRequest.test.ts:97
```

The production clamp is correct and is **not** touched here. Only the two examples move onto a pinned clock (March 15, 2026), matching what the rest of the describe block already does for the two overflow-specific examples further down, and they now assert the full date rather than just month and year — the day-of-month is the part that was silently unchecked.

## Test Results

`npx vitest run app/javascript/components/Checkout/applePayRecurringPaymentRequest.test.ts`

```
 ✓ app/javascript/components/Checkout/applePayRecurringPaymentRequest.test.ts (24 tests) 23ms

 Test Files  1 passed (1)
      Tests  24 passed (24)
```

24/24 green, including the two Jan-31 and Oct-31 clamping examples that prove the production behaviour this change leaves alone is still exercised.

Mutation check: the failure is self-proving in the other direction — the *unfixed* spec is red in CI on every open PR today (e.g. run 30592277423 on #6558) and was green yesterday with no code change between, which is only explicable by the date dependency.

## QA media

Not applicable: no user-visible surface changes. This is a spec-only change to two assertions; the CI logs linked above are the evidence.

## AI disclosure

Written by an AI agent (gumclaw). The diagnosis, the fix and the local verification above are mine. Worth a human eye on one judgement call: a partially-written change on this branch had also *deleted* the `addMonthsClamped` clamping logic from the production module, which would have made the specs pass by reintroducing the January-31 overflow bug. I discarded that and fixed only the expectations, on the grounds that the clamp is the intended behaviour and the specs were the thing that was wrong.

### DIFF
--- a/app/javascript/components/Checkout/applePayRecurringPaymentRequest.test.ts
+++ b/app/javascript/components/Checkout/applePayRecurringPaymentRequest.test.ts
@@ -32,6 +32,19 @@ const membership = (overrides: Partial<Product> = {}) =>
   product({ nativeType: "membership", recurrence: "monthly", ...overrides });
 
 describe("getApplePayRecurringPaymentRequest", () => {
+  // End dates are derived from the current date, so every example that asserts one runs against a
+  // pinned clock. Left on the real clock they only pass on days whose target month is long enough
+  // to hold the same day-of-month, so they go red on the 31st of a month 11 or 9 months ahead of a
+  // 30-day one. Examples below that pin their own date override this.
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date(2026, 2, 15)); // March 15, 2026
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
   it("returns null for a one-time cart", () => {
     expect(getApplePayRecurringPaymentRequest([product()], MANAGEMENT_URL)).toBeNull();
   });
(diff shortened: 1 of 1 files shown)


gumclaw:
Independently verified this against `b2571a5` while triaging the same red Lint JS/TS job on #6558. The diagnosis holds and the fix is the right shape.

Root cause, confirmed on clean `origin/main` (`f4598f3`): the two examples recomputed the expected end date with a bare `setMonth()`, which is precisely the overflow `addMonthsClamped` exists to prevent. July 31 + 11 months naive → July 1 (month 6); the clamped source correctly yields June 30 (month 5). Hence `expected 5 to be 6`. It only fires on the 31st of a month whose target month has 30 days, so it reddens every open PR in the repo for the day.

Reproduction and verification:

- Clean `origin/main`, `TZ=UTC`: 2 failed / 22 passed — identical to CI. Same worktree on local PT (still July 30) passed, which is why this looked machine-specific.
- With this diff, `TZ` matrix UTC / America/Los_Angeles / Asia/Kolkata / Pacific/Kiritimati: 24 passed in all four.
- Mutation check: stripping the clamping out of `addMonthsClamped` reddens `clamps a fixed-duration membership end date...` and `clamps an installment plan end date...`. Coverage is preserved, not weakened — the expectations moved onto a pinned date and the assertions got tighter (exact year/month/day rather than a recomputed month index).
- `prettier` clean, `eslint` clean, `tsc --noEmit` reports nothing for this file.

One note for the record: the file already pinned the clock for its `month-boundary end dates` block, so the new suite-level `beforeEach` makes the whole file consistent with the convention that block established.

xrav3nz (approved):

gumclaw:
**`working` on, `awaiting-human` off.** @xrav3nz stays as assignee — you own the outcome on the Apple Pay date specs — but the label was claiming you were blocking, and you are not.

**What I owe here:** CI is still running (Test Fast shards pending). Nothing is asked of you until it settles. If it comes back green I will arm auto-merge rather than bounce this back for a courtesy review — a green clock-pinning spec fix does not need a second opinion, and `awaiting-human` should mean a named person has a real task.

For context on why this matters beyond its own diff: the two examples this pins were computing their expected end date with a bare `setMonth()` on the real clock, which overflows exactly the way `addMonthsClamped` exists to prevent (July 31 + 11 months naive → July 1). That is currently reddening `Lint JS/TS` on #6558, #6526 and #6495, none of which touched it. So this is the unblocker for three other PRs, which is why I am holding it rather than parking it.

If CI comes back red on something that *is* this branch, I will fix it and say so here.

PR #6659, opened 2026-07-31
Port the GDPR buyer erasure service spec to Minitest and fixtures

gumclaw:
## What

Ports `spec/services/gdpr_buyer_erasure_service_spec.rb` to the Minitest + YAML fixtures suite as `test/services/gdpr_buyer_erasure_service_test.rb`, deletes the RSpec original, and adds the `create_order`/`create_charge` builders to `test/support/model_factories.rb`.

Item 18 of the top-20 factory-heavy files on #5801. No production code changes.

## Why

This file was one of the heaviest FactoryBot consumers left in the suite. The port is deliberately mechanical — 37 original examples map 1:1 onto 37 Minitest tests — so the diff is reviewable as a coverage-preservation question rather than a rewrite.

## Test Results

```
37 runs, 122 assertions, 0 failures, 0 skips
```

Evidence in `qa-media/pr-6659-port-gdpr-erasure-test-results.png`.

## Fable-5 adversarial review: READY-TO-MERGE

Run against the branch before opening this PR, per policy. Because a test port is a no-op on production code, its entire risk is coverage silently shrinking — and CI cannot see that, since both suites are green by construction. The review was therefore scoped to a mechanical original→port mapping rather than style.

| 1 | P3 | `custom_fields.blank?` assertion was vacuous — `Purchase#custom_fields` maps the `purchase_custom_fields` association, not the column the service nulls. No `PurchaseCustomField` rows exist in the fixture, so it was blank before and after. Inherited verbatim from the RSpec original, so parity held. | **Fixed** in `aed7b84` |
| 2 | P3 | `create_order`/`create_charge` are byte-identical to the same hunk on sibling branch `test/port-purchase-export-service` — textual conflict for whichever merges second, zero semantic conflict | Accepted; whoever rebases second drops the duplicate hunk |
| 3 | P3 | Evidence PNG does not follow `pr-<number>-<description>` naming (no PR number existed yet) | Accepted, cosmetic |
| 4 | P3 | Two provenance comments restate history rather than a constraint | Accepted, kept the `#5801` pointer |

No P1s, no P2s.

**What the review confirmed safe** — these passed adversarial claims and matter as much as the findings:

- **Coverage mapping complete**: 37 original examples → 37 tests, every row `SAME`. No `MISSING`, no `WEAKER`.
- **Erasure pre-state audit**: for an erasure service, "field is now nil" passes trivially if the fixture never populated it. Every erased attribute was checked for a non-blank pre-state; all were `PRE-STATE-SET` except finding 1, now fixed.
- **Blast radius preserved**: the must-NOT-erase assertions (other buyers', other purchases', the seller's data) all survived the port. A port that keeps the positives and drops the negatives is the dangerous outcome; that did not happen.
- **No trait-fidelity regression** of the kind that hit the sibling branch, where a giftee purchase kept a default `purchase_state` that put it in the service's own scope.
- **No fixture leakage**: new rows do not perturb other tests that count rows or take `.first`/`.last`.
- **No ordering premises** left unasserted.

Finding 1's fix was verified by mutation rather than assertion: commenting out `custom_fields: nil` in the service (line 133) now reddens with `Expected "{"phone": "555-1234"}" to be nil`. The old assertion survived that mutant.

## QA steps

Test-only change; nothing to click through.

1. `bin/rails test test/services/gdpr_buyer_erasure_service_test.rb` — expect 37 runs, 0 failures.
2. Confirm `spec/services/gdpr_buyer_erasure_service_spec.rb` is gone and no other spec references it.
3. Sanity-check the coverage table above against `git show origin/main:spec/services/gdpr_buyer_erasure_service_spec.rb` if you want to re-derive the 37→37 mapping independently.

## AI disclosure

Written by gumclaw (AI agent). The fable-5 adversarial review was run by a separate headless model instance against the branch before this PR was opened; its verdict and findings are reproduced above, and finding 1's fix was verified empirically by mutation.


gumclaw:

```
```

gumclaw:
Renamed the evidence file to the CONTRIBUTING convention — `qa-media/port-gdpr-erasure-test-results.png` → `qa-media/pr-6659-port-gdpr-erasure-test-results.png`. It predated the PR number, which is why it shipped without one.

Pure rename (`R100`, 0 insertions, 0 deletions) plus the one-line body reference; no code changed, so the re-running CI covers a byte-identical tree. The PR stays yours.

PR #6662, opened 2026-07-31
Stop asking for chargeback evidence the submission page will not accept

gumclaw:
## What

Make the chargeback notice's ask for evidence agree with what the submission page will actually accept, and split the two questions the callers were really asking:

- `DisputeEvidence.accepting_evidence?` — "may I ask this seller and link the form?" Requires an **open window**.
- `DisputeEvidence.notice_worth_sending?` — "is the notice still worth sending at all?" Accepts an **unstamped** row.

## Why

The notice is rendered when the mail queue reaches it, not when it is enqueued, so the seller may have already submitted evidence or the dispute been resolved in between. The mailer only checked the window, while the enqueue gate in `Charge::Disputable` also checked submitted/resolved — so the two disagreed and the seller got a "🚨 Urgent: Action required" email whose button redirects straight back to the dashboard. The subject line had the same split: it shouted "Urgent" whenever a window had ever been opened, including when the body had already dropped the ask and the email contained nothing to act on.

The two predicates exist because the callers genuinely differ. A PayPal or Stripe Connect dispute has **no evidence surface at all**, never gets a window, and the plain notice is the only one it will ever receive — so an unstamped row must still send a notice while never being asked for evidence.

## Test Results

CI was green on the branch at `982121137` (76 checks) before the review fixes; the current head is running now.

⚠️ **The local suite is unusable on this machine** — `spec/models/dispute_evidence_spec.rb` fails in full, including examples this branch never touches, and the mailer spec fails across unrelated mailers. So I verified the core claim a different way, DB-free, since both predicates are pure class methods over three timestamps:

```
window  | submitted | resolved || ask? | notice? | divergence
absent  | false     | false    || false | true    |  <-- DIFFERS
absent  | false     | true     || false | false   |
absent  | true      | false    || false | false   |
absent  | true      | true     || false | false   |
open    | false     | false    || true  | true    |
open    | false     | true     || false | false   |
open    | true      | false    || false | false   |
open    | true      | true     || false | false   |
expired | * (all four rows)    || false | false   |

Rows where the two predicates differ: [["absent", false, false]]   CONFIRMED
ask?-without-notice? rows: []  (none — ask is strictly narrower, correct)
```

That is the whole safety argument in one table: the predicates differ on **exactly one** row — the unstamped, unsubmitted, unresolved row that PayPal/Connect disputes live on — and there is no row where we would ask for evidence inside a notice we decided not to send.

## Fable-5 adversarial review: CHANGES-REQUIRED → fixed

Run against the branch before opening this PR, per policy. Because `982121137` is the commit that *closes* round 1's findings, it was reviewed as its own delta first, per the round-chaining rule — a fix commit earns more suspicion than original code.

**No behavioral defect found.** The review proved both call sites are term-for-term equivalent to `main`'s inline expression, so the tip commit is a pure refactor at existing call sites; what it genuinely hardens is any *future* bare caller of `accepting_evidence?`, which in round 1 got `true` for an unstamped row.

| 1 | **P2** | The local in `Charge::Disputable` held `notice_worth_sending?` but was still named `accepting_evidence` — recreating the exact confusion this branch exists to remove | **Fixed** in `83f0c22` |
| 2 | P3 | Mailer comment claimed "read once, so the sentence cannot quote a different window than the gate just allowed" — false, since `accepting_evidence?` reads its own clock, so the two calls can straddle the window's end and quote "in the next 0 hours" | **Fixed** — `hours_left` read first, gate ANDed with it |
| 3 | P3 | Spec comment described an impossible flow ("fails to enqueue" yet "still queued") | **Fixed** — names the real push-succeeded-then-stamp-failed shape |
| 4 | P3 | New example in a `:vcr` describe has no committed cassette | Checked: no HTTP occurs on that path (the dispute-evidence creation is stubbed to nil) |

Finding 1 is the one worth reading twice. The name is not cosmetic: round 1's own history shows a maintainer plausibly "unifying" the gate onto the strict predicate, which would silence the plain notice for every PayPal and Connect dispute — the only notice they ever get. The new spec at `disputable_spec.rb:859` is the thing standing in the way, and it does redden under that change.

**Review also confirmed:** the permissive unstamped arm was probed for a race where a *stampable* row is briefly unstamped (which would send the no-ask notice for a dispute about to become fightable) — the stamping writer does not leave such a window on this path. No third string (preheader, translation key, mirror) still says "Action required" independently. No mailer signature changed, so no `have_enqueued_mail(...).with(...)` arity sweep is needed.

## QA steps

Copy changes for real states, so worth rendering rather than reasoning about:

1. A dispute with an **open** window, nothing submitted: subject is "🚨 Urgent: Action required…", body asks for information with a working "Submit additional information" button and a positive hour count.
2. Same dispute after the seller **submits**: subject falls back to "A sale has been disputed", body has no ask and no button.
3. A dispute whose window has **expired**: same as (2) — this is the case that used to shout "Urgent" with a button that bounces to the dashboard.
4. **PayPal dispute** (no evidence surface, never stamped): still receives the plain "A PayPal sale has been disputed" notice. This is the regression to watch — if it stops arriving, the split predicate has been collapsed.
5. A dispute **resolved** before the queued notice is delivered: no ask.

Assigning @nyomanjyotisa, who already owns the neighbouring chargeback-notice work in gumroad-private#1573 and gumroad#6628, and applying `awaiting-human`: this changes what sellers are told about a deadline that costs them real money if missed, and I cannot run the mailer specs locally to render (1)–(5) myself.

## AI disclosure

Written by gumclaw (AI agent). The fable-5 adversarial review was run by a separate headless model instance against the branch before this PR was opened; the truth table above was produced by an independent DB-free probe of the two predicates rather than taken from the review's reasoning.


gumclaw:

```
```

